PR #6180, opened 2019-09-22
Model system rework

gigaherz:
This is the initial implementation of my new model system idea.

This feature expands the vanilla model loading with a new key: "loader". If this key is found in the model, forge will derive a model instance from any necessary values seen in the json.

Example, `models/item/test_fluid_bucket.json` 
```json
{
  "parent": "forge:item/bucket",
  "loader": "forge:bucket",
  "fluid": "new_fluid_test:test_fluid"
}
```

The "parent" key is handled by the vanilla logic, which in this case takes care of providing the required default textures for the bucket. The "forge:bucket" loader constructs a custom model that uses a base texture and some mask textures, to produce a custom bucket model for the given fluid in the "fluid" key.

### I need people to test this! Here's how:

1. Clone/checkout the forge repository `git clone https://github.com/MinecraftForge/MinecraftForge.git`
2. Fetch and checkout my branch `git fetch origin pull/6180/head:model-system-rework` + `git checkout model-system-rework`
3. Run `gradlew setup`
4. Run `gradlew publishToMavenLocal`
5. Go to `<forge folder>/projects/forge/build/libs/` and write down the exact version number eg. `forge-1.14.4-28.1.90-model-system-rework.jar`
6. Edit your project's build.gradle, and in your repositories (NOT the repositories block from `buildscript`!!), add `mavenLocal()` as a repository, eg.
```gradle
repositories {
    mavenLocal()
}
```
7. Edit your minecraft dependency to match the version as seen in the build results, eg.
```gradle
minecraft 'net.minecraftforge:forge:1.14.4-28.1.90-model-system-rework'
```
8. Refresh/reimport your gradle project into your IDE, and wait for the setup to complete.
9. Rerun `gradlew gen<YOURIDE>Runs`
10. Implement your models under the new system and tell me of any issue you encounter or any feature you think would be important to add.

PR #6534, opened 2020-02-26
[1.14.x] Make MultiModel pass its extra model data down to its parts

TBoshoven:
This was implemented in several other model loaders, but never this one, even though it is still in use in 1.14.

My use case is a composite model with a dynamic model part that uses model data.

[Comparable change in `MultiLayerModel`](https://github.com/MinecraftForge/MinecraftForge/commit/fc189c9aafd9b9db7ebd19d348b58b7d8eaaaca5#diff-67d7802b8bffc62b50cd0d1808413ed1L150-L165)

### DIFF
--- a/src/main/java/net/minecraftforge/client/model/MultiModel.java
+++ b/src/main/java/net/minecraftforge/client/model/MultiModel.java
@@ -196,11 +196,11 @@ public final class MultiModel implements IUnbakedModel
             ImmutableList.Builder<BakedQuad> quads = ImmutableList.builder();
             if (base != null)
             {
-                quads.addAll(base.getQuads(state, side, rand));
+                quads.addAll(base.getQuads(state, side, rand, extraData));
             }
             for (IBakedModel bakedPart : parts.values())
             {
-                quads.addAll(bakedPart.getQuads(state, side, rand));
+                quads.addAll(bakedPart.getQuads(state, side, rand, extraData));
             }
             return quads.build();
         }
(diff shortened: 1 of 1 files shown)


ichttt:
Please make this PR for 1.15.x as well

gigaherz:
@ichttt MultiModel doesn't exist in 1.15. We got rid of it alongside all the rest of the old model system, because everything it did could be achieved with vanilla features and/or the new CompositeModel.

ichttt:
Oh ok, thanks for the heads up

TBoshoven:
Thanks gigahertz. 
The implementation in `CompositeModel` uses a more advanced form with `SubModelData` (see [here](https://github.com/MinecraftForge/MinecraftForge/blob/1.15.x/src/main/java/net/minecraftforge/client/model/CompositeModel.java#L67-L78), where the submodel can use separate data from the main model.

I'd be happy to backport that to `MultiModel` but, since it's deprecated in 1.14 anyway, I figured a simpler implementation would be enough.
My reasoning for this is that the `MultiModel` is coming from the blockstate definition (at least for me), which means that it will not need to read its submodel data (there's no way to do so from the JSON definition).
As long as the submodels do not use any external Forge models that require model data that might somehow clash, this is going to be sufficient.
Random clashes very unlikely thanks to the `ModelProperty` mechanism and the fact that there are no builtin Forge model properties. The only use case I see that really won't work here is including the same part twice, with different properties.
It'll definitely cover my use case.

TBoshoven:
Since I got questions about why this is needed (class is deprecated), [this still gets called by Forge](https://github.com/MinecraftForge/MinecraftForge/blob/dc145a3/src/main/java/net/minecraftforge/client/model/BlockStateLoader.java#L242) in the 1.14 branch.

TBoshoven:
I did not realize #6180 was merged. That may make this whole thing unnecessary.

TBoshoven:
I ended up solving my problem through variant blockstates (multipart don't pass down the extra data) and composite models.
I'd have preferred to use multiparts, but that would require patching those and I'm not sure that's something I want to get into.

Feel free to either merge this (it should work) or close it (it won't apply in 1.15).

PR #6536, opened 2020-02-26
Add projection matrix to RenderWorldLastEvent

Nightenom:
In our mod, we need to render structures during RenderWorldLastEvent. Those structures can go big (even far beyond render distance) so to reduce the load in the client thread we are aiming to mimic optimization used in vanilla code as much as possible.
This matrix is used for ClippingHelperImpl class and is passed into WorldRenderer from GameRenderer, it's a local variable and not stored anywhere.

EDIT: fixed partialTicks (finishTimeNano was passed in instead), added finishTimeNano

### DIFF
--- a/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
+++ b/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
@@ -172,9 +172,9 @@ public class ForgeHooksClient
         return MinecraftForge.EVENT_BUS.post(new DrawHighlightEvent(context, info, target, partialTicks, matrix, buffers));
     }
 
-    public static void dispatchRenderLast(WorldRenderer context, MatrixStack mat, float partialTicks)
+    public static void dispatchRenderLast(WorldRenderer context, MatrixStack mat, float partialTicks, Matrix4f projectionMatrix)
     {
-        MinecraftForge.EVENT_BUS.post(new RenderWorldLastEvent(context, mat, partialTicks));
+        MinecraftForge.EVENT_BUS.post(new RenderWorldLastEvent(context, mat, partialTicks, projectionMatrix));
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -172,9 +172,9 @@ public static boolean onDrawBlockHighlight(WorldRenderer context, ActiveRenderIn
         return MinecraftForge.EVENT_BUS.post(new DrawHighlightEvent(context, info, target, partialTicks, matrix, buffers));
     }
 
-    public static void dispatchRenderLast(WorldRenderer context, MatrixStack mat, float partialTicks)
+    public static void dispatchRenderLast(WorldRenderer context, MatrixStack mat, float partialTicks, Matrix4f projectionMatrix, long finishTimeNano)
```
ichttt, line 181: This is a binary breaking change, please keep the old method and mark it for removal in 1.16

src/main/java/net/minecraftforge/client/event/RenderWorldLastEvent.java
```diff
@@ -21,18 +21,24 @@
...
 public class RenderWorldLastEvent extends net.minecraftforge.eventbus.api.Event
 {
     private final WorldRenderer context;
     private final MatrixStack mat;
     private final float partialTicks;
-    public RenderWorldLastEvent(WorldRenderer context, MatrixStack mat, float partialTicks)
+    private final Matrix4f projectionMatrix;
+    private final long finishTimeNano;
+
+    public RenderWorldLastEvent(WorldRenderer context, MatrixStack mat, float partialTicks, Matrix4f projectionMatrix, long finishTimeNano)
```
ichttt, line 45: Also binary breaking. Please keep the old constructor as well and add something like `@Deprecated //TODO Remove in 1.16` to the old constructor

ichttt (changes requested):

src/main/java/net/minecraftforge/client/event/RenderWorldLastEvent.java
Nightenom: Done

PR #6537, opened 2020-02-27
Supply matrixstack to IRenderHandler where possible. Closes #6439

### LINKED ISSUE #6439: IRenderHandler needs access to the MatrixStack
**Minecraft Version:** 1.15

williewillus:
Sky renderhandlers are completely useless currently because you don't have the matrix stack representing the current transform.

This isn't very clean and I'd rather just blow up irenderhandler and split it up, but an RB has already been released, so hacks it is.

### DIFF
--- a/patches/minecraft/net/minecraft/client/renderer/WorldRenderer.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/WorldRenderer.java.patch
@@ -73,19 +73,39 @@
              IVertexBuilder ivertexbuilder2 = irendertypebuffer$impl.getBuffer(RenderType.func_228659_m_());
              this.func_228429_a_(p_228426_1_, ivertexbuilder2, p_228426_6_.func_216773_g(), d0, d1, d2, blockpos, blockstate);
           }
-@@ -1380,6 +1390,11 @@
+@@ -1380,6 +1390,15 @@
     }
  
     public void func_228424_a_(MatrixStack p_228424_1_, float p_228424_2_) {
 +      net.minecraftforge.client.IRenderHandler renderer = this.field_72769_h.func_201675_m().getSkyRenderer();
 +      if (renderer != null) {
-+         renderer.render(this.field_72773_u, p_228424_2_, this.field_72769_h, this.field_72777_q);
++         if (renderer instanceof net.minecraftforge.client.SkyRenderHandler) {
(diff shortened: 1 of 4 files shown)


src/main/java/net/minecraftforge/client/IRenderHandler.java
```diff
@@ -19,14 +19,27 @@
...
 {
     @OnlyIn(Dist.CLIENT)
+    @Deprecated // Use MatrixStack override, TODO 1.16 remove
     void render(int ticks, float partialTicks, ClientWorld world, Minecraft mc);
+
+    /**
+     * @param matrixStack The matrix stack. Can be null when called at sites where vanilla does not supply one.
+     */
+    @OnlyIn(Dist.CLIENT)
+    default void render(int time, float partialTicks, ClientWorld world, @Nullable MatrixStack matrixStack) {
```
LexManos: 1) This breaks the @FunctionalInterface on certian compilers IIRC. As it only works with one method.
2) Why remove the MC argument? I understand you can get it from Minecraft.getInstane, but having it passed in seems like better practice.

src/main/java/net/minecraftforge/client/IRenderHandler.java
```diff
@@ -19,14 +19,27 @@
...
 
+import com.mojang.blaze3d.matrix.MatrixStack;
 import net.minecraftforge.api.distmarker.Dist;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.world.ClientWorld;
 import net.minecraftforge.api.distmarker.OnlyIn;
 
+import javax.annotation.Nullable;
+
+// TODO 1.16 Split this apart for the various uses instead of stuffing everything into one interface.
```
LexManos: Why not just do that now and make this class as deprecated, and have the children extend it?

src/main/java/net/minecraftforge/client/IRenderHandler.java
tterrag1098: It won't, plenty of stdlib functional interfaces have lots of default methods. Just look at Comparator.

williewillus:
updated to address comments.
added three new subinterfaces of IRenderHandler, with the eventual goal of killing IRenderHandler completely. patched worldrenderer to do instanceof and call the more specific method if possible.

Note: WeatherRenderHandler is empty because there is no matrix stack there, so its signature matches that of IRenderHandler, which it inherits. Once IRenderHandler is deleted, its method will be moved down into WeatherRenderHandler.

HellFirePvP:
Is there a chance we'd be seeing this for 1.15 - or is this estimated to only be included in 1.16+ going forward?

Issue #6514, opened 2020-02-11
[1.15] Block loot table for shears should use tag


yanny7:
It will be very helpful if loot tables for shearing will be changed:

- now there is predicate for shear item
```
{
  "condition": "minecraft:match_tool",
  "predicate": {
    "item": "minecraft:shears"
  }
}
```
- with tag e.g. forge:shears developers can easely implement shearing for custom shears without adding new loot tables.
```
{
  "condition": "minecraft:match_tool",
  "predicate": {
    "tag": "forge:shears"
  }
}
```

bloche1871:
There is another issue related to this that will not be solved by the above: items with the shears tag will trigger tripwires when breaking them, unlike vanilla shears.

Shadows-of-Fire:
This issue was resolved by #6765
A separate issue should be opened relating to the tripwire problem, it is out of the scope of this issue.
Do note that the tripwire problem has been solved in 1.17 by https://github.com/MinecraftForge/MinecraftForge/pull/7997

PR #6540, opened 2020-02-29
[1.14.x] Created shears tag (#6514)

yanny7:
see #6514


l-Luna:
This is still relevant, I think?

LexManos:
Do not bump without useful information, no, this pr will. ever be accepted.
As we do not allow manually crafted jsons. It much use the data gen system.
It also has very little explanation and apparently no support from other modders.

Issue #6189, opened 2019-09-25
Custom world types specified in server.properties are not recognized.


Dooglamoo:
**Minecraft Version:** 1.14.4

**Forge Version:** 1.14.4-28.1.11

**Logs:** NA

**Steps to Reproduce:**
 1. Create a mod with a new world type
 2. Specify the world type in server.properties
 3. run the "Server" configuration

**Description of issue:**
The new world type is not used. The server runs with "default" world type instead.
It appears that the server.properties file is loaded before the mod is loaded, so the new world type that the mod creates is not yet in the world types list at the time that server.properties is loaded.

StigLennart:
fwiw, I found a workarround: 

You can create a dooglamoo world in singleplayer, then simply copy/paste that savedata towards your server/world folder. 

from here on I found that Minecraft will generate the terrain accordingly, even though the properties file will still reset to 'default'. This way you also have the added bonus of being able to customize the dooglamoo world gen through the singleplayer menu options

PR #6545, opened 2020-03-03
Defer writing the server.properties file until after mods have loaded

dags-:
Proposed fix for #6189 

This prevents the server from overwriting the server.properties file until after mods have loaded, allowing their world-types etc to be used in it.

It simply re-reads the properties file after mod-loading and updates the server's copy. The call to `ServerPropertiesProvider.func_219033_a(..)` also performs a save, ensuring the file is created and contains valid values.

This also maintains the vanilla behaviour of creating the settings file when the server stops early due to the eula not being accepted, the initSettings flag being present, or an invalid world being specified.

Hope this is ok!

### DIFF
--- a/patches/minecraft/net/minecraft/server/MinecraftServer.java.patch
+++ b/patches/minecraft/net/minecraft/server/MinecraftServer.java.patch
@@ -172,20 +172,16 @@
  
        try {
           OptionSet optionset = optionparser.parse(p_main_0_);
-@@ -888,10 +913,11 @@
+@@ -888,7 +913,7 @@
  
           Path path = Paths.get("server.properties");
           ServerPropertiesProvider serverpropertiesprovider = new ServerPropertiesProvider(path);
 -         serverpropertiesprovider.func_219035_b();
-+         if (!Files.exists(path)) serverpropertiesprovider.func_219035_b();
++         if (optionset.has(optionspec1) || !Files.exists(path)) serverpropertiesprovider.func_219035_b();
           Path path1 = Paths.get("eula.txt");
           ServerEula servereula = new ServerEula(path1);
           if (optionset.has(optionspec1)) {
-+            serverpropertiesprovider.func_219035_b();
-             field_147145_h.info("Initialized '" + path.toAbsolutePath().toString() + "' and '" + path1.toAbsolutePath().toString() + "'");
-             return;
-          }
-@@ -910,6 +936,10 @@
+@@ -910,6 +935,10 @@
           GameProfileRepository gameprofilerepository = yggdrasilauthenticationservice.createProfileRepository();
           PlayerProfileCache playerprofilecache = new PlayerProfileCache(gameprofilerepository, new File(s, field_152367_a.getName()));
           String s1 = Optional.ofNullable(optionset.valueOf(optionspec9)).orElse(serverpropertiesprovider.func_219034_a().field_219021_o);
@@ -196,7 +192,7 @@
           final DedicatedServer dedicatedserver = new DedicatedServer(new File(s), serverpropertiesprovider, DataFixesManager.func_210901_a(), yggdrasilauthenticationservice, minecraftsessionservice, gameprofilerepository, playerprofilecache, LoggingChunkStatusListener::new, s1);
           dedicatedserver.func_71224_l(optionset.valueOf(optionspec7));
           dedicatedserver.func_71208_b(optionset.valueOf(optionspec10));
-@@ -927,6 +957,7 @@
+@@ -927,6 +956,7 @@
           Thread thread = new Thread("Server Shutdown Thread") {
              public void run() {
                 dedicatedserver.func_71263_m(true);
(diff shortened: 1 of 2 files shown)


patches/minecraft/net/minecraft/server/dedicated/DedicatedServer.java.patch
```diff
@@ -1,46 +1,56 @@
 --- a/net/minecraft/server/dedicated/DedicatedServer.java
 +++ b/net/minecraft/server/dedicated/DedicatedServer.java
-@@ -97,6 +97,7 @@
+@@ -15,6 +15,7 @@
+ import java.net.InetAddress;
+ import java.net.Proxy;
+ import java.nio.charset.StandardCharsets;
++import java.nio.file.Paths;
```
bl4ckscor3: Import in patch file. It's not even used.

dags-: It's used on line [25](https://github.com/MinecraftForge/MinecraftForge/blob/77a55ae3180a22ee698d7395ec800a9fd2a4d992/patches/minecraft/net/minecraft/server/dedicated/DedicatedServer.java.patch#L25)

bl4ckscor3: Sorry, I missed that. Either way, patches shouldn't contain imports. You should always directly reference a class in patches, if it's not already imported, to minimize patch size. So line 25 needs to be `this.field_71340_o.func_219033_a(properties -> ServerProperties.func_218985_a(java.nio.file.Paths.get("server.properties")));`

DaemonUmbra: Regardless of if it’s used or not, patch files should NEVER have imports, always use Fully Qualified Names instead.

dags-: Okedoke, thanks for clarifying

patches/minecraft/net/minecraft/server/MinecraftServer.java.patch
```diff
@@ -172,26 +172,45 @@
...
+          Path path1 = Paths.get("eula.txt");
+          ServerEula servereula = new ServerEula(path1);
+          if (optionset.has(optionspec1)) {
++            serverpropertiesprovider.func_219035_b();
+             field_147145_h.info("Initialized '" + path.toAbsolutePath().toString() + "' and '" + path1.toAbsolutePath().toString() + "'");
+             return;
+          }
+ 
+          if (!servereula.func_154346_a()) {
++            serverpropertiesprovider.func_219035_b();
```
Alpvax: Am I correct in thinking that this wipes any changes to the properties if you have forgotten to accept the EULA (which I always forget to do)?
Sorry, it's a bit hard to read patch files on my phone, so I could be wrong about it, but I would rather it didn't affect the file if the EULA hasn't been accepted. Otherwise there's a risk that you change some settings, run the server, accept the EULA then re-run expecting your changed settings to apply and they won't.

dags-: It'd behave the same as the unmodified code does now; reset the `level-type` property to default if it was set to anything non-vanilla.

Other properties (providing they're valid) would remain as you configured them.

Alpvax: Could we not leave this out?
It would decrease the patch size, leave user changes intact, and any invalid settings would be corrected next run.

dags-: I'd assume that for most people's first run of a server there wouldn't be a server.properties file and they'd expect it to be generated. I realise that's technically what the init flag is for but this has been the behaviour since I can remember.

A better solution would probably be to check if the file exists before deciding whether to write it or not, but would increase the patch size by a couple of lines rather than decrease it - not sure how much of a concern that would be.

Alpvax: Yeah, wasn't really thinking about that, it probably should be generated even if the EULA hasn't been accepted. For some reason I was thinking that the file was already there, when in most cases it wouldn't be.

Alpvax: ~~I think this should definitely not be here anymore, because if it didn't exist it would have been created by the line above, but if it didn't it shouldn't be overwritten just because the EULA isn't accepted.~~

EDIT: Either I misread the patch or it didn't display correctly on my phone, I thought this line was still being added in addition to the if exist change.

dags-: This is outdated, please see 2513a10.

patches/minecraft/net/minecraft/server/MinecraftServer.java.patch
```diff
@@ -172,6 +172,15 @@
  
        try {
           OptionSet optionset = optionparser.parse(p_main_0_);
+@@ -888,7 +913,7 @@
+ 
+          Path path = Paths.get("server.properties");
+          ServerPropertiesProvider serverpropertiesprovider = new ServerPropertiesProvider(path);
+-         serverpropertiesprovider.func_219035_b();
++         if (optionset.has(optionspec1) || !Files.exists(path)) serverpropertiesprovider.func_219035_b();
```
LexManos, line 180: Typically we just inject the IF above the line so the patch file doesn't have a removal line.

dags-: Noted, thanks

PR #6554, opened 2020-03-07
Add "type" field to global loot modifier jsons and fix tests.

gigaherz:
To avoid binary breaking, "type" defaults to be the same as the filename.
Still sick so I'm going to need a sanity check on this. :P

PR #6561, opened 2020-03-13
Removed import IGlobalLootModifierSerializer

mooviies:
Compilation error on test project because of non existing IGlobalLootModifierSerializer
Changed it to use GlobalLootModifierSerializer instead. 
Changed usage of LootModifier.Serializer for GlobalLootModifierSerializer since LootModifier.Serializer doesn't exist.


AfterRebelion:
This is already part of https://github.com/MinecraftForge/MinecraftForge/pull/6554

mooviies:
Oh indeed it is.

PR #6562, opened 2020-03-13
Add renderer to RenderNameplateEvent

naqaden:
Adds renderer access previously provided by RenderLivingEvent.Specials.Pre, and includes packedLight to the same end: for mods to cancel/deny the event and render their own nameplate.

[forum thread](https://www.minecraftforge.net/forum/topic/81771-1152-rendernameplateevent-has-no-renderer/)

### DIFF
--- a/src/main/java/net/minecraftforge/client/event/RenderNameplateEvent.java
+++ b/src/main/java/net/minecraftforge/client/event/RenderNameplateEvent.java
@@ -60,7 +60,7 @@ public class RenderNameplateEvent extends EntityEvent
     private final IRenderTypeBuffer renderTypeBuffer;
     private final int packedLight;
 
-    @Deprecated
+    @Deprecated //TODO 1.16: upon removal, also remove @Nullable on getEntityRenderer(), and update its javadoc
     public RenderNameplateEvent(Entity entity, String content, MatrixStack matrixStack, IRenderTypeBuffer renderTypeBuffer)
     {
         this(entity, content, null, matrixStack, renderTypeBuffer, 0);
--- a/src/main/java/net/minecraftforge/client/event/RenderNameplateEvent.java
+++ b/src/main/java/net/minecraftforge/client/event/RenderNameplateEvent.java
@@ -21,6 +21,7 @@ package net.minecraftforge.client.event;
 
 import com.mojang.blaze3d.matrix.MatrixStack;
 import net.minecraft.client.renderer.IRenderTypeBuffer;
+import net.minecraft.client.renderer.entity.EntityRenderer;
 import net.minecraft.entity.Entity;
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.event.entity.EntityEvent;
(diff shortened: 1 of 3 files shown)


src/main/java/net/minecraftforge/client/event/RenderNameplateEvent.java
```diff
@@ -32,33 +33,41 @@
...
 
     private String nameplateContent;
     private final String originalContent;
+    private final EntityRenderer<T> entityRenderer;
     private final MatrixStack matrixStack;
     private final IRenderTypeBuffer renderTypeBuffer;
+    private final int packedLight;
 
-    public RenderNameplateEvent(Entity entity, String content, MatrixStack matrixStack, IRenderTypeBuffer renderTypeBuffer)
+    public RenderNameplateEvent(Entity entity, String content, EntityRenderer<T> entityRenderer, MatrixStack matrixStack, IRenderTypeBuffer renderTypeBuffer, int packedLight)
```
diesieben07: This breaks binary compatibility.

src/main/java/net/minecraftforge/client/event/RenderNameplateEvent.java
```diff
@@ -32,33 +33,41 @@
...
  * This event has a result. {@link HasResult}. <br>
  * ALLOW will force-render name plate/tag, DEFAULT will ignore the hook and continue using the vanilla check
  * & DENY will prevent name plate/tag from rendering<br>
  * <br>
  * This event is fired on the {@link MinecraftForge#EVENT_BUS}.
+ * @param <T> the entity type of the renderer
  **/
 @Event.HasResult
-public class RenderNameplateEvent extends EntityEvent
+public class RenderNameplateEvent<T extends Entity> extends EntityEvent
```
diesieben07: Type parameters on events are meaningless, unless `IGenericEvent` is used, which provides very limited filtering and is not applicable here.

diesieben07 (changes requested):

src/main/java/net/minecraftforge/client/event/RenderNameplateEvent.java
naqaden: I recall it wasn't useful in RenderLivingEvent.Specials.Pre either. Shall I strip it from the class and @SuppressWarnings("rawtypes") instead?

src/main/java/net/minecraftforge/client/event/RenderNameplateEvent.java
```diff
@@ -32,33 +35,47 @@
...
  * This cannot be changed by mods.<br>
+ * {@link #packedLight} contains the sky and block light values used in rendering the name plate/tag.<br>
  * <br>
  * This event has a result. {@link HasResult}. <br>
  * ALLOW will force-render name plate/tag, DEFAULT will ignore the hook and continue using the vanilla check
  * & DENY will prevent name plate/tag from rendering<br>
  * <br>
  * This event is fired on the {@link MinecraftForge#EVENT_BUS}.
  **/
+@SuppressWarnings("rawtypes")
```
ichttt: Use `EntityRenderer<?>` instead of raw `EntityRenderer`, don't just suppress the warning

ichttt (approved):

PR #6567, opened 2020-03-14
[1.15] allow different item attribute modifier based on the applied entity

Cheaterpaul:
this allows for an custom item to override IForgeItem#getAttributeModifiers to return different attribute modifier based on the entity the modifier should be applied to.

### DIFF
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
@@ -67,7 +67,7 @@ public interface IForgeItem
     /**
      * ItemStack sensitive version of getItemAttributeModifiers
      */
-    @SuppressWarnings("deprecation")
+    @Deprecated //TODO 1.16 remove in favor of overloaded method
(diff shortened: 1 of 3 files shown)


diesieben07 (changes requested):
These changes are not binary compatible.

Cheaterpaul:
hope this is binary compatible.

not sure about the IForgeItem, maybe inline that for 1.16

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -73,6 +73,16 @@ default Item getItem()
         return getItem().getAttributeModifiers(slot);
```
williewillus, line 73: Can you mark this old method as deprecated for removal in 1.16?

williewillus: (not this line, but the method it's in)

LexManos:
Yes.. but why?
Also you need an exec file to minimize your patch as you've got hunks that are just name changes.

PR #6570, opened 2020-03-17
[1.15.2] Allow ForgeConfigSpec to better support strings and lists

Senmori:
This PR allows ForgeConfigSpec to better support strings and lists in configuration files.

This introduces two new `ConfigValue` subclasses. `StringValue` and `ListValue<T>`.

The problem with the current implementation is that Strings are not supported as well as the primitive values.
This would allow users to migrate from `ConfigValue<String>` to `StringValue`.

Lists are the primary reason for this PR. List support in `ConfigValue` is a bit lacking.
This allows list elements to be converted to their proper type via a given converter.

Unfortunately, I had to put `defineString` instead of just using `define` because of type erasure in the suppliers/lists. I still think this is acceptable because it's still clear as to what the user is defining.

Proof of test working: https://i.imgur.com/abuwqeC.png


src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
```diff
@@ -329,41 +364,93 @@ public Object correct(Object value) {
                 }
             }, defaultSupplier);
         }
+        /** @see #define(List, Supplier, Predicate, Function) **/
+        public <T> ListValue<T> define(String path, List<T> defaultValue, Function<Object, T> elementConverter) {
```
LexManos: These should all be defineList, like all the list related siblings

PR #6571, opened 2020-03-17
Modded Hill Biomes

valoeghese:
A few patches to minecraft which allows modders to override a `getHill` method in their custom biome class in order to specify a custom hill biome in overworld generation.

This method can be used to create more variation in biomes in the vanilla generator.

This method can also be called by modded world generators in order to add these modded custom hill biomes to their custom world generators.

I have tested the code. Please specify any changes that should be made. Thanks. :)

### DIFF
--- a/patches/minecraft/net/minecraft/world/biome/Biome.java.patch
+++ b/patches/minecraft/net/minecraft/world/biome/Biome.java.patch
@@ -41,7 +41,7 @@
                 return true;
              }
           }
-@@ -375,6 +375,12 @@
+@@ -375,6 +375,35 @@
        return this.field_185364_H;
     }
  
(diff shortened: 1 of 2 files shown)


hYdos (approved):
this comment is not controversial but the commit is actually good i approve.

patches/minecraft/net/minecraft/world/biome/Biome.java.patch
```diff
@@ -50,11 +50,34 @@
...
++      if (this == Biomes.field_222370_aw) return Biomes.field_222371_ax;
++      if (this == Biomes.field_76771_b) return Biomes.field_150575_M;
++      if (this == Biomes.field_203615_U) return Biomes.field_203618_X;
++      if (this == Biomes.field_203616_V) return Biomes.field_203619_Y;
++      if (this == Biomes.field_76776_l) return Biomes.field_203620_Z;
++      if (this == Biomes.field_76770_e) return Biomes.field_150580_W;
++      if (this == Biomes.field_150588_X) return Biomes.field_150587_Y;
++      if (net.minecraft.world.gen.layer.LayerUtil.func_202826_a(Registry.field_212624_m.func_148757_b(this), Registry.field_212624_m.func_148757_b(Biomes.field_150607_aa))) return Biomes.field_150589_Z;
++      if ((this == Biomes.field_150575_M || this == Biomes.field_203618_X || this == Biomes.field_203619_Y || this == Biomes.field_203620_Z) && rand.func_202696_a(3) == 0) return rand.func_202696_a(2) == 0 ? Biomes.field_76772_c : Biomes.field_76767_f;
++      return this;
```
ichttt: This should default to a simple `return null;`

patches/minecraft/net/minecraft/world/gen/layer/HillsLayer.java.patch
```diff
@@ -0,0 +1,11 @@
+--- a/net/minecraft/world/gen/layer/HillsLayer.java
++++ b/net/minecraft/world/gen/layer/HillsLayer.java
+@@ -97,7 +97,7 @@
+             l = field_202809_p;
+          } else if ((i == LayerUtil.field_202830_a || i == LayerUtil.field_203636_g || i == LayerUtil.field_203637_i || i == LayerUtil.field_203638_j) && p_215723_1_.func_202696_a(3) == 0) {
+             l = p_215723_1_.func_202696_a(2) == 0 ? field_202812_s : field_202803_j;
+-         }
++         } else l = Registry.field_212624_m.func_148757_b(Registry.field_212624_m.func_148745_a(i).getHill(p_215723_1_));
```
ichttt: This should be changed to use the biome provided from `getHill` if it doesn't return null, and use the vanilla code path otherwise

ichttt (changes requested):
I think this is a useful feature, but there are some things this can be made more universal.
The suggested changes would allow registry overrides to change the hills as well, and we don't have to copy vanilla code (and keep it synced)

patches/minecraft/net/minecraft/world/biome/Biome.java.patch
valoeghese: do you mean to just replace the `return this;`, or to replace the whole method body?

ichttt: The entire method body, to avoid copying vanilla code

valoeghese: from your previous comment I assume you mean to replace the whole method body

valoeghese: ah I see, thanks for the confirmation

valoeghese:
updated code for requested changes

patches/minecraft/net/minecraft/world/gen/layer/HillsLayer.java.patch
```diff
@@ -0,0 +1,13 @@
+--- a/net/minecraft/world/gen/layer/HillsLayer.java
++++ b/net/minecraft/world/gen/layer/HillsLayer.java
+@@ -59,7 +59,9 @@
+ 
+       if (p_215723_1_.func_202696_a(3) == 0 || k == 0) {
+          int l = i;
+-         if (i == field_202799_f) {
++         Biome hill = Registry.field_212624_m.func_148745_a(i).getHill(p_215723_1_);
```
ichttt: `getByValue` can be null, so this should check for `null` before the getHill call as well, like this:
```
Biome biome = Registry.BIOME.getByValue(i);
Biome hill = biome == null ? null : biome.getHill(p_215723_1_);
```

patches/minecraft/net/minecraft/world/biome/Biome.java.patch
```diff
@@ -50,11 +50,15 @@
 +      if (this == Biomes.field_76789_p || this == Biomes.field_76788_q) return Biomes.field_76788_q;
 +      return Biomes.field_76781_i;
 +   }
++
++   public Biome getHill(net.minecraft.world.gen.INoiseRandom rand) {
```
ichttt, line 55: Should be marked with `@Nullable`

ichttt (changes requested):
Two small additional changes, otherwise looking good now

ichttt (approved):

Earthcomputer (approved):

Issue #6286, opened 2019-10-30
PlantType.create(String) NullPointerException


Oshi41:
**Minecraft Version:** 1.14.4

**Forge Version:** 28.1.0

**Logs:** https://gist.github.com/Oshi41/4a09a2674b7bba3d1962b908a79bd3e7

**Steps to Reproduce:**
 1. Creating custom PlantType to restrict flowers be planted on vanilla block types
 2. Run game and plant your flower with special PlantType

**Description of issue:**
PlantType.create(String) always returns null
**Block.canSustainPlant** method use **switch** for PlantType which is breaks the game

Any workaround how to create flower can be planted only on non-vanilla blocks?

Issue #5371, opened 2019-01-17
[1.14.4] Discussion about Plant changes


Shadows-of-Fire:
As mentioned in #5362 1.13 will be a good time to refactor the plant system that forge currently has in place.  Currently, we have EnumPlantType, IPlantable, and IGrowable.  These three pieces aren't sufficient to create a proper ecosystem for plants to exist and interact with machines and each other.

To resolve this issue, we need to figure out what it is modders need from a plant system.  The most basic change would be to start out with changing IGrowable and IPlantable.

The first change that should be made is EnumPlantType should no longer be an Enum.  Additionally, certain extra defaults (if suggested) could be added.

IPlantable should be changed to have additional contexts, specifically ItemStack, maybe EntityPlayer and EnumHand if possible.  This would allow for mods that need extra data to determine plant state to use this system.

IGrowable should be renamed to IPlant, and have additional methods for harvesting added to it.
Possible ideas for such methods include `boolean isMature(World, BlockPos, IBlockState)` and `void harvest(NonNullList<ItemStack>, World, BlockPos, IBlockState)`.  This would allow for much easier determination of when and how to harvest a planted object, which is currently impossible.

Currently, the problem is determining the needed contexts for people who make plants, seeds, and harvesters.  I'd like anyone who interfaces with plants to state what they would like to see from a redesigned plant system, so that this can be designed and completed before the breaking change window is over.

Shadows-of-Fire:
Another thing that can be done is add a `EnumActionResult plant(World, Pos, ItemStack)` to IPlantable, to allow custom placement.  This could be helpful for taller plants.

Buuz135:
The change to IGrowable -> IPlant looks good as most of the time they are crops (except sapplings). My only requests are `isMature` and `harvest` as it will make life much easier by not having to implement a custom handler for every crop every mod has.

Maybe `harvest` having an extra boolean to replant itself.

DarkGuardsman:
Add a check for 'Can view harvest level [type]' I know a few mods do not like there plants to be automated. However, still would like for UI features or NPCs to check the levels. So we could add either a context based input into isMature or a separate method to solve for this case. 

I don't recall all that was in the old methods and can't check while at work. What I would like to see:

- get all the growth state of a plant
- get the starting state
- get the end state
- get the drops of each state
- get the death state (edge case for mods that wish to poison or destroy crops, default to air)
- get the seed
- get a list of supported blocks/materials
[...]

PR #6573, opened 2020-03-17
Add fix for issue 6286. Changed null return to throw exception as per…

patrick-aoc:
Linked to issue #6286 

PlantType now implements IExtensibleEnum, allowing for custom plant types to be dynamically added into the game via mods.

PlantType.create() has been refactored to throw an exception if the enum has not been extended (as opposed to returning null and yielding the Null Pointer Exceptions from before). On inspecting how other Enums implemented dynamic creation of values, we came across IExtensibleEnum. IExtensibleEnum is an interface which allows for the Forge code to use ASM to dynamically add values into Enums. On further investigation of the docstring of IExtensibleEnum as well as how other classes implemented dynamic Enums, (for example net.minecraft.enchantment.EnchantmentType, net.minecraft.enchantment.EntityClassification, net.minecraft.enchantment.Rarity),  each of these Enums also implement IExtensibleEnum in the same way (by throwing an IllegalStateException which is then caught and replaced by ASM at runtime to create new values).

In addition, The CustomPlantTypeTest is a test mod that registers a custom block and a custom plant of a custom type into the game. The goal of this test mod is to check and see a if custom block can support a custom plant type. 
Note: the “custom plant” will look exactly like the custom block but will function as a flower.

### DIFF
--- a/src/main/java/net/minecraftforge/common/PlantType.java
+++ b/src/main/java/net/minecraftforge/common/PlantType.java
@@ -22,7 +22,7 @@ package net.minecraftforge.common;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.IWorldReader;
 
-// TODO 1.16: This should not be an enum. Change it to ToolType instead
+// TODO 1.16: This should not be an enum. Change it to something that functions similarly to ToolType
 public enum PlantType implements IExtensibleEnum
 {
     Plains,
--- a/src/main/java/net/minecraftforge/common/PlantType.java
+++ b/src/main/java/net/minecraftforge/common/PlantType.java
@@ -22,7 +22,7 @@ package net.minecraftforge.common;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.IWorldReader;
 
-public enum PlantType
+public enum PlantType implements IExtensibleEnum
 {
     Plains,
     Desert,
@@ -45,5 +45,8 @@ public enum PlantType
      * @param name the name of the type of plant, you had better follow the style above
      * @return the acquired {@link PlantType}, a new one if not found.
      */
-    public static PlantType create(String name){ return null; }
+    public static PlantType create(String name)
+    {
+        throw new IllegalStateException("Enum not extended");
+    }
 }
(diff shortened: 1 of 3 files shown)


3TUSK:
No, the fact that it is still an `enum` is a mistake. 
There should have been an interface, probably named `IPlantType`, and the current `PlantType` implements that interface. All methods should refer to the new interface instead.

… or we should consider the possibility of using vanilla tag system.

See #5371.

ichttt:
We can't make change it from an enum now, since that would be binary breaking... But we should leave a TODO there to change it to something like `ToolType` in future versions

src/main/java/net/minecraftforge/common/PlantType.java
```diff
@@ -22,7 +22,7 @@
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.IWorldReader;
 
-public enum PlantType
+public enum PlantType implements IExtensibleEnum
```
ichttt, line 26: Please add a todo comment like `TODO 1.16: This shouldn't be an enum, change it to something like ToolType`

src/test/java/net/minecraftforge/debug/block/CustomPlantTypeTest.java
```diff
@@ -0,0 +1,124 @@
...
+import net.minecraftforge.fml.common.Mod.EventBusSubscriber.Bus;
+import net.minecraftforge.registries.ObjectHolder;
+
+@Mod(CustomPlantTypeTest.MODID)
+@Mod.EventBusSubscriber(bus = Bus.MOD)
+public class CustomPlantTypeTest
+{
+    static final String MODID = "custom_plant_type_test";
+    static final String CUSTOM_SOIL_BLOCK = "test_custom_block";
+    static final String CUSTOM_PLANT_BLOCK = "test_custom_plant";
```
ichttt: `CUSTOM_SOIL_BLOCK ` and `CUSTOM_PLANT_BLOCK ` can be private

ichttt (commented):
Some minor comments. Also, please make this PR for 1.15 as well, as this is the primary branch right now.

Issue #6325, opened 2019-11-19
[1.14.4] IPlantable defaults SeaGrass and SeaPickle blocks to Plains plant type


Wabbit0101:
**Minecraft Version:** 1.14.4

**Forge Version:** 28.1.87

**Steps to Reproduce:**
 1. Check PlantType of either SeaGrassBlock or SeaPickleBlock

**Description of issue:**
The default implementation for IPlantable should probably be extended to handle SeaGrassBlock and SeaPickleBlock to return PlantType.water for these two.

PR #6574, opened 2020-03-17
Added checks for SeaGrass and SeaPickle blocks in IPlantable so that they can be recognized to be of the Water plant type

patrick-aoc:
Linked to issue #6325 

Extra checks have been added for the SEAGRASS and SEA_PICKLE blocks in the IPlantable interface so that they can be recognized to be of the Water plant type instead of just Plains.

Sea Pickles and Sea Grass blocks are blocks that one would generally find in bodies of water, i.e. you wouldn’t find these blocks in forests or caves. At the moment, calling getPlantType on these particular blocks returns the Plains type. Having this sort of behaviour could allow players to place Sea Grass and Sea Pickles on top of blocks that aren’t underwater, which doesn’t make logical sense (unless one wanted this sort of behaviour). 

The SeaPlantPlaceTest is a test mod that has a test method (onBlockPlaced) that is subscribed to the event in which an entity places a block in the game. When an entity places a block, this method gets triggered first checks to see if the block placed was Sea Grass or a Sea Pickle. It then checks to see if the block in which those sea plants are sitting can sustain them.


noobanidus:
Doesn't the sea pickle by vanilla defaults have a different visual and texture when placed on a non-waterlogged blocks, implying that the ability to place them outside of water is intended?

TeamSpen210:
Well, the out-of-water models are called `dead`, and they can only grow while underwater even with bonemeal. So they can be placed there, but it's not an appropriate location and it won't grow.

noobanidus:
> Having this sort of behaviour could allow players to place Sea Grass and Sea Pickles on top of blocks that aren’t underwater

I was referring to this statement which implied the PR would prevent Sea Pickles from being placed on blocks that aren't underwater.

LexManos:
1) 1.15 branch first
2) Sea Pickles and Sea Grass are submerged plants not water plants like lily pads.

PR #6575, opened 2020-03-18
Use the cached isEmpty value in isEmpty()

Shadows-of-Fire:
This change makes isEmpty significantly less computationally expensive, whilist retaining all previous behaviors.

isEmpty() is invoked very frequently by complex item handling mods such as Applied Energistics and Refined Storage, so the speedup could be significant in heavy item processing scenarios.

In general, this is a positive change, with no downside whatsoever,.

### DIFF
--- a/patches/minecraft/net/minecraft/item/ItemStack.java.patch
+++ b/patches/minecraft/net/minecraft/item/ItemStack.java.patch
@@ -31,7 +31,7 @@
        this.field_151002_e = p_i48204_1_ == null ? null : p_i48204_1_.func_199767_j();
        this.field_77994_a = p_i48204_2_;
        if (this.field_151002_e != null && this.field_151002_e.func_77645_m()) {
-@@ -100,6 +106,7 @@
+@@ -100,14 +106,17 @@
        }
  
        this.func_190923_F();
@@ -39,7 +39,9 @@
     }
  
     private void func_190923_F() {
-@@ -108,6 +115,8 @@
+       this.field_190928_g = false;
+-      this.field_190928_g = this.func_190926_b();
++      this.field_190928_g = this.getItemRaw() == null || this.getItemRaw() == Items.field_190931_a || this.field_77994_a <= 0;
     }
  
     private ItemStack(CompoundNBT p_i47263_1_) {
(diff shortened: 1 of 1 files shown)


LexManos:
This looks like a microoptomization with no justification. We typically dont accept those.
At  the end of the day the real world usecase for this is just "optimizing' out a `if_icmple` and `iconst_0` operation... which is just the definition of micro.

Shadows-of-Fire:
I do have a couple skyfactory snapshots where isEmpty() was around 5% total cpu time, let me go dig those up.

LexManos:
Sky Factory is 1.12, and thus we don't care about it.
Beyond that, as stated you're optimizing out two operations. It is most likely some OTHER issue causing isEmpty to be called far more often then it should be.

Shadows-of-Fire:
SkyFactory may be 1.12, but the method remains unchanged from that version, thus this change still holds value.

Anyway, the complexity decrease is as follows:

A call to isEmpty() will perform the following instructions:
```
  public boolean isEmpty();
    Code:
       0: aload_0
       1: getstatic     #220                // Field EMPTY:Lnet/minecraft/item/ItemStack;
       4: if_acmpne     9
       7: iconst_1
       8: ireturn
       9: aload_0
      10: invokespecial #226                // Method getItemRaw:()Lnet/minecraft/item/Item;
      13: ifnull        39
      16: aload_0
      17: invokespecial #226                // Method getItemRaw:()Lnet/minecraft/item/Item;
      20: getstatic     #231                // Field net/minecraft/item/Items.AIR:Lnet/minecraft/item/Item;
      23: if_acmpeq     39
      26: aload_0
      27: getfield      #124                // Field count:I
      30: ifgt          37
      33: iconst_1
      34: goto          38
      37: iconst_0
      38: ireturn
      39: iconst_1
      40: ireturn
```

Obviously, the cost is significantly reduced if this is the empty stack, but that is not always the case, and the distribution of empty stack as `ItemStack.EMPTY` vs other empty stacks is variable.

For all other stacks, almost every instruction is run, as no other stack should, in a normal environment, ever have an item of null or air, as that is what the empty stack is for.

Most importantly, this method is also called from all invocations of setCount(), not just isEmpty(), so rapid manipulation of stack counts also invokes this cost, but the only way to free that cost up from setCount would be to disallow stacks of air or null other than the empty stack and change the empty state update to `this == EMPTY || count <= 0`

The reduced method only has two instructions, getfield and ireturn.

williewillus:
I would examine the compiled machine code instead of the bytecode if you're looking to actually quantify the performance benefits

LexManos:
Closing as this is a micro optimization that has no real bench marking or justification.

PR #6579, opened 2020-03-22
Allow colors of tooltip to be specified in GuiUtils tooltip methods

williewillus:
In Vazkiis mods we have giant helper methods to draw tooltips when we could just be using Forge's, all because we need custom colors. I imagine the same for many other large mods that use tooltips with custom-colored borders.

The methods in GuiUtil already support changing the colors using an event, so it makes little sense to not allow them to take the colors as parameters too.

Added three parameters for the colors and add overloads for the old signatures for backward compatibility.


Kamefrede:
Bump. Would be good to have!

PR #6583, opened 2020-03-24
Add ability to create custom (client-side only) chat commands

henne90gen:
I was looking for a way to create custom chat commands.
I think this simple fix would enable people to add custom chat commands to their mods.

All this patch does, is hook into the standard way Minecraft registers chat commands and make it expandable.
This is done by using a static list of commands that should be registered.
A mod simple adds their commands to this list during the initialization and they'll be added to the game.

Another way of doing it would be by creating an annotation and scanning the classes of the mod for this annotation and add the custom commands that way.
This would be a lot more convenient for the mod creator, but also much more complicated to implement.

### DIFF
--- a/patches/minecraft/net/minecraft/command/Commands.java.patch
+++ b/patches/minecraft/net/minecraft/command/Commands.java.patch
@@ -1,6 +1,23 @@
 --- a/net/minecraft/command/Commands.java
 +++ b/net/minecraft/command/Commands.java
-@@ -194,7 +194,15 @@
+@@ -94,6 +94,7 @@
+ import net.minecraft.util.text.TranslationTextComponent;
+ import net.minecraft.util.text.event.ClickEvent;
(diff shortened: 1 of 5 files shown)


tterrag1098:
There is no point to this, it's already perfectly possible to register commands.

henne90gen:
Could you elaborate on how to do that?

tterrag1098:
https://github.com/MinecraftForge/MinecraftForge/blob/83ebdcd8f25c9bc7a99cc121d3ae67bb88995503/src/main/java/net/minecraftforge/common/ForgeMod.java#L136-L137

Next time, seek out help on the forum or Discord first.

henne90gen:
This does not work for client side only mods, though.
I can try to find a solution that looks similar to the one that is available for the server side.

tterrag1098:
Neither does your change, it just adds a weird sideways hook into the same CommandDispatcher.

PR #6586, opened 2020-03-25
Reworked AnvilUpdateEvent for more flexibility and to solve existing problems

Fuzss:
So here is a reworked version of the `AnvilUpdateEvent`. Why is this necessary in the first place?

**Problems with the current implementation**
- Client desync as `RepairContainer::detectAndSendChanges` is not being called, a manual call is impossible; this causes e. g. the `Too Expensive!` text to not display when it should
- No access to whether the player is in creative mode; this is used for disabling the level cap in vanilla, modders might want to do something similar
- Modders always need to copy most of the `RepairContainer::updateRepairOutput` method, even when doing such simple things as adding / changing the output
&nbsp;

Solving all three of those problems has most notably lead to changes to `AnvilUpdateEvent` as well as `RepairContainer.java.patch`. For better understanding here is a deobfuscated version of the most relevant part of the changes to `RepairContainer.java.patch`.
```
          this.materialCost = 0;
+         boolean flag = false;
+         net.minecraftforge.event.AnvilUpdateEvent evt = net.minecraftforge.common.ForgeHooks.onAnvilUpdate(itemstack, itemstack2, this.repairedItemName, this.player.abilities.isCreativeMode);
+         this.materialCost = evt.getMaterialCost();
+         i += evt.getCost();
+         if (evt.isCanceled()) {
+            this.outputSlot.setInventorySlotContents(0, evt.getOutput());
+            this.maximumCost.set(i);
+            if (!evt.getOutput().isEmpty()) this.detectAndSendChanges();
+            return;
+         }
+         if (!evt.getOutput().isEmpty()) {
+            itemstack1 = evt.getOutput();
+            map = EnchantmentHelper.getEnchantments(evt.getOutput());
+         } else
          if (!itemstack2.isEmpty()) {
```

**Three ways of using the new implementation**
- Cancel the event
**Behaviour:** output slot remains empty
**Possible before?** Yes
**Changes:** `outputSlot` and `maximumCost` are properly reset
- Cancel the event and set your own output
**Behaviour:** no vanilla logic runs, modder has full control
**Possible before?** Yes
**Changes:** `RepairContainer::detectAndSendChanges` is called as intended and there's access to whether the player is in creative mode
- Change the output without canceling the event
**Behaviour:** vanilla code for setting the output will be skipped, but the rest of `RepairContainer::updateRepairOutput` executes as usual (renaming, calculating cost, checking level cap, etc.)
**Possible before?** No
&nbsp;

Some questions that might arise.

**Why has the call to this event moved outside of the `if (!itemstack2.isEmpty())`-block?**
A bunch of mods use this event for removing the vanilla level cap. Previously that type of removal would only apply when combing two items. Now it always applies, even when renaming an item.

**Why is the event now solely handled in the vanilla class and not in `ForgeHooks`?**
Access to multiple local variables is required (itemstack1 = output, i = maximum cost, map = enchantments, needed later).

### DIFF
--- a/src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
+++ b/src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
@@ -27,32 +27,37 @@ import javax.annotation.Nonnull;
 
 /**
  * 
- * AnvilUpdateEvent is fired when a player places items in both the left and right slots of a anvil.
- * If the event is canceled, vanilla behavior will not run, and the output will be set to null.
- * If the event is not canceled, but the output is not null, it will set the output and not run vanilla behavior.
- * if the output is null, and the event is not canceled, vanilla behavior will execute.
+ * This event is fired in {@link net.minecraft.inventory.container.RepairContainer#updateRepairOutput} as long as an item stack is present in the left slot of the anvil.
+ * Maximum cost and material cost will always be set from this event, even when not canceled.
+ * 
+ * If the output stack remains empty and the event is not canceled, vanilla behavior will be executed.
+ * If the output stack is changed it will be set from this event.
+ * If the event is then canceled no vanilla behaviour will apply.
+ * If the event is not canceled vanilla logic for creating the output will be skipped, but the rest of the vanilla method will execute as normal.
  */
 @Cancelable
 public class AnvilUpdateEvent extends Event
 {
     @Nonnull
-    private final ItemStack left;  // The left side of the input
+    private final ItemStack left;               // The left side of the input
     @Nonnull
-    private final ItemStack right; // The right side of the input
-    private final String name;     // The name to set the item, if the user specified one.
+    private final ItemStack right;              // The right side of the input, can be empty
+    private final String name;                  // The name of the left item stack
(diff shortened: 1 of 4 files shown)


pupnewfster:
Removing ATs/methods is not something that you want to be doing in the middle of a minecraft version as it will break binary compatibility. You also should be doing it in a way that does not break backwards compat with any mods that may be using the existing ForgeHooks method

Fuzss:
Thanks, I've now added a separate hook and updated the old one.

src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
```diff
@@ -27,39 +27,45 @@
...
+        this.cost = 0;
+        this.materialCost = 0;
     }
 
     @Nonnull
     public ItemStack getLeft() { return left; }
     @Nonnull
     public ItemStack getRight() { return right; }
     public String getName() { return name; }
+    public String isCreativePlayer() { return creativePlayer; }
```
loordgek: String ? that needs to be a boolean

Fuzss:
Huh, can't see how that slipped through, thanks.

PR #6599, opened 2020-03-31
[1.15.x] Add fix for issue 6286. Changed null return to throw exception 

patrick-aoc:
This was to address the comments made in #6573


src/main/java/net/minecraftforge/common/PlantType.java
```diff
@@ -22,7 +22,8 @@
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.IWorldReader;
 
-public enum PlantType
+// TODO 1.16: This should not be an enum. Change it to ToolType instead
```
ichttt: Sorry, seems like I didn't express myself clear. ToolType is an entirely different thing that should not be mixed with this, but this enum should be changed to a class that functions like ToolType (has a create method, caches instances, etc).

patrick-aoc: Updated. My apologies for misinterpreting!

PR #6600, opened 2020-04-02
Don't resolve symbolic links in ModDirTransformerDiscoverer

LXGaming:
This PR fixes an issue that occurs when the Game Directory is located behind a symbolic link.

The [ModDirTransformerDiscoverer](https://github.com/MinecraftForge/MinecraftForge/blob/0d2a0deef396a46ceadd8cdaeb0deca36770188f/src/fmllauncher/java/net/minecraftforge/fml/loading/ModDirTransformerDiscoverer.java) scans jar files checking if they contain ITransformationService or IModLocator services, If the jar does contain any of these services the path is added to a list, This list is later used by the [ModsFolderLocator](https://github.com/MinecraftForge/MinecraftForge/blob/0d2a0deef396a46ceadd8cdaeb0deca36770188f/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModsFolderLocator.java) which will exclude any jars in this list from being loaded as a Mod.

The issue is when resolving the path to the jar file the [ModDirTransformerDiscoverer](https://github.com/MinecraftForge/MinecraftForge/blob/0d2a0deef396a46ceadd8cdaeb0deca36770188f/src/fmllauncher/java/net/minecraftforge/fml/loading/ModDirTransformerDiscoverer.java) calls `toRealPath` but the [ModsFolderLocator](https://github.com/MinecraftForge/MinecraftForge/blob/0d2a0deef396a46ceadd8cdaeb0deca36770188f/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModsFolderLocator.java) uses `toAbsolutePath` as seen in [FMLPaths](https://github.com/MinecraftForge/MinecraftForge/blob/0d2a0deef396a46ceadd8cdaeb0deca36770188f/src/fmllauncher/java/net/minecraftforge/fml/loading/FMLPaths.java), The difference here is that `toRealPath` will resolve symbolic links where `toAbsolutePath` doesn't, This causes jars that were excluded to get loaded anyway due to the list containing the incorrect path.

The changes I've made remove the call to `toRealPath` and instead make the `modsDir` variable store an absolute path instead of a relative one using the same method as [FMLPaths#L73](https://github.com/MinecraftForge/MinecraftForge/blob/0d2a0deef396a46ceadd8cdaeb0deca36770188f/src/fmllauncher/java/net/minecraftforge/fml/loading/FMLPaths.java#L73) to resolve the path.

Related Issues:
- https://github.com/gorilla-devs/GDLauncher-Next-Issues/issues/15
- https://github.com/LXGaming/MixinBootstrap/issues/10

### DIFF
--- a/src/fmllauncher/java/net/minecraftforge/fml/loading/ModDirTransformerDiscoverer.java
+++ b/src/fmllauncher/java/net/minecraftforge/fml/loading/ModDirTransformerDiscoverer.java
@@ -53,7 +53,7 @@ public class ModDirTransformerDiscoverer implements ITransformerDiscoveryService
     }
 
     private static void scan(final Path gameDirectory) {
-        final Path modsDir = gameDirectory.resolve(FMLPaths.MODSDIR.relative());
+        final Path modsDir = gameDirectory.resolve(FMLPaths.MODSDIR.relative()).toAbsolutePath().normalize();
         transformers = new ArrayList<>();
         locators = new ArrayList<>();
         if (!Files.exists(modsDir)) {
@@ -74,9 +74,9 @@ public class ModDirTransformerDiscoverer implements ITransformerDiscoveryService
         if (LamdbaExceptionUtils.uncheck(() -> Files.size(path)) == 0) return;
         try (ZipFile zf = new ZipFile(new File(path.toUri()))) {
             if (zf.getEntry("META-INF/services/cpw.mods.modlauncher.api.ITransformationService") != null) {
-                transformers.add(path.toRealPath());
+                transformers.add(path);
             } else if (zf.getEntry("META-INF/services/net.minecraftforge.forgespi.locating.IModLocator") != null) {
-                locators.add(path.toRealPath());
+                locators.add(path);
             }
         } catch (IOException ioe) {
             LogManager.getLogger().error("Zip Error when loading jar file {}", path, ioe);


MingStar:
I can confirm resolving symlink check was the issue, as I had the client mods folder symlinked to a local server mods folder to keep them in sync.

The issue was gone after I created a real mods folder on client side, but I prefered the synced mods folder setup

mcejp:
This seems to also affect 1.14.4, which creates a tricky situation, because the classic way of installing mods by direct injection of classes into minecraft.jar doesn't work with the official launcher (it will overwrite the jar file with a vanilla version)

DaemonUmbra:
> This seems to also affect 1.14.4, which creates a tricky situation, because the classic way of installing mods by direct injection of classes into minecraft.jar doesn't work with the official launcher (it will overwrite the jar file with a vanilla version)

That method of modding is ancient, barbaric, a compatibility nightmare worse than coremodding, and is in no way supported, encouraged, or enabled by Forge.

mcejp:
> That method of modding is ancient, barbaric, a compatibility nightmare worse than coremodding, and is in no way supported, encouraged, or enabled by Forge.

That's great, and how exactly does this add value to a discussion about a situation where a Forge bug makes it _not work_?

I will gladly backport the fix to 1.14.4, if there is a non-zero chance of having another release there, but from what I've understood, that's against Forge's deprecation policy.

PR #6601, opened 2020-04-03
Add generic type to DataSerializerEntry

thedarkcolour:


tterrag1098:
Why? Please justify your changes.

thedarkcolour:
It would be helpful to not have to cast DataSerializerEntry.getSerializer() each time I want to use a custom data serializer.

thedarkcolour:
I can't see the error so I'm not sure what's going on.

LexManos:
The error is all the other references to DataSerializerEntry that is not valid anymore now that we must know the type explicitly. It was specifically designed with an unbound type due to restrictions in how the registry system works.

PR #6602, opened 2020-04-04
Added doRenderWorld to RenderTickEvent

18107:
**minecraft.gameRenderer.updateCameraAndRender** has a new argument **renderWorldIn** that is used to decide whether the world should be rendered or not. It appears to be false when the world does not exist (when joining or leaving a world) or when out of memory, and true otherwise.

This commit adds **renderWorldIn** to the **RenderTickEvent** so that mods have access to the relevant arguments when calling **updateCameraAndRender**.

### DIFF
--- a/src/main/java/net/minecraftforge/fml/hooks/BasicEventHooks.java
+++ b/src/main/java/net/minecraftforge/fml/hooks/BasicEventHooks.java
@@ -68,15 +68,15 @@ public class BasicEventHooks
         MinecraftForge.EVENT_BUS.post(new PlayerEvent.ItemSmeltedEvent(player, smelted));
     }
 
-    public static void onRenderTickStart(float timer)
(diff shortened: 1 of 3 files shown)


src/main/java/net/minecraftforge/fml/hooks/BasicEventHooks.java
```diff
@@ -68,15 +68,15 @@ public static void firePlayerSmeltedEvent(PlayerEntity player, ItemStack smelted
         MinecraftForge.EVENT_BUS.post(new PlayerEvent.ItemSmeltedEvent(player, smelted));
     }
 
-    public static void onRenderTickStart(float timer)
+    public static void onRenderTickStart(float timer, boolean doRenderWorld)
```
LexManos, line 71: All of these need defaults, we can't break binary compatibility after a recommended build.

PR #6604, opened 2020-04-05
Update README.md

hYdos:
Since other... modloaders exist i think we should have this renamed :)


hYdos:
leaving this in is spreading misinformation
you shouldn't be closing it

PR #6605, opened 2020-04-06
[1.15.x] Fixed getPartialTicks() in RenderWorldLastEvent

EmanueleGiannuzzi:
getPartialTicks() from RenderWorldLastEvent is returning "p_228378_2_" which is a long called "finishTimeNano" (according to mappings '20200406-1.15.1') 
"p_228378_1_" is "partialTicks"


ichttt:
This PR is wrong, `getPartialTicks` returns the correct result. It was patched in #6536. Make sure to use build 1.15.2-31.1.19 or newer to get the correct result for `getPartialTicks`

PR #6606, opened 2020-04-08
Patch Various Entity Classes to allow Modded Bows and Crossbows

HenryRichard:
There are now new versions of ProjectileHelper.getHandWith and MobEntity#isHolding. These take a Predicate<Item> instead of an Item; this lets the vanilla calling classes use instanceof to check for modded bows and crossbows instead of just vanilla ones.

A test mod (get_hand_with_test) that adds a BowItem and CrossbowItem is included. When a modded bow is given to a Skeleton, Wither Skeleton, Stray, or Illusioner, it properly uses it; when a modded crossbow is given to a pillager it properly uses it as well.

### DIFF
--- a/patches/minecraft/net/minecraft/entity/MobEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/MobEntity.java.patch
@@ -63,16 +63,12 @@
                    this.field_70170_p.func_72960_a(playerentity, (byte)30);
                 }
              }
-@@ -1215,7 +1226,12 @@
+@@ -1214,7 +1225,7 @@
+       }
  
     }
- 
-+   @Deprecated
+-
++   public boolean isHolding(java.util.function.Predicate<Item> itemPredicate) { return itemPredicate.test(this.func_184614_ca().func_77973_b()) || itemPredicate.test(this.func_184592_cb().func_77973_b()); }
     public boolean func_213382_a(Item p_213382_1_) {
        return this.func_184614_ca().func_77973_b() == p_213382_1_ || this.func_184592_cb().func_77973_b() == p_213382_1_;
     }
-+
-+   public boolean isHolding(java.util.function.Predicate<Item> itemPredicate) {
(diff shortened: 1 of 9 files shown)


patches/minecraft/net/minecraft/entity/MobEntity.java.patch
```diff
@@ -63,3 +63,12 @@
                    this.field_70170_p.func_72960_a(playerentity, (byte)30);
                 }
              }
+@@ -1214,7 +1225,7 @@
+       }
+ 
+    }
+-
```
ichttt, line 70: Please don't remove this newline

patches/minecraft/net/minecraft/entity/projectile/ProjectileHelper.java.patch
```diff
@@ -9,3 +9,12 @@
                    if (d0 == 0.0D) {
                       entity = entity1;
                       vec3d = vec3d1;
+@@ -156,7 +156,7 @@
+    public static Hand func_221274_a(LivingEntity p_221274_0_, Item p_221274_1_) {
+       return p_221274_0_.func_184614_ca().func_77973_b() == p_221274_1_ ? Hand.MAIN_HAND : Hand.OFF_HAND;
+    }
+-
```
ichttt, line 16: Same here, keep the newline

ichttt (commented):
Some patch style nitpicks, otherwise LGTM

patches/minecraft/net/minecraft/entity/projectile/ProjectileHelper.java.patch
```diff
@@ -9,3 +9,12 @@
...
++   public static Hand getHandWith(LivingEntity living, Predicate<Item> itemPredicate) { return itemPredicate.test(living.func_184614_ca().func_77973_b()) ? Hand.MAIN_HAND : Hand.OFF_HAND; }
```
LexManos, line 17: For both of these methods, 1) You can use multiple lines for full functions to make it not run off the edge.
2) Honestly this shouldn't be predicate. It should be getRangedWeapon or something like that that looks for bows/crossbows. So that all the callers don't have to specify the type every call.

PR #6610, opened 2020-04-09
Backport: Fix ChunkDataEvent.Load not firing, it now fires in the aync thread. …

phit:
backports c72d6e9

### DIFF
--- a/src/main/java/net/minecraftforge/event/world/ChunkDataEvent.java
+++ b/src/main/java/net/minecraftforge/event/world/ChunkDataEvent.java
@@ -22,6 +22,7 @@ package net.minecraftforge.event.world;
 import net.minecraft.nbt.CompoundNBT;
 import net.minecraft.world.World;
 import net.minecraft.world.chunk.Chunk;
+import net.minecraft.world.chunk.ChunkStatus;
 import net.minecraft.world.chunk.IChunk;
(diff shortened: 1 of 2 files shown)


src/main/java/net/minecraftforge/event/world/ChunkDataEvent.java
```diff
@@ -64,9 +65,17 @@ public CompoundNBT getData()
      **/
     public static class Load extends ChunkDataEvent
     {
-        public Load(IChunk chunk, CompoundNBT data)
+        private ChunkStatus.Type status;
+
+        public Load(IChunk chunk, CompoundNBT data, ChunkStatus.Type status)
```
ichttt, line 78: This is a binary breaking change, please leave the old constructor and deprecate it

phit: what ChunkStatus.Type can I realistically default to in that case?

ichttt: I'd just set it to null, and mark `getStatus` as nullable. Mods shouldn't fire this event so it will be nonnull in 99% of the time, but it's better to make modders that use the new type account for a possible null then break old mods

PR #6615, opened 2020-04-12
Provide extra context to RightClickBlock, create OnItemUse

Shadows-of-Fire:
This PR Re-Adds the proper ray trace result context to PlayerInteractEvent.RightClickBlock that was lost in the 1.13 update (previously we had access to the Vec3d hitvec).

It also adds a new event that is fired when onItemUse would normally be called.  This event is useful for when you want to add additional right click behaviors to arbitrary items, but do not want to manaully recreate all of the logic for handling the block being clicked, as is currently required (and impossible, without the first half of this PR).

### DIFF
--- a/src/main/java/net/minecraftforge/event/entity/player/PlayerInteractEvent.java
+++ b/src/main/java/net/minecraftforge/event/entity/player/PlayerInteractEvent.java
@@ -239,7 +239,6 @@ public class PlayerInteractEvent extends PlayerEvent
          * @return The ray trace result from clicking on this block.  May be null, if called from mod code using the old constructor.
          * Mods should be phasing away from the old constructor, and the nullable annotation will be removed in a future version.
          */
-        @Nullable
         public BlockRayTraceResult getRayTraceResult()
         {
             return res;
--- a/src/main/java/net/minecraftforge/event/entity/player/PlayerInteractEvent.java
+++ b/src/main/java/net/minecraftforge/event/entity/player/PlayerInteractEvent.java
@@ -27,6 +27,7 @@ import net.minecraft.util.ActionResultType;
 import net.minecraft.util.Direction;
 import net.minecraft.util.Hand;
 import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.BlockRayTraceResult;
 import net.minecraft.util.math.Vec3d;
 import net.minecraft.world.World;
 import net.minecraftforge.common.MinecraftForge;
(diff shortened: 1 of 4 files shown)


src/main/java/net/minecraftforge/event/entity/player/PlayerInteractEvent.java
```diff
@@ -198,6 +218,35 @@ public void setCanceled(boolean canceled)
...
+        public OnItemUse(PlayerEntity player, Hand hand, BlockPos pos, BlockRayTraceResult res) {
+            super(player, hand, pos, res.getFace());
+            this.res = res;
+        }
+
+        /**
+         * @return The ray trace result from clicking on this block.  May be null, if called from mod code using the old constructor.
+         * Mods should be phasing away from the old constructor, and the nullable annotation will be removed in a future version.
+         */
+        @Nullable
```
ichttt: This class is new, and you call a method on the `BlockRayTraceResult` in the constructor, so this has to be `Nonnull`

ichttt (changes requested):

src/main/java/net/minecraftforge/event/entity/player/PlayerInteractEvent.java
```diff
@@ -198,6 +218,34 @@ public void setCanceled(boolean canceled)
...
+        private final BlockRayTraceResult res;
+
+        public OnItemUse(PlayerEntity player, Hand hand, BlockPos pos, BlockRayTraceResult res) {
+            super(player, hand, pos, res.getFace());
+            this.res = res;
+        }
+
+        /**
+         * @return The ray trace result from clicking on this block.  May be null, if called from mod code using the old constructor.
+         * Mods should be phasing away from the old constructor, and the nullable annotation will be removed in a future version.
```
ichttt: Please cleanup this javadoc as well

src/main/java/net/minecraftforge/event/entity/player/PlayerInteractEvent.java
Shadows-of-Fire: Yep, that's what I get for copying the RightClickBlock class

Shadows-of-Fire:
Why'd this get moved back from assigned to triage?

DaemonUmbra:
"Split ``onItemUse`` into its own PR and explain/justify it better, what does it do that ``onRightClick`` doesn't?" is more or less what was said in the Triage channel.

Shadows-of-Fire:
Both uses of this PR are really just adding OnItemUse, the change to RightClickBlock was simply because it was thrown out for no reason in porting.  I honestly don't need that part, but I won't make a separate PR to add it.

Anyway, the explicit purpose of OnItemUse is to add new right click use on block behaviors to items, when the _item_ would normally be used, instead of when the block would be used.

At present, to emulate this behavior, I have to use RightClickBlock.  The problem with this is either I can fully emulate the block use to determine if the block has precedence or not, or force the item to always take precedence.  Neither situation is a good one.

For example, I allow a pickaxe to place a torch on right click in the presence of an NBT tag.  If i take the first scenario, then I have to copypaste a lot of vanilla code to emulate normal behavior if the pickaxe were to normally place torches on right click.  If I take the second scenario, then even if you use this pickaxe on say, a chest, you always place a torch and don't open the chest unless you can't place a torch there.  There's technically a third scenario, where you place the torch _and_ open the chest, that is equally undesirable.

Shadows-of-Fire:
Well, this has been ignored for a month now, so w/e

Issue #6300, opened 2019-11-05
Tag for vanilla ladder functionality.


ZeroLevels:
At current, basic ladders implemented through forge are missing a small piece of functionality that vanilla ladders have.  Vanilla ladders allow you to climb open trapdoors that are directly above them.

This functionality is not currently tied to a tag either through trapdoors, nor is there a tag for ladders currently.  I figured a tag for ladders could tie in this functionality.

PR #6617, opened 2020-04-12
Trapdoor ladder compat 1.15.2

stfwi:
Small patch allowing to climb open trap doors from modded ladders inheriting from `LadderBlock` (the change is only  reference to `Blocks.LADDER` --> `instanceof LadderBlock`.

Test mod is included in the PR (thanks again to @tterrag1098 for that hint), adding a simple modded ladder with the vanilla blockstate/model.

*Further references*: Open issue #6300, proposing to do this with tags. In this patch this is not considered, as tagging ladder blocks from pack configurations may lead to practical issues. E.g. the shapes of ladders have to be compatible with the trap doors, which is better explicitly (immutable) defined in the mod codes, where inheriting from `LadderBlock` is also possible.


src/test/java/net/minecraftforge/debug/block/TrapdoorTest.java
```diff
@@ -0,0 +1,80 @@
...
+@Mod(TrapdoorTest.MODID)
+public class TrapdoorTest
+{
+    public static final String MODID = "trapdoor_test";
+
+    // Represents modded ladders to be supported
+    private static class TrapdoorTestLadder extends LadderBlock
+    {
+        private TrapdoorTestLadder(Block.Properties builder)
+        { super(builder); }
```
LexManos: Your code style is horrible... It's just a test mod so it doesn't really matter, but do a lot better next time...

src/test/java/net/minecraftforge/debug/block/TrapdoorTest.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private static final Block MODDED_TEST_LADDER = (new TrapdoorTestLadder(
+      // Using different block properties as the model is the vanilla ladder model.
```
LexManos: No, static creation of registry objects is a BIG NO

src/test/resources/assets/trapdoor_test/blockstates/trapdoor_test_ladder.json
```diff
@@ -0,0 +1,8 @@
+{
```
LexManos: DataGens are preferred , even for test mods.

patches/minecraft/net/minecraft/block/TrapDoorBlock.java.patch
```diff
@@ -8,7 +8,7 @@
 +   public boolean isLadder(BlockState state, net.minecraft.world.IWorldReader world, BlockPos pos, net.minecraft.entity.LivingEntity entity) {
 +      if (state.func_177229_b(field_176283_b)) {
 +         BlockState down = world.func_180495_p(pos.func_177977_b());
-+         if (down.func_177230_c() == net.minecraft.block.Blocks.field_150468_ap)
++         if (down.func_177230_c() instanceof net.minecraft.block.LadderBlock)
```
LexManos, line 11: This by itself isn't bad. Would be nice if we could get a more detailed check using isLadder, but we don't have a directional argument for it...

stfwi: Hi lex, yes, isLadder() was my first check, this is the minimal alternative for a vanilla patch.

src/test/resources/assets/trapdoor_test/blockstates/trapdoor_test_ladder.json
stfwi: applied.

src/test/java/net/minecraftforge/debug/block/TrapdoorTest.java
stfwi: applied

src/test/java/net/minecraftforge/debug/block/TrapdoorTest.java
stfwi: applied.

stfwi:
Had to pull and merge to resolve a conflict in the mods.toml.

tterrag1098:
Your rebase is messed up, I suggest starting fresh and cherry picking your commits in.

stfwi:
Ok, no idea why the CI failed here. The changes are exactly as committed before.

gigaherz:
Not your fault. Gradle has really bad support for shared caches, and it really hates it when two PRs build at the same time. I restarted the build.

stfwi:
Ok thanks giga. Man that was a hard ride to flip a bit in vanilla ;)

LexManos:
This is pretty much ready to pull, except you still havent run the data generator. As the files would be in generated_test not test.

stfwi:
Hi lex, I moved them so they are in the actual client test. Should I leave the gen files? Respectively rebuild and push?

LexManos:
delete the jsons
Run the test data gen run config
It should regenerate the files in the correct folder.
If it does not you've either written the data gen wrong, or run the wrong task.

Leave them where they are generated because that's the point of data gens...

stfwi:
Ok done, rebuilt leaving the outfiles in the original path (`/src/generated_test/resources`). 
In a `forge_test_client` run these files are not implicitly included however, so I suppose the build args for the datagen were not yet properly settled (`args '--mod', 'trapdoor_test', '--all', '--output', rootProject.file('src/generated_test/resources/'), '--validate',
                        '--existing', sourceSets.main.resources.srcDirs[0]`) ?

stfwi:
Updated the PR to the datagen changes in 1.15.2-31.2.7, the datagen now saves the resources in the correct directory. Re-tested with a client run, still okay.

PR #6620, opened 2020-04-14
[1.15] Add Mushroom Tags

cech12:
This PR adds multiple tags for mushrooms. Mushrooms, its caps and stems can be grouped by color and functions. The "mushrooms/edible" tag is for food mods that deal with mushroom recipes. Some mods add mushooms with poisonous effect or other effects that should not be eaten.
The Stem and Cap tags are for mods (like mine) which are changing the behaviour of mushrooms. These tags should make it possible to be compatible with all mods that are adding mushrooms.


cech12:
It is still an active PR

PR #6628, opened 2020-04-16
ChunkDataEvent.Save: If chunk's world is null, use world from calling context

LordMonoxide:
If a chunk is unloaded due to a player moving away from it, the chunk save event will be posted with an instance of `ChunkPrimer`. The event uses `IChunk.getForgeWorld()`, but `ChunkPrimer` does not override `getForgeWorld`, so the default `getForgeWorld()` in the interface returns null. This means the save event's `getWorld()` method will also return null.

The method that posts the chunk save event has a world object available. This PR simply allows the chunk save event to be posted with a specific world instance, rather than always relying on `IChunk.getForgeWorld()`. If a chunk's `getForgeWorld()` does not evaluate to null, its value will still be used. If it does evaluate to null, the world object from the method will be used.

This issue was easily reproducible in all tests that I did. All you have to do is walk in a straight line, and as chunks are unloaded behind you, the event will be posted with a null world instance. Simply listen for the `ChunkDataEvent.Save` event, and log the world to the console.

I have included a test mod that shows the event is fixed. Without the fix, the console will be filled with messages like this:
```
[22:33:42] [Server thread/INFO]: Chunk at [29, 2] had null world
[22:33:43] [Server thread/INFO]: Chunk at [29, 1] had null world
[22:33:44] [Server thread/INFO]: Chunk at [29, 0] had null world
[22:33:46] [Server thread/INFO]: Chunk at [29, -1] had null world
[22:33:47] [Server thread/INFO]: Chunk at [29, -2] had null world
[22:33:49] [Server thread/INFO]: Chunk at [29, -3] had null world
[22:33:52] [Server thread/INFO]: Chunk at [29, -4] had null world
```
After the fix, no such messages are logged.

Access to the chunk's world object is important because without it, you can know its coordinates, but not which dimension it came from. This makes storing extra data about that chunk effectively impossible.


ichttt (changes requested):
Please deprecate the now unused old contructor of `ChunkDataEvent.Save`, and add a comment like `TODO remove in 1.16`

LordMonoxide:
Done

PR #6630, opened 2020-04-17
Patch EarlyProgressVisualization on mac

RogueLogix:
Moves GLFW commands to correct thread for MacOS.

glfwPollEvents cannot be called from the secondary thread on MacOS, so its not called at all after window creation. While this not preferable the window will be marked as not responding, but there is no prompt and swapping buffers still updates what is shown.

Unchanged behavior on Linux and Windows, still uses secondary thread for glfw including event polling

### DIFF
--- a/src/fmllauncher/java/net/minecraftforge/fml/loading/progress/ClientVisualization.java
+++ b/src/fmllauncher/java/net/minecraftforge/fml/loading/progress/ClientVisualization.java
@@ -21,6 +21,7 @@ package net.minecraftforge.fml.loading.progress;
 
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.LogManager;
+import org.lwjgl.glfw.GLFW;
 import org.lwjgl.glfw.GLFWErrorCallback;
 import org.lwjgl.glfw.GLFWImage;
 import org.lwjgl.glfw.GLFWVidMode;
@@ -42,7 +43,6 @@ import java.util.function.IntSupplier;
 import java.util.function.LongSupplier;
 import java.util.function.Supplier;
 
-import static org.lwjgl.glfw.Callbacks.glfwFreeCallbacks;
 import static org.lwjgl.glfw.GLFW.*;
 import static org.lwjgl.glfw.GLFW.glfwCreateWindow;
 import static org.lwjgl.opengl.GL11.*;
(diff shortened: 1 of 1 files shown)


src/fmllauncher/java/net/minecraftforge/fml/loading/progress/EarlyProgressVisualization.java
```diff
@@ -30,9 +30,8 @@
 
     public void accept(final Dist dist) {
         if (visualization != null) return;
-        // We don't show the window on Mac because mac is super mega triple shit and can't handle anything out of the ordinary
         final boolean ismac = System.getProperty("os.name").toLowerCase(Locale.ROOT).contains("mac");
-        visualization = dist.isClient() && !ismac && Boolean.parseBoolean(System.getProperty("fml.earlyprogresswindow", "true")) ? new ClientVisualization() : new NoVisualization();
+        visualization = new ClientVisualization(ismac);
```
LexManos: This still needs the disable flag, because there is some edgecases like macs in this case that have issues. So there needs to be a way to work around it.

Plus the entire point of this being off thread is to not block vanilla loading... Having the GL context be created in the main thread has no side effects to Minecraft crating its own GL context in the same thread?

RogueLogix: I removed the NoVisualization default/backup as I didn't see a point of keeping a dead and empty class, but that could be added back in if you believe it should still be there. That flag is used to switch between glfwPollEvents being called on the secondary thread or it not being called at all and on which thread the window and context are created on (as Win32 requires polling to be on the same thread that it was created on). If glfwPollEvents could be called during the loading then these two paths could be unified under the glfw's defined behavior.

There is no additional blocking going on from these commits, all that is moved to the main thread is the creation and deletion of the context, previously deletion was waited upon anyway so only creation is new work/wait time for the main thread. Actual rendering and context binding still always takes place on a second thread.

In regards to the GL context creation, tldr: no. The OpenGL specification describes behavior with multiple contexts and shared contexts, and in this case the entire context and glfw are cleaned up before Minecraft creates it's context. With GLFW it is actually undefined behavior to create an OpenGL context on any thread _but_ the main thread.

RogueLogix: oh, i see the disable flag you are talking about, i didnt even realize that was there, let me correct that.

RogueLogix: disable flag added back in.

cpw:
I don't understand this issue. The `Main` thread is being used for render tick. It wouldn't work otherwise. So, what is the problem here?

RogueLogix:
Unless I was informed by Gigaherz wrong in regards to you having issues with this working on Cocoa/MacOS, the reason it was broken is because you were using glfw commands documented as main thread only on a secondary thread. I moved those specific glfw commands to the main thread, allowing it to work on MacOS (tested with 10.15).

PR #6631, opened 2020-04-17
Fix precheck for if an ItemHandler is full

pupnewfster:
This PR fixes the precheck that is based on Vanilla's hopper implementation for if an inventory is full.

Currently the hopper's IItemHandler integration considers a slot full if it is at the stacks max size. This is a problem for any slots that want to allow past a stack's max stack size to be inserted (for example Mekanism's bins). I am changing this to checking to only consider it as full if it is at the slot's limit. This change does not break any logic due to the fact that it is a precheck, so simulation still happens to ensure the destination can actually accept a stack, but does allow for this simulation and insertion to take place if a container can accept past the max stack size in a slot.

I tested to ensure this does not break any vanilla interactions, including ensuring that it doesn't try to stack items past their max stack size and it continues to properly work as intended. I also tested and this fix does work as intended, allowing for ItemHandler implementations that allow for items stacking past their max stack size to be inserted.

### DIFF
--- a/src/main/java/net/minecraftforge/items/VanillaInventoryCodeHooks.java
+++ b/src/main/java/net/minecraftforge/items/VanillaInventoryCodeHooks.java
@@ -225,7 +225,7 @@ public class VanillaInventoryCodeHooks
         for (int slot = 0; slot < itemHandler.getSlots(); slot++)
         {
             ItemStack stackInSlot = itemHandler.getStackInSlot(slot);
-            if (stackInSlot.isEmpty() || stackInSlot.getCount() != stackInSlot.getMaxStackSize())
+            if (stackInSlot.isEmpty() || stackInSlot.getCount() < itemHandler.getSlotLimit(slot))
             {
                 return false;
             }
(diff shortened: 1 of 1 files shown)


ichttt:
I don't know this code very well, but it seems reasonable to me. Can you fix the TODO in `insertStack` as well? It seems like the cooldown reduction got lost for quite a while...

pupnewfster:
> I don't know this code very well, but it seems reasonable to me. Can you fix the TODO in `insertStack` as well? It seems like the cooldown reduction got lost for quite a while...

Probably not any time soon, due to the fact that I am not quite sure how the cooldown reduction is supposed to be implemented and for the next couple weeks I am going to be spending most of my time working on school stuff so will be a bit too busy to figure out how that works. (Ideally that bug fix won't have to wait until I have more time and am able to look into how to fix that existing TODO as well)

pupnewfster:
Got a chance to look at this some, and addressed the TODO you asked about @ichttt . I had not initialized realized how simple of a change it was to re-enable.

PR #6633, opened 2020-04-19
Fix issue with providing a null ItemStack to onPlayerDestroyItem

smith61:
This PR fixes an issue where when the last block in a stack is placed into the world the resulting PlayerDestroyItemEvent contains a null ItemStack for the original stack.

I believe the original intent was to clone the original ItemStack if the item is NOT damageable, rather than when it is damageable, then not send the event if the copy is null. This PR makes the "safer" change to just always clone in I missed a case where this would be called for damageable items.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -606,7 +606,7 @@ public class ForgeHooks
         if (!(itemstack.getItem() instanceof BucketItem)) // if not bucket
             world.captureBlockSnapshots = true;
 
-        ItemStack copy = itemstack.isDamageable() ? itemstack.copy() : null;
+        ItemStack copy = itemstack.copy();
         ActionResultType ret = itemstack.getItem().onItemUse(context);
         if (itemstack.isEmpty())
             ForgeEventFactory.onPlayerDestroyItem(player, copy, context.getHand());


pupnewfster:
Wouldn't it make more sense to just replace the null with ItemStack.EMPTY?

smith61:
From my understanding the original item is meant to be the stack that was just destroyed. So in this case it would be the stack of blocks that was just emptied. I don't see how ItemStack.EMPTY would be better if a mod was wanting to know which stack was just emptied.

LexManos:
So, this one is a bit weird, the original logic was obviously to not fire PlayerDestroyItemEvent for standard placements. As I wouldn't consider placing a block to be 'destroying' the item. Mainly thinking only damageable items can be destroyed. But, the other calls to this event don't follow that, so guess this should be fired for everything being consumed. We'll need to re-think the event, and better define it in the javadocs. This is another old event with tons of hooks for it that some don't agree with eachother.

smith61:
@LexManos In that case it might be better structured to have a "PlayerConsumeItemEvent" or "PlayerConsumerItemStackEvent" which is fired every time a player "consumes" an item from the inventory. Be it by placing the last block, eating the last item, or destroying it. Then having "PlayerDestroyItemEvent" be a subclass of that that is used whenever its the latter. That's likely a backwards compatibility nightmare though

chylex:
I hit this when porting from 1.12, I don't think back then it fired for block placements at all, so the ported event handler started crashing once it started getting nulls. Besides that, iirc the event still isn't firing when armor gets destroyed, which was also surprising as the event name suggests that it should.

If the "block consumption" is required as an event, I would split it into two - a consume stack event that fires for everything, and a destroy event that fires specifically when a durability item (held tool, shield, armor) breaks.

PR #6637, opened 2020-04-21
[1.15.x] Added 'optional' flag to ConditionalRecipe

stfwi:
This PR adds an "optional" flag to conditional recipe JSONs:

```
{
  "type": "forge:conditional",
  "optional": true,             <---------------
  "recipes": [
    ...
  ]
}
```

... where the "optinoal" key is not required and defaults to `false`.

Main purpose is to allow mod authors and packers to explicitly specify that an unmatched conditional recipe is considered and intended, supressing log warnings "Skipped ... as its serializer returned null".

To keep changes in the vanilla patch as short as possible, a scoped exception from the serializer is used to indicate an unmatched conditional recipe (the recipe is generically typed, which prevents e.g. returning a constant reference to a `ConditionalRecipe.EMPTY` or the like). A notice is still registered in the debug log file. The datagen is extended accordingly:

```
  ConditionalRecipe.builder()
    .addCondition(
      ...
    )
    .setOptional()    <---------
    .addRecipe(
      ...
    )
    .build(consumer, recipeId);
```

The module test is `net.minecraftforge.debug.crafting.ConditionalRecipeTest` (MODID `conditional_recipe_test`), for the runtime test one matching and one unmatched recipe was added in the resources.

The vanilla patch also includes the change in the related PR#6621 by Darkhax to prevent later merge conflicts. In combination with this PR, authors can choose between internal (e.g. due to feature opt-out) log discards or logging warnings for missing recipe dependencies.


src/main/java/net/minecraftforge/common/crafting/ConditionalRecipe.java
```diff
@@ -91,6 +88,11 @@ public T read(ResourceLocation recipeId, JsonObject json)
                     return (T)RecipeManager.deserializeRecipe(recipeId, JSONUtils.getJsonObject(ele.getAsJsonObject(), "recipe"));
                 idx++;
             }
+            if (JSONUtils.hasField(json, "optional"))
+                if (!JSONUtils.isBoolean(json, "optional"))
+                    throw new JsonSyntaxException("Invalid contitional recipe, the 'optional' field value must be Boolean (true/false).");
+                else if (JSONUtils.isBoolean(json, "optional"))
```
TeamSpen210, line 94: This seems either redundant or the wrong method, since it's the same as above.

stfwi: Dang it, manual merging error. This has to be `getBoolean()`.

stfwi:
@LexManos is this PR closed due to review or relevancy/wrong approach? 
(Shall I make a changed PR or reopen this, or let it be?)

PR #6484, opened 2020-01-31
[1.14] Fix IForgeBlock.removedByPlayer not firing on the client.

covers1624:
Fairly simple patch, similar to what's being done in PlayerInteractionManager.
This is needed on the client for mods that override default destroy logic, otherwise, we end up with desyncs between client and server about the state of the block in world, i.e. canceling the block break and having some other logic run.

PR #6638, opened 2020-04-22
[1.15] Fix IForgeBlock.removedByPlayer not firing on the client.

covers1624:
1.15 version of #6484

Fairly simple patch, similar to what's being done in PlayerInteractionManager.
This is needed on the client for mods that override default destroy logic, otherwise, we end up with desyncs between client and server about the state of the block in world, i.e. canceling the block break and having some other logic run.

### DIFF
--- a/patches/minecraft/net/minecraft/client/multiplayer/PlayerController.java.patch
+++ b/patches/minecraft/net/minecraft/client/multiplayer/PlayerController.java.patch
@@ -8,7 +8,7 @@
        if (this.field_78776_a.field_71439_g.func_223729_a(this.field_78776_a.field_71441_e, p_187103_1_, this.field_78779_k)) {
           return false;
        } else {
-@@ -103,7 +104,7 @@
+@@ -103,12 +104,11 @@
              Block block = blockstate.func_177230_c();
              if ((block instanceof CommandBlockBlock || block instanceof StructureBlock || block instanceof JigsawBlock) && !this.field_78776_a.field_71439_g.func_195070_dx()) {
                 return false;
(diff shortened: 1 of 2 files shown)


ichttt (changes requested):
This changes vanilla behavior, as removedByPlayer will call removeBlock, which uses different flags when calling setBlockState then the vanilla call (Vanilla uses `11` in this code path, removeBlock uses `3`)

covers1624:
Ah, I missed that. The missing bit appears to be the `RENDER_MAIN_THREAD` bit, we could either change `removedByPlayer` to do something like `return world.setBlockState(pos, fludState.getBlockState(), world.isRemote ?  3  : 11 );` Or create a separate method called client side. Not really sure which would be more acceptable.

ichttt:
I would change the removedByPlayer method, as you would otherwise have to add a new method to both blockstate and block, and I guess existing mods or mods ported from older versions expect this behavior as well

LexManos:
I am so sick of this whackamole client and server desync crap. This like the 50th PR for this.

Vanilla please fix your broken as hell user interactions.
Modders, start moving your things to server side only as they should be.

covers1624:
I agree all interactions should be server-side but in this context server-side only isn't possible due to vanilla processing the interaction on the client and the server. With this hook, changing the default behaviour of a block break _only_ on the server results in the desync because the client uses the default logic (just delete the block) and would require extra packets from the mod (or a patch) to re-update the client-side with what actually happened on the server (example, breaking a subpart of the block, or just cancelling it altogether). This patch just allows the mod to handle things on the client the same way it happens on the server, without the need for sending additional packets / visual glitches from the interaction being updated.

PR #6640, opened 2020-04-22
LTS back port for Trapdoor climbing support for modded ladders.

stfwi:
The changes and tests of this PR are identical to the 1.15.x patch in PR#6617.
- Reference check changed to type check,
- Test adds a modded ladder inherited from `LadderBlock` for verification.


LexManos:
Wait for the original to be pulled, and then cherry pick. So you don't have to sync the two branches.

PR #6645, opened 2020-04-24
Expose new constructor to allow child classes to provide TileEntityType

Yamahari:
Currently there is no constructor for LecternTileEntity that exposes a way to provide a custom TileEntityType.


ichttt:
I don't think that's the way to go, as we would have patch 17 TileEntites to make this work for all tile entites.
Maybe it would be better to create a protected method called `getTileEntityType` that defaults to null, and use this if the result is not null, and otherwise use the value supplied by the constructor parameter

Yamahari:
there's already a function `getType()`, the problem with overriding it occurs when capabilities are attached. `this.gatherCapabilities()` is called in the TileEntity constructor and if someone then calls `getType()` the child class is not constructed yet

ichttt:
Are you sure? Have you tested this behaviour? I'm pretty sure the method in the child class is called even if it hasn't been fully constructed yet...

Yamahari:
yes it's called but you can't use any fields of the class because they are not initialized

ichttt:
You can still access static fields, and you should only have one `TileEntityType` per `TileEntity` class either way, so I don't get your problem. Either way, this is not a viable solution to begin with. Feel free to ask for help in discord or the forums.

Issue #6603, opened 2020-04-04
Model "transform" block rotation misbehaving


TBoshoven:
**Minecraft Version:** 1.15.2
**Forge Version:** 1.15.2-31.1.35

**Steps to Reproduce:**
 1. Create a composite model for a block with one of the parts rotated. I used the following for testing:
```json
{
  "loader": "forge:composite",
  "parts": {
    "anvil1": {
      "parent": "minecraft:block/anvil"
    },
    "anvil2": {
      "parent": "minecraft:block/anvil",
      "transform": {
        "rotation": [{"y": 90}]
      }
    }
  }
}
```
 2. Place a block with this model in the world

**Expected behavior:**
Block looks like two anvils crossing each other.

**Actual behavior:**
One of the two anvils is displaced.

**Description of issue:**
It seems like rotation goes by corner base where in 1.14.4 (through the backported system) it rotates with a center base. This is what I would expect as the "correct" behavior.
Furthermore, it seems to be behaving inconsistently when loaded through a variant blockstate with a rotation in it as well.

Anvil configuration with no rotation in the blockstate:
![2020-04-04_12 40 41](https://user-images.githubusercontent.com/1949124/78456420-8734fc00-7671-11ea-84cf-4216556a9ab5.png)

Anvil configuration with a 90-degree rotation over the Y axis in the blockstate:
![2020-04-04_12 37 20](https://user-images.githubusercontent.com/1949124/78456359-2c030980-7671-11ea-8558-25dd0d3f45c4.png)

(ignore the stuff rendering underneath; I just threw this model on an existing block in my mod)

gigaherz:
Fixed in c23ea1e

PR #6649, opened 2020-04-25
Fix interaction of transforms with variant transforms

TBoshoven:
This relates to https://github.com/MinecraftForge/MinecraftForge/issues/6603 .

When using model transforms and variant transformations at the same time, the transformations are applied in reverse order. This PR fixes that.

**Example**:
Assume the following composite model:
```json
{
  "loader": "forge:composite",
  "parts": {
    "wideanvil": {
      "parent": "minecraft:block/anvil",
      "transform": {
        "scale": [2, 1, 1]
      }
    }
  }
}
```
This is supposed to produce a single anvil submodel with double the width (over the X axis).
When combined with a blockstate containing `"y": 90` (a 90 degree rotation over the Y axis), this produces unexpected results:
scale-without-patch

The anvil at the top of the image has the rotation in the blockstate and the one to the right does not.
I would expect both anvils to be the same, but rotated.
The reason this is not working as expected is that the order of transformations is reversed (first the variant blockstate is applied, then the submodel transformation). As a result, the block is first rotated, then stretched, leading to this effect.

This PR changes that order, leading to the correct result:
scale-with-patch

**Impact**:
This is clearly a breaking change.
However, I think it is important to make it because there is no good way to work around this issue within the system. The only true workaround I see is to make completely separate models for all rotations.

### DIFF
--- a/src/main/java/net/minecraftforge/client/model/ModelLoaderRegistry.java
+++ b/src/main/java/net/minecraftforge/client/model/ModelLoaderRegistry.java
@@ -249,7 +249,7 @@ public class ModelLoaderRegistry
         IModelGeometry<?> customModel = blockModel.customData.getCustomGeometry();
         IModelTransform customModelState = blockModel.customData.getCustomModelState();
         if (customModelState != null)
-            modelTransform = new ModelTransformComposition(customModelState, modelTransform, modelTransform.isUvLock());
+            modelTransform = new ModelTransformComposition(modelTransform, customModelState, modelTransform.isUvLock());
 
         if (customModel != null)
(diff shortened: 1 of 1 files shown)


gigaherz:
Oooh, that was the bug? 🤦‍♂️ my bad. I didn't get around to fixing it because I assumed it was something more... mathy.

TBoshoven:
There was another one ;) See #6650 

When I realized the two issues described in #6603 were separate, it became clear that the ordering was wrong. It just took me forever poking at mathy things to figure out where exactly these were composed.

gigaherz:
Merged as part of c23ea1e733

PR #6650, opened 2020-04-25
Model transforms should use center coordinates

TBoshoven:
This relates to https://github.com/MinecraftForge/MinecraftForge/issues/6603 .

Model transforms behave rather oddly. In particular, using rotation displaces blocks.

**Example**:
Assume the following composite model:
```json
{
  "loader": "forge:composite",
  "parts": {
    "rotatedanvil": {
      "parent": "minecraft:block/anvil",
      "transform": {
        "rotation": [{"y":  90}]
      }
    }
  }
}
```
This is supposed to produce a single anvil submodel with a 90 degree rotation over the Y axis.
This produces some (in my opinion) unexpected results:
rotation-without-patch
I marked the blocks straight under the models to demonstrate where I would expect the models to be.
The top anvil has an additional 90-degree rotation in its variant blockstate, see also https://github.com/MinecraftForge/MinecraftForge/pull/6649

Clearly, both anvils are displaced. It seems like they use corner coordinates for their rotation rather than center coordinates. This is not a very intuitive or practical system. It's also different from the implementation of these transformations in 1.14 .
If this is an intentional decision (I see a few reasons why corner coordinates might be preferred), it might make sense to implement the inverse of this in the 1.14 branch, to keep models portable between versions.

This PR removes that coordinate transformation, leading to the following result:
rotation-with-patch
(they look off center, but this is just the perspective)

**Impact**:
I'm not sure if or how this change should be applied, since I assume people are already using this or working around this.
There could be some kind of flag, or maybe just some documentation explaining how to properly form these transformations.
I made the PR as is to get an idea of whether this should be fixed and how/where.

### DIFF
--- a/src/main/java/net/minecraftforge/client/extensions/IForgeTransformationMatrix.java
+++ b/src/main/java/net/minecraftforge/client/extensions/IForgeTransformationMatrix.java
@@ -93,16 +93,7 @@ public interface IForgeTransformationMatrix
      */
     default TransformationMatrix blockCenterToCorner()
     {
-        TransformationMatrix transform = getTransformaion();
-        if (transform.isIdentity()) return TransformationMatrix.identity();
-
-        Matrix4f ret = transform.getMatrix();
-        Matrix4f tmp = Matrix4f.makeTranslate(.5f, .5f, .5f);
-        ret.multiplyBackward(tmp);
-        tmp.setIdentity();
-        tmp.setTranslation(-.5f, -.5f, -.5f);
-        ret.mul(tmp);
-        return new TransformationMatrix(ret);
+        return applyOrigin(new Vector3f(-.5f, -.5f, -.5f));
     }
 
     /**
@@ -110,16 +101,24 @@ public interface IForgeTransformationMatrix
      */
     default TransformationMatrix blockCornerToCenter()
     {
+        return applyOrigin(new Vector3f(.5f, .5f, .5f));
+    }
+
+    /**
+     * Apply this transformation to a different origin.
+     * Can be used for switching between coordinate systems.
+     * Parameter is relative to the current origin.
+     */
+    default TransformationMatrix applyOrigin(Vector3f origin) {
         TransformationMatrix transform = getTransformaion();
         if (transform.isIdentity()) return TransformationMatrix.identity();
 
         Matrix4f ret = transform.getMatrix();
-        Matrix4f tmp = Matrix4f.makeTranslate(-.5f, -.5f, -.5f);
+        Matrix4f tmp = Matrix4f.makeTranslate(-origin.getX(), -origin.getY(), -origin.getZ());
         ret.multiplyBackward(tmp);
         tmp.setIdentity();
-        tmp.setTranslation(.5f, .5f, .5f);
+        tmp.setTranslation(origin.getX(), origin.getY(), origin.getZ());
         ret.mul(tmp);
         return new TransformationMatrix(ret);
     }
-
 }
--- a/src/main/java/net/minecraftforge/client/extensions/IForgeTransformationMatrix.java
+++ b/src/main/java/net/minecraftforge/client/extensions/IForgeTransformationMatrix.java
@@ -89,19 +89,19 @@ public interface IForgeTransformationMatrix
     }
 
     /**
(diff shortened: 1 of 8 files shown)


gigaherz:
Ah! **This** is what I was... never getting around to fixing. For the same concern you have.
My best idea was to add a temporary (as in, until 1.16) setting in the "transform" block, like `"origin": "center" / "corner"` or similar.

TBoshoven:
I think that's not a bad idea in general actually.
There are some benefits to corner coordinates (it's what models themselves use so it's easier to position submodels I guess) but they're often quite difficult to work with for these types of transformations.

I also didn't validate *which* corner is currently being used (I assume it's the 0,0 one of the model but I didn't actually check).

tterrag1098:
I think adding the "origin" key is the best option, since it solves the problem while also being a value-add. It could even be made to accept arbitrary x/y/z coordinates.

TBoshoven:
I'd be up for implementing this `"origin"` key.
Is there a preference for the "string" version or the "coordinate" version mentioned above (I would be fine with either)?
Do we need any further discussion?

gigaherz:
I think the ideal thing would be to support both. If string, accept "corner" and "center" and use the appropriate vector, if array, accept [x,y,z] with float values, and construct the vector from it.
I'm not aware of any other concern. If you have any doubt with how to integrate this value into the model system, I'm around on discord.

TBoshoven:
Implemented the above. Still working on adding a test block for this stuff.

TBoshoven:
While writing a test model, I came across another peculiarity. It seems like the current origin coordinate is actually not the 0,0,0 corner, but the 1,1,1 corner.

Luckily, I just wrote this whole thing to choose the origin, so this should be an easy fix.

In addition to the coordinate origin, there will be *three* string ones:
* `center`
* `corner`, the `(0,0,0)` corner
* `opposing-corner` (default), the `(1,1,1)` corner

I made the last one an option so people can easily keep their models compatible if it is ever (1.16?) decided to move to a different default coordinate system.

TBoshoven:
Implemented the above and added a test mod with a model featuring various types of transformations to verify that everything works as expected.
I added rotations in the blockstate, because it clearly demonstrates why https://github.com/MinecraftForge/MinecraftForge/pull/6649 is needed.

gigaherz:
Merged as part of c23ea1e733

Angular-Angel:
Yeah I'm having this issue again, I don't know what the deal is. :/

image

Angular-Angel:
Okay I fixed it, I just needed to add 
```        "origin": "center",```
To my transform block. I'm gonna say though, that should really be the default. :/

gigaherz:
It was supposed to be. When writing the code we decided to use the same origin as vanilla, which is on a corner.... but then later when porting to 1.15 I messed up the order of transforms and ended up with the wrong corner. Problem was the format was already established when we figured out why it didn't work, and we didn't want to break existing mod jsons, so we added the "origin" option instead. Lex's opinion at the time was that we shouldn't change json file formats without having a "version" value that people can specify, so changing the transform default would mean you would have to add `"version":2` instead to get the new default.

Angular-Angel:
Ah, okay! So, if I add "version": 2 then the center will be the default? Where in the code do I need to add that? To level? :/

gigaherz:
No 🤦‍♂️ my bad wasn't clear enough. I meant that as a hypothetical.

Angular-Angel:
Ah, okay, gotcha. XD

TBoshoven:
I think it would be fine to just keep this system intact and change the default to `"center"` (from `"opposing-corner"`) in some major release.
As I mentioned above, having the option to use a corner (or any coordinate) as an original could be quite useful.
Instead of having to add a version, people would just need to add `"origin": "opposing-corner"` to their model if they want the old behavior. Same amount of effort.

Guess this could be added to the list of things to fix for 1.18? I think the ship has probably sailed for 1.17.

KnightMiner:
The point of the version is to prevent a silent breaking. If the default in the JSON changes, people might not be aware that it changed just from the log. In other words, you have to opt in to the newest system, not opt in to keeping the old one. Even for a breaking changes release having to update JSON just to keep valid JSON working is not idea.

I think making the origin required is more likely to happen than changing the default.

PR #6653, opened 2020-04-26
Validate hivepos based on blocktag instead of tileentity type

JaisDK:
It is currently not possible to make custom beehives with different tileentity types without bees losing interest in them. This change will fix that.


LexManos:
No, We're not going to add tags as logic. We've had issues with this in the past.
If you wish to expand hives, we can do it in a different way. It's worth discussing with others on discord or the forums before making a PR.

JaisDK:
@LexManos The BlockTag check is used several places already, which is why I suggested it, but this can also be done with a simple instanceof BeehiveTileEntity instead as I've seen done from another mod.
Would such a change be OK or do you suggest that there is a way to solve this without changes to MC code? The other mod implementing a change in this method, also has the same problem of bees leaving the hive.

PR #6656, opened 2020-04-26
[1.15.2] Beehive validate in livingtick

JaisDK:


JaisDK:
#6653

JaisDK:
No other input on this issue from Discord. If you know of another way to achieve this, I am all ears. The problem is that vanilla bees use this check to validate if the saves beehive pos is still a beehive, and if not it will discard it. This means that vanilla bees will never be able to stay in modded beehives using a subclass of the BeehiveTileEntity.

JaisDK:
@ichttt Will this be reviewed and merged?

JaisDK:
@LexManos Can you take a look at this again?

PR #5009, opened 2018-07-02
Add a way for attached capabilitites to be included in sync.

gigaherz:
Implemented for ItemStacks only for now. Might do the rest if people agree with this approach.

## Reasoning
Currently, synchronizing capability data of 3rdparty items with the client is extremely annoying (own item capabilitites are doable now, after #4932 was merged). It requires custom packets, which are not guaranteed to reach the client at the right time compared with vanilla interactions.

This PR adds logic to allow capabilitites to provide extra data to include in share tags, which are then deserialized on the client. The majority of the logic is agnostic to the type of object, and could be used to add sync support to TileEntity

## Notes
- I moved a couple bits of code from PacketBuffer to ItemStack, to avoid adding more accessors that are just used for that one task, but I would be happy to move it back and create those accessors, if it's wanted.
- I had to add a new function to PacketUtil, because the existing assumption wouldn't work anymore, for ItemStacks sent from the client to the server (creative menu).
- I added a new set of arrays to the CapabilityDispatcher so that capabilitites with share tag don't need to be serializable, and vice versa. If it is a concern to have so many allocations, the class could be reshuffled to use less overall arrays, at a potential cost in performance.

## TODO

- [ ] Sync share tags for TileEntities (~~should be easily doable~~ maybe not THAT easy >_<)
- [ ] Sync share tags for other things (maybe, I didn't look if it's feasible)
- [ ] Tests

PR #6667, opened 2020-04-30
Serialize capabilities when sending ItemStacks to the client

josephtarbit:
Currently, when `ItemStack`s are sent to the client (e.g. sending the contents of a `Container`) , capability NBT data is not sent which leads to strange behaviour where capability NBT data is out of sync between the client and server. This pull request resolves the issue by simply modifying the method the PacketBuffer uses to serialize and deserialize `ItemStack`s.

### DIFF
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
@@ -145,7 +145,7 @@ public interface IForgeItem
     @Nullable
     default CompoundNBT getShareTag(ItemStack stack)
     {
-        return stack.getTag();
+        return stack.serializeNBT();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


tterrag1098:
Does this not break vanilla clients?

josephtarbit:
> Does this not break vanilla clients?

What exactly do you mean by that?

tterrag1098:
When a vanilla client joins a forge server, would this not cause issues when it tries to deserialize the extra data?

gigaherz:
Although this would solve the issue, I think it ignores the fact that capability data could be rather large, and it would be extremely wasteful to send it unconditionally, since capabilities can't control the behaviour. On top of that, item implementations can override getShareTag, which would easily ignore capability data.
I wrote a PR 2 years ago, that I never got around to updating for 1.13+. 
I might be worth updating, if you are interested: #5009

josephtarbit:
> I wrote a PR 2 years ago, that I never got around to updating for 1.13+.
> I might be worth updating, if you are interested: #5009


I think that's probably best. As @tterrag1098 mentioned, this PR would break vanilla client compatibility too. I'll close this and put a bit of thought into designing a more elegant solution.

PR #6668, opened 2020-05-01
Capability share tags

josephtarbit:
This is a reimplementation of #5009 and follow up to #6667. It allows capabilities to specify a custom share tag to be sent when synchronization occurs between server and client via the newly implemented `ICapabilityShareTag`. The tag is appended to the existing vanilla share tag (`ItemStack::getTag`) so is fully compatible with vanilla clients. The majority of the logic is agnostic to the type of object, and could be used to add sync support to `TileEntity`s and various other capability providers.

### DIFF
--- a/patches/minecraft/net/minecraft/item/ItemStack.java.patch
+++ b/patches/minecraft/net/minecraft/item/ItemStack.java.patch
@@ -243,3 +243,16 @@
     public SoundEvent func_226629_F_() {
        return this.func_77973_b().func_225520_U__();
     }
+@@ -922,4 +975,12 @@
+    public SoundEvent func_226630_G_() {
+       return this.func_77973_b().func_225519_S__();
+    }
++
++   public CompoundNBT serializeShareTag() {
++      return this.serializeShareTags();
++   }
++
++   public void deserializeShareTag(CompoundNBT nbt) {
++	  this.deserializeShareTags(nbt);
++   }
+ }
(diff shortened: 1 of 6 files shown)


josephtarbit:
Anyone have any opinions on this? May of got lost in the 1.16 update efforts.

alcatrazEscapee:
This would help me a lot. As it stands, there is no effective way to ensure that item stack capabilities are synced, in the specific instance where you apply a capability to generic items. In my case, that means making all foods have expiration dates, or all items able to be heated. Both of which use capabilities to store data about the item. Both of which need to be player visible via tooltips or other aesthetic differences (and thus require syncing)

For other objects, e.g. tile entities, or entities, (generally) enough control is given to the modder to be able to sync things manually. However with capabilities, due to how vanilla will sync and overwrite item stacks via completely untraceable methods, trying to consistently ensure an item stack capability is synced involves listening to every container, on every change, duplicating existing vanilla logic for checking if an item stack should sync, and then running that on every player tick (this is from experience, there are very annoying edge cases where omitting one of the above causes it not to work).

The existing infrastructure for capability syncing on item stacks works in cases where you either control the item stack (use share tag), or control the container (use custom sync logic). When you're trying to work outside those restrictions, it's very painful.

gigaherz:
Since 1.16 is the main branch now, this would need to be updated for 1.16.
Since 1.16 isn't a recommended build, there's still a time window where breaking chances can be considered, in case it's needed to do this right (eg. defaulting to NOT synchronize capabilities).

josephtarbit:
> eg. defaulting to NOT synchronize capabilities.

Are you proposing a modification here?

josephtarbit:
Still in progress

josephtarbit:
Closing until I have the time to finish

PR #6669, opened 2020-05-01
1.15.x recipe book categories stuff

nils-prommersberger:
The Modder is now able to create his own RecipeBookCategories.
He can add them to MinecraftForge during the FMLClientSetupEvent with RecipeBookCategories#addAdditionalCategory. There he have also to add the condition. The Modder can also use that method with condition to add his own Recipes to Vanillas RecipeBookCategories when they do not fit in automatically. 
Modder is also able to use vanillas recipe unlock stuff.

PS: 
I do not know where to place my example mod, so I placed it in the test_module


patches/minecraft/net/minecraft/client/util/ClientRecipeBook.java.patch
```diff
@@ -0,0 +1,74 @@
+--- a/net/minecraft/client/util/ClientRecipeBook.java
++++ b/net/minecraft/client/util/ClientRecipeBook.java
+@@ -7,6 +7,8 @@
+ import java.util.Collections;
+ import java.util.List;
+ import java.util.Map;
++import java.util.function.Function;
```
diesieben07: No imports in patch files.

patches/minecraft/net/minecraft/client/util/ClientRecipeBook.java.patch
```diff
@@ -0,0 +1,74 @@
...
++
```
diesieben07: don't add newlines

patches/minecraft/net/minecraft/client/util/RecipeBookCategories.java.patch
```diff
@@ -0,0 +1,79 @@
...
+-   FURNACE_BLOCKS(new ItemStack(Blocks.field_150348_b)),
+-   FURNACE_MISC(new ItemStack(Items.field_151129_at), new ItemStack(Items.field_151166_bC)),
+-   BLAST_FURNACE_SEARCH(new ItemStack(Items.field_151111_aL)),
+-   BLAST_FURNACE_BLOCKS(new ItemStack(Blocks.field_150450_ax)),
+-   BLAST_FURNACE_MISC(new ItemStack(Items.field_151037_a), new ItemStack(Items.field_151149_ai)),
+-   SMOKER_SEARCH(new ItemStack(Items.field_151111_aL)),
+-   SMOKER_FOOD(new ItemStack(Items.field_151147_al)),
+-   STONECUTTER(new ItemStack(Items.field_221782_da)),
+-   CAMPFIRE(new ItemStack(Items.field_151147_al));
++public class RecipeBookCategories {
```
diesieben07: No, just no.
Use `IExtensibleEnum`.

diesieben07 (changes requested):

gigaherz (changes requested):
Besides the already mentioned issues in the code, this is missing a critical feature if mods are allowed to add new categories: an improved GUI which is capable of handling more than 5 categories without overflowing the vertical space.

For reference, I already implemented an equivalent feature back in January, which I didn't push because of the lack of an improved GUI. 

It also shows how to make use of the extensible enum system: https://github.com/MinecraftForge/MinecraftForge/compare/1.15.x...gigaherz:custom-machine-recipebook

nils-prommersberger:
@gigaherz I just want the ability for modders to add their own RecipeBookCategories. To implement these in vanilla guis or their own is left to the modders.

In the ClientRecipeBook class I changed the name of func_216767_a but I don't believe that is the right way. What would be the right way if it is not?

So I implemented IExtensibleEnum (I mean it's cool, thanks) - but how to get access to the factory in my test mod?

tterrag1098:
"compatibility is left to modders" is not good enough for forge. If this is going to be officially supported it needs to be compatible out of the box, and that includes handling more than 5 categories.

nils-prommersberger:
I think I said it wrong. 

Until now, when unlocking a recipe in Minecraft via "minecraft:unlock_recipe", the recipe is always displayed in the crafting table. First I just wanted to open the possibility to use "minecraft:unlock_recipe" and the RecipeBook, not the graphics. But if this is still not enough for you after this explanation, I will try to get an extended implementation.

nils-prommersberger:
ok it's  not ready to merge. But I would like to know if everything is good so far or if I should change something.

patches/minecraft/net/minecraft/client/gui/recipebook/RecipeBookGui.java.patch
```diff
@@ -9,7 +39,113 @@
           this.field_193018_j.add(new RecipeTabToggleWidget(recipebookcategories));
        }
  
-@@ -417,7 +417,7 @@
+@@ -189,24 +202,68 @@
+    }
+ 
+    private void func_193949_f() {
```
pupnewfster: Rather than making such a large/"messy" patch. It may make more sense to make the new logic in its own method at the bottom of the RecipeBookGui file, and then just patch calls to this method to instead call that new method.

niprow:
ups sorry
I'll take care of it

ichttt:
Screenshots would be nice.
Also, replacing the GUI isn't that good of a solution because you have to copy the vanilla file and redistribute it with minor changes, and it breaks resource packs. Maybe you can get away with using and in-code rotating the existing arrow textures

PR #6673, opened 2020-05-03
[1.14.4] Automatically remove tile entities when blockstate no longer has them

### LINKED ISSUE #6664: Blockstate-specific hasTileEntity behavior isn't clearly defined
**Minecraft Version:** 1.14.4

**Forge Version:** 1.14.4-28.2.3

**Steps to Reproduce:**
 1. Create a block with two states
 2. Define `hasTileEntity` such that it is true in one of the two cases 
 3. Have an activation function that swaps between the two states
 4. Observe that the tile entity gets created as expected for he block state that uses it, but when the block is swapped, it will exist, even though `hasTileEntity` is false.

**Description of issue:**
The behavior for when a single block switches from a state with a tile entity to one without is not clearly defined. Right now, we end up with a block in the world that has a tile entity even though it says it doesn't have one. This even causes the tile entity to stick around after the block is broken and more such shenanigans.

I came up with two possible "solutions":
* Get rid of this functionality and stick to the block-level one. I don't see a strong use case since this can be achieved by using separate blocks anyway.
* Fix this behavior by patching Vanilla code. From what I can tell, `Block#onReplaced` and/ord `Chunk#setBlockState` would have to be changed in a minor way to remove the tile entity when `hasTileEntity` flips to `false`.

TBoshoven:
Fixes #6664 
Tested using the demo mod linked in the issue.

### DIFF
--- a/patches/minecraft/net/minecraft/block/Block.java.patch
+++ b/patches/minecraft/net/minecraft/block/Block.java.patch
@@ -58,7 +58,7 @@
     @Deprecated
     public void func_196243_a(BlockState p_196243_1_, World p_196243_2_, BlockPos p_196243_3_, BlockState p_196243_4_, boolean p_196243_5_) {
 -      if (this.func_149716_u() && p_196243_1_.func_177230_c() != p_196243_4_.func_177230_c()) {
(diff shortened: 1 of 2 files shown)


TBoshoven:
Reminder for this since the 1.15 version (#6674) was merged.

PR #6675, opened 2020-05-04
1.14.x the "can spawn" hook does not fire until later in the spawn check code. 

Whums:
the "can spawn" hook does not fire until later in the spawn check code. Making it impossible to have mobs spawn on modded blocks. 

The spawning of entities is controlled by EntitySpawnPlacementRegistry and there is currently no way to override it in Forge.


ichttt:
Create an issue, not a pull request...

PR #6676, opened 2020-05-04
Improve Fabric mod error message

Prospector:
I've tried to improve this error message here because "cannot be loaded" is vague and makes it seem like an unexpected error and doesn't tell why it cannot be loaded. This should help with the confusion I think.


pau101:
This message reads as implying there can be a compatible Fabric mod and this one just happens to be be incompatible. I would think this suffices:
`File {2} cannot be loaded because it is a mod for Fabric, not Forge`

PoolloverNathan (changes requested):
I think the message should be shorter.

jamierocks (changes requested):
Should make the equivalent change for LiteLoader too.

Prospector:
Lex already vetoed this on discord anyway, but IMO this does not read like some fabric mods are compatible at all, rather that them being fabric is the cause for the incompatibility. Perhaps it could've used a comma though.

PR #6684, opened 2020-05-09
[1.15]Fixed FluidTank.drain calling onContentsChanged when simulating

ShetiPhian:
onContentsChanged is currently being called for both execute and simulate.
Having it called on simulate is odd as nothing changed, ~but just in case some mod is expecting that behavior I opted to make a new method that is aware of the FluidAction, which in turn calls the original to keep binary compatibility.~

### DIFF
--- a/src/main/java/net/minecraftforge/fluids/capability/templates/FluidTank.java
+++ b/src/main/java/net/minecraftforge/fluids/capability/templates/FluidTank.java
@@ -148,7 +148,7 @@ public class FluidTank implements IFluidHandler, IFluidTank {
         if (fluid.isEmpty())
         {
             fluid = new FluidStack(resource, Math.min(capacity, resource.getAmount()));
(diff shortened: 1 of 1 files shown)


LordMonoxide:
It's only called in both ways for drain. Fill only calls it for `EXECUTE`. The new method you added isn't getting called for `fill`/`SIMULATE`.

ShetiPhian:
Well shit; I completely missed that simulate exited early into the method.

Now any change could effect a mod.
Either simulated fills start calling it, or simulated drains stop calling it.
I guess a boolean flag could be used to stop the new method from calling the old when filling.....

I'm going to go logical, the method is onContentsChanged, simulated filling does not call it, thus simulated drain calling it must not be intended.
So I'll change this to correct that.

@ichttt  In light of my oversight the feature tag is no longer valid, it just became a bugfix

PR #6686, opened 2020-05-09
[1.15.x] Add Coal Tag

dipo33:
I'm missing Coal tag, which would group all kinds of Coal like regular coal, charcoal and other mod coals. So I think this would be a good to add.


pupnewfster:
Doesn't this already exist? `minecraft:coals`

PR #6687, opened 2020-05-09
[1.15] Extensible Raid.WaveMember and RaidEvent

seymourimadeit:
this pr makes it so the Raid.WaveMember inner class implements IExtensibleEnum so modders can have their own entities spawn in raids, and a RaidEvent class that is called when a raid starts which also allows modders to do stuff in raids.

included is a test mod that shows off these features.


src/main/java/net/minecraftforge/event/world/RaidEvent.java
```diff
@@ -0,0 +1,23 @@
+package net.minecraftforge.event.world;
+
+import net.minecraft.world.raid.Raid;
+import net.minecraftforge.eventbus.api.Cancelable;
+import net.minecraftforge.eventbus.api.Event;
+/*
+ * This event is called right before a raid starts.
+ */
```
diesieben07, line 9: 1. Please use proper Javadoc formatting.
2. This documentation is not accurate. You call this event every tick while the raid is active. It would, imho. be more logical to have `RaidEvent.Tick` (not sure if cancelling that is even useful, does that mean "skip this tick" or "stop the raid"?).

diesieben07 (changes requested):
This is missing some stuff. Like an event for when the raid begins instead of just every tick.

src/main/java/net/minecraftforge/event/world/RaidEvent.java
seymourimadeit: noted, added some sub events

ByteLaw5:
Damn I really need this one

patches/minecraft/net/minecraft/world/raid/Raid.java.patch
```diff
@@ -0,0 +1,56 @@
...
+                if (blockpos != null) {
++            	  if (net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.world.RaidEvent.RaidStart(this)));
+                   this.field_221349_m = true;
+                   this.func_221294_b(blockpos);
+                   if (!flag3) {
+@@ -462,6 +465,7 @@
+ 
+          for(int l = 0; l < j; ++l) {
+             AbstractRaiderEntity abstractraiderentity = raid$wavemember.field_221285_g.func_200721_a(this.field_221348_l);
++            if (net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.world.RaidEvent.RaidSpawnRaidersEvent(this))) return;
```
diesieben07: Why not include the entity in the event? Why is it called "SpawnRaidersEvent", when its really an event for when a single raider is spawned?
Also, there are more raiders spawned below in this method, why do they not fire the event?

src/main/java/net/minecraftforge/event/world/RaidEvent.java
```diff
@@ -0,0 +1,70 @@
+package net.minecraftforge.event.world;
+
+import net.minecraft.world.raid.Raid;
+import net.minecraftforge.eventbus.api.Cancelable;
+import net.minecraftforge.eventbus.api.Event;
+
+/**
+ *  RaidEvent along with its subevents gets fired whenever a {@link Raid} occurs.
+ */
```
diesieben07, line 9: This is still inaccurate. `RaidEvent` is never fired.

src/main/java/net/minecraftforge/event/world/RaidEvent.java
```diff
@@ -0,0 +1,70 @@
...
+@Cancelable
+public class RaidEvent extends Event
+{
+   public final Raid raid;
```
diesieben07: Why is this public?

src/main/java/net/minecraftforge/event/world/RaidEvent.java
```diff
@@ -0,0 +1,70 @@
...
+   }
+   
+   public Raid getRaid()
+   {
+	   return raid;
+   }
+   
+   /**
+    * This event is fired for every {@link Raid} tick.
+    * This event is {@link Cancelable}.
```
diesieben07, line 27: What happens if you cancel it?

diesieben07 (changes requested):

src/main/java/net/minecraftforge/event/world/RaidEvent.java
seymourimadeit: skips ticks

diesieben07: I know that. I mean you need to write that in the documentation. That's what it's there for.

patches/minecraft/net/minecraft/world/raid/Raid.java.patch
seymourimadeit: included the entity in the event, renamed the event, and made the getPotentialBonusSpawns fire that event

src/main/java/net/minecraftforge/event/world/RaidEvent.java
seymourimadeit: fixed that

src/main/java/net/minecraftforge/event/world/RaidEvent.java
seymourimadeit: clarified it

src/main/java/net/minecraftforge/event/world/RaidEvent.java
```diff
@@ -0,0 +1,81 @@
...
+   
+   public Raid getRaid()
+   {
+	   return raid;
+   }
+   
+   /**
+    * This event is fired for every {@link Raid} tick. 
+    * This event is {@link Cancelable}.
+    * If this event is cancelled, each tick will be skipped.
```
diesieben07: That is not true... If the event is cancelled _that tick_ will be skipped. Not each of them.

diesieben07 (changes requested):

solarsonata (changes requested):
I think this still needs further clarification:
`**
    * This event is fired for every {@link Raid} tick. 
    * This event is {@link Cancelable}.
    * If this event is cancelled, each tick will be skipped.
    */
   @Cancelable
   public static class Tick extends RaidEvent
   {`
Does 'Each tick will be skipped' mean only the cancelled tick, or every tick in the raid from that point onwards?

solarsonata:
> 
> 
> I think this still needs further clarification:
> `** * This event is fired for every {@link Raid} tick. * This event is {@link Cancelable}. * If this event is cancelled, each tick will be skipped. */ @Cancelable public static class Tick extends RaidEvent {`
> Does 'Each tick will be skipped' mean only the cancelled tick, or every tick in the raid from that point onwards?

_Aaaaa I forgot how to properly quote lines for changes... oh well._

diesieben07:
To me this looks good. Anyone else have comments?

seymourimadeit:
ignore that, accidentally misclicked

seymourimadeit:
should i split this pr?

patches/minecraft/net/minecraft/world/raid/Raid.java.patch
```diff
@@ -0,0 +1,72 @@
...
+       this.field_221353_q = false;
++      if (net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.world.RaidEvent.Stop(this))) return;
+       this.field_221355_s.func_201360_b();
+       this.field_221360_x = Raid.Status.STOPPED;
+    }
+ 
+    public void func_221307_p() {
+       if (!this.func_221310_d()) {
++    	 if (net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.world.RaidEvent.Tick(this))) return;
+          if (this.field_221360_x == Raid.Status.ONGOING) {
```
LexManos, line 15: These tick events are highly costly, and should be designed to not need to be called every tick. What are you expecting to happen on people's implementations?
Perhaps a Wave start event or some system of queueing things to happen in the next tick would be a better approach?

src/main/java/net/minecraftforge/event/world/RaidEvent.java
```diff
@@ -0,0 +1,81 @@
+package net.minecraftforge.event.world;
+
+import net.minecraft.entity.monster.AbstractRaiderEntity;
+import net.minecraft.world.raid.Raid;
+import net.minecraftforge.eventbus.api.*;
+
+/**
+ *  RaidEvent is the base class of all Raid events.
+ */
```
LexManos, line 9: There is no specification in any of these events for what canceling the event does. 
Is the root event cancelable? What does canceling the raid member event do? Documentation is severly lacking.

patches/minecraft/net/minecraft/world/raid/Raid.java.patch
```diff
@@ -0,0 +1,72 @@
...
++      if (net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.world.RaidEvent.RaidersSpawn(this, p_221335_1_.field_221285_g.func_200721_a(field_221348_l)))) return 0;
+       boolean flag = difficulty == Difficulty.EASY;
+       boolean flag1 = difficulty == Difficulty.NORMAL;
+       int i;
+@@ -682,7 +687,6 @@
+       default:
+          return 0;
+       }
+-
+       return i > 0 ? p_221335_2_.nextInt(i + 1) : 0;
```
LexManos, line 47: @diesieben07

patches/minecraft/net/minecraft/world/raid/Raid.java.patch
```diff
@@ -0,0 +1,72 @@
...
+          this.field_221285_g = p_i50602_3_;
+          this.field_221286_h = p_i50602_4_;
+       }
++      
++      public static WaveMember create(String name, EntityType<? extends AbstractRaiderEntity> typeIn, int[] waveCountsIn) {
++         throw new IllegalStateException("Enum not extended");
++      }
++      
++      public void init() {
++    	  field_221284_f = values();
```
LexManos, line 69: @diesieben07 spacing

src/main/java/net/minecraftforge/event/world/RaidEvent.java
```diff
@@ -0,0 +1,81 @@
...
+@Cancelable
+public class RaidEvent extends Event
+{
+   private final Raid raid;
+   
+   public RaidEvent(Raid raid)
+   {
+	  this.raid = raid;
```
LexManos, line 17: @diesieben07 This class is filled with tabs.

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -180,4 +180,5 @@ private-f net.minecraft.world.server.ChunkHolder field_219320_o # block update l
 public net.minecraft.world.server.ServerChunkProvider field_186029_c # chunkGenerator
 public net.minecraft.world.server.ServerChunkProvider field_73251_h # worldObj
 private-f net.minecraft.world.storage.loot.LootPool field_186455_c # rolls
-private-f net.minecraft.world.storage.loot.LootPool field_186456_d # bonusRolls
\ No newline at end of file
+private-f net.minecraft.world.storage.loot.LootPool field_186456_d # bonusRolls
+public-f net.minecraft.world.raid.Raid$WaveMember field_221284_f # VALUES
```
LexManos, line 184: Why is the VALUES field make publicly setable?
Are we intentionally letting random people screw with other waves entries?
Would it be better to have a add/removeMemeber function?

src/test/java/net/minecraftforge/debug/world/RaidEventTest.java
```diff
@@ -0,0 +1,69 @@
...
+@Mod("raid_event_and_more_test")
+@Mod.EventBusSubscriber
+public class RaidEventTest 
+{
+	public static final boolean ENABLE = false;
+	private static Logger LOGGER = LogManager.getLogger(RaidEventTest.class);
+	
+    public RaidEventTest()
+    {
+        if (ENABLE)
```
LexManos, line 26: @diesieben07 Again, tabs, also not that big of a deal, but simpler way would just to not use @EventBusSubscriber, but instead just register the handler gated behind a single ENABLE. Just cleaner code that way.

LexManos (changes requested):

patches/minecraft/net/minecraft/world/raid/Raid.java.patch
seymourimadeit: will address this the next time i try a raid event pr

PR #6690, opened 2020-05-11
Implement ICustomPacket directly instead of using UnsafeHacks…

TheGlitch76:
...reflection.

This pull request drastically reduces the complexity of ICustomPacket by implementing the interface directly instead of using UnsafeHacks reflection. In addition to making this code easier to understand (and, in theory maintain), this PR reduces Forge's reliance on deprecated classes and reflection tactics.

I implemented the methods in the style I have seen other parts of Forge used, but if there is a new preferred style that differs from what I have please mention it in a review.
The Fields enum was removed as all of its useful additions were package-private or private, and I assumed it was an implementation detail. If someone uses this enum in their mod, I can restore it.

### DIFF
--- a/patches/minecraft/net/minecraft/network/login/server/SCustomPayloadLoginPacket.java.patch
+++ b/patches/minecraft/net/minecraft/network/login/server/SCustomPayloadLoginPacket.java.patch
@@ -3,7 +3,7 @@
 @@ -8,7 +8,7 @@
  import net.minecraftforge.api.distmarker.Dist;
  import net.minecraftforge.api.distmarker.OnlyIn;
- 
+
 -public class SCustomPayloadLoginPacket implements IPacket<IClientLoginNetHandler> {
 +public class SCustomPayloadLoginPacket implements IPacket<IClientLoginNetHandler>, net.minecraftforge.fml.network.ICustomPacket<SCustomPayloadLoginPacket> {
     private int field_209919_a;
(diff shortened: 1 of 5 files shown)


patches/minecraft/net/minecraft/network/play/server/SCustomPayloadPlayPacket.java.patch
```diff
@@ -9,3 +9,28 @@
...
     public static final ResourceLocation field_209913_d = new ResourceLocation("debug/path");
     public static final ResourceLocation field_209914_e = new ResourceLocation("debug/neighbors_update");
+@@ -68,4 +68,24 @@
+    public PacketBuffer func_180735_b() {
+       return new PacketBuffer(this.field_149171_b.copy());
+    }
++
++	@Override
++	public PacketBuffer getInternalData() {
++		return field_149171_b;
```
LexManos: Tabs are bad

LexManos:
Yes, this could be a change, however, @cpw felt like implementing it this way. And Unsafe is a `community standard`  isn't going anywhere until they replace it with a normal API. Which doesn't seem like they are going to anytime soon.

So either way this is a rather meh PR. It has no real world value beyond more verbose, but slightly less fancy code.

TheGlitch76:
Oops, I'll fix those tabs. Yeah, cpw, if you're not interested in changes like these please tell--I have a couple other minor refactors to networking code that makes it more readable lined up, but there's no reason to PR them if they won't be merged.

patches/minecraft/net/minecraft/network/play/server/SCustomPayloadPlayPacket.java.patch
TheGlitch76: Fixed in latest commit

cpw:
I did it this way because maintaining four more patches is a pain in the arse, when we could just do it like this. It didn't break in the 1.14 or 1.15 upgrades. It'd be nice if it could completely go away but the changes needed in vanilla look unlikely to happen sadly.

patches/minecraft/net/minecraft/network/login/server/SCustomPayloadLoginPacket.java.patch
```diff
@@ -3,9 +3,44 @@
 @@ -8,7 +8,7 @@
  import net.minecraftforge.api.distmarker.Dist;
  import net.minecraftforge.api.distmarker.OnlyIn;
- 
+
```
TheGlitch76, line 6: My IDE must of had a heart attack at seeing FF's code style. I've seen it and will fix when I have a chance.

PR #6693, opened 2020-05-14
Fire the LevelChange event after having enchanted an item

Cyborgmas:
The onEnchant method is responsible for reducing the xp level after an enchant.
However, it does not call addExperienceLevel which is the method that fires the LevelChange event.
This changes one line to make it pass through this method to fire the event. A test mod also logs the event, and it does fire after enchanting.
addExperienceLevel also plays a sound, but only if the level change is positive. SInce enchantment consumes levels, this should not happen unless mods play with the amount.


src/test/resources/META-INF/mods.toml
```diff
@@ -65,7 +67,7 @@ loaderVersion="[28,)"
     modId="custom_plant_type_test"
 [[mods]]
     modId="chunk_data_event_save_null_world_test"
-[[dependencies.global_loot_test]]
+[[dependencies.level_change_event_on_enchant_test]]
     modId="forge"
```
LexManos: Why?

src/test/java/net/minecraftforge/debug/entity/player/LevelChangeEventOnEnchantTest.java
```diff
@@ -0,0 +1,21 @@
...
+@Mod("level_change_event_on_enchant_test")
+@Mod.EventBusSubscriber()
+public class LevelChangeEventOnEnchantTest {
+   private static final boolean ENABLE = true;
+   private static final Logger logger = LogManager.getLogger(LevelChangeEventOnEnchantTest.class);
+
+   @SubscribeEvent
+   public static void onPlayerLevelChange(PlayerXpEvent.LevelChange event)
+   {
+      if (!ENABLE) return;
```
pupnewfster: What is the point of this? If you want a flag like that for a one line body just use `if (ENABLED)`. Inverting it just to return seems silly.

src/test/resources/META-INF/mods.toml
Cyborgmas: did this trying to get the test mod launching, but I had used the wrong run config

src/test/java/net/minecraftforge/debug/entity/player/LevelChangeEventOnEnchantTest.java
Cyborgmas: copy pasted from another xp test mod.

src/test/resources/META-INF/mods.toml
LexManos: Then don't include it in your PR. Only include intended changes.

src/test/java/net/minecraftforge/debug/entity/player/LevelChangeEventOnEnchantTest.java
LexManos: Why not just add the event handler to that test mod? As they are both PlayerXPEvent...

Cyborgmas: ah yeah could have tested using the other mod instead huh... /oops

Cyborgmas:
should i remove my test mod then? since it is unecessary?

Cyborgmas:
Removed it, so only the patch needs to be merged. Did check with the old test mod it worked

PR #6697, opened 2020-05-15
[1.15] Remove duplicate post of ChunkEvent.Load.

covers1624:
This call to `ChunkEvent.Load` is a duplicate, it is called _directly_ after `ChunkDataEvent.Load`, and _only_ when a chunk is loaded from disk. Whereas the other call to `ChunkEvent.Load` is fired _after_ the chunk has been marked as loaded and successfully generated or loaded from disk.  
This duplicate call was also fired for ProtoChunks in the middle of generating if generation was halted at a specific step and saved then loaded from disk.  

I wrote a [small test mod](https://gist.github.com/covers1624/3c933d05a7960d59d0165ca14e02013b) (i threw it into an existing mod of mine) to track this down, I can add this to the PR if requested. The test mod is horrible, but does well enough to demonstrate the duplicated calls to the Load event. Without this change, log output from the test mod would be similar to [this](https://gist.github.com/covers1624/c92919d0bb7508b93b7e9dd90309cf8e), and when flying around near(relative to the chunk load distance around the player) edges of generated chunks [this](https://gist.github.com/covers1624/59ffd08d569135d8331d22642066fee6).

### DIFF
--- a/patches/minecraft/net/minecraft/world/server/ChunkManager.java.patch
+++ b/patches/minecraft/net/minecraft/world/server/ChunkManager.java.patch
@@ -16,15 +16,7 @@
                 }
  
                 this.func_219229_a(p_219185_5_);
-@@ -482,6 +484,7 @@
-                if (flag) {
-                   IChunk ichunk = ChunkSerializer.func_222656_a(this.field_219255_i, this.field_219269_w, this.field_219260_n, p_223172_1_, compoundnbt);
-                   ichunk.func_177432_b(this.field_219255_i.func_82737_E());
-+                  net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.world.ChunkEvent.Load(ichunk));
-                   return Either.left(ichunk);
-                }
- 
-@@ -594,6 +597,7 @@
+@@ -594,6 +596,7 @@
                 if (list != null) {
                    list.forEach(chunk::func_76622_b);
                 }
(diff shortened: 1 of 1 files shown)


LexManos:
The original intention was to give you protochunks when they were loaded and saved to disc as well. The duplicate calls without a unload in between however is an issue.
I would have to look into why chunk finalization is also firing load.

covers1624:
I see, We do however have `ChunkDataEvent.Load`(and Save), that is fired for ProtoChunks and LevelChunks. The duplicate calls without unload were due to this event location, as the call path for loading or generating a chunk through the mess of CompletableFutures, lead to that event, so that event would fire when it was _loaded_, then later once the futures were evaluated the event would fire a second time once the chunk was _acutally loaded_.

LexManos:
Ah completeable futures... got to live them.
Anyways, did more thinking and ya the data event is the important one to get protoss as we want to allow for extra data management. Are protochunks ever accessible from wolrd.getChunk?  We don't want cases where a chunk ids ticked before a load is called.

covers1624:
No ProtoChunks don't appear to be accessible through any of the world.getChunk calls, Vanilla appears to always evaluate the future down to a complete LevelChunk, but don't quote me on that one.

covers1624:
It should also be noted, that modifying the test mod as such
```java
            } else {
                LOGGER.info("LevelChunk Load: {}", pos);
                IChunk other = chunk.getWorldForge().getChunk(pos.x, pos.z);
                LOGGER.info("Yay no thread lock!");
            }
```
Causes the thread to hang indefinitely, due to the event being fired _inside_ the future..
Unfortunately, some methods on Chunk access the world and then again access the chunk, if a deferredTileEntity exists at the location of a call to `Chunk.getTileEntity`, this will cause the Chunk call `setDeferredTileEntity` and call through to `World.setTileEntity` that intern tries to access the chunk again. I tried to look for a better place to smack these events to not cause this, but the Future system there isnt that friendly. :/

LexManos:
Yes, it hanging in this event is an expected result, not much we can do about it. Modders just need to be careful. Loading is mainly for data gathering and monitoring not for major logic.

Issue #6095, opened 2019-08-29
Clicking "No" on missing registry entries screen prevents loading of other worlds


bl4ckscor3:
**Minecraft Version:** 1.15.2

**Forge Version:** 31.1.18

**Log:** https://gist.github.com/bl4ckscor3/bd30857d5082ba75eddd82160bd42ad8

**Steps to Reproduce:**
 1. Load a world which was previously loaded with mods that are no longer present
 2. Click "No" when prompted
 3. Try to load any world

**Description of issue:** When loading a world that has missing registry entries and the player chooses the "No" option, worlds can no longer be loaded and throw you back to the main menu.
Visual description: https://gfycat.com/parchedsnoopygoldfinch

cpw:
I suspect the embedded server didn't properly shutdown with no.

bl4ckscor3:
This issue is still present in Forge 31.1.18. I have updated the log accordingly.

PR #6699, opened 2020-05-15
Fix StartupQuery not working more than once on the client

diesieben07:
The `aborted` flag was used / intended to make the client return to the main menu. But it already does that when the server ends, which is what aborting the query does.
What now happened is that the client would abort the query, it would reset itself and then it would abort. This causes `abort` to stay `true`, causing further queries to always immediately "pseudo-abort", where the client thinks the query is done, but the server keeps waiting anyways.

Since the flag is not needed anyways (the clue is that the server stopped), this fixes it.


bl4ckscor3:
This fixes #6095

LexManos:
This has some threading implications that need to be fully dug into before being accepted.

PR #6701, opened 2020-05-17
[1.15] Re-implement MultiPart entities

Tarig0:
Currently I am looking to see if this implementation of a multipart entitiy that mirrors, but not replaces, how MC handles the ender dragon.

I know that these changes can co-exists with the ender dragon with no crashes but before I test with my own mod I wanted to ensure this implementation makes sense to others.


Tarig0:
I will also squish and rebase after the initial feedback

Tarig0:
Any interest/issues with this PR?

PR #6705, opened 2020-05-18
Make Material.Builder public to allow creation of custom materials

Sirttas:
It is currently impossible to create a custom `net.minecraft.block.material.Material` because a lot of methods in `net.minecraft.block.material.Material.Builder` ar private/protected.


diesieben07:
Use an AccessTransformer.

PR #6709, opened 2020-05-21
Sound.json Data Generator

Mysterious-Dev:
This PR adds a data generator provider to generate the sounds.json file more quickly.

Included is a test mod that shows off these features.


pupnewfster:
This is missing a few things such as allowing redirecting to an existing sound event instead of just using sound files. Additionally this should make use of the existing file helper to validate the sound files actually exist to help prevent typos.

An example that I wrote (but haven't had the free time to make a PR/adjust some of the methods to be better for forge's use (such as using strings instead of helper objects we have in Mekanism) can be found here:
- https://github.com/mekanism/Mekanism/blob/1.15x/src/datagen/main/java/mekanism/client/sound/BaseSoundProvider.java
- https://github.com/mekanism/Mekanism/blob/1.15x/src/datagen/main/java/mekanism/client/sound/SoundEventBuilder.java
- https://github.com/mekanism/Mekanism/blob/1.15x/src/datagen/main/java/mekanism/client/sound/SoundBuilder.java
With example usage [here](https://github.com/mekanism/Mekanism/blob/1.15x/src/datagen/main/java/mekanism/client/sound/MekanismSoundProvider.java). Feel free to reference it for figuring out how to use the existing file helper, or just porting it into your PR if you feel that it is more complete/better structured.

src/main/java/net/minecraftforge/common/data/SoundsProvider.java
```diff
@@ -0,0 +1,164 @@
...
+
+    @Override
+    public String getName() {
+        return "Sounds.json";
+    }
+
+    private void save(DirectoryCache cache, Object object, Path target) throws IOException {
+        String data = GSON.toJson(object);
+        System.out.println(data);
+        data = JavaUnicodeEscaper.outsideOf(0, 0x7f).translate(data); // Escape unicode after the fact so that it's not double escaped by GSON
```
tterrag1098: This is unnecessary.

src/main/java/net/minecraftforge/common/data/SoundsProvider.java
```diff
@@ -0,0 +1,164 @@
...
+    @Override
+    public void act(DirectoryCache cache) throws IOException {
+        addSounds();
+        if (!data.isEmpty())
+            save(cache, data, this.gen.getOutputFolder().resolve("assets/" + modid + "/sounds.json"));
+    }
+
+    @Override
+    public String getName() {
+        return "Sounds.json";
```
tterrag1098: just "Sounds" is fine.

src/main/java/net/minecraftforge/common/data/SoundsProvider.java
```diff
@@ -0,0 +1,164 @@
...
+
+    public static class SoundsFile{
+
+        public String name;
+        public Float volume;
+        public Float pitch;
+        public Integer weight;
+        public Boolean stream;
+        public Integer attenuation_distance;
+        public Boolean preload;
```
tterrag1098: Why are you using boxed types here? There's no way this serializes correctly.

tterrag1098 (changes requested):

src/main/java/net/minecraftforge/common/data/SoundsProvider.java
Mysterious-Dev: I use this so that when the person has not set one of the fields with a function it will not serialize so that it is not too heavy

tterrag1098: That would still serialize the default if it was explicitly set. Better to just check that it's not the default.

Mysterious-Dev: Yes but only worries is that for example if the player wants to set the field preload to false and I verify that this is different from false ... knowing that a boolean variable has a value by default and I cannot verify so whether it was set or not since the! = null does not work

pupnewfster: Just set them to the defaults of if they are not there?? For [example](https://github.com/mekanism/Mekanism/blob/1.15x/src/datagen/main/java/mekanism/client/sound/SoundBuilder.java#L116-L152). You also don't validate the dev is setting things to valid numbers, so I believe setting attenuation distance to negative will cause the json you create to crash the game/print out errors.

Mysterious-Dev: And if I made an adapter type to check the fields before serializing them, it would work?

src/main/java/net/minecraftforge/common/data/SoundsProvider.java
Mysterious-Dev: But what is unnecessary? the "System.out.println(data);" line ?

Mysterious-Dev:
So I want to know, is there anything else to change?

Issue #6713, opened 2020-05-23
1.15 ExtendedButton renders golden text, a 1.14 effect


Cyborgmas:
1.15.2

31.2.0

Implement a button that extends ExtendedButton (a forge impl)

While hovering over said button, the text will take a golden color. This is a nice touch from forge, however when active=false, the gold still renders. This is wrong because when the button is inactive hoevering/clicking should do nothing.

This is probably one line of code just if(!active). Will probably submit a PR but made an issue to not forget.

Cyborgmas:
Got the line even https://github.com/MinecraftForge/MinecraftForge/blob/1.15.x/src/main/java/net/minecraftforge/fml/client/gui/widget/ExtendedButton.java#L59

Cyborgmas:
Changed the issue, see my comment in the PR

PR #6714, opened 2020-05-23
Remove the yellow coloring on hover in extendedbutton

Cyborgmas:
See #6713. The test mod replaces the option button in the menu and makes it inactif. Hovering over it won't make the text change color.


Cyborgmas:
So opened up 1.14 to try and see if the bug also existed there. Found out that all vanilla buttons in 1.14 had the yellow glowing effect. Checked in vanilla to make sure:
1.15 hovering: white outline
1.14 hovering: yellow text

This is why i am changing the intent of my PR to completely remove the yellow highlighting from ExtendedButton. this will put it in line with vanilla's style

PR #6715, opened 2020-05-23
Fix some buttons not using alpha values when rendering.

Cyborgmas:
Realized this bug when replacing the button in the main menu screen with Forge's. ExtendedButton did not slowly fade in as the normal did, since the drawing it did using GUIUtils,drawContinuousTexturedBox did not account for alpha at all. 
Added an overload to GUIUtils to incorporate a custom alpha value. Changed all of the widgets using the function that ignored alpha to use alpha instead.

Also, the 2 ImageButton in the main menu (Language and Accessibility) did not fade in. To correct that, added an alpha blending. This is completely controllable to modders since it relies on the GUI's current alpha value.


Cyborgmas:
the commits are a mess sorry
2 things; not a rendering expert but I think the rendersystem calls i did are fine, not 100% sure.
Also might delete the test mod, since its almost identical to the one in #6714 , its purpose is to show that the replaced extended button for the Options button does fade in

Cyborgmas:
Resolved the conflict that occured because of the merge of #6714

PR #6716, opened 2020-05-25
IBlockState -> BlockState in Javadocs, plus minor improvements/update…

xbony2:
…s/fixes to related Javadocs

While I was updating my mod, I saw IBlockState referenced in a Javadoc, so I became pretty confused. This pull request will update some of the documentation.

A few things I ran into while making this PR that aren't resolved:
* The main Javadoc for SleepingLocationCheckEvent appears to be outdated, since there is no isBed method (and no very obvious replacement for it)
* func_226942_a_ is the current function name of "generateTree" under the current MCP mappings for whatever reason, unless my mappings are broken which is possible.
* The JavaDoc for IForgeBlock#getRayTraceResult appears to be outdated, since there is no collisionRayTrace method (and no very obvious replacement for it)


xbony2:
Sad!

PR #6717, opened 2020-05-25
Using DeferredRegister with custom registries

Cyborgmas:
I do not expect this to be PR'd, I open this mostly for comments and to maybe have a place to record anything such a feature must and can't have.

At first I created a separate class (DeferredCustomRegistry), but now implemented the solution into the DeferredRegister class directly.

Unsure if everything I used is necessary or if there is a simpler way instead of LazyOptional. 
I got it to work at first using LazyOptional<RegistryObject<T>> but now put a wrapper around it to simplify access to the T value and make it a supplier of T as well. 
But there could be something else to use here.

Any comments appreciated

### DIFF
--- a/src/main/java/net/minecraftforge/registries/DeferredRegister.java
+++ b/src/main/java/net/minecraftforge/registries/DeferredRegister.java
@@ -20,11 +20,16 @@
 package net.minecraftforge.registries;
 
 import net.minecraft.util.ResourceLocation;
+import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.common.util.LazyOptional;
 import net.minecraftforge.common.util.NonNullSupplier;
 import net.minecraftforge.event.RegistryEvent;
 import net.minecraftforge.eventbus.api.IEventBus;
 import net.minecraftforge.fml.RegistryObject;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.Marker;
+import org.apache.logging.log4j.MarkerManager;
 
 import java.util.Collection;
 import java.util.Collections;
@@ -59,6 +64,8 @@ import java.util.stream.Collectors;
  */
 public class DeferredRegister<T extends IForgeRegistryEntry<T>>
 {
+    private static final Logger LOGGER = LogManager.getLogger();
+    private static final Marker MARKER = MarkerManager.getMarker("DeferredRegister");
     private IForgeRegistry<T> type;
     private Supplier<IForgeRegistry<T>> typeSup;
     private final String modid;
@@ -100,7 +107,16 @@ public class DeferredRegister<T extends IForgeRegistryEntry<T>>
         Objects.requireNonNull(name);
         Objects.requireNonNull(sup);
         final ResourceLocation key = new ResourceLocation(modid, name);
-        RegistryObject<I> ret = RegistryObject.of(key, this.type);
+        RegistryObject<I> ret;
+        try
+        {
+            ret = RegistryObject.of(key, this.type);
+        }
+        catch (IllegalArgumentException e)
+        {
+            LOGGER.error("Errored while trying to register an entry. This likely happened due to registering for a custom registry, use lazyRegister instead.");
+            throw new IllegalArgumentException("Null registry used in DeferredRegister");
+        }
         if (entries.putIfAbsent((RegistryObject<T>) ret, () -> sup.get().setRegistryName(key)) != null) {
             throw new IllegalArgumentException("Duplicate registration " + name);
         }
@@ -121,7 +137,8 @@ public class DeferredRegister<T extends IForgeRegistryEntry<T>>
         Objects.requireNonNull(sup);
         final ResourceLocation key = new ResourceLocation(modid, name);
         LazyRegistryObject<I> lazyRet = new LazyRegistryObject<>(()->RegistryObject.of(key, this.type));
-        if (lazyEntries.putIfAbsent(lazyRet.cast(), () -> sup.get().setRegistryName(key)) != null) {
+        if (lazyEntries.putIfAbsent(lazyRet.cast(), () -> sup.get().setRegistryName(key)) != null)
+        {
             throw new IllegalArgumentException("Duplicate registration " + name);
         }
         return lazyRet;
--- a/src/main/java/net/minecraftforge/registries/DeferredRegister.java
+++ b/src/main/java/net/minecraftforge/registries/DeferredRegister.java
@@ -108,7 +108,7 @@ public class DeferredRegister<T extends IForgeRegistryEntry<T>>
     }
 
     /**
-     * Adds a new supplier to the list of entries to be registered, and returns a {@link DeferredCustomRegistry.LazyRegistryObject}
+     * Adds a new supplier to the list of entries to be registered, and returns a {@link LazyRegistryObject}
      * that will be populated with the created entry automatically and lazily.
      *
      * @param name The new entry's name, it will automatically have the modid prefixed.
--- a/src/main/java/net/minecraftforge/registries/DeferredRegister.java
+++ b/src/main/java/net/minecraftforge/registries/DeferredRegister.java
@@ -85,8 +85,12 @@ public class DeferredRegister<T extends IForgeRegistryEntry<T>>
     /**
      * Custom registries are only created after the {@link DeferredRegister} constructor is called due to static init.
      * To solve this, the {@link IForgeRegistry} and {@link RegistryObject}s must be evaluated only when registry events are fired.
-     * The passed in supplier will only be called once {@link #addEntries} is invoked.
-     * When it is invoked, it is guaranteed that the registry is created since the {@link RegistryEvent.NewRegistry} event is fired beforehand.
+     * The passed in supplier will be called automatically when {@link #addEntries} is invoked.
+     * When that happens, it is guaranteed that the registry is created since the {@link RegistryEvent.NewRegistry} event is fired beforehand.
+     *
+     * However, this can mess up if modders call {@link #getEntries()} too quickly. It can be called only after the NewRegistry has fired.
+     * Calling it then won't crash, but will result in empty RegistryObject that will update when the appropriate RegistryEvent is fired.
+     *
      * @param supReg A nonnull supplier of an IForgeRegistry of the desired type.
      */
     public DeferredRegister(NonNullSupplier<IForgeRegistry<T>> supReg, String modid)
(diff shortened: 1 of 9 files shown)


src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -56,17 +64,36 @@
...
     }
 
+    /**
+     * Custom registries are only created after the {@link DeferredRegister} constructor is called due to static init.
+     * To solve this, the {@link IForgeRegistry} and {@link RegistryObject}s must be evaluated only when registry events are fired.
+     * The passed in supplier will only be called once {@link #addEntries} is invoked.
+     * When it is invoked, it is guaranteed that the registry is created since the {@link RegistryEvent.NewRegistry} event is fired beforehand.
+     * @param supReg supplier of an IForgeRegistry of the desired type.
+     */
+    public DeferredRegister(Supplier<IForgeRegistry<T>> supReg, String modid)
```
pupnewfster: I am not quite sure given looking at usages it feels like this should use forge's `NonNullSupplier`, as the one place it is used `getFromSupplier` is assumed to be nonnull due to the call point being `this.getFromSupplier().getRegistrySuperType()`. Though maybe this is right just being Supplier given strictly speaking with custom registries the supplier would be supplying a null if called immediately.

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -80,13 +107,43 @@ public DeferredRegister(IForgeRegistry<T> reg, String modid)
...
+        RegistryObject<I> ret;
+        try
+        {
+            ret = RegistryObject.of(key, this.type);
+        }
+        catch (IllegalArgumentException e)
+        {
+            LOGGER.error("Errored while trying to register an entry. This likely happened due to registering for a custom registry, use lazyRegister instead.");
+            throw new IllegalArgumentException("Null registry used in DeferredRegister");
+        }
```
pupnewfster: Instead of using a try/catch here it may make more sense to check if `this.type` is `null` directly.

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -80,13 +107,43 @@ public DeferredRegister(IForgeRegistry<T> reg, String modid)
...
+     * @param name The new entry's name, it will automatically have the modid prefixed.
+     * @param sup A factory for the new entry, it should return a new instance every time it is called.
+     * @return A {@link LazyRegistryObject} that will be lazily updated with when the entries in the registry change.
+     */
+    public <I extends T> LazyRegistryObject<I> lazyRegister(final String name, final Supplier<? extends I> sup)
+    {
+        Objects.requireNonNull(name);
+        Objects.requireNonNull(sup);
+        final ResourceLocation key = new ResourceLocation(modid, name);
+        LazyRegistryObject<I> lazyRet = new LazyRegistryObject<>(()->RegistryObject.of(key, this.type));
```
pupnewfster: Instead of giving it `this.type` in the supplier for the lazy registry object, shouldn't it be `getFromSupplier()` to have slightly less reliance on making sure that `getFromSupplier()` was called before the supplier is ran. I believe this may be the case if `getEntries()` is called after the new registry event but before the entries are actually added in `addEntries` (Not that I can think of a case for why someone would call that in that stage)

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -99,24 +156,111 @@ public void register(IEventBus bus)
...
+     * Since the view is unmodifiable, can't cache it really. Gotta stream and collect each time
+     * This should not be a costly operation however, since the {@link RegistryObject} is cached.
+     * The modders can also cache the entries themselves.
      * @return The unmodifiable view of registered entries. Useful for bulk operations on all values.
      */
     public Collection<RegistryObject<T>> getEntries()
     {
+        if(entriesView.isEmpty())
+        {
+            return lazyEntriesView.stream().map(LazyRegistryObject::getRegistryObject).collect(Collectors.toSet());
```
pupnewfster: Given there is no caching going on and `lazyEntriesView` is not exposed externally anyways would it make more sense to remove `lazyEntriesView` all together and just do `lazyEntries.keySet()` (with the remainder of the line being as it currently is).

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -99,24 +156,111 @@ public void register(IEventBus bus)
...
     }
+
+    /**
+     * If the IForgeRegistry is null, gets it from the supplier. This is sure to be called from a classic register event.
+     * It is even guaranteed to be called from the Register Block event, since it is the first one called.
+     * This also makes sure that the IForgeRegistry is correct when updating the RegistryObjects.
+     *
+     * This is only useful for custom registries, when the {@link IForgeRegistry} does not exist when static init happens
+     */
+    private IForgeRegistry<T> getFromSupplier(){
```
pupnewfster: I feel like this should have a better name. Not sure which of these three are best but they are the ones that make the most sense to me:
- `getRegistryType`
- `getRegistry`
- `getType`

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -99,24 +156,111 @@ public void register(IEventBus bus)
...
+    }
+
+    /**
+     * A way of lazily evaluating registry objects but in a nicer way than using {@link LazyOptional<RegistryObject>}
+     *
+     * Also a supplier, to mimic {@link RegistryObject} as best as possible
+     */
+    public static class LazyRegistryObject<T extends IForgeRegistryEntry<? super T>> implements Supplier<T>
+    {
+        private final LazyOptional<RegistryObject<T>> lazyRet;
```
pupnewfster: Instead of using `LazyOptional` which I feel like is being misused here, this should use [`net.minecraftforge.common.util.NonNullLazy`](https://github.com/MinecraftForge/MinecraftForge/blob/1.15.x/src/main/java/net/minecraftforge/common/util/NonNullLazy.java).

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -99,24 +156,111 @@ public void register(IEventBus bus)
...
+        }
+        return type;
+    }
+
+    /**
+     * A way of lazily evaluating registry objects but in a nicer way than using {@link LazyOptional<RegistryObject>}
+     *
+     * Also a supplier, to mimic {@link RegistryObject} as best as possible
+     */
+    public static class LazyRegistryObject<T extends IForgeRegistryEntry<? super T>> implements Supplier<T>
```
pupnewfster: This potentially should be implementing [`net.minecraftforge.common.util.NonNullSupplier`](https://github.com/MinecraftForge/MinecraftForge/blob/1.15.x/src/main/java/net/minecraftforge/common/util/NonNullSupplier.java) as `RegistryObject#get` is Nonnull.

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -99,24 +156,111 @@ public void register(IEventBus bus)
...
+        /**
+         * Helper function to cast, since it is needed in {@link #lazyRegister} to work with
+         * objects not directly implementing the type of this registry
+         * But now that i think about it, pretty useless...
+         */
+        @SuppressWarnings("unchecked")
+        private  <X extends IForgeRegistryEntry<? super X>> LazyRegistryObject<X> cast()
+        {
+            return (LazyRegistryObject<X>) this;
+        }
```
pupnewfster: I feel like this would be cleaner doing it like how forge does it for the registry object and just inlining it `lazyRegister`.

pupnewfster (commented):
I added a few comments throughout the modified deferred register implementation; I didn't bother looking at the `DeferredCustomRegistry` implementation given as I understand from your comment it is relatively similar.

One thing that may be worth creating a test for (though I am not sure how easy it would be to do with how forge's test mods are setup) would be to make a second test mod that acts as an "addon" to the first test mod you have and make sure that it is properly able to use the first test mod's custom registry in the deferred register system.

src/main/java/net/minecraftforge/registries/DeferredRegister.java
Cyborgmas: indeed woops

src/main/java/net/minecraftforge/registries/DeferredRegister.java
Cyborgmas: I can test that. I put this.type directly to be sure to get an error since that would mean something went wrong compared to the intention

src/main/java/net/minecraftforge/registries/DeferredRegister.java
Cyborgmas: Seems indeed better. hesitated to use Lazy directly for some reason, will use forges versions

src/main/java/net/minecraftforge/registries/DeferredRegister.java
Cyborgmas: chose getType for now

src/main/java/net/minecraftforge/registries/DeferredRegister.java
Cyborgmas: Might be good yeah, though the goal is to have the set be unmodifiable so I added a Collectors.unmodifiableSet() to it
probably are better ways of doing this

src/main/java/net/minecraftforge/registries/DeferredRegister.java
Cyborgmas: so yeah... it indeeds prevent a crash if you getEntries() in between the NewRegistry Event and the the first registry event. so changed it. but would be a weird occurence since you have a list of registry objects that are effectively empty

src/main/java/net/minecraftforge/registries/DeferredRegister.java
pupnewfster: Wouldn't the unmodifiable part need to be done at the end of collecting them in a new set, and not the set you are streaming over?

Cyborgmas: This is what i meant
`Collections.unmodifiableSet(lazyEntries.keySet().stream().map(LazyRegistryObject::getRegistryObject).collect(Collectors.toSet()))`
so yeah i do it at the end

Cyborgmas:
So you did well not checking DeferredCustomRegistry. 

Did the addon mod, was quick. No problem arose, worked like a charm!

Cyborgmas:
I tried D:

PR #6722, opened 2020-05-26
Provide matrix and buffers to MapDecoration.render

williewillus:
As title. Current method is useless because we're in batched mode but the matrix and buffers aren't passed.

### DIFF
--- a/patches/minecraft/net/minecraft/world/storage/MapDecoration.java.patch
+++ b/patches/minecraft/net/minecraft/world/storage/MapDecoration.java.patch
@@ -4,7 +4,7 @@
        return i;
     }
  
-+   @Deprecated // Use overload with matrix stack and buffers. TODO 1.15 remove this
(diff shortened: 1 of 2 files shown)


patches/minecraft/net/minecraft/world/storage/MapDecoration.java.patch
```diff
@@ -1,17 +1,23 @@
 --- a/net/minecraft/world/storage/MapDecoration.java
 +++ b/net/minecraft/world/storage/MapDecoration.java
-@@ -83,6 +83,16 @@
+@@ -83,6 +83,22 @@
        return i;
     }
  
++   @Deprecated // Use overload with matrix stack and buffers. TODO 1.15 remove this
```
pupnewfster: Is this TODO supposed to say 1.16 and not 1.15?

williewillus: yes, thanks

PR #6725, opened 2020-05-27
Delete unused events

Cyborgmas:
Unsure if these need/should be fired. Mostly want to point out that these events are not fired by forge.

From all the events, for me only the NameFormat could be relevant.
CanUpdate said it fired on entity creation. Now the only reference seems to be in Entity#read
For GetCollisions, could not find a method with that intent in World nor Client/Server World . 
The ActionPerformed event seems useless. I believe the MouseInput events are more precise versions of that event.

There is also the FMLFingerprintViolationEvent that is not fired, but I think it is out of my league. (did not include it)


pupnewfster:
Even though these events aren't fired I am pretty sure this still counts as a binary breaking change, so the proper thing to do here may just be to add TODOs for removal for 1.16, though I will let someone on the forge team verify/deny that.

3TUSK:
For what's worth, the patch of firing `NameFormat` was lost during 1.13 update. See #757 for details about the original PR and rationale.  
Possible use case of `NameFormat` event is a simple `/nick` command that allow you change your nickname in a specific server (think Discord).

Cyborgmas:
If any event have a usecase, ill remove the deletion and possibly implement them. 
Pretty sure deleting events will have to wait for 1.16, this PR could be marked as 1.16 todo, since the work has already been done :D

ichttt:
Yeah, it would be better if these events could be ported to 1.15 if they are still applicable instead of deleting them... And yes, this is a binary breaking change

Cyborgmas:
Displayname was using getName() in and not caring about the player entity's custom name. Made it instead check for a customname first before using getName() directly which looks up the gameProfile name. Also prefixes and suffixes are a thing so that should be used first.
problem with customname is that its last one wins since there can only be one name.
Which might be the reason why it wasnt used?

Cyborgmas:
Started this while doing the javadocs - makes no sense to keep it here on 1.15 so i'll reopen a similar PR for 1.16

PR #6378, opened 2019-12-19
[1.15] Adding Categories within JigsawPattern and option to add/remove buildings

Cheaterpaul:
(while there is no option to add buildings to a village)

2 events are fired:

- JigsawPatternInitEvent - fired in constructor of JigsawPattern. There you can add/remove JigsawPieces

- StructureJigsawPoolInitEvent - fired when a jigsaw based structure inititizes its JigsawPatterns. There you can add your own JigsawPattern before the structure need them.

Problem:
when adding more houses to a village chances of spawning e.g. a farm are shrinking.

Resolved by adding JigsawCategories:
the categories preserves the weight of the type of JigsawPieces inside it.

The JigsawPatternInitEvent first creates JigsawCategories from the JigsawPattern's weighted JigsawPiece based on the pieces registryname and groups the pieces inside the category.

Now when subscribed to the JigsawPatternInitEvent you can add/remove JigsawPieced based on there registry from a category or add/remove whole categories from the JigsawPattern. The changes are cached before there are applied all together to the pre created JigsawCategories of the JigsawPattern.

optionally i added a registrylike console output for all changes (JigsawPatternInitEvent#printChanges())

old PR's #6324 (same but for 1.14), #6142 (same without the whole categories for 1.14)

PR #6732, opened 2020-05-29
1.15 Create a JigsawPiece event

Cyborgmas:
Add an event which fires when a Jigsaw Block is looking for all possible jigsaw connections from a pool.

This means that when generating plain villages (which are the biggest) there is about 500 to 600 event fired.

The event contains the list of jigsawpieces which can be modified in any way. 
It also contains 3 fields to be able to pinpoint which jigsaw pieces to replace.

There is also a test mod which replace the iron golem spawn with a wither one. This is only in plains biome, so to test use buffet generation. 

This however does not allow for putting custom JigsawPatterns in the village pools.
However, it is (should be... did not check) still possible to use custom pools by using a JigsawPiece containing a jigsaw block containing a custom pool. that pool having been registered to the JigsawPatternRegistry in the JigsawManager


Cyborgmas:
i was told this PR exists also https://github.com/MinecraftForge/MinecraftForge/pull/6378 
But i was already in too deep, and hopefully this solution is a bit better

GotoLink:
I don't speak for the Forge team here, but i think the javadoc goes into "code guide" territory.
I would stick to the facts:
-when the event is fired (and possibly a simple use-case)
-what the fields mean (and possibly where to use them).
From a coding-style point of view, i dislike the enum here, you just use it as string array..so why no string array ?

Cyborgmas:
Yeah you are right, went a little overboard with the javadocs I will admit that. 
What I propose is to leave them as is for now so that anyone reviewing this PR has access to that information. When it comes time to PR ill make it more concise and submit everything to the Forge Docs.

I like enums because they are flashy and serve as a code reference. You can see all the types just by typing its name instead of the array list which you have to check inside the event and the strings.
Still see why it is pretty useless to make it as an enum, so I made it an immutable list. 

Also corrected some things. for instance some jigsaw patterns can be empty, so added that.
JigsawPatterns can also be invalid, however that is checked before the event so can never happen.

Cyborgmas:
Removed the javadoc and added a PR to the documentation, ofc for now the link in the javadocs is broken.

filips123:
I have a use-case for such event (but with some modifications) for my custom structure for my mod. I posted more details to [forums](https://forums.minecraftforge.net/topic/87580-1161-partly-changing-jigsawmanagers-functionality/).

src/main/java/net/minecraftforge/event/world/JigsawPieceEvent.java
```diff
@@ -0,0 +1,69 @@
...
+    */
+   public JigsawPieceEvent(List<JigsawPiece> pieces,  int currentDepth, ResourceLocation patternLocation)
+   {
+      this.pieces = pieces;
+      this.currentDepth = currentDepth;
+      String pool = patternLocation.getPath();
+      pool = pool.replace("village/", "");
+      this.generalType = fromString(pool);
+      pool = pool.replace(this.generalType+"/", "");
+      this.specificType = pool.replace("zombie/", "");;
```
Cheaterpaul, line 54: With a custom jigsawbased structure that adds a patternlocation that contain `village` or `zombie`. `specificType` will be filled with a useless string. (additionally modid is missing)

src/main/java/net/minecraftforge/event/world/JigsawPieceEvent.java
```diff
@@ -0,0 +1,69 @@
...
+{
+   //These can be looked at under data/minecraft/structures/village (or /pillager_outpost). All correspond to the folder name except
+   // "empty", which indicates an empty pattern, and "modded", which indicates the path was not found.
+   public static final List<String> STRUCTURE_TYPES = ImmutableList.of("plains", "desert", "snowy", "taiga", "savanna", "decays", "common", "pillager_outpost", "empty", "modded");
+
+   public List<JigsawPiece> pieces;
+   public final String generalType;
+   public final String specificType;
+   public final boolean isZombie;
+   public final int currentDepth;
```
Cheaterpaul, line 35: There should be a field that holds the whole patternLocation. And then `generalType` and `specificType` are not really needed.

Cheaterpaul (changes requested):
At whole this seem like a good approach. But this does not work for non village & pillage structures.

Not sure if that are to much events for only one village at world gen, because the result for throwing the event for the same patternlocation is always the same (except `currentDepth`, but i dont know how it should help).

filips123:
I tried to port this PR to 1.16, make it more friendly for modded structures and implement some of the things I mentioned in my forum topic. I think it works but I still need to modify and test it a bit. Is it ok if I open a PR for that in a few days?

Cyborgmas:
Ah dang, dont know how i missed the updates from cheaterpaul - will address them when i have the time (not this week prob). you can if you want but i was going to port this eventually fyi

Cyborgmas:
#6939 should be backported instead, no sense making 2 separate PRS

Issue #6726, opened 2020-05-27
Forge Dynamic Bucket model generates inner quads upside down


KnightMiner:
**Minecraft Version:** 1.15.2

**Forge Version:** 31.2.5

**Steps to Reproduce:**

The model JSON below will reproduce this bug. Placing the bucket in an item frame will make the issue easier to visualize.

```json
{
  "loader": "forge:bucket",
  "fluid": "minecraft:empty",
  "textures": {
    "particle": "item/bucket",
    "base": "item/bucket"
  }
}
```

**Description of issue:**

The Forge dynamic bucket model generates the inner quads of the model using the opposite Y coordinate, causing gaps to appear in the model.
[This screenshot](https://user-images.githubusercontent.com/12155288/82991664-a42be280-9fc3-11ea-850c-ae19f57f6d19.png) demonstrates the issue well.

**Possible solution:**

Replacing [this line](https://github.com/MinecraftForge/MinecraftForge/blob/1.15.x/src/main/java/net/minecraftforge/client/model/ItemLayerModel.java#L127) with 
```java
int alpha = sprite.getPixelRGBA(f, u, vMax - v - 1) >> 24 & 0xFF;
```
appears to fix the issue, as shown in [this screenshot](https://user-images.githubusercontent.com/12155288/82991805-d9d0cb80-9fc3-11ea-9ee7-9840eb8d9254.png). However, I would like some feedback from @gigaherz as he has more experience with this area of the code (and will have a better idea if that is the right place to patch). I am willing to make a pull request if that fix is considered correct.

gigaherz:
Fixed by 28f7b5d

PR #6733, opened 2020-05-31
Fix incorrect generation of render coords and texture UVs for side BakedQuads created by ItemLayerModel

iChun:
The side baked quads generated by getQuadsForSprite had the wrong texture coords and UVs causing the sides to render improperly like so:

image
image
image
image

These changes fixes that:
image
image
image
image

### DIFF
--- a/src/main/java/net/minecraftforge/client/model/ItemLayerModel.java
+++ b/src/main/java/net/minecraftforge/client/model/ItemLayerModel.java
@@ -326,7 +326,7 @@ public final class ItemLayerModel implements IModelGeometry<ItemLayerModel>
         int height = sprite.getHeight();
 
         float x0 = (float) u / width;
-        float y0 = (float) v / height;
+        float y0 = 1f - (float) v / height;
         float x1 = x0, y1 = y0;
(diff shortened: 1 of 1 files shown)


KnightMiner:
Just want to crosslink #6726. I only reference the dynamic bucket model there as no other loader uses item layer yet. This fix is probably better than mine because I just was looking for the minimal change to get it working.

iChun:
@KnightMiner your fix seemed to work just as well as mine, really. The issue seems to stem from a flip of the rendering on the y axis somewhere around MC 1.14. You flipped the pixel detection to fix it, and I flipped the texture/uv coords. I like yours since it looks much cleaner though, I'll let @gigaherz decide on what to do next.

gigaherz:
After a quick look, it seems to me like the two fixes are equivalent. iChun's change achieves the same exact purpose, by flipping the input coords, but also flipping the output UVs, which has the indirect consequence of flipping the color query.
So I think considering it's the same effective change, but KnightMiner's is less involved, I'm leaning toward doing that change.
Since I'm working on another change to the layer models right now, I'll include the fix and credit you both on that commit.

gigaherz:
Fixed by 28f7b5d88

Issue #6734, opened 2020-06-01
Advancement data provider not reusable by modders


percivalalb:
**Minecraft Version:** 1.15.2

**Forge Version:** 31.2.5

**Description of issue:**
The the data provider `net.minecraft.data.AdvancementProvider` class is hard coded to generate vanilla advancements. Also `net.minecraft.advancements.Advancement.Builder` could be made to be more usable for modders.

**Possible solutions:**
Alter advancement data provider to accept custom advancement generators.
Make a `DisplayInfo` builder, and add methods to `Advancement.Builder` that take Supplier and ResourceLocation params.

PR #6735, opened 2020-06-01
[WIP] Make Advancement data generators reusable by modders

percivalalb:
See #6734

### DIFF
--- a/patches/minecraft/net/minecraft/data/AdvancementProvider.java.patch
+++ b/patches/minecraft/net/minecraft/data/AdvancementProvider.java.patch
@@ -1,0 +1,20 @@
+--- a/net/minecraft/data/AdvancementProvider.java
++++ b/net/minecraft/data/AdvancementProvider.java
+@@ -47,12 +47,16 @@
+          }
+       };
+ 
+-      for(Consumer<Consumer<Advancement>> consumer1 : this.field_204283_d) {
++      for(Consumer<Consumer<Advancement>> consumer1 : this.getAdvancements()) {
+          consumer1.accept(consumer);
+       }
+ 
+    }
+ 
++   protected List<Consumer<Consumer<Advancement>>> getAdvancements() {
++      return field_204283_d;
++   }
++
+    private static Path func_218428_a(Path p_218428_0_, Advancement p_218428_1_) {
+       return p_218428_0_.resolve("data/" + p_218428_1_.func_192067_g().func_110624_b() + "/advancements/" + p_218428_1_.func_192067_g().func_110623_a() + ".json");
+    }


Cyborgmas (commented):
the patch size is 4 lines long, you can reduce it to 3 at least. 
even 1 but i think that is a bit too much

Edit - mb its 6 lines long, but meant the method you added has a new line, which you can remove

DaemonUmbra:
I may be misunderstanding, are you planning to add your own advancements to the vanilla advancement provider rather than using the GatherDataEvent?

percivalalb:
@Cyborgmas I get that reducing patch size is desirable, but the point is to reduce random blocks of code throughout the file so it is easier to maintain when an update is required. So in this case I don't think it matters. 3484200f8038cbb2e3c548bed31b163e65ca8fd4 is a similar example which has a new line.

@DaemonUmbra The point is to make adding advancements for mods easier, currently you need to create you own data provider to be called in the `GatherDataEvent` as the vanilla one is hardcoded to only generate it's advancements. The idea is you extend `AdvancementProvider` and override getAdvancements so you can generate your own advancements.

DaemonUmbra:
~~Took a closer look and it does look like something needs to be done, however I don't think extending AdvancementProvider is the answer, I think a utility with static helper methods would be more suitable~~

percivalalb:
As in the end user overrides AdvancementProvider in their mod, this is how most the other vanilla providers work

Cyborgmas:
yeah but vanilla hardcoded all of its advancements.
Maybe making a ForgeAdvProvider designed to be extended would make more sense?
In it you could expose the builder stuff found in EndAdvancements to juste name one.

DaemonUmbra:
Yeah I'd forgotten how this system worked, for some reason I thought it was built around static methods and forgot so many of the helpers are non-static in each provider.

tterrag1098 (approved):

percivalalb:
Been busy the last week so little progress. Before I continue further, wanted to share my plan and check it is ok, something along the lines of...

- `DisplayInfoBuilder`: https://gist.github.com/percivalalb/66716721a1bdbc9aae22bd607642ab63

- Create another builder called `AdvancementBuilder` which will pretty much be a copy of `net.minecraft.advancements.Advancement.Builder` but with extra helper methods aimed at modders.

- Change the Advancement provider to be similar to the recipe provider: E.g

```
protected void registerAdvancements(Consumer<Advancement> consumer) {
    consumer.accept($CREATE_YOUR_ADVANCEMENT$);
}
```
instead of current changes fb04e1c6397ff4df461b05e4ddd480a7cd696b94

Is adding these builders beyond Forge's mission and would create too much extra code to be maintained?

PR #6737, opened 2020-06-02
Re-implement forge.logging.noAnsi...

DaemonUmbra:
...for use with consoles that do not support ANSI

Windows CMD and Powershell are detected as consoles for the purpose of enabling ANSI, despite their ANSI support being disabled by default, resulting in raw ANSI symbols being printed.

(Also includes a license fix)

### DIFF
--- a/src/fmllauncher/resources/log4j2.xml
+++ b/src/fmllauncher/resources/log4j2.xml
@@ -20,10 +20,10 @@
     <Appenders>
         <TerminalConsole name="Console">
             <PatternLayout>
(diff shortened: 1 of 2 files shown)


ichttt:
I'm not sure if this is the correct way to fix this issue. TCA already knows if the terminal can support colors or not, we just need a way to make log4j's HighlightConverter aware of the TCA state. I already opened two PRs on minecrell's repo that propose a possible change to TCA to achive the same result on supported terminals and auto-disable on unsupported ones, but I think he is still not happy with them.

PR #6739, opened 2020-06-02
Back to gradle 4.9 because fg works better with that version.

Codetoil:


LexManos:
As sated on discord, you have to actually make a argument.

Issue #6563, opened 2020-03-13
[1.14.4/1.15.2] Userdev Login provides incorrect userProperties launch argument


Choonster:
Originally reported [here](https://www.minecraftforge.net/forum/topic/76829-1144-runclient-crash-with-username/) for 1.14.4, another user reported it for 1.15.2 a few hours ago. The problem code appears to be in both versions.

Running the client in userdev with the `--username` and `--password` argument produces [this error](https://scans.gradle.com/s/uiopxxndtcji2/console-log#L764):
```
 java.lang.RuntimeException: java.lang.reflect.InvocationTargetException	
 	at cpw.mods.modlauncher.LaunchServiceHandlerDecorator.launch(LaunchServiceHandlerDecorator.java:39)	
 	at cpw.mods.modlauncher.LaunchServiceHandler.launch(LaunchServiceHandler.java:50)
 	at cpw.mods.modlauncher.LaunchServiceHandler.launch(LaunchServiceHandler.java:68)	
 	at cpw.mods.modlauncher.Launcher.run(Launcher.java:80)	
 	at cpw.mods.modlauncher.Launcher.main(Launcher.java:65)	
 	at net.minecraftforge.userdev.LaunchTesting.main(LaunchTesting.java:101)	
 Caused by: java.lang.reflect.InvocationTargetException	
 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)	
 	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)	
 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)	
 	at java.lang.reflect.Method.invoke(Unknown Source)	
 	at net.minecraftforge.userdev.FMLUserdevClientLaunchProvider.lambda$launchService$0(FMLUserdevClientLaunchProvider.java:55)	
 	at cpw.mods.modlauncher.LaunchServiceHandlerDecorator.launch(LaunchServiceHandlerDecorator.java:37)	
 	... 5 more	
 Caused by: com.google.gson.JsonSyntaxException: com.google.gson.stream.MalformedJsonException: Use JsonReader.setLenient(true) to accept malformed JSON at line 1 column 3 path $.	
 	at com.google.gson.internal.Streams.parse(Streams.java:60)	
 	at com.google.gson.internal.bind.TreeTypeAdapter.read(TreeTypeAdapter.java:65)	
[...]

PR #6740, opened 2020-06-02
Potential userProperties Fix

DaemonUmbra:
This PR *should* fix the serialization issues with userProperties in userdev for people whose accounts have userProperties attached.

I can't test it myself as my account does not return any userProperties during auth, searching for testers.

### DIFF
--- a/src/userdev/java/net/minecraftforge/userdev/LaunchTesting.java
+++ b/src/userdev/java/net/minecraftforge/userdev/LaunchTesting.java
@@ -20,9 +20,12 @@
 package net.minecraftforge.userdev;
 
 import com.google.common.base.Strings;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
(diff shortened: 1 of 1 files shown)


jamierocks:
This would fix GH-6563.

> I can't test it myself as my account does not return any userProperties during auth, searching for testers.

I've actually got a branch locally with effectively the same changes, but was also unable to verify it actually fixed launch. Though I am in no real doubt it does.

DaemonUmbra:
I've been burned by "this will definitely work" before...

jamierocks:
> I've been burned by "this will definitely work" before...

That's why I didn't make a pr, I had intended to look at how it used to be implemented to verify - just didn't get around to it.

DaemonUmbra:
Someone [on the forums](https://www.minecraftforge.net/forum/topic/86244-intellij-unable-to-use-minecraft-account/?do=findComment&comment=405587) reports that it works.

PR #6742, opened 2020-06-04
Add custom registry type support to DeferredRegister.

LexManos:
Still needs javadocs, and some way of making a soft DeferredRegistry.
But I couldn't think of a way to make sure type saftey in a soft reference, as we need the base class.

### DIFF
--- a/src/main/java/net/minecraftforge/registries/DeferredRegister.java
+++ b/src/main/java/net/minecraftforge/registries/DeferredRegister.java
@@ -21,6 +21,7 @@ package net.minecraftforge.registries;
 
 import net.minecraft.util.ResourceLocation;
 import net.minecraftforge.event.RegistryEvent;
+import net.minecraftforge.eventbus.api.EventPriority;
 import net.minecraftforge.eventbus.api.IEventBus;
 import net.minecraftforge.fml.RegistryObject;
 
@@ -32,6 +33,8 @@ import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
 import java.util.function.Supplier;
+
+import com.google.common.reflect.TypeToken;
 
 /**
  * Utility class to help with managing registry entries.
@@ -56,15 +59,35 @@ import java.util.function.Supplier;
  */
 public class DeferredRegister<T extends IForgeRegistryEntry<T>>
 {
-    private final IForgeRegistry<T> type;
+    public static <B extends IForgeRegistryEntry<B>> DeferredRegister<B> create(IForgeRegistry<B> reg, String modid)
+    {
+        return new DeferredRegister<B>(reg, modid);
+    }
+
+    public static <B extends IForgeRegistryEntry<B>> DeferredRegister<B> create(Class<B> base, String modid)
+    {
+        return new DeferredRegister<B>(base, modid);
+    }
+
+    private final Class<T> superType;
     private final String modid;
     private final Map<RegistryObject<T>, Supplier<? extends T>> entries = new LinkedHashMap<>();
     private final Set<RegistryObject<T>> entriesView = Collections.unmodifiableSet(entries.keySet());
 
+    private IForgeRegistry<T> type;
+    private Supplier<RegistryBuilder<T>> registryFactory;
+
+    private DeferredRegister(Class<T> base, String modid)
+    {
+        this.superType = base;
+        this.modid = modid;
+    }
+
+    @Deprecated // Make private in 1.16, Use create. Constructors are an implementation detail.
     public DeferredRegister(IForgeRegistry<T> reg, String modid)
     {
+        this(reg.getRegistrySuperType(), modid);
         this.type = reg;
-        this.modid = modid;
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -56,15 +59,35 @@
  */
 public class DeferredRegister<T extends IForgeRegistryEntry<T>>
 {
-    private final IForgeRegistry<T> type;
+    public static <B extends IForgeRegistryEntry<B>> DeferredRegister<B> create(IForgeRegistry<B> reg, String modid)
```
Cyborgmas, line 65:
```suggestion
    /**
     * Use for vanilla/forge registries. See example above.
     */
    public static <B extends IForgeRegistryEntry<B>> DeferredRegister<B> create(IForgeRegistry<B> reg, String modid)
```

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -56,15 +59,35 @@
...
+    {
+        return new DeferredRegister<B>(reg, modid);
+    }
+
+    public static <B extends IForgeRegistryEntry<B>> DeferredRegister<B> create(Class<B> base, String modid)
```
Cyborgmas, line 73:
```suggestion
    /**
     * Use for custom registries that are made during the NewRegistry event. 
     */
    public static <B extends IForgeRegistryEntry<B>> DeferredRegister<B> create(Class<B> base, String modid)
```

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -80,13 +103,32 @@ public DeferredRegister(IForgeRegistry<T> reg, String modid)
...
+            throw new IllegalStateException("Could not create RegistryObject in DeferredRegister");
+
         if (entries.putIfAbsent((RegistryObject<T>) ret, () -> sup.get().setRegistryName(key)) != null) {
             throw new IllegalArgumentException("Duplicate registration " + name);
         }
+
         return ret;
     }
 
+    public Supplier<IForgeRegistry<T>> makeRegistry(final String name, final Supplier<RegistryBuilder<T>> sup) {
```
Cyborgmas, line 137:
```suggestion
    /**
     * For custom registries only, fills the {@link #registryFactory} to be called later see {@link #register(IEventBus)}
     *
     * Calls {@link RegistryBuilder#setName} and {@link RegistryBuilder#setType} automatically.
     *
     * @param name  Path of the registry's {@link ResourceLocation}
     * @param sup   Supplier of the RegistryBuilder that is called to fill {@link #type} during the NewRegistry event
     * @return      A supplier of the {@link IForgeRegistry} created by the builder.
     */
    public Supplier<IForgeRegistry<T>> makeRegistry(final String name, final Supplier<RegistryBuilder<T>> sup) {
```

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -80,13 +103,32 @@ public DeferredRegister(IForgeRegistry<T> reg, String modid)
...
+        if (this.type != null || this.registryFactory != null)
+            throw new IllegalStateException("Cannot create a registry for a type that already exists");
+
+        this.registryFactory = () -> sup.get().setName(new ResourceLocation(modid, name)).setType(this.superType);
+        return () -> this.type;
+    }
+
     /**
      * Adds our event handler to the specified event bus, this MUST be called in order for this class to function.
      * See the example usage.
```
Cyborgmas, line 149:
```suggestion
     * See the example usage.
     *
     * If using this for a custom registry, creates the the registry for you in {@link #createRegistry} using the {@link RegistryEvent.NewRegistry} event
     * If no {@link #registryFactory} was provided, listens to the NewRegistry event on {@link EventPriority.LOWEST} to check if the registry was properly created.
```

Cyborgmas (commented):
Added javadocs. No idea if there was a cleaner way to do this

src/main/java/net/minecraftforge/registries/DeferredRegister.java
pupnewfster: In java it isn't `@see THING` it is `{@link THING}`

Cyborgmas: Hmm it worked though. Gave me a link to the function.

pupnewfster: Ah turns out java also has a `@see` I just have never actually seen it used before so didn't realize it was actually a thing.

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -108,7 +156,7 @@ public void register(IEventBus bus)
 
     private void addEntries(RegistryEvent.Register<?> event)
     {
-        if (event.getGenericType() == this.type.getRegistrySuperType())
+        if (this.type != null && event.getGenericType() == this.type.getRegistrySuperType())
```
pupnewfster, line 174: As a reference to `Class<T> superType` is now stored in `DeferredRegister` would it make more sense to just compare `event.getGenericType()` to that rather than querying the registry for it?

LexManos: No, because the intent is to not have superType be in here/definitive.
As one of the intentions is to be able to be used as a soft link.
By referring to the registry by name or some other method.

PR #6744, opened 2020-06-05
TileEntityType: Allow registration via passing an ImmutableSet directly

hanetzer:
Lex was showing me another way of going about this with Suppliers and ResourceLocations,
but as I didn't really understand the subject matter I could not keep up with the edits. This was
my original idea.

### DIFF
--- a/patches/minecraft/net/minecraft/tileentity/TileEntityType.java.patch
+++ b/patches/minecraft/net/minecraft/tileentity/TileEntityType.java.patch
@@ -9,3 +9,14 @@
     private static final Logger field_206866_A = LogManager.getLogger();
     public static final TileEntityType<FurnaceTileEntity> field_200971_b = func_200966_a("furnace", TileEntityType.Builder.func_223042_a(FurnaceTileEntity::new, Blocks.field_150460_al));
     public static final TileEntityType<ChestTileEntity> field_200972_c = func_200966_a("chest", TileEntityType.Builder.func_223042_a(ChestTileEntity::new, Blocks.field_150486_ae));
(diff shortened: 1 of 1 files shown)


diesieben07:
I am not quite sure which problem this solves.

hanetzer:
[SmelteryBlocks.java#L85](https://github.com/SlimeKnights/TinkersConstruct/blob/bf61f051600f7bbf637ac4d62fa8065f0822d87a/src/main/java/slimeknights/tconstruct/blocks/SmelteryBlocks.java#L85) [SmelteryTileEntities.java#L27](https://github.com/SlimeKnights/TinkersConstruct/blob/1.15/src/main/java/slimeknights/tconstruct/tileentity/SmelteryTileEntities.java#L27)

We create a ImmutableSet of valid smeltery blocks in the first link, used to do the
multiblock creation logic, and in the second one we pass 36 different suppliers to
the TE reg code (12 block variants, fullblock, slab, and stairs variants). With this
commit we could just pass the one ImmutableList.

KnightMiner:
We can just pass an immutable set as the set, it implements set afterall

hanetzer:
> commit we could just pass the one ~~ImmutableList.~~


> We can create an immutable set just as easily.

Typo. ImmutableSet

KnightMiner:
My point is, there is no reason we have to use the builder, we can pass the set directly into the constructor as that appears to be public. Fewer Forge patches needed then, especially for something few people will use.

hanetzer:
I have no idea why I didn't even see that. That is a valid option, as far as I can tell.

PR #6745, opened 2020-06-06
[1.15] Add Modded Ocean Biome Support

SmellyModder:
This PR allows for new ocean biomes to be added to work properly with Vanilla's Layers involving oceans.

Many biome mods want to add Ocean Biomes, but they can't add them properly currently, and this PR fixes that issue. 

There is a new method in Forge's `BiomeManager` called `addOceanBiome()`, which is used to add new ocean biomes.


src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -36,14 +37,8 @@
 {
     private static TrackedList<BiomeEntry>[] biomes = setupBiomes();
 
-    public static List<Biome> oceanBiomes = new ArrayList<Biome>();
```
ichttt, line 39: This is a binary breaking change, as this field was public

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -36,14 +37,8 @@
...
-
-    static
-    {
-        oceanBiomes.add(Biomes.OCEAN);
-        oceanBiomes.add(Biomes.DEEP_OCEAN);
-        oceanBiomes.add(Biomes.FROZEN_OCEAN);
-    }
+    private static List<Biome> deepOceanBiomes = new ArrayList<Biome>();
```
ichttt: Why are deep ocean biomes special cased? If you already add something special for deep oceans, why not for "normal" oceans as well?

ichttt (commented):
Not a complete review, just what I noticed while skimming through this

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: `oceanBiomes` was never actually called externally in 1.13+, so keeping it private would be ok in this case. 

I made the ocean biome entry system similar to the normal biome entry system, which is why I kept it private. 

Since `oceanBiomes` was never actually used truly in 1.13+, I thought it would be okay to convert to a system more like the normal biome entry system.

Making it public would allow people to set a new value on the field also.

If you still think I should make it public, that's fine, I just wanted to give my reasoning.

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: Accidently left that part over from a different way I attempted doing the ocean biome entry system.

Will fix 👍

src/main/java/net/minecraftforge/common/BiomeManager.java
pupnewfster: ichttt's point is that once there is an RB binary breaking changes are not allowed. So if some modder somewhere was calling `oceanBiomes` (even if it wasn't actually doing anything), it would cause their mod to start crashing.

SmellyModder: I see, point taken. I will make it public 👍

Although again I will say outside of this PR, in 1.13+ versions of forge, manipulating this field in any way will not affect anything in-game, so I'd be surprised if anyone would still be using it in 1.15

marchermans: Should we not mark this as @Deprecated then?

valoeghese: marking the field as `@Deprecated` seems like the best course of action

patches/minecraft/net/minecraft/world/gen/layer/IslandLayer.java.patch
```diff
@@ -0,0 +1,11 @@
+--- a/net/minecraft/world/gen/layer/IslandLayer.java
++++ b/net/minecraft/world/gen/layer/IslandLayer.java
+@@ -10,7 +10,7 @@
+       if (p_215735_2_ == 0 && p_215735_3_ == 0) {
+          return 1;
+       } else {
+-         return p_215735_1_.func_202696_a(10) == 0 ? 1 : LayerUtil.field_202832_c;
++         return p_215735_1_.func_202696_a(10) == 0 ? 1 : net.minecraft.util.registry.Registry.field_212624_m.func_148757_b(OceanLayer.getWeightedBiomeEntry(net.minecraftforge.common.BiomeManager.OceanType.NORMAL, p_215735_1_).biome);
```
valoeghese: This part of the codebase isn't to specify specific biomes, but to specify the general placement of continents and oceanic areas. So I don't think adding the ocean biome entry here is the best course of action.
Especially because this is on a much larger scale/zoom-out than the continental BiomeLayer.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -123,6 +146,48 @@ public static void removeBiome(BiomeType type, BiomeEntry entry)
...
+    public static ImmutableList<OceanBiomeEntry> getOceanBiomes(OceanType type)
+    {
+        int idx = type.ordinal();
+        List<OceanBiomeEntry> list = idx >= oceanBiomes.length ? null : oceanBiomes[idx];
+
+        return list != null ? ImmutableList.copyOf(list) : null;
+    }
+
+    @Nullable
+    public static ImmutableList<OceanBiomeEntry> getOceanBiomesForBiome(Biome biome, @Nullable OceanType type, boolean excludeWarm)
```
valoeghese: This method could have a better name imo. Something showing that it specifically gets the complete biome entries which contain this biome as the primary biome, rather than this vague modder-unfriendly name

patches/minecraft/net/minecraft/world/gen/layer/OceanLayer.java.patch
```diff
@@ -0,0 +1,25 @@
...
+       } else if (d0 < -0.4D) {
+-         return LayerUtil.field_202831_b;
++         return net.minecraft.util.registry.Registry.field_212624_m.func_148757_b(getWeightedBiomeEntry(net.minecraftforge.common.BiomeManager.OceanType.FROZEN, p_215735_1_).biome);
+       } else {
+-         return d0 < -0.2D ? LayerUtil.field_203634_d : LayerUtil.field_202832_c;
++         return d0 < -0.2D ? net.minecraft.util.registry.Registry.field_212624_m.func_148757_b(getWeightedBiomeEntry(net.minecraftforge.common.BiomeManager.OceanType.COLD, p_215735_1_).biome) : net.minecraft.util.registry.Registry.field_212624_m.func_148757_b(this.getWeightedBiomeEntry(net.minecraftforge.common.BiomeManager.OceanType.NORMAL, p_215735_1_).biome);
+       }
+    }
++
++   public static net.minecraftforge.common.BiomeManager.OceanBiomeEntry getWeightedBiomeEntry(net.minecraftforge.common.BiomeManager.OceanType type, INoiseRandom context) {
```
valoeghese: Wouldn't it be better to have a smaller patch by moving this method to an impl class? IIRC forge prefers their patch files to be nice and small so they can fix patches easier across updates.

valoeghese: Or is there a reasoning for having the method here?

valoeghese (changes requested):

patches/minecraft/net/minecraft/world/gen/layer/OceanLayer.java.patch
marchermans: The general consensus is to not override in patches if the base class can define the logic.
To be fair I had to ask as well.

But since this is at the bottom of the patch file it does not matter.

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: Yes, `@Deprecated` does seem like the best decision.

patches/minecraft/net/minecraft/world/gen/layer/IslandLayer.java.patch
SmellyModder: Looked into this more and yes, this does seem to be the case, I'll just remove the patch then 👍

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: I was thinking the same thing recently and felt like changing it, so I will change it.

patches/minecraft/net/minecraft/world/gen/layer/OceanLayer.java.patch
SmellyModder: I noticed `BiomeLayer` was doing a very similar thing to what I needed for `OceanLayer` at the bottom of the class, so I thought it'd be ok.

So, what do you guys think, keep it in the class or move it to `BiomeManager`?

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -123,6 +146,48 @@ public static void removeBiome(BiomeType type, BiomeEntry entry)
         return list != null ? ImmutableList.copyOf(list) : null;
     }
 
+    @Nullable
+    public static ImmutableList<OceanBiomeEntry> getOceanBiomes(OceanType type)
```
diesieben07, line 203: Not threadsafe, missing documentation.

src/main/java/net/minecraftforge/common/BiomeManager.java
diesieben07: Not threadsafe, missing documentation.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -114,6 +119,24 @@ public static void removeBiome(BiomeType type, BiomeEntry entry)
...
     }
 
+    public static void addOceanBiome(OceanType type, OceanBiomeEntry entry)
+    {
+        int idx = type.ordinal();
+        List<OceanBiomeEntry> list = idx > oceanBiomes.length ? null : oceanBiomes[idx];
+        if (list != null) list.add(entry);
+    }
+
+    public static void removeOceanBiome(OceanType type, OceanBiomeEntry entry)
```
diesieben07, line 157: Not threadsafe, missing documentation.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -114,6 +119,24 @@ public static void removeBiome(BiomeType type, BiomeEntry entry)
         }
     }
 
+    public static void addOceanBiome(OceanType type, OceanBiomeEntry entry)
```
diesieben07, line 141: Not threadsafe, missing documentation.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -80,6 +74,17 @@
         return currentBiomes;
     }
 
+    private static List<OceanBiomeEntry>[] setupOceanBiomes()
```
diesieben07: EnumMap would be a much cleaner solution here, no?

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -123,6 +146,48 @@ public static void removeBiome(BiomeType type, BiomeEntry entry)
...
+                if(entry.biome == biome && (excludeWarm ? type != OceanType.WARM : true))
+                {
+                    list.add(entry);
+                }
+            }
+        }
+        return list.isEmpty() ? null : ImmutableList.copyOf(list);
+    }
+
+    public static boolean isOceanBiome(Biome biome, boolean shallow)
```
diesieben07, line 261: Not threadsafe, lacking documentation.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -142,6 +207,11 @@ public static BiomeType create(String name) {
         }
     }
 
+    public static enum OceanType
```
diesieben07, line 302: Lacking documentation.

diesieben07 (changes requested):
None of your methods have documentation (what do they do, when should you call them?)
None of your methods are threadsafe, but your test mod implies you intend for people to call them during parallel mod loading.

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: Yes, this definitely a valid suggestion and I'll get around to fixing it 👍

This is also my response to the similar suggestions regarding thread safety and lack of documentation below.

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: In my opinion, it would be a much cleaner solution, yes. 

The thought of having it be an `EnumMap` did cross my mind when I was first making the change to the field, but it would be inconsistent with the already present `biomes` field in the class, so I decided not to do it and stay consistent.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -123,6 +166,83 @@ public static void removeBiome(BiomeType type, BiomeEntry entry)
...
+                        list.add(entry);
+                    }
+                }
+            }
+            return list.isEmpty() ? null : ImmutableList.copyOf(list);
+        }
+    }
+
+    /**
+     * Checks if a biome is present in the ocean biome entries for every OceanType.
```
diesieben07: _any_ `OceanType`, not every.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -123,6 +166,83 @@ public static void removeBiome(BiomeType type, BiomeEntry entry)
...
+    @Nullable
+    public static ImmutableList<OceanBiomeEntry> getOceanBiomesContainingBiome(Biome biome, @Nullable OceanType type, boolean excludeWarm)
+    {
+        OceanType[] values = type != null ? new OceanType[] {type} : OceanType.values();
+        List<OceanBiomeEntry> list = new ArrayList<>();
+        synchronized (BiomeManager.class)
+        {
+            for(OceanType types : values)
+            {
+                for(OceanBiomeEntry entry : getOceanBiomes(types))
```
diesieben07: `getOceanBiomes` can return null, this will throw a NPE.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -114,6 +119,44 @@ public static void removeBiome(BiomeType type, BiomeEntry entry)
...
+     * @param type The type of ocean for this OceanBiomeEntry.
+     * @param entry The OceanBiomeEntry to remove.
+     */
+    public static void removeOceanBiome(OceanType type, OceanBiomeEntry entry)
+    {
+        int idx = type.ordinal();
+        synchronized (BiomeManager.class)
+        {
+            List<OceanBiomeEntry> list = idx > oceanBiomes.length ? null : oceanBiomes[idx];
+            if (list != null && list.contains(entry))
```
diesieben07: You don't need this `contains` check.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -114,6 +119,44 @@ public static void removeBiome(BiomeType type, BiomeEntry entry)
...
+     *
+     * @param type The type of ocean for this OceanBiomeEntry.
+     * @param entry The OceanBiomeEntry to add.
+     */
+    public static void addOceanBiome(OceanType type, OceanBiomeEntry entry)
+    {
+        int idx = type.ordinal();
+        synchronized (BiomeManager.class) {
+            List<OceanBiomeEntry> list = idx > oceanBiomes.length ? null : oceanBiomes[idx];
+            if (list != null) list.add(entry);
```
diesieben07: Should this not be an error condition? When can the index be out of bounds so that list is null?

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -35,15 +36,8 @@
...
-    public static List<Biome> oceanBiomes = new ArrayList<Biome>();
-
-    static
-    {
-        oceanBiomes.add(Biomes.OCEAN);
-        oceanBiomes.add(Biomes.DEEP_OCEAN);
-        oceanBiomes.add(Biomes.FROZEN_OCEAN);
-    }
+    @Deprecated //Don't call this method directly, use the proper methods in this class
+    public static List<OceanBiomeEntry>[] oceanBiomes = setupOceanBiomes();
```
diesieben07: This is still a binary breaking change as you have changed the type of the field.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -123,6 +166,83 @@ public static void removeBiome(BiomeType type, BiomeEntry entry)
         return list != null ? ImmutableList.copyOf(list) : null;
     }
 
+    /**
+     * Gets the ocean biome entries for a type of ocean.
+     * This method should be called to get the ocean biome entries for a type of ocean.
+     * This method is threadsafe.
+     *
+     * @param type The type of ocean to get the entries for.
+     * @return An ImmutableList of the ocean biome entries for the ocean type or null if there are no entries.
```
diesieben07: This is incorrect, if a biome is added and then removed so that the list becomes empty, an empty list is returned instead of `null` even though there are no entries.
This makes it so your patch code above will fail, because it checks for null specifically.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -123,6 +166,83 @@ public static void removeBiome(BiomeType type, BiomeEntry entry)
...
+     * @param shallow If the biome is shallow (i.e. not a deep ocean)
+     * @return True if the biome is an ocean or is a shallow ocean if shallow is true
+     */
+    public static boolean isOceanBiome(Biome biome, boolean shallow)
+    {
+        synchronized (BiomeManager.class)
+        {
+            for(OceanType type : OceanType.values())
+            {
+                for(OceanBiomeEntry entry : getOceanBiomes(type))
```
diesieben07: getOceanBiomes can be null.
Also, it might be worth caching this result? Do we expect a lot of biomes in here?

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -123,6 +166,83 @@ public static void removeBiome(BiomeType type, BiomeEntry entry)
...
+            {
+                for(OceanBiomeEntry entry : getOceanBiomes(types))
+                {
+                    if(entry.biome == biome && (excludeWarm ? type != OceanType.WARM : true))
+                    {
+                        list.add(entry);
+                    }
+                }
+            }
+            return list.isEmpty() ? null : ImmutableList.copyOf(list);
```
diesieben07: You are building this list and then just create a copy of it. Why? Use `ImmutableList.builder`.

patches/minecraft/net/minecraft/world/gen/layer/DeepOceanLayer.java.patch
```diff
@@ -0,0 +1,32 @@
...
+-               return LayerUtil.field_203637_i;
+-            }
+-
+-            if (p_202748_6_ == LayerUtil.field_202831_b) {
+-               return LayerUtil.field_203638_j;
+-            }
+-
+-            return LayerUtil.field_202830_a;
++            java.util.List<net.minecraftforge.common.BiomeManager.OceanBiomeEntry> oceanBiomes = net.minecraftforge.common.BiomeManager.getOceanBiomesContainingBiome(net.minecraft.util.registry.Registry.field_212624_m.func_148745_a(p_202748_6_), null, false);
++            return oceanBiomes != null ? net.minecraft.util.registry.Registry.field_212624_m.func_148757_b((net.minecraft.util.WeightedRandom.func_180166_a(oceanBiomes, p_202748_1_.func_202696_a(net.minecraft.util.WeightedRandom.func_76272_a(oceanBiomes))).deepOcean)) : LayerUtil.field_202830_a;
```
diesieben07: Might be worth to put this in a method in `ForgeHooks`, to make it more readable.

patches/minecraft/net/minecraft/world/gen/layer/MixOceansLayer.java.patch
```diff
@@ -0,0 +1,50 @@
...
+                }
+             }
+          }
+ 
+          if (i == LayerUtil.field_202830_a) {
+-            if (j == LayerUtil.field_203633_b) {
+-               return LayerUtil.field_203636_g;
++            java.util.List<net.minecraftforge.common.BiomeManager.OceanBiomeEntry> entries = net.minecraftforge.common.BiomeManager.getOceanBiomesContainingBiome(jBiome, null, true);
++            if (entries != null) {
++               return net.minecraft.util.registry.Registry.field_212624_m.func_148757_b(net.minecraft.util.WeightedRandom.func_180166_a(entries, p_215723_1_.func_202696_a(net.minecraft.util.WeightedRandom.func_76272_a(entries))).mixOcean);
```
diesieben07: The patches in this class should almost definitely be a method in `ForgeHooks` imho.

patches/minecraft/net/minecraft/world/gen/layer/OceanLayer.java.patch
```diff
@@ -0,0 +1,25 @@
...
+       } else {
+-         return d0 < -0.2D ? LayerUtil.field_203634_d : LayerUtil.field_202832_c;
++         return d0 < -0.2D ? net.minecraft.util.registry.Registry.field_212624_m.func_148757_b(getWeightedBiomeEntry(net.minecraftforge.common.BiomeManager.OceanType.COLD, p_215735_1_).biome) : net.minecraft.util.registry.Registry.field_212624_m.func_148757_b(getWeightedBiomeEntry(net.minecraftforge.common.BiomeManager.OceanType.NORMAL, p_215735_1_).biome);
+       }
+    }
++
++   private net.minecraftforge.common.BiomeManager.OceanBiomeEntry getWeightedBiomeEntry(net.minecraftforge.common.BiomeManager.OceanType type, INoiseRandom context) {
++      java.util.List<net.minecraftforge.common.BiomeManager.OceanBiomeEntry> biomeList = net.minecraftforge.common.BiomeManager.getOceanBiomes(type);
++      return net.minecraft.util.WeightedRandom.func_180166_a(biomeList, context.func_202696_a(net.minecraft.util.WeightedRandom.func_76272_a(biomeList)));
++   }
```
diesieben07: Same as above, this stuff should be in ForgeHooks.

diesieben07 (changes requested):
Yes, things are technically threadsafe now. Not a fan of just chugging `synchronized` on everything, but I guess it will work.

Your idea of returning null if there are no entries is causing you problems already, this is somewhat of an antipattern in Java. Why not return an empty list?

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: Will change 👍

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: Will fix 👍 
Also, I'll make the methods that can return null just return an empty list instead.

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: Looking back at it, yes, it should really be an error condition.
Will fix 👍

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: True.
In that case, the best option would be to deprecate the old field and make a new field called `forgeOceanBiomes`(or some other similar name) to handle the ocean biomes, correct?
(I've mentioned this before, but in 1.13+ outside of this PR, the old field isn't even called outside of `BiomeManager` in forge)

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: Will fix. 
As said above, I will be making the nullable lists return an empty list instead.

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: First thing I've already answered above.

Caching it would be good if a lot of biomes get added, although it's just cycling through a copy of the ocean biomes for a type, so it shouldn't be that performance impacting.

If you still think caching it is a good idea I can do that 👍

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: True, I will fix that.

patches/minecraft/net/minecraft/world/gen/layer/DeepOceanLayer.java.patch
SmellyModder: Would make it more readable yeah, will do that.
(_Also this is my response to the similar suggestions below_)

src/main/java/net/minecraftforge/common/BiomeManager.java
diesieben07: Yes, make a new (private) field and deprecate the old one.
It doesn't matter if it's used or not, the point is it is public and so someone _could_ use it, so it has to stay.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -35,15 +36,8 @@
 public class BiomeManager
 {
     private static TrackedList<BiomeEntry>[] biomes = setupBiomes();
-
-    public static List<Biome> oceanBiomes = new ArrayList<Biome>();
-
-    static
```
valoeghese, line 41: Additionally, we don't want to change the default contents of `oceanBiomes` in case someone was using them .

valoeghese (changes requested):
What diesieben said, but I have one more nitpick to add on

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -80,6 +81,17 @@
...
 
+    private static List<OceanBiomeEntry>[] setupOceanBiomes()
+    {
+        List<OceanBiomeEntry>[] currentBiomes = new List[OceanType.values().length];
+        currentBiomes[OceanType.WARM.ordinal()] = Lists.newArrayList(new OceanBiomeEntry(Biomes.WARM_OCEAN, Biomes.DEEP_WARM_OCEAN, Biomes.LUKEWARM_OCEAN, 10));
+        currentBiomes[OceanType.LUKEWARM.ordinal()] = Lists.newArrayList(new OceanBiomeEntry(Biomes.LUKEWARM_OCEAN, Biomes.DEEP_LUKEWARM_OCEAN, Biomes.DEEP_LUKEWARM_OCEAN, 10));
+        currentBiomes[OceanType.FROZEN.ordinal()] = Lists.newArrayList(new OceanBiomeEntry(Biomes.FROZEN_OCEAN, Biomes.DEEP_FROZEN_OCEAN, Biomes.COLD_OCEAN, 10));
+        currentBiomes[OceanType.COLD.ordinal()] = Lists.newArrayList(new OceanBiomeEntry(Biomes.COLD_OCEAN, Biomes.DEEP_COLD_OCEAN, Biomes.DEEP_COLD_OCEAN, 10));
+        currentBiomes[OceanType.NORMAL.ordinal()] = Lists.newArrayList(new OceanBiomeEntry(Biomes.OCEAN, Biomes.DEEP_OCEAN, Biomes.DEEP_OCEAN, 10));
+        return currentBiomes;
```
diesieben07: This is still super ugly. You need a raw type here, too, because Arrays don't work with generics and it's just overall ugly.
Please use an `EnumMap`.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -114,6 +126,43 @@ public static void removeBiome(BiomeType type, BiomeEntry entry)
...
+    public static void removeOceanBiome(OceanType type, OceanBiomeEntry entry)
+    {
+        int idx = type.ordinal();
+        synchronized (BiomeManager.class)
+        {
+            if(type == OceanType.NORMAL && forgeOceanBiomes[idx].size() <= 1)
+            {
+                throw new IllegalStateException("At least one NORMAL type ocean biome entry must be present");
+            }
+            forgeOceanBiomes[idx].remove(entry);
```
diesieben07: This requires you to implement `equals` and `hashCode` in `OceanBiomeEntry`, otherwise this method is hard to use (you have to call `getOceanBiomes` first and find the entry you want to remove there...).
Your size check also does not take into account the fact that one might pass in an entry that is not actually contained in the list.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -114,6 +126,43 @@ public static void removeBiome(BiomeType type, BiomeEntry entry)
...
+     * This method is threadsafe.
+     *
+     * @param type The type of ocean for this OceanBiomeEntry.
+     * @param entry The OceanBiomeEntry to add.
+     */
+    public static void addOceanBiome(OceanType type, OceanBiomeEntry entry)
+    {
+        synchronized (BiomeManager.class)
+        {
+            forgeOceanBiomes[type.ordinal()].add(entry);
```
diesieben07: There should probably be some conflict checking here... What if two people add the same set of biomes but with a different weight? Do we allow that? Throw an exception? Only add the one with the higher weight? Add the weights?

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1196,4 +1201,31 @@ public static synchronized void updateBurns()
         return list;
     }
 
+    public static BiomeManager.OceanBiomeEntry getWeightedOceanBiomeEntry(BiomeManager.OceanType type, INoiseRandom context)
+    {
+        List<BiomeManager.OceanBiomeEntry> biomeList = BiomeManager.getOceanBiomes(type);
+        if(biomeList.isEmpty()) biomeList = BiomeManager.getOceanBiomes(BiomeManager.OceanType.NORMAL);
```
diesieben07: Formatting (missing space after if)

diesieben07 (changes requested):
Some minor changes still.
Also, I still think some of this stuff needs to be cached. In `getMixedOceansForOcean` you call `getOceanBiomesContainingBiome` three times which will each time copy the whole list of entries and then iterate it only to add it all to _yet another_ list (every time).

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: I'll change it to use an `EnumMap` 👍

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: Hm yeah, those are true. 
Although, I'm thinking of removing this method so that there'll be fewer conflict issues in mods, thoughts?

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: It's fine if someone decides to add the same set of biomes with a different weight. 
A. because someone might want ocean biome entries that are the same, but exist for different ocean types, and B. because someone might want two very similar ocean biome entries that have different weights.

src/main/java/net/minecraftforge/common/ForgeHooks.java
SmellyModder: Will fix 👍

SmellyModder:
@diesieben07 Hmm, yes, I would agree that caching some things to make the `getOceanBiomesContainingBiome` method have better performance is a good idea. 
Although I'm wondering how such a thing should be done. Since each parameter in `getOceanBiomesContainingBiome` affects the returned list.

A way I can think of doing it would be creating a map for the biomes and have the value(`V`) for that map be an `EnumMap` of a new enum I'd create for the variations of the parameters I used for that method that would store the lists. 
It's a bit hard to explain, but it'd be like this: `Map<Biome, EnumMap<OceanFinderType, List<OceanBiomeEntry>>>`

Any other ideas on how it could be done?

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -35,11 +36,11 @@
...
 {
     private static TrackedList<BiomeEntry>[] biomes = setupBiomes();
-
+    @Deprecated //Don't call this field, use the proper methods in this class to add and get ocean biomes
     public static List<Biome> oceanBiomes = new ArrayList<Biome>();
+    private static List<OceanBiomeEntry>[] forgeOceanBiomes = setupOceanBiomes();
 
-    static
-    {
+    static {
```
valoeghese: no need to change the formatting of this line.

valoeghese (changes requested):

src/main/java/net/minecraftforge/common/BiomeManager.java
SmellyModder: Yup will fix 👍 
Must've autoformatted.

ghost:
As long as this PR is stable I would love it merged ASAP, I need to add some ocean biomes to my mod :)

TheCurle:
Changes still need to be made, "i want this pls merge thx" does not help us.
Patience would be appreciated right now.

pupnewfster:
Not to mention this PR is going to have to be ported to 1.16 and merged into that before it can be backported to 1.15 due to how the LTS system works

SmellyModder:
Going to make sure this PR is reviewed well and then I'll work on porting it to 1.16

diesieben07:
> Going to make sure this PR is reviewed well and then I'll work on porting it to 1.16

That is not how it works, unfortunately.

PR #6523, opened 2020-02-19
Add a way to create Music Discs without requiring a SoundEvent

TheSilkMiner:
### Abstract
Creating a `MusicDiscItem` requires the user to supply a `SoundEvent`, which is a `ForgeRegistryEntry`. For this reason, these entries should only be created and registered in their `RegistryEvent.Register`. In this case, though, the registry event for `SoundEvent`s fires after the one for `Item`s, making it effectively impossible to create a `MusicDiscItem`.

### The Changes
What this PR does is change this requirement so that the user can use a `Supplier<SoundEvent>`, so that the actual `SoundEvent` can be retrieved and loaded as needed. Vanilla has also been patched to account for this behavior.
Additionally, `WorldRenderer` has been patched not to be stupid, rendering the `RECORDS` map effectively useless, since the look up is now performed only if the two-parameter version of `playRecord` is called with a non-`null` `SoundEvent`, which vanilla doesn't call anymore. In fact, the `MusicDiscItem` that vanilla retrieves from `playEvent` isn't discarded anymore, but it's passed to Forge's three-parameter version.
This PR also provides a test mod that demonstrates the usage of the newly added hook.

PR #6746, opened 2020-06-06
Add data generators for sounds.json

TheSilkMiner:
### Abstract

The goal of this PR is to provide a data generator for the `sounds.json` file, as previously requested in #6523 ([comment](https://github.com/MinecraftForge/MinecraftForge/pull/6523#discussion_r425950102)).
All restrictions in place have been taken directly from vanilla's `SoundListSerializer`, which is used to convert this file into an actual list of sounds.

### Questions
- Validations of vanilla sounds is, at the moment, not possible, since the "additional" resource pack that gets downloaded by the launcher isn't loaded by the `ExistingFileHelper`. At the moment I am logging a warning and letting it pass. Should I approach this differently?


src/main/java/net/minecraftforge/client/sounds/generators/SoundDefinitionsProvider.java
```diff
@@ -0,0 +1,245 @@
...
+    private boolean validate(final String name, final SoundDefinition def) {
+        return def.soundList().stream().anyMatch(it -> this.validate(name,it));
+    }
+
+    private boolean validate(final String name, final SoundDefinition.Sound sound) {
+        switch (sound.type()) {
+            case SOUND: return this.validateSound(name, sound.name());
+            case EVENT: return this.validateEvent(name, sound.name());
+        }
+        throw new AssertionError();
```
marchermans: This should probably be a different exception here, at least give it a proper message.

src/main/java/net/minecraftforge/client/sounds/generators/SoundDefinitionsProvider.java
```diff
@@ -0,0 +1,245 @@
...
+
+    private boolean validateSound(final String soundName, final ResourceLocation name) {
+        final boolean valid = this.helper.exists(name, ResourcePackType.CLIENT_RESOURCES, ".ogg", "sounds");
+        if (!valid) {
+            final String path = name.getNamespace() + ":sounds/" + name.getPath() + ".ogg";
+            if (!"minecraft".equals(name.getNamespace())) {
+                LOGGER.warn("Unable to find corresponding OGG file '{}' for sound event '{}'", path, soundName);
+            } else {
+                LOGGER.warn("Unable to find original Minecraft sound OGG file '{}' for event '{}': assuming it's part of the data pack downloaded by the launcher", path, soundName);
+                return false;
```
marchermans: So a missing minecraft sound event is marked as not valid.
Did you not specify in your PR message that it lets this slide cause it might be missing due to ExternalFileHelper not seeing those?

marchermans (changes requested):
Generally a good PR, personally I would like to see the AssertionError be something else.
This is not an assert but more like an Illegal or unsupported case, it should be documented properly, and at least have a message when ever it is reached.

Also you are currently marking missing vanilla sounds as invalid. Is this intentional, even though you specified in your PR message that you let it slide, due to missing entries in the file helper.

src/main/java/net/minecraftforge/client/sounds/generators/SoundDefinitionsProvider.java
marchermans: Okey after a discussion on discord, this message can be disregarded I misread the methods intentions and return values.
A bit weirdly named, to be fair. But this works as intended.

TheSilkMiner: After discussion on Discord, I decided to flip the logic so that it is easier to understand. Double negatives aren't that good.

src/main/java/net/minecraftforge/client/sounds/generators/SoundDefinitionsProvider.java
TheSilkMiner: Replaced with a `RuntimeException` with a message outlining the error and added a comment to explain why this situation isn't simply logged and the sound is marked as invalid.

PR #6747, opened 2020-06-07
Another attempt at multi-layer item rendering.

gigaherz:
This is the third internal redesign, but the first one I feel comfortable showing.

## Intended use cases

1. Drawing multi-layer BlockItems in an equivalent way to the placed blocks.
2. Improving the fullbright item support by allowing fullbright quads to be drawn using an unlit render type.
3. Improving custom model support (eg. dynamic bucket models) by allowing models to be drawn without transparency sorting.

## What these use cases require

1. The ability to draw quads in multiple render types. (multi-layer blocks, custom models)
2. The ability to repeat render types in a well-defined order. (layered items)

## The solution

By returning a `List<Pair<IBakedModel, RenderType>>`, the code allows both use cases of the same model being drawn with multiple render types (needed by the existing multi-layer block system), and other models being drawn with a specific render type (needed by layered items with mixed fullbright quads).

### DIFF
--- a/patches/minecraft/net/minecraft/client/renderer/ItemRenderer.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/ItemRenderer.java.patch
@@ -9,7 +9,7 @@
  
        for(Item item : Registry.field_212630_s) {
           if (!field_195411_c.contains(item)) {
-@@ -91,7 +91,7 @@
+@@ -91,9 +91,11 @@
              p_229111_8_ = this.field_175059_m.func_178083_a().func_174953_a(new ModelResourceLocation("minecraft:trident#inventory"));
           }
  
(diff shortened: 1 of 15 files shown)


pupnewfster:
Can this system be used to fix the bucket model loader having the fluid make the bucket transparent?

gigaherz:
Dynamic bucket model is now capable of showing both transparent fluids:
transparent fluid

And illuminated fluids:
illuminated fluid

And multi-layer block models will work as intended when used in the corresponding BlockItem:
multi-layer
(left: vanilla slime block, right: custom test model)

cech12:
Is there a plan to merge it? :)

gigaherz:
A derivative of this was merged into 1.16 a while ago. In order to merge this, we would need to make sure it lines up with the current implementation in 1.16, while also making sure it doesn't introduce any breaking change into the LTS version.
I don't currently have the energy/interest to do that backport.

Edited to add: If someone wants to start from the 1.16 code with this as a reference, and see if they can make a proper backport, feel free!

gigaherz:
Oops I forgot about this. I don't really have any interest in backporting, so no sense keeping this open.

PR #6750, opened 2020-06-08
RFC for 1.16: Make LazyOptional#map/filter return an Optional<T> to indicate they are no longer lazy.

gigaherz:
As a bonus, they also clearly indicate the returned object does in fact not support listeners/invalidation.

### DIFF
--- a/src/main/java/net/minecraftforge/common/BiomeDictionary.java
+++ b/src/main/java/net/minecraftforge/common/BiomeDictionary.java
@@ -16,6 +16,7 @@
  * License along with this library; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
  */
+
 /* Biomes are completely redone in 1.16.2, reevaluate
 package net.minecraftforge.common;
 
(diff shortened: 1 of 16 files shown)


gigaherz:
It has been pointed out to me, that the `map` method is in fact still lazy.
I still believe this change makes sense, as the usefulness of the lazy status of the object returned from `map`  is questionable at best, so I'll leave it up to gather more feedback.

pupnewfster:
> It has been pointed out to me, that the `map` method is in fact still lazy.
> I still believe this change makes sense, as the usefulness of the lazy status of the object returned from `map` is questionable at best, so I'll leave it up to gather more feedback.

Would it make more sense in that case to have two methods, one that maps it to an optional (like this PR changes it to) and one that maps more lazily (like it currently does)?

gigaherz:
Makes sense to me, but I don't know what the naming could be. I'd be OK with keeping the existing behaviour as `lazyMap(Function)` maybe others would prefer the default `map` to be lazy one.

marchermans:
I would prefer this to be as clear as possible:
1) `map(...)` maps directly to a none lazy optional: `Optional<T>`
2) `lazyMap(...)`maps lazily to a lazy optional: `LazyOptional<T>`

My guess would be that currently 99% of the people are already using the current `map` as if it was 1) so they would not need to change anything. Everybody else will probably get a compiler error and it is a small adaptation.

KnightMiner:
I would also appreciate being able to get a regular optional from a lazy one, as there is a lot more functionality for chaining there. Likewise, I think it would be beneficial to have a flatMap method to work with other methods that return optionals (or perhaps lazy optionals)

PR #6752, opened 2020-06-09
RFC for 1.16: make getRegistryName NonNull, add getRegistryNameNullable

ichttt:
Most of the time modders call getRegistryName, they assume that the regname is not null, as it is already in a registry, so there are pointless IDE warnings as you don't check for null, but in reality there is no way it can be null at that stage.
Open for comments.

### DIFF
--- a/patches/minecraft/net/minecraft/block/Block.java.patch
+++ b/patches/minecraft/net/minecraft/block/Block.java.patch
@@ -158,7 +158,7 @@
  
     public String toString() {
 -      return "Block{" + Registry.field_212618_g.func_177774_c(this) + "}";
-+      return "Block{" + getRegistryName() + "}";
++      return "Block{" + getRegistryNameNullable() + "}";
     }
  
     @OnlyIn(Dist.CLIENT)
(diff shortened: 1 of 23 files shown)


tterrag1098:
If we're going to change this, it's worth thinking about a more creative solution. I changed `RegistryObject#get` to nonnull in 1.15 and it mostly caused confusion since the method name did not change.

Perhaps we should go with `getId` to match `RegistryObject` and better fit what the value actually is. Since it's not the "name of the registry" it's the ID of the object.

I would also question how useful `getRegistryNameNullable` really is, since it looks like it's mostly just used to format strings. Couldn't that job be done by `toString`?

gigaherz:
We maybe could leave getRegistryName() as it is now, and make getId() the non-nullable helper?

LordMonoxide:
For brevity's sake, I think `getId()` would be good, and maybe more descriptive. Matching up with `RegistryObject` is also a bonus.

ichttt:
`getId` could easily lead to conflicts in some classes that implement the interface, maybe `getRegistryId` would be better

PR #6757, opened 2020-06-10
Add EndermanAttack Event

Mysterious-Dev:
This RP adds an event that is called when a player looks an enderman in the eye.

This event could be used for example to add conditions so that the enderman does not attack the player.


tterrag1098:
Please just fix this PR and don't open another.

Mysterious-Dev:
Ok

PR #6758, opened 2020-06-10
Added ProjectileImpactEvent.FireworkRocket

DaemonUmbra:
Event is cancellable, causes firework to ignore the current hit and continue on its journey.
Requested by folks on the forums


diesieben07 (changes requested):
> Event is cancellable, causes firework to ignore the current hit and continue on its journey.

This should be the code documentation.

diesieben07 (approved):

LexManos:
One of these days we need to make ProjectileImpactEvent generic. All these subclasses just to change the return type of getEntity is bad.

PR #6759, opened 2020-06-10
[1.15.2] Add function to add items with the same behavior as the pumpkin for enderman

Mysterious-Dev:
This RP adds an event that is called when a player looks an enderman in the eye.

This event could be used for example to add conditions so that the enderman does not attack the player.

This PR is a recreation of an [old one](https://github.com/MinecraftForge/MinecraftForge/pull/6757) which was closed because I could not reopen the old one given the commands I had to do to solve some problem.


src/main/java/net/minecraftforge/event/entity/EndermanAttackEvent.java
```diff
@@ -0,0 +1,31 @@
+package net.minecraftforge.event.entity;
+
+import net.minecraft.entity.Entity;
+import net.minecraft.entity.monster.EndermanEntity;
+import net.minecraft.entity.monster.ZombieEntity;
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraftforge.eventbus.api.Cancelable;
+
+@net.minecraftforge.eventbus.api.Cancelable
```
diesieben07: Why the FQN here?
And please add some documentation: What happens when the event is cancelled? When is this fired?

src/main/java/net/minecraftforge/event/entity/EndermanAttackEvent.java
```diff
@@ -0,0 +1,31 @@
...
+public class EndermanAttackEvent extends EntityEvent {
```
diesieben07: Why does this not extend `LivingEvent`?

src/main/java/net/minecraftforge/event/entity/EndermanAttackEvent.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public EndermanAttackEvent(EndermanEntity entity, PlayerEntity player)
+    {
+        super(entity);
+        this.player = player;
+        this.entity = entity;
+    }
+
+    @Override
+    public Entity getEntity() {
```
diesieben07: You should probably give access to `EndermanEntity` here.

diesieben07 (changes requested):

src/main/java/net/minecraftforge/event/entity/EndermanAttackEvent.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public EndermanAttackEvent(EndermanEntity entity, PlayerEntity player)
+    {
+        super(entity);
+        this.player = player;
+        this.entity = entity;
+    }
+
+
+    public EndermanEntity getEntity() {
```
diesieben07: Missing `@Override`.

diesieben07 (changes requested):

diesieben07 (approved):

LexManos:
Doesn't vanilla already prevent the attack when wearing a pumpkin on your head? Why not hook into that instead of throwing this event?

Mysterious-Devs:
This event was to add a possibility of control over the conditions before the enderman attacks, for example if the player has a level below so much (for example a skill level in capabilty) and well the enderman attacks  otherwise it remains passive

Mysterious-Dev:
@LexManos This event was to add a possibility of control over the conditions before the enderman attacks, for example if the player has a level below so much (for example a skill level in capabilty) and well the enderman attacks  otherwise it remains passive

Mysterious-Devs:
But on the other hand I don't really understand what you mean?

Mysterious-Dev:
So what should I do now?

LexManos:
Address my question. Vanilla already has a mechanic for this, if I'm not mistaken. Why add another hook in a completely different place? I understand you want to have some random logic run. But why does it need to be here. Why does it need to be every tick. Why does it need to be this way instead of vanilla's method?

Mysterious-Dev:
Well this is the basic function of MC to which I added the event but to change so that it is not to execute each tick I do not really see how we could do

pupnewfster:
What Lex is saying that instead of adding an event, modify `EndermanEntity#shouldAttackPlayer` to replace the check for if the worn item is a carved pumpkin with a new method in `IForgeItem` along the lines of `hidesEyesFromEnderman` (probably with a better name than that) and defaulting it to true for the carved pumpkin.

Mysterious-Devs:
Ah, ok I did not understand at the beginning but it is clear.

Mysterious-Devs:
Yes but for example if a modder wants to make sure that the player no longer has the aggro of the enderman if he has a certain level with a capa, well with this method it is not too possible?

Mysterious-Dev:
It's good I changed for a function

Mysterious-Devs:
But there is still a bug, I will try to correct it (the pumpkins are no longer taken into account)

LordMonoxide:
I can't comment on the code directly from my phone for some reason, but the isEndermanMask method on IForgeItem docs are wrong - they reference beacon payment

Mysterious-Dev:
@LordMonoxide It's fix, thanks.

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -619,6 +619,17 @@ default boolean isBeaconPayment(ItemStack stack)
         return Tags.Items.BEACON_PAYMENT.contains(stack.getItem());
     }
 
+    /**
+     * Whether this Item can be used as a payment to activate the vanilla beacon.
```
TeamSpen210: This looks like a copy-paste error:
```suggestion
     * Whether this Item can be used to hide player head for enderman.
```

src/test/resources/META-INF/mods.toml
```diff
@@ -65,3 +65,11 @@ loaderVersion="[28,)"
     modId="stencil_enable_test"
 [[mods]]
     modId="deferred_registry_test"
+[[mods]]
+    modId="ender_mask_test"
+[[dependencies.global_loot_test]]
```
marchermans: I am not sure this makes sense here.
Should that not be dependencies.ender_mask_test?

marchermans (approved):
Generally a good addition, especially functionality wise.
As such nice work.

How ever i have some questions:
1) Why is this turned into an additional method on the ItemStack? Do you have a particular use case?
2) Why is this not additionally enhanced by using a custom Tag which is querried by default?
and
3) Why does the mods.toml change adapt a test mod that you did not touch?

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
Mysterious-Dev: It's change in the last commit

diesieben07:
> Why is this not additionally enhanced by using a custom Tag which is querried by default?

We can't use tags for behavior, that causes issues with vanilla servers.

Mysterious-Dev:
> Why is this turned into an additional method on the ItemStack? Do you have a particular use case?

Because this is also the case for isBeaconPayment method.

Mysterious-Dev:
> Why does the mods.toml change adapt a test mod that you did not touch?

I have a test mod class which adds an item for the test

src/test/resources/META-INF/mods.toml
Mysterious-Dev: It's fix

Mysterious-Devs:
What can i do now ?

patches/minecraft/net/minecraft/entity/monster/EndermanEntity.java.patch
```diff
@@ -16,6 +16,15 @@
     }
  
     protected void func_70088_a() {
+@@ -159,7 +159,7 @@
+ 
+    private boolean func_70821_d(PlayerEntity p_70821_1_) {
+       ItemStack itemstack = p_70821_1_.field_71071_by.field_70460_b.get(3);
+-      if (itemstack.func_77973_b() == Blocks.field_196625_cS.func_199767_j()) {
++      if (itemstack.isEnderMask()) {
```
LexManos: Since we have it, it is probably worth passing in both the enderman entity and player entity to allow more power.

Mysterious-Dev: I don't really understand what you mean by that?

TheCurle: In the function parameters, he wants to pass the Enderman entity as well as the Player entity. That's how I read it.

Mysterious-Dev: So basically do a generic method shouldAttackPlayer?

TheCurle: Hmm. I guess my guess must have been wrong, because I assumed he wanted
 `shouldAttackPlayer(PlayerEntity player, EndermanEntity enderman)`

But then you'd have something like 
`if (entity instanceof EndermanEntity) {
    (EndermanEntity)entity.shouldAttackPlayer(player, (EndermanEntity)entity)`

Passing an object to a function of itself is usually bad practice, so i'm not sure what he meant.

Mysterious-Dev: Yes it's not wrong @TheCurle

Mysterious-Dev: @LexManos , I can't understand what you're suggesting

pupnewfster: Lex is suggesting that you make `isEnderMask` be like `isEnderMask(EndermanEntity enderman, PlayerEntity player)`

Mysterious-Dev: Okay, but what's the use of doing it like this? Because I do not see the possible use in this form.

pupnewfster: Lex's point is that given we have the information anyways we may as well pass it. That way if someone does have use case like they only want specific players to have it work as an ender mask, or only specific endermen (or subspecies of enderman that get added by some mod)

Mysterious-Dev: So basically put just 2 arguments in the function, ok I understand now.

Mysterious-Dev:
It's good, it's done, thank you anyway @pupnewfster

src/test/java/net/minecraftforge/debug/item/EnderMaskTest.java
```diff
@@ -0,0 +1,43 @@
...
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+
+@Mod(EnderMaskTest.MODID)
+public class EnderMaskTest {
+
+
+
+    public static final String MODID = "ender_mask_test";
```
diesieben07, line 17: Extraneous whitespace

src/test/java/net/minecraftforge/debug/item/EnderMaskTest.java
```diff
@@ -0,0 +1,43 @@
...
+import net.minecraftforge.debug.client.model.CompositeModelTest;
+import net.minecraftforge.eventbus.api.IEventBus;
+import net.minecraftforge.fml.RegistryObject;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+
+@Mod(EnderMaskTest.MODID)
+public class EnderMaskTest {
```
diesieben07: Formatting (braces)

src/test/java/net/minecraftforge/debug/item/EnderMaskTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+
+
+    public static final String MODID = "ender_mask_test";
+
+    public static final DeferredRegister<Item> ITEMS = new DeferredRegister<>(ForgeRegistries.ITEMS, MODID);
+
+    public static RegistryObject<Item> ender_mask = ITEMS.register("ender_mask", () ->
+            new ArmorItem(ArmorMaterial.LEATHER, EquipmentSlotType.HEAD, (new Item.Properties().group(ItemGroup.MISC))) {
```
diesieben07: Formatting (braces)

src/test/java/net/minecraftforge/debug/item/EnderMaskTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+                @Override
+                public boolean isEnderMask(ItemStack stack, PlayerEntity player, EndermanEntity endermanEntity) {
```
diesieben07: Formatting (braces)

src/test/java/net/minecraftforge/debug/item/EnderMaskTest.java
```diff
@@ -0,0 +1,43 @@
...
+                    return player.experienceLevel > 10;
+                }
+            }
+    );
+
+    public EnderMaskTest() {
```
diesieben07: Formatting (braces)

diesieben07 (changes requested):

src/test/java/net/minecraftforge/debug/item/EnderMaskTest.java
Mysterious-Devs: It's change

src/test/java/net/minecraftforge/debug/item/EnderMaskTest.java
Mysterious-Dev: It's change

src/test/java/net/minecraftforge/debug/item/EnderMaskTest.java
Mysterious-Dev: It's fix in the last commit

Issue #6756, opened 2020-06-10
ModDiscoverer assumes Forge implementations


jamierocks:
Forge's [ModDiscoverer](https://github.com/MinecraftForge/MinecraftForge/blob/5dbea9ea8dbda01ad11d91a713eb29cc973c10eb/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModDiscoverer.java) and surrounding classes (e.g. [LoadingModList](https://github.com/MinecraftForge/MinecraftForge/blob/1.15.x/src/fmllauncher/java/net/minecraftforge/fml/loading/LoadingModList.java)) relies on implementations of interfaces from [Forge SPI's language providers](https://github.com/MinecraftForge/ForgeSPI/tree/a8b4d9299e7b9ad59edc9d78ae342e3a855d725c/src/main/java/net/minecraftforge/forgespi/language) being the Forge implementations.

I'm still understanding the whole system, however this is no trivial change to make - as implementation details of Forge's implementation are relied on:
- [ModFile::identifyMods()](https://github.com/MinecraftForge/MinecraftForge/blob/5dbea9ea8dbda01ad11d91a713eb29cc973c10eb/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModDiscoverer.java#L116)
- [ModFileInfo::getFile()](https://github.com/MinecraftForge/MinecraftForge/blob/5dbea9ea8dbda01ad11d91a713eb29cc973c10eb/src/fmllauncher/java/net/minecraftforge/fml/loading/LoadingModList.java#L85)
- possible more, I found these from a cursory look

Resolving this may well require additions to Forge SPI, and perhaps require happening on a version bump (perhaps 1.16?).

Effectively the following changes needs to be made to resolve this:
- ModFile -> IModFile
- ModFileInfo -> IModFileInfo

cpw:
Why? Why don't you use the language loader system and feed mods in that way? It would probably help quite a LOT if you explained what on earth you are trying to do... This isn't an extension point in FML.

jamierocks:
> Why?

I was experimenting with implementing a [IModLocator](https://github.com/MinecraftForge/ForgeSPI/blob/master/src/main/java/net/minecraftforge/forgespi/locating/IModLocator.java), to use a custom mod metadata format. i.e. a mod that doesn't use `mods.toml`.

> Why don't you use the language loader system and feed mods in that way?

[...]

PR #6761, opened 2020-06-11
Allow mods to override ModFileInfo

jamierocks:
This exists as a stop-gap solution (somewhat) to GH-6756 by allowing
services to override the one hold-out that cannot be overriden.

This is only a solution to a symptom, and not a solution to the actual
issue though - GH-6756 is still absolutely valid.

### DIFF
--- a/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
+++ b/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
@@ -47,7 +47,7 @@ public class ModFileInfo implements IModFileInfo
     private final List<IModInfo> mods;
     private final Map<String,Object> properties;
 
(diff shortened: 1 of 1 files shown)


cpw:
This will break a critical classloading barrier and will not be accepted, sorry. ModFileInfo is an internal class. The interfaces are there so that external language loaders and other things that plug in to FML at that level, don't have to hard depend on internal implementations, but rather interfaces.

jamierocks:
> This will break a critical classloading barrier and will not be accepted, sorry.

Right, in that case - perhaps its worth giving `ModFile` and co package-private constructors too, so they can't be extended outside of Forge?

> The interfaces are there so that external language loaders and other things that plug in to FML at that level, don't have to hard depend on internal implementations, but rather interfaces.

I know, currently GH-6756 prevents that from being the reality though - I tried to be clear that this wasn't much of a solution, but would allow something to be done in the meantime (though if this is going to mess with classloading, then not appropriate).

PR #6762, opened 2020-06-12
Fixed debug.log not being generated in dev

DaemonUmbra:
Not 100% sure why the debug.log doesn't generate during dev, but I would assume it has to do with the xml not being loaded or not being loaded soon enough, as manually specifying it as a property fixes the issue.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -261,7 +261,9 @@ project(':forge') {
                 environment 'FORGE_SPEC', SPEC_VERSION
                 environment 'FORGE_VERSION', project.version.substring(MC_VERSION.length() + 1).toString()
                 environment 'LAUNCHER_VERSION', SPEC_VERSION
+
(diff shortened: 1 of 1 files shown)


marchermans (approved):
Nice catch.
This is only for ForgeDev-Time correct? Not for the ModDev-Time.

marchermans:
@DaemonUmbra Is this still relevant for 1.16?

DaemonUmbra:
Yes it does.
But I'm not so sure this is the "fix" I want to submit anymore.

TheCurle:
Take your time, there seem to be a few ways to do this.

TheCurle:
You've taken some time, I assume you've figured out how you want to take this forward? :P

DaemonUmbra:
No actually... the issue seems to be inconsistent

PR #6765, opened 2020-06-12
[1.15.x] DataGenerator for LootTable, replacing shears item to shears tag

yanny7:
Created ForgeLootTableProvider that replace matching item to tag. Flexible solution with possibility to extend.

Referenced in #6514, #6624

### DIFF
--- a/src/main/java/net/minecraftforge/common/data/ForgeLootTableProvider.java
+++ b/src/main/java/net/minecraftforge/common/data/ForgeLootTableProvider.java
@@ -1,0 +1,178 @@
+package net.minecraftforge.common.data;
+
+import com.mojang.datafixers.util.Pair;
+import net.minecraft.advancements.criterion.ItemPredicate;
+import net.minecraft.data.DataGenerator;
+import net.minecraft.data.LootTableProvider;
+import net.minecraft.item.Item;
+import net.minecraft.item.Items;
+import net.minecraft.tags.Tag;
+import net.minecraft.util.ResourceLocation;
+import net.minecraft.world.storage.loot.*;
+import net.minecraft.world.storage.loot.conditions.Alternative;
+import net.minecraft.world.storage.loot.conditions.ILootCondition;
+import net.minecraft.world.storage.loot.conditions.Inverted;
+import net.minecraft.world.storage.loot.conditions.MatchTool;
+import net.minecraftforge.common.Tags;
+
+import java.lang.reflect.Field;
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+public class ForgeLootTableProvider extends LootTableProvider {
+
+    public ForgeLootTableProvider(DataGenerator gen) {
+        super(gen);
+    }
+
+    @Override
+    protected void validate(Map<ResourceLocation, LootTable> map, ValidationTracker validationtracker) {
+        // do not validate against all registered loot tables
+    }
+
+    @Override
+    protected List<Pair<Supplier<Consumer<BiConsumer<ResourceLocation, LootTable.Builder>>>, LootParameterSet>> getTables() {
+        return super.getTables().stream().map(pair -> {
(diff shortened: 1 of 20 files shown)


src/main/java/net/minecraftforge/common/data/ForgeLootTableProvider.java
```diff
@@ -0,0 +1,178 @@
...
+        return false;
+    }
+
+    private boolean checkMatchTool(MatchTool lootCondition, Item expected) {
+        ItemPredicate predicate = getField(MatchTool.class, lootCondition, 0);
+        Item item = getField(ItemPredicate.class, predicate, 4);
+        return item != null && item.equals(expected);
+    }
+
+    private <T, R> void setField(Class<T> clz, T inst, int index, R value) {
```
percivalalb: Using access transformers might be better here, but involves removing final keyword

yanny7: I dont think it will be better, if that changes will be visible for modders

LexManos: probably better to add a getter, and then replace the whole Inverted instance instead of just the term inside.

yanny7: Removed setField method as proposed

src/main/java/net/minecraftforge/common/data/ForgeLootTableProvider.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    private boolean checkMatchTool(MatchTool lootCondition, Item expected) {
+        ItemPredicate predicate = getField(MatchTool.class, lootCondition, 0);
+        Item item = getField(ItemPredicate.class, predicate, 4);
+        return item != null && item.equals(expected);
+    }
+
+    @SuppressWarnings("unchecked")
+    private <T, R> R getField(Class<T> clz, T inst, int index)
```
ichttt: Depending on field order doesn't feel very safe, is there a reason this is used over ObfuscationReflectionHelper or a simple Access Transformer?

src/main/java/net/minecraftforge/common/data/ForgeLootTableProvider.java
```diff
@@ -0,0 +1,162 @@
...
+
+import java.lang.reflect.Field;
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+public class ForgeLootTableProvider extends LootTableProvider {
```
ichttt, line 29: Should probably note that this is used for shears only right now

ichttt (changes requested):
Some thoughts on the data generator

src/main/java/net/minecraftforge/common/data/ForgeLootTableProvider.java
TheSilkMiner: That's also what's happening in `ForgeRecipeProvider`, where obtaining the list or map of ingredients relies on the field ordering (why?). There should be a common consensus on the approach to follow.
IMO `ObfuscationReflectionHelper` should be preferred over ATs (since these data Gens don't run at runtime and we don't really need to expose data) and field ordering.

yanny7: I was using `ForgeRecipeProvider` as example, but I can replace that by method from `ObfuscationReflectionHelper`

src/main/java/net/minecraftforge/common/data/ForgeLootTableProvider.java
yanny7: Added comment

malte0811:
This PR has broken any mod still using the `LootTableLoadEvent` to add drops to e.g. grass. See https://github.com/BluSunrize/ImmersiveEngineering/issues/4046. I'm moving IE to global loot modifiers now, but I'm not sure whether there are cases where the event has to be used (in which case this would be an actual problem).

Issue #6755, opened 2020-06-10
Wolf pathing finding does not match vanilla


percivalalb:
**Minecraft Version:** 1.15.2

**Forge Version:** 32.2.9

**Steps to Reproduce:**
 1. Tame a vanilla wolf
 2. Walk away from the wolf so that the direct path between it and you crosses lava
 3. Watch the dog make no attempt to path around and kill itself

**Description of issue:**
Wolves in forge make no attempt to avoid lava whereas vanilla wolves will path around it.

Forge: https://gfycat.com/lankysnarlingenglishpointer
Vanilla: https://gfycat.com/tinyaggravatingequine

I assume this applies to other entity's pathing too.

jamierocks:
perhaps something to do with the patches in https://github.com/MinecraftForge/MinecraftForge/blob/1.15.x/patches/minecraft/net/minecraft/pathfinding/WalkNodeProcessor.java.patch ?

percivalalb:
https://github.com/MinecraftForge/MinecraftForge/blob/5dbea9ea8dbda01ad11d91a713eb29cc973c10eb/patches/minecraft/net/minecraft/pathfinding/WalkNodeProcessor.java.patch#L17
Is missing the change in x, y, z arguments, should be 
`func_189553_b(p_193578_0_, i + p_193578_1_, j + p_193578_2_, k + p_193578_3_);`

This does not fix this issue but still needs fixing. I have found that removing this part of the patch does fix it.
https://github.com/MinecraftForge/MinecraftForge/blob/1.14.x/patches/minecraft/net/minecraft/pathfinding/WalkNodeProcessor.java.patch#L52-L53

percivalalb:
I think the problem is that `WalkNodeProcessor.getAiPathNodeType` returns `PathNodeType.LAVA` in vanilla and returns `PathNodeType.DANGER_FIRE`  in forge (this is when the block is a fluid contained in the lava tag)

PR #6767, opened 2020-06-12
Fix entities pathing through lava

percivalalb:
Related Issue: https://github.com/MinecraftForge/MinecraftForge/issues/6755

In vanilla, for a lava block, `func_189553_b/getPathNodeTypeRaw` returns `PathNodeType.LAVA` whereas in Forge the default implementation in `IForgeBlock#getAiPathNodeType` causes it returns `PathNodeType.DANGER_FIRE`.

`PathNodeType.DANGER_FIRE` only increases the cost of the path where as `PathNodeType.LAVA` classes the path as untraveled, hence why in forge entities will try to cross lava and vanilla not.

I have changed `IForgeBlock#getAiPathNodeType` to always return null by default so that vanilla logic takes over for fire and lava. This doesn't change the behaviour of fire but does fix lava.

The fix in `func_193578_a/checkNeighborBlocks` is because this method only wants to know if there are dangers nearby not if the blocks nearby are impassable. This change from `PathNodeType.LAVA` to `PathNodeType.DANGER_FIRE` makes this patch match vanilla for lava fluids.

A side note: `IForgeBlock#getAiPathNodeType` should possibly have an extra boolean param (`isNeighbour`) so that the returned `PathNodeType` can be dependent on if it is being check as a neighbouring block or the actual block. A use case would be for lava which wants to return `PathNodeType.LAVA` normally and `PathNodeType.DANGER_FIRE` when a neighbouring block.

### DIFF
--- a/patches/minecraft/net/minecraft/pathfinding/WalkNodeProcessor.java.patch
+++ b/patches/minecraft/net/minecraft/pathfinding/WalkNodeProcessor.java.patch
@@ -1,6 +1,6 @@
 --- a/net/minecraft/pathfinding/WalkNodeProcessor.java
 +++ b/net/minecraft/pathfinding/WalkNodeProcessor.java
-@@ -383,16 +383,9 @@
+@@ -383,15 +383,11 @@
              for(int j = -1; j <= 1; ++j) {
                 for(int k = -1; k <= 1; ++k) {
                    if (i != 0 || k != 0) {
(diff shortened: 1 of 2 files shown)


hYdos (approved):

hYdos:
this is important. Hope it doesn't go stale

LexManos:
This changes the behavior of anything that used burnable just fucking make lava return lava bother changes are needed.

PR #6772, opened 2020-06-14
Add Fire Source Tag to add fire source block

Mysterious-Dev:
This PR add a fire source tag to allow to specify the blocks like the netherrack which allows the fire to burn indefinitely.


LexManos:
We are not adding tags for in game logic. They are to be used for recipes.
You can already override this extension method in your own block.

PR #6775, opened 2020-06-15
[1.15.2] Added a new event called "RenderWorldPreTranslucentEvent"

nanorover59:
This event is very similar to RenderWorldLastEvent, but is instead called right before translucent blocks are rendered in the WorldRenderer class. The reason for this being that objects rendered from RenderWorldLastEvent are always invisible behind translucent blocks, and using an event called before these blocks are rendered will not lead to this rendering problem.
A while ago I encountered that problem while working on a project involving rendering "sub-world" objects made of blocks but separate from main world chunks. One suggestion I got on Forge Forums was to submit a pull request adding a new event.
Included in this pull request is a small example mod that demonstrates rendering a single "block" in the world with this new event.

### DIFF
--- a/patches/minecraft/net/minecraft/block/StemBlock.java.patch
+++ b/patches/minecraft/net/minecraft/block/StemBlock.java.patch
@@ -1,6 +1,6 @@
 --- a/net/minecraft/block/StemBlock.java
 +++ b/net/minecraft/block/StemBlock.java
-@@ -38,22 +38,24 @@
+@@ -40,22 +40,24 @@
  
     public void func_225534_a_(BlockState p_225534_1_, ServerWorld p_225534_2_, BlockPos p_225534_3_, Random p_225534_4_) {
        super.func_225534_a_(p_225534_1_, p_225534_2_, p_225534_3_, p_225534_4_);
@@ -30,7 +30,7 @@
           }
  
        }
-@@ -98,4 +100,10 @@
+@@ -101,4 +103,10 @@
     public StemGrownBlock func_208486_d() {
        return this.field_149877_a;
     }
--- a/patches/minecraft/net/minecraft/block/StemBlock.java.patch
+++ b/patches/minecraft/net/minecraft/block/StemBlock.java.patch
@@ -1,6 +1,6 @@
 --- a/net/minecraft/block/StemBlock.java
 +++ b/net/minecraft/block/StemBlock.java
-@@ -40,22 +40,24 @@
+@@ -38,22 +38,24 @@
  
     public void func_225534_a_(BlockState p_225534_1_, ServerWorld p_225534_2_, BlockPos p_225534_3_, Random p_225534_4_) {
        super.func_225534_a_(p_225534_1_, p_225534_2_, p_225534_3_, p_225534_4_);
@@ -30,7 +30,7 @@
           }
  
        }
-@@ -101,4 +103,10 @@
+@@ -98,4 +100,10 @@
     public StemGrownBlock func_208486_d() {
        return this.field_149877_a;
     }
--- a/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
+++ b/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
@@ -29,6 +29,7 @@ import static org.lwjgl.opengl.GL11.GL_VERTEX_ARRAY;
 import static org.lwjgl.opengl.GL11.glColorPointer;
 import static org.lwjgl.opengl.GL11.glDisableClientState;
 import static org.lwjgl.opengl.GL11.glEnableClientState;
+import static org.lwjgl.opengl.GL11.glMultMatrixf;
 import static org.lwjgl.opengl.GL11.glNormalPointer;
 import static org.lwjgl.opengl.GL11.glTexCoordPointer;
 import static org.lwjgl.opengl.GL11.glVertexPointer;
@@ -40,13 +41,21 @@ import java.io.File;
 import java.lang.reflect.Field;
 import java.nio.Buffer;
 import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Random;
 import java.util.Set;
 import java.util.stream.Stream;
 
 import javax.annotation.Nonnull;
 
+import net.minecraft.client.renderer.*;
+import net.minecraft.client.settings.KeyBinding;
+import net.minecraftforge.client.event.RenderHandEvent;
+import net.minecraftforge.client.model.pipeline.LightUtil;
+import net.minecraftforge.fml.loading.progress.StartupMessageManager;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.core.async.ThreadNameCachingStrategy;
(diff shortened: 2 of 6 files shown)


src/main/java/net/minecraftforge/client/event/RenderWorldPreTranslucentEvent.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * This event is called immediately before translucent blocks are rendered in the WorldRenderer class.
+ * 
+ * It is useful for rendering in a similar manner to RenderWorldLastEvent without the result being
+ * invisible behind translucent blocks such as stained glass or water.
+ * However, due to the nature of translucent block rendering, translucent blocks in the world
+ * will be invisible behind anything translucent rendered from this event.
+ */
+
+public class RenderWorldPreTranslucentEvent extends net.minecraftforge.eventbus.api.Event
```
covers1624: A fully qualified name for `Event` isn't necessary here.

src/test/java/net/minecraftforge/debug/client/rendering/RenderWorldPreTranslucentEventTest.java
```diff
@@ -0,0 +1,84 @@
...
+        Impl renderTypeBuffer = IRenderTypeBuffer.getImpl(Tessellator.getInstance().getBuffer());
+        
+        // Render the block at the coordinates (0, 128, 0).
+        renderBlock(matrixStack, renderTypeBuffer, world, blockState, new BlockPos(0, 128, 0), projectedView, new Vec3d(0.0, 128.0, 0.0));
+
+        renderTypeBuffer.finish();
+    }
+    
+    @SuppressWarnings("deprecation")
+	public static void renderBlock(MatrixStack matrixStack, Impl renderTypeBuffer, World world, BlockState blockState, BlockPos logicPos, Vec3d projectedView, Vec3d renderCoordinates)
```
covers1624: Tabs, use spaces.

src/test/java/net/minecraftforge/debug/client/rendering/RenderWorldPreTranslucentEventTest.java
```diff
@@ -0,0 +1,84 @@
...
+            return;
+        
+        // Get instances of the classes required for a block render.
+        MinecraftServer server = Minecraft.getInstance().getIntegratedServer();
+        World world = DimensionManager.getWorld(server, DimensionType.OVERWORLD, false, true);
+        MatrixStack matrixStack = event.getMatrixStack();
+
+        // Get the projected view coordinates.
+        @SuppressWarnings("resource")
+		Vec3d projectedView = Minecraft.getInstance().gameRenderer.getActiveRenderInfo().getProjectedView();
```
covers1624: Weird indent

covers1624 (changes requested):
Looks good! Minor formatting things i noticed.

nikita488:
In my opinion, it will be much more useful to have an event called before everything rather than just translucent, like post it right after IRenderTypeBuffer.Impl is created in updateCameraAndRender method. Also, it will be useful to have ActiveRenderInfo, ClippingHelperImpl, and IRenderTypeBuffer as getters in this event.

covers1624:
I think having an event _before_ everything might be useful in some cases, but directly before translucent is definitely more useful in general. You do bring up a good point,
we already have `ActiveRenderInfo` context where the Event is fired, which should probably be added to the Event for convenience.

nanorover59:
I agree that an event called before everything is rendered would still have some uses, but for the project I was working on before making the pull request it had some issues that rendering right before translucent blocks did not.

src/main/java/net/minecraftforge/client/event/RenderWorldPreTranslucentEvent.java
nanorover59: Fixed this and the other two issues.

patches/minecraft/net/minecraft/block/StemBlock.java.patch
```diff
@@ -30,7 +30,7 @@
           }
  
        }
-@@ -101,4 +103,10 @@
+@@ -98,4 +100,10 @@
```
covers1624: I'd just discard these changes since you didn't touch anything here.

src/main/java/net/minecraftforge/client/event/RenderWorldPreTranslucentEvent.java
```diff
@@ -0,0 +1,86 @@
...
+    {
+        return context;
+    }
+
+    public MatrixStack getMatrixStack()
+    {
+        return mat;
+    }
+    
+    public Impl getRenderTypeBuffers()
```
covers1624: I'd use `IRenderTypeBuffer.Impl` since this is an inner class reference. (same with field and parameter declarations)

src/test/java/net/minecraftforge/debug/client/rendering/RenderWorldPreTranslucentEventTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+        // Get instances of the classes required for a block render.
+        MinecraftServer server = Minecraft.getInstance().getIntegratedServer();
+        World world = DimensionManager.getWorld(server, DimensionType.OVERWORLD, false, true);
+        MatrixStack matrixStack = event.getMatrixStack();
+        Impl renderTypeBuffers = event.getRenderTypeBuffers();
+
+        // Get the projected view coordinates.
+        @SuppressWarnings("resource")
+        Vec3d projectedView = Minecraft.getInstance().gameRenderer.getActiveRenderInfo().getProjectedView();
```
covers1624: We have ActiveRenderInfo context where the event is fired, why not just pass it through from there?

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -129,20 +122,21 @@
...
 import net.minecraftforge.client.model.animation.Animation;
-import net.minecraftforge.client.model.data.EmptyModelData;
-import net.minecraftforge.client.model.pipeline.QuadGatheringTransformer;
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.common.model.TransformationHelper;
 import net.minecraftforge.eventbus.api.Event;
 import net.minecraftforge.fml.ModLoader;
 import net.minecraftforge.fml.VersionChecker;
 import net.minecraftforge.fml.client.registry.ClientRegistry;
+import net.minecraftforge.fml.loading.progress.StartupMessageManager;
```
covers1624: It might be worth trying to avoid changing so many imports here, they aren't really important changes and could throw off other PR's into this class.

covers1624 (changes requested):
Overall formatting still good, just a few changes.

patches/minecraft/net/minecraft/block/StemBlock.java.patch
nanorover59: This has been done.

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
nanorover59: I started again here with the original class and only added the two imports I needed.

src/test/java/net/minecraftforge/debug/client/rendering/RenderWorldPreTranslucentEventTest.java
nanorover59: The event now has direct access to WorldRenderer's ActiveRenderInfo.

covers1624 (approved):
Looks good

TheCurle:
If this is still a wanted feature, we would like you to update it to 1.16 for consideration.

Nightenom:
Note to Curle: #7225 is extended 1.16 version and should cover all needs of the author's problem:
> A while ago I encountered that problem while working on a project involving rendering "sub-world" objects made of blocks but separate from main world chunks.

TheCurle:
Hm. Okay, let's update that one and go from there.

TheCurle:
This PR has a lot of conceptual issues, those which led to the creation of #7661 among others.

The more versatile approach would be far too much work to properly integrate into 1.15 at this stage, so i'm going to close this and focus on the better approach for 1.16.

PR #6777, opened 2020-06-15
Post SoundLoadEvent on mod bus

josephtarbit:
Currently when `SoundLoadEvent` is initially called upon Minecraft's start-up, modders cannot handle the event because it is posted on the forge bus which isn't started until Minecraft finishes loading. This pull request simply modifies the posting of `SoundLoadEvent` to use the mod bus instead of the forge bus.


ichttt:
This is a bit of a problem, as this event is also fired when sounds reload, and this breaks mods that depend on it being called on sound reload... A possible solution would be to fire on both busses, but that's not great either

josephtarbit:
That’s what I thought, probably best to wait and merge it into 1.16.

Aeronica:
The last I checked some of the sound events are not hooked up anymore. Sound events need some cleanup. I've done work arounds so I can use a custom codec, but it's a hacky. The use the lambda's in the new sound code in certain places may make sit a challenge to add an events. For instance to specify you own codec, and submit a stream to it. Things I could before the sound rewrite/removal of the old Paulscode library. I'd be willing to put time in to improve sound events in the 1.15/1.16.

LexManos:
The mod event bus should be for lifecycle events only. Not random MC events.
Maybe the SoundLoadEvent needs to be redesigned to something like the Registry event.

josephtarbit:
> something like the Registry event.

What exactly do you have in mind when you say this? I took inspiration from `ParticleFactoryRegisterEvent` which seems to be posted on the mod bus too but if there's a more ideal solution I'm all ears.

josephtarbit:
This has been merged into 1.16 now #6955 so any updates?

TheCurle:
It feels like this feature would be better as two separate events; loading from data/resource pack during startup, which could be on the mod event bus, and a _re_loading event posted on the Forge bus. Thereby, the worries about breaking existing mods is irrelevant.

The PR was merged for 1.16, but a breaking change like this is not a good idea for the LTS version.
Some reconsideration is likely needed here.

PR #6782, opened 2020-06-17
FluidAttributes: switch to imperial bucket.

hanetzer:
[Forum Thread](https://forums.minecraftforge.net/topic/86964-bucket-volume-or-imperial-vs-metric/)
So. This one may be controversial. A fairly large amount of fluid use is for
'molten $foo', where foo can be vanilla or modded ores, and things of a similar
bent. 1000 units of fluid per bucket don't work terribly well mathmatically for
this use.

Metric is a great system, for the most part, but within minecraft crafting tends
towards a base-9 or base-4 system of crafting. If 1000mb of molten iron is cast
to 1 iron block, there is no even way to divide that evenly into 9 ingots or 81
nuggets. Using 1296, we have 1296mb for a 3x3 block, 144mb for an ingot, and
16mb for a nugget.

With 1.16 to be released soon it seems like now would be the time to make a
clean break, since modders are already going to be updating (hopefully).

Signed-off-by: Marty E. Plummer <hanetzer@startmail.com>

### DIFF
--- a/src/main/java/net/minecraftforge/fluids/FluidAttributes.java
+++ b/src/main/java/net/minecraftforge/fluids/FluidAttributes.java
@@ -58,7 +58,7 @@ import net.minecraft.world.biome.BiomeColors;
  */
 public class FluidAttributes
 {
-    public static final int BUCKET_VOLUME = 1000;
+    public static final int BUCKET_VOLUME = 1296;
 
     private String translationKey;
 


Lanse505:
Personal Opinion:

TLDR; While the intention is good, IMO this should have been a discussion issue rather than a PR. This has implications on how we decide fluid amounts and tank sizes in mods in the community as a whole. And thusly the "granularity" of fluid amounts.


Full;
While the intention is good, personally I feel this should be brought up as a discussion topic instead of as a straight PR change. Changing the scaled definition of what a bucket of fluid "is" in terms of volume, has farther reaching implications for mods than just "it scales properly" between nuggets, ingots, blocks.

This is in theory a "fundemental" change in theory to the average size of tanks, outputs of fluid recipes, etc. and thus is something that should be discussed before a PR aimed at targetting the "problem" should be done.


Caveat:
Most likely no solid consensus will be reached on this, but a discussion is always healthy to have. The discussion of "what's the logical volume of a bucket of fluid" has been had many a time, in both the forge community as well as other communities. Everyone has an opinion.
If a discussion is brought about, please lets keep it civil :P 


P.s.
If we follow the logic above, some values we could establish is (some based on vanilla behaviours):

1 Nugget = 16mB
1 Ingot = 144mB
1 Bottle = 432mB (1/3 Bucket)
1 Bucket = 1296mB
1 Block = 1296mB

hanetzer:
> If we follow the logic above, some values we could establish is (some based on vanilla behaviours):
> 
> 1 Nugget = 16mB
> 1 Ingot = 144mB
> 1 Bottle = 432mB (1/3 Bucket)
> 1 Bucket = 1296mB
> 1 Block = 1296mB

And one glass pane is 486mb, assuming we treat glass as a 3x3 block and not a 2x2
(those are 576mb assuming their component parts are ingot-ish)

LordMonoxide:
One major drawback, of course, is the mental math aspect, and this alone might be a deal breaker

FreneticScribbler:
What's the actual benefit of this over just treating 1 Bucket = 1000mB and 1 (3x3) Block = 1296mB, as is usual? (Other than a block fitting nicely into a vanilla bucket I suppose)

Lanse505:
> 
> 
> What's the actual benefit of this over just treating 1 Bucket = 1000mB and 1 (3x3) Block = 1296mB, as is usual? (Other than a block fitting nicely into a vanilla bucket I suppose)

So I think the logic is that currently 1 block is 1 bucket.
Which is 1000mB.

1000 isn't really nicely dividable down into 9 (aka ingots).
since 1000/9 is 111.11111... mB

Same goes if you divide that by nine again, which turns into 12.345679... mB

Meaning there is currently not a good granularity to how items/blocks and fluids interact in terms of scale.

With what he suggests, the theory is that we'd get a better granularity in terms of converting amounts up and down and in-between. But we'd also lose a bit of granularity since we're not counting in "clean" singular number anymore in theory.

PrincessOfEvil:
On the notion of the actual number, wouldn't it make more sense to make it
an antiprime, like 720 or even 5040? 1296 works with 2 and 3 but falls
apart when it has to work in a base 10 system.

>

hanetzer:
> One major drawback, of course, is the mental math aspect, and this alone might be a deal breaker

If by 'mental math' aspect you mean the ease of doing base10 conversions by moving
the decimal point, nothing is stopping us from using 1 kilobucket for 1000x 1296 buckets.

LexManos:
👋 We use the issue tracker exclusively for final bug reports and feature requests.
  However, this issue appears to be better suited for the
  [Forge Support Forums](https://forums.minecraftforge.net/forum/18-support-bug-reports/).
  Please create a new topic on the support forum with this issue, and the conversation can
  continue there.

ExpensiveKoala:
I'm pretty sure they meant like 8 buckets = 8000mb instead of 8 buckets = 10368mb

hanetzer:
[Forum thread started, OP linked](https://forums.minecraftforge.net/topic/86964-bucket-volume-or-imperial-vs-metric/)

PR #6785, opened 2020-06-19
[1.14.x] Extendible categories in recipe book

yanny7:
Basicaly allows modders to add recipe book into custom machines and add tabs into vanilla recipe book

- custom recipe book category can be registered via forge registries.
- constructor have `IRecipeType<?>` to set where recipe book category belongs
- modder can set if category is `search` - no filtering is applied for displayed recipes
- `setPredicate(IRecipe<?>)` non-search category with filtering
- `setIcon(ItemStack...)` recipe book category icon
```
    @SubscribeEvent
    public static void registerRecipeBookCategories(RegistryEvent.Register<RecipeBookCategories> event) {
        event.getRegistry().register(new RecipeBookCategories(TestRecipe.millstone).setSearch().setIcon(new ItemStack(Items.BELL)).setRegistryName("base_crafting_book_test", "test"));
    }
```
- Custom machine container must extends `RecipeBookContainer<C extends IInventory>` 
- Custom machine gui just render and handle recipe book as in e.g. AbstractFurnaceScreen


patches/minecraft/net/minecraft/client/util/RecipeBookCategories.java.patch
```diff
@@ -0,0 +1,151 @@
...
+-   FURNACE_MISC(new ItemStack(Items.field_151129_at), new ItemStack(Items.field_151166_bC)),
+-   BLAST_FURNACE_SEARCH(new ItemStack(Items.field_151111_aL)),
+-   BLAST_FURNACE_BLOCKS(new ItemStack(Blocks.field_150450_ax)),
+-   BLAST_FURNACE_MISC(new ItemStack(Items.field_151037_a), new ItemStack(Items.field_151149_ai)),
+-   SMOKER_SEARCH(new ItemStack(Items.field_151111_aL)),
+-   SMOKER_FOOD(new ItemStack(Items.field_151147_al)),
+-   STONECUTTER(new ItemStack(Items.field_221782_da)),
+-   CAMPFIRE(new ItemStack(Items.field_151147_al));
++@net.minecraftforge.registries.ObjectHolder("minecraft")
++public class RecipeBookCategories extends net.minecraftforge.registries.ForgeRegistryEntry<RecipeBookCategories> {
```
TheSilkMiner: This breaks binary compatibility

patches/minecraft/net/minecraft/client/gui/recipebook/RecipeBookGui.java.patch
```diff
@@ -1,6 +1,25 @@
 --- a/net/minecraft/client/gui/recipebook/RecipeBookGui.java
 +++ b/net/minecraft/client/gui/recipebook/RecipeBookGui.java
-@@ -89,7 +89,7 @@
+@@ -34,6 +34,8 @@
+ import net.minecraftforge.api.distmarker.Dist;
+ import net.minecraftforge.api.distmarker.OnlyIn;
+ 
++import net.minecraft.client.gui.widget.button.Button.IPressable;
```
TheSilkMiner: Don't use `import`s in patches

patches/minecraft/net/minecraft/client/gui/recipebook/RecipeBookGui.java.patch
```diff
@@ -18,3 +106,63 @@
...
+       }
+ 
+    }
++
++   private void updateButtons() {
++      this.up.visible = this.offset > 0 && this.field_193018_j.size() > 5;
++      this.down.visible = this.field_193018_j.size() > 5 && this.offset < this.field_193018_j.size() - 5;
++   }
++
++   private static class UpDownButton extends net.minecraft.client.gui.widget.button.Button {
```
TheSilkMiner: You should move this to another file IMO.
Patches are supposed to be minimal and adding classes is kinda against that

patches/minecraft/net/minecraft/inventory/container/AbstractFurnaceContainer.java.patch
```diff
@@ -0,0 +1,12 @@
+--- a/net/minecraft/inventory/container/AbstractFurnaceContainer.java
++++ b/net/minecraft/inventory/container/AbstractFurnaceContainer.java
+@@ -171,4 +171,9 @@
+    public boolean func_217061_l() {
+       return this.field_217064_e.func_221476_a(0) > 0;
+    }
++
++   @Override
++   public IRecipeType<?> getRecipeType() {
```
TheSilkMiner: You should move this over to the class where you added this method and use `instanceof` checks. It's not the best OOP design, but it saves up on patch count. Same for `PlayerContainer`, `WorkbenchContainer`

patches/minecraft/net/minecraft/inventory/container/RecipeBookContainer.java.patch
```diff
@@ -1,11 +1,37 @@
...
+ import net.minecraft.inventory.IInventory;
+ import net.minecraft.item.crafting.IRecipe;
++import net.minecraft.item.crafting.IRecipeType;
+ import net.minecraft.item.crafting.RecipeItemHelper;
+ import net.minecraft.item.crafting.ServerRecipePlacer;
+ import net.minecraftforge.api.distmarker.Dist;
+ import net.minecraftforge.api.distmarker.OnlyIn;
++import net.minecraftforge.registries.ForgeRegistries;
+ 
++import java.util.stream.Collectors;
```
TheSilkMiner: Don't use imports in patches

patches/minecraft/net/minecraft/client/util/RecipeBookCategories.java.patch
yanny7: Any idea how to do it without breaking compatibility?

patches/minecraft/net/minecraft/client/gui/recipebook/RecipeBookGui.java.patch
yanny7: removed imports from patches

patches/minecraft/net/minecraft/client/gui/recipebook/RecipeBookGui.java.patch
yanny7: Moved UpDownButton to new class

patches/minecraft/net/minecraft/inventory/container/AbstractFurnaceContainer.java.patch
yanny7: implemented default body for getRecipeType

patches/minecraft/net/minecraft/inventory/container/RecipeBookContainer.java.patch
yanny7: Fixed

patches/minecraft/net/minecraft/client/util/RecipeBookCategories.java.patch
yanny7: Fixed ABI

TheSilkMiner (changes requested):
There are still some breaking changes within this PR:
- `RecipeTabToggleWidget#field_193921_u`'s type has been changed, along with the constructor references
- `RecipeTabToggleWidget#func_201503_d`'s return type has been changed
- `ClientRecipeBook#field_197931_e` had its type changed
- `ClientRecipeBook#func_216767_a` had its parameters changed
- `ClientRecipeBook#func_202891_a` was deleted and replaced with another method

Also, I am not a fan of replacing constructor arguments inside an `enum`, since that leads to a much bigger set of changes in patches.

The way I'd approach this is rather have the recipe tab toggle widget category remain the same, but redirect all usages towards another `IForgeRecipeBookCategory<?>` , which gets assigned by vanilla in the constructor. Also, provide a secondary constructor, and patch `RecipeBookGui#initSearchBar` to use the new one. which would set `field_193921_u` to `null` and actually reference the category.
Meanwhile, `func_201503_d` can be deprecated and replaced with an additional function, patching the call when needed. 

Regarding `ClientRecipeBook`, I'd do the same: deprecate the current methods, try to patch vanilla to use the newly added ones, and have the old method redirect to the new method if possible, so that patch size can be lowered if needed (E.g. `func_216767_a` could be redirected to a `getRecipeByCategory` method with the same body as much as possible).

Regarding the `enum`, I'd hardcode the types directly within the constructor: this way there is no need to modify the constructor. This shouldn't matter much, but it may lead to lower patch sizes. Or probably just hardcoding the return type in the method, by making `IForgeRecipeBookCategory` provide default implementations for most of the methods you added (`isSearch`, `getPredicate`, `getRecipeType` and friends).

Nevertheless, these suggestions may need some further discussion around feasibility, since if they end up complicating the code too much, maybe it's not that worth it.

ichttt:
Just as a note: https://github.com/MinecraftForge/MinecraftForge/pull/6669 exists, but it seems abandond.
Also, new features should first be PRd to 1.15 instead of 1.14

yanny7:
I port this PR to 1.15 after will be approved for 1.14

diesieben07:
> I port this PR to 1.15 after will be approved for 1.14

That is the wrong direction. New features go in 1.15 first. 1.14 is LTS and receives backports after the fact.

yanny7:
Its enough for me that there will be no problem with code for 1.14. I dont want to waste my time by porting it right now into 1.15 when code is not polished. I have no problem that after approving that code is OK for merging this PR will be postproned until 1.15 port will be available

diesieben07:
A feature PR for 1.14 won't get approved if it's not a backport of a 1.15 feature.

yanny7:
Feature was ported to 1.15

yanny7:
Still WIP

TheCurle:
This 1.14 PR has a 1.15 version and is unnecessary. 
As a side note, the same applies as before. I don't see a 1.16 version, so you should target that before the 1.15 PR is considered.
Apologies for the hassle, but manpower is limited here...

CREEATION:
@yanny7 any news on this? 🐢

PR #6786, opened 2020-06-19
Use Fluid's fillSound rather than emptySound when filling buckets

shartte:
Probably just a simple mixup here, the getEmptySound() is used where BucketItem actually is picking up a fluid (thus, the fillSound should be played).

The two obfuscated names on the next line for the vanilla sounds map to ITEM_BUCKET_FILL_LAVA & ITEM_BUCKET_FILL.


marchermans (approved):

marchermans (approved):
Good catch.
Might be a bit small of a change for a single PR but this is also used by a different PR, so my question to @shartte Was there a reason why you did not make this and the fluid sounds PR #6787 into one?

ichttt:
Yeah IMO this and #6787 should be in the same PR, as both are related to fluid shounds and very small

shartte:
Well, they _are_ different bugs and while they both relate to Fluids, one might want to merge one, but not the other.

PR #6787, opened 2020-06-19
Provide empty/fill sounds for vanilla fluids

shartte:
This PR aims to make playing fluid empty/fill sounds easier or mods by providing the Vanilla sounds in the Vanilla fluid attributes. This removes the need to "special-case" this in the mod.

NOTE1: This changes behavior in that previously the "minecraft:lava" fluid tag would toggle the lava empty/fill sound for buckets (if the FluidAttributes did not specify a sound).
Removing the lava fluid from that tag would now no longer change the sound to the water empty/fill sounds.
NOTE2: Without #6786, this PR would incorrectly use the liquid attribute's empty sound when filling a bucket.


TheCurle (approved):
Quick and easy. Looks good.

PR #6788, opened 2020-06-19
[1.15] Redirect to new Forge model methods in vanilla multipart model

KnightMiner:
This pull request makes the vanilla multipart model redirect the Forge `BlockState` variants of `isAmbientOcclusion` and `getParticleTexture` to the first model in the multipart model, for consistency with the stateless variants.

In addition, `getQuads` now passes along the `IModelData` instance to each submodel. I currently have that logic pulled into `ForgeHooksClient` as it was quite large, let me know if there is a preferred place to put that code.

I simply redirected `getModelData` to the first model for consistency with the other methods, as that is sufficient for most needs, including my own. We could potentially call the method for each submodel, passing the previous as `tileData`, but I think its very unlikely to have two different model loaders/sets of model data in a single multipart model, and in many cases that will be inefficient as you run the same logic multiple times (as each child likely has the same loader).

----

Included in this pull request is an update to the model loader test mod, adding in `multipart_block`. This block uses the fence model, but the center post is replaced with a "dynamic" model that turns invisible when placed above white wool. Note it does not have the tag for the block to connect to itself, so place it near solid blocks to test (was not sure it was worth adding the tag since that is beyond the point of the test mod)

### DIFF
--- a/patches/minecraft/net/minecraft/client/renderer/model/MultipartBakedModel.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/model/MultipartBakedModel.java.patch
@@ -1,0 +1,38 @@
+--- a/net/minecraft/client/renderer/model/MultipartBakedModel.java
++++ b/net/minecraft/client/renderer/model/MultipartBakedModel.java
+@@ -27,10 +27,12 @@
+    protected final ItemCameraTransforms field_188624_d;
+    protected final ItemOverrideList field_188625_e;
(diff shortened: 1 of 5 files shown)


marchermans (approved):
This looks good.
This also helps me along thanks.

KnightMiner:
8f9e52c added this to the 1.16 branch in a slightly different form, and I have moved most of my development to 1.16 so I do not plan to backport tterrag's changes, so closing this.

PR #6790, opened 2020-06-20
[1.15.x] Extendible categories in recipe book

yanny7:
Basicaly allows modders to add recipe book into custom machines and add tabs into vanilla recipe book

- custom recipe book category can be registered via forge registries.
- constructor have `IRecipeType<?>` to set where recipe book category belongs
- modder can set if category is `search` - no filtering is applied for displayed recipes
- `setPredicate(IRecipe<?>)` non-search category with filtering
- `setIcon(ItemStack...)` recipe book category icon
```
    @SubscribeEvent
    public static void registerRecipeBookCategories(RegistryEvent.Register<RecipeBookCategories> event) {
        event.getRegistry().register(new RecipeBookCategories(TestRecipe.millstone).setSearch().setIcon(new ItemStack(Items.BELL)).setRegistryName("base_crafting_book_test", "test"));
    }
```
- Custom machine container must extends `RecipeBookContainer<C extends IInventory>` 
- Custom machine gui just render and handle recipe book as in e.g. AbstractFurnaceScreen

### DIFF
--- a/patches/minecraft/net/minecraft/client/world/ClientWorld.java.patch
+++ b/patches/minecraft/net/minecraft/client/world/ClientWorld.java.patch
@@ -69,3 +69,12 @@
        if (p_217384_1_ == this.field_73037_M.field_71439_g) {
           this.field_73037_M.func_147118_V().func_147682_a(new EntityTickableSound(p_217384_3_, p_217384_4_, p_217384_2_));
        }
+@@ -689,7 +706,7 @@
+    }
+ 
+    public double func_228331_m_() {
+-      return this.field_72986_A.func_76067_t() == WorldType.field_77138_c ? 0.0D : 63.0D;
++      return this.field_73011_w.getHorizonHeight();
+    }
+ 
+    public int func_228332_n_() {
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeDimension.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeDimension.java
@@ -26,9 +26,8 @@ import net.minecraft.client.renderer.Vector3f;
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.entity.player.ServerPlayerEntity;
 import net.minecraft.util.math.BlockPos;
-import net.minecraft.util.math.Vec3d;
 import net.minecraft.world.World;
-import net.minecraft.world.biome.Biome;
+import net.minecraft.world.WorldType;
 import net.minecraft.world.biome.Biomes;
 import net.minecraft.world.chunk.Chunk;
 import net.minecraft.world.dimension.Dimension;
@@ -237,6 +236,11 @@ public interface IForgeDimension
     default int getSeaLevel()
     {
         return 63;
+    }
+
+    default double getHorizonHeight()
+    {
+        return getWorld().getWorldInfo().getGenerator() == WorldType.FLAT ? 0.0D : 63.0D;
     }
 
     /**


patches/minecraft/net/minecraft/client/gui/recipebook/RecipeBookGui.java.patch
```diff
@@ -1,15 +1,118 @@
...
+-            return p_209505_1_.func_201503_d().equals(this.field_191913_x.func_201503_d());
++            return p_209505_1_.getRecipeBookCategory().equals(this.field_191913_x.getRecipeBookCategory());
+          }).findFirst().orElse((RecipeTabToggleWidget)null);
+       }
+ 
+@@ -104,7 +107,18 @@
+ 
+       this.field_191913_x.func_191753_b(true);
+       this.func_193003_g(false);
++      // remove tabs that does not contains recipes
```
marchermans: Patches should be without comments.
In my eyes the patch is clear in its intention, when you read the change.
As such it is a higher priority to keep the patches as small as possible and then comments are not really needed and just bloat the patch.

Could you remove it?

patches/minecraft/net/minecraft/client/gui/recipebook/RecipeTabToggleWidget.java.patch
```diff
@@ -0,0 +1,66 @@
...
+-      List<ItemStack> list = this.field_193921_u.func_202903_a();
++      List<ItemStack> list = this.recipeBookCategory.getIcon();
+       int i = this.field_191755_p ? -2 : 0;
+       if (list.size() == 1) {
+          p_193920_1_.func_180450_b(list.get(0), this.x + 9 + i, this.y + 5);
+@@ -101,12 +102,13 @@
+ 
+    }
+ 
++   @Deprecated
```
marchermans: Here however a comment would be nice.
You mark this method as Deprecated, however you do not explain anywhere what to do if you want to use the method, aka what the alternatives are, or where the new registry is located.

When you deprecated a field, method or class you should insert a comment what to do in case a modder is using it. This is a higher priority (at least in my eyes) then keeping the patches small.

patches/minecraft/net/minecraft/client/util/RecipeBookCategories.java.patch
```diff
@@ -0,0 +1,188 @@
...
++
++   @javax.annotation.Nullable
++   @Override
++   public net.minecraft.util.ResourceLocation getRegistryName() {
++      return registryName;
++   }
++
++   @Override
++   public Class<net.minecraftforge.common.extensions.IForgeRecipeBookCategory<? extends net.minecraft.item.crafting.IRecipeType<?>>> getRegistryType() {
++      //return RecipeBookCategories.class;
```
marchermans: I am not sure returning null here is a good idea, especially since this goes into an `IForgeRegistry`but i could be wrong here.
Anybody with more information on that system should be able to tell you if this is correct or not.

patches/minecraft/net/minecraft/inventory/container/RecipeBookContainer.java.patch
```diff
@@ -1,11 +1,72 @@
...
 +++ b/net/minecraft/inventory/container/RecipeBookContainer.java
-@@ -31,4 +31,8 @@
+@@ -31,4 +31,69 @@
  
     @OnlyIn(Dist.CLIENT)
     public abstract int func_203721_h();
 +
-+   public java.util.List<net.minecraft.client.util.RecipeBookCategories> getRecipeBookCategories() {
-+      return net.minecraft.client.util.ClientRecipeBook.func_216769_b(this);
++   public net.minecraft.item.crafting.IRecipeType<?> getRecipeType() {
```
marchermans: Just a question:
Why is this done using a bunch of if-switches here?

Somebody from the @Triage-Team: Is it better to make this like he implemented it here. Or should he override the method in the individual containers (like he says modders should do)? Patches would be bigger and have a higher count. However it would be easier to maintain and sets a better example for others.

src/main/java/net/minecraftforge/common/ForgeRecipeBookCategories.java
```diff
@@ -0,0 +1,54 @@
...
+import com.google.common.collect.ImmutableList;
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.crafting.IRecipe;
+import net.minecraft.item.crafting.IRecipeType;
+import net.minecraftforge.common.extensions.IForgeRecipeBookCategory;
+
+import java.util.List;
+import java.util.function.Predicate;
+
+public class ForgeRecipeBookCategories<T extends IRecipeType<?>> extends net.minecraftforge.registries.ForgeRegistryEntry<IForgeRecipeBookCategory<?>> implements IForgeRecipeBookCategory<T> {
```
marchermans: Should this not be named: `ForgeRecipeBookCategory` ? Since this is one category and not multiple ones?

src/main/java/net/minecraftforge/common/extensions/IForgeRecipeBookCategory.java
```diff
@@ -0,0 +1,16 @@
...
+
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.crafting.IRecipe;
+import net.minecraft.item.crafting.IRecipeType;
+import net.minecraftforge.registries.IForgeRegistryEntry;
+
+import java.util.List;
+import java.util.function.Predicate;
+
+public interface IForgeRecipeBookCategory<T extends IRecipeType<?>> extends IForgeRegistryEntry<IForgeRecipeBookCategory<?>> {
```
marchermans: Documentation is needed here.

JavaDoc or otherwise.

What does every method do.
You are working on an API that others use and need to maintain. You are not in a Patch, so be as detailed as you can be with this.

marchermans (changes requested):
Generally documentation issues which still need to be resolved.
As well as some issues with comments in patches.

Once the @Triage-Team gets to it: I left a question regarding the way he implemented the RecipeBookContainer patch for the different containers. AKA  A long list of if-switches in one patch of the base class. Or several smaller patches where he overrides the method from the base class. What is the goto way here?

patches/minecraft/net/minecraft/client/util/RecipeBookCategories.java.patch
```diff
@@ -0,0 +1,188 @@
...
++public enum RecipeBookCategories implements net.minecraftforge.common.extensions.IForgeRecipeBookCategory<net.minecraft.item.crafting.IRecipeType<?>> {
+    SEARCH(new ItemStack(Items.field_151111_aL)),
+    BUILDING_BLOCKS(new ItemStack(Blocks.field_196584_bK)),
+    REDSTONE(new ItemStack(Items.field_151137_ax)),
+@@ -36,4 +36,176 @@
+    public List<ItemStack> func_202903_a() {
+       return this.field_202904_j;
+    }
++
++   //--FORGE
```
marchermans: This is not correct, we use:
`/*================================ FORGE START ================================================*/`
as a separator,

You should probably add this on all files where you add a footer section with just forge changes.
Additionally it also needs an end marker:
`/*================================ FORGE END ================================================*/`

marchermans (changes requested):
Another small adaptation of something i overlooked in this PR, with regards to formatting.
The forge section markers are missing or wrong.

ExpensiveKoala:
Is there a reason why RecipeBookContainer isn't an interface? What if my machines container needs to extend a different container?

patches/minecraft/net/minecraft/inventory/container/RecipeBookContainer.java.patch
yanny7: As pointed in this comment in 1.14 implementation https://github.com/MinecraftForge/MinecraftForge/pull/6785#pullrequestreview-434422979 this will allow not to patch enum fields. I will add default statement for throwing exception for notifying when new field will be added

src/main/java/net/minecraftforge/common/ForgeRecipeBookCategories.java
yanny7: Right, I will change it

patches/minecraft/net/minecraft/client/gui/recipebook/RecipeBookGui.java.patch
yanny7: Removed comments

patches/minecraft/net/minecraft/client/gui/recipebook/RecipeTabToggleWidget.java.patch
yanny7: Added comments for deprecated methods

patches/minecraft/net/minecraft/inventory/container/RecipeBookContainer.java.patch
yanny7: It was request from @TheSilkMiner to do it on one place for vanilla containers

src/main/java/net/minecraftforge/common/extensions/IForgeRecipeBookCategory.java
yanny7: Added  JavaDoc

patches/minecraft/net/minecraft/client/util/RecipeBookCategories.java.patch
yanny7: Replaced with valid forge separator

patches/minecraft/net/minecraft/client/util/RecipeBookCategories.java.patch
yanny7: Added missing return value

patches/minecraft/net/minecraft/inventory/container/RecipeBookContainer.java.patch
TheSilkMiner: Since I was mentioned, my line of reasoning was less patches is better. Moreover, it seems Forge is taking the hardcoded approach in other patches and extension interfaces (refer to `IForgeBlock` and the patch for `Block`).

marchermans: I know, but I have seen it in other places the other way around, and since I am starting the reviewing of PRs, i was wondering. I mean we should technically set an example as well for other modders: How todo it right. And this is might or might not be the way we project this kind of solution as right. Hence the request to the Triage team to clarify how this is supposed to be handled.

This was not directed against @yanny7 but more as a question to the Triage team what the going rate here is? Is 4 if-switches okey, 5? 6? or only 3?.

src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -197,6 +198,7 @@ public static void init()
         makeRegistry(PAINTING_TYPES, PaintingType.class, new ResourceLocation("kebab")).create();
         makeRegistry(RECIPE_SERIALIZERS, IRecipeSerializer.class).disableSaving().create();
         makeRegistry(STAT_TYPES, StatType.class).create();
+        makeRegistry(RECIPE_BOOK_CATEGORIES, net.minecraftforge.common.extensions.IForgeRecipeBookCategory.class).disableSaving().create();
```
ichttt: Also wrong spot, put it below the `Custom forge registries` tab. You can also use imports here, this is not a patch

src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -126,6 +126,7 @@
     public static final ResourceLocation PAINTING_TYPES = new ResourceLocation("motive"); // sic
     public static final ResourceLocation RECIPE_SERIALIZERS = new ResourceLocation("recipe_serializer");
     public static final ResourceLocation STAT_TYPES = new ResourceLocation("stat_type");
+    public static final ResourceLocation RECIPE_BOOK_CATEGORIES = new ResourceLocation("recipe_book_categories");
```
ichttt: Also wrong spot, put it below the `Custom forge registries` comment. Additionally, is there any reason this lives in the minecraft namespace and not the forge namespace?

src/main/java/net/minecraftforge/registries/ForgeRegistries.java
```diff
@@ -81,6 +82,7 @@
     public static final IForgeRegistry<PaintingType> PAINTING_TYPES = RegistryManager.ACTIVE.getRegistry(PaintingType.class);
     public static final IForgeRegistry<IRecipeSerializer<?>> RECIPE_SERIALIZERS = RegistryManager.ACTIVE.getRegistry(IRecipeSerializer.class);
     public static final IForgeRegistry<StatType<?>> STAT_TYPES = RegistryManager.ACTIVE.getRegistry(StatType.class);
+    public static final IForgeRegistry<IForgeRecipeBookCategory<?>> RECIPE_BOOK_CATEGORIES = RegistryManager.ACTIVE.getRegistry(IForgeRecipeBookCategory.class);
```
ichttt: This is the wrong spot, put it below the `Custom forge registries` comment.

src/main/java/net/minecraftforge/common/ForgeRecipeBookCategory.java
```diff
@@ -0,0 +1,65 @@
...
+import com.google.common.collect.ImmutableList;
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.crafting.IRecipe;
+import net.minecraft.item.crafting.IRecipeType;
+import net.minecraftforge.common.extensions.IForgeRecipeBookCategory;
+
+import java.util.List;
+import java.util.function.Predicate;
+
+public class ForgeRecipeBookCategory<T extends IRecipeType<?>> extends net.minecraftforge.registries.ForgeRegistryEntry<IForgeRecipeBookCategory<?>> implements IForgeRecipeBookCategory<T> {
```
ichttt: No need to use fully qualified names, this is not a patch. Use imports instead

src/main/java/net/minecraftforge/common/ForgeRecipeBookCategory.java
```diff
@@ -0,0 +1,65 @@
...
+        return this.recipeType;
+    }
+
+    @Override
+    public List<ItemStack> getIcon() {
+        return this.icons;
+    }
+
+    public static List<IForgeRecipeBookCategory<?>> getValidCategories(IRecipe<?> recipe) {
+        return net.minecraftforge.registries.ForgeRegistries.RECIPE_BOOK_CATEGORIES.getValues().stream().filter(category -> {
```
ichttt: Again, use imports

src/main/java/net/minecraftforge/client/UpDownButton.java
```diff
@@ -0,0 +1,57 @@
...
+        if (isDown) {
+            blit(this.x, this.y, 1, 208 + i * 18, 11, 17);
+        } else {
+            blit(this.x, this.y, 15, 208 + i * 18, 11, 17);
+        }
+        this.renderBg(minecraft, mouseX, mouseY);
+    }
+
+    // re-implemented blit function for rotating and mirroring texture
+    public void blit(int x, int y, int u, int v, int w, int h) {
```
ichttt: This should not be needed, you can just rotate using `RenderSystem` and use the vanilla functions instead of copying and modifing the functions.

patches/minecraft/net/minecraft/util/registry/Registry.java.patch
```diff
@@ -164,8 +164,17 @@
 +   @Deprecated public static final Registry<Activity> field_218375_Q = forge("activity", Activity.class, () -> {
        return Activity.field_221366_b;
     });
++   @Deprecated public static final Registry<net.minecraftforge.common.extensions.IForgeRecipeBookCategory<?>> RECIPE_BOOK_CATEGORIES = forge("recipe_book_categories",
++           net.minecraftforge.common.extensions.IForgeRecipeBookCategory.class, Registry::getDefaultRecipeBook);
  
-@@ -246,6 +248,14 @@
++   private static net.minecraftforge.common.extensions.IForgeRecipeBookCategory<?> getDefaultRecipeBook() {
```
ichttt: `getDefaultRecipeBookCategory` would make more sense, although it's a very long name

patches/minecraft/net/minecraft/inventory/container/RecipeBookContainer.java.patch
```diff
@@ -1,11 +1,73 @@
...
++
++   public java.util.List<net.minecraftforge.common.extensions.IForgeRecipeBookCategory<?>> getRecipeBookCategories() {
++      return net.minecraftforge.registries.ForgeRegistries.RECIPE_BOOK_CATEGORIES.getValues().stream().filter(category -> {
++         return category.getRecipeType().equals(this.getRecipeType());
++      }).sorted(RecipeBookContainer::compareTo).collect(java.util.stream.Collectors.toList());
++   }
++
++   private static final java.util.Map<net.minecraftforge.common.extensions.IForgeRecipeBookCategory<?>, Integer> VANILLA_LIST  = new java.util.HashMap<>();
++
++   private static int compareTo(net.minecraftforge.common.extensions.IForgeRecipeBookCategory<?> from, net.minecraftforge.common.extensions.IForgeRecipeBookCategory<?> to) {
```
ichttt: I'm not too happy with the whole sorting thing. IMO, it should just use vanilla first, and then registration order. ItemGroups function similar to this

patches/minecraft/net/minecraft/inventory/container/RecipeBookContainer.java.patch
TheCurle: Having the logic in one centralised place has its merits - as does having the change spread out over smaller patches to where the change actually makes an effect.
It depends on the code style of the person writing it, and in this case I believe this fits the Forge style the best. If the list were any bigger i'd suggest splitting it out, and that might be worth a //TODO for when (and if) Mojang add any more blocks like these.

marchermans: Perfect, thanks for the clear up.

marchermans (approved):
This looks good now.
Patches are as far as I know of a good quality, and functionally they seem to do the job.

Nice work.

ichttt (commented):
Just some of my thoughts

patches/minecraft/net/minecraft/inventory/container/RecipeBookContainer.java.patch
yanny7: Ordering is definitely needed there, first I must provide correctly ordered vanilla categories and secondly to not break vanilla ordering, I have to provide as first category "search".

Other categories are then in registration order

LordMonoxide: Since setup events are fired concurrently, wouldn't that make the category order for mod categories indeterministic? It might be a better user experience to sort mod categories by name

src/main/java/net/minecraftforge/registries/GameData.java
yanny7: There is no reason, I just put it on wrong place

patches/minecraft/net/minecraft/inventory/container/RecipeBookContainer.java.patch
TheSilkMiner: ~~Setup~~ Registry events are not parallel, meaning this somewhat guarantees ordering. They may become parallel in the future, though.
But at the same time I don't see how this is an issue, considering the same happens for item groups, blocks, items themselves... There is a lot of reliance on registration order so I believe this is fine.

src/main/java/net/minecraftforge/registries/GameData.java
yanny7: Fixed

src/main/java/net/minecraftforge/registries/GameData.java
yanny7: Fixed

src/main/java/net/minecraftforge/registries/ForgeRegistries.java
yanny7: Fixed

src/main/java/net/minecraftforge/common/ForgeRecipeBookCategory.java
yanny7: Fixed

src/main/java/net/minecraftforge/common/ForgeRecipeBookCategory.java
yanny7: Fixed

src/main/java/net/minecraftforge/client/UpDownButton.java
yanny7: used RenderSystem

patches/minecraft/net/minecraft/util/registry/Registry.java.patch
yanny7: renamed

patches/minecraft/net/minecraft/inventory/container/RecipeBookContainer.java.patch
marchermans: @yanny7 What is the exact reasoning for the ordering? Does vanilla do an index based approach here and just assumes that "Search" is always index 0?

Or why does Vanilla want these pages to be first?

yanny7: In vanilla, order is hard-coded, but always in first tab are recipes without filtering, what makes sense. I just wanna make it consistent.

I named it wrong "search", because you can search in every tab. But as I mentioned, first tab is unfiltered list of recipes.

I rename it to match reality

marchermans: But where is it hard-coded? Is it just the way they add the tabs to the list? Are they accessing that list then again to get a tab back an rely on the unfiltered list being on 0?

Cause at the moment I do not see a reason for the ordering except for: Vanilla adds it this way, so that is the reason.

Do not get me wrong. Having ordering on something like this would be nice since it produces some consistency in a much used module.

yanny7: Please check old implementation:
```
public static List<RecipeBookCategories> func_216769_b(RecipeBookContainer<?> p_216769_0_) {
      if (!(p_216769_0_ instanceof WorkbenchContainer) && !(p_216769_0_ instanceof PlayerContainer)) {
         if (p_216769_0_ instanceof FurnaceContainer) {
            return Lists.newArrayList(RecipeBookCategories.FURNACE_SEARCH, RecipeBookCategories.FURNACE_FOOD, RecipeBookCategories.FURNACE_BLOCKS, RecipeBookCategories.FURNACE_MISC);
         } else if (p_216769_0_ instanceof BlastFurnaceContainer) {
            return Lists.newArrayList(RecipeBookCategories.BLAST_FURNACE_SEARCH, RecipeBookCategories.BLAST_FURNACE_BLOCKS, RecipeBookCategories.BLAST_FURNACE_MISC);
         } else {
            return p_216769_0_ instanceof SmokerContainer ? Lists.newArrayList(RecipeBookCategories.SMOKER_SEARCH, RecipeBookCategories.SMOKER_FOOD) : Lists.newArrayList();
         }
      } else {
         return Lists.newArrayList(RecipeBookCategories.SEARCH, RecipeBookCategories.EQUIPMENT, RecipeBookCategories.BUILDING_BLOCKS, RecipeBookCategories.MISC, RecipeBookCategories.REDSTONE);
      }
   }
```
You see that first tab is always SEARCH
I dont see problem with forcing this way, if there is problem with it I remove it

marchermans: @yanny7 I understand that this is the issue here, but my question to clarify:
Does it break if the order is changed: AKA Can I put EQUIPMENT first and SEARCH second?
Obviously this would switch around the tabs, but does it break something else?

yanny7: @OrionDevelopment it breaks anything, but we should stick with some conventions. People would expect that first tab is without filtering, cause all vanilla recipe books works same

marchermans: So pure from a technical standpoint the sorting is not required. It is just a "convention" okey, thanks for clearing that up.

yanny7:
forgot rebase

PR #4958, opened 2018-05-29
Add elytra flight attribute

TheIllusiveC4:
Remake of #4476 that was recently closed due to inactivity.

This PR allows for custom elytra flight through the use of entity attributes so that mods can add their own elytra items without compatibility conflicts. The included test mod is the same as the previous PR with Elytra Boots that grant elytra flight and an Anti-Elytra Potion to remove elytra flight ability.

I replaced the 3-case implementation from the previous PR with a 2-case implementation, opting to patch the attribute into the vanilla elytra item. Values >= 1 will grant elytra flight, otherwise it will not. I believe this to be cleaner overall, but if the previous 3-case is preferred then I can change this.

PR #4848, opened 2018-04-03
Add custom Elytra support

l-Luna:
This adds support for items to act as custom Elytras via implementing IElytra on the item. 

I know that https://github.com/MinecraftForge/MinecraftForge/pull/4476 exists, but this makes creating Elytras with custom properties significantly easier.

One use-case is a RF-powered Elytra with can speed up at the cost of more energy.

PR #4759, opened 2018-02-19
Added IElytra and a check for it in EntityLivingBase to be able to make your own Elytra

Minecraftschurli:
changed Elytra check from == to instanceof

PR #4476, opened 2017-10-21
1.12.x custom elytra

pWn3d1337:
This PR tries to allow for custom elytra flight through Entity Attributes. The idea is based on https://github.com/MinecraftForge/MinecraftForge/pull/3931 but for elytra flight. The associated vanilla bug with entity attributes has since been fixed in https://github.com/MinecraftForge/MinecraftForge/pull/4265.

The testmod adds ElytraBoots which are boots that grant elytra flight trough event handlers. They also provide superjump and no falldamage for easier fly-testing. 
The other added item is an anti-elytra potion that stops elytra flying (including vanilla).

The Entity Attribute has a base value of 1 and a range of 0 to 2. The logic behind this is: value of 1(default) allows elytra flight with vanilla elytra. >1 means elytra flight no matter if vanilla elytra is worn. <1 means no elytra flight at all, even with vanilla elytra.

PR #6791, opened 2020-06-21
[1.15.x] Add support for custom elytra

brandon3055:
There is currently no good way to add custom elytra or elytra type flight. The primary issue is that all of the elytra code is hard coded to work with `item == Items.ELYTRA`.
This PR aims to resolve this issue by adding a simple interface that can be added to any chest plate item. And because forge allows you specify an `EquipmentSlotType` for any item this does not have to be an armor item.

There have been a few similar PR's in the past. #4958 #4848 #4759 #4476 but none of these got merged so i figured i would take a crack at it. 
My goal was to make this PR as simple and easy to maintain as possible. 
Initially this was just one interface with two methods and 3 minor changes in vanilla code. Then i realized with just one more minor change i could add the ability to easily use the vanilla elytra model with custom texture support so i added that as well.  

There are a lot of other features that were covered by the previous pr's such as custom damage handling, velocity handling for things like a "boost" function as well as a few other things but all of that can already be done via other means. So instead i chose to just focus on the base issue which is that its not currently possible to add custom elytra type flight.  

The only real down side to this pr vs something like #4958 is that the interface added by this pr only works when used on an item that goes in the chest plate slot.
But i don't think that's to unreasonable.


50ap5ud5 (changes requested):
Nice work. To make your pull request more easily merged, I would suggest the following changes:

**PR Description:**

1. Add a [1.15.2] tag to the title of your Pull Request. This can save time for the Forge team when reviewing your PR
2. Elaborate on what issue your PR solves and what your solution provides. While your justification, `“There have been a few failed attempts to add this in the past so i decided to take a simpler approach.”` Is a valid statement, it can be quite vague. Could you provide a link to the previous failed attempts? Furthermore, please explain why is your version going to be more successful.
3. `“This is the bare minimum required to add custom elytra.”` - What features did you add in your patches? Perhaps provide a summary of your changes to allow for easy reviewing.

To add these in, I suggest taking a look at other pull requests. Gigaherz's multi-layer item rendering pull request does a good job of this by using headers and dot points.

**Patch files:**
1.	In line 42-45 of ClientPlayerEntity.java.patch, you have added an extra space which has caused the git diff to show that you’ve changed the Minecraft code. Please try and minimize the changes you make to the original code as to keep the patch sizes small, as requested by the Contributor’s Guidelines in the MinecraftForge Github Wiki.
2.	Documentation for new functions. Please provide a Javadoc for the method `public void updateSyncFields(ClientPlayerEntity old)` in ClientPlayerEntity patch. This can help modders and other reviewers better understand the purpose of this function.
3.	Other miscellaneous formatting enhancements:
 In LivingEntity patch, line 539, add a comment to indicate the end of your new additions. You have added “Forge Start” on line 487, but there is no “Forge End” comment. This is just a minor formatting nit-pick to ensure that the people viewing the class will know where your changes end.

I’m not too knowledgeable about the validity of other aspects of your patch, such as replacing vanilla behaviour with Forge hooks. I’m going to leave that to more experienced reviewers to comment upon.

Feel free to ask for advice from other modders in the Forge discord. These are just some common expectations that I've seen requested on most pull requests.

covers1624:
@50ap5ud5 Please leave your review comments as an actual review, everything you say about the patches doesnt line up with the changes in the pr.

marchermans (approved):
The code looks sound and the patches are good.
Please add some details however to your PR itself since the PR is not well justified at the moment.

brandon3055:
@OrionDevelopment Yea i got a little ahead of myself when i made the PR and completely forgot to add a description so that description was a little rushed. 

@50ap5ud5 I'm with covers on this one. You were right about my description needing work but most of your feedback regarding my changes has nothing to do with my changes.

covers1624 (approved):

50ap5ud5:
Hi brandon, to clarify my comments on your patch, in the ClientEntityPlayer patch I was suggesting that some documentation be provided for this method: https://github.com/MinecraftForge/MinecraftForge/blob/2a8287ec664fdb6d4926f93ad39a5dcf54a84665/patches/minecraft/net/minecraft/client/entity/player/ClientPlayerEntity.java.patch#L63

About your LivingEntity patch, the comment about adding "Forge End" was referring to this line: 
https://github.com/MinecraftForge/MinecraftForge/blob/2a8287ec664fdb6d4926f93ad39a5dcf54a84665/patches/minecraft/net/minecraft/entity/LivingEntity.java.patch#L487
There is a "Forge Start" but not "Forge End" line.

You didn't do anything wrong and I wasn't trying to attack you. I just noticed there were some minor aspects that could be looked into.

Hope that clears some things up. Sorry for the confusion about the review on your patch, I'll be sure to leave the review comments as an actual review in the future.

As a note, the links I provided can be accessed by clicking "Files Changed" tab, and then clicking "View file" button as shown in the screenshot below. It shows the whole file rather than a snippet.
image

brandon3055:
smh... Your looking at the raw patch file. Those changes have nothing to do with me or this PR. when you "view file" what your looking at is the full patch file which includes a bunch of changes made by forge or other contributors which includes the changes made by me. @covers1624 help me out here. I'm not sure how best to explain this.

covers1624:
@50ap5ud5 Your looking at a diff of a patch file. Not the entire Patch file. We only care about the changes here, anything else was added in other PR's or by the Forge team. If you are unsure how the Forge patch system works, please go and look at reviews on other PR's before commenting here.

Minecraftschurli (approved):
Looks good and I would appreciate it if this would be finally added as I did one of the previous attempts at these changes

src/main/java/net/minecraftforge/items/IElytra.java
```diff
@@ -0,0 +1,61 @@
...
+
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.ResourceLocation;
+
+public interface IElytra
+{
+
+    /**
+     * Called when the user attempts to start elytra flying and every tick while flying.
```
diesieben07: This is bad documentation, imho. You shouldn't specify when this is called, you should specify what it should do. In this case: return true to allow flying.

src/main/java/net/minecraftforge/items/IElytra.java
```diff
@@ -0,0 +1,61 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.items;
+
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.ResourceLocation;
+
+public interface IElytra
```
diesieben07: Maybe some Javadocs on the class explaining how to use this interface ("implement this on your item", etc.)

diesieben07 (changes requested):
Another note, do we really want to include textures for a test mod?

brandon3055:
I figured it would be fine to add textures as there are several other test mods that do that. But i can just use the vanilla textures if that is preferred. The intention was to make it clear that you can use a custom texture for the model but its not really that important.

diesieben07:
I am not sure about the texture, it was more of a question.

src/main/java/net/minecraftforge/items/IElytra.java
```diff
@@ -23,11 +23,19 @@
 import net.minecraft.item.ItemStack;
 import net.minecraft.util.ResourceLocation;
 
+/**
+ * Implementing this on an item will allow that item to act as a set of elytra.
+ * Any item implementing this must be able to be placed in the chest plate slot.
```
Minecraftschurli: Why does it need to be a chestplate that seems a bit restrictive could you change it so it works in any equipmentslot (for example elytra boots)

marchermans: As far as I know from looking at his patches, this PR is minimal and just replaces the vanilla itemstack and item check for something with extensible logic, instead of replacing the entire detection logic.

In my eyes this PR does exactly what it says on the tin.
And that is fine.
Another PR in the future can add the additional check that makes it possible from anywhere, let's start somewhere small and see how the community reacts.

brandon3055: Because this essentially just extends the vanilla `item == Items.ELYTRA` check which only checks the chest plate slot. I could extend it to the other equipment slots but i was trying to minimize changes to vanilla code. And if i am extending it to the other equipment slots what about an item that lets you fly when you are holding it? Or some sort of charm that works as long as its in your inventory like an angel ring? What about things like baubles? Where do you draw the line? 

This is why i decided to keep it as simple as possible. This is where something like the elytra entity attribute that was attempted in #4958 would be useful as its is much more flexible. But its also a lot more complicated requiring much more invasive changes to vanilla code which is probably why it never got merged. 

People have been trying to get elytra support merged since 2018. I just want to create a simple pr that might actually get merged.  

That said i can see how it may be worth while expanding this to the other equipment slots. I will let lex or mezz decide if its worth the extra complexity before i make the change.

LexManos:
Overall, this is a simple attempt that doesn't address any of the major concerns of the old PRs.
Combined with the fact that Forge has moved away from specialcase interfaces for things that only function on one object.
This belongs in IForgeItem/IForgeItemStack not a custom interface.
The changes to vanilla arnt bad, they are kept small and self contained. 

The problem with elytras is that they are a ugly hack on the vanilla code base that is a PITA to try and extend.

brandon3055:
@LexManos So it sounds like i have two options. I can stick with this simple approach but move it to IForgeItem. Or i could try to do a "minor overhaul" and turn elytra into something a little more mod friendly. for which i would probably use an attribute similar to #4958 (unless someone has a better idea) 
What would you prefer? 
Personally i think its probably best to take the simple approach for now just to get something working. The functionality can always be expanded with a more in depth pr later.

marchermans:
@brandon3055 I think it would be preferred if you went the `IForgeItem` _and_ `IForgeItemStack` route. Add the current player as a parameter to the method on the ItemStack as well and pass both the ItemStack and Item to the player.

That said. I also like the attribute based approach of the referenced stale PR. However its implementation needed more work.

TLDR: If you want to keep the PR small, use the Item/ItemStack route. If you feel like expanding this PR, use the attribute approach, where you can use the stale PRs code as a springboard for your own implementation.

brandon3055:
Switched to using `IForgeItem` and `IForgeItemStack` with some help from @covers1624 (he basically re wrote the entire thing xD)

covers1624:
@brandon3055 Please fix your conflicts.

brandon3055:
Resolved Conflict

covers1624 (approved):

PR #6793, opened 2020-06-22
DirectoryCache: use custom comparator.

### LINKED ISSUE #6783: Potential datagen sorting issue.
**Steps to Reproduce:**
 0.  must be run on a linuxy system (my case was gentoo ~amd64)
 1. `git clone https://github.com/SlimeKnights/TinkersConstruct`
 2. `gradlew runData`
 3. `git diff src/generated/resources/.cache/cache`

**Description of issue:**
One or more of the developers of Tinkers Construct use windows, and I use linux.
When I do runData for new recipe/loot table/etc I always get the following change:
```diff
diff --git a/src/generated/resources/.cache/cache b/src/generated/resources/.cache/cache
index 0e711a6af0..8f4b661397 100644
--- a/src/generated/resources/.cache/cache
+++ b/src/generated/resources/.cache/cache
@@ -64,13 +64,13 @@ f1062a73b4417d8772ebb9aa9189aa4064f649e0 data/forge/tags/items/ores/ardite.json
 05669b0f8399c31b028661957b82d60cc4b10a9c data/forge/tags/items/ores/cobalt.json
 f6d948da59e46d26cfc34188b9b140dd37f437d1 data/forge/tags/items/rods.json
 cb4d6289ec4c32cf5b57f3fa21419376486a7cae data/forge/tags/items/rods/stone.json
-2513ecf2fc1caedde8be0b41be0e6bf4e149dab6 data/forge/tags/items/slimeballs.json
 9ecc2c411681c0fc1a42acc3cdf7510783ffaec7 data/forge/tags/items/slimeball/blood.json
 8a9941607c1100a061f3c47d203b834dc91dc243 data/forge/tags/items/slimeball/blue.json
 a7629398a73eef2b5267342d79f1cebcf2b1eec4 data/forge/tags/items/slimeball/green.json
 1cb76aaacdb5b68a408abf6fac03a3b8a3a5d7d4 data/forge/tags/items/slimeball/magma.json
 caa063b792e2560c06105fd235ef20ffb152058f data/forge/tags/items/slimeball/pink.json
 cd0c3987c9be5c4d2fe8e9352734c581e1d60f15 data/forge/tags/items/slimeball/purple.json
+2513ecf2fc1caedde8be0b41be0e6bf4e149dab6 data/forge/tags/items/slimeballs.json
 67c5756a5424da1261b7e83f7ce0ced0763f5196 data/forge/tags/items/stained_glass.json
 ea3d6efc7f691cff6c7ad0eb59cf13ad9597afe1 data/forge/tags/items/storage_blocks.json
 74fb0e0d20e0081c34e64f3506802d8f4321ee01 data/forge/tags/items/storage_blocks/alubrass.json
```

This change of sorting always occurs, even on a clean checkout of the code. I think
this may do with the locale used to sort the second column.

[Related TiC issue/pr](https://github.com/SlimeKnights/TinkersConstruct/pull/4132)

hanetzer:
I think this fixes #6783 , user pupnewfster on discord showed only the
line-endings changing when they ran it.

### DIFF
--- a/src/test/java/net/minecraftforge/debug/world/MarkDimensionForDeletionTest.java
+++ b/src/test/java/net/minecraftforge/debug/world/MarkDimensionForDeletionTest.java
@@ -61,7 +61,7 @@ public class MarkDimensionForDeletionTest
         {
             if (!worldIn.isRemote)
             {
-                if (playerIn.isShiftKeyDown())
+                if (playerIn.isSneaking())
                 {
                     playerIn.sendMessage(new StringTextComponent("You are in dimension " + worldIn.dimension.getType().getRegistryName()));
                 }
--- a/patches/minecraft/net/minecraft/data/DirectoryCache.java.patch
+++ b/patches/minecraft/net/minecraft/data/DirectoryCache.java.patch
@@ -1,20 +1,13 @@
 --- a/net/minecraft/data/DirectoryCache.java
 +++ b/net/minecraft/data/DirectoryCache.java
-@@ -23,7 +23,7 @@
-    private final Path field_208326_c;
-    private int field_208327_d;
-    private final Map<Path, String> field_208328_e = Maps.newHashMap();
--   private final Map<Path, String> field_208329_f = Maps.newHashMap();
-+   private final Map<Path, String> field_208329_f = Maps.newTreeMap(); //Forge: TreeMap, makes the file output predictable/sorted.
-    private final Set<Path> field_218457_g = Sets.newHashSet();
- 
-    public DirectoryCache(Path p_i49352_1_, String p_i49352_2_) throws IOException {
-@@ -55,7 +55,7 @@
+@@ -55,8 +55,8 @@
        }
  
        IOUtils.writeLines(this.field_208329_f.entrySet().stream().map((p_208319_1_) -> {
 -         return (String)p_208319_1_.getValue() + ' ' + this.field_208325_b.relativize(p_208319_1_.getKey());
+-      }).collect(Collectors.toList()), System.lineSeparator(), writer);
 +         return (String)p_208319_1_.getValue() + ' ' + this.field_208325_b.relativize(p_208319_1_.getKey()).toString().replace('\\', '/'); //Forge: Standardize file paths.
-       }).collect(Collectors.toList()), System.lineSeparator(), writer);
++      }).sorted(java.util.Comparator.comparing(a -> a.split(" ")[1])).collect(Collectors.toList()), System.lineSeparator(), writer);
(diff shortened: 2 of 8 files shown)


src/test/java/net/minecraftforge/debug/world/MarkDimensionForDeletionTest.java
```diff
@@ -61,7 +61,7 @@
         {
             if (!worldIn.isRemote)
             {
-                if (playerIn.isShiftKeyDown())
+                if (playerIn.isSneaking())
```
marchermans, line 64: Why was this additional change made in this PR?

LexManos:
Your test means nothing. It wouldnt ever expose anything while run on a the same os over and over.
As for your solution, as I said in the linked issue, sorting the stream while writing is the best way to keep it localilized patch wise. It'd even let us delete our TreeMap change.
It just needs `.sorted((a,b) -> a.split(" ")[1].compareTo(b.split(" ")[1])`

src/test/java/net/minecraftforge/debug/world/MarkDimensionForDeletionTest.java
hanetzer: because mapping changes made the test not run.

hanetzer:
> Your test means nothing. It wouldnt ever expose anything while run on a the same os over and over.

It seemed like this would be the same kind of sorting issue as mentioned in the original issue,
a top-level tag foo**s**.json and nested tags under foo/stuff.json, but I guess I was
mistaken. I did have pup try it on their windows machine. Could have someone test it without
the comparator constructor on windows as well.

> As for your solution, as I said in the linked issue, sorting the stream while writing is the best way to keep it localilized patch wise. It'd even let us delete our TreeMap change.
> It just needs `.sorted((a,b) -> a.split(" ")[1].compareTo(b.split(" ")[1])`

Assuming you want that `.sorted((a,b) -> a.split(" ")[1].compareTo(b.split(" ")[1])`
in the same place as you mentioned [here](https://github.com/MinecraftForge/MinecraftForge/issues/6783#issuecomment-647029915), it doesn't seem to like it.
patch

hanetzer:
pup helped me sort out the sorting you suggested and it works on both my (linux)
and their (windows) machines with the same output.

hanetzer:
> Your test means nothing. It wouldnt ever expose anything while run on a the same os over and over.

So are you wanting that removed, or what? The reason I put it in is it, without the sorted
snippet inserted, produces different output on windows and linux, so a change which
causes the output to not be normalized the same could be noticed in theory.

Issue #6796, opened 2020-06-22
FluidUtil#tryEmptyContainer() is ignoring the doDrain parameter 


desht:
**Minecraft Version:** 1.15.2

**Forge Version:** 31.2.20

**Logs:** n/a

**Steps to Reproduce:**
 1. See https://github.com/TeamPneumatic/pnc-repressurized/issues/537

**Description of issue:**
`FluidUtil#tryEmptyContainer()` is ignoring the `doDrain` parameter and always transferring fluid to the destination fluid handler, even for simulated transfers.  This result in fluid dupe bugs under some circumstances (see linked issue report above).

gigaherz:
FluidUtil needs a **major** redesign/refactor, changing the fluid handler API in the process if necessary. The current API makes certain things effectively impossible, which I didn't realize at the time because I'm not the right person to work on fluids (I have never used the fluid handler API in any of my mods). If anyone wants to give FluidUtil a complete makeover for 1.16, please do so!

KingLemming:
What certain things are impossible right now? I'll see if I can actually get some time for 1.16...

gigaherz:
I can't remember all the details anymore, but: there was one of the methods, I think the one mentioned in this issue, for which "simulate" needed to know the resulting stack after draining, but in order to know the resulting stack, the code had to duplicate the input. But removing the duplication caused the return value to be incorrect.
At the time, I tried to think of how to fix it, and I really wanted to scrape the entire FluidUtils, and start over from a list of requirements, seeing exactly what the system *needs*, and not try to patch up what it currently does. I really wish I could explain why I went "ugh, this entire thing is a mess", but I can't remember anymore.

LordMonoxide:
I've worked/struggled with the fluid stuff in 1.15 a fair bit. Here's a couple issues:

- Things that are bucket-interactable implement `IBucketPickupHandler`. You can take the fluid from it by calling `pickupFluid`, but there's no way to check what fluid is inside before you do that. You get a fluid that you might not want, with no way to know how to put it back. For fluid source blocks, you can call `World.getFluidState`, but source blocks also implement `IBucketPickupHandler`, so it's possible that `World.getFluidState` could return something other than what `pickupFluid` returns.
- `pickupFluid` only returns a `Fluid`, so no amount or NBT
[...]

PR #6243, opened 2019-10-16
Fix FluidUtil#tryEmptyContainer causing duplicate destination filling

rubensworks:
While the source container was correctly being drained only once,
the destination fluid handler was incorrectly being filled twice.
This commit fixes that by passing the `doDrain` parameter to the internal fluid transfer method.

PR #6797, opened 2020-06-22
[1.15.x] Fix tryEmptyContainer() ignoring doDrain, and always filling the dest container

desht:
#6796


desht:
I believe this is still an issue (although I should raise a new request for 1.16..)

TheCurle:
@gigaherz @tterrag1098 @covers1624
This LTS feature seems like a useful addition.
I'm unsure which of you to assign to, so i'll leave this ping comment here.

TheCurle:
On second inspection, https://github.com/MinecraftForge/MinecraftForge/pull/6243#issuecomment-545140968 this comment was brought up to me.
Given that the same problems exist here, I'll close this until we can work out a better solution, if this PR is even needed any more.

PR #6803, opened 2020-06-25
Fix Forge experimental render pipeline lighting issue

aidancbrady:
This PR fixes the experimental lighting pipeline not properly taking unpacked BakedQuad lighting into account.

Specifically, VertexLighterFlat fails to scale light values unpacked via LightUtil from a scale of 0 -> (240 / Short.MAX_VALUE) to a scale of 0 -> 1, as is used by subsequent lighting logic. The proposed changes add this scaling to allow for proper comparison against scaled lighting of unpacked BakedQuads.


aidancbrady:
Leaving this open for 1.15, going to recreate for 1.16.

PR #6805, opened 2020-06-25
[1.15.x] Added event for when a particle is spawned in the world

ChristianANoeddeboe:
Adds a new event, which is used when a particle is spawned in the world.

This allows players to temporarily replace or change a spawned vanilla particle.

An example scenario is increasing the damage the player does. So far a high damage hit on an entity will display many damage particles, and this will help modders reduce these to a reasonable amount.

Testmod has been added as ParticleWorldHandleEventTest.java.

### DIFF
--- a/patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
+++ b/patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
@@ -306,7 +306,37 @@
        if (p_217401_2_ != null) {
           explosion.func_199592_a(p_217401_2_);
        }
-@@ -1377,4 +1421,15 @@
+@@ -1109,12 +1153,14 @@
+    }
+ 
+    public <T extends IParticleData> int func_195598_a(T p_195598_1_, double p_195598_2_, double p_195598_4_, double p_195598_6_, int p_195598_8_, double p_195598_9_, double p_195598_11_, double p_195598_13_, double p_195598_15_) {
+-      SSpawnParticlePacket sspawnparticlepacket = new SSpawnParticlePacket(p_195598_1_, false, p_195598_2_, p_195598_4_, p_195598_6_, (float)p_195598_9_, (float)p_195598_11_, (float)p_195598_13_, (float)p_195598_15_, p_195598_8_);
++      net.minecraftforge.event.ParticleWorldHandleEvent event = new net.minecraftforge.event.ParticleWorldHandleEvent(p_195598_1_, false, p_195598_8_, p_195598_2_, p_195598_4_, p_195598_6_, p_195598_9_, p_195598_11_, p_195598_13_, p_195598_15_);
++      if (net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(event)) return 0;
++      SSpawnParticlePacket sspawnparticlepacket = new SSpawnParticlePacket(event.getParticle(), event.isLongDistance(), (float)event.getXCoord(), (float)event.getYCoord(), (float)event.getZCoord(), (float)event.getXOffset(), (float)event.getYOffset(), (float)event.getZOffset(), (float)event.getParticleSpeed(), event.getParticleCount());
+       int i = 0;
+ 
+       for(int j = 0; j < this.field_217491_A.size(); ++j) {
+          ServerPlayerEntity serverplayerentity = this.field_217491_A.get(j);
+-         if (this.func_195601_a(serverplayerentity, false, p_195598_2_, p_195598_4_, p_195598_6_, sspawnparticlepacket)) {
++         if (this.func_195601_a(serverplayerentity, event.isLongDistance(), event.getXCoord(), event.getYCoord(), event.getZCoord(), sspawnparticlepacket)) {
+             ++i;
+          }
+       }
+@@ -1123,8 +1169,10 @@
+    }
+ 
+    public <T extends IParticleData> boolean func_195600_a(ServerPlayerEntity p_195600_1_, T p_195600_2_, boolean p_195600_3_, double p_195600_4_, double p_195600_6_, double p_195600_8_, int p_195600_10_, double p_195600_11_, double p_195600_13_, double p_195600_15_, double p_195600_17_) {
+-      IPacket<?> ipacket = new SSpawnParticlePacket(p_195600_2_, p_195600_3_, p_195600_4_, p_195600_6_, p_195600_8_, (float)p_195600_11_, (float)p_195600_13_, (float)p_195600_15_, (float)p_195600_17_, p_195600_10_);
+-      return this.func_195601_a(p_195600_1_, p_195600_3_, p_195600_4_, p_195600_6_, p_195600_8_, ipacket);
++      net.minecraftforge.event.ParticleWorldHandleEvent event = new net.minecraftforge.event.ParticleWorldHandleEvent(p_195600_2_, false, p_195600_10_, p_195600_4_, p_195600_6_, p_195600_8_, p_195600_11_, p_195600_13_, p_195600_15_, p_195600_17_);
++      if (net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(event)) return false;
++      IPacket<?> ipacket = new SSpawnParticlePacket(event.getParticle(), event.isLongDistance(), (float)event.getXCoord(), (float)event.getYCoord(), (float)event.getZCoord(), (float)event.getXOffset(), (float)event.getYOffset(), (float)event.getZOffset(), (float)event.getParticleSpeed(), event.getParticleCount());
++      return this.func_195601_a(p_195600_1_, event.isLongDistance(), event.getXCoord(), event.getYCoord(), event.getZCoord(), ipacket);
+    }
+ 
+    private boolean func_195601_a(ServerPlayerEntity p_195601_1_, boolean p_195601_2_, double p_195601_3_, double p_195601_5_, double p_195601_7_, IPacket<?> p_195601_9_) {
+@@ -1377,4 +1425,15 @@
           return p_229854_1_.func_175898_b(p_229855_1_.func_180328_a());
        });
     }
--- a/patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
+++ b/patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
@@ -306,45 +306,35 @@
        if (p_217401_2_ != null) {
           explosion.func_199592_a(p_217401_2_);
        }
-@@ -1109,22 +1153,25 @@
+@@ -1109,12 +1153,14 @@
     }
  
     public <T extends IParticleData> int func_195598_a(T p_195598_1_, double p_195598_2_, double p_195598_4_, double p_195598_6_, int p_195598_8_, double p_195598_9_, double p_195598_11_, double p_195598_13_, double p_195598_15_) {
 -      SSpawnParticlePacket sspawnparticlepacket = new SSpawnParticlePacket(p_195598_1_, false, p_195598_2_, p_195598_4_, p_195598_6_, (float)p_195598_9_, (float)p_195598_11_, (float)p_195598_13_, (float)p_195598_15_, p_195598_8_);
--      int i = 0;
--
--      for(int j = 0; j < this.field_217491_A.size(); ++j) {
--         ServerPlayerEntity serverplayerentity = this.field_217491_A.get(j);
++      net.minecraftforge.event.ServerParticleSpawnEvent event = net.minecraftforge.event.ForgeEventFactory.onWorldParticleSpawn(p_195598_1_, false, p_195598_2_, p_195598_2_, p_195598_6_, p_195598_8_, p_195598_9_, p_195598_11_, p_195598_13_, p_195598_15_);
++      if(event == null) return 0;
++      SSpawnParticlePacket sspawnparticlepacket = new SSpawnParticlePacket(event.getParticle(), event.isLongDistance(), event.getPosX(), event.getPosY(), event.getPosZ(), (float)event.getXOffset(), (float)event.getYOffset(), (float)event.getZOffset(), (float)event.getParticleSpeed(), event.getParticleCount());
+       int i = 0;
+ 
+       for(int j = 0; j < this.field_217491_A.size(); ++j) {
+          ServerPlayerEntity serverplayerentity = this.field_217491_A.get(j);
 -         if (this.func_195601_a(serverplayerentity, false, p_195598_2_, p_195598_4_, p_195598_6_, sspawnparticlepacket)) {
--            ++i;
-+      java.util.concurrent.atomic.AtomicInteger i = new java.util.concurrent.atomic.AtomicInteger();
-+      net.minecraftforge.event.ForgeEventFactory.onWorldParticleSpawn(p_195598_1_, false, p_195598_2_, p_195598_4_, p_195598_6_, p_195598_8_, p_195598_9_, p_195598_11_, p_195598_13_, p_195598_15_).ifPresent(event -> {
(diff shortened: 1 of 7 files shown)


GotoLink:
The event javadoc is talking about command, probably an error ?
The second argument is always false according to the patches. What happens if you force longDistance to true by using this event ?

ChristianANoeddeboe:
> 
> 
> The event javadoc is talking about command, probably an error ?
> The second argument is always false according to the patches. What happens if you force longDistance to true by using this event ?

Changing longDistance will just increase the range in which players can see the particle. Will not break anything, and may be useful, if you want to give the particle a bigger "effect".

patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
```diff
@@ -306,7 +306,37 @@
        if (p_217401_2_ != null) {
           explosion.func_199592_a(p_217401_2_);
        }
-@@ -1377,4 +1421,15 @@
+@@ -1109,12 +1153,14 @@
+    }
+ 
+    public <T extends IParticleData> int func_195598_a(T p_195598_1_, double p_195598_2_, double p_195598_4_, double p_195598_6_, int p_195598_8_, double p_195598_9_, double p_195598_11_, double p_195598_13_, double p_195598_15_) {
+-      SSpawnParticlePacket sspawnparticlepacket = new SSpawnParticlePacket(p_195598_1_, false, p_195598_2_, p_195598_4_, p_195598_6_, (float)p_195598_9_, (float)p_195598_11_, (float)p_195598_13_, (float)p_195598_15_, p_195598_8_);
++      net.minecraftforge.event.ParticleWorldHandleEvent event = new net.minecraftforge.event.ParticleWorldHandleEvent(p_195598_1_, false, p_195598_8_, p_195598_2_, p_195598_4_, p_195598_6_, p_195598_9_, p_195598_11_, p_195598_13_, p_195598_15_);
```
marchermans: Move the construction of the event, and the posting to the event bus to the `ForgeEventFactory`. This keeps all the events in one place and reduces the patch size here from 3 to 2, every little bit helps.

patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
```diff
@@ -306,7 +306,37 @@
...
+             ++i;
+          }
+       }
+@@ -1123,8 +1169,10 @@
+    }
+ 
+    public <T extends IParticleData> boolean func_195600_a(ServerPlayerEntity p_195600_1_, T p_195600_2_, boolean p_195600_3_, double p_195600_4_, double p_195600_6_, double p_195600_8_, int p_195600_10_, double p_195600_11_, double p_195600_13_, double p_195600_15_, double p_195600_17_) {
+-      IPacket<?> ipacket = new SSpawnParticlePacket(p_195600_2_, p_195600_3_, p_195600_4_, p_195600_6_, p_195600_8_, (float)p_195600_11_, (float)p_195600_13_, (float)p_195600_15_, (float)p_195600_17_, p_195600_10_);
+-      return this.func_195601_a(p_195600_1_, p_195600_3_, p_195600_4_, p_195600_6_, p_195600_8_, ipacket);
++      net.minecraftforge.event.ParticleWorldHandleEvent event = new net.minecraftforge.event.ParticleWorldHandleEvent(p_195600_2_, false, p_195600_10_, p_195600_4_, p_195600_6_, p_195600_8_, p_195600_11_, p_195600_13_, p_195600_15_, p_195600_17_);
```
marchermans: Again this can probably be handled better by using a method in `ForgeEventFactory` possibly returning an Optional which is empty when the event is cancelled, or contains the even if it is allowed to pass.

Would again reduce the patch size.

marchermans (changes requested):
You can reduce your patch sizes if you use create a relevant method in `ForgeEventFactory` and process it a bit there. Then you would only need to check the result of that method (which could be an optional, containing the fired event if it is allowed to pass) in the patched code. Reducing it from a multiline patch to a single line patch.

marchermans (changes requested):
Also a testmod is required, since this adds an event and modders should have an example on how to use it,

patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
ChristianANoeddeboe: EDIT: I've incorporated the changes, please tell me again if I need to change anything 😃 

I'm having a bit of trouble understanding how the optional will reduce patch size. Using an optional I get this in the ServerWorld.java:
```
public <T extends IParticleData> int spawnParticle(T type, double posX, double posY, double posZ, int particleCount, double xOffset, double yOffset, double zOffset, double speed) {
      AtomicInteger i = new AtomicInteger();
      ForgeEventFactory.onWorldParticleSpawn(type, false, posX, posY, posZ, particleCount, xOffset, yOffset, zOffset, speed).ifPresent(event -> {
         SSpawnParticlePacket sspawnparticlepacket = new SSpawnParticlePacket(event.getParticle(), event.isLongDistance(), (float)event.getXCoord(), (float)event.getYCoord(), (float)event.getZCoord(), (float)event.getXOffset(), (float)event.getYOffset(), (float)event.getZOffset(), (float)event.getParticleSpeed(), event.getParticleCount());
         for (ServerPlayerEntity serverplayerentity : this.players) {
            if (this.sendPacketWithinDistance(serverplayerentity, event.isLongDistance(), event.getXCoord(), event.getYCoord(), event.getZCoord(), sspawnparticlepacket)) {
               i.incrementAndGet();
            }
         }
      });
      return i.get();
   }

   public <T extends IParticleData> boolean spawnParticle(ServerPlayerEntity player, T type, boolean longDistance, double posX, double posY, double posZ, int particleCount, double xOffset, double yOffset, double zOffset, double speed) {
      AtomicBoolean result = new AtomicBoolean(false);
      ForgeEventFactory.onWorldParticleSpawn(type, longDistance, posX, posY, posZ, particleCount, xOffset, yOffset, zOffset, speed).ifPresent(event -> {
         SSpawnParticlePacket sspawnparticlepacket = new SSpawnParticlePacket(event.getParticle(), event.isLongDistance(), (float)event.getXCoord(), (float)event.getYCoord(), (float)event.getZCoord(), (float)event.getXOffset(), (float)event.getYOffset(), (float)event.getZOffset(), (float)event.getParticleSpeed(), event.getParticleCount());
         result.set(this.sendPacketWithinDistance(player, event.isLongDistance(), event.getXCoord(), event.getYCoord(), event.getZCoord(), sspawnparticlepacket));
      });
      return result.get();
   }
```

With the following in the ForgeEventFactory.java:
```
public static Optional<ParticleWorldHandleEvent> onWorldParticleSpawn(IParticleData particle, boolean longDistance, double posX, double posY, double posZ, int particleCount, double xOffset, double yOffset, double zOffset, double speed) {
        net.minecraftforge.event.ParticleWorldHandleEvent event = new net.minecraftforge.event.ParticleWorldHandleEvent(particle, longDistance, particleCount, speed, posX, posY, posZ, xOffset, yOffset, zOffset);
        if (net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(event)) return Optional.empty();
        return Optional.of(event);
    }
```

patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
```diff
@@ -306,7 +306,45 @@
...
++         for (ServerPlayerEntity serverplayerentity : this.field_217491_A) {
++            if (this.func_195601_a(serverplayerentity, event.isLongDistance(), event.getXCoord(), event.getYCoord(), event.getZCoord(), sspawnparticlepacket)) {
++               i.incrementAndGet();
++            }
+          }
+-      }
+-
+-      return i;
++      });
++      return i.get();
```
diesieben07: You can make this patch _way_ smaller by not using `Optional`. Make `onParticleSpawn` return the event, or null. Then just have one if statement that returns immediately if the event is cancelled (two lines so far). After that just re-assign the parameter variables using the event. Then your patch becomes only additions, which is much easier to maintain.

patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
```diff
@@ -306,7 +306,45 @@
...
+-      return this.func_195601_a(p_195600_1_, p_195600_3_, p_195600_4_, p_195600_6_, p_195600_8_, ipacket);
++      java.util.concurrent.atomic.AtomicBoolean result = new java.util.concurrent.atomic.AtomicBoolean(false);
++      net.minecraftforge.event.ForgeEventFactory.onWorldParticleSpawn(p_195600_2_, p_195600_3_, p_195600_4_, p_195600_6_, p_195600_8_, p_195600_10_, p_195600_11_, p_195600_13_, p_195600_15_, p_195600_17_).ifPresent(event -> {
++         SSpawnParticlePacket sspawnparticlepacket = new SSpawnParticlePacket(event.getParticle(), event.isLongDistance(), (float)event.getXCoord(), (float)event.getYCoord(), (float)event.getZCoord(), (float)event.getXOffset(), (float)event.getYOffset(), (float)event.getZOffset(), (float)event.getParticleSpeed(), event.getParticleCount());
++         result.set(this.func_195601_a(p_195600_1_, event.isLongDistance(), event.getXCoord(), event.getYCoord(), event.getZCoord(), sspawnparticlepacket));
++      });
++      return result.get();
```
diesieben07: See above.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -528,6 +529,12 @@ public static void onPlayerBrewedPotion(PlayerEntity player, ItemStack stack)
         MinecraftForge.EVENT_BUS.post(new PlayerBrewedPotionEvent(player, stack));
     }
 
+    public static Optional<ParticleWorldHandleEvent> onWorldParticleSpawn(IParticleData particle, boolean longDistance, double posX, double posY, double posZ, int particleCount, double xOffset, double yOffset, double zOffset, double speed) {
+        net.minecraftforge.event.ParticleWorldHandleEvent event = new net.minecraftforge.event.ParticleWorldHandleEvent(particle, longDistance, particleCount, speed, posX, posY, posZ, xOffset, yOffset, zOffset);
```
diesieben07: This is not a patch file, no need to use FQNs.

src/main/java/net/minecraftforge/event/ParticleWorldHandleEvent.java
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * ParticleWorldHandleEvent is fired before an {@link SSpawnParticlePacket} is sent.<br>
+ * <br>
+ * This event is {@link Cancelable}. <br>
+ * If the event is canceled, the execution of the command does not occur.<br>
+ * <br>
+ * This event is fired on the {@link MinecraftForge#EVENT_BUS}.<br>
+ **/
+@Cancelable
+public class ParticleWorldHandleEvent extends Event {
```
diesieben07: This is a bad name, imho. You should make it clear that:
1. This is a server-side event.
2. That this is about particle spawns.

My suggestion would be `ServerParticleSpawnEvent`.

diesieben07 (changes requested):

patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
marchermans: @ChristianANoeddeboe sorry to indeed have send you on a goose chase, i misread the PR and the patch. Using an Optional here is indeed not usefull. But you will need to have the method in ForgeEventFactory @diesieben07 is completely right here.

The best approach would probably be to combine what you had in the first place with the factory method either returning null to cancel the behaviour, or returning the processed event. This would indeed make it a two, possibly three, line patch.

Sorry for the inconvenience.

src/main/java/net/minecraftforge/event/ParticleWorldHandleEvent.java
marchermans: Probably good idea too, I initially thought it was fired on both sides, but I checked the code myself this morning and yeah a name indicating that this is a server side only event and what it is fired for, would be great.

ChristianANoeddeboe:
Thank you for your help @diesieben07 and @OrionDevelopment, I've changed the code now based on your feedback.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -528,6 +529,12 @@ public static void onPlayerBrewedPotion(PlayerEntity player, ItemStack stack)
         MinecraftForge.EVENT_BUS.post(new PlayerBrewedPotionEvent(player, stack));
     }
 
+    public static ServerParticleSpawnEvent onWorldParticleSpawn(IParticleData particle, boolean longDistance, double posX, double posY, double posZ, int particleCount, double xOffset, double yOffset, double zOffset, double speed) {
+        ServerParticleSpawnEvent event = new ServerParticleSpawnEvent(particle, longDistance, particleCount, speed, posX, posY, posZ, xOffset, yOffset, zOffset);
+        if (net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(event)) return null;
```
diesieben07: This is not a patch file. You don't need to use fully qualified names.

src/main/java/net/minecraftforge/event/ServerParticleSpawnEvent.java
```diff
@@ -0,0 +1,142 @@
...
+package net.minecraftforge.event;
+
+import net.minecraft.particles.IParticleData;
+import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.eventbus.api.Cancelable;
+import net.minecraft.network.play.server.SSpawnParticlePacket;
+import net.minecraftforge.eventbus.api.Event;
+
+/**
+ * ParticleWorldHandleEvent is fired before an {@link SSpawnParticlePacket} is sent.<br>
```
diesieben07: Missing name change here.

src/main/java/net/minecraftforge/event/ServerParticleSpawnEvent.java
```diff
@@ -0,0 +1,142 @@
...
+ * <br>
+ * This event is {@link Cancelable}. <br>
+ * If the event is canceled, the execution of the command does not occur.<br>
```
diesieben07: Copy-paste error?

diesieben07 (changes requested):
Few small changes needed still, other than that looks good now.

src/main/java/net/minecraftforge/event/ServerParticleSpawnEvent.java
ChristianANoeddeboe: Yep sorry about that, I've changed it now.

src/test/java/net/minecraftforge/debug/world/ServerParticleSpawnEventTest.java
```diff
@@ -0,0 +1,46 @@
...
+    public static final String MODID = "particle_world_handle_event_test";
+
+    /**
+     * The following will replace all {@link ParticleTypes#DAMAGE_INDICATOR} with {@link ParticleTypes#ANGRY_VILLAGER}
+     * as well as raising the Y-coord with 2, and doubling the particle speed.
+     * @param event
+     */
+    @SubscribeEvent
+    public static void onEvent(ServerParticleSpawnEvent event) {
+        if(event.getParticle().getParameters().equals(ParticleTypes.DAMAGE_INDICATOR.getParameters())) {
```
diesieben07: missing space after if

diesieben07 (changes requested):
Sorry, another thing I missed:
all your files are not formatted correctly, braces need to be on new lines.
please check the other files in Forge.

src/test/java/net/minecraftforge/debug/world/ServerParticleSpawnEventTest.java
ChristianANoeddeboe: Yeah, sorry I completely forgot to check what formatting you already used... I think I got them all corrected now.

diesieben07 (approved):

diesieben07:
1.15 is the current LTS version, please port your PR to 1.16. Features go in the current version first before being backported.

ChristianANoeddeboe:
I've created a 1.16 version of this now. @diesieben07

PR #6815, opened 2020-06-26
Fix z-level issues on item tooltips

Cyborgmas:
Vanilla changed its tooltip to be on zlevel 400 instead of 300.


covers1624 (approved):
Something i missed when updating GuiUtils

gigaherz:
The custom tooltip stuff needs to be redesigned to use the new MatrixStack logic.

PR #6818, opened 2020-06-26
[1.16.x] Store and expose language on ServerPlayerEntity

octylFractal:
This is needed by WorldEdit to have its strings translated according to the client's language, without needing WorldEdit to be installed on the client.

### DIFF
--- a/patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
@@ -15,7 +15,15 @@
  import javax.annotation.Nullable;
  import net.minecraft.advancements.CriteriaTriggers;
  import net.minecraft.advancements.PlayerAdvancements;
-@@ -464,6 +460,7 @@
+@@ -145,6 +141,7 @@
+    private boolean field_71147_cj = true;
+    private int field_71144_ck = -99999999;
+    private int field_147101_bU = 60;
++   public String language;
+    private ChatVisibility field_71143_cn;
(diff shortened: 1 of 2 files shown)


me4502 (approved):

covers1624 (approved):

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
```diff
@@ -15,15 +15,23 @@
  import javax.annotation.Nullable;
  import net.minecraft.advancements.CriteriaTriggers;
  import net.minecraft.advancements.PlayerAdvancements;
-@@ -464,6 +460,7 @@
+@@ -145,6 +141,7 @@
+    private boolean field_71147_cj = true;
+    private int field_71144_ck = -99999999;
+    private int field_147101_bU = 60;
++   public String language;
```
LexManos, line 22: Needs Nullable, and to not be public. We dont need random people trying to set the value and expecting it to do something. And having it a getter method means we can change API easier.

LexManos:
Needs more explanation, also translations don't exist on the server, the only thing the server should know is english.

octylFractal:
How do you suggest we translate WorldEdit then? We currently do not require it to be present on the client, so we map the `TranslatableComponent`s server-side using bundled translations, but we need this property added in Forge too to allow translating to work if WorldEdit is missing on the client-side.

me4502:
Also worth noting that this field existed beforehand, but in 1.16.1 Proguard removed it from the server jar file.

octylFractal:
Closing in favor of a future PR.

PR #6819, opened 2020-06-26
Retarget Block.Properties patch to the new AbstractBlock, reintroduce harvestLevel and harvestTool fields

TheCurle:
On request of one person on the forum, and one person on the Discord.
These used to be in the Block.Properties subclass before it got pulled out of Block and into AbstractBlock. 
The changes were still there, just pointing to the wrong location.


tterrag1098:
There's nothing assigning `lootTableSupplier` either, is that another missing patch?

TheCurle:
Ooh, you're right. That's a few lines below, forgot to snip it. Please hold..

marchermans (approved):
Looks good.
Compared it to the old code and patches and this indeed should fix the regression.

Nice work.

PR #6821, opened 2020-06-26
Alter default behaviour of World::isNightTime()

TheCurle:
By default, it checks if the world is the overworld - since this function is used by bees, it makes bees in a modded world more difficult.

isDayTime has a check for the Overworld which can be overriden, so this feels like the simplest way to solve this issue.

Solves https://forums.minecraftforge.net/topic/85661-1152-custom-dimensions-isnighttime/

### DIFF
--- a/patches/minecraft/net/minecraft/client/resources/Language.java.patch
+++ b/patches/minecraft/net/minecraft/client/resources/Language.java.patch
@@ -4,7 +4,7 @@
        this.field_135037_b = p_i1303_2_;
        this.field_135038_c = p_i1303_3_;
        this.field_135036_d = p_i1303_4_;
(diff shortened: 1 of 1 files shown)


TheCurle:
Whoops! Forgot how Github PRs work for a moment there. That's.. unfortunate

marchermans (approved):
At the moment this looks like the best approach. That said, does this not also change the behaviour of night and day time scheduling in modded dimensions? I am wondering how and if we are going to push this on into a dimension specific logic that modders can control, when and if the dynamic dimension registration logic is going to be added.

TheCurle:
Whoops, forgot to rebranch.

PR #6822, opened 2020-06-26
[1.16.x] Allow GlobalEntityTypeAttributes' EntityType -> AttributeModifierMap to be added to

itsmeow:
The only way to set attributes on an entity is by registering an `AttributeModifierMap` to `GlobalEntityTypeAttributes`. Unfortunately, this map is an `ImmutableMap` meaning it is impossible to set attributes on a modded entity. 

This PR remedies that by creating a second map, and adding a `put` method. It also avoids modifying the vanilla map so that the patch is more stable since it is likely the vanilla map will change.

This implementation also does get in the most efficient way I could come up with (`getOrDefault` will evaluate the second statement, and using `containsKey` requires going over one of the maps twice)

The `get` method (`func_233835_a_`) is only used in the constructor of `LivingEntity`, so it should be safe to register to until your entity is constructed.


TheCurle (approved):
Looks good.

marchermans (approved):
I do not like the way this is done in vanilla.
But at least this makes sense to patch it this way.

patches/minecraft/net/minecraft/entity/ai/attributes/GlobalEntityTypeAttributes.java.patch
```diff
@@ -0,0 +1,23 @@
...
++   private static final Map<EntityType<? extends LivingEntity>, AttributeModifierMap> FORGE_ATTRIBUTES = new java.util.HashMap<EntityType<? extends LivingEntity>, AttributeModifierMap>();
++   
++   public static AttributeModifierMap put(EntityType<? extends LivingEntity> type, AttributeModifierMap map) {
++       return FORGE_ATTRIBUTES.put(type, map);
++   }
+ 
+    public static AttributeModifierMap func_233835_a_(EntityType<? extends LivingEntity> p_233835_0_) {
+-      return field_233833_b_.get(p_233835_0_);
++       AttributeModifierMap map = field_233833_b_.get(p_233835_0_); return map != null ? map : FORGE_ATTRIBUTES.get(p_233835_0_);
```
gigaherz: There's no need to shove the return into the same line. Patches should be made as simple as possible but not to the extend it hurts readability unnecessarily

patches/minecraft/net/minecraft/entity/ai/attributes/GlobalEntityTypeAttributes.java.patch
```diff
@@ -0,0 +1,23 @@
...
++   private static final Map<EntityType<? extends LivingEntity>, AttributeModifierMap> FORGE_ATTRIBUTES = new java.util.HashMap<EntityType<? extends LivingEntity>, AttributeModifierMap>();
++   
++   public static AttributeModifierMap put(EntityType<? extends LivingEntity> type, AttributeModifierMap map) {
++       return FORGE_ATTRIBUTES.put(type, map);
```
gigaherz, line 10: Is it worth adding a sanity check to prevent copypasta calling put() twice on the same EntityType?

gigaherz (changes requested):

patches/minecraft/net/minecraft/entity/ai/attributes/GlobalEntityTypeAttributes.java.patch
itsmeow: There could be a case someone may want to override another mod's attribute map.

gigaherz: In that case, would we want to similarly allow adding properties to vanilla entities by favoring forge's output and using vanilla as a fallback?

itsmeow: Makes sense to me. Although if we were to do that it would probably require more in-depth features such as the ability to register on top of existing maps (they're builder based). This PR is mainly a quick addition so modders can actually USE attributes

itsmeow: Upon further inspection, it seems the maps have a bunch of public methods that allow this. So vanilla and modded can be modified. Although it appears some can be non mutable, so favoring Forge makes more sense. I'll do that.

itsmeow:
.. what? Why is it not building?

gigaherz:
Jenkins has issues with parallel execution, another PR was building at the same time, so this one errored. I'll give it a poke.

gigaherz (approved):
LGTM now.

PR #6824, opened 2020-06-26
Fix forge loading screen

ichttt:
- The early and the intermediary loading screen now have the new mojang background color
- The late loading gui now renders text correctly again (I'm not sure if it is the correct way to fix it, though)
- The MC log color is now white instead of black for better readability. Other colors may need to be changed as well to work better with the new red background


TheCurle (commented):
Looks good, i don't have any complaints about the way you chose to do the text rendering, STBEasy is used throughout Forge for other things.

PR #6825, opened 2020-06-26
Make Type classes safe to use in concurrent loading

ichttt:
This makes `ToolType` and `PlantType` threadsafe. It tries to avoid synchronisation as much as possible, and only does it when it needs to write a new value.

### DIFF
--- a/src/main/java/net/minecraftforge/common/PlantType.java
+++ b/src/main/java/net/minecraftforge/common/PlantType.java
@@ -19,7 +19,10 @@
 
 package net.minecraftforge.common;
 
+import net.minecraft.block.BlockState;
+import net.minecraft.util.Direction;
 import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.IBlockReader;
 import net.minecraft.world.IWorldReader;
 
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


src/main/java/net/minecraftforge/common/PlantType.java
```diff
@@ -41,24 +43,33 @@
...
      * @return the acquired {@link PlantType}, a new one if not found.
      */
     public static PlantType get(String name)
     {
-        return VALUES.computeIfAbsent(name, e -> {
-            if (INVALID_CHARACTERS.matcher(e).find())
-                throw new IllegalArgumentException("PlantType.get() called with invalid name: " + name);
-            return new PlantType(e);
-        });
+        PlantType type = VALUES.get(name);
```
gigaherz: I don't think calling get() while another thread is in the middle of computeIfAbsent is thread-safe

ichttt: So I did some reading, and it turns out computeIfAbset that ConcurrentHashMap actually allows the behaviour as I used it, however, computeIfAbsent is already enoguh, as the implementation for ConcurrentHashMap is written so it guranteeres that not race condition will occur

PR #6827, opened 2020-06-26
Fix creative screen GUI having the arrow buttons misaligned

Cyborgmas:
Simple fix, the person who ported the patch added the buttons before calling the gui's super.init(), meaning guiLeft and guiTop were not set.


TheCurle (approved):
Quick and easy. Looks good to me.

covers1624:
Derp, was me..

PR #6829, opened 2020-06-26
[1.16.x] Store and expose language on ServerPlayerEntity

me4502:
Prior to 1.16, this (private) field was available to mods. In 1.16, Proguard removed this field so this PR re-adds it.

The main use case here is to allow mods to handle localisation without having a client-side mod in place. WorldEdit uses this field inside our `sendMessage` method to strip out WorldEdit internal translation keys and replace them with the translated text. As we are generally server-only, this is required to provide a localised experience for non-English speakers.

This is useful for other mods that wish to do the same as we do, allowing servers to provide a localised experience without requiring a client mod to handle translations.

In vanilla Minecraft, only the English localisation is loaded onto the server, so no inbuilt system is set up to automatically handle custom translations here, meaning some manual effort is still required from the mod maker. This just makes the information required to do this available with a getter and stored field.

A future feature that adds a full translation system based on this field could potentially be added at a later date, but this feature would be a dependency of that either way. That would be a much larger undertaking, however, that we personally would be unable to make use of due to our requirement of being cross-platform.

If Forge did have a server-side translation bundle loading system, it would need to hook into a few places to strip out the components similar to what WorldEdit does, as there are numerous places where a TextComponent is sent via packets. The main of which is the chat packet.

Alternatives:
* We could hook Netty and listen to the CClientSettings packet

Supersedes https://github.com/MinecraftForge/MinecraftForge/pull/6818

### DIFF
--- a/patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
@@ -108,12 +108,20 @@
 +      //Allows mods to specify data that persists after players respawn.
 +      CompoundNBT old = p_193104_1_.getPersistentData();
 +      if (old.func_74764_b(PERSISTED_NBT_TAG))
-+          getPersistentData().func_218657_a(PERSISTED_NBT_TAG, old.func_74781_a(PERSISTED_NBT_TAG));
++         getPersistentData().func_218657_a(PERSISTED_NBT_TAG, old.func_74781_a(PERSISTED_NBT_TAG));
 +      net.minecraftforge.event.ForgeEventFactory.onPlayerClone(this, p_193104_1_, !p_193104_2_);
     }
  
     protected void func_70670_a(EffectInstance p_70670_1_) {
-@@ -1291,14 +1304,14 @@
+@@ -1177,6 +1190,7 @@
+    }
+ 
+    public void func_147100_a(CClientSettingsPacket p_147100_1_) {
++      this.language = p_147100_1_.field_149530_a;
+       this.field_71143_cn = p_147100_1_.func_149523_e();
+       this.field_71140_co = p_147100_1_.func_149520_f();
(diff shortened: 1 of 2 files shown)


patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
```diff
@@ -108,12 +108,20 @@
 +      //Allows mods to specify data that persists after players respawn.
 +      CompoundNBT old = p_193104_1_.getPersistentData();
 +      if (old.func_74764_b(PERSISTED_NBT_TAG))
-+          getPersistentData().func_218657_a(PERSISTED_NBT_TAG, old.func_74781_a(PERSISTED_NBT_TAG));
++         getPersistentData().func_218657_a(PERSISTED_NBT_TAG, old.func_74781_a(PERSISTED_NBT_TAG));
```
me4502, line 111: IDE corrected formatting

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
```diff
@@ -131,20 +139,35 @@
...
-@@ -1375,6 +1389,8 @@
+@@ -1375,7 +1390,9 @@
        if (itementity == null) {
           return null;
        } else {
+-         this.field_70170_p.func_217376_c(itementity);
 +         if (captureDrops() != null) captureDrops().add(itementity);
 +         else
-          this.field_70170_p.func_217376_c(itementity);
++            this.field_70170_p.func_217376_c(itementity);
```
me4502, line 157: Another IDE corrected formatting

LexManos:
It still needs to be addressed on how modders are supposed to use it, and get some form of generic server side translation system in place. However Vanilla translation components are used all over the place and would not be a simple system to automate. So this is a stop-gap until Mojang figures out server side content exists.

me4502:
Do you mean just further clarification in this PR? Or do you mean I need to write a system that handles that.

LexManos:
If you want to write it, it would be a good idea. But as discussed on discord, I doubt you will bother.

PR #6836, opened 2020-06-26
Quick and dirty server fix

Cyborgmas:
Right now, the server can not run at all.

This is due to the changes in how datapacks are loaded. The DedicatedServer's constructor now requires some datapack information and vanilla loads them before its creation. However, forge patches the loading to load mod datapacks as well. For that it needs the ModList, which is not generated yet, not until ModLoader runs. And ModLoader currently requires the DedicatedServer to begin creating a bit of hickup there.

So got the Server succesfully running, but did some ugly hacks for it (also my patches are ugly, might tidy them up later). 
What I did is strip out the problematic datapack stuff from Main and put it into DedicatedServer to be called on during # init() and passed in null into the DedicatedServer's constructor.

I am sure there are other cleaner ways to do it, but from what I understand it involves going a bit deeper into how mods are loaded and changing internal stuff.

### DIFF
--- a/patches/minecraft/net/minecraft/server/Main.java.patch
+++ b/patches/minecraft/net/minecraft/server/Main.java.patch
@@ -28,6 +28,94 @@
           SaveFormat saveformat = SaveFormat.func_237269_a_(file1.toPath());
           SaveFormat.LevelSave saveformat$levelsave = saveformat.func_237274_c_(s);
           MinecraftServer.func_240777_a_(saveformat$levelsave);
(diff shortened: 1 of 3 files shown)


Cyborgmas:
Hmmm it seems that I missed some stuff. There are still exceptions being thrown woops. Well, will investigate this a bit later

Cyborgmas:
Quick look at it seems to be a problem when re starting the server, with world data already existing, deleting the world folder resets it and it works fine the first time.

PR #6839, opened 2020-06-26
Replace ItemTags and BlockTags patches with ATs

JDLogic:
These should have been ATs not patches. Also `FluidTags#makeWrapperTag` was missed.


covers1624:
They _were_ when we were on the private branch.. I blame @illyohs

0xilly:
I stayed a mile away from that package :P

JDLogic:
Ughh.. going to change this to a patch cleanup PR because I have found other cases like this. I'll update when it is ready.

Also ItemTags was covers and BlockTags was Illy. I have proof :P

LexManos:
Yes, a full cleanup/sanity pass is be needed. This is also a nice spot to toss in the patch validator as well if someone wants to work on it.
Something that checks if patches:
Have tabs
Have Hunks that are JUST whitespace changes. Like the first hunk in BlockTags
Has import changes

pupnewfster:
`EntityTypeTags#func_232896_a_` (makeWrapperTag) was also missed.

PR #6795, opened 2020-06-22
[1.15] Fix being unable to add a custom milk fluid to vanilla milk buckets

KnightMiner:
Forge currently provides a fluid bucket wrapper that adds fluid support for vanilla buckets. However, on the milk bucket the wrapper does not properly work as Forge provides no milk fluid, and unlike in 1.12 there is no longer a global name for milk (#6769). This pull request removes the milk bucket wrapper as Forge has other systems in place that can meet the same need as that code was intended to meet.

To demonstrate the way modders can accomplish a custom milk and to make sure my changes work, this PR adds a test mod that adds a milk fluid and makes the milk bucket support it. You can test picking up milk by using `/setblock ~ ~ ~ milk_fluid_test:milk_block` then picking it up with a bucket. If you enable the fluid handler wrapper in the test mod, you can then right click the bucket and it will return the contained fluid.

Issue #6769, opened 2020-06-13
[1.15.2][1.16] Milk Bucket Fluid Wrapper forces the vanilla milk bucket to return no fluid


KnightMiner:
**Minecraft Version:** 1.15.2, 1.16.1

**Forge Version:** 31.2.9

**Description of issue:**

The Forge fluid bucket wrapper is currently hardcoded to only support `BucketItem`, however it is returned for the vanilla milk bucket. As a result, that bucket always [returns empty](https://github.com/MinecraftForge/MinecraftForge/blob/1.15.x/src/main/java/net/minecraftforge/fluids/capability/wrappers/FluidBucketWrapper.java#L71-L82). Normally this would not be a problem as we can just attach a new capability using the event, but the native capability provider [run before event added ones](https://github.com/MinecraftForge/MinecraftForge/blob/1.15.x/src/main/java/net/minecraftforge/common/capabilities/CapabilityDispatcher.java#L70-L78), causing the Forge one to win out. This means there is no way to add a milk fluid as a return for the bucket save a registry override on the milk bucket item.

[Other places](https://github.com/MinecraftForge/MinecraftForge/blob/1.15.x/src/main/java/net/minecraftforge/fluids/capability/wrappers/FluidBucketWrapper.java#L61-L68) in the code hardcode the fluid [to fill the milk bucket](https://github.com/MinecraftForge/MinecraftForge/blob/9114bec81d2bb705f8dbaa0ab70e4673fbfbb288/src/main/java/net/minecraftforge/fluids/FluidUtil.java#L644-L647) to `minecraft:milk`, which is equally bad as mods should not be registering fluids to the Minecraft namespace

**Solution:**

The logic in its current state is pretty useless, as there is no way for mods to extend it and the only remaining logic is redundant to the [default attribute getter](https://github.com/MinecraftForge/MinecraftForge/blob/9114bec81d2bb705f8dbaa0ab70e4673fbfbb288/src/main/java/net/minecraftforge/fluids/FluidUtil.java#L650). The logic could be fixed to actually be useful, such as using the `forge:milk` tag to find the proper fluid to return, but I think such a solution is overkill. The easiest solution I see is just removing all Forge patches related to the milk bucket fluid container, as its pretty easy for the mod adding milk to add their own capability to milk buckets.

I can make a pull request if desired, once I know the preferred solution. As I said, I am in favor of reducing patches and logic by just leaving milk buckets as not a fluid container unless a mod adds a capability.

PR #6840, opened 2020-06-26
[1.16] Fix being unable to add a custom fluid to vanilla milk buckets

KnightMiner:
Same as #6795, but for 1.16. Description copied below:

----

Forge currently provides a fluid bucket wrapper that adds fluid support for vanilla buckets. However, on the milk bucket the wrapper does not properly work as Forge provides no milk fluid, and unlike in 1.12 there is no longer a global name for milk (#6769). This pull request removes the milk bucket wrapper as Forge has other systems in place that can meet the same need as that code was intended to meet.

To demonstrate the way modders can accomplish a custom milk and to make sure my changes work, this PR adds a test mod that adds a milk fluid and makes the milk bucket support it. You can test picking up milk by using `/setblock ~ ~ ~ milk_fluid_test:milk_block` then picking it up with a bucket. If you enable the fluid handler wrapper in the test mod, you can then right click the bucket and it will return the contained fluid.


TheCurle (approved):
Looks good, and a quick fix for a well requested feature

ichttt:
Can you explain a little more why this code was there to begin with and how it is stopping modders to add milk. Is this allowing modders to register a new milk item with the properties of vanilla milk, or overwrite the vanilla milk item?

KnightMiner:
In 1.12 and before, fluids were not namespaced, so it made sense that any mod with a custom milk fluid would be registered as `milk`. Forge took advantage of this and special cased the vanilla milk bucket to return `milk`.

1.14+ has namespaced fluids, so there is no longer one sensible option. However, whoever ported the milk bucket patch decided that fluid buckets should fill from `minecraft:milk` (where mods should never register), and drain as `Fluids.EMPTY` (which makes the whole wrapper useless, this part is due to a lack of case for the milk bucket, in part because there is no sensible choice)

The introduction of capabilities as opposed to the old fluid container system means is now trivial for us to add our own fluid capability to the milk bucket to return our fluid. That is, it's trivial if it does not already have one. In the case of the milk bucket, since Forge added one (which as I said before is currently useless), it takes precedence over anything a mod could add. Thus, if you want to empty a milk fluid from the vanilla milk bucket, your only solution is a registry substitution on the vanilla item.

For the record, I explained most of this in the linked issue #6769 already. Not sure why this was marked "Work in progress" asking me to reiterate it again.

----

This PR is entirely about adding custom capabilities to existing containers, it has nothing to do with removing potion effects or other milk properties.

TheCurle:
This PR is ongoing, and needs further discussion with @KingLemming and the community to see if this feature is more useful over #7294.

KnightMiner:
#7294 was merged as the preferred solution, so closing this.

PR #6842, opened 2020-06-26
Fix TagsUpdateEvent not firing on the server

Cyborgmas:
The TagsUpdateEvent was firing too early, before mods were loaded and before the Forge EventBus had started, meaning it was shutdown and ignored the event.

There are also some changes in how tag Collections are managed, will tinker and submit more stuff here.


Cyborgmas:
NetworkTagManager#func_232980_f_ is responsible for updating all the Tags classes and firing the TagsUpdateEvent.

As such, I removed the firing of the event in the ClientPlayNetHandler, since it is using the new function, and already firing the event. I also added a call to this function inside NetworkTagManager#reload. 

I am uncertain of these changes, since there seems to be other ways Vanilla manages tags now (like the TagCollectionManager)

TheCurle:
This is becoming a high priority fix.
Can we get some reviews on this?

Cyborgmas:
Made changes that should avoid possible bugs/duplicate firing of the event. With what I did for other PRs, found out this makes more sense then what i previously did.

ghost (approved):

LexManos:
https://github.com/MinecraftForge/MinecraftForge/commit/3b64ea630514f89d9effcd10b841540157685710
Your approach and understanding of the issue was completely wrong.

PR #6844, opened 2020-06-26
Fix BreakSpeedEvent having pos always null

### LINKED ISSUE #6837: BreakSpeed event block pos always null
**Minecraft Version:** 1.16.1

**Forge Version:** 32.0.16

**Description of issue:**
The [position variable](https://github.com/MinecraftForge/MinecraftForge/blob/a1110b92d031f74a910f7be61483f918a9cfcb75/src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java#L119) from the BreakSpeed PlayerEvent is always null due to the fact that in the AbstractBlock class from the Minecraft code the method "getPlayerRelativeBlockHardness" now uses the method "getDigSpeed(BlockState)" from the PlayerEntity class instead of the "getDigSpeed(BlockState, BlockPos)" from the same class. This has the effect that the block position is always null as the first "getDigSpeed" method gives the secound "getDigSpeed" method null as the block position.

Cyborgmas:
Fixes #6837

### DIFF
--- a/patches/minecraft/net/minecraft/block/AbstractBlock.java.patch
+++ b/patches/minecraft/net/minecraft/block/AbstractBlock.java.patch
@@ -45,15 +45,17 @@
     }
  
     @Deprecated
-@@ -293,7 +297,7 @@
+@@ -293,8 +297,8 @@
        if (f == -1.0F) {
           return 0.0F;
        } else {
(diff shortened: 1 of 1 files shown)


TheCurle:
I hope you meant BreakSpeedEvent :P

Issue #6833, opened 2020-06-26
1.16.1 How to deal with fabulous graphics options for BlockItem entities


Cyborgmas:
While making #6832 noticed that the custom block item entity I debugged with was not getting the "fabulous" render type (Atlases.field_239279_C_) while vanilla ones were.

This is because in ItemRenderer#renderItem , flag1 is set to an instance check of BreakableBlock and StainedGlassPaneBlock and it is responsible for whether or not the entity will receive fancy or fabulous transparency.

This is of course an easy fix, I made the custom block extend BreakableBlock and it works, receiving teh correct transparency.

However, unsure if forge would want to make a better way of determining if a block should be Fabulously transparent. As in have a function in IForgeBlock and IForgeBlockState to do this instead.

TheCurle:
It may be worth refactoring BreakableBlock to TranslucentBlock, or ShaderBlock. 
Thoughts?

PR #6847, opened 2020-06-27
Add fabulous graphics helper when rendering a Block as ItemEntity

Cyborgmas:
What I think should be done to answer my own question in #6833. 

Modders can then override the new function and return true in blocks having "fabulous compatible" render types.


TheCurle (commented):
We need to do something about the atrocious naming of BreakableBlock. If it's more suited to handling some of this logic, we should move it into there.

I think you're on the right idea here, though.

covers1624:
Perhaps we should use RenderTypeLookup for this instead, it would be cleaner imo.

Cyborgmas:
Yeah it seems a bit of a mess though. Also it means more patches. I'll give it another look

covers1624:
I would rather have this over there, as thats where our existing patches for RenderType stuff is, id rather not resort to shouldRenderBlah methods like we had in 1.12.

Cyborgmas:
Okay using RenderTypeLookup now, tested it with a custom block and it works

TheCurle:
If covers is happy with this, we'll assign it

patches/minecraft/net/minecraft/client/renderer/RenderTypeLookup.java.patch
```diff
@@ -87,6 +92,12 @@
 +      setRenderLayer(block, type::equals);
 +   }
 +
++   public static void setRenderLayerWithFabulous(Block block, RenderType type) {
++      java.util.Objects.requireNonNull(type);
++      java.util.function.Predicate<RenderType> predicate = type::equals;
++      setRenderLayer(block, predicate.or(t -> t.equals(Atlases.func_239280_i_())));
```
covers1624: This will break if a mod already uses `setRenderLayer` right?

Cyborgmas: Not sure what you mean

covers1624: If my mod calls `RenderTypeLookup.setRenderLayer(MyBlocks.MY_BLOCK, e -> e == RenderType.getTransparent())`
then calls `RenderTypeLookup.setRenderLayerWithFabulous(MyBlocks.MY_BLOCK, Atlases.func_239280_i_())`
`setRenderLayerWithFabulous` will overwrite my  `setRenderLayer` call. Just store these in a seperate map instead of reusing setRenderLayer's map.

Cyborgmas: okay was not sure why that would be a problem, but it could if you want a predicate other than equals ? If thats the case i can just add another overload no?

Cyborgmas:
Made it a separate map, does seem a bit nicer. 
Something I am not sure about is I made setRenderLayer automatically do setFabulousRenderLayer. However, there is also the non-predicate version that can be used to overwrite it.

Cyborgmas:
Added a fix to the piston renderer which uses the fabulous render layer map

TheCurle:
Bump. Is there a reason this was closed?

TheCurle:
I was notified on Discord that this was closed because Cyborgmas was unhappy with the solution. Closing again

PR #6848, opened 2020-06-27
Fix access levels being hardcoded to private via patch overriding AT entry

pupnewfster:
I was trying to track down why these [two](https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/src/main/resources/META-INF/accesstransformer.cfg#L101-L102) AT entries were not working. And discovered it was due to them being [patched](https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/patches/minecraft/net/minecraft/data/loot/EntityLootTables.java.patch#L48-L53) back to private after the AT got performed.

### DIFF
--- a/patches/minecraft/net/minecraft/data/loot/EntityLootTables.java.patch
+++ b/patches/minecraft/net/minecraft/data/loot/EntityLootTables.java.patch
@@ -32,11 +32,10 @@
              if (resourcelocation != LootTables.field_186419_a && this.field_218587_b.remove(resourcelocation) != null) {
                 break;
              }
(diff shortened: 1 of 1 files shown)


LordMonoxide:
I think I may have run into this before, since I had one specific AT that I couldn't get to work for the life of me. Is there some sort of fix that might cover all cases?

LexManos:
JD is working on a [validator](https://github.com/MinecraftForge/MinecraftForge/pull/6851) that will check and verify these don't happen.
This happens when someone edits a AT before running genPatches without re-running setup.

Issue #6838, opened 2020-06-26
[1.16.1] No longer able to register custom reload listeners on server-side


BlayTheNinth:
**Minecraft Version:** 1.16.1

**Forge Version:** 32.0.17

The server now uses an instance of `DataPackRegistries` which creates its own `IReloadableResourceManager`, and on reload, creates a new instance of that (which also results in a new `IReloadableResourceManager`) via `func_240961_a_`.

This means that adding a custom reload listener in the existing server lifecycle events will no longer have any effect, as the list of listeners on reload will be a fresh list that no longer contains the custom listener previously added.

Before 1.16, custom reload listeners [could be added](https://github.com/blay09/FarmingForBlockheads/blob/1.15.x/src/main/java/net/blay09/mods/farmingforblockheads/FarmingForBlockheads.java#L86-L89) in `FMLServerAboutToStartEvent`, since ResourceManagers didn't get recreated back then.

There probably needs to be a new event fired after instantiation of `DataPackRegistries` or just before the reload to allow for mods to add custom listeners on the server side.

Use Case: Allowing data packs to be used for gameplay configuration, e.g. market trades, mod compat, etc.:
- https://github.com/blay09/KleeSlabs/blob/1.16.x/src/main/java/net/blay09/mods/kleeslabs/registry/json/JsonCompatLoader.java
- https://github.com/blay09/FarmingForBlockheads/blob/1.15.x/src/main/java/net/blay09/mods/farmingforblockheads/registry/market/MarketRegistryLoader.java
- https://github.com/blay09/CookingForBlockheads/blob/1.15.x/src/main/java/net/blay09/mods/cookingforblockheads/compat/json/JsonCompatLoader.java

PR #6849, opened 2020-06-27
Fix and tweak how resource reload listeners are added

Cyborgmas:
From #6838, the new resource manager is DataPackRegistries. However instead of creating it once, it is created on each reload.

Not sure if my fix is proper/in line with how forge does things.


src/main/java/net/minecraftforge/resource/IForgeDataPackRegistries.java
```diff
@@ -0,0 +1,22 @@
...
+
+import net.minecraft.resources.IFutureReloadListener;
+import net.minecraftforge.common.loot.LootModifierManager;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Stream;
+
+public interface IForgeDataPackRegistries {
+   List<IFutureReloadListener> resources = new ArrayList<>();
```
covers1624: Because this is in an interface, this is public, it should not be.

covers1624 (commented):
Why not just add an event to add things instead of this static 'registry'?

Cyborgmas:
Yeah the list being public felt weird. What are the thoughts about this way with the Event?

covers1624:
Yeah looks better this way, needs docs for when the event is fired, on what bus, and what its purpose is.

patches/minecraft/net/minecraft/resources/DataPackRegistries.java.patch
```diff
@@ -1,28 +1,10 @@
...
  
     public FunctionReloader func_240960_a_() {
-@@ -85,4 +88,8 @@
-    public void close() {
-       this.field_240952_b_.close();
-    }
-+
-+   public net.minecraftforge.common.loot.LootModifierManager getLootModifierManager() {
-+      return lootManager;
-+   }
```
bignelli, line 27: Wasn't this getter used by Global Loot Modifiers ? I think they are broken right now though.

Cyborgmas: woops ill re add a getter

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -23,8 +23,8 @@
 import net.minecraft.entity.ai.attributes.RangedAttribute;
 import net.minecraft.util.SoundEvent;
 import net.minecraft.world.storage.IServerConfiguration;
-import net.minecraft.world.storage.IWorldInfo;
```
pupnewfster: Why are the imports in this file changing?

Cyborgmas:
Fixed the things added docs. however not sure about how I handle getting the Instance in LootModifier. should be fine

src/main/java/net/minecraftforge/event/AddReloadListenerEvent.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * The main ResourceManager is recreated on each reload, through {@link DataPackRegistries}'s creation.
+ *
+ * The event is fired on each reload and lets modders add their own ReloadListeners.
+ * The event is fired on the {@link MinecraftForge#EVENT_BUS}
+ */
+public class AddReloadListenerEvent extends Event
```
KnightMiner, line 18: Might want to emphasize this is for server reload listeners. Only clue in the javadoc is the link to data pack registries, but I could see someone getting confused and trying to use this for the client side resource pack reload listeners.

pupnewfster:
Brief summary of the core part of the issue (as requested by lex to remember), in 1.16 every time `/reload` is ran the reload listeners list is nuked and rebuilt. This PR solves the issue of mods not being able to reliably add reload listeners by adding an event to add listeners to that is fired when the reload listener list is being rebuilt.

src/main/java/net/minecraftforge/common/loot/LootModifierManager.java
```diff
@@ -154,4 +161,17 @@ private IGlobalLootModifier deserializeModifier(ResourceLocation location, JsonE
         return registeredLootModifiers.values();
     }
 
+    @SubscribeEvent
+    public static void onResourceReload(AddReloadListenerEvent event)
```
LexManos: Forge has its own internal handler, so this shouldn't have this weird meta instance thing...
What is the point of people having access to the instance?

Cyborgmas: used here https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/src/main/java/net/minecraftforge/common/ForgeHooks.java#L1169 but commented for now. I can put this in the InternalHandler if its better suited.

pupnewfster: Yes this should be moved to the internal handler as there is no sense in exposing a modder facing API if we can avoid it when it comes to internal instances of things.

Cyborgmas:
Okay re-enabled global loot modifiers as well, this will close #6863

PR #6851, opened 2020-06-27
Add simple patch checker and cleanup patches

JDLogic:
Adds a simple patch verifier that checks for the following:
- Any tabs
- Any import changes
- Hunks that only have white space changes
- Disconnected white space changes (See `EntityRendererManager` patch change)
- Fields with access change and/or final modifier changed

There is also a task that will automatically fix whatever errors it can. Example output before these patch changes.

```
> Task :forge:checkPatchesAndFix
Patch contains white space change in valid hunk at 32 (cannot auto fix), file: patches/minecraft/net/minecraft/entity/monster/ShulkerEntity.java.patch
Removing white space hunk starting at 43, file: patches/minecraft/net/minecraft/entity/monster/ShulkerEntity.java.patch
*** Updating patch file. Please run setup then genPatches again. ***
Fixing tabs on line 9, file: patches/minecraft/net/minecraft/network/play/client/CClientSettingsPacket.java.patch
*** Updating patch file. Please run setup then genPatches again. ***
Removing white space hunk starting at 3, file: patches/minecraft/net/minecraft/world/gen/IWorldGenerationBaseReader.java.patch
Patch is now empty removing, file: patches/minecraft/net/minecraft/world/gen/IWorldGenerationBaseReader.java.patch
Patch contains white space change in valid hunk at 16 (cannot auto fix), file: patches/minecraft/net/minecraft/client/renderer/entity/EntityRendererManager.java.patch
Removing white space hunk starting at 84, file: patches/minecraft/net/minecraft/client/renderer/FirstPersonRenderer.java.patch
*** Updating patch file. Please run setup then genPatches again. ***
Patch contains white space change in valid hunk at 58 (cannot auto fix), file: patches/minecraft/net/minecraft/client/renderer/FogRenderer.java.patch
```
Edit: Mention field check


covers1624 (approved):

gradle/patchVerifer.gradle
```diff
@@ -0,0 +1,194 @@
+import java.util.regex.Pattern
+
+def ignore = [
+        "patches/minecraft/net/minecraft/client/renderer/ViewFrustum.java.patch",
+        "patches/minecraft/net/minecraft/data/BlockModelDefinition.java.patch"
```
LexManos: These should verify, but also just delete the patch silently if the patches are ONLY the S2S artifacts.

build.gradle
```diff
@@ -137,6 +137,8 @@ project(':forge') {
     apply plugin: 'net.minecraftforge.gradle.patcher'
     apply plugin: 'net.minecrell.licenser'
     apply plugin: 'de.undercouch.download'
+
+    apply from: "$rootDir/gradle/patchVerifer.gradle"
```
LexManos: The better way to do this would be to define the task in buildSrc similar to how we do things in MCPConfig.

patches/minecraft/net/minecraft/network/play/server/SJoinGamePacket.java.patch
```diff
@@ -1,10 +0,0 @@
---- a/net/minecraft/network/play/server/SJoinGamePacket.java
-+++ b/net/minecraft/network/play/server/SJoinGamePacket.java
-@@ -21,6 +21,7 @@
-    private boolean field_149204_b;
-    private GameType field_149205_c;
-    private GameType field_241785_e_;
-+   private int dimensionInt;
-    private Set<RegistryKey<World>> field_240811_e_;
```
LexManos, line 8: This seems incorrect, to be automatically fixed, but yes this int should also be deleted because we no longer use it. Was this and the others below this done manually?

LexManos (changes requested):
Can we make it so that the genPatches task is finalized bu this task

patches/minecraft/net/minecraft/network/play/server/SJoinGamePacket.java.patch
JDLogic: Yes, this was from a manual sweep over all the patches to catch little things like this that slipped in during the update process. I should have mentioned that.

PR #6853, opened 2020-06-27
Modded Nether Biomes

valoeghese:
Simple pull request which adds functionality for mods to edit the Nether Biomes that can generate in the nether.

The specific placement details for nether biomes are already handled by the vanilla `Biome.Settings` class, so all forge needs to do is add a way for modders to add their nether biomes to the list.

This functionality is useful in any biome mods which wish to enhance the nether :)


patches/minecraft/net/minecraft/world/biome/provider/NetherBiomeProvider.java.patch
```diff
@@ -0,0 +1,11 @@
+--- a/net/minecraft/world/biome/provider/NetherBiomeProvider.java
++++ b/net/minecraft/world/biome/provider/NetherBiomeProvider.java
+@@ -77,7 +77,7 @@
+    }
+ 
+    private static NetherBiomeProvider func_235285_d_(long p_235285_0_) {
+-      ImmutableList<Biome> immutablelist = ImmutableList.of(Biomes.field_235254_j_, Biomes.field_235252_ay_, Biomes.field_235253_az_, Biomes.field_235250_aA_, Biomes.field_235251_aB_);
++      ImmutableList<Biome> immutablelist = net.minecraftforge.common.BiomeManager.getNetherBiomes();
+       return new NetherBiomeProvider(p_235285_0_, immutablelist.stream().flatMap((p_235273_0_) -> {
```
ChloeDawn: These should be expression lambdas and it can all be on a single line

ChloeDawn (changes requested):

patches/minecraft/net/minecraft/world/biome/provider/NetherBiomeProvider.java.patch
pupnewfster: The lambda is not part of this PR, it is just showing due to how patches work. I believe at one point there was a ForgeFlower issue/PR related to lambdas that can be inlined which would address your comment. Regardless it doesn't actually have to do with this PR as it is't one of the lines modified.

ChloeDawn: Oh right my bad, I misread the diff

valoeghese: yeah it's a one line patch lol

TheCurle (approved):
Looks good to me.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -103,6 +105,11 @@ public static void addBiome(BiomeType type, BiomeEntry entry)
         if (list != null) list.add(entry);
     }
 
+    public static void addNetherBiome(Biome biome)
```
diesieben07: Lacking documentation, when should this be called? I assume you intended `FMLCommonSetupEvent`?

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -114,6 +121,11 @@ public static void removeBiome(BiomeType type, BiomeEntry entry)
         }
     }
 
+    public static void removeNetherBiome(Biome biome)
```
diesieben07: Lacking documentation, when should this be called? I assume you intended `FMLCommonSetupEvent`?

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -37,6 +38,7 @@
     private static TrackedList<BiomeEntry>[] biomes = setupBiomes();
 
     public static List<Biome> oceanBiomes = new ArrayList<Biome>();
+    private static List<Biome> netherBiomes = Lists.newArrayList(Biomes.field_235254_j_, Biomes.field_235252_ay_, Biomes.field_235253_az_, Biomes.field_235250_aA_, Biomes.field_235251_aB_);
```
diesieben07: This will most likely be accessed during parallel mod loading and thus needs to be threadsafe.

diesieben07 (changes requested):

patches/minecraft/net/minecraft/world/biome/provider/NetherBiomeProvider.java.patch
```diff
@@ -0,0 +1,11 @@
+--- a/net/minecraft/world/biome/provider/NetherBiomeProvider.java
++++ b/net/minecraft/world/biome/provider/NetherBiomeProvider.java
+@@ -77,7 +77,7 @@
+    }
+ 
+    private static NetherBiomeProvider func_235285_d_(long p_235285_0_) {
+-      ImmutableList<Biome> immutablelist = ImmutableList.of(Biomes.field_235254_j_, Biomes.field_235252_ay_, Biomes.field_235253_az_, Biomes.field_235250_aA_, Biomes.field_235251_aB_);
++      ImmutableList<Biome> immutablelist = net.minecraftforge.common.BiomeManager.getNetherBiomes();
```
diesieben07: You can simplify this patch and also avoid copying the biome list into the forge class:
`+ immutablelist = net.minecraftforge.common.BiomeManager.getNetherBiomes(immutablelist);`

src/main/java/net/minecraftforge/common/BiomeManager.java
valoeghese: I'll make it a synchronized list

patches/minecraft/net/minecraft/world/biome/provider/NetherBiomeProvider.java.patch
valoeghese: By copying the biome list into the forge class I give modders the ability to remove vanilla biomes if they wish, however.

Jonathing: Wouldn't giving modders the ability to remove vanilla biomes bring up the possibility of having mods conflict with each other?

valoeghese: That is true - in the case that one modder wishes to expand upon a biome that is then removed or replaced by another. Not sure whether it calls for removing the functionality entirely, however.

The other alternative is that modders who wish to remove or replace a vanilla biome would need to registry replace that biome or the nether dimension.

valoeghese: Or use a world type, which is by far a safer route, but that would make incompatibilities with other world type mods.

Jonathing: A registry could work, but something like removing integral things from vanilla should probably only be done via ASM so it can't be easily done. In any case, the only mods that should directly control what biomes are to be removed are mods that have their own world gen (i.e. Biomes O' Plenty, just to name one).

diesieben07 (changes requested):
A synchronized list is not enough.
`getNetherBiomes` copies (and thus iterates) the list. `synchronizedList` will only synchronize the call to `List#iterator`, the iteration itself is not guarded.
So if `removeNetherBiome` or `addNetherBiome` are called concurrently with `getNetherBiomes` you still have unsynchronized behavior.

patches/minecraft/net/minecraft/world/biome/provider/NetherBiomeProvider.java.patch
valoeghese: Should I remove the `removeNetherBiome` method as well, since it will only affect modded biomes and most mods won't need to remove their own biomes or other mod's biomes?

valoeghese:
Ah, ok.

patches/minecraft/net/minecraft/world/biome/provider/NetherBiomeProvider.java.patch
diesieben07: > A registry could work, but something like removing integral things from vanilla should probably only be done via ASM so it can't be easily done. 

No.
I agree though, removing things sounds like a bad idea.

Jonathing: > Should I remove the `removeNetherBiome` method as well, since it will only affect modded biomes and most mods won't need to remove their own biomes or other mod's biomes?

I personally would remove that method. It's the only criticism I have about this pull request. Other than that, it's great.

valoeghese:
Would switching from a synchronised list to directly using synchronised{} blocks be better? Or is there a better way you recommend?

patches/minecraft/net/minecraft/world/biome/provider/NetherBiomeProvider.java.patch
valoeghese: I've removed that method. Will be gone in my next commit.

valoeghese:
I'm gonna write a proper test mod

valoeghese:
Ok, should be ready for review!
I took some time to write a working test mod to go with the pr :).

ghost (approved):
Its looking good to me, short patch that clearly works well.

valoeghese:
Fixed conflicts

diesieben07 (approved):

diesieben07:
Okay, here is the thing...
The same thing needs to be done for overworld and end... Do we really just want to copy-paste all this stuff for overworld/nether/end?

ghost:
Yeah, we're gonna need a full biome system.  I was literally working on it, before I discovered that all of BiomeManager was nuked and has to be reconsidered.  While I could just reimplement it, there really should be proper discussions over what we want it to achieve and look like, especially as we need to support multiple dimensions.

valoeghese:
Unfortunately, the Overworld, Nether, and End all use different methods for placing biomes. Because of this they each need their own apis.

Jonathing:
iirc 1.16.2 is nuking the way biomes are done right now and are doing them as datapack JSONs, so it might not get much better. This PR is probably the best we got for 1.16.1, although feel free to correct me if I'm wrong.

tterrag1098:
"Nuking" is not the right word, please don't spread rumors about things that aren't that big of a deal. Biomes have been serializeable for some time now, this won't be an earth-shattering change.

valoeghese:
1.16.2 will make it serialisable. However, the patch will be different, nether stuff is initialised in a method that provides a biome provider with the default nether biomes.

valoeghese:
The patch will still be useful, as mods will likely want to add biomes to the nether on first world creation, before it is serialised into the data pack.

Jonathing:
Sweet. Thanks letting me know!

valoeghese:
When 1.16.2 comes out the patch will be slightly more complex. Perhaps it would be better to wait for 1.16.2 so I can move the patch to the new location.

0Navis0:
It would be good to have something similar for end biomes.

valoeghese:
End biomes are out of scope for this PR, and would be better suited in a separate PR.

0Navis0:
Which is why I suggest to make something similar for end biomes. ;)

valoeghese:
So 1.16.2 is out now. I'll update this pr soon

valoeghese:
I've started work on updating now. Since forge seems to want modders to use json biomes, I'm gonna make the methods take a `ResourceLocation`, so that the biome can be defined in json.

valoeghese:
It seems that vanilla uses a `RegistryKey<Biome>` instead, so I'll use that

valoeghese:
I've written the patch and methods, just need to update the testmod

valoeghese:
(and by written the patch I mean generated it with gradle after making a 2-line change)

valoeghese:
Can confirm I can add the plains biome to the nether in 1.16.2
image

valoeghese:
Ready for review on 1.16.2

CorgiTaco (approved):

bageldotjpg (approved):

ChampionAsh5357:
Since dimensions are data-driven, wouldn't it make more sense to expand this into the dynamic registry system and apply and appendage to the nether there?

valoeghese:
I'm injecting into the preset creation. It would probably be unneccesary work to replace this with datapacks if that's what you're talking about.

valoeghese:
Biomes themselves are still done with datapacks, and merely added to the preset by using the registry key with this method

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -123,7 +139,22 @@ public static void removeBiome(BiomeType type, BiomeEntry entry)
...
+*/
+    public static ImmutableList<Pair<Biome.Attributes, Supplier<Biome>>> getNetherBiomes(Registry<Biome> biomeRegistry, Pair<Biome.Attributes, Supplier<Biome>>... biomes)
+    {
+        ImmutableList.Builder<Pair<Biome.Attributes, Supplier<Biome>>> result = ImmutableList.builder();
+        result.add(biomes);
+
+        synchronized (moddedNetherBiomes)
+        {
+            for (Pair<RegistryKey<Biome>, Biome.Attributes> moddedBiome : moddedNetherBiomes) {
+                result.add(Pair.of(moddedBiome.getSecond(), () -> biomeRegistry.func_243576_d(moddedBiome.getFirst())));
```
Cyborgmas: You can do moddedBiome.swap().mapSecond(s -> ), a bit cleaner imo.

Cyborgmas, line 151: You can do moddedBiome.swap().mapSecond(s -> ), a bit cleaner imo.

Cyborgmas: A bit more relevant `func_243576_d` is `getOrThrow`, which is fine for vanilla biomes but will crash on datapack biomes.
This is because there are multiple registries that exist for the same type, vanilla only are the static ones in `WorldGenRegistries`, while datapack ones are held in `DatapackRegistries` and are rebuilt.
The way it works is that pure vanilla is loaded once, after the first "Create new world" button click, and a copy of the vanilla only registry will be passed here (through `DatapackRegistries`), making datapack biomes crash since they are not present.
When clicking the other "Create new world" button, the objects (like the biomeproviders) are built again but this time including the objects coming from datapacks (like the biomes).
Basically, datapack biomes should get loaded, but won't because it'll crash the first time.
You could simply use one of the function that returns an optional and check if present first, but it's a bit dirty.

Redstoneguy129 (approved):

src/main/java/net/minecraftforge/common/BiomeManager.java
valoeghese: So, in order to support datapack biomes, should I use a different method? I'm trying to make it work so that datapack biomes can be added with this.

valoeghese: I mean, it probably depends on when the supplier is resolved

Cyborgmas: I mean you can test it https://gist.github.com/Cyborgmas/fa514d3297125e98ffb8377865891c2b and see what happens and find a proper fix. Otherwise I am working on another system that'll allow to modify it in a more data-driven way.

SmellyModder (approved):

Vinetos (approved):

src/main/java/net/minecraftforge/common/BiomeManager.java
valoeghese: If you're making modded nether biomes data driven there's likely be no further need for this pr since forge afaik has said they want to head that way I think. Though I will maintain this PR until another, preferred PR opens

TheCurle:
If this feature is still wanted, please rebase and fix conflicts.

valoeghese:
Yeah I'll push a commit soon

Max094Reikeb:
Is it gonna be added soon?

TelepathicGrunt:
Same Max. I too would like to know as there's a severe conflict between nether mods at the moment. People's only option is pretty much to replace the vanilla nether biome provider with their own and try their best to import other mod's nether biomes to various degrees of success. It's pretty much chaos right now.

LexManos:
There are a bunch of conflicts that prevent this from being pulled.
It also has a rather dirty non-revival of the BiomeManager class. Leaving a ton of stuff commented out and interspersing their code in random spots.
That class itself just needs to be deleted as all of the Biome things have changed in 1.16.2 and from what I've seen elseware we have all the hooks needed to replace that entire class already.
Now, to address the attempted solution here directly.
Where are the modded biomes added to the overworld?
Why is adding biomes to the nether different then adding them to the overworld?
What happens if someone decides to remove a biome from the preset?
Does it corrupt worlds or is the nether biome list held in the world data?

TheCurle:
I'm working on this, along with a bunch of other biome hooks.
Along with the fact that this PR was pre-BiomeManager revival, it seems like this is going to be superseded soon.
I'll leave it open until that.

valoeghese:
I don't think I'll be working on this PR as I'm a lot more busy now and Curle already intends to re-implement this in a way that can support the rewrite

PR #6369, opened 2019-12-18
[1.15] Add player size and pose event

maxanier:
1.15 version of  #6166.
I went ahead and combined the size event with the eye height event, so only one instead of two events are thrown. Downsides are that the patch size is slightly larger and it breaks binary compatibility. 
If you don't like it I can split it up again.

I can commit a small test mod if you want.

PR #6059, opened 2019-08-15
Add way to change player pose and size

maxanier:
## Why
For a mod of mine I would like to transform the player into a bat. Therefore I need to modify the size of the player, so the collision boxes match the size of a bat.

## Issue
With MC 1.13/1.14 the system of determining the size of an entity has been changed to use different poses (`net.minecraft.entity.Pose` enum). For modded entities and probably even vanilla mobs this can be worked around. But the PlayerEntity class determines it's size by looking up the current pose in an immutable map. 

## Proposed solution
This PR makes the Pose enum extensible (`IExtensibleEnum`), adds a way to specify the size a player should have in a added pose and introduces an event allowing to set the pose a player is currently in.

There is a (disabled) test that makes the player 2 blocks wide in creative mode.

### Notes
- The entities pose is stored in the datamanager using a enum parameter. However that should not be an issue as the enum is just serialized using it's ordinal.
- It is not possible to override the size associated with vanilla poses (could be changed)
- If an overridden pose isn't clear (there is a collision in the world), the vanilla falls back to the SWIMMING pose, so it's not possible to force a pose even though there is no space (could be changed)
- I have created a new class PlayerSize to store the pose->size map. If there is a better place for this, I will move it.

### Alternatives
A simpler alternative would be to just add a hook and event in the `PlayerEntity#getSize` method and skip/ignore the entire pose concept. However, I think the pose system might be useful for other things as well and might be extended in future MC versions.

PR #6858, opened 2020-06-27
[1.16] Replace EntityHeight event with EntitySize event

maxanier:
1.16 version of #6369
For the reasoning see: #6059

This PR replaces the existing `EntityEvent#Height` with a `EntityEvent#Size` which allows modifying the entity size in addition to the eye height. This is binary breaking.

I personally only need to modify the player size, but as the existing `Height` event is posted for all entities, this system is therefore available for all entities.

If needed, I can commit a small test mod.

<details>
  <summary>Very basic test code</summary>

```
@Mod("player_size_test")
public class PlayerSizeTest
{
    public PlayerSizeTest(){
        MinecraftForge.EVENT_BUS.register(this);
    }


    @SubscribeEvent
    public void onEntitySize(EntityEvent.Size event){
        if(event.getEntity() instanceof PlayerEntity){
            if(((PlayerEntity) event.getEntity()).inventory !=null && ((PlayerEntity) event.getEntity()).getHeldItemMainhand().getItem() == Items.STICK){
                event.setNewSize(EntitySize.fixed(0.2f,0.2f));
                event.setNewEyeHeight(0.15f);
                ((PlayerEntity) event.getEntity()).setForcedPose(Pose.STANDING); //In theory only necessary once
            }
            else{
                ((PlayerEntity) event.getEntity()).setForcedPose(null); //In theory only necessary once
            }
        }
    }

}
```

This simple changes the player's size and pose while holding a stick (have to sneak to update the size here)
If adding a proper test mod, I would write something more suitable.
</details>

### DIFF
--- a/patches/minecraft/net/minecraft/entity/Entity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/Entity.java.patch
@@ -31,18 +31,20 @@
        this.field_200606_g = p_i48580_1_;
        this.field_70170_p = p_i48580_2_;
        this.field_213325_aI = p_i48580_1_.func_220334_j();
-@@ -211,7 +214,9 @@
+@@ -211,7 +214,11 @@
        this.field_70180_af.func_187214_a(field_189655_aD, false);
        this.field_70180_af.func_187214_a(field_213330_X, Pose.STANDING);
        this.func_70088_a();
 -      this.field_213326_aJ = this.func_213316_a(Pose.STANDING, this.field_213325_aI);
-+      this.field_213326_aJ = getEyeHeightForge(Pose.STANDING, this.field_213325_aI);
++      net.minecraftforge.event.entity.EntityEvent.Size sizeEvent = net.minecraftforge.event.ForgeEventFactory.getEntitySizeForge(this, Pose.STANDING, this.field_213325_aI, this.func_213316_a(Pose.STANDING, this.field_213325_aI));
++      this.field_213325_aI = sizeEvent.getNewSize();
++      this.field_213326_aJ = sizeEvent.getNewEyeHeight();
 +      net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.entity.EntityEvent.EntityConstructing(this));
 +      this.gatherCapabilities();
     }
  
     @OnlyIn(Dist.CLIENT)
-@@ -317,7 +322,13 @@
+@@ -317,7 +324,13 @@
     }
  
     public void func_70106_y() {
@@ -56,7 +58,7 @@
     }
  
     public void func_213301_b(Pose p_213301_1_) {
-@@ -330,6 +341,7 @@
+@@ -330,6 +343,7 @@
  
     public boolean func_233562_a_(Entity p_233562_1_, double p_233562_2_) {
        double d0 = p_233562_1_.field_233557_ao_.field_72450_a - this.field_233557_ao_.field_72450_a;
@@ -64,7 +66,7 @@
        double d1 = p_233562_1_.field_233557_ao_.field_72448_b - this.field_233557_ao_.field_72448_b;
        double d2 = p_233562_1_.field_233557_ao_.field_72449_c - this.field_233557_ao_.field_72449_c;
        return d0 * d0 + d1 * d1 + d2 * d2 < p_233562_2_ * p_233562_2_;
-@@ -560,7 +572,7 @@
+@@ -560,7 +574,7 @@
  
              this.field_70140_Q = (float)((double)this.field_70140_Q + (double)MathHelper.func_76133_a(func_213296_b(vector3d)) * 0.6D);
              this.field_82151_R = (float)((double)this.field_82151_R + (double)MathHelper.func_76133_a(d0 * d0 + d1 * d1 + d2 * d2) * 0.6D);
@@ -73,7 +75,7 @@
                 this.field_70150_b = this.func_203009_ad();
                 if (this.func_70090_H()) {
                    Entity entity = this.func_184207_aI() && this.func_184179_bs() != null ? this.func_184179_bs() : this;
-@@ -575,7 +587,7 @@
+@@ -575,7 +589,7 @@
                 } else {
                    this.func_180429_a(blockpos, blockstate);
                 }
@@ -82,7 +84,7 @@
                 this.field_191959_ay = this.func_191954_d(this.field_82151_R);
              }
           }
-@@ -591,8 +603,9 @@
+@@ -591,8 +605,9 @@
  
           float f2 = this.func_225515_ai_();
           this.func_213317_d(this.func_213322_ci().func_216372_d((double)f2, 1.0D, (double)f2));
@@ -94,7 +96,7 @@
           }) && this.field_190534_ay <= 0) {
              this.func_241209_g_(-this.func_190531_bD());
           }
-@@ -611,11 +624,10 @@
+@@ -611,11 +626,10 @@
        int j = MathHelper.func_76128_c(this.field_233557_ao_.field_72448_b - (double)0.2F);
        int k = MathHelper.func_76128_c(this.field_233557_ao_.field_72449_c);
        BlockPos blockpos = new BlockPos(i, j, k);
@@ -108,7 +110,7 @@
              return blockpos1;
           }
        }
-@@ -801,6 +813,7 @@
+@@ -801,6 +815,7 @@
     public void func_174829_m() {
        AxisAlignedBB axisalignedbb = this.func_174813_aQ();
        this.func_226288_n_((axisalignedbb.field_72340_a + axisalignedbb.field_72336_d) / 2.0D, axisalignedbb.field_72338_b, (axisalignedbb.field_72339_c + axisalignedbb.field_72334_f) / 2.0D);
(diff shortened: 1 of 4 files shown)


patches/minecraft/net/minecraft/entity/Entity.java.patch
```diff
@@ -31,18 +31,20 @@
...
        this.field_70170_p = p_i48580_2_;
        this.field_213325_aI = p_i48580_1_.func_220334_j();
-@@ -213,7 +216,9 @@
+@@ -213,7 +216,11 @@
        this.field_70180_af.func_187214_a(field_189655_aD, false);
        this.field_70180_af.func_187214_a(field_213330_X, Pose.STANDING);
        this.func_70088_a();
 -      this.field_213326_aJ = this.func_213316_a(Pose.STANDING, this.field_213325_aI);
-+      this.field_213326_aJ = getEyeHeightForge(Pose.STANDING, this.field_213325_aI);
++      net.minecraftforge.event.entity.EntityEvent.Size sizeEvent= getSizeForge(Pose.STANDING, this.field_213325_aI);
```
maxanier: I am not sure about this part.
The old eyeHeight method was placed in the constructor as well, so I kept it here.
However, I do not think this is ideal. If mods decide to change an entity's size/eyeHeight, they need something to base their decision upon. But at this point during entity construction neither capabilities nor held items nor anything else is available.
This means a) the event handler has to check if the entity is properly constructed and b) the mod has to trigger a size update anyway.
Only scenario this might be useful might be a mod which e.g. simply scales all cows to 200%. But that could also be done without the constructor patch, by just calling recalculateSize in EntityJoinWorld.

Therefore, I would suggest removing this ugly patch, making it much easier to use.

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
```diff
@@ -43,23 +44,31 @@
...
 +      net.minecraftforge.fml.hooks.BasicEventHooks.onPlayerPostTick(this);
     }
  
     public boolean func_226563_dT_() {
-@@ -550,6 +555,7 @@
+@@ -333,6 +339,7 @@
+    }
+ 
+    protected void func_213832_dB() {
++      if(forcedPose!=null){this.func_213301_b(forcedPose);return;}
```
ichttt: I think you can add a few newlines to make this patch more readable

ichttt (commented):
A small nitpick, otherwise looks good to me.
Regarding you question about firing in the contructor, I'm not too sure, someone else will have to decide this

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
TheCurle: Yeah. You should aim for conciseness in patches, but not to the degree of unreadability.

patches/minecraft/net/minecraft/entity/Entity.java.patch
LexManos: The issue is that we have to have SOME default value in the constructor or somewhere that is always called before any other methods in the entity is called because eyeHeight needs to be a sane value until the next time resize is called.

patches/minecraft/net/minecraft/entity/Entity.java.patch
```diff
@@ -310,9 +316,9 @@
 +      this.reviveCaps();
 +   }
 +
-+   private float getEyeHeightForge(Pose pose, EntitySize size) {
-+      net.minecraftforge.event.entity.EntityEvent.EyeHeight evt = new net.minecraftforge.event.entity.EntityEvent.EyeHeight(this, pose, size, this.func_213316_a(pose, size));
++   private net.minecraftforge.event.entity.EntityEvent.Size getSizeForge(Pose pose, EntitySize size) {
```
LexManos: honestly all of this should move to ForgeHooks/ForgeEvenntFadctory like all the other events instead of being special.

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
```diff
@@ -336,5 +348,9 @@
 +         else if (facing.func_176740_k().func_176722_c()) return playerEquipmentHandler.cast();
 +      }
 +      return super.getCapability(capability, facing);
++   }
++
++   public void setForcedPose(@Nullable Pose pose) {
```
LexManos, line 376: Why is this being forced at all? The data manager already syncs the pose, why add some backdoor to bypass it?

LexManos (changes requested):

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
maxanier: Unfortunately, the player entity (re)determines the players pose every tick and thereby overrides any manually set pose.
To determine the current pose the PlayerEntity uses hardcoded sizes and therefore e.g. falls back to the swimming pose when the player is in a 1x1 space (where they fit because of actual entity size). 
It is possible to override the value again right after that in the Post PlayerTickEvent, so the pose determined by vanilla does not affect rendering or game mechanics. However, every time the pose is changed `notifyDataManagerChange` calls `recalculateSize` causing an event to be posted.
Therefore, without this "backdoor" the event would be posted twice every tick instead of not at all.

If you still prefer this, I can remove this patch.

patches/minecraft/net/minecraft/entity/Entity.java.patch
maxanier: I would simply use the vanilla methods for size and eyeHeight. Thereby they are at sane values, and if desired a mod can still force a size recalculation once the entity is fully constructed.

patches/minecraft/net/minecraft/entity/Entity.java.patch
maxanier: moved

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -722,4 +724,11 @@ public static long onSleepFinished(ServerWorld world, long newTime, long minTime
         MinecraftForge.EVENT_BUS.post(event);
         return event.getNewTime();
     }
+
+    public static net.minecraftforge.event.entity.EntityEvent.Size getEntitySizeForge(Entity player, Pose pose, EntitySize size, float eyeHeight)
+    {
+        net.minecraftforge.event.entity.EntityEvent.Size evt = new net.minecraftforge.event.entity.EntityEvent.Size(player, pose, size, eyeHeight);
```
ichttt: You don't need to use fully qualified names here, just use `EntityEvent.Size` instead of `net.minecraftforge.event.entity.EntityEvent.Size`. Same in the line below

LexManos:
We're still waiting on @iChun to give his feedback because the eye height was done on request of Morph.
I'm still not a fan of the active event. But it doesn't seem to have a way to have it as a property of the pose or whatever you're using to determine the form.

iChun:
I've had a look see through the changes, and personally I'm in agreement of extending the EyeHeight event into a proper Size event since the addition of Pose to MC. 

I've also gone through #6059 briefly and I understand the challenges in extending Pose with IExtensibleEnum considering issues syncing with the clients as well as breaking vanilla, so I reckon this might be the next best option in regards to changing an entity/player's size. 

The only thing I'm not too sure about currently is the check for forcedPose in updatePose(). Although I can see why it's done earlier on to bypass vanilla checks, it also bypasses vanilla's fallback of using the smallest pose possible for the player. This opens things up to errors done by a mod dev eg forgetting to check for clearance like vanilla does, or forgetting to clear the forced pose. 

Some documentation reminding devs not to neglect that should be added, as well as a getter for forcedPose added for other devs to check that it is not already being forced by other mods. Mod conflicts with eyeHeight and entity size was always an issue in my experience with Morph, and it would be a nice note to add as well that mods should only change those when they should, instead of eg every tick. I understand wanting to keep patch sizes small, but as this is a new method added by Forge and with potential mod conflicts, it might be worthwhile adding at least a bit of documentation to the method.

maxanier:
Alright, @iChun thank you for your input.
I will add the requested documentation as well as the getter.
I agree that having some clearance check for the forcedPose would be good, but it would require posting another event.

Was there any specific reason for the "getEyeHeight" in the constructor? Otherwise I will remove it, as it mainly causes issues.
https://github.com/MinecraftForge/MinecraftForge/pull/6858#discussion_r446529339

iChun:
Check for forcedPose can be done when listening to the player tick. Another event is not necessary. It's just worth adding docs so that mod devs don't forget about it.

As to having getEyeHeight in the constructor, it's been a long while and eyeHeight has changed a lot since it was first added for Morph, but I think it's done in the constructor because not all entities will call recalculateSize in their lifetime, and it allows changing the eyeHeight before the entity is ticked. Having the default eye height, and then changing it a tick later, potentially causes issues to other mods (eg projectile firing mods) or rendering (especially if you're changing the player eye height on respawn).

One can simply check isAddedToWorld() or firstUpdate to see if it's been added yet/ticked yet. We pass in the default eyeHeight/size for reference to the dev. It also allows for the dev to see if the height has already been changed by another mod in the same event.

maxanier:
Ok, then I will keep it and add an appropriate comment to the event

maxanier:
Rebased and applied suggestions

maxanier:
@ichttt Anything Lex wants changed?

ichttt:
Please update to 1.16.2, otherwise this looks good to me and lex

maxanier:
Done

LexManos:
I am wholey not a fan of this hack, but the original was a hack as well. And ichun says its fine. So, here we go :/

PR #6860, opened 2020-06-27
Fixed handleServerAboutToStart being called too late on integrated servers

### LINKED ISSUE #6859: FMLServerAboutToStartEvent fires too late on integrated servers
**Minecraft Version:** 1.16.1

**Forge Version:** 32.0.23

**Description of issue:**
I used the `FMLServerAboutToStartEvent` to add a reload listener for some custom tags between vanilla's tag reload listener, and vanilla's recipe reload listener. In 1.16, `FMLServerAboutToStartEvent` is fired after recipes already start loading on an integrated server, but on a dedicated server this event is fired at the expected time and I am successfully able to add my listener and have it get fired before recipes load. I am not sure but I would not be fully surprised if the reload listeners don't do their initial firing at all on an integrated server due to the `FMLServerAboutToStartEvent` being fired after after the reload listeners start running.

pupnewfster:
Reopening this now that I have done more testing and realized I was entirely wrong about what was causing me to crash when I opened #6859 initially, and I now understand what this PR is fixing. `ServerLifecycleHooks#handleServerAboutToStart` sets the current server, on the dedicated server this is called ***before*** the world starts to be loaded (`func_240800_l__`). On an integrated server however, this is called ***after*** the server starts to load so `ServerLifecycleHooks#getCurrentServer` returns null when queried from places like `WorldEvent.Load` (which if you want to use the "global" save data for some information by grabbing the overworld, makes this useless in singleplayer as you aren't able to access it). This PR moves the call on the Integrated server to handling server about to start to before the world starts loading on the server.

### DIFF
--- a/patches/minecraft/net/minecraft/server/integrated/IntegratedServer.java.patch
+++ b/patches/minecraft/net/minecraft/server/integrated/IntegratedServer.java.patch
@@ -1,10 +1,11 @@
 --- a/net/minecraft/server/integrated/IntegratedServer.java
 +++ b/net/minecraft/server/integrated/IntegratedServer.java
-@@ -60,8 +60,9 @@
+@@ -59,9 +59,10 @@
+       this.func_71245_h(true);
        field_147148_h.info("Generating keypair");
        this.func_71253_a(CryptManager.func_75891_b());
-       this.func_240800_l__();
 +      if (!net.minecraftforge.fml.server.ServerLifecycleHooks.handleServerAboutToStart(this)) return false;
+       this.func_240800_l__();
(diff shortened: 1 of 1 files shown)


pupnewfster:
Changed this back to a draft given upon closer inspection I am not sure if this actually addresses the issue fully and I need to do some more testing.

pupnewfster:
For some reason this change does prevent a crash in my test from occurring but I am going ahead and closing this PR as I don't think it actually is an issue and I just had not realized how reload listeners have changed in 1.16 and I believe #6849 will work well enough for my needs.

cpw:
Ok I'll take a look this evening..

cpw:
b00c91af3c83a2946b2726147c1445b6c5ad75fe

Issue #6857, opened 2020-06-27
Unable to use existing tags in Data Generators


pupnewfster:
**Minecraft Version:** 1.16.1

**Forge Version:** 32.0.23

**Logs:** https://gist.githubusercontent.com/pupnewfster/b81e396d26730c5a3351de46e441d502/raw/dda5ebfca9467f2c25601d19b60fccb89060dc72/gistfile1.txt

**Steps to Reproduce:**
 1. Download the test mod https://github.com/pupnewfster/DataGenerator-Failure and try to run the data generator

**Description of issue:**
Due to the changes in the tag system it seems that it is no longer possible to use tags in recipe data generators if you are not adding to the tag yourself. and instead get the error message: `Tag x used before it was bound` for example in the test mod trying to use forge's cobblestone tag gives the error `Tag forge:cobblestone used before it was bound`.

TBoshoven:
I've been looking into this problem, but kind of gave up for now.
What I did find is that Forge itself has this issue as well. Running Forge datagen fails on a recipe with this exact error.

Reproduction steps:
* Set of Forge development
* Run the `forge_data` task in Gradle

```
java.lang.IllegalStateException: Tag minecraft:acacia_logs used before it was bound
at net.minecraft.tags.TagRegistry$NamedTag.func_232944_c_(TagRegistry.java:77)
```

covers1624:
God damnit, I ran into this _exact_ issue when I was fixing patches (only it was on launch not datagen). We need to build the TagCollections and do something similar to `NetworkTagManager.func_232980_f_` and call `BlockTags.setCollection` and so on. Tags now get a callback when the collection is updated to refresh their internal cache, and since the collection is never updated, you hit that exception.

Cyborgmas:
Yeah what i found is that in forge, there are calls to retrieve the Tag, using `NamedTag#func_232944_c_`  but vanilla never uses that

pupnewfster:
Resolved by https://github.com/MinecraftForge/MinecraftForge/commit/d5aa0c4a9121558f9c19db8c3a21b8b12217086a

PR #6865, opened 2020-06-27
1.16 Fix datagen resolving tags and exploding.

Cyborgmas:
Datagen is currently not working because of an issue with tags (#6857)

From what I understand, the problem is because forge uses actual Tag references in datagen, where vanilla does not. Therefore, vanilla does not generate/update tags.
The way vanilla does it (when running server/client) requires the tags to be there as files ofc and I mimic that somewhat. That is why all the tags have been added. this could be completely wrong though... would require another way to load/manage tags. 

There also was a crash due to now improper reflection in Ingredient, so fixed that.
I also added the clean_data task, was useful to debug stuff, and fixed clean_server as well. 

This is not complete at all. Datagen is running without crashing but it seems to be generating stuff incorrectly.
If anyone can submit tests, I would appreciate it.


pupnewfster:
This PR is adding a lot of files in `src/main/resources/data` but I believe forge's data generator is supposed to target `src/generated/resources/data`

Cyborgmas:
yeah, but those are not the result of datagen. that are tags needed to make it work. will try to find a workaround because its prob unecessary

Cyborgmas:
Okay, so after a bit more debugging found out that the problem lies in the creation of an Ingredient, forge has an isSimple field that gets every item of the ingredient. when its a taglist ingredient it tries to get the items the tag contains and fails since it was not updated.

Cyborgmas:
Okay, so the isSimple field is not used during datagen (only used if a recipe matches) so i force it to false when running datagen. 
the furnace recipe got deleted because vanilla now has a tag for it, called furnace_material 

The shears tag got deleted for some reason, will check why.

pupnewfster:
My guess is the code behind the shears tag might not have been properly ported due to it getting merged into the 1.15 branch in the final pass before 1.16 was released.

Cyborgmas:
okay just re-added.

williewillus:
Would be great to have this merged, our datagens can't run at the moment as a result of this.

PR #6867, opened 2020-06-27
Expand the range of depth values allowed when the game is drawing GUIs.

gigaherz:
The default value is 3000, which is very limiting for a lot of mods.
This expands the far plane to 30000, which still leaves enough precision to not disturb drawing higher detail objects.

### DIFF
--- a/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
+++ b/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
@@ -153,6 +153,10 @@ public class ForgeHooksClient
 {
     private static final Logger LOGGER = LogManager.getLogger();
 
+    // De-inlined to ensure
+    public static final double GUI_DEPTH_NEAR = 1000.0;
+    public static final double GUI_DEPTH_FAR = 30000.0; // Forge: expand from 3000 because quite a lot of mods have issues due to the limited range.
+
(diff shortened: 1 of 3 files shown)


src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -153,6 +153,10 @@
 {
     private static final Logger LOGGER = LogManager.getLogger();
 
+    // De-inlined to ensure
```
KnightMiner: Ensure what?

gigaherz: Oops. "To ensure the two patches don't get out of sync" :P

gigaherz:
As a note, we won't be merging this straight away.
Lex expressed his concern of using an arbitrary number like 30000 as opposite to something smarter.
We should discus the advantages and drawbacks of setting up some kind of layering system so that if a GUI wants to show a popup, it can use pushScreen(x)/popScreen(), instead of drawing them by hand.
This would allocate some extra space for each popup level, progressively increasing the number. Or alternatively clear the depth between layers.

TheCurle:
Should we move this to an issue for further discussion of the proposed system?

marchermans:
I think so.
I have a GUI library and I am running into this as well. Having a stacking system would be nice.

PR #6868, opened 2020-06-28
Edited canSustainPlant to match vanilla isValidGround methods

### LINKED ISSUE #6781: Forge re-introduces bug converting coarse dirt and podzol to dirt after sapling is grown, breaking tree farms
A user found a bug with Forge but was unable to post it here due to Forge's anti-spam measures. So I am reposting their issue report they had posted on Botania back to here where it properly belongs. 

Original issue report: https://github.com/Vazkii/Botania/issues/3329

Content of issue report:
--------
# Version Information
Forge version: forge-31.2.20
Botania version: Botania-r1.15-385.jar

# Further Information
https://bugs.mojang.com/browse/MC-57884 fixed back in 2014

Additionally according to this
Mycelium should NOT allow saplings to be planted on it (verified for vanilla 1.15.2) but once Forge is installed (tested up to latest forge-31.2.20)  mycelium allows the planting of saplings again. (also bad.)

Once the sapling grows, the Coarse Dirt, and Podzol (and apparently erroneously, Mycelium) get converted back to plain dirt again.

this breaks several modded operations/options not the least of which is Botania Tree Farms for which you use the Rannuncarpus to select the coarse dirt to plant the saplings on to differentiate from regular dirt or grass nearby.

screenshot: https://imgur.com/0NwIANg

I have made multiple attempts at filing this report with Forge this morning after confirming the bug, but cannot get past their blasted spam registration protection to create an account with and either search the forums properly to see if the bug is even reported yet, or submit my own.

Confirmed this bug exists SOLELY when 1.15.2 + forge-31.2.20 and no other mods installed. so NOT a botania issue, but directly affects botania, so you were my second choice for getting this reported upstream since I cannot register at minecraftforge.net. (grr. apologies.)

DaemonUmbra:
Should Fix #6781, but I'm not confident this is the correct way to do it. Posting as a draft to get feedback.

### DIFF
--- a/patches/minecraft/net/minecraft/block/Block.java.patch
+++ b/patches/minecraft/net/minecraft/block/Block.java.patch
@@ -81,7 +81,7 @@
     }
  
     @OnlyIn(Dist.CLIENT)
-@@ -419,6 +424,83 @@
+@@ -419,6 +424,41 @@
        return this;
     }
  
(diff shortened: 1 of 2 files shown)


TheCurle (approved):
I was working on this separately, and this is almost the exact same result i came to.
This works, in my testing. Fixes the sapling-on-mycelium bug that I set out to fix.

Maybe we should rewrite this logic to move away from Mojang's code though.

Cyborgmas:
Will that not break any custom plant/crop with a plantype? also there seems to be a check for water for cactus/sugar cane but it does not appear in the new version

ichttt:
Yeah I think this should still use `PlantType`, you may have to add/change the plant types to account for the new vanilla logic, though

PR #6876, opened 2020-06-29
1.16 - Fix TagsUpdateEvent firing too early.

Cyborgmas:
On servers, players can not enter any fuel into furnaces. However, it works after doing the /reload command.
This is because when the TagsUpdatedEvent is fired, forge sets up the burntime of fuels and uses that to determine if an item is a valid fuel. However, the TagsUpdateEvent currently fires before modloading end, while the EventBus is still shutdown and the event is ignored. 
the reload command fires the event, the burntimes are updated and everything works again.

This PR simply strips out the firing of the event from the NetworkTagManager's function and adds it to each relevant places as well as right after modloading ended. 
This is still a bit weird, because the event is firing after tags updated because it simply cant be fired at that time. 

The firing of the event can definitely be moved somewhere else in DedicatedServer, it really just needs to be fired at some point is all.

### DIFF
--- a/patches/minecraft/net/minecraft/client/Minecraft.java.patch
+++ b/patches/minecraft/net/minecraft/client/Minecraft.java.patch
@@ -278,7 +278,15 @@
     }
  
     private void func_184117_aA() {
-@@ -1676,6 +1715,10 @@
+@@ -1643,6 +1682,7 @@
+          try {
+             saveformat$levelsave.func_237287_a_(p_238195_2_, iserverconfiguration);
+             minecraft$packmanager.func_238225_b_().func_240971_i_();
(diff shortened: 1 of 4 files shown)


NR472:
Hi there, sorry I'm a programming noob, how do you actually do the reload command?

TheCurle:
In the game, type /reload into chat.

NR472:
I tried that and it said unknown or incomplete command :(

MelanX:
You need to enable cheats in Singleplayer or be op in Multiplayer to execute a command like this.

NR472:
Fixed it! Thank you so so much!

PR #6877, opened 2020-06-29
[1.16] Allow any armor to have custom knockback resistance

TheCurle:
This replaces a hardcoded check for Netherite armor, with a check for `IArmorMaterial.func_230304_f_ > 0`. 
This function is only used here, meaning it most likely means `getKnockbackResistance`.


KingLemming:
This can already be done actually via existing events (Equipment Change or Living Update), but this would certainly be an improvement. :)

PR #6882, opened 2020-06-30
[1.15] Add extra debugging information to the MDK

TheCurle:
This helps us diagnose issues with the JDK, without asking for command outputs and various other things to get information relevant to debugging.

### DIFF
--- a/mdk/build.gradle
+++ b/mdk/build.gradle
@@ -144,3 +144,5 @@ publishing {
         }
     }
 }
+
+System.out.println("Running with java ${System.getProperty("java.version")} (${System.getProperty("os.arch")})")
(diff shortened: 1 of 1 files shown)


mdk/build.gradle
```diff
@@ -143,4 +143,6 @@ publishing {
             url "file:///${project.projectDir}/mcmodsrepo"
         }
     }
-}
\ No newline at end of file
+}
+
+System.out.println("Running with java ${System.getProperty("java.version")} (${System.getProperty("os.arch")})")
```
marchermans: You should probably use a lifecycle logger here instead of the usual System.out.println

Something like project.getLogger().lifecycle(..) will do i think.

marchermans (changes requested):
See the line comment.
In my eyes we should avoid using the System.out inside gradle for logging.
Gradle has a perfectly fine logging system that should be used instead. It then dumps both to a file (if connected) and to the console at once.

TheCurle:
This was a feature explicitly requested by Lex and giga. See the mentioned PR, the 1.16 change has been made already.

I can quickly fix this one up though, if there's a more robust solution.

marchermans:
If that is already merged and it was requested then I see no objections.

mdk/build.gradle
marchermans: Disregard the previous review, has been cleared up as, requested and already merged in to 1.16.

marchermans (approved):

PR #6883, opened 2020-06-30
[1.16] Add extra debugging information to the MDK

TheCurle:
Same as #6882. Extra information useful for debugging Java issues.


TheCurle:
This is redundant due to  #ab29145

PR #6884, opened 2020-06-30
[1.16] Include model data in getQuads call

SuperMartijn642:
The model data wasn't included when getting quads from specific sides, but was when getting quads for side = null.


marchermans (approved):
This looks like a correct patch.
I am reading myself into the Rendering/Model system and this change is correct as far as I can see.

Can anyone on the @Rendering-Team verify that for me?

gigaherz:
Merged as 46c9dec4217

PR #6885, opened 2020-06-30
Pass in matrix stack in tooltip events

### LINKED ISSUE #6879: 1.16: Tooltip render events don't include a reference to the current MatrixStack

Cyborgmas:
Fix #6879

### DIFF
--- a/src/main/java/net/minecraftforge/client/event/RenderTooltipEvent.java
+++ b/src/main/java/net/minecraftforge/client/event/RenderTooltipEvent.java
@@ -24,6 +24,7 @@ import java.util.List;
 
 import javax.annotation.Nonnull;
 
+import com.mojang.blaze3d.matrix.MatrixStack;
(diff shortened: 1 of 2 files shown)


src/main/java/net/minecraftforge/client/event/RenderTooltipEvent.java
```diff
@@ -48,14 +49,16 @@
     @Nonnull
     protected final ItemStack stack;
     protected final List<? extends ITextProperties> lines;
+    protected final MatrixStack matrixStack;
```
pupnewfster, line 52: Correct me if I am wrong, but I don't believe you added a getter to actually *get* the matrix stack?

marchermans (approved):
Changes look good.
Makes sense to have the MatrixStack in this position as well.

gigaherz:
Merged as 45152c6073

PR #6505, opened 2020-02-08
Fix ITeleporter being ignored when teleporting from the end to the overworld

pupnewfster:
It seems the reimplementation of the `ITeleporter` interface in #6404 missed the case in `ServerPlayerEnttiy` where vanilla has specific code in place for teleporting from the end to the overworld. This PR adds a method similar to the one the 1.12 version of `ITeleporter` to determine if the vanilla code should be ran. Rather than calling it `isVanilla()` like it was in 1.12, I have named it `specialEndReturn()` (if anyone has a better name suggestion feel free to mention it). The reason I named it this is due to the new implementation of `ITeleporter` to allow more control/handling over how the teleportation actually happens it is only needed in this one place rather than having checks all over the place.

Additionally this PR fixes forge's `setdimension` command ignoring the position information.

Tested:
- Transitioning from the end to the overworld using vanilla's end portal
- Tested using forge's `setdimension` command going to the end from the overworld and back.

PR #6404, opened 2020-01-02
[1.15] Reimplemented the ITeleporter interface

mcenderdragon:
This is a port of https://github.com/MinecraftForge/MinecraftForge/pull/6017 with Lex requested changes. 

I have tested via the `/forge setdimension` command and players & entities worked.

### Why do we need this:
Currently each time you switch the dimension to a custom made one you will automatically spawn a nether portal. This PR gives you full control over the dimension change process.

Resolves #5990

PR #6886, opened 2020-06-30
Reimplement ITeleporter Patches

pupnewfster:
Ports the patches relating to ITeleporter to 1.16. (See #6404 and #6505 for more details). I tested via the forge `setdimension` command and also with a modified version of Mekanism to verify that the system is properly working.


ichttt (commented):
I think the exc file needs an update to account for new parameters. Otherwise this seem good

PR #6887, opened 2020-06-30
[1.15] Added AnvilDamageEvent and AnvilDamageEvent.Falling for cancelling Anvil damage

Lanse505:
This PR makes it so that you can cancel the anvil taking damage (going from Anvil -> Chipped -> Damaged).
The normal event targets taking damage in general while the Falling sub-event targets when the anvil is damaged through the FallingBlockEntity code.

Included is a test mod that shows off this feature.


Lanse505:
Closing this PR while I re-evaluate somethings relating to the PR.

PR #6158, opened 2019-09-15
Add hook for user connections at the stage of ban/whitelist:

soundlogic2236:
The hook consists of an event fired during the stage of ban/whitelist which can control if the connecting user is rejected and the message they are rejected with if they are, enabling custom access control mods to work properly.
  - Add a new event ServerTryConnectEvent to fire at this stage
  - Patch NetHandlerLoginServer to call the event and modify its behavior according to the result
  - Add canUserConnect to the ForgeEventFactory
  - Add tests for the hook as ServerModifyConnectTest
This resolves #4300.

PR #6889, opened 2020-06-30
Add PlayerConnectingEvent 

### LINKED ISSUE #4300: Event Request: Handle Login
I've looked through quite a few classes and can't find a way to hook in simply for a whitelist mod.

What I'm requesting is a post-auth and pre-login event that provides either the profile or network handler with the ability to set the result to deny or cancel the event.

phit:
fires right after whitelist and ban checks, setting a rejection message
or cancelling it will disconnect the player

based on GH-6158
resolves GH-4300

---

I tried to address the feedback in #6158 but not too familiar with the Forge event system in general and how cancelable is usually handled.
Also the event name is quite broad, any better name ideas?

### DIFF
--- a/patches/minecraft/net/minecraft/network/login/ServerLoginNetHandler.java.patch
+++ b/patches/minecraft/net/minecraft/network/login/ServerLoginNetHandler.java.patch
@@ -15,17 +15,15 @@
           this.func_147326_c();
        } else if (this.field_147328_g == ServerLoginNetHandler.State.DELAY_ACCEPT) {
           ServerPlayerEntity serverplayerentity = this.field_147327_f.func_184103_al().func_177451_a(this.field_147337_i.getId());
-@@ -91,6 +97,9 @@
+@@ -91,6 +97,7 @@
        }
  
        ITextComponent itextcomponent = this.field_147327_f.func_184103_al().func_206258_a(this.field_147333_a.func_74430_c(), this.field_147337_i);
-+
 +      itextcomponent = net.minecraftforge.event.ForgeEventFactory.onPlayerConnect(this.field_147327_f, this.field_147333_a, this.field_147337_i, itextcomponent);
-+
        if (itextcomponent != null) {
           this.func_194026_b(itextcomponent);
        } else {
-@@ -128,7 +137,7 @@
+@@ -128,7 +135,7 @@
           this.field_147328_g = ServerLoginNetHandler.State.KEY;
           this.field_147333_a.func_179290_a(new SEncryptionRequestPacket("", this.field_147327_f.func_71250_E().getPublic(), this.field_147330_e));
        } else {
@@ -34,7 +32,7 @@
        }
  
     }
-@@ -142,7 +151,7 @@
+@@ -142,7 +149,7 @@
           this.field_147335_k = p_147315_1_.func_149300_a(privatekey);
           this.field_147328_g = ServerLoginNetHandler.State.AUTHENTICATING;
           this.field_147333_a.func_150727_a(this.field_147335_k);
@@ -43,7 +41,7 @@
              public void run() {
                 GameProfile gameprofile = ServerLoginNetHandler.this.field_147337_i;
  
-@@ -151,11 +160,11 @@
+@@ -151,11 +158,11 @@
                    ServerLoginNetHandler.this.field_147337_i = ServerLoginNetHandler.this.field_147327_f.func_147130_as().hasJoinedServer(new GameProfile((UUID)null, gameprofile.getName()), s, this.func_191235_a());
                    if (ServerLoginNetHandler.this.field_147337_i != null) {
                       ServerLoginNetHandler.field_147332_c.info("UUID of player {} is {}", ServerLoginNetHandler.this.field_147337_i.getName(), ServerLoginNetHandler.this.field_147337_i.getId());
@@ -57,7 +55,7 @@
                    } else {
                       ServerLoginNetHandler.this.func_194026_b(new TranslationTextComponent("multiplayer.disconnect.unverified_username"));
                       ServerLoginNetHandler.field_147332_c.error("Username '{}' tried to join with an invalid session", (Object)gameprofile.getName());
-@@ -164,7 +173,7 @@
+@@ -164,7 +171,7 @@
                    if (ServerLoginNetHandler.this.field_147327_f.func_71264_H()) {
                       ServerLoginNetHandler.field_147332_c.warn("Authentication servers are down but will let them in anyway!");
                       ServerLoginNetHandler.this.field_147337_i = ServerLoginNetHandler.this.func_152506_a(gameprofile);
@@ -66,7 +64,7 @@
                    } else {
                       ServerLoginNetHandler.this.func_194026_b(new TranslationTextComponent("multiplayer.disconnect.authservers_down"));
                       ServerLoginNetHandler.field_147332_c.error("Couldn't verify username because servers are unavailable");
-@@ -185,6 +194,7 @@
+@@ -185,6 +192,7 @@
     }
  
     public void func_209526_a(CCustomPayloadLoginPacket p_209526_1_) {
--- a/patches/minecraft/net/minecraft/network/login/ServerLoginNetHandler.java.patch
+++ b/patches/minecraft/net/minecraft/network/login/ServerLoginNetHandler.java.patch
@@ -15,7 +15,17 @@
           this.func_147326_c();
        } else if (this.field_147328_g == ServerLoginNetHandler.State.DELAY_ACCEPT) {
           ServerPlayerEntity serverplayerentity = this.field_147327_f.func_184103_al().func_177451_a(this.field_147337_i.getId());
-@@ -128,7 +134,7 @@
+@@ -91,6 +97,9 @@
+       }
+ 
+       ITextComponent itextcomponent = this.field_147327_f.func_184103_al().func_206258_a(this.field_147333_a.func_74430_c(), this.field_147337_i);
++
++      itextcomponent = net.minecraftforge.event.ForgeEventFactory.onPlayerConnect(this.field_147327_f, this.field_147333_a, this.field_147337_i, itextcomponent);
++
+       if (itextcomponent != null) {
+          this.func_194026_b(itextcomponent);
+       } else {
+@@ -128,7 +137,7 @@
           this.field_147328_g = ServerLoginNetHandler.State.KEY;
           this.field_147333_a.func_179290_a(new SEncryptionRequestPacket("", this.field_147327_f.func_71250_E().getPublic(), this.field_147330_e));
        } else {
@@ -24,7 +34,7 @@
        }
  
     }
-@@ -142,7 +148,7 @@
+@@ -142,7 +151,7 @@
           this.field_147335_k = p_147315_1_.func_149300_a(privatekey);
           this.field_147328_g = ServerLoginNetHandler.State.AUTHENTICATING;
           this.field_147333_a.func_150727_a(this.field_147335_k);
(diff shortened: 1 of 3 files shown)


TheCurle:
How about PlayerConnectedEvent? ServerConnectEvent insinuates the server itself is connecting to something, like another server.

phit:
"connected" sounds like it's already done, but since the whole point of this event is the ability to stop the connecting early I named it "PlayerConnectingEvent" for now

src/main/java/net/minecraftforge/event/PlayerConnectingEvent.java
```diff
@@ -0,0 +1,103 @@
...
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.util.text.ITextComponent;
+import net.minecraft.util.text.TranslationTextComponent;
+import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.eventbus.api.Cancelable;
+
+import javax.annotation.Nullable;
+
+/**
+ * ServerConnectEvent is fired when a player is connecting to a server right after the ban and whitelist checks. <br>
```
phit: rename here before merge

patches/minecraft/net/minecraft/network/login/ServerLoginNetHandler.java.patch
```diff
@@ -15,7 +15,17 @@
...
        } else if (this.field_147328_g == ServerLoginNetHandler.State.DELAY_ACCEPT) {
           ServerPlayerEntity serverplayerentity = this.field_147327_f.func_184103_al().func_177451_a(this.field_147337_i.getId());
-@@ -128,7 +134,7 @@
+@@ -91,6 +97,9 @@
+       }
+ 
+       ITextComponent itextcomponent = this.field_147327_f.func_184103_al().func_206258_a(this.field_147333_a.func_74430_c(), this.field_147337_i);
++
++      itextcomponent = net.minecraftforge.event.ForgeEventFactory.onPlayerConnect(this.field_147327_f, this.field_147333_a, this.field_147337_i, itextcomponent);
++
```
diesieben07: Unnecessary whitespace changes.

src/main/java/net/minecraftforge/event/PlayerConnectingEvent.java
```diff
@@ -0,0 +1,103 @@
...
+    public void setRejectionMessage(@Nullable ITextComponent rejectionMessage)
+    {
+        if (rejectionMessage == null) {
+            this.setCanceled(false);
+            this.rejectionMessage = null;
+        } else {
+            this.setCanceled(true);
+            this.rejectionMessage = rejectionMessage;
+        }
+    }
```
diesieben07, line 98: One thing that was [noted in a previous PR](https://github.com/MinecraftForge/MinecraftForge/pull/6158#issuecomment-531923415) is that people should probably not be able to turn an already rejected login into an accepted one.
I agree with that. I might even go so far as to say to not even fire the event if vanilla rejects the player already: Do we really want mods to be able to bypass the vanilla bans and whitelist? Probably not.

diesieben07 (changes requested):
The name `PlayerConnectingEvent` seems fine to me.

src/main/java/net/minecraftforge/event/PlayerConnectingEvent.java
```diff
@@ -0,0 +1,103 @@
...
+ * If the event is canceled, the player will be disconnected with either the set rejection message or a generic disconnected message.<br>
+ * <br>
+ * This event is {@link net.minecraftforge.eventbus.api.Cancelable}.<br>
+ * <br>
+ * This event does not have a result. {@link HasResult}<br>
+ * <br>
+ * This event is fired on the {@link MinecraftForge#EVENT_BUS}.
+ **/
+@Cancelable
+public class PlayerConnectingEvent extends net.minecraftforge.eventbus.api.Event {
```
TheSilkMiner: Why's this a Fully Qualified Name?

phit: copy pasta :D https://github.com/MinecraftForge/MinecraftForge/blob/1.15.x/src/main/java/net/minecraftforge/event/ServerChatEvent.java#L46

src/main/java/net/minecraftforge/event/PlayerConnectingEvent.java
phit: Being able to modify the vanilla whitelist and bans is the main point why I made this PR, say you have a mod that adds additional whitelist.jsons from various sources, if the player isn't in the vanilla whitelist the mod wouldn't be able to do that.

diesieben07: The vanilla whitelist should just be disabled in that case.

dries007: Idealy the event would let you know if the vanilla action was to block and let you overwrite that.

> The vanilla whitelist should just be disabled in that case.

I get many, many support requests for my mod (MCLink) from people who enable the whitelist and expect my mod to then overwrite that setting. I could overwrite that option on every startup, overwrite the command, and do a whole bunch of other, non-user friendly things. But I'd prefer not to do that and just have the event always fire and let the modder handle it.

diesieben07: I don't like the idea of mods being able to bypass vanilla whitelist decisions.
If as a user I enable the whitelist I expect it to be enforced. If I don't want it enabled, I don't enable it.

dries007: You are not an average user though. An average user expects to install a "whitelist mod" on their server and everything to Just Work™. Explaining to them that they have to install a "whitelisting mod" and then _disable_ the whitelist results in confusion.

Then you also get the added bonus of the following scenario (has happend, I've received complaints about it):
- User has a whitelisted server
- User installs whitelist mod to do some automatic whitelist managent
- User has to disalbe the whitelist because of reasons
- Everything works for a wile
- User decides to remove the mod for any reason
- Server is now no longer protected by any whitelist because they had to disable it previously

I agree that this has potential to be abused. But so do many, many other things. And when they do get abused, people stop using the mod or they complain.
I'd say make it a "paranoid whitelist" config option, but honestly, that's gonna be even more confusing.

DeliciousJaffa: Having a mod-specific "lockout whitelist" is confusing to users by itself.
I'd expect to be able to enable the vanilla whitelist when maintenance on a server is required and not have to search through (potentially multiple) mod config files.
A mod might not even implement a method to have a "lockout" resulting in a user having to disable mods and restart a server to achieve a basic task which would be otherwise accomplished by typing `whitelist on`.

diesieben07:
After further discussion on Discord:

- This will _definitely_ not go in it's current form (mods should not be able to overrule vanilla whitelist, etc.).
- It is questionable why this needs to be an event at all. Why do multiple mods need to control the whitelist? Why do mods need to control the whitelist at all? Your use-case of MCLink (the only one provided here) is better solved by e.g. a cronjob which sets the whitelist.

Please provide any other use-cases you might have for this.

phit:
- first point is fine with me, I'll change it so it only fires if the player is allowed to join by vanilla
- multiple mods don't really need to control it, I suppose you are suggesting implementing this like permissions with one handler? cronjobs are useless it means you need control the underlying system, which isn't the case for most servers with server providers, and even if you do, you still need to somehow trigger the reload of the whitelist via e.g. RCON, having it all handled in a mod is a lot cleaner
here's an example mod I've used in the past to achieve similar results:
https://www.curseforge.com/minecraft/mc-mods/whitelister there's also quite a few examples on CF https://www.curseforge.com/minecraft/mc-mods/search?category=&search=Whitelist

dries007:
On the first: This is already possible, just not very convenient. I really don't understand why there are any legitimate conserns about this. Bukkit has been doing this for years. A mod can use any trigger to disable the whitelist all together, including a login attempt by someone, so if you wanted to you could already bypass the whitelist. If you cannot trust a mod with this kind of power, why would you trust them with full read/write/execute permissions on your computer? But enough of that.

On the second:
- A cron job relies on a periodic update. MCLink is live. It checks APIs on the fly which is a large part of the appeal. It works the moment you click the subscribe button. You can never have that with a periodic update.
- Most simple server hosts don't have cron jobs.
- People don't want the hassle of configuring cron jobs. They already ask me for simpler things than editing a config file because it's confusing.
- Separation of conserns. With a normal event 1 mod can be used per system, e.g. Discord, Patreon, Twitch, ...
- This works and exists in Bukkit, therefore people expect it to "just work" in Forge as well.

EDIT: Oh, just FYI: Feel free to @-me on discord (@dries007#8913), I'd love to discuss more if you prefer that platform or if you want to know more about MCLink.

diesieben07:
Having it as an event is not trivial.
Who decides on the order? If you have two whitelist mods, the first one rejects the player because they are not on the whitelist... then the 2nd one says "oh no, this guy is on the whitelist" and removes the rejection. But what if the first mod was a banlist mod and the overruling of the rejection makes no sense?

This is not a trivial problem that is not solved by simply firing and event and letting people fight over the contents of the rejection message.

dries007:
The priority system by Forge can be used for ordering, it's not very granular, but it does work.

The event tracks the origianal status (aka what vanilla would have done) so mods can act accordingly:
- A whitelist mod would only turn vanilla "non whitelist" rejections into OK and not touch any event with another initial state.
- A collective ban / reputation mod could be configured to block logins regardless of any previous state.
- A VIP list mod could overwrite the vanilla "full server" state only to allow sertain groups of people to join regardless.

I'd say, leave it up to the modders to decide based on the usecase they want to cover.

Altough it might be benefitial to add an enum that tracks the state, so the originalMessage doesn't have to be parsed/compared by every listener.

diesieben07:
> The priority system by Forge can be used for ordering, it's not very granular, but it does work.

it is not user controllable. What will happen is that mods will put highest priority on everything because their mod is the most important.

As for your other points: Yes, it is possible. But it is not solved by simply creating an event and allowing people to set the rejection message (which is the current state of the PR).

Claycorp:
I have a mod I just recently made that would make use of this also.

Many times I've needed to keep players out of a server while fixing something that has gone wrong while the server is live. I don't want to be constantly telling people to leave or mess with whitelist/server settings to lock out the server that may require a restart for a simple task. Thus the easy option is to prevent people from joining unless they are OP'ed. Easily activated and deactivated all while the server is running and causes the least interference/possible issues.


Also isn't the whole "Who decides on the order" argument moot? Most, if not all, modifiable events have the same issue yet they still exist with the same problem that makes adding this an issue. You can't have a mod that stops someone from doing X and have a mod that allows someone to do X at the same time or even some of them where you can't do two of the same thing at the same time. Users also don't have direct control over what's picked but devs are more than capable of fixing the issue or providing options around it. Otherwise users will find a new mod that does both or just not use one of them.

dries007:
> As for your other points: Yes, it is possible. But it is not solved by simply creating an event and allowing people to set the rejection message (which is the current state of the PR).

I'm willing to put in the effort to modify this PR with the author or create a new one which addresses this, but I first want some consensus on what it should approximately look like to avoid straight up rejection.

diesieben07:
The order argument is not mood. Most events exist not to modify vanilla things per-se but to add a hook to vanilla things _when your mod is involved_.
Take UseHoeEvent: You won't do anything here if the block being target isn't yours. As such: Order is a non-issue.
Here however the express purpose is to modify vanilla things and also do it from multiple mods at once.

@dries007 I'm not the one making the decision whether things go in or not. https://github.com/MinecraftForge/MinecraftForge/pull/6889#issuecomment-666619393 was me relaying to you the opinion of the project maintainer.

TheCurle:
There has been talk of this PR in other platforms, and the consensus is that this needs large architectural changes to work.
Not to mention it's stale. 

As such, I'll close it. If this is something that modders desperately need (it seems not, given that this is possible by other means) then you need to talk with the triage team.

PR #6893, opened 2020-06-30
[1.16] Added AnvilBreakEvent, AnvilDamageEvent & AnvilDamageEvent.Falling/Container

Lanse505:
This PR aims to add 4 new events.

AnvilBreakEvent - Fired whenever the Anvil breaks by conventional means (Container Damage or FallingBlockDamage).
AnvilDamageEvent - Catch-All event for when an anvil is damaged.
AnvilDamageEvent.Falling - Fired whenever the Anvil is being attempted to be damaged through the FallingBlockEntity Code.
AnvilDamageEvent.Container - Fired whenever the Anvil is being attempted to be damaged through the RepairContainer code.

So what purpose does these events fill?

**AnvilBreakEvent**
Previously we've been able to catch Anvil breaks through the world.
Currently this isn't possible so this event was created to fill that void.

**AnvilDamageEvent**
This event and it's sub-events is created so that you can control if/how the anvil is damaged through the normal means.
For example in the test mod, I've made it so the anvil isn't damaged if it lands on Wool or if the player is holding a stick in the case of the container. This could be used in the case of "RPG" mods to make it so there is a smaller chance for the anvil to take damage the higher the players "smithing" is or whatever.

Included is a test mod that shows off these events.

### DIFF
--- a/patches/minecraft/net/minecraft/entity/item/FallingBlockEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/item/FallingBlockEntity.java.patch
@@ -1,19 +1,17 @@
 --- a/net/minecraft/entity/item/FallingBlockEntity.java
 +++ b/net/minecraft/entity/item/FallingBlockEntity.java
-@@ -153,7 +153,7 @@
-                               ((FallingBlock)block).func_176502_a_(this.field_70170_p, blockpos1, this.field_175132_d, blockstate, this);
-                            }
+@@ -198,10 +198,12 @@
+                entity.func_70097_a(damagesource, (float)Math.min(MathHelper.func_76141_d((float)i * this.field_145816_i), this.field_145815_h));
+             }
  
--                           if (this.field_145810_d != null && block instanceof ITileEntityProvider) {
-+                           if (this.field_145810_d != null && this.field_175132_d.hasTileEntity()) {
-                               TileEntity tileentity = this.field_70170_p.func_175625_s(blockpos1);
-                               if (tileentity != null) {
-                                  CompoundNBT compoundnbt = tileentity.func_189515_b(new CompoundNBT());
 -            if (flag && (double)this.field_70146_Z.nextFloat() < (double)0.05F + (double)i * 0.05D) {
-+            if (flag && (double)this.field_70146_Z.nextFloat() < (double)0.05F + (double)i * 0.05D && !net.minecraftforge.common.ForgeHooks.onAnvilDamageFalling(this.field_175132_d, this.field_70122_E ? this.func_130014_f_().func_180495_p(this.func_226270_aj_()) : null)) {
++            if (flag && (double)this.field_70146_Z.nextFloat() < (double)0.05F + (double)i * 0.05D && net.minecraftforge.common.ForgeHooks.onAnvilDamageFalling(this.field_175132_d, this.func_226268_ag_(), this.func_130014_f_().func_180495_p(func_226270_aj_()), this.func_226270_aj_(), this, this.func_130014_f_())) {
                 BlockState blockstate = AnvilBlock.func_196433_f(this.field_175132_d);
                 if (blockstate == null) {
 -                  this.field_145808_f = true;
-+                  if (net.minecraftforge.common.ForgeHooks.onAnvilBreak(this.field_175132_d, field_70170_p, func_226268_ag_(), this)) {
++                  if (net.minecraftforge.common.ForgeHooks.onAnvilBreak(this.field_175132_d, this.func_130014_f_(), this.func_226268_ag_(), this, null)) {
 +                     this.field_145808_f = true;
 +                  }
+                } else {
+                   this.field_175132_d = blockstate;
+                }
--- a/patches/minecraft/net/minecraft/entity/item/FallingBlockEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/item/FallingBlockEntity.java.patch
@@ -11,3 +11,6 @@
                                   CompoundNBT compoundnbt = tileentity.func_189515_b(new CompoundNBT());
 -            if (flag && (double)this.field_70146_Z.nextFloat() < (double)0.05F + (double)i * 0.05D) {
 +            if (flag && (double)this.field_70146_Z.nextFloat() < (double)0.05F + (double)i * 0.05D && !net.minecraftforge.common.ForgeHooks.onAnvilDamageFalling(this.field_175132_d, this.field_70122_E ? this.func_130014_f_().func_180495_p(this.func_226270_aj_()) : null)) {
+                BlockState blockstate = AnvilBlock.func_196433_f(this.field_175132_d);
+                if (blockstate == null) {
+                   this.field_145808_f = true;
--- a/patches/minecraft/net/minecraft/entity/item/FallingBlockEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/item/FallingBlockEntity.java.patch
@@ -13,4 +13,7 @@
 +            if (flag && (double)this.field_70146_Z.nextFloat() < (double)0.05F + (double)i * 0.05D && !net.minecraftforge.common.ForgeHooks.onAnvilDamageFalling(this.field_175132_d, this.field_70122_E ? this.func_130014_f_().func_180495_p(this.func_226270_aj_()) : null)) {
                 BlockState blockstate = AnvilBlock.func_196433_f(this.field_175132_d);
                 if (blockstate == null) {
-                   this.field_145808_f = true;
+-                  this.field_145808_f = true;
++                  if (net.minecraftforge.common.ForgeHooks.onAnvilBreak(this.field_175132_d, field_70170_p, func_226268_ag_(), this)) {
++                     this.field_145808_f = true;
++                  }
--- a/patches/minecraft/net/minecraft/entity/item/FallingBlockEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/item/FallingBlockEntity.java.patch
@@ -1,17 +1,11 @@
 --- a/net/minecraft/entity/item/FallingBlockEntity.java
 +++ b/net/minecraft/entity/item/FallingBlockEntity.java
-@@ -198,10 +198,12 @@
-                entity.func_70097_a(damagesource, (float)Math.min(MathHelper.func_76141_d((float)i * this.field_145816_i), this.field_145815_h));
-             }
- 
--            if (flag && (double)this.field_70146_Z.nextFloat() < (double)0.05F + (double)i * 0.05D) {
-+            if (flag && (double)this.field_70146_Z.nextFloat() < (double)0.05F + (double)i * 0.05D && net.minecraftforge.common.ForgeHooks.onAnvilDamageFalling(this.field_175132_d, this.func_226268_ag_(), this.func_130014_f_().func_180495_p(func_226270_aj_()), this.func_226270_aj_(), this, this.func_130014_f_())) {
-                BlockState blockstate = AnvilBlock.func_196433_f(this.field_175132_d);
-                if (blockstate == null) {
--                  this.field_145808_f = true;
-+                  if (net.minecraftforge.common.ForgeHooks.onAnvilBreak(this.field_175132_d, this.func_130014_f_(), this.func_226268_ag_(), this, null)) {
-+                     this.field_145808_f = true;
-+                  }
-                } else {
-                   this.field_175132_d = blockstate;
-                }
+@@ -153,7 +153,7 @@
+                               ((FallingBlock)block).func_176502_a_(this.field_70170_p, blockpos1, this.field_175132_d, blockstate, this);
+                            }
+
+-                           if (this.field_145810_d != null && block instanceof ITileEntityProvider) {
++                           if (this.field_145810_d != null && this.field_175132_d.hasTileEntity()) {
+                               TileEntity tileentity = this.field_70170_p.func_175625_s(blockpos1);
+                               if (tileentity != null) {
+                                  CompoundNBT compoundnbt = tileentity.func_189515_b(new CompoundNBT());
--- a/patches/minecraft/net/minecraft/entity/item/FallingBlockEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/item/FallingBlockEntity.java.patch
@@ -9,3 +9,5 @@
                                TileEntity tileentity = this.field_70170_p.func_175625_s(blockpos1);
                                if (tileentity != null) {
                                   CompoundNBT compoundnbt = tileentity.func_189515_b(new CompoundNBT());
+-            if (flag && (double)this.field_70146_Z.nextFloat() < (double)0.05F + (double)i * 0.05D) {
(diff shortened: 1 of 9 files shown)


src/main/java/net/minecraftforge/event/anvil/AnvilDamageEvent.java
```diff
@@ -0,0 +1,87 @@
...
+public class AnvilDamageEvent extends Event
+{
+    /**
+     * The current state of the Anvil pre-damage
+     */
+    @Nonnull
+    private final BlockState currentState;
+
+    /**
+     * A simple event fired whenever code is fired to attempt to damage the Anvil.
```
diesieben07: What does this even mean? "whenever code is fired"?
Also, this documentation should be on the class, not the constructor.

src/main/java/net/minecraftforge/event/anvil/AnvilDamageEvent.java
```diff
@@ -0,0 +1,87 @@
...
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+@Cancelable
+public class AnvilDamageEvent extends Event
+{
+    /**
+     * The current state of the Anvil pre-damage
+     */
```
diesieben07: This documentation should be on the getters, not the fields. The getters are the public API.

src/main/java/net/minecraftforge/event/anvil/AnvilDamageEvent.java
```diff
@@ -0,0 +1,87 @@
...
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.world.World;
+import net.minecraftforge.eventbus.api.Cancelable;
+import net.minecraftforge.eventbus.api.Event;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+@Cancelable
+public class AnvilDamageEvent extends Event
```
diesieben07, line 21: Why does this event not include the world and position of the anvil?

src/main/java/net/minecraftforge/event/anvil/AnvilDamageEvent.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Nonnull
+    public BlockState getCurrentState() { return currentState; }
+
+    /**
+     * A simple event fired whenever code is fired from the FallingBlockEntity to damage the Anvil.
+     * If the event is canceled, vanilla behaviour will not run and the Anvil's BlockState will not change (It won't be damaged).
+     * If the event is not canceled, vanilla behaviour will run and the Anvil will be damaged.
+     */
+    public static class Falling extends AnvilDamageEvent
```
diesieben07, line 42: Why not include the falling entity and world here? And maybe even the position of the "ground".

src/main/java/net/minecraftforge/event/anvil/AnvilDamageEvent.java
```diff
@@ -0,0 +1,87 @@
...
+    {
+        /**
+         * The current blockstate beneath the FallingBlockEntity for the Anvil.
+         */
+        @Nullable
```
diesieben07, line 44: Docs should be on the getter.

src/main/java/net/minecraftforge/event/anvil/AnvilDamageEvent.java
```diff
@@ -0,0 +1,87 @@
...
+        private final World world;
+
+        /**
+         * A simple event fired whenever code is fired from the RepairContainer to damage the Anvil.
+         * If the event is canceled, vanilla behaviour will not run and the Anvil's BlockState will not change (It won't be damaged).
+         * If the event is not canceled, vanilla behaviour will run and the Anvil will be damaged.
+         *
+         * @param currentState
+         */
+        public Container(@Nonnull BlockState currentState, PlayerEntity player, World world)
```
diesieben07: Docs should be on the class, not constructor.
Why does this not include the `RepairContainer`?

src/main/java/net/minecraftforge/event/anvil/AnvilUpdateEvent.java
```diff
@@ -17,7 +17,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
  */
 
-package net.minecraftforge.event;
+package net.minecraftforge.event.anvil;
```
diesieben07, line 20: Is this really necessary?

src/main/java/net/minecraftforge/event/anvil/AnvilBreakEvent.java
```diff
@@ -0,0 +1,49 @@
...
+import net.minecraftforge.eventbus.api.Cancelable;
+import net.minecraftforge.eventbus.api.Event;
+
+@Cancelable
+public class AnvilBreakEvent extends Event
+{
+    /**
+     * This is the state of the block before it broke.
+     */
+    private final BlockState statePreBreaking;
```
diesieben07, line 18: Docs should be on the getter.

src/main/java/net/minecraftforge/event/anvil/AnvilBreakEvent.java
```diff
@@ -0,0 +1,49 @@
...
+
+import net.minecraft.block.BlockState;
+import net.minecraft.entity.Entity;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+import net.minecraftforge.eventbus.api.Cancelable;
+import net.minecraftforge.eventbus.api.Event;
+
+@Cancelable
+public class AnvilBreakEvent extends Event
```
diesieben07, line 15: Missing documentation.

src/main/java/net/minecraftforge/event/anvil/AnvilBreakEvent.java
```diff
@@ -0,0 +1,49 @@
...
+    private final World world;
+
+    /**
+     * This is the BlockPos of where the block broke.
+     */
+    private final BlockPos pos;
+
+    /**
+     * This is the entity involved with the break.
+     * This can either be a FallingBlockEntity or a PlayerEntity.
```
diesieben07: These should be separate then and properly named. Don't dump unrelated things into the same field.

diesieben07 (changes requested):
Your patches are broken, how did you generate them?

src/main/java/net/minecraftforge/event/anvil/AnvilDamageEvent.java
Lanse505: The general purpose catch-all event is called from "AnvilBlock#damage" which has and provides no contexts outside of the previous Blockstate. This event is just as stated a "general catch all" event and provides an entry point for anyone who wants to just stop anvil damage without context. Falling and Container sub-events are the only ones that can and does provide additional context.

I've adressed your points in regard to missing contexts such as:
Falling now has the FallingBlockEntity Context and World Context.
Container now have the RepairContainer Context.

src/main/java/net/minecraftforge/event/anvil/AnvilUpdateEvent.java
Lanse505: I can move the Anvil events back out to the main event folder again if so pleases.
I decided to create a anvil package seeing as there are now 3 Anvil related events.

src/main/java/net/minecraftforge/event/anvil/AnvilDamageEvent.java
diesieben07: Is that event even useful at all then? There is no context at all...
At least let people change the state, maybe?

src/main/java/net/minecraftforge/event/anvil/AnvilBreakEvent.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * @return Returns the BlockState before it broke.
+     */
+    @Nonnull
+    public BlockState getStatePreBreaking() { return statePreBreaking; }
+
+    /**
+     * This is either supplied by the FallingBlockEntity or the PlayerEntity.
```
diesieben07, line 44: Wrong docs

src/main/java/net/minecraftforge/event/anvil/AnvilBreakEvent.java
```diff
@@ -0,0 +1,67 @@
...
+    public BlockPos getPos() { return pos; }
+
+    /**
+     * @return Returns the FallingBlockEntity, This can return Null!
+     */
+    @Nullable
+    public FallingBlockEntity getFallingBlockEntity() { return fallingBlockEntity; }
+
+    /**
+     * @return Returns the PlayerEntity, This can return Null!
```
diesieben07, line 63: Add info about when it is null and when not

src/main/java/net/minecraftforge/event/anvil/AnvilBreakEvent.java
```diff
@@ -0,0 +1,67 @@
...
+    public World getWorld() { return world; }
+
+    /**
+     * @return Returns the position of where the Break took place.
+     */
+    @Nonnull
+    public BlockPos getPos() { return pos; }
+
+    /**
+     * @return Returns the FallingBlockEntity, This can return Null!
```
diesieben07, line 57: Add info about when it is null and when not

src/main/java/net/minecraftforge/event/anvil/AnvilDamageEvent.java
Lanse505: I mean that's fair enough.
I'll remove it then and just keep falling and container.

patches/minecraft/net/minecraft/entity/item/FallingBlockEntity.java.patch
```diff
@@ -1,11 +1,17 @@
...
+                entity.func_70097_a(damagesource, (float)Math.min(MathHelper.func_76141_d((float)i * this.field_145816_i), this.field_145815_h));
+             }
  
--                           if (this.field_145810_d != null && block instanceof ITileEntityProvider) {
-+                           if (this.field_145810_d != null && this.field_175132_d.hasTileEntity()) {
-                               TileEntity tileentity = this.field_70170_p.func_175625_s(blockpos1);
-                               if (tileentity != null) {
-                                  CompoundNBT compoundnbt = tileentity.func_189515_b(new CompoundNBT());
+-            if (flag && (double)this.field_70146_Z.nextFloat() < (double)0.05F + (double)i * 0.05D) {
++            if (flag && (double)this.field_70146_Z.nextFloat() < (double)0.05F + (double)i * 0.05D && net.minecraftforge.common.ForgeHooks.onAnvilDamageFalling(this.field_175132_d, this.func_226268_ag_(), this.func_130014_f_().func_180495_p(func_226270_aj_()), this.func_226270_aj_(), this, this.func_130014_f_())) {
```
diesieben07: You don't need to pass all these parameters here.
- fallTile can be accessed through getBlockState.
- getEntityWorld can be accessed

So you only need to pass `this`, and the two positions.

patches/minecraft/net/minecraft/entity/item/FallingBlockEntity.java.patch
```diff
@@ -1,11 +1,17 @@
...
+                BlockState blockstate = AnvilBlock.func_196433_f(this.field_175132_d);
+                if (blockstate == null) {
+-                  this.field_145808_f = true;
++                  if (net.minecraftforge.common.ForgeHooks.onAnvilBreak(this.field_175132_d, this.func_130014_f_(), this.func_226268_ag_(), this, null)) {
```
diesieben07: Don't use braces and don't indent the assignment, then this patch becomes a 1 line addition.
And same with the parameters.

patches/minecraft/net/minecraft/entity/item/FallingBlockEntity.java.patch
```diff
@@ -1,11 +1,17 @@
...
-                            }
+@@ -198,10 +198,12 @@
+                entity.func_70097_a(damagesource, (float)Math.min(MathHelper.func_76141_d((float)i * this.field_145816_i), this.field_145815_h));
+             }
  
--                           if (this.field_145810_d != null && block instanceof ITileEntityProvider) {
-+                           if (this.field_145810_d != null && this.field_175132_d.hasTileEntity()) {
-                               TileEntity tileentity = this.field_70170_p.func_175625_s(blockpos1);
-                               if (tileentity != null) {
-                                  CompoundNBT compoundnbt = tileentity.func_189515_b(new CompoundNBT());
```
diesieben07, line 11: You completely broke this patch file because you removed the existing patch.
Compare:
Before: https://github.com/MinecraftForge/MinecraftForge/blob/bf1a0461cb321f48f0738503d0f028ba6f054d11/patches/minecraft/net/minecraft/entity/item/FallingBlockEntity.java.patch
After: https://github.com/MinecraftForge/MinecraftForge/blob/4a63773d657edb5ee1f913885e1f253c25cdcdba/patches/minecraft/net/minecraft/entity/item/FallingBlockEntity.java.patch

patches/minecraft/net/minecraft/inventory/container/RepairContainer.java.patch
```diff
@@ -1,15 +1,26 @@
...
++         if (!p_230301_1_.field_71075_bZ.field_75098_d && blockstate.func_235714_a_(BlockTags.field_200572_k) && p_230301_1_.func_70681_au().nextFloat() < 0.12F  && !net.minecraftforge.common.ForgeHooks.onAnvilDamageContainer(blockstate, p_230301_1_, p_230301_1_.func_130014_f_(), this)) {
              BlockState blockstate1 = AnvilBlock.func_196433_f(blockstate);
-@@ -96,8 +98,11 @@
+             if (blockstate1 == null) {
+-               p_234633_1_.func_217377_a(p_234633_2_, false);
+-               p_234633_1_.func_217379_c(1029, p_234633_2_, 0);
++               if (net.minecraftforge.common.ForgeHooks.onAnvilBreak(blockstate, p_230301_1_.func_130014_f_(), p_234633_2_, null, p_230301_1_)) {
++                  p_234633_1_.func_217377_a(p_234633_2_, false);
++                  p_234633_1_.func_217379_c(1029, p_234633_2_, 0);
++               }
```
diesieben07, line 19: Don't change the indentation here.

diesieben07 (changes requested):

src/main/java/net/minecraftforge/event/anvil/AnvilUpdateEvent.java
KnightMiner: Moving an existing event will break things for anyone using the old event. While it may be more organized, it makes an unneeded breaking change. You could just mark it with a `TODO: move to net.minecraftforge.event.anvil` and get it moved next major update.

pupnewfster: Not sure if this should be moved or not, but given there has not been a RB yet so breaking changes are still allowed, this ***is*** the next major update.

Lanse505: Will move everything back and add todos then.
Currently re-evaluating some stuff either way.

PR #6895, opened 2020-07-01
Fix patch error in ItemEntity

### LINKED ISSUE #6854: [1.16.1] ItemEntity patch port error, readAdditional code is in writeAdditional
**Minecraft Version:** 1.16.1

**Forge Version:** 32.0.23

**Logs:** N/A

**Steps to Reproduce:**
```
	@SubscribeEvent
	public void onItemJoinWorld(EntityJoinWorldEvent e) {
		World world = e.getWorld();
		if (world.isRemote) {
			return;
		}
		
		Entity entity = e.getEntity();
		if (entity instanceof ItemEntity == false) {
			return;
		}
		
		ItemEntity ie = (ItemEntity)entity;
		CompoundNBT nbtc = ie.serializeNBT();
		nbtc.putInt("Lifespan", 20); // set lifespan in ticks
		ie.read(nbtc);
	}
```

**Description of issue:**
Hi! First of all, thank you for the quick update to 1.16. Appreciate all the work you do very much.

In 1.15, this code was able to change the lifespan of all ItemEntitities in the world. By changing the NBT value. In 1.16 I can't seem to change the lifespan, even though I don't see any changes in the immediate code. Not experienced enough with the Forge format to make a pull request, hope this can be of some use though.

ichttt:
Fixes #6854 read being called in `writeAdditional` instead of `readAdditional`
Also regenerate Patch for HoeItem as context changed due to a new AT for the constructor

### DIFF
--- a/patches/minecraft/net/minecraft/entity/item/ItemEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/item/ItemEntity.java.patch
@@ -78,14 +78,14 @@
        if (this.func_200214_m() != null) {
           p_213281_1_.func_186854_a("Thrower", this.func_200214_m());
        }
-@@ -259,6 +277,7 @@
-       if (this.func_200215_l() != null) {
-          p_213281_1_.func_186854_a("Owner", this.func_200215_l());
+@@ -272,6 +290,7 @@
+       if (p_70037_1_.func_74764_b("PickupDelay")) {
+          this.field_145804_b = p_70037_1_.func_74765_d("PickupDelay");
        }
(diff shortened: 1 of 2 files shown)


marchermans (approved):
Looks okey.

Just a question: Patches are applied after the AT is done correct? Hence the new HoeItem patch, because it is now in an AT?

TheCurle:
Correct. Patches are after ATs, and can override them.

TheCurle:
This was one of the "most important" patches on the to-do list, i assume Lex just hasn't gotten around to this yet: 
https://gist.github.com/TheCurle/5a966b92ad0c8bdb300516c6e8f8fa30

PR #6896, opened 2020-07-01
[1.16.x] Add support for custom Nether Portal Blocks

artman41:
- Modify accessibility so that `NetherPortalBlock$Size` can be extended, allowing for custom `NetherPortalBlock`s
- Allow `NetherPortalBlock$Size` to be instantiated with a given `portalBlock`

### DIFF
--- a/patches/minecraft/net/minecraft/block/NetherPortalBlock.java.patch
+++ b/patches/minecraft/net/minecraft/block/NetherPortalBlock.java.patch
@@ -45,3 +45,64 @@
                 this.field_150862_g = 0;
                 break;
              }
+
+@@ -217,19 +217,25 @@ public class NetherPortalBlock extends Block {
+    }
+
+    public static class Size {
+-      private final IWorld world;
+-      private final Direction.Axis axis;
+-      private final Direction rightDir;
+-      private final Direction leftDir;
+-      private int portalBlockCount;
++      protected final IWorld world;
++      protected final Direction.Axis axis;
++      protected final Direction rightDir;
++      protected final Direction leftDir;
++      protected int portalBlockCount;
+       @Nullable
+-      private BlockPos bottomLeft;
+-      private int height;
+-      private int width;
++      protected BlockPos bottomLeft;
++      protected int height;
++      protected int width;
++      protected Block portalBlock;
+
+       public Size(IWorld worldIn, BlockPos pos, Direction.Axis axisIn) {
++         this(worldIn, pos, axisIn, Blocks.NETHER_PORTAL);
++      }
++
++      public Size(IWorld worldIn, BlockPos pos, Direction.Axis axisIn, Block portalBlock) {
+          this.world = worldIn;
+          this.axis = axisIn;
++         this.portalBlock = portalBlock;
+          if (axisIn == Direction.Axis.X) {
+             this.leftDir = Direction.EAST;
+             this.rightDir = Direction.WEST;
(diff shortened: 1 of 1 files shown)


pupnewfster:
Isn't this already possible via `IForgeBlock#isPortalFrame`? https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java#L582

artman41:
@pupnewfster if I'm not mistaken, that's referring to the Obsidian outer frame correct?

`IForgeBlock.java:574`
```java
    /**
     * Determines if this block can be used as part of a frame of a nether portal.
     *
     * @param state The current state
     * @param world The current world
     * @param pos Block position in world
     * @return True, to support being part of a nether portal frame, false otherwise.
     */
    default boolean isPortalFrame(BlockState state, IWorldReader world, BlockPos pos)
    {
        return state.isIn(Blocks.OBSIDIAN);
    }
```

pupnewfster:
That is correct, is this PR for the actual portal itself?

artman41:
It is

patches/minecraft/net/minecraft/block/NetherPortalBlock.java.patch
```diff
@@ -45,3 +45,64 @@
...
+-      private final IWorld world;
+-      private final Direction.Axis axis;
+-      private final Direction rightDir;
+-      private final Direction leftDir;
+-      private int portalBlockCount;
++      protected final IWorld world;
++      protected final Direction.Axis axis;
++      protected final Direction rightDir;
++      protected final Direction leftDir;
++      protected int portalBlockCount;
```
diesieben07, line 62: Please use an AT

patches/minecraft/net/minecraft/block/NetherPortalBlock.java.patch
```diff
@@ -45,3 +45,64 @@
...
+       @Nullable
+-      private BlockPos bottomLeft;
+-      private int height;
+-      private int width;
++      protected BlockPos bottomLeft;
++      protected int height;
++      protected int width;
```
diesieben07, line 69: Please use an AT

patches/minecraft/net/minecraft/block/NetherPortalBlock.java.patch
```diff
@@ -45,3 +45,64 @@
...
++      protected Block portalBlock;
+
+       public Size(IWorld worldIn, BlockPos pos, Direction.Axis axisIn) {
++         this(worldIn, pos, axisIn, Blocks.NETHER_PORTAL);
++      }
++
++      public Size(IWorld worldIn, BlockPos pos, Direction.Axis axisIn, Block portalBlock) {
+          this.world = worldIn;
+          this.axis = axisIn;
++         this.portalBlock = portalBlock;
```
diesieben07, line 79: Move this assignment up (first thing in the constructor) to get a consecutive patch.

diesieben07 (changes requested):
Your patch file is completely broken, did you edit it by hand?

patches/minecraft/net/minecraft/block/NetherPortalBlock.java.patch
artman41: @diesieben07 could I get a definition on `at`?

TheSilkMiner: Access Transformer

Issue #4624, opened 2017-12-26
GuiOpenEvent should be enhanced to pass in "old" GUI as well


jabelar:
Currently the GuiOpenEvent passes in only the "new" GuiScreen that is intended to be opened. However, at the point where the Minecraft#displayGuiScreen() fires the event it has just recorded whether any current screen was active. Here's the current code:
`        GuiScreen old = this.currentScreen;
        net.minecraftforge.client.event.GuiOpenEvent event = new net.minecraftforge.client.event.GuiOpenEvent(guiScreenIn);

        if (net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(event)) return;

        guiScreenIn = event.getGui();
        if (old != null && guiScreenIn != old)
        {
            old.onGuiClosed();
        }
`

There are cases when it would be good to know what the previous GUI was, so it would be cool is the "old" value was also passed (in an immutable way) to the event.

marvin-roesch:
Well, while it is technically mutable, `Minecraft#currentScreen` contains exactly that information. But a PR to include the old GUI in the event shouldn't take more than a minute :P

jabelar:
Actually, I guess that is valid point. The screen is always called from within the Minecraft class on the client side, so it is safe to grab it again with Minecraft#currentScreen. So never mind, this isn't needed.

PR #6897, opened 2020-07-02
Allows the old screen to be accessed in GuiOpenEvent  

deniszitu:
This PR allows, when replacing a screen in GuiOpenEvent, the parent screen can be passed as a reference to new screen.

old issue #4624 lol

### DIFF
--- a/patches/minecraft/net/minecraft/client/Minecraft.java.patch
+++ b/patches/minecraft/net/minecraft/client/Minecraft.java.patch
@@ -136,7 +136,7 @@
        }
  
 +      Screen old = this.field_71462_r;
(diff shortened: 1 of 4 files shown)


diesieben07:
You can already do this by accessing `Minecraft#currentScreen` when the event is fired. Why is this needed?

deniszitu:
> You can already do this by accessing `Minecraft#currentScreen` when the event is fired. Why is this needed?

I could have sworn it was private 😅

`Minecraft.getInstance().currentScreen`

`event.getGui().getMinecraft().currentScreen`

I think with this change it’s more "intuitive"

`event.getOldGui()`

PR #6899, opened 2020-07-02
Re-add HoverChecker

sciwhiz12:
Re-adds the `HoverChecker` class from the 1.15.x branch. 
I presume it was removed because of the conflicts of SRG methods/fields of `Widget`.

I have not changed any functionality within the class, nor have I changed the javadocs or author tag. I have only translated the calls to methods/fields of Widget to their equivalent 1.16 SRG names.
I am unsure if I should make a test mod for this, but I will if it is requested.

### DIFF
--- a/src/main/java/net/minecraftforge/client/gui/HoverChecker.java
+++ b/src/main/java/net/minecraftforge/client/gui/HoverChecker.java
@@ -1,6 +1,6 @@
 /*
  * Minecraft Forge
- * Copyright (c) 2016-2019.
+ * Copyright (c) 2016-2020.
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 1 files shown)


src/main/java/net/minecraftforge/client/gui/HoverChecker.java
```diff
@@ -0,0 +1,106 @@
+/*
+ * Minecraft Forge
+ * Copyright (c) 2016-2019.
```
KnightMiner: It's 2020 now, I assume this should be updated?

gigaherz: Yes but it needs updating repository-wide, otherwise the license checker script will complain.

sciwhiz12: The file containing that is the LICENSE-header.txt at the root; I ran the task to format the license so it was automatically added. Another PR needs to be done that updates both the license header template and all files' header.

LexManos:
It was removed because we nuked a bunch of unused classes.
Why do you need this? It's a big class for what boils down to ~10 lines you could implement yourself.

sciwhiz12:
Ah, I thought it was removed because of the reason I said above.
I made this PR when @bl4ckscor3 was asking in Discord what happened to the class.
If this is the case, then I'll gladly close the PR.

Issue #6904, opened 2020-07-02
GuiOpenEvent being fired twice


deniszitu:
**Minecraft Version:** 1.16.1

**Forge Version:** 32.0.44

**Description of issue:**
GuiOpenEvent is fired twice whenever the ingame screen ((Screen) null) is displayed.

List of tested versions: 1.16.x, 1.15.x, 1.14.x

PR #6905, opened 2020-07-02
Fixed GuiOpenEvent being fired twice

deniszitu:
See #6904 

`MouseHelper#grabMouse()` unnecessarily calls `Minecraft#displayGuiScreen()` and consequently fires GuiOpenEvent twice.


LexManos:
There are quite a few places that call display(null) and then grabMouse.
I don't consider the multiple calls to display a bug we should fix.
It doesn't hurt anything. It has no performance impact, it has weird edge cases which would be noticeable to the user.
There just isn't any compelling reason to do this.

PR #6913, opened 2020-07-03
Load Modded Datapacks in DatapackScreen, before world creation

### LINKED ISSUE #6898: Datapack validation failed! 1.16.1
**Minecraft Version:** 1.16.1

**Forge Version:** 32.0.38

**Logs:** [debug.log](https://github.com/MinecraftForge/MinecraftForge/files/4862493/debug.log)

**Steps to Reproduce:**
 1. Create a new World
 2. Try to add a datapack rough the button in the caw screen

**Description of issue:**
If you try to add a datapack trough the Create a World screen. i says "datapack validation failed!"
problem

Cyborgmas:
There is a new Datapack selection menu when creating worlds, but it is currently impossible to add custom datapacks to it.
This is because a ResourcePackList is created, and it needs the ResourcePackLoader.loadResourcePacks hook added to find mod datapacks. Without mod datapacks, some tags get defined in code but are not found in datapack files.
This alone fixes #6898 .
However, I also polished up additional things. 
First, at that point Forge's Tags was not yet init, and its tags were not all being added to their tag collection. I am not sure what impact tags not being init could have for the datapack screen, but it seems wrong.
Secondly, the forge:dyes tags are created twice, I removed their creation from Tags and referenced DyeColor instead.
Thirdly, vanilla checks default datapack using a DatapackCodec so I created one which incorporates mods. (This is QOL only, if default and actual dont match, a verification occurs that loads up all datapacks but otherwise no impact)


Cyborgmas:
I also modified the default datapack check to incorporate modded dimensions. Datapacks were being verified only against "vanilla", which launched a check of all datapacks the first time "done" was clicked in the DatapackScreen, this was reset when re-entering the CreateWorldScreen, re-checking again.

patches/minecraft/net/minecraft/client/gui/screen/CreateWorldScreen.java.patch
```diff
@@ -0,0 +1,12 @@
+--- a/net/minecraft/client/gui/screen/CreateWorldScreen.java
++++ b/net/minecraft/client/gui/screen/CreateWorldScreen.java
+@@ -384,7 +384,9 @@
+       if (path != null) {
+          File file1 = path.toFile();
+          ResourcePackList<ResourcePackInfo> resourcepacklist = new ResourcePackList<>(ResourcePackInfo::new, new ServerPackFinder(), new FolderPackFinder(file1, IPackNameDecorator.field_232625_a_));
++         net.minecraftforge.fml.packs.ResourcePackLoader.loadResourcePacks(resourcepacklist, net.minecraftforge.fml.server.ServerLifecycleHooks::buildPackFinder);
+          resourcepacklist.func_198983_a();
++         this.field_238933_b_ = DatapackCodec.getModdedDefault();
```
LexManos: This needs to 
1) Not be a random method shoved onto DatapackCodec
2) Make sure all uses of the DatapackCodec.field_234880_a_ uses our default mod pack list.
3) Not change this entry every time the button is pressed, but instead do it once in the initalizer and use the argument passed into the constructor like vanilla does.

patches/minecraft/net/minecraft/util/datafix/codec/DatapackCodec.java.patch
```diff
@@ -0,0 +1,12 @@
+--- a/net/minecraft/util/datafix/codec/DatapackCodec.java
++++ b/net/minecraft/util/datafix/codec/DatapackCodec.java
+@@ -29,4 +29,9 @@
+    public List<String> func_234887_b_() {
+       return this.field_234883_d_;
+    }
++
++   @Deprecated //Forge internal
++   public static DatapackCodec getModdedDefault() {
```
LexManos: Again, why are you adding a random method onto this random class instead of say... anywhere else we stick random functions like ForgeHooks?

src/main/java/net/minecraftforge/common/Tags.java
```diff
@@ -34,6 +35,8 @@
 
 public class Tags
 {
+    public static void init () {}; //Dummy init, create the tags so that they are added to their collections.
```
LexManos: This does absolutely nothing. Congrats you run the static tantalizer of the outer class. It doesn't do anything to the inner classes.

src/main/java/net/minecraftforge/common/Tags.java
```diff
@@ -153,22 +156,22 @@
...
+        public static final ITag.INamedTag<Item> DYES_CYAN = DyeColor.CYAN.getTag();
+        public static final ITag.INamedTag<Item> DYES_LIGHT_GRAY = DyeColor.LIGHT_GRAY.getTag();
+        public static final ITag.INamedTag<Item> DYES_GRAY = DyeColor.GRAY.getTag();
+        public static final ITag.INamedTag<Item> DYES_PINK = DyeColor.PINK.getTag();
+        public static final ITag.INamedTag<Item> DYES_LIME = DyeColor.LIME.getTag();
+        public static final ITag.INamedTag<Item> DYES_YELLOW = DyeColor.YELLOW.getTag();
+        public static final ITag.INamedTag<Item> DYES_LIGHT_BLUE = DyeColor.LIGHT_BLUE.getTag();
+        public static final ITag.INamedTag<Item> DYES_MAGENTA = DyeColor.MAGENTA.getTag();
+        public static final ITag.INamedTag<Item> DYES_ORANGE = DyeColor.ORANGE.getTag();
+        public static final ITag.INamedTag<Item> DYES_WHITE = DyeColor.WHITE.getTag();
```
LexManos: I mean, is there a reason for this change? Ya I guess it's a little bit of cleanup making only one instance referenced multiple places. But that doesn't matter. The game will update both just fine.

src/main/resources/pack.mcmeta
```diff
@@ -1,6 +1,6 @@
 {
    "pack": {
-      "pack_format": 4,
+      "pack_format": 5,
       "description": "Forge resource pack"
```
LexManos, line 4: Unrelated but sure whatever

Cyborgmas: yeah, the datapack icon was just saying out of date, so fixed it

Cyborgmas:
DatapackCodec.field_234880_a_ is now updated with mods as soon as possible. 
On the server both the codec passed in and field_234880_a_ is updated.
On the client, datapacks are loaded pretty late, either on world load or in the datapack screen. However, resource packs are loaded early and have identical naming to datapacks (mod:modid). Using that, makes field_234880_a_ update early enough to avoid changing a lot of its references.

PR #6914, opened 2020-07-04
Custom Item integration with Piglins

pupnewfster:
Allows custom items to be marked as "currency" for piglin, so that when they are given them they use their loot table and give something random back. This PR also allows marking an item when worn in an armor slot as an item that can "pacify" the piglin and keep them neutral while worn.

### DIFF
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
@@ -36,7 +36,10 @@ import net.minecraft.entity.LivingEntity;
 import net.minecraft.entity.ai.attributes.Attribute;
 import net.minecraft.entity.ai.attributes.AttributeModifier;
 import net.minecraft.entity.item.ItemEntity;
+import net.minecraft.entity.monster.piglin.PiglinTasks;
 import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.item.ArmorItem;
+import net.minecraft.item.ArmorMaterial;
 import net.minecraft.item.Items;
 import net.minecraft.inventory.EquipmentSlotType;
 import net.minecraft.item.Item;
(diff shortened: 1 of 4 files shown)


Cyborgmas:
Have not checked the code, but can modders already add different blocks that when mined anger piglins? if not might be worth implementing a method for it here as well?

pupnewfster:
That is possible via tags that vanilla adds.

gigaherz:
A reminder that questions like that can easily be asked on Discord/forums and don't belong here. :P

TeamSpen210:
For the same reasons as in #6759, you might want to make the hook methods receive the `PiglinEntity` and the wearer (for the armor check). That way they can do something context-sensitive.

pupnewfster:
As it is in `PiglinTasks` I don't believe the piglin entity is easily accessible (at least without a messy patch), but I will go ahead and add the wearer to the methods.

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -113,6 +116,26 @@ default ActionResultType onItemUseFirst(ItemStack stack, ItemUseContext context)
         return ActionResultType.PASS;
     }
 
+    /**
+     * Called by Piglins when checking to see if they will give an item or something in exchange for this item.
+     *
+     * @return True if this item can be used as "currency" by piglins
+     */
+    default boolean isPiglinCurrency(ItemStack stack) {
```
diesieben07: Formatting (braces on newlines)

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -113,6 +116,26 @@ default ActionResultType onItemUseFirst(ItemStack stack, ItemUseContext context)
...
+    }
+
+    /**
+     * Called by Piglins to check if a given item marks a player as "honorable" and should not be attacked.
+     *
+     * @param wearer The entity wearing this ItemStack
+     *
+     * @return True if piglins are neutral to players wearing this item in an armor slot
+     */
+    default boolean pacifiesPiglins(ItemStack stack, LivingEntity wearer) {
```
diesieben07: Formatting (braces on newlines)

src/main/java/net/minecraftforge/common/extensions/IForgeItemStack.java
```diff
@@ -450,4 +450,24 @@ default boolean isRepairable()
     {
         return getStack().getItem().isRepairable(getStack());
     }
+
+    /**
+     * Called by Piglins when checking to see if they will give an item or something in exchange for this item.
+     *
+     * @return True if this item can be used as "currency" by piglins
+     */
+    default boolean isPiglinCurrency() {
```
diesieben07: Formatting (braces on newlines)

src/main/java/net/minecraftforge/common/extensions/IForgeItemStack.java
```diff
@@ -450,4 +450,24 @@ default boolean isRepairable()
...
+    }
+
+    /**
+     * Called by Piglins to check if a given item marks a player as "honorable" and should not be attacked.
+     *
+     * @param wearer The entity wearing this ItemStack
+     *
+     * @return True if piglins are neutral to players wearing this item in an armor slot
+     */
+    default boolean pacifiesPiglins(LivingEntity wearer) {
```
diesieben07: Formatting (braces on newlines)

diesieben07 (changes requested):

diesieben07 (approved):

PR #6918, opened 2020-07-04
[1.15.x / LTS version] Make ArgumentList a public class

alcatrazEscapee:
This is a useful util class used by forge run launchers, that would be nice if it could be used by external launch configs without a hassle. My personal use case is a custom run config that launches JUnit tests from within a dedicated server launch config. I don't see any particular API reason why this class isn't already public other than nothing within forge needs it outside if the package.

### DIFF
--- a/src/userdev/java/net/minecraftforge/userdev/ArgumentList.java
+++ b/src/userdev/java/net/minecraftforge/userdev/ArgumentList.java
@@ -33,7 +33,7 @@ import org.apache.logging.log4j.Logger;
  * A class that attempts to parse command line arguments into key value pairs to allow addition and editing.
  * Can not use JOptSimple as that doesn't parse out the values for keys unless the spec says it has a value.
  */
-class ArgumentList {
+public class ArgumentList {
     private static final Logger LOGGER = LogManager.getLogger();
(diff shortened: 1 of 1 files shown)


TheCurle:
~~Why is this a patch and not an AT?~~

Scratch that, i completely misread the filename.

LexManos:
Nothing in the userdev package is meant for end modder/user consumption. I can understand your use case, we should probably look into some form of unit test system for MC/Forge. And yes this is a nice little utility. But I do not want to expose it.

alcatrazEscapee:
That's understandable. I do realize I've been poking in things that weren't intended to be poked in, but in the interest of getting tests working...

> we should probably look into some form of unit test system for MC/Forge

If it's any help, I have (and thus the reason why I made this request) managed to get JUnit tests running within a dedicated server run configuration, using a little [library mod](https://github.com/alcatrazEscapee/mcjunitlib). I don't know you would find any of it useful, because frankly I don't really know what I'm doing most of the time, but it does work. The server launches within the transforming class loader environment, runs tests outputting results to the console, Minecraft source code can be used and everything is properly initialized. So if you're interested in an actual Junit implementation built in to forge, there's a working proof of concept.

PR #6919, opened 2020-07-04
[1.16] Add event for registering datapacks

DoubleNegation:
This PR adds a new server lifecycle event called `FMLServerDatapackDiscoveryEvent`.  
This event is fired before the server loads its datapacks. It allows handlers to add `IPackFinder`s to the server's `ResourcePackList`.

The new event allows mods to register datapacks which are generated at runtime.  
This used to be possible without this event before Minecraft 1.16 by adding the pack finder in `FMLServerAboutToStartEvent`. Since datapacks are now loaded a lot earlier, this new event is necessary to do the same thing.

#### Why, for the love of god, would you want to register a datapack at runtime?
Most prominentely, mods such as ResourceLoader, OpenLoader and TheLoader used to provide this functionality to mod pack creators. These mods are widely used and well-known, many modpacks rely on them. However, in the 1.16 update, they have to resort to techniques such as coremodding, which is generally not recommendable, to offer the same features as they did previously.  
This PR solves this issue by re-adding a way to register datapacks using Forge.

A test mod is also included with this PR to demonstrate how this event may be used. After loading into the game with the mod active, run `/datapack list` to verify that it's working.


Darkhax:
I think that putting the event inside Forge's `ResourcePackLoader.loadResourcePacks` method and then simply collect `IPackFinder` from mods using the event would be a much cleaner approach. Were there any specific reasons for doing things the way you have?

DoubleNegation:
`ResourcePackLoader.loadResourcePacks` is called by the server and the client. At the point in time when it is called from the client, however, the Forge event bus is shut down, so the event would never fire for the client anyway.

The clear separation of the server code from the common code therefore doesn't serve any functional purpose. Instead, it is supposed to make it clearer to a reader of the code that this event is only fired server-side. Additionally, there is a chance that initialization order changes in the future so that the Forge bus is started before this method is called client-side. In such a case, the behavior of the event would change unexpectedly for no immediately obvious reason in the simpler case, but not in the more complex one with the separate method.

That's why I suggested this solution. If turns out that the simpler way is preferred, I would of course update this PR accordingly.

Darkhax:
Ah I see, makes sense to me then.

diesieben07:
Some of your use-cases are bad, registering things like blocks dynamically is not a good idea.
I can however see something like craft tweaker using this, allowing basically a different format for recipe files, etc.

Darkhax:
I think most of the use cases are wrong/misguided but there is still merit in this PR. 

> A modpack utility that loads a folder from an irregular location as a datapack (similar to Resource Loader)

This is a good use case for this PR. Many mods such as OpenLoader, TheLoader, and ResourceLoader have had this functionality in previous versions but can no longer do this using the Forge/MC API. 

> A mod which adds custom versions of other mods' blocks/items/etc. (similar to Dense Ores)

Datapacks are not used to define content. Things like recipes that are only present when another mod or config option is set can be done using recipe conditions. This is not a good use case for this PR.

> A mod which allows changing the game via script files (similar to Mod Tweaker)

CraftTweaker/ModTweaker/MineTweaker does not use data packs at all and add the recipes directly via code. This use case is not supported by Forge and your PR would not be needed by their mod or help their mod become more legitimate in Forge's eyes.

> A modpack utility that allows adding simple data-defined blocks/items

Data packs can not define new content and this PR would not help to change that. There are many mods that attempt to do this already such as my WIP project Jragon. The solution for allowing these types of content packs to include normal data pack files is to just treat them as mods. This solves the issue of non-deterministic content and Forge already loads mods as data packs.

TeamSpen210:
Another use case which might be palatable is to load parts of a mod's datapack as an additional entry, so that it can be enabled/disabled. That would let you have easier/harder versions of recipes, perhaps also triggering other behaviours like hiding/showing blocks/items. As you mentioned that could be done with recipe conditions and a config, but having it as a seperate datapack would be more clear and more prominent in the interface.

DoubleNegation:
Thanks for all the feedback on the use cases! I updated the PR description accordingly.  

> Another use case which might be palatable is to load parts of a mod's datapack as an additional entry, so that it can be enabled/disabled. That would let you have easier/harder versions of recipes, perhaps also triggering other behaviours like hiding/showing blocks/items. As you mentioned that could be done with recipe conditions and a config, but having it as a seperate datapack would be more clear and more prominent in the interface.

With the current implementation of the PR, this is not really possible in a user-friendly way.  
This is because the datapack is only registered when the server starts up, so it doesn't show up in the Create New World -> Data Packs screen.  
The PR could be updated to allow for this functionality if this is a desirable feature for Forge.

TheCurle:
This PR seems to be mostly irrelevant now.
Custom datapack jsons can be loaded with a system similar to vanilla's discovery mechanisms.
The *Loader mods mentioned seem to have found their own ways to work around this, as have modpacks.

Given that this event fires for servers only, it's usability is limited, and the arguments for its inclusion are weak at best.
If you want to take another crack at this, I suggest looking at the aforementioned vanilla discovery systems and working on some form of wrapper to make that usable for modders.

In the mean time, this PR is closed.

Darkhax:
@TheCurle 

> The *Loader mods mentioned seem to have found their own ways to work around this, as have modpacks.

Please don't cite my mod as reasons to reject this PR. All these mods are relying on mixins to the same target while we wait for this PR to be properly reviewed by the Forge team.

> Given that this event fires for servers only, it's usability is limited, and the arguments for its inclusion are weak at best.

Mods can already add client pack finders without this PR, the point of this PR is to cover server pack finders which is a gap currently. Forge itself requires this special patch to inject the mod datapacks, this is just broadening said patch to allow other pack finder sources.

LexManos:
Yes we know that hacks can hack things. That's a given.
I've looked into it, and aside from bad field/param/method names this PR is fine.
There is no real good explained usecase for this. But it's a "stop everyone from hacking the same shit" type PR.

sciwhiz12:

PR #6926, opened 2020-07-06
Fix debug world not generating modded blocks

### LINKED ISSUE #6924: No modded blocks in debug world
**Minecraft Version:** 1.16.1

**Forge Version:** forge-1.16.1-32.0.47

**Logs:** [debug.log](https://gist.github.com/SE7-KN8/915604617b74ae8bc50cfaed81f9d2f2)

**Steps to Reproduce:**
 1. Create a new world with "Debug" type
 2. See that there are no modded blocks

**Description of issue:**
In the Debug world type, there are no modded blocks (only tested in dev-environment). In 1.15 this was not the case and all modded blocks were in the debug world too.

No blocks visible:

But they are registred:

sciwhiz12:
This PR fixes #6924.
See [my comment on the issue][explaincomment] for the explanation on why the issue exists, and the solutions I found for that issue.

_TL;DR:_ `DebugChunkGenerator` used to be instance-based, now it's singleton. Static reference to `Block` registry means that nothing after that class is loaded will be generated. `private static final` fields; One field is list of blockstates, two others are integers dependent on contents of the list.

This PR does the following:
  * Adds 3 lines to the Forge AccessTransformer to definalize the three fields.
  * Adds a patch to `DebugChunkGenerator`: a method that recalculates the information in the three fields in exactly the same way they are statically initialized.
  * Adds 2 lines (1 import) to the `GameData` class, specifically the `BlockCallbacks` inner class, to call the aforementioned method in `DebugChunkGenerator` when the registry is baked. As only one bake callback can be present for a registry, and `BlockCallbacks` takes that spot, I added to the bake callback.

~~I did find another solution for this problem ([debug_world_fix_1 branch on my fork][firstsolution]), but that requires adding a new patch which modifies the fields to public and introduces a new static method, which I felt to be too intrusive and may break future versions if that class is changed again.~~

Note: I did not put the `/* === FORGE START === */` and end markers, because when I asked on Discord, [tterag responded][tteragquote] that its "not required anymore".

[firstsolution]: https://github.com/sciwhiz12/MinecraftForge/tree/debug_world_fix_1
[explaincomment]: https://github.com/MinecraftForge/MinecraftForge/issues/6924#issuecomment-654184760
[tteragquote]: https://discordapp.com/channels/313125603924639766/725850371834118214/737770783060918285

### DIFF
--- a/patches/minecraft/net/minecraft/world/gen/DebugChunkGenerator.java.patch
+++ b/patches/minecraft/net/minecraft/world/gen/DebugChunkGenerator.java.patch
@@ -1,15 +1,13 @@
 --- a/net/minecraft/world/gen/DebugChunkGenerator.java
 +++ b/net/minecraft/world/gen/DebugChunkGenerator.java
-@@ -101,4 +101,12 @@
+@@ -101,4 +101,10 @@
  
        return blockstate;
     }
 +   
 +   public static void initValidStates() {
-+      field_177464_a = StreamSupport.stream(Registry.field_212618_g.spliterator(), false).flatMap((p_236067_0_) -> {
-+         return p_236067_0_.func_176194_O().func_177619_a().stream();
-+      }).collect(Collectors.toList());
-+      field_177462_b = MathHelper.func_76123_f(MathHelper.func_76129_c((float)field_177464_a.size()));
-+      field_181039_c = MathHelper.func_76123_f((float)field_177464_a.size() / (float)field_177462_b);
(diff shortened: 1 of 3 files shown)


ichttt:
This is fine in general, however it would be better if we could do this without making the methods public. A rebake method patched into the class would be a bit cleaner.

sciwhiz12:
So, the better solution would be akin to my [debug_world_fix_1 branch][firstfix]: a public method called from `BlockCallbacks` that rebakes the list and integers. Will do, then.

[firstfix]: https://github.com/sciwhiz12/MinecraftForge/tree/debug_world_fix_1

sciwhiz12:
Updated to be a patched-in public method, called when the `Block`s registry is baked.

PR #6931, opened 2020-07-07
Added func_240719_a_ (setFontId) to the side annotation stripper.

Darkhax:
This PR adds Style.setFontId to the side annotation stripper config. In 1.16 Style is designed to be immutable and the constructor is not visible. The font data exists on the server vanilla just never sets it there using this method. Use cases for calling this on the server would be for applying font to item display names or applying it to things via commands.


LexManos:
Add func_240723_c_ as it's the only other method in the class that's sided, and has no side effects. Also make sure to run checkSAS

Darkhax:
Added the new method. checkSAS had no errors or additions.

PR #6932, opened 2020-07-07
Add a way to stitch custom TextureAtlasSprite's

nikita488:
Right now it's not possible to create TextureAtlasSprite via code and stitch it to the AtlasTexture. This PR adds an easy way to do this.


diesieben07:
1.15.2 is the current LTS version. New features should go into 1.16 first before being backported.

Additionally, this goes against the data-driven approach Minecraft and Forge are continuously moving towards. Instead of specifying the custom loading entirely in code, instead we should add a way to trigger the custom loading from a metadata file for the texture, similar to how animation metadata is handled.
What I am imagining is a `json` file for a texture which can specify a loader type (a `ResourceLocation`) and that loader will the be used to load the texture instead of the normal approach.

nikita488:
I don't see how data-driven textures would be useful. If we will go with the **mcmeta** route and specify loader there, I still need texture file, otherwise, **.mcmeta** won’t be recognized. I can just create a full white (dummy) texture file and create **.mcmeta** for this file, then specify some kind of “**loader**” field in it (like we have now in models) and it will stitch this texture file to the atlas, then I will be able to deserialize my “**loader**” field from **mcmeta**, but what’s next? I just don’t see what can be done next.
My use case is to change all pixels of the texture every tick. Back in the days, I was able to override the **hasAnimationMetadata** method in **TextureAtlasSprite** to return true and then use the **updateAnimation** method to update the pixel array every tick. 
So I just need own sprite class which extends **TextureAtlasSprite**, but after **1.15.*** the constructor requires all the atlas stuff parameters (like **atlasWidth**, **atlasHeight**, **x**, **y** positions of this sprite in the atlas) so I can’t just create it and stitch easily anymore.
Also, you can see how [TwilightForest](https://github.com/TeamTwilight/twilightforest/blob/1.15.x/src/main/disabled/client/texture/GradientMappedTexture.java) used custom sprites to do fancy stuff.

KnightMiner:
Another case that has used custom atlas sprites before is Tinkers Construct. We used it to generate part material textures so if one add-ons adds a material and another a part, we have a texture for that combination. There is no way to do that through a "texture loader" as we don't know what textures are needed until we load the game with all the add-ons.

nikita488:
After thinking about it, I came to one idea:
* I can have **“loader”** field in **mcmeta** file which will be set like:
`“loader”: “mymod:my_sprite_loader”`
* I can patch **TextureAtlasSprite.Info** class to have an additional argument in the constructor, like **ISpriteLoader**;
* Inside **AtlasTexture#makeSprites** method I can deserialize my Loader from **mcmeta** file and pass it to the **TextureAtlasSprite.Info** (if **“loader”** field don’t exist in **mcmeta**, I will pass **DEFAULT** Loader).
* Inside **AtlasTexture#loadSprite** instead of creating new **TextureAtlasSprite** instance I will patch it to use my **ISpriteLoader#load** method instead and pass all the arguments that **loadSprite** has and **NativeImage** instance.

Is this will be better than I have now? I also need to create some kind of **SpriteLoaderRegistry** class where we will be registering our loaders (just like **Model Loader**'s).

diesieben07:
nikita, that is indeed what I had in mind.
But if there are even more dynamic use cases (like the one pointed out by KnightMiner) we need an even more complex system...

nikita488:
Yeah, actually I like mcmeta solution, it will work for my use case and a lot easier to maintain. But it will not work for TiCon, because it needs like thousands of dummy texture files + mcmeta files for them. So at this point, I'm open to suggestions.

KnightMiner:
At this point, I am not sure the best solution for our usecase. I think it might be possible for us to register a custom `AtlasTexture` and generate the extra sprites in that instead of attempting to inject them into the block atlas (since models can internally choose an atlas it seems), but I'm also half inclined to ditch the whole texture generation and just use vanilla tinting as fallbacks.

nikita488:
I'm not sure, but i think Blocks will only use Block atlas texture and you can't choose other than that. Also i don't think it's possible to add custom sprites to custom atlas, because vanilla only allow us to load sprites from files, so you still need files to upload them to custom atlas.

PR #6933, opened 2020-07-07
Expose the DataPackRegistries instance to the AddReloadListenerEvent

jaredlll08:
This makes it that reload listeners can use the data inside the managers that DataPackRegistries holds (in my case I need the RecipeManager specifically)

### DIFF
--- a/src/main/java/net/minecraftforge/event/AddReloadListenerEvent.java
+++ b/src/main/java/net/minecraftforge/event/AddReloadListenerEvent.java
@@ -18,8 +18,13 @@ import java.util.List;
 public class AddReloadListenerEvent extends Event
 {
    private final List<IFutureReloadListener> listeners = new ArrayList<>();
-
-   /**
+   private final DataPackRegistries dataPackRegistries;
+    
+    public AddReloadListenerEvent(DataPackRegistries dataPackRegistries) {
+        this.dataPackRegistries = dataPackRegistries;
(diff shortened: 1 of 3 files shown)


src/main/java/net/minecraftforge/event/AddReloadListenerEvent.java
```diff
@@ -18,8 +18,13 @@
 public class AddReloadListenerEvent extends Event
 {
    private final List<IFutureReloadListener> listeners = new ArrayList<>();
-
-   /**
+   private final DataPackRegistries dataPackRegistries;
+    
+    public AddReloadListenerEvent(DataPackRegistries dataPackRegistries) {
```
pupnewfster: Bracket should be on the next line.

src/main/java/net/minecraftforge/event/AddReloadListenerEvent.java
```diff
@@ -31,4 +36,8 @@ public void addListener(IFutureReloadListener listener)
    {
       return ImmutableList.copyOf(listeners);
    }
+    
+    public DataPackRegistries getDataPackRegistries() {
```
pupnewfster: Bracket should be on the next line.

Darkhax:
Yeah, this is definitely going to fix a major hole in some of my datapack stuff.

Issue #6935, opened 2020-07-07
Grindstone repairing does not work


sciwhiz12:
**Minecraft Version:** 1.16.1

**Forge Version:** 32.0.57

**Logs:** _not applicable, in-game issue_

**Steps to Reproduce:**
 1. Obtain two identically damaged diamond pickaxes (`/give @s minecraft:diamond_pickaxe{Damage:768s} 2`)
 2. Put the two items into the grindstone's interface
 3. _Observe that there is an invalid result (non-stackable item is stacked and not repaired)_
 4. Break a block with one of the diamond pickaxes _(so they are not identically damaged)_
 5. Put the two items into the grindstone's interface
 6. _Observe that there is no result_

**Description of issue:**
When using the grindstone to repair two tools of the same item, the result is bugged and not correct.

The correct result with the given steps for both 1-3 and 4-6 is a fully repaired diamond pickaxe.
The actual result from following the given steps on the given Forge version differs:
- Steps 1-3 result in a stack of the given pickaxes, with the count of 2. (also happens with enchanted pickaxes)
- Steps 4-6 result in nothing.

[Original bug report on forums _(this is one part of the report)_][bugreport]

<details>
    <summary>Screenshot of result of steps 1-3 on given Forge version</summary>
    <img src="https://user-images.githubusercontent.com/21304337/86803339-78af1400-c0a8-11ea-86a5-42d0f6d7a157.png"/>
</details>
<details>
    <summary>Screenshot of result of steps 1-3 on given Forge version, using enchanted pickaxes</summary>
    <img src="https://user-images.githubusercontent.com/21304337/86803585-b449de00-c0a8-11ea-8e2f-68c55bdd6ce2.png"/>
</details>
<details>
    <summary>Screenshot of result of steps 4-6 on given Forge version (damage values are exaggerated to visually differentiate the items)</summary>
    <img src="https://user-images.githubusercontent.com/21304337/86803488-9e3c1d80-c0a8-11ea-94ba-b1bc3dc8a96a.png"/>
</details>

[bugreport]: https://forums.minecraftforge.net/topic/88079-mczombie1991

PR #6936, opened 2020-07-07
Fix canRepair not being set true as default

### LINKED ISSUE #6934: Repairing tools does not work in crafting window
**Minecraft Version:** 1.16.1

**Forge Version:** 32.0.57

**Logs:** _not applicable, ingame behaviour issue_

**Steps to Reproduce:**
  1. Obtain two damaged diamond pickaxes (`/give @s minecraft:diamond_pickaxe{Damage:768s} 2`)
  2. Put them into the player inventory's crafting grid
  3. _Observe that there is no crafting result_

**Description of issue:**
In vanilla, two damaged tools of the same item can be repaired in the crafting window, which removes enchantments and then combines their remaining uses into a new item.

I have tested that, in vanilla 1.16.1, the given steps do give the expected result: a fully repaired diamond pickaxe. I have also tested the above steps in the given Forge version using diamond pickaxes enchanted with Efficiency 1, which also results in no crafting result.

[Original bug report on forums _(this is one part of the report)_][bugreport]

<details>
    <summary>Screenshot of result of given steps on given Forge version</summary>
    
</details>

[bugreport]: https://forums.minecraftforge.net/topic/88079-mczombie1991

sciwhiz12:
This PR closes #6934 and #6935.

`Item.Properties#canRepair` was not being set to `true` as default, so all items, regardless of whether they call `setNoRepair()` or not, were always not repairable (`canRepair == false`), leading to issues where items are being repaired. This PR simply adds the default value of `true`.

I have tested that this fixes the above linked issues using the steps provided.

### DIFF
--- a/patches/minecraft/net/minecraft/item/Item.java.patch
+++ b/patches/minecraft/net/minecraft/item/Item.java.patch
@@ -125,7 +125,7 @@
        private Rarity field_208104_e = Rarity.COMMON;
        private Food field_221541_f;
        private boolean field_234688_g_;
-+      private boolean canRepair;
++      private boolean canRepair = true;
 +      private java.util.Map<net.minecraftforge.common.ToolType, Integer> toolClasses = Maps.newHashMap();
 +      private java.util.function.Supplier<java.util.concurrent.Callable<net.minecraft.client.renderer.tileentity.ItemStackTileEntityRenderer>> ister;
  


diesieben07:
But this causes all items to be repairable by default, which is not correct.
The default in vanilla for this is `isDamageable`, so we should default to that, if its not set, right?

sciwhiz12:
`Item.Properties#canRepair` is a Forge-added variable, and is transferred to `Item#canRepair`, which is `final`. Futhermore, the only access to `Item#canRepair` is through `Item#isRepairable(ItemStack)`, which does call `Item#isDamageable()`:
```java
// Item.class, 1.16.1-32.0.57, L307-310
@Override
public boolean isRepairable(ItemStack stack) {
   return canRepair && isDamageable();
}
```

diesieben07:
> which does call Item#isDamageable():

That's the part I missed, thanks.

diesieben07 (approved):

diesieben07:
Just checked the git history, this indeed just a missed patch from the 1.16 update: https://github.com/MinecraftForge/MinecraftForge/commit/6d15febdb35b5676b1795b2965d6eb527bb336e6#diff-d0164446501bbeeb653e67a161393d00L128-R128

PR #6944, opened 2020-07-09
[1.15.2] Render local player when the renderViewEntity is not the local player

XFactHD:
# Description

# Use-cases

If requested, I will also make a PR for this for 1.16.


diesieben07:
1.15.2 is LTS, new features should go into 1.16 first before being backported.

XFactHD:
> 
> 
> 1.15.2 is LTS, new features should go into 1.16 first before being backported.

Done

TheCurle:
This PR is awaiting #7216.

XFactHD:
I don't need this anymore, the mod this was for has moved on to 1.16.

PR #6945, opened 2020-07-09
[1.16.x] Render local player when the renderViewEntity is not the local player

XFactHD:
# Description

# Use-cases


marchermans:
I would love to see this in as well.
It is a very usefull change we could use to make animated building in a 3D world possible

XFactHD:
Is there anything I need to fix before this is merged?

marchermans (approved):
This looks good.

- [x] Functionality seems to be there.
- [x] Formatting of patches seems to be correct.
- [x] Testmod and documentation exists.
- [x] Formatting of changes to none patch files seem to be okey.

XFactHD:
I will redo this PR because I just fucked it up completely...

PR #6952, opened 2020-07-11
[1.16.x] Allow command to be registered for client side execution

GotoLink:
Update of #6670 for 1.16
The old test mod was removed so it is 'new' again.

### DIFF
--- a/patches/minecraft/net/minecraft/command/Commands.java.patch
+++ b/patches/minecraft/net/minecraft/command/Commands.java.patch
@@ -25,3 +25,11 @@
           } catch (CommandException commandexception) {
              p_197059_1_.func_197021_a(commandexception.func_197003_a());
              return 0;
+@@ -276,6 +285,7 @@
+                argumentbuilder.executes((p_197053_0_) -> {
+                   return 0;
+                });
++               argumentbuilder.executes(null);// FORGE Client Commands support for lan or single-player, strips execution part, see ForgeHooksClient
+             }
+ 
+             if (argumentbuilder instanceof RequiredArgumentBuilder) {
--- a/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
+++ b/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
@@ -22,6 +22,8 @@ package net.minecraftforge.client;
 import com.google.common.collect.ImmutableList;
 import com.mojang.blaze3d.matrix.MatrixStack;
 import com.mojang.blaze3d.vertex.IVertexBuilder;
+import com.mojang.brigadier.ParseResults;
+import com.mojang.brigadier.exceptions.CommandSyntaxException;
 import net.minecraft.client.MainWindow;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.MouseHelper;
@@ -49,6 +51,8 @@ import net.minecraft.client.renderer.vertex.VertexFormatElement;
 import net.minecraft.client.renderer.vertex.VertexFormatElement.Usage;
 import net.minecraft.client.resources.I18n;
 import net.minecraft.client.settings.KeyBinding;
+import net.minecraft.command.CommandException;
+import net.minecraft.command.ISuggestionProvider;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.LivingEntity;
 import net.minecraft.entity.player.PlayerEntity;
@@ -61,6 +65,7 @@ import net.minecraft.util.Direction;
 import net.minecraft.util.Hand;
 import net.minecraft.util.MovementInput;
 import net.minecraft.util.ResourceLocation;
+import net.minecraft.util.Util;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.BlockRayTraceResult;
 import net.minecraft.util.math.EntityRayTraceResult;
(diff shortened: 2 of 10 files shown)


src/main/java/net/minecraftforge/client/event/CommandRegistryEvent.java
```diff
@@ -0,0 +1,46 @@
...
+package net.minecraftforge.client.event;
+
+import com.mojang.brigadier.CommandDispatcher;
+import net.minecraft.command.ISuggestionProvider;
+import net.minecraftforge.eventbus.api.Event;
+
+/**
+ * Fired when client {@link CommandDispatcher} has all of its commands synced from the server
+ */
+public class CommandRegistryEvent extends Event
```
ichttt, line 29: As we have a `RegisterCommandsEvent`, it might be good to call this something similar like `RegisterClientCommandsEvent` and have similar context. I would, however, not extend `RegisterCommandsEvent` so that people that listen for `RegisterCommandsEvent` don't accidentally register for the client as well

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -794,4 +795,52 @@ public static void drawItemLayered(ItemRenderer renderer, IBakedModel modelIn, I
...
+                {
+                    details.func_230529_a_(new StringTextComponent(syntax.getInput().substring(position)).func_240701_a_(TextFormatting.RED, TextFormatting.UNDERLINE));
+                }
+                details.func_230529_a_(new TranslationTextComponent("command.context.here").func_240701_a_(TextFormatting.RED, TextFormatting.ITALIC));
+                Minecraft.getInstance().player.sendMessage(new StringTextComponent("").func_230529_a_(details).func_240699_a_(TextFormatting.RED), Util.field_240973_b_);
+            }
+        }
+        catch (Exception generic)
+        {
+            StringTextComponent message = new StringTextComponent(generic.getMessage() == null ? generic.getClass().getName() : generic.getMessage());
```
ichttt, line 803: IMO any non-command exception thrown by a client command should be logged

patches/minecraft/net/minecraft/command/Commands.java.patch
```diff
@@ -15,3 +15,11 @@
              return this.field_197062_b.execute(stringreader, p_197059_1_);
           } catch (CommandException commandexception) {
              p_197059_1_.func_197021_a(commandexception.func_197003_a());
+@@ -273,6 +281,7 @@
+                argumentbuilder.executes((p_197053_0_) -> {
+                   return 0;
+                });
++               argumentbuilder.executes(null);// FORGE Client Commands support
```
ichttt: What does this do? Why is this needed for client commands to work? Please add some more description to the comment

src/main/java/net/minecraftforge/client/event/CommandRegistryEvent.java
GotoLink: The event you refer to has a very different purpose (and was proposed later than this).
I would argue it to be less confusing when the name are not too similar.

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
GotoLink: This is a paste from the vanilla Commands class error handling.
Behavior is intentionally identical with server commands.

patches/minecraft/net/minecraft/command/Commands.java.patch
GotoLink: This is explained in the commit message:
"Patch Commands to remove empty lambdas of server commands from being
sent to client (and thus execute there when they should not)."
The line right above this is the one where "empty lambdas" are set into the builder and then sent to the client by SCommandListPacket. They will then get registered by CommandDispatcher, and would be trapped by the hook of ForgeHooksClient in ChatScreen.
Without this patch, server commands are executed as client commands which do nothing since empty lambdas are always successful, effectively trapping all server commands on the client side.

src/main/java/net/minecraftforge/client/event/CommandRegistryEvent.java
ichttt: How this this event different in purpose from the RegisterCommandsEvent? Both are intended to be used to register new commands, one for client-side commands, one for server-side commands. There are fired at different times and have a completly different implementation, but to the modder they serve the same purpose for different sides or am I wrong?

patches/minecraft/net/minecraft/command/Commands.java.patch
ichttt: I somewhat understand this, but please add a brief version of this to the code comment so that when someone needs to port patches they understand why this is there as well

coehlrich:
Currently when using this, when connecting to a non forge and running a server command it doesn't get sent to the server and when running a client command the command doesn't get added to the client's sent history.

src/main/java/net/minecraftforge/client/event/CommandRegistryEvent.java
GotoLink: The RegisterCommandsEvent is there to fix registration originating from vanilla datapacks, and triggers on both client and server side. The registered command won't execute on the client and may or may not execute on the server, depending on various conditions.

While this PR is intended to be a long-term feature supported by Forge, and a new instalment of what was supported in older versions a.k.a _client-only_ 'registered **and executed**' commands, the execution is guaranteed not conditional.
And as you said, they trigger at different times so if a modder used those for a generic listener, he would not get the same behaviour with both event.

GotoLink:
Yes there is a patch in a server side class included in this PR, which exists for the sole purpose of fixing this issue.
The client sent history isn't updated for a client command because the message is never sent to the server when it is processed by the client, as intended.

GotoLink:
I have added more precise comments, proper javadoc in the client hook method.
And another patch to fix what coehlrich mentionned above: in case of vanilla server, clients can strip the empty lambdas on their own.

TheCurle:
If someone has the time to review this, i'd appreciate it.

QWERTY770:
Please remove the lable "Stale".

coehlrich:
I made an updated version of this PR: #7754 
It also makes a couple of changes.

sciwhiz12:
Hello! @GotoLink, please leave a comment whether you wish to continue this PR for 1.16. Note that there is already a 1.18 version of the feature intended by this PR at #7754, so it is your decision whether to discontinue updating this PR in favor of that or to continue with your own implementation. (We suggest the former so there is a concentration of effort in one PR rather than fragmented across multiple.) If we do not hear back from you after some time, this PR will be closed for inactivity.

sciwhiz12:

As a note, PR #7754 which implements a similar feature to this has been merged since the previous comment.

PR #6954, opened 2020-07-11
[1.16] Capability share tags 

josephtarbit:
Update of #6668 for 1.16


src/main/java/net/minecraftforge/common/extensions/IForgeItemStack.java
```diff
@@ -372,7 +374,17 @@ default ITextComponent getHighlightTip(ITextComponent displayName)
     @Nullable
     default CompoundNBT getShareTag()
     {
-        return getStack().getItem().getShareTag(getStack());
+	    CompoundNBT shareTag = getStack().getItem().getShareTag(getStack());
```
ichttt: Weird indentation here

src/main/java/net/minecraftforge/common/extensions/IForgeItemStack.java
```diff
@@ -372,7 +374,17 @@ default ITextComponent getHighlightTip(ITextComponent displayName)
...
+        CompoundNBT capabilityShareTags = this.serializeShareTag();
+        if(capabilityShareTags != null)
```
ichttt: Missing space between if and bracket here and below a few times

ichttt (commented):
Some style notes

patches/minecraft/net/minecraft/item/ItemStack.java.patch
```diff
@@ -243,3 +243,18 @@
     public SoundEvent func_226629_F_() {
        return this.func_77973_b().func_225520_U__();
     }
+@@ -946,4 +999,14 @@
+    public SoundEvent func_226630_G_() {
+       return this.func_77973_b().func_225519_S__();
+    }
++
++   @Override
++   public CompoundNBT serializeShareTag() {
```
ichttt, line 252: This can be moved to the extension interface

src/main/java/net/minecraftforge/common/extensions/IForgeItemStack.java
```diff
@@ -384,6 +396,15 @@ default CompoundNBT getShareTag()
      */
     default void readShareTag(@Nullable CompoundNBT nbt)
     {
+        if (nbt != null && nbt.contains("ForgeCaps", Constants.NBT.TAG_COMPOUND))
+        {
+            this.deserializeShareTag(nbt.getCompound("ForgeCaps"));
+            nbt.remove("ForgeCaps");
+            if (nbt.size() == 0)
+            {
+                return;
```
ichttt, line 405: Why does this early exist if no forge caps are present?

ichttt (commented):
Some general notes:
First, this looks like this is a special case for ItemStack capabilities, can we make this so it is usable for Entites/TileEntites as well?
Makes the new interface really sense, or would it be better to a a new paramter like `boolean network` to the already existing methods that deal with serialization?

src/main/java/net/minecraftforge/common/extensions/IForgeItemStack.java
josephtarbit: It exists early if the NBT compound is empty after reading the forge caps.

patches/minecraft/net/minecraft/item/ItemStack.java.patch
josephtarbit: If I moved this to the extension interface I'd have to make the capability provider serialize method public which it shouldn't be as it's an internal method.

josephtarbit:
I've been thinking about this and I'm now considering adding a bit more functionality to the API. What do people think of having a system that automatically syncs 'dirty' capabilities - it'd be fairly easy to implement as vanilla already does this for its own data in `ItemStack`s, `Entity`s, `TileEntity`s etc. The reason I'm considering this is because having a simple 'share tag' system is only really applicable to `ItemStack`s. A fuller API would be much more universal, easier and useful to modders. Specifically what I intend is an interface that `ICapabilityProvider`s implement which provides a method to check if the provider is dirty and methods to read and write to a `PacketBuffer` with a flag parameter specifying whether to only write dirty data or a full write. This would also allow easy integration with vanilla's `DataManager` which could them be used inside capabilities.

josephtarbit:
Any insights on this?

ichttt:
I think that's better then something specific for ItemStacks, but I would not call them dirty, but rather `synced` or something else that implies the only difference is S->C sync

josephtarbit:
Closing until I have the time to finish

PR #6960, opened 2020-07-12
Add dataprovider for Global Loot Modifiers

Cyborgmas:
Added a provider to manage GLM datagen. Also incorporated it in the test mod and verified that the output was the same as the pre-existing files.

### DIFF
--- a/src/main/java/net/minecraftforge/common/data/GlobalLootModifierProvider.java
+++ b/src/main/java/net/minecraftforge/common/data/GlobalLootModifierProvider.java
@@ -31,10 +31,8 @@ import net.minecraft.loot.conditions.ILootCondition;
 import net.minecraft.util.ResourceLocation;
 import net.minecraft.util.Tuple;
 import net.minecraftforge.common.loot.GlobalLootModifierSerializer;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
+import net.minecraftforge.common.loot.LootModifier;
 
-import javax.annotation.Nullable;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


src/main/java/net/minecraftforge/common/data/GlobalLootModifierProvider.java
```diff
@@ -0,0 +1,106 @@
...
+   public void addModifier(String modifier, GlobalLootModifierSerializer<?> serializer, List<ILootCondition> conditions, @Nullable Consumer<JsonObject> extraProperties)
+   {
+      this.modifiers.put(modifier, new Tuple<>(serializer, conditions));
+      this.extraProperties.put(modifier, extraProperties);
+   }
+
+   @Override
+   public String getName()
+   {
+      return "Global Loot Modifiers" + modid;
```
MelanX: You should add a space after "Modifiers" because "modid" would stick on it.

src/main/java/net/minecraftforge/common/data/GlobalLootModifierProvider.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+
+public abstract class GlobalLootModifierProvider implements IDataProvider
+{
+   private static final Logger LOGGER = LogManager.getLogger();
```
ichttt: In Forge code we use 4 spaces for indentation, not 3

ichttt (commented):
Indentation is wrong, otherwise looking good to me

Cyborgmas:
Added documentation, as requested on discord.

ichttt:
In addition to the changes you just did, it would be better if you could write an actual serializer instead of just handing down a json serailizer in addModifier

Cyborgmas:
If you mean instead of passing of having a json consumer making something like GlobalLootModifierSerializer#write I avoided that on purpose.
Because the keys can be anything the modder chooses and because if I just passed a JsonObject in #write, then there would be no way to gen two modifiers using the same serializer but different values.

Cyborgmas:
So, @ichttt I was a bit dumb when i answered, hopefully this is better.

PR #6961, opened 2020-07-12
[1.16.x] Fix ChunkDataEvents using different data tags

### LINKED ISSUE #6957: [1.15.2] ChunkDataEvent.{Load/Save} use different data tags
**Minecraft Version:** 1.15.2

**Forge Version:** 31.2.27

**Description of issue:**
When attempting to use the chunk data events, I noticed that my data was always lost. After some debugging I found that the Save event gets called for the "root" chunk tag, which has the two tags shown in step 2. When the Load event gets called however, the "Level" tag (present in the root tag) is used for the event, shown by the output described in step 3.

In net.minecraft.world.chunk.storage.ChunkSerializer#read, line 63, the following can be found: 
`CompoundNBT compoundnbt = compound.getCompound("Level");`
Forge then uses the "compoundnbt" for the ChunkDataEvent.Load, instead of "compound"

**Steps to Reproduce:**
 1. Create an eventlistener listening to ChunkDataEvent.Save and ChunkDataEvent.Load, print the keys inside the data:

```
@SubscribeEvent
public void chunkLoadFromDisk(ChunkDataEvent.Save event) {
        System.out.println(event.getData().keySet());
}

@SubscribeEvent
public void chunkLoadFromDisk(ChunkDataEvent.Load event) {
        System.out.println(event.getData().keySet());
}
```

 2. When saving the output will be: `[Level, DataVersion]`
 3. When loading, the output will be: `[Status, zPos, LastUpdate, Biomes, InhabitedTime, xPos, Heightmaps, TileEntities, Entities, isLightOn, TileTicks, Sections, PostProcessing, Structures, LiquidTicks]`

sciwhiz12:
_This PR fixes #6957, for 1.16.x._

The linked issue's author summed up the problem nicely: `ChunkDataEvent.Save` and `ChunkDataEvent.Load` gave different data tags; `Save` gives the whole chunk tag, both the `Level` tag and the `DataVersion`, while `Load` gives only the `Level` subtag.

This PR does one thing: modifies the patch to `ChunkManager` so the two calls to post the `Load` event use the root tag instead of the `Level` subtag.

My previous commit was changing the `Save` event to conform to the `Load` event, but as [stated by LexManos][lexcomment], modders could use the `ChunkDataEvent`s to store custom chunk data.

[lexcomment]: https://github.com/MinecraftForge/MinecraftForge/pull/6961#issuecomment-657744142

### DIFF
--- a/patches/minecraft/net/minecraft/world/chunk/storage/ChunkSerializer.java.patch
+++ b/patches/minecraft/net/minecraft/world/chunk/storage/ChunkSerializer.java.patch
@@ -21,7 +21,7 @@
        }
  
        if (chunkstatus$type == ChunkStatus.Type.LEVELCHUNK) {
-+         net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.world.ChunkDataEvent.Load(ichunk, compoundnbt, chunkstatus$type));
++         net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.world.ChunkDataEvent.Load(ichunk, p_222656_4_, chunkstatus$type));
           return new ChunkPrimerWrapper((Chunk)ichunk);
        } else {
           ChunkPrimer chunkprimer1 = (ChunkPrimer)ichunk;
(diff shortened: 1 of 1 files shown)


LexManos:
Passing in the extra version info is entirely intentional.
The theory was that modders could add whatever stuff outside the vanilla data it wanted to.
Honestly the other end should be the one to change, pass in the full NBT to the read function.

sciwhiz12:
Whoops, accidentally closed while rebasing and force-pushing new branch. Will open when I've made the new commit

PR #6962, opened 2020-07-12
[1.15.2] Fix ChunkDataEvents using different data tags

### LINKED ISSUE #6957: [1.15.2] ChunkDataEvent.{Load/Save} use different data tags

sciwhiz12:
_This PR fixes #6957._

This PR is the same as #6961, but for 1.15.2. See that issue and PR for details.

### DIFF
--- a/patches/minecraft/net/minecraft/world/server/ChunkManager.java.patch
+++ b/patches/minecraft/net/minecraft/world/server/ChunkManager.java.patch
@@ -28,7 +28,7 @@
  
              this.field_219255_i.func_217381_Z().func_230035_c_("chunkSave");
              CompoundNBT compoundnbt1 = ChunkSerializer.func_222645_a(this.field_219255_i, p_219229_1_);
-+            net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.world.ChunkDataEvent.Save(p_219229_1_, p_219229_1_.getWorldForge() != null ? p_219229_1_.getWorldForge() : this.field_219255_i, compoundnbt1));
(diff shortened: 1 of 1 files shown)


pupnewfster:
Honestly to an extent I am unsure if this is "too late" to include into 1.15 or not, just from the point of view that it is *technically* a "breaking change" in that any mods that are currently manually getting the level tag because different tags are passed, will no longer have it be pointing at the correct direction.

sciwhiz12:
I do agree that this is a breaking change because it causes the incompatibility with any existing mods. However, the linked issue is made for 1.15, and not for 1.16, so I thought to at least make a PR for this. Do you recommend that this PR be closed because of this being a breaking change? If so, then the linked issue should be closed when the 1.16 PR is merged.

LexManos:
This is a breaking change so sadly can't be accepted for 1.15.

PR #6963, opened 2020-07-12
[1.16.x] Make ArgumentList a public class

alcatrazEscapee:
This is a active-development-non-LTS-version of #6918 since it has to be merged into 1.16.x before the possibility of getting merged for 1.15.x. The same reasoning still applies from that PR.


LexManos:
https://github.com/MinecraftForge/MinecraftForge/pull/6918#issuecomment-657797184

PR #6964, opened 2020-07-12
Introduce a JSON data structure combiner

gigaherz:
This will allow processing the json files as they load, before they are passed into the corresponding deserializer.

## How to use

This feature is available on specific supported json files (eg, loot tables).

Json objects can contain a "_forge_combine" key, which should itself be a json object. 

By default, all json files start in "overwrite" mode, which returns the file as-is, maintaining the default vanilla behaviour. 

Certain alternative modes are available depending on the element type.

### Objects:

- Combine: The two objects are combined, processing the child elements recursively.

### Arrays:

- Combine: The two arrays are combined, processing the child elements recursively. When in combine mode, array elements can have special modes to choose what to replace, or where to insert.
- Zip: The two arrays are combined pairwise, combining each pair of elements into a new element.

### Child elements within arrays:

- Append (default): The element is added to the end of the list.
- Overwrite: The element with either the specified index, or the specified search pattern, gets replaced.
- Combine: Same as replace, but combines the existing value instead of overwriting.
- Insert: The element is inserted at the position with the specified index.

### How it's processed

The process begins at the root object, which defaults to "overwrite". If this object is set to "combine" instead, the merger will iterate through the key sets of both the original and the new, combining values for keys that exist on both objects. Children of combined objects inherit the combine mode, but they can choose to overwrite instead, by specifying this explicitly.

For arrays, the combine operation adds elements to the array, without combining them. However those child elements can specify their own mode, in which case the element will either be inserted at a chosen location, will replace (overwrite) and existing value, or will be combined with an existing value. This behaviour is guided either by index, or by search pattern.

For all other data types (primitives, null), overwrite and combine act the same: the element is returned as-is.

### Controlling the insertion behaviour for non-object elements

Since only json objects can contain the "_forge_combine" key, another feature is available: any object which has the "_forge_combine" key, and inside has a "value" key, the contents of this value key are used, instead of the object. This allows inserting/replacing arrays, strings, or other primitives, contained within other arrays, or controlling the mode of values in individual keys within an object.

That is, the following object: 

```json
{
  "_forge_combine": { "value": 1 }
}
```

is equivalent to the primitive `1`.

## Todo

- [ ] Tie into dimension json import logic, which isn't based on datapack reload listeners.
- [x] Add support for a DELETE mode, which would work within children of objects and arrays.

### DIFF
--- a/src/main/java/net/minecraftforge/common/util/JsonMerger.java
+++ b/src/main/java/net/minecraftforge/common/util/JsonMerger.java
@@ -253,7 +253,7 @@ public class JsonMerger
                 if (settings.has("index"))
                 {
                     if (mode == MergeMode.APPEND)
-                        throw new JsonSyntaxException("Index is noy available in append mode.");
+                        throw new JsonSyntaxException("Index is not available in append mode.");
                     index = settings.get("index").getAsInt();
                 }
 
@@ -262,7 +262,7 @@ public class JsonMerger
                     if (index != null)
                         throw new JsonSyntaxException("Array replace option 'find' can not be used at the same time as 'index'.");
                     if (mode == MergeMode.APPEND)
-                        throw new JsonSyntaxException("Find is noy available in append mode.");
+                        throw new JsonSyntaxException("Find is not available in append mode.");
                     find = settings.get("find");
                 }
 
--- a/src/main/java/net/minecraftforge/common/util/JsonMerger.java
+++ b/src/main/java/net/minecraftforge/common/util/JsonMerger.java
@@ -123,6 +123,7 @@ public class JsonMerger
         JsonElement find = null;
         boolean silent = false;
         boolean exact = false;
+        boolean after = false;
         if (second.isJsonObject())
         {
             JsonObject secondObj = second.getAsJsonObject();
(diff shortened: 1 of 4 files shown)


src/main/java/net/minecraftforge/common/util/JsonMerger.java
```diff
@@ -0,0 +1,468 @@
...
+            a.add(combineJsonElements(first.get(i), second.get(i), MergeMode.COMBINE));
+        }
+        return null;
+    }
+
+    private static JsonElement concatenateArrays(JsonArray first, JsonArray second)
+    {
+        List<JsonElement> elements = Lists.newArrayList();
+        for(int i=0;i<first.size();i++)
+            elements.add(first.get(i));
```
TeamSpen210, line 214: Would it be worth it to add a check to these elements to ensure none of them accidentally contain `_forge_combine`? It shouldn't happen since I assume this will always be a vanilla file, but if used between two mods or something it could have some benefits.

gigaherz: I used `"_forge_combine"` specifically to not have to worry about vanilla content using that key. While I could filter out those keys, I'm not sure it's worth the trouble.

src/main/java/net/minecraftforge/common/util/JsonMerger.java
```diff
@@ -242,92 +252,93 @@ private static void processJsonArrayElement(List<JsonElement> parent, JsonElemen
 
                 if (settings.has("index"))
                 {
+                    if (mode == MergeMode.APPEND)
+                        throw new JsonSyntaxException("Index is noy available in append mode.");
```
Iuhegi: Not important, but there's a typo in the exception message.
```suggestion
                        throw new JsonSyntaxException("Index is not available in append mode.");
```

src/main/java/net/minecraftforge/common/util/JsonMerger.java
```diff
@@ -242,92 +252,93 @@ private static void processJsonArrayElement(List<JsonElement> parent, JsonElemen
...
                     index = settings.get("index").getAsInt();
                 }
 
                 if (settings.has("find"))
                 {
                     if (index != null)
                         throw new JsonSyntaxException("Array replace option 'find' can not be used at the same time as 'index'.");
+                    if (mode == MergeMode.APPEND)
+                        throw new JsonSyntaxException("Find is noy available in append mode.");
```
Iuhegi: Same typo here
```suggestion
                        throw new JsonSyntaxException("Find is not available in append mode.");
```

PR #6967, opened 2020-07-13
[1.16.x] Use default config values if a parsing exception is thrown

sciwhiz12:
Currently, if a config file is corrupted that `nightconfig` cannot parse it, it crashes the game with `ParsingException: Not enough data available`. This has an easy solution: delete the offending config file. However, it is easier to implement a fix such that if a config file cannot be parsed, then the default config values are used.

This PR does the following:
* Modifies `FMLConfig` so any `ParsingException`s are caught, and an attempt is made to delete the corrupted `fml.toml` file and then load the config. The default config from `defaultfmlconfig.toml` will then be copied and saved by the save operation. If the delete fails, a message is logged, and the successive config load will most likely error.
* Modifies `ConfigTracker` so any `ParsingException`s are caught, and the default values are saved to disk, overwriting the corrupted config file.

Attached is the [corrupted config file][corrupted] that I used to test the PR, packaged in a ZIP archive.

[corrupted]: https://github.com/MinecraftForge/MinecraftForge/files/4913991/corrupted.zip

### DIFF
--- a/src/fmllauncher/java/net/minecraftforge/fml/loading/FMLConfig.java
+++ b/src/fmllauncher/java/net/minecraftforge/fml/loading/FMLConfig.java
@@ -21,11 +21,14 @@ package net.minecraftforge.fml.loading;
 
 import com.electronwill.nightconfig.core.ConfigSpec;
 import com.electronwill.nightconfig.core.file.CommentedFileConfig;
+import com.electronwill.nightconfig.core.io.ParsingException;
 import com.electronwill.nightconfig.core.io.WritingMode;
 import net.minecraftforge.api.distmarker.Dist;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


LexManos:
The problem with that is that we intentionally do not overwrite the entire config file when it errors so that data isn't lost and users can go in and fix whatever they broke in the config.
Yes this results in some tech support posts about the same thing in the forums.
We could probably use a better error message showing the exact file that is erroring.
But deleting data is not an option.

The other side of this is we need to understand what causes the 'corruption'. Is it user error? Are they typing toml wrong? Or is something screwing up during saving?

sciwhiz12:
From what I tested, `ParsingException`s are thrown for any error in parsing the document, whether it be corrupted ("Not enough data available") or a unexpected character due to user error or more corruption. So, as you've said, my current PR would overwrite the corrupted config file, and any user-entered values will be erased.

I have thought of an alternate solution: if a config file is corrupted, then it is moved with a suffix (`.bak` or `_old`) and a new config file is written, so the game can load properly and users can manually correct settings if needed. A screen could also be shown to the user, informing them that corrupted configs were found and backed up. (Maybe it can be part of a config GUI impl.?)
Another alternative is to catch `ParsingException`s and wrap them around another, better named exception, like `ConfigLoadingException: Config file {} for modid {} failed to load`, so users can self-diagnose the issue.

sciwhiz12:
After thinking on this a while, I think the best solution (and the one I'm going to re-PR for) will be the one above: wrapping the `ParsingException`s into an exception like `ConfigLoadingException` with an informative message with the modid and file.

There's two places where this can be implemented: the initial config load from `ConfigTracker` which immediately crashes, and the automated config reload from `ConfigWatch` which logs but ignores the exception.

This doesn't overwrite user configs, and helps users easily diagnose the problem.

PR #6969, opened 2020-07-13
Added an event for registering commands. closes #6968

### LINKED ISSUE #6968: Commands should no longer be registered in FMLServerStartingEvent
**Minecraft Version:** 1.16.1

**Forge Version:** 32.0.63

**Logs:** n/a

**Steps to Reproduce:**
 1. Register a command using `FMLServerStartingEvent#getCommandDispatcher`
 2. Go into game
 3. run `/reload`
 4. try run the command

Commands are now handled by DataPackRegistries, so when a reload occurs, it scraps all the previous commands and re-registers them from the DPR Commands listener, this means that when you register a command in the FMLServerStartingEvent event, the command will only work until a `/reload` happens.

I'm not sure the best way to approach this, but I feel removing `FMLServerStartingEvent#getCommandDispatcher()` would be a good start and changing the javadoc to match, since currently the javadoc is:

```
 * This event allows for customizations of the server, such as loading custom commands, perhaps customizing recipes or
 * other activities.
```
but neither of those given examples actually work in the event anymore.

jaredlll08:


patches/minecraft/net/minecraft/command/Commands.java.patch
```diff
@@ -1,6 +1,15 @@
 --- a/net/minecraft/command/Commands.java
 +++ b/net/minecraft/command/Commands.java
-@@ -202,6 +202,14 @@
+@@ -183,6 +183,8 @@
+       if (p_i232148_1_.field_237219_d_) {
+          PublishCommand.func_198581_a(this.field_197062_b);
+       }
++
```
pupnewfster: Extraneous white space being added the line above the line you are adding.

pupnewfster:
Also if I recall the conversation that was had on discord, shouldn't this PR also be removing the `FMLServerStartingEvent#getCommandDispatcher` and also adjusting the JavaDocs on that class?

Darkhax:
This looks good to me.

src/main/java/net/minecraftforge/fml/event/server/FMLServerStartingEvent.java
```diff
@@ -19,14 +19,13 @@
...
-import net.minecraft.command.CommandSource;
 import net.minecraft.server.MinecraftServer;
 
 /**
  * Called after {@link FMLServerAboutToStartEvent} and before {@link FMLServerStartedEvent}.
- * This event allows for customizations of the server, such as loading custom commands, perhaps customizing recipes or
- * other activities.
+ * This event allows for customizations of the server.
+ *
+ * If you need to add commands or change recipes, see {@link net.minecraftforge.event.RegisterCommandsEvent} and {@link net.minecraftforge.event.AddReloadListenerEvent} respectively.
```
LexManos: Telling people to change recipes in code is a non-starter.

jaredlll08: Only mentioned recipes since that is what was initially there. 

Updated it though.

Issue #6973, opened 2020-07-15
Commands are broken/unreadable


Darkere:
**Minecraft Version:** {Minecraft version}
1.16.1
**Forge Version:** {Forge version. *Version number, not latest/rb*}
32.0.67
**Logs:** {Link(s) to GitHub Gist with full latest.log and/or crash report}
https://gist.github.com/Darkere/6288e280bd22826fdf4f07f9708a52e6
**Steps to Reproduce:**
 1. run commands

**Description of issue:**
/forge tps is unreadable
/forge dimensions creates an error
/forge mods has broken characters
/forge track crashes (see crashlog)

Darkere:
Examples: 
/forge tps
![image](https://user-images.githubusercontent.com/4283717/87608232-9eb76280-c6ff-11ea-8189-956561d6dfd6.png)

/forge dimensions
![image](https://user-images.githubusercontent.com/4283717/87608251-ae36ab80-c6ff-11ea-97ac-2dc601d38dd3.png)

/forge mods (only in a dev environment)
![image](https://user-images.githubusercontent.com/4283717/87608292-cd353d80-c6ff-11ea-8847-b5b08755864c.png)


I have not been able to reproduce the crash. Entirely unsure how it happened.

PR #6976, opened 2020-07-15
Fixed commands not being sent to the client on reload close #6974

### LINKED ISSUE #6974: New Commands registered during reload don't get suggested
**Minecraft Version:** {Minecraft version}
1.16.1
**Forge Version:** {Forge version. *Version number, not latest/rb*}
32.0.67

**Steps to Reproduce:**
 1. Register a new command during /reload (issue is not present if the command exists before this)
 2. The new command works but is not recognized as option for autocomplete

**Description of issue:**
The command works but it does not get recognized properly, showing this error: 

image

Paging @jaredlll08 who made the PR for it.

jaredlll08:


jaredlll08:
This doesn't fix the modlist part of https://github.com/MinecraftForge/MinecraftForge/issues/6973, no idea what is going on with the formatting.

I myself would just change the character, but I tested with `+` and `*` and neither looked that great

PR #6977, opened 2020-07-16
[1.16] Re-add patch for PlayerSetSpawnEvent

diesieben07:
Re-adds patch for PlayerSetSpawnEvent, also fixes the JavaDoc and include the Dimension info in the event.


Mysterious-Dev (approved):
It's look good

PR #6346, opened 2019-12-03
Cull particles outside of view frustum.

Unnoen:
This is a relatively simple patch that prevents particles we can't see from rendering.

https://puu.sh/ELfhJ/55eba95262.png - Control (Looking at particles)
https://puu.sh/ELfj1/41fa8559de.png - Current (Particles still rendered)
https://puu.sh/ELfjI/5622df57bb.png - Changed (Particles not rendered)

We're just passing icamera to the renderParticles method, where we check if the bounding box of the particle is inside it.

As you can see in the screenshots, when we aren't rendering the particles we get a decent FPS boost. This will change from PC to PC, but will be a free gain nonetheless.

PR #6981, opened 2020-07-17
Add particle culling

ichttt:
This adds culling for particles, and is a refresh of #6346
Before this PR:
java_7D1BxVYfLV
java_jwMO5HBQ8F
java_Hap9cqt0Lg
With this PR:
java_tmqbdscj1P
java_Ivy2k95mD1
java_xfo2jmhB6L
You can see that the FPS drops a little bit when all particles that are present in the world are on the screen. However, these drops are really minor, and this is a worst-case scenario where all particles are visible. When only roughly1/4th of the particles is visible, the performance improvement of this PR is really visible. This also improves underwater performance by a lot, cause there are many underwater particles that are not on the screen.
This PR also allows mods to opt out of particle culling, in case their particles rendering can't match their bounding box


Tfarcenim:
This is going to be great for the new nether particles.

Mysterious-Dev (approved):

BedrockLegends:
Cant wait for this to be added to forge looks sweet

PR #6982, opened 2020-07-17
[1.16] Add data generators for sounds.json

TheSilkMiner:
### Abstract

This is the 1.16.1 port of PR #6746, whose goal is to provide a data generator for the `sounds.json` file, as previously requested in #6523 ([comment](https://github.com/MinecraftForge/MinecraftForge/pull/6523#discussion_r425950102)).
All restrictions in place have been taken directly from vanilla's `SoundListSerializer`, which is used to convert this file into an actual list of sounds.

To account for sounds being downloaded separately from the game, a modification of the arguments needed for data generation and therefore the main entry point for user dev (`LaunchTesting`) has been changed to accommodate the need for additional parameters. This has been done as a completely separate commit from the rest, so that it can be easily dropped in case the changes are deemed too invasive.


src/main/java/net/minecraftforge/client/sounds/generators/SoundDefinition.java
```diff
@@ -0,0 +1,449 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.client.sounds.generators;
```
ichttt: Please move this and all related classes to the data gen package.

src/generated_test/resources/assets/data_gen_test/sounds.json
```diff
@@ -0,0 +1,125 @@
+{
+  "ambient.underwater.loop.additions": {
+    "sounds": [
+      "minecraft:ambient/underwater/additions/bubbles1",
```
ichttt: it would be nice if this could match the vanilla file, so if you could auto-remove the minecraft prefix if that would be great!

ichttt (commented):
In addition to the comments on the code, It would be nice if we could setup the asset index to be sent into data gens. If that's too much work though it's fine as well

src/main/java/net/minecraftforge/client/sounds/generators/SoundDefinition.java
TheSilkMiner: To make sure, do you refer to `net.minecraftforge.common.data` or is there another package?

ichttt: Yes that package should be fine

TheSilkMiner:
The asset index is now being sent into data-gens as per request. Since these changes required some tweaks in how the data-gen gets loaded and how the run gets created, the whole set of changes has been placed into a separate commit, so that it can be dropped if needed.

TheSilkMiner:
Rebased on top of commit ce6c664acdeae7635ff5d86c5a270792e6c3f99b

patches/minecraft/net/minecraft/data/Main.java.patch
```diff
@@ -26,7 +28,7 @@
 +         Collection<Path> existingPacks = optionset.valuesOf(existing).stream().map(Paths::get).collect(Collectors.toList());
 +         java.util.Set<String> mods = new java.util.HashSet<>(optionset.valuesOf(mod));
 +         boolean isFlat = mods.isEmpty() || optionset.has(flat);
-+         net.minecraftforge.fml.DatagenModLoader.begin(mods, path, inputs, existingPacks, flag2, flag1, flag3, flag4, flag5, isFlat);
++         net.minecraftforge.fml.DatagenModLoader.begin(mods, path, inputs, existingPacks, optionset.valueOf(assetIndex), optionset.valueOf(assetsDir).toPath(), flag2, flag1, flag3, flag4, flag5, isFlat);
```
LexManos: This will NPE if someone doesn't specify the assetsDir, its not marked as required, and it shouldn't be. So you can't do .toPath directly. You need a null check

src/main/java/net/minecraftforge/common/data/ExistingFileHelper.java
```diff
@@ -48,17 +52,24 @@
     private final SimpleReloadableResourceManager clientResources, serverData;
     private final boolean enable;
 
-    public ExistingFileHelper(Collection<Path> existingPacks, boolean enable) {
+    public ExistingFileHelper(Collection<Path> existingPacks, @Nullable final String assetIndex, @Nullable final File assetsDir, boolean enable) {
```
LexManos: What is your facination with adding arguments to the middle of the argument list?
New arguments go on the end, and ideally keep the old method bouncing to the new with default values.

TheCurle:
I have not seen any discussion about this recently. Can someone check over this for me?

src/test/java/net/minecraftforge/debug/DataGeneratorTest.java
```diff
@@ -201,6 +205,83 @@ protected void registerRecipes(Consumer<IFinishedRecipe> consumer)
...
+            super(generator, MODID, helper);
+        }
+
+        @Override
+        public void registerSounds()
+        {
+            // Taken from vanilla's 'sounds.json'. These are supposed to be 1-to-1 copies, but currently
+            // automatic testing of those copies cannot be performed due to limitations in the data-gen
+            // system, where the additional data pack that vanilla loads from the launcher's assets/
+            // directory isn't actually loaded when running data-gen.
```
ichttt: Don't we have that set up now? We should be able to compare the generated file with the original file

src/main/java/net/minecraftforge/fml/DatagenModLoader.java
```diff
@@ -39,13 +40,13 @@ public static boolean isRunningDataGen() {
         return runningDataGen;
     }
 
-    public static void begin(final Set<String> mods, final Path path, final java.util.Collection<Path> inputs, Collection<Path> existingPacks, final boolean serverGenerators, final boolean clientGenerators, final boolean devToolGenerators, final boolean reportsGenerator, final boolean structureValidator, final boolean flat) {
+    public static void begin(final Set<String> mods, final Path path, final Collection<Path> inputs, Collection<Path> existingPacks, final String assetIndex, final File assetsDir, final boolean serverGenerators, final boolean clientGenerators, final boolean devToolGenerators, final boolean reportsGenerator, final boolean structureValidator, final boolean flat) {
```
ichttt: Although run from datagen, this is a binary break. Please just leave the old contructor, and deprecate it. Also, please add the new parameters to the end of the parameter list

ichttt (changes requested):
Sorry that it took so long! I completly missed that PR...
Please rebase and take a look at my comments, other than that it should be good.

ichttt (approved):
Seems good now. Thanks!

src/main/java/net/minecraftforge/common/data/SoundDefinitionsProvider.java
```diff
@@ -0,0 +1,270 @@
...
+    public String getName()
+    {
+        return "Sound Definitions";
+    }
+
+    // Quick helpers
+    /**
+     * @see SoundDefinition#definition()
+     */
+    protected static SoundDefinition definition()
```
tterrag1098, line 99: It seems a bit unnecessary to define bouncers for all these methods when they could just be defined in this class entirely. What purpose does defining them inside the implementation serve?

tterrag1098 (changes requested):
This looks good aside from my one note. Kudos on writing a proper test.

src/main/java/net/minecraftforge/common/data/SoundDefinitionsProvider.java
TheSilkMiner: I defined the factory methods in the implementation classes so that everything related to them would be in the same place, and to keep the constructors private. These bouncer methods exist mainly as a convenience, so that people wouldn't need to specify the class every time they want to create a new sound or a new definition, without using `static` imports.

tterrag1098: Why do the constructors need to be private? They aren't for most other data builder classes.

TheSilkMiner: I consider constructors as an implementation detail and, as such, it doesn't need to be exposed. If this is not the desired convention, though, I can definitely change it. I based myself off the change that had been made to `DeferredRegister`, which had seen its main constructor's visibility reduced to `private` for the same reason.

If the constructor were to be made public, though, would all the other factory methods be removed, converted into constructors, or something else?

tterrag1098: You are correct in general, and maybe I'm in the minority here, but I feel like when you have to start duplicating methods in two different places just to make your design work, that design has failed. Sometimes you have to compromise on what is theoretically right vs what is easier to write and maintain. Mostly these data builder classes will never be directly instantiated, even if the constructors are public. If we need to change the ctor to a static factory later, it can be done with little API interruption between versions. If you want to leave a *single* static factory on the class and just move all the overloads out, that would also be fine.

TheSilkMiner: Fair enough; will do.

src/main/java/net/minecraftforge/client/model/generators/ModelProvider.java
```diff
@@ -357,7 +356,7 @@ public T nested()
         ret.assertExistence();
         return ret;
     }
-    
+
```
TheCurle: What's with these whitespace changes?

src/main/java/net/minecraftforge/common/data/ExistingFileHelper.java
```diff
@@ -218,7 +233,7 @@ public void trackGenerated(ResourceLocation loc, IResourceType type) {
      * <p>
      * This represents a <em>promise</em> to generate the file later, since other
      * datagen may rely on this file existing.
-     * 
+     *
```
TheCurle: w-w-w-whitespace!

TheCurle (commented):
Minor nitpicks.

I'm, again, not familiar enough with the datagen system to weigh in on this, but it seems like this will just be a gigantic nightmare for maintenance in the future, especially with the changes to the sound system that 1.17 will be bringing.

Maybe someone else can weigh in on that, but for the mean time I need someone that knows this system to look at this. Yes, that includes you, if you're reading this.

marchermans (approved):

tterrag1098 (approved):

PR #6985, opened 2020-07-17
[1.16] Rework RenderWorldLastEvent

Nightenom:
The old way became less programmer-friendly as of recent rendering updates in vanilla code.
The new way adds events before finishing of important render buffers and creates an event for each rendered block layer. The old RenderWorldLastEvent was moved to the end of World rendering method.

I'm not sure if this PR needs a test mod since it's only event change.
I have not carried over profiler section calls because of patch lengths. If you think it's better to separate forge events from vanilla profiling, I can add them into client forge hooks.

`gradlew build` was complaining about missing license headers so I've added them


src/main/java/net/minecraftforge/client/event/RenderWorldEvent.java
```diff
@@ -0,0 +1,157 @@
...
+    {
+        return mStack;
+    }
+
+    public float getPartialTicks()
+    {
+        return partialTicks;
+    }
+
+    public ClippingHelper getClippinghelper()
```
ichttt: should be `getClippingHelper`

src/main/java/net/minecraftforge/client/event/RenderWorldEvent.java
```diff
@@ -0,0 +1,157 @@
...
+    private final ActiveRenderInfo activeRenderInfoIn;
+    private final RenderTypeBuffers renderTypeBuffers;
+
+    private RenderWorldEvent(WorldRenderer context, MatrixStack mStack, float partialTicks, ClippingHelper clippinghelper, ActiveRenderInfo activeRenderInfoIn, RenderTypeBuffers renderTypeBuffers)
+    {
+        this.context = context;
+        this.mStack = mStack;
+        this.partialTicks = partialTicks;
+        this.clippinghelper = clippinghelper;
+        this.activeRenderInfoIn = activeRenderInfoIn;
```
ichttt: just use `activeRenderInfo` instead of `activeRenderInfoIn`

src/main/java/net/minecraftforge/client/event/RenderWorldEvent.java
```diff
@@ -0,0 +1,157 @@
...
+    {
+        return partialTicks;
+    }
+
+    public ClippingHelper getClippinghelper()
+    {
+        return clippinghelper;
+    }
+
+    public ActiveRenderInfo getActiveRenderInfoIn()
```
ichttt: Same here and a few time below, remove the `In` suffix

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -138,9 +139,37 @@ public static boolean onDrawBlockHighlight(WorldRenderer context, ActiveRenderIn
         return MinecraftForge.EVENT_BUS.post(new DrawHighlightEvent(context, info, target, partialTicks, matrix, buffers));
     }
 
-    public static void dispatchRenderLast(WorldRenderer context, MatrixStack mat, float partialTicks, Matrix4f projectionMatrix, long finishTimeNano)
+    private static float partialTicksHolder;
+    private static ClippingHelper clippinghelperHolder;
+    private static ActiveRenderInfo activeRenderInfoInHolder;
```
ichttt: This seems weird, do we really need this context on thw WorldBlockLayerEvent?

ichttt (commented):
Just a basic code-style review, I'm not familiar enough with the actual event to review the actual code changes.

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
Nightenom: ClippingHelper and partialTicks are questionable, you definitely need ActiveRenderInfo for positioning - might make it event special and pass in d0-d2 from vanilla only (since that's usually the only thing you need)

Nightenom:
Whoops, will recreate...

diesieben07:
Please do not create a new PR. Simply re-push to the same branch.

Nightenom:
Sorry, wasn't aware of being able to recreate branch, since I deleted it I guess it's too late to reopen (github doesn't allow me to do so)

diesieben07:
You cannot reopen it because the branch no longer exists.
You should be able to reopen it once you create the branch again.

Nightenom:
Nope
image

diesieben07:
Welp, I did not know that. That is a stupid feature. Oh well.

PR #6990, opened 2020-07-18
Fixed modded overworld biomes not spawning [1.16.x]

ghost:
Adapted the same patch on 1.15.2 to 1.16.1 Forge to allow modded forge biomes from `BiomeManager` to be used in `BiomeLayer` with weighted lists. Should be straight forward and crucial patch, of course let me know if there are changes required to the styling or similar.

PR will allow modded biomes to spawn in the overworld.  Test mod `modded_overworld_biomes_test` included, adding the biome `red_biome` with bright red grass so it can be found easily.

### DIFF
--- a/patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
+++ b/patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
@@ -1,9 +1,21 @@
 --- a/net/minecraft/world/gen/layer/BiomeLayer.java
 +++ b/net/minecraft/world/gen/layer/BiomeLayer.java
-@@ -28,12 +28,32 @@
-    private static final int[] field_202746_t = new int[]{field_202730_d, field_202729_c, field_202741_o, field_202736_j};
-    private static final int[] field_202747_u = new int[]{field_202731_e, field_202731_e, field_202731_e, field_202742_p};
-    private int[] field_151623_c = field_202744_r;
+@@ -22,18 +22,34 @@
+    private static final int field_202740_n = Registry.field_212624_m.func_148757_b(Biomes.field_76780_h);
+    private static final int field_202741_o = Registry.field_212624_m.func_148757_b(Biomes.field_76768_g);
+    private static final int field_202742_p = Registry.field_212624_m.func_148757_b(Biomes.field_150584_S);
+-   private static final int[] field_202743_q = new int[]{field_202728_b, field_202730_d, field_202729_c, field_202740_n, field_202736_j, field_202741_o};
+-   private static final int[] field_202744_r = new int[]{field_202728_b, field_202728_b, field_202728_b, field_202739_m, field_202739_m, field_202736_j};
+-   private static final int[] field_202745_s = new int[]{field_202730_d, field_202738_l, field_202729_c, field_202736_j, field_202727_a, field_202740_n};
+-   private static final int[] field_202746_t = new int[]{field_202730_d, field_202729_c, field_202741_o, field_202736_j};
+-   private static final int[] field_202747_u = new int[]{field_202731_e, field_202731_e, field_202731_e, field_202742_p};
+-   private int[] field_151623_c = field_202744_r;
++   @Deprecated private static final int[] field_202743_q = new int[]{field_202728_b, field_202730_d, field_202729_c, field_202740_n, field_202736_j, field_202741_o};
++   @Deprecated private static final int[] field_202744_r = new int[]{field_202728_b, field_202728_b, field_202728_b, field_202739_m, field_202739_m, field_202736_j};
++   @Deprecated private static final int[] field_202745_s = new int[]{field_202730_d, field_202738_l, field_202729_c, field_202736_j, field_202727_a, field_202740_n};
++   @Deprecated private static final int[] field_202746_t = new int[]{field_202730_d, field_202729_c, field_202741_o, field_202736_j};
++   @Deprecated private static final int[] field_202747_u = new int[]{field_202731_e, field_202731_e, field_202731_e, field_202742_p};
++   @Deprecated private int[] field_151623_c = field_202744_r;
 +   private java.util.List<net.minecraftforge.common.BiomeManager.BiomeEntry>[] biomes = new java.util.ArrayList[net.minecraftforge.common.BiomeManager.BiomeType.values().length];
  
     public BiomeLayer(boolean p_i232147_1_) {
@@ -11,18 +23,14 @@
 -         this.field_151623_c = field_202743_q;
 +      for (net.minecraftforge.common.BiomeManager.BiomeType type : net.minecraftforge.common.BiomeManager.BiomeType.values()) {
 +         com.google.common.collect.ImmutableList<net.minecraftforge.common.BiomeManager.BiomeEntry> biomesToAdd = net.minecraftforge.common.BiomeManager.getBiomes(type);
-+         int idx = type.ordinal();
-+
-+         if (biomes[idx] == null) biomes[idx] = new java.util.ArrayList<net.minecraftforge.common.BiomeManager.BiomeEntry>();
-+         if (biomesToAdd != null) biomes[idx].addAll(biomesToAdd);
++         if (biomes[type.ordinal()] == null) biomes[type.ordinal()] = new java.util.ArrayList<net.minecraftforge.common.BiomeManager.BiomeEntry>();
++         if (biomesToAdd != null) biomes[type.ordinal()].addAll(biomesToAdd);
        }
  
 +      int desertIdx = net.minecraftforge.common.BiomeManager.BiomeType.DESERT.ordinal();
-+
 +      biomes[desertIdx].add(new net.minecraftforge.common.BiomeManager.BiomeEntry(Biomes.field_76769_d, 30));
 +      biomes[desertIdx].add(new net.minecraftforge.common.BiomeManager.BiomeEntry(Biomes.field_150588_X, 20));
 +      biomes[desertIdx].add(new net.minecraftforge.common.BiomeManager.BiomeEntry(Biomes.field_76772_c, 10));
-+
 +      if (p_i232147_1_) {
 +         biomes[desertIdx].clear();
 +         biomes[desertIdx].add(new net.minecraftforge.common.BiomeManager.BiomeEntry(Biomes.field_76769_d, 10));
@@ -35,7 +43,7 @@
     }
  
     public int func_202726_a(INoiseRandom p_202726_1_, int p_202726_2_) {
-@@ -46,21 +66,21 @@
+@@ -46,21 +62,21 @@
                 return p_202726_1_.func_202696_a(3) == 0 ? field_202733_g : field_202734_h;
              }
  
@@ -61,15 +69,17 @@
           default:
              return field_202735_i;
           }
-@@ -68,4 +88,11 @@
+@@ -68,4 +84,13 @@
           return p_202726_2_;
        }
     }
 +
++   /* ======================================== FORGE START =====================================*/
 +   protected net.minecraftforge.common.BiomeManager.BiomeEntry getWeightedBiomeEntry(net.minecraftforge.common.BiomeManager.BiomeType type, INoiseRandom context) {
 +      java.util.List<net.minecraftforge.common.BiomeManager.BiomeEntry> biomeList = biomes[type.ordinal()];
 +      int totalWeight = net.minecraft.util.WeightedRandom.func_76272_a(biomeList);
-+      int weight = net.minecraftforge.common.BiomeManager.isTypeListModded(type)?context.func_202696_a(totalWeight):context.func_202696_a(totalWeight / 10) * 10;
++      int weight = net.minecraftforge.common.BiomeManager.isTypeListModded(type) ? context.func_202696_a(totalWeight) : context.func_202696_a(totalWeight / 10) * 10;
 +      return (net.minecraftforge.common.BiomeManager.BiomeEntry)net.minecraft.util.WeightedRandom.func_180166_a(biomeList, weight);
 +   }
++   /* ========================================= FORGE END ======================================*/
  }
--- a/patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
+++ b/patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
@@ -69,7 +69,7 @@
 +   protected net.minecraftforge.common.BiomeManager.BiomeEntry getWeightedBiomeEntry(net.minecraftforge.common.BiomeManager.BiomeType type, INoiseRandom context) {
 +      java.util.List<net.minecraftforge.common.BiomeManager.BiomeEntry> biomeList = biomes[type.ordinal()];
 +      int totalWeight = net.minecraft.util.WeightedRandom.func_76272_a(biomeList);
-+      int weight = context.func_202696_a(totalWeight);
++      int weight = net.minecraftforge.common.BiomeManager.isTypeListModded(type)?context.func_202696_a(totalWeight):context.func_202696_a(totalWeight / 10) * 10;
 +      return (net.minecraftforge.common.BiomeManager.BiomeEntry)net.minecraft.util.WeightedRandom.func_180166_a(biomeList, weight);
 +   }
  }
--- a/patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
+++ b/patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
@@ -1,21 +1,9 @@
 --- a/net/minecraft/world/gen/layer/BiomeLayer.java
 +++ b/net/minecraft/world/gen/layer/BiomeLayer.java
-@@ -22,18 +22,34 @@
-    private static final int field_202740_n = Registry.field_212624_m.func_148757_b(Biomes.field_76780_h);
-    private static final int field_202741_o = Registry.field_212624_m.func_148757_b(Biomes.field_76768_g);
-    private static final int field_202742_p = Registry.field_212624_m.func_148757_b(Biomes.field_150584_S);
--   private static final int[] field_202743_q = new int[]{field_202728_b, field_202730_d, field_202729_c, field_202740_n, field_202736_j, field_202741_o};
--   private static final int[] field_202744_r = new int[]{field_202728_b, field_202728_b, field_202728_b, field_202739_m, field_202739_m, field_202736_j};
--   private static final int[] field_202745_s = new int[]{field_202730_d, field_202738_l, field_202729_c, field_202736_j, field_202727_a, field_202740_n};
--   private static final int[] field_202746_t = new int[]{field_202730_d, field_202729_c, field_202741_o, field_202736_j};
--   private static final int[] field_202747_u = new int[]{field_202731_e, field_202731_e, field_202731_e, field_202742_p};
--   private int[] field_151623_c = field_202744_r;
-+   @Deprecated private static final int[] field_202743_q = new int[]{field_202728_b, field_202730_d, field_202729_c, field_202740_n, field_202736_j, field_202741_o};
-+   @Deprecated private static final int[] field_202744_r = new int[]{field_202728_b, field_202728_b, field_202728_b, field_202739_m, field_202739_m, field_202736_j};
-+   @Deprecated private static final int[] field_202745_s = new int[]{field_202730_d, field_202738_l, field_202729_c, field_202736_j, field_202727_a, field_202740_n};
-+   @Deprecated private static final int[] field_202746_t = new int[]{field_202730_d, field_202729_c, field_202741_o, field_202736_j};
-+   @Deprecated private static final int[] field_202747_u = new int[]{field_202731_e, field_202731_e, field_202731_e, field_202742_p};
-+   @Deprecated private int[] field_151623_c = field_202744_r;
+@@ -28,12 +28,28 @@
+    private static final int[] field_202746_t = new int[]{field_202730_d, field_202729_c, field_202741_o, field_202736_j};
+    private static final int[] field_202747_u = new int[]{field_202731_e, field_202731_e, field_202731_e, field_202742_p};
+    private int[] field_151623_c = field_202744_r;
 +   private java.util.List<net.minecraftforge.common.BiomeManager.BiomeEntry>[] biomes = new java.util.ArrayList[net.minecraftforge.common.BiomeManager.BiomeType.values().length];
  
     public BiomeLayer(boolean p_i232147_1_) {
--- a/patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
+++ b/patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
@@ -1,6 +1,6 @@
 --- a/net/minecraft/world/gen/layer/BiomeLayer.java
 +++ b/net/minecraft/world/gen/layer/BiomeLayer.java
-@@ -22,18 +22,35 @@
+@@ -22,18 +22,32 @@
     private static final int field_202740_n = Registry.field_212624_m.func_148757_b(Biomes.field_76780_h);
     private static final int field_202741_o = Registry.field_212624_m.func_148757_b(Biomes.field_76768_g);
     private static final int field_202742_p = Registry.field_212624_m.func_148757_b(Biomes.field_150584_S);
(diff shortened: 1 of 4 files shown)


marchermans (commented):
In general this looks good.
However due to the nature of the biome and worldgen systems in 1.16, I am not sure this will be pulled into the main branch before those systems stabilize. 

This is not a rejection of the pr, just a heads up that it might take a bit and that tu ou might need to keep it up to date for a bit while other changes are merged. Up until the point either we think it can be merged or untill mojang decides that the biome and worldgen systems is ready and stable. 

Greets orion

marchermans:
Additionally things like proper documentation of why this is needed, and test mods that show the system in action are missing.

@Xerousso

ghost:
Yeah, I'll see what I can do about some test mods, I was just reading the contribution guide and it said for new hooks test mods are required, which this isn't as the hooks already exist in `BiomeManager`, this patch is just to make them actually work.

The PR is required as currently the information given to `BiomeManager` by adding biomes and similar is not actually taken into consideration for the `BiomeLayer` when creating the world, meaning that right now people can only add biomes to the overworld by making their own world type and layers, which is rather poor, or achieve it through coremodding which as we all know is extremely incompatible.

Xerousso:
Also I was told that the failure of the jenkins build was on jenkins end, and I'm hoping that it won't negatively impact my PR by making people think that it doesn't build correctly?  If anyone has the ability to rerun that it would be appreciated :)

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
```diff
@@ -0,0 +1,79 @@
...
+@@ -22,18 +22,35 @@
+    private static final int field_202740_n = Registry.field_212624_m.func_148757_b(Biomes.field_76780_h);
+    private static final int field_202741_o = Registry.field_212624_m.func_148757_b(Biomes.field_76768_g);
+    private static final int field_202742_p = Registry.field_212624_m.func_148757_b(Biomes.field_150584_S);
+-   private static final int[] field_202743_q = new int[]{field_202728_b, field_202730_d, field_202729_c, field_202740_n, field_202736_j, field_202741_o};
+-   private static final int[] field_202744_r = new int[]{field_202728_b, field_202728_b, field_202728_b, field_202739_m, field_202739_m, field_202736_j};
+-   private static final int[] field_202745_s = new int[]{field_202730_d, field_202738_l, field_202729_c, field_202736_j, field_202727_a, field_202740_n};
+-   private static final int[] field_202746_t = new int[]{field_202730_d, field_202729_c, field_202741_o, field_202736_j};
+-   private static final int[] field_202747_u = new int[]{field_202731_e, field_202731_e, field_202731_e, field_202742_p};
+-   private int[] field_151623_c = field_202744_r;
```
pupnewfster: Is there a reason these fields are being deleted rather than just no longer using them?

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
```diff
@@ -0,0 +1,79 @@
...
++   private java.util.List<net.minecraftforge.common.BiomeManager.BiomeEntry>[] biomes = new java.util.ArrayList[net.minecraftforge.common.BiomeManager.BiomeType.values().length];
+-   public BiomeLayer(boolean p_i232147_1_) {
+-      if (p_i232147_1_) {
+-         this.field_151623_c = field_202743_q;
++   public BiomeLayer(boolean p_i232147_1_)
++   {
```
pupnewfster: There is no reason to be changing the formatting of vanilla classes (such as moving the bracket to the next line). And the formatting should match the rest of the file, which means that for patches brackets go on the same line not on the next one (there are a lot of cases below where you are putting them on the next one).

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
```diff
@@ -0,0 +1,79 @@
...
++         biomes[desertIdx].add(new net.minecraftforge.common.BiomeManager.BiomeEntry(Biomes.field_76772_c, 10));
++         biomes[desertIdx].add(new net.minecraftforge.common.BiomeManager.BiomeEntry(Biomes.field_76768_g, 10));
++      }
+    }
+    public int func_202726_a(INoiseRandom p_202726_1_, int p_202726_2_) {
+@@ -46,21 +63,21 @@
+                return p_202726_1_.func_202696_a(3) == 0 ? field_202733_g : field_202734_h;
+             }
+-            return this.field_151623_c[p_202726_1_.func_202696_a(this.field_151623_c.length)];
++            return Registry.field_212624_m.func_148757_b(getWeightedBiomeEntry(net.minecraftforge.common.BiomeManager.BiomeType.DESERT, p_202726_1_).biome);
```
pupnewfster: I am not actually positive on the stance of Forge Registry vs Vanilla Registry use in patches, but is there a reason this is using the vanilla registry instead of the Forge Registry for biomes?

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
ghost: As I said this is an adaption of a previous patch that was present in 1.15.2, I can add them back in if you would like, although they do have no use.

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
Xerousso: There is not a reason, again it was an adaptation.

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
ghost: Ah thanks, this is my first time contributing so I was not sure.

LexManos:
```> Task :forge:applyPatches FAILED

  Apply Patch: net/minecraft/world/gen/layer/BiomeLayer.java.patch

    Hunk #1 Failed: Cannot find hunk target

    Hunk #2 Failed: Cannot find hunk target```

File: patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch (hunk 3 above)
Comment (ghost) [reply patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch]: Looking at it closer, it appears that the forge registries do not have an option to get the numerical id of a biome, correct me if I'm wrong.  This is why vanilla registries are required here.

File: patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch (hunk 2 above)
Comment (ghost) [reply patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch]: Fixed

File: patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch (hunk 3 above)
Comment (Shadows-of-Fire) [reply patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch]: They... do, but that requires a cast to the actual underlying type (ForgeRegistry) rather than the field type of IForgeRegistry

Comment (ghost) [reply patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch]: Ah, as such?:
```java
return ((ForgeRegistry) ForgeRegistries.BIOMES).getID(getWeightedBiomeEntry(net.minecraftforge.common.BiomeManager.BiomeType.ICY, context).biome);
```
I suppose I should replace it with this?  Let me know what you think

ghost: > and I do not prefer the ForgeRegistry cast hack
> as it ties things to one implementation
> - tterrag

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
ghost: Re-added

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
```diff
@@ -0,0 +1,75 @@
...
++      for (net.minecraftforge.common.BiomeManager.BiomeType type : net.minecraftforge.common.BiomeManager.BiomeType.values()) {
++         com.google.common.collect.ImmutableList<net.minecraftforge.common.BiomeManager.BiomeEntry> biomesToAdd = net.minecraftforge.common.BiomeManager.getBiomes(type);
++         int idx = type.ordinal();
++
++         if (biomes[idx] == null) biomes[idx] = new java.util.ArrayList<net.minecraftforge.common.BiomeManager.BiomeEntry>();
++         if (biomesToAdd != null) biomes[idx].addAll(biomesToAdd);
+       }
+ 
++      int desertIdx = net.minecraftforge.common.BiomeManager.BiomeType.DESERT.ordinal();
++
```
valoeghese: Not sure if extra whitespace in a patch is a problem or not, since forge prefers patches to be as small as possible. But it's probably fine.

valoeghese (approved):

ghost:
By the way, if the PR is held back because things change in 1.16.2, feel free to hit me up to port my PRs, I'll be more than happy to do so :)

TelepathicGrunt (changes requested):
Looks good! The only thing I would change is line 72 and remove the `context.func_202696_a(totalWeight / 10) * 10` as vanilla biomes are already multiples of 10 which makes the isTypeListModded check useless. So that line can just be `int weight = context.func_202696_a(totalWeight)` for a slightly smaller patch and 1 less check.

TelepathicGrunt (approved):

ghost:
Lmao what?  I change the title, it rechecks and this time fails?

srs-bsns:
> Looks good! The only thing I would change is line 72 and remove the `context.func_202696_a(totalWeight / 10) * 10` as vanilla biomes are already multiples of 10 which makes the isTypeListModded check useless. So that line can just be `int weight = context.func_202696_a(totalWeight)` for a slightly smaller patch and 1 less check.

The vanilla biome weights are multiples of 10 because Forge adds them in the BiomeManager to have a weight that is 10x their normal incidence so that mods can add biomes to be much rarer (weight < 10). The call to `func_202696_a` (`LazyAreaLayerContext#random`), when there are no mod-added biomes present, reduces the total weight value so that the call to `#random` properly mimics vanilla behavior.

Mysterious-Dev (approved):

ghost:
> 
> 
> > Looks good! The only thing I would change is line 72 and remove the `context.func_202696_a(totalWeight / 10) * 10` as vanilla biomes are already multiples of 10 which makes the isTypeListModded check useless. So that line can just be `int weight = context.func_202696_a(totalWeight)` for a slightly smaller patch and 1 less check.
> 
> The vanilla biome weights are multiples of 10 because Forge adds them in the BiomeManager to have a weight that is 10x their normal incidence so that mods can add biomes to be much rarer (weight < 10). The call to `func_202696_a` (`LazyAreaLayerContext#random`), when there are no mod-added biomes present, reduces the total weight value so that the call to `#random` properly mimics vanilla behavior.

How will reducing all weights by the same factor affect how `#random` does vanilla behavior?

srs-bsns:
> How will reducing all weights by the same factor affect how `#random` does vanilla behavior?

When calculating a pseudorandom number with a provided bound, the random number that is returned is the next pseudorandom number in the squence that is put through a modulus operation.
So, in this case, for example: `(12345678 % 6) * 10` results in `0` and `12345678 % 60` results in `18`, which would lead to the return of different items from the list. Thus, dividing the total weight by 10 before getting the next random, (in the case that the biome list is unmodified,) preserves the vanilla selection of biome. Without doing this, any given world, even when there are no added biomes, would have a drastically different biome layout than a vanilla world using the same seed. (This was likely @Adubbz intention, and fallows the Forge axiom of preserving vanilla behaviour.)

As I commented in your other PR (https://github.com/MinecraftForge/MinecraftForge/pull/7000#issuecomment-662168707) it would probably be prudent to wait until 1.16.2 has been released in a couple of weeks, since this code will likely become vestigial.

TelepathicGrunt:
Yeah the suggestion I made on the random was my mistake but as for the prudent part, that's really unfounded concerns as BiomeLayer remains untouched in 1.16.2. This is a 1.15.2 hook that can be re-implemented into 1.16.1 and 1.16.2 with no change and without this hook, many biome mods are unable to port at all making this a high impact issue. 

It is highly unlikely that Mojang will change biomelayers between the current snapshot and the 1.16.2 full release. If for some god-forsaken reason they do change it, we can make another pr to re-implement the hook in a new form without burdening the forge maintainers trying to update Forge. Plus, we don't even know when 1.16.2 could come out. Some people say a week. others say two weeks. Maybe more. We simply don't know and my take is that trying to predict what Mojang is doing is not very wise and most likely isn't going to turn out right.

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
```diff
@@ -0,0 +1,75 @@
...
++            return Registry.field_212624_m.func_148757_b(getWeightedBiomeEntry(net.minecraftforge.common.BiomeManager.BiomeType.ICY, p_202726_1_).biome);
+          default:
+             return field_202735_i;
+          }
+@@ -68,4 +88,11 @@
+          return p_202726_2_;
+       }
+    }
++
++   protected net.minecraftforge.common.BiomeManager.BiomeEntry getWeightedBiomeEntry(net.minecraftforge.common.BiomeManager.BiomeType type, INoiseRandom context) {
```
marchermans: You need to add the Forge method markers. Additionally please check that this is at the end of the file.

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
```diff
@@ -0,0 +1,75 @@
...
+    private int[] field_151623_c = field_202744_r;
++   private java.util.List<net.minecraftforge.common.BiomeManager.BiomeEntry>[] biomes = new java.util.ArrayList[net.minecraftforge.common.BiomeManager.BiomeType.values().length];
+ 
+    public BiomeLayer(boolean p_i232147_1_) {
+-      if (p_i232147_1_) {
+-         this.field_151623_c = field_202743_q;
++      for (net.minecraftforge.common.BiomeManager.BiomeType type : net.minecraftforge.common.BiomeManager.BiomeType.values()) {
++         com.google.common.collect.ImmutableList<net.minecraftforge.common.BiomeManager.BiomeEntry> biomesToAdd = net.minecraftforge.common.BiomeManager.getBiomes(type);
++         int idx = type.ordinal();
++
```
marchermans: Remove the not needed whiteline.

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
```diff
@@ -0,0 +1,75 @@
...
+    private static final int[] field_202747_u = new int[]{field_202731_e, field_202731_e, field_202731_e, field_202742_p};
+    private int[] field_151623_c = field_202744_r;
++   private java.util.List<net.minecraftforge.common.BiomeManager.BiomeEntry>[] biomes = new java.util.ArrayList[net.minecraftforge.common.BiomeManager.BiomeType.values().length];
+ 
+    public BiomeLayer(boolean p_i232147_1_) {
+-      if (p_i232147_1_) {
+-         this.field_151623_c = field_202743_q;
++      for (net.minecraftforge.common.BiomeManager.BiomeType type : net.minecraftforge.common.BiomeManager.BiomeType.values()) {
++         com.google.common.collect.ImmutableList<net.minecraftforge.common.BiomeManager.BiomeEntry> biomesToAdd = net.minecraftforge.common.BiomeManager.getBiomes(type);
++         int idx = type.ordinal();
```
marchermans: You could probably get rid of this line as well.
Just use `type.ordinal()` in the couple of places you are using it.

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
```diff
@@ -0,0 +1,75 @@
+--- a/net/minecraft/world/gen/layer/BiomeLayer.java
++++ b/net/minecraft/world/gen/layer/BiomeLayer.java
+@@ -28,12 +28,32 @@
+    private static final int[] field_202746_t = new int[]{field_202730_d, field_202729_c, field_202741_o, field_202736_j};
```
marchermans, line 4: Are these original fields still in use?
If not: Put a  `@Deprecated` in front of them (on the same line) and a comment behind them referencing their replacement (also on the same line)

Eventhough these fields might be private, if vanilla code becomes unused by forge changes. We should keep track of this. Not just for people using the code, but also for porters who port the patch in the future. It is then always good to know what happens with a field.

Do not remove the field since that can also cause issue. As I said a `@Deprecated` and a comment should suffice.

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
```diff
@@ -0,0 +1,75 @@
...
++         if (biomes[idx] == null) biomes[idx] = new java.util.ArrayList<net.minecraftforge.common.BiomeManager.BiomeEntry>();
++         if (biomesToAdd != null) biomes[idx].addAll(biomesToAdd);
+       }
+ 
++      int desertIdx = net.minecraftforge.common.BiomeManager.BiomeType.DESERT.ordinal();
++
++      biomes[desertIdx].add(new net.minecraftforge.common.BiomeManager.BiomeEntry(Biomes.field_76769_d, 30));
++      biomes[desertIdx].add(new net.minecraftforge.common.BiomeManager.BiomeEntry(Biomes.field_150588_X, 20));
++      biomes[desertIdx].add(new net.minecraftforge.common.BiomeManager.BiomeEntry(Biomes.field_76772_c, 10));
++
```
marchermans: Please remove this whiteline.
It is not needed.

marchermans (approved):
This is starting to look good.
@Xerousso Could you check the Randomization issue that @TelepathicGrunt mentioned, and make sure that, if no modded biomelayer is registered, the behaviour is identical to vanilla?

@Xerousso Additionally it would be amazing if you updated your initial PR opening message. We are trying to get our documentation to a better place. And as such it would be nice if you could provide a minimal example (a minimal amount of line suffices) that shows the usage, or reference a test mod, if it exists. If no testmod exists, please additionally create one.

Greets,
Orion

P.S.: We discussed this, since it does not look like that this will be touched by the 1.16.2 changes, we will probably pass this on for merging as soon as you did the cleanup and documentation.

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
ghost: This is at the end of the file, what are Forge method markers @OrionDevelopment ?

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
ghost: Sure, will do

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
ghost: Sure, will do

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
ghost: Sure, will do

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
ghost: I am not quite sure as to what you mean @OrionDevelopment ?  Which line?

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
marchermans: Look at other patched vanilla classes which have methods that have been added by forge.
There are two comments: ///////// FORGE START /////////// and /////////// FORGE END //////////// or something in that direction.
These enclose the methods that forge added to that class, and allow us to easily identify the methods which are not from the vanilla code base and are very handy during porting.

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
marchermans: See this line 14 of your patch file?. It defines the variable idx as type.ordinal().
And is then only used todo look ups 3 times.
This variable as such is not really needed, and can be removed to reduce the patch size.

ghost: Ah, will remove

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
ghost: Ah, makes sense.  Done.

ghost:
Still need to find a place to learn how to make a test mod, I was told to look at other PRs but I haven't found anything but patches.

srs-bsns:
> that's really unfounded concerns as BiomeLayer remains untouched in 1.16.2.

My opinion was more than just about this patch, but whatever, it's not that important.

With respect to this patch, `#getWeightedBiomeEntry` probably no longer needs to be here. The method was patched into `GenLayerBiome` back in 1.7.10 because it required access to the protected superclass method `GenLayer#nextInt`. This is not the case anymore, since in 1.14 (or maybe it was 1.13, but I don't have any old artifacts kicking around to check.) RNG is handled by an external `LazyAreaLayerContext` object that can simply be passed to an external method call.

I suggest that `getWeightedBiomeEntry` be moved to the Forge BiomeManager where it belongs, and external calls be made to it to select a biome at random. This will reduce the overall patch size, and any changes that need to be made to this process in the future should be exclusive to Forge classes.

ghost:
https://gist.github.com/Xerousso/e135fd69efa617444a11ac9e9cdfc2b9
Would adding it to `BiomeManager` as seen in this gist be preferable?

srs-bsns:
> 
> 
> https://gist.github.com/Xerousso/e135fd69efa617444a11ac9e9cdfc2b9
> Would adding it to `BiomeManager` as seen in this gist be preferable?

Basically that, yes.

ilja615 (approved):
I approve this PR and I hope this PR will be added to forge so that we can make ModdedBiome, I am looking forward to creating my own ModdedBiomes in 1.16.1 as well as to see what amazing creations the other modders would  make!
- Greetings, 
Ilja 615

CorgiTaco (approved):
I approve This PR is rather essential because custom world types in 1.16 have been rather defective and world types for small biome adding mods isn't necessarily efficient.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -154,6 +155,16 @@ public BiomeEntry(Biome biome, int weight)
...
 
+    public static BiomeEntry getWeightedBiomeEntry(BiomeType type, INoiseRandom context)
+    {
+        List<BiomeEntry> biomeList = biomes[type.ordinal()];
+
+        int totalWeight = WeightedRandom.getTotalWeight(biomeList);
+        int weight = isTypeListModded(type) ? context.random(totalWeight) : context.random(totalWeight / 10) * 10;
+
+        return WeightedRandom.getRandomItem(biomeList, weight);
+    }
```
srs-bsns, line 166: This method can probably be renamed to `getWeightedBiome` and just return the `Biome`, since the weight value is only needed here and not in BiomeLayer.

ghost: True, but it may be useful for other things that we haven't thought about to use this method and get the weight

srs-bsns: Even if the weight value is useful for other processes, those other processes should probably be handled within `BiomeManager` or in other processes called from it, and not done within `BiomeLayer`, which would also require further patching it. At the point that this method returns, it should only be the `Biome`, since that it all that is consequential in `BiomeLayer`.

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
```diff
@@ -0,0 +1,71 @@
...
+-   private static final int[] field_202746_t = new int[]{field_202730_d, field_202729_c, field_202741_o, field_202736_j};
+-   private static final int[] field_202747_u = new int[]{field_202731_e, field_202731_e, field_202731_e, field_202742_p};
+-   private int[] field_151623_c = field_202744_r;
++   @Deprecated private static final int[] field_202743_q = new int[]{field_202728_b, field_202730_d, field_202729_c, field_202740_n, field_202736_j, field_202741_o};
++   @Deprecated private static final int[] field_202744_r = new int[]{field_202728_b, field_202728_b, field_202728_b, field_202739_m, field_202739_m, field_202736_j};
++   @Deprecated private static final int[] field_202745_s = new int[]{field_202730_d, field_202738_l, field_202729_c, field_202736_j, field_202727_a, field_202740_n};
++   @Deprecated private static final int[] field_202746_t = new int[]{field_202730_d, field_202729_c, field_202741_o, field_202736_j};
++   @Deprecated private static final int[] field_202747_u = new int[]{field_202731_e, field_202731_e, field_202731_e, field_202742_p};
++   @Deprecated private int[] field_151623_c = field_202744_r;
```
marchermans: Please add a comment stating why they are deprecated and what their replacement is, if any.

src/main/java/net/minecraftforge/common/BiomeManager.java
marchermans: @Xerousso You can leave it like this.
It does not hurt to have it return `BiomeEntry` instead of biome.

src/test/java/net/minecraftforge/debug/world/ModdedOverworldBiomesTest.java
```diff
@@ -0,0 +1,52 @@
...
+    {
+        FMLJavaModLoadingContext.get().getModEventBus().addListener(this::setup);
+        IEventBus modBus = FMLJavaModLoadingContext.get().getModEventBus();
+        BIOMES.register(modBus);
+    }
+
+    private void setup(final FMLCommonSetupEvent event)
+    {
+        BiomeManager.addBiome(BiomeType.WARM, new BiomeEntry(redBiome.get(), 200));
+    }
```
marchermans: See the other mods on how todo these. Most notably missing is a way to disable a given testmod.

marchermans (changes requested):
Some small nitpickings.

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
Xerousso: @OrionDevelopment Should I do a javadoc for each one or just a standard comment?

src/test/java/net/minecraftforge/debug/world/ModdedOverworldBiomesTest.java
Xerousso: I've checked 10 different test mods and have not found a way to disable a given one in their code?

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
ghost: LexManos and diesieben07 said its pointless to do a comment or deprecate these at all as they are private fields.

src/test/java/net/minecraftforge/debug/world/ModdedOverworldBiomesTest.java
ghost: Done.

marchermans:
@Xerousso Why did you remove the  `@Deprecated` markers?
`Has been resolved on Discord`Sio disregard

marchermans (approved):
Readup on the discord discussion. Sorry missed that.
Okey this PR looks good in my eyes.

Assigning it to lex.

PR #6992, opened 2020-07-18
[1.16] Re-added PlayerEvent.NameFormat

Tschipp:
Since it seems that `PlayerEvent.NameFormat` wasn't ported to more recent versions of Forge, I have reimplemented it here.
The behavior is identical to how it used to work:
- The `PlayerEvent.NameFormat` is fired properly and can change the display name of a player.
- This affects the name tag in the world and in the chat.
- This does not affect the name in the player list (tab). This is intended and how it used to work.
- `EntityPlayer#refreshDisplayName()` needs to be called to re-fire the `PlayerEvent.NameFormat` event.

Changes:
- Didn't readd `EntityPlayer#getDisplayNameString()`. Instead, the `displayname` field is now set in `EntityPlayer#getDisplayName()`
- Changed `PlayerEvent.NameFormat` and all methods that rely on it to use `ITextComponent` instead of `String`.

- This PR also contains a test mod


src/test/resources/META-INF/mods.toml
```diff
@@ -71,4 +71,6 @@ loaderVersion="[28,)"
 [[mods]]
     modId="ender_mask_test"
 [[mods]]
-    modId="raid_enum_test"
\ No newline at end of file
+    modId="raid_enum_test"
+[[mods]]
+	modId="player_name_event_test"
```
Yunus1903: Weird spacing here

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
```diff
@@ -288,19 +288,23 @@
...
+@@ -1702,9 +1743,13 @@
+    public Scoreboard func_96123_co() {
+       return this.field_70170_p.func_96441_U();
     }
- 
+-
++   
     public ITextComponent func_145748_c_() {
 -      IFormattableTextComponent iformattabletextcomponent = ScorePlayerTeam.func_237500_a_(this.func_96124_cp(), this.func_200200_C_());
++      if(this.displayname == null) this.displayname = net.minecraftforge.event.ForgeEventFactory.getPlayerDisplayName(this, this.func_200200_C_());
```
Yunus1903: if statement has no space between the `if` and `(`

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
```diff
@@ -313,6 +317,14 @@
...
++   private ITextComponent displayname = null;
++   /**
++    * Force the displayed name to refresh
++    */
++   public void refreshDisplayName() {
++       this.displayname = net.minecraftforge.event.ForgeEventFactory.getPlayerDisplayName(this, this.func_200200_C_());
++   }
 +
 +   private final net.minecraftforge.common.util.LazyOptional<net.minecraftforge.items.IItemHandler>
 +         playerMainHandler = net.minecraftforge.common.util.LazyOptional.of(
```
Yunus1903: Minecraft tabs need to be 3 spaces, some of them are longer here

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -152,26 +153,26 @@ public BreakSpeed(PlayerEntity player, BlockState state, float original, BlockPo
      **/
     public static class NameFormat extends PlayerEvent
     {
-        private final String username;
-        private String displayname;
+        private final ITextComponent username;
+        private ITextComponent displayname;
 
-        public NameFormat(PlayerEntity player, String username) {
+        public NameFormat(PlayerEntity player, ITextComponent username) {
```
Yunus1903: Bracket should be on the next line here (for consistency in the class)

Yunus1903 (changes requested):

Tschipp:
Fixed the formatting issues you mentioned

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
```diff
@@ -288,19 +288,23 @@
...
        if (this.field_71068_ca < 0) {
           this.field_71068_ca = 0;
-@@ -1704,7 +1745,10 @@
+@@ -1702,9 +1743,13 @@
+    public Scoreboard func_96123_co() {
+       return this.field_70170_p.func_96441_U();
     }
- 
+-
++
```
diesieben07: Unnecessary patch.

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
```diff
@@ -313,6 +317,14 @@
 +   public Collection<IFormattableTextComponent> getSuffixes() {
 +       return this.suffixes;
 +   }
++   
++   private ITextComponent displayname = null;
++   /**
++    * Force the displayed name to refresh
++    */
```
diesieben07: That JavaDoc explains almost nothing. You should at least link to the event name and clarify that this will cause the event to be re-fired.

diesieben07 (changes requested):

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
Tschipp: Fair point

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
Tschipp: Not sure what happened there...

diesieben07:
Any reason you changed `refreshDisplayName` from just setting the field to `null` (that's how it was when the event was still in)? Other than that looks good now.

Tschipp:
Are you sure? As far as I can tell, the method was last present in 1.12 and looks pretty much exactly like it is in this implementation:
https://github.com/MinecraftForge/MinecraftForge/blob/1.12.x/patches/minecraft/net/minecraft/entity/player/EntityPlayer.java.patch#L615

diesieben07:
Well, I remembered that wrong then, nvm

Tschipp:
Changed the implementation to _hopefully_ be non-breaking.

ichttt:
You don't have to make this non-breaking, forge is still in beta (has no recommeneded build), so breaking changes are allowed.

Tschipp:
Yes I know, but I was hoping this could perhaps be backported to 1.15 as well (I could also make a PR for that)

ichttt:
I would suggest you would leave this PR as a breaking one, and implement it in a non-breaking way in a different PR for 1.15, so we can get rid of legacy code in 1.16

Tschipp:
Hm yeah you're probably right. I'll try to revert the commit.

darentuzi:
@LexManos does this look good? Would be a cool feature to have.

PR #6994, opened 2020-07-19
[1.16] Reimplement ICloudRenderHandler, ISkyRenderHandler and IWeatherRenderHandler

alcatrazEscapee:
These seem to have been dropped during the 1.16 update cycle as they got lost in dimension changes, and without them there is no way to change cloud / sky / weather rendering of existing dimensions.

- I've re-added the calls to each interface in the same locations as they were in 1.15 (as that has been unchanged).
- The render handlers themselves have been moved from `Dimension` to `DimensionRenderInfo` as that serves the same purpose that `Dimension` was before in this case.
- `IRenderHandler` has been removed, and instead split into three specific interfaces for each render hook, as was described in #6537, since the different renderers need different arguments.


src/main/java/net/minecraftforge/client/IWeatherRenderHandler.java
```diff
@@ -19,14 +19,14 @@
...
+/**
+ * Call {@link net.minecraft.client.world.DimensionRenderInfo#setWeatherRenderHandler(IWeatherRenderHandler)}, obtained from a {@link ClientWorld} with an implementation of this to override all weather rendering with your own.
+ * This includes rain and snow.
+ */
 @FunctionalInterface
-public interface IRenderHandler
-{
-    @OnlyIn(Dist.CLIENT)
+public interface IWeatherRenderHandler {
     void render(int ticks, float partialTicks, ClientWorld world, Minecraft mc);
```
ichttt: Now that we have seperate interfaces for everything, we might as well add more specific context, so pass in the LightTexture and the position here

src/main/java/net/minecraftforge/client/ICloudRenderHandler.java
```diff
@@ -24,13 +24,10 @@
...
  */
-public interface SkyRenderHandler extends IRenderHandler {
-	@Override
-	default void render(int ticks, float partialTicks, ClientWorld world, Minecraft mc) {}
-
-	void render(int ticks, float partialTicks, MatrixStack matrixStack, ClientWorld world, Minecraft mc);
-
+@FunctionalInterface
+public interface ICloudRenderHandler {
+    void render(int ticks, float partialTicks, MatrixStack matrixStack, ClientWorld world, Minecraft mc);
```
ichttt: You can also pass the pos in here as an additional parameter

ichttt (changes requested):
A few notes, otherwise looking good to me

src/main/java/net/minecraftforge/client/IWeatherRenderHandler.java
alcatrazEscapee: Makes sense. Both are still obtainable (via `mc.gameRenderer.getLightTexture()` and `mc.gameRenderer.getActiveRenderInfo().getProjectedView()`), but it's simpler if the user doesn't have to track those down as I did :)

ichttt (approved):

alcatrazEscapee:
@tterrag1098 @gigaherz Is there anything I need to do / change / rework here? I'm hoping this can be merged before 1.16.2 (the RB part, not the minecraft release part) without having to revert to non-breaking changes as this was the breaking change which would've been done in 1.15 in #6537

Alpvax:
How do these changes interact with fabulous graphics? (I've not looked at 1.16 code yet, so I'm just asking the question)

alcatrazEscapee:
Merge conflicts fixed, 1.16.2 updated etc. etc. If there's anything I can do here please let me know. I would really benefit from these hooks being here, both for the additive reason but also for the ability to stop vanilla's rendering.

@Alpvax this doesn't have any bearing on your graphics fancyness setting. Vanilla / the modder can call `Minecraft#isFancyGraphicsEnabled`, or `Minecraft#func_238218_y_` (is fabulous graphics enabled) and make changes accordingly whenever they see fit. These hooks simply allow a modder to add rendering at a given stage of the pipeline, but more importantly, cancel vanilla behavior that would otherwise run here.

50ap5ud5:
Thanks for making this, it will be really useful for my use cases!

alcatrazEscapee:
@ Triage people: Merge conflicts fixed, again. Updated to 1.16.3 (*if there's a label for that*). Also this is **no longer a breaking change**, seeing that 1.16.3 is almost certainly going to get RB very soon. So all the old (unused mind you) stuff still exists. I've marked it all `@Deprecated` with links to the replacement interfaces, with `todo`s marking them to be removed during 1.17's update cycle.

tterrag1098:
Seems straightforward enough, though personally I'd be fine with removing the deprecated classes. But I'll leave that decision up to lex.

PR #6997, opened 2020-07-19
Fix ClimberPathNavigator spinning when width is small. Closes #6993

### LINKED ISSUE #6993: Entities using ClimberPathNavigator spinning in navigation when width is small
**Minecraft Version:** 1.16.1

**Forge Version:** 32.0.70

**Logs:** N/A

**Description of issue:**
Entities which use `ClimberPathNavigator` (like a spider) with small width's (something like `0.2F`) start spinning when their path navigation is finished.

~~This issue was fixed before in #6091 but disappeared between the update from 1.15 to 1.16~~
It's caused by a mojang bug (https://bugs.mojang.com/browse/MC-94054), which is still unresolved.

GIF shows the bug in action (In this case I've changed the spider's width to be `0.2F`)

ezgif com-optimize

Yunus1903:
This PR is a fix (also kind of a hack) for #6993 

Entities using `ClimberPathNavigator` (like a spider) keep spinning at the end of their path navigation when their width is 0.9 or smaller.
This PR makes sure when their width is checked for navigation that the width is never lower then 1.

`.withinDistance(this.entity.getPositionVec(), (double)this.entity.getWidth())`
is now
`.withinDistance(this.entity.getPositionVec(), Math.max((double)this.entity.getWidth(), 1.0D))`

### DIFF
--- a/patches/minecraft/net/minecraft/pathfinding/ClimberPathNavigator.java.patch
+++ b/patches/minecraft/net/minecraft/pathfinding/ClimberPathNavigator.java.patch
@@ -1,10 +1,11 @@
 --- a/net/minecraft/pathfinding/ClimberPathNavigator.java
 +++ b/net/minecraft/pathfinding/ClimberPathNavigator.java
-@@ -38,7 +38,7 @@
+@@ -38,7 +38,8 @@
           super.func_75501_e();
        } else {
           if (this.field_179696_f != null) {
 -            if (!this.field_179696_f.func_218137_a(this.field_75515_a.func_213303_ch(), (double)this.field_75515_a.func_213311_cf()) && (!(this.field_75515_a.func_226278_cu_() > (double)this.field_179696_f.func_177956_o()) || !(new BlockPos((double)this.field_179696_f.func_177958_n(), this.field_75515_a.func_226278_cu_(), (double)this.field_179696_f.func_177952_p())).func_218137_a(this.field_75515_a.func_213303_ch(), (double)this.field_75515_a.func_213311_cf()))) {
++            // FORGE: Fix MC-94054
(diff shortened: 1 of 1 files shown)


patches/minecraft/net/minecraft/pathfinding/ClimberPathNavigator.java.patch
```diff
@@ -0,0 +1,11 @@
...
++++ b/net/minecraft/pathfinding/ClimberPathNavigator.java
+@@ -38,7 +38,7 @@
+          super.func_75501_e();
+       } else {
+          if (this.field_179696_f != null) {
+-            if (!this.field_179696_f.func_218137_a(this.field_75515_a.func_213303_ch(), (double)this.field_75515_a.func_213311_cf()) && (!(this.field_75515_a.func_226278_cu_() > (double)this.field_179696_f.func_177956_o()) || !(new BlockPos((double)this.field_179696_f.func_177958_n(), this.field_75515_a.func_226278_cu_(), (double)this.field_179696_f.func_177952_p())).func_218137_a(this.field_75515_a.func_213303_ch(), (double)this.field_75515_a.func_213311_cf()))) {
++            if (!this.field_179696_f.func_218137_a(this.field_75515_a.func_213303_ch(), Math.max((double)this.field_75515_a.func_213311_cf(), 1.0D)) && (!(this.field_75515_a.func_226278_cu_() > (double)this.field_179696_f.func_177956_o()) || !(new BlockPos((double)this.field_179696_f.func_177958_n(), this.field_75515_a.func_226278_cu_(), (double)this.field_179696_f.func_177952_p())).func_218137_a(this.field_75515_a.func_213303_ch(), Math.max((double)this.field_75515_a.func_213311_cf(), 1.0D)))) {
```
ichttt, line 9: A proper fix would be nicer then just enforing a min value of 1.0, but I guess that would be a too complicated patch?
Also, please add a comment saying `FORGE: Fix MC-94054 Entities using ClimberPathNavigator spinning in navigation when width is small` or something like that.

Yunus1903: Enforcing the min value didn't really change anything in the pathfinding other than fix the spinning from what I could see. But yes I also think fixing it the proper way would be more complicated.

LexManos:
This isn't a proper fix, it should actually make the target position a Vec2D set to the center of the block, or the specific point it's trying to get to. If that was indeed the case then the bug would be gone. But sure, this fixes it within an acceptable margin of error.

PR #6999, opened 2020-07-20
Add missing copyright / license headers.

alcatrazEscapee:
I noticed this when the gradle `checkLicense` task was failing. This is what I ended up doing to make that task pass. I'm assuming these are just oversights?


DaemonUmbra:
``gradlew updateLicenses`` does this automatically

alcatrazEscapee:
Ah. Well then just ignore my idiocy.

DaemonUmbra:
If it's not mentioned in the docs that updatedLicenses should be run before making PRs to Forge perhaps you could PR that to the Docs

alcatrazEscapee:
Yes sir right away sir as you wish sir.

PR #7000, opened 2020-07-20
Added hook for custom edge biomes [1.16.x]

ghost:
Added a hook to `Biome` so that people can add edge biomes to their biome, to blend between biomes or to make a more natural border.
Example biome class: https://hatebin.com/jqacmqwywl
```java
@Override
public Biome getEdge(INoiseRandom rand, Biome northBiome, Biome westBiome, Biome southBiome, Biome eastBiome) {
    return Biomes.BADLANDS;
}
```
Very simple hook and patch, let me know any changes that would be required.  Keep in mind that custom biomes cannot be added to spawning in the overworld currently without my other PR: https://github.com/MinecraftForge/MinecraftForge/pull/6990

### DIFF
--- a/patches/minecraft/net/minecraft/world/biome/Biome.java.patch
+++ b/patches/minecraft/net/minecraft/world/biome/Biome.java.patch
@@ -36,7 +36,7 @@
                 return true;
              }
           }
-@@ -589,6 +589,17 @@
+@@ -589,6 +589,22 @@
        }
     }
  
@@ -48,6 +48,11 @@
 +
 +   @Nullable
 +   public Biome getHill(net.minecraft.world.gen.INoiseRandom rand) {
++      return null;
++   }
++
++   @Nullable
++   public Biome getEdge(net.minecraft.world.gen.INoiseRandom rand, Biome northBiome, Biome westBiome, Biome southBiome, Biome eastBiome) {
 +      return null;
 +   }
 +
--- a/patches/minecraft/net/minecraft/world/gen/layer/ShoreLayer.java.patch
+++ b/patches/minecraft/net/minecraft/world/gen/layer/ShoreLayer.java.patch
@@ -1,0 +1,12 @@
+--- a/net/minecraft/world/gen/layer/ShoreLayer.java
++++ b/net/minecraft/world/gen/layer/ShoreLayer.java
+@@ -34,6 +34,9 @@
+ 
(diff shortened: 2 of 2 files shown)


Yunus1903 (approved):

valoeghese (approved):
This is one of two places that “edge” biomes are made. (The other injection point is at a larger scale and mostly focuses on blending biomes of different climates like desert and swamp)

A hook for the other type of edge biome might be nice, but likely out of scope for this pr

ghost:
The other edge biome type is also very easy, not sure whether to include it in this PR.  Should I add it?

TelepathicGrunt (approved):
Looks good to me! I like how the modded biome has the random and the 4 neighboring biomes. This gives quite a lot of flexibility and interesting edges people can make for their biome.

srs-bsns:
It may not be prudent to add any additional patches for biomes or dimensions until after `1.16.2` has been released.
There are some significant changes being made that will likely require a rewrite of the BiomeManager that will change how biomes are registered with it to respect the new datapack properties for `multi_noise` `biome_source`s. (`temperature`, `humidity`, `altitude`, `weirdness`, `offset`)
Special case biomes, such as custom oceans, edge biomes, beaches, or rare mutated versions of biomes can all likely be handled in a BiomeManager rewrite with minimal patch changes to vanilla classes.

ghost:
I am more than willing to update the patch myself when 1.16.2 rolls around.  From what I've seen the changes in 1.16.2 aren't very drastic and from my conversation I had with slicedlime it does not sound like it will take any longer than 1 day or 2 tops to complete this part of the rewrite, which I am happy to sacrifice to ensure fellow modders and myself get access to such hooks.

This specific hook is so incredibly tiny that it will have a very low chance of affecting 1.16.2 at all, and it can also just be left behind and added at a later date to 1.16.2 if necessary so that modders on earlier versions can still enjoy the changes.

valoeghese:
@srs-bsns the MultiNoise biome provider is used in the nether rather than the overworld, so that will affect my Nether Biomes PR #6853 more.

valoeghese:
I highly doubt the overworld will move to multi noise, as it would cause large borders between old and new world gen, and the existing system fits the overworld well.

Mysterious-Dev (approved):

Sindarin27 (approved):

ghost:
How could this possibly have broken

TheCurle:
Jenkins' concurrent builds.

PR #7002, opened 2020-07-20
[1.16] Restructuring of the Config system

Senmori:
***This PR is a breaking change.*** 
Just wanted to get that out there before I go any further.

I'd really love feedback on what I could do better or what needs/should to be changed. Thanks!

## Justification
The current Configuration system works, and it supports almost everything that mods can realistically want.
However, it lacks certain features that would be nice and/or useful. There are three(3) main features that either need better support or actual implementation. These features are String support, List support, and custom object support.
Features such as String support without having to use `ConfigValue<String>` instead of just `StringValue` isn't a big deal, but it just makes the code a bit easier to read and understand.
The same thing cannot be said about lists. The current `ConfigValue<List<? super T>>` means you will always lose the type contained in the list. This PR introduces the `ListValue<T>` which will retain it's type.
There is currently no support for custom objects. This isn't Forge's fault, as night-config's support for those things is limited as well.

This PR will be broken down into three main parts. Changes in how configs are registered, the new `ModConfigSpec` class, and custom object support.

## Registration
Mods now register their configs via the same `ModLoadingContext#registerConfig` methods. However, these methods' signatures have changed (all old methods will be broken).
The new methods now allow for mods to a `Function` that will be called *after* their respective `ModConfig` has finished instantiation. This allows mods to register and create their configuration instances in a much cleaner way then previously.
`ModLoadingContext.get().registerConfig(ModConfig.Type.CLIENT, ForgeConfig.Server::new);`
The following are the changed/new register methods.
1. `registerConfig(ModConfig.Type)` will just return the default `ModConfigSpec` that Forge creates. Mods will have to create their config instances using the returned `ModConfigSpec`.
2. `registerConfig(ModConfig.Type, IConfigSpecFactory, Function)`
	* This will use the given `IConfigSpecFactory` to create the specification for the configuration and when the `ModConfig` is finished constructing the `Function` will be called using the completed spec that `IConfigSpecFactory` created.
3. `registerConfig(ModConfig.Type, IConfigSpecFactory, String, Function)`
	* This will do the same as above but the `String` will be the config's file name.

## ModConfigSpec
The new `ModConfigSpec` (name can be changed, I'm not 100% sold on it) is built on top of `AbstractConfigSpec` which simply extends night-config's `ConfigSpec`.
This is good because it allows mods to take advantage of the Night Config API. Yes, the API that the Forge config system has is good enough for almost all cases, however if a mod does need to use the Night Config API they can do that without any extra hassle.
However, this also means that the Forge config API cannot use the `define` methods. This isn't terrible, but it is a bit annoying.
I have kept many of the builder pattern methods the same for ease of use (and because it's super useful).
However, some of the methods did have to change.
For example, to define a generic `ConfigValue<?>` you used to be able to do `define(...)` however that has changed to `defineObject`. In return I have implemented support for most primitives and objects.
To cut down on the length of `ModConfigSpec` I have broken the API down into several interfaces and only implemented the method that *actually* defines(registers) the new configuration entry.
Yes, this means there's more interfaces but it means that the actual `ModConfigSpec` implementation is much cleaner and doesn't have any unnecessary methods in it (i.e. methods that only overload other methods).

Finally, `ForgeConfigSpec` is deprecated. As well, everything in it has been deprecated to let anything using those objects know to change them because they can't be used in `ModConfigSpec`
## Custom Object Support
So, this is the part that I'm not the most confident about because the only way to support custom object serialization/de-serialization was to just give the mods a sub-config to write their data to and hope they do it in a way that Night Config supports.
Custom objects are defined in `ModConfigSpec#_defineMod`. Those objects have to be converted to a format that Night Config supports. The only thing that would provide the most options for mods as well as long-term support is the `Config` option. This `Config` is *not* the actual configuration on file, it is a sub config created using `ConfigSpec` (i.e. `Config#createSubConfig`).
To save/load objects from config needs a converter, this is what `ModObjectConverter` is for (I don't like the name but I can't think of another one, feel free to provide one). These sub configs are then passed into the converter so that the mods can save their object's data to the configuration.
Custom objects are then stored in the config like any other config value. `ModValue` is this config value type (Again I'm not sold on the name).
I chose not to just use `ConfigValue<CustomObject>` because this would mean that there would have to be support for both regular objects AND custom objects in the same class. I believe this would just get messy over time.

**Tests**
There is a test for the new config system. `ForgeConfigTest` tests every value type implemented.
There are two tests for custom objects. One is for a class inside the test itself (`ConfigModObject`) and `BlockPos`. Both tests were successful.

LOG: https://gist.github.com/Senmori/53294dcd4923977de611dd77a6c751fb

The configs are the same but here's links to both of them.
Client Config: https://gist.github.com/Senmori/33409cb88ae6cef08b47ec1c9eeda7db
Server Config: https://gist.github.com/Senmori/72160cc67c1f709eb10ac92f9efbc763

**Thoughts**
I was able to join both a local world and a server without any errors showing up.
I made sure to keep the instantiation of `ForgeConfig` basically the same (e.g. public static final in a static block). To accomplish that I moved the registering of the configs from `ForgeMod` to `ForgeConfig` so I could maintain the `final` modifier.

There is a hard dependency on Guava's `Range`. I don't think this is a deal breaker personally but it could be removed somewhat easily if need be. I chose Guava's `Range` because it was already set up to use. I did not have to implement every detail (i.e. toString showing the correct range, the ability to check if a number is within that range, etc).

### DIFF
--- a/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/NightConfigWrapper.java
+++ b/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/NightConfigWrapper.java
@@ -39,8 +39,8 @@ public class NightConfigWrapper implements IConfigurable {
         this.config = config;
     }
 
-    NightConfigWrapper setFile(IModFileInfo file) {
-        this.file = file;
+    public NightConfigWrapper setFile(IModFileInfo file) {
+        if (this.file == null) this.file = file; // Don't overwrite existing IModFileInfo
         return this;
     }
 
--- a/src/main/java/net/minecraftforge/common/ForgeConfig.java
+++ b/src/main/java/net/minecraftforge/common/ForgeConfig.java
@@ -21,17 +21,27 @@ package net.minecraftforge.common;
 
 import static net.minecraftforge.fml.loading.LogMarkers.FORGEMOD;
 
+import net.minecraftforge.common.config.ModConfigSpec;
+import net.minecraftforge.common.config.RangeType;
+import net.minecraftforge.common.config.values.BooleanValue;
+import net.minecraftforge.common.config.values.DoubleValue;
+import net.minecraftforge.common.config.values.IntValue;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.ModLoadingContext;
 import net.minecraftforge.fml.config.ModConfig;
-import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.LogManager;
 
-import net.minecraftforge.common.ForgeConfigSpec.BooleanValue;
-import net.minecraftforge.common.ForgeConfigSpec.DoubleValue;
-import net.minecraftforge.common.ForgeConfigSpec.IntValue;
 
 public class ForgeConfig
 {
+    static {
+        SERVER = ModLoadingContext.get().registerConfig(ModConfig.Type.SERVER, ForgeConfig.Server::new);
+        CLIENT = ModLoadingContext.get().registerConfig(ModConfig.Type.CLIENT, ForgeConfig.Client::new);
+    }
+    public static void init() { }
+    public static final Server SERVER;
+    public static final Client CLIENT;
+
     public static class Server {
         public final BooleanValue removeErroringEntities;
         public final BooleanValue removeErroringTileEntities;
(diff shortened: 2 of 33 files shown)


Senmori:
More work needs to be done to this. Will reopen when complete.

Issue #6870, opened 2020-06-28
1.16 - How to approach new tags 


Cyborgmas:
Tags need to be created/updated/removed with 1.16, I propose to discuss them here as to what should be done.

My suggestions are here: https://gist.github.com/Cyborgmas/60203cda033297b1f84b2293499bdca5 , comment on what sounds good/bad/terrible and what I missed.

TBoshoven:
I think there are at least some that so far everyone has agreed with. We could start by making a PR to sort those out.

Alpvax:
Re vanilla gold ore tag, I would suggest making the vanilla tag inherit from the forge one, so that modders still add to the forge tag for consistency

Issue #6190, opened 2019-09-26
Duplicate Tags between Forge and Minecraft


pupnewfster:
**Minecraft Version:** 1.14.4

**Forge Version:** 28.1.22

**Description of issue:**

There are tags that are both declared by forge and minecraft.

Exact duplicates:
- `forge:arrows`, `minecraft:arrows`
- `forge:music_discs`, `minecraft:music_discs`

These tags overlap to an extent except the forge variant has sub versions of them as well, but it makes it unclear which one should be used:
- `forge:fences`, `minecraft:fences`
  - `forge:fences/wooden` overlaps with `minecraft:wooden_fences`
- `forge:sand`, `minecraft:sand`. The forge variant inherits from the subvariants colorless and red

LexManos:
In most cases it's probably due to tags being added in vanilla after tags being added in Forge. Or.. Just missing some.
Simplest solution would be to keep both, but have Forge's include vanilla's.
Especially in cases where Forge's are more refined then vanilla's.

pupnewfster:
That makes sense, though it may also make sense then in some places in the code (such as for `FenceBlock` where it checks for `BlockTags.FENCES` then should also be rerouted to the forge tag so that mods that only add to the more refined forge variant of a tag still have the expected vanilla behavior.

tterrag1098:
In those cases we could just override the vanilla tag, as long as the default values don't change, right?

LexManos:
I am not a real fan of tags being used as game logic, but I can see it's uses.
Rather then patch all the massive amount of usages in code, we'd probably just have to inject our tags into the Vanilla one. IF we can make sure that the Forge tag we inject is the intended one. The problem is the side effects could be not what we intend due to Forge's tags being more refined. 
Each would need to be taken on a case, by case.

liach:
In my opinion, a mechanism to merge tags (i.e. two resource locations refer to the same tag) may be helpful to redirect all forge/vanilla usages to the same tag when using two different resource locations to point to a tag of the same purpose. Should be a few easy patches in the tag manager.

pupnewfster:
Not quite sure how much sense that makes (unless it is just for forge/vanilla tags), as mods may use common names for things they want to process a specific way and then the names may overlap. (Unless you mean if the contents are identical then merge them, which at that point may not help as soon as someone adds a single item to one of them.

KingLemming:
[...]

PR #7004, opened 2020-07-21
1.16 forge tags

Cyborgmas:
Adds some basic 1.16 tags following current forge tags. Added netherite ingot to the recipe overrides as well, only lodestone changed. 
Also when running datagen a bunch of overrides were still incorrect so they are fixed now. (no longer, since it was already pushed to forge)


Cyborgmas:
This addresses #6870. Does not answer everything though, the rest should be changed if a need arises for it.

bageldotjpg:
Is the forge tag for Music Discs needed anymore as vanilla now has two, #music_discs and #creeper_drop_music_discs?

pupnewfster:
Shouldn't nether gold ore be in the gold ores tag one way or another? (Not sure of the proper hierarchy of tags for it. Maybe `forge:ores/gold/nether` that gets added to `forge:ores/gold`). Especially given if you think about it the ore itself has no real difference than gold ore besides spawning in the nether, given even though mining it normally drops gold nuggets, if you use silk touch you get the ore which can be smelted for a gold ingot just like normal gold ore.

Cyborgmas:
Okay so did not see the music disk tags. removed the forge one.
Also made forge:ores/gold simply contain #minecraft:gold_ores. 
If something more complex is desirable, I'll adjust

pupnewfster:
Thinking back about how we changed the music disc tag, this may be a good time to re-evaluate https://github.com/MinecraftForge/MinecraftForge/issues/6190 and decide how some of those other "duplicate" tags should be handled.

Cyborgmas:
Yeah it could be good to do. However the wooden fences are defaulted with blocks and i am not sure what i should to replace that. Should i incorporate the possibility to default to a minecraft tag?

TheLMiffy1111:
Should ancient debris be contained in forge:ores/netherite? In my opinion ancient debris feels more like the ore for netherite scrap as it smelts directly to it

KnightMiner:
I agree it should be tagged something like `forge:ores/netherrite_scrap`. Will be a lot easier to write data gen or automatic recipe support if we don't have to special case to avoid ancient debris smelting directly into netherrite.

pupnewfster:
> Yeah it could be good to do. However the wooden fences are defaulted with blocks and i am not sure what i should to replace that. Should i incorporate the possibility to default to a minecraft tag?

Not quite sure on how this should be properly handled as in that issue it seemed that there were multiple differing opinions on how they should be handled. The one other tag the issue listed however of tags that are exact duplicates are `minecraft:arrows` and `forge:arrows`. I propose that we remove the `forge:arrows` tag as it isn't even being actively used by anything.

src/main/java/net/minecraftforge/common/data/ForgeBlockTagsProvider.java
```diff
@@ -74,21 +75,22 @@ public void registerTags()
         func_240522_a_(GRAVEL).func_240534_a_(Blocks.GRAVEL);
         func_240522_a_(NETHERRACK).func_240534_a_(Blocks.NETHERRACK);
         func_240522_a_(OBSIDIAN).func_240534_a_(Blocks.OBSIDIAN);
-        func_240522_a_(ORES).addTags(ORES_COAL, ORES_DIAMOND, ORES_EMERALD, ORES_GOLD, ORES_IRON, ORES_LAPIS, ORES_REDSTONE, ORES_QUARTZ);
+        func_240522_a_(ORES).addTags(ORES_COAL, ORES_DIAMOND, ORES_EMERALD, ORES_GOLD, ORES_IRON, ORES_LAPIS, ORES_REDSTONE, ORES_QUARTZ, ORES_NETHERITE_SCRAP);
         func_240522_a_(ORES_COAL).func_240534_a_(Blocks.COAL_ORE);
         func_240522_a_(ORES_DIAMOND).func_240534_a_(Blocks.DIAMOND_ORE);
         func_240522_a_(ORES_EMERALD).func_240534_a_(Blocks.EMERALD_ORE);
-        func_240522_a_(ORES_GOLD).func_240534_a_(Blocks.GOLD_ORE);
+        func_240522_a_(ORES_GOLD).func_240531_a_(BlockTags.field_232866_P_);
```
ichttt, line 82: Should we keep this? It would make things more consistent as all ores have a tag under forge namespace, but it's also just a copy of the vanilla tag

Cyborgmas: I am not sure as well, I just think it is just a tad easier, avoids having to add to a minecraft tag only for gold ore.

LexManos: We should keep all of our tags. Having ours extend from vanilla's by default.
It keeps things organized, and it makes things not just dissapear in the middle of a MC version.
We could possibly make it deprecated. But honestly it's a data object it doesn't really matter how long we keep it.

src/main/java/net/minecraftforge/common/data/ForgeItemTagsProvider.java
```diff
@@ -54,7 +53,6 @@ public void registerTags()
         super.registerTags();
         filter = this.tagToBuilder.entrySet().stream().map(e -> e.getKey()).collect(Collectors.toSet());
 
-        func_240522_a_(Tags.Items.ARROWS).func_240534_a_(Items.ARROW, Items.TIPPED_ARROW, Items.SPECTRAL_ARROW);
         func_240522_a_(Tags.Items.BONES).func_240534_a_(Items.BONE);
```
LexManos, line 56: Same here, it shouldn't just dissapear, if we want to remove it, mark it for removal in 1.16.2.

src/main/java/net/minecraftforge/common/Tags.java
```diff
@@ -240,10 +242,10 @@ private static void init(){}
         public static final ITag.INamedTag<Item> INGOTS_BRICK = tag("ingots/brick");
         public static final ITag.INamedTag<Item> INGOTS_GOLD = tag("ingots/gold");
         public static final ITag.INamedTag<Item> INGOTS_IRON = tag("ingots/iron");
+        public static final ITag.INamedTag<Item> INGOTS_NETHERITE = tag("ingots/netherite");
         public static final ITag.INamedTag<Item> INGOTS_NETHER_BRICK = tag("ingots/nether_brick");
         public static final ITag.INamedTag<Item> LEATHER = tag("leather");
         public static final ITag.INamedTag<Item> MUSHROOMS = tag("mushrooms");
-        public static final ITag.INamedTag<Item> MUSIC_DISCS = tag("music_discs");
```
LexManos: All of these are binary breaks, so no.

halvors:
Is this ready to be merged? Really annoying not having these tags yet.

halvors:
@LexManos Any reason this is not merged for 1.16.2?

Cyborgmas:
Fixed the conflicts and removed the arrow and musics disks tag. The breaking change should be fine for 1.16.2.

PR #7007, opened 2020-07-22
[1.16] [LivingEvent]: LivingFindAmmoEvent and Forgehook /w patches to inline and fix unruley minecraft crossbow behaviour.

Lanse505:
This PR aims to add a hook event for when bows, crossbows and other shootable items attempt to find ammo using the PlayerEntity#findAmmo method.

To keep with vanilla convention, this event fires after the main ammo checks in the order of:
Hand -> PlayerInventory -> Event -> Creative Check.

This event is aimed at providing "arrow" ammo to the vanilla bow and crossbow.
But it's also applicable to anything calling "findAmmo" on the player.

For my use-case, I will be using this to provide ammo to the player through an equippable itemstack inventory "quiver".
And do alterations on the provided ammo from the quiver, based on player "feats" and "skills". Like increasing knockback power, providing potion effects, etc.

In essence this event provides the ability for mods to provide additional ways/places to provide ammo from.
A few examples is:
  - A Quiver inventory providing Arrows.
  - A mod adding a daily "free arrow" amount.
  - A mod adding an item to provide free arrows at the cost of energy.

A test mod has been provided and tested on both a client world and a server world using a vanilla ender chest as the external storage medium. And it works just fine on both ends.


src/test/java/net/minecraftforge/debug/entity/player/PlayerFindAmmoEventTest.java
```diff
@@ -0,0 +1,40 @@
...
+            ItemStack stack = inventory.getStackInSlot(i);
+            if (event.getAmmoPredicate().test(stack)) {
+                logger.info("Found valid ammo in player Ender Chest");
+                ammo = stack;
+                break;
+            }
+        }
+        if (!ammo.isEmpty())
+            logger.info("Set found ammo to Event");
+            event.setAmmo(ammo);
```
ExpensiveKoala: Is this event.setAmmo(ammo); line supposed to be in the if statement? Otherwise why indent?

src/test/java/net/minecraftforge/debug/entity/player/PlayerFindAmmoEventTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+@Mod("player_find_ammo_event_test")
+@Mod.EventBusSubscriber
+public class PlayerFindAmmoEventTest
+{
+    private static final boolean ENABLE = true;
+    private static final Logger logger = org.apache.logging.log4j.LogManager.getLogger(PlayerFindAmmoEvent.class);
+
+    @SubscribeEvent
+    public static void onPlayerFindAmmo(PlayerFindAmmoEvent event)
```
ExpensiveKoala: The event says it is cancelable and needs to be canceled for it to apply. Why is it not happening here?

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -733,4 +735,11 @@ public static void onCommandRegister(CommandDispatcher<CommandSource> dispatcher
         RegisterCommandsEvent event = new RegisterCommandsEvent(dispatcher, environment);
         MinecraftForge.EVENT_BUS.post(event);
     }
+
+    public static ItemStack onFindAmmo(PlayerEntity playerEntity, ItemStack shootable, Predicate<ItemStack> ammoPredicate)
+    {
```
ExpensiveKoala, line 740: You should add the logic for @cancelable here

src/test/java/net/minecraftforge/debug/entity/player/PlayerFindAmmoEventTest.java
Lanse505: Yes, I've fixed this now to be included inside the if statement.

src/test/java/net/minecraftforge/debug/entity/player/PlayerFindAmmoEventTest.java
Lanse505: I've added the cancellation and cancelled check now.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -733,4 +735,11 @@ public static void onCommandRegister(CommandDispatcher<CommandSource> dispatcher
...
+        PlayerFindAmmoEvent event = new PlayerFindAmmoEvent(playerEntity, shootable, ammoPredicate);
+        MinecraftForge.EVENT_BUS.post(event);
```
ExpensiveKoala, line 777: You aren't doing any checks for if it is cancelled here.

ExpensiveKoala (commented):
Also, small question. Does the test mod result in the ammo being consumed from the ender chest? I don't have my PC set up yet so I can't run the test myself.

Lanse505:
@ExpensiveKoala 
Yes, the test mod causes the arrows to get consumed from the supplied stack from the event.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
Lanse505: That's because the cancellation doesn't change the behaviour in the factory.
The cancellation is for other mods to know if they should be continuing expected behaviour or if they should ignore the event if a stack has been provided by another mod.

I wasn't sure to use the result or cancellation system here, and haven't gotten a response from my usual group of suspects that has yielded a better option/answer.

ExpensiveKoala:
Then my question is should the event be the one to handle ammo consumption? Either similar to how the ItemFishedEvent does it, or by just making the event use a copy of the resulting itemstack.

Lanse505:
I'm personally of the belief it shouldn't, this event is strictly for providing ammo through the findAmmo event.
I don't want this to run wild and handle several different things. 

The consumption of ammo is done by the shootable item.
Usually handled through the "onPlayerStoppedUsing" method.
In the case of BowItem it'd done at the very end by this statement
```
               if (!flag1 && !playerentity.abilities.isCreativeMode) {
                  itemstack.shrink(1);
                  if (itemstack.isEmpty()) {
                     playerentity.inventory.deleteStack(itemstack);
                  }
               }
```

If the forge team deems this event too narrow in scope, then it can be considered adding.
But I'm not adding another event that's outside of the scope of the base of the PR.

Lanse505:
After talking to Gigaherz, I'll be making some alterations to the event in the form of:
- Making it not cancellable

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
Lanse505: This is invalid since cancellation is now gone.

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
```diff
@@ -300,7 +300,18 @@
...
  
-@@ -1956,4 +2000,37 @@
+@@ -1912,6 +1956,10 @@
+                }
+             }
+ 
++            ItemStack ammo = net.minecraftforge.event.ForgeEventFactory.onFindAmmo(this, p_213356_1_, predicate);
++            if (!ammo.func_190926_b())
++               return ammo;
++
```
diesieben07: This patch should be two lines: No line break on the if and no white-space changes.

src/main/java/net/minecraftforge/event/entity/player/PlayerFindAmmoEvent.java
```diff
@@ -0,0 +1,54 @@
...
+import net.minecraftforge.eventbus.api.Cancelable;
+
+import javax.annotation.Nonnull;
+import java.util.function.Predicate;
+
+/**
+ * PlayerFindAmmoEvent is fired when a shootable item tries to find ammo on the player.
+ * This event is fired whenever a shootable item attempts to find ammo using {@link PlayerEntity#findAmmo(ItemStack)}.<br>
+ * <br>
+ * This event fires after the original vanilla order, aka: Hand -> PlayerInventory -> This.<br>
```
diesieben07:
```suggestion
 * This event fires after the original vanilla order, aka: Hand -> PlayerInventory -> This event.<br>
```

src/main/java/net/minecraftforge/event/entity/player/PlayerFindAmmoEvent.java
```diff
@@ -0,0 +1,54 @@
...
+ * <br>
+ * {@link #shootable} contains the {@link net.minecraft.item.ShootableItem} in it's ItemStack form.
+ * {@link #ammoPredicate} contains the {@link ShootableItem#getAmmoPredicate()} instance.<br>
+ * <br>
+ * This event is not {@link net.minecraftforge.eventbus.api.Cancelable}.<br>
+ * <br>
+ * This event does not have a result. {@link net.minecraftforge.eventbus.api.Event.HasResult}<br>
```
diesieben07:
```suggestion
 * This event does not have a {@linkplain net.minecraftforge.eventbus.api.Event.HasResult result}<br>
```

diesieben07 (changes requested):
This is incomplete and your test mod even shows this. Consuming of the ammo needs to be handled as well, as vanilla only tries to find the used ammo in the player inventory.
You can see this with your test mod by putting one arrow in the ender chest: It will never be deleted.

src/main/java/net/minecraftforge/event/entity/player/PlayerFindAmmoEvent.java
Lanse505: This should have been resolved now

src/main/java/net/minecraftforge/event/entity/player/PlayerFindAmmoEvent.java
Lanse505: This should have been resolved now

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
Lanse505: This should have been resolved now in theory?

Lanse505:
Okey so discussion time.
I've been talking a bit to gigaherz and a few others describing the issue at hand about stack consumption in inventories.

So there are a few possible solutions:

Solution 1 (Messier Option):
We can split the event into - PlayerFindAmmoEvent and PlayerConsumerAmmoEvent

The first one would work as it does now.
The second would work by passing in some contexts, and the ammo stack.
Mods would then check their inventories against the stack and if found, handle the decrement/clear the slot.

This would in theory keep more with vanilla behaviour.

Solution 2 (Cleaner Option but 'Hackier'):
We could add a Consumer<ItemStack> to the PlayerFindAmmoEvent and change the return type of findAmmo to Pair<ItemStack, Consumer<ItemStack>>, this would involve a larger patch to CrossbowItem and BowItem in theory. But would provide a more "elegant" solution, since it'd only involve 1 event, it'd work for pretty much any inventory, and would full-fill the original purpose.
This would in theory not keep with vanilla behaviour, but would provide better modding support.

diesieben07:
Yeah, it's ugly. Because `findAmmo` is not designed to do this.
The two-event option is probably the best since it reflects how vanilla operates. We just need a way to pass information from the first to the second event, since it's no longer obvious where the ammo came from.

Lanse505:
Yea, that's the issue that came up in discussion.
Basically the issue falls to the point of: 
- finding ammo is done in one place
- consuming ammo is done somewhere else.

So either we'd have to split the event and provide the ammo stack on the consume side but with no container context.
Which means additional logic and searching for every mod involved which is not very optimal.

Or we could create a unified object that would get passed/returned, issue with this would be that it'd change the return type which is a bit "yikes", since it'd involve 3 small/medium method patches in theory to "core" bow logic to get it working.

Question is if the second is done or a mix of the two are done, how should it be gone about.
Either it could return a Pair object of ItemStack (ammo) and Consumer<ItemStack> (ammo consumer).
Or we could create a custom object like has been suggested in pr talks with Gigaherz like "IAmmoProvider".
Which has a few methods like get(); and remove(); / consume();.

KingLemming:
Hrm...this is interesting as I currently use capabilities for this. Unfortunately I think with my Volley enchant, I'll still have to.

I currently work around this with the arrow nock/ arrow loosed events, but I basically hijack the entire process. As you've noticed, the find/consume is split in a sub-optimal way.

Lanse505:
Indeed, and either way the solution is put forth it either involves a "larger" patch or two kind of sub-optimal events.
I think personally that the most optimal solution for mod support would be to do the "larger" patch, which at the moment I've counted to about 4-5 smaller patches that need to be done in theory. That is if we only want this behaviour to apply to players.

findAmmo overall is a part of PlayerEntity, LivingEntity and MonsterEntity. Which means if we want to go this way of changing the return signature and doing the patches. Currently with the 4-5 small patches it'd apply only to player.
I think to get it equal between all instances of findAmmo it'd be quite a few more "smaller" sized patches in theory.

However I don't think that's neccesary to patch the entire chain in theory since the monster entity findAmmo call is rather simple and just gives them an arrow pretty much.

So the approach with more patches gives modders more power, since we'd be able to handle both the providing of ammo as well as the consumption of ammo for the bow/crossbow.

And with Arrow Nock and Arrow Volley we'd be able to control the rest of the bow chain.

This approach would involve the following patches in theory:
- Changing "findAmmo" in PlayerEntity to return either a IAmmoProvider object with a "get" and a "consume" method
- Patching onItemRightClick and onPlayerStoppedUsing in BowItem
- Patching onItemRightClick, hasAmmo and func_220023_a in CrossbowItem

All of the above methods and patches are only called in the player chain of calls so it should be safe to change them in theory, though I'd have to look more into it and attempt to poke around with it.

Once again this is probably the most invasive approach in terms of patches.
But it'd allow us to solve the issue of the find/consume split being in such a disasterous manner.

The other solution of splitting the event is the most lightweight option and requires the least number of patches, however it does involve the fact that people who provide ammo would all have to subscribe to the PlayerConsumerAmmoEvent and loop through all of their inventories to check if the stack is theirs.

P.s. I fully acknowledge that changing the method signature is a big nono and that that approach most likely won't end-up being done, but I'm just airing the question.

Lanse505:
Another option that Gigaherz just brought up in conversation to me was to create a new ForgeHook and use that as the replacement since that wouldn't change the method signature on PlayerEntity.findAmmo, which I think could work really well.
Hadn't even thought of the possibility to add a ForgeHook tbh

Lanse505:
I've converted this PR and the 1.15 PR to drafts for now, while I work out a good solution.
Currently after talking to a few people, I'm considering going the ForgeHook approach, to provide a replacement hook for the findAmmo in PlayerEntity.

Lanse505:
To pre-emptively answer a few questions:

Q: "Why is the two methods deprecated in CrossbowItem and is there a point to inlining the methods".
A: Yes, It inlines the split behaviour from two methods into a single unified method. This means we can properly let the event handle providing and consumption providing logic. Since that's normally split across 2 methods in CrossbowItem.

Q: "Why isn't the new ForgeHook merging the two methods, optimized in it's behaviour"
A: This is because I don't want to alter existing vanilla behaviour since it works and isn't poorly optimized to begin with really. Don't fix something that ain't broke!

Lanse505:
Debugging some wierd crossbow behaviour atm.

patches/minecraft/net/minecraft/item/BowItem.java.patch
```diff
@@ -1,17 +1,23 @@
...
 +         int i = this.func_77626_a(p_77615_1_) - p_77615_4_;
 +         i = net.minecraftforge.event.ForgeEventFactory.onArrowLoose(p_77615_1_, p_77615_2_, playerentity, i, !itemstack.func_190926_b() || flag);
 +         if (i < 0) return;
 +
-          if (!itemstack.func_190926_b() || flag) {
-             if (itemstack.func_190926_b()) {
-                itemstack = new ItemStack(Items.field_151032_g);
++         if (!pair.getKey().func_190926_b() || flag) {
++            if (pair.getKey().func_190926_b()) {
++               pair = org.apache.commons.lang3.tuple.Pair.of(new ItemStack(Items.field_151032_g), pair.getValue());
```
diesieben07: This whole thing would be a lot cleaner if you didn't change those last 3 lines unnecessarily. The vanilla logic is fine here because:
1. `itemstack` is equal to `pair.getKey()` here.
2. `pair.getKey()` is never accessed beyond this point, so this is actually _wrong_ (because `itemstack` is used below).

patches/minecraft/net/minecraft/item/BowItem.java.patch
```diff
@@ -26,18 +32,38 @@
...
+-                  if (itemstack.func_190926_b()) {
+-                     playerentity.field_71071_by.func_184437_d(itemstack);
+-                  }
++                   pair.getValue().accept(itemstack);
++                   if (false) {
++                      itemstack.func_190918_g(1);
++                   if (itemstack.func_190926_b()) {
++                      playerentity.field_71071_by.func_184437_d(itemstack);
++                   }
++                   }
```
diesieben07: The `if (false)` hack is useless if it makes the patch larger. Just removing the code is the best bet here I think.

patches/minecraft/net/minecraft/item/CrossbowItem.java.patch
```diff
@@ -0,0 +1,39 @@
...
++      if (f >= 1.0F && !func_220012_d(p_77615_1_) && net.minecraftforge.common.ForgeHooks.handleCrossbowCharging(pair, p_77615_1_, p_77615_3_)) {
+          func_220011_a(p_77615_1_, true);
+          SoundCategory soundcategory = p_77615_3_ instanceof PlayerEntity ? SoundCategory.PLAYERS : SoundCategory.HOSTILE;
+          p_77615_2_.func_184148_a((PlayerEntity)null, p_77615_3_.func_226277_ct_(), p_77615_3_.func_226278_cu_(), p_77615_3_.func_226281_cx_(), SoundEvents.field_219610_bB, soundcategory, 1.0F, 1.0F / (field_77697_d.nextFloat() * 0.5F + 1.0F) + 0.2F);
+@@ -82,6 +83,8 @@
+ 
+    }
+ 
++   /**@deprecated Forge: This method has been inlined to get control over the ammo consumption chain, Use {@link net.minecraftforge.common.ForgeHooks#handleCrossbowCharging(org.apache.commons.lang3.tuple.Pair, ItemStack, LivingEntity)}*/
++   @Deprecated
```
diesieben07, line 33: Do we need to deprecated private methods?

patches/minecraft/net/minecraft/item/CrossbowItem.java.patch
```diff
@@ -0,0 +1,39 @@
...
+    private static boolean func_220021_a(LivingEntity p_220021_0_, ItemStack p_220021_1_) {
+       int i = EnchantmentHelper.func_77506_a(Enchantments.field_222192_G, p_220021_1_);
+       int j = i == 0 ? 1 : 3;
+@@ -107,6 +110,8 @@
+       return true;
+    }
+ 
++   /**@deprecated Forge: This method has been inlined to get control over the ammo consumption chain, Use {@link net.minecraftforge.common.ForgeHooks#handleCrossbowCharging(org.apache.commons.lang3.tuple.Pair, ItemStack, LivingEntity)}*/
++   @Deprecated
```
diesieben07, line 42: Do we need to deprecated private methods?

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1160,6 +1166,92 @@ public static synchronized void updateBurns()
...
+    {
+        if (stack.getItem() instanceof ShootableItem)
+        {
+            ShootableItem shootable = (ShootableItem) stack.getItem();
+            Pair<ItemStack, Consumer<ItemStack>> findAndConsume = ForgeEventFactory.onFindAmmo(shooter, stack, shootable.getAmmoPredicate());
+            if (!findAndConsume.getLeft().isEmpty())
+                return findAndConsume;
+            if (shooter instanceof PlayerEntity)
+            {
+                // Vanilla Player behaviour
```
diesieben07, line 1198: Should probably link to where this logic comes from.

src/main/java/net/minecraftforge/event/entity/living/LivingFindAmmoEvent.java
```diff
@@ -0,0 +1,65 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.ShootableItem;
+
+import javax.annotation.Nonnull;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+
+/**
+ * LivingFindAmmoEvent is fired when a shootable item tries to find ammo on the player.
+ * This event is fired whenever a shootable item attempts to find ammo using {@link PlayerEntity#findAmmo(ItemStack)}.<br>
```
diesieben07: This is no longer true.

src/main/java/net/minecraftforge/event/entity/living/LivingFindAmmoEvent.java
```diff
@@ -0,0 +1,65 @@
...
+ * <br>
+ * This event fires after the original vanilla order, aka: Hand -> PlayerInventory -> This event.<br>
```
diesieben07: This is no longer true.

src/main/java/net/minecraftforge/event/entity/living/LivingFindAmmoEvent.java
```diff
@@ -0,0 +1,65 @@
...
+ * <br>
+ * {@link #shootable} contains the {@link net.minecraft.item.ShootableItem} in it's ItemStack form.
+ * {@link #ammoPredicate} contains the {@link ShootableItem#getAmmoPredicate()} instance.<br>
+ * <br>
+ * This event is not {@link net.minecraftforge.eventbus.api.Cancelable}.<br>
+ * <br>
+ * This event does not have a result. {@linkplain net.minecraftforge.eventbus.api.Event.HasResult}<br>
+ * <br>
+ * This event is fired from {@link LivingEntity#findAmmo(ItemStack)}.<br>
```
diesieben07: No longer true and also redundant with the above.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -21,10 +21,13 @@
 
 import java.io.File;
 import java.util.*;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
+
```
diesieben07: Unnecessary whitespace change.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -66,7 +69,6 @@
 import net.minecraft.util.ResourceLocation;
 import net.minecraft.util.SoundCategory;
 import net.minecraft.util.SoundEvent;
-import net.minecraft.util.math.AxisAlignedBB;
```
diesieben07, line 69: Why?

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -77,7 +79,6 @@
 import net.minecraft.world.IWorld;
 import net.minecraft.world.World;
 import net.minecraft.world.server.ServerWorld;
-import net.minecraft.world.WorldSettings;
```
diesieben07, line 80: Why?

diesieben07 (changes requested):

patches/minecraft/net/minecraft/item/CrossbowItem.java.patch
Lanse505: I was told to throw a deprecated on them in-case someone was using them :P 
If this is wrong, then I can remove the deprecated and comments ^^

patches/minecraft/net/minecraft/item/CrossbowItem.java.patch
Lanse505: I was told to throw a deprecated on them in-case someone was using them :P 
If this is wrong, then I can remove the deprecated and comments ^^

patches/minecraft/net/minecraft/item/BowItem.java.patch
Lanse505: This should be resolved now?

patches/minecraft/net/minecraft/item/BowItem.java.patch
Lanse505: This should be resolved now?

src/main/java/net/minecraftforge/event/entity/living/LivingFindAmmoEvent.java
Lanse505: This should be resolved now?

src/main/java/net/minecraftforge/event/entity/living/LivingFindAmmoEvent.java
Lanse505: This should be resolved now?

src/main/java/net/minecraftforge/event/entity/living/LivingFindAmmoEvent.java
Lanse505: This should be resolved now?

Lanse505:
@diesieben07 That should be it?

diesieben07 (approved):
The crossbow patch is still stupid ugly, but it's necessary for this functionality unfortunately.

Lanse505:
After lots of discussion and between me, and others while trying to adress some concernes raised by Lex in terms of how it functioned before. 

The suggested implementation can't really function safely as an api for modders, since syncing of both the found itemstack for the event and the consumer to a capability on the item can't be done in an acceptable way.

If anyone else wants to take an attempt at this PR feel free to in the future.
And if you find this message be aware of these pitfalls:
- The event should only be called once, and since logic is split across methods you'll need a way of syncing the consumer and the found ammo, this in a manner that is safe for both client and server-side.
- It needs to keep in mind that ArrowNock needs to be considered for removal if implemented, since it fills no real function.
- It needs to also keep in mind that interaction with the ArrowLoose event needs to be perserved and maybe even extended to support a pre and post state, for cancelling the whole logic chain and replacing the logic. And if the post state would be for final changes the arrow entity object.

PR #7009, opened 2020-07-22
Re-implemented get/remove/add methods for LootEntry's

Yunus1903:
Originally there were methods to get/remove/add `LootEntry`'s from a `LootPool`. (https://gist.github.com/Yunus1903/9dd6d14ca5f5b4a88c1e69cc8f14be10)

In 1.14 Mojang removed the names of the `LootEntry`'s so there was no more a way to use these methods. (They were commented out in 1.15 and fully removed in 1.16)

This PR adds get/remove/add methods for 4 `LootPoolEntryType`'s:
- `"item"`
- `"loot_table"`
- `"dynamic"`
- `"tag"` 

If/when everything in this PR looks good I will make a 1.15 version of this PR. If wanted I can also make a 1.14 version.

### DIFF
--- a/patches/minecraft/net/minecraft/loot/DynamicLootEntry.java.patch
+++ b/patches/minecraft/net/minecraft/loot/DynamicLootEntry.java.patch
@@ -1,0 +1,13 @@
+--- a/net/minecraft/loot/DynamicLootEntry.java
++++ b/net/minecraft/loot/DynamicLootEntry.java
+@@ -26,6 +26,10 @@
+       p_216154_2_.func_216034_a(this.field_216166_h, p_216154_1_);
+    }
+ 
++   public ResourceLocation getName() {
++      return field_216166_h;
++   }
++
+    public static StandaloneLootEntry.Builder<?> func_216162_a(ResourceLocation p_216162_0_) {
+       return func_216156_a((p_216164_1_, p_216164_2_, p_216164_3_, p_216164_4_) -> {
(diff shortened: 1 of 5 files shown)


diesieben07:
Why are the modifying methods needed? I think the idea is that things should be data-driven. To modify loot tables they can either be overwritten or modified using global loot modifiers.

Yunus1903:
> Why are the modifying methods needed? I think the idea is that things should be data-driven. To modify loot tables they can either be overwritten or modified using global loot modifiers.

Well first of all these methods already existed before 1.14 and `LootTable` still has methods to modify `LootPool`'s.
I think it would be nice and useful to be able to add entry's trough code to existing vanilla `LootPool`'s.

GirafiStudios:
Isn't the Global Loot Modifiers only a replacement for the LivingDropsEvent & BlockDropsEvent?

diesieben07:
> Isn't the Global Loot Modifiers only a replacement for the LivingDropsEvent & BlockDropsEvent?

Global loot modifiers lets you apply a modifier to all loot tables. So no, it is not.
> I think it would be nice and useful to be able to add entry's trough code to existing vanilla LootPool's.

Why? Minecraft in general and Forge are moving to a data-driven system.

Yunus1903:
> Why? Minecraft in general and Forge are moving to a data-driven system.

I do see some useful use cases in this.  Also if the rest is there why not make it more complete.

diesieben07:
"Just because" is not a valid argument.
If you do see some useful use cases, please provide some as example and why they cannot be done in a data-driven way.

Yunus1903:
I'll close this as it's not needed >.>

PR #7010, opened 2020-07-23
[1.15] [Tag] Add a new Tag for fluid.

FledgeXu:
Add a new Tag for fluid, this tag will improve the compatibility about Fluids.

### DIFF
--- a/src/generated/resources/data/forge/tags/fluids/fluid.json
+++ b/src/generated/resources/data/forge/tags/fluids/fluid.json
@@ -1,0 +1,9 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:water",
+    "minecraft:flowing_water",
+    "minecraft:lava",
+    "minecraft:flowing_lava"
+  ]
+}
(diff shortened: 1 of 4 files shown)


TeamSpen210:
What purpose would this serve exactly? All fluids are, well, fluids. If you want to check for the presence of one, just check that it's not `empty`. If you want to see if it's vanilla you could check the namespace.

marchermans:
@FledgeXu What purpose does this PR bring you?
You added tags for Fluids. Cool but why do you need it.
What is you use case? Why did you only add a default Forge Tag that tracks all fluids......
Additionally you did not provide any documentation, examples and test mods to verify the changes both now that you created a PR, as well as in the future.

Even though this PR is rather trivial, a lot of things are not ready for this to be even considered:

1. Consider more use-cases (more tags maybe?). In general I do not see a good use case since you can use the fluid registry for this.
2. Fix your formatting.
3. Add a lot more documentation
4. Add a test mod
5. Update your initial PR message, with points 1) 3) and 4).

FledgeXu:
You are right. I should rethink this pr.

PR #7123, opened 2020-07-23
[1.16] Re-introduce "outdated" notification on Mods button in main menu

diesieben07:
1.12 and earlier showed the emerald "outdated" icon on the Mods button in the main menu when any mod is outdated.
This was removed in 1.13.

### DIFF
--- a/src/main/java/net/minecraftforge/fml/client/ClientModLoader.java
+++ b/src/main/java/net/minecraftforge/fml/client/ClientModLoader.java
@@ -56,6 +56,7 @@ import net.minecraftforge.eventbus.api.Event;
 import net.minecraftforge.fml.BrandingControl;
 import net.minecraftforge.fml.LoadingFailedException;
 import net.minecraftforge.fml.LogicalSidedProvider;
+import net.minecraftforge.fml.ModList;
 import net.minecraftforge.fml.ModLoader;
 import net.minecraftforge.fml.ModLoadingStage;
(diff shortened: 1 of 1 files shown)


src/main/java/net/minecraftforge/fml/client/ClientModLoader.java
```diff
@@ -138,7 +139,12 @@ private static void finishModLoading(Executor executor)
 
     public static VersionChecker.Status checkForUpdates()
     {
-        return VersionChecker.Status.UP_TO_DATE;
+
+        boolean anyOutdated = ModList.get().getMods().stream()
+                .map(VersionChecker::getResult)
+                .map(result -> result.status)
+                .anyMatch(status -> status == VersionChecker.Status.OUTDATED || status == VersionChecker.Status.BETA_OUTDATED);
+        return anyOutdated ? VersionChecker.Status.OUTDATED : null;
```
ichttt, line 147: What's the reason behind chaning the default from `VersionChecker.Status.UP_TO_DATE` to `null`

diesieben07: Because `UP_TO_DATE` is not necessarily right. If some mods are `AHEAD` or `BETA` then we certainly aren't up to date.
The reason for choosing `null` here is that that's what previous versions did.

diesieben07:
The LTS Backport label was intentional, it serves as "request to backport" (see description of the label).

PR #7173, opened 2020-07-25
[1.15] Re-added PlayerEvent.NameFormat

Tschipp:
1.15 Backport of #6992 
Since it seems that `PlayerEvent.NameFormat` wasn't ported to more recent versions of Forge, I have reimplemented it here.
The behavior is identical to how it used to work:
- The `PlayerEvent.NameFormat` is fired properly and can change the display name of a player.
- This affects the name tag in the world and in the chat.
- This does not affect the name in the player list (tab). This is intended and how it used to work.
- `EntityPlayer#refreshDisplayName()` needs to be called to re-fire the `PlayerEvent.NameFormat` event.

Changes:
- Didn't readd `EntityPlayer#getDisplayNameString()`. Instead, the `displayname` field is now set in `EntityPlayer#getDisplayName()`

- This PR also contains a test mod

This PR should be non-breaking, for all those that used `PlayerEvent.NameFormat` in their mods (but didn't notice that it is never fired).

### DIFF
--- a/patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
@@ -451,10 +451,10 @@
  
     public ITextComponent func_145748_c_() {
 -      ITextComponent itextcomponent = ScorePlayerTeam.func_200541_a(this.func_96124_cp(), this.func_200200_C_());
-+      if (this.displayname == null) this.displayname = net.minecraftforge.event.ForgeEventFactory.getPlayerDisplayName(this, this.func_200200_C_().func_150261_e());
++      if (this.displayname == null) this.displayname = net.minecraftforge.event.ForgeEventFactory.getPlayerDisplayName(this, this.func_200200_C_());
 +      ITextComponent itextcomponent = new StringTextComponent("");
 +      prefixes.forEach(e -> itextcomponent.func_150257_a(e));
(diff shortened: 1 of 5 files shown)


diesieben07 (changes requested):
I think you should expose the original `ITextComponent` in the event as well and allow setting it as well. This will require some deprecation of the methods that use `String`.

Tschipp:
I thought about it, but decided against it, simply for the reason that I would have to make different names for the `ITextComponent` methods, and those names wouldn't be used in the 1.16 version anymore (at least how my PR stands right now). Thus when updating from 1.15 to 1.16 the modders have to either rename the method (So from `getDisplaynameComponent` to `getDisplayName`) Or they just account for the return type of those methods no longer being a `String` but a `ITextComponent`. What do you think is the better approach?

diesieben07:
Yes, but this prevents people from doing proper formatting.

Tschipp:
That is true. Though they could always use `ChatFormatting` for color codes and styles I guess. What other things could be customized? Click events don't really make sense in this context.

diesieben07:
Raw use of text formatting codes is ugly and deprecated.
`ITextComponent` would also allow using for example `TranslationTextComponent`.

Tschipp:
Yeah, that makes sense. Alright I'll make an augment of this PR and include methods for `ITextComponent`s

Tschipp:
Alright, changed it

diesieben07 (approved):

PR #7174, opened 2020-07-25
Fix/rails rotation

Raycoms:
Vanilla was missing two cases in their switch statement for the rails and powered rails rotation.

This PR adds the remaining two cases to fix the rotation bug.


marchermans (approved):

diesieben07:
Marked as WIP because 1.16 should be the primary target, 1.15 is LTS and should receive a backport. Considering this bug still exists in 1.16, please make a 1.16 version as well.

Raycoms:
Good thing, I found a better way to handle this too. Will make a 1.16 PR now too and update this one here.

diesieben07 (approved):

LexManos:
Alright to reproduce this you need to load the structure twice, it'll have the rail rotated 90 degrees.
However, for future reference, the comment should be on the same line if its small.

Issue #6073, opened 2019-08-21
Expose hooks for axe log stripping


waitingtocompile:
Currently log stripping behaviour is determined by AxeItem.BLOCK_STRIPPING_MAP, which is a protected static final field. Some sort of hook for registering new blocks and/or letting individual blocks define their own stripping behaviour would be useful.

williewillus:
You can just do this in your own block's onBlockActivated method, or use PlayerInteractEvent

TeamSpen210:
That’s true, but it’s rather ad hoc. Each block would then need to replicate the bit of logic in the axe item, and it’s not introspectable - if you wanted to make a “log stripping machine” for example it’d have no choice but to place a block, simulate a right click, post the event maybe and then pick up the block after.

Draco18s:
> it’d have no choice but to place a block

Good luck doing that and not bollocksing something up, too. I have to do a similar thing with a machine to determine valid item stack insertion, and *use an entire god damn dimension* to avoid...
- overwriting blocks in the world (e.g. breaking the bedrock floor/ceiling)
- letting the spawned block be visible (The End?)
- item duplication (if the player can find the spawned block...)
etc.

waitingtocompile:
While handling things on the log's activation is *doable*, I do agree with Draco and Teamspen that it potentially intoduces some problems with things like the theoretical "log stripping machine", or something similar. Not to mention other problems, where a mod might add some special log stripping tool that, while able to handle vanilla logs fine, doesn't play nicely with your mod's way of implementing log stripping. That's not to mention that it doesn't seem to me unresonable for forge to introduce, for instance, a recipe hook that let's you define log stripping in json, and including that in the mapping that vanilla uses, or some other way of registering that functionality.

Shadows-of-Fire:
All it needs is a change from protected to public and from ImmutableMap to HashMap.  Maybe a further change from Map<Block, Block> to IBlockState or IRegistryDelegate, but nothing beyond that.  Custom axes can simply call Items.(any axe item).onItemUse to replicate the logic.

You can do it pretty simply right now in code with an AT,

waitingtocompile:
While that's not *too* hard I do still think it would be better to actually expose some sort of proper system.

Alpvax:
Making it a recipe type would be the best approach IMO. Would allow for more modpack customisation.

Shadows-of-Fire:
Thats not within the specification of IRecipe.

[...]

PR #7176, opened 2020-07-26
[1.16] IForgeBlock#getToolModifiedState (Handling Axe, Hoe, and Shovel Right Click)

ChampionAsh5357:
This deals with an issue provided by #6073. The main purpose of this PR is to create a registry based hook system for programmers to add their own axe stripping maps to the game. Originally, this was a simple registry hook for only axes. It has been currently reworked to be included within IForgeBlock and used within the axe, hoe, and shovel.

Currently, the method check makes the mappings for each tool public and checks to see if its ToolType is on the item. If so, it tries to apply the current mapping to the block. If it fails and there are no other tool types, it returns null. If it fails and there is another tool type, it moves onto the next method to check the other tool types. This allows multitools to function correctly as well.


diesieben07:
_If_ we do a registry system like this it should be data-driven, i.e. loaded from JSON and accessible to datapacks.
However I question if a complex system like this is really needed for such a simple thing. What's wrong with a method `IForgeBlock#getAxeStrippedState(World, BlockPos, BlockState, PlayerEntity, ItemStack)`?

Lanse505:
I honestly believe that what diesieben is suggesting would be a simpler and better solution than a full blow axe stripping registry (In terms of the IForgeBlock#getAxeStrippedState). 

The only concern I there though is that this discussion should really be extended to the rest of the hardcoded tool conversions. Axe Stripping, Hoe Tilling ande Shovel Pathing.

I know when I tackled it in my own mod I did a custom recipe type for those types and then if none was found I defaulted ot vanilla behaviour, however that isn't a good forge-level solution.

TLDR;
A method in IForgeBlock would be the best solution **if** we were only talking about Axe Stripping, however personally I think this should be extended to not **only** target axe stripping but also target the other "tool internal" interactions. 

P.s. Note I'm just a pleb not affiliated with forge.

diesieben07:
> however personally I think this should be extended to not only target axe stripping but also target the other "tool internal" interactions

The options here are really either multiple methods in `IForgeBlock` or _one_ method which takes some kind of "mode" parameter that indicates "axe stripping", "shovel flatten", etc. For Hoes there is already `UseHoeEvent`, but we might want to include it here as an alternative means to achieve it?

pupnewfster:
Also I feel like if something is being implemented for the axe lookup map while we are at it, something should be implemented for the hoe, and shovel lookup maps as if memory serves they are basically the same concept.

Edit: Upon closer inspection it seems lanse has already suggested this. (Missed that fact when initially skimming it)

ChampionAsh5357:
> The options here are really either multiple methods in `IForgeBlock` or _one_ method which takes some kind of "mode" parameter that indicates "axe stripping", "shovel flatten", etc. For Hoes there is already `UseHoeEvent`, but we might want to include it here as an alternative means to achieve it?

> _If_ we do a registry system like this it should be data-driven, i.e. loaded from JSON and accessible to datapacks.
> However I question if a complex system like this is really needed for such a simple thing. What's wrong with a method `IForgeBlock#getAxeStrippedState(World, BlockPos, BlockState, PlayerEntity, ItemStack)`?

So in one scenario, it would be accessed through IForgeBlock and referenced in IForgeBlockState for shovel pathing, axe stripping, and item tilling. We can pass in either the suggested parameters or ItemUseContext. AxeItem would need a bit more work since it's basically a property mapper doing a Block -> Block system compared to a Block -> BlockState mapping. However, that is simple to implement and fix. We could reference the tool used through their tool type by default, or we can create a tag that holds the viable tools that can affect the block when setting. To be able to default back to the original map lookup, it would probably either use a ternary operator if we want to specify the original mappings in class. This could be done by returning Blocks.AIR.getDefaultState() if wanting to default to vanilla or null to cancel altogether. If not and specifically in IForgeBlock, the lookup maps would need to be public. This would allow everything to be referenced in one place as to either return the state to be used or null. UseHoeEvent would just be replaced by returning null within the IForgeBlock method as it would return the same thing. I do believe it should be replaced with a global event within IForgeBlock to disable normal vanilla behavior.

On the other hand, it would become an json file to supply the mappings instead through data packs. If this were the path, the system would be highly limited at the start since it would need to incorporate blockstates and property copying, something not normally handled. However, the system could be extended to include furnace fuel anything other that can be mapped with a one to one system.

Personally, the first idea is better in a much greater sense. I was thinking a bit too limited in scope and forgot about the extensible systems implemented on classes. I will start implementation on the first one; however, I would like to know whether the lookup maps should be made public or just handled internally through the AIR/null solution and if the event should allow for global disabling or removed altogether.

ChampionAsh5357:
Reworked using IForgeBlock with the suggested parameters from diesieben07. I used the access transformer to make the original maps public to be used in the function such that there wouldn't need to be any fancy workarounds with determining their final result. The function also allows multitools to use the different mappings.

I haven't removed or done anything with UseHoeEvent yet, I'm still not sure whether reusing or removing entirely is the best option.

Lanse505:
So what I mentioned to @pupnewfster in a DM was the possibility of doing something of the lines of:
Having a defaulted method in IForgeBlock called something along the lines of "getToolModifiedState" (Like you suggested).

Which would fire an event like "BlockToolInteractEvent", if the event returned the same blockstate that was passed in i.e. the original blockstate, then it'd default to vanilla behaviour.

What this would do is accomplish a couple of things in theory.
- It'd keep with vanilla behaviour and either require minimal to no patching on the tools themselves side.
- It'd provide an event hook for mods wanting to alter vanilla/other mods blocks interaction with tools.
- It'd provide a in-block method for specifying tool interaction behaviour.
- It'd remove the specific need for UseHoeEvent.

The event and method would take in something like the:
- User
- Original Blockstate
  - {
    - Position
    - Hit Side
    - Etc
    }
- Or
  - BlockRayTraceResult
- Tool-Type

If you feel that makes sense?

TLDR;
Mix IForgeBlock method with an event and default to providing the input blockstate.

ChampionAsh5357:
> So what I mentioned to @pupnewfster in a DM was the possibility of doing something of the lines of:
> Having a defaulted method in IForgeBlock called something along the lines of "getToolModifiedState" (Like you suggested).
> 
> Which would fire an event like "BlockToolInteractEvent", if the event returned the same blockstate that was passed in i.e. the original blockstate, then it'd default to vanilla behaviour.
> 
> What this would do is accomplish a couple of things in theory.
> 
> * It'd keep with vanilla behaviour and either require minimal to no patching on the tools themselves side.
> * It'd provide an event hook for mods wanting to alter vanilla/other mods blocks interaction with tools.
> * It'd provide a in-block method for specifying tool interaction behaviour.
> * It'd remove the specific need for UseHoeEvent.
> 
> The event and method would take in something like the:
> 
> * User
> * Original Blockstate
>   
>   * {
>     
>     * Position
>     * Hit Side
>     * Etc
>       }
> * Or
>   
>   * BlockRayTraceResult
> * Tool-Type
> 
> If you feel that makes sense?

I understand that. I think what I've done so far basically supports the gist of it. The only difference is that it would add an event within the IForgeBlockState call either before or after to determine the final block state before defaulting into vanilla behavior. It seems like a clear implementation. I would just pass in the ItemUseContext into the function to be called and check as you said if the blockstate changed. Determining whether its called before of after is the question. Before only takes the current blockstate into account while after can check both the current and resulting state instead. Also, would it be better to call this event in IForgeBlockState or IForgeBlock is another question. Either way, it allows modders to alter the vanilla result as you said.

UseHoeEvent has already been replaced as this point. I haven't removed the method since it does allow user to damage the current item and return a successful use without changing the block. The question there lies in whether we care if the action returns successful when the block hasn't changed. If we want that 'result', we can apply it to all classes such that it damages the item and returns a successful result. If we don't we can just remove the hook altogether. Of course it will probably be deprecated instead and removed in the next minor version if we do go down that route.

Lanse505:
I'd say fire it before vanilla behaviour.

So basically it'd work:
Mod Override Behaviour -> super call -> Event -> Vanilla

Also I think you'd need to add it to both IForgeBlockState and IForgeBlock.
Since IForgeBlockState methods tends to call state -> block -> block method.

So you have one "generic" and one blockstate sensitive.
I **might** be wrong on this, honestly right now I'm a bit woozey, but I think that's how it works.
You'd probably want to place the event inside of the BlockState sensitive one.
Though I'd like a triage member to chime in on this.

ChampionAsh5357:
> I'd say fire it before vanilla behaviour.
> 
> So basically it'd work:
> Mod Override Behaviour -> super call -> Event -> Vanilla
> 
> Also I think you'd need to add it to both IForgeBlockState and IForgeBlock.
> Since IForgeBlockState methods tends to call state -> block -> block method.
> 
> So you have one "generic" and one blockstate sensitive.
> I **might** be wrong on this, honestly right now I'm a bit woozey, but I think that's how it works.
> You'd probably want to place the event inside of the BlockState sensitive one.
> Though I'd like a triage member to chime in on this.

Well I'll get started on the event while we wait. As for the IForgeBlockState, if you check out the latest files, I already implemented that. What about your thoughts on UseHoeEvent?

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -887,4 +890,37 @@ default boolean shouldDisplayFluidOverlay(BlockState state, IBlockDisplayReader
...
+    	if(tool.contains(ToolType.HOE)) {
+    		BlockState modifiedState = HoeItem.HOE_LOOKUP.get(state.getBlock());
+    		if(modifiedState == null && tool.size() == 1) return null;
+    		if(modifiedState != null) return modifiedState;
+    	}
+    	if(tool.contains(ToolType.SHOVEL)) {
+    		BlockState modifiedState = ShovelItem.SHOVEL_LOOKUP.get(state.getBlock());
+    		return modifiedState;
+    	}
+    	return null;
```
Lanse505: Return the original state if you can't find a modified one, not null.

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -887,4 +890,37 @@ default boolean shouldDisplayFluidOverlay(BlockState state, IBlockDisplayReader
...
+     * @param world The world
+     * @param pos The block position in world
+     * @param player The player clicking the block
+     * @param stack The stack being used by the player
+     * @return The resulting state after the action has been performed
+     */
+    @Nullable
+    default BlockState getToolModifiedState(BlockState state, World world, BlockPos pos, PlayerEntity player, ItemStack stack)
+    {
+    	Set<ToolType> tool = stack.getToolTypes();
```
Lanse505: I'd say fire the event here, at the beginning and if it returns not the original state, then return that state.
Otherwise continue.

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
ChampionAsh5357: The reason I return null is that AxeItem, HoeItem, and ShovelItem all check if the block isn't in the list by comparing the result to null.

Lanse505: Hmm, that's fair enough.

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
ChampionAsh5357: I've added the event.

src/main/java/net/minecraftforge/event/world/BlockEvent.java
```diff
@@ -444,4 +444,47 @@ public PortalSpawnEvent(IWorld world, BlockPos pos, BlockState state, NetherPort
...
+
+    	private final PlayerEntity player;
+    	private final ItemStack stack;
+    	private BlockState finalState;
+    	
+		public BlockToolInteractEvent(IWorld world, BlockPos pos, BlockState originalState, PlayerEntity player, ItemStack stack) {
+			super(world, pos, originalState);
+			this.player = player;
+			this.stack = stack;
+			this.finalState = originalState;
```
Lanse505: There really should be a private final originalState, so there is always a clean reference to the originalState.

src/main/java/net/minecraftforge/event/world/BlockEvent.java
```diff
@@ -444,4 +444,47 @@ public PortalSpawnEvent(IWorld world, BlockPos pos, BlockState state, NetherPort
...
+     * This event is {@link Cancelable}. If canceled, this will prevent the tool
+     * from changing the block's state.
+     */
+    @Cancelable
+    public static class BlockToolInteractEvent extends BlockEvent
+    {
+
+    	private final PlayerEntity player;
+    	private final ItemStack stack;
+    	private BlockState finalState;
```
Lanse505: Should probably just be state. (Nitpicky)

src/main/java/net/minecraftforge/event/world/BlockEvent.java
```diff
@@ -444,4 +444,47 @@ public PortalSpawnEvent(IWorld world, BlockPos pos, BlockState state, NetherPort
...
+    	
+		public BlockToolInteractEvent(IWorld world, BlockPos pos, BlockState originalState, PlayerEntity player, ItemStack stack) {
+			super(world, pos, originalState);
+			this.player = player;
+			this.stack = stack;
+			this.finalState = originalState;
+		}
+		
+		public PlayerEntity getPlayer()
```
Lanse505: All the getter lines are wrong formatting should be same line getters.
i.e. 
```
public PlayerEntity getPlayer() { return player; }
```

src/main/java/net/minecraftforge/event/world/BlockEvent.java
```diff
@@ -444,4 +444,47 @@ public PortalSpawnEvent(IWorld world, BlockPos pos, BlockState state, NetherPort
...
+		{
+			return player;
+		}
+		
+		public ItemStack getHeldItemStack()
+		{
+			return stack;
+		}
+		
+		public void setFinalState(BlockState finalState)
```
Lanse505: I'd say as above change it to state and getState and setState.

src/main/java/net/minecraftforge/event/world/BlockEvent.java
```diff
@@ -444,4 +444,47 @@ public PortalSpawnEvent(IWorld world, BlockPos pos, BlockState state, NetherPort
...
+		{
+			this.finalState = finalState;
+		}
+		
+		public BlockState getFinalState()
```
Lanse505: I'd say as above change it to state and getState and setState.

src/main/java/net/minecraftforge/event/world/BlockEvent.java
ChampionAsh5357: I extended BlockEvent, which holds a private final originalState.

Lanse505: Ah fair enough, ignore me :P

src/main/java/net/minecraftforge/event/world/BlockEvent.java
ChampionAsh5357: Referenced as finalState to know the diference between getState and getFinalState in code. However, I can change to state. (less nitpicky)

src/main/java/net/minecraftforge/event/world/BlockEvent.java
Lanse505: You can resolve this

src/main/java/net/minecraftforge/event/world/BlockEvent.java
Lanse505: You can resolve this

src/main/java/net/minecraftforge/event/world/BlockEvent.java
ChampionAsh5357: Extends BlockEvent. Cannot use getState. Better to specify what it does though. Can rename setFinalState to setState though if still wanted.

src/main/java/net/minecraftforge/event/world/BlockEvent.java
Lanse505: You can resolve this, it's valid since there is already a getState, setState, and original state things.

src/main/java/net/minecraftforge/event/world/BlockEvent.java
Lanse505: You can resolve this

src/main/java/net/minecraftforge/event/world/BlockEvent.java
ChampionAsh5357: Most of the formatting for BlockEvent is incorrect then.

Lanse505: I might be wrong, but that's how I've been told my triage in the past to format the getters/setters.

src/main/java/net/minecraftforge/event/world/BlockEvent.java
ChampionAsh5357: Extends BlockEvent. Cannot use getState. Better to specify what it does though.

Lanse505: once again, you can mark all of these comments as resolved now Champion :P

pupnewfster:
I am not quite sure of the best way to go about this, but I think it would be useful if there was a way to specify a specific ToolType in `getToolModifiedState`. I am not sure about other people but I do know I have a couple mods that have a tool that can act as multiple tool types and in an AOE, but only as one type of tool at once (so you don't both till ground and make grass paths at the same time). With how this is currently setup there isn't a good way that I could go about and say that I want to get the tool modified state for a given block and position. Part of me also wonders in that case if `onToolUse` should specify what type of tool (or tools) it is being used as. I am not sure if this would be the best way to modify it (some other input would be appreciated), but potentially changing it so that the signature is:

```java
default BlockState getToolModifiedState(BlockState state, World world, BlockPos pos, PlayerEntity player, ItemStack stack, Set<ToolType> toolTypes)
```

Not sure if both constructors should exist for convenience of calling it without having to get the ToolTypes from the stack, or if only one should exist then to make sure people override the correct one.

Edit: Also I am not sure but if onToolUse is the more generic form of say onHoeUse and given we haven't had an RB yet for 1.16 it may make sense to remove the no longer needed onHoeUse event.

ChampionAsh5357:
> With how this is currently setup there isn't a good way that I could go about and say that I want to get the tool modified state for a given block and position

That makes sense for hoe/shovel since you don't know which will be prevalent with a hashset. I think that Set<ToolTypes> should be made parameter for the IForgeBlock version. In IForgeBlockState, it will instead have two methods, one that includes the tool types and one that doesn't. This way users can make a tool behave with the requested types and the other just to take them off the ItemStack.



> Also I am not sure but if onToolUse is the more generic form of say onHoeUse and given we haven't had an RB yet for 1.16 it may make sense to remove the no longer needed onHoeUse event.

As I stated above, I haven't removed this yet because I don't know whether we should keep the implementation that allows a hoe to not till but still be damaged. This would return an ActionResultType directly instead of a specific blockstate. So, if this were completely removed, we could had a semi 'variation' of it; however, it wouldn't be able to return an result and instead defaulting to whatever a failed action produces. It's just whether this feature should be included as is with a result, without a result, or removed altogether.

pupnewfster:
> I think that Set should be made parameter for the IForgeBlock version. In IForgeBlockState, it will instead have two methods, one that includes the tool types and one that doesn't. This way users can make a tool behave with the requested types and the other just to take them off the ItemStack.

I think that is actually a pretty good solution, that way when people are overriding it in their block they only have one method to override they can override so they don't accidentally override the wrong one, and it also makes it so that when people are calling the method there can be the "helper" method that auto passes all tool types on the stack without forcing users to also pass all the tooltypes of the stack if they don't want it to act as a singular specific type.

src/test/java/net/minecraftforge/debug/block/ToolInteractTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.event.world.BlockEvent.BlockToolInteractEvent;
+import net.minecraftforge.fml.common.Mod;
+
+@Mod("tool_interact_test")
+public class ToolInteractTest {
```
pupnewfster: Brackets should be on a new line, and four spaces should be used instead of using tabs.

src/main/java/net/minecraftforge/event/world/BlockEvent.java
```diff
@@ -444,4 +447,39 @@ public PortalSpawnEvent(IWorld world, BlockPos pos, BlockState state, NetherPort
...
+    
+    /**
+     * Fired when when this block is right clicked by a tool to change its state.
+     * Used to determine if an axe can strip, a shovel can path, or a hoe can till.
+     * 
+     * This event is {@link Cancelable}. If canceled, this will prevent the tool
+     * from changing the block's state.
+     */
+    @Cancelable
+    public static class BlockToolInteractEvent extends BlockEvent
```
pupnewfster, line 457: Brackets should be on a new line, and four spaces should be used instead of using tabs. This added event has a mix of tabs/spaces and also brackets on the same line some of which having the bracket the body and the closing bracket even being on the same line.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -358,6 +360,13 @@ public static int onHoeUse(ItemUseContext context)
         }
         return 0;
     }
+    
+    @Nullable
+    public static BlockState onToolUse(BlockState originalState, World world, BlockPos pos, PlayerEntity player, ItemStack stack, Set<ToolType> toolTypes) {
```
pupnewfster: Brackets should be on a new line, and four spaces should be used instead of using tabs. Also change `Set<ToolType> toolTypes` to `ToolType toolType` (see my main comment for more info)

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -358,6 +360,13 @@ public static int onHoeUse(ItemUseContext context)
...
+    	BlockToolInteractEvent event = new BlockToolInteractEvent(world, pos, originalState, player, stack, toolTypes);
+    	if(MinecraftForge.EVENT_BUS.post(event)) return null;
+    	return event.getFinalState();
```
pupnewfster: This should probably be like other events in format:
```java
return MinecraftForge.EVENT_BUS.post(event) ? null : event.getFinalState();
```

src/main/java/net/minecraftforge/common/extensions/IForgeBlockState.java
```diff
@@ -778,4 +779,39 @@ default boolean shouldDisplayFluidOverlay(IBlockDisplayReader world, BlockPos po
...
+     * @param pos The block position in world
+     * @param player The player clicking the block
+     * @param stack The stack being used by the player
+     * @param toolTypes The tool types to be considered when performing the action
+     * @return The resulting state after the action has been performed
+     */
+    @Nullable
+    default BlockState getToolModifiedState(World world, BlockPos pos, PlayerEntity player, ItemStack stack, Set<ToolType> toolTypes)
+    {
+    	return getBlockState().getBlock().getToolModifiedState(getBlockState(), world, pos, player, stack, toolTypes);
```
pupnewfster: This line is a mix of spaces and tabs, and should only be spaces.

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -887,4 +890,38 @@ default boolean shouldDisplayFluidOverlay(BlockState state, IBlockDisplayReader
...
+     *
+     * @param state The current state
+     * @param world The world
+     * @param pos The block position in world
+     * @param player The player clicking the block
+     * @param stack The stack being used by the player
+     * @return The resulting state after the action has been performed
+     */
+    @Nullable
+    default BlockState getToolModifiedState(BlockState state, World world, BlockPos pos, PlayerEntity player, ItemStack stack, Set<ToolType> toolTypes)
```
pupnewfster: The implementation of this method is a mess of spaces and tabs, should be all tabs. Brackets should be on their own lines and there should be a space after the if statements before the parenthesis. Also change `Set<ToolType> toolTypes` to `ToolType toolType` (see my main comment for more info)

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -244,6 +244,9 @@ public net.minecraft.item.MusicDiscItem <init>(ILnet/minecraft/util/SoundEvent;L
 public net.minecraft.item.PickaxeItem <init>(Lnet/minecraft/item/IItemTier;IFLnet/minecraft/item/Item$Properties;)V
 public net.minecraft.item.ToolItem <init>(FFLnet/minecraft/item/IItemTier;Ljava/util/Set;Lnet/minecraft/item/Item$Properties;)V
 #endgroup
+public net.minecraft.item.AxeItem field_203176_a # BLOCK_STRIPPING_MAP
+public net.minecraft.item.HoeItem field_195973_b # HOE_LOOKUP
+public net.minecraft.item.ShovelItem field_195955_e # SHOVEL_LOOKUP
```
pupnewfster: After editing the access transformer file the `checkATs` task should be ran. It doesn't matter quite as much in this case given these ATs should be removed (see my main review comment for more of an explanation)

src/main/java/net/minecraftforge/event/world/BlockEvent.java
```diff
@@ -444,4 +447,39 @@ public PortalSpawnEvent(IWorld world, BlockPos pos, BlockState state, NetherPort
             return size;
         }
     }
+    
+    /**
+     * Fired when when this block is right clicked by a tool to change its state.
+     * Used to determine if an axe can strip, a shovel can path, or a hoe can till.
```
pupnewfster: This should list it as an example (`For example: Used to determine...`), as mods can add other tool types so these are not the only cases that may exist.

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -887,4 +890,38 @@ default boolean shouldDisplayFluidOverlay(BlockState state, IBlockDisplayReader
     {
         return state.getBlock() instanceof BreakableBlock || state.getBlock() instanceof LeavesBlock;
     }
+    
+    /**
+     * Returns the state that this block should transform into when right clicked by a tool.
+     * Used to determine if an axe can strip, a shovel can path, or a hoe can till.
```
pupnewfster: This should list it as an example (`For example: Used to determine...`), as mods can add other tool types so these are not the only cases that may exist.

src/main/java/net/minecraftforge/common/extensions/IForgeBlockState.java
```diff
@@ -778,4 +779,39 @@ default boolean shouldDisplayFluidOverlay(IBlockDisplayReader world, BlockPos po
...
+     */
+    @Nullable
+    default BlockState getToolModifiedState(World world, BlockPos pos, PlayerEntity player, ItemStack stack)
+    {
+    	return getToolModifiedState(world, pos, player, stack, stack.getToolTypes());
+    }
+    
+    /**
+     * Returns the state that this block should transform into when right clicked by a tool.
+     * Used to determine if an axe can strip, a shovel can path, or a hoe can till.
```
pupnewfster: This should list it as an example (`For example: Used to determine...`), as mods can add other tool types so these are not the only cases that may exist.

src/main/java/net/minecraftforge/common/extensions/IForgeBlockState.java
```diff
@@ -778,4 +779,39 @@ default boolean shouldDisplayFluidOverlay(IBlockDisplayReader world, BlockPos po
...
+     * Return null if vanilla behavior should be disabled.
+     *
+     * @param world The world
+     * @param pos The block position in world
+     * @param player The player clicking the block
+     * @param stack The stack being used by the player
+     * @return The resulting state after the action has been performed
+     */
+    @Nullable
+    default BlockState getToolModifiedState(World world, BlockPos pos, PlayerEntity player, ItemStack stack)
```
pupnewfster: This method should be removed (see my main comment for more info)

src/main/java/net/minecraftforge/common/extensions/IForgeBlockState.java
```diff
@@ -778,4 +779,39 @@ default boolean shouldDisplayFluidOverlay(IBlockDisplayReader world, BlockPos po
...
+     *
+     * @param world The world
+     * @param pos The block position in world
+     * @param player The player clicking the block
+     * @param stack The stack being used by the player
+     * @param toolTypes The tool types to be considered when performing the action
+     * @return The resulting state after the action has been performed
+     */
+    @Nullable
+    default BlockState getToolModifiedState(World world, BlockPos pos, PlayerEntity player, ItemStack stack, Set<ToolType> toolTypes)
```
pupnewfster: Change `Set<ToolType> toolTypes` to `ToolType toolType` (see my main comment for more info)

src/main/java/net/minecraftforge/event/world/BlockEvent.java
```diff
@@ -444,4 +447,39 @@ public PortalSpawnEvent(IWorld world, BlockPos pos, BlockState state, NetherPort
...
+    @Cancelable
+    public static class BlockToolInteractEvent extends BlockEvent
+    {
+
+    	private final PlayerEntity player;
+    	private final ItemStack stack;
+    	private final Set<ToolType> toolTypes;
+    	private BlockState state;
+    	
+		public BlockToolInteractEvent(IWorld world, BlockPos pos, BlockState originalState, PlayerEntity player, ItemStack stack, Set<ToolType> toolTypes) {
```
pupnewfster: Change `Set<ToolType> toolTypes` to `ToolType toolType` (see my main comment for more info)

pupnewfster (commented):
Discussed a few things with lex about this PR some of they key points that should get changed that are a bit more broad than being able to put comments on direct parts of the code so I am going to write them here:
- The internal lookup maps for axes, hoes, and shovels should not be exposed via an AT as they are mutable and this then exposes them as a "public API". Alternatively even though it will make the patches a little verbose you should add static helper methods to the axe, hoe, and shovel items that returns null or the new state, this will keep the mutability from being exposed.
- The HoeUseEvent should be deprecated and marked for removal in 1.17. Something like `//TODO 1.17 Remove`, for now there is no rush even though we haven't had an RB yet for 1.16 to remove it.
- Unlike what I previously suggested of using a set for the tool types, lex pointed out that it would make more sense to just pass a singular tool type that the stack is acting as. This way the caller can specify which order it wants the stack's tool types to act in rather than it being unordered.
- What is the purpose of `BlockToolInteractEvent` existing instead of just using the generic item use event? The main reason for this question is with how the tool interact event is currently implemented all it can be used for is cancelling vanilla interactions, this is different than the purpose the HoeEvent was serving where it would also allow for replacing the logic of what happened.

ChampionAsh5357:
> * What is the purpose of `BlockToolInteractEvent` existing instead of just using the generic item use event?

I think lanse brought up the point of being able to edit the return type of vanilla behavior. With the current thought process, this would allow the code the adjust the return type of the blockstate. However, with your recommendations, each class having a static helper would make it more logical to use PlayerInteractEvent$RightClickItem event If we wanted to still have a potential altered final state in the event, we can create an extension of the above event to allow this to happen. This would also be able to support the functionality in UseHoeEvent where it damages the weapon but doesn't perform the check. However, BlockToolInteractEvent is still needed in some capacity to return an altered final state.

> * Alternatively even though it will make the patches a little verbose you should add static helper methods to the axe, hoe, and shovel items that returns null or the new state, this will keep the mutability from being exposed.

I found it strange that only AxeItem was immutable after reviewing this. Both HoeItem and ShovelItem are made using an immutable map and then just transferred over to a standard hashmap. We can remove the newHashMap call to make the other two maps immutable as well to support the original idea. However, the static helpers are a better implementation regardless.

ChampionAsh5357:
Since the main branch has updated, I'm currently working on rebasing this without having to close and reopen. Apologies for any unnecessary commits.

pupnewfster:
I think something got screwed up in your merging because if you look at the diff there are a lot of "changes" that have nothing to do with this PR and I believe are things that are already merged into forge, and it makes it a bit hard to re-review the changes being made by this PR.

ChampionAsh5357:
> I think something got screwed up in your merging because if you look at the diff there are a lot of "changes" that have nothing to do with this PR and I believe are things that are already merged into forge, and it makes it a bit hard to re-review the changes being made by this PR.

I see that now, I'm gonna try to revert the commits and just leave as is for right now. I've managed to fix it; however, I still need to update the mods.toml file to get the test mod to work which requires some more rebasing. Apologies for the inconvenience.

patches/minecraft/net/minecraft/item/AxeItem.java.patch
```diff
@@ -9,3 +9,34 @@
...
+-            world.func_180501_a(blockpos, block.func_176223_P().func_206870_a(RotatedPillarBlock.field_176298_M, blockstate.func_177229_b(RotatedPillarBlock.field_176298_M)), 11);
++            world.func_180501_a(blockpos, block, 11);
+             if (playerentity != null) {
+                p_195939_1_.func_195996_i().func_222118_a(1, playerentity, (p_220040_1_) -> {
+                   p_220040_1_.func_213334_d(p_195939_1_.func_221531_n());
+@@ -52,4 +52,15 @@
+          return ActionResultType.PASS;
+       }
+    }
++
```
pupnewfster: There is random white space being added to this line, it should just be a blank line.

patches/minecraft/net/minecraft/item/AxeItem.java.patch
```diff
@@ -9,3 +9,34 @@
...
++            world.func_180501_a(blockpos, block, 11);
+             if (playerentity != null) {
+                p_195939_1_.func_195996_i().func_222118_a(1, playerentity, (p_220040_1_) -> {
+                   p_220040_1_.func_213334_d(p_195939_1_.func_221531_n());
+@@ -52,4 +52,15 @@
+          return ActionResultType.PASS;
+       }
+    }
++   
++   /* ======================================== FORGE START =====================================*/
```
pupnewfster: You can remove the forge start/end comments as forge has been moving away from having them for the most part especially when it is just adding a method at the end of a file.

patches/minecraft/net/minecraft/item/HoeItem.java.patch
```diff
@@ -12,9 +12,26 @@
...
++         BlockState blockstate = getHoeTillingState(world.func_180495_p(blockpos), world, blockpos, p_195939_1_.func_195999_j(), p_195939_1_.func_195996_i());
           if (blockstate != null) {
              PlayerEntity playerentity = p_195939_1_.func_195999_j();
+             world.func_184133_a(playerentity, blockpos, SoundEvents.field_187693_cj, SoundCategory.BLOCKS, 1.0F, 1.0F);
+@@ -47,4 +49,14 @@
+ 
+       return ActionResultType.PASS;
+    }
++   
++   /* ======================================== FORGE START =====================================*/
```
pupnewfster: You can remove the forge start/end comments as forge has been moving away from having them for the most part especially when it is just adding a method at the end of a file.

patches/minecraft/net/minecraft/item/ShovelItem.java.patch
```diff
@@ -9,12 +9,30 @@
...
 +         if (blockstate1 != null && world.func_175623_d(blockpos.func_177984_a())) {
              world.func_184133_a(playerentity, blockpos, SoundEvents.field_187771_eN, SoundCategory.BLOCKS, 1.0F, 1.0F);
              blockstate2 = blockstate1;
           } else if (blockstate.func_177230_c() instanceof CampfireBlock && blockstate.func_177229_b(CampfireBlock.field_220101_b)) {
+@@ -67,4 +67,14 @@
+          }
+       }
+    }
++   
++   /* ======================================== FORGE START =====================================*/
```
pupnewfster: You can remove the forge start/end comments as forge has been moving away from having them for the most part especially when it is just adding a method at the end of a file.

patches/minecraft/net/minecraft/item/HoeItem.java.patch
```diff
@@ -12,9 +12,26 @@
...
-          BlockState blockstate = field_195973_b.get(world.func_180495_p(blockpos).func_177230_c());
++         BlockState blockstate = getHoeTillingState(world.func_180495_p(blockpos), world, blockpos, p_195939_1_.func_195999_j(), p_195939_1_.func_195996_i());
           if (blockstate != null) {
              PlayerEntity playerentity = p_195939_1_.func_195999_j();
+             world.func_184133_a(playerentity, blockpos, SoundEvents.field_187693_cj, SoundCategory.BLOCKS, 1.0F, 1.0F);
+@@ -47,4 +49,14 @@
+ 
+       return ActionResultType.PASS;
+    }
++
```
pupnewfster: There is random white space being added to this line, it should just be a blank line.

patches/minecraft/net/minecraft/item/ShovelItem.java.patch
```diff
@@ -9,12 +9,30 @@
...
 -         if (blockstate1 != null && world.func_180495_p(blockpos.func_177984_a()).func_196958_f()) {
 +         if (blockstate1 != null && world.func_175623_d(blockpos.func_177984_a())) {
              world.func_184133_a(playerentity, blockpos, SoundEvents.field_187771_eN, SoundCategory.BLOCKS, 1.0F, 1.0F);
              blockstate2 = blockstate1;
           } else if (blockstate.func_177230_c() instanceof CampfireBlock && blockstate.func_177229_b(CampfireBlock.field_220101_b)) {
+@@ -67,4 +67,14 @@
+          }
+       }
+    }
++
```
pupnewfster: There is random white space being added to this line, it should just be a blank line.

patches/minecraft/net/minecraft/item/AxeItem.java.patch
```diff
@@ -9,3 +9,34 @@
...
++   @javax.annotation.Nullable
++   protected static BlockState getAxeStrippingState(BlockState originalState, World world, BlockPos pos, PlayerEntity player, ItemStack stack)
++   {
```
pupnewfster: Brackets in patches should be on the same line, not the next line.

patches/minecraft/net/minecraft/item/HoeItem.java.patch
```diff
@@ -12,9 +12,26 @@
...
+             world.func_184133_a(playerentity, blockpos, SoundEvents.field_187693_cj, SoundCategory.BLOCKS, 1.0F, 1.0F);
+@@ -47,4 +49,14 @@
+ 
+       return ActionResultType.PASS;
+    }
++   
++   /* ======================================== FORGE START =====================================*/
++   @javax.annotation.Nullable
++   protected static BlockState getHoeTillingState(BlockState originalState, World world, BlockPos pos, PlayerEntity player, ItemStack stack)
++   {
```
pupnewfster: Brackets in patches should be on the same line, not the next line.

patches/minecraft/net/minecraft/item/ShovelItem.java.patch
```diff
@@ -9,12 +9,30 @@
...
           } else if (blockstate.func_177230_c() instanceof CampfireBlock && blockstate.func_177229_b(CampfireBlock.field_220101_b)) {
+@@ -67,4 +67,14 @@
+          }
+       }
+    }
++   
++   /* ======================================== FORGE START =====================================*/
++   @javax.annotation.Nullable
++   protected static BlockState getShovelPathingState(BlockState originalState, World world, BlockPos pos, PlayerEntity player, ItemStack stack)
++   {
```
pupnewfster: Brackets in patches should be on the same line, not the next line.

src/main/java/net/minecraftforge/event/world/BlockEvent.java
```diff
@@ -444,4 +445,55 @@ public PortalSpawnEvent(IWorld world, BlockPos pos, BlockState state, NetherPort
...
+    @Cancelable
+    public static class BlockToolInteractEvent extends BlockEvent
+    {
+
+        private final PlayerEntity player;
+        private final ItemStack stack;
+        private final ToolType toolType;
+        private BlockState state;
+
+		public BlockToolInteractEvent(IWorld world, BlockPos pos, BlockState originalState, PlayerEntity player, ItemStack stack, ToolType toolType)
```
pupnewfster: The spacing of this class is still rather screwed up and I believe is a mixture of tabs and spaces

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -359,6 +361,12 @@ public static int onHoeUse(ItemUseContext context)
         }
         return 0;
     }
+    
+    @Nullable
+    public static BlockState onToolUse(BlockState originalState, World world, BlockPos pos, PlayerEntity player, ItemStack stack, ToolType toolType) {
```
pupnewfster: Bracket should be on the next line

src/main/java/net/minecraftforge/event/world/BlockEvent.java
```diff
@@ -444,4 +445,55 @@ public PortalSpawnEvent(IWorld world, BlockPos pos, BlockState state, NetherPort
...
+
+	    public void setFinalState(BlockState finalState)
+	    {
+	        this.state = finalState;
+	    }
+
+	    public BlockState getFinalState()
+	    {
+	        return state;
+	    }
```
pupnewfster: It probably would be nice to java doc all the methods in this event, but at the very least these two methods should have an explanation of what they do.

pupnewfster (commented):
I believe there was some confusion about the purpose of the static helper methods. The idea was to call the static helper methods from the default implementation of `getToolModifiedState` based on the passed in `ToolType`, not to offload/special case the logic into the tools themselves. Basically the static helper methods should just perform the map lookup, and in the case of the axe it probably makes sense to have it also properly copy the rotation as that is part of the lookup.

Also in regards to your comments on the event, the reason for the question was more that given how it is currently setup (in a defaulted method), that if mods just override the method they likely won't bother calling the event and people may wonder why they are not able to use it to "reliably" change the interaction with other mod's blocks.

ChampionAsh5357:
> Basically the static helper methods should just perform the map lookup, and in the case of the axe it probably makes sense to have it also properly copy the rotation as that is part of the lookup.

Ah, ok. My bad. I will fix that with all the changes suggested now.

> Also in regards to your comments on the event, the reason for the question was more that given how it is currently setup (in a defaulted method), that if mods just override the method they likely won't bother calling the event and people may wonder why they are not able to use it to "reliably" change the interaction with other mod's blocks.

Then it would be better to reference the event in the IForgeBlockState call so that it requires the event to be called. Otherwise, it wouldn't be able to be reliably used to change the interaction with the other mod's block.

patches/minecraft/net/minecraft/item/HoeItem.java.patch
```diff
@@ -12,9 +12,21 @@
...
           if (blockstate != null) {
              PlayerEntity playerentity = p_195939_1_.func_195999_j();
+             world.func_184133_a(playerentity, blockpos, SoundEvents.field_187693_cj, SoundCategory.BLOCKS, 1.0F, 1.0F);
+@@ -47,4 +49,9 @@
+ 
+       return ActionResultType.PASS;
+    }
++
++   @javax.annotation.Nullable
++   public static BlockState getHoeTillingState(BlockState originalState, World world, BlockPos pos, PlayerEntity player, ItemStack stack) {
```
pupnewfster: A lot of these parameters are unused and serve no purpose.

patches/minecraft/net/minecraft/item/AxeItem.java.patch
```diff
@@ -9,3 +9,29 @@
...
+             if (playerentity != null) {
+                p_195939_1_.func_195996_i().func_222118_a(1, playerentity, (p_220040_1_) -> {
+                   p_220040_1_.func_213334_d(p_195939_1_.func_221531_n());
+@@ -52,4 +52,10 @@
+          return ActionResultType.PASS;
+       }
+    }
++
++   @javax.annotation.Nullable
++   public static BlockState getAxeStrippingState(BlockState originalState, World world, BlockPos pos, PlayerEntity player, ItemStack stack) {
```
pupnewfster: A lot of these parameters are unused and serve no purpose.

patches/minecraft/net/minecraft/item/ShovelItem.java.patch
```diff
@@ -9,12 +9,25 @@
...
              world.func_184133_a(playerentity, blockpos, SoundEvents.field_187771_eN, SoundCategory.BLOCKS, 1.0F, 1.0F);
              blockstate2 = blockstate1;
           } else if (blockstate.func_177230_c() instanceof CampfireBlock && blockstate.func_177229_b(CampfireBlock.field_220101_b)) {
+@@ -67,4 +67,9 @@
+          }
+       }
+    }
++
++   @javax.annotation.Nullable
++   public static BlockState getShovelPathingState(BlockState originalState, World world, BlockPos pos, PlayerEntity player, ItemStack stack) {
```
pupnewfster: A lot of these parameters are unused and serve no purpose.

patches/minecraft/net/minecraft/item/HoeItem.java.patch
ChampionAsh5357: Sometimes you forget that these methods are only called for a specific case. My bad.

PR #7188, opened 2020-07-28
[1.16.x] Add Method for blocks to decide if they solidify concrete powder.

mrp-v2:
Adds a method to `IForgeBlock` and `IForgeBlockState` to allow blocks to decide whether concrete powder should solidify when they are touching on a certain side. Includes parameters for the `BlockState`, `Direction`, `IBlockReader`, and `BlockPos`.
Includes a test mod.

### DIFF
--- a/patches/minecraft/net/minecraft/block/ConcretePowderBlock.java.patch
+++ b/patches/minecraft/net/minecraft/block/ConcretePowderBlock.java.patch
@@ -1,0 +1,35 @@
+--- a/net/minecraft/block/ConcretePowderBlock.java
++++ b/net/minecraft/block/ConcretePowderBlock.java
+@@ -34,7 +34,7 @@
+    }
+ 
+    private static boolean func_230137_b_(IBlockReader p_230137_0_, BlockPos p_230137_1_, BlockState p_230137_2_) {
+-      return func_212566_x(p_230137_2_) || func_196441_b(p_230137_0_, p_230137_1_);
++      return causesSolidify(p_230137_2_, null, p_230137_0_, p_230137_1_) || func_196441_b(p_230137_0_, p_230137_1_);
+    }
+ 
+    private static boolean func_196441_b(IBlockReader p_196441_0_, BlockPos p_196441_1_) {
+@@ -43,14 +43,12 @@
+ 
+       for(Direction direction : Direction.values()) {
+          BlockState blockstate = p_196441_0_.func_180495_p(blockpos$mutable);
+-         if (direction != Direction.DOWN || func_212566_x(blockstate)) {
+             blockpos$mutable.func_239622_a_(p_196441_1_, direction);
+             blockstate = p_196441_0_.func_180495_p(blockpos$mutable);
+-            if (func_212566_x(blockstate) && !blockstate.func_224755_d(p_196441_0_, p_196441_1_, direction.func_176734_d())) {
++            if (causesSolidify(blockstate, direction.func_176734_d(), p_196441_0_, blockpos$mutable)) {
+                flag = true;
+                break;
+             }
+-         }
+       }
+ 
+       return flag;
+@@ -59,6 +57,7 @@
+    private static boolean func_212566_x(BlockState p_212566_0_) {
+       return p_212566_0_.func_204520_s().func_206884_a(FluidTags.field_206959_a);
+    }
++   private static boolean causesSolidify(BlockState p_212566_0_, @javax.annotation.Nullable Direction side, IBlockReader world, BlockPos pos) { return p_212566_0_.causesConcretePowderToSolidify(side, world, pos); }
+ 
+    public BlockState func_196271_a(BlockState p_196271_1_, Direction p_196271_2_, BlockState p_196271_3_, IWorld p_196271_4_, BlockPos p_196271_5_, BlockPos p_196271_6_) {
+       return func_196441_b(p_196271_4_, p_196271_5_) ? this.field_200294_a : super.func_196271_a(p_196271_1_, p_196271_2_, p_196271_3_, p_196271_4_, p_196271_5_, p_196271_6_);
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
@@ -45,6 +45,7 @@ import net.minecraft.item.ItemStack;
 import net.minecraft.item.ShovelItem;
 import net.minecraft.pathfinding.PathNodeType;
 import net.minecraft.tags.BlockTags;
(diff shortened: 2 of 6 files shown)


src/test/java/net/minecraftforge/debug/block/ConcretePowderConversionEventTest.java
```diff
@@ -0,0 +1,15 @@
+package net.minecraftforge.debug.block;
+
+import net.minecraftforge.event.world.BlockEvent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
+
+@Mod("concrete_powder_conversion_event_test")
+@Mod.EventBusSubscriber
+public class ConcretePowderConversionEventTest {
```
diesieben07: Formatting (braces on new lines)

src/test/java/net/minecraftforge/debug/block/ConcretePowderConversionEventTest.java
```diff
@@ -0,0 +1,15 @@
...
+
+    @SubscribeEvent
+    public static void concretePowderConversion(final BlockEvent.ConcretePowderConversionEvent event){
```
diesieben07: Formatting (braces on new lines)

src/main/java/net/minecraftforge/event/world/BlockEvent.java
```diff
@@ -444,4 +444,20 @@ public PortalSpawnEvent(IWorld world, BlockPos pos, BlockState state, NetherPort
             return size;
         }
     }
+
+    /* Fired when a concrete powder block is about to be converted to a concrete block from
```
diesieben07: This is not proper Javadoc formatting

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -729,10 +729,15 @@ public static long onSleepFinished(ServerWorld world, long newTime, long minTime
         MinecraftForge.EVENT_BUS.post(event);
         return event.getListeners();
     }
-    
+
```
diesieben07: Unnecessary whitespace change

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -729,10 +729,15 @@ public static long onSleepFinished(ServerWorld world, long newTime, long minTime
...
     public static void onCommandRegister(CommandDispatcher<CommandSource> dispatcher, Commands.EnvironmentType environment)
     {
         RegisterCommandsEvent event = new RegisterCommandsEvent(dispatcher, environment);
         MinecraftForge.EVENT_BUS.post(event);
     }
+
+    public static boolean onTryConvertConcretePowder(IWorld world, BlockPos pos)
+    {
+        return MinecraftForge.EVENT_BUS.post(new BlockEvent.ConcretePowderConversionEvent(world, pos, world.getBlockState(pos)));
```
diesieben07: Why call `getBlockState` here instead of passing in the state being converted?

diesieben07 (changes requested):
As it is implemented currently this event is very limited, as it only allows prevention of the conversion.
The event should instead provide a setter for the resulting block state and not just be cancellable but have a `Result`:
- `Result.DENY`: Do not convert.
- `Result.ALLOW`: Convert to block state given in event (which by default is the solidified state), regardless of vanilla checks.
- `Result.DEFAULT`: Apply vanilla logic.

It's also questionable if this should be an event at all. Is it not enough to ask the block which (potentially) causes the conversion? i.e. add a method to `IForgeBlockState` and `IForgeBlock`, called for example `convertConcretePowder`.

mrp-v2:
I have now modified the event to have a result instead of being `cancelable`.

> It's also questionable if this should be an event at all. Is it not enough to ask the block which (potentially) causes the conversion? i.e. add a method to `IForgeBlockState` and `IForgeBlock`, called for example `convertConcretePowder`.

I'm don't think I understand. Are you saying a method should be added to `IForgeBlock` and `IForgeBlockState` that `ConcretePowderBlock` then calls to see if it should convert?

diesieben07:
Yes, that way a custom fluid could for example decide to turn concrete powder into something else.
What else is your use-case for this being an event?

mrp-v2:
I have a mod in which I wanted to give players the option of disabling the vanilla concrete conversion mechanic.

diesieben07 (changes requested):
Now you have some sort of strange hybrid between the two suggested approaches. And you didn't implement the blockstate replacement.
And you still have unnecessary whitespace changes.

mrp-v2:
I fixed my IDE settings so the whitespace changes will stop happening. Also I think I may have reached a decent combination of the `IForge...` methods and the conversion event.

patches/minecraft/net/minecraft/block/ConcretePowderBlock.java.patch
```diff
@@ -0,0 +1,60 @@
...
+    private static boolean func_196441_b(IBlockReader p_196441_0_, BlockPos p_196441_1_) {
+@@ -43,25 +45,24 @@
+ 
+       for(Direction direction : Direction.values()) {
+          BlockState blockstate = p_196441_0_.func_180495_p(blockpos$mutable);
+-         if (direction != Direction.DOWN || func_212566_x(blockstate)) {
+             blockpos$mutable.func_239622_a_(p_196441_1_, direction);
+             blockstate = p_196441_0_.func_180495_p(blockpos$mutable);
+-            if (func_212566_x(blockstate) && !blockstate.func_224755_d(p_196441_0_, p_196441_1_, direction.func_176734_d())) {
++            if (causesSolidify(blockstate, direction.func_176734_d(), p_196441_0_, blockpos$mutable)) {
```
diesieben07, line 20: Why does suddenly only check the method? And not the event? You have to be consistent.

diesieben07 (changes requested):
My point was not to make the event _and_ method in IForgeBlock... Either one or the other. Both together just confuses things.

mrp-v2:
It is now only the `IForge...` methods, which should still allow canceling vanilla behavior by replacing vanilla's water block.

TheCurle:
Please rebase to clean up that commit log and fix the conflicts.

mrp-v2:
Done. Also cleaned up leftover code from the event.
The labels are outdated now, as this is no longer a new event.

sciwhiz12:
Hello! @mrp-v2, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

sciwhiz12:

PR #7190, opened 2020-07-29
Have ForgeRegistry implement Codec, add new ForgeRegistry usecase and better debug dynamic object creation.

Cyborgmas:
Vanilla registries are already Codecs but any custom registry won't be and can't be used for that purpose.
I pretty much copied vanilla's implementation. I tested with both the normal and compressed Json ops, but in reality the COMPRESSED JsonOps is never used in mc.
Which is probably for the best since I think that relying on the integer ids as vanilla does could cause trouble if some objects are removed in between worlds. Not sure how to handle that, I think cpw talked on stream about this and doing some things with Lifecycle.


Cyborgmas:
Okay, changed up some stuff a bit, added some support for custom mod data in DimensionType's by creating a new ForgeRegistry.
The DimensionExtraCodec -> DimensionExtra pair is similar to the GlobalLootModifierSerializer -> IGlobalLootModifier pair , except it uses codecs. It is a bit ugly because DimensionExtraCodec should in theory be a functional interface, but that can't be until a pretty substantial rewrite to the registry system occurs.

Small problem I encountered : custom objects coming from mods would not load until I caused a shift in DataPacks in the datapack menu... trying to find a solution but I think I caused it with my DatapackCodec shenanigans here:
https://github.com/MinecraftForge/MinecraftForge/pull/6913

Cyborgmas:
Okay so my last assessment was wrong, custom dimensions load properly even without trying to update something in the datapack menu.
However, mod data packs won't get tested until pretty late and the message is not super informative, so I provide a way to test them and to get some debug information.
2 things i am not sure of:
- How to properly handle the errors, including when to actually perform the check.
- How intensive this check will be on big modpacks... because right now with no objects it is at ~9 seconds. Perhaps a way to do this only in dev would be ideal? But then conflicts might arise from inter-modpacks behaviour although currently that is not very possible for the json objects.

Suggestions welcome, would like to handle the error more a bit more nicely.

PR #7191, opened 2020-07-29
[1.16] Add constructor of ClientWorld for rendering purposes

Nightenom:
This pr is follow up of #7225.

Situation prior 1.16: our mod is rendering template previews - for that you need to have fake World instance with few overrides so it works the same as normal World rendering, extending World class was sufficient and World constructor does not trigger any special things

Situation in 1.16: extending World class is no longer sufficient (see attached picture), we need to extend ClientWorld class which constructor trigger WorldLoad event (you definitely don't want to trigger it just for rendering purposes)

Solution: I've added two protected constructors to allow no-init of ClientWorld and World class for rendering purposes. Both of these constructors are setting null wherever possible, users are asked to override methods causing NPEs - my idea for this PR was also to create somewhat template classes (extending ClientWorld and Chunk) which would override necessary methods and provide somewhat documentation over what you should do when trying to render previews using vanilla World rendering pipeline <- just ask me to do that if you think it's a good idea

I would like to provide a test mod once #7225 will be merged.

Picture showing the problematic method

### DIFF
--- a/patches/minecraft/net/minecraft/client/world/ClientWorld.java.patch
+++ b/patches/minecraft/net/minecraft/client/world/ClientWorld.java.patch
@@ -62,3 +62,10 @@
        if (p_217384_1_ == this.field_73037_M.field_71439_g) {
           this.field_73037_M.func_147118_V().func_147682_a(new EntityTickableSound(p_217384_3_, p_217384_4_, p_217384_2_));
        }
+@@ -872,4 +889,6 @@
+          return this.field_239146_c_ ? 1.0D : 0.03125D;
+       }
+    }
++   @Deprecated // Only use for "fake" world rendering purposes, warning: some methods may cause NPE when using this constructor
++   protected ClientWorld() { super(); this.field_73035_a = null; this.field_217430_d = null; this.field_239130_d_ = null; this.field_239131_x_ = null; this.field_239129_E_ = null; }
+ }
--- a/patches/minecraft/net/minecraft/world/World.java.patch
+++ b/patches/minecraft/net/minecraft/world/World.java.patch
@@ -385,21 +385,43 @@
                    blockstate.func_215697_a(this, blockpos, p_175666_2_, p_175666_1_, false);
                 }
              }
-@@ -1075,6 +1137,18 @@
+@@ -1075,6 +1137,40 @@
        return this.field_226689_w_;
     }
  
++   /* ==== FORGE START ==== */
++
 +   private double maxEntityRadius = 2.0D;
++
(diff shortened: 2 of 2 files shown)


josephtarbit:
Seems like a good idea to me, I think the fake work template class is quite good too. This would be very useful.

TheCurle:
This has merge conflicts to resolve, and probably needs more community feedback. 
There are ways to create in-world and in-gui previews of a world with current systems, so this is arguably just a quality of life system.

marchermans:
@TheCurle For rendering in world (like overlays etc) you actually need a world instance as of 1.16. At least if you want models a chance to update their state properly to their fake and real neighbors.

For example Fences would only render in the default state otherwise, since you need a call to updatePostPlacement, which requires a world.

TheCurle:
Correct, but there are a few approaches as to where to get that instance from.
One mod i've seen temporarily swaps out a section of land near the borders for what they want to render, load that area, translates the camera and renders it into a framebuffer and displays that on the screen.
Ergo, they use the existing world instance.

There are other mods that create rendering dimensions that aren't accessible by other means, etc.
I mean, sure, this approach makes it easier, but it's _possible_ without this. Hence, needs more community input.
Is this wanted? Would it save headaches? Is it easier to maintain?

Alpvax:
Imo swapping out existing land seems like a bad idea, as you have to take into account other players potentially being near that area, as well as generating those chunks needlessly.
Having a rendering dimension would work, and we could add hooks to add one which multiple mods could use. It's still additional unnecessary server load though.

I think this would be the best solution of the three.

TheCurle:
I didn't say it was a flawless solution, i said it's what some mods are doing right now.
And, yes, this would probably be the better solution, but without community engagement we have no idea :)

50ap5ud5:
Hi, is this pull request still being worked on? The issue you mentioned would affect one of my mods, which uses a copy of the ClientWorld to render a chunk in another dimension.

LexManos:
The PR is not being considered, there are methods that allow you to render things using whatever world you want. You could also just make a full world and not have to deal with all the potential NPE explosions that this introduces.
If your concern is that you don't want people confused by a WorldLoad event. People shouldn't. WorldLoad events happen for a bunch of different worlds, making it an assumption that there is only one ClientWorld seems .. wrong.

And introducing API that breaks all of the null contracts of a ton of fields in a major core of the game is not something we're gunna do.

TLDR: You should be able to render whatever you want as is. If not you'll need to dig into something more specific then this hack.

PR #7192, opened 2020-07-29
Fix modproperties property in mods.toml causing exception

sciwhiz12:
There is a little known and undocumented property in `mods.toml`: the `modproperties` property. It allows modders to add custom data to their `IModInfo`, accessible through `IModInfo#getModProperties`, which returns a `Map<String, Object>`.

A commit was made to both [Forge][mcforgecommit] and [ForgeSPI][forgespicommit] that modified how modfiles were loaded, changing, among other things, how the `modproperties` property was loaded. This change causes a `java.lang.ClassCastException: com.electronwill.nightconfig.core.SimpleCommentedConfig cannot be cast to java.util.Map` on line 99 of `ModInfo`.

The exact same problem also will occur on line 79 of `ModFileInfo`, for the (also little known and undocumented) `properties` property, which exists for adding custom data to the `IModFileInfo` (the `mods.toml` for Forge), accessible through `IModFileInfo#getFileProperties`, which also returns a `Map<String, Object>`.

The only reason that this crash was discovered and this PR was created is because [JTK222 asked for help on Discord][discordconvo] regarding this property, which worked fine before on their 1.15 workspace but crashes on 1.16, and the underlying issue was found: a missing call to `UnmodifiableConfig#valueMap` to convert it from `UnmodifiableConfig` to a `Map<String, Object>`.

This PR adds the aforemetioned method to the `Optional` call chains of both `modproperties` and `properties`, fixing the issue. I have tested this fix locally on my machine with the `modproperties` values given in the Discord conversation.

[mcforgecommit]: https://github.com/MinecraftForge/MinecraftForge/commit/20f78ac724164110d80c30a19473ee0bc1d2ecc1
[forgespicommit]: https://github.com/MinecraftForge/ForgeSPI/commit/d73b00cc9e868d99951c7dd22d73dbc310b400b7
[discordconvo]: https://discordapp.com/channels/313125603924639766/725850371834118214/738138530445918298

### DIFF
--- a/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
+++ b/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
@@ -75,8 +75,7 @@ public class ModFileInfo implements IModFileInfo, IConfigurable
             .orElseThrow(()->new InvalidModFileException("Missing License, please supply a license.", this));
          */
         this.showAsResourcePack = config.<Boolean>getConfigElement("showAsResourcePack").orElse(false);
-        this.properties = config.<UnmodifiableConfig>getConfigElement("properties").
-                map(UnmodifiableConfig::valueMap).orElse(Collections.emptyMap());
+        this.properties = config.<Map<String, Object>>getConfigElement("properties").orElse(Collections.emptyMap());
         this.modFile.setFileProperties(this.properties);
         this.issueURL = config.<String>getConfigElement("issueTrackerURL").map(StringUtils::toURL).orElse(null);
         final List<? extends IConfigurable> modConfigs = config.getConfigList("mods");
--- a/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/NightConfigWrapper.java
+++ b/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/NightConfigWrapper.java
@@ -45,8 +45,14 @@ public class NightConfigWrapper implements IConfigurable {
     }
 
     @Override
+    @SuppressWarnings("unchecked")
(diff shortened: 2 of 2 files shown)


ichttt:
The latest 1.15 build should also be broken then, as this change is also in the 1.15 version. But this looks correct, as it's pretty much the same in 1.14: https://github.com/MinecraftForge/MinecraftForge/blob/1.14.x/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModInfo.java#L89-L90

jamierocks:
This was an intentional change, `ModInfo` was altered to allow for factory creation for none-`mods.toml` formats.

You need to fix the implementation of `IConfigurable`.

sciwhiz12:
Changed `NightConfigWrapper`, the default implementation of `IConfigurable` for Forge, to convert any `UnmodifiableConfig`s to their value-maps of `Map<String, Object>`.

~~I will update~~ Updated the 1.15 PR with this new fix.

JTK222:
Is there anything else missing here?
As I was the one initially affected by this, I cannot really update my mod till this gets fixed.
Therefore, I'd really appreciate if this could get merged asap.
(Btw, again thank you a lot sciwhiz12 :3)

jamierocks:
Yes, https://github.com/MinecraftForge/MinecraftForge/blob/8f9e52c26030e1033e66517271131c02dcb5e02f/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java#L79 will need correcting too.

sciwhiz12:
Also fixed that; I will note that the `properties` property is a separate (also undocumented) property which is for the whole `IModFileInfo` (in this case, it's for the `mods.toml`).

PR #7193, opened 2020-07-29
[1.16.x] [Cleanup] Removed obsolete HarvestDropsEvent

Yunus1903:
I removed the `HarvestDropsEvent` event because it has been left unused since **1.14** and mostly (if not fully) replaced by `GlobalLootModifier`.

~~If wanted I can make PR's for **1.14** & **1.15** too.~~

### DIFF
--- a/src/main/java/net/minecraftforge/event/ForgeEventFactory.java
+++ b/src/main/java/net/minecraftforge/event/ForgeEventFactory.java
@@ -252,13 +252,6 @@ public class ForgeEventFactory
         return event.getDisplayname();
     }
 
-    public static float fireBlockHarvesting(NonNullList<ItemStack> drops, World world, BlockPos pos, BlockState state, int fortune, float dropChance, boolean silkTouch, PlayerEntity player)
-    {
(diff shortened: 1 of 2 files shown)


Cyborgmas:
You can't for 1.14/1.15, it is a breaking change.

Yunus1903:
> You can't for 1.14/1.15, it is a breaking change.

If it's an event that doesn't even work, would it really count as breaking?

Cyborgmas:
Yup

gigaherz:
To be specific: any mod that happens to have a HarvestDropsEvent handler in their code, and they didn't remove it because it wasn't hurting the mod to keep it there, would suddenly start crashing. That's pretty much the definition of a breaking change. Just because *forge* isn't calling it, doesn't mean it won't break mods.

Yunus1903:
> To be specific: any mod that happens to have a HarvestDropsEvent handler in their code, and they didn't remove it because it wasn't hurting the mod to keep it there, would suddenly start crashing. That's pretty much the definition of a breaking change. Just because _forge_ isn't calling it, doesn't mean it won't break mods.

Yeah I get it. This PR is just some cleanup anyway, doesn't really matter for previous versions.

gigaherz (approved):

PR #5464, opened 2019-02-13
Make BannerPattern extensible

tterrag1098:
Added two factory methods to `BannerPattern` matching the two used constructors, one for ItemStack based, and one using pattern strings. New impl of `RuntimeEnumExtender` handles this case fine :smile:.

Will wait to merge this until after 1.13.2 update is complete, then rebase on top of that. I doubt there will be conflicts.

This closes #5447

PR #7195, opened 2020-07-30
[1.16.x] Add possibility to create custom Game Rule and Category

Mysterious-Dev:
This PR adds the possibility of creating custom Game Rules as well as the categories that go with it.

A test mod is included

### DIFF
--- a/patches/minecraft/net/minecraft/world/GameRules.java.patch
+++ b/patches/minecraft/net/minecraft/world/GameRules.java.patch
@@ -1,0 +1,23 @@
+--- a/net/minecraft/world/GameRules.java
++++ b/net/minecraft/world/GameRules.java
+@@ -219,7 +219,7 @@
+       }
+    }
+ 
+-   public static enum Category {
++   public static enum Category implements net.minecraftforge.common.IExtensibleEnum {
+       PLAYER("gamerule.category.player"),
+       MOBS("gamerule.category.mobs"),
+       SPAWNING("gamerule.category.spawning"),
+@@ -238,6 +238,11 @@
+       public String func_234908_a_() {
+          return this.field_234907_h_;
+       }
++
++      public static Category create(String name, String p_i231613_3_) {
++         throw new IllegalStateException("Enum not extended");
++      }
++
+    }
+ 
+    interface IRule<T extends GameRules.RuleValue<T>> {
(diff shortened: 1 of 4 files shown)


diesieben07:
The problem with this is that game-rules are not namespaced, so there will be conflicts. The question is, why does this need to exist at all? The equivalent in mods would be configs. A command to modify mod configs would be much more useful instead, no?

Mysterious-Dev:
Yes it is a good idea for the command but the only problem is that it applies to all the worlds of the game, while a Gamerule only applies to the world. The two solutions can be complementary

diesieben07:
There are world-scoped configs as of 1.13.

Mysterious-Dev:
Oh okay, I didn't know that

Mysterious-Dev:
But then in this case we could make a gui like the gamerules screen for the creation world.

Mysterious-Dev:
What do you think @diesieben07 ?

diesieben07:
Sure, that would be cool. There have been efforts to port forward the config GUI.

Mysterious-Dev:
But in this case it would be a GUI at the creation of the world.

diesieben07:
Same idea, the most of the work that needs to be done is actually building a GUI for the configs.

ExpensiveKoala:
I don't agree with @diesieben07 that this is useless because per-world configs exist.
Making it so you need a custom command to change your config is how you get hundreds of commands that just clutter /help anyways.

If you mean for there to be a system to dynamically create commands for the configspecs, then that in itself seems redundant. Which is why a pre-world create config screen was suggested. That doesn't work for values people want to change multiple times throughout the playthrough. Especially so if the point is to make it command block accessible, or even allowing a questing book mod to run the command on completion.

A gamerule extension cuts the command clutter, significantly reduces the overhead needed to create those commands.

Also, the suggestion that it not being namespaced which causes clashes can easily be solved with a slightly bigger patch.

There are also many enums that were extended, but didn't include making them namespaced. Like: https://github.com/MinecraftForge/MinecraftForge/pull/5464

I can definitely see the other side of the argument that configs do everything that gamerules already do, but is that enough of a reason to locking out modders from extending upon vanilla behavior?

Darkere:
Someone already made a mod to change any config (except for client-configs) via command. https://www.curseforge.com/minecraft/mc-mods/commands4configs

This means modders don't even need to implement a command, which is a much better solution IMO.

PR #7198, opened 2020-07-31
Added Xdoclint:none to javadoc task (browser docs)

TheArcaneBrony:
It's useful to be able to have documentation in browser so you can see an overview of things outside of your IDE.
This change allows you to compile them without all the "reference not found" errors etc.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1546,3 +1546,13 @@ task setup() {
 }
 
 
+allprojects {
+    repositories {
+        jcenter()
+    }
+    gradle.projectsEvaluated {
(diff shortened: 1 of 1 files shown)


diesieben07:
How about instead we fix the errors?

TheCurle:
diesieben is correct, this is not the way to go about doing this. 
I agree with you that browser based javadocs are useful, but this is like gluing together a broken plate. 
We should focus instead on not throwing the plate across the room.

If you want to actually fix this, I suggest making a new PR.

PR #7199, opened 2020-07-31
[1.16.x] Fixed custom models ignoring "gui_light" model tag

nikita488:
Custom models always used "SIDE" gui light completely ignoring "gui_light" model tag.
Before:
image
After:
image

### DIFF
--- a/src/main/java/net/minecraftforge/client/model/BlockModelConfiguration.java
+++ b/src/main/java/net/minecraftforge/client/model/BlockModelConfiguration.java
@@ -105,8 +105,9 @@ public class BlockModelConfiguration implements IModelConfiguration
     }
 
     @Override
-    public boolean isShadedInGui() {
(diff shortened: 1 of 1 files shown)


gigaherz:
Sorry that I didn't take a look sooner, summer heat left me stupid.

This PR makes no sense. `isShadedInGui` is not meant to be the source for the `gui_light` parameter. That's what `isSideLit` is for. which is already implemented appropriately. Since 1.15, the use of the `isGui3d()` method (which is what the `isShadedInGui()` parameter is meant to be used for), seems to have been reduced to a single call in `ItemRenderer` which chooses how to **translate** the model during rendering, no rotation or lighting involved.
However! When taking a look at your issue that you were trying to fix, I have noticed some inconsistency in how the methods are used, which I'll need to verify and fix.

PR #7201, opened 2020-08-01
Added Basic Class-Level SAS Capability to checkSAS, added Vector4f to SAS

DaemonUmbra:
Feel free to cherry pick just the checkSAS change


build.gradle
```diff
@@ -882,11 +882,18 @@ project(':forge') {
                     def (cls, desc) = (line.trim() + '    ').split(' ', -1)
                     cls = cls.replaceAll('\\.', '/')
                     desc = desc.replace('(', ' (')
-                    if (desc.isEmpty() || json[cls] == null || json[cls]['methods'] == null || json[cls]['methods'][desc] == null) {
+
+                    //Class SAS
+                    if(desc.isEmpty()) {
+                        lines.add(cls + (comment == null ? '' : ' ' + comment))
```
LexManos: Space after if, aand use the json to make sure the class exists. It's the point of validating...
Also, should this propagate to children?
Should this propagate all overridden methods? Such as a Sided interface being stripped, thus stripping the needing the implementations of that interface to be revived. I'm thinking IChestLid here.

DaemonUmbra: I was wondering if it should propagate to children.
At the time I just wanted to be able to get Vector4f common before RB, should have marked this as a draft

DaemonUmbra: Would be nice to have it propagate but I'm not sure checkSAS has enough info to be sure that the child of one SASed interface doesn't implement other sideonly interfaces that prevent it from being a common class

LexManos: It has enough information to make sure you can propagate it to the specific methods inside the class you're unsiding.
It doesn't need to unside the whole class of the client. Just the methods.

DaemonUmbra:
Well that merge was an unmitigated disaster

PR #7204, opened 2020-08-02
Make baking of blockstate-id map reentrant

diesieben07:
Fixes issues with "Open to LAN" if a resource pack is included in the world. A world resourcepack will cause bake to be called again due to reloading of resources.
This will cause the BlockState-ID map to become completely messed up, as it now contains every state twice. Someone joining the LAN world will see a completely empty map, as they receive completely wrong state IDs. For some reason it still seems to work locally.

To reproduce this issue create a world with a world resourcepack (must be called `resources.zip` in the world folder). Open the world to LAN and join with a 2nd game instance.


diesieben07:
After further investigation with Lex on Discord it seems that resource reloading reloads mods, which is not desired. Lifecycle events fire again, etc.

PR #7207, opened 2020-08-03
[1.16] Fix possible crash when using rendering regionCache

### LINKED ISSUE #7197: [1.15 & 1.16] Possible crash when using rendering regionCache
**Minecraft Version:** 1.15 & 1.16

**Forge Version:** both latest (1.15.2 - 31.2.33, 1.16.1 - 32.0.98)

**Logs:**
crash log: 
https://gist.github.com/loudwhitenoise/301f5a98f974eeb3eb16fb73d0d1da62 
latest.log: 
https://gist.github.com/loudwhitenoise/f598067f53ec34521cd61f4ece7fa795

**Description of issue:**
CacheLoader [(code link)](https://github.com/MinecraftForge/MinecraftForge/blob/638d1d31deabd8a83cf4b84496549703bc527ab4/src/main/java/net/minecraftforge/client/MinecraftForgeClient.java#L116) may cause crash when returning null. This shouldn't happen in vanilla code (as the cache is filled properly by vanilla), but can happen when rendering fake worlds which are not using ChunkRenderCache properly.

**Proposed solution:**
Make cache value Optional<ChunkRenderCache> and adjust methods accordingly

I can make PR when the final solution is determined

Nightenom:
Fixes #7197 using Optional as cache value wrapper

[This](https://github.com/MinecraftForge/MinecraftForge/pull/7207/files#diff-6b2a01e94e48c3cec54af7d282a1abd0L116) may be null when a chunk is completely empty (air blocks only)

### DIFF
--- a/src/main/java/net/minecraftforge/client/MinecraftForgeClient.java
+++ b/src/main/java/net/minecraftforge/client/MinecraftForgeClient.java
@@ -116,7 +116,7 @@ public class MinecraftForgeClient
         if (cache == null)
             regionCache.invalidate(Pair.of(world, position));
         else
-            regionCache.put(Pair.of(world, position), Optional.ofNullable(cache));
+            regionCache.put(Pair.of(world, position), Optional.of(cache));
     }
 
     public static ChunkRenderCache getRegionRenderCache(World world, BlockPos pos)
--- a/src/main/java/net/minecraftforge/client/MinecraftForgeClient.java
+++ b/src/main/java/net/minecraftforge/client/MinecraftForgeClient.java
@@ -28,6 +28,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 import net.minecraft.client.renderer.RenderType;
 import org.apache.commons.lang3.tuple.Pair;
@@ -119,6 +120,7 @@ public class MinecraftForgeClient
             regionCache.put(Pair.of(world, position), Optional.of(cache));
     }
 
+    @Nullable
     public static ChunkRenderCache getRegionRenderCache(World world, BlockPos pos)
     {
         return getRegionRenderCacheOptional(world, pos).orElse(null);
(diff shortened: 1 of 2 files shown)


src/main/java/net/minecraftforge/client/MinecraftForgeClient.java
```diff
@@ -122,10 +123,10 @@ public static void onRebuildChunk(World world, BlockPos position, ChunkRenderCac
         if (cache == null)
             regionCache.invalidate(Pair.of(world, position));
         else
-            regionCache.put(Pair.of(world, position), cache);
+            regionCache.put(Pair.of(world, position), Optional.ofNullable(cache));
     }
 
-    public static ChunkRenderCache getRegionRenderCache(World world, BlockPos pos)
+    public static Optional<ChunkRenderCache> getRegionRenderCache(World world, BlockPos pos)
```
diesieben07: This is a binary breaking change. Is there a reason you are not just adding a null check on the call-site?

diesieben07 (changes requested):

src/main/java/net/minecraftforge/client/MinecraftForgeClient.java
Nightenom: Guave Cache does not allow null values as mentioned in linked issue (#7197)

diesieben07: That doesn't mean this method needs to return an Optional.

Nightenom: Surely I can rename this to getRegionRenderCacheOptional and make the old method link it

diesieben07: Or just not change the method at all.

Nightenom: Done

Nightenom:
@diesieben07 anything else what needs to be addressed? I'm gonna make backport once this is ready

src/main/java/net/minecraftforge/client/MinecraftForgeClient.java
```diff
@@ -115,10 +116,15 @@ public static void onRebuildChunk(World world, BlockPos position, ChunkRenderCac
         if (cache == null)
             regionCache.invalidate(Pair.of(world, position));
         else
-            regionCache.put(Pair.of(world, position), cache);
+            regionCache.put(Pair.of(world, position), Optional.ofNullable(cache));
```
marchermans: You can use a simple `Optional.of(...)` here since `cache` can not be null.

marchermans (approved):
One minor nitpicking then this looks good.

PR #7210, opened 2020-08-03
[1.15.2] Fix misaligned patch for ServerWorld#updateEntity

### LINKED ISSUE #6516: Entity#canUpdate ignored in ServerWorld#updateEntity
**Minecraft Version: 1.15.2**

**Forge Version: 31.1.1**

**Steps to Reproduce:**
 1. Try setting the flag canUpdate
 2. see how entities still update on the server

**Description of issue:**
this is what i think is the problem: this is the current code
```java
if (entityIn.canUpdate())
    iprofiler.func_230035_c_("tickNonPassenger");
entityIn.tick();
```
and this is what i think it should be:
```java
if (entityIn.canUpdate()) {
    iprofiler.func_230035_c_("tickNonPassenger");
    entityIn.tick();
}
```

sciwhiz12:
_This PR fixes #6516._

As the title says; a misaligned patch for `ServerWorld#updateEntity` makes the `Entity#canUpdate` variable and method useless, as the `if` statement only affects the call to `IProfiler`.

This PR simply moves the `if` statement down to affect the `Entity#tick` method instead. _(or, depending on your viewpoint, moves the call to `IProfiler` up to not be affected by the `if`)_

### DIFF
--- a/patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
+++ b/patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
@@ -143,17 +143,18 @@
           i += this.field_73012_v.nextInt(8) - this.field_73012_v.nextInt(8);
           j += this.field_73012_v.nextInt(8) - this.field_73012_v.nextInt(8);
           ++k;
-@@ -584,8 +593,9 @@
+@@ -584,9 +593,10 @@
              ++p_217479_1_.field_70173_aa;
              IProfiler iprofiler = this.func_217381_Z();
              iprofiler.func_194340_a(() -> {
 -               return Registry.field_212629_r.func_177774_c(p_217479_1_.func_200600_R()).toString();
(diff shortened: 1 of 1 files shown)


diesieben07:
Note: Patch was fixed in 1.16 already.

LexManos:
This should of been assigned to the LTS team, also link the commit in 1.16 that this is a backport of.

sciwhiz12:
The commit that fixed this patch in 1.16 is 631230b298b958136c74c8a4687246ad878d669f, which is an amalgamation of patch fixes by tterag.

LexManos:
Ya I know, I was the one who fixed it. That commit is just a squash of a lot of our work in the update.
I'm just saying to you and Dies that for future reference things like this should have links to what it's backporting.
https://github.com/MinecraftForge/MinecraftForge/commit/631230b298b958136c74c8a4687246ad878d669f#diff-46083ba4c2dc13cceb01bc37bf449598R87 is the proper link.

PR #7214, opened 2020-08-03
Add user-friendly exceptions when config loading fails

sciwhiz12:
Currently, when a config file fails to be parsed by NightConfig, it throws a `ParsingException`. The exception includes no details of what file errored, and in the case of corruption of the file (which is the most reported error), has the vague message of `Not enough data available`.

Ideally, the solution would be to find the underlying cause of config corruption and fix it. However, there is no known way to reproduce the issue _(outside of trying it thousands of times and hoping for the best)_, so a fix cannot be attempted. 
But,at least make the exception message can be made more helpful, so users and support can pinpoint the file causing the crash (because `ParsingException`s are also thrown on invalid TOML files) without losing user-defined config settings.

This PR does the following:
  * Wraps any `ParsingException` thrown during the loading of `FMLConfig` with a `RuntimeException` to have a more helpful message: 
      `Failed to load FML config from {configFile.toString()}"`;
  * Creates a new `ConfigLoadingException`, which has a helpful message: 
      `Failed loading config file <config.getFullPath().toString()> with type <config.getType()> for modid <config.getModId()>`
  * Wraps any `ParsingException` thrown during config loading (at game startup) and re-loading (from file changes) with a `ConfigLoadingException`.

_Notes:_
  * Only a `ParsingException` thrown during game startup from `FMLConfig` and `ConfigTracker` makes the game crash. A `ParsingException` thrown when config re-loading due to file changes does not crash the game, presumably because the `ConfigWatcher` runs on a separate thread from the game.
  * This is not a solution for the problem of config corruption; that must be tackled separately when the underlying issue is found. In the meanwhile, this PR presents a stopgap measure to allow both end-users and support to better help the user _(otherwise the advice is mostly "delete all your configs")_.
  * `ConfigLoadingException` was not used for `FMLConfig` because there is no `ModConfig` available, and I wanted the exception to be clear as to what it represents: a config loading failure caused by a **mod**.


src/main/java/net/minecraftforge/fml/config/ConfigLoadingException.java
```diff
@@ -0,0 +1,37 @@
...
+    public ConfigLoadingException(ModConfig config, Exception cause)
+    {
+        super(cause);
+        this.config = config;
+    }
+
+    @Override
+    public String getMessage()
+    {
+        return "Failed loading config file " + config.getFullPath().toString() + " with type " + config.getType() + " for modid " + config.getModId();
```
ichttt: This seems weird, why not just print pass the message as a super argument?

sciwhiz12: I believe that is because I looked at `LoadingFailedException` as a reference for how exceptions and their messages were constructed. Will change.

LexManos:
Does the config watcher thread break everything else? Is there a way to detect that and make sure we are not in a state where configs are frozen or not written because the watcher thread is dead.

sciwhiz12:
`ConfigWatcher` is a `Runnable` that is run by [`FileWatcher`][file_watcher] on its associated daemon `WatcherThread` whenever the given watched file changes on disk. Exceptions thrown by the `Runnable` are [given to the exception handler][exception_handling] set on creation of the `FileWatcher` instance.

Because we use the default instance of `FileWatcher`, we use [the default exception handler][default_handler], which prints then disregards the exception, ensuring that the `WatcherThread` will always continue running, and any further file changes will still be detected and loaded.

Also, since the parser writes the config data immediately when parsed, the config may contain inconsistent values, because any futher entries in the file after the exception is thrown will not be parsed, which may cause mod behavior inconsistencies. _(The associated `ModConfig.Reload` event will also not be fired, so any mods which bake config values using that event will not be affected)_

Possible solutions to the above would be (a) making our own instance of `FileWatcher` with a custom exception handler, and (b) loading two instances of `CommentedConfig`, one being the _real_ file config and the other a value-holder that is given to the mod, and only copying the values from the _real_ file config to the value-holder when loading succeeds. _(However, I believe that those changes would be out-of-scope of this PR)_

[file_watcher]: https://github.com/TheElectronWill/night-config/blob/78d9b1cbfc700bacddab903576693f7b89306a80/core/src/main/java/com/electronwill/nightconfig/core/file/FileWatcher.java
[exception_handling]: https://github.com/TheElectronWill/night-config/blob/78d9b1cbfc700bacddab903576693f7b89306a80/core/src/main/java/com/electronwill/nightconfig/core/file/FileWatcher.java#L180-L184
[default_handler]: https://github.com/TheElectronWill/night-config/blob/78d9b1cbfc700bacddab903576693f7b89306a80/core/src/main/java/com/electronwill/nightconfig/core/file/FileWatcher.java#L49-L51

LexManos:
Interesting, I understand the small scope of this PR, however it seems to be a fundamental config issue. So can you look into option (b), as that seems the best course of action. Keeping the game stable while reading errors happen. 
So what i would suggest is that you move the new exception class to a inner class to keep it from being public API in case you need to change it into something more standard in the PR regarding Option B. And I can pull this as a informative change to help debugging while you look into the proper fix.

src/main/java/net/minecraftforge/fml/config/ConfigFileTypeHandler.java
```diff
@@ -120,4 +120,12 @@ public void run() {
             }
         }
     }
+
+    public static class ConfigLoadingException extends RuntimeException
+    {
```
LexManos, line 125: private

sciwhiz12: Whoops. Fixed.

PR #7216, opened 2020-08-03
[1.16.x] Render local player when the renderViewEntity is not the local player

XFactHD:
This is the same as #6945 because I blew that one up...

# Description

# Use-cases


XFactHD:
This PR is now ready for 1.16.2.

XFactHD:
@OrionDevelopment Would you mind taking a look at this again? I blew up the original PR because of a stupid mistake...

XFactHD:
Rebased onto latest changes

Earthcomputer:
Maybe add a comment in the patch saying what's changed?

XFactHD:
Done

diesieben07:
Lex voiced some concern on Discord regarding this.
I have deciphered this mess of a condition. It boils down to this, in vanilla:
An entity is rendered if all of the following is true:

1. The entity is inside the view frustrum or is being ridden by or riding the player.
2. The entity is not the view entity or, if it is, third person mode is active or the entity is sleeping.
3. The entity is not a player or is the view entity.

The last point is changed here to not filter out the client player, but only if they are not a spectator. The spectator check is needed to preserve vanilla behavior (you cannot see spectator mode players who are spectating an entity).

PR #7218, opened 2020-08-04
Allow to connect using TLS and TLS with SNI

ctron:
This change allows a client to connect to Minecraft server with TLS enabled.

I don't want to argue about then necessity of the security TLS bring. To me, the main reason to enable TLS, is the use of SNI (see https://en.wikipedia.org/wiki/Server_Name_Indication). In a nutshell, it is similar to a "vhost" header, just on the TLS level.

This allows you to connect to a Minecraft server, which is running behind a load balancer/proxy. Using SNI, the load balancer can distinguish the target host, and then properly forward the connection.

So this even makes sense in the "insecure" mode. Which is not less secure than the normal Minecraft protocol. But still allows to use TLS SNI.

The changes are:

* Allow the user to enable TLS (also in insecure mode)
* Respect the TLS settings when connecting
* A new field in the "add/edit server" screen:
  image

It does not make any changes on the server side!

In order to test this, you need to run a normal Minecraft server, and e.g. an instance of `stunnel`. Assuming the minecraft server is running on port 1234:

* Install `openssl` and `stunnel`
* Create a self signed certificate:
  ~~~
  openssl req -new -x509 -days 365 -nodes -out stunnel.pem -keyout stunnel.pem
  # accept the defaults and enter some random information
  openssl gendh 2048 >> stunnel.pem
  ~~~
* Create a config file for stunnel (`mctunnel.conf`):
  ~~~
  foreground = yes
  pid=
  
  [mc]
  accept  = 5678
  connect = 1234
  cert = /home/user/stunnel.pem
  ~~~
* Run stunnel:
  ~~~
  stunnel mctunnel.conf
  ~~~

With this change, you can now connect to port `5678` if you select "insecure TLS" as an option.

### DIFF
--- a/patches/minecraft/net/minecraft/client/gui/screen/AddServerScreen.java.patch
+++ b/patches/minecraft/net/minecraft/client/gui/screen/AddServerScreen.java.patch
@@ -1,0 +1,38 @@
+--- a/net/minecraft/client/gui/screen/AddServerScreen.java
++++ b/net/minecraft/client/gui/screen/AddServerScreen.java
+@@ -24,6 +24,7 @@
+    private TextFieldWidget field_146308_f;
+    private TextFieldWidget field_146309_g;
+    private Button field_152176_i;
++   private Button buttonTlsMode;
+    private final Screen field_228179_g_;
+    private final Predicate<String> field_181032_r = (p_210141_0_) -> {
+       if (StringUtils.func_151246_b(p_210141_0_)) {
+@@ -72,10 +73,14 @@
+          this.field_146311_h.func_152584_a(ServerData.ServerResourceMode.values()[(this.field_146311_h.func_152586_b().ordinal() + 1) % ServerData.ServerResourceMode.values().length]);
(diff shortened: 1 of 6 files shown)


diesieben07:
In my opinion this is not something that should be in Forge. This is something a mod should add.

TheCurle:
The security advantages of using TLS are practically nullified by being used on a game like Minecraft. All of the sensitive information involved (Mojang auth server tokens) is already encrypted. There is no other valuable information involved.

SNI can be added separately by a mod, like 7 said. This is beyond Forge's scope, imo.

ctron:
So how would it be possible to intercept the connection creation and extend the "add server" screen in a mod. I don't see any extension points for that. But as I am new to Minecraft modding, maybe that is just me and you can give me some pointers. That would be great.

Unnoen:
This won't be accepted into Forge as it's out of scope and befits a mod implementation better.

If you want some assistance on how to implement this in a mod, the [Forge Discord](https://discord.gg/UvedJ9m) might be of help.

PR #7219, opened 2020-08-04
[1.16.x] Add mod URLs to the Missing Mods error screen.

### LINKED ISSUE #6799: Show a URL so the user can download missing Dependencies
As #6798 was badly worded:
when a dependncy is missing show the user this screen but with a link to a download page of the mod:
img

Of course if the mod is installed that URL takes priority if there. **NO** auto-downloding, just a quick link for the user. 
A TOML example could be:
```
[[dependencies.examplemod]]
    modId="forge" #mandatory
    mandatory=true #mandatory
    versionRange="[28.0.19,)"
    ordering="NONE"
    side="BOTH"
    url="http://files.minecraftforge.net/"
```

mrp-v2:
Resolves #6799
Adds a new line to the missing mods error screen giving a mods homepage.
Allows `displayURL` to be defined in the `dependencies` section of a `mods.toml`.

The only thing I couldn't figure out is how to make the URLs clickable.


src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModInfo.java
```diff
@@ -243,6 +245,11 @@ public IModInfo getOwner()
         {
             return owner;
         }
+
+        public String getDisplayURL()
+        { 
+            return displayURL;
+        }
```
diesieben07, line 264: This should not be like this... This should be implemented from the interface, so this class doesn't need to be made public and you don't need to do those ugly instanceof checks.

diesieben07 (changes requested):

src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModInfo.java
jamierocks: The instanceof checks aren't needed anyway, FML already hard assumes the implementation classes.

mrp-v2: I've now made a pull request for the interface. See MinecraftForge/ForgeSPI#6

mrp-v2:
I've now updated the PR to implement the changes in the interface. This means it now has errors in it until the interface is updated.

cpw:
The lack of a screenshot showing the result disturbs me.

mrp-v2:
image
Here's what the error screen looks like with this PR.

mcenderdragon:
@mrp-v2 what is the current state of this PR? You got it working so is the draft state still needed?

mrp-v2:
This is a draft PR because it requires [this PR](https://github.com/MinecraftForge/ForgeSPI/pull/6) in ForgeSPI to be merged before this one can be.

Shadows-of-Fire:
The relevant ForgeSPI PR has been merged, thus this PR can move forward.
@mrp-v2 To proceed, you will need to rebase onto 1.18.x.

PR #7221, opened 2020-08-05
[1.16.x] bugfix: ItemStack#isDamageable is not calling Item#isDamagable

mcenderdragon:
Backstory: I made an Item getting more Durability with an Enchantment, so from getMaxDurability -> get enchanemtnent list -> other mods added dynamic enchanments -> they checked if the item was damage able -> itemStack#isDamagebale calls Item#getMaxDamage -> stackoverflow.

I talked with diesieben07 with this and this is the solution we came up with.


KnightMiner:
One concern I have is isDamagable is not stack sensitive. What if a tool is only damagable bases on NBT, like our Tinkers Construct pickaxes that have both a broken and a unbreakable tag? Does this handle the vanilla unbreakable tag?

Might be worth making a stack sensitive isDamagable method to solve that.

mcenderdragon:
@KnightMiner I have added a ItemStack paramaeter and patched the places where this got called. But it increases patch size alot, then again I total understand the need of this.

diesieben07:
> Does this handle the vanilla unbreakable tag?

It is already handled in `ItemStack#isDamageable`. I am not sure this patch is worth it. Many places in vanilla already call the stack-less `isDamageable` and apparently the tinkers tools work fine with that in place.

KnightMiner:
They work fine because we bypass most of vanillas methods when damaging them, which means we have to make sure another mod never tries to damage our tools (a particular problem with our "shears" or similar right click action tools). It would be a lot easier if we did not have to bypass vanilla logic and other mods can safely damage it.

diesieben07:
There is already a hook into the "damage item" functionality: `IForgeItem#damageItem`. If that returns 0 then no damage is performed at all, even if the item is damageable. Doesn't that solve the "other mods damaging" issue?

KnightMiner:
I am not certain if that handled the need or not since I am mostly used to our 1.12 code, where there was no damage item hook, just set damage. However, we have also had cases where mods that do thing beside damage tools want to know if its damagable, such as that one mod that modifies how durability bars display to show numbers.

Even ignoring Tinkers Construct though, my main comment was they were replacing a call to a stack sensitive method with a non-stack sensitive method, which means mods now have less context to determine if something can be damaged. A new method solves that problem by keeping the same amount of context in the method.

diesieben07:
> my main comment was they were replacing a call to a stack sensitive method with a non-stack sensitive method

That's not what's going on here though. What's going on here is that `ItemStack#isDamageable` calls `Item#getMaxDamage() > 0` incorrectly instead of `Item#isDamageable`. Whether `Item#isDamageable` is stack sensitive or not is a different issue.

KnightMiner:
[Yes, that is what the original commit did](https://github.com/MinecraftForge/MinecraftForge/pull/7221/commits/010e16d7156b60a7a6c1d137c8dee81284e27fc7). The call to `Item::getMaxDamage(ItemStack) > 0` was replaced with `Item::isDamageable()`. Yes, the second is more correct, but it removed the item stack context that was previously added in a Forge patch.

----

If we wanted to do this without adding a new stack sensitive method, and also in a way unlikely to break existing mods, one simple solution would be switching the condition to `item.isDamageable() && item.getMaxDamage(stack) > 0`, so no new method needed and minimal overall patches, but all existing context remains from before. However, pretty sure this does not fix mcenderdragon's problem with the stack overflow, hence my original suggestion.

mcenderdragon:
> If we wanted to do this without adding a new stack sensitive method, and also in a way unlikely to break existing mods, one simple solution would be switching the condition to item.isDamageable() && item.getMaxDamage(stack) > 0, so no new method needed and minimal overall patches, but all existing context remains from before. However, pretty sure this does not fix mcenderdragon's problem with the stack overflow, hence my original suggestion.

@KnightMiner actually no, the stackoverflow excpetion which triggered this PR would not be resolved as it would still call `getMaxDamage` which triggered the loop.

When I patched these calls I noticed how largly inconsistent vanilla was when calling `ItemStack#isDamageable` and `Item#isDmageable` so most likely the original intension of the methods are different.

mcenderdragon:
in 1.12 there is [Item#setDamage](https://github.com/MinecraftForge/MinecraftForge/blob/1.12.x/patches/minecraft/net/minecraft/item/Item.java.patch#L562) which can be used to fully control of a item takes damage or not.
And in 1.15+ (possibly earlier) we have `IForgeItem#damageItem` as diesieben07 pointet out. So I will undo the patch to add a ItemStack context to isDamageable. As it is not needed and everything can be done from other methods.

diesieben07:
After feedback from Discord, please revert to patching isDamageable to have ItemStack context.

mcenderdragon:
@diesieben07 ok done.

patches/minecraft/net/minecraft/item/MerchantOffer.java.patch
```diff
@@ -0,0 +1,11 @@
+--- a/net/minecraft/item/MerchantOffer.java
++++ b/net/minecraft/item/MerchantOffer.java
+@@ -178,7 +178,7 @@
+          return true;
+       } else {
+          ItemStack itemstack = p_222201_1_.func_77946_l();
+-         if (itemstack.func_77973_b().func_77645_m()) {
++         if (itemstack.func_77973_b().isDamageable(itemstack)) {
+             itemstack.func_196085_b(itemstack.func_77952_i());
```
LexManos, line 9: We typically use the ItemStack bouncer, as it makes cleaner code and prevents any potential issues of the argument being wrong, but this is fine for now.

PR #7222, opened 2020-08-05
[1.15.x] bugfix: ItemStack#isDamageable did not call Item#isDamageable

mcenderdragon:
Backport of #7221

### DIFF
--- a/patches/minecraft/net/minecraft/item/ItemStack.java.patch
+++ b/patches/minecraft/net/minecraft/item/ItemStack.java.patch
@@ -118,7 +118,7 @@
  
     public boolean func_77984_f() {
 -      if (!this.field_190928_g && this.func_77973_b().func_77612_l() > 0) {
-+      if (!this.field_190928_g && this.func_77973_b().getMaxDamage(this) > 0) {
++      if (!this.field_190928_g && this.func_77973_b().func_77645_m()) {
           CompoundNBT compoundnbt = this.func_77978_p();
           return compoundnbt == null || !compoundnbt.func_74767_n("Unbreakable");
        } else {


patches/minecraft/net/minecraft/item/ItemStack.java.patch
```diff
@@ -118,7 +118,7 @@
  
     public boolean func_77984_f() {
 -      if (!this.field_190928_g && this.func_77973_b().func_77612_l() > 0) {
-+      if (!this.field_190928_g && this.func_77973_b().getMaxDamage(this) > 0) {
++      if (!this.field_190928_g && this.func_77973_b().func_77645_m()) {
```
gigaherz, line 121: I'm not sure about this. The call to getMaxDamage is itemstack-aware, but it's being patched with a call that is **not** stack-aware, so it seems counter-productive?

mcenderdragon: Did I mess up the Backport ? Ahh, back then we did not have the isDamageable I think.

gigaherz: You added isDamageable(ItemStack) in the same PR, but didn't add that stack-sensitive method here, so the change is just kindof standing on one foot.

covers1624: @mcenderdragon Any update on these comments?

mcenderdragon: right now I dont have the energy to port/change this, It will hopefully get better at the end of april. then again it is in 1.16 so it is not as important if 1.15 is missing this.

gigaherz:
Closing this as the backport wouldn't make sense without the other changes done to 1.16, and those were breaking changes.

PR #7223, opened 2020-08-05
[1.15.x] Post EntityJoinWorldEvent when Lightning is added in ServerWorld.

L3MON4D3:
The EntityJoinWorldEvent for Lightning isn't posted in addLightningBolt. EntityConstructing is posted, however at that time the LightningBoltEntity isn't fully initialized and eg. position is missing.


diesieben07 (approved):

diesieben07:
Not an issue in 1.16 anymore as "global entities" are no longer separate there.

L3MON4D3:
Blocks around the impact got ignited even when the Event was canceled and the Lightning never actually added to the World, should be fixed now.

PR #7225, opened 2020-08-06
[1.16] Rework RenderWorldLastEvent

Nightenom:
The old way became less programmer-friendly as of recent rendering updates in vanilla code.
The new way adds events before finishing of important render buffers and creates an event for each rendered block layer. The old RenderWorldLastEvent was moved to the end of World rendering method.

I'm not sure if this PR needs a test mod since it's only event change.
~I have not carried over profiler section calls because of patch lengths. If you think it's better to separate forge events from vanilla profiling, I can add them into client forge hooks.~ done

_Sequel of #6985 as I failed to push into it_
_Superseded by and closes #7661_

### DIFF
--- a/patches/minecraft/net/minecraft/client/renderer/GameRenderer.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/GameRenderer.java.patch
@@ -27,7 +27,7 @@
              } catch (Throwable throwable) {
                 CrashReport crashreport1 = CrashReport.func_85055_a(throwable, "Rendering screen");
                 CrashReportCategory crashreportcategory1 = crashreport1.func_85058_a("Screen render details");
-@@ -595,9 +597,16 @@
+@@ -595,6 +597,11 @@
        Matrix4f matrix4f = matrixstack.func_227866_c_().func_227870_a_();
        this.func_228379_a_(matrix4f);
        activerenderinfo.func_216772_a(this.field_78531_r.field_71441_e, (Entity)(this.field_78531_r.func_175606_aa() == null ? this.field_78531_r.field_71439_g : this.field_78531_r.func_175606_aa()), !this.field_78531_r.field_71474_y.func_243230_g().func_243192_a(), this.field_78531_r.field_71474_y.func_243230_g().func_243193_b(), p_228378_1_);
@@ -39,12 +39,7 @@
        p_228378_4_.func_227863_a_(Vector3f.field_229179_b_.func_229187_a_(activerenderinfo.func_216777_e()));
        p_228378_4_.func_227863_a_(Vector3f.field_229181_d_.func_229187_a_(activerenderinfo.func_216778_f() + 180.0F));
        this.field_78531_r.field_71438_f.func_228426_a_(p_228378_4_, p_228378_1_, p_228378_2_, flag, activerenderinfo, this, this.field_78513_d, matrix4f);
-+      this.field_78531_r.func_213239_aq().func_219895_b("forge_render_last");
-+      net.minecraftforge.client.ForgeHooksClient.dispatchRenderLast(this.field_78531_r.field_71438_f, p_228378_4_, p_228378_1_, matrix4f, p_228378_2_);
-       this.field_78531_r.func_213239_aq().func_219895_b("hand");
-       if (this.field_175074_C) {
-          RenderSystem.clear(256, Minecraft.field_142025_a);
-@@ -709,4 +718,9 @@
+@@ -709,4 +716,9 @@
     public OverlayTexture func_228385_m_() {
        return this.field_228375_t_;
     }
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@
 **/.classpath
 *.project
 
+#vscode
+/.vscode
+
 #In case people make their workspace in the repo
 /eclipse/
 
--- a/patches/minecraft/net/minecraft/client/renderer/WorldRenderer.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/WorldRenderer.java.patch
@@ -137,7 +137,7 @@
        RenderSystem.disableBlend();
        RenderSystem.popMatrix();
        FogRenderer.func_228370_a_();
-+      net.minecraftforge.client.ForgeHooksClient.dispatchRenderWorldLast(this, p_228426_1_, p_228426_2_, clippinghelper, p_228426_6_, field_228415_m_);
++      net.minecraftforge.client.ForgeHooksClient.dispatchRenderWorldLast(this, p_228426_1_, p_228426_2_, clippinghelper, p_228426_6_, field_228415_m_, p_228426_9_, p_228426_3_);
     }
  
     private void func_228423_a_(MatrixStack p_228423_1_) {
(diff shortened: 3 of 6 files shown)


diesieben07:
I am not sure if "fire even more events every single frame" is a good idea.

Nightenom:
The problem here is vanilla has started implementing shaders. Also slowly moving to OpenGL 3+ mechanisms - adding more framebuffers etc. Thus we need to hook more precisely into rendering pipeline than before.
I've thought about some kind of registry-like approach with rendering hooks as an interface. There is already block-breaking event and weather interfaces, so it would be around 10 hooks.

Nightenom:
List of hookable places in world rendering in order as they appear in rendering pipeline (ticked being what we have now):
- [x] sky rendering, status: open PR of old patches #6994
- [ ] terrain setup + culling, usage: delay setting up your render things
- [ ] block layers, usage: rendering of non-existing things (eg. structure previews), special rendering
- [ ] entities, usage: same as block layers
- [ ] tile entities, usage: same as block layers
- [ ] destroy progress (blocks), usage: none?
- [x] highlighted block, status: DrawHighlightEvent
- [ ] source buffer end, usage: line boxes, glint render types, ...
- [x] particles, status: has standalone manager useable by mods
- [x] cloud rendering, status: open PR of old patches #6994
- [x] weather rendering, status: open PR of old patches #6994
- [x] last render hook, status: RenderWorldLastEvent (needs moving to WorldRenderer from GameRenderer)

malte0811:
What is the current state of this PR? From the amount of linked issues/PRs quite a few people (including me :) ) are interested in this feature or subsets of it, I think someone from MineColonies was interested as well.

marchermans:
Yeah this comes from our team.
@Nightenom What is the state here?

Nightenom:
I can rebase it at any time. All I am/was waiting for is whether all of those events make sense, Diesieben and also me are wondering whether it's too much or not to fire 12 events per ~tick~frame just because of rendering purposes

gigaherz:
We'd ideally need to have some benchmark numbers. Eg. like, build a forge from that, and use it in a large modpack world, and see what the numbers look like side by side with the base version of forge (like, 35.0.7 vs 35.0.7+renderevenrs).
I'd love to have these events, but there's already enough people who feel like you can't play modpacks without optimizations mods, so we'd rather not actually make it worse.

malte0811:
> use it in a large modpack world

I feel like that's not the right test case. In a large pack there's already so much going on each frame that I doubt you'd be able to see this *at all*, it would just get lost in the noise. To me the right test case would be something like a newly generated world with no mods.

Nightenom:
**Changes to PR:**
added profiler calls
make entire PR non-breaking with deprecated marks

**Also, I've made testing:**
generated a new vanilla world in forge environment, during first 30 seconds after load I've been rotating with the camera to force-load and setup render cache, after that, I kept rotating with the camera for 1 minute and took the following results (everything in nanoseconds):
- flat world with nothing but flat grass and few entities, BEFORE event patch:
` Min: 1164300 | Max: 05149300 | Avg: 1342970,64 `
- normal world with the camera being roughly 30 blocks above the ground, BEFORE event patch:
` Min: 2545800 | Max: 16928900 | Avg: 4328035,70 `
- normal world with the camera being roughly 30 blocks above the ground, AFTER event patch:
` Min: 2986800 | Max: 28499000 | Avg: 5472149,16 `

**The result is rendering with all events being called is roughly 1ms slower**
Note that the test definitely wasn't as clean as possible

Setup info:
- BEFORE patch:
latest main branch (1.16.4 - 35.0.15) with updateCameraAndRender and old RenderWorldLastEvent wrapped by time measuring code
- AFTER patch:
same as above + all events being called and all events were subscribed by one dummy listener with simple logger message producing this output per frame:
```
[Render thread/INFO]: RenderWorldTerrainUpdateEvent
[Render thread/INFO]: RenderWorldBlockLayerEvent
[Render thread/INFO]: RenderWorldBlockLayerEvent
[Render thread/INFO]: RenderWorldBlockLayerEvent
[Render thread/INFO]: RenderWorldEntitiesEvent
[Render thread/INFO]: RenderWorldTileEntitiesEvent
[Render thread/INFO]: RenderWorldEndEvent
[Render thread/INFO]: RenderWorldBlockLayerEvent
[Render thread/INFO]: RenderWorldBlockLayerEvent
[Render thread/INFO]: RenderWorldPreWeatherEvent
[Render thread/INFO]: RenderWorldLastEvent
```

CPU Intel(R) Core(TM) i5-3350P CPU @ 3.10GHz 3.70 GHz, RAM 16.0 GB, GPU GTX 750 Ti, w10 pro 20H2, Java: 11.0.8 JVM: 11.0.8+10(AdoptOpenJDK) Arch: amd64

malte0811:
1 ms is quite a lot when compared to the "target" frame time (16.6 ms). Can you test with something faster than debug printouts? I would guess that they account for a majority of that 1 ms. Maybe increment one counter per event and log when the counter is divisible by 512?  
Most implementations of these events would probably be something like "check if list x is nonempty, in the (relatively rare) case that it is run some rendering for each list entry", which is much faster in the typical (empty list) case.

Nightenom:
Here you go:
before
Min: 3059400 | Max: 14908600 | Avg: 4587204,83
after
Min: 3061700 | Max: 18209400 | Avg: 4734080,33

As you can note, both averages changed significantly (I would say my nearly 7 years old pc is quite unstable for such precise measurements). Even though I agree on loggers might being overkill I've used them to try to simulate a few mods using a random set of these events.

Nightenom:
_Note: my pc seems to be limited by 100ns precision on System#nanoTime()_
However, measuring posting 11 events per frame during 6000 frames (100 seconds on 60Hz) resulted in `Min: 5500 | Max: 8511400 | Avg: 11799,63`. Event listeners were made of a few math ops and conditional logging.
From this, I would say it's pointless to worry about how much few events can take. ~<1050ns per event posting is nothing but micro-optimization

edit: Side note: I've learned from this that people who care about huge modpacks being slow should yell (after an investigation) at mod programmers which should take a look at the heaviest parts of mods and try actually to optimize them. Or just don't make modpacks with an enormous amount of mods...

patches/minecraft/net/minecraft/client/renderer/GameRenderer.java.patch
```diff
@@ -39,12 +39,7 @@
        p_228378_4_.func_227863_a_(Vector3f.field_229179_b_.func_229187_a_(activerenderinfo.func_216777_e()));
        p_228378_4_.func_227863_a_(Vector3f.field_229181_d_.func_229187_a_(activerenderinfo.func_216778_f() + 180.0F));
        this.field_78531_r.field_71438_f.func_228426_a_(p_228378_4_, p_228378_1_, p_228378_2_, flag, activerenderinfo, this, this.field_78513_d, matrix4f);
-+      this.field_78531_r.func_213239_aq().func_219895_b("forge_render_last");
-+      net.minecraftforge.client.ForgeHooksClient.dispatchRenderLast(this.field_78531_r.field_71438_f, p_228378_4_, p_228378_1_, matrix4f, p_228378_2_);
```
pupnewfster, line 43: I am slightly unsure if this would constitute a breaking change or not, given at least for fast/fancy (where this *does* work), any mod that is currently listening to the existing RenderLastEvent would stop having it fired and I believe have their code get "broken". So we may need to leave this in, or if the new render last event is equivalent in terms of when/how it fires, have the new one still just fire the old event instead of the new one until 1.17

Nightenom: The new RenderWorldLastEvent is a 1:1 replacement, will take a look at how to merge them - possibly can fire the old one using the new firing method but projectionMatrix in the old event would have to be null (I doubt anyone managed to use it except me after I've added it) other parameters are available, but for sure we can fire both of them until 1.17

Nightenom: My bad - all arguments are available, event fully revived and fired after the new one.

TheCurle:
This PR is superseded, so i'll close it until further notice.

PR #7228, opened 2020-08-08
Improve performance and cleanup code for DelegatingResourcePack

### LINKED ISSUE #7227: Performance of DelegatingResourcePack in larger modpacks
**Minecraft Version:** 1.14, 1.16 (presumably 1.15 as well, but not tested there)

**Forge Version:** 32.0.106, but probably affects every available version on 1.14, 1.15 and 1.16.

**Steps to Reproduce:**
 1. Load a small modpack (2-3 mods)
 2. Measure the speed of `DelegatingResourcePack#getInputStream` for existing files in different mod namespaces
3. Load a larger modpack (>50 mods with their own resource packs)
4. Measure again

**Description of issue:**
`DelegatingResourcePack#getInputStream` and `DelegatingResourcePack#resourceExists` check every delegate even when only very few of them contain the relevant namespace. This can lead to significant slowdown even in medium-size modpacks. Most of this time ends up in `java.nio.file.Files#exists`.  
The Immersive Engineering manual is loaded from resource and data packs the first time the manual is opened. In small modpacks this does not cause a noticeable lag spike (after some optimization in the latest version), but a user reported a 2 second spike with 40 mods. In my own tests it wasn't as bad (34 seconds for loading the manual 100 times in a larger pack), but still a lot worse than in my dev setup (<10 seconds). By checking the mod owning the namespace first I was able to speed this up to 3-4 seconds in both cases (Note: we still support 1.14, so I had to implement a workaround on our side in any case).

In Forge the fix should probably be implemented differently, similar to what is already done for vanilla resource packs: Keep a map of all packs/mods containing resources for a namespace and only check those. For most mod namespaces this should mean checking just a single delegate.

diesieben07:
- `DelegatingResourcePack` will now only query relevant packs for resources, based on `getResourceNamespaces`.
- Removed a lot of the synchronization in `DelegatingResourcePack`, `addDelegate` is deprecated.
- `DelegatableResourcePack` is deprecated and basically unused now. To be removed in 1.17.

Fixes #7227.


src/main/java/net/minecraftforge/fml/packs/DelegatingResourcePack.java
```diff
@@ -84,69 +111,86 @@ public String getName()
...
     {
-        synchronized (delegates)
-        {
-            return delegates.stream()
-                    .flatMap(r -> r.getResourceNamespaces(type).stream())
-                    .collect(Collectors.toSet());
-        }
+        return delegates.stream()
+                .flatMap(r -> r.getResourceNamespaces(type).stream())
+                .collect(Collectors.toSet());
```
malte0811: Could this be changed to use `namespaceAssets.keySet()`/`namespaceData.keyset()`? IIRC `ModFileResourcePack#getResourceNamespaces` does file system operations, so it's relatively slow.

malte0811 (commented):
Looks good aside from (possibly) one missed minor optimization. Thank you for implementing this so quickly!

src/main/java/net/minecraftforge/fml/packs/DelegatingResourcePack.java
diesieben07: You're right, changed that.

PR #7232, opened 2020-08-09
Add hook for structure separation settings

### LINKED ISSUE #6996: [1.16.x] It is not possible to add/modify StructureSeparationSettings for Structures
**Minecraft Version:** 1.16.1

**Forge Version:** 32.0.67

Minecraft changed from a getter method in each Structure to get the structure generation seperation to a map at `DimensionStructuresSettings#field_236191_b_` which is final & Immutable, so it can not be modified (except with incompatible methods).

But if you can edit it, i does not help because the field is copied and used before even the mod constructors are constructed.

This results in every custom structure is generated in every chunk or more, because there exists no StructureSeparationSettings (DimensionStructuresSettings#42).

To be able to modify add temporally it is possible to add the settings eg. to `DimensionSettings.Preset.field_236122_b_ ` but that is not threadsafe

diesieben07:
Fixes #6996.


patches/minecraft/net/minecraft/world/gen/DimensionSettings.java.patch
```diff
@@ -0,0 +1,11 @@
+--- a/net/minecraft/world/gen/DimensionSettings.java
++++ b/net/minecraft/world/gen/DimensionSettings.java
+@@ -104,7 +104,7 @@
+    }
+ 
+    private static DimensionSettings func_242741_a(DimensionStructuresSettings p_242741_0_, BlockState p_242741_1_, BlockState p_242741_2_, ResourceLocation p_242741_3_) {
+-      Map<Structure<?>, StructureSeparationSettings> map = Maps.newHashMap(DimensionStructuresSettings.field_236191_b_);
++      Map<Structure<?>, StructureSeparationSettings> map = net.minecraftforge.common.ForgeHooks.getAllStructureSeparationSettings();
```
ichttt, line 8: A problem that might occur with this is that the method is called during `clinit`, and if a mod causes classloading of this before things are registered it will break

PR #7233, opened 2020-08-09
[1.16.x] DrawHighlightEvent$HighlightEntity Firing

ChampionAsh5357:
`DrawHighlightEvent$HighlightEntity` is never fired by default. The only reference to the event is via `ForgeHooksClient::onDrawBlockHighlight` which as the name implies is only fired for block highlighting. Highlighting is defined differently for an entity compared to a block. For a block, highlighting is defined as the selection box that is rendered when the block is moused over. ~~Entities (as I have interpreted) refer to when a highlight effect is applied via spectral arrow or some other effect. Since this is applied in a loop rather than as a result of an entity hit, it is never considered or attached to be canceled.~~

~~This PR fixes that issue by adding the hook such that the entity box can be applied and canceled. To also allow for the event to be used rather than making an entirely new one, an `EntityRayTraceResult` is passed in requiring as little modification as need to the actual patch.~~

~~I did nothing to cancel the debug collision boxes as that already has its own method via `F3 + B`. ~~

I'm not sure whether it would be better to pass in a null value and remove the ability to grab the ray trace result as the entity was never technically hit. However, since that would cause a breaking change, I just passed in a new instance containing the entity.~~

I reimplemented the method for its original case that entities act like a block to have a selection drawn for it.


ichttt:
This really is a weird event now that I think about it... Do you actually need the event? Otherwise I would just remove it

pau101:
The semantics of the event should remain as originally intended, enabling drawing custom selection boxes for any mouse over hit result type (which has been possible for the 8 years of its existence). I have a mod which used this feature which when porting to 1.15 I had to reinstate myself to retain the behavior. Here is my selection box drawing for my mod entity:

image

ChampionAsh5357:
Since there is a use-case, I have updated this to 1.16.2.

pau101:
@ChampionAsh5357 I think you misunderstood my comment.

ChampionAsh5357:
> @ChampionAsh5357 I think you misunderstood my comment.

@pau101 I probably have. Since you can create the selection box without the need of the event, I'm guessing the event is just pointless. You were probably just mentioning that the idea should remain intact; however, the event is not necessary to implement in this case. I could be wrong on my interpretation again. Entities don't technically have a selection box anyways as they are usually determined as the highlight around so this implementation might just not be relevant and should be removed.

pau101:
@ChampionAsh5357 The purpose of the `DrawHighlightEvent` event is to allow replacing or removing the drawing of a selection box for whatever the mouse over hit result is. In my mod's case I have an entity which acts like a block so I create my own selection box rendering to visualize how the entity can be interacted with.

Trying to render a selection box otherwise will not appear as one should given draw order and interaction with visual enhancement mods. I'll like to repeat this event has operated this way for 8 years and reinterpreting it instead as an entity outline shader event is an quite an odd change.

ChampionAsh5357:
@pau101 Ah ok. My misinterpretation. My choice to do it this way was mainly an interpretation of how a highlight event for an entity would be. There are only two boxes an entity has a "selection" for whether its either highlighted by an arrow or enabled via debug. `DrawHighlightEvent` is ambiguous in this regard by what you're interpreting it as. If it was supposed to be only for the hit result of the mouse over, it should have been referenced as `DrawSelectionEvent` rather than highlight.

Personally, what I'm arguing is a mute point. There is technically no reason to have the event for entities as they don't really have a mouse over selection box. As ichttt said, if there's no need for the event, it can just be removed. And since the event is not needed or will probably never be implemented in the way you describe, it might just be better to remove it.

So its more of an argument whether or not to interpret `DrawHighlightEvent` as what it is or rename in a later version and remove the unused entity implementation altogether.

pau101:
@ChampionAsh5357 This PR could reinstate the event to original behavior but if you're not up to that yourself that is fine. The problem is the event was just never added again when porting to 1.15. The appropriate action to me is to put the patch back in to reenable the event as it worked for the 8 years since creation.

ChampionAsh5357:
@pau101 I guess I'll get started on that then. Makes more sense than the current usage. I just need to do some testing to see how it was originally implemented. I'll also leave a comment to rename `DrawHighlightEvent` to `DrawSelectionEvent` for less ambiguity as its two subclasses both already specify its for highlighting the block or entity.

PR #7234, opened 2020-08-10
[1.16.x] SoundEvent Deprecations For Removal

ChampionAsh5357:
For anyone who misunderstands, I am talking about forge's `SoundEvent` used to hook into the `SoundEngine`. As @sciwhiz12 has reviewed, a few of the events are currently either unused or don't have any purpose.

First, ~`SoundLoadEvent` and~ `SoundSetupEvent` serve no purpose. `SoundSetupEvent` was taken over by `SoundLoadEvent` entirely and was never updated to the mod event bus only. This causes the method to be executed on both event buses. ~`SoundLoadEvent` no longer serves any purpose though since custom `ISound`s can just be created and executed. The reload only attaches the sounds from `SoundEvent`s anyways which negates the purpose of both events.~ `SoundLoadEvent` has a use case for initializing and reloading access to the `SoundEngine`.

Second, the `getName` methods serve little purpose as well. They only output the path of the sound which is just bad in general. Instead, it should be called via `ISound::getSoundLocation` with the instance provided.

These methods/events have been marked deprecated and for removal in the next version. I did include a patch for `SoundLoadEvent` to fix the ModLoader issue though. Even if it's marked for removal, it still should function correctly in this version.


josephtarbit:
SoundLoadEvent is only posted on the mod bus  #6955. Also the purpose of the SoundLoadEvent is to gain access to the sound engine which isn’t accessible anywhere else which I use to load in the EFX extension for OpenAL. The alternative without using the SoundLoadEvent would be ATing or using reflection to access the sound engine and then initialising your sound stuff in the client tick event as no load events are fired after the sound engine loads.

ChampionAsh5357:
> Also the purpose of the SoundLoadEvent is to gain access to the sound engine which isn’t accessible anywhere else which I use to load in the EFX extension for OpenAL.

Ah okay. sci and I were trying to find a use case yesterday and it seems this was it. It's better than having to go through the other events each time a sound is played too.

> SoundLoadEvent is only posted on the mod bus #6955.

Well half correct. `SoundLoadEvent` is posted on the mod bus in the constructor. However, it is also used to reload the sounds when the reloadable listener is called. That part is called on the forge event bus. Because of that, I updated the event bus for the reload method to the mod event bus.

josephtarbit:
> However, it is also used to reload the sounds when the reloadable listener is called.

Ahh must of missed that. Also you might want to make the `SoundLoadEvent` implement `IModBusEvent` as that wasn't a thing when I wrote my PR so I missed that out.

ChampionAsh5357:
> Ahh must of missed that. Also you might want to make the `SoundLoadEvent` implement `IModBusEvent` as that wasn't a thing when I wrote my PR so I missed that out.

I believe that was already done by someone else since its already merged with forge.

josephtarbit:
> > Ahh must of missed that. Also you might want to make the `SoundLoadEvent` implement `IModBusEvent` as that wasn't a thing when I wrote my PR so I missed that out.
> 
> I believe that was already done by someone else since its already merged with forge.

I'm blind 🤦‍♂️

ichttt:
Please update for 1.16.2

ChampionAsh5357:
Should be all good now

Mysterious-Dev (approved):
That look good

ichttt:
I see `TODO: 1.16.2 Remove`, but this PR is targeting 1.16.2

ChampionAsh5357:
Apologies, PR was for 1.16.1 to be fixed in 1.16.2. Updated to 1.17 removal.

TheCurle:
This all seems to still be in order.
Having more reminders as we get closer to the 1.17 update can only help.

ChampionAsh5357:
Apologies, exploded, trying to revert and adjust.

ChampionAsh5357:
I'll create a new one to fix this later, accidentally just broke this.

PR #7235, opened 2020-08-10
[1.16.x] Environment tick event

mrp-v2:
This event provides a way for mods to perform weather and random tick based actions at the same time as vanilla (not actually the same time, either just before or just after).
Test mod included.

### DIFF
--- a/patches/minecraft/net/minecraft/world/server/ServerChunkProvider.java.patch
+++ b/patches/minecraft/net/minecraft/world/server/ServerChunkProvider.java.patch
@@ -1,6 +1,6 @@
 --- a/net/minecraft/world/server/ServerChunkProvider.java
 +++ b/net/minecraft/world/server/ServerChunkProvider.java
-@@ -353,7 +_,7 @@
+@@ -353,13 +_,15 @@
                 if (optional1.isPresent()) {
                    Chunk chunk = optional1.get();
                    ChunkPos chunkpos = p_241099_7_.func_219277_h();
@@ -9,6 +9,14 @@
                       chunk.func_177415_c(chunk.func_177416_w() + j);
                       if (flag1 && (this.field_217246_l || this.field_217247_m) && this.field_73251_h.func_175723_af().func_177730_a(chunk.func_76632_l())) {
                          WorldEntitySpawner.func_234979_a_(this.field_73251_h, chunk, worldentityspawner$entitydensitymanager, this.field_217247_m, this.field_217246_l, flag2);
+                      }
+ 
++                     net.minecraftforge.fml.hooks.BasicEventHooks.onPreEnvironmentTick(this.field_73251_h,chunk, k);
+                      this.field_73251_h.func_217441_a(chunk, k);
++                     net.minecraftforge.fml.hooks.BasicEventHooks.onPostEnvironmentTick(this.field_73251_h, chunk, k);
+                   }
+                }
+             }
 @@ -427,6 +_,14 @@
  
     public <T> void func_217222_b(TicketType<T> p_217222_1_, ChunkPos p_217222_2_, int p_217222_3_, T p_217222_4_) {
--- a/src/main/java/net/minecraftforge/fml/hooks/BasicEventHooks.java
+++ b/src/main/java/net/minecraftforge/fml/hooks/BasicEventHooks.java
@@ -25,6 +25,7 @@ import net.minecraft.inventory.IInventory;
 import net.minecraft.item.ItemStack;
 import net.minecraft.util.RegistryKey;
 import net.minecraft.world.World;
+import net.minecraft.world.chunk.Chunk;
 import net.minecraftforge.client.model.animation.Animation;
 import net.minecraftforge.common.MinecraftForge;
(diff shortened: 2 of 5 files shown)


KnightMiner:
Why not make the pre event cancelable, canceling can be used to prevent the random tick from being processed in vanilla logic.

Also, can you give an example usage of this event?

mrp-v2:
Canceling the event would also cancel lightning, skeleton horse traps, water freezing in cold biomes, snow layers generating during snow, blocks like cauldrons filling with water during rain, as well as random ticks. Also, other tick events aren't cancelable, so it would be inconsistent for this one to be.

An example usage would be if a mod that adds custom biomes/world generation and wanted to have a custom precipitation that caused a block to form on the surface of the world, like snow causes snow to form. Or if a mod wanted to give blocks extra random ticks. If a mod wants something to be caused by weather, this event would be the way to do it.

The specific inspiration for this event is that I have a mod in which I want to modify snowy weather so that snow layers can accumulate multiple layers instead of being limited to just 1 layer.

patches/minecraft/net/minecraft/world/server/ServerChunkProvider.java.patch
```diff
@@ -0,0 +1,13 @@
+--- a/net/minecraft/world/server/ServerChunkProvider.java
++++ b/net/minecraft/world/server/ServerChunkProvider.java
+@@ -358,8 +358,9 @@
+                      if (flag1 && (this.field_217246_l || this.field_217247_m) && this.field_73251_h.func_175723_af().func_177730_a(chunk.func_76632_l())) {
+                         WorldEntitySpawner.func_234979_a_(this.field_73251_h, chunk, worldentityspawner$entitydensitymanager, this.field_217247_m, this.field_217246_l, flag2);
+                      }
+-
```
ichttt: Please keep the newline to reduce patch size

src/test/java/net/minecraftforge/debug/world/EnvironmentTickEventTest.java
```diff
@@ -0,0 +1,25 @@
...
+import net.minecraftforge.fml.common.Mod;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+@Mod("environmenttickeventtest")
+public class EnvironmentTickEventTest
+{
+    private final Logger LOGGER;
+    
+    public EnvironmentTickEventTest(){
```
ichttt: Please put the brackets on a new line

src/test/java/net/minecraftforge/debug/world/EnvironmentTickEventTest.java
```diff
@@ -0,0 +1,25 @@
...
+        LOGGER = LogManager.getLogger();
+        MinecraftForge.EVENT_BUS.register(this);
+    }
+    
+    @SubscribeEvent
+    public void environmentTickEvent(final TickEvent.WorldTickEvent.EnvironmentTickEvent event){
```
ichttt: Same here. please fix the code style

ichttt (commented):
Some code style notes

diesieben07:
Extending WorldTickEvent is a breaking change. Mods subscribed to WorldTickEvent will now receive this event as well, which will throw off tick-based timing.

mrp-v2:
Should I extend TickEvent instead?

src/main/java/net/minecraftforge/fml/hooks/BasicEventHooks.java
```diff
@@ -119,4 +120,14 @@ public static void onPostServerTick()
...
     }
+
+    public static void onPreEnvironmentTick(World world, Chunk chunk, int randomTickSpeed)
+    {
+        MinecraftForge.EVENT_BUS.post(new TickEvent.WorldTickEvent.EnvironmentTickEvent(LogicalSide.SERVER, TickEvent.Phase.START, world, chunk, randomTickSpeed));
+    }
+    
+    public static void onPostEnvironmentTick(World world, Chunk chunk, int randomTickSpeed)
+    {
+        MinecraftForge.EVENT_BUS.post(new TickEvent.WorldTickEvent.EnvironmentTickEvent(LogicalSide.SERVER, TickEvent.Phase.END, world, chunk, randomTickSpeed));
```
alcatrazEscapee:
```java
new TickEvent.WorldTickEvent.EnvironmentTickEvent
```
looks wrong here, seeing that `EnvironmentTickEvent` now directly extends `TickEvent`. It should just be `new TickEvent.EnvironmentTickEvent()` for simplicity, even if this does technically compile just fine.

sciwhiz12:
Hello! @mrp-v2, please leave a comment whether you wish to continue this PR for 1.16. Note that there is already an updated version of the feature intended by this PR at #8314, so it is your decision whether to discontinue updating this PR in favor of that or to continue with your own implementation. (We suggest the former so there is a concentration of effort in one PR rather than fragmented across multiple.) If we do not hear back from you after some time, this PR will be closed for inactivity.

sciwhiz12:

As a note, PR #8314 seems to implement a similar feature to this PR.

PR #7236, opened 2020-08-10
Removed sidedness from PacketBuffer methods.

DaemonUmbra:
Affected methods:
* readLongArray (x2)
* readSectionPos
* readString (no-arg variant)


pau101:
There should be little reason to use the no-arg `PacketBuffer#readString()` on a server, requiring an explicit max length sounds appropriate for server-side validation. Same goes for `readLongArray`.

ichttt:
I agree with @pau101, the server should only use the readString and readLongArray variants with maxSize, so the server can pre-validate the max expected length

TheCurle:
There are a surprising amount of mods still making this mistake. @DaemonUmbra, is this still something you want added?

TheCurle:
Amusingly, 1.17 includes this change as a side effect of them removing class-level code stripping.

PR #7239, opened 2020-08-12
[1.16.x] Fixes for Bamboo, Enchantments and Conduits, small adjustements to bring some patches closer to the vanilla code.

EdwinMindcraft:
**Fixes:**
BambooBlock:
The growth check (random.nextInt(3) == 0) was fired twice.
ConduitTileEntity: The check to see if a block was a conduit frame ran 4 times, one for each block in the array, effectively counting valid blocks 4 times.
EnchantmentHelper: Minecraft 1.16 added a new condition to check for when applying enchantments, that condition was discarded in the update process.

**Adjustments:**
AbstractBlock: Replaced a != with func_203425_a (isIn)
AbstractRailBlock: Same thing. (railShape2 is still unused)
TurtleEgg: Removed a redundant check for LivingEntity.


LexManos:
I think the rebase derped. Sorry, needs updating.

EdwinMindcraft:
Sorry about that, it should be correct now.

PR #7241, opened 2020-08-13
Print mod file name when erroring from ModFileInfo

spnda:
I just tried switching from 1.16.1 to 1.16.2 and in 1.16.2 a license inside a mod is being enforced. Now, when I got the error I had no idea what mod that came from. And I can imagine that probably many people will have to go through all mods before they can find a culprit. Printing the mod file name when printing the error should make this quite a lot easier. I have also added it to other errors inside of ModFileInfo.java.

If anyone has any suggestions how to make this error message better, let me know. Also, this is completely untested.

This is what the error currently looks like:
```
net.minecraftforge.fml.loading.moddiscovery.InvalidModFileException: Missing License, please supply a license.
 	at net.minecraftforge.fml.loading.moddiscovery.ModFileInfo.lambda$new$2(ModFileInfo.java:68)
[stacktrace]
```

### DIFF
--- a/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/InvalidModFileException.java
+++ b/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/InvalidModFileException.java
@@ -27,7 +27,7 @@ public class InvalidModFileException extends RuntimeException
 
     public InvalidModFileException(String message, IModFileInfo modFileInfo)
     {
-        super(message);
+        super(String.format("%s (%s)", message, ((ModFileInfo)modFileInfo).getFile().getFileName()));
         this.modFileInfo = modFileInfo;
     }
 }


LexManos:
Probably better to just do it in InvalidModFileException itself. And catch all the other places.

spnda:
> And catch all the other places.

Do you mean adding this to various other error messages related to mods?

LexManos:
They all run through the same exception class. And they all pass in the mod info.
So editing the string in the exception class seems like a logical place instead of all over the place.

spnda:
> 
> 
> They all run through the same exception class. And they all pass in the mod info.
> So editing the string in the exception class seems like a logical place instead of all over the place.

I had that anyway. From your first reply I thought you meant all other mod exceptions other than InvalidModFileException should also get the treatment. I think this is good to go now though.

TheCurle:
It looks to me like that's his intention, yes. In the place where all the exceptions are handled, handle all the exceptions. Tell the user which mod failed, because some of them are still cryptic for other reasons.

LexManos:
```
/var/jenkins_home/workspace/forge_forge_PR-7241/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/InvalidModFileException.java:30: error: cannot find symbol
        super(String.format("%s (%s)", message, (ModFileInfo)modFileInfo.getFile().getFileName()));
                                                                        ^
  symbol:   method getFile()
  location: variable modFileInfo of type IModFileInfo```

PR #7245, opened 2020-08-13
Fix Forge 1.16.2 Server connection

Cyborgmas:
Misapplied patch skipped brigadier argument registration. And tag registry creation seemingly.


LexManos:
Yay patcher just randomly deciding this spot looks nice...
Covers is working on a fixup for that.

PR #7246, opened 2020-08-13
1.16.2 Remove forge optional tags

Cyborgmas:
Vanilla now has support for optional tag entries, as such forge optional tags aren't needed anymore. see https://minecraft.gamepedia.com/Tag#JSON_format.
However, this means previous tags made this way would be incompatible. So perhaps waiting for 1.17 might be desired?


LexManos:
Could we add an adapter like Mojang does for old data pack versions?

Cyborgmas:
Woops i actually didnt remove the "optional" entry deserialization (https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/src/main/java/net/minecraftforge/common/ForgeHooks.java#L1081-L1091) 
So in theory it will read the optional array and create the needed tags from it. But I would have to test.
Not sure how adapted you want, is being able to parse existing tag files enough? (even though codewise, all will break).

LexManos:
If we can parse existing tags, and translate them to the format vanilla wants. Then that'd be golden.
I dont remember off hand the actual formats, Forge vs vanilla.
But could be just moving our optional entries into whatever section vanilla is looking for them?

Cyborgmas:
Okay tested and it gets parsed no problem.
For reference, the diff between forge and vanilla opt : https://gist.github.com/Cyborgmas/9104ab2715151fbd3433345ea88fc712
Deleted the biome test mod, since it was crashing the test run, fixed up the data gen test mod for tags as well.

PR #7250, opened 2020-08-14
[1.16.2] Mark PointOfView as an extensible enum

Exopandora:
This PR marks PointOfView as an extensible enum as the title says. It allows modders to create a custom POV/perspective. This is required because otherwise there is no good way for modders to create a new POV/perspective in 1.16.2.

Sample mod without any real use:
```
import net.minecraft.client.settings.PointOfView;
import net.minecraftforge.fml.common.Mod;

@Mod("point_of_view_enum_test")
public class PointOfViewEnumTest 
{
    private static final boolean ENABLE = false;
	
    public PointOfViewTest()
    {
        if (ENABLE)
            PointOfView.create("first_person_front", true, true);
    }
}
```


ichttt:
I don't know if this is a good way to do this, as the only possible new value that makes sense is first person front...

Exopandora:
I want to add a new third person mode without overriding the vanilla 3pp. This PR looks like the only good way to implement it. It would also simplify a lot of things for me.

Issue #7249, opened 2020-08-14
(1.16.2) Nether Portals can't be created


ghost:
Minecraft: 1.16.2
Forge Version: 1.16.2-33.0.2
Log: https://gist.github.com/TheAGamePlayer/92994f63970e09bfe0c2c3b86666f94a
Steps to Reproduce:
1) Make a nether portal
2) Attempt to ignite it
3) Wonder why it doesn't work >.>

Description:
When updating my mod to the new 1.16.2 version, I realized you could not go into the nether. Nether Portals would not ignite regardless of what I would use to ignite it. I provided a gif as video evidence containing the broken feature.
![bug](https://user-images.githubusercontent.com/58047374/90209428-44fa9480-dda0-11ea-945f-0126fd57b40a.gif)

LexManos:
https://github.com/MinecraftForge/MinecraftForge/commit/680c4062f49c07eb1b53a39ade89df217f7b2223

PR #7251, opened 2020-08-14
Fixed nether portal not spawning in 1.16.2

Yunus1903:
As the title says.
IEventBus#post returns true if the event is canceled. This means the return values in ForgeEventFactory#onTrySpawnPortal were swapped and thus the portal wouldn't spawn.

this also closes issue https://github.com/MinecraftForge/MinecraftForge/issues/7249


LexManos:
Ya, I fixed this shortly after waking up this morning.
Currently looking into other things before pushing.

Yunus1903:
Fixed in https://github.com/MinecraftForge/MinecraftForge/commit/680c4062f49c07eb1b53a39ade89df217f7b2223

PR #7252, opened 2020-08-14
1.16.2 Custom Tag Type Support

Cyborgmas:
Added a system for mods to add their own tag types. Not sure however if this is the best implementation possible, could be better to overtake the whole tag system or almost all of it.
What is most annoying in this is how to create the TagRegistry, not finding any better ways currently, but any ideas are welcome.
The test is very simple, I'll do some more complex ones a bit later.


Cyborgmas:
pupnewfster has been working on this, so closing this PR.

PR #7255, opened 2020-08-14
Events cleanup

### LINKED ISSUE #6394: EntityEvent.CanUpdate is never called
**Minecraft Version:** 1.14, 1.15

**Forge Version:** all 1.14 and 1.15 versions up to and including https://github.com/MinecraftForge/MinecraftForge/commit/e8ce61d4cd2de3250a27b1d4b8aa20d88e53c0d7

**Logs:** not applicable

**Steps to Reproduce:**
 1. Search for construction of `EntityEvent.CanUpdate` and see that it is only ever created in `ForgeEventFactory#canEntityUpdate`
2. Search for uses `ForgeEventFactory#canEntityUpdate` and see that no patch ever calls it

**Description of issue:**

It appears that this event was forgotten during the move to 1.13. To avoid confusion, it should be commented out until it is properly implemented, or if there are no plans to implement it, it should be removed.

sciwhiz12:
## Superseded by #7916.

_TL;DR: Events documentation are outdated, event class/methods reference dead things, and other old things in event code. **Breaking changes PR**, to prepare for future javadocs non-breaking PR._

Forge has many events added to its API, to hook into vanilla behaviour and gameplay. These events were implemented by different authors for their usecases, and expanded for general use by other modders. Over the years, as Minecraft and Forge changes their internal structures, these events are updated and modified, or in some cases outright removed if they serve little purpose.

However, there are some flaws in events and their javadocs currently; here are the most notable:
  1. As the mappings for Minecraft are updated, class and member name changes cause references in javadocs to break; especially, the [large class renames for 1.14][class_renames] breaking most javadocs from pre-1.14.
  2. Some, if not most events have insufficient documentation on their functions and usages.
  3. The method names across some of the events do not match: one event may have `#getRenderPartialTick()`, another has `#getPartialTicks()`, `#getPartialRenderTicks()`, etc.
  4. Some events do not encapsulate the fields, so field renaming (which is, in my opinion, an implementation and non-public facing operation) can break accessors.

As 1.16.2 was recently released (at the time of writing), we are currently in a phase where breaking changes are permitted. This PR remedies points #&zwj;3 and #&zwj;4 by doing a cleanup of most of the event code, which causes **a lot of breaking changes** related to events.

Points #&zwj;1 and #&zwj;2 will be fixed by a future PR of mine to create and update event javadocs. This PR does all of the breaking changes to the public-facing API of events, so a future PR may clean the formatting and internal implementation details of events and be merged outside of the breaking changes window.

_Note #&zwj;1: I have not changed/updated the documentation present on these classes, as my future PR will rewrite the javadocs anyway, so any changes I make will eventually be removed._

_Note #&zwj;2: The methods within the event factories and hooks have not been renamed, as I have dedicated this PR for breaking changes relating to the public-facing API. Those event factories and hooks are (to my understanding) not for public consumption, and can be safely changed/renamed outside of the breaking changes window. _

### Non-exhaustive list of changes

  - **Added package-info.java for defaulting nullness annotations**
    - Adding the non-nullness annotations (`mcp.MethodsReturnNonnullByDefault` and `javax.annotation.ParametersAreNonnullByDefault`) to the packages through a `package-info.java` file in each package allows the removal of `@Nonnull` annotations for individual fields/methods/parameters. This is for users of static analysis tools (or IDEs equipped with those tools) to safely spot event code that use `null`s incorrectly.
    - Future PRs can explictly add the nullness annotations (`javax.annotation.Nullable`) for fields/methods/parameters that accept and hold `null` values.

  - **Removed nonnullness annotations from fields/parameters/methods relating to `ItemStack`s**
    - Related to the previous point; `ItemStack`s should never be `null` (only empty or not), so this annotation is now unneeded.
  - **Moved render events to `net.minecraftforge.client.event.render` sub-package**
    - Although the render events can be identified with the `Render*` prefix in the classname, combining these classes into a dedicated sub-package allows easier finding of these events and organizes the root client-only event package.
  - **Moved events in `.furnace` and `.brewing` sub-packages to root event package**
    - The `FurnaceFuelBurnTimeEvent`, `PlayerBrewedPotionEvent`, and `PotionBrewEvent` are not significant enough to warrant having a sub-package in the root events package.
  - **Renamed methods to be the same across similar events**
    - Standardizes method names between events, so future events have a baseline of method naming, and modders may gain an intuition on relating the return type to the method name.
  - **Renamed events with class names that reference outdated MC names**
    - `GuiScreenEvent` to `ScreenEvent`, reference to `GuiScreen`
    - `GuiContainerEvent` to `ContainerScreenEvent`; reference to `GuiContainer`
    - `GuiOpenEvent` to `ScreenOpenEvent`; reference to `GuiScreen`
    - `InputUpdateEvent` to `MovementUpdateEvent`; may be confused with `InputEvent`
    - `FOVUpdateEvent` to `FOVModifierEvent`; changes the hand FOV modifier (fires after `FOVUpdate`)
    - `EntityRenderViewEvent.FOVModifier` to `.FOVUpdate`; updates the FOV value before all modifers are applied
  - **Removed unnecessary qualifications of event classes**
    - As these are Forge classes and not in patches, these full qualifications of classes are unneeded.
  - **Removed `EntityEvent.CanUpdate`**
    - Fixes #6394. There is no hook for this event, and there is no good usecase for this event: any potential use can be achieved by calling `Entity#canUpdate(boolean)` during a `TickEvent`. 
    - _**Please comment if there is a valid usecase for this event, so the event may be re-introduced, and reimplemented by another PR.**_
  - **Changed public fields to private and add getters**
    - By encapsulating the fields with getters, future events can change the field names without a breaking change.
  - **Removed unnecessary static modifiers to enums**
    - The [Java™ Language Specification, section 8.9][jls-8.9] states that: _"A nested enum type is implicitly `static`."_ Therefore, the `static` modifier for nested enums are unnecessary.
  - **Extended `PotionApplicableEvent` to apply to wither skeletons and withers**
    - As `PotionApplicableEvent` fires for the `SpiderEntity` and the poison effect (as spiders are normally immnue to poison), the event will now also fire for the `WitherEntity` and `WitherSkeletonEntity`.
  - **Renamed `EnderTeleportEvent` to `LivingTeleportEvent`**
    - Since the event is also fired for shulkers, which don't fit under the 'ender' theme (enderman, ender pearl).

...and more small fixes and renames. 

[class_renames]: https://github.com/ModCoderPack/MCPBot-Issues/issues/814
[jls-8.9]: https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.9


src/main/java/net/minecraftforge/client/event/ClientChatReceivedEvent.java
```diff
@@ -66,8 +65,7 @@ public ChatType getType()
      * The UUID of the player or entity that sent this message, or null if not known.
      * This will be equal to {@link net.minecraft.util.Util#field_240973_b_} for system messages.
      */
-    @Nullable
-    public UUID getSenderUUID()
+    public UUID getUUID()
```
diesieben07: What is the reason behind this name change? `getUUID` is fairly unspecific...

src/main/java/net/minecraftforge/client/event/ContainerScreenEvent.java
```diff
@@ -26,17 +26,17 @@
 /**
  * Event class for handling GuiContainer specific events.
```
diesieben07: Missing name change

src/main/java/net/minecraftforge/client/event/ContainerScreenEvent.java
```diff
@@ -47,7 +47,7 @@ public ContainerScreen getGuiContainer()
      * This is useful for any slot / item specific overlays.
      * Things that need to be on top of All GUI elements but bellow tooltips and dragged stacks.
```
diesieben07:
```suggestion
     * Things that need to be on top of All GUI elements but below tooltips and dragged stacks.
```

src/main/java/net/minecraftforge/client/event/ContainerScreenEvent.java
```diff
@@ -89,7 +89,7 @@ public int getMouseY()
      * This event is fired directly after the GuiContainer has draw any background elements,
```
diesieben07: Missing name change

src/main/java/net/minecraftforge/client/event/FOVModifierEvent.java
```diff
@@ -24,35 +24,35 @@
...
     }
 
-    public float getNewfov()
+    public float getFOV()
     {
         return newfov;
     }
 
-    public void setNewfov(float newfov)
+    public void setFOV(float newfov)
```
diesieben07: Might want to rename the parameter, too.

src/main/java/net/minecraftforge/client/event/FOVModifierEvent.java
```diff
@@ -24,35 +24,35 @@
 import net.minecraft.util.math.MathHelper;
 import net.minecraftforge.eventbus.api.Event;
 
-public class FOVUpdateEvent extends Event
+public class FOVModifierEvent extends Event
 {
     private final PlayerEntity entity;
     private final float fov;
     private float newfov;
```
diesieben07: Might want to rename this field, too.

src/main/java/net/minecraftforge/client/event/MovementUpdateEvent.java
```diff
@@ -27,11 +27,11 @@
  * This event is fired after player movement inputs are updated.<br>
  * Handlers can freely manipulate {@link MovementInput} to cancel movement.<br>
  */
-public class InputUpdateEvent extends PlayerEvent
+public class MovementUpdateEvent extends PlayerEvent
```
diesieben07: I think `MovementInputUpdateEvent` would be better, since `MovementUpdateEvent` could mean all kinds of things.

src/main/java/net/minecraftforge/client/event/ScreenEvent.java
```diff
@@ -203,7 +201,7 @@ public Post(Screen gui, MatrixStack mStack, int mouseX, int mouseY, float render
      * This event fires at the end of {@link GuiScreen#drawBackground(int)} and before the rest of the Gui draws.
```
diesieben07: Wrong link

src/main/java/net/minecraftforge/client/event/ScreenEvent.java
```diff
@@ -45,24 +43,24 @@
...
     private final Screen gui;
 
-    public GuiScreenEvent(Screen gui)
+    public ScreenEvent(Screen gui)
     {
         this.gui = gui;
     }
 
     /**
      * The GuiScreen object generating this event.
```
diesieben07: Reference to GuiScreen

src/main/java/net/minecraftforge/client/event/ScreenOpenEvent.java
```diff
@@ -19,9 +19,9 @@
 
 package net.minecraftforge.client.event;
 
+import net.minecraft.client.gui.screen.Screen;
 import net.minecraftforge.eventbus.api.Cancelable;
 import net.minecraftforge.eventbus.api.Event;
-import net.minecraft.client.gui.screen.Screen;
 
 /**
  * This event is called before any Gui will open.
```
diesieben07:
```suggestion
 * This event is called before any Screen will open.
```

src/main/java/net/minecraftforge/client/event/ScreenOpenEvent.java
```diff
@@ -19,9 +19,9 @@
 
 package net.minecraftforge.client.event;
 
+import net.minecraft.client.gui.screen.Screen;
 import net.minecraftforge.eventbus.api.Cancelable;
 import net.minecraftforge.eventbus.api.Event;
-import net.minecraft.client.gui.screen.Screen;
```
diesieben07: Why?

diesieben07 (changes requested):
I have to say, some of the method name changes seem rather arbitrary, to be honest. For example why does `PlayerSleepInBedEvent` need to have a method called `setSleepResult`? What's wrong with `setResult`? Of course it's going to be the sleeping result... Just as one example.

src/main/java/net/minecraftforge/client/event/ScreenOpenEvent.java
sciwhiz12: This is the result of letting IntelliJ organize imports. Will fix.

src/main/java/net/minecraftforge/client/event/ClientChatReceivedEvent.java
sciwhiz12: Yeah, on hindsight, the original method is better named. Will fix.

src/main/java/net/minecraftforge/client/event/ContainerScreenEvent.java
sciwhiz12: While I will fix this and the other review issues to ensure the javadocs are consistent with the class names, I will state that I deliberately did not touch the javadocs _(excepting those that were touched by removing qualifications and the like)_ because a future PR of mine will completely rewrite those javadocs, making it (at least, to me) a pointless endeavour. 

_(In fact, I have already rewritten the javadocs for this event, see the [event_docs branch on my fork][event_docs_branch])_

[event_docs_branch]: https://github.com/sciwhiz12/MinecraftForge/tree/1.16/event_docs

src/main/java/net/minecraftforge/client/event/ContainerScreenEvent.java
sciwhiz12: See my previous comment: https://github.com/MinecraftForge/MinecraftForge/pull/7255#discussion_r471100490.

src/main/java/net/minecraftforge/client/event/ContainerScreenEvent.java
sciwhiz12: See my previous comment: https://github.com/MinecraftForge/MinecraftForge/pull/7255#discussion_r471100490.

src/main/java/net/minecraftforge/client/event/ContainerScreenEvent.java
diesieben07: Oh, I thought you wanted to do it all in one go.

src/main/java/net/minecraftforge/client/event/FOVModifierEvent.java
sciwhiz12: My intention with making this PR is to do as much of the breaking changes needed, and only breaking changes as much as possible. A rename of a method parameter does not affect the method's signature and (to my understanding) does not break binary compatibility, and therefore can be done in a later PR outside of the breaking changes window. _(My intention is to have this PR to be merged as soon as possible within the breaking changes window, and another PR such as my events javadocs PR can rename the internal details, such as field and method parameter names)_
But, to be consistent with fixing all of the issues flagged by review, I will fix this.

src/main/java/net/minecraftforge/client/event/FOVModifierEvent.java
sciwhiz12: See my previous comment on field/parameter renaming: https://github.com/MinecraftForge/MinecraftForge/pull/7255#discussion_r471101124.

src/main/java/net/minecraftforge/client/event/MovementUpdateEvent.java
sciwhiz12: That does sound better, and more reflects the purpose of the event. Will do.

src/main/java/net/minecraftforge/client/event/ScreenEvent.java
sciwhiz12: See my previous comment: https://github.com/MinecraftForge/MinecraftForge/pull/7255#discussion_r471100490.

src/main/java/net/minecraftforge/client/event/ScreenEvent.java
sciwhiz12: See my previous comment: https://github.com/MinecraftForge/MinecraftForge/pull/7255#discussion_r471100490.

src/main/java/net/minecraftforge/client/event/ScreenOpenEvent.java
sciwhiz12: See my previous comment: https://github.com/MinecraftForge/MinecraftForge/pull/7255#discussion_r471100490.

sciwhiz12:
Although I do not see any other arbitrary method changes (please clarify, so I can see if I did (re)name something wrongly), I can comment on `PlayerSleepInBedEvent` and `setSleepResult`: the original was `getResultStatus` which is ambiguous on first glance _("status of a result? what?")_, and renaming it to `setResult` could cause confusion with `Event#setResult` (not in a binary compat. sense, but overloading a unrelated method could cause confusion).

diesieben07:
For example: getPlayer -> getPlayerEntity. getAmount -> getDamageAmount, etc.

sciwhiz12:
The `getPlayer` -> `getPlayerEntity` is to be more descriptive of the return type as `PlayerEntity`. 
`getAmount` -> `getDamageAmount` is to better describe what the amount returned represents, without having to check the javadoc constantly. _("getAmount? We're in... LivingAttackEvent, so maybe it's the damage amount? Lemme check... ah, it **is** the damage amount." vs "getDamageAmount? Ah, its the amount of damage." In my opinion, it's better to have self-explanatory method names, than having to explain everything in the javadocs)_

Alpvax:
I feel that the first is unnecessary. What else could getPlayer be, other than a PlayerEntity? Again with getAmount, what else could it be. I think it's great to rename to avoid confusion, but if there's no confusion I would rather stick with the less verbose name.

sciwhiz12:
The `getPlayerEntity` rename was partly to be more descriptive about it returning a `PlayerEntity` (similar to `Entity` for `getEntity`), and partly of my own preference to the rename. If you believe that conciseness would be better in this instance, then I shall revert the commit.
In my opinion, the `getDamageAmount` rename better fits in that it self-documents; rather than a generic `amount` with context, it can be understood as an `amount of damage`, even without the context of the event's name.

On a related point, I am second-guessing my decision to override `getEntity` in `LivingEvent` and `PlayerEvent` to return their respective entity types. (so, `LivingEvent#getEntity` now returns `LivingEntity`) Should I keep this rename as-is? Or would it be better to keep the old way of having a separate method that returns the explicit event's type? (`LivingEvent#getLivingEntity`)

ChampionAsh5357:
~~My opinion is that if you’re gonna override `EntityEvent::getEntity`, it should be generic so that you don’t have to cast to an already known entity extension. Otherwise, the old way should be followed with the specified signature. However, it’s up to them how they would want it done.~~

I forgot covariance in Java was a thing, my bad.

sciwhiz12:
Updated to 1.16.3. I do still hope that this PR will be merged before 1.16 gets a Recommended Build, so I can start work on creating /updating the events documentation. 
It has been ~1 month since this PR was made, so I hope most concerns are now okay, but please review the PR / message on Discord if there are any more issues with this.

sciwhiz12:
Just to inform anyone stumbling upon this PR, I was informed weeks ago (before the first RB for 1.16) that this PR will be considered for 1.17 instead, due to the amount of breaking changes that this PR introduces to events.

As such, this PR will lie dormant until 1.17 Forge is released. I will keep updating this PR to latest commits to avoid being labeled as Stale, but I will hold off on making new changes.

sciwhiz12:
This PR has been superseded by #7916 for 1.17. I chose to make a new PR for 1.17's changes instead of repurposing this PR in order to preserve the work I've already done, so I may port over the changes to the new PR bit-by-bit.

PR #7256, opened 2020-08-16
Improve startup time by caching the manifest data for mod jars

### LINKED ISSUE #55: Signed mod jars significantly slow down class loading
I recently noticed that the "mod construction" phase took quite long for Immersive Engineering while profiling startup for a modpack, even when compared to mods of a similar size. Most of that time is spent loading classes. With help from @sfPlayer1 I tracked this down to jar signing:  
ModLauncher parses the manifest every time it loads a class from a jar file ([here](https://github.com/cpw/modlauncher/blob/966c6986244239e51f22b91eb0acce6d7c37de7c/src/main/java/cpw/mods/modlauncher/TransformingClassLoader.java#L230-L243)). We seem to be one of the few teams that still uses signed jars in 1.16, so IE has a much larger manifest than most mods (roughly 735 kB compared to 25 bytes without signing). Storing the loaded manifest would significantly improve load times for signed mods and at least slightly for unsigned ones.

Startup times (time between the message "ModLauncher running[...]" and the last message of the form "[minecraft/AtlasTexture]: Created: [...]"):  
Without any mods: 42-43 s (2 launches)
With signed IE jar: 75-76 s or 32-34 additional seconds (3 launches)  
With unsigned IE jar: 55-58 s or 12-16 additional seconds (2 launches)

Forge version: 32.0.107  
ModLauncher version: 6.1.1+74+master.966c698

malte0811:
Fixes cpw/modlauncher#55

 - I originally thought this was a ModLauncher issue because I didn't look at the stack in detail after it entered ModLauncher. Looking at it again the time I thought was in ModLauncher is actually in the method I'm changing in this PR.
 - I'm using `AtomicReference` since a) it needs to be thread safe (the method is called in classloading), b) AFAIK accessing an atomic is faster than acquiring and releasing a lock and c) parsing the manifest on multiple threads will not cause any issues and can only happen once per mod.
 - I'll happily accept suggestions for getting rid of the `Optional<Optional<…>>`. IntelliJ puts out a warning when comparing an `Optional` to null and I agree that nullable `Optional`s are bad design, but I'm not sure this is better. This is in `fmllauncher`, so using `Lazy<T>` is not an option.

The table below shows launch time in seconds with patch applied to Forge 106. The setup is slightly different from the one I used in the ModLauncher issue, but is mostly similar. I'm using three launches for each configuration to get more reliable values. The Forge builds for both tests are unsigned (didn't set up signing for Forge). With the official 106 Forge build start time without mods is ~50 s, which would probably also be reduced by my changes.
|With PR|Mods|First launch|Second launch|Third launch|
|---|---|---|---|---|
|No|Immersive Engineering, signed|77|73|73|
|Yes|Immersive Engineering, signed|52|48|49|
|No|Immersive Engineering, unsigned|51|55|56|
|Yes|Immersive Engineering, unsigned|51|56|52|
|No|None|38|39|43|
|Yes|None|38|39|38|

### DIFF
--- a/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
+++ b/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
@@ -56,9 +56,8 @@ public class ModFileInfo implements IModFileInfo, IConfigurable
     private final Map<String,Object> properties;
     private final String license;
     // Caches the manifest of the mod jar as parsing the manifest can be expensive for
-    // signed jars. The manifest needs to be an Optional<Manifest>, so a nullable Optional
-    // is probably the best option.
-    private volatile Optional<Manifest> manifest = null;
+    // signed jars.
+    private final Optional<Manifest> manifest;
 
     ModFileInfo(final ModFile modFile, final IConfigurable config)
     {
@@ -88,6 +87,7 @@ public class ModFileInfo implements IModFileInfo, IConfigurable
                 this.modFile::getFileName,
                 () -> this.mods.stream().map(IModInfo::getModId).collect(Collectors.joining(",", "{", "}")),
                 () -> this.mods.stream().map(IModInfo::getVersion).map(Objects::toString).collect(Collectors.joining(",", "{", "}")));
+        this.manifest = modFile.getLocator().findManifest(modFile.getFilePath());
     }
 
     @Override
@@ -121,13 +121,7 @@ public class ModFileInfo implements IModFileInfo, IConfigurable
 
     public Optional<Manifest> getManifest()
     {
-        Optional<Manifest> result = manifest;
-        if (result == null)
-        {
-            result = modFile.getLocator().findManifest(modFile.getFilePath());
-            manifest = result;
-        }
-        return result;
+        return manifest;
     }
 
     @Override
--- a/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
+++ b/src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
@@ -77,7 +77,8 @@ public class ModFileInfo implements IModFileInfo, IConfigurable
         this.modFile.setFileProperties(this.properties);
         this.issueURL = config.<String>getConfigElement("issueTrackerURL").map(StringUtils::toURL).orElse(null);
         final List<? extends IConfigurable> modConfigs = config.getConfigList("mods");
-        if (modConfigs.isEmpty()) {
+        if (modConfigs.isEmpty())
+        {
             throw new InvalidModFileException("Missing mods list", this);
         }
         this.mods = modConfigs.stream()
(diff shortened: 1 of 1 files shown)


diesieben07 (changes requested):
My suggestions:

- Field should just be `volatile Optional<Manifest>` and then you can use double-checked locking. This should be the ideal performance here (`AtomicReference` isn't the fasted either).
- Yes, a "nullable optional" is not ideal, but imho the right choice here. Nesting the optionals is just ugly.

malte0811:
I forgot `volatile` existed, I haven't done much multithreading in Java. Do I actually need to use double-checked locking here or would just writing `if (x==null) x = …;` work as well once `x` is volatile (asking since you explicitly mentioned it)? As far as I can tell the locking is only necessary to prevent multiple threads from doing the same work, which isn't a problem here.

diesieben07:
Yeah I guess if you don't care about the work potentially happening multiple times then just `volatile` alone is fine.
You just need to ensure you only read the variable once. E.g. don't do `if (myField != null) { return myField; }`, since that is not guaranteed to be the same value then.

malte0811:
Can the value ever go "backwards" in time? In this case the value is only set once (or multiple times to equivalent values), so once I see a non-`null` value in the variable it should never go back to `null` again. So the code you wrote should technically be fine if the value never goes backwards. Minimizing accesses is probably still a good idea, since any access will need some amount of special treatment.

diesieben07:
Yeah, true, you might just in theory see two different `Manifest` values.

src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
```diff
@@ -115,12 +118,12 @@ public VersionRange getModLoaderVersion()
     }
 
     public Optional<Manifest> getManifest() {
-        Optional<Optional<Manifest>> manifest = this.manifest.get();
-        if (!manifest.isPresent()) {
-            manifest = Optional.of(modFile.getLocator().findManifest(modFile.getFilePath()));
-            this.manifest.set(manifest);
+        Optional<Manifest> result = manifest;
+        if (result == null) {
```
diesieben07: Braces on new lines (might want to fix the one on the method signature as well while you're at it).

diesieben07 (changes requested):
Other than the formatting change looks good now.

src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
```diff
@@ -54,6 +55,10 @@
     private final List<IModInfo> mods;
     private final Map<String,Object> properties;
     private final String license;
+    // Caches the manifest of the mod jar as parsing the manifest can be expensive for
+    // signed jars. The manifest needs to be an Optional<Manifest>, so a nullable Optional
+    // is probably the best option.
+    private volatile Optional<Manifest> manifest = null;
```
ichttt: If it doesn't matter that two threads hold different instances of the manifest (with the same content), this does not need to be volatile

diesieben07: Yes it does. As soon as more than one thread can access a variable you _must_ use volatile (or synchronization). If you just use a normal field, it would be perfectly okay for other threads than the writing thread to _never_ see this new value.

ichttt:
Is it also possible to look up the manifest in the ctor and caching that? That way the whole lazy init and volatile stuff could be avoided

src/fmllauncher/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
```diff
@@ -37,6 +37,7 @@
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
```
diesieben07: Pointless import, other than this looks good now.

diesieben07 (changes requested):

diesieben07 (approved):

PR #7257, opened 2020-08-16
Add DataProviders for WorldGen

Cyborgmas:
Added dataproviders for all of the possible registries, formatting is still incorrect for now, I'll be fixing it a bit later.

Codecs are the key feature here and I explain a lot of it in the javadocs. 

An issue I encountered is that most world gen objects will inline any other worldgen object that is not present in the registry and not reference them using a resourcelocation. To prevent that, RegistryKeyCodec must be used and encoded using WorldGenSettingsExport.
Furthemore, some providers depend on the output of other providers, it would be a pain to have to gen the first batch of files, put them in the mod resources, then gen the second batch (and then a third batch even!).
To help with that, each new provider will cause the ExistingFileHelper to reload the ResourceManager. Since it has no listeners it is pretty efficient and only serves to discover the newly created files. This is then used by the WorldSettingsImport$IResourceAccess (its anonymous class, which bases itself off of a ResourceManager).

There was also some issues with how enums are decoded because of IEnumExtensible, added some debug information for that.

Questions welcome and if some extra helper could be added, do tell.


Cyborgmas:
Okay, so made it much easier now to reference other objects through RegistryOpsHelper, including making it work with vanilla objects!
However, the downside is that I register things during datagen to a vanilla registry... Which is probably not ideal, but since it is purely in dev, it should be okay for now?

Have not tested all of the providers I created, I'll get to that soon.

Redstoneguy129 (approved):

Cyborgmas:
This PR contains some minor breaking changes, only related to internal stuff however. (DataModLoader and GatherDataEvent).
However, the main thing is that it allows for properly encoding object references instead of the whole objects, something that is needed to make datagen for #7287

Cyborgmas:
The latest commit also removed the breaking in change in GatherDataEvent

patches/minecraft/net/minecraft/entity/EntityClassification.java.patch
```diff
@@ -21,7 +21,7 @@
 @@ -57,6 +57,16 @@
        return this.field_82707_i;
     }
- 
++
 +   public static EntityClassification create(String name, String id, int maxNumberOfCreatureIn, boolean isPeacefulCreatureIn, boolean isAnimalIn, int despawnDistance) {
 +      throw new IllegalStateException("Enum not extended");
 +   }
```
ChampionAsh5357, line 27: Is this a misapplied patch since it seems to be unrelated to this PR?

src/main/java/net/minecraftforge/common/data/worldgen/CodecBackedProvider.java
```diff
@@ -0,0 +1,66 @@
...
+public abstract class CodecBackedProvider<P> implements IDataProvider
+{
+    protected final static Logger LOGGER = LogManager.getLogger();
+    protected final RegistryOpsHelper regOps;
+    protected final Codec<P> codec;
+    protected final Gson gson;
+
+    protected CodecBackedProvider(Codec<P> codec, RegistryOpsHelper regOps)
+    {
+        this(codec, regOps, new GsonBuilder().setPrettyPrinting().registerTypeAdapter(ResourceLocation.class, new ResourceLocation.Serializer()).create());
```
ChampionAsh5357, line 46: Why create a new GSON instance for every provider?

src/main/java/net/minecraftforge/common/data/worldgen/RegistryBackedProvider.java
```diff
@@ -0,0 +1,54 @@
...
+    protected RegistryBackedProvider(Codec<P> codec, RegistryOpsHelper regOps, RegistryKey<? extends Registry<P>> regKey)
+    {
+        super(codec, regOps);
+        this.regKey = regKey;
+    }
+
+    protected void saveAndRegister(P instance, ResourceLocation name, DirectoryCache cache, Path path) throws IOException
+    {
+        if(instance instanceof IForgeRegistryEntry) //necessary for biomes which encode the registry name.
+            ((IForgeRegistryEntry<?>) instance).setRegistryName(name);
```
ChampionAsh5357, line 50: I don't believe this is necessary as the codec field itself is optional. Adding this field wouldn't do anything anyways as it seems Forge replaces it with the current registry name regardless.

src/main/java/net/minecraftforge/common/data/worldgen/BiomeDataProvider.java
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * The {@link BiomeMaker} class also provides helper functions for creating specific biomes.
+ *
+ * To use newly created objects from other DataProviders, add this provider after the others.
+ * See {@link RegistryOpsHelper#getObject} for usage.
+ *
+ * See <a href=https://minecraft.gamepedia.com/Custom_world_generation#Biome>the wiki</a> for more details
+ * on biome parameters.
+ */
+public abstract class BiomeDataProvider extends RegistryBackedProvider<Biome>
```
ChampionAsh5357, line 47: Most of this class could be abstracted into `RegistryBackedProvider`.

- `#map` and `#generator` can be stored within the registry backed provider
- `#act` and `#put` could also be implemented back there as the location is stored within the registry key path
- `#modid` is only used for `#getName`, it probably doesn't need to be a field here.
- The entire class needs to be javadoced more clearly since you have to observe the code to be able to figure out how to get it working.

src/main/java/net/minecraftforge/common/data/worldgen/ConfiguredCarverProvider.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+public abstract class ConfiguredCarverProvider extends RegistryBackedProvider<ConfiguredCarver<?>>
+{
+    private final DataGenerator generator;
+    private final String modid;
+    protected final Map<ResourceLocation, ConfiguredCarver<?>> builders = new HashMap<>();
+
+    public ConfiguredCarverProvider(DataGenerator generator, RegistryOpsHelper helper, String modid)
+    {
+        //TODO This codec is dispatched for the vanilla CARVER registry, and won't affect any mod added carvers.
```
ChampionAsh5357, line 45: Is this comment out of date along with the rest? `RegistryOpsHelper` registers to the vanilla registry along with any other mod carvers. This event executes after `FMLCommonSetupEvent` which by then all mods have registered to the builtin registries. The only one this isn't true for is the `#DIMENSION_TYPE_REGISTRY` iirc, but those seem to be added from the `RegistryOpsHelper`.

ChampionAsh5357 (changes requested):
So I have one major concern that's not really specified. This forces an order of registration meaning that Carvers must be registered before Biomes and so on. This order is not made clear for any consumer trying to use these events.

In other news, please update this PR to 1.17 as well.

sciwhiz12:
Hello! @Cyborgmas, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

Cyborgmas:
Will try to update, but not sure. Will close for now, and open a new PR in the case I do update

PR #7261, opened 2020-08-20
Movement Handles for Custom Scaffolding

ChampionAsh5357:
Ladders are hooked through mc using the tag "climbable" and implemented for more specific uses using `IForgeBlock::isLadder`. This handles fine for scaffolding in most cases as climbing a scaffolding is pretty much the same as a ladder. The only difference in how a user climbs down. For ladders, it is assumed that the player would just stop holding a button. However, scaffolding requires the player to shift, which is not possible for a standard ladder. The value is also hardcoded for `Blocks::SCAFFOLDING` specifically.

This PR focuses on adding this movement handle in by adding another method into `IForgeBlock`. There is also code to add in a forge tag as part of the base to allow users the same flexibility as the ladder for a simple or more specific implementation.


src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -911,4 +911,18 @@ default BlockState getToolModifiedState(BlockState state, World world, BlockPos
...
+    /**
+     * Checks if a player or entity handles movement on this block like scaffolding.
+     *
+     * @param state The current state
+     * @param world The current world
+     * @param pos The block position in world
+     * @param entity The entity on the scaffolding
+     * @return True if the block should act like scaffolding
+     */
+    default boolean isScaffolding(BlockState state, IWorldReader world, BlockPos pos, LivingEntity entity)
```
ichttt, line 924: Why pass all these additional paramteres in when they can be (and are in fact) accessed from `LivingEntity`?

ichttt (commented):
In addition to my other comment, can you provide a data generator for the test json files?

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
ChampionAsh5357: I did this mainly to mimic `IForgeBlock::isLadder`. It's implementation is pretty much similar where everything could be passed in through the `LivingEntity`. I figured since the idea is to mimic the ladder usage with scaffolding (as the ladder method determines if the thing is climable), it would be reasonable to keep the same parameters. Of course, if this is a bad judgement, I shall adjust.

build.gradle
```diff
@@ -364,6 +364,7 @@ project(':forge') {
                 args '--flat', '--all', '--validate',
                     '--mod', 'data_gen_test', 
                     '--mod', 'piston_event_test', 
+					'--mod', 'scaffolding_test',
```
covers1624: Indent is wonky

ChampionAsh5357: Accidentally tabbed instead of spaced, fixed now.

src/test/resources/data/forge/tags/blocks/scaffolding.json
```diff
@@ -0,0 +1,6 @@
+{
```
ichttt: This file should no longer be needed, now that you have the generated version

src/test/resources/assets/scaffolding_test/blockstates/scaffolding_method_test.json
```diff
@@ -0,0 +1,10 @@
+{
```
ichttt: Can you also provide a blockstate data gen?

ChampionAsh5357:
ichttt, should I still remove the parameters from the block or keep it the same for consistency purposes?

src/test/java/net/minecraftforge/debug/block/ScaffoldingTest.java
```diff
@@ -0,0 +1,99 @@
...
+    {
+        IEventBus modBus = FMLJavaModLoadingContext.get().getModEventBus();
+        BLOCKS.register(modBus);
+        modBus.addListener(this::gatherData);
+    }
+
+    private void gatherData(final GatherDataEvent event)
+    {
+        DataGenerator gen = event.getGenerator();
+    	if(event.includeClient()) gen.addProvider(new ScaffoldingBlockState(gen, MODID, event.getExistingFileHelper()));
```
ichttt: wrong indentation

src/test/java/net/minecraftforge/debug/block/ScaffoldingTest.java
```diff
@@ -0,0 +1,99 @@
...
+    	if(event.includeServer()) gen.addProvider(new ScaffoldingTagsProvider(gen));
+    }
+
+    static class ScaffoldingTagsProvider extends BlockTagsProvider
+    {
+        public ScaffoldingTagsProvider(DataGenerator generatorIn)
+        {
+		    super(generatorIn);
```
ichttt: tabs

ichttt (approved):

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -911,4 +911,18 @@ default BlockState getToolModifiedState(BlockState state, World world, BlockPos
...
+    {
+        return state.isIn(Blocks.SCAFFOLDING) || state.getBlock().isIn(Tags.Blocks.SCAFFOLDING);
```
LexManos: We **DO NOT** add tags for game logic.

ChampionAsh5357: Apologies, Lex. Removed the in-game tag logic directly and moved it to the specific test.

src/test/java/net/minecraftforge/debug/block/ScaffoldingTest.java
```diff
@@ -35,8 +36,9 @@
 {
     static final String MODID = "scaffolding_test";
     static final DeferredRegister<Block> BLOCKS = DeferredRegister.create(ForgeRegistries.BLOCKS, MODID);
+    static final IOptionalNamedTag<Block> SCAFFOLDING = BlockTags.createOptional(new ResourceLocation("forge", "scaffolding"));
 
-    static final RegistryObject<Block> SCAFFOLDING_TAG_TEST = BLOCKS.register("scaffolding_tag_test", () -> new ScaffoldingBlock(AbstractBlock.Properties.create(Material.MISCELLANEOUS, MaterialColor.SAND).doesNotBlockMovement().sound(SoundType.SCAFFOLDING).variableOpacity()));
+    static final RegistryObject<Block> SCAFFOLDING_TAG_TEST = BLOCKS.register("scaffolding_tag_test", () -> new ScaffoldingTagTestBlock(AbstractBlock.Properties.create(Material.MISCELLANEOUS, MaterialColor.SAND).doesNotBlockMovement().sound(SoundType.SCAFFOLDING).variableOpacity()));
```
LexManos: Remove all of this entirely. We do not support or encourage people using custom tags for game logic. So there is no need for this 'test'

ChampionAsh5357: Ok, should be done now.

PR #7262, opened 2020-08-21
Remove ForgeHooks.canToolHarvestBlock

### LINKED ISSUE #7258: ForgeHooks.canToolHarvestBlock returns incorrectly for multiple blocks
**Minecraft Version:** 1.16.1

**Forge Version:** 1.16.1-32.0.106

**Steps to Reproduce:**
 1. Create code that uses ForgeHooks.canToolHarvestBlock
2. Break Nether Quartz or Oak Log with any pickaxe/axe or fist respectively
3. canToolHarvestBlock will always return false.

**Description of issue:**
canToolHarvestBlock returns incorrectly for multiple blocks. I was only able to test Nether Quartz with a fist and all tiers of pickaxe. canToolHarvestBlock returned false for each pickaxe, yet the item and XP still dropped in game. The same occured with all tiers of an axe and a fist against Oak Logs.

**Test code:**
`@SubscribeEvent
    public static void breakBlock(BlockEvent.BreakEvent e) {
System.out.println("ForgeHooks.canToolHarvestBlock returns " + ForgeHooks.canToolHarvestBlock(e.getWorld().getWorld(), e.getPos(), e.getPlayer().getHeldItemMainhand()) + " for the block broken.");
}`

sciwhiz12:
_This PR fixes #7258._

_From my comment on the linked issue:_
From my investigation, `ForgeHooks.canToolHarvestBlock` seems to be code leftover from a time when tool classes were `Strings` instead of `ToolType`. There is no code calling this method, and it seems that this method is the only reason why the `IForgeItemStack#getHarvestLevel` has the `@Nullable` annotation in its parameters.
See commit 263e9193840a288d68667b83263a2229593123b3 for the original commit that added the method.

This PR does the folllowing:
  * Removes `ForgeHooks.canToolHarvestBlock`;
  * Removes the `@Nullable` annotations from `Item#getHarvestLevel` and `IForgeItemStack#getHarvestLevel`;
  * Changes the javadoc of `IForgeItemStack#getHarvestLevel`;
  * Changes `ForgeHooks.canHarvestBlock` from calling `IForgeItem#getHarvestLevel` to call `IForgeItemStack#getHarvestLevel`.


LexManos:
The intention of the nullables were that modders could call them as general informative handlers. And get the result generically for any player/block. Useful for tooltips/machines/whatever.

PR #7263, opened 2020-08-21
Don't process packets sent by dead players

LXGaming:
The Minecraft server doesn't check the players health when processing packets, A modified client can exploit this to perform certain actions while the player is dead which includes:
- Moving (Includes loading chunks)
- Block Interaction (Includes breaking, placing and inventory transactions with chests)
- Item Interaction (Includes damaging entities)
- Using Portals
- Using Commands

While the player is dead they are invisible to all other entities including players, this also prevents them from picking up Item or XP Orbs.

I reported this issue to Mojang (MC-164255) back in October 2019 but it was closed as a duplicate of MC-87170 (Because the issue ids are sequential I was able to find that this has been known since August 2015), I then submitted a PR to Sponge to fix the issue https://github.com/SpongePowered/SpongeCommon/pull/2401

The fix I've implemented for Forge is pretty much the same as Sponge, I added logic to check if the player is alive in `PacketThreadUtil` as all the packets (KeepAlive Packet being an exception) at this point are being processed on the Main Thread, The fix allows `CCustomPayloadPacket` for mods and the `CClientStatusPacket` for respawning, Everything else is ignored as when the player dies the only options they have are to respawn or rage quit.

It should also be noted that this issue has been fixed in Bukkit / Spigot for years now and was fixed in Sponge last year.

### DIFF
--- a/patches/minecraft/net/minecraft/network/PacketThreadUtil.java.patch
+++ b/patches/minecraft/net/minecraft/network/PacketThreadUtil.java.patch
@@ -1,0 +1,40 @@
+--- a/net/minecraft/network/PacketThreadUtil.java
++++ b/net/minecraft/network/PacketThreadUtil.java
+@@ -16,6 +16,7 @@
+       if (!p_218797_2_.func_213162_bc()) {
+          p_218797_2_.execute(() -> {
+             if (p_218797_1_.func_147298_b().func_150724_d()) {
++               if (shouldProcess(p_218797_0_, p_218797_1_)) // Fix: MC-164255 (Duplicate of private)
+                p_218797_0_.func_148833_a(p_218797_1_);
+             } else {
+                field_225384_a.debug("Ignoring packet due to disconnection: " + p_218797_0_);
+@@ -25,4 +26,29 @@
+          throw ThreadQuickExitException.field_179886_a;
+       }
+    }
++
++   public static <T extends INetHandler> boolean shouldProcess(IPacket<T> packetIn, T processor) {
++      if (processor instanceof net.minecraft.network.play.ServerPlayNetHandler) {
++         net.minecraft.entity.player.ServerPlayerEntity player = ((net.minecraft.network.play.ServerPlayNetHandler) processor).field_147369_b;
++         if (!player.func_70089_S()) {
++            // Allow CustomPayloads as mods communicate here
++            if (packetIn instanceof net.minecraft.network.play.client.CCustomPayloadPacket) {
++               packetIn.func_148833_a(processor);
++               return true;
++            }
++
++            // Allow the player to respawn
++            if (packetIn instanceof net.minecraft.network.play.client.CClientStatusPacket && ((net.minecraft.network.play.client.CClientStatusPacket) packetIn).func_149435_c() == net.minecraft.network.play.client.CClientStatusPacket.State.PERFORM_RESPAWN) {
++               packetIn.func_148833_a(processor);
++               return true;
++            }
++
++            // Deny anything else
++            return false;
++         }
++      }
++
++      // Anything else is valid
++      return true;
++   }
+ }
--- a/patches/minecraft/net/minecraft/network/PacketThreadUtil.java.patch
+++ b/patches/minecraft/net/minecraft/network/PacketThreadUtil.java.patch
@@ -8,25 +8,23 @@
                 p_218797_0_.func_148833_a(p_218797_1_);
              } else {
                 field_225384_a.debug("Ignoring packet due to disconnection: " + p_218797_0_);
-@@ -25,4 +26,29 @@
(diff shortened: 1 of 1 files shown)


patches/minecraft/net/minecraft/network/PacketThreadUtil.java.patch
```diff
@@ -0,0 +1,40 @@
...
+       }
+    }
++
++   public static <T extends INetHandler> boolean shouldProcess(IPacket<T> packetIn, T processor) {
++      if (processor instanceof net.minecraft.network.play.ServerPlayNetHandler) {
++         net.minecraft.entity.player.ServerPlayerEntity player = ((net.minecraft.network.play.ServerPlayNetHandler) processor).field_147369_b;
++         if (!player.func_70089_S()) {
++            // Allow CustomPayloads as mods communicate here
++            if (packetIn instanceof net.minecraft.network.play.client.CCustomPayloadPacket) {
++               packetIn.func_148833_a(processor);
```
ichttt: This is being called twice when true is being returened

patches/minecraft/net/minecraft/network/PacketThreadUtil.java.patch
```diff
@@ -0,0 +1,40 @@
...
++               return true;
++            }
++
++            // Allow the player to respawn
++            if (packetIn instanceof net.minecraft.network.play.client.CClientStatusPacket && ((net.minecraft.network.play.client.CClientStatusPacket) packetIn).func_149435_c() == net.minecraft.network.play.client.CClientStatusPacket.State.PERFORM_RESPAWN) {
++               packetIn.func_148833_a(processor);
```
ichttt: Same here, returning true already causes `func_148833_a` (`processPacket`) to be called

patches/minecraft/net/minecraft/network/PacketThreadUtil.java.patch
```diff
@@ -0,0 +1,40 @@
...
++               if (shouldProcess(p_218797_0_, p_218797_1_)) // Fix: MC-164255 (Duplicate of private)
+                p_218797_0_.func_148833_a(p_218797_1_);
+             } else {
+                field_225384_a.debug("Ignoring packet due to disconnection: " + p_218797_0_);
+@@ -25,4 +26,29 @@
+          throw ThreadQuickExitException.field_179886_a;
+       }
+    }
++
++   public static <T extends INetHandler> boolean shouldProcess(IPacket<T> packetIn, T processor) {
```
ichttt: Why is this public?

XFactHD: Wouldn't it be a good idea to move this into a forge class like NetworkHooks to make the patch smaller?

covers1624: No, it's fine here as its all additions, and moving it elsewhere would mean it's, required to be public, and its clearly not meant to be.

patches/minecraft/net/minecraft/network/PacketThreadUtil.java.patch
LXGaming: Fixed, I missed removing those lines when I moved the logic to the separate method.

covers1624 (approved):

patches/minecraft/net/minecraft/network/PacketThreadUtil.java.patch
```diff
@@ -0,0 +1,38 @@
...
+                p_218797_0_.func_148833_a(p_218797_1_);
+             } else {
+                field_225384_a.debug("Ignoring packet due to disconnection: " + p_218797_0_);
+@@ -25,4 +26,27 @@
+          throw ThreadQuickExitException.field_179886_a;
+       }
+    }
++
++   private static <T extends INetHandler> boolean shouldProcess(IPacket<T> packetIn, T processor) {
++      if (processor instanceof net.minecraft.network.play.ServerPlayNetHandler) {
```
Vinetos, line 17: Using an early return could increase the lisibility by decreasing bloc indentation by one

Vinetos (approved):

ichttt:
After disussion in the forge team, we came to the conclusion that this is not in forge's scope. This could have many implications in existing code, and there are many places (in mods) where valid packets at invalid times can cause harm.

LexManos:
More specifically, I am worried about a lot of the side effects of this 'catch all' type approach.
There are plenty of mods out there that allow, say, chatting while dead. Or walking around as a ghost.
On top of this, it's most definitely a thing that Mojang themselves should address. So keep campaigning on that end. We fix *some* small bugs in vanilla. Things that don't have many side effects. This one just has to many unknown side effects for me to consider pulling it.

PR #7264, opened 2020-08-21
Display a more friendly error message for mods that cannot be loaded due to missing license 

ichttt:
Currently, when a mod is missing a license, forge just crashes. This now provides a nicer error, so the user at least knows what files causes the error without reading a log file.
I'm not sure if the error message is good enough, if someone has a better string please tell me.
Screenshot of the new screen:
javaw_OxLdSd4aeU


Mysterious-Dev (approved):

pupnewfster:
Superseded by #7350 which has now been merged

PR #1758, opened 2015-03-23
Added a Food/Hunger API

Parker8283:
This adds various events and utilities to help with the manipulation of Food and the Hunger system.

This code has been lifted from AppleCore, and documentation on all this can be found here: https://github.com/squeek502/AppleCore/wiki/Using-the-AppleCore-API under the first two bullet points in the Events section, and the Accessor section.

Also, tagging @squeek502, as this is his original code. I just took it out of his mod and lifted it into forge for all to have! (with permission of course :P)

PR #7266, opened 2020-08-21
[1.16] Added a Hunger System API

Parker8283:
**This is an event-driven API that allows for control of the hunger system in the game. Things like food, health regen, hunger regen, exhaustion, starvation, etc. This was actually a PR I had made 5 years ago (#1758), but broke and never fixed. Until now!**

The goal of this system is to allow modders to change values/behaviors that were previously hard-coded in the game. Some examples include:

* Changing the hunger and saturation values of foods eaten
* Allow foods to have different hunger and saturation values per-player
* Change the amount of exhaustion added for each exhausting action the player does
* Control how and when health and hunger regeneration occurs
* Control how and when starvation occurs
* and more!

There is a lot that is allowed by this system. A good overview of everything would be to check the test mods that are included with this PR.

Other more technical things that this PR does include:

* `FoodStats` now takes a `PlayerEntity` upon construction that is used throughout the class when querying values. This allows `FoodStats` to operate more player-dependently than before, allowing for operations to occur based on what player is trying to perform them.
* The `foodTimer` in `FoodStats` has been separated into the `foodTimer` and `starveTimer`. This allows starvation to happen independently from the exhaustion/regeneration timer.
* A `FoodValues` class is provided to allow for easy use of hunger and saturation values for food objects.
* An `IEdible` interface is provided that can be applied to non-standard food objects so that they can be included in the system. For a vanilla example, Cake now implements this interface. It is still recommended and preferred to include a `Food` object when constructing your item over implementing this interface.
  * The original author of this system, @squeek502, notes that he was never completely satisfied with this approach of including non-standard food items and wonders if there is a better solution. If you have an idea, please let me know!
* Fire events while in peaceful mode to allow modders to control how health/hunger regen happens in peaceful
* Removes an `OnlyIn` annotation over a simple setter for the saturation level
  * ~~Was unsure if I should leave the imports for that or not. I seem to remember an exception for this one specific case. Or maybe I'm crazy.~~ EDIT: Found the new-ish (at least to me) SAS system. Kinda cool. This comment is no longer relevant.

---

This system would be used by mods such as [Hunger Overhaul](https://www.curseforge.com/minecraft/mc-mods/hunger-overhaul) and [The Spice Of Life](https://www.curseforge.com/minecraft/mc-mods/the-spice-of-life) to fundamentally change how the hunger system works in the game.

This system was originally written by @squeek502 for [AppleCore](https://www.curseforge.com/minecraft/mc-mods/applecore), and was first attempted to be upstreamed by me way back in the 1.7.10/1.8 days (as noted above). He has offered his code into the [Public Domain](https://github.com/squeek502/AppleCore/blob/1.12.2/LICENSE), and as such I am re-attempting to put this system into Forge! Since this code was originally written for a different Minecraft version, there is a good chance that there are better ways to do some of what we are trying to do. If that is the case, please let me know and I will gladly learn the new systems and how best to be doing what we are attempting. (I haven't really been actively modding Minecraft since the 1.8 days, haha)

### DIFF
--- a/patches/minecraft/net/minecraft/block/Block.java.patch
+++ b/patches/minecraft/net/minecraft/block/Block.java.patch
@@ -48,6 +48,15 @@
     public float func_149638_a() {
        return this.field_235689_au_;
     }
+@@ -300,7 +_,7 @@
+ 
+    public void func_180657_a(World p_180657_1_, PlayerEntity p_180657_2_, BlockPos p_180657_3_, BlockState p_180657_4_, @Nullable TileEntity p_180657_5_, ItemStack p_180657_6_) {
+       p_180657_2_.func_71029_a(Stats.field_188065_ae.func_199076_b(this));
+-      p_180657_2_.func_71020_j(0.005F);
++      p_180657_2_.func_71020_j(net.minecraftforge.event.ForgeEventFactory.onExhaustionAdded(p_180657_2_, net.minecraftforge.event.hunger.ExhaustionEvent.ExhaustingActions.HARVEST_BLOCK, 0.005F));
+       func_220054_a(p_180657_4_, p_180657_1_, p_180657_3_, p_180657_5_, p_180657_2_, p_180657_6_);
+    }
+ 
 @@ -332,6 +_,7 @@
        p_176216_2_.func_213317_d(p_176216_2_.func_213322_ci().func_216372_d(1.0D, 0.0D, 1.0D));
     }
--- a/patches/minecraft/net/minecraft/block/CakeBlock.java.patch
+++ b/patches/minecraft/net/minecraft/block/CakeBlock.java.patch
@@ -1,0 +1,40 @@
+--- a/net/minecraft/block/CakeBlock.java
++++ b/net/minecraft/block/CakeBlock.java
+@@ -19,7 +_,7 @@
+ import net.minecraft.world.IWorldReader;
+ import net.minecraft.world.World;
+ 
+-public class CakeBlock extends Block {
++public class CakeBlock extends Block implements net.minecraftforge.common.hunger.IEdible {
+    public static final IntegerProperty field_176589_a = BlockStateProperties.field_208173_Z;
+    protected static final VoxelShape[] field_196402_b = new VoxelShape[]{Block.func_208617_a(1.0D, 0.0D, 1.0D, 15.0D, 8.0D, 15.0D), Block.func_208617_a(3.0D, 0.0D, 1.0D, 15.0D, 8.0D, 15.0D), Block.func_208617_a(5.0D, 0.0D, 1.0D, 15.0D, 8.0D, 15.0D), Block.func_208617_a(7.0D, 0.0D, 1.0D, 15.0D, 8.0D, 15.0D), Block.func_208617_a(9.0D, 0.0D, 1.0D, 15.0D, 8.0D, 15.0D), Block.func_208617_a(11.0D, 0.0D, 1.0D, 15.0D, 8.0D, 15.0D), Block.func_208617_a(13.0D, 0.0D, 1.0D, 15.0D, 8.0D, 15.0D)};
+ 
+@@ -48,11 +_,11 @@
+    }
+ 
+    private ActionResultType func_226911_a_(IWorld p_226911_1_, BlockPos p_226911_2_, BlockState p_226911_3_, PlayerEntity p_226911_4_) {
+-      if (!p_226911_4_.func_71043_e(false)) {
++      if (!p_226911_4_.func_71043_e(this.canAlwaysEat())) {
+          return ActionResultType.PASS;
+       } else {
+          p_226911_4_.func_195066_a(Stats.field_188076_J);
+-         p_226911_4_.func_71024_bL().func_75122_a(2, 0.1F);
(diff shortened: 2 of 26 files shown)


Parker8283:
Looks like it can't be built because of missing license headers? At least that's what I see locally. I didn't touch any of the files it's reporting.

Insane96:
So for usage [AppleCore documentation](https://github.com/squeek502/AppleCore/wiki) can be used?

Parker8283:
For the most part, yes. I can migrate a lot of it to the Forge Documentation if this gets merged so that it's in more of an official place if people would like.

Parker8283:
Rebased and found SAS system, so used that instead of manually removing an OnlyIn. Everything still appears to work, at least from my testing.

Parker8283:
Rebased onto current HEAD.

Parker8283:
Rebased onto current HEAD. If there's anything else missing/needs fixing, please let me know. I would love to get this merged soon so that we can update our mods properly.

EEEureka (approved):

GeorgeOrwell1:
Why isn't this getting merged?

Parker8283:
Forge is made up of a team of volunteers for the most part. There's about 30 gajillion other issues and PRs open on this project and all associated projects. And with 1.17 just dropping, I'm sure they're stretched pretty thin right now. Patience my friend :)

marchermans:
@Parker8283 I ran initial tests on this, and it seems to work.

But I am wondering how much thought you put into backwards compatibility with the existing systems.
What if a mod is depending on the existing framework of how food works?

Can you put some more time into documenting this all a bit better, since right now it is a bit of a mess and hard for me to understand where the old system ends and the new one begins.

Parker8283:
Hi Orion, thanks for looking this over!

For API/Binary compatibility, I believe I have not made backwards incompatible changes, but let me know if I missed any.

For the overall system, it is designed in a way that modders should not have to change _anything_ in order to keep the same functionality that they already have in their mods. I think perhaps instead of thinking of it like an old vs new system thing, it's likely better to see this as an augmentation of _the_ system; the only real changes posed by this system are firing events during hunger-related events that already occurred in code, but were untouchable by Forge modders due to its hard-coded nature. Therefore, the system as it stands today hasn't changed, and to modders, unless they look for it, they won't tell a difference in the end result. They can continue doing what they already have been, and it will work as expected.

The only two cases I can think of where modders _might_ need to put in some extra work to be fully functional with this system are these two:
* The modder creates a non-standard "food item" (as in, doesn't add a `Food` object to their Item during building/registration, and instead implements adding hunger/saturation in their own way). Even in this case, their item would continue to work as expected with this system, it might just not fire _all_ of the events it's supposed to. On top of that, it won't be picked up by the `FoodValues` class as a "food item" because it doesn't use the `Food` object on the item, or implement the `IEdible` class introduced by this PR. But as stated, even with this, it would still work, just not use the new system completely like the mod author might want it to. The simple fix for this would be to implement the `IEdible` interface and make sure they're calling the proper method for food consumption (`FoodStats.eat(Item, ItemStack)` instead of `FoodStats.eat(int, float)`). This is exactly what the patch to `CakeBlock` demonstrates.
* The modder has hard-coded values of the expected vanilla behavior. For example, they always expect an Apple to add max 4 hunger to the player, or health regen to always occur when the player's hunger is >=18. Instead of querying the proper places to get those values, they just make those assumptions in whatever parts of their code pertain to that. This one is a little hazy for me even writing it out now, but I guess since the current hunger system is hard-coded to always operate this way, a modder might have just made these assumptions in their code. I'm not sure how realistic this case is, but it's just one I thought of while thinking of ways this PR might cause incompatibilities. Fix here would be to listen to the proper events or call the proper methods that they probably should've been doing in the first place :)

Overall, this system just augments the current system to fire various events at different points of the hunger system lifecycle, allowing modders to easily hook-in and change the default behavior if they so desire. But it should not require any changes from the modders themselves to keep their mods working as they always did. Perhaps this would be a good point to add a test for to prove? But I'm not entirely sure how to implement that, so I'm all ears if that is desired. Otherwise, if the code is confusing, please let me know which parts and I can help to decipher it. It's been a minute since I really dug into it, so I might even need a refresher, haha.

sciwhiz12:
Hello! @Parker8283, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

Parker8283:
I plan on making a 1.18 version. It's on the TODO list for me. Might not get to it until the new year. If anyone has feedback on the content of this PR in the meantime, that would help when I do make the 1.18 version of this PR. Otherwise, I'll port this to 1.18 soon and make a new PR and leave this one open as the LTS version if/when that 1.18 version is merged. Thanks for checking in!

Parker8283:
1.18 PR made (#8405). This has become the LTS version of the PR, if the 1.18 version is pulled. If/when that happens, I will update this PR with the changes needed. Until then, **do not merge this PR**. :)

MWall541:
If you do not want it merged until then, consider **drafting** this PR so the triage team knows not to interact with it for now.

Parker8283:
Ah, they made that option as small and hidden as possible. I did originally look for that, but couldn't find it. Thanks.

PR #7267, opened 2020-08-21
Client Biped/Horse Armor Models

ChampionAsh5357:
Currently, biped armor models are attached through a specific item using `IForgeItem::getArmorModel` and `IForgeItem::getArmorTexture`. The problem with this is that `getArmorModel` has a return type that is not physically available on the dedicated server. This is countered by the usage of `OnlyIn`; however, it should still be separated for this case. For `IForgeItem::getArmorTexture`, it does not have the same problem since it returns a `ResourceLocation`. However, since it's only meant to be called on the client, it should be moved for the same reasons.

The goal is to move these two methods into an `Item` -> `IModelProvider` map where `IModelProvider` holds two methods: `getModel` and `getTexture`. This will create two separate registries to suffice the use cases of biped and horse armor models. There is also a remake of `HorseModel` that users should use for custom models since all of its `ModelRenderer`s were private making it inaccessible without the use of reflection.

`IModelProvider` is a functional interface where it requires that `getModel` be set to some value or null. `getTexture` can be overriden to set a custom location for the resource or else it will just default to the one provided by the armor items.


ChampionAsh5357:
Outdated solution and probably won't be coming back to it for a while. Closing for these reasons.

PR #7268, opened 2020-08-22
Re-implemented GuiUtils#drawHoveringText

Yunus1903:
The `drawHoveringText` methods inside `GuiUtils` had been commented out in the update to 1.16.2 due to the addition of `IReorderingProcessor`.

I tried to make these methods use the `List<? extends IReorderingProcessor>` parameter first instead of the original `List<? extends ITextProperties>` (Lex had already done this so I assumed that's what he intended ). This was a problem because wrapping the lines would not have been possible this way.

So I just left the original `List<? extends ITextProperties>` parameter and made changes inside the method itself.
I believe this is better to do because editing the lines in the `RenderTooltipEvent` would otherwise not be possible.

### DIFF
--- a/src/main/java/net/minecraftforge/fml/client/gui/GuiUtils.java
+++ b/src/main/java/net/minecraftforge/fml/client/gui/GuiUtils.java
@@ -21,18 +21,28 @@ package net.minecraftforge.fml.client.gui;
 
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.gui.AbstractGui;
+import net.minecraft.client.gui.FontRenderer;
+import net.minecraft.client.renderer.IRenderTypeBuffer;
 import net.minecraft.client.renderer.Tessellator;
 import net.minecraft.client.renderer.BufferBuilder;
 import net.minecraft.client.renderer.vertex.DefaultVertexFormats;
 import net.minecraft.item.ItemStack;
 import net.minecraft.util.ResourceLocation;
 import net.minecraft.util.math.vector.Matrix4f;
+import net.minecraft.util.text.ITextProperties;
+import net.minecraft.util.text.LanguageMap;
+import net.minecraft.util.text.Style;
+import net.minecraftforge.client.event.RenderTooltipEvent;
+import net.minecraftforge.common.MinecraftForge;
+
 import org.lwjgl.opengl.GL11;
 
 import com.mojang.blaze3d.matrix.MatrixStack;
 import com.mojang.blaze3d.systems.RenderSystem;
 
 import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * This class provides several methods and constants used by the Config GUI classes.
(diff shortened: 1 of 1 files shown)


marchermans:
Why did you change the type from `IReorderingProcessor` to `ITextProperties` ?

Yunus1903:
> Why did you change the type from `IReorderingProcessor` to `ITextProperties` ?

Those methods originally used `ITextProperties`, I tried to migrate the methods to `IReorderingProcessor` but because you can't convert this type back into a `ITextProperties` things like wrapping the text and editing the text inside the events would not be possible.

Also this would make existing calls to the method still work.

marchermans:
@Yunus1903 This is not supposed to be used to rewrap lines etc.
IReorderingProcessors are already wrapped and trimmed lines.

MC Made the change away from ITextProperties to IReorderingProcessors to enable BiDi-Language support, like support for arabic writing which requires RTL.
This behaviour should be followed here. I do not see a valid reason or use case to make those render methods any different. Use the font renderer to split the lines -> this results already in a list of IReorderingProcessors which you can then pass to this method.

TLDR: I would love to see this just being an IReorderingProcessor instead of ITextProperties.

ichttt:
One of the major points of this funtion is that it supports wrapping lines, to wrap depending on the screen size (vanilla can't do that), which is very handy when mods define large tooltips.
So yes, this unfortunally needs to be `ITextProperties` to enable the old behavior.

(Btw this PR is not complete as it is missing the patch for the Screen class)

Yunus1903:
> (Btw this PR is not complete as it is missing the patch for the Screen class)

Ah yes forgot about this. I can't really think of a way to re-implement that method call though as you can't convert `IReorderingProcessor` back into a `ITextProperties` AFAIK.

KnightMiner:
Can always make two methods. One that takes the reordering processor list and draws directly, one that takes ITextProperties to call the split method

ichttt:
That would be pointless, though. The only point of the patch was (AFAICT) to allow tooltips to be wrapped when there is not enough space on the screen.

marchermans:
@ichttt You can already do that yourself though.
You know the screenwidth, you know where you want to render. So just split your strings properly before rendering.

For just the auto string splitting, reverting the ability of RTL, which mojang is pushing, is not really a good idea.

raoulvdberge:
I don't want to do the splitting myself, this utility exists for a reason.

Yunus1903:
I'd say we should leave the method as it is (like I did in my commit). But I'm not sure how to handle the patch in the `Screen` class. I believe it's only there to trigger the event (it passes a `maxTextWidth` of `-1`), correct me if I'm wrong.

ichttt:
I agree with @raoulvdberge, and additionally, this still supports RTL, as the splitted string is still converted into IReorderingProcessor before passing it to the font renderer, so it should be fine.
And no, the patch is not only there to trigger the event, it is also there to split large tooltips if they don't fit on the screen. -1 just means no additional max-width requirement, but screen size max width still applies.

Yunus1903:
> And no, the patch is not only there to trigger the event, it is also there to split large tooltips if they don't fit on the screen. -1 just means no additional max-width requirement, but screen size max width still applies.

Oh you're right, didn't think of that. I can't really think of a way to re-implement that feature though. Should I just leave it out for this PR? Then at least `GuiUtils#drawHoveringText` would be useful again. Unless you or someone else has a suggestion.

ichttt:
Sure, leave it out for now, but I think we should fix GuiUtils soon-ish, as it is really useful for a bunch of mods.

Crimix:
Any news on this, is it missing something?

Yunus1903:
Other than the patch in the `Screen` class (which is a whole issue by itself), this PR should be ready to be merged.

Vazkii:
@cpw this is what I told you on twitter was missing.

marchermans (approved):
Since people seem to want this.
I reviewed it again.

Works on my instance, and code seems to be in order.

marchermans:
@Yunus1903 Fix merge conflicts before I can assign it please.

raoulvdberge:
@OrionDevelopment  You suggested that an `IReorderingProcessor` is already trimmed and wrapped:
> This is not supposed to be used to rewrap lines etc.
IReorderingProcessors are already wrapped and trimmed lines.

Is that true or not?

@ichttt suggests otherwise:

> One of the major points of this funtion is that it supports wrapping lines, to wrap depending on the screen size (vanilla can't do that), which is very handy when mods define large tooltips.
So yes, this unfortunally needs to be ITextProperties to enable the old behavior.

Yunus1903:
@OrionDevelopment It should be ok now

PR #7271, opened 2020-08-25
Add validation via resources for tag providers

tterrag1098:
Removes the hack of running the vanilla tag datagen and then filtering out its content.

Adds an (optional) `ExistingFileHelper` param to `TagsProvider`. If present, this will be used to search for existing tags on the resource path.
Adds a new (default) method to `ITagEntry`: `exists`, which accepts a `String` (the tag "type" or folder name) and an `ExistingFileHelper` and returns true if this entry exists. For item entries, this checks the item registry. For tag entries, it looks in the existing data files. Optional entries always return true (the default impl).

Updates all the forge code to use the new ctors on tag providers, and re-ran the datagen test.

Moved `ExistingFileHelper` out of client package.


patches/minecraft/net/minecraft/tags/ITag.java.patch
```diff
@@ -41,18 +41,37 @@
...
+    public static class ItemEntry implements ITag.ITagEntry {
+@@ -178,6 +191,12 @@
        public String toString() {
           return this.field_232969_a_.toString();
        }
 +      @Override public boolean equals(Object o) { return o == this || (o instanceof ITag.ItemEntry && java.util.Objects.equals(this.field_232969_a_, ((ITag.ItemEntry) o).field_232969_a_)); }
++
++      @Override
++      public boolean exists(String type, net.minecraftforge.common.data.ExistingFileHelper existingFileHelper) {
++         return net.minecraftforge.registries.ForgeRegistries.ITEMS.containsKey(field_232969_a_);
```
Cyborgmas: Pretty sure `ItemEntry` is meant for a single entry, `EntityTypeTags` and `FluidTags` are also `ItemEntry`'s.
The `DataGeneratorTest` uses `BlockTags` but I am guessing it worked on the fact that blocks and items share the same name.

patches/minecraft/net/minecraft/data/TagsProvider.java.patch
```diff
@@ -1,34 +1,56 @@
...
  
              try {
                 String s = field_200437_e.toJson((JsonElement)jsonobject);
-@@ -78,7 +85,7 @@
+@@ -70,7 +85,7 @@
+                field_200436_d.error("Couldn't save tags to {}", path, ioexception);
+             }
+ 
+-         }
++//         }
```
ichttt: I would just remove the else and keep the brackets, as this commented out curly bracket is floating around quite freely otherwise (which might cause issues when porting patches)

patches/minecraft/net/minecraft/data/TagsProvider.java.patch
```diff
@@ -1,34 +1,56 @@
 --- a/net/minecraft/data/TagsProvider.java
 +++ b/net/minecraft/data/TagsProvider.java
-@@ -28,11 +28,17 @@
+@@ -28,11 +28,19 @@
     protected final DataGenerator field_200433_a;
     protected final Registry<T> field_200435_c;
     protected final Map<ResourceLocation, ITag.Builder> field_200434_b = Maps.newLinkedHashMap();
 +   protected String modId;
++   net.minecraftforge.common.data.ExistingFileHelper existingFileHelper;
```
ichttt: should be private or protected, but package-private makes no sense here

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -20,6 +20,7 @@
 package net.minecraftforge.client.model.generators;
 
 import net.minecraft.util.ResourceLocation;
+import net.minecraftforge.common.data.ExistingFileHelper;
```
ichttt, line 23: Unnecessary import here and a few times below

ichttt (commented):
Some small notes, looking good to me otherwise.

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
tterrag1098: It's not unnecessary...the class moved.

patches/minecraft/net/minecraft/tags/ITag.java.patch
tterrag1098: That's going to pose a problem then. This method is probably going to move to a static with hardcoded type checks since these 4 entry types are fixed, and I can accept the registry as a parameter there.

tterrag1098: Thinking about it more, this code is completely unnecessary. An `ItemEntry` will never reach this point because validating the registry contains the entry is handled by `func_232963_b_`. I'm going to push a new version that only cares about `TagEntry` and hardcodes it into `TagsProvider`.

PR #7276, opened 2020-08-26
Fixing logic in getCurrentRailPosition to correctly identify the rail…

ghost:
… block that a given Minecart is on


ichttt:
For reference: In 1.12.2, the logic was still ok (https://github.com/MinecraftForge/MinecraftForge/blob/f1cca475eadea9d15c996e521bf9b6b5970425e9/patches/minecraft/net/minecraft/entity/item/EntityMinecart.java.patch#L268), it seems be a porting mistake from https://github.com/MinecraftForge/MinecraftForge/commit/b6ddb0a7cf2b1ff353ff450c895c09f55753c7e9#diff-c50b449473370ca9f8a79f0b8efb0e74

PR #7279, opened 2020-08-27
]1.16] Optimize PlayerList#getPlayerByUsername

IzzelAliz:
The vanilla logic can be slow when lots of players are present.

### DIFF
--- a/patches/minecraft/net/minecraft/server/management/PlayerList.java.patch
+++ b/patches/minecraft/net/minecraft/server/management/PlayerList.java.patch
@@ -1,14 +1,15 @@
 --- a/net/minecraft/server/management/PlayerList.java
 +++ b/net/minecraft/server/management/PlayerList.java
-@@ -110,6 +110,7 @@
+@@ -110,6 +110,8 @@
     private GameType field_72410_m;
     private boolean field_72407_n;
     private int field_72408_o;
 +   private final List<ServerPlayerEntity> playersView = java.util.Collections.unmodifiableList(field_72404_b);
++   private final Map<String, ServerPlayerEntity> playersByName = Maps.newHashMap();
  
     public PlayerList(MinecraftServer p_i231425_1_, DynamicRegistries.Impl p_i231425_2_, PlayerData p_i231425_3_, int p_i231425_4_) {
        this.field_72400_f = p_i231425_1_;
-@@ -146,6 +147,7 @@
+@@ -146,6 +148,7 @@
        IWorldInfo iworldinfo = serverworld1.func_72912_H();
        this.func_72381_a(p_72355_2_, (ServerPlayerEntity)null, serverworld1);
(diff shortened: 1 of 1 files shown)


ichttt:
While performance is always a good aim, this seems kind of arbitary. Please provide some numbers why this should be a hot spot

IzzelAliz:
In most cases it would be a command player lookup. On my end (Arclight) this method is called frequently though I don't have statistics or benchmarks.
I would say this is far from a hot spot but I  (or other dev lookup players by name) would benefit from it.

diesieben07:
Your last message is evidence that this is almost purely premature optimization.
Patches to Minecraft code are debt that needs to be carried forward every update. Unless there is a good reason for this, I don't think this should be added.

Darkhax:
I did some basic benchmarks under simulated conditions using 10k players.

## Data
| Test      | Run 1 (ms)      | Run 2 (ms)      | Run 3 (ms)      | Avrg (ms)    |
|-----------|---------|---------|---------|---------|
| Loop (A)  | 1052.83 | 1003.48 | 1007.71 | 1021.34 |
| Cache (A) | 75.55   | 74.62   | 68.55   | 72.90   |
| Loop (B)  | 973.02  | 872.43  | 992.57  | 946.00  |
| Cache (B) | 80.96   | 78.32   | 103.90  | 87.72   |

## Procedure
1. Generated a world with 10k player instances. Each with random UUID and name.
2. Created a shuffled set containing the names of all 10k players.
3. Warmup by sequentially looked up each name in the random set 10k times with both vanilla and PR code. (200m lookups)
4. Sequentially look up each name in the random set 3 times using Vanilla, timing the duration of each. (A)
5. Sequentially look up each name in the random set 3 times using PR, timing the duration of each. (A)
6. Sequentially look up each name in the random set 3 times using PR, timing the duration of each. (B)
7. Sequentially look up each name in the random set 3 times using Vanilla, timing the duration of each. (B)

## Conclusion
While there is a clear improvement in execution time this PR is very unnecessary. Other bottlenecks would get in your way long before you hit 10k concurrent players, but even under such a high load this method's individual execution time is well below 1ms and is well within the range for acceptable performance. This PR would only improve performance for things that consistently do massive amounts of name lookups like a TE checking for their owner every tick, however those use cases should be using UUIDs instead.

sciwhiz12:
Hello! @IzzelAliz, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.
Additionally, the above benchmarks by @Darkhax show that this PR seems like unnecessary and a premature optimization. If you do update this PR, please provide concrete benchmarks and statistics to confirm the viability of this PR.

sciwhiz12:

Issue #6692, opened 2020-05-13
GL_FRAMEBUFFER_INCOMPLETE_ATTACHMENT


Darkhax:
**Minecraft Version:** 1.15.2

**Forge Version:** 31.1.87

**Logs:** https://gist.github.com/Darkhax/029b35d1c26340cfdf59aee8ec923b82

**Steps to Reproduce:**
 1. Try to start the game.
 2. Sometimes this happens.

**Description of issue:** When I launch the game it will sometimes throw this exception and crash. This specific log is from a dev environment with basically nothing in it. I've reproduced it across a few different projects however not consistently.

BedrockLegends:
I have ran into this same issue on my packs

LexManos:
Unable to reproduce in any of our tests. You're gunna have to dig into this.

TheCurle:
Small update: it looks like this is caused by the early launch window, something about the state of the thread just before starting the game - meaning it's basically random. Will look into adjustments to fix this, and narrow down the exact cause

josephtarbit:
I’m having this issue too, it’s very intermittent.

Darkhax:
I am trying to get indepth OpenGL logs for this issue however attaching the program RogueLogic mentioned is not realistic to run 24/7. I did manage to reproduce it on 1.16.1. Here is a [1.16.1 log](https://gist.github.com/Darkhax/ae979215dadd056f2375c0fe7d9aa4f7).

TheCurle:
We have sufficient logs (about 20 each for 1.15 and 1.16). We know it's caused by the early progress window so a `-Dfml.earlyprogresswindow=false` will work as a quick workaround. 

Without a proper GL debugging trace, though, we can't do much. Try as I might, most of the computers i can cause this error reliably on, are simply not fast enough to run the trace and cause the error before filling its drive. 

We *need* a proper trace, and we're trying. Expect updates on this.

mrp-v2:
I consistently get this error, even on the latest 1.16.4 forge version, whenever I minimize the early loading window.

TheCurle:
Looking into this with the provided dumps. https://mega.nz/file/9ZQyUD4I#9CNgPYHh1rXxahnuPVP-P5pXW14agVZmbWwX8OILpcE - Early Window Disabled
[...]

PR #7280, opened 2020-08-27
Try to fix GL_FRAMEBUFFER_INCOMPLETE_ATTACHMENT error

ichttt:
- Only allow polling events if the render thread is currently not rerendering
- Stop the render thread and poll events before any other GLFW code is run on the main thread

Could be a possible fix for #6692, as it looks like some kind of race condition, but it's hard to tell if this fixes it as it's hard to reproduce.
The issue definitly exists in 1.16.2 as well: https://pastebin.com/gyJpbbg7


mrp-v2:
I consistently get this error, even on the latest 1.16.4 forge version, whenever I minimize the early loading window.

mrp-v2:
Even with the changes in this PR I still consistently get the error when I minimize the early loading window.

ichttt:
cpw found the real issue in 47e076071b37ee285a1439b554c0d14c7a735bc4, closing

PR #7287, opened 2020-08-29
Add data driven biome modification system.

Cyborgmas:
This is not quite finished yet, but it's good enough to share progress I think and I would like comments. 
The major problem in 1.16.2 is that it is currently impossible to simply add features or mobs in a biome or to add biomes to a dimension.

To solve this, I hook into the parsing done by `WorldSettingsImport` and recreate the object with the additions.
This one hook will have all objects of the 9 registries covered by `DynamicRegistries`, for now only `Biome`s are handled from it, but if need be it can handle other objects.

The issue is that re-creating a whole Biome is not wanted, only re-creating some precise objects, such as the `Map<EntityClassification, List<MobSpawnInfo.Spawners>>` in `MobSpawnInfo` or the `List<List<Supplier<ConfiguredFeature<?, ?>>>>` in `BiomeGenerationSettings`.

To handle that, `Codec`'s are used which copy all the unecessary objects (such as primitives), but goes into the objects that contain the wanted objects, such as `BiomeGenerationSettings`. When at the desired object, it is then handled in WorldGenAdditions.

For biomes, there are currently 4 additions implemented: `Structures`, `Features`, `Spawners`, `SpawnCosts`. To add some features to a biome, simply add a `json` file in `additions/biomes/features` (subject to change).
In it, a "targets" key is needed which will be a list of strings. These strings can be a `Biome.Category` or a `Biome`'s registry name.
```
  "targets" : [
    "savanna",
    "minecraft:plains",
    "forest"
  ]
```
To differentiate categories and biomes, the `minecraft` namespace is required here. "savanna" refers to the category and "minecraft:plains" refers to the biome. This will probably be adapted to use biome tags in the future, for even more flexibility.

For now, Features and Structures are in the same `json` under a "features" and "starts" field respectively. Spawners and SpawnCosts as well are in the same `json`, they are under `additions/biomes/mob_spawns`. They need a "spawners" and "spawn_costs" key respectively.
This might change, since it requires to have them be an `optionalFieldOf` otherwise a file with only "features" would error since no "starts" field is defined. The problem being that there is no way to differentiate with an actual key error.

For adding biomes to dimensions: things get a lot more messy. `Dimension`s are not part of `DynamicRegistries` and their entire registry is created from `DimensionGeneratorSettings`.
The codec part is a bit different, but in the end same thing: the object to modify is buried deep and through codecs it gets "extracted".
For now, only the biomes of `NetherBiomeProvider` (called MultiNoise on the mc wiki) can be added to, using a similar `json` structure, with a "targets" key that contains the dimension's registry name. (minecraft:the_nether being the only vanilla one). This makes it possible to add to another mod's dimensions as well.

Where it gets difficult is for the `OverworldBiomeProvider`, which does require a `List<Biome>` in its `super` but does not depend on it for which biomes to generate. it only uses a `Layer` which is a pretty complicated mess. It is mainly defined in `LayerUtil#func_237216_a_()`. From what I can gather, it would need to be hooked into for mods to add their own layers. However, the order of the layers seem to be pretty freaking important, but would have to be undefined depending on the mods that exists. 
I don't have many ideas on how to deal with the Overworld right now so ideas are welcome. I'll explore more on the idea once biome tags are added (hopefully), which will help with some stuff that only uses the int id of biomes.


alcatrazEscapee:
> I would like comments.

I can do that!

With respect to the `targets`, it is fairly straightforward but is lacking a few useful features:

- The usage of biome categories as a stand-in for biome dictionary / tags is good, but lacking more fine grained control. For example, the ability to use combinational logic (given a `hot` and `dry` tag, apply to biomes that are `hot` **and not** `dry`). I'd like to see conditions such as `and`, `not` supported, in addition to the ability to define tags (e.g. `#forge:hot`)
- This may take the form (similar to recipe conditions) of having a target type, e.g.
```json
"targets": {
    "type": "forge:and",
    "targets": [
        "#forge:hot",
        {
            "type": "forge:not",
            "target": "#forge:dry"
        }
    ]
}

```
- (More of a convenience), a way to list "all biomes" would be useful. Right now the next best thing would be listing all biome categories as it's a fixed enum.

The other thing I think is lacking here is there is no ability (that I can see) for customization outside of straight additions. For example:

- Removing vanilla features based on a condition, such as removing vanilla ore generation and replacing it with something else.
- Modifying properties of a biome (changing temperature, downfall, effects, etc.)
- Selectively adding features based on other properties of the biome (e.g. only add special emerald ore features in biomes that already spawn emerald ores.)

Now any *modification* as opposed to a straight up addition has the potential to conflict if two mods want to overwrite the same thing. In this case, I think it would be reasonable to assume that the last modification would "win" per say (order being determined by datapack load order).

Again, what might be more useful in this case would be a type driven system, where modders have the ability to define custom "biome modifications", in addition to standard ones used by Forge.  For example, adding features may use the "forge:add_feature" type:
`example_plains_modifier.json`
```json
"targets": [ "minecraft:plains" ],
"type": "forge:add_features",
"features": [ "generictechmod:copper_ore" ]
```
This would then pass all biomes of interest to the `forge:add_feature` type instance, and it would perform modifications as it sees fit. (A more similar system to how global loot modifiers are implemented, in this case it would be global biome modifiers). Modders can define custom types, and inspect / modify the biome as they need at that stage.

Cyborgmas:
Regarding the conditional `targets` , I think it would be best to have the logic be inside of tags directly perhaps?
So you would add a `hot_but_not_dry` tag in it and that tag would have the actual logic and when running it would contain all the elements in the `hot` tag that are not in the `dry` tag. 
If that seems excessive I can surely add the logic just in `targets`.
For listing all biomes, you can simply have an empty array  `"targets" : [],` .

For the modifications, I would argue there is no need for forge there. You can override a biome completely already (`data/minecraft/worldgen/biome/plains.json` -> overrides plains biome). This will indeed conflict but the winner will be determined by datapack order (which, as you said, is acceptable)

For replacing a vanilla feature again that should be doable by copying the configured feature's registry name and putting it in the minecraft namespace. I say should because I have not tested it, but pretty positive that will work.
For straight out removing a feature you could also just replace the feature with a NOOP one (again have not tested). This only covers complete removals though. For per-biome, I could perhaps add something for that.

> Selectively adding features based on other properties of the biome (e.g. only add special emerald ore features in biomes that already spawn emerald ores.)

Not sure how useful that would be, I'd say it would be best to do that with a biome tag. For this example you could have a `has_emerald` tag that would reference the `mountain` tag by default. With time, you would come to know other modded biomes that are not mountain but have emerald and you can add them as optional to the tag. 
The issue with your idea (I think), it's pretty impossible to know if a feature "spawns emerald", for vanilla there is only one of course but what if a mod adds some that adds a cave of gems, including emeralds? At least this way it is easily editable by anybody.

> Modders can define custom types, and inspect / modify the biome as they need at that stage.

The issue I find with `global biome modifiers` is that the difference with loot tables modifying a biome can have some much bigger and unpredictable consequences. I think (I might be completely wrong) that it is preferable to have additions be easy and compatible but modifications be left to the vanilla way (override the whole biome). I say best to leave biome modifications to mods that have that explicit purpose.

alcatrazEscapee:
If I understand correctly, the aim that you're going for here is for tags to be the ultimate identifiers for any kind of biome replacement (which I don't disagree with). My primary issue is that is means nothing is automatic (both for the modder and from the user perspective).

Consider  `#hot_but_not_dry`: How does this tag come to exist, *and be used by mods*? I cannot assume a mod will add their biomes to an arbitrary tag of my creation. And unless biome tags themselves can have and/or/not logic, this just pushes the problem elsewhere:

**A given mod, that wants conditional checks on biomes, *must* be manually specifying biomes for these logical checks to operate on (whether that be in `targets`, or in the definition for a tag) at some level**. This is something (if possible) I'd very much like to avoid.

Here's my example use case for the modifications in particular. I have a mod which converts the *entire* world to a post-apocalyptic nuclear winter version of itself (working in 1.15 and 1.16.1) via these type of biome modifications. It currently
1) Does not require the user to manually configure which biomes should be winter-ized
2) Automatically works with any modded biomes that are added, forcefully converting them to winter versions.
3) Has a config setting to disable all invasive modifications if they are undesirable by a pack maker.

This is an ideal system for 3) to be automatic, but 1) and 2) are just not possible without explicitly "overriding" every biome, which is something I would very much like to avoid as it creates more work for me, less out-of-the-box compatibility for users, and incompatibilities with mods that need to do similar things.

Cyborgmas:
For the `hot_but_not_dry` tag, you missed my point. 
There would be a `forge:hot` tag and `forge:dry` tag, just instead of doing the logic inside of `targets` it's done inside a custom tag `modid:hot_but_not_dry`. Which will be updated accordingly, depending on the content of `forge:hot` and `forge:dry`, which are 2 tags that you could expect modders to use.

For the other part I'll get back to you, got some thinking to do.

Drullkus:
How feasible would you think it'd be to set a feature or structure-start filter based on the world id? Unfortunately this kinda thing isn't even in Vanilla itself but I have a few friends who would much appreciate a such feature

Cyborgmas:
So I am thinking that worldgen modification would be split in 2 categories: safe stuff and unsafe stuff.
Safe stuff would be basically everything that I currently have done: additions. This would be data-driven for sure.

However, unsafe stuff would (I think) not be datadriven. Example: if you want to do heavy modifications to the features of a biome, I would make it be an event (or some sort of pseudo-registry) where you would get the `RegistryKey<Biome>` and would need to give something like a `UnaryOperator<List<List<Supplier<ConfiguredFeature<?, ?>>>>>`  AND your mod id. That way, the "unsafe" modifications can be tracked and if 2 or more conflict (modify the same object), a big warning can be logged and if a weird world is created and a user reports it, debugging can be done easily.

Some aspects of "unsafe" modifications could probably be data-driven, but all possible modifications would be a bit too much. The other solution would be a `global biome modifier` system I guess?

Also would this be too much? Basically right now I am at allow everything but have it be tracked.

(Pretty sure it would not be straightforward Drulkus, but I'll check it more in depth at some point.)

alcatrazEscapee:
I would ideally want unsafe modifications to be *more* data driven than safe ones - because these are ones that are more important for a pack maker to be able to enable / disable / modify as they have a higher chance of causing conflicts. Identifying it is only part of the problem - being able to disable or even modify it so it does not conflict short of "remove the mod" would give pack makers a lot more fine grained control in this matter.

I think a type driven system would be the ideal solution to these issues. For example, say mods can define a modifier in code which (essentially) is a `UnaryOperator<Biome>`. This can be setup to be loaded via json, so additional configuration elements can be added as needed (for example, in a forge provided `forge:add_feature` type reading the list of feature ids).

Tracking modifiers is very simple: the modifers are all json, their targets are all json. Inherently, these include the mod IDs in question. The order is viewable from the user end (data pack order), and any manner of log messages can be spat out when multiple modifiers modify the same biome.

But moreover, any kind of possible biome modification is very unlikely to generate *hard incompatibilities* as all that's really being done is modifying data. Mods *should expect* that their biomes may not have exactly the data they were defined with, as the end user is entirely free to override them using a datapack. The more likely *conflict* would simply be weird or unintended behavior as two mods with very different purposes try and modify a biome. And in this case, there's no system that can prevent that, but seeing as it shouldn't outright *break* anything, the simple solution is to 1) allow it, 2) make it obvious what's happening (logs), and 3) make it data driven enough that the pack maker/user can edit it.

Cyborgmas:
Okay completely rewrote the system. Hopefully a lot better and more versatile, went with a `type` system instead.

I add 2 registries (hopefully that's fine) one for `BiomeConditionType`s and one for `BiomeModifierType`s. Both are simply codec holders (i.e. serializers/deserializers) for their respective objects.
BiomeConditions are a Biome Predicate, that say which Biome will be modified. This allows modifying all biomes that have a temperature > 0.8 for example.
The `BiomeModifier` needs to return a `BiomeModifications` object, which just holds operators for each possible field to modify.
This is to be able to chain with other modifications on the same biome (and depend/adapt on any potential modifications done before yours). Following this, I made it so that the first modifier in the first datapack will be executed last. Since that is where it'll have the most effect/power.

However, I decided to reduce the scope of the PR. This now only adds possibilities for modifying biomes. Adding biomes to dimensions will have to come a bit later.

Comments welcome!

TelepathicGrunt:
> How feasible would you think it'd be to set a feature or structure-start filter based on the world id? Unfortunately this kinda thing isn't even in Vanilla itself but I have a few friends who would much appreciate a such feature

Drullkus, dimension/world blacklisting should be done within the structure/feature itself. Well, structures is more complicated as they cannot get the world passed in where it is needed to do dimension/world blacklisted but maybe something can be done about the structure spacing in the world's chunkgenerator's dimensionsettings. Anyway, the point is that's more of a thing for modders to implement themselves and provide a config for. This pr is for specifically biomes themselves and what they hold which is not dimension/world specific (multiple worlds can use the same biomes).

TelepathicGrunt (approved):

TelepathicGrunt:
Just looked over and it seems to be for the most part, good. However, I have one main concern about this. I get the main benefit of this system is to allow players to use datapack to override what biomes the mod's stuff gets added to. But how will the player know about this system? How to use it? And what entries are in the registry that they can use for conditions? I say this as unlike some other PRs for APIs into forge, this one seems to be not just for modders but more for players/modpack makers which means a different approach will need to be done instead of just merging and walking away. 

Still, be sure to put how to use this API on the Forge Docs so modders know how to make conditions and all. But as for players, idk what the best solution is to make sure this system is visible to them. I bring this up because an API really only matters when people actually use it and the last thing I want is for your hard work to go unused and forgotten.

src/main/java/net/minecraftforge/common/world/biomes/ForgeBiomeModifiers.java
```diff
@@ -0,0 +1,34 @@
...
+    public static final RegistryObject<BiomeConditionType<BiomeInvertedCondition>> INVERTED = BIOME_CONDITIONS.register("inverted", () -> new BiomeConditionType<>(BiomeInvertedCondition.CODEC));
+    public static final RegistryObject<BiomeConditionType<BiomeAndCondition>> AND_COMBINED = BIOME_CONDITIONS.register("and_combined", () -> new BiomeConditionType<>(BiomeAndCondition.CODEC));
+    public static final RegistryObject<BiomeConditionType<BiomeOrCondition>> OR_COMBINED = BIOME_CONDITIONS.register("or_combined", () -> new BiomeConditionType<>(BiomeOrCondition.CODEC));
+    public static final RegistryObject<BiomeConditionType<NonVoidBiomeCondition>> NON_VOID = BIOME_CONDITIONS.register("non_void", () -> new BiomeConditionType<>(NonVoidBiomeCondition.CODEC));
+
+    public static final RegistryObject<BiomeModifierType<SimpleFeaturesAdditions>> SIMPLE_FEATURES = BIOME_MODIFIERS.register("features_additions", () -> new BiomeModifierType<>(SimpleFeaturesAdditions.CODEC));
+    public static final RegistryObject<BiomeModifierType<SimpleSpawnsAdditions>> SIMPLE_SPAWNS = BIOME_MODIFIERS.register("spawns_additions", () -> new BiomeModifierType<>(SimpleSpawnsAdditions.CODEC));
```
TelepathicGrunt, line 26: Oh yeah, I noticed you only have additions BiomeModifierTypes registered by default. Maybe add replacement BiomeModifierTypes and removal BiomeModifierTypes so players always has replacement/removal of features/spawns as an option without needing another mod to add them to the registry.

Cyborgmas:
I'd say that the PR is mostly for modders first, as of now there is no way to modify biomes in a clean way. The data driven part is secondary, it's mainly just to stay in line with how vanilla does things and also it gives the *possibility* of modpack makers and even users to modify, tweak or just remove some of the modifiers. Will it be used a lot? who knows, but it's there for those that need it.

And for the forgedocs, im kinda at my limit here heh, so maybe, but no promises, I put already quite a few javadocs/comments.

src/main/java/net/minecraftforge/common/world/biomes/ForgeBiomeModifiers.java
Cyborgmas: Sure, i could.

Cyborgmas:
Datagen will have to wait to do it properly, kinda need https://github.com/MinecraftForge/MinecraftForge/pull/7257 (or similar, but dont feel like deduplicating it), to properly encode any registered references.
So for now, I'd say it's looking pretty good.

src/main/java/net/minecraftforge/common/world/biomes/conditions/BiomeOrCondition.java
```diff
@@ -0,0 +1,39 @@
...
+import net.minecraftforge.common.world.biomes.ForgeBiomeModifiers;
+import net.minecraftforge.common.world.biomes.conditions.base.BiomeConditionType;
+import net.minecraftforge.common.world.biomes.conditions.base.IBiomeCondition;
+
+public class BiomeOrCondition implements IBiomeCondition
+{
+    public static final MapCodec<BiomeOrCondition> CODEC = RecordCodecBuilder.mapCodec(inst ->
+            inst.group(
+                    IBiomeCondition.INNER_CODEC.fieldOf("first").forGetter(or -> or.first),
+                    IBiomeCondition.INNER_CODEC.fieldOf("second").forGetter(or -> or.second)
```
alcatrazEscapee: Same here, using a list instead of two fields would reduce unnecessary nesting in json.

src/main/java/net/minecraftforge/common/world/biomes/conditions/BiomeAndCondition.java
```diff
@@ -0,0 +1,39 @@
...
+import net.minecraftforge.common.world.biomes.ForgeBiomeModifiers;
+import net.minecraftforge.common.world.biomes.conditions.base.BiomeConditionType;
+import net.minecraftforge.common.world.biomes.conditions.base.IBiomeCondition;
+
+public class BiomeAndCondition implements IBiomeCondition
+{
+    public static final MapCodec<BiomeAndCondition> CODEC = RecordCodecBuilder.mapCodec(inst ->
+            inst.group(
+                    IBiomeCondition.INNER_CODEC.fieldOf("first").forGetter(or -> or.first),
+                    IBiomeCondition.INNER_CODEC.fieldOf("second").forGetter(or -> or.second)
```
alcatrazEscapee: A suggestion: Instead of two fields, use `IBiomeCondtion.INNER_CDEC.listOf()` and `or` together all the elements. That would avoid unnecessary nesting of multiple conditions in json.

alcatrazEscapee (approved):

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -306,4 +306,6 @@ public net.minecraft.world.gen.layer.LayerUtil func_202829_a(JLnet/minecraft/wor
 private-f net.minecraft.world.raid.Raid$WaveMember field_221284_f # VALUES
 public net.minecraft.world.server.ServerChunkProvider field_186029_c # chunkGenerator
 public net.minecraft.world.server.ServerChunkProvider field_73251_h # worldObj
-public net.minecraft.world.storage.FolderName <init>(Ljava/lang/String;)V # constructor
\ No newline at end of file
+public net.minecraft.world.storage.FolderName <init>(Ljava/lang/String;)V # constructor
+public net.minecraft.world.biome.BiomeGenerationSettings <init>(Ljava/util/function/Supplier;Ljava/util/Map;Ljava/util/List;Ljava/util/List;)V
+public net.minecraft.world.biome.MobSpawnInfo <init>(FLjava/util/Map;Ljava/util/Map;Z)V
```
pupnewfster: Make sure to run the `checkATs` task as I am pretty sure that it will modify this given if memory serves the file should be alphabetical and the `b` in `biome` comes before the `s` in `storage`.

src/main/java/net/minecraftforge/common/world/biomes/BiomeModifications.java
```diff
@@ -0,0 +1,188 @@
...
+    public BiomeModifications modifyPlayerSpawnFriendly(UnaryOperator<Boolean> playerSpawnFriendly)
+    {
+        this.playerSpawnFriendly = playerSpawnFriendly;
+        return this;
+    }
+
+    static Biome performModifications(List<BiomeModifications> modifications, Biome base)
+    {
+        return modifications.isEmpty() ? base :
+                new Biome.Builder()
```
pupnewfster: I feel like this method would be more readable if you just split out the `? :` into an if statement

Lothrazar:
Just for reference, slicedlime (Mojang developer) has called system adding/replacing with JSON-worldgen system in 1.16.3 release "experimental".  Meaning literally everything could change in a major or minor release in the future.  

"It does use those files, but they’re not included in the jar yet. You can still define your own files to add/replace, but it is considered experimental."
https://twitter.com/slicedlime/status/1305056397420306432

maxanier:
As far as I can tell this is probably superseded by the new biome event. However, I think this is the more suitable solution once the vanilla JSON-worldgen system is considered stable by Mojang (1.17?).

Cyborgmas:
Okay updated so it has no conflicts. 
I made the BiomeLoadingEvent redirect towards this new system. I was hoping to be able to call the event simply at the appropriate time but an event is way too messy to handle ordering.

Instead of hooking directly when the biome is created I hook only into when the WorldSettingsImport is created with a ResourceManager before world creation. I have it use a custom IResourceAccess which will tell what object it can load and load those objects.
This IResourceAccess will first check the datapacks for the biomes there and then it will check the BiomeModifierManager for all biomes that meet the conditions to be modified and add those as well. It can then properly load up both biomes completely from disc and normally registereds ones, before applying the modifications.

The good thing about this method is that it will happen in between datapack loading and worldgen. so by the point i check that the conditions are met, tags will have been loaded. 

This is still a bit WIP, I haven't gotten to actually testing a tag condition, but in *theory* it works.
I also have some tweaking to do with managing the biomeloadingevent. because as it stands it has the most priority if a biome is modified using both the event and the json system. (+ other stuff, i just realized i didnt do oops)

However one main question: is this still under consideration? (at least this version of a json system), because this PR will essentially deprecate BiomeLoadingEvent..., as I can't find a sane way to use it.

I don't have much time currently. so my work on this is sporadic. If anyone would like to help move things along faster, testing the tags as i mentioned above and just general stress testing of both systems would be appreciated.

Cyborgmas:
Was fun to do, but probably needs to be re-written entirely now and the alternative seems to be working okay, so not seeing much use for this.

PR #7288, opened 2020-08-30
[1.16.x] Added model data to ItemStack rendering

CJMinecraft01:
As the title suggests I have added functionality so that items can provide `IModelData` to pass onto the `getQuads` method of the `IBakedModel` for that item using the provided ItemStack.
This allows items to specify model data properties using NBT data and other information provided by the ItemStack parameter.
I have also provided a relevant test which allows an item to use another item's model by specifying that item in the model data object

### DIFF
--- a/patches/minecraft/net/minecraft/client/renderer/ItemRenderer.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/ItemRenderer.java.patch
@@ -1,14 +1,6 @@
 --- a/net/minecraft/client/renderer/ItemRenderer.java
 +++ b/net/minecraft/client/renderer/ItemRenderer.java
-@@ -47,6 +47,7 @@
- import net.minecraft.world.World;
- import net.minecraftforge.api.distmarker.Dist;
- import net.minecraftforge.api.distmarker.OnlyIn;
-+import net.minecraftforge.client.model.data.IModelData;
- 
- @OnlyIn(Dist.CLIENT)
- public class ItemRenderer implements IResourceManagerReloadListener {
-@@ -59,7 +60,7 @@
+@@ -59,7 +59,7 @@
  
     public ItemRenderer(TextureManager p_i46552_1_, ModelManager p_i46552_2_, ItemColors p_i46552_3_) {
        this.field_175057_n = p_i46552_1_;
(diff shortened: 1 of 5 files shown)


patches/minecraft/net/minecraft/client/renderer/ItemRenderer.java.patch
```diff
@@ -1,6 +1,14 @@
 --- a/net/minecraft/client/renderer/ItemRenderer.java
 +++ b/net/minecraft/client/renderer/ItemRenderer.java
-@@ -59,7 +59,7 @@
+@@ -47,6 +47,7 @@
+ import net.minecraft.world.World;
+ import net.minecraftforge.api.distmarker.Dist;
+ import net.minecraftforge.api.distmarker.OnlyIn;
++import net.minecraftforge.client.model.data.IModelData;
```
TeamSpen210: In patches, you don't want to add imports. Instead use the fully-qualified names.

CJMinecraft01: Oh okay. Thank you for letting me know. Haven't messed with patches before so it's really appreciated. I'll try to get a fix done today, thanks

KnightMiner:
We currently can swap out item models based on NBT using vanilla ItemOverrideList. You can override it to be more dynamic than JSON models allow. What benefits does this approach provide over that?

I can say the benefit of ItemOverrideList is we do not have to create a new data object every time it is rendered. In a TE we can cache based on the TE instance, but in an item you have no cache location in a stack. Plus, this requires you to override a method in the stack to provide data, while the override list works on any item (even if you wanted to add a custom model to vanilla based on a vanilla item's NBT)

CJMinecraft01:
Those are two very valid points you have made. The main benefit I see with using this over the ItemOverrideList is the support for any data type. I may be incorrect about this but as far as I can see the override list only supports floats? The idea here is to enable any data type to be used which increases the flexibility further.

Regarding caching, that is very true and I'm not 100% certain on the performance impacts of this and if there is a way to cache it. Perhaps the item singleton can have a cache? Or perhaps a stack can contain a cache instead. I'm not too sure on the implications of this will be but I appreciate your points

KnightMiner:
JSON item override list only supports floats, which is why I specified you can override it to be more dynamic than is supported by JSON. The actual method it uses to get model overrides accepts simply an item stack as the parameter. See [the DynamicBucketModel](https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/src/main/java/net/minecraftforge/client/model/DynamicBucketModel.java#L245) for an example of a NBT based override.

gigaherz:
I'm going to close this. As stated, you can already do this by extending ItemOverrideList, and overriding its getOverrideModel method. The IModelData system is a workaround for being able to provide state from the TileEntity, and from world neighbours, neither of which have any vanilla means to be done.

PR #7289, opened 2020-08-30
Add support for custom tag types

pupnewfster:
Loosely based on/a continuation of #7252

This PR adds support for registries to be able to define that they support custom tag types by adding a `tagFolder` method to the registry builder, as well as, add support to allow data generators to be made for the custom tag types. Additionally this PR is marking the biome registry as a registry that supports tags, and it should be very simple to add support for more tag types.

I tested the following environments to ensure that tags get properly synced that that vanilla compat did not get accidentally broken:
- Forge Singleplayer
- Forge Client to Forge Server
- Forge Client (with custom tag support) to Forge Server (from before this PR)
- Forge Client (from before this PR) to Forge Server (with custom tag support)
- Forge Client to Vanilla Server
- Vanilla Client to Forge Server


src/main/java/net/minecraftforge/fml/ModLoader.java
```diff
@@ -198,6 +198,8 @@ public void gatherAndInitializeMods(final ModWorkManager.DrivenExecutor syncExec
         CapabilityManager.INSTANCE.injectCapabilities(modList.getAllScanData());
         statusConsumer.ifPresent(c->c.accept("Populating registries"));
         dispatchAndHandleError(ModLoadingStage.LOAD_REGISTRIES, syncExecutor, parallelExecutor, periodicTask);
+        statusConsumer.ifPresent(c->c.accept("Adding custom tag types"));
+        GameData.setCustomTagTypesFromRegistries();
```
cpw, line 202: Since this is not trying to call mod supplied code, but is rather, just parsing out metadata from the registries into structure, this is the best solution.

halvors:
+1 @cpw seems you're ok with this PR? Merge please? xD

TheCurle:
CPW was futhering a conversation that was being had in internal channels.
Do not take that as sign of approval, and do not nag for PRs to be merged.
There is a process we need to go through.

Cyborgmas (approved):
Looking good!
Small commnent: pretty sure it would be a good thing to have some `forge:` biome tags, so should this PR add them as well directly ? Or should it be adressed in another separate one, when this gets merged?

halvors:
@TheCurle Sorry, it was not my intension to nag for the PR to be merged, just check what the status of this was.

PR #7294, opened 2020-09-01
[1.16] Optionally add milk as a fluid to the vanilla bucket

### LINKED ISSUE #6769: [1.15.2][1.16] Milk Bucket Fluid Wrapper forces the vanilla milk bucket to return no fluid

KnightMiner:
This pull request supersedes #6840, in order to fix #6769. After discussion on Discord, it was determined the best solution for mods to add milk as a fluid was for Forge to register the fluid, while mods just set a Forge flag to enable it, similar to the universal bucket back in 1.12

To enable Forge adding milk, you simply need to call `ForgeMod.enableMilkFluid()` during mod construction. We can move this function to a different location if there is a better one, but this location kept the logic behind the flag clean.

A few details on implementation:
* It was suggested on Discord that we use `minecraft:milk` and alike rather than the `forge` namespace for the fluids.
* A tag `forge:milk` was added containing the still and flowing milk like the vanilla water and lava tags. I am not sure if we want to add the tag to the `minecraft` namespace instead for consistency with the fluid's namespace, but I know some mods are using `forge:milk` for compat right now so it probably is better to stick with that.
* No fluid block is included to prevent the need for Forge to change the behavior of the vanilla bucket. I could add one that is simply unused by default if preferred.
* ~~I had to add a listener to the texture stitch event to ensure fluid textures are registered (KingLemming can give more details on why that is required). I am not sure if there is a better place than a client event listener subclass of `ForgeMod`, but I can say it does not cause issues when launching a server there.~~
* I did a bit of cleanup regarding milk in the bucket wrapper logic as the checks for milk are redundant to the fallback. The water and lava checks there really are redundant as well, but as they are just field lookups instead of object creation and are unrelated to this PR, I left them as is.

Included in this PR is `ForgeMilkTest` that will enable the Forge milk fluid and cause a milk bucket to display the name of the contained fluid on right click. You can enable the test mod using the boolean `ENABLE` flag.

Finally, credits to Rorax for the textures for the milk fluid. Used for this PR with permission.


KnightMiner:
Updated to resolve conflicts with be58851

KnightMiner:
Updated to resolve conflicts with c75eb9e

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -229,4 +278,18 @@ public void registerRecipeSerialziers(RegistryEvent.Register<IRecipeSerializer<?
...
+        @SubscribeEvent
+        public static void registerFluidTextures(TextureStitchEvent.Pre event)
+        {
+            if (enableMilkFluid && PlayerContainer.LOCATION_BLOCKS_TEXTURE.equals(event.getMap().getTextureLocation()))
+            {
+                event.addSprite(STILL_MILK_TEXTURE);
+                event.addSprite(FLOWING_MILK_TEXTURE);
+            }
+        }
+    }
```
pupnewfster: Why is this needed? `FluidModel#getTextures` calls `ForgeHooksClient#getFluidMaterials` which gets the textures from the FluidAttributes.

KnightMiner: Is that new for 1.13+? (since the addition of vanilla fluids probably) I recall a discussion with KingLemming that this was required, and did not hear of it being resolved since then.

pupnewfster: The fluid system got rewritten in 1.14+ (1.13 didn't have one at all), and if memory serves giga made it so it auto adds the textures defined in FluidAttributes to the correct atlas so that modders don't have to have a bunch of boiler plate doing so.

KnightMiner: Removed the event listener

Rosethorns:
I give my permission for Forge and all associated to use the milk fluid textures I created as linked.

KnightMiner:
Updated for current Forge mappings on the main repo, and resolved conflicts with `ForgeMod` again.

KingLemming:
This seems fine to me - were we waiting on anything else?

LexManos:
No, It's fluids so you and cpw are the only one that touch it. If it doesn't break things I don't care.

cpw:
Why is this PR a requirement?

Buuz135:
Minecraft's milk bucket doesn't handle the fluid properly without a registry replacement and multiple modders might need milk as a fluid so this way we fix unnecesary fluid conflicts and registry replacements.

cpw:
I approve and would have merged but the language changes need a local staging I cannot provide right now.

PR #7296, opened 2020-09-02
[1.16.2] Fix vanilla causing modded teleporters to crash

mrp-v2:
Vanilla gets a `PortalInfo` outside of the vanilla teleport lambda, which causes a crash if `Entity#setPortal` has not been called, breaking modded teleporters. ~~This patch hides the method call behind a boolean checking if the teleporter is an instance of the vanilla teleporter.~~ This patch re-adds the `ITeleporter#isVanilla` method and uses it for all necessary checks. This means the `instanceof` checks can also be replaced.

~~Uses a boolean variable so that the same `instanceof` check is not performed multiple times.~~

Tested with a mod I'm currently developing (not included).


Tamaized:
Instead of completely disregarding PortalInfo entirely I feel ITeleporter should be reworked to include a PortalInfo context.
Right now this just looks like an ugly workaround to maintain an old system instead of following vanilla with their changes.

mrp-v2:
The point of this PR is because the vanilla method for getting the portal info requires the entity to have been in a nether portal at some point, which isn’t the case with moded teleporters, which causes the game to crash because of a NPE.

ichttt:
In 1.12.2, we had this code in `ITeleporter`: https://github.com/MinecraftForge/MinecraftForge/blob/87a63bc5e08f7f1e7085fc62c7800a4071c94291/src/main/java/net/minecraftforge/common/util/ITeleporter.java#L56
I would suggest we re-add this method and use it, makes the code a little bit cleaner and adds a little more flexibility

McJty:
Note that the current fix is invalid and will cause guaranteed NPE's. Consider the current code in ServerPlayerEntity.changeDimension:

```
         PortalInfo portalinfo = teleporter.isVanilla() ? this.func_241829_a(server) : null;
         if (portalinfo != null || !teleporter.isVanilla()) {
            Entity e = teleporter.placeEntity(this, serverworld, server, this.rotationYaw, spawnPortal -> {//Forge: Start vanilla logic
            serverworld.getProfiler().startSection("moving");
            if (registrykey == World.OVERWORLD && server.getDimensionKey() == World.THE_NETHER) {
               this.enteredNetherPosition = this.getPositionVec();
            } else if (spawnPortal && server.getDimensionKey() == World.THE_END) {
               this.func_242110_a(server, new BlockPos(portalinfo.pos));
            }

            serverworld.getProfiler().endSection();
            serverworld.getProfiler().startSection("placing");
            this.setWorld(server);
            server.addDuringPortalTeleport(this);
            this.setRotation(portalinfo.rotationYaw, portalinfo.rotationPitch);

```

If the teleporter is not vanilla then portalinfo will be null. The if test will succeed because the teleporter is not vanilla. However, a bit later (last line in the code I pasted) portalInfo is actually referenced causing a NPE.

mrp-v2:
Here is the code from my **working** teleporter class using the current fix (from this PR):
https://gist.github.com/mrp-v2/0abb6c02facdd20e9287c6379acad033

Yes I realize this isn't the way other people use the `ITeleporter` interface. I think it would be great if #7317 got merged, as it would greatly decrease the amount of code in the `ITeleporter` linked above.

PR #7299, opened 2020-09-03
Bump pack version to 6

sciwhiz12:
1.16 changed the blockstates for walls, meaning that resource packs for 1.15 are not compatible with 1.16. However, 1.16 did not bump the version number from `5` to `6`, so 1.15 resource packs still loaded without warnings in 1.16.

An [issue was filed on Mojang's bug tracker][issue-197275], and a fix was released with 1.16.2: bumping the pack version from `5` to `6`. 
However, Forge did not update it's `pack.mcmeta` nor the internal constant used for resource pack delegation. This was eventually [reported on the Discord][discord_report].

This PR simply updates the `pack.mcmeta` and that internal constant to `6`.

_Note: I did think of maybe using `SharedConstants` directly to bypass the need of a hardcoded constant, but [as said by tterag1908 on Discord][tterag] (to which I concur), that may cause unforseen side-effects, and the resources are manually updated when needed anyway._

[issue-197275]: https://bugs.mojang.com/browse/MC-197275
[discord_report]: https://discordapp.com/channels/313125603924639766/437001959950778368/751103750180962367
[tterag]: https://discordapp.com/channels/313125603924639766/437001959950778368/751108702190305362

### DIFF
--- a/src/main/java/net/minecraftforge/fml/client/ClientModLoader.java
+++ b/src/main/java/net/minecraftforge/fml/client/ClientModLoader.java
@@ -260,7 +260,7 @@ public class ClientModLoader
             }
         }
         final ResourcePackInfo packInfo = ResourcePackInfo.createResourcePack("mod_resources", true, () -> new DelegatingResourcePack("mod_resources", "Mod Resources",
-                new PackMetadataSection(new TranslationTextComponent("fml.resources.modresources", hiddenPacks.size()), 5),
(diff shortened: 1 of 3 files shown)


pupnewfster:
It may make sense to update the test module's pack.mcmeta while you are at it: https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/src/test/resources/pack.mcmeta (Not sure if any of the test mods add some form of wall that would need to be updated as well)

sciwhiz12:
There doesn't seem to be any test mods adding custom walls, so just a pack version bump should be fine.

PR #7300, opened 2020-09-03
Fixed crash with zip paths

Cyborgmas:
path.toFile() throws for a ZipPath before the ResourcePackFileNotFoundException is thrown which is just ignored silently afterward

### DIFF
--- a/src/main/java/net/minecraftforge/fml/packs/ModFileResourcePack.java
+++ b/src/main/java/net/minecraftforge/fml/packs/ModFileResourcePack.java
@@ -27,6 +27,7 @@ import net.minecraft.util.ResourceLocation;
 import net.minecraftforge.fml.loading.moddiscovery.ModFile;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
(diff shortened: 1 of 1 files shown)


src/main/java/net/minecraftforge/fml/packs/ModFileResourcePack.java
```diff
@@ -66,7 +67,7 @@ protected InputStream getInputStream(String name) throws IOException
     {
         final Path path = modFile.getLocator().findPath(modFile, name);
         if(!Files.exists(path))
-            throw new ResourcePackFileNotFoundException(path.toFile(), name);
+            throw new FileNotFoundException(String.format("'%s' in ResourcePack '%s'", path, name));
```
ichttt: If I read vanilla code correctly, the path should not point to the zip entry, but to the file of the pack (which is the mod file in this case). So you can do `modFile.getFilePath().toFile()` which should be fine as the mod jar always exists, and use the vanilla exception

src/main/java/net/minecraftforge/fml/packs/ModFileResourcePack.java
```diff
@@ -27,6 +27,7 @@
 import net.minecraftforge.fml.loading.moddiscovery.ModFile;
 
 import java.io.File;
+import java.io.FileNotFoundException;
```
ToMe25: This import isn't needed anymore since the used exception is `ResourcePackFileNotFoundException` again, right?

PR #7301, opened 2020-09-04
Do not try to get the bounding box of an empty voxel shape

malte0811:
` IForgeTileEntity#getRenderBoundingBox` currently calls `getBoundingBox` on the collision shape of the relevant block. If the block has an empty collision shape this will throw an `UnsupportedOperationException`. This doesn't crash because the call is wrapped in a try-catch related to bukkit issues, but still isn't great and, at least for me, wastes 0.5% of the render thread's time.

### DIFF
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeTileEntity.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeTileEntity.java
@@ -28,6 +28,7 @@ import net.minecraft.nbt.CompoundNBT;
 import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.math.AxisAlignedBB;
 import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.shapes.VoxelShape;
 import net.minecraft.world.World;
 import net.minecraftforge.api.distmarker.Dist;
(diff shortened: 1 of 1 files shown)


ichttt:
The only problem I see with this code (though you didn't introduce this) is that the default bounding box start at max negative and ends at max positive. This bounding box is then passed into `ClippingHelper.isBoundingBoxInFrustum`, which adds/subtracts from the max number, depending on the camera. It probably works due to doubles having positive/negative infinity, but I would like a design where `getRenderBoundingBox` is nullable and null means "always render" better, so we can skip a few cpu cycles on null cases and be certain that the code is correct

malte0811:
Do you want me to change that in this PR or was that just a general "should be done at some point"? It seems a bit outside the scope of this PR, but I can include it if you want (and assuming making it nullable doesn't turn into a giant mess).

LexManos:
Honestly, this change should be to crash the game if modders provide invalid bounding boxes.
The fact that it's hiding this exception is not intended. Its a dirty hack around Bukkit servers sending invalid tile entitites.
But, I don't think there is a sane way to crash when modders screw up, but also not crash when users connect to bad servers.

PR #7308, opened 2020-09-07
Fix crash when dumping crash report for an exception that has a null cause

### LINKED ISSUE #7306: [1.16.2] Mod Loading causes a crash if a dependency is not loaded, instead of showing the error screen.
**Minecraft Version:** 1.16.2

**Forge Version:** 33.0.37

**Logs:** https://gist.github.com/Shadows-of-Fire/11907ea010e7127851222abecba5469c

**Steps to Reproduce:**
 1. Install any mod that has a required dependency without that dependency.
 2. Launch
 3. The game will crash.

**Description of issue:**  This appears to be caused by some change that was made to how mods are loaded in 1.16, as this will not happen in 1.15.  Any dependency error causes a crash (and any startup error might?  Not sure just yet).  This does not happen when using 1.16.2-33.0.22, so some change between 22 and 37 would have caused the problem.

ichttt:
Fixes #7306
Null cause should be handeled properly now.
Also converts all possible details to `ICrashCallable` by turning them into lambdas, so the crash report code can log any exception instead of crash reporting code crashing

### DIFF
--- a/src/main/java/net/minecraftforge/fml/CrashReportExtender.java
+++ b/src/main/java/net/minecraftforge/fml/CrashReportExtender.java
@@ -84,12 +84,13 @@ public class CrashReportExtender
                 category.addDetail("Caused by "+(depth++), cause +"\n\t\t"+ Arrays.stream(cause.getStackTrace()).map(Objects::toString).collect(Collectors.joining("\n\t\t")));
                 cause = cause.getCause();
             }
-            category.applyStackTrace(cause);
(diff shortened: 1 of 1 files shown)


cpw:
What can still be null?

ichttt:
The cause throwable. You can't pass a null cause to applyStackTrace. The rest is safty in case anything else should ever throw an exception. Vanilla does it this way as well

LexManos:
Thats about what we came up with, the exception is fine, as it doesnt have any null protection. But the other lines are all optionals that should never throw.
But, doesn't really matter.

Issue #7307, opened 2020-09-07
Brewing Stand does not respect container items.


Darkhax:
**Minecraft Version:** 1.16.2

**Forge Version:** 33.0.37

**Logs:** N/A

**Steps to Reproduce:**
 1. Create a recipe with a container item as the reagent ingredient.
 2. Craft the recipe in the game.
 3. The item is not replaced with it's containing item.

**Description of issue:**
When using a container item such as a lava bucket as the reagent (top slot) of a brewing recipe, the item will be completely consumed rather than being replaced with the container item. In other words the bucket is consumed along with the lava.

Here is a test recipe you can use to verify this issue.

```java
BrewingRecipeRegistry.addRecipe(Ingredient.fromItems(Items.ICE), Ingredient.fromItems(Items.LAVA_BUCKET), new ItemStack(Items.OBSIDIAN));
````

PR #7314, opened 2020-09-09
Fixed container item being consumed in brewing stand. #7307

Darkhax:
It seems there was already code to support container items but decreasing the stack size before the checks would make the item empty and fail the various checks. This PR restructures how and where the stack is shrunk.


diesieben07:
This will fix #7307 (adding link here so the PR shows up in the issue as well).

Darkhax:
Thanks

PR #7317, opened 2020-09-10
[1.16.3] Extended ITeleporter Interface

SilverDavidMC:
This PR aims to expand the functionality of the ITeleporter interface and make it possible to use your own in Entity.changeDimension. The previous issues with changeDimension were related to the ITeleporter not being passed into methods that interacted with it, and instead inconsistently using the vanilla one. To fix this, a couple of methods have been modified to have a default version and a modded teleporter safe version. To accompany this, the ITeleporter interface gained methods for finding and placing a portal (or whatever applies given the mod). 

To take use of this new system, all that needs to be done is calling Entity.changeDimension(ServerWorld, ITeleporter). In the past, you would have to store, replace, and then restore the old default teleporter. That is no longer required.

Within the repository is a new test mod under `net.minecraftforge.debug.world.TeleporterTest`. To use the test mod, set a "teleporter_test:red_teleporter" or a "teleporter_test:blue_teleporter". Landing on these will cause them to teleport you to their respective "red" and "blue" dimensions. The "red" dimension is scaled the same as the nether while "blue" is 1 to 1.


patches/minecraft/net/minecraft/entity/Entity.java.patch
```diff
@@ -262,7 +261,29 @@
...
+    }
+ 
+-   @Nullable
++   @Nullable @Deprecated // Use ITeleporter safe version
+    protected PortalInfo func_241829_a(ServerWorld p_241829_1_) {
++      return this.func_241829_a(p_241829_1_, p_241829_1_.func_85176_s());
++   }
++
++   @Nullable
++   protected PortalInfo func_241829_a(ServerWorld p_241829_1_, net.minecraftforge.common.util.ITeleporter teleporter) {
```
pupnewfster: This method should have a proper name and be added to the `forge.exc` file so that the `ServerWorld` param can be properly remapped.

SilverDavidMC: Taken care of. The new method has been named `getPortalInfo` and exists within the `forge.exc` for it's `ServerWorld` param.

mrp-v2:
>This PR aims to expand the functionality of the ITeleporter interface and make it possible to use your own in Entity.changeDimension.

Custom teleporters can already be used just fine. I have a working teleporter implementation in 1.16.3 where I call `Entity#changeDimension` and pass the destination world and my `ITeleporter` implementation and it works perfectly.

Tamaized:
The problem with the PR that got merged is it's just a hack that works **around** vanilla instead of **with** vanilla like this PR does. I'd much rather be able to provide a PortalInfo and allow vanilla code to handle its information instead of completely ignoring vanilla code and doing it all myself.

This new PR does exactly what I want and cleanly.

mrp-v2:
The intent of my comment was not to say that this PR is unnecessary, I meant that the description makes it sound like all teleporters are broken, when in fact you can still get them to work, it is just a lot of work. I'm all for making the `ITeleporter` interface more user-friendly.

~~One problem with this PR however is that modded teleporters cannot control certain teleports involving vanilla dimensions, I have fixed this in SilverDavidMC#1 by adding more ITeleporter#isVanilla() checks~~. Other than that everything looks great.

mrp-v2:
One problem with this PR however is that modded teleporters cannot control certain teleports involving vanilla dimensions, I have fixed this in SilverDavidMC/MinecraftForge#1 by adding more `ITeleporter#isVanilla()` checks. Other than that everything looks great.

src/main/java/net/minecraftforge/common/util/ITeleporter.java
```diff
@@ -58,10 +75,68 @@
...
+    // States if this teleporter is the vanilla instance
     default boolean isVanilla()
     {
         return getClass() == Teleporter.class;
     }
+
+    // Replicated from the vanilla code for interface implementation. Return an empty optional if no portal was found.
+    default Optional<TeleportationRepositioner.Result> findPortal(ServerWorld fromWorld, ServerWorld toWorld, Entity entity)
+    {
+        PointOfInterestManager pointofinterestmanager = toWorld.getPointOfInterestManager();
```
ichttt: Copying vanilla code should always be avoided if possible. Please check if you can't replace this with a call to the original method if it is a clean copy or patch and call the original method if the change is small enough.

src/main/java/net/minecraftforge/common/util/ITeleporter.java
```diff
@@ -58,10 +75,68 @@
...
+    // Returns the portal info for the result. Default returns the tpResult position and zero values for everything else.
+    default PortalInfo getPortalInfo(TeleportationRepositioner.Result tpResult)
+    {
+        return new PortalInfo(new Vector3d(tpResult.field_243679_a.getX(), tpResult.field_243679_a.getY(), tpResult.field_243679_a.getZ()), Vector3d.ZERO, 0, 0);
+    }
+    
+    // Scales the given blockpos based on the worlds passed in.
+    default BlockPos getScaledPos(World fromWorld, World toWorld, BlockPos originalPos)
+    {
+    	WorldBorder worldborder = toWorld.getWorldBorder();
```
ichttt: Same here, looks like vanilla copied code, can we remove this copy somehow?

ichttt (commented):
I can't say much over the concrete implementation as I'm not familiar with this section of the code, but some general review notes.
Also please double-check your formatting, it seems you are only using two spaces in some places instead of four

src/main/java/net/minecraftforge/common/util/ITeleporter.java
mrp-v2: Looks like we can't, it is embedded in another method. See `Entity#getPortalInfo(ServerWorld, ITeleporter)`

src/main/java/net/minecraftforge/common/util/ITeleporter.java
mrp-v2: fixed in https://github.com/SilverDavidMC/MinecraftForge/pull/3

mrp-v2:
~~Another thing worth considering is that any modded teleporters that override one of the `ITeleporter`s default methods would not be able to use `super` to reference that default implementation. This problem could be resolved by making `ITeleporter` abstract, or more elegantly, by refactoring `ITeleporter` into `ITeleporter` and an `AbstractTeleporter`, as in https://github.com/SilverDavidMC/MinecraftForge/pull/2.~~

pupnewfster:
I am pretty sure that is not true and they could be referenced by doing `ITeleporter.super.method(args)`

mrp-v2:
Hmm that does work.

mrp-v2:
~~Two last thing I think could be improved: the current version of the adjusted `ITeleporter` assumes that modded teleporters will be based on `PointOfInterestType`s, and the `getScaledPos` method can be made static and moved to a helper class, since it won't need to be overriden and isn't neccesarily specific to teleporters. See https://github.com/SilverDavidMC/MinecraftForge/pull/4.~~

src/main/java/net/minecraftforge/common/util/ITeleporter.java
```diff
@@ -19,49 +19,80 @@
...
     {
-        return getClass() == Teleporter.class;
+        return false;
+    }
+
+    /**
+     * Finds a portal to teleport to and creates a {@link TeleportationRepositioner.Result} for it.
+     * Return an empty {@link Optional} if no portal was found.
+     */
+    Optional<TeleportationRepositioner.Result> findPortal(ServerWorld fromWorld, ServerWorld toWorld, Entity entity);
```
pupnewfster: This is a breaking change as any mods currently using ITeleporter will not have this method implemented if it is not defaulted.

src/main/java/net/minecraftforge/common/util/ITeleporter.java
```diff
@@ -19,49 +19,80 @@
...
+
+    /**
+     * Creates a portal if one doesn't exist and returns the {@link TeleportationRepositioner.Result Result}.
+     */
+    Optional<TeleportationRepositioner.Result> createAndGetPortal(ServerWorld fromWorld, ServerWorld toWorld,
```
pupnewfster: This is a breaking change as any mods currently using ITeleporter will not have this method implemented if it is not defaulted.

src/main/java/net/minecraftforge/common/util/ITeleporter.java
mrp-v2: `Optional.empty()` for both of these?

pupnewfster: Yes and then all the changes to the CommandSetDimension command should be able to be removed

src/main/java/net/minecraftforge/common/util/ITeleporter.java
```diff
@@ -19,49 +19,80 @@
...
     }
-    
-    //used internally to handle vanilla hardcoding
+
+    /**
+     * Is this teleporter the vanilla instance.
+     */
     default boolean isVanilla()
     {
-        return getClass() == Teleporter.class;
```
pupnewfster, line 65: Why was this changed as it adds an extra line in the patch to Teleporter.java

mrp-v2: I figured returning true/false was more performant than checking for class equality.

src/main/java/net/minecraftforge/common/util/ITeleporter.java
mrp-v2: And the 2 lines at the end of the vanilla `Teleporter`

src/main/java/net/minecraftforge/common/util/ITeleporter.java
mrp-v2: But I don't know java well enough to know if it makes a difference.

SilverDavidMC:
Went through and fixed the breaking changes, undid the isVanilla check, cleaned up some code, and it overall should be more compatible with existing teleporters.

src/main/java/net/minecraftforge/common/util/ITeleporter.java
SilverDavidMC: Fixed

patches/minecraft/net/minecraft/entity/Entity.java.patch
```diff
@@ -262,7 +261,24 @@
           }
        } else {
           return null;
-@@ -2320,9 +2357,10 @@
+@@ -2123,8 +2160,15 @@
+       this.field_70128_L = true;
+    }
+ 
+-   @Nullable
++   @Nullable @Deprecated // Use ITeleporter safe version
```
pupnewfster: If you put the Deprecated and comment on the next line or the previous line then this patch will just be an addition of one line instead of an addition and removal

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
```diff
@@ -8,7 +8,27 @@
        boolean flag = this.field_70170_p.func_82736_K().func_223586_b(GameRules.field_223609_l);
        if (flag) {
           ITextComponent itextcomponent = this.func_110142_aN().func_151521_b();
-@@ -614,13 +615,14 @@
+@@ -602,10 +603,16 @@
+       return this.field_71133_b.func_71219_W();
+    }
+ 
+-   @Nullable
++   @Nullable @Deprecated // Use ITeleporter safe version
```
pupnewfster: Same here

src/main/java/net/minecraftforge/common/util/ITeleporter.java
```diff
@@ -19,49 +19,89 @@
...
+        return this.getClass() == Teleporter.class;
+    }
+
+    /**
+     * Finds a portal to teleport to and creates a {@link TeleportationRepositioner.Result} for it. Defaults to calling vanilla Teleporter methods.
+     */
+    default Optional<TeleportationRepositioner.Result> findPortal(ServerWorld fromWorld, ServerWorld toWorld, Entity entity)
+    {
+    	Teleporter teleporter = toWorld.getDefaultTeleporter();
```
pupnewfster: Spacing is screwed up.

src/main/java/net/minecraftforge/common/util/ITeleporter.java
```diff
@@ -19,49 +19,89 @@
...
+            teleporter = (Teleporter) this;
+        return teleporter.func_242957_a(TeleporterHelper.getScaledPos(fromWorld, toWorld, new BlockPos(entity.getPositionVec())), toWorld.func_234923_W_() == World.field_234919_h_);
+    }
+
+    /**
+     * Creates a portal if one doesn't exist and returns the {@link TeleportationRepositioner.Result Result}. Defaults to calling vanilla Teleporter methods.
+     */
+    default Optional<TeleportationRepositioner.Result> createAndGetPortal(ServerWorld fromWorld, ServerWorld toWorld, Entity entity)
+    {
+    	Teleporter teleporter = toWorld.getDefaultTeleporter();
```
pupnewfster: Spacing is screwed up. Also is this changing the default of ITeleporter to spawning a portal instead of having a custom ITeleporter not have a portal? If so that is a breaking change from the logical standpoint even if not binary

patches/minecraft/net/minecraft/entity/Entity.java.patch
```diff
@@ -262,7 +261,23 @@
...
+    }
+ 
+    @Nullable
++   @Deprecated // Forge: Use ITeleporter safe version
+    protected PortalInfo func_241829_a(ServerWorld p_241829_1_) {
++      return this.getPortalInfo(p_241829_1_, p_241829_1_.func_85176_s());
++   }
++
++   @Nullable
++   protected PortalInfo getPortalInfo(ServerWorld p_241829_1_, net.minecraftforge.common.util.ITeleporter teleporter) {
```
pupnewfster: `teleporter` should probably be annotated as nullable given the line below "handles" it being null

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
```diff
@@ -8,7 +8,26 @@
...
+    @Nullable
++   @Deprecated // Forge: Use ITeleporter safe version
+    protected PortalInfo func_241829_a(ServerWorld p_241829_1_) {
+-      PortalInfo portalinfo = super.func_241829_a(p_241829_1_);
+-      if (portalinfo != null && this.field_70170_p.func_234923_W_() == World.field_234918_g_ && p_241829_1_.func_234923_W_() == World.field_234920_i_) {
++      return this.getPortalInfo(p_241829_1_, p_241829_1_.func_85176_s());
++   }
++   
++   @Nullable
++   protected PortalInfo getPortalInfo(ServerWorld p_241829_1_, net.minecraftforge.common.util.ITeleporter teleporter) {
```
pupnewfster: Nullable here as well? Unless it shouldn't be nullable anywhere given it doesn't seem null safe two lines down from here

patches/minecraft/net/minecraft/entity/Entity.java.patch
```diff
@@ -231,9 +231,8 @@
           this.func_213319_R();
           this.field_70170_p.func_217381_Z().func_76320_a("reposition");
 -         PortalInfo portalinfo = this.func_241829_a(p_241206_1_);
--         if (portalinfo == null) {
-+         PortalInfo portalinfo = teleporter.isVanilla() ? this.func_241829_a(p_241206_1_) : null;
-+         if (portalinfo == null && teleporter.isVanilla()) {
++         PortalInfo portalinfo = teleporter.isVanilla() ? this.func_241829_a(p_241206_1_) : this.getPortalInfo(p_241206_1_, teleporter);
```
pupnewfster: This can just be `PortalInfo portalinfo = this.getPortalInfo(p_241206_1_, teleporter);`

patches/minecraft/net/minecraft/entity/Entity.java.patch
```diff
@@ -245,7 +244,7 @@
                 entity.func_213317_d(portalinfo.field_222506_b);
                 p_241206_1_.func_217460_e(entity);
 -               if (p_241206_1_.func_234923_W_() == World.field_234920_i_) {
-+               if (spawnPortal && p_241206_1_.func_234923_W_() == World.field_234920_i_) {
++               if (spawnPortal && p_241206_1_.func_234923_W_() == World.field_234920_i_ && teleporter.isVanilla()) {
```
pupnewfster: What is the purpose of this teleporter is vanilla check? If a modded teleporter passes false for spawn portal then it won't spawn the end spawn platform anyways.

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
```diff
@@ -26,28 +45,28 @@
...
           playerlist.func_187243_f(this);
 -         serverworld.func_217434_e(this);
 -         this.field_70128_L = false;
 -         PortalInfo portalinfo = this.func_241829_a(p_241206_1_);
--         if (portalinfo != null) {
 +         serverworld.removeEntity(this, true); //Forge: the player entity is moved to the new world, NOT cloned. So keep the data alive with no matching invalidate call.
 +         this.revive();
-+         PortalInfo portalinfo = teleporter.isVanilla() ? this.func_241829_a(p_241206_1_) : null;
-+         if (portalinfo != null || !teleporter.isVanilla()) {
++         PortalInfo portalinfo = teleporter.isVanilla() ? this.func_241829_a(p_241206_1_) : this.getPortalInfo(p_241206_1_, teleporter);
```
pupnewfster: This can just be `PortalInfo portalinfo = this.getPortalInfo(p_241206_1_, teleporter);`

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
```diff
@@ -26,28 +45,28 @@
...
+          if (portalinfo != null) {
 +            Entity e = teleporter.placeEntity(this, serverworld, p_241206_1_, this.field_70177_z, spawnPortal -> {//Forge: Start vanilla logic
              serverworld.func_217381_Z().func_76320_a("moving");
-             if (registrykey == World.field_234918_g_ && p_241206_1_.func_234923_W_() == World.field_234919_h_) {
+-            if (registrykey == World.field_234918_g_ && p_241206_1_.func_234923_W_() == World.field_234919_h_) {
++            if (registrykey == World.field_234918_g_ && p_241206_1_.func_234923_W_() == World.field_234919_h_ && teleporter.isVanilla()) {
```
pupnewfster: I am not sure about this isVanilla check being added, as the enteredNetherPosition is used for checking if a player has travelled to the nether from what I can tell.

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
```diff
@@ -26,28 +45,28 @@
...
                 this.field_193110_cw = this.func_213303_ch();
 -            } else if (p_241206_1_.func_234923_W_() == World.field_234920_i_) {
-+            } else if (spawnPortal && p_241206_1_.func_234923_W_() == World.field_234920_i_) {
++            } else if (spawnPortal && p_241206_1_.func_234923_W_() == World.field_234920_i_ && teleporter.isVanilla()) {
```
pupnewfster: What is the purpose of this teleporter is vanilla check? If a modded teleporter passes false for spawn portal then it won't spawn the end spawn platform anyways.

src/main/java/net/minecraftforge/common/util/ITeleporter.java
```diff
@@ -19,49 +19,93 @@
...
+     * Finds a portal to teleport to and creates a {@link TeleportationRepositioner.Result} for it.
+     * Calls {@link Teleporter} methods if this is an instance of Teleporter.
+     * Defaults to the scaled entity's position.
+     */
+    default Optional<TeleportationRepositioner.Result> findPortal(ServerWorld fromWorld, ServerWorld toWorld, Entity entity)
+    {
+        if (this instanceof Teleporter)
+            return ((Teleporter) this).func_242957_a(TeleporterHelper.getScaledPos(fromWorld, toWorld, new BlockPos(entity.getPositionVec())), toWorld.func_234923_W_() == World.field_234919_h_);
+        else
+            return Optional.ofNullable(new TeleportationRepositioner.Result(TeleporterHelper.getScaledPos(fromWorld, toWorld, new BlockPos(entity.getPositionVec())), 0, 0));
```
pupnewfster: This can just be `Optional.of(` and I believe you can use `entity.func_233580_cy_()` instead of `new BlockPos(entity.getPositionVec())`;

src/main/java/net/minecraftforge/common/util/ITeleporter.java
```diff
@@ -19,49 +19,93 @@
...
+     * Creates a portal if one doesn't exist and returns the {@link TeleportationRepositioner.Result Result}.
+     * Calls {@link Teleporter} methods if this is an instance of Teleporter.
+     * Defaults to the scaled entity's position.
+     */
+    default Optional<TeleportationRepositioner.Result> createAndGetPortal(ServerWorld fromWorld, ServerWorld toWorld, Entity entity)
+    {
+        if (this instanceof Teleporter)
+            return ((Teleporter) this).func_242956_a(TeleporterHelper.getScaledPos(fromWorld, toWorld, new BlockPos(entity.getPositionVec())), entity.getHorizontalFacing().getAxis());
+        else
+            return Optional.ofNullable(new TeleportationRepositioner.Result(TeleporterHelper.getScaledPos(fromWorld, toWorld, new BlockPos(entity.getPositionVec())), 0, 0));
```
pupnewfster: This can just be `Optional.of(` and I believe you can use `entity.func_233580_cy_()` instead of `new BlockPos(entity.getPositionVec())`;

src/main/java/net/minecraftforge/server/command/CommandSetDimension.java
```diff
@@ -75,6 +76,20 @@ public Entity placeEntity(Entity entity, ServerWorld currentWorld, ServerWorld d
...
+            {
+                return Optional.empty();
+            }
+
+            @Override
+            public Optional<TeleportationRepositioner.Result> createAndGetPortal(ServerWorld fromWorld,
+                    ServerWorld toWorld, Entity entity)
+            {
+                return Optional.empty();
+            }
```
pupnewfster: I believe this isn't necessary because the methods are defaulted in `ITeleporter` and because of not allowing breaking changes the functionality should be how it is currently anyways. Unless you would rather be overriding one/both of these methods to return the result for the correct destination position and remove the setPositionAndUpdate call from the placeEntity method above.

src/test/resources/META-INF/mods.toml
```diff
@@ -85,4 +85,6 @@ license="LGPL v2.1"
 [[mods]]
     modId="custom_tag_types_test"
 [[mods]]
-    modId="biome_loading_event_test"
\ No newline at end of file
+    modId="biome_loading_event_test"
+[[mods]]
+	modId="teleporter_test"
```
pupnewfster: This needs to be rebased anyways given there is a merge conflict, but also this should use four spaces like the rest of the file instead of using a tab.

src/test/java/net/minecraftforge/debug/world/TeleporterTest.java
```diff
@@ -0,0 +1,190 @@
...
+            int scale = TeleporterHelper.getPortalSearchRadius(fromWorld, toWorld);
+            BlockPos scaledPos = TeleporterHelper.getScaledPos(fromWorld, toWorld, new BlockPos(entity.getPositionVec()));
+            poiManager.ensureLoadedAndValid(toWorld, scaledPos, scale);
+            Optional<PointOfInterest> optional = poiManager.getInSquare((poiType) -> {
+               return poiType == this.poi;
+            }, scaledPos, scale, PointOfInterestManager.Status.ANY).sorted(Comparator.<PointOfInterest>comparingDouble((poi) -> {
+               return poi.getPos().distanceSq(scaledPos);
+            }).thenComparingInt((poi) -> {
+               return poi.getPos().getY();
+            })).findFirst();
```
pupnewfster: There is no reason to be using extended lambdas (with the braces) here it can just be `poiType -> poiType == this.poi` for example. Also instead of using `sorted` and `findFirst()` you can just use `min` where the current `sorted` call is and remove the `findFirst()` call

src/test/java/net/minecraftforge/debug/world/TeleporterTest.java
```diff
@@ -0,0 +1,190 @@
...
+            
+            return optional.map((poi) -> {
+               BlockPos poiPos = poi.getPos();
+               toWorld.getChunkProvider().registerTicket(TicketType.PORTAL, new ChunkPos(poiPos), 3, poiPos);
+               BlockState blockstate = toWorld.getBlockState(poiPos);
+               return TeleportationRepositioner.func_243676_a(poiPos, entity.getHorizontalFacing().getAxis(), 21, Direction.Axis.Y, 21, (pos) -> {
+                  return toWorld.getBlockState(pos) == blockstate;
+               });
```
pupnewfster: No need for the extended lambda here either

pupnewfster (commented):
Finally had a chance to review this in more depth and overall I think it is designed pretty well to solve the issue that custom ITeleporter's don't work currently due to bypassing the portal info check. Just a lot of minor things that need to be cleaned up before I can assign it to lex.

patches/minecraft/net/minecraft/entity/Entity.java.patch
SilverDavidMC: Going to opt to not mark it as nullable since it still technically shouldn't be null and it isn't really "handled" properly if it is.

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
SilverDavidMC: Same reason as Entity.java.patch. It technically could be null and I could add extra checks in this case, but teleporter really should never be null in the first place since that would cause many logic issues in game.

patches/minecraft/net/minecraft/entity/Entity.java.patch
pupnewfster: It might be worth removing the != null check then to make it a bit more readable and obvious that it follows the parameters are nonnull by default and that it wasn't just a missed nullable annotation

patches/minecraft/net/minecraft/entity/Entity.java.patch
SilverDavidMC: Not sure actually. Probably got left here by mistake while working on the logic.

src/test/java/net/minecraftforge/debug/world/TeleporterTest.java
SilverDavidMC: Didn't realize `min` was a thing. Nifty!

patches/minecraft/net/minecraft/entity/Entity.java.patch
SilverDavidMC: Yeah that's a valid point. I wasn't sure if I should leave it there or not.

src/main/java/net/minecraftforge/server/command/CommandSetDimension.java
```diff
@@ -25,17 +25,18 @@
...
 import net.minecraft.world.server.ServerWorld;
 import net.minecraftforge.common.util.ITeleporter;
-
 import com.mojang.brigadier.builder.ArgumentBuilder;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
 import com.mojang.brigadier.exceptions.DynamicCommandExceptionType;
 import com.mojang.brigadier.exceptions.SimpleCommandExceptionType;
 
 import java.util.Collection;
+import java.util.Optional;
```
pupnewfster: These import changes aren't needed

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
pupnewfster: My guess is this comment of mine got missed because of how github truncates things, but still should be addressed.

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
pupnewfster: My guess is this comment of mine got missed because of how github truncates things, but still should be addressed.

pupnewfster:
@SilverDavidMC This is looking relatively good, (there are still two comments you haven't addressed about changes in `ServerPlayerEntity`). One thing that you may find helpful for making sure you don't accidentally miss comments is once you address something and "resolve" it you can hit a button on github to "resolve conversation" to make that specific back and forth of comments auto fold and be shown as "resolved". (That also reduces how much scrolling people end up needing to go through)

patches/minecraft/net/minecraft/entity/Entity.java.patch
mrp-v2: I think it is required for teleporters to be able to control their own teleports to the end, instead of using the vanilla obsidian platform.

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
mrp-v2: The purpose is if modded teleporters want to spawn their own portal in the end and not the obsidian platform.

patches/minecraft/net/minecraft/entity/Entity.java.patch
pupnewfster: Which is the reason the spawnPortal boolean is already being added to this line, changing it to also require the teleporter being vanilla makes it *impossible* for mods that may want to use vanilla's obsidian platform.

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
pupnewfster: As I said on the other comment, that  is the reason the spawnPortal boolean is already being added to this line, changing it to also require the teleporter being vanilla makes it *impossible* for mods that may want to use vanilla's obsidian platform.

patches/minecraft/net/minecraft/entity/Entity.java.patch
mrp-v2: so for modded teleporters to make their own portals in the end, they pass false and then make their own portal?

pupnewfster: Yes:
- Modded teleporter that wants to use vanilla "portal": pass true
- Modded teleporter that wants to use no portal: pass false
- Modded teleporter that wants to use a custom portal: pass false, and make their own
- Vanilla teleporter: passes true by default
There is no need to have a check to see if it is the vanilla teleporter as all that does is make the first case not possible, and makes the spawnPortal parameter pointless to begin with.

SilverDavidMC:
Oh shoot I didn't realize that was a thing. Thanks! I'll go through and do that

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
SilverDavidMC: Removing this isVanilla check since it's used for advancements and the check can break that

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
SilverDavidMC: Removing this isVanilla check since you can control it with the spawnPortal boolean

LexManos:
Was gunna let this fester until you guys decided to figure it out.
However, you've gone down a rather dumb rabbit hole.
The ONLY issue is that we have stupid mechanics which force the portal to null if its not a vanilla teleporter. Which means they can never use the normal movement mechanics as they NPE.

All this needs is
```
    // Return null to prevent teleportation
    @Nullable
    default PortalInfo getPortalInfo(Entity entity, ServerWorld destWorld, Function<ServerWorld, PortalInfo> defaultSupplier) {
        return defaultSupplier.apply(destWorld);
    }
```
Which reduce the patches to this:
``` PortalInfo portalinfo = teleporter.getPortalInfo(this, p_241206_1_, this::func_241829_a);```
Keep it simple folks.

**Comment by SilverDavidMC** [comment]
You know what. That's actually a really good way to do it and a lot more simple than what I was thinking. I can't do anything with it at the moment, but I'll for sure look into changing things around later today.

File: patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch [hunk 25]
```diff
@@ -19,7 +19,7 @@
        ServerWorld serverworld = this.func_71121_q();
        RegistryKey<World> registrykey = serverworld.func_234923_W_();
 -      if (registrykey == World.field_234920_i_ && p_241206_1_.func_234923_W_() == World.field_234918_g_) {
-+      if (registrykey == World.field_234920_i_ && p_241206_1_.func_234923_W_() == World.field_234918_g_ && teleporter.isVanilla()) { //Forge: Fix non-vanilla teleporters triggering end credits
++      if (registrykey == World.field_234920_i_ && p_241206_1_.func_234923_W_() == World.field_234918_g_) { //Forge: Fix non-vanilla teleporters triggering end credits
```
KnightMiner: Probably don't need the comment here anymore, since you are no longer changing code on this line

SilverDavidMC: Yeah just saw that. Still going through things

pupnewfster: Are you sure this line doesn't need to be modified still? Try using `forge setdimension` to go from the end to the overworld. That is why I initially had to make a PR to add the isVanilla check to this line (Though I believe it might have been an inlined class check back then)

pupnewfster: Are you sure this line doesn't need to be modified still? Try using forge setdimension to go from the end to the overworld. That is why I initially had to make a PR to add the isVanilla check to this line (Though I believe it might have been an inlined class check back then)

src/main/java/net/minecraftforge/common/util/TeleporterHelper.java
```diff
@@ -0,0 +1,36 @@
+package net.minecraftforge.common.util;
+
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.MathHelper;
+import net.minecraft.world.DimensionType;
+import net.minecraft.world.World;
+import net.minecraft.world.border.WorldBorder;
+
+public class TeleporterHelper
```
pupnewfster: There is no reason for this stuff to exist anymore, at least not as public forge API, just put the methods in your test mod if they are still needed by it.

SilverDavidMC: Fair enough. I wasn't sure if I should've kept it or not

pupnewfster: Ya remove it, it serves no purpose

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
SilverDavidMC: Keeping the isVanilla check here so that it doesn't show end credits

src/main/java/net/minecraftforge/server/command/CommandSetDimension.java
```diff
@@ -75,6 +73,12 @@ public Entity placeEntity(Entity entity, ServerWorld currentWorld, ServerWorld d
                 repositionedEntity.setPositionAndUpdate(pos.getX(), pos.getY(), pos.getZ());
                 return repositionedEntity;
             }
+            
+            @Override
+            public PortalInfo getPortalInfo(Entity entity, ServerWorld destWorld, Function<ServerWorld, PortalInfo> defaultPortalInfo)
+            {
+                return new PortalInfo(entity.getPositionVec(), Vector3d.ZERO, entity.rotationYaw, entity.rotationPitch);
+            }
```
pupnewfster: I am guessing this is required to make it work properly and otherwise it just doesn't teleport (instead of how it currently just silently null pointers)?

SilverDavidMC: It doesn't throw a crash, but the game will lock up since the vanilla method would be used here by default. I'm not exactly sure where it locks up, but it does, so I needed to add this. If your teleporter was doing everything in the ITeleporter.placeEntity method, it could get around this, but that's a rather unwieldy way of doing things.

pupnewfster: I figured this was the case I just wanted to confirm there wasn't being added due to some breaking change being introduced.

src/main/java/net/minecraftforge/common/util/ITeleporter.java
```diff
@@ -29,39 +32,62 @@
...
+     * @param entity The entity teleporting before the teleport
+     * @param destWorld The world the entity is teleporting to
+     * @param defaultPortalInfo A reference to the vanilla method for getting portal info. You should implement your own logic instead of using this
+     * 
+     * @return The location, rotation, and motion of the entity in the destWorld after the teleport
+     */
+    @Nullable
+    default PortalInfo getPortalInfo(Entity entity, ServerWorld destWorld, Function<ServerWorld, PortalInfo> defaultPortalInfo)
+    {
+        return defaultPortalInfo.apply(destWorld);
```
pupnewfster: Discussed this with lex some and this should instead return:
```java
return isVanilla() ? defaultPortalInfo.apply(destWorld) : new PortalInfo(entity.getPositionVec(), Vector3d.ZERO, entity.rotationYaw, entity.rotationPitch);
```
so that current mods implementing ITeleporter using work arounds continue to work and don't get broken by this change.

src/main/java/net/minecraftforge/server/command/CommandSetDimension.java
```diff
@@ -41,7 +42,7 @@
 public class CommandSetDimension
 {
     private static final SimpleCommandExceptionType NO_ENTITIES = new SimpleCommandExceptionType(new TranslationTextComponent("commands.forge.setdim.invalid.entity"));
-    private static final DynamicCommandExceptionType INVALID_DIMENSION = new DynamicCommandExceptionType(dim -> new TranslationTextComponent("commands.forge.setdim.invalid.dim", dim));
```
pupnewfster: Don't remove this, it is here to remember to reimplement it at some point. (Also with the requested change to the default implementation of `getPortalInfo` you should be able to revert all changes being made to this class.)

pupnewfster (commented):
There are a couple more small changes to make this not break any existing mods causing their code to just not teleport due to not running. Additionally, also you should remove the test mod (and all the files added by it), as with the change of making this just have a `getPortalInfo` method instead of a bunch of new methods all the test mod does is add clutter and the SetDimension command is a good enough "example mod" for people to use.

PR #7323, opened 2020-09-12
WIP: Fix forge light pipeline applying block offsets twice

### LINKED ISSUE #7321: Experimental Lighting Pipeline causes flowers to get excessive offset
**Minecraft Version:** {Minecraft version}
1.16.2
**Forge Version:** {Forge version. *Version number, not latest/rb*}
33.0.61

**Steps to Reproduce:**
 1. enable experimentalLightingPipeline in config
 2. place flowers or bamboo
 3. ...

**Description of issue:**
image

image

diesieben07:
Fixes #7321.

In 1.15 vanilla switched from applying block offsets individually to each quad in `renderQuadsSmooth`/`renderQuadsFlat` to simply doing a matrix transform in `renderModel`. Forge's pipeline still applies the offset to every quad again though, causing models to be shifted twice.

I have marked this WIP because I do not know if this is the proper fix or if instead the matrix transform needs to be disabled when the forge pipeline is enabled.

### DIFF
--- a/src/main/java/net/minecraftforge/client/model/pipeline/BlockInfo.java
+++ b/src/main/java/net/minecraftforge/client/model/pipeline/BlockInfo.java
@@ -51,8 +51,6 @@ public class BlockInfo
 
     private boolean full;
 
-    private float shx = 0, shy = 0, shz = 0;
-
     private int cachedTint = -1;
     private int cachedMultiplier = -1;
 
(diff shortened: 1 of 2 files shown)


covers1624:
Looks fine to me, shouldn't need to touch the matrix as in both pipelines the offset is applied after the quad has been lit.

PR #7327, opened 2020-09-12
Add an event mechanism for intercepting runtime biome creation..

cpw:
As discussed on stream. Uses an event mechanism to intercept the "runtime" biome creation mechanism, allowing for injection of customizations at that point. Should fire after all biome data is loaded from disk and other sources.

Should be an easy way to hook in JSON data-driven biome customization as well..


cpw:
@Cyborgmas something for you to review perhaps..

Yunus1903:
Here's the stream VOD for reference: https://www.twitch.tv/videos/739143705

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -306,4 +306,5 @@ public net.minecraft.world.gen.layer.LayerUtil func_202829_a(JLnet/minecraft/wor
 private-f net.minecraft.world.raid.Raid$WaveMember field_221284_f # VALUES
 public net.minecraft.world.server.ServerChunkProvider field_186029_c # chunkGenerator
 public net.minecraft.world.server.ServerChunkProvider field_73251_h # worldObj
-public net.minecraft.world.storage.FolderName <init>(Ljava/lang/String;)V # constructor
\ No newline at end of file
+public net.minecraft.world.storage.FolderName <init>(Ljava/lang/String;)V # constructor
+public net.minecraft.world.biome.Biome$Climate
```
pupnewfster: I am pretty sure if you call checkATs it will reorder this as biome comes before storage alphabetically

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -839,6 +845,18 @@ public static FluidAttributes createVanillaFluidAttributes(Fluid fluid)
         throw new RuntimeException("Mod fluids must override createAttributes.");
     }
 
+    public interface BiomeCallbackFunction {
```
pupnewfster, line 848: Would it make sense to add the `@FunctionalInterface` annotation to this to make it more clear that is only has the one method?

src/main/java/net/minecraftforge/event/BiomeLoadingEvent.java
```diff
@@ -0,0 +1,82 @@
+package net.minecraftforge.event;
```
pupnewfster, line 20: This file also needs its license header.

Cyborgmas:
It seems pretty solid yeah and it should indeed be easy to adapt to.

Main thing that I am concerned about is if the order in which biomes are processed should be well-defined for the scope of this PR. The event bus is not made for having a user defined order (or modpack defined order) like a datapack would.
The order is pretty important in cases where someone wants to do xyz to all tree features, it would be more effective if it runs at the end, after mods added all of their trees.
Of course this means that if 2 mods add features X and need to modify features X its impossible to satisfy both, which I would say is a fine edge case.
But, it could be just a thing that needs to be added at some point and for now have mods modify biomes somewhat non-deterministically.

A small thing to arrange would be to have the Builders be able to remove objects since that is wanted by many. Right now it would still require reflection/at.

cpw:
> It seems pretty solid yeah and it should indeed be easy to adapt to.

> Main thing that I am concerned about is if the order in which biomes are processed should be well-defined for the scope of this PR. The event bus is not made for having a user defined order (or modpack defined order) like a datapack would.

Ordering seems undefined in vanilla. We'd need to figure out a way to impose an ordering. This would be problematic, I expect. 

EventBus cares not a jot about ordering. The information about the biome to be loaded _is_ present - any mod that wanted to affect more than one biome could easily do so.

> The order is pretty important in cases where someone wants to do xyz to all tree features, it would be more effective if it runs at the end, after mods added all of their trees.

That kind of ordering can be controlled by the EventPriority mechanism of EventBus. That is quite doable.

> Of course this means that if 2 mods add features X and need to modify features X its impossible to satisfy both, which I would say is a fine edge case.
> But, it could be just a thing that needs to be added at some point and for now have mods modify biomes somewhat non-deterministically.
> 
> A small thing to arrange would be to have the Builders be able to remove objects since that is wanted by many. Right now it would still require reflection/at.

Yes. I think we'll probably look at not using the vanilla builder directly, but rather, creating a subclass and using that instead. That would remove the groody patches they both got. Probably a thing for tomorrow. That's why this isn't right into Forge BTW: trying to quickly iterate a solution sketch.

Cyborgmas:
> That kind of ordering can be controlled by the EventPriority mechanism of EventBus. That is quite doable.

Yeah, thinking about it should be possible to say in the doc like
- additions: high priority
- removal: normal priority
- modifications: low priority

It would probably cover a lot of cases if people stick to that, user/modpack defined order could come later. It is not hard to do, just has to use something else than an event, so another time.

patches/minecraft/net/minecraft/util/registry/WorldSettingsImport.java.patch
```diff
@@ -0,0 +1,10 @@
+--- a/net/minecraft/util/registry/WorldSettingsImport.java
++++ b/net/minecraft/util/registry/WorldSettingsImport.java
+@@ -231,6 +231,7 @@
+ 
+          public <E> DataResult<Pair<E, OptionalInt>> func_241879_a(DynamicOps<JsonElement> p_241879_1_, RegistryKey<? extends Registry<E>> p_241879_2_, RegistryKey<E> p_241879_3_, Decoder<E> p_241879_4_) {
+             JsonElement jsonelement = this.field_244349_a.get(p_241879_3_);
++            jsonelement.getAsJsonObject().addProperty("forge:resourcelocation", p_241879_3_.func_240901_a_().toString());
```
marchermans: You need a NULL check here.
If the file does not exist or is malformed this jsonElement is null.

marchermans (changes requested):
Logically speaking this looks good.
One minor thingy is the fact that you need a null check when trying to mutate the jsonElement that gets parsed.

If the file is misformed in a DataPack or actually missing for what ever reason, the element will be null, and that might create some weird, hard to trace errors.

pupnewfster:
One of my use cases in Mek for this would be to add baby mob spawns based on a percentage rate of the "adult" mob spawns in a given biome. So baby skeletons spawn a lot more frequently in the soul soil nether biome because skeletons have a very high spawn rate there, than say how often they spawn in plains. To accomplish this it would be of use to have at least read access of `MobSpawnInfo.Builder.field_242567_a` so that I can read what the current values are for a given mob (if they are even there) and base my new values I would be adding via `MobSpawnInfo.Builder#func_242575_a(EntityClassification.MONSTER, myValues)` on them.

src/main/resources/META-INF/accesstransformer.cfg
cpw: it did

src/main/java/net/minecraftforge/common/ForgeHooks.java
cpw: yup, though it is 100% a no-op to the Java nowadays. It's just a dev signifier

src/main/java/net/minecraftforge/event/BiomeLoadingEvent.java
cpw: Indeed

src/main/java/net/minecraftforge/mixin/MixinBiomeGenerationSettingsAccessor.java
```diff
@@ -0,0 +1,36 @@
...
+import net.minecraft.world.gen.carver.ConfiguredCarver;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+import java.util.List;
+import java.util.Map;
+import java.util.function.Supplier;
+
+@Mixin(BiomeGenerationSettings.class)
+public interface MixinBiomeGenerationSettingsAccessor {
```
jamierocks, line 33: I recommend giving some thought about how you want to organise mixins going forward. Sponge has an indepth style guide they made early in their 1.15 update (see https://github.com/SpongePowered/SpongeCommon/wiki/Implementation-Policy:-Bridges-and-Mixins).

In this specific case, I would recommend the package be `net.minecraftforge.mixin.world.gen` and the class `BiomeGenerationSettingsAccessor`.

LexManos: We are working with mumfrey to address some ideas. Ideally these mixins would be private interfaces in BiomeGenerationSettingsBuilder classes so that it isn't exposed to modders. Currently these are just ideas. For now I think i'm gunna re-write this to just use ATs as we want to get Biomes functioning for 1.16, but keeping ideas for mixins in the future.

cpw: Yes, as much as anything this was a great proof of concept for how mixins would work in a forge context. The toolchain needs a touch more work to probably be good for all three audiences but that's stuff that's happening. It was a good exercise for me..

LexManos:
https://github.com/MinecraftForge/MinecraftForge/commit/11e3281b3fe5661bdcc3cc2b2aa6309ed3613a0f

PR #7329, opened 2020-09-15
Allows json based biomes to be loaded.

Thelonedevil:
This PR aims to address an issue caused by the biome loading event addition.

Datapack biomes would error because only the biomes loaded from the Biome Registry would have the `forge:registry_name` property injected into the biome json.
`Error loading registry data: No key forge:registry_name in MapLike[...]`
This is because there are 2 implementations of `WorldSettingsImport.IResourceAccess`, `RegistryAccess` which was used when you open the new world screen and loaded from the registry, and an anonymous class in `WorldSettingsImport.IResourceAccess` which is used for datapacks.
`RegistryAccess` had already been patched to inject `forge:registry_name` property into the json.
This patches the anonymous class to do the same.

Adds a test mod that when enabled will log the names of all biomes that go through the BiomeLoadedEvent and a sample json biome based on the vanilla desert biome under the mod namespace to demonstrate the differences between registry biomes and datapack biomes. The sample biome is not generated as it is not included in any dimension.

### DIFF
--- a/patches/minecraft/net/minecraft/util/registry/WorldSettingsImport.java.patch
+++ b/patches/minecraft/net/minecraft/util/registry/WorldSettingsImport.java.patch
@@ -1,6 +1,14 @@
 --- a/net/minecraft/util/registry/WorldSettingsImport.java
 +++ b/net/minecraft/util/registry/WorldSettingsImport.java
-@@ -231,6 +231,7 @@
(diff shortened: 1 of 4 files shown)


cpw:
Thanks for this and the test case.

marchermans (approved):

marchermans:
Works on my end.

PR #7330, opened 2020-09-15
[1.16.x] Add support for custom world generators

Won-Ton:
The BiomeGeneratorTypeScreens class defines each of the user-selectable world generator types that are available in the 'More World Options' screen (similar to the WorldType class from 1.15 and below).

The class is not extensible, so modders have no way to let users select their generator(s) when creating a world:
1. The constructor is private
2. Generator types that have an 'edit screen' are defined in an immutable map

This PR does the following:
1. Changes the constructor access to protected
2. Adds an "extension"? interface to BiomeGenTypeScreens that modders can optionally override to provide their edit screen
3. Modifies WorldOptionsScreen to make use of said interface before falling back to the immutable map for the vanilla stuff
4. Adds a test mod to demo custom generator types with & without an edit screen


Won-Ton:
Superseded by #7353

PR #7332, opened 2020-09-17
[1.16.x] Fix Recipe Providers Throwing NPE for Non-Grouped Items

ChampionAsh5357:
If an item is not in an `ItemGroup`, the data provider results in an NPE as it tries to pull the group path for the advancement location. To get around this, a null check has been added before constructing the advancement id that returns an empty string if null now. Since all the recipes use the same advancement id creator, I just generalized the id into a single forge hook and passed in the required information there.

### DIFF
--- a/patches/minecraft/net/minecraft/data/CookingRecipeBuilder.java.patch
+++ b/patches/minecraft/net/minecraft/data/CookingRecipeBuilder.java.patch
@@ -1,0 +1,11 @@
+--- a/net/minecraft/data/CookingRecipeBuilder.java
++++ b/net/minecraft/data/CookingRecipeBuilder.java
+@@ -68,7 +68,7 @@
+    public void func_218635_a(Consumer<IFinishedRecipe> p_218635_1_, ResourceLocation p_218635_2_) {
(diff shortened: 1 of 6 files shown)


marchermans:
Talked about this internally: All Items which are a result of a Recipe have an ItemGroup.
Forge is just following convention here, and as such we do not think this PR would be setting a good example.

For what is this even needed?

ChampionAsh5357:
Although all recipes have an `ItemGroup`, you can still create recipes without one assuming it is not generated. Only generated ones throw that error during verification. The convention is true as long as everything is datagened. Otherwise, it's not considered at all. It's just a path variable to the advancement recipe itself when running.

In my opinion, if forge recommends pushing everything to data generators with groups, then yes, this should not be merged and closed. Otherwise, it just fixes an oversight that comes about with having an `Item` result that's not within an `ItemGroup`.

LexManos:
> In my opinion, if forge recommends pushing everything to data generators with groups, then yes, this should not be merged and closed.

Following Vanilla's lead on this.

PR #7333, opened 2020-09-17
[1.15.x] Fix Recipe Providers Throwing NPE for Non-Grouped Items

ChampionAsh5357:
This is the 1.15.x backport of #7332. If an item is not in an ItemGroup, the data provider results in an NPE as it tries to pull the group path for the advancement location. To get around this, a null check has been added before constructing the advancement id that returns an empty string if null now. Since all the recipes use the same advancement id creator, I just generalized the id into a single forge hook and passed in the required information there.


ChampionAsh5357:
Closing PR based on Lex's reponse.

PR #7336, opened 2020-09-17
Make Biome.Climate and BiomeAmbiance usable from BiomeLoadingEvent

### LINKED ISSUE #7335: Biome.Climate in BiomeLoadingEvent is unusable
In the new `BiomeLoadingEvent`, there is an option for `setClimate(Biome.Climate)`. The intent here is to modify existing climates, however this is unusable for several reasons:

- The `Biome.Climate` instance provided by the event is immutable and does not expose any builder or mutators, meaning the only way to modify the climate would be creating a new one and invoking `BiomeLoadingEvent#setClimate`. However...
- The constructor for `Biome.Climate` is private, meaning you cannot construct a new `Climate` object outside of using the codec (which is really quite gross)
- There are no accessors for any of the fields in `Biome.Climate`. So if you want to modify only one field of the climate (and use the current values for the rest), you cannot access the current values to create a new climate object.

alcatrazEscapee:
This fixes #7335 by adding a builder to the climate settings.

Vanilla does not have a builder, nor a public constructor, nor any accessors or getters unlike the other biome objects. This PR:

- Adds a new builder for the climate settings, which invokes the now public constructor.
- Replaces the `Biome.Climate` instance on `BiomeLoadingEvent` with the builder as the existing climate was unusable (no public getters, only private final fields)
- I've tried to not expose any unnecessary internals to modders and only provide the forge modifiable hookable points.

*No I do not consider "serialize the climate to json using the codec, then write to the `JsonElement`, then deserialize it again" a usable method for biome modifications*


alcatrazEscapee:
I've also noticed a lot of these same concerns apply to `BiomeAmbiance` although I'm not sure what the best way to approach this one is (or if I should make it a separate follow-up PR, as the patches will almost certainly conflict with this one):

- In `BiomeLoadingEvent`, there's a setter and getter for effects. So the intent is to re-create your own effects instance, likely re-using the old one for any properties that you don't need to re-use.
- However, all the accessors for the old effects instance are client only, and thus effectively unusable. And due to a lot of the fields on `BiomeAmbiance` being optionals, this will create a lot of unnecessary code duplication when copying a biome ambiance to a builder (when a mod is doing it).

Ideally forge just exposes a builder rather than the raw unusable vanilla pojo's. But on the stream where cpw was architecting the system I was told there was no need for builders for these as they only contained simple data. Some guidance here would be appreciated.

patches/minecraft/net/minecraft/world/biome/Biome.java.patch
```diff
@@ -16,7 +16,7 @@
 -      })).apply(p_235064_0_, Biome::new);
 +      }), ResourceLocation.field_240908_a_.fieldOf("forge:registry_name").forGetter(Biome::getRegistryName))
 +      .apply(p_235064_0_, (climate, category, depth, scale, effects, gen, spawns, name) ->
-+          net.minecraftforge.common.ForgeHooks.enhanceBiome(name, climate, category, depth, scale, effects, gen, spawns, p_235064_0_, Biome::new));
++          net.minecraftforge.common.ForgeHooks.enhanceBiome(name, new net.minecraftforge.common.world.BiomeClimateBuilder(climate.field_242460_b, climate.field_242461_c, climate.field_242462_d, climate.field_242463_e), category, depth, scale, effects, gen, spawns, p_235064_0_, Biome::new));
```
LexManos: This should be inside the hook like everything else is...

alcatrazEscapee: The reason I did it this way was to avoid ATing and/or patching accessors for all the fields in `Biome.Climate` but I can change it if that's preferred.

LexManos: They are final fields, so you could just AT them. It is better practice to add getters, but this is minecraft we have to have a happy middle. Plus, we need access to these values somehow to let modders make decisions on it.

alcatrazEscapee: If that's the case, I could axe the entire builder, AT all the fields and expose the constructor - would that be a preferred solution (for this and potentially `BiomeAmbiance`)? Or would it be better to expose a builder in the event rather than not.

LexManos: Im not opposed to the builder, as it lets people just edit one value without having to rebuild the object. But it is a large section of code that we'd have to keep maintaining. 
So I think public {still final} fields, and public constructor is good for now.
If modders wanna replace just one value they can copy the others from the existing one in the event.

alcatrazEscapee:
Alright then, so in the interest of "least code for forge to maintain", this is now very simple:

- Add AT entries to both `Biome.Climate` and `BiomeAmbiance` so they can be recreated completely during `BiomeLoadingEvent`. Both have setters on the event, the latter has a builder, and the former's constructor is now public.

This may not be the nicest API for modders to use but given the situation (and the half baked biome stuff) it'll do.

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -299,6 +299,23 @@ public net.minecraft.world.World field_73004_o #rainingStrength
...
+public net.minecraft.world.biome.Biome$Climate <init>(Lnet/minecraft/world/biome/Biome$RainType;FLnet/minecraft/world/biome/Biome$TemperatureModifier;F)V
+public net.minecraft.world.biome.Biome$Climate field_242460_b # precipitation
+public net.minecraft.world.biome.Biome$Climate field_242461_c # temperature
+public net.minecraft.world.biome.Biome$Climate field_242462_d # temperatureModifier
+public net.minecraft.world.biome.Biome$Climate field_242463_e # downfall
+public net.minecraft.world.biome.BiomeAmbience field_235205_b_ # fogColor
+public net.minecraft.world.biome.BiomeAmbience field_235206_c_ # waterColor
+public net.minecraft.world.biome.BiomeAmbience field_235207_d_ # waterFogColor
+public net.minecraft.world.biome.BiomeAmbience field_235208_e_ # particle
+public net.minecraft.world.biome.BiomeAmbience field_235209_f_ # ambientSound
```
LexManos: All of these have getters, you need to remove the sided annotations with the SAS file.

alcatrazEscapee: Had no idea that was a thing. Well, TIL.

PR #7337, opened 2020-09-17
Fix modded EntityClassifications

maxanier:
The `EntityClassification` enum can be extended thanks to Forge's `IExtensibleEnum`.
However, then included `VALUES_MAP` is not updated with the new additions, which causes issues.

This PR fixes this by adding new elements to the map in the constructor.
Also updates the CODEC

### DIFF
--- a/patches/minecraft/net/minecraft/entity/EntityClassification.java.patch
+++ b/patches/minecraft/net/minecraft/entity/EntityClassification.java.patch
@@ -9,7 +9,16 @@
     MONSTER("monster", 70, false, false, 128),
     CREATURE("creature", 10, true, true, 128),
     AMBIENT("ambient", 15, true, false, 128),
-@@ -57,6 +57,17 @@
+@@ -14,7 +14,7 @@
+    WATER_AMBIENT("water_ambient", 20, true, false, 64),
+    MISC("misc", -1, true, true, 128);
+ 
+-   public static Codec<EntityClassification> field_233667_g_ = IStringSerializable.func_233023_a_(EntityClassification::values, EntityClassification::func_233670_a_);
++   public static Codec<EntityClassification> field_233667_g_ = net.minecraftforge.common.IExtensibleEnum.createCodecForExtensibleEnum(EntityClassification::values, EntityClassification::func_233670_a_);
+    private static final Map<String, EntityClassification> field_220364_f = Arrays.stream(values()).collect(Collectors.toMap(EntityClassification::func_220363_a, (p_220362_0_) -> {
+       return p_220362_0_;
+    }));
+@@ -57,6 +57,16 @@
        return this.field_82707_i;
     }
  
@@ -21,7 +30,6 @@
 +   @Deprecated
 +   public void init() {
 +      field_220364_f.put(this.field_220365_j, this);
-+      field_233667_g_ = IStringSerializable.func_233023_a_(EntityClassification::values, EntityClassification::func_233670_a_);
 +   }
 +
     public int func_233671_f_() {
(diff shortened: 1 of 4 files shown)


maxanier:
NVM this is a stupid implementation, will fix in a moment

maxanier:
Ok, better implementation copied from #7257.
Not sure if that one is merged before RB or at all. If not, please consider this PR as a simple fix for a simple problem.

patches/minecraft/net/minecraft/entity/EntityClassification.java.patch
```diff
@@ -9,13 +9,20 @@
...
  
 +   public static EntityClassification create(String name, String id, int maxNumberOfCreatureIn, boolean isPeacefulCreatureIn, boolean isAnimalIn, int despawnDistance) {
 +      throw new IllegalStateException("Enum not extended");
 +   }
++
++   @Override
++   @Deprecated
++   public void init() {
++      field_220364_f.put(this.field_220365_j, this);
++      field_233667_g_ = IStringSerializable.func_233023_a_(EntityClassification::values, EntityClassification::func_233670_a_);
```
pupnewfster: I have my doubts if this second line is actually needed as the Codec seems to reference `Enum#values` and not the cached `VALUES_MAP`.

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -315,4 +315,5 @@ public net.minecraft.world.gen.layer.LayerUtil func_202829_a(JLnet/minecraft/wor
 private-f net.minecraft.world.raid.Raid$WaveMember field_221284_f # VALUES
 public net.minecraft.world.server.ServerChunkProvider field_186029_c # chunkGenerator
 public net.minecraft.world.server.ServerChunkProvider field_73251_h # worldObj
-public net.minecraft.world.storage.FolderName <init>(Ljava/lang/String;)V # constructor
\ No newline at end of file
+public net.minecraft.world.storage.FolderName <init>(Ljava/lang/String;)V # constructor
+public-f net.minecraft.entity.EntityClassification field_233667_g_ # CODEC
```
pupnewfster: For future reference (or if I am wrong about modifying the codec being wrong), when you modify the access transformer file you should run the `checkATs` task as it orders things alphabetically and `entity` should be way before `world`.

patches/minecraft/net/minecraft/entity/EntityClassification.java.patch
Cyborgmas: yes but it caches it instantly. so a better way might be to change that simply (in IStringSerializable). remove the first line and do .get() on the function directly.

maxanier: the latter would probably be the better solution as it would also resolve this for other extensible enums that use this (currently only `net.minecraft.world.gen.feature.jigsaw.JigsawPattern.PlacementBehaviour` though).
But it is more "intrusive", although I can't see any issues on first sight/test), I am not going to make that call

LexManos: Makeing the codec a public non-final is a non-starter.
Perhaps the better solution would be to introduce a new codec type for extensible enums?
And then change the codec to use that instead of the default enum one.
Or, we could change the default enum one {if its in MC code} to not cache the values.
Either option is better then letting all modders change the entire codec object willy nilly.

maxanier: ok, removed the caching as Cybogmas suggested, but just for the two enums that are `IExtensibleEnum`

patches/minecraft/net/minecraft/entity/EntityClassification.java.patch
```diff
@@ -9,13 +9,28 @@
...
     }
  
 +   public static EntityClassification create(String name, String id, int maxNumberOfCreatureIn, boolean isPeacefulCreatureIn, boolean isAnimalIn, int despawnDistance) {
 +      throw new IllegalStateException("Enum not extended");
 +   }
++
++   @Override
++   @Deprecated
++   public void init() {
++      field_220364_f.put(this.field_220365_j, this);
```
ichttt: vanilla calls getName instead of using the raw field, we should mirror this (even though it shouldn't make a difference)

PR #7338, opened 2020-09-17
Fix ToggleableKeyBinding differences to vanilla.

diesieben07:
A `ToggleableKeyBinding` can be active (`isKeyDown`) even when it's conflict context or modifier are not active. Instead those must be checked when the toggle happens.

A symptom of this is the player stopping to sneak while a GUI is open when sneaking-toggle is enabled.

[Reported on the forums](https://forums.minecraftforge.net/topic/91437-sneak-takes-off-in-chat-menu/).

### DIFF
--- a/src/main/resources/META-INF/accesstransformer.cfg
+++ b/src/main/resources/META-INF/accesstransformer.cfg
@@ -176,6 +176,7 @@ private-f net.minecraft.client.renderer.model.ModelBakery field_217853_J # field
 protected net.minecraft.client.renderer.model.ModelBakery func_177594_c(Lnet/minecraft/util/ResourceLocation;)Lnet/minecraft/client/renderer/model/BlockModel; # loadModel
 private-f net.minecraft.client.renderer.tileentity.PistonTileEntityRenderer field_178462_c # blockRenderer - it's static so we need to un-finalize in case this class loads to early.
 public net.minecraft.client.renderer.tileentity.TileEntityRendererDispatcher field_147557_n # fontRenderer - needed for rendering text in TESR items before entering world
+default net.minecraft.client.settings.KeyBinding field_74513_e # pressed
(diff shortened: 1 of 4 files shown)


src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -176,6 +176,7 @@ private-f net.minecraft.client.renderer.model.ModelBakery field_217853_J # field
 protected net.minecraft.client.renderer.model.ModelBakery func_177594_c(Lnet/minecraft/util/ResourceLocation;)Lnet/minecraft/client/renderer/model/BlockModel; # loadModel
 private-f net.minecraft.client.renderer.tileentity.PistonTileEntityRenderer field_178462_c # blockRenderer - it's static so we need to un-finalize in case this class loads to early.
 public net.minecraft.client.renderer.tileentity.TileEntityRendererDispatcher field_147557_n # fontRenderer - needed for rendering text in TESR items before entering world
+default net.minecraft.client.settings.KeyBinding field_74513_e # pressed
```
pupnewfster, line 179: Shouldn't this be public or protected or something given it is in a class and not an interface (if default even works in ATs in interfaces, no idea)

diesieben07: I made it default, because the field should not be used by modders.

LexManos: The access you want is protected, so that the subclass ToggleableKeyBinding can access it, but others cant.

diesieben07: I know what protected does. I specifically did _not_ make it protected because (like I said above) mods should not mess with this field, imho.

LexManos: Works for me.

PR #7341, opened 2020-09-19
Make Biome not a ForgeRegistryEntry

maxanier:
This is mainly meant for discussion. I haven't found any kind of ongoing discussion regarding the new dynamic biome system, but I think there should be one.

The new dynamic registry system introduced by Mojang results in Biome objects being not unique anymore. This does not really work with Forge's registry system since it is expected that there is always just one object for a registry key.
Lex still made the biome a `ForgeRegistryEntry` again in https://github.com/MinecraftForge/MinecraftForge/commit/d8eaeabf3cf03b2de05c23ce88c3b75548720014 to allow modders to register new biomes similar to before for now.

This works ok, but there are a few downsides.
For example you can no longer compare a biome object retrieved from the world to the one you created as they are not the same object even if they are the same biome (pointer/reference equality). You can even compare `getRegistryName()` because the biome objects in the world don't have this set.
Therefore it is misleading to modders that the Biome class still extends `ForgeRegistryEntry`.

Without the forge registry events, modders are still able to register new biomes using data packs (probably need #7257 ), although users will receive a warning on world load.

BTW: The correct way to retrieve the registry name of a biome in a world is `world.func_241828_r().func_243612_b(Registry.BIOME_KEY).getKey(biome)` now.


GirafiStudios:
Pupnewfster have been working on resolving the issue in this PR https://github.com/MinecraftForge/MinecraftForge/pull/7316

maxanier:
As far as I can tell that just adds a way to identify biomes using tags, but does not modify the fact that biome objects are not "singleton per biome registry name" and the fact that `getRegistryName` returns null for biomes in a world. Or did I overlook something?

GirafiStudios:
You're correct that it does not currently, I just had the impression that he was working towards that too - could be wrong though!

pupnewfster:
It is something I was experimenting with but after various discussions with lex we decided to kill off trying to have tags for biomes because trying to pair them with the dynamic registry system is just flawed and the main use case of them would be for filtering stuff in the BiomeLoadingEvent to only add stuff to specific biome types, but vanilla's own biomes that go Object -> Json -> Object have that happen quite a while before there is any form of datapack access to even be able to manually read tags.

LexManos:
Biomes being in a Forge Registry is a hack that exists because modders needed something and until Mojang figures out their data driven system, this is it. Yes, there are things that could not have names when in world as they are a part of the dynamic registry and not part of the defaults registry.
However, getRegistryName is Nullable for a reason, https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/src/main/java/net/minecraftforge/registries/IForgeRegistryEntry.java#L51
When working with the new dynamic registries you'll need to keep this in mind.

However yes, the entire registry system needs to be rewritten because of how Mojang decided to manage their dynamic registries, I'm really hoping I can convince them to do something a bit more sane in future versions.

maxanier:
Gimpansor has pointed out on Discord that the core benefit of biomes being ForgeRegistryEntries is raw-id persistance. JSON biomes do not have this, but it is required to make sure a world can be loaded correctly again.

Therefore this is required until Mojang changes their system

PR #7343, opened 2020-09-19
Added patch (and test mod) for creating custom boat entity types.

rehwinkel:
This patch adds the ability to easily add a new Boat type. This means modders can easily create custom boats without having to copy the BoatEntity class or anything.


ChampionAsh5357:
This is not really a patch, you just rewrote most of Minecraft's code. Instead of replacing everything, the enum should extend `IExtenisbleEnum` or whatever it's called. Anything else is unnecessary. You can create your own `BoatRenderer` and `IDispenseItemBehavior`. The `BoatEntity` and `BoatItem` would be handled the same. Maybe the only other thing you would need to patch is `BoatEntity#func_184455_j`; however, that could be done with a simple return statement at the without removing anything else.

PR #7344, opened 2020-09-19
Make it possible to modify the spawn lists for structures

pupnewfster:
In 1.16.2 Mojang switched the four structures that are hardcoded (Swamp Huts, Pillager Outposts, Ocean Monuments, and Nether Fortresses) as additional things to check for mob spawns to immutable lists from array lists. This PR modifies how NoiseChunkGenerator gets the spawn lists based on structures to be able to check all structures (if they have spawns for the given classification), instead of being hardcoded to the four structures vanilla uses to supporting any structure. This PR also adds an event to allow modifying this list for any structure including the four hardcoded vanilla ones so that spawns can be added to/removed from the hardcoded structures.


halvors:
+1

marchermans (approved):
This looks good.
Tested and working my machine.

marchermans (approved):
Code looks good. 
Question: Why are we using a static class/method based manager instead of a singleton based on?

All in all. Works on my machine. Code formatting seems to be okey, and implemented based on feedback. Assigning.

TelepathicGrunt:
Hey, could you please add documentation on how to use this PR? Maybe even add it to the Forge Docs too?
It would really help modders learn and adopt this hook into their code and use it properly

pupnewfster:
I don't currently have time to write up something for the Forge Docs, which is why I made sure to java doc the vast majority of methods and stuff in this PR. What is the part that you are having trouble understanding?

TelepathicGrunt:
I was trying to figure out the event through the commits and it wasn't clear which file was the part players would see. But I understand it now and it's much easier to follow when actually using it in Forge lol. 

I did add the event into my [StructureTutorialMod](https://github.com/TelepathicGrunt/StructureTutorialMod/blob/890c0b2a42cd892a65f0ef3e656a25fda7b1a234/src/main/java/com/telepathicgrunt/structuretutorial/StructureTutorialMain.java#L101-L120) that I show people so that more people will know about the event. 

Though there is one thing that I am still unsure about. In my quick test, it seemed like normal mobs still spawn in my structure alongside the new mobs I added. Is there a way for users to cancel the regular mobs and only spawn that structure's mobs instead or was my testing flawed? Some people might want to have structure spawns be the only spawns for their structure and to stop the Biome's mobs from being in it. But this is kind of more of a specific use case so maybe it's best to leave that to the other mobspawnevents

pupnewfster:
Only mobs that spawn in your structure spawn (though other mobs may spawn nearby and walk into it). Also assuming you are adding the spawns to your own structure I would recommend overwriting `IForgeStructure#getDefaultSpawnList()` instead so that you don't have to listen to the event and the values are in there by default when the event fires. The use of the event is more for adding to or removing from structures that are not your own.

TelepathicGrunt:
Ah. Thank you! Didn't know it already hooked up to the structure spawn list. I'll update my tutorial to that then

PR #7001, opened 2020-07-20
Added hook for custom edge biomes [1.15.x]

ghost:
> Added a hook to `Biome` so that people can add edge biomes to their biome, to blend between biomes or to make a more natural border.
> Example biome class: https://hatebin.com/jqacmqwywl
> ```java
> @Override
> public Biome getEdge(INoiseRandom rand, Biome northBiome, Biome westBiome, Biome southBiome, Biome eastBiome) {
>     return Biomes.BADLANDS;
> }
> ```
> Very simple hook and patch, let me know any changes that would be required.
Direct port of #7000 to 1.15.x.

PR #7346, opened 2020-09-21
Reintroduce BiomeManager to allow modded biomes in the overworld

maxanier:
The BiomeManager, which allows registering biomes to be generateed in the overworld, was commented out during the 1.16.2 update to evaluate the biome changes.

However, in the biome generation code almost nothing changed. The only import change is the introduction of dynamic/JSON biomes. But they are not supported in Overworld worldgen in Vanilla either.
Additionally and most importantly, it appears to be impossible to allow the inclusion of (modded) JSON biomes on Forge's end, because the vanilla system relies on integer ids for biome selection. So, if the id of a dynamic biome changes (as they are not persistent I think) the biome layer will break. `ForgeRegistryEntry` biome ids are managed and stored by Forge.

Therefore, I think there is no reason the change the system until Mojang changes their own system (to e.g. incorporate dynamic biomes better). So, I have updated the `BiomeManager`/patches.

~Additionally, I swapped the `ImmutableList` in `OverworldBiomeProvider` for a normal array list. The field is just used once and immediately copied to a new immutable list (via streams) anyways. Mods should probably make use of the new work queue to ensure sequential access.~
Since the biome list in `OverworldBiomeProvider`/`BiomeProvider` should include all biomes that might be generated in the world, I also added a way to register any biome that might be generated using the biome entries or otherwise.

I left out the hooks added #7001, #5969 and #6571. But I will reintroduce them if desired and after the concept (of just reintroducing `BiomeManager` has been approved by Lex or similar).
#6853 and #6745 are also related, but out of my scope.

I have tested this PR with a mod of mine and the biome generates just as expected. 
Additionally, I tested what happens when my mod/biome is removed/not registered: No crash, already generated biomes might behave strangely as expected though


patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
```diff
@@ -0,0 +1,77 @@
+--- a/net/minecraft/world/gen/layer/BiomeLayer.java
```
valoeghese, line 1: this patch can probably be cleaned up a bit since we're readding it

valoeghese (changes requested):

patches/minecraft/net/minecraft/world/biome/provider/OverworldBiomeProvider.java.patch
```diff
@@ -0,0 +1,11 @@
...
...omes.field_76787_r, Biomes.field_76786_s, Biomes.field_76785_t, Biomes.field_76784_u, Biomes.field_76783_v, Biomes.field_76782_w, Biomes.field_76792_x, Biomes.field_150574_L, Biomes.field_150575_M, Biomes.field_150576_N, Biomes.field_150577_O, Biomes.field_150583_P, Biomes.field_150582_Q, Biomes.field_150585_R, Biomes.field_150584_S, Biomes.field_150579_T, Biomes.field_150578_U, Biomes.field_150581_V, Biomes.field_150580_W, Biomes.field_150588_X, Biomes.field_150587_Y, Biomes.field_150589_Z, Biomes.field_150607_aa, Biomes.field_150608_ab, Biomes.field_203614_T, Biomes.field_203615_U, Biomes.field_203616_V, Biomes.field_203617_W, Biomes.field_203618_X, Biomes.field_203619_Y, Biomes.field_203620_Z, Biomes.field_185441_Q, Biomes.field_185442_R, Biomes.field_185443_S, Biomes.field_185444_T, Biomes.field_150590_f, Biomes.field_150599_m, Biomes.field_185445_W, Biomes.field_185446_X, Biomes.field_185447_Y, Biomes.field_185448_Z, Biomes.field_185429_aa, Biomes.field_185430_ab, Biomes.field_185431_ac, Biomes.field_185432_ad, Biomes.field_185433_ae, Biomes.field_185434_af, Biomes.field_185435_ag, Biomes.field_185436_ah, Biomes.field_185437_ai, Biomes.field_185438_aj, Biomes.field_185439_ak);
```
LexManos: No, Why is this needed? And why is it public/mutable?
You removed the functions in BiomeManager that messes with theses.
It's also probably better to move this into accessor methods in BiomeManager not make it publicly mutable without any way to track {Yes I know the old patch did but the old patch isn't always best}

patches/minecraft/net/minecraft/world/gen/layer/BiomeLayer.java.patch
maxanier: Can you be more specific?
I would like to keep the structure as closely to Vanilla as reasonable.
Could probably get rid of the null checks (and use empty lists in the biome manager where necessary) though.
Could move the legacy biome stuff into BiomeManager as well, but I think it is more appropriate here as it shouldn't be modified by modders

patches/minecraft/net/minecraft/world/biome/provider/OverworldBiomeProvider.java.patch
shartte: This list should be updated with any modded biome that is added to the overworld. It's passed on to the ctor of the biome source, which uses it for lookups such as "hasStructure". An Accessor should be fine to do this, an alternative is to patch OverworldBiomeProvider's constructor and "enrich" the list there before it is passed to the parent constructor.

maxanier: Yes, (a copy of) the list is used for various purposes and should include all biomes generated in the overworld.
I have adapted the patch to leave this unmodified and instead enrich the list in the constructor as shartte suggested

maxanier:
The biome list in the `BiomeProvider` should include all biomes that may be generated in the world. This is not limited to the biome entries added to the `BiomeManager` and used by the `BiomeLayer`.
I changed the patch to a hopefully better variant.

`BiomeManager#addBiomes` and `BiomeManager#addBiomeEntry` are kept separate on purpose because as stated above the all biome list might include additional biomes. If you want, I can make `addBiomeEntry` also add the biome to the biome list so modders can't forget this.

A few "style" ideas I can implement if desired:
- Rename `BiomeManager` to `ForgeBiomeManager` or `OverworldBiomeManager` or similar because there already is a vanilla BiomeManager (that does something different)
- Rename the methods to include `overworld` to add better clarity if #6853 is merged
- Change to `BiomeManager#addBiomes` to accept a list instead of a vararg as this creates a warning due to the used generics
- Filter the final BiomeKey list passed to the BiomeProvider for duplicates if modders should somehow mess up and add vanilla biomes here as well. Shouldn't be an issue though

shartte:
Does the ForgeRegistry for Biomes take care of `net.minecraft.world.biome.BiomeRegistry#idToKeyMap` (field_244202_c), or does the PR somehow circumvent its use? IIRC it is only used for layering purposes.

maxanier:
Forge already patches the `BiomeRegistry#func_244203_a` (getBiomeForId or similar) to use the ForgeRegistry instead of the `idToKeyMap`
Therefore, as far as I can tell, the `BiomeRegistry#idToKeyMap` is not used anymore.
(It is not updated though)

LexManos:
I've worked on this, and reintroduced the BiomeLayer patches in https://github.com/MinecraftForge/MinecraftForge/commit/d1e3567c6b1c4ccd94ff9bf469c62e7933ec2bec
I don't know what we need to do for the provider. We can look more into that in a new PR if you want to dive into it.

PR #7347, opened 2020-09-22
Standard interface for multi-output crafting

waitingtocompile:
There's currently no fixed standard for recipes that output multiple items, which makes interoperability between mods that want to do that needlessly awkward. This interface makes that less of a pain.

The code is shamelessly stolen from Gigaherz who posted it during the discussion about it and left the actual PR for other people to do.

### DIFF
--- a/src/main/java/net/minecraftforge/common/crafting/IMultiOutputRecipe.java
+++ b/src/main/java/net/minecraftforge/common/crafting/IMultiOutputRecipe.java
@@ -1,0 +1,73 @@
+package net.minecraftforge.common.crafting;
+
+import com.mojang.datafixers.util.Pair;
+import net.minecraft.inventory.IInventory;
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.crafting.IRecipe;
+import net.minecraft.util.NonNullList;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Random;
+
+public interface IMultiOutputRecipe<T extends IInventory> extends IRecipe<T> {
+    @Deprecated
+    @Override
+    default ItemStack getCraftingResult(T inputs) {
+        return getCraftingResults(inputs).get(0);
+    }
+
+    /**
+     * Gets the additional results from the recipe.
+     * @return The list of results.
+     *    The float value on the pair is the chance for that result to happen.
+     *    A chance of 1.0 means 100%.
+     */
+    List<Pair<ItemStack, Float>> getAdditionalOutputs();
+
+    /**
+     * Gets the final results
+     * @param inputs The inventory exposing the input slots
+     * @return The list of outputs from the crafting.
+     */
+    NonNullList<ItemStack> getCraftingResults(T inputs);
+
+    abstract class Impl<T extends IInventory> implements IMultiOutputRecipe<T>
+    {
+        protected static final Random rand = new Random();
+
+        private final ItemStack primaryOutput;
+        private final List<Pair<ItemStack, Float>> additionalOutputs;
+
+        protected Impl(ItemStack primaryOutput, List<Pair<ItemStack, Float>> additionalOutputs)
+        {
+            this.primaryOutput = primaryOutput;
+            this.additionalOutputs = additionalOutputs;
+        }
+
+        @Override
+        public ItemStack getRecipeOutput()
+        {
+            return primaryOutput;
+        }
+
+        @Override
+        public List<Pair<ItemStack, Float>> getAdditionalOutputs()
+        {
+            return Collections.unmodifiableList(additionalOutputs);
+        }
+
+        @Override
+        public NonNullList<ItemStack> getCraftingResults(T inputs)
+        {
+            NonNullList<ItemStack> outputs = NonNullList.create();
+            outputs.add(primaryOutput);
+            for(Pair<ItemStack, Float> additional : additionalOutputs)
+            {
+                if (rand.nextFloat() < additional.getSecond())
+                    outputs.add(additional.getFirst());
+            }
+            return outputs;
+        }
+    }
+}


KnightMiner:
What is the benefit of this? I have never needed to know that another mod's recipe returns multiple items. For my own recipes, its trivial to make such an interface.

Also, does every "multi output" recipe really need "additional outputs"? All the multioutput recipes I have written are not chance based, so adding that is additional overhead I do not need. That method really needs a better name in any case, such as `getRandomOutputs` or `getChanceOutputs`.

waitingtocompile:
The specific use case that was being discussed is if your mod adds a crusher, and you want the crushing recipes from other mods to work in it cleanly, you might know the recipetypes that you want to check the registry for. But any recipes that do have multiple outputs will require you to manually fiddle around with whatever system that mod author used to handle that, with no guarentee of consistency or that there will even be a usable API for that mod. This interface fixes that problem.

As for the additional outputs, that's mostly there for things like JEI wanting to accurately display the outputs beyond the first. There are going to be times when modders want those tied to random chance and so anything asking for that information should be able to also get those probabilities. If your mod doesn't apply randomness then you should just give `1.0f`. It's not for _just_ the randomized outputs, it's the counterpart to `getRecipeOutput` for _all_ of the crafting outputs beyond the first, that may or may not be tied to randomization. Hell, the interface doesn't even demand that it's even accurate, since the the _actual_ crafting result from any given actual craft should instead be provided by `getCraftingResults`

ChampionAsh5357:
That use case can be provided for many different crafting types for many different mods. It doesn't necessarily explain why Forge would need to have this. As KnightMiner explained, this a trivial interface to make. Multi-Compatibility between mods if up for the authors to decide. I could PR one for multiple furnace outputs and it would still be missing the same reasoning.

As for JEI to accurately to display outputs, that's once again up to the mod author for compatibility. You can have the additional outputs display just as many mods have done by creating their own interface to compat with JEI.

Plus, the issue more refers to a lack of standards for an `IRecipeType`. If you wanted to do this, you should leave documents for recipe types namespaced with `forge`. Otherwise, this whole API would be non-applicable without some cast check which might have differing generics.

gigaherz:
Separating the results into getExtraOutputs + getChanceOutputs would certainly be an option. When I wrote that gist I was thinking to minimize the complexity of the implementation, so having non-chance outputs with chance=1.0 seemed appropriate.
As I said on Discord, I don't personally have an use case for multi-output recipes *yet*, so whether or not this is useful depends on people's use cases, and the use case being discussed was how to share recipes across mods, without relying on some 3rdparty library.

EDIT: I didn't see Ash's message until I sent my reply:
I strongly believe that "mods can do that on their own forge isn't needed" is the worst reason for dismissing a PR. If we start from that frame of mind, 99% of forge wouldn't exist. "Who needs capabilities, mods can just cooperate to use each other's interfaces", "Who needs tags/oredict, mods can just cooperate to find each other's blocks", ...

alcatrazEscapee:
I would argue that this interface does not provide anything generically useful. For instance, possible other extensions of `IRecipe` may want to provide:

- Additional outputs conditional on some property of the machine or the input
- Fluid outputs, or both Fluid and Item outputs
- Outputs of any other manner (energy, or even something mod specific)
- Block / blockstate outputs, for those using this interface to codify in-world crafting operations (I am.)

There's *definitely* no reason for forge to provide all of those base implementations, so why should it include this one? I think "sharing recipes across mods" is going to require knowing what the type or implementation of the recipe is and this just arbitrarily choosing `List<Pair<ItemStack, Float>>` as the most general/useful form of this when it is in fact, not.

waitingtocompile:
As giga said, the point is that with this I can grab from, say, half a dozen different recipe types as needed, hell I could even offload what recipeTypes are handled by what machine into a config if I was so inclined and thus make it possible for any user to make any machine capable of any recipeType. I don't need to know what the actual implementation is or how it works, I can just do an `instanceof` check to determine if it is a multi output recipe, and if so ask it for it's full list of outputs. No need to rummage through a third party API that might not even be provided.

Alcatraz commented while I was typing this up. In an ideal world I'd love some extensive and flexible set of standards for those kinds of things to be included in forge. I don't know what that would actually look like in practice or if it's remotely practical to do something like that. The use case I have for this interface or something similar to it is something that's been tugging at the back of my mind for a while now as something I wanted to have, and upon seeing other people say that they thought it was something worthwhile and giga put forward a possible implementation of it I figured it was worth doing. If someone does have a better idea of how to standardize secondary outputs of recipes I'd be more than happy to at least experiment with it and see what I can put together.

alcatrazEscapee:
If the intent is "I want to know what output this recipe of an unknown type provides", then a possible generic solution may look like:
```java
interface ISpecializedRecipe<T, C extends IInventory> extends IRecipe<C> {
    // Deprecate the ItemStack output method and replace it one that returns a generic T
    RecipeOutputType<T> getOutputType();
```

And then have a type class capable of representing any arbitrary output type, with a simple registry system in the manner that `RecipeType` uses:
```java
// T is the actual type of the output, whether it be ItemStack, List<ItemStack>, Pair<ItemStack, FluidStack>, whatever
class RecipeOutputType<T> {
    static RecipeOutputType<ItemStack> ITEM_STACK = register("item_stack", ItemStack.class);
    static RecipeOutputType<List<ItemStack>> ITEM_STACKS = register("item_stacks", new TypeToken<List<ItemStack>>(){}.getType());
    static RecipeOutputType<FluidStack> FLUID_STACK = register("fluid_stack", FluidStack.class);
    // etc.

    public static <T> RecipeOutputType<T> register(...) {
    // return a unique instance based on some identifying traits, possibly class or type token, or identifier idk
}
```

If you wanted to handle recipes that produce multiple items, you'd check the output type against `ITEM_STACKS` for example. But I think this gets back to my original point: You *cannot* have a system for "handle the possible outputs of this recipe that I do not know the implementation of", because modders (myself included) may be doing very unintended (or unsupported things) with the vanilla interface.

For the use case of JEI, this is (I would assume) the reason why they have an interpreter / bridge system between raw mod recipes, and how to display them. In which case power is given to the modder to make any display or other transformations they need.

sciwhiz12:
Hello! @waitingtocompile, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after a few days, this PR will be closed for inactivity.

Technici4n:
This is something that would help mods such as AE2 discover the outputs of a recipe when encoding it. Otoh, grabbing them from the REI/JEI display is also a viable solution that doesn't require forge nor modder work.

TheCurle:
This is a single-class change with merge conflicts (via jenkins config) targetting the LTS version with a broken github check, so i'm going to close this.

If you want this in 1.18, you may open a new PR and hope the GitHub gods shine down upon you.

PR #7350, opened 2020-09-23
2nd take at better license error

ichttt:
This is an alternative to  #7264
The main difference user-side is that it shows the correct error when the version range is not satisfied (in #7264 it would show the missing license screen, here it correctly shows the wrong version screen).
This is done by deferring the crash, so fmllauncher runs cleanly and once modloading is about to start the error is populated.
Some screenshots:
Mod for 1.15.2 running on 1.16.3 with this PR:
java_KCeelLfykr
Mod for 1.16.1, but with version range `[1.16,1.17)` running in 1.16.3 with this PR:
java_JQAVCMFxTK
Both would hard-crash without any screen before, leaving most users clueless behind

### DIFF
--- a/src/fmllauncher/java/net/minecraftforge/fml/loading/EarlyLoadingException.java
+++ b/src/fmllauncher/java/net/minecraftforge/fml/loading/EarlyLoadingException.java
@@ -19,6 +19,8 @@
 
 package net.minecraftforge.fml.loading;
 
+import net.minecraftforge.forgespi.language.IModInfo;
+
 import java.util.List;
 
 /**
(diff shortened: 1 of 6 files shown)


src/fmllauncher/java/net/minecraftforge/fml/loading/EarlyLoadingException.java
```diff
@@ -29,10 +31,19 @@
...
 
 
+        private final IModInfo modInfo;
         private final String i18message;
         private final Object[] args;
+
         public ExceptionData(final String message, Object... args) {
             this.i18message = message;
+            this.modInfo = null;
+            this.args = args;
```
pupnewfster: This might be a bit cleaner as `this(message, null, args);`

Issue #7324, opened 2020-09-12
[1.16] Bug: Axe is not effective tool for Logs & more


cech12:
**Minecraft Version:** 1.16
**Forge Version:** 33.0.61

**Description of issue:**
The method `BlockState.isToolEffective(ToolType.AXE)` returns wrong values for the axe tool type. Logs and other blocks that are defined only via its materials in `net.minecraft.item.AxeItem` return false for `BlockState.isToolEffective(ToolType.AXE)`. 

I researched a bit and I found a TODO in `net.minecraftforge.common.ForgeHooks::initTools` that could reference to this issue:
```
        blocks.forEach(block -> blockToolSetter.accept(block, ToolType.PICKAXE, 0));
        blocks = getPrivateValue(ShovelItem.class, null, 0);
        blocks.forEach(block -> blockToolSetter.accept(block, ToolType.SHOVEL, 0));
        //TODO Axes check Material and Blocks now.
        blocks = getPrivateValue(AxeItem.class, null, 1);
        blocks.forEach(block -> blockToolSetter.accept(block, ToolType.AXE, 0));
```

I think there are two possibilities to fix this issue:
1. Add at the position of this TODO also a loop that adds the axe tool type to every block with a fitting material. Con: It could be difficult to get a list of all blocks to check their material.
2. Add a material check to the `net.minecraftforge.common.extensions.IForgeBlock::isToolEffective` method. Con: There could be a problem to get the private field of the AxeItem.

Which fix is the better one?

pupnewfster:
Closing this as the PR fixing it was merged.

PR #7351, opened 2020-09-23
Support effective materials of vanilla AxeItem

cech12:
This PR fixes the bug found in https://github.com/MinecraftForge/MinecraftForge/issues/7324

It adds a material check to the `getHarvestLevel` and `getHarvestTool` of the `Block` class to support also the defined effective materials in the AxeItem class.

### DIFF
--- a/patches/minecraft/net/minecraft/block/Block.java.patch
+++ b/patches/minecraft/net/minecraft/block/Block.java.patch
@@ -90,7 +90,7 @@
     }
  
     @OnlyIn(Dist.CLIENT)
-@@ -410,6 +417,83 @@
+@@ -410,6 +417,87 @@
        return this;
     }
  
@@ -108,12 +108,16 @@
 +   @Nullable
 +   @Override
 +   public net.minecraftforge.common.ToolType getHarvestTool(BlockState state) {
++      if (this.harvestTool == null && net.minecraft.item.AxeItem.field_234662_c_.contains(state.func_185904_a()))
++         return net.minecraftforge.common.ToolType.AXE;
 +      return harvestTool; //TODO: RE-Evaluate
 +   }
 +
 +   @Override
 +   public int getHarvestLevel(BlockState state) {
-+     return harvestLevel; //TODO: RE-Evaluate
++      if (this.harvestLevel < 0 && net.minecraft.item.AxeItem.field_234662_c_.contains(state.func_185904_a()))
++         return 0;
++      return harvestLevel; //TODO: RE-Evaluate
 +   }
 +
 +   @Override
--- a/patches/minecraft/net/minecraft/block/Block.java.patch
+++ b/patches/minecraft/net/minecraft/block/Block.java.patch
@@ -90,7 +90,7 @@
     }
  
     @OnlyIn(Dist.CLIENT)
-@@ -410,6 +417,87 @@
+@@ -410,6 +417,83 @@
        return this;
     }
  
@@ -108,15 +108,11 @@
 +   @Nullable
 +   @Override
 +   public net.minecraftforge.common.ToolType getHarvestTool(BlockState state) {
-+      if (this.harvestTool == null && net.minecraft.item.AxeItem.field_234662_c_.contains(state.func_185904_a()))
-+         return net.minecraftforge.common.ToolType.AXE;
 +      return harvestTool; //TODO: RE-Evaluate
 +   }
 +
 +   @Override
 +   public int getHarvestLevel(BlockState state) {
-+      if (this.harvestLevel < 0 && net.minecraft.item.AxeItem.field_234662_c_.contains(state.func_185904_a()))
-+         return 0;
 +      return harvestLevel; //TODO: RE-Evaluate
 +   }
 +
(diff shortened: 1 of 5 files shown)


pupnewfster:
If I am understanding what this is doing correctly (I haven't looked into it in much detail yet), I feel like a better approach would be to modify the canHarvest in `AxeItem` (and `ShovelItem` and `HoeItem`) similar to how there is currently a patch currently in `PickaxeItem` that checks the harvest tool and harvest level.

ChampionAsh5357:
Checking materials might not be the best idea itself. This was the same issue the pickaxe was having with a material defining whether it could be effective or mined. It shouldn't be required that a specific material determines whether an axe is effective against it or not. What if I wanted wood that was resistant to axes? I would need to make an entire new material instead of just having a harvest tool and harvest level.

cech12:
The problem is, that the public block methods `getHarvestLevel` and `getHarvestTool` for logs for example return wrong values for the axe tooltype, because the harvest tool and harvest level are not set for them. Changing the `canHarvest` method of the AxeItem does not change this.

cech12:
I cannot find the message of CampionAsh, but here is my answer to it. Minecraft implemented the material check for the efficiency calculation by itself. So, this PR adds to all these fitting blocks the harvest tool by default. It can be overriden by any mod block by overriding the two methods or set it via item properties. There is no need of creating a new material for it.

ChampionAsh5357:
Apologies, I deleted my reply after I looked a bit more into the file since I overlooked something which negated my point.

patches/minecraft/net/minecraft/block/Block.java.patch
```diff
@@ -108,12 +108,16 @@
 +   @Nullable
 +   @Override
 +   public net.minecraftforge.common.ToolType getHarvestTool(BlockState state) {
++      if (this.harvestTool == null && net.minecraft.item.AxeItem.field_234662_c_.contains(state.func_185904_a()))
```
LexManos: This is a no for a few reasons, 1) Don't make the field public/modifiable for other people. It's not good API design. The better option would be to add a 'is' method in the AxeItem class.
2) Honestly I think you're attacking the wrong target, you should probably go after 'isToolEffective' not getHarvestTool.
Perhaps we should change the default implementation of isToolEffective to `return getHavestTool(state) == null || getHavestTool(state) == tool`

patches/minecraft/net/minecraft/block/Block.java.patch
```diff
@@ -108,12 +108,16 @@
...
++         return net.minecraftforge.common.ToolType.AXE;
 +      return harvestTool; //TODO: RE-Evaluate
 +   }
 +
 +   @Override
 +   public int getHarvestLevel(BlockState state) {
-+     return harvestLevel; //TODO: RE-Evaluate
++      if (this.harvestLevel < 0 && net.minecraft.item.AxeItem.field_234662_c_.contains(state.func_185904_a()))
```
LexManos: This logic is just wrong.. it doesn't take into account levels that are already set.
However, this does look like getHarvestLevel is wrong, the javadocs say it should have a tool class passed in. Perhaps this got lost somewhere in the shuffle. We should re-add it. That way a block could require a iron shovel, or a wooden pickaxe.

patches/minecraft/net/minecraft/block/Block.java.patch
cech12: 1) You are right, adding a static `isEffectiveMaterial(BlockState)` is better than making this field public. I will change that.
2) All blocks that have a material where the axe is effective have not set a harvest tool. So, only adding the check `getHavestTool(state) == null` to `isToolEffective` would not change that and pickaxes would also be an effective tool for logs for example. So, changing the `getHarvestTool` should also be done.

patches/minecraft/net/minecraft/block/Block.java.patch
cech12: The check `this.harvestLevel < 0 &&...` takes that into account, I think. Or what do you mean? 
I could re-add the tool type to this method, if you want. :)

patches/minecraft/net/minecraft/block/Block.java.patch
LexManos: 2) If that's the case, then you should do your first idea and fill in the vanilla blocks with their tools based on the material.
You can easily iterate all blocks using the registry. So it shouldn't be hard.
Vanilla blocks do not have state based materials so you don't have to do anything special. 
Modded blocks should override isToolEffective, Honestly getHarvestTool should be deleted. It's mainly still existing for JEI. Considering multiple tools may work on the same state the concept of only having one is not valid anymore.

cech12: Should I use the ForgeRegistries.BLOCKS registry for this?
Should I remove that method in this PR?

LexManos: Sure
No, 1.17 would be the time, we need to talk about it more.

cech12:
I changed the code like discussed hope everything is fine now. :)

PR #7352, opened 2020-09-24
Feature - default config screen

AlexiyOrlov:
I have designed an optional config screen for mods. It can handle boolean, integer, long, double and string properties.
Its every method can be overridden, allowing extension or upgrade. A test mod is included.
The build fails because some files (not mine) are missing license text.
I also made a generic `IPressable` sub-interface for getting lambda parameter with right class.


src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -19,57 +19,49 @@
...
-import net.minecraftforge.common.data.ForgeItemTagsProvider;
-import net.minecraftforge.common.data.ForgeLootTableProvider;
-import net.minecraftforge.common.data.ForgeRecipeProvider;
-import net.minecraftforge.common.model.animation.CapabilityAnimation;
-import net.minecraftforge.energy.CapabilityEnergy;
-import net.minecraftforge.event.RegistryEvent;
-import net.minecraftforge.fml.event.lifecycle.GatherDataEvent;
-import net.minecraftforge.fluids.capability.CapabilityFluidHandler;
-import net.minecraftforge.items.CapabilityItemHandler;
-import net.minecraftforge.fml.common.Mod;
```
diesieben07, line 72: Why are you reorganizing all these imports?

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -223,4 +215,10 @@ public void registerRecipeSerialziers(RegistryEvent.Register<IRecipeSerializer<?
         event.getRegistry().register(new ConditionalRecipe.Serializer<IRecipe<?>>().setRegistryName(new ResourceLocation("forge", "conditional")));
 
     }
+
+    @OnlyIn(Dist.CLIENT)
```
diesieben07, line 219: Do not use `@OnlyIn`.

src/main/java/net/minecraftforge/fml/client/gui/IPressHandler.java
```diff
@@ -0,0 +1,42 @@
...
+package net.minecraftforge.fml.client.gui;
+
+import net.minecraft.client.gui.widget.button.Button;
+
+import javax.annotation.ParametersAreNonnullByDefault;
+
+/**
+ * Sub-interface with generic parameter to allow using correct class in lambda
+ */
+public interface IPressHandler<W extends Button> extends Button.IPressable {
```
diesieben07, line 29: I am not sure this is needed. It's ugly (unchecked casts) and the only place you use it you don't even need the specific `Button` subclass (you are only calling `setMessage`, which is a method on `Widget`.

src/main/java/net/minecraftforge/fml/client/gui/screen/DefaultConfigScreen.java
```diff
@@ -0,0 +1,250 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.util.Map;
+
+/**
+ * Default screen for configs from the mod menu. Can handle boolean, integer, long, double and string properties.
+ */
+@OnlyIn(Dist.CLIENT)
```
diesieben07, line 41: Do not use `@OnlyIn`.

src/main/java/net/minecraftforge/fml/client/gui/screen/DefaultConfigScreen.java
```diff
@@ -0,0 +1,250 @@
...
+     * Builds screen elements from the config
+     *
+     * @return vertical offset for the next screen element
+     */
+    @SuppressWarnings("unchecked")
+    protected int buildElements(int spacing, int x, UnmodifiableConfig actualValues, String k, Object v) {
+        if (v instanceof ForgeConfigSpec.ValueSpec) {
+            ForgeConfigSpec.ValueSpec valueSpec = (ForgeConfigSpec.ValueSpec) v;
+            Object specValue = actualValues.get(k);
+            Object valueRange = valueSpec.getRange();
```
diesieben07, line 91: Why on earth are you typing this as `Object`? Adjust the visibility on `Range` if you need it.

src/main/java/net/minecraftforge/fml/client/gui/screen/DefaultConfigScreen.java
```diff
@@ -0,0 +1,250 @@
...
+
+    protected int handleString(int spacing, int x, String key, String value, ForgeConfigSpec.ConfigValue<String> stringConfigValue) {
+        TextFieldWidget textField = new TextFieldWidget(font, x + font.getStringWidth(key), spacing, width - font.getStringWidth(key) - 20, 18, key);
+        textField.setText(value);
+        textField.setResponder(s -> {
+            if (!s.isEmpty()) {
+                try {
+                    stringConfigValue.set(s);
+                } catch (Exception e) {
+                    e.printStackTrace();
```
diesieben07, line 141: No.

diesieben07 (changes requested):
Your formatting is all wrong, please follow the guidelines.
Moreover, I don't like this:
> boolean, integer, long, double and string 
What about enums? What about lists?

src/main/java/net/minecraftforge/common/ForgeMod.java
AlexiyOrlov: I used it it because the server was crashing. Ok I will just remove this.

diesieben07: https://mcforge.readthedocs.io/en/latest/concepts/sides/

PR #7353, opened 2020-09-24
Add support modded world generation

Won-Ton:
There's basically no way for modders to define new world generators due to the relevant vanilla code being inextensible and/or hardcoded.

**The problem:**
- On the client, mods have no way to add their custom chunk generators as selectable options in the world creation menu:
  - BiomeGeneratorTypeScreens has a private constructor
  - The list of BiomeGeneratorTypeScreens isn't safe to modify inside ClientSetup, an the map of BiomeGeneratorTypeScreens to EditScreens is immutable
  - There's a couple of other access issues that would cause problems for mods such as updating the generator's settings on closing the EditScreen
- On the server the chunk generator types are hardcoded in a switch statement so users can't use mod generator names as a value on the `level-type` entry in server.properties

**This PR adds:**
- the LevelType class (similar to WorldType in older versions) and new ForgRegistries registry for it.
- LevelType implementations for the vanilla generation options
- the LevelTypeOptions class used by in the ui patches to interact with and display level-types
- a server-side hook to try and use the new LevelType class for creating new worlds if an entry exists for the  `level-type` property in server.properties (falls back to the vanilla way if it fails)
- the `LevelTypeEvent.DefaultLevel` event for modifying the default LevelType, and `LevelTypeEvent.CreateLevel` for modifying the initial Dimension registry used by the world
- a test mod to demonstrate the above features


src/main/java/net/minecraftforge/client/GeneratorTypeManager.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    /**
+     * Set the default GeneratorType
+     *
+     * @param generatorType The GeneratorType to use as the default
+     */
+    public synchronized void setDefaultGeneratorType(@Nonnull BiomeGeneratorTypeScreens generatorType) {
+        this.defaultGeneratorType = generatorType;
+    }
```
srs-bsns: This is probably a bad idea in that a default is a single object, and it is in an environment with multiple actors.
This is nothing more than a race condition where multiple mods will be competing to be the loser to set the default.
The default should be hard-set to the vanilla default (`field_239066_a_`) only.

Cyborgmas: I'd say its fine, if (for example) modpacks want to have a default value for some reason. I would perhaps just add logging after the first override.

alcatrazEscapee: To second cyborg - there is no possible *breaking conflict* that would be caused by having two mods compete for a default. One would win, the other would be just fine. If anything, this avoids two mods doing the same mixin / coremod (of which I am doing in lieu of this PR) to do this which is going to be much worse if two mods do the same thing. This pattern exists all over (see: registry replacement, recipe / json overwriting, etc.)

And as to why this would be useful, for people writing complete world gen overhauls, I get a lot of bug reports and confused users by "I installed this mod that's supposed to completely change world gen but I got a vanilla world". This kind of default helps deal with that.

src/main/java/net/minecraftforge/client/GeneratorTypeManager.java
```diff
@@ -0,0 +1,196 @@
...
+        return true;
+    }
+
+    public static GeneratorTypeManager get() {
+        return INSTANCE;
+    }
+
+    // Used to access the list/maps of GeneratorTypes in BiomeGeneratorTypeScreens so that they can be
+    // copied to the manager.
+    private static class VanillaGeneratorTypeAccess extends BiomeGeneratorTypeScreens {
```
alcatrazEscapee: A possible simplification: this class exists only to access two fields, `field_239068_c_` and `field_239069_d_`, both of which could just be AT'd to public instead. This feels like a very roundabout way to do that, using the quirk of a protected static field. It's also named as if it was a mixin `@Accessor` which confused me at first. Might be simpler just to AT both fields as there's not really any potential implementation hiding since any mod subclassing `BiomeGeneratorTypeScreens` will be able to access them anyway.

Won-Ton: True. In my personal projects as a rule of thumb if I can achieve something without AT's I go with that even if it's a bit jank-looking. I can see how simple == maintainable might be preferable in this project.

src/main/java/net/minecraftforge/client/GeneratorTypeManager.java
Won-Ton: Huh, thought I put note about about 'no guarantees' on the javadoc.

But yeah, this is something I've seen quite a few mods do with coremods below 1.16. Only one can win out in that situation too (assuming they haven't exploded trying).

src/main/java/net/minecraftforge/client/GeneratorTypeManager.java
Won-Ton: Honestly, if anything, I feel like these fields should be made private and their contents copied to the GeneratorTypeManager in a static block or something. This'd make the patch larger and possibly be considered breaking.

Making something public that no-one else is supposed to use doesn't _feel_ like the right move either. Not sure what the correct way forward is :/ More guidance would be appreciated

src/main/java/net/minecraftforge/client/GeneratorTypeManager.java
Won-Ton: A second idea might be to say whichever entry is earliest in the list wins as default. That way the number of clicks to select the loser's option is minimized (were it the other way round the user would have to cycle back round through the options to get to a previous spot)

alcatrazEscapee:
One possible additional consideration for this PR: can the existing infrastructure be adapted to *also* work for servers?

Basically in order to add a hook to server creation based off the `level-type` field, all that would be necessary would be to patch `DimensionGeneratorSettings.func_242753_a`, and delegate any unknown `level-type` fields to a system like this.

The Manager class could be expanded to provide both possible objects (after all, the `BiomeGeneratorTypeScreens` is just a `ChunkGenerator` factory, same as the hook for server world creation, and could be treated as such) and be side agnostic.

Won-Ton:
Oh lordy that method... I agree, makes sense to try come up with something unified 👍

src/main/java/net/minecraftforge/client/GeneratorTypeManager.java
Won-Ton: Marking as resolved as this is now outdated. I've added the logging as suggested in the changes

src/main/java/net/minecraftforge/client/GeneratorTypeManager.java
Won-Ton: Marking as resolved as this now outdated, however, I've stuck with the pattern in the revisions (see GeneratorOptionList) so feel free to raise again for discussion/guidance

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -216,6 +218,12 @@ public void missingSoundMapping(RegistryEvent.MissingMappings<SoundEvent> event)
         }
     }
 
+    @SubscribeEvent
+    public void registerLevelTypes(RegistryEvent.Register<LevelType> event)
+    {
+        LevelTypes.forEach(event.getRegistry()::register);
+    }
```
Won-Ton, line 226: Not sure if this is the appropriate way to do this. With there being no vanilla registry to wrap, the minecraft namespaced instances need to be added somewhere. Guidance would be appreciated.

Won-Ton:
rip

PR #7355, opened 2020-09-25
[1.16.x] Add event for player changing game mode

Ocelot5836:
**The problem at hand:**
There is currently no way to check if a player's game mode has changed besides having code that caches player states every tick using one of the tick events. This can be slow in the case of having to manage hundreds of players in one tick.
- On server side the event is fired after packets are sent to clients notifying the change.
- On client side the event is fired when the client receives a packet changing the local client player game mode.

**This PR adds:**
- A new event for when the player changes from an old game type to the current type
- Patches to PlayerInteractionManager and NetworkPlayerInfo
- A test mod

**An example of how this event could be used:**
```java
    @SubscribeEvent
    public void onEvent(PlayerEvent.PlayerChangeGameModeEvent event)
    {
        // Event is only fired when a player's game mode changes, 
        // which for this specific case is usually once every 10-20 minutes
        PlayerEntity player = event.getPlayer();
        if (this.gameRunning && event.getNewGameMode() != GameType.ADVENTURE)
            this.killPlayer(player, DamageSource.GENERIC);
    }
```
**Previously had to do something similar to:**
```java
    private final Map<UUID, GameType> lastPlayerGameTypes = new HashMap<>();

    @SubscribeEvent
    public void onEvent(TickEvent.ServerTickEvent event)
    {
        // Basic implementation of checking
        this.gameManager.getServer().getPlayerList().getPlayers().forEach(player ->
        {
            UUID uuid = player.getUniqueID();
            GameType gameType = player.interactionManager.getGameType();
            if (this.lastPlayerGameTypes.getOrDefault(uuid, GameType.NOT_SET) != gameType)
            {
                // Mark player as already being checked
                this.lastPlayerGameTypes.put(player.getUniqueID(), player.interactionManager.getGameType());
               
                // Process player
                if (this.gameRunning && gameType != GameType.ADVENTURE)
                    this.killPlayer(player, DamageSource.GENERIC);
            }
        });
    }

    @SubscribeEvent
    public void onEvent(PlayerEvent.PlayerLoggedOutEvent event)
    {
        // Required to make sure the code will be executed consistently
        this.lastPlayerGameTypes.remove(event.getPlayer().getUniqueID()); 
    }
```


src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -510,4 +511,30 @@ public PlayerChangedDimensionEvent(PlayerEntity player, RegistryKey<World> fromD
             return this.toDim;
         }
     }
+
+    /**
+     * On the server side fired when any player has their game mode changed via {@link net.minecraft.server.management.PlayerInteractionManager#setGameType(GameType)}.
```
diesieben07: This "via" info makes no sense, the event docs are in need of changes and new events should not follow the existing guides.
You should describe when the event fires, not which method is responsible for it.

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -510,4 +511,30 @@ public PlayerChangedDimensionEvent(PlayerEntity player, RegistryKey<World> fromD
...
+     * On the client side fired when only when the client player changes their game mode.
+     */
+    public static class PlayerChangeGameModeEvent extends PlayerEvent {
```
diesieben07: Formatting (braces on newlines)

patches/minecraft/net/minecraft/client/network/play/NetworkPlayerInfo.java.patch
```diff
@@ -0,0 +1,13 @@
+--- a/net/minecraft/client/network/play/NetworkPlayerInfo.java
++++ b/net/minecraft/client/network/play/NetworkPlayerInfo.java
+@@ -50,7 +50,9 @@
+    }
+ 
+    protected void func_178839_a(GameType p_178839_1_) {
+-      this.field_178866_b = p_178839_1_;
++       GameType oldGameType = this.field_178866_b;
++       this.field_178866_b = p_178839_1_;
++       if(Minecraft.func_71410_x().field_71439_g != null && this.field_178867_a.getId().equals(Minecraft.func_71410_x().field_71439_g.func_110124_au())) net.minecraftforge.common.ForgeHooks.onChangeGameMode(Minecraft.func_71410_x().field_71439_g, p_178839_1_, oldGameType);
```
diesieben07: 1. Formatting (don't change indentation)
2. Extract this logic into the `ForgeHooks` method, that's what its there for. You can have this be a much simpler patch by doing:
`this.gameType = ForgeHooks.onClientChangeGameType(gameMode, this.gameType, ...)`

diesieben07 (changes requested):
Some extra thoughts:
- Maybe this should be cancelable? At least on the server.
- Maybe server and client event should be separate? That way you could always fire the client event and just not include the `PlayerEntity`.

patches/minecraft/net/minecraft/client/network/play/NetworkPlayerInfo.java.patch
```diff
@@ -0,0 +1,12 @@
+--- a/net/minecraft/client/network/play/NetworkPlayerInfo.java
++++ b/net/minecraft/client/network/play/NetworkPlayerInfo.java
+@@ -50,7 +50,8 @@
+    }
+ 
+    protected void func_178839_a(GameType p_178839_1_) {
+-      this.field_178866_b = p_178839_1_;
++       if(Minecraft.func_71410_x().field_71439_g != null && this.field_178867_a.getId().equals(Minecraft.func_71410_x().field_71439_g.func_110124_au()) && this.field_178866_b != p_178839_1_) net.minecraftforge.common.ForgeHooks.onClientChangeGameMode(Minecraft.func_71410_x().field_71439_g, this.field_178866_b, p_178839_1_);
++       this.field_178866_b = p_178839_1_;
```
diesieben07: Indentation still wrong. Also missing space after `if`.
Also, my point of separating client and server event was so that you can fire the event on the client _always_, by not including the `PlayerEntity`.

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
```diff
@@ -130,15 +130,23 @@
     }
  
     protected void func_70670_a(EffectInstance p_70670_1_) {
-@@ -1187,6 +1208,7 @@
+@@ -1139,6 +1160,7 @@
+    }
+ 
+    public void func_71033_a(GameType p_71033_1_) {
++      if(!net.minecraftforge.common.ForgeHooks.onChangeGameMode(this, this.field_71134_c.func_73081_b(), p_71033_1_)) return;
```
diesieben07: Formatting, space after `if`.

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -510,4 +511,65 @@ public PlayerChangedDimensionEvent(PlayerEntity player, RegistryKey<World> fromD
             return this.toDim;
         }
     }
+
+    /**
+     * Fired when the game type of a server player is changed to a unique value before listeners are fired.
```
diesieben07: I have no idea what you mean to say with this sentence. "changed to a unique value"? And what listeners are you talking about?
You also need to add a note about what happens when the event is cancelled.

diesieben07 (changes requested):

patches/minecraft/net/minecraft/client/network/play/NetworkPlayerInfo.java.patch
Ocelot5836: When the event is fired client side without the player check there are two problems:

1. The event is fired for all other tracked players of the local client. If a player is too far away it won't be fired.
2. Out of the tracked players only the ones not in spectator are fired. If another player comes out of spectator mode the local client won't know about the change and the event is not fired.
3. Sometimes neither of these conditions are met and the event *still* fires.
4. A reliable alternative is to just have the mod send a packet to the client about their mod specific status updates. Because the server knows about all players, the mod dev can control the scope of the event.

By adding the player, the goal is to add consistent results for when the event is fired.

diesieben07: That is not true. The packet is sent to all players.
Have you actually verified these claims you are making?

Ocelot5836: I just redid me experiment and found different results. I think it may have been due to poor placement of one of the patches but my claims are not true. All players are always notified of all game mode changes, even for players they have never or currently cannot see.

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
diesieben07: You have not fixed this.

diesieben07 (changes requested):

diesieben07 (approved):

PR #7356, opened 2020-09-25
Background Music Selection

ChampionAsh5357:
This PR adds exactly what the title says, a way to set the background music currently playing in Minecraft. This hooks into `MusicTicker` where the selector method is called.
There is also a test mod included that logs the available sounds and sets the current selector to the end credits to verify.


src/main/java/net/minecraftforge/client/event/sound/BackgroundMusicSelectionEvent.java
```diff
@@ -0,0 +1,94 @@
...
+import net.minecraft.client.Minecraft;
+import net.minecraft.client.audio.BackgroundMusicSelector;
+import net.minecraft.client.audio.ISound;
+import net.minecraft.client.audio.MusicTicker;
+import net.minecraft.client.audio.SimpleSound;
+import net.minecraftforge.eventbus.api.Event;
+
+/**
+ * Fired whenever {@link MusicTicker} tries to get the current
+ * {@link BackgroundMusicSelector} for the current situation.
```
diesieben07: Double "current".

src/main/java/net/minecraftforge/client/event/sound/BackgroundMusicSelectionEvent.java
```diff
@@ -0,0 +1,94 @@
...
+ * 
+ * When the setter below is called, the current background music selector
+ * is replaced with the user set one. The selector will wait until the current
+ * music is finished playing followed by a random delay specified by the new
+ * selector or if {@link BackgroundMusicSelector#shouldReplaceCurrentMusic}
+ * on the new selector is true.
+ * 
+ * If the setter is never called, the background music selector will default to
+ * the client call from {@link Minecraft#getBackgroundMusicSelector}.
```
diesieben07: This is a very roundabout way of explaining this. And it does not really belong in this event. This event should simply be "okay, what music should be playing right now".
Also, those `{@link}`s are incorrect (wrong name for the methods with Forge's current mappings).

src/main/java/net/minecraftforge/client/event/sound/BackgroundMusicSelectionEvent.java
```diff
@@ -0,0 +1,94 @@
...
+ * */
+public class BackgroundMusicSelectionEvent extends Event
+{
+    private final MusicTicker musicTicker;
+    private final Minecraft mc;
```
diesieben07: Why is this needed here?

src/main/java/net/minecraftforge/client/event/sound/BackgroundMusicSelectionEvent.java
```diff
@@ -0,0 +1,94 @@
...
+    public BackgroundMusicSelectionEvent(MusicTicker musicTicker, Minecraft mc, @Nullable ISound currentMusic, int timeUntilNextMusic, BackgroundMusicSelector _default)
+    {
+        this.musicTicker = musicTicker;
+        this.mc = mc;
+        this.currentMusic = currentMusic;
+        this.timeUntilNextMusic = timeUntilNextMusic;
+        this.defaultSelector = _default;
+    }
+
+    /**Gets the music ticker that handled background music.*/
```
diesieben07: Please use proper Javadoc formatting.
This doc is also pointless.

src/main/java/net/minecraftforge/client/event/sound/BackgroundMusicSelectionEvent.java
```diff
@@ -0,0 +1,94 @@
...
+ * If the setter is never called, the background music selector will default to
+ * the client call from {@link Minecraft#getBackgroundMusicSelector}.
+ * */
+public class BackgroundMusicSelectionEvent extends Event
+{
+    private final MusicTicker musicTicker;
+    private final Minecraft mc;
+    @Nullable
+    private final ISound currentMusic;
+    private final int timeUntilNextMusic;
```
diesieben07: Why is this needed? Which music should be played does not depend on this field at all.

src/main/java/net/minecraftforge/client/event/sound/BackgroundMusicSelectionEvent.java
```diff
@@ -0,0 +1,94 @@
...
+     * has finished playing or if the selector instantly
+     * replaces the current music being played.
+     * */
+    public int getTimeUtilNextMusic()
+    {
+        return this.timeUntilNextMusic;
+    }
+
+    /**Gets the default selector if the setter is not called.*/
+    public BackgroundMusicSelector getCurrentBackgroundMusicSelector()
```
diesieben07: Should be `getDefaultSelector`

src/main/java/net/minecraftforge/client/event/sound/BackgroundMusicSelectionEvent.java
```diff
@@ -0,0 +1,94 @@
...
+ * */
+public class BackgroundMusicSelectionEvent extends Event
+{
+    private final MusicTicker musicTicker;
+    private final Minecraft mc;
+    @Nullable
+    private final ISound currentMusic;
+    private final int timeUntilNextMusic;
+    private final BackgroundMusicSelector defaultSelector;
+    private BackgroundMusicSelector newSelector;
```
diesieben07: This should be `defaultSelector` by default, which saves you all the null checks.

diesieben07 (changes requested):
Why did you patch this in `MusicTicker` instead of patching `Minecraft#func_238178_U_`?
Your indentation is wonky in places (using tabs).
Also please use the same names for things. Sometimes you call it `_default`, sometimes you call it `defaultSelector`, sometimes you call it "currentBackgroundMusicSelector`.

ChampionAsh5357:
> Why did you patch this in `MusicTicker` instead of patching `Minecraft#func_238178_U_`?

This is the only place where the method is called and it allows the user to switch based on the selector output as well. However, I could always remove the call and do an instant check at the head of the method. The default wouldn't be known then though.

> Your indentation is wonky in places (using tabs).

Yeah, I figured that would be the case. I can check this a million times and still miss it.

> Also please use the same names for things. Sometimes you call it `_default`, sometimes you call it `defaultSelector`, sometimes you call it "currentBackgroundMusicSelector`.

Ok, will standardize.

diesieben07:
> This is the only place where the method is called and it allows the user to switch based on the selector output as well. However, I could always remove the call and do an instant check at the head of the method. The default wouldn't be known then though.

Yeah you are right, because that method uses early returns. Probably better the way you did this.

src/main/java/net/minecraftforge/client/event/sound/BackgroundMusicSelectionEvent.java
ChampionAsh5357: It doesn't. Removed the reference to music ticker as well since none of those methods should directly be called here. The only thing that should be specified is whether to use the default or the user set selector.

src/main/java/net/minecraftforge/client/event/sound/BackgroundMusicSelectionEvent.java
ChampionAsh5357: Redone it so that the default selector is on the modifiable field and can be set to the returned value. Saves the need for two separate fields.

ichttt (approved):

diesieben07 (approved):

diesieben07:
This shouldn't be a tick event.
Please explain what you are trying to achieve with this event and try to rework the event so that it maybe only fires once a song finishes playing.

ChampionAsh5357:
Thanks for the explanation.
So here is the background music selector tasks:
- [x] Play music (duh)
- [x] Replace current music immediately
- [x] Specify a delay timer with min and max delay

Issue is here that the immediate count makes the music ticker check every tick which stops caching. Also, the delay with min max makes it difficult to hook in when a song finishes playing as the delay might overextend or vice versa.

So, just as a brainstorm, what if instead of an event, additional music selectors are added as a registry that holds some sort of boolean function -> music selector. The event could then be delegated to just being used to replace vanilla music in a specific circumstance when being selected only getting called once the song finishes and the time expires (e.g. MusicTicker#selectRandomBackgroundMusic). Let me know how this sounds.

diesieben07:
A registry of sorts does sound like a plausible option.

ChampionAsh5357:
I removed the event altogether since it makes relatively no sense to use it. It breaks down if a sound overrides a vanilla selector that replaces the current music. The vanilla music will always go first regardless. The replacement is just a simple map for a selector -> condition. The condition takes an instance of the vanilla selector to default to and to pass into the condition.

There is still a relative problem of a O(n) when handling because of it being called every tick. However, it's much better than using the event dispatcher.

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -96,6 +96,7 @@
 import org.lwjgl.opengl.GL13;
 
 import javax.annotation.Nonnull;
+
```
pupnewfster: No reason for this whitespace only change in this file.

ChampionAsh5357: Agh, thanks for the catch.

sciwhiz12:
Hello! @ChampionAsh5357, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

ChampionAsh5357:
Probably gonna close this for now. If I have more time, I may come back to it later.

PR #7357, opened 2020-09-25
Add codec extra objects, implementable on any codec and implement it on DimensionType

Cyborgmas:
Dimension types are pretty much a random collection of data about a dimension, however that "random collection" is fixed. This PR adds a small "type" serializer system, which allows for any amount of data to be linked to the dimension type and be modifiable by users, using a datapack.


src/main/java/net/minecraftforge/common/world/IDimensionExtra.java
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Errors if the list contains 2 objects with the same "type" key.
+     */
+    static DataResult<List<IDimensionExtra>> ensureUnique(List<IDimensionExtra> extras)
+    {
+        Set<ResourceLocation> toTest = new HashSet<>();
+        for(IDimensionExtra extra : extras)
+        {
+            if(!toTest.add(extra.getType().getRegistryName()))
+                return DataResult.error("Duplicate founds in forge extras!", extras);
```
alcatrazEscapee:
```suggestion
                return DataResult.error("Duplicate dimension type extra entries were found: " + extra.getType().getRegistryName(), extras);
```
Mojang's DFU errors are already horrible to understand, providing a little more information would be super nice here. Even better if the dimension type registry name could be appended to this string to identify *where* the extras were found.

alcatrazEscapee (commented):
So based on what I see here, this seems very similar to the system of biome extensions I had written up for #7364 - if I remove the extra patches for method context, it's essentially the same extension. Would it be worth consolidating these two "extra/extensions" variants into one PR that hits all the dynamic registry objects? That way there could be one consistient style (`XExtra` vs `FooExtension`, the key of the forge extra/extension field in json, etc.)

Another thing to consider would just be having *one* extension type - "dynamic registry extension", which can be applied to any dynamic registry and reduce duplicating this across all of them. The codec could have a verification step where each `DynamicRegistryExtensionType` reports what objects (Biomes, dimension types, etc.) it would be valid to attach to.

src/main/java/net/minecraftforge/common/world/IDimensionExtra.java
Cyborgmas: well the dimension type is being created when this called. so no real way to know what its registry name is.

Cyborgmas:
Problem with doing trying to do something for all worldgen objects is it would have to be an interface right? And we need to modify the constructor for this...

A way around that could be to have setters/getters in the interface and have a reference equality map (==) that could be used to hold the extras. this would require parsing the extras after the object (say a biome) is created which is possible just gotta take the time to figure it out.
I could give it a go in the future (not near :P), and get back to you

patches/minecraft/net/minecraft/world/DimensionType.java.patch
```diff
@@ -0,0 +1,54 @@
...
+       }
+    }
++
++   public java.util.List<net.minecraftforge.common.world.IDimensionExtra> getExtras(){ return extras; }
++   public Optional<net.minecraftforge.common.world.IDimensionExtra> getExtra(ResourceLocation name) {
++      return getExtras().stream().filter(e -> e.getType().getRegistryName().equals(name)).findAny();
++   }
++   public Optional<net.minecraftforge.common.world.IDimensionExtra> getExtra(net.minecraftforge.common.world.DimensionExtraType<?> codec) {
++      return getExtras().stream().filter(e -> e.getType().equals(codec)).findAny();
++   }
```
alcatrazEscapee: This could do the unchecked cast here to make the return type nicer. Every time a modder uses this they're going to have to cast down from `IDimensionExtra` anyway, and if you only allow querying by type it's even simpler.

```java
public <E extends net.minecraftforge.common.world.IDimensionExtra> Optional<E> getExtra(net.minecraftforge.common.world.DimensionExtraType<E> type) {
    return getExtras().stream().filter(e -> e.getType().equals(type)).findAny().map(e -> (E) e);
}
```

alcatrazEscapee (commented):
The patches would need to be dimension-specific, yes. But unless I've missed something, there's nothing in `DimensionExtraType` or `IDimensionExtra` that is actually specific to dimension types. They both quite easily could be renamed and reused for all manner of registry objects, and have the advantage that you'd be making the system for adding extensions consistent across all registry types, both from the modder perspective, and from the datapack exposed API.

patches/minecraft/net/minecraft/world/DimensionType.java.patch
alcatrazEscapee: Actually if you do this it might as well be a `Map<DimensionExtraType, IDimensionExtra>` if you're already requiring them to be unique per type.

Cyborgmas:
So abstracted it to a Codec Extra object, seems to work pretty well. Would need to check that nothing weird happens when there are plenty of objects or with overrides.

patches/minecraft/net/minecraft/world/DimensionType.java.patch
```diff
@@ -0,0 +1,32 @@
...
++      }), Codec.BOOL.fieldOf("has_skylight").forGetter(DimensionType::hasSkyLight), Codec.BOOL.fieldOf("has_ceiling").forGetter(DimensionType::func_236037_d_), Codec.BOOL.fieldOf("ultrawarm").forGetter(DimensionType::func_236040_e_), Codec.BOOL.fieldOf("natural").forGetter(DimensionType::func_236043_f_), Codec.doubleRange((double)1.0E-5F, 3.0E7D).fieldOf("coordinate_scale").forGetter(DimensionType::func_242724_f), Codec.BOOL.fieldOf("piglin_safe").forGetter(DimensionType::func_241509_i_), Codec.BOOL.fieldOf("bed_works").forGetter(DimensionType::func_241510_j_), Codec.BOOL.fieldOf("respawn_anchor_works").forGetter(DimensionType::func_241511_k_), Codec.BOOL.fieldOf("has_raids").forGetter(DimensionType::func_241512_l_), Codec.intRange(0, 256).fieldOf("logical_height").forGetter(DimensionType::func_241513_m_), ResourceLocation.field_240908_a_.fieldOf("infiniburn").forGetter((p_241508_0_) -> {
```
Cyborgmas: I have no idea why this is here... tried deleting the patch file completely then reran gradlew setup. After that ran gradlew genPatches and got nothing...
I add the lines back in (not touching this whatsoever ofc...) and this diff is still here... will debug more later.

marchermans: Looks like methods got renamed.

sciwhiz12:
Hello! @Cyborgmas, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after a few days, this PR will be closed for inactivity.

Cyborgmas:
yeah, will try to update soon

alcatrazEscapee:
Might I suggest the use cases of this PR are better off addressed by #8263. It's a more generic solution to this problem (Defining extra world generation datapack types) that can be simply addressed by matching resource keys (i.e. if you actually want something that is specific to dimension types, or biomes), that is also much more composable (i.e. if two mods add a dimension type extension, that still has to be defined in the same json file)

Cyborgmas:
Not exactly sure how, Would a mod create a new datapack object (say dimension_type_extension), and that would be composable if mods all agree on the same name?

alcatrazEscapee:
As is, this PR gives mods the following capabilities:

- For a dimension type, mod X may "attach" some arbitrary extra data that can be defined in json.
- At runtime, mods may query the dimension type to obtain this extra data.
- Multiple mods can attach different pieces of data and all may be loaded.

With #8263, here's how you'd accomplish the same thing:

- mod X defines a new world gen registry type, "mod_x_dimension_type_data", which can be defined in data
  - This data, in addition to the data itself, contains a registry name of a dimension type
- At runtime, all mods may query the dimension type's **name**, then cross reference that with a corresponding entry in the "mod_x_dimension_type_data" to obtain this extra data
- Multiple mods may attach different pieces of data, by using their own registries (mod_x_..., mod_y_...), and they may all be loaded.

---

But, moreover, the capabilities of #8263 are much more broad and usable than this PR is.

- It's not limited to just defining data on a single dimension type, it can be defined for any world gen object (via the "referencing" method described above)
- It can define completely custom data (a use case not covered by this PR)
- It is generally more composable (imagine if two mods define the same extension, but they are overriding the same file - one will win. As opposed to two mods defining different extra data registries - both may coexist and one can be overridden at a time.

Cyborgmas:
Okay great thanks for the explainer, does seem like a better way.

PR #7360, opened 2020-09-26
Add modded biomes to overworld biome list

maxanier:
## Why?
The `BiomeProvider` class has a list containing all biomes that might be generated in the related dimension.
This list is used for several things:
- Determine if a structure might be generated in this dimension (used by the locate command)
- Determine possible surface blocks in the dimension (used e.g. when looking for a world spawn point) 
- Used for Stronghold generation
- Potentially used by mods as this list is publicly available
This use cases where deduced from code, the only thing I tested is:
A structure (that only generates in a modded biome) cannot be found using the `/locate` command unless the biome is added to the list.

Currently one has to AT that list and replace it using a modified copy (using the **sequential** mod loading queue to avoid race conditions).

## Implementation
This PR simply adds a list to Forge's `BiomeManager` that modders can add biomes to and which is merged with the vanilla biome list in the `OverworldBiomeProvider` on construction.
Biomes added as `BiomeEntry` to the `BiomeManager` are automatically added to this list as well. Additionally, one can manually add biomes in case they are generated in a different way (e.g. as hill variant of another biome).

### DIFF
--- a/src/main/java/net/minecraftforge/common/BiomeManager.java
+++ b/src/main/java/net/minecraftforge/common/BiomeManager.java
@@ -34,6 +34,7 @@ import net.minecraft.world.biome.Biome;
 public class BiomeManager
 {
     private static TrackedList<BiomeEntry>[] biomes = setupBiomes();
+    private static final List<RegistryKey<Biome>> additionalOverworldBiomes = new ArrayList<>();
 
     private static TrackedList<BiomeEntry>[] setupBiomes()
     {
@@ -97,11 +98,23 @@ public class BiomeManager
     }
     */
 
+    /**
+     * Add biomes that you add to the overworld without using {@link BiomeManager#addBiome(BiomeType, BiomeEntry)}
+     */
+    public static void addAdditionalOverworldBiomes(List<RegistryKey<Biome>> biomes){
+        additionalOverworldBiomes.addAll(biomes);
+    }
+
     public static boolean addBiome(BiomeType type, BiomeEntry entry)
     {
         int idx = type.ordinal();
         List<BiomeEntry> list = idx > biomes.length ? null : biomes[idx];
-        return list == null ? false : list.add(entry);
+        if(list != null)
+        {
+            additionalOverworldBiomes.add(entry.key);
+            return list.add(entry);
+        }
+        return false;
     }
 
     public static boolean removeBiome(BiomeType type, BiomeEntry entry)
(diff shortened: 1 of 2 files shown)


src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -97,11 +98,26 @@ public static void removeSpawnBiome(Biome biome)
     }
     */
 
+    /**
+     * Add biomes that you add to the overworld without using {@link BiomeManager#addBiome(BiomeType, BiomeEntry)}
+     */
+    public static void addAdditionalOverworldBiomes(RegistryKey<Biome> biome){
+        if (!"minecraft".equals(biome.func_240901_a_().getNamespace()) && additionalOverworldBiomes.stream().noneMatch(entry -> entry.compareTo(biome) == 0))
```
maxanier, line 105: Unfortunately, `RegistryKey` does not support `equals`.

src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -97,11 +98,26 @@ public static void removeSpawnBiome(Biome biome)
     }
     */
 
+    /**
+     * Add biomes that you add to the overworld without using {@link BiomeManager#addBiome(BiomeType, BiomeEntry)}
+     */
+    public static void addAdditionalOverworldBiomes(RegistryKey<Biome> biome){
+        if (!"minecraft".equals(biome.func_240901_a_().getNamespace()) && additionalOverworldBiomes.stream().noneMatch(entry -> entry.func_240901_a_().equals(biome.func_240901_a_())))
```
maxanier, line 105: Unfortunately, `RegistryKey` does not support `equals`.

pupnewfster: That is true but I believe you can use `RegistryKey#func_244356_a` to compare them which is likely to make this line a bit cleaner

maxanier: That method does something weird/different. It compares the registry name/id with a registry key of a registry, I think?

pupnewfster: Ah you are correct I misread which resource locations that method compares.

TelepathicGrunt (approved):
I like what I see here. Especially the biome list concat in the OverworldBiomeProvider as this fixes a Forge bug that I am certain is still around today. The bug was that if you add a structure only to modded biomes and add those biomes to the overworld using Forge's biome manager, the structure will not spawn at all. This is because forge forgot to add modded biomes to that list you patched and that list is used by MC to know if a structure can even spawn in the world before it start checking spawn locations and all. So this PR fixes a pretty weird and nasty bug too!

LexManos:
Honestly the entire BiomeManager needs to die, and people need to use the data driven system. But that's beating a dead horse that will never change unless Mojang gets the concept that their current data system is inflexible as hell.

PR #7361, opened 2020-09-26
Adding LivingBurnBySunlightEvent

PellucidDice:
Adding LivingBurnBySunlightEvent, which triggers in LivingTick() in ZombieEntity, PhantomEntity, and AbstractSkeletonEntity.
This event is based on the LivingEvent, with an extra parameter "duration" which stands for the ticks the entity is trying to burn.
This event allows users to change the duration of the burn effect apply to the entity by sunlight, including setting it to 0 without putting out existing burn effect that can be caused by some other source other than sunlight.

An example mod is added, it sets the burn effect by sunlight on stray to 0.


patches/minecraft/net/minecraft/entity/monster/AbstractSkeletonEntity.java.patch
```diff
@@ -1,6 +1,26 @@
 --- a/net/minecraft/entity/monster/AbstractSkeletonEntity.java
 +++ b/net/minecraft/entity/monster/AbstractSkeletonEntity.java
-@@ -155,7 +155,7 @@
+@@ -45,6 +45,8 @@
+ import net.minecraft.world.DifficultyInstance;
+ import net.minecraft.world.IServerWorld;
+ import net.minecraft.world.World;
++import net.minecraftforge.common.MinecraftForge;
++import net.minecraftforge.event.entity.living.LivingBurnBySunlightEvent;
```
Cyborgmas: Cant have import in patches.

patches/minecraft/net/minecraft/entity/monster/AbstractSkeletonEntity.java.patch
```diff
@@ -1,6 +1,26 @@
...
+ public abstract class AbstractSkeletonEntity extends MonsterEntity implements IRangedAttackMob {
+    private final RangedBowAttackGoal<AbstractSkeletonEntity> field_85037_d = new RangedBowAttackGoal<>(this, 1.0D, 20, 15.0F);
+@@ -109,7 +111,9 @@
+          }
+ 
+          if (flag) {
+-            this.func_70015_d(8);
++            LivingBurnBySunlightEvent burnEvent = new LivingBurnBySunlightEvent(this, 8);
++            MinecraftForge.EVENT_BUS.post(burnEvent);
++            this.func_70015_d(burnEvent.getDuration());
```
Cyborgmas: Make this a method in ForgeEventFactory instead, to have a one line patch instead of 3 each time.

Cyborgmas (commented):
Only commented on one file, but it applies to the others as well
Also the genPatches task shouldve errored and tell you you had imports, best to read what it says :)

ChampionAsh5357:
Adding onto what Cyborgmas said, you should also javadoc the event so that every knows what it does and how to properly use it.

As a side note, why do you allow the user to change the duration instead of just making the event cancelable and letting the user handle their own logic? At least it should have the annotation of `@HasResult` and handle whether to default and do the original intention, deny to prevent it, or allow and set your own value. This is my opinion anyways.

PellucidDice:
Sounds workable, thanks for the suggestions.

patches/minecraft/net/minecraft/entity/monster/AbstractSkeletonEntity.java.patch
```diff
@@ -1,6 +1,23 @@
 --- a/net/minecraft/entity/monster/AbstractSkeletonEntity.java
 +++ b/net/minecraft/entity/monster/AbstractSkeletonEntity.java
-@@ -155,7 +155,7 @@
+@@ -45,6 +45,7 @@
+ import net.minecraft.world.DifficultyInstance;
+ import net.minecraft.world.IServerWorld;
+ import net.minecraft.world.World;
++import net.minecraftforge.event.ForgeEventFactory;
```
ChampionAsh5357: Use the fully qualified name to call your method instead of the `import` keyword.

patches/minecraft/net/minecraft/entity/monster/PhantomEntity.java.patch
```diff
@@ -1,6 +1,23 @@
 --- a/net/minecraft/entity/monster/PhantomEntity.java
 +++ b/net/minecraft/entity/monster/PhantomEntity.java
-@@ -214,7 +214,7 @@
+@@ -41,6 +41,7 @@
+ import net.minecraft.world.gen.Heightmap;
+ import net.minecraftforge.api.distmarker.Dist;
+ import net.minecraftforge.api.distmarker.OnlyIn;
++import net.minecraftforge.event.ForgeEventFactory;
```
ChampionAsh5357: Once again, use the fully qualified name to call your method instead of the `import` keyword.

patches/minecraft/net/minecraft/entity/monster/ZombieEntity.java.patch
```diff
@@ -1,6 +1,23 @@
 --- a/net/minecraft/entity/monster/ZombieEntity.java
 +++ b/net/minecraft/entity/monster/ZombieEntity.java
-@@ -268,12 +268,16 @@
+@@ -64,6 +64,7 @@
+ import net.minecraft.world.World;
+ import net.minecraft.world.server.ServerWorld;
+ import net.minecraft.world.spawner.WorldEntitySpawner;
++import net.minecraftforge.event.ForgeEventFactory;
```
ChampionAsh5357: Once again, use the fully qualified name to call your method instead of the `import` keyword.

patches/minecraft/net/minecraft/entity/monster/ZombieEntity.java.patch
```diff
@@ -19,15 +36,15 @@
              for(int l = 0; l < 50; ++l) {
                 int i1 = i + MathHelper.func_76136_a(this.field_70146_Z, 7, 40) * MathHelper.func_76136_a(this.field_70146_Z, -1, 1);
                 int j1 = j + MathHelper.func_76136_a(this.field_70146_Z, 7, 40) * MathHelper.func_76136_a(this.field_70146_Z, -1, 1);
-@@ -284,6 +288,7 @@
+@@ -284,6 +289,7 @@
                 if (WorldEntitySpawner.func_209382_a(entityspawnplacementregistry$placementtype, this.field_70170_p, blockpos, entitytype) && EntitySpawnPlacementRegistry.func_223515_a(entitytype, serverworld, SpawnReason.REINFORCEMENT, blockpos, this.field_70170_p.field_73012_v)) {
                    zombieentity.func_70107_b((double)i1, (double)j1, (double)k1);
                    if (!this.field_70170_p.func_217358_a((double)i1, (double)j1, (double)k1, 7.0D) && this.field_70170_p.func_226668_i_(zombieentity) && this.field_70170_p.func_226669_j_(zombieentity) && !this.field_70170_p.func_72953_d(zombieentity.func_174813_aQ())) {
 +                     if (livingentity != null)
```
ChampionAsh5357: This shouldn't be in the patch. I'm not sure why it was added.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -748,4 +749,11 @@ public static void onCommandRegister(CommandDispatcher<CommandSource> dispatcher
         MinecraftForge.EVENT_BUS.post(evt);
         return evt;
     }
+
+    public static int onLivingBurnBySunlight(LivingEntity living, int duration)
+    {
+        LivingBurnBySunlightEvent event = new LivingBurnBySunlightEvent(living, duration);
+        MinecraftForge.EVENT_BUS.post(event);
+        return event.getResult() == Result.DEFAULT ? duration : event.getResult() == Result.ALLOW ? event.getDuration() : 0;
+    }
```
ChampionAsh5357: You replace the line completely in the above patches. It seems a bit pointless to call `Entity#func_70015_d` for if the result is 0. You can handle that internally here.

src/main/java/net/minecraftforge/event/entity/living/LivingBurnBySunlightEvent.java
```diff
@@ -0,0 +1,36 @@
...
+ * LivingBurnBySunlightEvent is triggered when an entity is trying
+ * to burn (caused by sunlight) during {@link LivingEntity#livingTick()}
+ * through {@link net.minecraft.entity.Entity#setFire(int)}
+ *
+ * This event has a result. {@link HasResult}
+ * {@link Result#ALLOW} Burn the entity for custom duration
+ * {@link Result#DENY} Prevent the entity from burning
+ * {@link Result#DEFAULT} Burn the entity for original duration
+ */
+@Event.HasResult
```
ChampionAsh5357: This can just be `@HasResult`.

src/test/java/net/minecraftforge/debug/LivingBurnBySunlightEventTest.java
```diff
@@ -0,0 +1,28 @@
...
+    public LivingBurnBySunlightEventTest()
+    {
+        MinecraftForge.EVENT_BUS.addListener(this::onLivingBurnBySunlight);
+    }
+
+    public void onLivingBurnBySunlight(LivingBurnBySunlightEvent event)
+    {
+        if (event.getEntityLiving() instanceof PhantomEntity)
+            event.setResult(Event.Result.DENY);
+        if (event.getEntityLiving() instanceof StrayEntity)
```
ChampionAsh5357: This is more me being picky, but use an `else if` here. Makes no sense to check if you know it can't be both.

patches/minecraft/net/minecraft/entity/monster/ZombieEntity.java.patch
ChampionAsh5357: Cyborgmas pointed out that this patch line is normal, so you can mark this as resolved.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
PellucidDice: I managed to call `Entity#func_70015_d` in `onLivingBurnBySunlight(),` but I kept the return of duration there.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -748,4 +749,14 @@ public static void onCommandRegister(CommandDispatcher<CommandSource> dispatcher
         MinecraftForge.EVENT_BUS.post(evt);
         return evt;
     }
+
+    public static int onLivingBurnBySunlight(LivingEntity living, int duration)
```
ChampionAsh5357: This return time has become pointless now if everything is handled internally. Should be `void`.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -748,4 +749,14 @@ public static void onCommandRegister(CommandDispatcher<CommandSource> dispatcher
...
+    {
+        LivingBurnBySunlightEvent event = new LivingBurnBySunlightEvent(living, duration);
+        MinecraftForge.EVENT_BUS.post(event);
+        duration = event.getResult() == Result.DEFAULT ? duration : event.getResult() == Result.ALLOW ? event.getDuration() : 0;
+        if (duration != 0)
+            living.setFire(duration);
+        return duration;
```
ChampionAsh5357: This can be handled better imo. It's very messy for what it can be.

Maybe something like if the result isn't deny do logic.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
PellucidDice: The reason that I return the duration there is I think other users may need the updated value when they are making their custom entities.

ichttt: That doesn't make much sense, the modder could just call `getFlameTimer` to get the actual value or create their own version of this method that fires the event. Just return void here.

src/main/java/net/minecraftforge/event/entity/living/LivingBurnBySunlightEvent.java
```diff
@@ -0,0 +1,36 @@
...
+ * LivingBurnBySunlightEvent is triggered when an entity is trying
+ * to burn (caused by sunlight) during {@link LivingEntity#livingTick()}
+ * through {@link net.minecraft.entity.Entity#setFire(int)}
+ *
+ * This event has a result. {@link HasResult}
+ * {@link Result#ALLOW} Burn the entity for custom duration
+ * {@link Result#DENY} Prevent the entity from burning
+ * {@link Result#DEFAULT} Burn the entity for original duration
+ */
+@HasResult
```
ichttt: Not quite sure if this should really have a result. If you call `setDuration`, you 100% want the custom duration, if you don't want the entity to burn set the duration to zero. Maybe cancelable would make sense, but even that is questionable

PellucidDice: I think cancelable does make sense, as setFire(0) will do nothing.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
PellucidDice: Changed the event from HasResult to Cancelable, this makes the factory cleaner.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
PellucidDice: Made the factory return void.

src/test/java/net/minecraftforge/debug/LivingBurnBySunlightEventTest.java
```diff
@@ -0,0 +1,28 @@
...
+{
+    public LivingBurnBySunlightEventTest()
+    {
+        MinecraftForge.EVENT_BUS.addListener(this::onLivingBurnBySunlight);
+    }
+
+    public void onLivingBurnBySunlight(LivingBurnBySunlightEvent event)
+    {
+        if (event.getEntityLiving() instanceof PhantomEntity)
+            event.setResult(Event.Result.DENY);
```
ichttt: This needs to be updated to cancel the event, as it no longer has a result

src/test/java/net/minecraftforge/debug/LivingBurnBySunlightEventTest.java
```diff
@@ -0,0 +1,28 @@
...
+        else if (event.getEntityLiving() instanceof StrayEntity)
+        {
+            event.setResult(Event.Result.ALLOW);
```
ichttt: Same here, as the event is now cancelable, you don't need this

ichttt (changes requested):
Sorry for the massive delay. Mostly fine now, the testmod just needs ajusting

LexManos:
To to love moving, things have slipped through the cracks.
This event isn't "bad" I am just not a fan of adding even more events that get fired literally every tick.
We already have the tick events for that.
Perhaps this would be better suited for an attribute or something that prevents burning during daylight. Kinda like Skeletons can wear helmets.

Why are you trying to make Strays not burn in the sun? Whats the lore behind it? Perhaps we can come up with better way that isn't event spam.

Also the Duration means literally nothing. As this event would be fired every tick, and thus just keep extending the durations.

Perhaps this would be better as a entity attribute, that starts as 0 for most entities, but 8 for vanilla burnable mobs. And then Helmets/Whatever you're wanting can add attribute modifiers to adjust the time/damage? This way the event isn't fired literally every tick for every entity.

PellucidDice:
Hi, sorry for the late reply.
My original target was just to prevent monsters from being burn (by sunlight only) when some different conditions are reached. Entity attribute sounds good, but now I'm struggling what should the value of the attribute represent.
This is my thoughts:
1. Burn duration - Modifiers can modify the duration of burning caused by sunlight, this can be used when you want the monster to get rid of the fire/burn as soon as they find a shelter.
2. Chance of burning - Modifiers can modify the chance for every living entity to burn, default value vanilla mobs should be 0, and monsters that should burn be default will be 1 or 100 (something like that).

The reason I did not include burn damage in it is that the cause of mob burning can be something else other than sunlight (lava, fire block, lightning...etc).
What do you think?

PellucidDice:
Found an alternative way...

PR #7362, opened 2020-09-27
[1.16.x] [WIP] Add the ability to dynamically add worlds.

mrp-v2:
Currently all this does is add a method that allows worlds to be added to a server (integrated or dedicated) using a `RegistryKey<Dimension>` and a `Dimension` object. 

The current problem is that worlds added in this manner are not loaded automatically by vanilla on startup, these worlds have to be added in this way every time the game restarts. If you have ideas on a way to fix this, please share!

Feel free to provide suggestions on how this could be done better or things that are missing, etc.

### DIFF
--- a/patches/minecraft/net/minecraft/server/MinecraftServer.java.patch
+++ b/patches/minecraft/net/minecraft/server/MinecraftServer.java.patch
@@ -1,6 +1,10 @@
 --- a/net/minecraft/server/MinecraftServer.java
 +++ b/net/minecraft/server/MinecraftServer.java
-@@ -225,7 +225,7 @@
+@@ -222,10 +222,11 @@
+    private DataPackRegistries field_195576_ac;
+    private final TemplateManager field_240765_ak_;
+    protected final IServerConfiguration field_240768_i_;
++   private IChunkStatusListener iChunkStatusListener;
  
     public static <S extends MinecraftServer> S func_240784_a_(Function<Thread, S> p_240784_0_) {
        AtomicReference<S> atomicreference = new AtomicReference<>();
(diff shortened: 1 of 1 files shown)


50ap5ud5:
Hi, thanks for adding this, this is exactly what I need for one of my mods, along with many others such as RFTools Dimensions and Mystcraft.

I am also working on a version of this system. I've found the lifecycle processes for dimensions can be quite interesting. 
Here's some notes that a previous modder made on the dimensions system, if you're interested. It may help provide insight into how we can load these worlds on startup.

https://gist.github.com/waitingtocompile/47ea59812662f91a51c7de7ce64cc124

Redstoneguy129:
Exactly what we need during these hard times!

Draco18s:
A couple of things a PR like this should be aware of include:
 - Making sure that dimension chunks are loaded and unloaded properly (I assume this happens already, just one of those things to check)
 - Allowing dimensions to be unloaded from memory (they still exist and can be traveled to, but do not receive ticks; dims can't be unloaded if a player is present)
 - Having a hook to *deregister* a dimension entirely. I think RF Tools recycles its dimensions, but there's an ongoing conversation about Mystcraft ages and being able to delete them to free up disk space. One of those "planned if we can figure out the mechanics right" things.

mrp-v2:
- As far as I can tell, vanilla handles the chunkloading automatically once the worlds are added, but some edge cases may come up related to that.
- Once again, I believe vanilla handles this already, but I'm not sure.
- This should be doable, the tricky part will probably be determining when it is 'safe' to delete a world.

mrp-v2:
This functionality of this PR is included in #7371.

PR #7364, opened 2020-09-27
System for Behavioral Biome Modifications

alcatrazEscapee:
## Problem

The problem this PR is trying to address is twofold:

1. Since `Biome`s are inextensible, prone to copying operations from deep within vanilla `Codec` based code, bad at storing their registry name (from `IForgeRegistryEntry#getRegistryName`), they are *very difficult* to use if a modder wants to modify any of their behavior. For example, what used to be acomplished simply by overriding `Biome#getPrecipitation()` in your own `Biome` subclass and registering that is now near impossible.

2. Secondly, since all these fields on `Biome` are now just holding data, there is very rarely additional context passed in to these methods such as a `World` or `BlockPos`. As Vanilla shoves more world behavior onto data driven properties of `Biome`s, this makes having dynamic behavior very difficult indeed.

## Solution

1. Mods can register `BiomeExtensionType`s, which are type objects for any manner of biome extension. These are:
- Exposed as part of the datapack json for a biome via the `forge:extensions` tag.
- Attached to all dynamic biomes as they are created
- Available to be attached to other biomes via `BiomeLoadingEvent`
- Are able to intercept call paths for different biome based behavior, and use additional context for those cases.

In order to take advantage of this system, a modder simply registers a behavior with the same registry name as their `Biome`s. Forge is able to handle attaching the behavior to the `Biome` whenever one is created through the biome Codec (something that mod code will be *very* hard pressed to try and intercept in a compatible and mod-friendly way - the biome constructor does not have a registry name available, and it's invocation path is an absolute mess of DFU nonsense).

2. Since these methods on `IForgeBiome` exist, they are able to take additional context where it is available. Call sites throughout the vanilla code base have been patched to pass in both a `World` and `BlockPos` where possible. This allows mods to have proper dynamic position and/or world dependent behavior.

## Now What

- I personally have a lot of use cases for a system like this. In <1.16.2 I overrode these methods (among others) in order to make the temperature and rain dependent on a number of factors, including the latitude, current season, in game time, etc.
- There is a test mod which demonstrates this capability by making the plains biome only rain in odd-numbered x coordinates.
- I have only added temperature and rain types initially, as I am unsure if this is the best (or has any chance at being) the solution to the outlined problem. If this is a good initiative I can look into other places where this could be taken advantage of.

## Feedback Wanted

1. Is this in any way a good solution to the problem? Are there better ways to accomplish this?
2. If so, what other possible additions to `IForgeBiome` would be good? What other context should be added to these methods where possible?
3. The method of "patch all callsites" has a tendency to generate *a lot* of patches. Is there a better way to go about this?


alcatrazEscapee:
So based on feedback in discord, I've modified this to be a more generic "extension" object, which is now:
- usable in `BiomeLoadingEvent`
- usable via datapacks
- has possibility for multiple non-conflicting extensions to be applied per biome.

This also gives modders the ability to define biome extensions for unrelated mod behavior - it eliminates the need for mods to track biomes by registry key and instead allows them to attach arbitrary data *and* behavior to biomes.

patches/minecraft/net/minecraft/world/biome/Biome.java.patch
```diff
@@ -5,22 +5,75 @@
...
-@@ -200,7 +202,7 @@
+@@ -83,9 +86,9 @@
+          return p_242434_0_.field_76749_E;
+       }), BiomeAmbience.field_235204_a_.fieldOf("effects").forGetter((p_242429_0_) -> {
+          return p_242429_0_.field_235052_p_;
+-      })).apply(p_242432_0_, (p_242428_0_, p_242428_1_, p_242428_2_, p_242428_3_, p_242428_4_) -> {
+-         return new Biome(p_242428_0_, p_242428_1_, p_242428_2_, p_242428_3_, p_242428_4_, BiomeGenerationSettings.field_242480_b, MobSpawnInfo.field_242551_b);
+-      });
++      }), ResourceLocation.field_240908_a_.fieldOf("forge:registry_name").forGetter(Biome::getRegistryName),
++      net.minecraftforge.common.world.BiomeExtensionType.CODEC.listOf().optionalFieldOf("forge:behaviors", Collections.emptyList()).forGetter(Biome::getExtensions))
```
Cyborgmas, line 32: is it needed for the client to know about these objects? I would think these things are only handled server side?

src/main/java/net/minecraftforge/common/world/BiomeExtensionType.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.mojang.serialization.Codec;
+import net.minecraft.util.ResourceLocation;
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.ForgeRegistryEntry;
+import net.minecraftforge.registries.IForgeRegistryEntry;
+
+public class BiomeExtensionType extends ForgeRegistryEntry<BiomeExtensionType>
+{
+    public static final Codec<IBiomeExtension> CODEC = ResourceLocation.field_240908_a_.<IBiomeExtension>dispatchMap(ex -> ex.getType().getRegistryName(), id -> ForgeRegistries.BIOME_EXTENSION_TYPES.getValue(id).getCodec()).codec();
```
Cyborgmas: why the dispatch map ? havent looked but wont just .dispatch do the trick and eliminate the .codec at the end?

src/main/java/net/minecraftforge/common/world/BiomeExtensionType.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private final Codec<? extends IBiomeExtension> codec;
+
+    public BiomeExtensionType(Codec<? extends IBiomeExtension> codec)
```
Cyborgmas: here i would suggest requiring a MapCodec instead and do immediately .codec() on it. This is because the KeyDispatchCodec that .dispatch() creates does an instance check on MapCodecCodec. 
pretty much all codecs have a mapcodec equivalent and most codec impl just call that mapcodec equivalent and do .codec() on it (Creating a MapCodecCodec) so this will work fine in 95% of cases, so this is for the other cases that are a pain to debug (which i know because I did :P)

patches/minecraft/net/minecraft/world/biome/Biome.java.patch
alcatrazEscapee: Yes, for example in the included test mod. The client behavior (rendering rain) changes based on the `RainType` returned.

alcatrazEscapee:
Looking at this with new updates in mind:

- #7357 addresses the main concern (attaching additional data to biomes - or any dynamic registry type - as determined through JSON) much better and more generically imo.
- Biomes store their registry name more reliably as of #7434 now, so the problem as outlined is simpler to solve by mods alone.
- Finally, the other unique purpose of this PR (patching call sites for various `Biome` method calls to add context), is probably a very obscure use case, and after discussion in the Forge discord upon ways to work around this, are probably not suited for Forge.

PR #7373, opened 2020-09-28
Fix toggleable keybindings still being active while in GUI

### LINKED ISSUE #7238: [1.16.x] No difference between AnimalEntity.func_204701_dC() and AgeableEntity.func_213743_em()
**Minecraft Version:** 1.16

**Forge Version:**  32.0.108

**Description of issue:**
There is a naming conflict between AnimalEntity#func_204701_dC and AgeableEntity#func_213743_em since both have the same name (canBreed) in the mappings forge is built against despite being different in the original code, which means that currently, func_204701_dC is removed in favor of func_213743_em.

### LINKED ISSUE #7370: Sneaking while in GUI causes movement
**Minecraft Version:** 1.16.3
**Forge Version:** 34.1.0
**Logs:** _not applicable, in-game issue_

**Steps to Reproduce:**
  1. Open a world in creative mode
  2. Start flying above the ground _(above 5 blocks or so, to see the issue clearly)_
  3. Open the creative menu
  4. Press and hold the sneak keybind _(defaults to <kbd>Shift</kbd>)_
  5. _Observe that the player descends to the ground_
  6. Repeat steps 1-4 on a vanilla instance
  7. _Observe that the player does not descend to the ground_

**Description of issue:**
When pressing the sneak keybind (default is <kbd>Shift</kbd>) while inside a GUI, its actions still apply to the player (in this case, the player descends from flying). I suspect this may be related to an [earlier PR by diesieben07][keybind_pr], which fixed a reported issue about toggleable keybinds being triggerable while in a GUI.

[keybind_pr]: https://github.com/MinecraftForge/MinecraftForge/pull/7338

sciwhiz12:
_This PR fixes #7370._

An [earlier PR by diesieben07][earlier_pr] fixed the issue of the player's sneak stopping when the player opens a GUI and the sneak toggle is on. However, this led to another issue, as described by the linked issue: when the sneak keybind is set to hold, and a GUI is open, the sneak keybind still takes effect.

This PR fixes that by requiring a toggleable keybind to be active if it's pressed down, and either the key confilict context is active or the keybind is set to toggle.
This PR also includes a cleanup commit to fix the license headers of `SkyRenderHandler` and `BiomeLoadingEventTest`, so the `checkLicenses` task does not error.

[earlier_pr]: https://github.com/MinecraftForge/MinecraftForge/pull/7338

### DIFF
--- a/patches/minecraft/net/minecraft/client/settings/ToggleableKeyBinding.java.patch
+++ b/patches/minecraft/net/minecraft/client/settings/ToggleableKeyBinding.java.patch
@@ -13,5 +13,5 @@
        }
  
     }
-+   @Override public boolean func_151470_d() { return this.field_74513_e; }
++   @Override public boolean func_151470_d() { return this.field_74513_e && (isConflictContextAndModifierActive() || field_228053_a_.getAsBoolean()); }
  }
--- a/src/main/java/net/minecraftforge/client/SkyRenderHandler.java
+++ b/src/main/java/net/minecraftforge/client/SkyRenderHandler.java
@@ -1,4 +1,3 @@
-  
 /*
  * Minecraft Forge
  * Copyright (c) 2016-2020.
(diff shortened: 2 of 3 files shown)


Thelonedevil:
This looks like it should fix 7370 rather than 7238 unless I am mistaken?

sciwhiz12:
Whoops, my mistake. Must've copied the wrong number while writing the issue; thankfully the commit message is correct. Thank you for noticing it!

PR #7374, opened 2020-09-28
Fix invalid license headers

Gamebuster19901:
These files have missing or incorrect license headers and are causing my builds to fail.


ChampionAsh5357:
This is currently within #7373.

PR #7378, opened 2020-09-30
Add back patches to Screen allowing for item tooltips to be auto wrapped again

pupnewfster:
When #7268 was merged it skipped adding back the patches in Screen to call GuiUtils#drawHoveringText because of types not matching. This PR patches the calls that have text components to still go through GuiUtils#drawHoveringText to ensure that tooltips get automatically wrapped and don't end up falling off the screen. There are still a few places that call Screen#func_238654_b_ but they seem to be mostly just internal implementations of some of vanilla's screens most of which not even having access to something extending ITextProperties without a decent amount of rewriting of their implementations so after discussing it with @gigaherz we decided they probably don't need to be patched.


marchermans (approved):
Code looks good.
Tested on my machine and with C&B as a test mod. Works.

Issue #7326, opened 2020-09-12
Additional Tag Types - TIleEntityType, Enchantments, and possibly others


Tfarcenim:
Recently, Forge has added the ability to create custom tag registries for any ForgeRegistry.  
However, mods can only create new ones if they are the owner of the Registry they came from and cannot create new vanilla ones such as TileEntityType.
Currently, Biome is the only supported one, I would like to see TileEntityType and Enchantment as well.

KnightMiner:
I think for adding any new tag type, it is important to describe usecases for the tag if you want it added. As in, why do we need tile entity type tags and enchantment tags?

----

For enchantments, I could see it potentially being useful to mark an enchantment as an axe enchantment, a sword enchantment, etc. Though it may be better to do that the other way, to where an enchantment has a tag of items that are acceptable. Either way requires a small rework to how enchants work and would require a lot more discussion. Beyond these I cannot think of another good use.

For tile entity type, I cannot think of one case that is not handled already by block tags, as you cannot have a tile entity without blocks. Especially with the fact that tile entity types have a closed list of blocks, you are not going to be seeing new blocks being added with an existing tile entity type that often.

The one other case I can think of being useful is potions (the forge registry calls this by the old name of `POTION_TYPE`, not referencing the effects). Could be used to tag all variants of a potion for instance. However, I think I am the only mod that uses "potions" as a crafting input, and you can always write my recipes with a list of potions instead, so I am not sure that is enough case for this.

----

Overall, this may be a discussion we should have on the forum first, then come back here with reasoning for specific tag types.

Tfarcenim:
> For enchantments, I could see it potentially being useful to mark an enchantment as an axe enchantment, a sword enchantment, etc. Though it may be better to do that the other way, to where an enchantment has a tag of items that are acceptable. Either way requires a small rework to how enchants work and would require a lot more discussion. Beyond these I cannot think of another good use.

My usecase was providing a user friendly way of adding enchants to a specific item (spikes in this case).  Another usecase that I saw earlier is for a common tag space for enchants that should be mutually exclusive like soulbound.

For example if 3 mods add soulbound, then it would be possible to mark incompatibility with other soulbound enchants without hardcoding via
enchantment.isIn(EnchTags.SOULBOUND), instead of having to rely on a list of strings or grabbing every soulbound mod's id.

[...]

PR #7379, opened 2020-09-30
Adds tag support for Enchantments, Potions, and TileEntityTypes

pupnewfster:
See #7326 for the discussion/request about some of the various use cases for the different tag types.


patches/minecraft/net/minecraft/enchantment/Enchantment.java.patch
```diff
@@ -18,7 +34,7 @@
     }
  
     public void func_151368_a(LivingEntity p_151368_1_, Entity p_151368_2_, int p_151368_3_) {
-@@ -139,6 +139,24 @@
+@@ -139,6 +148,24 @@
```
TheCurle, line 37: What's up with these hunks being changed randomly?
I don't see a reason for these to be changed, from looking at the files

pupnewfster: iirc it is just because that represents what line it is starting on so because there are more lines being injected above that means that anything being added lower down gets added *later* line number wise.

TheCurle: I don't see any other patches in these files (with these hunk indexes), these should not be changed?
Else every single line would have one of these..

TheCurle: Same with the first one in this file. You don't change anything above it, yet it is still changed.

pupnewfster: https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/patches/minecraft/net/minecraft/enchantment/Enchantment.java.patch#L21-L42

TheCurle: Okay. that one was me being blind, my bad.
Now explain this: https://github.com/MinecraftForge/MinecraftForge/pull/7379/files#diff-9b95d3b9cf5c13f0915da0244317d899R3

RANKSHANK: There are 9  lines added to the file and the patch lines below are shifted by 9, seems pretty sane to me

pupnewfster: Because a line is being added here curle? https://github.com/MinecraftForge/MinecraftForge/pull/7379/files#diff-9b95d3b9cf5c13f0915da0244317d899R9

TheCurle: The first one. There is nothing being added above it, and the line being added inside it shouldn't affect the position.
Whatever, i'll just chalk it down to Git being weird.
I redact my first statement :P

PR #7383, opened 2020-10-01
Added an ItemModifierEvent.

Darkhax:
The goal of this PR is to improve the consistency of modded attribute modifier sources. It does this by providing an ItemModifierEvent which allows listeners to add/remove modifiers as needed.

**Use Cases**
1. A mod that prevents standard tools and armor from disappearing when broken but removes all functionality and modifiers until repaired.
2. A mod that allows standard tools to level up and unlock additional modifiers.
3. An enchantment which gives tools a modifier, similar to Sharpness.

The current roadblock for these use cases comes from mixing multiple sources together. For example an enchantment provided modifier may be applied during entity tick like a potion effect. This would mean the broken tool mod could not reliably detect and remove that modifier automatically. If both mods were to use this system the removal mod could simply clear the map.

Another challenge these mods have is making sure the item tooltips provide accurate information to the player. While not impossible currently this PR would seamlessly integrate the modifiers into the vanilla tooltip code.

While NBT modifiers do exist, when the vanilla code detects that NBT tag it will disable the built in modifiers. Mods have tried to get around that by adding the built-in modifiers to the NBT as well however this causes compatibility issues with stack sensitive built-in modifiers.


patches/minecraft/net/minecraft/item/ItemStack.java.patch
```diff
@@ -214,16 +214,20 @@
...
              }
           }
        } else {
 -         multimap = this.func_77973_b().func_111205_h(p_111283_1_);
 +         multimap = this.func_77973_b().getAttributeModifiers(p_111283_1_, this);
        }
- 
+-
++      multimap = com.google.common.collect.ArrayListMultimap.create(multimap);
++      net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.ItemModifierEvent(this, p_111283_1_, multimap));
```
pupnewfster: Part of me wonders if there is some good way that we could maybe "cache" the result of this as of 1.16 Vanilla caches in an ImmutableMultiMap the attributes items themselves return. Alternatively should this at least be returning an immutable map from this method in ItemStack after making it mutable for purposes of firing the event so that people don't accidentally "inject" attributes outside of using the event onto item stacks.

src/test/resources/META-INF/mods.toml
```diff
@@ -87,4 +87,6 @@ license="LGPL v2.1"
 [[mods]]
     modId="custom_tag_types_test"
 [[mods]]
-    modId="biome_loading_event_test"
\ No newline at end of file
+    modId="biome_loading_event_test"
+[[mods]]
+	modId="item_modifier_test"
```
pupnewfster: This is a tab, and should probably be four spaces to match the rest of the file.

patches/minecraft/net/minecraft/item/ItemStack.java.patch
Darkhax: I don't think this can be reliably cached as the initial map is not deterministic. Making it immutable again is fine however it would add more overhead.

pupnewfster: Ya I wasn't sure if there is a good way to cache it which is why I was mentioning wondering. I do think  having it be immutable would be better though, or at least unnmodifiable given nothing would have access to the local ArrayListMultimap to accidentally mutate it. Though I am not sure if there is a unnmodifiable view for multimaps.

Shadows-of-Fire:
This sounds like an excellent solution to the problem that is "having to use nbt modifiers".  I have to do precisely as you've described and copy those modifiers into the nbt, but that still breaks with some mods, like silent gear.

The only other thing I would like to see would be some way to sort tooltips, as currently there is no sorting order, and I believe it would be ideal traits such as attack damage / attack speed were always presented first.  Scenarios such as this just look weird:

Darkhax:
@Shadows-of-Fire Better tooltip code for attributes in general is needed but I think they're out of scope for this PR. One thing I would like to see is giving the attribute the ability to supply it's own ITextComponent.

Shadows-of-Fire:
Yeah, tooltips could stand to be in their own PR.  They would certainly be appreciated, but this needs to happen first.

Oh, and @Darkhax you may want to resolve the merge conflict with the mods.toml

Darkhax:
Thanks for the heads up!

Darkhax:
I've merged a PR by KnightMiner which adds some useful improvements to this PR.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -152,6 +157,7 @@
 import net.minecraftforge.registries.GameData;
 import net.minecraftforge.registries.IRegistryDelegate;
 
+import org.antlr.v4.runtime.misc.MultiMap;
```
Shadows-of-Fire, line 160: Is this being used?

KnightMiner: Oh, you can blame me for that one. Was importing Multimap (the one Mojang uses) and clicked the wrong import. In my normal workspaces I auto clean up imports on save, but I have that disabled when I work on Forge to prevent unneeded patches

patches/minecraft/net/minecraft/item/ItemStack.java.patch
```diff
@@ -214,15 +214,18 @@
...
-@@ -817,7 +841,7 @@
+@@ -817,9 +841,9 @@
              }
           }
        } else {
 -         multimap = this.func_77973_b().func_111205_h(p_111283_1_);
 +         multimap = this.func_77973_b().getAttributeModifiers(p_111283_1_, this);
        }
- 
+-
```
KnightMiner, line 224: Something got a little weird in the patch here. I don't recall seeing it in my PR because I was looking at a patch of a patch.

If I am reading this right, it is removing an empty line between `}` and `multimap = ...`?

Shadows-of-Fire: Yeah, that's what that is, Patch "`+ -`" is "this line was removed in this commit"

Darkhax:
This PR has been superseded by https://github.com/MinecraftForge/MinecraftForge/pull/7484 because I don't have the time to keep maintaining it.

PR #7385, opened 2020-10-02
Added Basic Class-Level SAS Capability to checkSAS, added Vector4f to SAS

DaemonUmbra:
Opening this PR as a successor to #7201 
The intentions are the same but without the compounded mistakes I made attempting to update the previous

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1062,5 +1062,3 @@ task setup() {
     dependsOn ':clean:extractMapped'
     dependsOn ':forge:extractMapped' //These must be strings so that we can do lazy resolution. Else we need evaluationDependsOnChildren above
 }
-
-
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -1062,5 +1062,3 @@ task setup() {
     dependsOn ':clean:extractMapped'
     dependsOn ':forge:extractMapped' //These must be strings so that we can do lazy resolution. Else we need evaluationDependsOnChildren above
 }
-
-
```
pupnewfster, line 1066: Is this meant to be part of the diff? As it is the only changes in the file.

DaemonUmbra: It wasn't intended to be but idk why there's a double newline at the end of it as it is

PR #7386, opened 2020-10-02
Implement LivingConversionEvent event

PellucidDice:
This is an event can help the users to handle mob conversion.
If pre is cancelled the mob conversion will be also cancelled.
Post provides the original entity (likely to be removed right after the event), and the new entity.


src/main/java/net/minecraftforge/event/entity/living/LivingConversionEvent.java
```diff
@@ -0,0 +1,63 @@
...
+import net.minecraftforge.eventbus.api.Cancelable;
+
+/**
+ * LivingConversionEvent is triggered when an entity is trying
+ * to replace itself with other entity
+ *
+ * This event is {@link Cancelable}
+ * If cancelled, the replacement will not occur
+ */
+@Cancelable
```
ChampionAsh5357: The base event should not be cancelable. Only make the needed events cancelable.

src/main/java/net/minecraftforge/event/entity/living/LivingConversionEvent.java
```diff
@@ -0,0 +1,63 @@
...
+
+        public EntityType<? extends LivingEntity> getOutcome()
+        {
+            return outcome;
+        }
+
+        public void setOutcome(EntityType<? extends LivingEntity> outcome)
+        {
+            this.outcome = outcome;
+        }
```
ChampionAsh5357: None of these seem to do anything.

src/main/java/net/minecraftforge/event/entity/living/LivingConversionEvent.java
```diff
@@ -0,0 +1,63 @@
...
+
+        public LivingEntity getOutcome()
+        {
+            return outcome;
+        }
+
+        public void setOutcome(LivingEntity outcome)
+        {
+            this.outcome = outcome;
+        }
```
ChampionAsh5357: These also seem pointless.

src/main/java/net/minecraftforge/event/entity/living/LivingConversionEvent.java
```diff
@@ -0,0 +1,63 @@
+package net.minecraftforge.event.entity.living;
```
ChampionAsh5357, line 20: Add Forge license for header.

src/test/java/net/minecraftforge/debug/entity/living/LivingConversionEventTest.java
```diff
@@ -0,0 +1,31 @@
+package net.minecraftforge.debug.entity.living;
```
ChampionAsh5357, line 20: Add Forge license for header.

Alpvax:
Do we want to expand this to include creeper -> charged creeper?
Or is it only for when the EntityType changes?

cech12:
A mooshroom can be transformed by a lightning. The EntityType does not change but the color (brown).
Maybe the Sheep can also be part of it when it is colored?

PellucidDice:
I feel like the things you've mentioned (change in data) should probably handle in another new event as if we add those changes in this event, it will make things a bit messed up unless we create a list of conditions to decide clearly if a data change can be count as a conversion.

marchermans (approved):
Good addition.

Personally I think we should not trigger this event for raw data changes.
We could add additional events for this, but a single event, like EntityStateChanged could also suffice.

That said, this PR works, code looks good and well formatted.

src/main/java/net/minecraftforge/event/entity/living/LivingConversionEvent.java
```diff
@@ -0,0 +1,76 @@
...
+    {
+        private final EntityType<? extends LivingEntity> outcome;
+
+        public Pre(LivingEntity entity, EntityType<? extends LivingEntity> outcome)
+        {
+            super(entity);
+            this.outcome = outcome;
+        }
+
+        public EntityType<? extends LivingEntity> getOutcome()
```
marchermans, line 64: Please add javadoc to this entry.

src/main/java/net/minecraftforge/event/entity/living/LivingConversionEvent.java
```diff
@@ -0,0 +1,76 @@
...
+    {
+        private final LivingEntity outcome;
+
+        public Post(LivingEntity entity, LivingEntity outcome)
+        {
+            super(entity);
+            this.outcome = outcome;
+        }
+
+        public LivingEntity getOutcome()
```
marchermans, line 103: Please add javadoc to this entry.

marchermans (changes requested):
Code okey, javadoc missing.

LexManos:
Needs rebasing, but pullable after that.

PellucidDice:
Haven't play Minecraft for a while, there may be some leftovers.

Entities conversion types we currently have in this pr:
Villager --> Witch
Zombie --> Drowned
Villager <-> ZombieVIllager
Piglin    -->  Zombified Piglin
Pig        -->  Zombified Piglin
Hoglin   -->  Zoglin

bl4ckscor3:
Hoglin --> Zoglin is one that is missing.

sciwhiz12:
There is also conversion of red mooshrooms into brown mooshrooms and vice-versa by being struck by lightning.

PellucidDice:
> Hoglin --> Zoglin is one that is missing.

Thank you.

> There is also conversion of red mooshrooms into brown mooshrooms and vice-versa by being struck by lightning.

LivingConversionEvent is currently looking for conversions that have changing in EntityType, but mooshrooms are pretty special that they have their own type, there's a discussion on this in previous comments.

I'm also thinking should we reset the timer when the event is cancelled, for example:
- Reset drownedConversionTime to 0 if the event is cancelled.
- Reset field_234358_by_ in HoglinEntity to 0 if the event is cancelled.

Otherwise the event can be triggered every tick even if the event was cancelled last tick, an alternative way is I can create another boolean to let the users decide whether the timer should be reset. What do you think?

Alpvax:
It might be worth making the timer available in the event so modders can delay the transformation by a set number of ticks. For transformations without a timer it could be set to -1 and ignored

LexManos:
That was my initial thought myself, but not all conversions are based on timer. So it'd be better to access accessors to the entities themselves, and let the mods call whatever logic they want to delay.

PellucidDice:
Maybe something like this will work?

https://pastebin.com/UqQr5Dky

Alpvax:
> Maybe something like this will work?
> 
> https://pastebin.com/UqQr5Dky

Except that will not reset the cooldown when the event is cancelled. You need to remove the else from the else if.

PellucidDice:
> Except that will not reset the cooldown when the event is cancelled. You need to remove the else from the else if.

Are you sure about that, I may be wrong but isn't if the event is cancelled, 
boolean cancelled = true, 
then !cancelled = false, 
therefore else if will execute which resets the timer if needed?

Alpvax:
My mistake, if it is not cancelled, the timer will not reset. But of course in that instance the transformation will continue, so the timer will be irrelevant. I had not thought through all the side effects.

PellucidDice:
Yeah, the reason I did not reset if the event is not cancelled is that the entity itself is likely going to be removed right after that.

LexManos:
Perhaps pass in a Consumer<Integer> to the hook to reset the timer to a numeric value?
So we dont have to add accessors. And so that the patches can be tiny `delay -> this.convertDelay = delay` as an argument to the hook.

Also, its a difference between 'reset the time' and 'set the timer to a value'
The second part is more flexible so you should go that route.

patches/minecraft/net/minecraft/entity/monster/ZombieEntity.java.patch
```diff
@@ -1,23 +1,24 @@
...
 +++ b/net/minecraft/entity/monster/ZombieEntity.java
-@@ -186,7 +186,7 @@
+@@ -186,7 +186,8 @@
        if (!this.field_70170_p.field_72995_K && this.func_70089_S() && !this.func_175446_cd()) {
           if (this.func_204706_dD()) {
              --this.field_204708_bE;
 -            if (this.field_204708_bE < 0) {
-+            if (this.field_204708_bE < 0 && net.minecraftforge.event.ForgeEventFactory.canLivingConvert(this, EntityType.field_200725_aD)) {
++
++            if (this.field_204708_bE < 0 && net.minecraftforge.event.ForgeEventFactory.canLivingConvert(this, EntityType.field_200725_aD, (timer) -> this.drownedConversionTime = timer)) {
```
LexManos: Something is broken here.. did you edit this file by hand? The field should of been mapped to the srg name. I mean.. it's fixed in the first part of the line, but the lamba is broken...

patches/minecraft/net/minecraft/entity/monster/ZombieVillagerEntity.java.patch
```diff
@@ -1,14 +1,15 @@
...
 +++ b/net/minecraft/entity/monster/ZombieVillagerEntity.java
-@@ -117,6 +117,7 @@
+@@ -116,7 +116,7 @@
+       if (!this.field_70170_p.field_72995_K && this.func_70089_S() && this.func_82230_o()) {
           int i = this.func_190735_dq();
           this.field_82234_d -= i;
-          if (this.field_82234_d <= 0) {
-+            if (!net.minecraftforge.event.ForgeEventFactory.canLivingConvert(this, EntityType.field_200756_av)) return;
+-         if (this.field_82234_d <= 0) {
++         if (this.field_82234_d <= 0 && net.minecraftforge.event.ForgeEventFactory.canLivingConvert(this, EntityType.field_200756_av, (timer) -> this.conversionTime = timer)) {
```
LexManos: Same here...

patches/minecraft/net/minecraft/entity/monster/piglin/AbstractPiglinEntity.java.patch
```diff
@@ -1,14 +1,15 @@
...
-@@ -84,6 +84,7 @@
+@@ -83,7 +83,7 @@
+          this.field_242334_c = 0;
        }
  
-       if (this.field_242334_c > 300) {
-+         if (!net.minecraftforge.event.ForgeEventFactory.canLivingConvert(this, EntityType.field_233592_ba_)) return;
+-      if (this.field_242334_c > 300) {
++      if (this.field_242334_c > 300 && net.minecraftforge.event.ForgeEventFactory.canLivingConvert(this, EntityType.field_233592_ba_, (timer) -> this.field_242334_c = timer)) {
           this.func_241848_eP();
```
LexManos, line 9: This one obfed correctly

LexManos (changes requested):
Sorry for the delay, moving houses is a pain. Anyways, seems there are some obf issues with your latest commit. @MinecraftForge/issuemanager Can someone keep an eye on this and poke it when it's resolved. All my other concerns have been addressed.

PellucidDice:
It seems like I did not run genPatches correctly, sorry for wasting your time.

LexManos (approved):

PR #7387, opened 2020-10-02
Reimplement nametag distance attribute, fix attribute translation keys

### LINKED ISSUE #7372: [1.16.3] Forge attributes don't have proper translation key
**Minecraft Version:** 1.16.3

**Forge Version:** 34.1.0

**Description of issue:**
The forge attributes (reach, swimming speed, ...) don't have a proper translation key (or have an incorrect one) and thus don't display properly when shown in a tooltip, such as the "When in Main Hand" of swords via `getAttributeModifiers` or on potion effects, as seen here (with a vanilla attribute to compare):

image

sciwhiz12:
_This PR closes #7372._

First, this PR addresses the lack of translation keys for the Forge-added attributes. For this, I mirrored how vanilla has their attributes and translation keys, and added the English (`en_us.json`) translation keys as a default. I do not know how Forge integrates with [Crowdin][crowdin] and the crowdsourced translations, and I have received no word from any of the Forge/Triage team on how it is setup. 
I request any member of the Forge/Triage team to contact me if there are further steps to be taken on my end to add these translation keys to the Crowdin site.

Second, this PR addresses the lack of functionality of the `forge:nametag_distance` attribute. From the name, I have presumed that this attribute controls the distance from which the nametag of a living entity can be observed from. I implemented a hook as `ForgeHooksClient#isNameplateInRenderDistance(...)`, which is patched into `EntityRender#renderName(...)`.

However, due to restrictions on entity rendering, all nameplates are not rendered if they exceed 64 blocks from the camera's position (see `Entity#isInRangeToRenderDist()`). In my opinion, the patches required to workaround that for this attribute may cause unintended consequences, so I went with the safe option of limiting the attribute's max value to 64 blocks (as any higher value functions the same as `64.0d`).

_Note: In my opinion, these changes should be considered non-breaking: for the first point, the changed string is only used (as far as my investigation led me) for translations; for the second point, any value higher than `64.0d` has the same function as that value; the changed maximum value simply ensures that it will never be some arbitrarity high number._

[crowdin]: https://crowdin.com/project/minecraft-forge

### DIFF
--- a/src/main/resources/assets/forge/lang/en_us.json
+++ b/src/main/resources/assets/forge/lang/en_us.json
@@ -159,5 +159,9 @@
   "attribute.name.forge.swim_speed": "Swim Speed",
   "attribute.name.forge.nametag_distance": "Nametag Render Distance",
   "attribute.name.forge.entity_gravity": "Entity Gravity",
-  "attribute.name.generic.reach_distance": "Reach Distance"
+  "attribute.name.generic.reach_distance": "Reach Distance",
+  "forge.swimSpeed": "Swim Speed",
+  "forge.nameTagDistance": "Nametag Render Distance",
+  "forge.entity_gravity": "Entity Gravity",
+  "generic.reachDistance": "Reach Distance"
 }
--- a/src/main/resources/assets/forge/lang/en_us.json
+++ b/src/main/resources/assets/forge/lang/en_us.json
@@ -156,6 +156,8 @@
   "forge.controlsgui.alt": "ALT + %s",
 
   "forge.container.enchant.limitedEnchantability": "Limited Enchantability",
-  "attribute.name.generic.reachDistance": "Reach Distance",
-  "attribute.name.forge.swimSpeed": "Swim Speed"
+  "attribute.name.forge.swim_speed": "Swim Speed",
+  "attribute.name.forge.nametag_distance": "Nametag Render Distance",
+  "attribute.name.forge.entity_gravity": "Entity Gravity",
+  "attribute.name.generic.reach_distance": "Reach Distance"
 }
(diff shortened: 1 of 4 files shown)


src/main/resources/assets/forge/lang/en_us.json
```diff
@@ -156,6 +156,8 @@
   "forge.controlsgui.alt": "ALT + %s",
 
   "forge.container.enchant.limitedEnchantability": "Limited Enchantability",
-  "attribute.name.generic.reachDistance": "Reach Distance",
-  "attribute.name.forge.swimSpeed": "Swim Speed"
+  "attribute.name.forge.swim_speed": "Swim Speed",
+  "attribute.name.forge.nametag_distance": "Nametag Render Distance",
+  "attribute.name.forge.entity_gravity": "Entity Gravity",
+  "attribute.name.generic.reach_distance": "Reach Distance"
```
pupnewfster: Is the naming change *needed* for it to work, or can it work with the old naming? The reason I ask is if memory serves if the keys get renamed stuff needs to happen in crowdin/split as forge currently is using the same lang instance on crowdin between 1.15 and 1.16 (and I think maybe even 1.14)

sciwhiz12: It does work with the old naming; my reasoning for changing it is to make it consistent with how vanilla names the translation keys for their attributes. If you wish, I can use the old naming (_though I really wish we can make it consistent with vanilla_).

marchermans: @sciwhiz12 You will probably need both entries in there. Since the data is indeed pulled for both 1.15 and 1.16 so this would be a breaking change in both branches.

sciwhiz12: So I've added both translation entries for the attributes to the en_us.json file, but the code only uses the old translation keys for now. This should prevent this from being a breaking change now.

marchermans (approved):

PR #7394, opened 2020-10-03
Add ability to get the boxes of entity models.

Gamebuster19901:
Note: This PR contains breaking changes, and involves rendering.

With the rendering changes in 1.16, it has become impossible to get the cubes of a model because all of the ModelRenderer instances in most vanilla models are not in a collection or array, so they would all have to be retrieved manually via reflection. Additionally, the boxes from modded models are impossible to obtain.

This PR adds the method `getModelParts()` to the `Model` class so the model boxes are obtainable again. Mods that have models will need to implement this method.

This would allow modders to render other things onto models from other mods, and onto vanilla models.

An example where this would be useful is rendering arrows or other projectiles that have struck an entity. Vanilla does this for players, but not for other entities.

I intend on using this feature to render sticky grenades onto entities, like in halo: https://www.youtube.com/watch?v=T1Wxd-48XPY

Here is an image showing the first use case:
image

TODO:

- [ ] Update this post
- [ ] Javadocs for new events
- [x] Ability to retrieve boxes from models
- [x] Cancelable event that fires when a layer is added
- [x] ~~Cancelable event that fires when a layer is removed~~
- [x] Make an event specifically for adding layers
- [x] Create test mod
    - [x] Test layer add event
    - [ ] ~~Test layer remove event~~
    - [x] Test renderer creation event


gigaherz:
You say minor, but this literally breaks every single released mod which has even just one entity model in it. So I'd call that major.
If you hope to be able to get this merged to 1.16.x, you may want to make that method nullable, default it to return null, with a TODO comment to make it abstract in 1.17

Gamebuster19901:
That's fair, I assumed that mods having to implement a getter wasn't that big of a breaking change.

I think making the method return an empty collection instead of null would be better though, as it wouldn't require the annotation which would save a line in the patch, and mods using it wouldn't have to use null checks.

gigaherz:
You don't need to annotate all the overrides, only the main method in the Model class, and speaking of, instead of adding the method inside Model, you may want to create an extension interface. `net.minecraftforge.client.extensions.IForgeModel` would be the name based on the existing convention, this way you only have one patch line, regardless of how complex the added method is.

ExpensiveKoala:
Why not implement vanilla's methods like those used for the player model like you said?
Patching `Model#accept` should achieve the same effect wouldn't it?

Currently `PlayerModel` is the only one that overrides it, so you would only need to patch `Model` and `PlayerModel` (and maybe `ModelRenderer`).
`Model#accept` gets passed the ModelRenderer that gets created. This way, as long as the model is using one of the constructors that calls it, you can get it to work on all models without having to patch them. That includes modded models.

Gamebuster19901:
> Why not implement vanilla's methods like those used for the player model like you said?
> Patching `Model#accept` should achieve the same effect wouldn't it?
> 
> Currently `PlayerModel` is the only one that overrides it, so you would only need to patch `Model` and `PlayerModel` (and maybe `ModelRenderer`).
> `Model#accept` gets passed the ModelRenderer that gets created. This way, as long as the model is using one of the constructors that calls it, you can get it to work on all models without having to patch them. That includes modded models.

1. Model#accept is used to add new ModelRenderers to a Model
2. Adding new ModelRenderers and ModelBoxes directly to the model is out of scope for this PR, this PR is for retrieving model boxes from the model, and adding/removing LayerRenderers
3. PlayerModel is one of the only models that stores it's ModelRenderers in a collection, so it actually can accept new ModelRenderers. Most models have their ModelRenderers hardcoded individually in `private final` fields, so to make Model#accept work, every model would need a patch to add their ModelRenderers into a collection.

ExpensiveKoala:
> * Model#accept is used to add new ModelRenderers to a Model
> * Adding new ModelRenderers to a model directly is out of scope for this PR, this PR is for retrieving model boxes from the model, and adding/removing LayerRenderers
> * PlayerModel is one of the only models that stores it's ModelRenderers in a collection, so it actually can accept new ModelRenderers. Most models have their ModelRenderers hardcoded as a `private final` field, so to make Model#accept work, every model would need a patch to add their ModelRenderers into a collection.

1. `Model#accept` is called by the constructor of `ModelRenderer`. It's not used to add new `ModelRenderer`s to a `Model`, but rather an empty method that gets called when a new `ModelRenderer` is created.

1. It's not adding new `ModelRenderer`s. `PlayerModel` uses `Model#accept` to store the `ModelRenderer`s that get instantiated to return its `ModelRenderer`s as a collection.

1. You would only have to patch `Model` and `PlayerModel`, as all Models extend `Model` in one way or the other. `PlayerModel` being the only one that overrides the accept method, so you would have to patch `PlayerModel`.

-------------

The only issue I can see, is this won't work for models such as Shulker, EnderCrystal, Trident, or TileEntityRenderers which call the constructors which don't get passed a `Model`

Gamebuster19901:
Ah, I didn't realize Model#accept was called in some of the constructors of ModelRenderer.

Unfortunately, that also won't work for modded models that don't use the constructors which call Model#accept. I could patch the ModelRenderer constructors to require a Model, but then all of the vanilla calls to those would also have to be patched, defeating the purpose.

 I still think this is the best approach for compatibility.

ExpensiveKoala:
It wouldn't work for modded models with the way you're doing it without them updating either. You would be patching in a method for getting the collection of `ModelRenderer` anyways so have them override that. Besides, if they are making a MobModel, they should be following the vanilla convention of using the proper constructor that takes in the `Model` anyways.

Most mods use tools to generate their model code. Those models already use the correct constructor.

Again, the only models this method would exclude are TileEntityRenderers, which don't extend Model and models such as the ShulkerModel, BookModel, and TridentModel.

Also, since the list of models that would be excluded include TileEntityRenderers, you can't just blindly patch the constructors to take a `Model` either and I don't think you should do that either.

Even if you were to patch them to have them override the method that returns the collection, it would still be far less patches than patching every single model class. It also has the benefit of already working with most models from mods without them having to update their mod to include that method you're adding.

Gamebuster19901:
You make some very good points, I'll try doing it that way.

PR #6467, opened 2020-01-22
 [1.15.x] Re-implement the Config GUI

Cadiboo:
PR containing only the Config GUI implementation from #6421. 
- Adds the mods list button to the in game menu screen
- Re-implements the config GUI
- Adds a test mod for the config system
- Mostly re-implements FML's dummy config GUI

Screenshots:

PR #6464, opened 2020-01-22
[1.15.x] Extend the config system

Cadiboo:
- Fixes and standardises the comments in `ForgeConfigSpec`
- Makes `ForgeConfigSpec`'s utility methods public (in prep for the Config GUI PR)
- Fixes various small issues in `ForgeConfigSpec`
- Adds support for Bytes, Shorts and Floats (`ForgeConfigSpec`)
- Adds info to `FMLConfig` telling Modders to use Forge's system and not base their config of FML's
- Improves the documentation on `ModConfig.Type`

PR #7395, opened 2020-10-04
[1.16.x] Re-implement the Config GUI

Cadiboo:
Supersedes #6467.

- Re-implements the config GUI
- Adds a test mod for the config system
- Makes some changes to other Forge classes

This was designed to be super easily extensible by other mods wanting to add extra stuff to their Config GUI.
The test mod serves as an example of how to do this.
More thought was put into this than writing clean or highly performant GUI code.

### Forge API changes (no breaking changes)
 - Increase visibility of `ForgeConfigSpec.Range` from `private` to `public` (To allow use in the Config GUI)
- Add a public method to `ConfigTracker` to get all `ModConfig` registered to a mod
- Add a public method to `ModConfig` to save and reload it's contents (see <> for why)
- Makes some changes/improvements to `ForgeConfigSpec` (Needs a talk with someone from the Forge team, preferably cpw, preferably on discord)
  - Need to talk with one of the Forge team about how translation keys for config categories should be handled
  - Remove explicit comment handling and use `CommentedConfig` instead
  - (Minor) Add "allowed values" to the comments of entries created with `defineInList`

### Other changes
- Adds the mods list button to the in game menu screen
- Add back `MessageDialogScreen`, a copy of the `DisconnectedScreen` but with custom text
- Changes/improvements to `ForgeConfigSpec`
  - Comment cleanup & typo fixes
  - Store "allowed values" for entries created with `defineInList` and `defineEnum` for use by the Config GUI
  - Make `ConfigValue`s implement their respective `Supplier` counterparts ([Discussed on discord](https://discordapp.com/channels/313125603924639766/725850371834118214/740768957342154762) and was super easy to implement)
  - Implement the rest of `ForgeConfigSpec.Range#toString()`
- Improve the comments on `ModConfig.Type` (from #6464)
- Add doc/info comment to `FMLConfig` (from #6464)
- Formatting fixes to `ForgeConfig.java`
- Comment typo fixes to `DistExecutor`
- Cleanup to the `Slider` widget
- Change `UnicodeGlyphButton` from rendering the result of `getNarrationMessage` as it's title to using `getMessage`
- Fixes up some translations for Forge's config


src/main/java/net/minecraftforge/client/gui/config/CategoryConfigScreen.java
```diff
@@ -0,0 +1,294 @@
...
+        tooltip.addAll(translateCommentWithFallback(commentTranslationKey, comment));
+        return tooltip;
+    }
+
+    @Override
+    protected ConfigElementList makeConfigElementList() {
+        return new CategoryConfigElementList(this, field_230706_i_, categoryInfo);
+    }
+
+    public interface ConfigCategoryInfo {
```
ichttt, line 103: Maybe move the inner classes it into their own files? Or is there a good reasoning this is an inner class?

src/main/java/net/minecraftforge/client/gui/config/CategoryConfigScreen.java
```diff
@@ -0,0 +1,294 @@
...
+            };
+        }
+
+        Collection<String> elements();
+
+        Object getValue(String key);
+
+        Object getSpec(String key);
+
+        /** Special case categories because reasons */
```
ichttt: Would be nice to have a little more understandable comment than `because reasons`

src/main/java/net/minecraftforge/client/gui/config/ConfigScreen.java
```diff
@@ -0,0 +1,232 @@
...
+ * TODO:
+ * - Get someone's review on changes to ForgeConfigSpec for comments
+ * - Make ConfigScreen header size be smaller if it has a null subTitle?
+ * - Translucent scissored list?
+ * - Do I need to deal with sub-configs?
+ * - Fix lists
+ * -> Change indices instead of reloading the entire view
+ * -> Drag and drop movement?
+ * -> Maybe try and refactor it all to be less confusing and twisted
+ * - Client -> Server syncing?
```
ichttt, line 54: If a server config is being changed, it needs to be synced to the server and all clients. It should also (IMO) only be changeable by owners. So yes, we need syncing, but we already have a packet for this (although we might need a new one for in-game changes to config). Most likely it should also resync when the file is changed by the filesystem, which IIRC it doesn't do right now as well.

src/main/java/net/minecraftforge/client/gui/config/ControlCreator.java
```diff
@@ -0,0 +1,519 @@
...
+        else if (value instanceof Number)
+            createNumericInteractionWidget((Interactor<? extends Number>) interactor);
+        else if (value instanceof Enum<?>)
+            createEnumInteractionWidget((Interactor<Enum>) interactor);
+        else if (value instanceof String)
+            createStringInteractionWidget((Interactor<String>) interactor);
+        else if (value instanceof List<?>)
+            createListInteractionWidget((Interactor<List<?>>) interactor);
+        else
+            throw new IllegalStateException("Don't know how to make a widget for " + value);
```
ichttt, line 183: How likely is this going to happen? Can this happen with a mod using custom objects?

src/main/java/net/minecraftforge/client/gui/config/ControlCreator.java
```diff
@@ -0,0 +1,519 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    public <T> T[] createHolderForUseInLambda(T value) {
```
ichttt: We typically use AtomicWhatever or MutableWhatever (from apache common langs) for use in lambdas

ichttt (commented):
Some early notes. I know this is still a draft and heavy WIP (I saw many Todos :P), so it's just a very rough review from a very quick look

maxanier:
I didn't look into the code yet, but I tested this with a mod of mine:
It works great out of the box, only:
- If you have multiple configuration specs/files per type only the latter are displayed in the menu.
```
        ModLoadingContext.get().registerConfig(ModConfig.Type.SERVER, serverSpec);
        ModLoadingContext.get().registerConfig(ModConfig.Type.SERVER, balanceSpec, "vampirism-balance.toml");
```
Here only options from balanceSpec are displayed
- Cannot edit server configuration while hosting a SP world to LAN (maybe that is intentional or would require to much work/hacks)

I might look into the code when  have more time

src/main/java/net/minecraftforge/client/gui/config/CategoryConfigScreen.java
Cadiboo: Ya true.
In Forge's configs, all values are stored as `ConfigValue<T>`s with their corresponding `ValueSpec` stored in a sibling map.
Except for categories, they're stored as raw `SimpleConfig`s (`SimpleCommentedConfig`s after my changes) with no corresponding `ValueSpec`.
This means that comments are handled separately (in a map `levelComments` before my changes and that translation keys are not possible).
I'll add this explanation into the code.
It was initially so vague because I want to talk through my changes with someone (that improve comments) and also talk about how translation keys can be implemented and how to also allow them to store extra data in the future.

src/main/java/net/minecraftforge/client/gui/config/CategoryConfigScreen.java
Cadiboo: It's a mostly internal thing that is likely change a bit before I merge this PR. I don't see any reason for this to be in it's own class as it's just used like a `Record` to pass around functions. I should probably make this into a POJO rather than an interface though...

src/main/java/net/minecraftforge/client/gui/config/ConfigScreen.java
Cadiboo: > If a server config is being changed, it needs to be synced to the server and all clients.
> Most likely it should also resync when the file is changed by the filesystem, which IIRC it doesn't do right now as well.

Yes, I was assuming that I would just need to send a packet to the server, it would save & reload the config and then that the server -> all clients syncing was already implemented. If it isn't, it shouldn't be too hard to implement.

> It should also (IMO) only be changeable by owners

Definitely, I was thinking changeable by OPs

> we already have a packet for this

We have one for Server -> Client sync during the handshake, I think this one could be repurposed for in-game updates but I'm not sure... We definitely need a new one for Client -> Server config update request.
I've left out the code for syncing right now because I know it will be a bit of work to implement.

src/main/java/net/minecraftforge/client/gui/config/ControlCreator.java
Cadiboo: Unlikely.
I've covered all the common values (booleans, numbers, strings, enums) + lists. I haven't yet implemented `Config`s + if something similar to #6464 is merged this (specifically `createNumericInteractionWidget`) will need to be extended to support the new types.
However, any `Exception`s thrown by this method are caught in the `tryCreateInteractor` methods in `CategoryConfigScreen` and `ListConfigScreen`.

Mod's using custom objects may also trigger this, but any exceptions will be caught and a simple, uneditable, `toString()` representation will be displayed instead of a widget with a tooltip telling the user to edit the value in the config file.

src/main/java/net/minecraftforge/client/gui/config/ControlCreator.java
Cadiboo: `MutableWhatever` seems like that class I couldn't find. I'll fix this up, ty!

Cadiboo:
> If you have multiple configuration specs/files per type only the latter are displayed in the menu.

I'll look into fixing this. A glance at `ConfigTracker` shows that it wasn't set up to handle multiple configs. In fact, If you run the command `/config showfile SERVER` you'll see that only one file is printed. I'm not sure you're meant to be using multiple configs like that.

> Cannot edit server configuration while hosting a SP world to LAN (maybe that is intentional or would require to much work/hacks)

It's intentional right now. SP non-lan works because there's only one player and I can ignore the syncing (maybe, it might be an issue there too but it *appears* to work) and just reach across logical sides (ew!). I need to actually add some packets n stuff to make it work in non-singleplayer. See ichittt's comment https://github.com/MinecraftForge/MinecraftForge/pull/7395#discussion_r499278792.

src/main/java/net/minecraftforge/client/gui/config/ConfigScreen.java
Minecraftschurli: > Definitely, I was thinking changeable by OPs

Would be cool if you would utilize the forge permission api for that, so that people are encouraged to use it more

ChampionAsh5357:
Are the allowed values only specified on the tooltip and then validated on save? That's what I see anyways. If so, then maybe something similar to `ISuggestionHelper` in commands should be used on the text field widgets to allow autofill on the fields and to notify the user if an entered text is invalid.

Cadiboo:
> Are the allowed values only specified on the tooltip and then validated on save?

The widgets content is validated whenever it changes (and then also on save) but semantics. If it’s invalid, the widget will turn red. I can add some code to also make the tooltip say it’s invalid and make it’s label turn red (`addUpdateResponder` in the config category list class).

> If so, then maybe something similar to `ISuggestionHelper` in commands should be used on the text field widgets to allow autofill on the fields and to notify the user if an entered text is invalid.

Sounds great to me. Would require a couple minor changes to ForgeConfigSpec, but I’m not sure how hard it would be to implement in the text widgets. I’m not going to do it if it’s very hard to make auto completion work.

src/main/java/net/minecraftforge/client/gui/config/CategoryConfigScreen.java
```diff
@@ -233,6 +273,42 @@ protected void addExtraTooltipInfo(ValueSpec valueInfo, List<ITextProperties> to
             }
         }
 
+        protected void addInvalidityWarningHandler(Interactor<?> interactor, List<ITextProperties> tooltip) {
+            final ITextComponent tooltipWarning = new TranslationTextComponent("forge.configgui.tooltip.invalidValue").func_240701_a_(TextFormatting.BOLD, TextFormatting.RED);
+            MutableObject<Boolean> warningActive = new MutableObject<>(false);
```
ichttt, line 278: `MutableBoolean` and similar classes exist for primitive types

src/main/java/net/minecraftforge/client/gui/config/CategoryConfigScreen.java
```diff
@@ -0,0 +1,370 @@
...
+        }
+
+        Collection<String> elements();
+
+        /**
+         * Warning: Can return null if you call it with a key that isn't in the collection returned from {@link #elements()}
+         *
+         * @return Either a {@link Config} (if the key corresponds to a category) or a {@link ConfigValue} (if the key corresponds to a value).
+         */
+        Object getValue(String key);
```
ichttt, line 137: Can we use DFU's Either here to make things a little cleaner?

src/main/java/net/minecraftforge/client/gui/config/ControlCreator.java
```diff
@@ -0,0 +1,525 @@
...
+        else if (value instanceof String)
+            createStringInteractionWidget((Interactor<String>) interactor);
+        else if (value instanceof List<?>)
+            createListInteractionWidget((Interactor<List<?>>) interactor);
+        else
+            throw new IllegalStateException("Don't know how to make a widget for " + value);
+    }
+
+    public void createBooleanInteractionWidget(Interactor<Boolean> interactor) {
+        MutableObject<Boolean> state = new MutableObject<>();
```
ichttt, line 187: Again, MutableBoolean

src/main/java/net/minecraftforge/client/gui/config/ControlCreator.java
```diff
@@ -0,0 +1,525 @@
...
+                throw new IllegalStateException("Don't know how to make a widget for Non-ranged Number " + value);
+        }
+    }
+
+    /**
+     * @return null if a slider should not be created for the interactor.
+     */
+    @SuppressWarnings("unchecked")
+    @Nullable
+    public <T extends Number, C extends Number & Comparable<? super C>> Range<C> getRangeForSliderCreation(Interactor<T> interactor) {
```
ichttt, line 240: Should be able to inline `T`

src/main/java/net/minecraftforge/client/gui/config/ControlCreator.java
```diff
@@ -0,0 +1,525 @@
...
+        Range<C> range = (Range<C>) interactor.getData(RANGE_KEY);
+        if (range == null)
+            return null;
+        double min = range.getMin().doubleValue();
+        double max = range.getMax().doubleValue();
+        if (max - min > 10)
```
ichttt, line 246: This seems like an arbitary bound... At least put it into a `private static final int` and add some clarification why 10 seems like a good value

src/main/java/net/minecraftforge/client/gui/config/ControlCreator.java
```diff
@@ -0,0 +1,525 @@
...
+        MutableObject<String> currentlyRespondingTo = new MutableObject<>();
+        textField.setResponder(newText -> {
+            if (Objects.equals(currentlyRespondingTo.getValue(), newText))
+                // We are being called from inside visualsUpdater, don't recurse & stackoverflow
+                return;
+            T newValue;
+            try {
+                newValue = parser.apply(newText);
+            } catch (NumberFormatException ignored) {
+                ((TextFieldWidget) interactor.control).setTextColor(RED);
```
ichttt, line 290: Shouldn't this get a red border? In case the text is empty

src/main/java/net/minecraftforge/client/gui/config/ControlCreator.java
```diff
@@ -0,0 +1,525 @@
...
+    @SuppressWarnings("unchecked")
+    public static <T> T copyMutable(T o) {
+        // Immutable objects can be returned as is
+        if (o instanceof Boolean || o instanceof Number || o instanceof String || o instanceof Enum<?>)
+            return o;
+        if (o instanceof List<?>)
+            return (T) new ArrayList<>((List<?>) o);
+        if (o instanceof UnmodifiableConfig)
+            return (T) ((AbstractConfig) o).clone();
+        return o;
```
ichttt, line 457: Should we log something here, or are there legitimate cases where the code flow could reach here?

src/main/java/net/minecraftforge/client/gui/config/ControlCreator.java
```diff
@@ -0,0 +1,525 @@
...
+    public TextFieldWidget createTextField(Interactor<?> interactor) {
+        return new ConfigElementTextField(interactor.title);
+    }
+
+    public <T extends Number> Slider createSlider(Interactor<T> interactor, Range<? extends Number> range, Slider.ISlider iSlider) {
+        T value = interactor.initialValue;
+        double min = range.getMin().doubleValue();
+        double max = range.getMax().doubleValue();
+        boolean showDecimals = value instanceof Double || value instanceof Float;
+        // I don't think a slider should be this hard to set up properly, maybe it could do with a refactor?
```
ichttt, line 473: Agreed, it's annoying. Maybe add a TODO 1.17 comment, and leave it as-is now

src/main/java/net/minecraftforge/client/gui/config/ListConfigScreen.java
```diff
@@ -0,0 +1,168 @@
...
+        return new ListConfigElementList(this, field_230706_i_);
+    }
+
+    protected void setAndNotify(int index, Object newValue) {
+        value.set(index, newValue);
+        onModified(value);
+    }
+
+    protected void onAddRemove(int index) {
+        // TODO: Don't take the lazy way out
```
ichttt, line 64: Shouldn't matter that much, doesn't seem like a performance hot spot

src/main/java/net/minecraftforge/client/gui/config/ListConfigScreen.java
```diff
@@ -0,0 +1,168 @@
...
+            ControlCreator creator = configScreen.getControlCreator();
+            Interactor<?> interactor = new Interactor<>(title, item);
+            interactor.addData(ControlCreator.INITIAL_VALUE_KEY, item);
+            interactor.addData(ControlCreator.DEFAULT_VALUE_KEY, item);
+            interactor.addSaver(newValue -> setAndNotify(index, newValue));
+            try {
+                creator.createAndInitialiseInteractionWidget(interactor);
+                return interactor;
+            } catch (Exception e) {
+                return null;
```
ichttt, line 113: Catching raw `Exception` and silently discarding it isn't the nicest thing, why is this needed?

src/main/java/net/minecraftforge/client/gui/config/ModConfigScreen.java
```diff
@@ -0,0 +1,202 @@
...
+        for (ModConfig modConfig : allConfigs) {
+            if (modConfig.getType() == ModConfig.Type.SERVER && !canPlayerEditServerConfig())
+                continue;
+            configsToDisplay.add(modConfig);
+        }
+        return new ConfigElementList(this, field_230706_i_) {
+            {
+                for (ModConfig modConfig : configsToDisplay) {
+                    String translationKey = "forge.configgui.modConfigType." + modConfig.getType().name().toLowerCase();
+                    IFormattableTextComponent title = CategoryConfigScreen.translateWithFallback(translationKey, StringUtils.capitalize(modConfig.getType().name().toLowerCase()));
```
ichttt, line 116: Always use `Locale.ROOT` when using toLowerCase is such context

src/main/java/net/minecraftforge/common/ForgeConfig.java
```diff
@@ -52,62 +52,62 @@
 
         Server(ForgeConfigSpec.Builder builder) {
             builder.comment("Server configuration settings")
-                   .push("server");
```
ichttt, line 55: Please leave all the indentation as-is, it just bloats the already huge diff

src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
```diff
@@ -403,17 +414,18 @@ public Object correct(Object value) {
...
+            context.setComment(ObjectArrays.concat(context.getComment(), "Allowed Values: " + allowedValues));
+            return new EnumValue<>(this, define(path, new ValueSpec(defaultSupplier, validator, context), defaultSupplier).getPath(), defaultSupplier, converter, clazz);
         }
 
         //boolean
         public BooleanValue define(String path, boolean defaultValue) {
             return define(split(path), defaultValue);
         }
         public BooleanValue define(List<String> path, boolean defaultValue) {
-            return define(path, (Supplier<Boolean>)() -> defaultValue);
```
ichttt, line 416: These casts may be needed for eclipse compiler, please leave them

src/main/java/net/minecraftforge/fml/DistExecutor.java
```diff
@@ -46,7 +46,7 @@
     private DistExecutor() {}
 
     /**
-     * Run the callable in the supplier only on the specified {@link Side}.
+     * Run the callable in the supplier only on the specified {@link Dist}.
```
ichttt, line 49: Please leave this class alone, it is not in the scope of this PR

src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
```diff
@@ -60,23 +61,20 @@
...
 
-/*
+/**
  * Like {@link com.electronwill.nightconfig.core.ConfigSpec} except in builder format, and extended to accept comments, language keys,
  * and other things Forge configs would find useful.
  */
-public class ForgeConfigSpec extends UnmodifiableConfigWrapper<UnmodifiableConfig> //TODO: Remove extends and pipe everything through getSpec/getValues?
+public class ForgeConfigSpec extends UnmodifiableConfigWrapper<UnmodifiableCommentedConfig> //TODO: Remove extends and pipe everything through getSpec/getValues?
 {
-    private Map<List<String>, String> levelComments = new HashMap<>();
```
ichttt, line 69: About translations for categories: We could add a map for cat path -> translation key, just like we had with `levelComments` before

ichttt (changes requested):
This should be a fairly complete review of this PR.
About the changes to ForgeConfigSpec: Those seem mostly fine to me, I added a suggestion on how to handle translation keys for categories.
Mostly looking good IMO though

src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
Cadiboo: My thoughts as well

src/main/java/net/minecraftforge/fml/DistExecutor.java
Cadiboo: I figured it would be easier to bundle these small docs changes in this PR, Is it really worth splitting it off into its own PR?

src/main/java/net/minecraftforge/client/gui/config/ControlCreator.java
Cadiboo: Yeah, that would be much better solution than changing the text color. I’ll see if I can do that. If I can I’ll also do it for buttons - the text color is super ugly IMO

src/main/java/net/minecraftforge/client/gui/config/ControlCreator.java
Cadiboo: I’m not sure. I figured it was better to fail silently and have possible weird behaviour with mutables than fail causing the game to crash when the config screen was opened or logging a message that would always be ignored.

src/main/java/net/minecraftforge/client/gui/config/ControlCreator.java
Cadiboo: It’s very arbitrary, I’m not unhappy about it though - if people don’t like it I want them to change it by supplying their own ControlCreator for their config screen. The aim then is to make it as easy as possible to change which I think I have done.

src/main/java/net/minecraftforge/client/gui/config/ListConfigScreen.java
Cadiboo: It’s not, the issue is that whenever someone adds/removes an element the screen refreshes from scratch, losing the “undo” ability for any changes they’d made before adding/removing.
I plan to generify this class so it can be a superclass for both List screens and Config screens (raw ones from NightConfig, effectively `Map<String, ?>`s).

src/main/java/net/minecraftforge/client/gui/config/ListConfigScreen.java
Cadiboo: The user won’t care about whatever went wrong and they won’t want their game to crash. Catching the exception and making the element just say “please edit this in the config file” seemed clean to me. It catches the exceptions from being unable to create a widget (e.g. an unknown value gets passed in, there’s a bug in my code or a bug in a mod that uses the ControlCreator api’s code)

src/main/java/net/minecraftforge/common/ForgeConfig.java
Cadiboo: Heh, you should see my old PR

src/main/java/net/minecraftforge/client/gui/config/CategoryConfigScreen.java
Cadiboo: I can’t find out what a DFU is, I assume it’s not a Device Firmware Update which is all google will show me.

TheCurle: Data Fixer-Upper, it's a Mojang tool that... is hard to explain.

ichttt: There is a class in the DFU lib that's called Either, it can be used instead of raw object if exactly two completly different types are possible

James-E-A:
Will this PR break compatibility with current mods' programmed configs?

e.g. this currently blocks squeek502/AppleSkin#85 — how much (_if any_) refactoring will this and similar devs have to do of [their existing code](https://github.com/squeek502/AppleSkin/blob/1.12/java/squeek/appleskin/ModConfig.java) to get their config GUIs working again?

Will you support the "legacy" API?

GirafiStudios:
> Will this PR break compatibility with current mods' programmed configs?
> 
> e.g. this currently blocks [squeek502/AppleSkin#85](https://github.com/squeek502/AppleSkin/issues/85) — how much (_if any_) refactoring will this and similar devs have to do of [their existing code](https://github.com/squeek502/AppleSkin/blob/1.12/java/squeek/appleskin/ModConfig.java) to get their config GUIs working again?
> 
> Will you support the "legacy" API?

I think you've misunderstood something. There is currently NO code for the in-game config in Forge, so nothing will break.
Mods will have to add support for this on their end, when the PR gets accepted though.

James-E-A:
> > Will you support the "legacy" API?
> 
> I think you've misunderstood something. There is currently NO code for the in-game config in Forge, so nothing will break.
Mods will have to add support for this on their end, when the PR gets accepted though.

I believe I have not misunderstood anything:

I'm referring to [the GUI code](https://github.com/MinecraftForge/MinecraftForge/blob/1.12.x/src/main/java/net/minecraftforge/fml/client/config/GuiConfig.java) that was dropped in 1.13 — the "legacy" config GUI API from 1.12 — some mods (e.g. AppleSkin, linked previously) retain their GUI config declarations which were coded against this API.

I am asking if this PR will resume support, "drop-in", non-breaking (or minimally-breaking) for [such legacy code](https://github.com/squeek502/AppleSkin/blob/1.12/java/squeek/appleskin/ModConfig.java) _[that is still coded against the 1.12 config GUI API]_, or if all mod developers will be required to refactor their GUI config declarations against the new config GUI API.

ChampionAsh5357:
The GUI is associated with the configuration toml. If they haven't already rewritten their configs, then of course they would need to be rewritten. The GUI itself has no bounds on the config as its dependent on it. Comparing pre-1.13 to post-1.13 will most likely include a full rewrite of the mod anyways.

So, yes, I believe you did misunderstand. This it not to mention the fact that Forge doesn't support 'legacy' versions.

XFactHD:
@JamesTheAwesomeDude The code you are referencing can not be used with any 1.13+ version of Forge because the config system itself has been rewritten completely. Therefore any "existing" config GUI code will need to be rewritten.

@GirafiStudios That is not true. Forge provides an ExtensionPoint where you can register your own implementation of a Config GUI, you do not need this PR for that. The "only" thing this PR does is providing a default implementation of a Config GUI.

Cadiboo:
I’ll fix up the merge conflicts on this but IIRC I’m still waiting for Lex/cpw’s comments about the changes to ForgeConfigSpec and about how translation keys on categories should be supported.

LexManos:
Honestly, this is a major thing that will break every Minecraft version.
It's better to be split off into a custom mod/library that is managed by someone who has the time and the patience to deal with rendering crap.
The config systems that we use are designed to have all of the metadata necessary for these generic systems to exist.
If we do not have the hooks in place {I believe we do} to allow such a mod to provide the GUI's for other mods that don't provide their own. Then we can add that.
Also, IF this mod does come into existence, it could be published with our installer as a default option. But that's something we'd have to talk to the author about.

Basically, Config GUIs are nice. But Mojang likes to break their rendering system to much for us to sanely manage them.

PR #7398, opened 2020-10-05
Additions to the mdk build.gradle to accommodate datagens.

Flanks255:
Adds the --existing entry to the args of runData, and adds the generated resources folder to the sourceSets.

To give people a better time getting started using datagens.


mdk/build.gradle
```diff
@@ -19,6 +19,8 @@ archivesBaseName = 'modid'
 
 sourceCompatibility = targetCompatibility = compileJava.sourceCompatibility = compileJava.targetCompatibility = '1.8' // Need this here so eclipse task generates correctly.
 
+sourceSets.main.resources { srcDir 'src/generated/resources' }
```
marchermans: This requires both `srcDir 'src/generated/resources'` as well as `srcDir 'src/main/resources'`.
Since right now you only allow the generated resources to be used.

marchermans (changes requested):

mdk/build.gradle
Flanks255: This line is adding the generated resources to  sourceset, not replacing. The normal resources are in the sourceset by default, adding them a second time would be redundant.

marchermans: See that is what i am not sure about. I remember needing += as an operator inbetween.
Let me check.

marchermans: Checked the API, this is indeed correct: https://github.com/gradle/gradle/blob/master/subprojects/core-api/src/main/java/org/gradle/api/file/SourceDirectorySet.java#L50

marchermans (approved):

marchermans (changes requested):
@Flanks255 We will need comments on these additions explaining what each line does.
Even if previously PRs where merged that might not have added comments on the lines you touched, I would ask you to add them there.

raoulvdberge:
This comment is kinda useless:
```java
//Include generated resources.
sourceSets.main.resources { srcDir 'src/generated/resources' }
```

What about `// Include generated resources by datapacks.`?

Flanks255:
I'm not quite sure what the generated resources have todo with datapacks?

ahaim5357:
This PR should wait in my opinion. If you want to do this, you should figure out how to add all the files necessary such that the `ExistingFileHelper` can verify them properly, including forge ones. This just writes code that should be present based on where the generated file locations are.

ChampionAsh5357:
This PR should wait in my opinion. If you want to do this, you should figure out how to add all the files necessary such that the ExistingFileHelper can verify them properly, including forge ones. This just writes code that should be present based on where the generated file locations are.

Apologies, posted with the incorrect account.

raoulvdberge:
> I'm not quite sure what the generated resources have todo with datapacks?

Sorry, meant datagens.

marchermans:
@ChampionAsh5357 This has nothing todo with where forge files are.

The data gen run is related to your mods datageneration, aka its datapack it has no bearing on the forge datapack nor on any other mods datapack. If you include in your datapack entries in the forge namespace then this will work, and this change will pick them up. As such I see no reason to wait here. This is a good change and something newer modders run into on a daily basis.

marchermans (approved):

PR #7400, opened 2020-10-06
Throw IOException when mod file not found in ModJarUrlHandler

IzzelAliz:
An IOException should thrown when resources are missing, instead of a NPE.

Also return a non null Optional in getManifest.

### DIFF
--- a/src/fmllauncher/java/net/minecraftforge/fml/loading/ModJarURLHandler.java
+++ b/src/fmllauncher/java/net/minecraftforge/fml/loading/ModJarURLHandler.java
@@ -19,6 +19,7 @@
 
 package net.minecraftforge.fml.loading;
 
+import net.minecraftforge.fml.loading.moddiscovery.ModFileInfo;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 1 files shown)


cpw:
Why?

IzzelAliz:
> 
> 
> Why?

It is common to write like that

```java
try {
  byte[] bytes = readAll(new URL("some stuff").openStream());
  // do some stuff
} catch (IOException e) {
  // handle here
}
```
because IDE will generate this.

My code wants to recognize extensions by specific file in jar, and I found this url handler. When I tried to read this I got an unexpected NPE.

Though it is easy for me to add another catch, I think throwing IOException here would be nice. Also some third party library that eats an url only expect throwing IOException, it is hard to change their code.

src/fmllauncher/java/net/minecraftforge/fml/loading/ModJarURLHandler.java
```diff
@@ -47,21 +48,25 @@ protected URLConnection openConnection(URL url) {
...
                 // trim first char
                 modpath = url.getPath().substring(1);
-                resource = FMLLoader.getLoadingModList().getModFileById(modid).getFile().findResource(modpath);
-                manifest = FMLLoader.getLoadingModList().getModFileById(modid).getManifest();
+                ModFileInfo modFileInfo = FMLLoader.getLoadingModList().getModFileById(modid);
+                if (modFileInfo == null) {
+                    throw new IOException("No mod file found for " + modid);
+                }
+                resource = modFileInfo.getFile().findResource(modpath);
+                manifest = modFileInfo.getManifest().orElse(null);
```
KnightMiner, line 69: Why do you unpack this then just repack it as an optional in the getter? Does calling `orElse` possibly throw an IOException? If so since we return an optional later anyways, it may make sense to repack the return here and keep the field type the same

Alterantively, are you protecting against connect never being called/failing above? Maybe it makes more sense to default `manifest` to `Optional.empty()` in that case.

LexManos:
Why are you manually creating the url for mod files with invalid mod ids?
It seems the only thing that can return null is the call to getModFileById.
Also, using exceptions are code flow is a bad idea. There are ways to check if a file exists before just blinding trying to read it.

IzzelAliz:
> 
> 
> Why are you manually creating the url for mod files with invalid mod ids?
> It seems the only thing that can return null is the call to getModFileById.
> Also, using exceptions are code flow is a bad idea. There are ways to check if a file exists before just blinding trying to read it.

So I checked it now, but I think it would be nice if it is throwing proper exceptions.

PR #7401, opened 2020-10-06
[1.15] Throw IOException when mod file not found

IzzelAliz:
Backport fix of #7400


marchermans:
The 1.16 was closed and not accepted, see #7400 for more details.

PR #7403, opened 2020-10-08
Fix (NonNull)Lazy.Concurrent using a global lock

malte0811:
This prevented multiple threads from resolving independent `Lazy`s at the same time. Implementation is as discussed in the Forge Discord yesterday, starting at [this](https://discordapp.com/channels/313125603924639766/725850371834118214/763492119313907714) message.

### DIFF
--- a/src/main/java/net/minecraftforge/common/util/Lazy.java
+++ b/src/main/java/net/minecraftforge/common/util/Lazy.java
@@ -78,7 +78,7 @@ public interface Lazy<T> extends Supplier<T>
      */
     final class Concurrent<T> implements Lazy<T>
     {
-        // Also used as the lock to reduce object size
+        private volatile Object lock = new Object();
         private volatile Supplier<T> supplier;
         private volatile T instance;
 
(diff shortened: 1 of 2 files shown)


diesieben07:
Suppliers are usually created using lambdas, which are not intended to be synchronized on usually as they have no clearly defined object identity (i.e. it is not specified whether you will get a new object every time or not, etc.).

malte0811:
I guess I learned something new today, I always assumed every lambda was its own object. But even if that was the case it would be valid to create multiple `Lazy`s from the same `Supplier`. I see two possible solutions:
 - Use a `lock` member, and possibly implement the "loose the lock" idea suggested in the Discord conversation I linked to
 - Create a new supplier object in the constructor, using an anonymous class rather than a lambda to ensure a new object for each `Lazy`.

The first solution is a lot cleaner, and the only advantage of the second one would be a slightly reduced size of `Lazy` after the value has been resolved. Unless someone objects to it I will implement the first alternative later today.

KnightMiner:
Isn't the lazy instance already a new supplier? Why not synchronize over `this`?

malte0811:
That was my first idea too. But if you google it it's apparently considered to be bad practice, because API users can also decide to synchronize on that object, potentially causing deadlocks. The recommended behavior is to synchronize on an internal object which is never seen by "outside" code (which my current implementation admittedly does not do either).

gigaherz:
Oops forgot about this.

PR #7404, opened 2020-10-08
Add new way to specify a custom background texture

ichttt:
Currently, you can only specify a custom name, as you would have to overwrite a client-only method to change the namespace, which isn't quite nice to do.
This patch adds a new way to specify the complete RL, so modders don't need to ship their textures in the minecraft namespace.

In addition to the obvious, this also speeds up the loading of the mods by a tiny bit, as the only thing in the minecraft namespace in assets in mods is mostly the tab background texture, and we have an optimization in DelegatingResourcePack that skips Files.exists if a mod doesn't have any resources in a certain namespace (as Files.exist can take up to a millisecond depending on the setup)

### DIFF
--- a/patches/minecraft/net/minecraft/item/ItemGroup.java.patch
+++ b/patches/minecraft/net/minecraft/item/ItemGroup.java.patch
@@ -1,6 +1,14 @@
 --- a/net/minecraft/item/ItemGroup.java
 +++ b/net/minecraft/item/ItemGroup.java
-@@ -107,12 +107,15 @@
+@@ -101,18 +101,23 @@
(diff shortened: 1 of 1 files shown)


ChloeDawn:
Why were the String methods deprecated in this patch? They are still useful to modders, for example `this.setBackgroundImageName("item_search.png")`. This is now reported as a deprecated call, expecting the modder to use the ResourceLocation method and re-specify the entire `minecraft` path...

LexManos:
Because we change what modders SHOULD be using. So they need a notification to use the more specific version. Even if they wanna target the minecraft namespace.

PR #7410, opened 2020-10-09
[1.16.3] Provide an entry point for custom data systems to sync.

Shadows-of-Fire:
**The issue:**
At present, custom data pack systems have no effective entry point to sync their data.  It can either be synced from within the reload listener, or from some other arbitrary live handler at a later point in time.  This PR provides an event so that mods that add custom systems can sync their data appropriately at the same time as vanilla.  This includes syncing on login (for initial) as well as during the reload to all active players.

**Why two subclasses?**
Some mods custom systems may depend on certain data being available to the client.  The issue that prompted this PR was that I have a system where enchanting stats (for blocks, things such as enchanting power bonus) is loaded from datapacks.  This information is displayed on item tooltips.  However, on dedicated servers, I cannot effectively relay this tooltip information as it cannot be provided early enough to be indexed (JEI calculates search trees in the RecipesUpdatedEvent). 

The first subclass (BeforeRecipes) allows data to be sent before the server sends out the recipes and tag packets.  This would ensure that sent data is received before the client processes recipe and tag packets.
The second subclass (AfterRecipes) allows data to be sent out after the server sends out this data, which is important if the client would require this information in order to be able to deserialize custom system data.

**An aside:**
When doing this, I noticed that upon initial connection, vanilla syncs the recipe data before the tag data, but during datapack reload, tags are synced before the recipes are.  I am not sure if this actually matters, but it is something that could be changed if desired.


Shadows-of-Fire:
This is still required to resolve the issue of sending data before recipe sync

Shadows-of-Fire:
Made obsolete by #7985

PR #7412, opened 2020-10-11
Create fakeplayer handling

Raycoms:
At the moment there is a serious mismatch between permission mods and the use of fakeplayers.

The problem mostly arises from the fact that mods reuse the same fakeplayer instance for all their actors.
This way there is no way for a permission mod to check the owner of a fakeplayer in an appropriate way.

I, therefore, propose those changes.
The idea is that developers may store the owner in their tileEntities and before using the fakeplayer in an action may store the owners UUID temporarily in the global fakeplayer (the next tile would then change that and so forth).

### DIFF
--- a/src/main/java/net/minecraftforge/common/util/FakePlayer.java
+++ b/src/main/java/net/minecraftforge/common/util/FakePlayer.java
@@ -19,6 +19,7 @@
 
 package net.minecraftforge.common.util;
 
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 import com.mojang.authlib.GameProfile;
@@ -26,6 +27,7 @@ import com.mojang.authlib.GameProfile;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.entity.player.ServerPlayerEntity;
+import net.minecraft.nbt.CompoundNBT;
 import net.minecraft.network.play.client.CClientSettingsPacket;
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.management.PlayerInteractionManager;
@@ -42,6 +44,11 @@ import java.util.UUID;
 //Preliminary, simple Fake Player class
 public class FakePlayer extends ServerPlayerEntity
 {
+    /**
+     * UUID of the player responsible for the FakePlayers actions.
+     */
+    private UUID owner = null;
+
     public FakePlayer(ServerWorld world, GameProfile name)
     {
         super(world.getServer(), world, name, new PlayerInteractionManager(world));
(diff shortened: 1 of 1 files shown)


SquidDev:
I feel a cleaner approach would just for mods to construct their `FakePlayer` on a per-tile basis (rather than globally), and instantiate it with the owning player's `GameProfile`. 

I believe this is what mods like BuildCraft and ComputerCraft do, and it seems to work pretty well. More importantly, it doesn't require any special handling of fake players by permission mods. I know this approach worked very well with Sponge in the past (though less of a concern on 1.16).

As you point out, the issue is that most mods don't do this right now. Though improving documentation to encourage this approach may be better than adding _another_ way to provide a game profile/UUID.

KnightMiner:
It is important to note that some fake players are not associated with a player intentionally, as they are meant to mimic behavior such as a piston which does not tie actions to a player. Whatever solution is made needs a way for permission mods to handle that. Really, a permission mod if it detects a fake game profile should just handle it the same way they do pistons, and if the action is tied to a player the mod writer should use a real game profile.

Raycoms:
@SquidDev We had this discussion with Lex 5 years ago. Back then he said that these mods should create a fakeplayer per player and give the fakeplayer the players UUID. However, so far NO mod at all follows this approach. I discussed this with a few tech-mod authors and this was the compromise everyone could agree on.

@KnightMiner exactly, that's why the owner field is nullable after all.

Shadows-of-Fire:
I don't like this approach.  It has significant shortcomings when people are using a shared fakeplayer (which appears to be the entire point...?) as that fakeplayer would no longer be "the same" shortly after execution if there are multiple things executing by different owners at once.  Having each tile construct it's own fakeplayer should be the ideal approach, as well as having mods do one of the following for "global" non-player actions that are internal:
A) All "internal" actions are executed by [Minecraft].
B) Each mod declares a [\<modid>] named player, and executes it's global actions via that one, so permission mods can declare explicit mod actions as permitted or not.
C) Even more specifically, each specific "internal" action has it's own fake player, like [\<modid>_\<action>] and then those are whitelisted on an individual basis.

Adding this field doesn't solve the problem that people are not respecting the way that these actions need to be handled, and essentially complicates it even more, by then requiring extra checks for fake players.

Raycoms:
The idea of this approach is this:
TileEntity of modA gets FakePlayer from public static field.
TileEntity calls setOwner
TileEntity executes action with the FakePlayer
Next TileEntity grabs FakePlayer -> does the same.


Especially B) Doesn't work. Permission mods do NOT want to whitelist fakeplayers globally. I don't want people to place machines on claim borders and allow the machines to enter the claims.

And, as I already said, no one creates a new fakeplayer per tileEntity etc (besides it being a lot of overhead too to do that).

Raycoms:
Currently the approach I described above doesn't work, because you can't easily change the fakeplayers own UUID without side effects in the world. However, with the owner field that is sideeffect free and performs very well.

LexManos:
We are not going to add a threading nightmare just because modders do things wrong. And it shouldn't be a Fake Player per te. It should be a fake player per player you care about. Cache it based on uuid.

PR #7415, opened 2020-10-13
Fix worldgen object lists erroring

Cyborgmas:
This causes the "not a JSON object" error in the log, and is caused because the codec used wants a list of either ALL resource locations or ALL json objects. This PR changes the codec to have the either be applied to each individual element (instead of the whole list). From the mojang name of this function, this goes against the vanilla intention, which could have unintented consequences.

Also, the resource location decoding should happen fine as long as all objects are registered (to the vanilla registry) correctly. There have been some cases where I was not able to pinpoint the error, so this might still be needed. This fix could also hide potential problems in mod code, and not error.

The purpose of the PR is mainly to let people know the problem is known, and the fix for now is (theoretically) to register the objects.

### DIFF
--- a/patches/minecraft/net/minecraft/util/registry/RegistryKeyCodec.java.patch
+++ b/patches/minecraft/net/minecraft/util/registry/RegistryKeyCodec.java.patch
@@ -1,0 +1,23 @@
+--- a/net/minecraft/util/registry/RegistryKeyCodec.java
++++ b/net/minecraft/util/registry/RegistryKeyCodec.java
+@@ -19,17 +19,17 @@
+    }
+ 
+    public static <E> Codec<List<Supplier<E>>> func_244328_b(RegistryKey<? extends Registry<E>> p_244328_0_, Codec<E> p_244328_1_) {
+-      return Codec.either(func_244325_a(p_244328_0_, p_244328_1_, false).listOf(), p_244328_1_.<Supplier<E>>xmap((p_244329_0_) -> {
++      return Codec.either(func_244325_a(p_244328_0_, p_244328_1_, false), p_244328_1_.<Supplier<E>>xmap((p_244329_0_) -> {
+          return () -> {
+             return p_244329_0_;
+          };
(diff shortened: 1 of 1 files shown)


TelepathicGrunt:
To be clear for everyone, this affect ConfiguredFeatures, StructureFeatures, and ConfiguredCarvers. My understanding is this PR will fix the issue of any of them being unregistered causing them to wipe out registered ones from the biome. 

Though I do want to point out that some mods would like all the stuff to always be registered. For example, World Blender allows people to blacklist features to import from other biomes into its own dimension. If the feature is not registered, the resourcelocation based blacklist will not work. Some other mods might use the resourcelocation for other stuff as well such as detecting if a specific biome has a certain mod's structure on (for whatever reason).

I'm totally fine with this PR but I think it should also detect unregistered stuff and just advise the modder that registering anyway is a good idea.

TelepathicGrunt:
After trying to help tons and tons of people register their configuredfeatures, I think it's almost a lost cause at this point. That or I am burned out lol. I think it might be a good time to see about getting this pr merged and end this issue once and for all. Modders can register later if they wish for players to use the modded configured stuff in custom datapacks

TheCurle:
This PR being merged will not end this issue. 
This PR may end up making the issue worse.
The solution is still to register these features.

Cyborgmas:
Yeah I opened this as a PR but its rly just to say that the root cause is known, but this is in no way the proper fix. 
For example, it breaks mojang's codec format. Something that worked in vanilla will work with this codec no problem, but a vanilla client would not have this format so it would fail to parse it if it connected to a forge server. So the question is, will a vanilla client actually parse such a list received from the server. (NetworkCodecs omit some info). 
That would need testing as well as a bunch of other things before this could actually be considered (in my opinion). (and I am not *that* motivated to do the tests)

Cyborgmas:
Tried making a better version of the codec: #7582 . Objects will still need to be registered, but it is those unregistered that will pay the price (and not be in the world) which is a better solution imo.

PR #7417, opened 2020-10-15
[1.16.x] Add event for fabulous graphics rendering

Ocelot5836:
**The problem at hand**
After the introduction of the Fabulous Graphics setting, it has been near impossible to use the `RenderLastEvent` without experiencing some sort of visual issues. This is due to the fabulous shader being rendered before the event is fired so new render changes are never drawn to the screen.
**This PR adds:**
- A new event for rendering that allows rendering into fabulous frame buffers
- Patches WorldRenderer
- A test mod

I considered just moving `RenderLastEvent`, but that could have unintended consequences so I opted to make a new event. Technically `RenderLastEvent` _can_ be moved but it would have the chance of breaking mods that use it.

### DIFF
--- a/patches/minecraft/net/minecraft/client/renderer/WorldRenderer.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/WorldRenderer.java.patch
@@ -87,7 +87,7 @@
           RenderState.field_239237_T_.func_228549_b_();
        } else {
           iprofiler.func_219895_b("translucent");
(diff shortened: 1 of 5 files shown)


Nightenom:
Please consider #7225 as it should solve your issue, your event should be equal to RenderWorldPreWeatherEvent

Ocelot5836:
That other PR is far more extensive than this one is so I'm going to close this for now.

Issue #7421, opened 2020-10-18
[1.14+] Nullpointer Exception in FluidUtil.tryPlaceFluid


BluSunrize:
**Minecraft Version:** 1.15.2

**Forge Version:** 1.15.2-31.2.30, but likely to affect any other releases since 1.14.4 and up to 1.16.3

**Logs:** https://gist.github.com/BluSunrize/6d411b4bb8fdf1aa62e52841fbd55c8f

**Steps to Reproduce:**
 1. Register any Bucket with Forge's DispenseFluidContainer behavior
 2. Put that item in a dispenser
 3. Trigger the dispenser

**Description of issue:**
This was previously reported in #6275 and is not exclusive to that dispenser behavior.

The root cause lies in `FluidUtil.tryPlaceFluid` ([see here](https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/src/main/java/net/minecraftforge/fluids/FluidUtil.java#L535)), which has its PlayerEntity argument annotated as `@Nullable` when it most certainly is not.
The method proceeds to initialize an ItemUseContext with that player ([see here](https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/src/main/java/net/minecraftforge/fluids/FluidUtil.java#L553)) and that constructor makes a call to `player.world`, and will therefor result in a NullPointerException.

Dispenser behaviors trigger this, but so would basically anyone who tries to call `tryPlaceFluid` with null for a player. My personal recommendation here is to use the other constructor of `ItemUseContext`, which accepts a World as a separate argument.

PR #7422, opened 2020-10-18
Fix a few FluidUtil issues that were causing DispenseFluidContainer to not function properly

pupnewfster:
This PR addresses #7421 for 1.16 by ATing the two constructors that allow for nullable players in `ItemUseContext` and `BlockItemUseContext` to be public so that modders (and forge) can make use of them. Additionally this PR fixes some other issues throughout FluidUtil that caused `DispenseFluidContainer` to not properly work.

- Fixed FluidUtil#getFluidContained so that if the contained fluid is empty it returns an empty optional instead of an optional with an empty fluid stack, as all places that call this method assume this implementation.
- Made FluidUtil#tryPickUpFluid properly handle calling the methods of `IFluidBlock` instead of just looking at the tile in that position, and also make it able to handle `IBucketPickupHandler` so that it can properly de-fluidlog blocks
- Made FluidUtil#tryPlaceFluid properly be able to handle `ILiquidContainer`s so that it can properly place fluids into fluid loggable blocks.

### DIFF
--- a/src/main/java/net/minecraftforge/fluids/FluidUtil.java
+++ b/src/main/java/net/minecraftforge/fluids/FluidUtil.java
@@ -24,6 +24,8 @@ import javax.annotation.Nullable;
 
 import com.google.common.base.Preconditions;
 import net.minecraft.block.Block;
+import net.minecraft.block.IBucketPickupHandler;
+import net.minecraft.block.ILiquidContainer;
 import net.minecraft.block.material.Material;
 import net.minecraft.block.BlockState;
 import net.minecraft.entity.player.PlayerEntity;
@@ -31,7 +33,6 @@ import net.minecraft.fluid.Fluid;
 import net.minecraft.fluid.Fluids;
 import net.minecraft.item.BlockItemUseContext;
 import net.minecraft.item.ItemStack;
-import net.minecraft.item.ItemUseContext;
 import net.minecraft.item.Items;
 import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.*;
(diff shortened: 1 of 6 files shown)


src/main/java/net/minecraftforge/fluids/FluidUtil.java
```diff
@@ -481,20 +488,29 @@ public static FluidActionResult tryPickUpFluid(@Nonnull ItemStack emptyContainer
...
+        }
+        else
+        {
+            Optional<IFluidHandler> fluidHandler = getFluidHandler(worldIn, pos, side).resolve();
+            if (!fluidHandler.isPresent())
             {
-                return tryFillContainer(emptyContainer, targetFluidHandler, Integer.MAX_VALUE, playerIn, true);
+                return FluidActionResult.FAILURE;
             }
+            targetFluidHandler = fluidHandler.get();
```
pupnewfster, line 507: I moved this handling from IFluidBlock (as it is not the proper handling for an IFluidBlock) to here in an else branch, but I am unsure if this branch should not just be removed instead as querying a tile for its fluid handler doesn't seem like it has to do with getting a fluid from *in world*.

LexManos:
https://github.com/MinecraftForge/MinecraftForge/commit/d97f28141cb6bd4ffd0b232343f2975da8e1915f

BluSunrize:
I'm late to the party since this got merged already, but y'all realize that this PR full on modified the expected behavior of `FluidUtil.getFluidContained`, right?

Previously, it would only return `Optional.empty()` when it was given an empty ItemStack. And now it does it for the contained FluidStack being empty to. That's a *massive* change in its behavior, and results in a plethora of issues for any mod that opted to call `orElseThrow` on that optional.

Developers wrote their code expecting to be allowed to throw an exception if an empty Optional was returned here, on account of that empty Optional indicating a faulty state - an empty ItemStack. But now, an empty FluidStack (which is far more common!) outputs the same result and causes devs to falsely assume a faulty state, resulting in an exception thrown.

That's not good. That's a pretty large modification to expected method behavior, and it was introduced on a minor version update. I realize it's likely to late to fix this, nor would anyone want to because "The whole Fluid system needs a refactor anyway". But even so, this definitely deserves a highlight, as well as a consideration for future Pull Requests, because methods should not change their behavior like that on a running version.

pupnewfster:
Ya I agree that part wasn't ideal, truth be told I forgot that that part technically is a breaking change as the other places in forge that call it assumed that it had the behavior it has now. It also as the name is getFluidContained but returning an optional never crossed my mind when I made this PR like four months before it got merged that modders may be using that method as a utility method to check if an itemstack is empty (instead of based on the name doing what forge did in various places of assuming the implementation and it being an optional that is returned that it actually only has an optional if it well has a fluid contained)... So hadn't considered it as a breaking change and by the time lex was asking me if the PR had any breaking changes I did a brief glance of things and didn't think so as I had remembered there not being any functionality changes beyond the one I commented on earlier which after discussing we agreed that it functionality and case it adjusted should never have really been possible beforehand.

PR #7426, opened 2020-10-21
Fix regression in RightClickBlock

Shadows-of-Fire:
Somewhere between 1.12 and 1.16 (this may have happened in any version between 1.13-1.16), the patches that enabled the ALLOW results to have any purpose were lost.  This PR restores them, and additionally updates the javadocs on RightClickBlock to be a bit more descriptive on what the consequences of each result is.


Shadows-of-Fire:
Update as of the third commit, there was another regression with this event regarding the lack of information available.  This was lost somewhere between 1.12-1.15 (was attempted to be readded in #6615).  In 1.12 the hitvec was passed as Vec3d, but it is now available as a BlockRayTraceResult which is more contextual (and requires no additional work to access, the hitvec had to be manually recomputed in 1.12).

Shadows-of-Fire:
These issues still persist, so this isn't stale.

PR #7427, opened 2020-10-22
Fixed Configs not correcting when hot-loaded from changes on disk

### LINKED ISSUE #7367: [1.16.3] Config values range bug
**Minecraft Version:** 1.16.3

**Forge Version:** 34.1.0

**Logs:** https://gist.github.com/Extegral/49da83149b9dc246f227dbb93a4b127e

**Steps to Reproduce:**
 1. Make some config via `ForgeConfigSpec`;
 2. Add the value with defined range;
 3. Make this value to be used somewhere in the game;
 4. Lauch the game and enter into any world;
 5. Change the value in the config to be beyond defined limit and save it;
 6. Wait until it's called;
 7. Behold!

**Description of issue:**
As I have noticed, Forge automatically updates config values after they are changed in config file, without requiring to reload the world or whole client or server. I've used what I've found out as default config implementation, it basically looks as following:

```
public class ConfigHandler {
    public static ForgeConfigSpec COMMON;
    public static final ForgeConfigSpec.IntValue IMPORTANT_VALUE;
    
     ...
    
static {
        final ForgeConfigSpec.Builder common = new ForgeConfigSpec.Builder();

        common.comment("Very important stuff down there").push("Important Options");
        
        IMPORTANT_VALUE = common
                .comment("Important value. VERY.")
                .translation("configGui.somecoolmod.important_value")
                .define("importantValue", 4, 1, 100);

       common.pop();

       COMMON = common.build();

 }

}
```

```
public SomeCoolMod() {

       ...

       ModLoadingContext.get().registerConfig(ModConfig.Type.COMMON, ConfigHandler.COMMON, "somecoolmod-common.toml");

}
```

But when the values are updated after file changing, they are not checked to be within the range that they should. So, when you define a value like shown above, and simply call it somewhere like so:

`ConfigHandler.IMPORTANT_VALUE.get()`

The integer it returns may end up being beyond expected bounds, like zero, negative or bigger than 99. Values that are out of bounds are corrected upon [re]launching the client/server, but other than that, they effectively persist. Needless to say, it may break some things badly if one fully relies on the values bounds they have defined.

Even more than that, values could be defined as object they should not represent at all. You can easily enter boolean value in place of integer and end up with `ClassCastException` as soon as it get's called, instead of showing up the parser error in log file. Invalid keys like this are also corrected on startup, but for whatever reason this is not done when config changes after, naively trusting user's competence. The log refenced above shows an example of crash caused by this.

This issue is the exact duplicate of #6181, which I reported two major Minecraft versions ago, and which have never received any attention, later being marked as stale and closed. It seems to have never been addressed since.

DaemonUmbra:
Closes #7367

### DIFF
--- a/src/main/java/net/minecraftforge/fml/config/ConfigFileTypeHandler.java
+++ b/src/main/java/net/minecraftforge/fml/config/ConfigFileTypeHandler.java
@@ -30,7 +30,6 @@ import org.apache.commons.io.FilenameUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
-import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -100,32 +99,28 @@ public class ConfigFileTypeHandler {
 
     public static void backUpConfig(final CommentedFileConfig commentedFileConfig, final int maxBackups)
     {
-        File bakFileLocation = commentedFileConfig.getFile().getParentFile();
+        Path bakFileLocation = commentedFileConfig.getNioPath().getParent();
         String bakFileName = FilenameUtils.removeExtension(commentedFileConfig.getFile().getName());
         String bakFileExtension = FilenameUtils.getExtension(commentedFileConfig.getFile().getName()) + ".bak";
-        File bakFile = new File(bakFileLocation, bakFileName + "-1" + "." + bakFileExtension);
+        Path bakFile = bakFileLocation.resolve(bakFileName + "-1" + "." + bakFileExtension);
         try
         {
             for(int i = maxBackups; i > 0; i--)
             {
-                File oldBak = new File(bakFileLocation, bakFileName + "-" + i + "." + bakFileExtension);
-                if(oldBak.exists())
+                Path oldBak = bakFileLocation.resolve(bakFileName + "-" + i + "." + bakFileExtension);
+                if(Files.exists(oldBak))
                 {
-                    if(i == maxBackups)
-                    {
-                        if (oldBak.delete()) continue;
-                        LOGGER.warn(CONFIG, "Failed to back up config file {} as the oldest backup, {}, could not be deleted", commentedFileConfig.getFile().getName(), oldBak.getAbsolutePath());
-                        return;
-                    }
-                    if (!oldBak.renameTo(new File(bakFileLocation, bakFileName + "-" + (i + 1) + "." + bakFileExtension)))
-                        LOGGER.warn(CONFIG, "Failed to back up config file {} as an old backup, {}, could not be renamed", commentedFileConfig.getFile().getName(), oldBak.getAbsolutePath());
+                    if(i >= maxBackups)
+                        Files.delete(oldBak);
+                    else
+                        Files.move(oldBak, bakFileLocation.resolve(bakFileName + "-" + (i + 1) + "." + bakFileExtension));
                 }
             }
-            Files.copy(commentedFileConfig.getNioPath(),bakFile.toPath());
+            Files.copy(commentedFileConfig.getNioPath(), bakFile);
         }
         catch (IOException exception)
         {
-            LOGGER.warn(CONFIG, "Failed to back up config file {}", commentedFileConfig.getFile().getName());
+            LOGGER.warn(CONFIG, "Failed to back up config file " + commentedFileConfig.getNioPath(), exception);
         }
     }
 
--- a/src/main/java/net/minecraftforge/fml/config/ConfigFileTypeHandler.java
+++ b/src/main/java/net/minecraftforge/fml/config/ConfigFileTypeHandler.java
@@ -120,7 +120,7 @@ public class ConfigFileTypeHandler {
         }
         catch (IOException exception)
         {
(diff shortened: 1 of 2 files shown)


cpw:
This looks good...

src/main/java/net/minecraftforge/fml/config/ConfigFileTypeHandler.java
```diff
@@ -91,6 +93,35 @@ private boolean setupConfigFile(final ModConfig modConfig, final Path file, fina
...
 
+    public static void backUpConfig(final CommentedFileConfig commentedFileConfig) {
+        int maxBackups = 5;
+        File bakFileLocation = commentedFileConfig.getFile().getParentFile();
+        String bakFileName = FilenameUtils.removeExtension(commentedFileConfig.getFile().getName());
+        String bakFileExtension = FilenameUtils.getExtension(commentedFileConfig.getFile().getName()) + ".bak";
+        File bakFile = new File(bakFileLocation, bakFileName + "-1" + "." + bakFileExtension);
+        try
+        {
+            for(int i = maxBackups; i > 0; i--){
```
sciwhiz12: Braces on newlines, please.

src/main/java/net/minecraftforge/fml/config/ConfigFileTypeHandler.java
```diff
@@ -91,6 +93,35 @@ private boolean setupConfigFile(final ModConfig modConfig, final Path file, fina
...
+                File oldBak = new File(bakFileLocation, bakFileName + "-" + i + "." + bakFileExtension);
+                if(oldBak.exists()){
```
sciwhiz12: Braces on newlines, please.

src/main/java/net/minecraftforge/fml/config/ConfigFileTypeHandler.java
```diff
@@ -91,6 +93,35 @@ private boolean setupConfigFile(final ModConfig modConfig, final Path file, fina
         return true;
     }
 
+    public static void backUpConfig(final CommentedFileConfig commentedFileConfig) {
+        int maxBackups = 5;
```
sciwhiz12: Just asking, wouldn't this be better as a static constant, instead of a local constant?

sciwhiz12 (changes requested):
A bit of formatting.
<sup><sup>~pls no hurt me~</sup></sup>

src/main/java/net/minecraftforge/fml/config/ConfigFileTypeHandler.java
DaemonUmbra: Added an overload that allows the caller to choose maxBackups and a TODO requesting ideas on how to allow mods to set maxBackups for their own config

cpw:
How is this going to interact with automatically saving file editors? Are we going to spam a bunch of broken backups whenever it breaks?

Also, we use NIO, not java.io.File in general. It should make the code a lot cleaner...

src/main/java/net/minecraftforge/fml/config/ConfigFileTypeHandler.java
```diff
@@ -91,6 +93,42 @@ private boolean setupConfigFile(final ModConfig modConfig, final Path file, fina
...
+                {
+                    if(i == maxBackups)
+                    {
+                        if (oldBak.delete()) continue;
+                        LOGGER.warn(CONFIG, "Failed to back up config file {} as the oldest backup, {}, could not be deleted", commentedFileConfig.getFile().getName(), oldBak.getAbsolutePath());
+                        return;
+                    }
+                    if (!oldBak.renameTo(new File(bakFileLocation, bakFileName + "-" + (i + 1) + "." + bakFileExtension)))
+                        LOGGER.warn(CONFIG, "Failed to back up config file {} as an old backup, {}, could not be renamed", commentedFileConfig.getFile().getName(), oldBak.getAbsolutePath());
+                }
```
cpw, line 117: Use NIO, not java.io.File. This will be a lot cleaner..

src/main/java/net/minecraftforge/fml/config/ConfigFileTypeHandler.java
```diff
@@ -91,6 +92,38 @@ private boolean setupConfigFile(final ModConfig modConfig, final Path file, fina
...
+                        Files.delete(oldBak);
+                    else
+                        Files.move(oldBak, bakFileLocation.resolve(bakFileName + "-" + (i + 1) + "." + bakFileExtension));
+                }
+            }
+            Files.copy(commentedFileConfig.getNioPath(), bakFile);
+        }
+        catch (IOException exception)
+        {
+            LOGGER.warn(CONFIG, "Failed to back up config file " + commentedFileConfig.getNioPath(), exception);
```
DaemonUmbra: Before anyone comments on this part, Log4J doesn't seem to have a method for both {} params and throwable printing at the same time.

pupnewfster: Actually it does, I looked it up at one point, and it turns out log4j has "hidden" functionality that makes it treat it as such if it has one more passed to it than the parameters and is a throwable.

DaemonUmbra: TIL

PR #7428, opened 2020-10-22
[1.16.x] Add loot table ID in Global Loot Modifiers

TheSilkMiner:
## Abstract
Global Loot Modifiers provide an efficient way to react to loot drops and modify the generated loot without altering loot tables, which proves to be incompatible with other mods as only the "top-most" one wins. Some loot tables, though, cannot be reliably identified via GLMs due to context not being available. One such example is dungeon loot tables, as in there is no way to reliably recognize whether a given loot table represents a dungeon OR which type of dungeon it is. This PR aims to make this possible.

## The Solution
**NOTE:** After the feedback, the "Alternative" solution has been adopted. The old solution description has been kept here for history.
The approached solution consists in assigning an identifier to every loot table, namely the RL that identifies its location inside a datapack. This is a good choice because two loot tables cannot coexist within the same location, otherwise the second would override the first. The given identifier is then passed over to `ForgeHooks`, which injects it into the loot context prior to passing the loot to the Global Loot Modifier system.
The loot table ID can be queried directly through the context via the getter or via the new `forge:loot_table_id` `LootCondition`. 
Backwards compatibility and null-safety has been kept by adding a "bouncing" method in `ForgeHooks` that uses a custom ResourceLocation (`forge:unknown`) to identify loot tables without an ID. Moreover, the same identifier is returned by `LootContext` instead of a null value, allowing mods not to null-check for the extremely rare case of a loot table not having an associated ID.

<details><summary> Old solution</summary>

The approached solution consists in assigning an identifier to every loot table, namely the RL that identifies its location inside a datapack. This is a good choice because two loot tables cannot coexist within the same location, otherwise the second would override the first. The given identifier is then passed on to the hook that generates the loot and from there as a parameter to the Global Loot Modifier.
Backwards-compatibility is preserved on the side of Global Loot Modifiers by having the method with the added resource location redirect by default to the original method. The same applies to the default implementation (`LootModifier`). The old method has also been scheduled for removal in 1.17.
As an example of backwards-compatibility being preserved, one of the examples GLMs has not been updated to use the new method. A new GLM is also provided that illustrates the usage of the new parameter.

</details>


Tslat:
This basically completes the GLM system in my eyes.
Without being able to identify which table you're modifying for, the GLM system is missing a _huge_ portion of its functionality, effectively forcing modders to go back to ugly static modifications of loot tables

Cyborgmas:
Well GLM being loot table agnostic is not bad, that's what conditions are for. The alternative proposed, of putting the information in the loot context would be more proper imo and then making a condition which checks that would be cleaner. But it's probably a mess patch wise (if memory serves) so this is a good alternative.

TeamSpen210:
It could be done just by copying the `lootTableId` patch to the context, and then setting it in `modifyLoot`. Doing it via conditions would be far more desirable, since then users would be able to for example expand a condition to apply to a modded dungeon, or disallow it from applying to that.

TheSilkMiner:
According to given feedback, the PR has been reworked to use the combination of `LootContext` + custom `ILootCondition`

src/main/java/net/minecraftforge/common/loot/LootTableIdCondition.java
```diff
@@ -0,0 +1,81 @@
...
+import net.minecraft.loot.ILootSerializer;
+import net.minecraft.loot.LootConditionType;
+import net.minecraft.loot.LootContext;
+import net.minecraft.loot.conditions.ILootCondition;
+import net.minecraft.loot.conditions.LootConditionManager;
+import net.minecraft.util.JSONUtils;
+import net.minecraft.util.ResourceLocation;
+
+public class LootTableIdCondition implements ILootCondition {
+    public static final ResourceLocation UNKNOWN_LOOT_TABLE = new ResourceLocation("forge", "unknown");
```
Cyborgmas: I would make this name a bit more unique perhaps, like forge:__unknown__ or something

src/main/java/net/minecraftforge/common/loot/LootTableIdCondition.java
```diff
@@ -0,0 +1,81 @@
...
+import com.google.gson.JsonSerializationContext;
+import net.minecraft.loot.ILootSerializer;
+import net.minecraft.loot.LootConditionType;
+import net.minecraft.loot.LootContext;
+import net.minecraft.loot.conditions.ILootCondition;
+import net.minecraft.loot.conditions.LootConditionManager;
+import net.minecraft.util.JSONUtils;
+import net.minecraft.util.ResourceLocation;
+
+public class LootTableIdCondition implements ILootCondition {
```
Cyborgmas: braces { should be on new line for a forge class

patches/minecraft/net/minecraft/loot/conditions/LootConditionManager.java.patch
```diff
@@ -0,0 +1,10 @@
+--- a/net/minecraft/loot/conditions/LootConditionManager.java
++++ b/net/minecraft/loot/conditions/LootConditionManager.java
+@@ -24,6 +24,7 @@
+    public static final LootConditionType field_237471_n_ = func_237475_a_("weather_check", new WeatherCheck.Serializer());
+    public static final LootConditionType field_237472_o_ = func_237475_a_("reference", new Reference.Serializer());
+    public static final LootConditionType field_237473_p_ = func_237475_a_("time_check", new TimeCheck.Serializer());
++   public static final LootConditionType LOOT_TABLE_ID = func_237475_a_("forge:loot_table_id", new net.minecraftforge.common.loot.LootTableIdCondition.Serializer()); // FORGE
```
Cyborgmas: Not sure if this should be patched in? Probably better in ForgeMod or something no?

Cyborgmas (commented):
Seems pretty good, definitely not that big a mess that i thought. It is more patches but I still thinks its a better solution. Left some minor comments

src/main/java/net/minecraftforge/common/loot/LootTableIdCondition.java
TheSilkMiner: `forge:unknown_loot_table`?

patches/minecraft/net/minecraft/loot/conditions/LootConditionManager.java.patch
TheSilkMiner: That could be an alternative solution, but since we don't have a registry for `ILootCondition`s, when would be the best way to initialize it?
Mod constructor? In a new event for loot tables specifically?

Cyborgmas: Oh right... well nvm then.

TheSilkMiner: I figured that vanilla registers them statically on bootup (-.-).
I reworked the addition into an AT and I register the class statically when Forge is initialized. A nice side effect of this is that other mods can now register their own conditions without having to AT or reflect themselves.
Of course, if an alternative ends up being available, like an event, the static init can be dropped in favor of the event.

Cyborgmas:
ATing the register function might be a nono, would have to verify with other people.

TheCurle:
Code looks good. I'm not familiar with the GLM system in depth, so i'll call upon Cyborgmas one more time to check this over.

Cyborgmas:
That part seems good, I am just a tad concerned about https://github.com/MinecraftForge/MinecraftForge/pull/7428/files#diff-9b2699706a1e994f28ae7ddc53ee517d6f57457a4ca51a7df9e3ba897c82d689R36 . 
Afaik, it's best to register it inside still a RegistryEvent, like set it to null and non final then do it? (not its own registryevent ofc, as it doesnt have one. I think giga used IRecipeSerializer?) Also the `LootConditionManager.register` AT is uneeded, as it is just a helper function for `Registry.register(Registry.LOOT_CONDITION_TYPE, new ResourceLocation(registryName), new LootConditionType(serializer))`

50ap5ud5:
Thanks for making this! I encountered similar limitations with GLM, glad someone made a fix for it!

Cyborgmas (approved):

TelepathicGrunt (approved):

Tslat (approved):

NicosaurusRex99 (approved):

patches/minecraft/net/minecraft/loot/LootTable.java.patch
```diff
@@ -59,6 +59,12 @@
 +         throw new RuntimeException("Attempted to modify LootTable after being finalized!");
 +   }
 +
++   private ResourceLocation lootTableId;
++   public void setLootTableId(final ResourceLocation id) {
++      if (this.lootTableId == null && id != null) this.lootTableId = id;
```
gigaherz: Is this condition necessary?

src/main/java/net/minecraftforge/common/loot/LootTableIdCondition.java
```diff
@@ -0,0 +1,93 @@
...
+    public static final ResourceLocation UNKNOWN_LOOT_TABLE = new ResourceLocation("forge", "unknown_loot_table");
+
+    private final ResourceLocation targetLootTableId;
+
+    private LootTableIdCondition(final ResourceLocation targetLootTableId)
+    {
+        this.targetLootTableId = targetLootTableId;
+    }
+
+    public static Builder builder(final ResourceLocation targetLootTableId)
```
gigaherz: This method feels out of place here. It should either be at the top after the constants, or below right before the Builder.

gigaherz (commented):
Just a couple nitpicky comments from a quick look. 
Otherwise LGTM. I can't think of a better way to achieve this.

patches/minecraft/net/minecraft/loot/LootTable.java.patch
TheSilkMiner: Not necessary, but more so as a failsafe. This way a malicious mod cannot alter the name of the loot table after it has been loaded, preventing weird and hard to track down issues with desyncs between what vanilla thinks and what loot modifiers perceive.

gigaherz: I'm wondering then, if we should instead have a proper check.
```java
    if (this.lootTableId != null) throw new IllegalStateException("Loot table ID can not be set more than once.");
    this.lootTableId = Objects.requireNonNull(id);
```

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -327,4 +327,4 @@ public net.minecraft.world.gen.treedecorator.TreeDecoratorType <init>(Lcom/mojan
 private-f net.minecraft.world.raid.Raid$WaveMember field_221284_f # VALUES
 public net.minecraft.world.server.ServerChunkProvider field_186029_c # chunkGenerator
 public net.minecraft.world.server.ServerChunkProvider field_73251_h # worldObj
-public net.minecraft.world.storage.FolderName <init>(Ljava/lang/String;)V # constructor
\ No newline at end of file
+public net.minecraft.world.storage.FolderName <init>(Ljava/lang/String;)V # constructor
```
gigaherz: I guess after rebasing, it left an unintended change in this file?

gigaherz (commented):
If you undo this it will LGTM.

PR #7432, opened 2020-10-24
Reduce log's level and add comments.

FledgeXu:
The MDK's default log level is `debug` that seems too much for modding. 
In the `debug` level, It will generate about **2.1m size** and **12480 lines** in console (Example Mod and Just booting game):
image
image
In this default log level, It's hard to find useful logs like json not found. Because they are submerged in spam logs.

This PR includes:
- Reduce log's level from `debug` to `info`. That will cut off a lot of spam logs.
  image
  image
- Add explains about log setting in `build.gradle`

This PR is not a huge improvement, but nice to have.

### DIFF
--- a/mdk/build.gradle
+++ b/mdk/build.gradle
@@ -42,12 +42,12 @@ minecraft {
             // "SCAN": For mods scan.
             // "REGISTRIES": For firing of registry events.
             // "REGISTRYDUMP": For getting the contents of all registries.
-            property 'forge.logging.markers', 'SCAN,REGISTRIES,REGISTRYDUMP'
+            property 'forge.logging.markers', 'REGISTRIES'
 
             // Recommended logging level for the console
             // You can set various levels here.
             // Please read: https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels
-            property 'forge.logging.console.level', 'info'
+            property 'forge.logging.console.level', 'debug'
 
             mods {
                 examplemod {
@@ -64,12 +64,12 @@ minecraft {
             // "SCAN": For mods scan.
             // "REGISTRIES": For firing of registry events.
             // "REGISTRYDUMP": For getting the contents of all registries.
-            property 'forge.logging.markers', 'SCAN,REGISTRIES,REGISTRYDUMP'
+            property 'forge.logging.markers', 'REGISTRIES'
 
             // Recommended logging level for the console
             // You can set various levels here.
             // Please read: https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels
-            property 'forge.logging.console.level', 'info'
+            property 'forge.logging.console.level', 'debug'
 
             mods {
                 examplemod {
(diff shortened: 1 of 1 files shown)


XFactHD:
I am all for reducing log spam but I think that this is the wrong way to go about it. I did a similar test to what you did in a mod that does absolutely nothing on startup but with three test cases:
- log level set to info
- log level set to debug
- log level set to debug with the three log markers enabled

My results are as follows:
- Log level info: 27 lines (17 KB)
- Log level debug: 549 lines (94 KB)
- Log level debug including markers: 5835 lines (1030 KB)

Therefore I would recommend to disable the log markers instead of disabling the debug logging completely.

FledgeXu:
I have considered this. However, I don’t which one is better. Is 549 lines still too long?

sciwhiz12:
Instead of disabling all the logging markers, I would suggest only removing the `SCAN` and `REGISTRYDUMP` markers, as they make up the bulk of the extraneous data (esp. the `REGISTRYDUMP`, which dumps the contents of all registries after their freezing). `REGISTRY` is related to the registry events, and is still useful when trying to debug registry-related behaviors.
_(I personally disable both logging markers, both to reduce log/console spam and to speedup the loading process.)_

FledgeXu:
You are right. And If the log level is `info` the markers are all disable.
I will fix it.

XFactHD:
The firing of the registry events is still logged when the marker is disabled though there might be more data on those when the marker is enabled.

XFactHD:
I just checked what the `REGISTRIES` log marker actually does: it only logs `ObjectHolder` lookups and registry freezing, therefore I don't see a real reason to keep it enabled.

FledgeXu:
We can keep it as an example.

FledgeXu:
Actually, There are a branch of markers, my friend tells me.
https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/src/fmllauncher/resources/log4j2.xml

DaemonUmbra:
Personally I feel like things are fine as they are, remember the default is not enforced and you can change it for your own projects.

Darkere:
I can imagine rare cases where it is useful but I've only used the log for debugging it once. Just throwing a breakpoint into the registration code is much more flexible and also tells me where the problem is exactly. 

I imagine that in 90% of cases it's a "oh did I forget to register this?" in which case checking your own code is much faster than going digging in the log file. 

So I'm all for removing the registry from the log. 

Debug in comparison can be useful. Seeing which events fired when error happened can be crucial information.

raoulvdberge:
Agreed. The default is too noisy, and it's not THAT useful.

TheCurle:
This is a good idea. 

As it stands, i'm willing to pass this on, but after some discussion the idea was floated of adding two extra tasks that optionally enable this extended logging for client and server, so modders do not have to touch the markers to debug simple issues.

Thoughts?

FledgeXu:
It’s a good idea.
However, I don’t know where we should add these tasks.
In `build.gradle` or in ForegGradle?

TheCurle:
Some quick experimenting should provide the answer

FledgeXu:
After my experiments, I find that it seems not an elegant way to achieve this goal.
I have to write, a least, 3 tasks in `build.gradle` which is not necessary and redundancy in my opinion.
However, those tasks also should not be provided by ForgeGradle.
Anyway, I’m not professor of groovy, There are may better ways to write tasks.

raoulvdberge:
I think it's easier to just modify the logging markers than to introduce new tasks...
If you add a comment that contains what markers are possible (what this PR does..), I don't see the value of a task.

TheCurle:
The idea is that:
If someone is having trouble registering items, we (Forge Support) can ask them to
A) modify the build.gradle file to add back these markers and look at the output OR
B) Run the hypothetical new debug task
With A there is a considerable risk of the user forgetting to change the gradle file back, ergo completely making this change moot.

It's an ease of use thing. But having a list of markers would be good in its own right, maybe as part of the Documentation.

PR #7434, opened 2020-10-25
Fix data pack biomes failing to load in SingleBiomeProvider

### LINKED ISSUE #7433: SingleBiomeProvider fails to generate world with data pack biomes
**Minecraft Version:** 1.16.3

**Forge Version:** 1.16.3 - 34.1.25

**Logs:** https://hatebin.com/rlrsvbygzg

**Steps to Reproduce:**

1. In the create new world screen, add the following data pack:  [test2pack.zip](https://github.com/MinecraftForge/MinecraftForge/files/5436126/test2pack.zip)
2. In "more world options", choose  "Single biome", then scroll down and select "test2:plains"
3. Hit "create world"
4. Click accept to the experimental settings warning screen
5. In vanilla: this works fine. In forge, data packs fail to load with this error: https://hatebin.com/rlrsvbygzg

**Description of issue:**

Forge patches a `ResourceLocation` onto the `Biome` codec. This is all well and good, except it's inadequate because nowhere does it *set* the registry name that it expects to be there. Thus round trip serialization (`biome == encode(decode(biome))`) is now broken. This matters because apparently vanilla does this while loading data packs. Who knew.

- This is related to #7329 although it is a separate issue.
- It is also closely related to https://bugs.mojang.com/browse/MC-197616 - as it stands, this creates massive log spam on the client because vanilla does not properly serialize these biomes over the network. However, with forge, it doesn't even load the world. This should really be fixed in forge if not in vanilla.

alcatrazEscapee:
The registry name on biomes is now *actually stored on the biome*. Since we can't just call `setRegistryName`, as it spams the log about "Dangerous override blah blah", this is called anywhere and everywhere so it can't just be assumed to be mod code, we add an additional field to `Biome` storing the registry name.

This patches the two important call sites for new biome creation (the codec, and `FlatGenerationSettings`.) In all other cases (for instance using `Biome.Builder`, it's either a vanilla biome (fine), or a mod biome which `setRegistryName` is expected to be called.

This has the added benefit, that `Biome#getDynamicRegistryName` *actually* will hold the registry name, even when it is a dynamic instance! Horray! No need to delve through the `MinecraftServer` -> `DynamicRegistries` -> `Registry<Biome>` every time!

This fixes #7433


Cyborgmas (commented):
Seems well done, just got a few leftover import changes in ForgeHooks

alcatrazEscapee:
Much obliged, fixed.

TelepathicGrunt (approved):

PR #7439, opened 2020-10-27
A possible modification to #7434 to make it a non-breaking change.

alcatrazEscapee:
The gist, from discord:

> So TIL that https://github.com/MinecraftForge/MinecraftForge/pull/7434 may have included a binary compat break by changing Biome extends ForgeRegistryEntry<Biome> to Biome implements IForgeRegistryEntry<Biome>
> 
> A friend mentioned to me that code that worked pre 34.1.32 now crashes with
> 
> > java.lang.NoSuchMethodError: net.minecraft.world.Biome.setRegistryName(Lnet/minecraft/util/ResourceLocation;)Lnet/minecraftforge/registries/IForgeRegistryEntry;
> 
> So there's two things here
> - Removing ForgeRegistryEntry accidentally involved removing two overloads of setRegistryName which take a String name or String modid, String name, respectively.
> - And (this I did not know), apparently narrowing the type of a return value actually constitutes a breaking change. (Although the only "fix" that is required would be to recompile against the new version, but that's still a binary break)

So, this is one possibility to re-implement `ForgeRegistryEntry`, and avoid the original idea using `get/set/DynamicRegistryName`, and avoid the same pitfalls that were fallen into in the original idea.

### DIFF
--- a/patches/minecraft/net/minecraft/world/biome/Biome.java.patch
+++ b/patches/minecraft/net/minecraft/world/biome/Biome.java.patch
@@ -5,7 +5,7 @@
  import org.apache.logging.log4j.Logger;
  
 -public final class Biome {
-+public final class Biome implements net.minecraftforge.registries.IForgeRegistryEntry<Biome> {
++public final class Biome extends net.minecraftforge.registries.ForgeRegistryEntry.DynamicRegistryEntry<Biome> {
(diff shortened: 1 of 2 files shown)


LexManos:
This should be fine, Sorry guys, the previous acceptance was a derp on my end.
This should be binary compatible with anyone who compiled against the previous versions.

PR #7441, opened 2020-10-27
Fix translucent blocks not rendering properly when moved by pistons on fabulous

### LINKED ISSUE #6915: [1.16] Translucent blocks do not render when being moved by a piston
**Minecraft Version:** 1.16.1

**Forge Version:** 32.0.47

**Logs:** N/A

**Steps to Reproduce:**
 1. Put a sticky piston on a clock with stained glass
 2. Turn on clock
 3. Glass disappears when in motion
 4. Repeat in vanilla and note it stays rendered

**Description of issue:**
https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/patches/minecraft/net/minecraft/client/renderer/tileentity/PistonTileEntityRenderer.java.patch#L7

The patched-out call to `func_239221_b_` is important as it checks if the block is for the TRANSLUCENT layer and returns in its stead TRANSLUCENT_MOVING_BLOCK (yarn name). Not using that layer means it doesn't render (for some reason)

Cyborgmas:
Fixes #6915.

This is only on fabulous settings and the vanilla blocks affected are any stained glass, honey, slime and ice.
The fix is simply checking for translucent and passing in translucent moving block instead


patches/minecraft/net/minecraft/client/renderer/tileentity/PistonTileEntityRenderer.java.patch
```diff
@@ -1,11 +1,12 @@
...
 +++ b/net/minecraft/client/renderer/tileentity/PistonTileEntityRenderer.java
-@@ -61,8 +61,12 @@
+@@ -61,8 +61,13 @@
     }
  
     private void func_228876_a_(BlockPos p_228876_1_, BlockState p_228876_2_, MatrixStack p_228876_3_, IRenderTypeBuffer p_228876_4_, World p_228876_5_, boolean p_228876_6_, int p_228876_7_) {
 -      RenderType rendertype = RenderTypeLookup.func_239221_b_(p_228876_2_);
-+      net.minecraft.client.renderer.RenderType.func_228661_n_().stream().filter(t -> RenderTypeLookup.canRenderInLayer(p_228876_2_, t)).forEach(rendertype -> {
++      RenderType.func_228661_n_().stream().filter(t -> RenderTypeLookup.canRenderInLayer(p_228876_2_, t)).forEach(rendertype -> {
++      rendertype = RenderTypeLookup.canRenderInLayer(p_228876_2_, RenderType.func_228645_f_()) ? RenderType.func_239269_g_() : rendertype;
```
ichttt: This does not seem correct. This replaces every render type with `TranslucentMovingBlock` if the block can render in the Translucent render type, but I think it should only replace the `Translucent` render layer with `TranslucentMovingBlock`, or am I wrong?

Cyborgmas: Indeed, I'll add a check that this is already in translucent.

LexManos:
1) This is rendering so should be assigned to tterrage/giga
2) This is rendering so it should at the very least have a screenshot showing the before and after
3) The change is.. odd to say the least. Shouldn't it just be a map on the stream in the original patch? I'm not seeing how the function that was pointed out is getting integrated into this new system any better then the old.

Cyborgmas:
[Bug Video][bvid]

[Fixed Video][fvid]

As for the change, it was pretty subjective, the loop inside the function felt weird and odd to read, just refactoring the behaviour in another function with proper formatting is more legible, I can revert to just changing the patch.

[bvid]:https://streamable.com/klkn8p
[fvid]:https://streamable.com/k0of0f

Cyborgmas:
In the fixed video, the side of the ice and stained glass disappears, I thought this was a vanilla issue but tested it again and it isn't will debug and update accordingly

Cyborgmas:
Okay, I am quite dumb I re checked vanilla and it is indeed the same behaviour.
This could probably be fixed though.
[vanilla][v]

[v]:https://streamable.com/fl96a4

PR #7442, opened 2020-10-27
Add Player context to AnvilUpdateEvent

Shadows-of-Fire:
This PR adds the player using the Anvil to the context of the AnvilUpdateEvent.

It also updates the documentation of the event significantly, as it was pretty bare before.  Most of the changes involve the comments (the changes to add the player to the event were rather minor).

The old constructor and old ForgeHooks method have been deprecated as they cannot be removed without causing a breaking change.

The reason for having the player available is as follows:
I was attempting to recreate the vanilla behavior of using the anvil, but with slightly lower numbers as my mod allows for higher enchantments levels to be obtained through normal gameplay.  These higher enchantment levels make it very difficult to use the anvil, but currently, it is impossible to retain vanilla behavior for creative players and still make this modification.  The player using the anvil is required to still allow creative players to do whatever they want (such as putting enchantments on items that normally may not receive them), while still being able to change the vanilla calculations.

Another possible solution to this problem would be to add another event that fires when vanilla's implementation has completed, to allow for that result to be changed or edited.  However, I believe this solution is superior, as there are likely more uses for the player context than my scenario.

### DIFF
--- a/src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
+++ b/src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
@@ -36,7 +36,8 @@ import javax.annotation.Nullable;
  * if the output is empty, and the event is not canceled, vanilla behavior will execute. <br>
  */
 @Cancelable
-public class AnvilUpdateEvent extends Event {
+public class AnvilUpdateEvent extends Event
+{
 
     private final ItemStack left;
     private final ItemStack right;
@@ -65,7 +66,7 @@ public class AnvilUpdateEvent extends Event {
     }
 
     /**
-     * @return ihe Item in the left input (leftmost) anvil slot.
+     * @return The item in the left input (leftmost) anvil slot.
      */
     public ItemStack getLeft()
     {
@@ -92,10 +93,11 @@ public class AnvilUpdateEvent extends Event {
     }
 
     /**
-     * This is the output as a result of this event, not of the two items. <br>
+     * This is the output as determined by the event, not by the vanilla behavior between these two items. <br>
      * If you are the first receiver of this event, it is guaranteed to be empty. <br>
-     * It will only be non-empty if changed by an event handler.
-     * @return The item in the output (rightmost) anvil slot.
+     * It will only be non-empty if changed by an event handler. <br>
+     * If this event is cancelled, this output stack is discarded.
+     * @return The item to set in the output (rightmost) anvil slot.
      */
     public ItemStack getOutput() 
     {
--- a/src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
+++ b/src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
@@ -131,7 +131,7 @@ public class AnvilUpdateEvent extends Event {
     }
 
     /**
-     * The material cost is how many right inputs are consumed.
+     * The material cost is how many units of the right input stack are consumed.
      * @return The material cost of this anvil operation.
      */
     public int getMaterialCost()
--- a/src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
+++ b/src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
@@ -19,52 +19,146 @@
 
 package net.minecraftforge.event;
 
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.inventory.container.RepairContainer;
+import net.minecraft.item.ItemStack;
 import net.minecraftforge.eventbus.api.Cancelable;
 import net.minecraftforge.eventbus.api.Event;
-import net.minecraft.item.ItemStack;
 
-import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 /**
  * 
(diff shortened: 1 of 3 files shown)


src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
```diff
@@ -19,52 +19,146 @@
...
-    @Nonnull
-    private final ItemStack right; // The right side of the input
-    private final String name;     // The name to set the item, if the user specified one.
-    @Nonnull
-    private ItemStack output;      // Set this to set the output stack
-    private int cost;              // The base cost, set this to change it if output != null
-    private int materialCost; // The number of items from the right slot to be consumed during the repair. Leave as 0 to consume the entire stack.
-
-    public AnvilUpdateEvent(@Nonnull ItemStack left, @Nonnull ItemStack right, String name, int cost)
+public class AnvilUpdateEvent extends Event {
```
diesieben07: Formatting (braces on new lines)

src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
```diff
@@ -19,52 +19,146 @@
...
-    public String getName() { return name; }
-    @Nonnull
-    public ItemStack getOutput() { return output; }
-    public void setOutput(@Nonnull ItemStack output) { this.output = output; }
-    public int getCost() { return cost; }
-    public void setCost(int cost) { this.cost = cost; }
-    public int getMaterialCost() { return materialCost; }
-    public void setMaterialCost(int materialCost) { this.materialCost = materialCost; }
+    /**
+     * @return ihe Item in the left input (leftmost) anvil slot.
```
diesieben07: Typo

src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
```diff
@@ -19,52 +19,146 @@
...
+     * @return The name that the output item will be set to, if applicable.
+     */
+    @Nullable
+    public String getName()
+    {
+        return name;
+    }
+
+    /**
+     * This is the output as a result of this event, not of the two items. <br>
```
diesieben07: Maybe it's because I am not a native English speaker, but to me this sentence does look correct.

src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
```diff
@@ -19,52 +19,146 @@
...
+     * If you are the first receiver of this event, it is guaranteed to be empty. <br>
+     * It will only be non-empty if changed by an event handler.
+     * @return The item in the output (rightmost) anvil slot.
```
diesieben07: Given the above description, the label for the return value is wrong (it's the item that should be put in that slot, not the one currently in it).

src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
```diff
@@ -19,52 +19,146 @@
...
+     * Changes the level cost of this operation.
+     * @param cost The new level cost.
+     */
+    public void setCost(int cost)
+    {
+        this.cost = cost;
+    }
+
+    /**
+     * The material cost is how many right inputs are consumed.
```
diesieben07:
```suggestion
     * The material cost is how many units of the right input stack are consumed.
```

src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
```diff
@@ -19,52 +19,146 @@
...
+    {
+        return materialCost;
+    }
+
+    /**
+     * Sets how many right inputs are consumed. <br>
+     * A material cost of zero consumes the entire stack. <br>
+     * A material cost higher than the count of the right stack
+     * consumes the entire stack. <br>
+     * It does not block the operation.
```
diesieben07: What is "it" here? What is not being blocked?

diesieben07 (changes requested):

src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
Shadows-of-Fire: The intentional meaning of this statement is that "this is the output as determined by the event, not by the vanilla behavior between these two items", which may be more sufficient.

src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
Shadows-of-Fire: Yeah that's vague.  The output being obtainable is not prevented by the right stack count being lower than the material cost.

Essentially the material cost has no bearing on whether or not the output is available.

src/main/java/net/minecraftforge/event/AnvilUpdateEvent.java
diesieben07: Yes, that sounds much clearer!

Shadows-of-Fire:
Comments should be a bit more concise now (and formatting / typo errors are resolved).

diesieben07 (approved):

PR #7444, opened 2020-10-28
Add an optional biome palette to chunk data.

alcatrazEscapee:
Okay, so I think the original issue here missed the point. Here is the specific impact of this issue to mods (mine, among others):

1. If a modder *changes the order of their biome registrations*, worlds created before the change may have shuffled biomes.
2. If another mod is added, or removed, worlds created before the change may have shuffled biomes.
3. If a mod adds a biome (without adding or removing mods), worlds created before the change may have shuffled biomes.

In contrast to popular opinion, you can show that this **is not** influenced by forge registry's biome ID mapping as that is done too late to matter in the dynamic registry. I have tested this several times and gotten identical results on this. If there's a better way to fix the behavior as described by the test mod and steps below, (e.g. perhaps making this mapping work?) I'm all ears.

1. Create a new world (Let's call it "First World"). Change the generation to "Single Biome", and select `biome_id_order_test:first_biome`.
2. Observe in F3, the biome is labeled as `biome_id_order_test:first_biome`
3. Quit the game and close Minecraft
4. Change the `REVERSED = false` to `REVERSED = true` in the test mod (or vice versa). All this does is change the order in which the two biomes are registered!
5. Launch the game
6. Load "First World"
7. **Without this fix**: Observe in F3, the biome is now `biome_id_order_test:second_biome`. **With this fix:** the biome is saved and restored correctly.

This addresses <https://bugs.mojang.com/browse/MC-202036>.

### DIFF
--- a/patches/minecraft/net/minecraft/world/chunk/storage/ChunkSerializer.java.patch
+++ b/patches/minecraft/net/minecraft/world/chunk/storage/ChunkSerializer.java.patch
@@ -1,5 +1,14 @@
 --- a/net/minecraft/world/chunk/storage/ChunkSerializer.java
 +++ b/net/minecraft/world/chunk/storage/ChunkSerializer.java
+@@ -67,7 +67,7 @@
+          field_222658_a.error("Chunk file at {} is in the wrong location; relocating. (Expected {}, got {})", p_222656_3_, p_222656_3_, chunkpos);
+       }
+ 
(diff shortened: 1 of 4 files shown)


alcatrazEscapee:
Looking into this some more, this should probably be rewritten in such a way that the forge Biome registry saving + sync actually works as intended (as currently it doesn't really do squat as the save + ID match happens too late and doesn't have any effect on the dynamic registry). As per #7451. I will need to think about this some more.

PR #7446, opened 2020-10-28
Fix food bar not rendering when non-living entities are mounted

diesieben07:
Reported [on the forums](https://forums.minecraftforge.net/topic/92896-hunger-bar-disappears-when-in-boat/).


TheCurle (approved):
What an oversight.
This is a compatible fix, i'm all for it

SmellyModder (approved):

src/main/java/net/minecraftforge/client/gui/ForgeIngameGui.java
```diff
@@ -112,7 +112,7 @@ public void func_238445_a_(MatrixStack mStack, float partialTicks)
         this.scaledHeight = this.mc.getMainWindow().getScaledHeight();
         eventParent = new RenderGameOverlayEvent(mStack, partialTicks, this.mc.getMainWindow());
         renderHealthMount = mc.player.getRidingEntity() instanceof LivingEntity;
-        renderFood = mc.player.getRidingEntity() == null;
+        renderFood = !(mc.player.getRidingEntity() instanceof LivingEntity);
```
KnightMiner: This is just the inverse of the above line. Might be easier to just "not" it.

PR #7448, opened 2020-10-29
Implement feature for mods to define new world types

gigaherz:
This PR overlaps with #7353. I realized that other PR exists, after I already had written this code.

Unlike 7353, this PR does not attempt to implement any stock GUI for level-type options, nor does it wrap nor allow customizing of the vanilla world types, I considered those to be out of the scope of my idea.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeConfig.java
+++ b/src/main/java/net/minecraftforge/common/ForgeConfig.java
@@ -108,8 +108,30 @@ public class ForgeConfig
                     .comment("Fix advancement loading to use a proper topological sort. This may have visibility side-effects and can thus be turned off if needed for data-pack compatibility.")
                     .translation("forge.configgui.fixAdvancementLoading")
                     .define("fixAdvancementLoading", true);
+
             builder.pop();
         }
+    }
+
+    /**
+     * General configuration that doesn't need to be synchronized but needs to be available before server startup
+     */
+    public static class Common {
+        public final ForgeConfigSpec.ConfigValue<? extends String> defaultWorldType;
+
+        Common(ForgeConfigSpec.Builder builder)
+        {
+            builder.comment("General configuration settings")
+                    .push("general");
+
+            defaultWorldType = builder
+                    .comment("Defines a default world type to use. The vanilla default world type is represented by 'default'.")
+                    .translation("forge.configgui.defaultWorldType")
+                    .define("defaultWorldType", "default");
+
+            builder.pop();
+        }
+
     }
 
     /**
@@ -193,6 +215,15 @@ public class ForgeConfig
         final Pair<Client, ForgeConfigSpec> specPair = new ForgeConfigSpec.Builder().configure(Client::new);
         clientSpec = specPair.getRight();
         CLIENT = specPair.getLeft();
+    }
+
+
+    static final ForgeConfigSpec commonSpec;
+    public static final Common COMMON;
+    static {
+        final Pair<Common, ForgeConfigSpec> specPair = new ForgeConfigSpec.Builder().configure(Common::new);
+        commonSpec = specPair.getRight();
+        COMMON = specPair.getLeft();
     }
 
 
--- a/src/main/java/net/minecraftforge/common/ForgeConfig.java
+++ b/src/main/java/net/minecraftforge/common/ForgeConfig.java
@@ -125,7 +125,8 @@ public class ForgeConfig
                     .push("general");
 
             defaultWorldType = builder
-                    .comment("Defines a default world type to use. The vanilla default world type is represented by 'default'.")
+                    .comment("Defines a default world type to use. The vanilla default world type is represented by 'default'.",
+                             "The modded world types are registry names which should include the registry namespace, such as 'examplemod:example_world_type'.")
                     .translation("forge.configgui.defaultWorldType")
                     .define("defaultWorldType", "default");
 
--- a/src/main/java/net/minecraftforge/common/world/ForgeWorldType.java
+++ b/src/main/java/net/minecraftforge/common/world/ForgeWorldType.java
@@ -1,0 +1,121 @@
+/*
+ * Minecraft Forge
+ * Copyright (c) 2016-2020.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation version 2.1
+ * of the License.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.common.world;
+
+import net.minecraft.util.ResourceLocation;
+import net.minecraft.util.StringUtils;
+import net.minecraft.util.Util;
+import net.minecraft.util.registry.DynamicRegistries;
+import net.minecraft.util.registry.Registry;
+import net.minecraft.util.text.ITextComponent;
+import net.minecraft.util.text.TranslationTextComponent;
+import net.minecraft.world.DimensionType;
+import net.minecraft.world.biome.Biome;
+import net.minecraft.world.gen.ChunkGenerator;
+import net.minecraft.world.gen.DimensionSettings;
+import net.minecraft.world.gen.settings.DimensionGeneratorSettings;
+import net.minecraftforge.common.ForgeConfig;
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.ForgeRegistryEntry;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class ForgeWorldType extends ForgeRegistryEntry<ForgeWorldType>
+{
+    private static final Logger LOGGER = LogManager.getLogger();
+
+    public static ForgeWorldType getDefaultWorldType()
+    {
+        String defaultWorldType = ForgeConfig.COMMON.defaultWorldType.get();
+
+        if (StringUtils.isNullOrEmpty(defaultWorldType) || "default".equals(defaultWorldType))
+            return null; // use vanilla
+
+        ForgeWorldType def = ForgeRegistries.WORLD_TYPES.getValue(new ResourceLocation(defaultWorldType));
+        if (def == null)
+        {
+            LOGGER.error("The defaultWorldType '{}' specified in the forge config has not been registered. The vanilla default generator will be used.", defaultWorldType);
+        }
+
+        return def;
+    }
+
+    private final IChunkGeneratorFactory factory;
+
+    public ForgeWorldType(IChunkGeneratorFactory factory)
+    {
+        this.factory = factory;
+    }
+
+    public ForgeWorldType(IBasicChunkGeneratorFactory factory)
+    {
+        this.factory = factory;
+    }
+
+    public String getTranslationKey()
+    {
+        return Util.makeTranslationKey("generator", getRegistryName());
+    }
+
+    public ITextComponent getDisplayName()
+    {
+        return new TranslationTextComponent(getTranslationKey());
+    }
+
+    /**
+     * Called from both the dedicated server and the world creation screen in the client.
+     * to construct the DimensionGEneratorSettings:
+     * @return The constructed chunk generator.
+     */
+    public ChunkGenerator createChunkGenerator(Registry<Biome> biomeRegistry, Registry<DimensionSettings> dimensionSettingsRegistry, long seed, String generatorSettings)
+    {
+        return this.factory.createChunkGenerator(biomeRegistry, dimensionSettingsRegistry, seed, generatorSettings);
+    }
+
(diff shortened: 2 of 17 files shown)


alcatrazEscapee:
Between this and #7353, (and previously #7330), it would be nice if we could get one of these into forge as the lack of ability to define custom world generators is quite painful for those doing world generation overhauls. This seems like a nice lightweight implementation (both from a forge maintenance perspective, and in terms of the public modder facing API.

*That aside*, there is one feature I'd still like to see here that's currently missing, and that is the ability to define a specific world type as the new "default" world type. As someone writing a world generation overhaul mod, we get frequent issues / complaints why our world generation does not apply by default. Being able to set a new default (Note: this does not need to *guarantee* that our default is applied over other mods which do the same thing - it just needs to take precedence over vanilla) helps alleviate these complaints, and makes the whole situation much smoother.

Ideally this default is both equally settable as the default generator type in `CreateWorldScreen`, and also as the default field for `sever.properties` when that file is created.

Won-Ton:
This only allows mods to define the chunk generator for the overworld which is a bit too limiting imo. Mods would need to be able to override creation of the `DimensionGeneratorSettings` object (or at-least the Dimension registry) to be able to provide their own generators for the Nether, End, and other dims.

gigaherz:
@Won-Ton Yes that's the big difference between our PRs, I don't think mine is *better*, just different scope.
The goal for this one is to tie into the existing level-type option, and in mc this option is only available for the overworld.

Whether or not this will be preferred over yours, I can not say yet.

@alcatrazEscapee I tried implementing that feature, but I'm not convinced by it. It's too "loose" (not well defined) if I make the default code-based, because there could be more than one mod adding a world type. It probably would need to be a forge config, to choose a default so modpack authors are the ones to pick, instead of the mods.

Cyborgmas:
Having different methods of generating the overworld is a vanilla mechanic, there is no equivalent for the nether/end. 
You can now however use a datapack to override the end / nether dimension completely. If multiple mods do this, the user will just need to order the datapacks in the datapack menu screen to be able to determine who wins. Its not ideal but the possibility is there

srs-bsns:
> `@`alcatrazEscapee I tried implementing that feature, but I'm not convinced by it. It's too "loose" (not well defined) if I make the default code-based, because there could be more than one mod adding a world type. It probably would need to be a forge config, to choose a default so modpack authors are the ones to pick, instead of the mods.

This is precisely what I opined in the other PR. (https://github.com/MinecraftForge/MinecraftForge/pull/7353#pullrequestreview-496031539)

Features should not be added to Forge that create a situation where multiple mods will fight to set some default. There should be a feature, as giga has stated, that allows for a mod pack author to set the default manually, or such a feature should be left to a specialized mod that does this specifically for mod packs. (Much like what CustomMainMenu does to customize the main menu, which was another proposed Forge feature that has been denied on several occasions in the past.)

Won-Ton:
> The goal for this one is to tie into the existing level-type option, and in mc this option is only available for the overworld.

There's `BiomeGeneratorTypeScreens::func_241220_a_` - the "single_biome_caves" overrides it. That could be forwarded to a method on `ForgeWorldType.Factory` with a default implementation the same as `BiomeGeneratorTypeScreens`'s.

Just thinking that this PR has more of a chance of getting looked-at/pulled so it'd be good to have that^ in if possible.

alcatrazEscapee:
> I tried implementing that feature, but I'm not convinced by it. It's too "loose" (not well defined) if I make the default code-based, because there could be more than one mod adding a world type

I don't believe this should be an issue. I think the ideal feature here is that, Forge supports setting a default, with the explicit contract that *this does not guarantee your mod the default!* All it does is put your default above vanilla in priority. This could be exposed to mods, a forge config, exposed to pack makers, to whatever - I don't particularly mind. But having *some standard way* of doing this - despite not being a guarantee -  should be good for the vast majority of world generation overhaul mods. Because **I am not** requesting a "set my mod as default 100% of the time on pain of death" feature. I just want a "use my mod over vanilla, if possible" option.

> Features should not be added to Forge that create a situation where multiple mods will fight to set some default.

As was outlined in the other issue, this is all nice and well in principle, but is also *all over forge*. e.g. registry replacement (last wins). or listening to events and cancelling them (first wins).

Having a common hook with a well understood ordering principle is *much better* than having no hook at all - forcing mods to try and bodge it in themselves if the feature is desired (this defaulting over vanilla avoids *a lot* of headache from users in my case).

gigaherz:
> There's `BiomeGeneratorTypeScreens::func_241220_a_` - the "single_biome_caves" overrides it. That could be forwarded to a method on `ForgeWorldType.Factory` with a default implementation the same as `BiomeGeneratorTypeScreens`'s.

My initial implementation did forward that, and all the other places DimensionGeneratorSettings is constructed. I decided against it because it complicated the implementation and I didn't see the advantage. However, I did not realize that there was at least one existing case of this being overriden, so I'll take another look.

srs-bsns:
> As was outlined in the other issue, this is all nice and well in principle, but is also _all over forge_. e.g. registry replacement (last wins). or listening to events and cancelling them (first wins).

This is not a valid argument for this feature. Runtime level events are not neccessarily mutually exclusive like setting the default generation type is.

> Having a common hook with a well understood ordering principle is _much better_ than having no hook at all - forcing mods to try and bodge it in themselves if the feature is desired (this defaulting over vanilla avoids _a lot_ of headache from users in my case).

This is not a valid argument countering my above opinion. Forge not having this feature is **_not an excuse_** for mods to try to add this feature in themselves. The whole point is that this **_is not something that any one mod should get to decide_**. This feature is pure narcissism.

If your mod, that adds a custom generator type, requires that it's generator type be used for the mod to function, then you should **_clearly state in the mods instructions to do just that_**.

Rongmario:
> This is not a valid argument countering my above opinion. Forge not having this feature is **_not an excuse_** for mods to try to add this feature in themselves. The whole point is that this **_is not something that any one mod should get to decide_**. This feature is pure narcissism.
> 
> If your mod, that adds a custom generator type, requires that it's generator type be used for the mod to function, then you should **_clearly state in the mods instructions to do just that_**.

Good luck getting people to follow simple instructions. It'll just mixins to the rescue again, do you want that?

eerussianguy:
> If your mod, that adds a custom generator type, requires that it's generator type be used for the mod to function, then you should **_clearly state in the mods instructions to do just that_**.

"Mod instructions"? People expect stuff to work when they install a mod. It's pretty simple. Look at any world gen mod -- the BOP curseforge has in big red letters USE OUR WORLD TYPE and people still don't do it. For TFC people come to Discord asking every. single. day. Why their worldgen doesn't work. There's your ivory tower of how mods ought to be, and then there's the reality that users can't read instructions.

Won-Ton:
> My initial implementation did forward that, and all the other places DimensionGeneratorSettings is constructed. I decided against it because it complicated the implementation and I didn't see the advantage. However, I did not realize that there was at least one existing case of this being overriden, so I'll take another look.

Sweet, thanks. Another approach would be to just pass the DimensionGeneratorSettings through a method like this:
```java
default DimensionGeneratorSettings modify(DimensionGeneratorSettings settings, DynamicRegistries registries)
{
        return settings;
}
```
That should be all a mod needs to construct its own more advanced configuration of dimensions and whatnot. Keeps the boiler plate out of the api

srs-bsns:
> Good luck getting people to follow simple instructions. It'll just mixins to the rescue again, do you want that?

Forge doesn't exist to be a moderator for jerks.
You can add a mixin all you want. Then some other modder that gets annoyed with your mod getting set as the default, and not theirs, will then set their mod as default at a later time in the lifecycle until someone sets the default right at the moment that the GUI opens. After this the only solution would be for one mod to crash the game citing incompatibility with another mod forcing users to choose loading one or the other.
The real solution is *Don't be a jerk!*


> "Mod instructions"? People expect stuff to work when they install a mod. It's pretty simple. Look at any world gen mod -- the BOP curseforge has in big red letters USE OUR WORLD TYPE and people still don't do it. For TFC people come to Discord asking every. single. day. Why their worldgen doesn't work. There's your ivory tower of how mods ought to be, and then there's the reality that users can't read instructions.

This is another pointless argument. As someone that has maintained a world gen mod for 1.7-1.12 that requires a custom world type be used, I can say unequivocally, that the amount of people that require extra instruction beyond what is provided is so small that it's not worth counting. There are other things that mods can do that don't require such a feature, such as having a GUI pop-up in the new world screen that states the mods instructions.

gigaherz:
There's no need to be heated. I think I have come up with a reasonable solution that may prevent mods fighting each other.
Commit in a minute.

src/test/java/net/minecraftforge/debug/world/ForgeWorldTypeTest.java
```diff
@@ -0,0 +1,57 @@
...
+            @Override
+            protected void init()
+            {
+                super.init();
+
+                addButton(new Button(0, 0, 120, 20, new StringTextComponent("close"), btn -> {
+                    Minecraft.getInstance().displayGuiScreen(returnTo);
+                }));
+            }
+        });
```
Won-Ton, line 80: The customization screens need a way of sending the modified `DimensionGeneratorSettings` back to the `WorldOptionsScreen`. The built-in BiomeGeneratorTypeScreens use `WorldOptionsScreen::func_239043_a_` which has protected access (I just AT'd it in my PR)

gigaherz: True. I have also added that AT line since it makes sense to use the existing vanilla accessor.

src/main/java/net/minecraftforge/common/ForgeConfig.java
```diff
@@ -108,10 +108,35 @@
...
+            builder.comment("General configuration settings")
+                    .push("general");
+
+            defaultWorldType = builder
+                    .comment("Defines a default world type to use instead of the vanilla 'default' default.",
+                            "Special values:",
+                            "\t\"default\": Use the vanilla default, ignore mod-provided defaults.",
+                            "\t\"proposed-default\": Use the mod-provided default, or vanilla if none present (default).")
+                    .translation("forge.configgui.defaultWorldType")
+                    .define("fixAdvancementLoading", "proposed-default");
```
alcatrazEscapee:
```
 .define("defaultWorldType", "proposed-default");
```
typo?

gigaherz: I'm not seeing the typo?

EDIT: Oh wait copypasta 🤦

Won-Ton:
I've had a chance to test things out, make my own world-type etc and I can do everything I'd like to be able to 👍
I look forward to this being taken out of draft ;]

LexManos:
I'm vetoing the default management for all the egomaniacal reasons listed above and the experience over the last 10 years that Modders can't play nice with each other when it comes to defaults. I'm not opposed to it being a config option that mod pack creators can set. But I AM opposed to modders setting it via code. This is purely a end user choice. Modder's shouldn't be deciding for the end user. 

Also, final warning, "I'll just hack it with a coremod" is not a option/threat we will suffer. Any talk of that may result in blocks/bans. We are here to do the best we can for the most people.

src/main/java/net/minecraftforge/common/ForgeConfig.java
```diff
@@ -108,10 +108,32 @@
...
+    public static class Common {
+        public final ForgeConfigSpec.ConfigValue<? extends String> defaultWorldType;
+
+        Common(ForgeConfigSpec.Builder builder)
+        {
+            builder.comment("General configuration settings")
+                    .push("general");
+
+            defaultWorldType = builder
+                    .comment("Defines a default world type to use instead of the vanilla 'default' default.")
```
sciwhiz12: Might I suggest a rewording of this, to avoid the doubled-up "default" word?
```suggestion
                    .comment("Defines a default world type to use. The vanilla default world type is represented by 'default'.")
```

src/main/java/net/minecraftforge/common/world/ForgeWorldType.java
```diff
@@ -0,0 +1,121 @@
...
+    {
+        String defaultWorldType = ForgeConfig.COMMON.defaultWorldType.get();
+
+        if (StringUtils.isNullOrEmpty(defaultWorldType) || "default".equals(defaultWorldType))
+            return null; // use vanilla
+
+        ForgeWorldType def = ForgeRegistries.WORLD_TYPES.getValue(new ResourceLocation(defaultWorldType));
+        if (def == null)
+        {
+            LOGGER.error("The defaultWorldType specific in the forge config '{}' has not been registered. The vanilla default generator will be used.", defaultWorldType);
```
sciwhiz12: Just rearranging and fixing a probable misspell.
```suggestion
            LOGGER.error("The defaultWorldType '{}' specified in the forge config has not been registered. The vanilla default generator will be used.", defaultWorldType);
```

gigaherz: oops, I didn't English very well there.

src/main/java/net/minecraftforge/common/ForgeConfig.java
```diff
@@ -108,10 +108,32 @@
...
+    public static class Common {
+        public final ForgeConfigSpec.ConfigValue<? extends String> defaultWorldType;
+
+        Common(ForgeConfigSpec.Builder builder)
+        {
+            builder.comment("General configuration settings")
+                    .push("general");
+
+            defaultWorldType = builder
+                    .comment("Defines a default world type to use. The vanilla default world type is represented by 'default'.")
```
Won-Ton: Would it be worth noting in the comment that modded world-types must be specified with a namespace? Modders will need to relay that information anyway (for server.properties) but there could still be confusion since "default" isn't prefixed with "minecraft:"

gigaherz: Good idea. It may be confusing otherwise for those who are used to writing "biomesoplenty", which would mean "minecraft:biomesoplenty" with this change.

PR #7453, opened 2020-10-31
Expanded EarlyLoaderGUI and moved LoadingErrorScreen initialization to allow for flexibility in rendering and potential future upgrades.

AtelierAmber:
Abstracting out the definition of the EarlyLoaderGUI and allowing for protected access to critical methods and fields allows for flexibility in the rendering pipeline. For example, running on hardware that may require specific opengl calls can now be handled with a subclass of EarlyLoaderGUI instead of having to do context checks in the EarlyLoaderGUI itself. 

This also applies to LoadingErrorScreen, however, this can be expanded more in the future if conditional ErrorScreen's were to be implemented. For example, a warning screen that displays warnings but does not kill the game (registry warnings, missing textures), or a verbose version of the LoadingErrorScreen. 

An argument could be made for implementing these features primarily in ClientModLoader but that could end up creating a large number of unnecessary parameters to begin and completeModLoading.


AtelierAmber:
Moved to experimental implementation on personal fork. 
@cpw

PR #7455, opened 2020-10-31
Make possibility to register ConfiguredFeature and StructureFeature with ForgeRegistry

Mysterious-Dev:
This PR adds the possibility of registering ConfiguredFeature  and StructureFeature  with the DeferredRegistry rather than with the Minecraft method.


patches/minecraft/net/minecraft/world/gen/feature/ConfiguredFeature.java.patch
```diff
@@ -0,0 +1,11 @@
+--- a/net/minecraft/world/gen/feature/ConfiguredFeature.java
++++ b/net/minecraft/world/gen/feature/ConfiguredFeature.java
+@@ -15,7 +15,7 @@
+ import org.apache.logging.log4j.LogManager;
+ import org.apache.logging.log4j.Logger;
+ 
+-public class ConfiguredFeature<FC extends IFeatureConfig, F extends Feature<FC>> implements IDecoratable<ConfiguredFeature<?, ?>> {
++public class ConfiguredFeature<FC extends IFeatureConfig, F extends Feature<FC>> extends net.minecraftforge.registries.ForgeRegistryEntry<ConfiguredFeature<?, ?>>  implements IDecoratable<ConfiguredFeature<?, ?>> {
```
pupnewfster: There is an extra space being added before implements

src/main/java/net/minecraftforge/common/world/StructureSpawnManager.java
```diff
@@ -99,7 +99,7 @@ private static void gatherEntitySpawns(Map<Structure<?>, StructureSpawnInfo> str
             Structure<?> structure = entry.getKey();
             StructureSpawnInfo spawnInfo = entry.getValue();
             //Note: We check if the structure has spawns for a type first before looking at the world as it should be a cheaper check
-            if (spawnInfo.spawns.containsKey(classification) && structureManager.getStructureStart(pos, spawnInfo.insideOnly, structure).isValid())
+            if (spawnInfo.spawns.containsKey(classification) && structureManager.func_235010_a_(pos, spawnInfo.insideOnly, structure).isValid())
```
pupnewfster: This looks like you are building against the version from before mappings got updated a few days ago, you should update forgedev as I am pretty sure this change will cause it not to build properly if merged.

src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -166,9 +168,11 @@ public static void init()
         makeRegistry(WORLD_CARVERS, c(WorldCarver.class)).disableSaving().disableSync().create();
         makeRegistry(SURFACE_BUILDERS, c(SurfaceBuilder.class)).disableSaving().disableSync().create();
         makeRegistry(FEATURES, c(Feature.class)).addCallback(FeatureCallbacks.INSTANCE).disableSaving().create();
+        makeRegistry(CONFIGURED_FEATURES, c(ConfiguredFeature.class)).addCallback(FeatureCallbacks.INSTANCE).disableSaving().create();
```
pupnewfster, line 171: I believe both the configured features and configured structure features are also dynamic registries so this potentially need to be done in a more similar way to biomes being done, unless the reason biomes are like they are forge registry wise because vanilla only has a dynamic registry for them and not both types of registries.

Mysterious-Dev: So I would have to try to integrate this like this [commit ](https://github.com/MinecraftForge/MinecraftForge/commit/d8eaeabf3cf03b2de05c23ce88c3b75548720014) ?

pupnewfster: I think so

patches/minecraft/net/minecraft/world/gen/feature/ConfiguredFeature.java.patch
Mysterious-Dev: Fix

src/main/java/net/minecraftforge/common/world/StructureSpawnManager.java
Mysterious-Dev: Fix

src/main/java/net/minecraftforge/registries/GameData.java
Mysterious-Dev: I changed some thing, not the GameData but by testing with a DataPack adding structures and features, it works nicely

TelepathicGrunt:
Dynamic Registries shouldn't be touched by Forge. They exist mainly for per-world registries which Forge does not handle very well. I would encourage that none of the Dynamic Registries be made into Forge Registries as this would cause massive issues. 

One being that since everyone uses Registry.register for these right now, all worldgen mods will break as Forge Registries lock out that registry method. 

Second being that since these registries are for the Dynamic Registry, a Forge registry would not be helpful because during worldgen, the Forge registry would be incorrect since it holds the stuff registered before the world is entered and not the new JSON entries. For example, right now Forge keeps the Biome registry a Forge registry despite it being a Dynamic Registry because legacy reason. If during worldgen you try to use the Forge registry of biomes to try and get, say, a biome's ID... datapack biomes and mods with JSON biomes will not be in that registry. Instead, people should only be using the Dynamic Registry once a world is made. In fact, one of the biggest problems I keep seeing from mods is people are making Biome Providers but then use the Forge Registry of biomes to try and convert IDs to biomes or pass in list of biomes into the constructor. Doing this will cause "Unknown biome ID: -1" issues and /locate breaking for other mod's structures added to biomes. And so, by making the ConfiguredFeature and StructureFeature into Forge registries, more of this misuse will occur and cause mod conflicts and datapack conflicts.

Sorry if I came off a bit harsh but this one PR is a hard no from me. In fact, I would argue it would be better that the Forge Biome registry should be removed to try and force people to use the Dynamic Registry properly but this would break biome mods currently as they had to switch to the Forge Registry. 

In all, it's just not a good time for worldgen modders lol

Mysterious-Dev:
@TelepathicGrunt Then if this is a problem, what we can do is put the register functions of the Features and StructureFeatures classes in public with accesstransformer so as not to have to rewrite them in our classes.

Mysterious-Dev:
It will break less thing

TelepathicGrunt:
Just in case there was a misunderstanding, you do not need to AT those methods at all. You can simply use 
```
Registry.register(WorldGenRegistries.CONFIGURED_STRUCTURE_FEATURE, new ResourceLocation(StructureTutorialMain.MODID, "configured_run_down_house"), CONFIGURED_RUN_DOWN_HOUSE);
```
to register StructureFeatures and ConfiguredFeatures.

Mysterious-Dev:
Yes it's true

Mysterious-Dev:
To avoid polluting and as I realize that it can break a lot of generation mod, I close this pr

PR #7456, opened 2020-11-02
Add support for referencing forge's resources and specifying existing mods to data generators

pupnewfster:
This PR adds support for declaring that a mod's resources should be added to the existing file helper's resource check b adding an `--existing-mod` argument that takes a modid. This PR also auto-adds forge as an existing mod so that modders can reference forge resources from their data generators without having running into errors, similar to how vanilla resources can be referenced.

### DIFF
--- a/patches/minecraft/net/minecraft/data/Main.java.patch
+++ b/patches/minecraft/net/minecraft/data/Main.java.patch
@@ -1,10 +1,11 @@
 --- a/net/minecraft/data/Main.java
 +++ b/net/minecraft/data/Main.java
-@@ -21,8 +21,12 @@
+@@ -21,8 +21,13 @@
        OptionSpec<Void> optionspec6 = optionparser.accepts("all", "Include all generators");
        OptionSpec<String> optionspec7 = optionparser.accepts("output", "Output folder").withRequiredArg().defaultsTo("generated");
        OptionSpec<String> optionspec8 = optionparser.accepts("input", "Input folder").withRequiredArg();
 +      OptionSpec<String> existing = optionparser.accepts("existing", "Existing resource packs that generated resources can reference").withRequiredArg();
++      OptionSpec<String> existingMod = optionparser.accepts("existing-mod", "Existing mods that generated resources can reference the resource packs of").withRequiredArg();
 +      OptionSpec<java.io.File> gameDir = optionparser.accepts("gameDir").withRequiredArg().ofType(java.io.File.class).defaultsTo(new java.io.File(".")).required(); //Need by modlauncher, so lets just eat it
 +      OptionSpec<String> mod = optionparser.accepts("mod", "A modid to dump").withRequiredArg().withValuesSeparatedBy(",");
 +      OptionSpec<Void> flat = optionparser.accepts("flat", "Do not append modid prefix to output directory when generating for multiple mods");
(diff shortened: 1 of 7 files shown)


cpw:
Does this affect forge's datagen? We use this to generate forge's resources too..

pupnewfster:
It seems to mostly work.

The way I tested it was by adding the line I am changing in the data generator test and validating it crashed. Then I added forge as existing by specifying it with the new `--existing-mod` argument, and it properly worked. Next I removed the manual specification of forge existing and uncommented to make forge be automatically added (similar to vanilla being added), which still worked. The reason I believe it works for forge's own data generators still is that I also at the end tested by running forge's data generator and it ran fine, didn't end up changing any of the files or crash.

I also just tested by removing some of forge's generated resources to see if it properly is unable to find them on the initial run or if it thinks they exist for one run before it then deletes them and it seems to find them. Do you think a reasonable solution to this would be to make it so that forge only gets automatically added if it has not been specified as a `--mod`?

patches/minecraft/net/minecraft/data/Main.java.patch
```diff
@@ -24,9 +25,10 @@
 -         datagenerator.func_200392_c();
 +         Collection<Path> inputs = optionset.valuesOf(optionspec8).stream().map(Paths::get).collect(Collectors.toList());
 +         Collection<Path> existingPacks = optionset.valuesOf(existing).stream().map(Paths::get).collect(Collectors.toList());
++         java.util.Set<String> existingMods = new java.util.HashSet<>(optionset.valuesOf(existingMod));
 +         java.util.Set<String> mods = new java.util.HashSet<>(optionset.valuesOf(mod));
 +         boolean isFlat = mods.isEmpty() || optionset.has(flat);
-+         net.minecraftforge.fml.DatagenModLoader.begin(mods, path, inputs, existingPacks, flag2, flag1, flag3, flag4, flag5, isFlat);
++         net.minecraftforge.fml.DatagenModLoader.begin(mods, path, inputs, existingPacks, existingMods, flag2, flag1, flag3, flag4, flag5, isFlat);
```
TheCurle, line 31: Why is this in the middle of the argument list?

src/main/java/net/minecraftforge/fml/DatagenModLoader.java
```diff
@@ -39,14 +40,19 @@ public static boolean isRunningDataGen() {
...
+    @Deprecated //TODO: Remove in 1.17
+    public static void begin(final Set<String> mods, final Path path, final Collection<Path> inputs, Collection<Path> existingPacks, final boolean serverGenerators, final boolean clientGenerators, final boolean devToolGenerators, final boolean reportsGenerator, final boolean structureValidator, final boolean flat) {
+        begin(mods, path, inputs, existingPacks, Collections.emptySet(), serverGenerators, clientGenerators, devToolGenerators, reportsGenerator, structureValidator, flat);
+    }
+
+    public static void begin(final Set<String> mods, final Path path, final Collection<Path> inputs, Collection<Path> existingPacks, Set<String> existingMods, final boolean serverGenerators, final boolean clientGenerators, final boolean devToolGenerators, final boolean reportsGenerator, final boolean structureValidator, final boolean flat) {
```
TheCurle, line 48: Middle of the argument list..
This is a Forge class yes, but why?

TheCurle (commented):
I'm going to test this a little. 
Looks okay, apart from some weirdness with the API changes.

src/main/java/net/minecraftforge/fml/DatagenModLoader.java
pupnewfster: Because argument wise it makes more sense going next to existing packs than at the end after a ton of random boolean values, and once we hit 1.17 and the deprecated old one can be removed without worrying that someone is calling it for some reason from somewhere else, it really won't matter as it won't be the "middle" it will just be the "normal" place for it.

PR #7457, opened 2020-11-02
[1.17.x] Remove forge command 'setdimension'

### LINKED ISSUE #7254: Game hangs upon switching dimension with `forge setdimension`
In 1.16.2, the game hangs when you invoke `forge setdimension @p minecraft:nether`, but using `execute in minecraft:the_end run teleport ~ ~ ~` works just fine. Since vanilla commands now support switching dimensions I recommend removing this command or executing the command by proxy

mrp-v2:
Removes 'forge setdimension', as a vanilla alternative exists. Marked as 1.17 because it is a breaking change.


KnightMiner:
Until 1.17, think it would be good for the Forge command to print a chat message when using it directing users to the vanilla command? Something like "This command will be removed in a future version, use `/vanillacommand syntax`"

mrp-v2:
Yes, that is a good idea, but that would be a separate PR.

alcatrazEscapee:
@mrp-v2 this would be a good time to make this PR not a draft :)

sciwhiz12:
1.17 is now outside of the supported versions. The class removed by this PR has been removed during the 1.18 process, and the class cannot be removed in 1.16 due to the breaking changes policy. For these reasons, I am closing this PR.

PR #7461, opened 2020-11-03
[1.16.x] Make forge build script inject versions into MDK mods.toml

### LINKED ISSUE #7460: 1.16.4 Example mods.toml loaderVersion and dependency version not up to date.
In Version 1.16.4 - 35.0.1of the forge MDK the loadVersion and the dependency version range of forge are 34 and not 35.
And the version of the Minecraft dependency is 1.16.3 instead of 1.16.4

Here the zipped mods.toml:
[mods.zip](https://github.com/MinecraftForge/MinecraftForge/files/5482466/mods.zip)

mrp-v2:
Fixes #7460.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -879,7 +879,8 @@ project(':forge') {
                 FORGE_NAME: project.name,
                 MC_VERSION: MC_VERSION,
                 MAPPING_CHANNEL: MAPPING_CHANNEL,
-                MAPPING_VERSION: MAPPING_VERSION
+                MAPPING_VERSION: MAPPING_VERSION,
+                LOADER_VERSION: SPEC_VERSION.substring(0, SPEC_VERSION.indexOf("."))
             ])
             rename 'gitignore\\.txt', '.gitignore'
             rename 'gitattributes\\.txt', '.gitattributes'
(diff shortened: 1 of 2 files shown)


KnightMiner:
Would be better to use Minecraft range `[1.16.3,1.16.4]` and keep the old loader and FML versions, code built for 1.16.4 works on 1.16.3. Arguably you leave out the strict range entirely and go `[1.16.3,)`

LexManos:
This needs to be fixed once and for all, by making the build script replace these tokens when building the MDK zip like we do for build.gradle

mrp-v2:
Correct me if I've done this incorrectly, but the build script now injects the versions.
I built the MDK locally to verify the versions are injected correctly.

raziel23x (approved):

TheCurle:
The version is hard-coded for a reason - Lex wants to prevent people from copy-pasting this code from eg. old versions of the mdk.
Just changing it to 1.16.4 is good enough for now.

mrp-v2:
I'm confused. Lex said to make the build script inject the Minecraft and Forge version, but now you want me to do it manually? To clarify, by buildscript I've been refering to the Forge buildscript, not the mdk build.gralde.

TheCurle:
Oh, i totally misread, this is indeed what he wanted.

pupnewfster:
https://github.com/MinecraftForge/MinecraftForge/commit/38a5400a8c878fe39cd389e6d4f68619d2738b88

PR #7463, opened 2020-11-04
Allow for custom command argument types

diesieben07:
Allow actually registering custom command argument types by using the new vanilla network filter system to filter them on vanilla connections.

After this PR Forge clients will see the following:
commands_forge

Vanilla clients will continue to see the following (improving this behavior is out of scope for this PR).
commands_vanilla

This is enabled by actually registering Forge's custom argument types (`ModIdArgument` and `EnumArgument`), this was previously disabled due to it preventing vanilla clients from connecting.

I have also removed the existing patch to `SCommandListPacket` for the following reasons:
1. If the argument type could not be found on the client it's data would have to be read first, before reading the ResourceLocation for the custom suggestion data. However we don't know what the argument type is, so we can't read it's data. Trying to read a ResourceLocation here will only produce garbled data.
2. This patch was completely ineffective anyways, because the condition it checked will never evaluate to true. `(i & 16) != 0` cannot be true, if `i` is initialized as `buf & 3`.

Potentially improving the behavior of custom argument types that are present on the server, but not on a _forge client_ is out of scope of this PR. That state will continue to cause errors and would need a separate system, where the client can skip over unknown argument type data. Probably a length would have to be prepended, however that requires altering the protocol.


patches/minecraft/net/minecraft/command/arguments/ArgumentTypes.java.patch
```diff
@@ -0,0 +1,13 @@
+--- a/net/minecraft/command/arguments/ArgumentTypes.java
++++ b/net/minecraft/command/arguments/ArgumentTypes.java
+@@ -220,4 +220,10 @@
+          this.field_197481_c = p_i48088_3_;
+       }
+    }
++
```
marchermans: Could you remove the empty lines here? 
Or is this policy to leave them here?
Also is a ForgeMarker not required here?

marchermans (changes requested):
Some nitpicking on the formatting, not sure if it is accurate or not.

patches/minecraft/net/minecraft/command/arguments/ArgumentTypes.java.patch
diesieben07: I don't think we have the Forge marker in very many places anymore, but I don't know. I'll remove the empty line.

patches/minecraft/net/minecraft/command/arguments/ArgumentTypes.java.patch
```diff
@@ -0,0 +1,12 @@
...
++++ b/net/minecraft/command/arguments/ArgumentTypes.java
+@@ -220,4 +220,9 @@
+          this.field_197481_c = p_i48088_3_;
+       }
+    }
++   @javax.annotation.Nullable public static ResourceLocation getId(ArgumentType<?> type) {
++      Entry<?> entry = func_201040_a(type);
++      return entry == null ? null : entry.field_197481_c;
++   }
++
```
marchermans: Another empty line.....
Sorry for being annoying about this. But smaller patches benefit everybody.

marchermans (changes requested):

patches/minecraft/net/minecraft/command/arguments/ArgumentTypes.java.patch
diesieben07: All good, I am just being an idiot.

marchermans (approved):

patches/minecraft/net/minecraft/network/play/server/SCommandListPacket.java.patch
```diff
@@ -1,12 +0,0 @@
---- a/net/minecraft/network/play/server/SCommandListPacket.java
-+++ b/net/minecraft/network/play/server/SCommandListPacket.java
-@@ -119,6 +119,9 @@
-          String s = p_197695_0_.func_150789_c(32767);
-          ArgumentType<?> argumenttype = ArgumentTypes.func_197486_a(p_197695_0_);
-          if (argumenttype == null) {
-+            if ((i & 16) != 0) { // FORGE: Flush unused suggestion data
```
LexManos, line 7: This must of changed in a MC update. It needs to be `buf & 16` not `i & 16`to match the case below it. And it does actually function as needed, at least when I wrote it.

diesieben07: I know that `buf & 16` was the original intention. However the ResourceLocation this patch is trying to read is not the next thing to read here. If `argumenttype` is `null` then the `ArgumentType` was not found, so whatever data it wrote is next in the buffer. _After_ that data the custom suggestion `ResourceLocation` is to be read. The patch was added by @tterrag1098 back in 2019, where it correctly checked `buf & 16`. However I am unsure how that is supposed to have ever fixed anything - for the reasons outlined above.

cpw:
NICE

PR #7464, opened 2020-11-04
[1.17.x] Remove positional isAir

### LINKED ISSUE #7409: [RFC][1.17] Removal of positional isAir.
Our hooks to allow tile entity based air have been around for a long time.
However, they were created in an age where we were limited to 4 bits for metadata.
A time where blocks themselves were limited to 256/4096. But now that both of those restrictions are removed I think it's time to align our hooks and modders more with vanilla's design.

As it sits, we require 50 patches for this one feature as is. Combined with another 130 that could be needed to deal with vanilla edge cases. This is very costly maintenance wise for a feature I am not sure anyone uses anymore.

So I'm proposing in 1.17, that we remove the positional arguments to isAir and use the state based system like most other block properties. Unless we get a valuable argument from this that isn't statistical fear mongering then it'll be headed for removal in 1.17. 

Update: As people seem confused, this is only asking people who OVERRIDE this function. It won't effect people who call this function. As callers do not need to know implementation of the isAir function.

mrp-v2:
Undeprecates `BlockState#isAir()` and removes `BlockState#isAir(IBlockReader, BlockPos)`.

Resolves #7409.


sciwhiz12:
Closing as implemented already in 1.17.

PR #7465, opened 2020-11-04
Updated MDK and README for 1.16.4

Mysterious-Dev:
The title is self-explanatory


Thelonedevil:
This looks to be being handled already by #7461 in a more future proof way (at least the mods.toml part)

PR #7440, opened 2020-10-27
[1.16] Fix Banner.toItem causing crash on server

TheCurle:
Forge uses the SAS to remove @OnlyIn(Dist.CLIENT) from BannerTileEntity.getItem:
`net/minecraft/tileentity/BannerTileEntity func_190615_l(Lnet/minecraft/block/BlockState;)Lnet/minecraft/item/ItemStack;`
But that function calls BannerBlock.forColor, which is also sided.

This PR adds BannerBlock.forColor to the Forge SAS, allowing servers to call BannerTileEntity.getItem again.

NOTE: i have not run checkSAS on this yet

PR #7467, opened 2020-11-05
Add annotation-based tools for dist-based implementation of methods

FoxSamu:
Today I ran into the annoying coding style issue that you have to create a complete new class just for calling a client-only method if available. Because this makes code very very ugly I created a different approach:

A new `@ImplementOn` annotation can specify a method in the same class to be overwritten by another method when on a specific `Dist`. For example:

```java
    public void myMethod() {

    }

    @ImplementOn(dist = Dist.CLIENT, method = "myMethod")
    private void myMethod_client() {
        System.out.println("Client implementation");
    }

    @ImplementOn(dist = Dist.DEDICATED_SERVER, method = "myMethod")
    private void myMethod_server() {
        System.out.println("Server implementation");
    }
```

Then, upon loading of this class, an ASM transformer (`net.minecraftforge.common.asm.RuntimeDistImplementor` to be precise) injects either `myMethod_client` or `myMethod_server` into the method `myMethod` and strips unused `@ImplementOn` methods so that they won't get loaded. The above example will, on `Dist.CLIENT` become:

```java
    public void myMethod() {
        this.myMethod_client();
    }

    @ImplementOn(dist = Dist.CLIENT, method = "myMethod")
    private void myMethod_client() {
        System.out.println("Client implementation");
    }
```

Note how the method is directly delegated: this keeps line numbers and debug info understandable in stack traces. The system works for methods with or without return values, static and virtual methods, and any set of arguments. For `@ImplementOn` to work properly, the following conditions are checked, with the intentions specified:

- The return types of the target method and the injected method must match **exactly**. This is because it finds the target method based on the name specified and the descriptor (including return type) of the injector method.
- The argument lists of the target method and the injected method must match **exactly**. This is because it finds the target method based on the name specified and the descriptor of the injector method.
- The `static` modifier must either be specified on both the target and injected method, or must be omitted on both methods. When one is static and the other isn't, the injection fails. This is because a static method can not reference an instance method. An instance method can access a static method but for the semantics of 'replacing the method' I forced the `static` modifiers to match exactly.
- The injected method **must** be private. This prevents overriding which can cause strange issues as unused injected methods are stripped (subclasses might be unable to load if they call `super`). This also prevents them being called from within another class (even inner classes can't call them - this goes via a bridge method).
- The injected method may never be called (only checked in a development environment, since it checks all instructions in a class and might drop performance a lot). This filters the cases where a method of the same class accesses the injector (including bridge methods), which must not happen either.
- There must be, for each separate `Dist`, only one injected method: you can specify injections for the same method but for different dists. This rule is obvious: you should not inject into a method twice because you will then overwrite your own injection.

Note that this works with method overloading: an `@ImplementOn` annotated method will be injected into the method with the name specified in the annotation's `method` parameter and the descriptor of the injected method. Hence this will work:

```java
    public void myMethod() {
        System.out.println("Common implementation");
    }

    @ImplementOn(dist = Dist.CLIENT, method = "myMethod")
    private void myMethod_client() {
        System.out.println("Client implementation");
    }

    public void myMethod(int argument) {
        System.out.println("Common implementation, argument = " + argument);
    }

    @ImplementOn(dist = Dist.CLIENT, method = "myMethod")
    private void myMethod_client(int argument) {
        System.out.println("Client implementation, argument = " + argument);
    }
```

What's also notable here is that an `@ImplementOn` method is only injected when the `dist` parameter matches the `Dist` of the environment. When the `Dist`s don't match, the target method is left untouched and the injected method is stripped from the class, if the `Dist`s _do_ match, the contents of the target method are stripped **completely** and are overwritten by the necessary bytecode.

---

As an alternative suggestion to the semantics I've implemented in the current pull request, it's also possible to make a `@DelegateOn` annotation which is put above the target method and using the `@OnlyIn` annotations for stripping sided methods. It would then look like this:

```java
    @DelegateOn(client = "myMethod_client", server = "myMethod_server")
    public void myMethod() {

    }

    @OnlyIn(Dist.CLIENT)
    private void myMethod_client() {
        System.out.println("Client implementation");
    }

    @OnlyIn(Dist.DEDICATED_SERVER)
    private void myMethod_server() {
        System.out.println("Server implementation");
    }
```

Drawback is here: the `myMethod_client` and `myMethod_server` render as unused and don't have an annotation indicating they are dynamically being injected, nor is there any visual reason not to call them.

---

I think this is a well-working alternative to `DistExecutor.safeCallWhenOn`, since it doesn't require the targeted method to be in a separate class, it allows the use of arguments and it works with return types.

---

Things that might still be necessary and which are missing now:

- Checking the signatures of the methods (so that generic types and `throws` clauses do match). This is only a semantical issue though, in bytecode it's perfectly safe if a checked exception is thrown without having it in the `throws` clause and due to type erasure generics will work.

### DIFF
--- a/src/fmllauncher/java/net/minecraftforge/common/asm/RuntimeSidedImplementor.java
+++ b/src/fmllauncher/java/net/minecraftforge/common/asm/RuntimeSidedImplementor.java
@@ -1,0 +1,405 @@
+/*
+ * Minecraft Forge
+ * Copyright (c) 2016-2020.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation version 2.1
+ * of the License.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.common.asm;
+
+import java.util.ArrayList;
+import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import net.minecraftforge.api.distmarker.Dist;
+import net.minecraftforge.fml.loading.AdvancedLogMessageAdapter;
+import net.minecraftforge.fml.loading.FMLEnvironment;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.Marker;
+import org.apache.logging.log4j.MarkerManager;
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.Type;
+import org.objectweb.asm.tree.AbstractInsnNode;
+import org.objectweb.asm.tree.AnnotationNode;
+import org.objectweb.asm.tree.ClassNode;
+import org.objectweb.asm.tree.InsnList;
+import org.objectweb.asm.tree.InsnNode;
+import org.objectweb.asm.tree.MethodInsnNode;
+import org.objectweb.asm.tree.MethodNode;
+import org.objectweb.asm.tree.VarInsnNode;
+
+import cpw.mods.modlauncher.serviceapi.ILaunchPluginService;
+
+/**
+ * Modifies specified enums to allow runtime extension by making the $VALUES field non-final and
+ * injecting constructor calls which are not valid in normal java code.
+ */
+public class RuntimeSidedImplementor implements ILaunchPluginService
+{
+
+    private static final Logger LOGGER = LogManager.getLogger();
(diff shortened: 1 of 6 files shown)


diesieben07:
I don't think we should be adding more ASM based hackyness because you are too lazy to write a separate class.

TheCurle:
I concur. Sided safe referencing is a Java pattern, and it is generally encouraged to separate code that will be used to build two different releases of the same program. In this case, client and server of a game.

Your solution is no better than just using OnlyIn, which is known to break a lot of things: see #7440.

FoxSamu:
Closed by the dictators who made `@Mod`, `@Mod.EventBusSubscriber`, `@SubscribeEvent`, `@OnlyIn`, `IExtensibleEnum`, ... neither are they java patterns in this case

TheCurle:
Welcome to hacking into a game that was never built to support mods. 
OnlyIn exists only to support vanilla, for the record.
SubscribeEvent is superceded recently by DeferredRegister, and your argument is overall invalid.

Also, yes, it is a pattern. This is standard procedure...

gigaherz:
A few things:
* OnlyIn is used to mark vanilla methods that are missing, it doesn't cause them to be missing. The OnlyIn transformer is a hack, and only exists for specific exceptional situations, it's not a convenience and will never be endorsed as one.
* `@Mod` and `@SubscribeEvent` are API markers, they are used to identify features, it's very much not the same as what you are proposing.
* IExtensibleEnum isn't an annotation? It's a hack needed by modders.

To use client-only code correctly you just need, at most, one (1) helper class, which has some forwarders and keeps all client-only references all hidden from the shared logic. You just need to follow two simple rules:
1. This helper class shouldn't have any outward references to the client-only code (no client-only extends/implement, params or fields)
2. The methods in this class should avoid "implicit casts" (eg, assigning Minecraft#player to a PlayerEntity variable), which cause the jvm to have to classload the types so it can verify they are compatible.

That's all. One class and some basic defensive coding.

PR #7471, opened 2020-11-07
Fix #7363 (Removing structures causes worlds to be unsaveable)

### LINKED ISSUE #7363: Removing structures leads to chunks being permanently unsaveable
This is a vanilla issue (MC-194811) but affects modded worlds a lot more often, so a fix should probably be made in forge.

In `ChunkSerializer.unpackStructureReferences`, the structure reference map `field_236365_a_` is queried but the resulting value is not null-checked. The result could be null if the mod that added that structure was later removed.

This means the returned map has a entry with key=null. Later on, when the game tries to save this (`ChunkSerializer.writeStructures`), it doesn't null check the keys of the map and blows up, meaning any progress made in this chunk is never saved.

Recommended solution: Add a null check to `unpackStructureReferences` and just skip it if the structure is not in the registry.

alcatrazEscapee:
See the issue, everything of note is documented there. This simply does as is recommended.

### DIFF
--- a/patches/minecraft/net/minecraft/world/chunk/storage/ChunkSerializer.java.patch
+++ b/patches/minecraft/net/minecraft/world/chunk/storage/ChunkSerializer.java.patch
@@ -17,6 +17,15 @@
                    chunkprimer.func_201637_h(blockpos);
                 }
              }
+@@ -167,7 +168,7 @@
+       Heightmap.func_222690_a(ichunk, enumset);
+       CompoundNBT compoundnbt4 = compoundnbt.func_74775_l("Structures");
+       ichunk.func_201612_a(func_235967_a_(p_222656_1_, compoundnbt4, p_222656_0_.func_72905_C()));
+-      ichunk.func_201606_b(func_227075_a_(p_222656_3_, compoundnbt4));
++      net.minecraftforge.common.ForgeHooks.fixNullStructureReferences(ichunk, func_227075_a_(p_222656_3_, compoundnbt4));
+       if (compoundnbt.func_74767_n("shouldSave")) {
+          ichunk.func_177427_f(true);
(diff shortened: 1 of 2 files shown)


50ap5ud5 (approved):

50ap5ud5:
This looks good. I can verify the fix works with a test mod I made: https://github.com/50ap5ud5/MinecraftForge/tree/chunk-fix-test. 

The way I tested this was: 

1. I created a new world and found one of my test structures
2. Saved the world save and closed it.
3. Disabled the registration of the structure in the test mod.
4. Opened up the game again and reopened the previous world.
5. The world is able to load successfully despite the console spamming `Unknown structure start: chunk_save_test:log_cabin`.
6. Placed some blocks around the chunk that the structure resided in.
7. Saved the world save and closed the game.
8. Reopen the game and reopen the previous world. Verified that the blocks placed were saved.

alcatrazEscapee:
Lovely! Works for me 👍

ichttt:
Why didn't you follow williewillus' suggenstion to just add a null check? As that's what vanilla seems to do everywhere else where it uses the map except at this place where it crashes

alcatrazEscapee:
A number of reasons:

- Adding a null check would avoid the "illegal structure" message. This is desired as this is fundamentally changing the chunk (i.e. these structures are being removed and won't be regenerated if you enter the world again with the structure mod intact.) Making this loud in the log is good to keep.
- The method in question does not get passed the chunk - and so if a null key is detected, it cannot mark the chunk as modified (which it has been, since the structures have been removed and should be saved)
- This avoids a larger patch in `ChunkSerializer` than adding the null check + possibly modifying logging logic + figuring out a way to set the chunk as modified.

PR #7472, opened 2020-11-08
Add getType to Capability.

eutro:
This PR adds a `getType` method to the `Capability` class, that returns the target class of the Capability, i.e. the parameter to the `CapabilityInject` annotation. This may be useful when the value of the capability is received from a source that cannot guarantee the correct type, for example if it's from a user-written script. This allows for checking the type safely and without causing heap pollution.

While currently the class can be retrieved by using `Class.forName(capability.getName())`, this is an awkward solution, as the name being the same as the class name is only an implementation detail.

### DIFF
--- a/src/main/java/net/minecraftforge/common/capabilities/Capability.java
+++ b/src/main/java/net/minecraftforge/common/capabilities/Capability.java
@@ -91,7 +91,7 @@ public class Capability<T>
     public String getName() { return name; }
 
     /**
-     * @return The target interface of this capability.
+     * @return The target interface of this capability, the value of the {@link CapabilityInject} annotation.
      */
     public Class<T> getType() { return type; }
 
@@ -149,9 +149,9 @@ public class Capability<T>
 
     // INTERNAL
     private final String name;
+    private final Class<T> type;
     private final IStorage<T> storage;
     private final Callable<? extends T> factory;
-    private final Class<T> type;
 
     Capability(String name, Class<T> type, IStorage<T> storage, Callable<? extends T> factory)
     {
--- a/src/main/java/net/minecraftforge/common/capabilities/Capability.java
+++ b/src/main/java/net/minecraftforge/common/capabilities/Capability.java
@@ -91,6 +91,11 @@ public class Capability<T>
     public String getName() { return name; }
 
     /**
+     * @return The target interface of this capability.
+     */
+    public Class<T> getType() { return type; }
+
+    /**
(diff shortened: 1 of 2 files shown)


LexManos:
No, the capability system is explicitly designed so you don't need to know and should not access the class itself. It would just cause people to do slow reflection processes when they should not.
The Capability Object are singletons for the class. So you shouldn't need to know the type as long as you just compare the capability instances.
So you're gunna have to explain yourself a lot better then "I want it, and I CAN hack it right now so I will if you don't accept this"

eutro:
That is not what I meant at all. My motivation is exactly what I said: I am receiving the value from a source that cannot guarantee the correct type, in this case a script written by the user.

LexManos:
And why can you not guarantee the correct type?

eutro:
The scripting language is dynamically typed.

LexManos:
And? In order for it to interact with standard java it needs to provide the correct types.
Also, as a side note, having scriptable capabilities is a horrible idea. The entire point is to be fast. Whereas interpreted languages are not. I have no idea what your project is or why you're trying to do it. But Minecraft is Java and you should encourage people to use Java.

eutro:
It can provide the types, I just cannot guarantee that they are correct, since I don't know what the required type of an arbitrary `Capability<T>` object is. The scripting language in question, while not quite as performant as Java, is still compiled, and fast enough.

LexManos:
The entire point of the capability objects is that it's a 1 to 1 of the requested type. If the get function returns the wrong object that is a modder error. If you're relying on your scripting engine to dynamically cast objects for you you're gunna have a bad time. It would be like calling `String getName()` and checking if the return is an instance of String. Combine that with the fact that you SHOULD be creating your objects outside of the get function l, there is no need for this. So, the answer is no. Use the system we designed how we designed it. We're not going to add extra functionality to explicitly allow you to do what we explicitly designed the system to not need to do.

Feel free to hop on the discord or forums to flesh out your usecase and figure out how to do it correctly. But this isnt going in.

eutro:
As you say, the point is that it's 1 to 1 with the requested type, I'm just passing through to the user script to an extent. Replace modder error with user error, and that's exactly what I'm trying to catch.

DaemonUmbra:
Can you please join [the Discord](https://discord.gg/UvedJ9m) for further discussion, and please provide examples of how this is even something that needs to be addressed.

Issue #6928, opened 2020-07-06
[1.16] Unable to add attributes to existing entity types


maxanier:
So with the new vanilla `GlobalEntityTypeAttributes` system it does not seem to be possible to add attributes to existing entity types (e.g. player entities) anymore.
Forge patches the corresponding static method in `PlayerEntity` to introduce its `REACH_DISTANCE` attribute, but for mods there currently isn't a solution. Right?

As all of the relevant code is static, this will get ugly to resolve.

So far, I see four alternatives
1) Replace the `ImmutableMap` Building Line `GlobalEntityTypeAttributes#73` with a call to Forge file that does the same, but additionally posts an event or checks a static registry before the attribute map is finalized.
Con: this means when MC updates this line, the changes have to be transferred to the Forge file
2) Patch `LivingEntity` and put a wrapper around the `AttributeModifierManager` which allows registering new attributes per Entity after creation
Con: Wrapper adds additional "bug-prone" layer and runtime overhead is introduced
3) Patch the `AttributeModifierMap` to remove the "immutability", allowing mods to register new attributes per entity type after the `GlobalEntityTypeAttributes` have been initialized
Con: Might be more error prone, if mods change things at the wrong time.
4) Patch each entity's static attribute builder method to post an event or check a static registry. 
Con: Number of patches to high -> non viable
Alternative: Just patch `LivingEntity` (to allow adding attributes to all interesting entities) and potentially `PlayerEntity` for player specific things

I tried discussing this on Discord, but didn't get any response. But I would like to get some feedback before I waste time with a PR for this

maxanier:
An alternative attempt is to use an AT entry to make the `AttributeModifierMap#field_233802_a_` (the map where the attributes (and their ModifieableInstance are stored) public and non-final.
Then, somewhere in the setup phase, one could copy that (immutable) map into a new (immutable) map, add the desired attributes and overwrite the original map.

The main issue with this, is the concurrency between mods (in the loading phase). If
```
mod1->copyOrginalMap
mod2->copyOriginalMap
mod1->writeNewMap
mod2->writeNewMap
```
The additions of mod1 will be ignored.
There aren't any non-parallel events at the end of the loading phase, right?

LexManos:
We don't want to add yet another event fired during every entity construction.
And making the map public non-final is a obviously bad idea.
We would have to have a end bake step for the whole registry system that rebuilds/merges these default maps.

ChampionAsh5357:
[...]

PR #7475, opened 2020-11-09
Entity attribute setup

Kovak:
This replaces #7420 because I messed up my rebase

The goal of this PR is to add an event where it is safe to modify LivingEntity attribute maps after registration and prior to the rest of mod loading.

At the moment it is not possible to introduce new attributes to existing entities in 1.16.x, a regression from previous versions (#6928).

After thinking about the pitfalls of trying to hook something into the inheritance chain of static initializers of these maps, I decided it might be simpler to wait until after all registration has taken place and then just copy all the attribute maps, make them mutable again, issue the EntityAttributeSetupEvent for other mods to modify those maps, and then refreeze them afterwards.


pupnewfster:
Didn't really look at/review this, but one thing I noticed when clicking it out of curiosity is that the codestyle for patches is wrong. They should match the same as the rest of the file, aka opening brackets on the same line, instead of the next line

Kovak:
Oh no problem I can fix that up

Kovak:
Hello uhh whoever is actually reading this. It has been 8 weeks since I first submitted this. I'm happy to make whatever changes are necessary to get this merged but I need some guidance on what those changes would look like. 

I'm currently stuck on 1.15 for my mod development because I absolutely need to be able to add attributes to move forward. At this rate forge 1.17 will be out before we get this regression addressed.

LexManos:
It isn't a regression, you can create your own attributes just fine we've had those hooks for ages.
it is completely possible to create new attributes for existing entities.
The FORGE_ATTRIBUTES take precedence over the vanilla ones. So anyone at any time can update it to be whatever they want.
Now, this is a bad design because anyone, at any time, can update it to be whatever they want. So a new system would be good.
However, yours currently is not any better then the existing one. Except for the fact that you have a 'do this exact thing right now' style event {Like our Register events}. Which I like because it tells modders the exact time to do it.

However, I just have to say it your implementation is horrid.
You're using the syncExecutor, which is good, but you're using a static field to pass the event map over to the post processor which... has an argument for the exact same event.. so you could just get it there...
You're using a static field to make sure all the mods get the same event instance, which means this can never be fired again which defeats the point we're trying to make with the lifecycle stuff. And it's also unneeded.

Your getMap function in AttributeModifierMap is unneeded, it's a inner class, you could just reference the field directly.
The other option would be to add a AttributeModifierMap.toMutable() but that would expose API that isn't meant for modders. But you can negate that by having a 'for internal use only' telling modders they shouldn't use it.


As for your changes to ForgeHooks:
1) We don't use wildcard imports
2) You unnecessarily changed the braces for fixNullStructureReferences
3) You are using the incorrect brace style for your new methods
4) Your copyAttributesToForgeMap/replaceForgeAttributes/postGenerateEntityAttributeEvent are unnecessary. You could just do `e.getMap().forEach(GlobalEntityTypeAttributes::put)` 
5) getCombinedAttributeModifierMap isn't so bad except A) it could be done in the event constructor, B) `map = Stream.concat(forge.entrySet().stream(), vanilla.entrySet().stream()).collect(Collectors.toMap(Entry::getKey, e -> e.getValue().toMutable(), (a,b) -> {});`


And the event itself. I am not a fan of just exposing the mutable map itself. The MutableAttribute is fine, as it just has an adder. But exposing the map allows for removes which isn't good That'd end up with things like Creepers not having any attributes at all.
So that needs to be changed to get(EntityType).createMutableAttribute(blah). So that we know it's only additive. {Side note createMutableAttribute is a horrible name, whoever named it that in the bot need to be smacked :P, it should just be 'add'}

Now, as for how these things could be enhanced.
The very basic one is that EntityType itself needs to have some method of `is this entity a *whatever*` So that we can get the inheritance effect like we used to have.
So you'd have to add a TypeToken to EntityType so that we could do stuff like ```for (EntityType<?> type : event.getTypes()) {
  if (type.getInstanceClass().instanceOf(BeeEntity.class))
    event.get(type).add(KNEES)
}```

Or, if you want to get stupid complex... attempt to reverse the inheritance list and then rebuild the merged lists at the end. So that we could just do `event.get(Types.BEE).add(KNEES)` and it'll populate it to any modded Bees... but I think that would be a complex system that may not be feasible.

Overall, your initial premise that you can't modify vanilla attributes is false.
But the idea that we need a better system is correct. Our current one is a hack to hold people over until someone cared.
I think i'll blame the length of this PR on pup being new and figuring things out. And nobody else in the community caring about it, as noted by the lack of anyone discussing it.

So ya, give it another pass, properly @Deprecate the current system we have for it {Global:put/FORGE_ATTRIBUTES} so we can remove that in 1.17, and then poke one of the triage team to do another pass of reviewing.

Kovak:
Thanks for taking the time to respond. I'll come back soon with that solution.

Kovak:
I've reorganized this PR to provide an api that only allows the addition of new attributes. I've encapsulated the state in at least a defined class with better cleanup this time around. I couldn't see anyway to encapsulate state similar to this until the post processing callback without some type of container outside of the loading stage. Is there some way to just say "give me all the events that were just dispatched in the postDispatchHook"? I feel like you referenced this above but I'm not seeing how it could be achieved.

Re the TypeToken / type erasure problem on EntityType. I looked into this a little bit and it seems to me like it is impossible to solve this problem without introducing some kind of change to every EntityType builder call. TypeToken's can only properly capture the type via anonymous class at the same physical time as the type is passed in. Otherwise it just captures T. 

This would have to look something like 

```java
register("test", new EntityType.Builder<ZombieEntity>(){});
```

At that point it seems like it'd be just as profitable and significantly less arcane to require the caller of Builder to submit a concrete class.

I'd like to propose the 2 issues should be solved separately instead of together. At this point in time I think a significant amount of use cases for attribute addition can be solved via this pr. At the moment a solution without the physical type information is still quite valuable. While it is possible to add attributes for an entity type at the moment, it is not possible to access the maps underlying the existing attributes without an access transformer. This makes it pretty messy to assemble a new attribute map containing all of the previous attributes plus whichever ones you want to add.

Kovak:
Any triage team wanna take a look at this?

Cyborgmas:
Not triage but I have encountered this problem as well and would like a solution as well.
However, I have a hard time following why all of this is necessary, the way I see it only a hook in `#getAttributesForEntity` is needed that will post an event that exposes the `AttributeModifierMap.MutableAttribute` by copying the already baked vanilla one as you did and then caches the result so that its only posted once for each type. (or just call it at some point during initialization for all entity types, in the end the attributes are only needed once the world is created which is pretty late). 
This seems complicated to have this run as a modbus event, unless i am missing something important?

Edit: did not see the issue with inheritance, my bad for skimming. However it can be solved easily since `getAttributesForEntity` is only called in one place, the `LivingEntity` constructor, which has inheritance data !

Kovak:
They are specifically wanting a solution that does not post per entity constructed and instead occurs as part of registration.

Cyborgmas:
so 1) i am not saying per entity constructed but just per entity type, post the event only on the first time this entity is constructed and then cache it. 
2) why does it need to be part of registration using a modbus event ? Why can it not be a normal forge bus event posted after all all modlifecycle is complete, because its only needed once a world is created. I am not against it being an event that does everything at once, just what is the reason behind putting it in the mod lifecycle?

Kovak:
In my opinion, we want entity attribute modification to occur as part of the registration phase of mod lifecycle as the entity attributes are effectively supposed to be frozen similar to the way literally every other asset is frozen during the mod lifecycle. This way by the time you get past the registry phase and into the various other mod lifecycle stuff it is possible to query the finalized attribute maps to make decisions.

With what you are suggesting you would literally have to wait until the first time an entity is spawned in a world in order to make such decisions. My question for you is why does that seem acceptable? Essentially you could not be certain of the final state of an entity's attributes until after it had been encountered for the first time every run of the game. 

It is both more logical to perform this step after registration but before any other phase in mod loading and more consistent. Your suggestion would still involve every single time you run the game and encounter a new entity unfreezing the attribute map for that entity, dispatching the attribute event, and then refreezing. It just doesn't seem necessary when you could just cleanly and orderly do it once as part of mod loading.

Cyborgmas:
> I am not against it being an event that does everything at once

Yes I also suggested to do it when the entity is first created but I can get why that is not wanted, even if we can get inheritance. 
However it still does not explain why it is done as a ModLoadingStage ? Why not do something similar to the ParticleFactoryRegistryEvent, still on the modbus but on one thread (forge bus is shutdown until after modloading). Would make your PR a lot smaller and the logic much less complicated.

My main issue is just that this seems like a lot for what in my mind is very little, but perhaps I am wrong and this is fine, in the end this does allow for adding the attributes.

Kovak:
Like I said previously: So that the actual finalized attribute maps are available in time for every mod bus event after the registration phase. Some of this is based off prior discussion on the associated issue (#6928). There is logically a missing step in-between registration right now and the rest of the mod life cycle that needs to allow for data modification like this. 

The final post in that issue thread is from Lex and says:

> We don't want to add yet another event fired during every entity construction.
> And making the map public non-final is a obviously bad idea.
> We would have to have a end bake step for the whole registry system that rebuilds/merges these default maps.

So the reason its an end bake step for the whole registry system that rebuilds/merges the maps in between registration and the rest of the mod lifecycle is because it was explicitly said that this is the desirable solution.

Cyborgmas:
Ah I completely agree with a registry end bake step, but then make it generic and extendable perhaps ? not just for attributes.

And even if it's unecessary what I meant by `doing something similar to ParticleFactoryRegistryEvent` is that you can post the event to all mod containers whenever. In this case in the finalActivityGenerator of the LOAD_REGISTRIES stage, so that it is done before common setup but without adding an extra stage.

Kovak:
Why not make it generic and more extendable? Because I'm having enough trouble getting anything reviewed by forge the last thing I'm going to do is extend this to encompass random things that no one has asked for and that we currently have no need for and thus make it even less likely to be reviewed.

Re: why is it a stage? Because it seemed cleaner and more explicit to me. Perhaps someone from the actual triage team could weigh in here. The introduction of a new stage was one of the few things Lex did not criticize at all so I'm inclined to believe that it is an acceptable approach. I appreciate you spit-balling with me here, but it also doesn't seem like you are really aware of the background discussion around this issue nor the shape of the desired solution.

Kovak:
Anyone? Been another month of this sitting here.

patches/minecraft/net/minecraft/entity/ai/attributes/GlobalEntityTypeAttributes.java.patch
```diff
@@ -1,15 +1,23 @@
...
 +   public static AttributeModifierMap put(EntityType<? extends LivingEntity> type, AttributeModifierMap map) {
 +       return FORGE_ATTRIBUTES.put(type, map);
 +   }
- 
++
++   public static AttributeModifierMap combineWithExisting(EntityType<? extends LivingEntity> livingEntity, AttributeModifierMap.MutableAttribute toCombine) {
++      AttributeModifierMap modifier = func_233835_a_(livingEntity);
++      AttributeModifierMap.MutableAttribute newMutable = modifier != null ? new AttributeModifierMap.MutableAttribute(modifier) : new AttributeModifierMap.MutableAttribute();
++      newMutable.combine(toCombine);
++      return FORGE_ATTRIBUTES.put(livingEntity, newMutable.func_233813_a_());
```
LexManos: Again, we don't want to add yet more random 'just shove things here at anytime' api. Deprecated the old put and put no replacement.

src/main/java/net/minecraftforge/event/entity/EntityAttributeSetupEvent.java
```diff
@@ -0,0 +1,56 @@
+package net.minecraftforge.event.entity;
```
LexManos: Header

src/main/java/net/minecraftforge/event/entity/EntityAttributeSetupEvent.java
```diff
@@ -0,0 +1,56 @@
...
+                (entityType) -> new AttributeModifierMap.MutableAttribute());
+        attributes.createMutableAttribute(attribute, value);
+    }
+
+    public void addAttributeToEntityType(EntityType<? extends LivingEntity> entity, Attribute attribute)
+    {
+        addAttributeToEntityType(entity, attribute, attribute.getDefaultValue());
+    }
+
+    public List<EntityType<? extends LivingEntity>> getTypes()
```
LexManos: Why is this needed? Consumers can have their own check if there is an existing one. Rebuilding this every call because people could add things between calls seems wasteful.

src/main/java/net/minecraftforge/event/entity/EntityAttributeSetupEvent.java
```diff
@@ -0,0 +1,56 @@
...
+{
+    private final Map<EntityType<? extends LivingEntity>, AttributeModifierMap.MutableAttribute> entityAttributes;
+
+    public EntityAttributeSetupEvent(ModContainer modContainer)
+    {
+        this.entityAttributes = new HashMap<>();
+    }
+
+    public boolean doesEntityTypeHaveAttribute(EntityType<? extends LivingEntity> entity, Attribute attribute)
+    {
```
LexManos: You really need to learn about letting arguments describe the method instead of adding a bunch of really long names.
And, why do these one line helper functions even exists? 
Tho, I could see if trying to use the map from other event handlers. That would be reasonable, But right now you just ask the Global so it's only half answering the question.

src/main/java/net/minecraftforge/event/entity/EntityAttributeSetupEvent.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class EntityAttributeSetupEvent extends Event implements IModBusEvent
+{
+    private final Map<EntityType<? extends LivingEntity>, AttributeModifierMap.MutableAttribute> entityAttributes;
+
+    public EntityAttributeSetupEvent(ModContainer modContainer)
+    {
+        this.entityAttributes = new HashMap<>();
```
LexManos: Cleaner code is to put simple field initializers in the field declaration.

src/main/java/net/minecraftforge/event/entity/EntityAttributeSetupEvent.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public boolean doesEntityTypeHaveAttribute(EntityType<? extends LivingEntity> entity, Attribute attribute)
+    {
+        return GlobalEntityTypeAttributes.getAttributesForEntity(entity).hasAttribute(attribute);
+    }
+
+    public void addAttributeToEntityType(EntityType<? extends LivingEntity> entity, Attribute attribute, double value)
```
LexManos: add

src/main/java/net/minecraftforge/event/entity/EntityAttributeSetupEvent.java
```diff
@@ -0,0 +1,56 @@
...
+    {
+        AttributeModifierMap.MutableAttribute attributes = entityAttributes.computeIfAbsent(entity,
+                (entityType) -> new AttributeModifierMap.MutableAttribute());
+        attributes.createMutableAttribute(attribute, value);
+    }
+
+    public void addAttributeToEntityType(EntityType<? extends LivingEntity> entity, Attribute attribute)
```
LexManos: add

src/main/java/net/minecraftforge/event/entity/EntityAttributeSetupEvent.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public List<EntityType<? extends LivingEntity>> getTypes()
+    {
+        return ForgeRegistries.ENTITIES.getValues().stream()
+                .filter(GlobalEntityTypeAttributes::doesEntityHaveAttributes)
+                .map(entityType -> (EntityType<? extends LivingEntity>) entityType)
+                .collect(Collectors.toList());
+    }
+
+    public Map<EntityType<? extends LivingEntity>, AttributeModifierMap.MutableAttribute> getEntityAttributes()
```
LexManos: Exposing the full mutable map defeating everything I said about not doing that.

src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -964,4 +996,26 @@ private static void forceRegistryName(IForgeRegistryEntry<?> entry, ResourceLoca
         }
 
     }
+
+    private static class AttributeSetupState
```
LexManos: This whole class is a no, just moving something I told you was bad to a different location doesn't make it any less bad.

The shared map should be held on the event object itself. That's it. 
We'll have to ask @cpw how he would like to handle something like this. 

Edit: OR What you could do is not make the contents of this class static.
```
public static class AttributeEventManager implements IStatefulModLoadingStage {
  private Map<EntityType<?>, AMM> attribs = copyVanillaAttributes();
  public Stream<EventGenerator<>> getGernerator() {
    return mc -> new Event(attribs)
  }
  public Future pre(...){}
  public Future post(...){}
}```
That would need a new ModLoadingState constructor like `ModLoadingStage(IStateful s){ this(stateful.getGen(), s::pre, s::post)}`

**Review by LexManos — CHANGES_REQUESTED** [review]
Needs work to address comments from my previous review, and this one.

File: src/main/java/net/minecraftforge/registries/GameData.java (hunk 9 above)
Comment (Kovak) [reply src/main/java/net/minecraftforge/registries/GameData.java]: Currently the way the loading stages work it does not seem possible to hold the state on the event itself without functionally doing the equivalent of what I am doing here. The suppliers are re-invoked for each phase of the stage. This works ok in the one place where it's really used because the registries hold the state instead. To be honest I was somewhat surprised that the post stage did not receive the same stream of events as passed into the stage itself. 

I would really appreciate some more guidance on what ya'll want that to look like because the functionality does not exist right now. The only reason I am doing the "thing that is bad" is because there is no other way to achieve that thing right now and it seemed even uglier to do some kind of supplier that stores the state inside itself (and would mostly be logically the exact same thing).

File: patches/minecraft/net/minecraft/entity/ai/attributes/GlobalEntityTypeAttributes.java.patch (hunk 1 above)
Comment (Kovak) [reply patches/minecraft/net/minecraft/entity/ai/attributes/GlobalEntityTypeAttributes.java.patch]: This is only supposed to be for internal forge usage. With the external put deprecated and FORGE_ATTRIBUTES being private there has to be some way for Forge package to modify the FORGE_ATTRIBUTES map.

File: src/main/java/net/minecraftforge/event/entity/EntityAttributeSetupEvent.java (hunk 3 above)
Comment (Kovak) [reply src/main/java/net/minecraftforge/event/entity/EntityAttributeSetupEvent.java]: I put in getTypes to help modders not fuck things up. ForgeRegistries contains EntityType<? extends Entity> but you can only add attributes to entities that extend LivingEntity. imo it was just here to make it easier/nicer for modders to add their attribute to every living entity without having to do the selection and casting themselves. I didn't cache it cause I didn't really think much of it as it's only ever gonna run a handful of times at initialization. Would you to prefer to not have it and leave people up to selecting the appropriate entity types themselves, or have it here but have it be less wasteful?

File: src/main/java/net/minecraftforge/event/entity/EntityAttributeSetupEvent.java (hunk 8 above)
Comment (Kovak) [reply src/main/java/net/minecraftforge/event/entity/EntityAttributeSetupEvent.java]: This is not the full mutable map, this is only the map per mod of the attributes they are adding. It is exposed so that the post-event consumer can then take all the various mod's additions and reassemble them into the new attribute maps. It's not really that useful for the users of the event themselves, although they cannot really do any damage with it.

To my knowledge because of the package structure there is not a way to mark it as package private and have GameData be able to access this map when it needs to.

**Comment by cpw** [comment]
> Re: why is it a stage? Because it seemed cleaner and more explicit to me. Perhaps someone from the actual triage team could weigh in here.

Well, that's a terrible idea. Don't use a stage just for this. whatever it is. Is it a global state all mods have to pass through? HELL NO. WTF is it doing as a stage...

There's lots of other smaller events that mods can opt into if they so care. Do it that same way. Others have explicitly said to do that (@Cyborgmas).

**Comment by Kovak** [comment]
My bad, I didn't quite realize it was safe to emit events from inside the GameData#checkForRevertToVanilla, I thought that mod loading stages were the direction to go in since they already have some fairly simple stages like the CommonSetupEvent stage. Here is another go at things that is a lot more direct.

I did do one thing as I ran into some theoretical issues when working on some recent features. At the moment there is no defined place to put your entity type's attributes and mod developers are sort of doing it all over, with the most common being either directly in the common setup event or in a deferred work queue scheduled then. If your goal is to modify existing maps this leaves you in a tough spot on determining when the right time is to get all mods maps. I have split the EntityAttributeSetupEvent into two separate events, EntityAttributeCreationEvent : where a mod should be adding attributes for its own entity types and EntityAttributeModificationEvent, where additional attributes can be added to existing entity maps. Without a clearly defined stage for putting the initial maps we won't be able to get an accurate sense of which entity types should be in the getTypes() list

EntityAttributeCreationEvent is a pretty thin wrapper over the old put api but it does throw an exception if you try to replace a map. EntityAttributeModificationEvent now has a full implementation of the has function and caches the entity types at the start so there's less computation cost there.

At the moment I deprecated the GlobalEntityTypeAttributes#put and GlobalEntityTypeAttributes#combineExisting functions to try to signal for modders not to call them, but is there better way for signaling forge internal use?

**Comment by gigaherz** [comment]
No, java doesn't have any way to indicate "for internal use only". Their solution to that was java9 modules: if something is internal, you don't export it from your module. However we can't make use of the module system due to needing to maintain java8 compatibility for the vanilla bundled jre.

**Comment by Kovak** [comment]
Well it wouldn't be a java thing anyway it'd be a forge thing. Because Forge and Minecraft are separate pacakges even in a java9 context it wouldn't work to not export it. It's why the put and combineExisting functions need to be public at all is that forge needs to be able to call them itself (otherwise FORGE_ATTRIBUTES is just a private map with no way to access it), however we need to discourage others from calling them and to go through the events instead.

File: patches/minecraft/net/minecraft/entity/ai/attributes/GlobalEntityTypeAttributes.java.patch (hunk 1 above)
Comment (LexManos) [reply patches/minecraft/net/minecraft/entity/ai/attributes/GlobalEntityTypeAttributes.java.patch]: There is...
The put method.
Its deprecated, and shouldn't be used by others, but for the time being its there and adding yet another PUT with even more logic is just asking for more headakes.

File: src/main/java/net/minecraftforge/event/entity/EntityAttributeModificationEvent.java [hunk 10]
```diff
@@ -0,0 +1,79 @@
...
+    private final Map<EntityType<? extends LivingEntity>, AttributeModifierMap.MutableAttribute> entityAttributes;
+    private final List<EntityType<? extends LivingEntity>> entityTypes;
+
+
+    public EntityAttributeModificationEvent(Map<EntityType<? extends LivingEntity>, AttributeModifierMap.MutableAttribute> mapIn)
+    {
+        this.entityAttributes = mapIn;
+        this.entityTypes = ForgeRegistries.ENTITIES.getValues().stream()
+                .filter(GlobalEntityTypeAttributes::doesEntityHaveAttributes)
+                .map(entityType -> (EntityType<? extends LivingEntity>) entityType)
```
LexManos, line 55: Again, this filter and list is wrong and should flat out be deleted.
The list becomes immediately outdated as soon as someone calls an add function.

src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -386,6 +392,10 @@ public static void revert(RegistryManager state, ResourceLocation registry, bool
                 revertTo(RegistryManager.VANILLA, false);
                 LOGGER.fatal("Detected errors during registry event dispatch, roll back to VANILLA complete");
             }
+            ModLoader.get().postEvent(new EntityAttributeCreationEvent());
```
LexManos: Why the hell did you pick this place to throw the events?
1) This is for checking for errors and reverying to vanilla if there is any
2) Why are you specifically adding 'lets make things not vanilla' in a method to revert to vanilla... without checking if you should revert to vanilla...
At the very lest this needs to be in a else statement.
But better would be to pull the entire thing out of the lifecycle events.

Kovak:
What else needs to be done for this? Is this new approach more acceptable?

Kovak:
...

Kovak:
Been 33 days since my last revision of this PR. Obviously there are still some changes ya'll want but you aren't saying? Kind of hard to divine intent from utter silence

marchermans:
There are still in unaddressed review comments open.

Either comment on them why you think the reviewer is wrong, or create a comment to fix them.

Kovak:
I did that and there was literally never a response?

and this particular point keeps being harped on without a satisfactory answer to my response throughout this thread. Ya'll have a private field in the minecraft package (FORGE_ATTRIBUTES). For the forge package to modify that private field there needs to be some type of api, in this case put and combineExisting. There is no way for the forge package to modify FORGE_ATTRIBUTES without such an api. If the desire is simply to move even more logic out of the minecraft package it can be done although in this case we're making use of package private access to the internals of the AttributeModifierMap and I think its cleaner to do it like this. 

Now I'm not a java wizard so maybe there is some thing I am unaware of that can overcome this particular limitation, but I'm fairly certain there is not. There has to be some public way to modify FORGE_ATTRIBUTES that we want to use from the forge package but don't really want available to modders but since they're separate packages there is only one way to do that and it will make them available to anyone and so the best we can do is mark them with some type of annotation (which I have defaulted to deprecated since there does not seem to be a better option).

LexManos (changes requested):
So.. apparently github errored in the submitting of this. No wonder it was stagnant for a while.

src/main/java/net/minecraftforge/event/entity/EntityAttributeModificationEvent.java
Kovak: You are incorrect here. Because of the split between EntityAttributeModificationEvent and EntityAttributeCreationEvent this list will not become outdated, as all entities with attributes will have had them added during EntityAttributeCreationEvent. This is partly why the 2 separate events exist at all. By the time this is called, every single map that will ever be added for entities will have been added at this point. The only thing we are doing is modifying the already existing ones in this event. There is no add function to call at this point. Only modify.

src/main/java/net/minecraftforge/registries/GameData.java
Kovak: You know it is so incredibly typical of you guys that I literally predicted this exact comment the day I submitted this PR 34 days ago.

So let us review.

1. I introduced a new mod loading stage between registration and FMLCommonSetup event as that is the proper time to complete entity attribute operations.
2. Ya'll hate that solution and LITERALLY tell me to do it in the post stage callback for the registration stage. Read up a few posts and you will see that being suggested in here as the alternative. Do you know how hard it is to get anything done up to ya'lls standards when those standards are a shifting mass of bad suggestions in between the good with no indication as to which you believe to be which, and with almost no communication at all when questions are asked?

SO WHICH IS IT

1. It is ok to do it in the post stage callback which is currently "revert to vanilla"

2. Do it in a mod loading stage that you have adamantly insisted is not correct.

As I have mentioned previously in this very thread about this feature, it is ideal for us to have this done before FMLCommonSetup because this entire entity attribute setup bit is basically part of the registration phase. One of the whole points of having this entity attribute api is to that by the time we get to FMLCommonSetup we will have a full picture of what entity attributes look like.

WHERE ELSE WOULD YOU SUGGEST IT GO TO FULFILL THAT?

src/main/java/net/minecraftforge/event/entity/EntityAttributeModificationEvent.java
LexManos: That you are actually correct about, I guess when I did the first review I misread the code.
doesEntityHaveAttributes is just a rather poor function that I misunderstood.

patches/minecraft/net/minecraft/entity/ai/attributes/GlobalEntityTypeAttributes.java.patch
```diff
@@ -1,15 +1,24 @@
...
 +   }
- 
++
++   @Deprecated // FORGE: Use net.minecraftforge.event.entity.EntityAttributeModificationEvent#add
++   public static AttributeModifierMap combineWithExisting(EntityType<? extends LivingEntity> livingEntity, AttributeModifierMap.MutableAttribute toCombine) {
++      AttributeModifierMap modifier = func_233835_a_(livingEntity);
++      AttributeModifierMap.MutableAttribute newMutable = modifier != null ? new AttributeModifierMap.MutableAttribute(modifier) : new AttributeModifierMap.MutableAttribute();
++      newMutable.combine(toCombine);
++      return FORGE_ATTRIBUTES.put(livingEntity, newMutable.func_233813_a_());
++   }
```
LexManos, line 10: Move this to ForgeHooks with a major *INTERNAL ONLY* warning.
Move FORGE_ATTRIBUTES to ForgeHooks as well, it's private it can change without us caring about breaking things.
Bounce put to a ForgeHooks function that logs a error in the console about deprecated call {Make it a small one liner because people will complain, but it'll just dissapear in 1.17}

src/main/java/net/minecraftforge/registries/GameData.java
LexManos: You can pick anywhere at all in the vanilla code to throw a fucking hook.
The lifecycle shit is a mess, it's cpw's baby I don't like touching it, so ask him.

But Hell You could even add a fucking 'if we are not reverting to vanilla because shit is broken, then lets do things' and it would be better.
Right now you're "Hey I know there is an error and we LITERALLY just reverted everything to vanilla to try and gracefully handle it... But lets modify more shit!"
Look, Vanilla minecraft is complete shit in this regard. The ONLY sane way to do this is to re-write the entire god damn thing so that vanilla isn't a mess of static maps and functions.
But that isn't going to happen. So at least apply some common sense and logical thinking about the code you're working near.

Kovak: I did originally think about this and my original thought was so we're reverting our registries but does that mean that this attribute modification step doesn't occur at all? All this revert does is return the registries to their original state, as far as I'm aware mods continue to execute otherwise. It seemed to me that doing it this way preserved the expected execution as much as possible.  

What if you wanted to add vanilla attributes to vanilla entities? That would still be entirely possible even after we revert the registries to vanilla, which is why I didn't make it conditional on not reverting originally. 

Obviously (now) it is your opinion that the opposite choice should be made and that this should not execute if we revert, and I'll get to putting together a solution that looks like that, but you must admit that the route I took is not fucking illogical it just is a different interpretation of what reverting to vanilla implies for this post-registration step of linking attributes to entity types. 

There's not really anywhere in "vanilla" to throw this fucking hook because it needs to be thrown after registration and before FMLCommonSetup. It needs to be part of the lifecycle shit. I didn't just come in here fucking willy-nilly not paying attention to shit. I read through (and understood even though I agree it is a mess) ya'lls lifecycle shit and tried to make sense of how it should be extended in this case. Just because ya'll dont like your own shit doesn't mean you need to take it out on me, a random guy just trying to contribute some sensible extensions to the lifecycle so that everyone doesn't have to rely on the much flakier AT and delayed execution approach that is necessary at the moment. Which by the way is what ya'll keep begging people to do instead of just coremod and AT shit to all hell.

TheCurle: The point of the revert to vanilla is to obtain a known good state to continue falling through the error path.
It's the only sane way to handle an error deep in the code without hard crashing.

After this, the event busses are shut down and mods stop running.

Adding further modifications to this known-good state defeats the point, and might, if a mod were coded badly enough, cause the game to hard-crash anyway.

The best option is to else this if statement so that it runs if, and only if, the game is still in a valid state.

TheCurle: To further that point, this could be literally one extra line and it'd be a practically suitable solution for this problem.
A single `else {`. That's all it needs. 

And this solution has been suggested no less than three separate times in this single thread.

Kovak: Thanks for being explicit about the intent of revert to vanilla curle. From just reading the lifecycle code it was not obvious what the final result should be as at this particular time in execution event busses are still active, mods are not unloaded, and theoretically it could still execute on only vanilla registry data. Surely ya'll can understand that from an outsider just reading the code and trying to piece together the various opinions dropped in this PR and associated issue perspective this was a valid decision to be made (albeit the unwanted one)? 

 Also I have acknowledged that I will be making this change. I'm not fighting for it not to be made. I am simply pointing out that I am not a fucking idiot and that it wasn't entirely clear which would be the better option.

The fundamental issue with this PR is that it involves doing something that currently does not have any equivalent. Mainly post registration modification of other global state. It is obvious that no one knew (or at least if they did know they did not bother to share) how they wanted this to look prior to me starting this PR and that feedback has been very flippant (and sometimes outright incorrect to the point of it being obvious that no one has bothered actually reading the relevant parts) with very little in the way of answering questions of clarification throughout the process. 

Look: I love forge, I love ya'll for developing and maintaining it. That's why I came here to try to contribute to an area of the code that was clearly being neglected. I would just encourage ya'll to read through this thread, make note of exactly how long the gaps are between responses (over 30 days on average, I originally opened this PR 4 months ago in October and literally multiple times have been given apologies of "sorry we should have gotten to this sooner/github bugged didnt see it"), exactly how many times a question just gets no response at all, exactly how many times there has been a given point of feedback that is literally entirely incorrect, and then exactly how many times needless aggression has been lobbied towards me for making decisions that have a non-obvious resolution. Also if it is not obvious enough, meeting ya'lls aggression with aggression has produced 100% better results than simply asking nicely and hoping someone will engage in good faith.

Kovak:
sorry everyone i fucked up the rebase again because of all the 1.16.5 changes and how long this has been sitting here. I'm going to close this one and open another clean PR in a bit.

PR #7479, opened 2020-11-11
Fix vanilla hardcoded sea level in World

alcatrazEscapee:
`World#getSeaLevel()` (`func_181545_F`) is hardcoded to return `63`. This method is called a lot within the vanilla code base, although only on logical server. This is also impossible for mods to override as it's a method in `World` if they want to change the sea level (me!).

However, mods can already change the sea level reported by `ChunkGenerator` subclasses via the `getSeaLevel` method there. And the `ServerWorld` has a reference to the chunk generator. So this just delegates the world's sea level to it's chunk generator. Horray!

### DIFF
--- a/patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
+++ b/patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
@@ -214,7 +214,7 @@
        explosion.func_77278_a();
        explosion.func_77279_a(false);
        if (p_230546_12_ == Explosion.Mode.NONE) {
-@@ -1410,4 +1446,14 @@
+@@ -1410,4 +1446,19 @@
           p_241121_0_.func_175656_a(p_241122_1_, Blocks.field_150343_Z.func_176223_P());
        });
     }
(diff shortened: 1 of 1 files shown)


50ap5ud5 (approved):

ichttt:
The problem with this is that `getSeaLevel` is a method defined in `World`, so when a `getSeaLevel` is called on the client, it still returns the default value. It should not be an issue with vanilla, but it could cause issues with mods

alcatrazEscapee:
While yes, mods could technically be calling `getSeaLevel` and receiving incorrect values, the client fundamentally does not know about the sea level outside of this hardcoding - the chunk generator does not exist / is synced to client. At the moment, Both `ChunkGenerator` and `World` have a hardcoded return value of 63. One of them mods can override.

If there's another better option I'm all ears. This brings the behavior closer to being consistent on servers (important for world gen). If it's not acceptable / too hardcoded in vanilla for this kind of solution I understand that.

ichttt:
You are semi-correct, I guess we already have an inconsistency because of this and this PR is the best we can do for now. But `NoiseChunkGenerator` overrides `getSeaLevel`, and both The End and The Nether have different values set for sea level in this case (32 for nether, 0 for end), so we would have to check that this doesn't break anything there.

alcatrazEscapee:
Oh damn, I didn't realize that. That would change vanilla behavior unfortunately 😢

PR #7482, opened 2020-11-12
Fix forge registry types that have private constructors (BlockStateProviderType, BlockPlacerType, FoliagePlacerType, TreeDecoratorType)

alcatrazEscapee:
This is pretty simple. `BlockStateProviderType` is a forge registry type with a private constructor. There exists a register method but it directly invokes `Registry.register()` (perish the thought). So to help every mod that wants to make a new `BlockStateProvider` (probably exactly one but ignore that *world gen is a mess sometimes okay?*), let's just AT this in Forge shall we?


ichttt (commented):
Sounds like it makes sense

alcatrazEscapee:
So upon further inspection I have found three other identical cases - forge registry entries, all for obscure world gen purposes, all with private constructors that are only invoked via a direct call to `Registry.register`. I have added them to this PR as they really are the same issue.

PR #7483, opened 2020-11-12
Fix concurrency issue in StartupMessageManager

### LINKED ISSUE #7474: ConcurrentModificationException when making call to StartupMessageManager#addModMessage
**Minecraft Version:** 1.15.2

**Forge Version:** 31.2.45

**Logs:** https://gist.github.com/Sm0keySa1m0n/50b6303047d5a9286dc70677101637e1

**Steps to Reproduce:**
Seems to be a race condition, can't reliably reproduce.

**Description of issue:**
During `FMLClientSetupEvent`, making a call to `StartupMessageManager#addModMessage` will sometimes produce a `ConcurrentModificationException`.

diesieben07:
Fixes #7474.

### DIFF
--- a/src/fmllauncher/java/net/minecraftforge/fml/loading/progress/StartupMessageManager.java
+++ b/src/fmllauncher/java/net/minecraftforge/fml/loading/progress/StartupMessageManager.java
@@ -32,7 +32,7 @@ import java.util.stream.Collectors;
 
 public class StartupMessageManager {
 
-    private static volatile Map<MessageType, List<Message>> messages = ImmutableMap.of();
+    private static volatile EnumMap<MessageType, List<Message>> messages = new EnumMap<>(MessageType.class);
 
     public static List<Pair<Integer,Message>> getMessages() {
         final long ts = System.nanoTime();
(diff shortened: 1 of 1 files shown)


src/fmllauncher/java/net/minecraftforge/fml/loading/progress/StartupMessageManager.java
```diff
@@ -88,22 +88,41 @@ long getTimestamp() {
...
+    {
+        Map<MessageType, List<Message>> messages = StartupMessageManager.messages;
+
+        List<Message> oldMessagesForType = messages.getOrDefault(type, ImmutableList.of());
+        List<Message> newMessagesForType = ImmutableList.<Message>builder()
+                .addAll(maxSize < 0 ? oldMessagesForType : oldMessagesForType.subList(0, Math.min(oldMessagesForType.size(), maxSize)))
+                .add(new Message(message, type))
+                .build();
+
+        ImmutableMap.Builder<MessageType, List<Message>> builder = ImmutableMap.builder();
```
ichttt: I would suggest keeping this an EnumMap, so we keep the enum-optimized behaviour. The map is not visible outside of this class anyway, so we don't have to worry about people misusing it.

PR #7369, opened 2020-09-27
Workaround for modded attributes causing warning on vanilla clients

diesieben07:
When modded attributes are added to vanilla entities those cause warning logs on vanilla clients every time they receive the entity attributes.
This is a possible workaround.

Fixes #7368.

PR #7485, opened 2020-11-12
[1.16] Vanilla packet splitting

### LINKED ISSUE #6703: `SRecipeBookPacket` needs patch to allow huge number of recipes
**Minecraft Version:** 1.15.2 (1.14.4 probably, too, but not tested there)

**Forge Version:** 31.2.0 and earlier

**Logs:** n/a, this issue is not visible in the logs.
image png 7b47f08be4cafdcec9962da5083f9a89

When there are too many recipes, `SRecipeBookPacket` gets too large for the network protocol and fails with the above error. Some kind of workaround patch is needed.

diesieben07:
This pull request is a workaround for the size limitations of the vanilla protocol, which causes issues when a huge number of recipes or tags is present.
To work around this, the system added in #7369 is enhanced to also act on forge<=>forge connections.
If it is detected that a packet is too large, it is split and sent as a series of custom payload packets instead.

This fixes #6703.


src/main/java/net/minecraftforge/network/VanillaPacketSplitter.java
```diff
@@ -0,0 +1,167 @@
...
+    private static boolean heuristicIsDefinitelySmallEnough(IPacket<?> packet)
+    {
+        // todo: improve this
+        return false;
+    }
+
+    private static final List<PacketBuffer> receivedBuffers = new ArrayList<>();
+
+    @SuppressWarnings("unchecked")
+    private static void onClientPacket(NetworkEvent.ClientCustomPayloadEvent event)
```
pau101: This event must be `ServerCustomPayloadEvent` to receive clientbound packets.

src/main/java/net/minecraftforge/network/VanillaPacketSplitter.java
```diff
@@ -0,0 +1,167 @@
...
+        ProtocolType protocol = ProtocolType.PLAY;
+
+        PacketBuffer buf = event.getPayload();
+
+        byte state = buf.readByte();
+        if (state == STATE_FIRST)
+        {
+            receivedBuffers.clear();
+        }
+        receivedBuffers.add(buf);
```
pau101, line 156: This buffer is released after this packet is processed*, either a copy of the buffer should be created or a switch to writing buffer bytes to a single cumulation buffer should be made.

\*should be released, as after confirming this, it is not in fact due to a patch port error introducing a memory leak for what it was intended to fix, new issue report in progress.

src/main/java/net/minecraftforge/network/VanillaPacketSplitter.java
```diff
@@ -0,0 +1,167 @@
...
+            {
+                try
+                {
+                    packet.readPacketData(full);
+                }
+                catch (IOException e)
+                {
+                    throw new UncheckedIOException(e);
+                }
+                ctx.enqueueWork(() -> ((IPacket<ClientPlayNetHandler>)packet).processPacket(Minecraft.getInstance().getConnection()));
```
pau101, line 183: A call to `ctx.setPacketHandled(true)` is missing, meaning the log will get filled with "unknown custom packet."

diesieben07:
Thank you for the review @pau101. I have fixed things up (this is what I get for writing PRs half dead at midnight).

alcatrazEscapee:
> I would like ideas if `heuristicIsDefinitelySmallEnough` can be improved (aka implemented at all) to avoid allocating a buffer for the packet.

So operating under the basic assumptions:
- This must **never** falsely categorize a packet which is too large as one that is too small, as then we've just implemented a probabilistic bug fix (and that's kind of stupid).
- This then needs to be be simple enough that it's not just serializing the packet, and checking if the length is small enough.

What is desired here is an upper bound on the total serialized length of the packet.

### Recipes

Alright, so the recipes packet is fairly easy to understand that it's impossible to upper bound. Mods are given direct write access to the buffer in custom recipes, and can thus technically write any amount of data to the buffer. Thus it's provably impossible to come up with an upper bound without a definitive upper bound on the total size of an arbitrary modded recipe packet.

### Tags

This packet has a deterministic protocol, and as such, we *can* upper bound it with knowledge of a bit of metadata.

- `STagsListPacket` writes four `ITagCollectionSupplier` instances for each of Blocks, Items, Entity Types, and Fluids.
- Each `ITagCollectionSupplier` consists of one var int (5 bytes), plus it writes each tag individually. Assuming we know the total number of tags, we can bound the maximum length of a single tag entry:
- Each tag entry consists of a `ResourceLocation` (which is encoded as a string, and thus has a maximum size of 32767 bytes) + one var int for the number of entries in the tag (5 bytes), plus one var int (the ID) for each entry in the tag (5 bytes each)
- A `ITag<T>` is just a glorified `Set<T>`. It cannot contain duplicates or unregistered elements, and thus we can upper bound the total number of entries in a tag by the total number of elements in the registry.

So, using all that, a primitive upper bound for the number of elements in say, the Blocks tag collection becomes:
```java
// 5 + (size of blocks tag collection) * (32767 + 5 + (size of blocks registry) * 5)

// Or in code, for block tags
ITagCollectionSupplier tags;
int maxBytes = 5 + tags.getBlockTags().getIDTagMap().size() * (32767 + 5 + ForgeRegistries.BLOCKS.getKeys().size() * 5)
``` 

Now, depending on how accurate we want to get here there's a number of things we can do to reduce this upper bound. Most of them involve iterating through the block tags collection and computing an individual bound on each tag as the largest contributors to this bound are going to be 1) the `ResourceLocation` length, and 2) The number of tag entries.

```java
// 5 + (for each tag, (size of tag name) + 5 + (number of tag entries) * 5)

// Or in code, for block tags
int maxBytes = 5;
ITagCollectionSupplier tags;
for (Map.Entry<ResourceLocation, ITag<Block>> entry : tags.getBlockTags().getIDTagMap().entrySet())
{
    maxBytes += entry.getKey().toString().length() + 5 + 5 * entry.getValue().getAllElements().size();
}
```

And finally, now we have the upper bounds, the heuristic is just comparing the upper bound to the maximum allowed length as per the network protocol.

Now: is this useful? Here are the counts for vanilla of the required numbers (as best I can), and computing the upper bounds using the simple version:
- 797 Blocks and 86 Block Tags - 3,161,107 bytes
- 976 Items and 54 Item Tags - 2,033,213 bytes
- 108 Entity Types and 5 Entity Tags - 166,565 bytes
- 5 Fluids and 2 Fluid Tags - 65,589 bytes

Total upper bound size for vanilla: 5,426,474 bytes. This is well over the protocol max 2,097,152, so we can definitively say **the simple upper bound is useless as a heuristic**.

Now, I'm not going to calculate an exact value for the iterative (second) heuristic . However, we can roughly approximate it using "reasonable" values to see if it's worth doing. Assuming:

- A `ResourceLocation` is on average 30 characters long
- A tag contains on average 10 entries.

Then with the same data from above, we arrive at:

- Blocks: 7,315 bytes
- Items: 4,595 bytes
- Entities: 430 bytes
- Fluids: 175 bytes.

For a total upper bound of 12,515 bytes. Now this heuristic is *much* closer to the actual size of the packet in question (and can be actually used to see that the tag is small enough to avoid splitting. There are two things here:

- For any given tag collection, this value can be precomputed and re-used (if efficiency is desired), as it's only dependent on constant properties about the tag collection.
- At some point, the heuristic will be too complicated to be useful over just writing the tag itself. I'm not sure if the second (iterative) one already qualifies for that. But I spent time computing it, so here it is.

TL;DR: Math. :D

diesieben07:
Resolved merge conflicts, removed "WIP" status.

campbellcole:
Had issues with a custom modpack throwing this error, compiled with this change and the error went away and everything works perfectly. +1 thank you. Lex, I read about your opinion on fixing this issue and I mostly agree, but I do think since this works and is a non-breaking and minor change, I ask that you consider merging it. Thank you everyone.

LexManos:
We're waiting on Disieben to update it and get back to us. Honestly this is a hacky fix that Mojang should do themselves. I'd LIKE for this to have some notification to bug Mojang to fix their things.
I keep poking them about it, but I can only do so much.

Prophious:
i know this might now help, but it seems back when minecracft had a limit on how many item/block ids as numbers there could be(cant remember the last version that did that) we did get a mod to handle more than at the time 4096(i think) items/blocks

if there was a way to make a mod to do exactly what is needed it might work better and sooner than adding it right in to forge, albeit a forge modification

i really hope this can be resolved in one way or another

best of luck to any modders that are willing to try to take up the task

could be done for a mod jam, with prizes for the one that comes up with the best working code to solve the issue(just an idea)

TheCurle:
There is a mod to do this already. But it is obsolete, because Forge itself now has this feature, as indicated by the "merged" label...

Bossos11:
hey, is this package splitting also including config files?(cause im still getting the badly compressed package error, even though im using the newest version of forgean it only started happening with pmmo[which has a very big config file])

diesieben07:
No, it does not include config files. However if you have a config file that is larger than 1048576 bytes (about 1MB) then you are doing something _very_ wrong.

Bossos11:
im not sure either but the last thing shown in the debug log was the handshake with the config files(before the player is disconnected)

emilsuperboi:
Does this work for forge 1.12.2?

PR #7488, opened 2020-11-13
change the AdvancementProvider to make it possabible for mods to hook into it

Unbekannt1998:
Since all common Data Providers have a way for mods to hook into, except for the `AdvancementProvider`. So i changed the `AdvancementProvider` so other Mods can use it to make there own Advancement. I also needed to hook in the `Advancement.Builder` and  add a way to give a `ResourceLocation` instead of just a string, since all new Advancements would be under the minecraft namespace. 

I included a Test Mod which create 2 new Advancements and overrides one of Minecrafts default Advancements.


ChampionAsh5357:
This should be using `ExistingFileHelper` to determine whether the parents are present before construction. As clarification, `ExistingFileHelper` should be used whenever you are trying to access existing json files that are included with the generator. In this case, since advancements have a parent, they should query the helper to verify that the file exists before continuing to create the advancement json. Since root files also have the ability to not have a parent, you would need to check that the parent tag is not null before verifying the existence using the helper.

patches/minecraft/net/minecraft/advancements/Advancement.java.patch
```diff
@@ -1,6 +1,23 @@
...
 +++ b/net/minecraft/advancements/Advancement.java
-@@ -309,6 +309,7 @@
+@@ -230,9 +230,13 @@
+          }
+       }
+ 
+-      public Advancement func_203904_a(Consumer<Advancement> p_203904_1_, String p_203904_2_) {
+-         Advancement advancement = this.func_192056_a(new ResourceLocation(p_203904_2_));
+-         p_203904_1_.accept(advancement);
++      public Advancement func_203904_a(Consumer<Advancement> p_203904_1_, String p_203904_2_){
```
Cyborgmas: YOu changed a space here after between the ){ creating 2 patch lines. But I would also suggest to not modify the constructor and simply add the other. This would actually be less patch lines overall in this case.

src/test/java/net/minecraftforge/debug/DataGeneratorTest.java
```diff
@@ -136,54 +153,55 @@ protected void registerRecipes(Consumer<IFinishedRecipe> consumer)
             ResourceLocation ID = new ResourceLocation("data_gen_test", "conditional");
 
             ConditionalRecipe.builder()
```
Cyborgmas: Maybe im reading this wrong but it seems to be a lot of unecessary line changes below, changing less lines helps better understand what you did and conserves blame.

src/test/java/net/minecraftforge/debug/DataGeneratorTest.java
```diff
@@ -32,28 +32,28 @@
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import net.minecraft.block.*;
```
Cyborgmas, line 43: The imports got rearranged which is unecessary diff if you could change that.

src/test/java/net/minecraftforge/debug/DataGeneratorTest.java
```diff
@@ -136,54 +153,55 @@ protected void registerRecipes(Consumer<IFinishedRecipe> consumer)
...
 
             ConditionalRecipe.builder()
                     .addCondition(
                             not(
-                                and(
-                                        not(modLoaded("minecraft")),
-                                        itemExists("minecraft", "dirt"),
-                                        FALSE()
-                                )
+                                    and(
```
Cyborgmas: Again here and below you add some tabs, might be better/more correct but its still unecessary.

src/test/java/net/minecraftforge/debug/DataGeneratorTest.java
```diff
@@ -435,7 +457,8 @@ protected void registerStatesAndModels()
                     );
 
             ModelFile barrel = models().cubeBottomTop("barrel", mcLoc("block/barrel_side"), mcLoc("block/barrel_bottom"), mcLoc("block/barrel_top"));
-            ModelFile barrelOpen = models().cubeBottomTop("barrel_open", mcLoc("block/barrel_side"), mcLoc("block/barrel_bottom"), mcLoc("block/barrel_top_open"));
+            ModelFile barrelOpen = models()
```
Cyborgmas: No need to change this

Cyborgmas (commented):
I think it would be better if you revert your style changes to `DataGeneratorTest`, it leads to unecessary diff and makes your changes seem a lot bigger then what they are.

Unbekannt1998:
i added the check for if the RL is a existing file. 
also changed the Formatting of the test file

patches/minecraft/net/minecraft/data/AdvancementProvider.java.patch
```diff
@@ -0,0 +1,38 @@
...
+    public void func_200398_a(DirectoryCache p_200398_1_) throws IOException {
+       Path path = this.field_204025_c.func_200391_b();
+       Set<ResourceLocation> set = Sets.newHashSet();
+@@ -47,12 +51,19 @@
+          }
+       };
+ 
++      registerAdvancement(consumer, fileHelper);
++
++      if(this.getClass() == AdvancementProvider.class)//Forge: Subclasses don't need this.
```
TheCurle: Are you 100% certain that this won't ever become necessary for other mods?

patches/minecraft/net/minecraft/data/AdvancementProvider.java.patch
```diff
@@ -0,0 +1,38 @@
...
+       for(Consumer<Consumer<Advancement>> consumer1 : this.field_204283_d) {
+          consumer1.accept(consumer);
+       }
+ 
+    }
+ 
++   protected void registerAdvancement(Consumer<Advancement> consumer, net.minecraftforge.common.data.ExistingFileHelper fileHelper){
++
```
TheCurle: Is this supposed to be empty?

src/test/java/net/minecraftforge/debug/DataGeneratorTest.java
```diff
@@ -32,7 +32,9 @@
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import net.minecraft.advancements.criterion.InventoryChangeTrigger;
 import net.minecraft.block.*;
+import net.minecraft.data.*;
```
TheCurle: Did you add this?
Try to avoid wildcard imports.

TheCurle (commented):
Mostly some questions to further understanding of what's going on here.
Some design decisions don't make much sense to me, but you have a chance to explain yourself :)
Overall, though, seems good.

patches/minecraft/net/minecraft/data/AdvancementProvider.java.patch
TheCurle: I understand this is for subclasses to override, but seems like you could at least put a javadoc or something here explaining that, as someone else reading this may not understand it.

patches/minecraft/net/minecraft/data/AdvancementProvider.java.patch
Unbekannt1998: yes since in the loop only vanilla Advancements will be generated and mods should use `registerAdvancement` which in essence does the same as the loop.

TheCurle: This should be evident from the code, not from looking at this conversation :)
", use registerAdvancement instead" would suffice

patches/minecraft/net/minecraft/data/AdvancementProvider.java.patch
Unbekannt1998: yes it should be empty since all vanilla Advancements are in a list and don't need to be generated again since they would be generated in the loop in the method above. Also I can't make the function abstract. 
I will add a javadoc to it any specifics i should add?

src/test/java/net/minecraftforge/debug/DataGeneratorTest.java
Unbekannt1998: Intellij does this automatically for me i will change this for forge

patches/minecraft/net/minecraft/advancements/Advancement.java.patch
```diff
@@ -1,6 +1,43 @@
 --- a/net/minecraft/advancements/Advancement.java
 +++ b/net/minecraft/advancements/Advancement.java
-@@ -309,6 +309,7 @@
+@@ -230,12 +230,36 @@
+          }
+       }
+ 
++      @Deprecated //Forge: use the other Method with the fileHelper
```
sciwhiz12: Maybe `//Forge: Use Advancement#register(Consumer<Advancement>, ResourceLocation, ExistingFileHelper)`?

patches/minecraft/net/minecraft/data/AdvancementProvider.java.patch
```diff
@@ -0,0 +1,44 @@
...
+ 
+    }
+ 
++   /**
++    * Subclasses should override this function if they want to generate Advancements.
++    * Just use {@see Advancement.Builder} to build your Advancements, you don't need an extra consumer like the vanilla classes.
++    * @param consumer used for the build function from {@see Advancement.Builder}
++    * @param fileHelper used for the build function from {@see Advancement.Builder}
++    */
++   protected void registerAdvancement(Consumer<Advancement> consumer, net.minecraftforge.common.data.ExistingFileHelper fileHelper){
```
sciwhiz12: I would suggest maybe a rename to `registerAdvancements`, to be consistent with the other similarly-functioning methods in datagen? (`registerTags` for `TagsProvider`, `registerRecipes` in `RecipeProvider`, `registerStatesAndModels` in `BlockStateProvider`, etc.)

Also, a suggestion for the javadoc:
```java
/**
 * Override this method for registering and generating custom {@link Advancement}s.
 * @param consumer Consumer for registering advancements to this provider
 * @param fileHelper Existing file helper
 * @see Advancement.Builder#build(Consumer<Advancement>, ResourceLocation, net.minecraftforge.common.data.ExistingFileHelper)
 */
```

src/test/java/net/minecraftforge/debug/DataGeneratorTest.java
```diff
@@ -645,6 +649,77 @@ public String getName() {
...
+                    new TranslationTextComponent(Items.COBBLESTONE.getTranslationKey()),
+                    new StringTextComponent("You got cobblestone"),
+                    new ResourceLocation("textures/gui/advancements/backgrounds/stone.png"),
+                    FrameType.TASK,
+                    false,
+                    false,
+                    false)
+                    .withCriterion("get_cobbleStone", InventoryChangeTrigger.Instance.forItems(Items.COBBLESTONE))
+                    .withParentId(new ResourceLocation("not_there/not_here"))
+                    .register(consumer, "illegalParent");*/
```
sciwhiz12: From the advancement's name, I assume this is for testing if a exception will happen when the parent is not present/does not exist? 
Please add a comment to clarify it (`// This should cause an error because of the parent not existing`), and also pass in the `ExistingFileHelper` to the `register(consumer, "illegalParent")` call to actually test if the `ExistingFileHelper` does complain about the missing parent.

sciwhiz12 (changes requested):
A few things I'd like to see changed; the big one being the method rename from `registerAdvancement` to `registerAdvancements`, to be consistent with other data providers.

patches/minecraft/net/minecraft/data/AdvancementProvider.java.patch
Unbekannt1998: i would like to keep the second part of my javadoc, since it is more clear what you should use, because if someone looks at the vanilla Classes they could be confused with the additional  consumer they use. Also Intellij does not show the last line when you look at the javadoc on the overridden Method.

Unbekannt1998:
can i get a triage member to look at this and maybe get some feedback on what i still need to change, if needed.

patches/minecraft/net/minecraft/advancements/Advancement.java.patch
```diff
@@ -1,6 +1,42 @@
...
+ 
++      public Advancement register(Consumer<Advancement> consumer, ResourceLocation id, net.minecraftforge.common.data.ExistingFileHelper fileHelper) {
++         Advancement advancement = this.build(id, fileHelper);
++         consumer.accept(advancement);
++         return advancement;
++      }
++
++      public Advancement build(ResourceLocation id, net.minecraftforge.common.data.ExistingFileHelper fileHelper) {
++         if (!this.func_192058_a((p_199750_0_) -> {
++            if(fileHelper.exists(p_199750_0_, net.minecraft.resources.ResourcePackType.SERVER_DATA, ".json", "advancements")) {
```
LexManos: This is too large for a patch. And it makes the code really hard to understand as its in SRG names.
You should move this to a ExistingAdvancementBuilder or something in the forge util package.

patches/minecraft/net/minecraft/data/AdvancementProvider.java.patch
```diff
@@ -0,0 +1,44 @@
...
+    private final DataGenerator field_204025_c;
+    private final List<Consumer<Consumer<Advancement>>> field_204283_d = ImmutableList.of(new EndAdvancements(), new HusbandryAdvancements(), new AdventureAdvancements(), new NetherAdvancements(), new StoryAdvancements());
++   private net.minecraftforge.common.data.ExistingFileHelper fileHelper;
+ 
++   @Deprecated
+    public AdvancementProvider(DataGenerator p_i48869_1_) {
+       this.field_204025_c = p_i48869_1_;
+    }
+ 
++   public AdvancementProvider(DataGenerator generatorIn, net.minecraftforge.common.data.ExistingFileHelper fileHelperIn){ this.field_204025_c = generatorIn; this.fileHelper = fileHelperIn;}
```
LexManos: Don't do crap like this.
It makes the line really long and hard to read.

patches/minecraft/net/minecraft/data/AdvancementProvider.java.patch
LexManos: Better option is to do what I did with the old version of the recipe provider.
And just move that for loop into this `accept(Consumer<Advancmenet>)` function.
So it's obvious that you should override the function if you want to not have the vanilla stuff.
This also fixes the annoying getClass() test you have to noop the vanilla stuff.

patches/minecraft/net/minecraft/data/AdvancementProvider.java.patch
```diff
@@ -0,0 +1,44 @@
+--- a/net/minecraft/data/AdvancementProvider.java
++++ b/net/minecraft/data/AdvancementProvider.java
+@@ -24,11 +24,15 @@
+    private static final Gson field_204024_b = (new GsonBuilder()).setPrettyPrinting().create();
+    private final DataGenerator field_204025_c;
+    private final List<Consumer<Consumer<Advancement>>> field_204283_d = ImmutableList.of(new EndAdvancements(), new HusbandryAdvancements(), new AdventureAdvancements(), new NetherAdvancements(), new StoryAdvancements());
++   private net.minecraftforge.common.data.ExistingFileHelper fileHelper;
```
LexManos: Should be protected to allow the child class to access it.

LexManos (changes requested):

patches/minecraft/net/minecraft/advancements/Advancement.java.patch
Unbekannt1998: i hope my latest commit is what you meant by this

patches/minecraft/net/minecraft/advancements/Advancement.java.patch
```diff
@@ -1,6 +1,37 @@
...
+       }
+ 
++      public Advancement register(Consumer<Advancement> consumer, ResourceLocation id, net.minecraftforge.common.data.ExistingFileHelper fileHelper) {
++         Advancement advancement = this.build(id, fileHelper);
++         consumer.accept(advancement);
++         return advancement;
++      }
++
++      public Advancement build(ResourceLocation id, net.minecraftforge.common.data.ExistingFileHelper fileHelper) {
++         if (!net.minecraftforge.common.data.AdvancementBuilderHelper.build(fileHelper, this)){
```
covers1624: Space before ending brace. `)){` -> `)) {`

patches/minecraft/net/minecraft/data/AdvancementProvider.java.patch
```diff
@@ -0,0 +1,38 @@
...
+    private final DataGenerator field_204025_c;
+    private final List<Consumer<Consumer<Advancement>>> field_204283_d = ImmutableList.of(new EndAdvancements(), new HusbandryAdvancements(), new AdventureAdvancements(), new NetherAdvancements(), new StoryAdvancements());
++   protected net.minecraftforge.common.data.ExistingFileHelper fileHelper;
+ 
++   @Deprecated
+    public AdvancementProvider(DataGenerator p_i48869_1_) {
+       this.field_204025_c = p_i48869_1_;
+    }
+ 
++   public AdvancementProvider(DataGenerator generatorIn, net.minecraftforge.common.data.ExistingFileHelper fileHelperIn){
```
covers1624: Again space before brace.

patches/minecraft/net/minecraft/data/AdvancementProvider.java.patch
```diff
@@ -0,0 +1,38 @@
...
++      registerAdvancements(consumer, fileHelper);
++   }
++
++   /**
++    * Override this method for registering and generating custom {@link Advancement}s. <br>
++    * Just use {@link Advancement.Builder} to build your Advancements, you don't need an extra consumer like the vanilla classes.
++    * @param consumer used for the register function from {@link Advancement.Builder}
++    * @param fileHelper used for the register function from {@link Advancement.Builder}
++    */
++   protected void registerAdvancements(Consumer<Advancement> consumer, net.minecraftforge.common.data.ExistingFileHelper fileHelper){
```
covers1624: Again space before brace.

src/main/java/net/minecraftforge/common/data/AdvancementBuilderHelper.java
```diff
@@ -0,0 +1,18 @@
+package net.minecraftforge.common.data;
```
covers1624: License header. Run the `updateLicense` Gradle task.

src/main/java/net/minecraftforge/common/data/AdvancementBuilderHelper.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.google.common.collect.Maps;
+import net.minecraft.advancements.Advancement;
+import net.minecraft.advancements.AdvancementRewards;
+import net.minecraft.resources.ResourcePackType;
+
+public class AdvancementBuilderHelper {
+
+    public static boolean build(ExistingFileHelper fileHelper, Advancement.Builder builder){
```
covers1624: Space before brace.

src/main/java/net/minecraftforge/common/data/AdvancementBuilderHelper.java
```diff
@@ -0,0 +1,18 @@
...
+        return builder.resolveParent((advancementId) -> {
+            if(fileHelper.exists(advancementId, ResourcePackType.SERVER_DATA, ".json", "advancements")) {
```
covers1624: space after `if` before the first parenthesis.

covers1624 (changes requested):

Unbekannt1998:
updated to the last Forge version and changed the names to the Mojang Mapping names

patches/minecraft/net/minecraft/advancements/Advancement.java.patch
```diff
@@ -1,5 +1,36 @@
...
++      public Advancement build(ResourceLocation id, net.minecraftforge.common.data.ExistingFileHelper fileHelper) {
++         if(!net.minecraftforge.common.data.AdvancementBuilderHelper.build(fileHelper, this)) {
++            throw new IllegalStateException("Tried to build incomplete advancement!");
++         } else {
++            if (this.field_192066_f == null) {
++               this.field_192066_f = this.field_199751_g.createRequirements(this.field_192065_e.keySet());
++            }
++            return new Advancement(id, this.field_192062_b, this.field_192063_c, this.field_192064_d, this.field_192065_e, this.field_192066_f);
++         }
++      }
```
ChampionAsh5357: So I'm going to reiterate Lex's point here. Most of this could be moved into a different class. Technically, all you should need to do here is have a method that takes in the advancement builder, data provider consumer, id, and the existing file helper. You could then call `#canBuild` within that method with the function to create a dummy advancement (which is what you are doing in `AdvancementBuilderHelper#build` and then `Builder#build` with no issues. Most you should need is three lines here, one for the method definition, one for the return statement to the outside method call, and one for the end bracket.

patches/minecraft/net/minecraft/data/AdvancementProvider.java.patch
```diff
@@ -0,0 +1,38 @@
...
+       Set<ResourceLocation> set = Sets.newHashSet();
+@@ -47,6 +_,16 @@
+          }
+       };
+ 
++      registerAdvancements(consumer, fileHelper);
++   }
++
++   /**
++    * Override this method for registering and generating custom {@link Advancement}s. <br>
```
ChampionAsh5357: I believe javadoc convention is to use an opening `<p>` on line breaks.

src/main/java/net/minecraftforge/common/data/AdvancementBuilderHelper.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class AdvancementBuilderHelper {
+
+    public static boolean build (ExistingFileHelper fileHelper, Advancement.Builder builder) {
+        return builder.canBuild((advancementId) -> {
+            if (fileHelper.exists(advancementId, ResourcePackType.SERVER_DATA, ".json", "advancements")) {
+                return new Advancement(advancementId, null, null, AdvancementRewards.EMPTY, Maps.newHashMap(), null);
+            }
+            return null;
+        });
```
ChampionAsh5357: When talking about the other comment. This is pretty much what I mean since you can do this and build the advancement in one method. You can ever return a more detailed error message specific to data providers.

ChampionAsh5357 (changes requested):
Mainly just need to lower the patches to `Advancement`, rest seems pretty good.

patches/minecraft/net/minecraft/data/AdvancementProvider.java.patch
```diff
@@ -0,0 +1,39 @@
...
+          }
+       };
+ 
++      registerAdvancements(consumer, fileHelper);
++   }
++
++   /**
++    * Override this method for registering and generating custom {@link Advancement}s.
++    * <p>
++    * Just use {@link Advancement.Builder} to build your Advancements, you don't need an extra consumer like the vanilla classes.
```
ChampionAsh5357: You can move them to the same line.

src/main/java/net/minecraftforge/common/data/AdvancementBuilderHelper.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.google.common.collect.Maps;
+import net.minecraft.advancements.*;
+import net.minecraft.resources.ResourcePackType;
+import net.minecraft.util.ResourceLocation;
+
+import java.util.Map;
+import java.util.function.Consumer;
+
+public class AdvancementBuilderHelper {
```
ChampionAsh5357: Braces on new line for everything in the class.

ChampionAsh5357 (commented):
I'll test this when I get home in a few hours to verify everything works. For now, just minor formatting fixes.

src/main/java/net/minecraftforge/common/data/AdvancementBuilderHelper.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.function.Consumer;
+
+public class AdvancementBuilderHelper
+{
+
+    public static Advancement build(ResourceLocation id, Consumer<Advancement> consumer, ExistingFileHelper fileHelper, Advancement.Builder builder)
+    {
+        boolean canBuild = builder.canBuild((advancementId) ->
+        {
+            if (fileHelper.exists(advancementId, ResourcePackType.SERVER_DATA, ".json", "advancements")) {
```
ChampionAsh5357: Rest of this looks good to me. Just this one brace on new line.

ChampionAsh5357:
Probably can remove `.cache` and `good_parent` since they are both changes during datagen.

ChampionAsh5357 (approved):

PR #7491, opened 2020-11-14
Append new event: EntityEmittedSoundEvent

EtiTheSpirit:
# The Problem This Solves
When attempting to override certain sounds, primarily those intercepted in `PlaySoundAtEntityEvent`, the lack of a source entity for these sounds can make contextual changes flat out impossible to perform in this event alone, often requiring external tracking or reimplementing some sounds yourself.

This event was designed for the express purpose of making contextual sound replacement easier, that is, replacement that requires a reference to the entity that emitted the sound to perform accurately. This is not possible via `PlaySoundAtEntityEvent` because the entity it passes in is the *listener*, not the entity that played the sound. This event resolves this problem by instead providing a reference to the entity that played the sound.

# Use Cases & Examples

#­1: The modder wishes to replace footstep sounds on a wide range of Block materials for players who have a specific capability registered. Let us assume that the local player has this capability that will override their footstep sounds. This will work for replacement of sounds locally just fine. However, assume that a remote player without this capability walks nearby. Because `PlaySoundAtEntityEvent` provides no source entity, detecting that the footstep sounds were *not* created by the local player entity is impossible. As a result, the remote player will emit footstep sounds based on the current material the local player is standing on.

#­2: The modder wishes to replace sounds should a supercharged creeper activate its fuse. They are able to detect that the TNT sound was emitted in the world via `PlaySoundAtEntityEvent`, however there is no way to detect if this sound was emitted by a charged creeper because there is no source entity. As a result, the modder must use alternative methods of detecting this, requiring much more work than desired.

# Implementation
The implementation of this change is straightforward. Alongside appending `EntityEmittedSoundEvent` (which is virtually identical to `PlaySoundAtEntityEvent` with the exception that it provides the source entity and allows changing the position of the sound in 3D space), `net.minecraft.entity.Entity` is modified to call a method newly appended to `IForgeWorld` rather than `world.playSound`. This new method fires `EntityEmittedSoundEvent`, performs any necessary alterations to the data, and then proceeds to call the standard `world.playSound` method as `Entity` once did.

# Included example
This mod alters the sound of a supercharged creeper's fuse. To test:
* Load into a new world with cheats enabled.
* Execute `/summon minecraft:creeper ~ ~ ~`
* Allow the creeper to explode. Sounds should be standard.
* Respawn (if necessary), and execute `/summon minecraft:creeper ~ ~ ~ {powered:true}`
* Allow the charged creeper to explode. The fuse sound should be replaced with a one second long sawtooth wave.


src/EXAMPLE_IMPLEMENTATION/java/etithespirit/UnitTesting.java
```diff
@@ -0,0 +1,62 @@
+package etithespirit;
```
sciwhiz12: Test mods should be placed under the `src/test` source set, under the `net.minecraftforge.debug` package (or one of its subpackages). _(in your case, it could be placed under `net.minecraftforge.debug.entity` since it affects entities)_

The class name should reflect the purpose of the test or the feature that is being tested, and suffixed with `Test`. _(in your case, it could be named as `EntityEmittedSoundEventTest`)_

And please run the `updateLicenses` task to add the license header after this class is moved and renamed.

src/EXAMPLE_IMPLEMENTATION/java/etithespirit/UnitTesting.java
```diff
@@ -0,0 +1,62 @@
...
+import net.minecraft.entity.monster.CreeperEntity;
+import net.minecraft.util.ResourceLocation;
+import net.minecraft.util.SoundEvent;
+import net.minecraft.entity.Entity;
+
+
+import java.util.stream.Collectors;
+
+// The value here should match an entry in the META-INF/mods.toml file
+@Mod("unittesting")
```
sciwhiz12: The modid should preferrably be a `snake_case` rendition of the class' name. _(for `EntityEmittedSoundEventTest`, `entity_emitted_sound_event_test`)_

And a corresponding `[[mods]]` entry should be placed in the `src/test/resources/META-INF/mods.toml` file, at the very bottom of the file.

src/EXAMPLE_IMPLEMENTATION/java/etithespirit/UnitTesting.java
```diff
@@ -0,0 +1,62 @@
...
+public class UnitTesting
+{
+    private static final SoundEvent SAWTOOTH_WAVE = new SoundEvent(new ResourceLocation("unittesting:implementation.sawtooth"));
```
sciwhiz12: As `SoundEvent` is a registrable object, please move this to using `DeferredRegister` and `RegistryObject`; and the registry name's path should be simplified to `sawtooth`.

src/EXAMPLE_IMPLEMENTATION/java/etithespirit/UnitTesting.java
```diff
@@ -0,0 +1,62 @@
...
+    
+    static {
+        SAWTOOTH_WAVE.setRegistryName("unittesting", "implementation.sawtooth");
+    }
+    
+    public UnitTesting() 
+    {
+        MinecraftForge.EVENT_BUS.register(this);
+        MinecraftForge.EVENT_BUS.register(Implementation.class);
+        GameRegistry.findRegistry(SoundEvent.class).register(SAWTOOTH_WAVE);
```
sciwhiz12: Please remove the use of that `Implementation` class; the event listener should be placed in the outermost class.
Remove the `GameRegistry.findRegistry` call, and replace (as stated above) with using `DeferredRegister` and `RegistryObject`.

All test mods should have a `private static final boolean ENABLE = true;` that controls if the event bus registration calls should be called (using an `if`-check).
This is used for disabling specific tests, such as for executing only a single test or default disabling tests which cause ~~horrible~~ unwanted side-effects in other tests. _(in your case, it can be default enabled, since it does not have side-effects)_

src/main/java/net/minecraftforge/event/entity/EntityEmittedSoundEvent.java
```diff
@@ -0,0 +1,86 @@
...
+ * Similarly, changing the {@link #position} will move the effective position of the sound. Setting this to a null value will cause the sound to not be played.<br>
+ * <br>
+ * This event is {@link Cancelable}.<br>
+ * If this event is canceled, the sound is not played.<br>
+ * <br>
+ * This event does not have a result. {@link HasResult} <br>
+ * <br>
+ * This event is fired on the {@link MinecraftForge#EVENT_BUS}.<br>
+ **/
+@net.minecraftforge.eventbus.api.Cancelable
```
sciwhiz12: Since this is a Forge class, this use of a fully qualified name should be replaced with an import.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -748,4 +749,11 @@ public static void onCommandRegister(CommandDispatcher<CommandSource> dispatcher
         MinecraftForge.EVENT_BUS.post(evt);
         return evt;
     }
+    
+    public static EntityEmittedSoundEvent entityEmittedSound(Entity source, double x, double y, double z, SoundEvent soundIn, SoundCategory category, float volume, float pitch)
+    {
+        EntityEmittedSoundEvent evt = new EntityEmittedSoundEvent(source, new net.minecraft.util.math.vector.Vector3d(x, y, z), soundIn, category, volume, pitch);
```
sciwhiz12: As this is a Forge class, this use of a fully qualified name should be replaced with an import.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -748,4 +749,11 @@ public static void onCommandRegister(CommandDispatcher<CommandSource> dispatcher
         MinecraftForge.EVENT_BUS.post(evt);
         return evt;
     }
+    
+    public static EntityEmittedSoundEvent entityEmittedSound(Entity source, double x, double y, double z, SoundEvent soundIn, SoundCategory category, float volume, float pitch)
```
sciwhiz12: To keep this consistent with the other event factory methods, I suggest this be renamed to `onEntityEmittedSound`.

sciwhiz12 (changes requested):
IMO, this is a good event to have, to selectively modify/cancel sounds from specific entities. (It may even allow better sound muffler mods.)

But there are some changes which needs to be done, the major one being moving the test mod  to the `src/test` source set and making it consistent with the other test mods.

EtiTheSpirit:
Everything should be good to go. All tested and functional as it should be.

50ap5ud5:
Thanks for making this! This will be useful for my use case, which is allowing a sound that is being played in one dimension to also be played in another dimension.

sciwhiz12:
Hello! @EtiTheSpirit, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

EtiTheSpirit:
Hey! For the time being, I think the best option is to close this PR and so I will be doing so. Additionally, there are some design flaws for the long term that have arisen and need to be addressed, such as what to do when an entity *causes* a sound (rather than emitting it themselves), and how that applies to this event since the entity is related to it.

Solving this issue requires a few more changes than the scope of this event initially aimed to encompass, and so the design of this PR needs to be scaled up a bit. I do have mixin implementations of this behavior in the 1.16 build of my mod, so in the future and after I am familiar with 1.18's workings, I may open a different PR with a slightly different event model better suited for serving the underlying goal of this event.

PR #7492, opened 2020-11-14
[1.16.x] Allow custom redstone wire to be fully interoperable with normal redstone wire

se7kn8:
When creating a subclass of redstone wire, there is a problem that redstone wire is not fully interoperable with the new wire (connections not working correctly). This is because checks in `RedstoneWireBlock` using a hardcoded reference to `Blocks.REDSTONE_WIRE` or `this`.

Situtation before this PR:
Before PR

To solve this problem, this PR creates new block tag (`forge:redstone_wire`) which contains by default only the vanilla redstone wire. All checks in `RedstoneWireBlock` that uses the hardcoded value have been replaced to check if the block is in the tag. (`getPower`, `notifyWireNeighborsOfStateChange`, `canConnectTo`). So if modders want to create a new wire they just have to add the wire to the tag (see example mod).

But while testing these changes, I found out that there was now a problem, that when there was a connection between modded and vanilla wire, the wire would never get turned off. This is because `RedstoneWireBlock` uses a field (`canProvidePower`) to switch the wire off when looking for powering connections. But since this value is per-instance, a custom wire does not know when it should change this value. This could be solved by using reflection to read the value from `Blocks.REDSTONE_WIRE`, but I thought it would be easier to make it static, since all subclasses of `RedstoneWireBlock` should have the same behavior.

Sitation after this PR:


patches/minecraft/net/minecraft/block/RedstoneWireBlock.java.patch
```diff
@@ -27,10 +50,12 @@
 -   }
 -
 -   protected static boolean func_176343_a(BlockState p_176343_0_, @Nullable Direction p_176343_1_) {
+-      if (p_176343_0_.func_203425_a(Blocks.field_150488_af)) {
 +   protected static boolean canConnectTo(BlockState p_176343_0_, IBlockReader world, BlockPos pos, @Nullable Direction p_176343_1_) {
-       if (p_176343_0_.func_203425_a(Blocks.field_150488_af)) {
++      if (p_176343_0_.func_235714_a_(net.minecraftforge.common.Tags.Blocks.REDSTONE_WIRE)) {
```
gigaherz: A tag is not valid for this. Using a tag means people could add random blocks to the redstone tag, which a vanilla client would have no idea of, and thus would cause a desync. This should only consider blocks that extend RedstoneWireBlock.

On top of that, I have a concern about this: what about all the blocks that want to implement "non-connecting" redstone? Like other colors of redstone?

EDIT: 
To clarify on the tags: Forge has a policy of remaining compatible with vanilla clients, and vanilla servers, if mods aren't marking the connection as incompatible. This means any change to logic that runs on both client and server can do things vanilla itself would do, if no mods are present.

While the tags themselves will exist and are synchronized, the code itself will not be patched in the vanilla side, so one of the sides would be reading from the tag, but the other wouldn't, meaning the presense of a datapack with values in this redstone tag would make the two sides run different logic.

gigaherz (changes requested):

patches/minecraft/net/minecraft/block/RedstoneWireBlock.java.patch
se7kn8: 1. I got your point, I will remove the tags and change it to `instanceof`

2. What about adding a method like `boolean canConnectToOtherRedstone()` in `RedstoneWireBlock` which returns `false` be default. Then for every check:
```java
if(blockState.isIn(Blocks.REDSTONE_WIRE){
  return true;
}
// I do the check here, so if the method returns false, it will still go through the other if branches
else if(blockState.getBlock() instanceof RedstoneWireBlock && ((RedstoneWireBlock) blockState.getBlock()).canConnectToOtherRedstone()){
  return true;
}else if {
// ...
}
```
This won't cause any breaking changes with existing mods, since the behavior will only change if they override the method and change the return value.

mrp-v2: I like that idea, it sounds like it would work fine with mods that currently add non-connecting variants of redstone wire. Example for reference: [More Wires](https://github.com/mrp-v2/MoreWires)

gigaherz: This sounds like a better option to me, however, the blocks would probably need to know the other block wants *them* specifically. Something like like:
```java
if(otherblock == thisblock || (otherblock instanceof RedstoneWireBlock && otherblock.canConnectToOther(this) && this.canConnectToOther(otherblock)))
```
Also, if you want this in 1.16.x, we need to make sure the default state doesn't break semantics. So by default canConnectToOther should behave the same way the old logic would have worked.

se7kn8:
I've done these change now in a way which doesn't break vanilla behavior. But for this the user of the new redstone wire need to overrride `getPower`, because otherwise it would break vanilla behavior. I've added an AT entry for this.

It's not possible to check if the other block wants a connection, because then `RedstoneWireBlock` would need to return `true`, which would change vanilla behavior.

patches/minecraft/net/minecraft/block/RedstoneWireBlock.java.patch
```diff
@@ -1,5 +1,14 @@
 --- a/net/minecraft/block/RedstoneWireBlock.java
 +++ b/net/minecraft/block/RedstoneWireBlock.java
+@@ -47,7 +47,7 @@
+    private final Map<BlockState, VoxelShape> field_235541_j_ = Maps.newHashMap();
+    private static final Vector3f[] field_235542_k_ = new Vector3f[16];
+    private final BlockState field_235543_o_;
+-   private boolean field_150181_a = true;
++   private static boolean field_150181_a = true;
```
pupnewfster: Why is this being changed to static? It strikes me like this could quite easily break a bunch of things.

se7kn8: Because Redstone wires check if there are powered by disabling themselves (through this field) and looking for other sources. If this is per instance (non static), subclasses would not know when they should disable powering, which results in infinite powering. To prevent this I've changed it to static, so subclasses also disable powering.

mrp-v2:
It is possible to check if the other block wants to connect if you make the default behavior return true if the blocks are the same.
`return blockstate.isIn(this)`
This would not change vanilla behavior.

patches/minecraft/net/minecraft/block/RedstoneWireBlock.java.patch
KnightMiner: If we want it to behave as before, you probably want an or condition between the two can connect calls. Or maybe make some sort of 3 way return for block, allow, and default.

mrp-v2: I think what giga suggested is already fine. If it was an or condition, then redstones that don't want to connect to other redstones would be forced to connect to restones that do want to connect to other redstones. I also think a 3-way return is uneccesary, as returning default would probably be no different from returning allow, unless I'm misunderstanding what you mean.

se7kn8:
`return blockstate.isIn(this)` would always return `true`, since the method is called: `state.getBlock().canConnectToOther(state)`. There is no way to get the blockstate of the caller (which would be a second parameter) without breaking anything.

mrp-v2:
Just add another parameter to `RedstoneWireBlock#canConnectTo` and `RedstoneWireBlock#canConnectToOther` to be the blockstate of the caller. The method is only used in `RedstoneWireBlock#recalculateSide`, and the info that that method has can be used to obtain both necessary blockstates.
```
   private RedstoneSide recalculateSide(IBlockReader reader, BlockPos pos, Direction direction, boolean nonNormalCubeAbove) {
      BlockPos blockpos = pos.offset(direction);
      BlockState blockstate = reader.getBlockState(blockpos);
      BlockState thisState = reader.getBlockState(pos);
      if (nonNormalCubeAbove) {
         boolean flag = this.canPlaceOnTopOf(reader, blockpos, blockstate);
         if (flag && canConnectTo(thisState, reader.getBlockState(blockpos.up()), reader, blockpos.up(), null) ) {
            if (blockstate.isSolidSide(reader, blockpos, direction.getOpposite())) {
               return RedstoneSide.UP;
            }

            return RedstoneSide.SIDE;
         }
      }

      return !canConnectTo(thisState, blockstate, reader, blockpos, direction) && (blockstate.isNormalCube(reader, blockpos) || !canConnectTo(reader.getBlockState(blockpos.down()), reader, blockpos.down(), null)) ? RedstoneSide.NONE : RedstoneSide.SIDE;
   }

   //pass both states so that wires have all the info they may need to decide if they want to connect to other wires
   protected static boolean canConnectTo(BlockState thisState, BlockState otherState, IBlockReader world, BlockPos pos, @Nullable Direction side) {
      if (otherState.isIn(Blocks.REDSTONE_WIRE)) {
         return true;
      } else if (otherState.getBlock() instanceof RedstoneWireBlock && ((RedstoneWireBlock)otherState.getBlock()).canConnectToOther(otherState, thisState) && ((RedstoneWireBlock)thisState.getBlock()).canConnectToOther(thisState, otherState)) {
         return true;
      } else if (otherState.isIn(Blocks.REPEATER)) {
         Direction direction = otherState.get(RepeaterBlock.HORIZONTAL_FACING);
         return direction == side || direction.getOpposite() == side;
      } else if (otherState.isIn(Blocks.OBSERVER)) {
         return side == otherState.get(ObserverBlock.FACING);
      } else {
         return otherState.canConnectRedstone(world, pos, side) && side != null;
      }
   }

   //pass both states so that wires have all the info they may need to decide if they want to connect to other wires
   public boolean canConnectToOther(BlockState thisState, BlockState otherState) {
      return otherState.isIn(this);
   }
```

mrp-v2:
I've made changes to make wire connections check that both wires want to connect to each other. See [this commit](https://github.com/mrp-v2/MinecraftForge/commit/fe74fcde3391204d6b803f2f929a515361365f92). I tried to PR https://github.com/SE7-KN8/MinecraftForge/tree/custom-redstone-wire but it wasn't working.

se7kn8:
I really like these changes, I can merge them tomorrow into this PR. But maybe you can change `canConnectToOther` with both blockstates as parameters, so that wires have the possibility to change their behavior based on their blockstate.

mrp-v2:
I've now changed `canConnectToOther` to take both blockstates, which enabled me to move it to an extension class, decreasing the patch size. Also fixed the formatting of the test mod class. See https://github.com/mrp-v2/MinecraftForge/tree/custom-redstone-wire.

se7kn8:
I merged your changes into my local repo, but found some problem while testing it:

As you see in the same, the custom wire is not fully interoperable with the redstone wire, and the state checking of the redstone seems to be broken.

mrp-v2:
Can you push it to a new branch on your github? Then I'll work on thoroughly testing and fixing those issues. Also can you look at the settings for your github repo and make sure there's not something there disabling PR's? That will make it easier for you to merge my changes.

se7kn8:
I've created a new branch, to which you should be able to PR [https://github.com/SE7-KN8/MinecraftForge/tree/redstone-wire](https://github.com/SE7-KN8/MinecraftForge/tree/redstone-wire)

mrp-v2:
It's still not letting me PR your branch, but I think I've finally fixed everything: https://github.com/mrp-v2/MinecraftForge/tree/custom-redstone-wire

ghost:
Is this is a mod?

DaemonUmbra:
> Is this is a mod?  

Is *what* a mod?
Do you understand the process going on right now?

se7kn8:
_To be honest, I don't understand why you cant PR to my fork. Nothing is disabled._
So I updated the PR with your commits. I didn't found any problem while testing your changes, LGTM!
@mrp-v2 If you have any changes in the next days, I suggest you open a new PR since I won't have much time the next days to always update the fork.

LexManos:
No. Redstone is a mess. If you want to be 'fully interoperable' with normal redstone. Use normal redstone.
Maybe when Mojang makes redstone sane we can re-address this.

PR #7496, opened 2020-11-15
Don't cache the blockstate twice, speeds up connecting to servers with many mods

ichttt:
Currently, when connecting to a server, the blockstate is cached twice: Once when the remote snapshot is injected, and once when tags have been synced correctly. Rebuilding the cache can take a lot of time with many mods though. This PR alters the behaviour, so when a remote snapshot gets injected or the registry is being frozen, the blockstates are not being cached. The existing cache is just being cleared, so that no old date remains in the cache.
Unfortunally, due to BakeCallback being a functional interface, the PR has a lot of places that can only be cleaned up after a breaking change.


LexManos:
This needs cleanup, you can have multiple methods in a functional interface, it just requires that only one be abstract.
As such adding a new function that bounces to the old is the best way to go for now, until the next breaking change. It isolates the changes to just the Builder.getBake function.

Also, I've spent a little time digging into the vanilla code, it seems that Mojang removed all the places where block state shapes were dependent on tags. So this re-caching probably isn't needed anymore. So we could just remove the call from the onBake handler and instead only do it once for every block. Tho the concern is existing mods at that point. 

Perhaps add a 'shapeDependsOnTags' flag to AbstractBlock/BlockProperties, and only invalidate the cache if that flag is set? We'd also need to change the cache from ALWAYS rebuilding when cacheState is called, to having a 'invalid' flag to gate it.

covers1624:
This has merge conflicts and needs to be rebased.

sciwhiz12:
Hello! @ichttt, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after a few days, this PR will be closed for inactivity.

sciwhiz12:

PR #7498, opened 2020-11-16
Speed up searching for classes/resources by removing calls to Files.exists

ichttt:
This PR speeds up the loading of resources and classes.
One of the problem is that in a pack with 200 mods, in a worst case we have to check 200 jars to find a class before we find the correct entry. This is done using `Files.exists`, which is not the fastest method as it has to go through the zip stack and do I/O stuff.
This PR introduces a cache of the top level dirs in a jar, so a jar can be ignored for a specified resource when the path start does not match any of the roots. If it does match, we still need to perform a `Files.exists` lookup as we only keep the first 2 levels of dirs, and up to 30 entries in total to reduce memory load and CPU time, but we can exclude many jars already based on this small cache. This has shown to provide a significant uplift in start time for large packs (see https://gist.github.com/ichttt/d28de6a89781da91863975f242762153).
I didn't do a client run yet, but the gains should be visible there as well, I just did not find the time to test it yet
Requires https://github.com/MinecraftForge/ForgeSPI/pull/7


LexManos:
Alright, seems cpw isn't going to get around to addressing this. I really don't like that nio's Files.exists() is an expensive operation. But there isn't anything we can do to fix java itself.

This isn't a bad change, however I think we should hide this behind a config value.
That config should have 3 values:
disable: Use the system that exists today, High CPU time, Low Memory
balance: Use your system, Medium CPU, Medium Memory
full: Screw it store a set of ALL files. Lowe CPU, High Memory.

Let the users decide which one they want.

cpw:
OK. I'll bite. Why do you make this assertion? Looking at your data in the gist file, the apparent speedup is ~10%. However, the variance on NEW is such that it's not possible to say it's statistically significant, and it is quite plausible that this apparent speedup is just a statistical anomaly.

Where else have you seen this "Files.exists" performance is bad information? I would like some supporting data with significant statistical analysis..

For our typical uses, it's literally a direct reference to existing data structures. It's not calling stat or anything, it's querying properties of the ZipFile entry directly...

cpw:
Plan:
1. Please produce more solid numbers, and validate the difference JDK version makes. I'd like to see JDK 14 vs JDK 8 numbers, to verify that this is NOT a problem on more recent JVMs.
2. A JMH test demonstrating the problem would be most excellent and would be much more useful than pack loading numbers to be honest.
2. Please provide memory impacts of this.

sciwhiz12:
Hello! @ichttt, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

sciwhiz12:

PR #7501, opened 2020-11-18
[1.16] Fix Entity EyeHeight and EntityEvent.Size problems.

### LINKED ISSUE #7399: Different sized slimes all have the same eye height
**Minecraft Version:** 1.16.3

**Forge Version:** 1.16.3-34.1.10

**Steps to Reproduce:**
 1. Press F3 + B to show hitboxes
 2. /summon minecraft:slime ~ ~ ~ {Size:10}
 3. Observe eye height very low (compared to vanilla where it's about halfway up)
 4. (Optional): summon more sizes of slimes, and all appear to have the same eye height

**Description of issue:**
I think the issue may be in Entity.recalculateSize(), where it calculates eye height using the old size instead of the new one when creating the EntityEvent.Size event. Aside from the slimes, when I mod entities to have different sizes, they always have the same eye height no matter their size.

mrp-v2:
Fixes #7399.

Fixes the new eye height from the event being incorrect by splitting the event into 2 pieces, a size event and a separate eye height event. This allows the eye height passed to the event to be the actual vanilla eye height, instead of a best guess eye height.

### DIFF
--- a/patches/minecraft/net/minecraft/entity/Entity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/Entity.java.patch
@@ -31,20 +31,19 @@
        this.field_200606_g = p_i48580_1_;
        this.field_70170_p = p_i48580_2_;
        this.field_213325_aI = p_i48580_1_.func_220334_j();
-@@ -211,7 +214,11 @@
+@@ -211,7 +214,10 @@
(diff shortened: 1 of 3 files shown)


mrp-v2:
Superseded by #7628.

PR #7502, opened 2020-11-19
[1.17] Remove mushroom feature patches

### LINKED ISSUE #7322: IForgeBlock.canBeReplacedByLogs patch for AbstractBigMushroomFeature break vanilla behavior
In AbstractBigMushroomFeature.generateStem, the vanilla code calls 
`if (!world.getBlockState(blockPos).isOpaqueCube(world, blockPos)) {`
which was patched with canBeReplacedByLogs instead (and is only used in AbstractBigMushroomFeature)
`if (world.getBlockState(blockPos).canBeReplacedByLogs(world, blockPos)) {`

https://github.com/MinecraftForge/MinecraftForge/blob/fe43088c96cf65c6793260eae96416871a9eb8f2/patches/minecraft/net/minecraft/world/gen/feature/AbstractBigMushroomFeature.java.patch#L7-L8

Beside the patched method's name being wonky, taking a closer look at how canBeReplacedByLogs reveals that it is not a suitable replacement for `! isOpaqueCube`

https://github.com/MinecraftForge/MinecraftForge/blob/c75eb9e7cfbdcde418f1a20175883cf61f10d900/src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java#L317-L321

Essentially, the method is suppose to make giant mushroom stems replace any block that is not opaque. The forge patch here make it unable to replace many non-opaque blocks such as ferns, flowers, bamboo, ice, cocoa, lily pads, chorus plants. Furthermore, it will also replace Dirt, Grass Block, and Logs which are opaque blocks which directly contradicts what the original method was. 

It might be better to just remove this patch for now as it is really specific to just AbstractBigMushroomFeature's stem at the moment and fails to preserve vanilla behavior for it. Or change canBeReplacedByLogs to canBeReplacedByMushroomStem and have it implement `! isOpaqueCube` by default and let other mod override it to allow/disallow mushroom stem replacement.

mrp-v2:
Resolves #7322
~~Adds new methods to `IForgeBlock` and `IForgeBlockState` that have more accurate names, and points the current methods to those methods. Changes the behavior of big mushroom growth block replacement to match vanilla, and marks the related `IForgeBlock` and `IForgeBlockState` methods as deprecated for removal in 1.17. Also fixes the javadocs for the methods being erroneous.~~ Removes the mushroom feature patches.


alcatrazEscapee:
This is marked 1.17 + breaking change, but 1.17 is also a target which is wholly undefined. Vanilla could do any number of changes in the meantime meaning this would have to be rewritten anyway. Basically, what is the point in making a PR six months in advance, *of even seeing the code you are going to be PR-ing against,* just to get stalebot to shut it down as why would this receive any activity before then?

Considering:

- This is a bug in 1.16. Fixing it should not involve "let's wait six months until 1.17".
- Forge will not accept breaking changes in 1.16
- There is no reason for this fix to *require* a breaking change.

For example, simply adding `@Deprecated` / cue for removal markings on the original method `canBeReplacedByLogs`, and *adding* `canBeReplacedByMushroomBlocks` makes this non breaking, able to be considered for 1.16, and probably not a waste of time.

*but that's just like, your opinion man.*

mrp-v2:
The changes are now non-breaking.

ichttt:
Tbh we should just drop these methods complety and remove the patches. It's now a very specific patch, and the fact nobody noticed this since 1.14 means it is likely to be unsued

LexManos:
Ya, the new worldgen stuff kinda screwed things up. I'm hoping that 1.17's changes will cleanup all this vanilla code. But we'll see.
However, if we do change vanilla worldgen we should patch it to not. Do you have a world see to demonstrate the issue?

mrp-v2:
After around an hour of searching a single biome world, I was unable to find any differences in giant mushroom generation between vanilla and forge. I found lots of places like this in both vanilla and forge (This shows that mushroom 'leaf' generation is the same.):

alcatrazEscapee:
The difference in behavior here is very simple to reproduce. Here is just one example:

1. Start with the following setup


2. Attempt to grow the huge mushroom using bone meal.

The result on Vanilla 1.16.4:



The result on Forge  31.1.15

mrp-v2:
Ah, I was under the impression that it only effected big mushrooms during world gen.

mrp-v2:
Should I adjust the PR to mark the methods as deprecated and make them do nothing, making forge behavior consistent with vanilla? Or should I adjust the logic in the methods to match vanilla behavior?

LexManos:
I'm thinking all of these 'canBeReplacedBy' functions should be scheduled for removal in 1.17. We need to re-think these as they are getting into the dalmations territory. IIRC they were intially introduced because of isAir. Which is going away in 1.17. So these can go as well.
On top of this, there are a ton of ~~bugs~~'features' that people use that depend on these glitchy mechanics in vanilla. 

Oh, this isn't pullable because you don't keep the separation of the two functions.
Trees and Mushrooms have different mechanics you can't delegate one to the other if you're separating them like this.

mrp-v2:
I've now updated this PR to only change the behavior of the methods to match vanilla, fix the erroneous javadocs, and mark the methods as deprecated for removal in 1.17.

sciwhiz12:
Closing as the `IForgeBlock` methods (and associated patches and `IForgeBlockState` redirects) were removed during the 1.17 update and patch cycle, rendering this 1.17-targeted PR obsolete.

PR #7490, opened 2020-11-13
[1.16.x] General review of patches, focused on style

mrp-v2:
I looked at each patch file, mostly looking for unnecessary whitespace changes. A few patches fixing old MC bugs that are now fixed in vanilla have also been removed, after checking that their removal did not cause the bug to reappear.

PR #7503, opened 2020-11-20
[1.16] Fix PointOfInterestType not being added to blockstate map when registered

mrp-v2:
When vanilla registers a `PointOfInterestType`, its blockstates are added to the map that is used to find POI's in the world. The forge registry does not add blockstates to that map. This PR resolves this by making the map a slave map (as suggested by tterrag) and using callbacks to add blockstates to it.

This will (probably) not break mods currently adding POIs because the method they are (probably) using to register their POI blockstates to the map now does nothing.

### DIFF
--- a/patches/minecraft/net/minecraft/village/PointOfInterestType.java.patch
+++ b/patches/minecraft/net/minecraft/village/PointOfInterestType.java.patch
@@ -9,3 +9,33 @@
     private static final Supplier<Set<PointOfInterestType>> field_234168_y_ = Suppliers.memoize(() -> {
        return Registry.field_218370_L.func_201756_e().map(VillagerProfession::func_221149_b).collect(Collectors.toSet());
     });
+@@ -36,7 +36,7 @@
+    }).filter((p_234173_0_) -> {
+       return p_234173_0_.func_177229_b(BedBlock.field_176472_a) == BedPart.HEAD;
+    }).collect(ImmutableSet.toImmutableSet());
+-   private static final Map<BlockState, PointOfInterestType> field_221073_u = Maps.newHashMap();
++   private static final Map<BlockState, PointOfInterestType> field_221073_u = net.minecraftforge.registries.GameData.getBlockStatePointOfInterestTypeMap();
+    public static final PointOfInterestType field_221054_b = func_226360_a_("unemployed", ImmutableSet.of(), 1, field_221071_s, 1);
+    public static final PointOfInterestType field_221055_c = func_226359_a_("armorer", func_221042_a(Blocks.field_222424_lM), 1, 1);
+    public static final PointOfInterestType field_221056_d = func_226359_a_("butcher", func_221042_a(Blocks.field_222423_lL), 1, 1);
+@@ -112,16 +112,14 @@
+    }
+ 
+    private static PointOfInterestType func_221052_a(PointOfInterestType p_221052_0_) {
+-      p_221052_0_.field_221075_w.forEach((p_234169_1_) -> {
+-         PointOfInterestType pointofinteresttype = field_221073_u.put(p_234169_1_, p_221052_0_);
+-         if (pointofinteresttype != null) {
+-            throw (IllegalStateException)Util.func_229757_c_(new IllegalStateException(String.format("%s is defined in too many tags", p_234169_1_)));
+-         }
+-      });
+       return p_221052_0_;
+    }
+ 
+    public static Optional<PointOfInterestType> func_221047_b(BlockState p_221047_0_) {
+       return Optional.ofNullable(field_221073_u.get(p_221047_0_));
+    }
++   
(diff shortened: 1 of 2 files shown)


src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -613,6 +618,38 @@ public void onCreate(IForgeRegistryInternal<Feature<?>> owner, RegistryManager s
...
+            Map<BlockState, PointOfInterestType> map = owner.getSlaveMap(BLOCKSTATE_TO_POINT_OF_INTEREST_TYPE, Map.class);
+            if (oldObj != null)
+            {
+                oldObj.getBlockStates().forEach(map::remove);
+            }
+            obj.getBlockStates().forEach((state) -> 
+            {
+                if (map.put(state, obj) != null)
+                {
+                    throw Util.pauseDevMode(new IllegalStateException(String.format("%s is defined in too many tags", state)));
```
diesieben07: We are not vanilla, might want to make this message a bit more useful and not use `Util.pauseDevMode`.

src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -615,6 +620,38 @@ public void onCreate(IForgeRegistryInternal<Feature<?>> owner, RegistryManager s
...
+                }
+            });
+        }
+
+        @Override public void onClear(IForgeRegistryInternal<PointOfInterestType> owner, RegistryManager stage)
```
diesieben07: Formatting: Put the annotation on its own line like everywhere else.

src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -615,6 +620,38 @@ public void onCreate(IForgeRegistryInternal<Feature<?>> owner, RegistryManager s
...
+        {
+            owner.getSlaveMap(BLOCKSTATE_TO_POINT_OF_INTEREST_TYPE, Map.class).clear();
+        }
+
+        @Override public void onCreate(IForgeRegistryInternal<PointOfInterestType> owner, RegistryManager stage)
```
diesieben07: Formatting

src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -77,6 +74,7 @@
 import net.minecraftforge.common.world.ForgeWorldType;
 import net.minecraftforge.event.RegistryEvent;
 import net.minecraftforge.event.RegistryEvent.MissingMappings;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
```
diesieben07: Why?

patches/minecraft/net/minecraft/village/PointOfInterestType.java.patch
```diff
@@ -9,3 +9,33 @@
...
+@@ -112,16 +112,14 @@
+    }
+ 
+    private static PointOfInterestType func_221052_a(PointOfInterestType p_221052_0_) {
+-      p_221052_0_.field_221075_w.forEach((p_234169_1_) -> {
+-         PointOfInterestType pointofinteresttype = field_221073_u.put(p_234169_1_, p_221052_0_);
+-         if (pointofinteresttype != null) {
+-            throw (IllegalStateException)Util.func_229757_c_(new IllegalStateException(String.format("%s is defined in too many tags", p_234169_1_)));
+-         }
+-      });
```
diesieben07, line 30: I liked the previous solution of `if (false)`, better. Its a smaller patch and you can add a comment explaining _why_ this is disabled.

diesieben07 (changes requested):

patches/minecraft/net/minecraft/village/PointOfInterestType.java.patch
mrp-v2: I opted to delete the code, so that if the code changes in a future Minecraft version, the patch breaks. I choose this because in #7490 it seemed lex prefers patches that will break if the vanilla code changes (see [this](https://github.com/MinecraftForge/MinecraftForge/pull/7490#discussion_r523277438) comment).

src/main/java/net/minecraftforge/registries/GameData.java
mrp-v2: Silly IDE formatting...

src/main/java/net/minecraftforge/registries/GameData.java
mrp-v2: Now explains which POI's conflict and why.

patches/minecraft/net/minecraft/village/PointOfInterestType.java.patch
diesieben07: Fair enough

diesieben07 (approved):

PR #7509, opened 2020-11-22
Fix crash when mods add goals to skeleton horses (MC-206338)

malte0811:
This turned up on ForgeCraft yesterday. The goal is modifying the goals, which works in vanilla because it's the last goal in the `LinkedHashSet`, but breaks with mods when additional goals are added (e.g. [here](https://github.com/TeamPneumatic/pnc-repressurized/blob/094bd99549d0111d829589be096624c20eee4894/src/main/java/me/desht/pneumaticcraft/common/event/EventHandlerPneumaticCraft.java#L138-L150), the added complexity was an attempt to fix this crash).  
[Vanilla report](https://bugs.mojang.com/browse/MC-206338) (edit: the report contains Mojang-mapped names).


desht:
Had some discussion with malte0811 about this (I added the original "fix" in PNC code which appeared to work at the time, but which I realise now is not the correct fix).  His analysis and fix is correct, in my opinion.

Vanilla gets away with modifying a collection being iterated, only because the element in question is the last in the collection, and it is `Iterator#next()` which does the CME test (specifically in this case, `HashMap.HashIterator#nextNode()`).  When mods are present, the element may no longer be the last, and that will trigger a CME.

diesieben07:
This is a very specialized workaround for a more generic problem (modifying goal list while ticking goals). We could instead patch `GoalSelector` to know whether it is updating goals currently, and if so, do a special handling in `addGoal` and `removeGoal` (store the modifications in a separate list then, which is processed after all goals have been ticked.

malte0811:
Is the generic problem relevant for any other vanilla goal? I'm generally not an entity-person, but so far this crash has only occurred with this specific goal (but twice for this one). And in modded goals I would say that the author is responsible for not modifying the goal list while the goal is running. Your suggestion would be a much larger patch, so I don't really want to add it without a good reason.

diesieben07:
Is there any possibility delaying the task by one tick will cause a change in behavior?

malte0811:
I guess it technically can/does, if you trigger a trap in its last tick in vanilla, the "original" horse stays, with the current change it doesn't. I'll take a look at how large the patches for your suggestion would be.

diesieben07 (approved):

LexManos:
So, your removal logic is a little wrong according to the vanilla remove function, you need to reset the tasks.
Then there is also the issues related to the OTHER streams over the goal list that could cause the CME.
So either this needs to be changed to wrap each of the streams in the tick event in this ticking flag. {As resetTask/shouldExecute/startExecuting could trigger the issue} Or we need to change getRunningGoals to return a snapshot. But that has performance implications. 

Doing the generic fix opens a lot more edge cases where we'd have issue.
If the goal list was inaccessible from outside GoalSelector it'd be one thing, but it's not. So we would have to take care of everyone else.

I'm actually more a fan of the original fix, where it just fixes the vanilla task that breaks things. This puts the responsibility of broken things onto the mod/whatever that breaks it and not on the system itself.

malte0811:
> I'm actually more a fan of the original fix, where it just fixes the vanilla task that breaks things.

So you're fine with the (minor) inconsistency with vanilla I mentioned above?
 > So, your removal logic is a little wrong according to the vanilla remove function, you need to reset the tasks.

Probably irrelevant if we're throwing out the new fix completely, but I think it's correct, since it uses the vanilla removal function when processing the `asyncRemoved` set. Or are you talking about more complex scenarios (e.g. goals being removed and re-added in the same async batch)?

LexManos:
> So you're fine with the (minor) inconsistency with vanilla I mentioned above?

Can you delay the entire content of the update instead of just the removal of the task? 
Can you explain this inconsistency a bit more? Is it duplicating horses.. or is it that one horse exists for a tick longer then it should?
 
> Probably irrelevant if we're throwing out the new fix completely, but I think it's correct, since it uses the vanilla removal function when processing the `asyncRemoved` set. Or are you talking about more complex scenarios (e.g. goals being removed and re-added in the same async batch)?

Humm maybe I read it wrong, go to love SRG patches y.y But ya it's moot as I don't want the generic fix as it opens a lot more holes.

malte0811:
>  Can you explain this inconsistency a bit more? Is it duplicating horses.. or is it that one horse exists for a tick longer then it should?

Neither. If you look at `SkeletonHorseEntity#livingTick` it first runs the super method (which includes goal ticking) and then removes the horse if it is a trap and has been a trap for at least 15 minutes.  
In vanilla if you trigger a trap in the last tick of those 15 minutes the horse will already be "untrapped" when the second part runs, so the horse will not be removed.  
With my initial patch `setTrap` will be called later in the server tick, so the horse is still marked as a trap when the removal check runs and the horse is removed.

 > Can you delay the entire content of the update instead of just the removal of the task?

So replace `TriggerSkeletonTrapGoal#tick` by
```java
ServerWorld serverworld = (ServerWorld)this.horse.world;
serverworld.getServer().enqueue(() ->
if (this.horse.removed) return;
/*existing code (not indented for patch size)*/
});
```
That should work, or at least not result in obviously different behavior even if it is triggered in the last tick. The trap time would be one tick lower, but I would guess that's acceptable?

PR #7513, opened 2020-11-23
Fix DimensionType not encode into registries in dedicated server saves

### LINKED ISSUE #7510: 1.16.4 - 35.1.0 : Ender Dragon does not spawn on a dedicated server.
Minecraft Version: 1.16.4
Forge Version: 35.1.0.
Logs: N/A (this isn't a crash and doesn't show up anywhere on logs when I read them).

This is a rather straight forward issue and a major one that is... Essentially, the Ender Dragon boss just doesn't spawn on a dedicated online server with Forge. The game and server still runs just fine but she just doesn't spawn. Her little bedrock pedestal doesn't spawn. Everything else in the End works. The Obsidian pillars spawn with the End Crystals and technically the Ender Dragon herself does work if you /summon her but she will have no health bar and defeating her like this doesn't spawn an exit portal or gateway, which means it is pointless.

She works as intended in singleplayer at least but in an online forge server...? She simply doesn't exist. I originally thought this was a mod until I tested my server with no mods installed, just Forge and yeah... She isn't there.

IzzelAliz:
This is introduced in #7445. When starting a (first run) server, the DimensionType instance in ServerWorld is not in the `DynamicRegistries`. I want to get the registry key of the dimType then find out this issue.

I print the instances in two registries with following code(breakpoint inside World.<init>, this.dimensionType is the world's type)

```java
for (Map.Entry<RegistryKey<Dimension>, Dimension> entry : ((ServerWorldInfo) info).getDimensionGeneratorSettings().func_236224_e_().getEntries()) {
     System.out.println(entry.getKey());
     System.out.println(entry.getValue());
     System.out.println(entry.getValue().getDimensionType());
}
System.out.println(this.dimensionType);
for (Map.Entry<RegistryKey<DimensionType>, DimensionType> entry : minecraftServer.func_244267_aX().func_230520_a_().getEntries()) {
     System.out.println(entry.getKey());
     System.out.println(entry.getValue());
}
```

And here are results, the first is the world copied from client saves(a valid save):
with level.dat
And this is from Forge 35.1.0 dedicated server:
without level.dat

The two different saves have different level.dat. The valid save one has encoded its dimension type to registry:
dat1
The corrupted one has not:
dat2

Passing a `WorldGenSettingsExport` instead of `WorldSettingsImport` could address this, which encodes instances to registry ResourceLocation.

Seems this fix #7510

### DIFF
--- a/patches/minecraft/net/minecraft/server/Main.java.patch
+++ b/patches/minecraft/net/minecraft/server/Main.java.patch
@@ -60,7 +60,7 @@
              }
  
 +            // Forge: Deserialize the DimensionGeneratorSettings to ensure modded dims are loaded on first server load (see SimpleRegistryCodec#decode). Vanilla behaviour only loads from the server.properties and deserializes only after the 2nd server load.
-+            dimensiongeneratorsettings = DimensionGeneratorSettings.field_236201_a_.encodeStart(worldsettingsimport, dimensiongeneratorsettings).flatMap(nbt -> DimensionGeneratorSettings.field_236201_a_.parse(worldsettingsimport, nbt)).getOrThrow(false, errorMsg->{});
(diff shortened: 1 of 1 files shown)


Cyborgmas:
Oof mb indeed, this does seem to be the proper fix. Thanks!

50ap5ud5 (approved):

PR #7517, opened 2020-11-24
Add PersistentForcedChunksLoaded event

pupnewfster:
Added a world event for PersistentForcedChunksLoaded so that mods can know when vanilla loads any saved forced chunks from memory and also load their custom tickets. The reason mods may have custom chunk loading tickets they keep track of that are persistent is so that then if two mods force load the same chunk, when one stops it properly allows the other one to keep having it force loaded as the ticket types are different. Previously I (and presumably other mod devs) just did this in `WorldEvent.Load`, except it turns out this is too early to safely do it as the rough order of events are:

```java
WorldEvent.Load
...
loadInitialChunks(...) {
    ...
    serverchunkprovider.registerTicket(TicketType.START, new ChunkPos(blockpos), 11, Unit.INSTANCE);
    while(serverchunkprovider.getLoadedChunksCount() != 441) {
        ...
    }
    ....
    for (ServerWorld serverworld1 : this.worlds.values()) {
        ... //Load any persistent forced chunks for the world
        WorldEvent.PersistentForcedChunksLoaded //(new as of this PR)
    }
    ....
}
```

The issue with doing it at `WorldEvent.Load`, is that depending on how many mods are doing this to add their own forced chunks, the number of loaded chunks may greatly exceed the `441` chunks vanilla waits for to load as the spawn, and then that while loop never gets exited.

### DIFF
--- a/patches/minecraft/net/minecraft/server/MinecraftServer.java.patch
+++ b/patches/minecraft/net/minecraft/server/MinecraftServer.java.patch
@@ -33,7 +33,15 @@
        ServerWorld serverworld = this.func_241755_D_();
        field_147145_h.info("Preparing start region for dimension {}", (Object)serverworld.func_234923_W_().func_240901_a_());
        BlockPos blockpos = serverworld.func_241135_u_();
-@@ -566,6 +569,7 @@
+@@ -480,6 +483,7 @@
+                serverworld1.func_72863_F().func_217206_a(chunkpos, true);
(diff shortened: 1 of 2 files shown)


Player3324:
Fixing the "is the spawn area loaded" check instead would be less error prone, replacing the "are there 441 chunks loaded in total" part with World.isAreaLoaded.

PR #6091, opened 2019-08-26
Fixed entity navigation to stop entities spinning

Wyn-Price:
Fixed an issue with entities spinning around on the last path node, and not being able to complete the path. This problem especially occurs to entities with large hitboxes, and is caused by the target position, and the checked position being different. I've changed entity offset in the path navigators to match the math in `Path#getVectorFromIndex`, to make the target and checked position the same. ~~This is disabled by default and can be turned on with the config option `fixEntityNavigation`.~~

For testing, I modified the vanilla Zombie entitiy to have a width of 2, and for it's wanding AI to be 4 times as likely.

Before:
forgespinning-before

After:
forgespinning-after

Related issue: https://bugs.mojang.com/browse/MC-94054

PR #7520, opened 2020-11-26
Re-implement part of pull 6091 - Fixed entity navigation to stop entities spinning

Corosauce:
https://github.com/MinecraftForge/MinecraftForge/pull/6091

The issue persists in 1.16.4 with entities of a larger AABB (tested 1.0 width). Updated Wyn-Prices fix for the land path following part factoring in the minor changes to the vanilla code.

The vanilla water path following (SwimmerPathNavigator) seems to be taking a different approach in 1.16.4 now so I left that part out, though I am unsure if water path following still has the issue still or not.

https://bugs.mojang.com/browse/MC-94054

### DIFF
--- a/patches/minecraft/net/minecraft/pathfinding/PathNavigator.java.patch
+++ b/patches/minecraft/net/minecraft/pathfinding/PathNavigator.java.patch
@@ -1,10 +1,11 @@
 --- a/net/minecraft/pathfinding/PathNavigator.java
 +++ b/net/minecraft/pathfinding/PathNavigator.java
-@@ -208,9 +208,9 @@
+@@ -208,9 +208,10 @@
        Vector3d vector3d = this.func_75502_i();
        this.field_188561_o = this.field_75515_a.func_213311_cf() > 0.75F ? this.field_75515_a.func_213311_cf() / 2.0F : 0.75F - this.field_75515_a.func_213311_cf() / 2.0F;
        Vector3i vector3i = this.field_75514_c.func_242948_g();
 -      double d0 = Math.abs(this.field_75515_a.func_226277_ct_() - ((double)vector3i.func_177958_n() + 0.5D));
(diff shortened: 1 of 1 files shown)


patches/minecraft/net/minecraft/pathfinding/PathNavigator.java.patch
```diff
@@ -0,0 +1,14 @@
+--- a/net/minecraft/pathfinding/PathNavigator.java
++++ b/net/minecraft/pathfinding/PathNavigator.java
+@@ -208,9 +208,9 @@
+       Vector3d vector3d = this.func_75502_i();
+       this.field_188561_o = this.field_75515_a.func_213311_cf() > 0.75F ? this.field_75515_a.func_213311_cf() / 2.0F : 0.75F - this.field_75515_a.func_213311_cf() / 2.0F;
+       Vector3i vector3i = this.field_75514_c.func_242948_g();
+-      double d0 = Math.abs(this.field_75515_a.func_226277_ct_() - ((double)vector3i.func_177958_n() + 0.5D));
++      double d0 = Math.abs(this.field_75515_a.func_226277_ct_() - ((double)vector3i.func_177958_n() + (this.field_75515_a.func_213311_cf() + 1) / 2D));
```
pupnewfster: This line, and the other changed line should both have a comment at the end of them along the lines of `//Forge: Fix MC-94054`

Corosauce: Ah right right, forgot to re-add that, fixed.

pupnewfster: No, add them on the same line to keep the patch size down, and add it to both of them

SmellyModder (approved):

PR #7521, opened 2020-11-28
Fix forge config option for default world type not applying unless you change the default world type

alcatrazEscapee:
Fixes an oversight in #7448.

Basically, when the options screen is created, the current selected dimension settings is passed to the screen, along with the preset. Forge replaces the preset but not the default settings, meaning if you don't change the world type, it uses the default settings (contrary to expectation, it does not create the settings when leaving the screen - it creates them every time you cycle the button)


gigaherz:
OOF, completely missed the fact the screen object was receiving an already-constructed value.

PR #7522, opened 2020-11-28
Added ability to do custom rain particle rendering

rehwinkel:
This makes it possible for modders to further customize rain/weather effects for their mods. Specifically it enables rendering ground splash particles and playing rain/weather sounds.


LexManos:
We really need to rethink all of these render handlers. Give ways to access the vanilla code, provide some system that isn't last person wins'

PR #7527, opened 2020-12-01
Make UnboundedMapCodec more lenient in decoding dimensions.

### LINKED ISSUE #7526: Removing Datapack Dimensions causes Nether and End Dimensions to be permanently deleted
**Minecraft Version:** 1.16-1.16.4 (Present in all 1.16.x versions)

**Forge Version:** 1.16.4-35.1.7

**Description of issue:**

The critical vanilla bug, https://bugs.mojang.com/browse/MC-197860 is still present in current versions of Minecraft.

This significantly affects Forge mods more than vanilla because users will be removing and adding new mods that contain dimensions written as a datapack.

I have investigated the cause of the issue and have found the following:

### How the Nether and End Dimensions are Removed

The removal of the nether and end relate to the:
•	Dimension Registration Order
•	Logic used to remove erroring dimensions in SaveFormat#getServerLifecyclePair 

**Dimension Registration Order**

When the Dynamic<T> dynamic = nbt.get("WorldGenSettings").orElseEmptyMap();  in SaveFormat#getServerLifecyclePair is being constructed, the order of dimension registration is as follows:
1.	Overworld
2.	Datapack loaded custom dimensions
3.	Nether
4.	End

**Logic used to remove erroring Dimensions**

Another user who looked into the issue correctly identified that the DimensionGeneratorSettings CODEC (GeneratorOptions.CODEC in Yarn names) will return missing dimensions.

However, that is only half of the story. The Dynamic<T> that is parsed into the DimensionGeneratorSettings CODEC originates from SaveFormat#getServerLifecyclePair.

When the custom dimension is removed, DimensionGeneratorSettings.CODEC fails to fully parse the Dynamic<T>, so the DataResult<DimensionGeneratorSettings> only return the Either#Right part of the result, which only includes the Overworld.

The crucial aspect is the Pair that is returned to the method uses the DataResult#resultOrPartial method, whose PartialResult removes ALL objects AFTER the erroring object, in addition to the erroring object. 
i.e. The data result removes the erroring custom dimension, the Nether and the End, leaving only the Overworld.
 
When these two factors are put into effect, the subsequent ServerWorldInfo, which defines the settings for the World, will use the incorrect data.

### Why the Nether and End Dimensions do not show after reinstating the missing datapack

When the world is loaded after the datapack is reinstated, by this time the Dynamic<T> no longer includes the nether and end in its list of dimensions. Hence, only the Overworld is present alongside the custom dimension.

### Potential Solutions

So far, I have not found a suitable solution. Some ideas I've had are:

- Patching the DataResult<DimensionGeneratorSettings> with a value returned from a method inside ForgeHooks. The method in ForgeHooks could remove only the erroring object, but not other ones like the Nether and End. 
 I'm not sure about the logic that should be used for this, and am not familiar enough yet with Mojang's Serialisation system to come up with the fix.
- Modifying the order of dimension registration so that the Overworld, Nether and End are registered before Datapack dimensions. This would require a larger and more complex patch, which I don't think is optimal.

Feel free to point out corrections, ask questions and discuss this issue. Thank you for your time.

alcatrazEscapee:
This may not be the shortest fix, but in lieu of actually patching something in DFU (e.g. see https://github.com/Mojang/DataFixerUpper/pull/55), this is (I think) the best solution to be had here. I have tested this with the example data pack TelepathicGrunt posted in the vanilla issue tracker and removing a datapack dimension removed the dimension, and only that dimension (the nether and end were still intact.)

Inspecting the call path for codec construction you can see that the `SimpleRegistry` patch is only invoked from the codec in `DimensionGeneratorSettings`.

Having a copy of `UnboundedMapCodec` was basically necessary as the class is `final`, and this is the simplest way to get at the actual underlying issue. Dealing with this after is impossible as the errors have already been discarded, and dealing with it before (i.e. as outlined in the issue report) is very painful, as the `Dynamic` as input to the codec will have to be initially parsed and understood without actually using the codec of which it's designed to do, erroring entries (read: can't just remove dimensions that don't exist, have to remove any entry that may error which basically involves parsing it via the codec, but without that.) So yeah, a lot of code that has potential for issue or other effect. (see: enderdragon not spawning :P )

Fixes #7526 / [MC-197860](https://bugs.mojang.com/browse/MC-197860)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -949,6 +949,7 @@ project(':forge') {
         exclude 'net/minecraftforge/fml/common/versioning/InvalidVersionSpecificationException.java'
         exclude 'net/minecraftforge/fml/common/versioning/Restriction.java'
         exclude 'net/minecraftforge/fml/common/versioning/VersionRange.java'
+        exclude 'net/minecraftforge/common/LenientUnboundedMapCodec.java'
 
         tasks {
             main {
--- a/patches/minecraft/net/minecraft/util/registry/SimpleRegistry.java.patch
+++ b/patches/minecraft/net/minecraft/util/registry/SimpleRegistry.java.patch
@@ -1,0 +1,12 @@
+--- a/net/minecraft/util/registry/SimpleRegistry.java
++++ b/net/minecraft/util/registry/SimpleRegistry.java
+@@ -212,7 +212,8 @@
+    }
+ 
+    public static <T> Codec<SimpleRegistry<T>> func_241745_c_(RegistryKey<? extends Registry<T>> p_241745_0_, Lifecycle p_241745_1_, Codec<T> p_241745_2_) {
+-      return Codec.unboundedMap(ResourceLocation.field_240908_a_.xmap(RegistryKey.func_240902_a_(p_241745_0_), RegistryKey::func_240901_a_), p_241745_2_).xmap((p_239656_2_) -> {
++      // FORGE: Fix MC-197860
++      return new net.minecraftforge.common.LenientUnboundedMapCodec<>(ResourceLocation.field_240908_a_.xmap(RegistryKey.func_240902_a_(p_241745_0_), RegistryKey::func_240901_a_), p_241745_2_).xmap((p_239656_2_) -> {
+          SimpleRegistry<T> simpleregistry = new SimpleRegistry<>(p_241745_0_, p_241745_1_);
+          p_239656_2_.forEach((p_239653_2_, p_239653_3_) -> {
+             simpleregistry.func_218381_a(p_239653_2_, p_239653_3_, p_241745_1_);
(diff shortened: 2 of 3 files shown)


50ap5ud5:
Thanks for making this! I can confirm your solution solves the issue.

Here were my findings:
- Vanilla already forces player to overworld if dimension is missing
- PR preserves vanilla dimensions
- PR also preserves working modded dimensions

Overall, this PR meets the Phase 1 requirements outline by Lex as per conversation on Discord:
image


More information about the testing process and related files:
https://drive.google.com/file/d/1TNqRLxnF9YVujFEvU3STZPzDwUuTUo0-/view?usp=drivesdk

50ap5ud5 (approved):

LexManos:
This is acceptable for phase 1, thanks 50ap5ud5 for actually doing what I asked and providing a test case.
I look forward to you guys digging into Phase 2/3.

PR #7528, opened 2020-12-01
[1.15.x] Add horizon hook

josephtarbit:
This adds a small method hook allowing a dimension to customise its horizon height. I would've added this to 1.16.x as well however the issue doesn't exist there due to the move to data driven dimensions.


valoeghese (approved):
simple and concise 👍 .

covers1624:
Sorry for the delay on this one.

Changes look good, I can definitely see valid uses for this.

PR #7529, opened 2020-12-02
Fix a few inconsistencies with capability invalidation

pupnewfster:
Made it so that `invalidateCaps` gets called when a tile unloads instead of only when a tile is removed, as after discussing it with Lex, it seems that it was overlooked. Adjusted the JavaDoc slightly on `LazyOptioanl#invalidate` to specify what constitutes a TE being invalidated, and also made it so that if someone calls `invalidate` on an already invalidated `LazyOptional` it doesn't fire the invalidation listeners again.


marchermans (approved):
This works.
Not much more to say about this really.

Issue #7519, opened 2020-11-25
[1.15/1.16] Adding an Entity to the same Chunk via `EntityJoinWorldEvent` during ChunkLoad


AterAnimAvis:
**Minecraft Version:** 1.16.4 (Also effects 1.15.2)
**Forge Version:** 1.16.4-35.1.4 (1.15.2-31.2.45)

**Logs:** N/A

**Steps to Reproduce:**
 1. Create a new world with https://github.com/AterAnimAvis/ForgeBugs/tree/create-entity-on-load
 2. Throw an item
 3. Exit the world
 4. Rejoin the world
 5. Deadlocks at 100%

**Description of issue:**
Adding an Entity during ChunkLoad causes a deadlock.

 1. `ChunkManager#func_219200_b` -> `ServerWorld#addEntityIfNotDuplicate` -> posts `EntityJoinWorldEvent`

That chunk is now waiting for EntityJoinWorldEvent to return.

 3. `EntityJoinWorldEvent` -> `ExampleMod#onEntityJoinWorld`
 4. `ExampleMod#onEntityJoinWorld` -> `ServerWorld#addEntity` -> `ServerWorld#addEntity0` -> `ServerWorld#getChunk` -> `ServerChunkProvider#getChunk()`
 5. `#getChunk()` -> `this.executor#driveUntil`

The `EntityJoinWorldEvent` is now waiting for our handler to return which is waiting on the `addEntity` which is waiting on the Chunk that is waiting on `EntityJoinWorldEvent`.

We are now in deadlock.

This can also be triggered by `IForgeItem#createEntity` via `ForgeInternalHandler#onEntityJoinWorld` if the upgrade from an `ItemEntity` to a new custom `Entity` occurs during ChunkLoad

Note I haven't had the time to fully debug this yet or try to come up with a potential fix.

Reminder for me: change the test mod so it triggers on initial world creation if possible.

Originally discovered/experienced by @FriederHannenheim (I like cats/Ich mag Katzen) on Discord

**Update:** Can occur during any ChunkLoad (Not specific to `MinecraftServer#loadInitialChunks` as originally reported)

pau101:
Just noting this issue was reported previously as #6770

AterAnimAvis:
Apologies, I found the bug fairly late in the night and completely forgot to check for existing Issues.

### First pass *"fix"* https://gist.github.com/AterAnimAvis/800eb413eac59434592387e5305bb5db
[...]

PR #7532, opened 2020-12-03
Fix Deadlock when adding Entities during Chunk Load (#7519)

AterAnimAvis:
NOTE: This PR was changed to only target the issues in forge's `IForgeItem`, this does not fix mods.

Original below:
---

Delays the firing of `EntityJoinWorldEvent` during chunk load until after the chunk is available.
This prevent deadlocks if a listener of the event tries to add an entity to the chunk being loaded.

- `ChunkEvent.Load` I've moved this so it still fires after Entity Removal (I'm not sure if this change is right).

Tested via https://github.com/MinecraftForge/MinecraftForge/issues/7519#issuecomment-735419698.

Seeing as we just saw someone have an issue with this in production (On Forge Discord), thought it was worth moving this to a PR.

A second look by someone who's had more experience with the `ChunkManager`'s code (and the mess of completablefutures) is probably worthwhile.
There's still probably a better fix available but this seems to work.

### DIFF
--- a/patches/minecraft/net/minecraft/world/server/ChunkManager.java.patch
+++ b/patches/minecraft/net/minecraft/world/server/ChunkManager.java.patch
@@ -8,11 +8,20 @@
                 }
  
                 this.func_219229_a(p_219185_5_);
-@@ -601,6 +_,7 @@
+@@ -582,6 +_,7 @@
+             if (this.field_219254_h.add(chunkpos.func_201841_a())) {
+                chunk.func_177417_c(true);
+                this.field_219255_i.func_147448_a(chunk.func_177434_r().values());
++               p_219200_1_.func_219276_a(ChunkStatus.field_222617_m, this).thenRun(() -> { //Forge: Delay EntityJoinWorld events until chunk is available to prevent deadlocks
+                List<Entity> list = null;
+                ClassInheritanceMultiMap<Entity>[] aclassinheritancemultimap = chunk.func_177429_s();
+                int i = aclassinheritancemultimap.length;
+@@ -601,6 +_,8 @@
                 if (list != null) {
                    list.forEach(chunk::func_76622_b);
                 }
 +               net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.world.ChunkEvent.Load(chunk));
++               });
              }
  
              return chunk;
--- a/patches/minecraft/net/minecraft/world/server/ChunkManager.java.patch
+++ b/patches/minecraft/net/minecraft/world/server/ChunkManager.java.patch
@@ -8,14 +8,14 @@
                 }
  
                 this.func_219229_a(p_219185_5_);
-@@ -582,6 +_,7 @@
+@@ -581,6 +_,7 @@
+             chunk.func_217318_w();
              if (this.field_219254_h.add(chunkpos.func_201841_a())) {
                 chunk.func_177417_c(true);
++               p_219200_1_.func_219276_a(ChunkStatus.field_222617_m, this).thenRun(() -> { //Forge: Delay until the chunk is available to prevent deadlocks
                 this.field_219255_i.func_147448_a(chunk.func_177434_r().values());
-+               p_219200_1_.func_219276_a(ChunkStatus.field_222617_m, this).thenRun(() -> { //Forge: Delay EntityJoinWorld events until chunk is available to prevent deadlocks
                 List<Entity> list = null;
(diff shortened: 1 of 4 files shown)


patches/minecraft/net/minecraft/world/server/ChunkManager.java.patch
```diff
@@ -8,23 +8,58 @@
...
+                      }
+                   }
+-               }
+ 
+-               if (list != null) {
+-                  list.forEach(chunk::func_76622_b);
+-               }
++                  if (list != null) {
++                     list.forEach(chunk::func_76622_b);
++                  }
```
pupnewfster: If you don't change the indentation of this inner stuff the patch will be a lot cleaner/smaller.

AterAnimAvis: Changed and pushed

pupnewfster: I think your change of this might have lost something? The description of the PR says you moved the load event, but it seems you are only adding a new one now, instead of also removing the old one.

AterAnimAvis: I think your reading the patch wrong. There's only one post of `ChunkEvent.Load`?

pupnewfster: Ya nevermind I was reading it wrong, because there was an extra line I think just because of all the lines earlier that I think was showing removed.

AterAnimAvis: ~~Literally woke up in the middle of the night and realised I missed a line from my fix. I’ll fix this in the morning (If I’m actually right)~~ - Thankfully the current patch doesn't need the fix (Prevents CME on iterating the entities for removal) which was necessary for a previous version.

IzzelAliz:
My opinion on this fix is to add a field to mark every `EntityJoinWorldEvent` from `addEntityIfNotDuplicate` as during world generation, and check this field in ForgeInternalHandler and call `addEntityIfNotDuplicate` instead of `addEntity`, and add them to chunk after this tick.
The `addEntityIfNotDuplicate` method is only called from ChunkManager and do not perform any chunk check operation, and I think it is only for world gen purpose.

Delay the transfer entity operation may cause many potential issue and may be not ideal.

AterAnimAvis:
Sorry for the late reply.

Your suggestion doesn't cover the underlying issue that the event gets fired before the chunk is available [1], whilst it would possible fix the issue with `IForgeItem#createEntity` it does nothing to handle the other cases (I.E. it would not fix the issues raised by my test case).

As I said previously:
> A second look by someone who's had more experience with the ChunkManager's code (and the mess of completablefutures) is probably worthwhile.
> There's still probably a better fix available but this seems to work.

Theres a test mod [provided here](https://github.com/AterAnimAvis/ForgeBugs/releases/tag/Forge7519-3) if you would like to easily test out my current proposed fix in production.

[1] - Discord `#squirrels`
> Unfortunately vanilla create the instances before they actually create the chunk they go into, hence the deadlock.. ideally we'd fix that so that the event only fires when the chunk is ready. - Curle

LexManos:
Ah, yes the CF that is chunk loading. So your change to move things to the next tick should be fine, It has a ton of potential side effects... Which is really annoying. There hasn't been really much feed back on this so i don't know if this is possible to pull without breaking a lot of mods.
However I do have one thing to note on it. You should probably move the tile entities into the delay as well. Because it goes all the way down into the chain of things that could call the world to get the chunk and cause a lock.

AterAnimAvis:
Updated.

A potential alternative would be to only pr a fix for the `IForgeItem`, add Javadocs with warnings in the necessary places and keeping this on hold till the next breaking changes phase (if it still applies then).

I've provided an updated test mod https://github.com/AterAnimAvis/ForgeBugs/releases/tag/Forge7519-4, which should allow people to test out the proprosed changes in production easily (Literally just drop the jar into your mods folder).

LexManos:
> A potential alternative would be to only pr a fix for the IForgeItem, add Javadocs with warnings in the necessary places and keeping this on hold till the next breaking changes phase (if it still applies then).

Honestly i'm more of a fan of that. As there are a lot of side effects with moving more and more to a delayed state.

AterAnimAvis:
I've migrated to fixing only the issue in `ForgeInternalHandler`.

Also added a small notice to both `ChunkEvent.Load` and `EntityJoinWorldEvent` about the potential deadlock issue.

PR #7534, opened 2020-12-03
Set WM_CLASS and WM_INSTANCE_NAME for early progress window

### LINKED ISSUE #7531: [Linux/X11] Forge messes up Minecraft Window Class breaking Window Manager Rules
**Minecraft Version:** 1.16.4

**Forge Version:** 35.1.0

**Logs:** N/A

**Steps to Reproduce:**
 1. Start Minecraft with Forge
 2. Read out window properties with `xprop`

**Description of issue:**
The `WM_CLASS` property reads "FML early loading progress", where it would be "Minecraft 1.16.4" for Vanilla Minecraft. This makes breaks any window manager rules targeting Minecraft windows and makes it impossible to differentiate based on the Minecraft version.

**Notes:**
When using Minecraft 1.14.4 with Forge 28.2.0, the window class starts out as "FML early loading progress", but then correctly switches to "Minecraft 1.14.4". This switch does not happen with the versions listed above.

diesieben07:
Emulate what we would get without the early progress window.
Vanilla Minecraft never sets these properties, so GLFW defaults to the first window title, which is "Minecraft* \<version\>".  Emulate this here by setting the properties explicitly to avoid GLFW using the early progress window title.

Fixes #7531.

### DIFF
--- a/src/fmllauncher/java/net/minecraftforge/fml/loading/FMLLoader.java
+++ b/src/fmllauncher/java/net/minecraftforge/fml/loading/FMLLoader.java
@@ -185,11 +185,12 @@ public class FMLLoader
         naming = commonLaunchHandler.getNaming();
         dist = commonLaunchHandler.getDist();
         production = commonLaunchHandler.isProduction();
-        progressWindowTick = EarlyProgressVisualization.INSTANCE.accept(dist, commonLaunchHandler.isData());
+
+        mcVersion = (String) arguments.get("mcVersion");
+        progressWindowTick = EarlyProgressVisualization.INSTANCE.accept(dist, commonLaunchHandler.isData(), mcVersion);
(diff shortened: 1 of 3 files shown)


LexManos:
Aside from all the binary breaks after a RB its fine.

alcatrazEscapee:
Correct me if I'm wrong - the only binary break I see is the signature of `EarlyProgressVisualization#accept`, adding a `String mcVersion` param.

Everything else is either private / package private: `Visualization` is package private, so changing the signature of `start` wouldn't be a break as nothing could be implementing it outside of this anyway, `ClientVisualization` is package private so changing the (public) method signatures there is fine, same with `NoVisualization`. Other method signatures that were changed are private.

diesieben07:
Fixed the binary break for `EarlyProgressVisualisation#accept`

PR #7535, opened 2020-12-04
Fix a few places where the vanilla method was being called instead of the forge one

pupnewfster:
- Moved default implementation of `IForgeBlock#getSlipperiness` method from `Block` to `IForgeBlock`
- For `IForgeBlock#isLadder`: Vanilla only checks against the climable tag now instead of also hardcoding specific types, so I believe if anything the strict overrides could cause issues if someone tries to remove them as "ladders" say on a vanilla server but the client still thinks it is a valid ladder.
- Fixed the following methods that had some calls not being switched to hooks properly:
 - Blocks:
  - `getLightValue` call for checking if mobs can spawn
  - `getSlipperiness` call from item entities
 - Items:
  - MerchantOffer calling stack insensitive max stack size

It also seems `observedNeighborChange`, not quite sure what it was used for or how to hook it back up. Should I mark it as deprecated and for removal in 1.17?


nikita488:
The `getFogColor`, `rotate` and `onPlantGrow` methods inside `IForgeBlock` and `IForgeBlockState` also never called like `observedNeighborChange`.

diesieben07:
I tracked `observedNeighborChange` all the way back to 1.12. This method used to be special to just observer blocks, Forge patched it to fire for all blocks. This method is now present for all blocks in vanilla and called `updatePostPlacement`. So yes, `observedNeighborChange` can be deprecated in Forge and removed in 1.17.

diesieben07:
The `getFogColor` patch probably got removed, because fog rendering was moved into its own class from `EntityRenderer`. `FogRenderer` should be updated to re-add the patch.

`rotate` is a forge extension and adds world and position. However the places where vanilla calls it don't have that information necessarily, so they call the deprecated version. Probably should make the parameters nullable / add information about them not necessarily being correct (sometimes rotate is called before the block is placed).

`onPlantGrow` probably can't be re-added, due to the move away from hardcoded handwritten world generation code. It was called during tree generation, etc. So another candiate for deprecation.

pupnewfster:
> `rotate` is a forge extension and adds world and position. However the places where vanilla calls it don't have that information necessarily, so they call the deprecated version. Probably should make the parameters nullable / add information about them not necessarily being correct (sometimes rotate is called before the block is placed).

This part feels relatively close to a breaking change to me given if people currently are assuming it is not null and we start passing null their code would break, so should I just put a TODO for it for 1.17?

Edit: I also am not quite sure how I would go about reintegrating the getFogColor method as I am not quite sure how FogRenderer even works, and from a brief glance I am pretty sure the default implementation will need to change/maybe even the params being passed to it.

marchermans (approved):

PR #7536, opened 2020-12-06
Override invalidateCaps instead of remove for vanilla TE capability invalidation

pupnewfster:
Addresses a small oversight of mine in https://github.com/MinecraftForge/MinecraftForge/commit/1a539ffebcbc3d9eba46cd489066c228a4e6fc2a where I did not realize that forge's patched in capabilities for vanilla TEs was hardcoded to happening in `remove` instead of in `invalidateCaps`.

### DIFF
--- a/patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
@@ -481,7 +481,7 @@
     }
  
     public boolean func_70094_T() {
-@@ -3082,6 +3138,60 @@
+@@ -3082,6 +3138,58 @@
        this.func_213361_c(p_213334_1_ == Hand.MAIN_HAND ? EquipmentSlotType.MAINHAND : EquipmentSlotType.OFFHAND);
     }
  
@@ -531,12 +531,10 @@
 +   }
 +
 +   @Override
-+   public void remove(boolean keepData) {
-+      super.remove(keepData);
-+      if (!keepData) {
-+         for (int x = 0; x < handlers.length; x++)
-+            handlers[x].invalidate();
-+      }
++   protected void invalidateCaps() {
++      super.invalidateCaps();
++      for (int x = 0; x < handlers.length; x++)
++         handlers[x].invalidate();
 +   }
 +
     @OnlyIn(Dist.CLIENT)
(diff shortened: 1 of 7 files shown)


patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
```diff
@@ -531,12 +531,10 @@
...
 +
 +   @Override
-+   public void remove(boolean keepData) {
-+      super.remove(keepData);
-+      if (!keepData) {
-+         for (int x = 0; x < handlers.length; x++)
-+            handlers[x].invalidate();
-+      }
++   protected void invalidateCaps() {
++      super.invalidateCaps();
```
marchermans, line 535: Why was the keepData check removed here?

patches/minecraft/net/minecraft/entity/item/minecart/ContainerMinecartEntity.java.patch
```diff
@@ -20,11 +20,10 @@
  
 -      super.func_70106_y();
 +      super.remove(keepData);
-+      if (!keepData) itemHandler.invalidate();
```
marchermans, line 23: Again why was this check removed?

patches/minecraft/net/minecraft/entity/passive/horse/AbstractHorseEntity.java.patch
```diff
@@ -43,9 +43,9 @@
 +   }
 +
 +   @Override
-+   public void remove(boolean keepData) {
-+      super.remove(keepData);
-+      if (!keepData && itemHandler != null) {
++   protected void invalidateCaps() {
++      super.invalidateCaps();
++      if (itemHandler != null) {
```
marchermans, line 48: Again removes the keepData check

marchermans (approved):
Generally approved.
Could you document why the keepData checks where removed.

patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
pupnewfster: `invalidateCaps` for Entities is called from `Entity#remove` when `keepData` is false. This extra check got removed as `invalidateCaps` only will be called when it should be invalidated.

patches/minecraft/net/minecraft/tileentity/LockableTileEntity.java.patch
```diff
@@ -10,16 +10,16 @@
 +      return new net.minecraftforge.items.wrapper.InvWrapper(this);
 +   }
 +
-+   @javax.annotation.Nullable
++   @javax.annotation.Nonnull
```
LexManos: This annotation can just be removed. It was Nullable to override the vanilla's package level ones. Back before the cap optional rewrite.

marchermans (approved):

PR #7538, opened 2020-12-07
Add a SoundType subclass that uses suppliers

sciwhiz12:
### The Chicken-and-Egg Problem
One of the difficulties in using custom `SoundType`s for modded blocks is the faux chicken-and-egg problem of `Block`, `SoundEvent`, and `SoundType`.

To add a custom sound type for a block, you have two options: override a method (which offers great flexibility in what sound is used, at the cost of _needing to subclass_), or passing the `SoundType` to the `AbstractBlock.Properties` constructor. Since the block properties is needed to construct the `Block` instance, the `SoundType` instance needs to be constructed before the block.

To construct a `SoundType` instance, you need to pass in 5 `SoundEvent` instances: the place sound, the break sound, the hit sound (when breaking the block), the step sound, and the fall sound (when falling on the block AFAIK).

The faux chicken-and-egg problem is thus: `SoundEvent`s must be registered. However, the `SoundEvent`s must also be available to construct the `SoundType`, which must be constructed before the `Block` using it is registered. However, `Block`s are loaded/registered wayy before `SoundEvent`s<sup id="1top">[1](#1)</sup>.

### Two Possible Solutions
The first solution would be to statically initialize the `SoundEvent`s and register them later. This is not preferrable due to the use of static initialization which is stated to be something that Forge is discouraging. Plus, this prevents the use of the `DeferredRegister` system (since `RegistryObject` errors when retrieving the inner object before the corresponding registry is loaded).

The other solution, which fits Forge's goal of reloadable registries, is extending `SoundType` to accept `Supplier<SoundEvent>`s instead. This is the more preferred and cleaner of the two, since it both does not need to use any sort of static initialization for `SoundEvent`, and allows the use of `DeferredRegister` and `RegistryObject`s (since `RegistryObject`s are `Supplier`s too).

### This Pull Request's Solution
However, there is no such subclass of `SoundType` in Forge, which means that mods who have custom `SoundType`s needs to all make their own subclass, and it is not obvious to new modders on how to add custom sound types to their block using `RegistryObject`s, meaning they will ask the same question again and again.

This PR remedies this deficiency, by **adding a new `ForgeSoundType` class under `net.minecraftforge.common.util`**. This class is a subclass of `SoundType`, which accepts `Supplier<SoundEvent>` instead of `SoundEvent` in the constructor.
 - `null` is passed into the superclass constructor, because no "blank/empty" `SoundEvent` exists, and using a predefined sound event may be confusing in the future if a bug ever occurs. This value should never be exposed, since the only access would be the getters, which are overriden by `ForgeSoundType`.
   - The casts were added by a review by LexManos, for the purpose of making sure that if Mojang were to change the parameters of the original `SoundType` constructor, this class's constructor will cause a compile-time error.
 - The constructor for `SoundType` was marked with `@Deprecated`, and a regular comment<sup id="2top">[2](#2)</sup> was placed alongside it, pointing modders to use `ForgeSoundType` for suppliers instead.
 - This PR includes a test mod (`custom_sound_type_test`) which was used to verify that `ForgeSoundType` works as expected.
    * The test mod adds a new block, the `custom_sound_type_test:test_block` which has a custom `ForgeSoundType` and `SoundEvent`, all of which point to the `entity/mob/zombie/hit1.ogg` sound (hitting, falling, breaking, placing, steps).

Checking each of the calls to the 5 getters will reveal that they are called during ingame, way after the sound events are registered. This means it should be safe to use this class with `RegistryObject`s.

#### TL;DR: _New class to Forge named `ForgeSoundType`, a replacement for `SoundType` that uses `Supplier<SoundEvent>` instead of `SoundEvent` directly in the constructor._

<sup><a id="1" href="#1top">^</a> **1:** Registry load order is currently defined as: `minecraft:block` first, `minecraft:item` second, then all other registries in alphabetical order according to their name (so `forge:*` goes first before `minecraft:*`).</sup>
<sup><a id="2" href="#2top">^<a/> **2:** I followed the established precedent of using regular comments instead of documentation comments.</sup>


src/main/java/net/minecraftforge/common/util/ForgeSoundType.java
```diff
@@ -0,0 +1,80 @@
...
+
+package net.minecraftforge.common.util;
+
+import net.minecraft.block.SoundType;
+import net.minecraft.util.SoundEvent;
+
+import javax.annotation.Nonnull;
+import java.util.function.Supplier;
+
+public class ForgeSoundType extends SoundType
```
pupnewfster, line 50: Might be worth throwing some java docs on this class that gives a brief summary of why this exists (given you wrote a very detailed thing for the PR explaining it you can probably just copy paste parts of that here)

sciwhiz12: Added a condensed/summarized version of the PR description as javadocs to the `ForgeSoundType` class. This should give modders more insight into why it exists.
(I thought about linking this PR's URL in the `@see` section, but decided against it for now. If modders are curious, they could always `git blame`/search on GitHub to find the commit and this PR).

patches/minecraft/net/minecraft/block/SoundType.java.patch
```diff
@@ -0,0 +1,10 @@
+--- a/net/minecraft/block/SoundType.java
++++ b/net/minecraft/block/SoundType.java
+@@ -3,6 +3,7 @@
+ import net.minecraft.util.SoundEvent;
+ import net.minecraft.util.SoundEvents;
+ 
++// Forge: Use {@link net.minecraftforge.common.util.ForgeSoundType} instead for suppliers
```
LexManos: I would move this comment to the Constructor, and then @Deprecate the constructor. As the class itself isn't deprecated, it's the new instances that are.

src/main/java/net/minecraftforge/common/util/ForgeSoundType.java
```diff
@@ -0,0 +1,102 @@
...
+{
+    private final Supplier<SoundEvent> breakSound;
+    private final Supplier<SoundEvent> stepSound;
+    private final Supplier<SoundEvent> placeSound;
+    private final Supplier<SoundEvent> hitSound;
+    private final Supplier<SoundEvent> fallSound;
+
+    public ForgeSoundType(float volumeIn, float pitchIn, Supplier<SoundEvent> breakSoundIn, Supplier<SoundEvent> stepSoundIn, Supplier<SoundEvent> placeSoundIn, Supplier<SoundEvent> hitSoundIn, Supplier<SoundEvent> fallSoundIn)
+    {
+        super(volumeIn, pitchIn, null, null, null, null, null);
```
LexManos: I know this sounds dumb, but can you add (SoundEvent) casts to all these nulls?
The other option is writing some form of dev-time validation that this class has the same methods/constructors as the parent.

The idea is to make it a error if Mojang changes any of the argument types.
Thus notifying us that we need to update this class.

patches/minecraft/net/minecraft/block/SoundType.java.patch
sciwhiz12: Moved the comment, and deprecated the constructor for `SoundType`.

src/main/java/net/minecraftforge/common/util/ForgeSoundType.java
sciwhiz12: Adding casts like that does makes sense, once the idea behind it is known. I've added the casts to the `null`s.

diesieben07 (approved):

PR #7544, opened 2020-12-11
Make shear interactions extend to subclasses

pupnewfster:
Makes it so less code has to be replicated in onItemUse checking the block used against to make custom shears function properly with some of the custom interactions vanilla has.


marchermans (approved):

LexManos:
We really need to write a generic `blah == Items.FOO` -> `blah instanceof FooItem` transformer. 
You should work on that @pupnewfster

PR #7545, opened 2020-12-11
Add generated resource tracking to ExistingFileHelper

tterrag1098:
Improve the interoperability of data generators by making `ExistingFileHelper` act as a central cache of all generated resources.

Data providers must now call a new method `trackGenerated` whenever a file *will* be generated in the future. Thus this method acts like a promise to generate the specific file later, which is reflected in the documentation.

Also added is a bit of convenience to avoid repetition of magic strings in `IResourceType`/`ResourceType`, which are just holders for a resource pack type (client resources / server data), a suffix (extension) and prefix. This encapsulates the three parameters that are passed to the current version of `exists` and avoids repeating them throughout the code since the same values must now be passed to `trackGenerated`

Finally this removes the hack in `ModelProvider` of using a special delegating `ExistingFileHelper` that refers to internal state, which was always quite awkward as an API since you needed to pass that instance to your item model provider (see https://github.com/MinecraftForge/MinecraftForge/commit/a57a240c134e5fc7294d4fdae3b283b5660598fd). This new system means that any generated data will be globally tracked and known about from any other data provider.

### DIFF
--- a/patches/minecraft/net/minecraft/data/TagsProvider.java.patch
+++ b/patches/minecraft/net/minecraft/data/TagsProvider.java.patch
@@ -1,11 +1,12 @@
 --- a/net/minecraft/data/TagsProvider.java
 +++ b/net/minecraft/data/TagsProvider.java
-@@ -28,10 +28,18 @@
+@@ -28,10 +28,20 @@
     protected final DataGenerator field_200433_a;
     protected final Registry<T> field_200435_c;
     protected final Map<ResourceLocation, ITag.Builder> field_200434_b = Maps.newLinkedHashMap();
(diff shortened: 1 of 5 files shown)


marchermans (approved):
This looks good.
Would it be possible for you to supply an example of how to use this properly in a gist or here in the PR, so that the Forge CookBook/Wiki can be updated and so that it can be used as an example in discord?

tterrag1098:
There's really not much to "use" for the average modder, this should just be a drop-in fix for the problem. The only time you need to interact with ExistingFileHelper directly is when making a custom data provider, which is pretty rarely done.

PR #7548, opened 2020-12-11
Add hook to allow modders to more easily control if endermen can place blocks on top of their blocks

pupnewfster:
This PR exposes a method in `IForgeBlock` to control if an Enderman can place a given block on top of a modder's block. My use case is to be able to deny Endermen from placing blocks on some of my blocks without having to listen to the full on block place event of all entities, though another thing this PR makes possible is for modders to relax the place requirements on their blocks to allow Endermen to place blocks on top of theirs even if it is "air" or opaque.


marchermans (approved):
A bit weird to test this, since a test mod did not make sense.

But it seems to be working, I created a endvoid world with a custom block and just bedrock and checked if they could place on my block.

Seems to be working.

pupnewfster:
Updated this PR based on comments from lex after discussing it on discord. A brief summary:

Original implementation is too specific, and is just asking for methods like `canMyChinchillaPlaceThisDirtOnTopOfYourBlockOnTuesday` to end up being added. Given the goal of the PR is to allow mods to blacklist blocks from being able to have endermen place things on them (similar to how vanilla hardcodes bedrock as invalid) and given the logic is server side only, tags fit well into this use case. This PR now adds a block tag `forge:enderman_place_on_blacklist` that allows modders to add their blocks to this blacklist.

PR #7549, opened 2020-12-12
Only apply the MC-68754 fix on Windows, closes #7546

### LINKED ISSUE #7546: Forge for 1.12.x crashes when exiting fullscreen on MacOS, when ran on a JDK compiled with newer versions of XCode (Oracle or OpenJDK 8u261 or higher).
**Minecraft Version:** 1.12.x (1.12 to 1.12.2)

**Forge Version:** 14.21.1.2403 to the most recent releases of Forge for 1.12.x (14.23.5.2854).

**Logs:** https://gist.github.com/NeRdTheNed/a037ab7e6357be7da5f8d37f4f166641

**Steps to Reproduce:**
 1. Download a JDK compiled with a newer version of Xcode (Xcode 4.5 was used until recently), and install it on a computer running MacOS. As Oracle and the OpenJDK community have started compiling JDK releases with Xcode 10.1, this can be achieved by downloading an offical Oracle or OpenJDK build of JDK 8u261 or newer. _(As proof that this is actually a XCode compilation related issue, Amazon's JDK 8 fork (Corretto 8) **does not** suffer from the same issues due to [having a different build process](https://github.com/corretto/corretto-8/blob/master/src/common/autoconf/toolchain.m4) which still uses Xcode 4, despite being based on the latest versions of OpenJDK (tested with Corretto 8u275))_. GraalVM 20.3.0 EE was used during testing, as well as AdoptOpenJDK 8u275-b01.
 2. Download and install Forge 14.21.1.2403 or higher (these issues still occur on Forge 14.23.5.2854. MultiMC was used during testing for version management, but this issue was also replicated on the official launcher).
 3. Run Forge with the newly installed JDK, enter fullscreen, then exit fullscreen. Installing any mod is not required to trigger this issue, it is 100% reproducible with just Forge.

**Description of issue:**

There is a somewhat complicated cause for this issue. The first thing I noticed when I encountered this issue was that the crash was reported to have happened during native MacOS execution, in liblwjgl.dylib:

```
2020-12-11 20:14:43.650 java[16985:258430] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'NSWindow drag regions should only be invalidated on the Main Thread!'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff403fbacd __exceptionPreprocess + 256
	1   libobjc.A.dylib                     0x00007fff6aaffa17 objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff40415629 -[NSException raise] + 9
	3   AppKit                              0x00007fff3d9ba5ca -[NSWindow(NSWindow_Theme) _postWindowNeedsToResetDragMarginsUnlessPostingDisabled] + 317
	4   AppKit                              0x00007fff3dabd43f -[NSThemeFrame setStyleMask:] + 137
	5   AppKit                              0x00007fff3dabd292 __25-[NSWindow setStyleMask:]_block_invoke + 1896
	6   AppKit                              0x00007fff3d972970 NSPerformVisuallyAtomicChange + 132
	7   AppKit                              0x00007fff3dabcadf -[NSWindow setStyleMask:] + 184
	8   liblwjgl.dylib                      0x0000000135e7bcca Java_org_lwjgl_opengl_MacOSXDisplay_nSetResizable + 90
	9   ???                                 0x0000000116691c87 0x0 + 4670954631
	10  ???                                 0x00000001166812bd 0x0 + 4670886589
)
libc++abi.dylib: terminating with uncaught exception of type NSException
Process crashed with exitcode 6.
```

...which is the worst sort of message you can get in a Java stack trace. I tracked the crash down to [this exact line of code in LWJGL 2](https://github.com/LWJGL/lwjgl/blob/master/src/native/macosx/org_lwjgl_opengl_Display.m#L627). By hacking together and compiling LWJGL 2 for newer versions of Mac's developer tools (don't do this, I wasted hours of my life trying to get it to compile) and running Forge with the resulting built native libraries (this was a huge pain, please don't bother doing this), I found that commenting out that line "fixed" the problem (I broke other things getting it to compile). You can find my very very bad hacks to make it build [here](https://github.com/NeRdTheNed/lwjgl/tree/temp-fix-dontuse), but they are not portable and will not compile outside of my laptop (though the first commit should give you a general idea of how to fix the build errors).

So, it’s a LWJGL bug, right? **No, because vanilla 1.12.x Minecraft does not crash when exiting fullscreen**. So what broke it? By running every version of Forge for 1.12 until my game started crashing when exiting fullscreen (less of a waste of time, GraalVM is amazing), I tracked the code causing the crash down in Forge to probably being [this line](https://github.com/MinecraftForge/MinecraftForge/blob/0215e2e128c8446e72a46431c6f8349cc8b1ddfe/patches/minecraft/net/minecraft/client/Minecraft.java.patch#L201), introduced by [this commit](https://github.com/MinecraftForge/MinecraftForge/commit/0215e2e128c8446e72a46431c6f8349cc8b1ddfe#diff-a23157af13640392a15eb64829751d419f2f25f1a02726e4d586ff1852f515aa), to fix [this bug](https://bugs.mojang.com/browse/MC-68754). A TL;DR of the code is that when fullscreen is exited, Forge toggles the ability to resize the window off, and then back on again, in order to work around a LWJGL bug where the window wouldn’t have been allowed to be resized.

So, why does this break everything? It's complicated. MacOS sort of forces developers to do all rendering / window (or NSWindow, as the crash log notes) modification related stuff in the "main" / "first" thread (I don't personally code native Mac applications, so I don't know much about this). Apparently, in older versions of MacOS's development tools, there was less stringent checks on this occurring in certain areas (I don't hate myself enough to find out more). By toggling the ability of the game window to be resized, _Forge is very likely inadvertently causing the NSWindow to be modified from a non-"main" thread_ (from the stack trace, the NSWindow might be updating window visuals as a result of it being resizable / non-resizable). The LWJGL code I hacked out is _safe to run form the main thread_, so **LWJGL is not the issue: the issue is when and where LWJGL's native interface code is called**. As a result of compiling with a newer version of Xcode, this bug, which was previously lying dormant, was suddenly revealed, and the full wrath of Apple was brought down mercilessly upon anyone who just wanted to play modded Minecraft with updated, faster, and safer JDK 8 builds.

Suggested fix: ~~replace the patch with something that communicates to the "main" thread that the resisability of the window should be toggled, and have the main thread toggle the resisability of the window. If any further testing is requested or questions require answers, I'll be asleep for a while, but feel free to leave comments on this issue.~~ I've come up with a much easier solution, can’t believe I didn’t think of this earlier. This only occurs as a result of trying to fix the LWJGL bug, and the LWJGL bug only happens on Windows. Much simpler fix: only toggle the resisability of the window on Windows. ~~PR nearly done, just finishing testing.~~ PR submitted (#7549).

As a side note, this issue might heavily affect Apple silicon users in the future, as if they are using JDKs compiled for Apple's M1 ARM architecture, they would be using newly compiled JDKs.

I really hate MacOS.

NeRdTheNed:
MC-68754 only affects Minecraft on Windows, as per the comments on Mojang's bug tracker, and the corresponding LWJGL bug report (https://github.com/LWJGL/lwjgl/issues/142). This change simply checks if the OS running Minecraft is Windows before applying the fix. This was tested by me on Windows 10, MacOS Mojave, and Lubuntu (5.8.0-31), and all platforms allowed the window to be resized as expected after exiting fullscreen. As the MC-68754 fix was the cause of #7546, this issue is fixed by applying this patch.

### DIFF
--- a/patches/minecraft/net/minecraft/client/Minecraft.java.patch
+++ b/patches/minecraft/net/minecraft/client/Minecraft.java.patch
@@ -223,7 +223,7 @@
              }
  
              Display.setFullscreen(this.field_71431_Q);
-+            if (!this.field_71431_Q) //Forge: Fix MC-68754, Screen is not resizeable after exiting fullscreen due to LWJGL bug https://github.com/LWJGL/lwjgl/issues/142 which is fixed, but not in the version MC ships
++            if (!this.field_71431_Q && (Util.func_110647_a() == Util.EnumOS.WINDOWS)) //Forge: Fix MC-68754, Screen is not resizeable after exiting fullscreen due to LWJGL bug https://github.com/LWJGL/lwjgl/issues/142 which is fixed, but not in the version MC ships
 +            {
 +                Display.setResizable(false);
 +                Display.setResizable(true);


NeRdTheNed:
@tterrag1098 Sorry to ping you, as I known you've already seen this PR! Just wanted to know if there's anything on my end that's preventing this from being merged. If this PR is low priority or requires review from a team member who doesn't have time to review it right now, that's totally fine, but if there's anything I can do to speed the process up, please let me know!

PR #7543, opened 2020-12-11
Add support for stack sensitive food

pupnewfster:

PR #7552, opened 2020-12-12
Fix getFood

GoogeTan:
I did it: https://forums.minecraftforge.net/topic/91558-1152-edit-minecraft-source-code-or-more-simply-hooks/?tab=comments
GetFood method now has an item stack

### DIFF
--- a/patches/minecraft/net/minecraft/item/Item.java.patch
+++ b/patches/minecraft/net/minecraft/item/Item.java.patch
@@ -22,6 +22,9 @@
     }
  
     public void func_219972_a(World p_219972_1_, LivingEntity p_219972_2_, ItemStack p_219972_3_, int p_219972_4_) {
+@@ -108,1 +112,1 @@
+-         if (p_77659_2_.func_71043_e(this.func_219967_s().func_221468_d())) {
++         if (p_77659_2_.func_71043_e(this.getFood(itemstack).func_221468_d())) {
 @@ -120,10 +124,12 @@
        return this.func_219971_r() ? p_77654_3_.func_213357_a(p_77654_2_, p_77654_1_) : p_77654_1_;
     }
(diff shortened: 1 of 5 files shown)


ChampionAsh5357:
Already a pr #7543

patches/minecraft/net/minecraft/item/Item.java.patch
```diff
@@ -121,6 +127,17 @@
...
++   @Deprecated
+   @Nullable
+   public Food func_219967_s() {
+      return this.field_219974_q;
+   }
++
++   @Nullable
++   public Food getFood(ItemStack stack) {
++      return this.field_219974_q;
++   }
```
pupnewfster, line 140: a. New methods like this should be in IForgeItem
b. This is a breaking change as if any mod overrides `getFood` this then doesn't gracefully fallback to that override

patches/minecraft/net/minecraft/item/Item.java.patch
```diff
@@ -121,6 +127,17 @@
     public boolean func_219970_i(ItemStack p_219970_1_) {
        return p_219970_1_.func_77973_b() == Items.field_222114_py;
     }
+@@ -306,4 +349,10 @@
++   @Deprecated
```
pupnewfster, line 131: When we add deprecation in forge to vanilla methods we put a comment that says what to use instead.

pupnewfster (commented):
I am going to close this PR given #7543 is already open, assigned and more complete, and does not have any merge conflicts (though I did notice from looking at this that a couple patch files I made didn't commit which I fixed now). But for future reference if you make further PRs I went ahead and reviewed this so you know how some things work/that would need to be changed.

Additionally some of your patch files look a bit weird with end of file line changing, and not as much surrounding code being added almost as if you edited them manually (which is a no go).

PR #7554, opened 2020-12-16
[1.16] Added multi part entity support

brandon3055:
Effectively a simplified version of #6701

This pr expands the vanilla logic used by the Ender Dragon and opens it up so other can make similar multipart entities. This is done via two new methods on `IForgeEntity` `isMultipartEntity` and `getParts`, All of the vanilla logic for handling dragons has been switched to these methods, There is now also a `PartEntity` which is extended by `EnderDragonPartEntity`.

Uses for this include creating large entities which require custom sub hit boxes. An obvious example being custom dragons as its now no longer possible to extend the vanilla dragon.


covers1624 (approved):
Patches are concise, clean, and minimal.

PR #7555, opened 2020-12-17
[1.16.x] Fixed the PotentialSpawnEvent to match functionality from previous versions.

### LINKED ISSUE #7309: [1.16.2] Setting mob spawns with WorldEvent.PotentialSpawns doesn't work.
Minecraft Version: 1.16.2

Forge Version: 33.0.22

Steps to Reproduce:
 1. Register a listener to `WorldEvent.PotentialSpawns`
 2. Modify the spawn list to only contain a non-vanilla spawn entry.
 3. Observe as no mobs are able to spawn.

Description of issue:
When trying to set mob spawns using `WorldEvent.PotentialSpawns`, any mobs set that are not the default for the area in question are not spawned. This is caused by `WorldEntitySpawner.func_234976_a_` returning false if the mob spawner in question is not one of the normal ones for the area, preventing mobs from spawning where it gets called. A simple fix would be to return true here instead of doing the check since, as far as I can tell, it has no apparent actual function. Of course an event could be added to let modders validate spawns manually.

Here's a code block for an event if you want to copy and paste it for testing.
`public void potentialSpawnsEvent(WorldEvent.PotentialSpawns event)
{
	event.getList().clear();
	event.getList().add(new MobSpawnInfo.Spawners(EntityType.ZOMBIFIED_PIGLIN, 1, 4, 4));
}`

GoryMoon:
This fix is equivalent to what this part of the patch changed in 1.15.x and is missing in 1.16.x. https://github.com/MinecraftForge/MinecraftForge/blob/60eb1c25e9f01fc4ccc1c1bd9d63fbae813b037d/patches/minecraft/net/minecraft/world/spawner/WorldEntitySpawner.java.patch#L52-L57 
The doc for the event has been updated to point to where it's actually fired from now.
I've added a test for the event to verify that it works as intended, it will try to replace all spawns with zombified piglins even if the biome normally doesn't allow them to spawn there.
This fixes #7309

### DIFF
--- a/patches/minecraft/net/minecraft/world/spawner/WorldEntitySpawner.java.patch
+++ b/patches/minecraft/net/minecraft/world/spawner/WorldEntitySpawner.java.patch
@@ -27,7 +27,7 @@
                                return;
                             }
  
-@@ -253,6 +255,7 @@
+@@ -253,12 +255,13 @@
           return null;
        } else {
           List<MobSpawnInfo.Spawners> list = func_241463_a_(p_234977_0_, p_234977_1_, p_234977_2_, p_234977_3_, p_234977_5_, biome);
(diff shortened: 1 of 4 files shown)


LexManos:
Honestly I think this event should be removed, we have data driven biomes and spawnlists now. The known uses for this event was to add mobs to other biomes.
What is the usecase for doing this in code still? Is there a better way to do what you want that isn't a global event like this?

GoryMoon:
Don't have a real need for it as I add my mobs to the biomes I want with the `BiomeLoadingEvent` atm, but only after I noticed the `PotentialSpawnEvent` didn't work fully from testing/debugging as that was the most straightforward way of adding spawns based on previous versions. I found that you could add spawning from the `BiomeLoadingEvent` from other mods when I got stuck on this not working fully.
I can't say what others have for usecase of this event, would need input from people that had that special use previously and hasn't found another way to resolve it.

GoryMoon:
Tagging the creators of the two issues related to this to see if they still have a usecase of this event that isn't solved with what's available today. @SilverDavidMC @McJty 
Like I said I don't have a specific usecase atm as I've found another way with `BiomeLoadingEvent`, just to get a discussion if needed. If non there might not be much of use anymore, also as this hasn't worked in 1.16.x most probably found another way.

LexManos:
Gunna hold off on pulling this until people get some justifications in.
I'd rather mark it as deprecated and remove it in 1.17, as I'm convinced we already cover the usecases with our other events.

sciwhiz12:
Hello! @GoryMoon, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

GoryMoon:
As this hasn't worked in 1.16.x people have most likely moved on to use `BiomeLoadingEvent` and the fact that it got removed in the 1.17 port I think this can be closed.
Also as no one has justified why this should be in as Lex was waiting for it's safe to say it's not needed anymore.
The related issue could also be closed I guess.

PR #7560, opened 2020-12-21
Fix entity properties not being sent to vanilla clients at all

diesieben07:
Sorry, my `SEntityPropertiesPacket` filter was utterly broken and just sent out packets with a 0 entity ID.
This causes vanilla clients to not see attribute changes at all.

### DIFF
--- a/src/main/java/net/minecraftforge/network/VanillaConnectionNetworkFilter.java
+++ b/src/main/java/net/minecraftforge/network/VanillaConnectionNetworkFilter.java
@@ -20,6 +20,7 @@
 package net.minecraftforge.network;
 
 import java.util.AbstractMap;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


src/main/java/net/minecraftforge/network/VanillaConnectionNetworkFilter.java
```diff
@@ -84,7 +85,7 @@ public static void injectIfNecessary(NetworkManager manager)
     @Nonnull
     private static SEntityPropertiesPacket filterEntityProperties(SEntityPropertiesPacket msg)
     {
-        SEntityPropertiesPacket newPacket = new SEntityPropertiesPacket();
+        SEntityPropertiesPacket newPacket = new SEntityPropertiesPacket(msg.getEntityId(), Collections.emptyList());
```
ichttt, line 88: Won't using `Collections.emptyList` throw an `UnsupportedOperationException` when adding the snapshots further down the line? Shouldn't we use a new ArrayList?

diesieben07: `SEntityPropertiesPacket` always initializes its `snapshot` field as `Lists.newArrayList`. This constructor parameter is only used in a for loop to fill that snapshots list. However as we already have a list of snapshots here (and not a list of `ModifiableAttributeInstance`, I pass in an empty list so the constructor does not add anything to the snapshots list - we will do that later manually.

ichttt: You are right, for some reason I misread the code there.

PR #7563, opened 2020-12-21
Add supplier variant to `AbstractBlock.Properties#lootFrom`

alcatrazEscapee:
Forge patches `Block` / `AbstractBlock`s loot table handling to use a supplier of the loot table name. However, there's no way to directly set this through the properties a. la. vanilla - even though `lootFrom` internally uses a supplier, you have to pass a `Block` into it.

This adds a proper overload of `lootFrom` which takes a `Supplier<? extends Block>` and lazily maps it to a `Supplier<ResourceLocation>` as is required for the loot table, so modders can now use `lootFrom` without having to worry about creating a registration order dependency.

### DIFF
--- a/patches/minecraft/net/minecraft/block/AbstractBlock.java.patch
+++ b/patches/minecraft/net/minecraft/block/AbstractBlock.java.patch
@@ -144,7 +144,7 @@
        public AbstractBlock.Properties func_200941_a(float p_200941_1_) {
           this.field_200961_i = p_200941_1_;
           return this;
-@@ -933,7 +971,7 @@
+@@ -933,10 +971,15 @@
        }
  
        public AbstractBlock.Properties func_222379_b(Block p_222379_1_) {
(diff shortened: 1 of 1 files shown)


diesieben07 (approved):

ChampionAsh5357:
I would also deprecate the original `lootFrom` method and leave a comment or javadoc reminding users to use the supplier version instead.

alcatrazEscapee:
Ah bugger, I meant to add that deprecation + comment but evidently forgot, thanks for pointing it out :)

patches/minecraft/net/minecraft/block/AbstractBlock.java.patch
```diff
@@ -144,12 +144,22 @@
...
           return this;
-@@ -933,7 +971,7 @@
+@@ -932,11 +970,17 @@
+          return this;
        }
  
++      @Deprecated // FORGE: Use the variant that takes a Supplier below
        public AbstractBlock.Properties func_222379_b(Block p_222379_1_) {
 -         this.field_222381_j = p_222379_1_.func_220068_i();
 +         this.lootTableSupplier = () -> p_222379_1_.delegate.get().func_220068_i();
```
LexManos, line 154: Should of been implemented as a `return lootFrom(p_222379_1_.delegate);` but whatever.

PR #7566, opened 2020-12-23
Fix crash when beehive is broken by fake player

SquidDev:
When a beehive is broken, every nearby bee targets a random player. However, if there are no nearby players, the game crashes.

This should not occur under normal (vanilla) conditions. However, if a beehive is broken by a fake player (or potentially, a player with an extended reach attribute), there are no players in range and so we see a crash.

**Edit:** Apologies for the force-push spam. Used the wrong git config to commit.


artheus (approved):
LGTM :+1:

PR #7569, opened 2020-12-27
Add ForgeHooks.canEntitySpawn to instances of ISpecialSpawner

ShroomAgent27:
This can be used to allow mods to cancel Phantom spawn events.

An example of this would be TorchMaster which dynamically blocks entity spawning if they are in range of a MegaTorch block

https://github.com/Xalcon/TorchMaster/blob/e1a5fd037fd3a1677109d7146376b01bc7f9ef9f/src/main/java/net/xalcon/torchmaster/common/logic/entityblocking/EntityBlockingEventHandler.java#L22-L46

No edits would need to be made to TorchMaster's code to start using this hook.

I'm not too sure if this is the best way to do this, so let me know if there's a better way.

### DIFF
--- a/patches/minecraft/net/minecraft/world/spawner/CatSpawner.java.patch
+++ b/patches/minecraft/net/minecraft/world/spawner/CatSpawner.java.patch
@@ -1,0 +1,10 @@
+--- a/net/minecraft/world/spawner/CatSpawner.java
++++ b/net/minecraft/world/spawner/CatSpawner.java
+@@ -80,6 +80,7 @@
+       if (catentity == null) {
+          return 0;
+       } else {
++         if(net.minecraftforge.common.ForgeHooks.canEntitySpawn(catentity, p_221122_2_, p_221122_1_.func_177958_n(), p_221122_1_.func_177956_o(), p_221122_1_.func_177952_p(), null, SpawnReason.NATURAL) == -1) return 0;
+          catentity.func_213386_a(p_221122_2_, p_221122_2_.func_175649_E(p_221122_1_), SpawnReason.NATURAL, (ILivingEntityData)null, (CompoundNBT)null);
+          catentity.func_174828_a(p_221122_1_, 0.0F, 0.0F);
+          p_221122_2_.func_242417_l(catentity);
--- a/patches/minecraft/net/minecraft/world/spawner/PatrolSpawner.java.patch
+++ b/patches/minecraft/net/minecraft/world/spawner/PatrolSpawner.java.patch
@@ -1,0 +1,10 @@
+--- a/net/minecraft/world/spawner/PatrolSpawner.java
++++ b/net/minecraft/world/spawner/PatrolSpawner.java
+@@ -94,6 +94,7 @@
+          return false;
+       } else {
+          PatrollerEntity patrollerentity = EntityType.field_220350_aJ.func_200721_a(p_222695_1_);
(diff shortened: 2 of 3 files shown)


Xalcon:
Thanks for the headsup. Would be amazing it this gets merged.,

diesieben07:
The other `ISpecialSpawner` implementations have the same issue, they should all be addressed at once, imho.
In addition, ~this implementation is not correct~, since `canEntitySpawn` has _three_ results. If `1` is returned, the vanilla checks must not be done and the entity must be spawned regardless.

Edit: Implementation _is_ correct, because there are no additional checks here in vanilla. However it still needs to be considered for the other spawners.

ShroomAgent27:
I'll go ahead and edit this PR to add the hooks for the owner spawners too.

patches/minecraft/net/minecraft/world/spawner/CatSpawner.java.patch
```diff
@@ -0,0 +1,10 @@
+--- a/net/minecraft/world/spawner/CatSpawner.java
++++ b/net/minecraft/world/spawner/CatSpawner.java
+@@ -80,6 +80,7 @@
+       if (catentity == null) {
+          return 0;
+       } else {
++         if(net.minecraftforge.common.ForgeHooks.canEntitySpawn(catentity, p_221122_2_, p_221122_1_.func_177958_n(), p_221122_1_.func_177956_o(), p_221122_1_.func_177952_p(), null, SpawnReason.NATURAL) == -1) return 0;
```
ShroomAgent27, line 7: There's lots of other calculations in this class that are used to determine where to spawn a Cat. I didn't want to make a large patch to bypass that code, so I just stuck it in the `spawnCat` method.

patches/minecraft/net/minecraft/world/spawner/PatrolSpawner.java.patch
```diff
@@ -0,0 +1,10 @@
+--- a/net/minecraft/world/spawner/PatrolSpawner.java
++++ b/net/minecraft/world/spawner/PatrolSpawner.java
+@@ -94,6 +94,7 @@
+          return false;
+       } else {
+          PatrollerEntity patrollerentity = EntityType.field_220350_aJ.func_200721_a(p_222695_1_);
++         if(net.minecraftforge.common.ForgeHooks.canEntitySpawn(patrollerentity, p_222695_1_, p_222695_2_.func_177958_n(), p_222695_2_.func_177956_o(), p_222695_2_.func_177952_p(), null, SpawnReason.PATROL) == -1) return false;
```
ShroomAgent27, line 7: Same for PatrolSpawner. `func_230253_a_` is the main method that is used to determine if a Patrol can spawn. I just hooked into `spawnPatroller` at the very end to look at the end result if the game decides it wants to spawn a Patrol or not.

ShroomAgent27 (commented):
* Added for CatSpawner
* Added for PatrolSpawner
* NOT added for `VillageSiege`. It is already handled by `VillageSiegeEvent`
* NOT added for `WanderingTraderSpawner`. It calls `EntityType#spawn` which is already handled by `ForgeEventFactory#doSpecialSpawn`

ShroomAgent27:
I made an alternate branch that achieves the same goals of adding hooks to instances of `ISpecialSpawner` with a new event. You can view the commit here. https://github.com/shroomdog27/MinecraftForge/commit/6b87526b4b9b9a30939e39b3c09e0341842a5605  

If that branch is better for this purpose than this one, I'll close this PR and open a new one with that branch instead.

A test mod for that branch can be found here. <https://github.com/shroomdog27/TorchMaster/tree/forge/CheckSpecialSpawner> It is a fork of TorchMaster that uses this new event to dynamically cancel mobs from spawning if its in range of a Torch or dread lamp.

PR #7571, opened 2020-12-28
Remove armour stand double add to world

Mgazul:
After placing an ArmorStand, the console appears: 
`Trying to add entity with duplicated UUID *************. Existing minecraft:armor_stand#***, new: minecraft:armor_stand#***`


diesieben07:
First of all, this is a vanilla bug, so should be reported to the vanilla bugtracker. Any vanilla bugs Forge fixes need to be documented in the patch with the Minecraft ticket number.
Your patch is also unnecessarily large (could be a single deletion).

That said, I do not think this is something we should fix in Forge. This bug has no negative implications, except a warning in the log for a quite rare occurrence (someone placing an armor stand using an armor stand item).

PR #7573, opened 2020-12-29
Fix/Replace player visibility event

maxanier:
The player visibility event introduced with https://github.com/MinecraftForge/MinecraftForge/commit/f0d9bf7886e53fc02d8b9c445086eb1e343c2420 back in 1.10/12 has stopped working in 1.14+ because the vanilla code changed and the respective patch hasn't been adjusted/replaced.

Changes in Vanilla (from 1.12 to 1.16):
- Visiblity can be checked for any `LivingEntity` instead of just `PlayerEntity` now
- The check is no longer directly tied to `World#getNearestAttackablePlayer` or its successors. Instead it is part of `LivingEntity`

Changes made by this PR:
- migrate the event from `PlayerEvent` to `LivingEvent` and modified it to suit the new vanilla functionality.
- deprecate and mark old `PlayerEvent.Visiblity` with `//TODO 1.17 remove`, it did and does not have any effect.
-   deprecate the relevant `ForgeHooks` method, even though it can probably be removed right away.

### DIFF
--- a/patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
@@ -71,6 +71,15 @@
           this.func_82142_c(this.func_70644_a(Effects.field_76441_p));
        }
  
+@@ -747,7 +752,7 @@
+             d0 *= 0.5D;
+          }
+       }
+-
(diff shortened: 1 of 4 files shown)


LexManos:
Honestly, this event should just go away as nobody is using it. But, that's a breaking change so i'll try and remember it in the 1.17 update.

PR #7582, opened 2021-01-04
Improve RegistryKeyCodec decoding + add extra debug info

Cyborgmas:
The current problem is that when the list is a mix of registered objects and inlined objects, the error is not descriptive and misleading. 
It also leads to the list (in some cases) to contain only the unregistered/inlined features (otherwise it is empty). 

The fix is quite annoying, because the problem is the mix of behaviour between `EitherCodec`, `ListCodec` and `#promotePartial`.

The `EitherCodec` tries its first codec then fallsback to the second. However this means that any partial result is that of the 2nd codec at all times. This is important since the list of `ConfiguredFeature`s in `BiomeGenerationSettings` has `#promotePartial` on it. The result is that only the inlined objects are added to the list. 
Another issue is that the `ListCodec` stops adding results (partial or not) after the first encountered complete error (no partial result).
This means that the order of the list will change the returned partial result.
So switching the `EitherCodec` is not sufficient (also encoding relies on the current order).
So to remedy this, another list codec was needed with improved decoding logic to get a better partial result.
Then if the registered list codec errored, check first with the inlined codec and if that errored too, discard it and return the registered list codec with a promoted partial. This makes it so that a mixed list of registered and unregistered objects will always result in a decoded list with only the registered objects.

I have also added a few patch lines to improve some error messages, I think they can be extremely helpful but if it's not wanted I will remove them.

Current hold: I have not yet tested with a unregistered object added by the `BiomeLoadingEvent` as I could not get it to work easily. Once that is done, the pr will no longer be a draft

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -950,6 +950,7 @@ project(':forge') {
         exclude 'net/minecraftforge/fml/common/versioning/Restriction.java'
         exclude 'net/minecraftforge/fml/common/versioning/VersionRange.java'
         exclude 'net/minecraftforge/common/LenientUnboundedMapCodec.java'
+        exclude 'net/minecraftforge/common/util/ImprovedListCodec.java'
 
         tasks {
             main {
--- a/patches/minecraft/net/minecraft/util/registry/DynamicRegistries.java.patch
+++ b/patches/minecraft/net/minecraft/util/registry/DynamicRegistries.java.patch
@@ -9,3 +9,12 @@
        MutableRegistry<E> mutableregistry = p_243607_0_.func_243612_b(registrykey);
  
        for(Entry<RegistryKey<E>, E> entry : registry.func_239659_c_()) {
+@@ -141,7 +143,7 @@
+       });
+       DataResult<SimpleRegistry<E>> dataresult = p_243610_0_.func_241797_a_(simpleregistry, p_243610_2_.func_243622_a(), p_243610_2_.func_243623_b());
+       dataresult.error().ifPresent((p_243603_0_) -> {
+-         field_243598_a.error("Error loading registry data: {}", (Object)p_243603_0_.message());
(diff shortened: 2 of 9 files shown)


alcatrazEscapee:
Well this is interesting. One things stands out:

> Another issue is that the ListCodec stops adding results (partial or not) after the first encountered complete error (no partial result).

Your implementation of `ImprovedListCodec` seems to be exactly the same change made in #7527 to `BaseMapCodec` (and in theory, in https://github.com/Mojang/DataFixerUpper/pull/55 ), and could be a much simpler PR to DFU - although the  chance of it getting merged before 1.17 is low.

The root cause being DFU's approach to error recovery is on the whole... not great.

Cyborgmas:
Oh thanks for the info! Yeah was going to file an issue to DFU, but the PR now seems simpler. Had a hard time debugging `apply2stable` and the `ap`'s so steared clear of that for the fix but your change is closer to the og so will do that instead.

Changed it, but maintained my added "withPartials" not 100% sure it's needed, but I think it is better to have that controllable and use it for registry objects.

Cyborgmas:
Okay a bit more debugging and it seems to be working, tested with BYG which had a innate problem with its own CF.
Before the fix the logged error was:
`Not a json object: "freeze_top_layer"`
And the biome using the CF had an empty list in the corresponding layer:

byg_before_fix

With this fix the errror becomes:
`Error decoding object list: ["minecraft:freeze_top_layer",{"config":{"feature":"byg:black_ice_snow","decorator":{"config":{},"type":"minecraft:nope"}},"type":"minecraft:decorated"}] (Inline definitions not allowed here)`

And the biome will still contain the registered feature:

byg_after_fix

Note that if the ListCodec was already tracking errors a bit better, then, without the other fixes, the resulting list would contain the unregistered features.

LexManos:
This should be upstreamed, not us.

PR #7585, opened 2021-01-05
Add Render(Tile)EntityEvent

Meldexun:
This allows modders to skip rendering of entities.

An example would be [https://github.com/Meldexun/EntityCulling](https://github.com/Meldexun/EntityCulling). It's a 1.12 coremod and with these events I could update it to 1.16 without making it a coremod again.


DaemonUmbra:
Did you create these files in an IDE or on GitHub's webpage?

marchermans:
I think this won't get merged.
We are already fireing a good amount of events for different things, and this is such a niche request that I think this is not warrented, since this would fire for each TE and as such would probably cause more performance hurt then you could gain from it.

Additionally MC already does camera based culling of TEs that are outside of its rendering range, which is probably a good middle ground between BTTW Performance via Raytracing and not testing them at all, with regards to complexity and benefits of the complexity both in terms of code maintanance as well as actual game performance.

If you can come back with numbers and show us that the event is useable we might consider adding this.
But for now I will close this PR.

Meldexun:
> Did you create these files in an IDE or on GitHub's webpage?

Kind of. I created the classes in my mod workspace and then used the github webpage to upload them.

> I think this won't get merged.
> We are already fireing a good amount of events for different things, and this is such a niche request that I think this is not warrented, since this would fire for each TE and as such would probably cause more performance hurt then you could gain from it.
> 
> Additionally MC already does camera based culling of TEs that are outside of its rendering range, which is probably a good middle ground between BTTW Performance via Raytracing and not testing them at all, with regards to complexity and benefits of the complexity both in terms of code maintanance as well as actual game performance.
> 
> If you can come back with numbers and show us that the event is useable we might consider adding this.
> But for now I will close this PR.

So I will go into more detail what my mod is doing and what the expected benefits are. From my testing the best case is something like a roguelike dungeon which yields an fps improvement of ~50%. In open areas (for example a flat surface) the mod has no benefit but I haven't noticed any major fps losses (only like 3 fps less at most).
Though I have to say there is also an "issue". I'm only raytracing from the camera position to the eye position the entity. That means when the eye position of the entity is hidden behind a block it will skip the rendering of that entity although for example it's arms are visible. To (at least mostly) avoid this problem I raytrace both directions and then I compare the positions that were hit and only skip the rendering when the distance is greater than 16 blocks. The issue can still occur but it's pretty rare and not really noticable. (When reducing the 16 blocks limit the fps improvement is even bigger which might be helpful for people who want to play a modpack with a bad pc)

PR #7589, opened 2021-01-08
[1.16.x] Add Spawn Egg Item Interaction to IForgeItem

ChampionAsh5357:
The spawn egg item has been problematic to users for a number of reasons, specifically its desire to rely on an IdentityHashMap making implementations of supplied entity types even more complex than it should be. Technically the spawn egg could be represented as a new Item besides for one method, `#getChildToSpawn`. This method is called within `MobEntity` to spawn a child when a mob is right clicked with its associated spawn egg. This could be handled with `PlayerInteractEvent$EntityInteract`; however, the optional supplied calls a protected method that would need to be accessed using reflection for AT.

To get around this to make spawn eggs fully implementable from the base item, a new method is added. This essentially takes in a new mob entity to create and spawns it in the world. From there, the optional is mapped to its corresponding entry and calls the protected method. Finally, if its present, it returns `SUCCESS` on the server. Note that this method is called only on the logical server similarly to the `SpawnEggItem` implementation.

As an alternative solution, I could instead make the method an interaction for all mob entities that returns an `ActionResultType` circumventing the need for the event altogether besides for vanilla items interacting with vanilla entities. This still makes it impossible for the protected method to be called without AT. However, the only method call this would be applicable to is within `FoxEntity`. It more or less depends on how much should the entity interaction with items be reliant on the event.

### DIFF
--- a/patches/minecraft/net/minecraft/entity/MobEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/MobEntity.java.patch
@@ -50,3 +50,11 @@
           }
        } else {
           return EquipmentSlotType.HEAD;
+@@ -1055,6 +1066,7 @@
+             }
+          }
+ 
++         if(this.field_70170_p instanceof ServerWorld && itemstack.getMobToSpawn(p_233661_1_, this).map(child -> { this.func_213406_a(p_233661_1_, child); return child;}).isPresent()) return ActionResultType.SUCCESS;
+          if (itemstack.func_77973_b() instanceof SpawnEggItem) {
+             if (this.field_70170_p instanceof ServerWorld) {
+                SpawnEggItem spawneggitem = (SpawnEggItem)itemstack.func_77973_b();
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
@@ -19,6 +19,7 @@
 
 package net.minecraftforge.common.extensions;
 
+import java.util.Optional;
 import java.util.Set;
 import java.util.function.Consumer;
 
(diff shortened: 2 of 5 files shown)


pupnewfster:
Personally I am not sure how worthwhile this solution is given it still leaves out a lot of functionality modders need to implement/override to have fully working spawn eggs. Currently what is needed to implement spawn eggs that properly support their entity types being registry replaced:

- Pass null or some prebaked instance of your entity type to the `SpawnEggItem`, and then override `SpawnEggItem#getType` to return your entity at the proper times. [Example](https://github.com/mekanism/Mekanism/blob/1.16.x/src/additions/java/mekanism/additions/common/item/AdditionsSpawnEggItem.java)
- If using vanilla's method of coloring the egg and their model, [adding](https://github.com/mekanism/Mekanism/blob/1.16.x/src/additions/java/mekanism/additions/client/AdditionsClientRegistration.java#L181-L182) a color handler as it is a bit inconsistent from my experience for if custom eggs will get a handler added or not by vanilla.
- [Adding](https://github.com/mekanism/Mekanism/blob/1.16.x/src/additions/java/mekanism/additions/common/entity/SpawnHelper.java#L79-L93) a copy or equivalent version of the dispense behavior vanilla does for the spawn egg item.
- [Overriding](https://github.com/mekanism/Mekanism/blob/1.16.x/src/additions/java/mekanism/additions/common/entity/baby/EntityBabySkeleton.java#L71-L73) `IForgeEntity#getPickedResult` in your entity to return your spawn egg item.

If I am reading this PR correctly, all it would *really* do is basically making it so you need to do is override a different method instead of `SpawnEggItem#getType`, whereas the big issue with spawn eggs currently because of how early things like dispensers are done is that the `SpawnEggItem.EGGS` map can't easily be substituted out with having a supplier/delegate based version that then both pick block on entities, dispenser behavior, and even maybe the color handlers could query thus allowing for custom spawn eggs to be implemented by basically just calling the constructor with a custom entity type, instead of still having to override a ton of other methods.

ChampionAsh5357:
So, it's not that it leaves functionality out, it's more so that all other functionality has been currently implemented. Every single method except `SpawnEggItem#getChildToSpawn` has an equivalent method that requires no override except if you wanted to make a system similar to the existing spawn egg. This functionality can be implemented using an event where the player interacts with the mob holding this egg, allowing the player to spawn a child version of the mob when right clicked with the spawn egg. All this method does is add in that missing functionality that when specified, it will spawn a mob when right clicking an entity.

Yes, what you said is true if you wanted to use the default spawn egg implementation. My discussion is more for a method that would need to be called in an event each time rather than through the item directly. This is why I mentioned in the alternative solution that this could technically be written as a method that could handle how this item should interact when right clicked on an entity. This can become an expansive method for modded items that interact with vanilla entities if need be.

If we were to do a proper implementation, it would probably be to move the spawn egg dispenser registration to during common setup and have the eggs store a lazily mapped value that would be resolved to that map during that time frame. Although, if that was the case, it would be more appropriate to move all dispensers to during common setup because of how early they occur.

I would like to change this PR to a draft for the time being as the implementation of what this should entail might need a bit more discussion than just adding a method to cover all use-cases with my current point. The end goal is to either make the `SpawnEggItem` class useable or make it so all use cases could be covered if extending a regular `Item`.

PR #7590, opened 2021-01-08
Fix for MacOS Shift/Scroll Issue

desagas:
This works because A. minecraft and forge currently do not us xoffset as it relates to scrollCallback(), and B. Because the default  MAC config is to HScroll when pressing shift and scrolling at the same time. This change very simply combines them together, which will complete fix the issue.

- GLFW does in fact see and output both offsets, so this is not a GLFW problem, this is a minecraft problem.

### DIFF
--- a/patches/minecraft/net/minecraft/client/MouseHelper.java.patch
+++ b/patches/minecraft/net/minecraft/client/MouseHelper.java.patch
@@ -1,5 +1,14 @@
 --- a/net/minecraft/client/MouseHelper.java
 +++ b/net/minecraft/client/MouseHelper.java
+
+import net.minecraft.client.util.NativeUtil;
++ import net.minecraft.util.Util;
+import net.minecraft.util.math.MathHelper;
+
+private final Minecraft minecraft;
++ public static final boolean IS_RUNNING_ON_MAC = Util.getOSType() == Util.OS.OSX;
+private boolean middleDown;
+
 @@ -71,6 +71,7 @@
              this.field_198042_g = -1;
           }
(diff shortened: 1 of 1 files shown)


Cyborgmas:
Yeah you arent supposed to manually edit patch files... you have to fork the repo, run `gradlew setup` in the console then you can go and edit the minecraft source file (in the `forge` project not the `clean` one) and run `gradlew genPatches`. 
The patch file will be generated automatically. Note that it is desired they are as small as possible so you want to not put any imports in them and instead always use the FQN.

desagas:
Okay, so, with no import, should I just put the "get os operating system" function in manually, to get passed that?

Cyborgmas:
its just instead of doing
`Class.doStuff()` go with `net.package.mine.Class.doStuff()`
the first will require an import the 2nd not

desagas:
Okay, so I would want to combine the Util package, and use it for all three references that will be imported in total.

desagas:
Gotcha. Thanks

desagas:
Hmmm, gradle didn't manager to install the net/minecraft files. I can't quite figure out how to get the to show up.

desagas:
Is it normal for the `gradlew setup` to delete a line of code. In the most recent build (38), `public boolean isMiddleDown() {
      return this.middleDown;
   }` has been removed from the MouseHelper code ... so< i guess, temporarily, I added it back so that I could test my code, as ForgeClientHooks can not run without that one line of code.

desagas:
Something went wrong, I am just going to cancel this, refork, and resubmit.

PR #7591, opened 2021-01-09
Add possibility for mods to add optional resource / data packs to their mod.

Cyborgmas:
The new pack finder will consider all packs in main/resources/packs/

This makes it possible for a mod author to promote any user made packs or showoff some of the possibilities of the mod using different packs.


KnightMiner:
This seems pretty niche to me, its not hard for a mod to add their own pack finder, I've done so in a couple of my mods. So not sure this needs to be in Forge personally.

Cyborgmas:
Okay but how is it niche if you then state that you use something similar ? Would it have not been easier to do if forge already provided the way to do it?

KnightMiner:
One of my two cases this PR would be insufficient to handle. The other would save me a few lines of code at most, and I was able to get away with a much simpler implementation anyways

PR #7593, opened 2021-01-09
Fixes MACOS SHIFT SCROLL converting to HSCROLL

desagas:
This fix simply adds the never used GLFW xoffset from the scrollCallback to the yoffset, to include both in the scrolling functionality. As xoffset is used no where else in the code, it does not break anything else. Further, I have compiled and tested, and it works fine.

### DIFF
--- a/patches/minecraft/net/minecraft/client/MouseHelper.java.patch
+++ b/patches/minecraft/net/minecraft/client/MouseHelper.java.patch
@@ -1,71 +1,20 @@
 --- a/net/minecraft/client/MouseHelper.java
 +++ b/net/minecraft/client/MouseHelper.java
-@@ -71,6 +71,7 @@
-             this.field_198042_g = -1;
-          }
+@@ -116,7 +116,12 @@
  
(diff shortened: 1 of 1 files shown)


LexManos:
Alright, 1) Stop making multiple PRs, you can update your branch on your end and update the existing PR
2) This is not the place to learn git
3) Review your changes before submitting. We will never accept something that a) breaks the build and b) removes every other change to the file.

desagas:
My changes are final. No more. This is the final code. Are you not allowing me to request the PR anymore? I only made changes because that person made me think that I had to, I learn now that what I did was fine, and he was not affiliated with Forge. I apologize!

DaemonUmbra:
You're not doing it correctly and from what I can see haven't even commited your changes via git, the most basic instruction I gave you. So I think you've worn out your PR welcome for now.

desagas:
My change works fine!! This one did, the first one did, all the ones in between did!

DaemonUmbra:
It doesn't... you didn't follow the correct process and you are ignorantly refusing to do so.

LexManos:
I have blocked him from the org for the next week.
Please take this week to learn how to use git and how to read your own code.
If you want to come back after that and re-attempt your PR, feel free.

PR #7594, opened 2021-01-09
Fix typos

PepperCode1:
I see these log messages every time I start my Forge server, so I had to make this PR.


covers1624:
Small PR's are not accepted for fixing typos. If you would like to create a larger PR across the entire repo for cleaning them up, it may be considered.

PR #7599, opened 2021-01-11
Fix worlds that are missing vanilla dimensions in the level.dat

Cyborgmas:
This fix will inject the vanilla dimensions back inside the nbt that is used to create the `DimensionGeneratorSettings`. #7527 prevents this bug from happening, this PR will fix any worlds that had the bug happen previous to that fix.

### DIFF
--- a/patches/minecraft/net/minecraft/world/storage/SaveFormat.java.patch
+++ b/patches/minecraft/net/minecraft/world/storage/SaveFormat.java.patch
@@ -1,5 +1,14 @@
 --- a/net/minecraft/world/storage/SaveFormat.java
 +++ b/net/minecraft/world/storage/SaveFormat.java
+@@ -95,7 +95,7 @@
+          }
+       }
+ 
+-      Dynamic<T> dynamic1 = p_237259_1_.update(TypeReferences.field_233375_y_, dynamic, p_237259_2_, SharedConstants.func_215069_a().getWorldVersion());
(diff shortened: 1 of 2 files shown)


Cyborgmas:
[phase2_test_world.zip](https://github.com/MinecraftForge/MinecraftForge/files/5797980/phase2_test_world.zip)
Here is a test world that has the end "deleted". If loading it in vanilla, the end is not available, using this fix it is.

50ap5ud5:
Hi Cyborgmas, thanks for providing this solution! I have conducted tests using your solution and provided resources and I can confirm it solves the issue.

Here were my findings and feedback:
-Worlds created before the Phase 1 fix (Vanilla dimensions that have already been "deleted") are able to restore access to all vanilla dimensions using this solution.
-This solution does not restore access to modded dimensions that have been removed, which is good as this is within scope of Phase 2.

For more information about my testing procedure, please see attached document:
https://drive.google.com/file/d/1aU_bfKTYLuBqYqwyKD1tH47eS9xJ6qcy/view?usp=sharing

50ap5ud5 (approved):

TheCurle (approved):
Works from my testing:
- Install a datapack that adds a dimension (several have been provided in various PRs)
- Create a world with a forge version in the interval 35.0.15><35.1.32
- Visit the End then return to the overworld
- Remove the datapack
- Observe that the End portal no longer works
- Launch with this fix, and observe that the End portal works again

I'm happy to push for the inclusion of this PR

PR #7605, opened 2021-01-17
Prevent NullPointerException when adding new loot pool.

noobanidus:
The `name` parameter of LootPool's init was added in a previous version and was required in order to create a new LootPool. With the switch to builders as of 1.14.4 (if not earlier), specifying a name became "optional" and is not enforced as part of the builder contruct.

Thus the `name` field became nullable and null by default. Any pool injected into a table using this method without specifically calling the `name` method of the builder returns null when `e.getName()` is called, but this is immediately dereferenced with an `equals` check.

This becomes extremely confusing to debug, however, as it is always a *previously added* pool without a name that causes a crash, rather than the pool being added by the mod. This makes it very difficult to track down.

Additionally, this causes the loot table being affected to not load and thus result in the table being considered empty.

This PR simply adds a null check before dereferencing `e.getName()`, preventing the NullPointerException.

### DIFF
--- a/patches/minecraft/net/minecraft/loot/LootTable.java.patch
+++ b/patches/minecraft/net/minecraft/loot/LootTable.java.patch
@@ -76,7 +76,7 @@
 +
 +   public void addPool(LootPool pool) {
 +      checkFrozen();
-+      if (field_186466_c.stream().anyMatch(e -> e == pool || e != null && e.getName().equals(pool.getName())))
++      if (field_186466_c.stream().anyMatch(e -> e == pool || e.getName() != null && e.getName().equals(pool.getName())))
 +         throw new RuntimeException("Attempted to add a duplicate pool to loot table: " + pool.getName());
 +      this.field_186466_c.add(pool);
 +   }
--- a/patches/minecraft/net/minecraft/loot/LootTable.java.patch
+++ b/patches/minecraft/net/minecraft/loot/LootTable.java.patch
@@ -76,7 +76,7 @@
 +
 +   public void addPool(LootPool pool) {
 +      checkFrozen();
-+      if (field_186466_c.stream().anyMatch(e -> e == pool || e.getName().equals(pool.getName())))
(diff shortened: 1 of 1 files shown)


TelepathicGrunt:
Maybe not this PR but it would be nice if unnamed pools would be logged by Forge so players know which mod is adding unnamed pools and devs know when they made their pool incorrectly. Enforcing named pools would be a breaking change which could be excessive but logging should help devs transition to better practices over time without causing crashes for players updating.

noobanidus:
@TelepathicGrunt That's an extremely valid point, although that would be something to patch into LootPool's builder/init instead. I'm not sure how the best way to go about doing that would be.

ChloeDawn:
This change won't prevent null names from causing an NPE, you are null checking the entry not the name. It would be `getName().equals` where the dereferencing would occur

noobanidus:
> This change won't prevent null names from causing an NPE, you are null checking the entry not the name. It would be `getName().equals` where the dereferencing would occur

I'm literally just pushing this now. My fault for getting nervous and trying to do my first Forge PR at 6am.

ChloeDawn:
Never make Forge PR's before a strong cup of coffee 😁

ChloeDawn (approved):

XFactHD:
> 
> 
> Maybe not this PR but it would be nice if unnamed pools would be logged by Forge so players know which mod is adding unnamed pools and devs know when they made their pool incorrectly. Enforcing named pools would be a breaking change which could be excessive but logging should help devs transition to better practices over time without causing crashes for players updating.

A few days ago, someone on the discord provided a log which contained some output related to this from a mod, [here](https://github.com/Extegral/Enigmatic-Legacy/commit/4a12f4b550a0ee4c36015a5bd4734dc445b481cc) is the commit that added that feature to the specific mod, maybe a solution similar to their Mixin but cleaner and non-crashing could be implemented in Forge.

PR #7606, opened 2021-01-18
Fix MC-176559 Breaking animation and process resets when a pickaxe enchanted with Mending mends by XP / Mending slows down breaking blocks again

BlueAgent:
Potential fixes for [MC-176559](https://bugs.mojang.com/browse/MC-176559)

Just opening this PR for comments on the best way to fix this.
It seems to have been caused by the removal of metadata and then damage being stored in NBT instead.
~~Then again apparantly this bug was reintroduced in 1.15.2, so might take a look at how vanilla fixed it before that.~~
Turns out it's been an issue since as mentioned above, the issue tracker has only started tracking it since 1.15.2 though.

Currently, the check always fails at the term `ItemStack.areItemStackTagsEqual(newStack, oldStack)` if the damage is different so it never gets to `(newStack.isDamageable() || newStack.getDamage() == oldStack.getDamage())`.

As for a test mod for this PR (if it is needed) thinking of a command that drops experience orbs on the player and gives them a pickaxe with mending to mine stuff with. Either that or a pickaxe that is pre-damaged with mending that spawns experience orbs on the player while they are holding it.

### Methods considered:

**1) Use `ItemStack.areItemsEqualIgnoreDurability(oldStack, newStack)`**
Not sure if it is okay because that method actually ignores nbt completely rather than just damage/durability.
`IForgeItem#canContinueUsing` does use it though so maybe it's okay to also use it?

**2) Copy stack but change damage to match if the item is damagable (currently in this PR)**
Doesn't modify the existing vanilla code but `ItemStack.copy()` is probably bad to use here...

**3) Create a method to compare item stacks including NBT but ignoring the Damage tag.**
I think this is probably the best option but unsure where the code should go for this, maybe in `IForgeItemStack`?
Not quite sure what to name it because that's how I thought `areItemsEqualIgnoreDurability` should work but is already taken.

### DIFF
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
@@ -19,6 +19,8 @@
 
 package net.minecraftforge.common.extensions;
 
+import java.util.HashSet;
+import java.util.Objects;
 import java.util.Set;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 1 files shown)


sciwhiz12:
Hello! @BlueAgent, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after a few days, this PR will be closed for inactivity.

BlueAgent:
Was unsure of which option to take when fixing this. Will probably end up going with option 3 though, and update the PR to 1.18 (assuming it's still an issue that is).

BlueAgent:
Ended up going with option 1 (using `ItemStack.isSameIgnoreDurability`) even though it technically doesn't do what the method says it does. It essentially just compares if the item is the same. Doesn't check NBT at all.

Figured it would be okay as `canContinueUsing` is using it for things like drawing a bow I think.
Let me know if this is not okay or if there's a better idea for how to fix this till Mojang fixes it officially.

BlueAgent:
Recorded a [Before](https://youtu.be/Cpi-99Zk2JY) and [After](https://youtu.be/M8Td5xLfNW8) video.

Some helpful commands:
```
execute unless entity @e[type=minecraft:experience_orb] run summon minecraft:experience_orb ~ ~1 ~ {Value:1s}
give Dev minecraft:diamond_pickaxe{Damage:1500}
enchant Dev minecraft:mending
```

LexManos:
Option 3 would be the way to go, except for the fact that we have a bunch of these hooks already. We want to reset things when NBT changes because it's used for everything not just damage.

BlueAgent:
> Option 3 would be the way to go, except for the fact that we have a bunch of these hooks already.

To clarify, does this mean we're going with Option 3 but without adding a helper method and just having the whole implementation in the hook? (Or do you mean there's already an existing helper method we can use?)

Will revisit this when I can and do the former unless it turns out I intepreted wrong.

BlueAgent:
Implemented Option 3 without a helper method.

[Video of 3rd Revision](https://www.youtube.com/watch?v=ubHPCiGTNSU)

- Show two damaged Diamond Pickaxes with Mending and the second one with additional Fortune enchantment.
- Demonstrate that mining progress resets if you swap between them while mining.
- Demonstrate that mining progress does not reset if you pick up experience while mining.

TheCurle:
Looks good to me! 
Going to do some final tests and will see about merging in the next few days.

@sciwhiz12 you're on poking duty.

sciwhiz12:
:shipit: Published as **1.18.2-40.0.16**.

PR #7608, opened 2021-01-18
Add support for caching of configuration values.

marchermans:
After running into the case where I forgot to cache the forge configuration value myself and it being discussed many times in discord even some people creating coremods for something so trivial, I decided to look into this. 

Even though it is simple enough to store the configuration value in a static field and update it during the reloading event, many people do not do this, and call `get()` on the ConfigValue to get the value of the configuration entry when ever they need it. Sometimes even during performance critical pieces of code, like during rendering calls.

The lookups of the values in the config system can become a performance bottleneck here, and as such caching of that value is actually needed. This can be done (as described above) in a static field easily enough. But that would be something modders would need to take care of themselfs, and is not common knowledge. This little PR adds a caching value directly in the ConfigValue while also adding a callback to the reloading system to clear them when the value is loaded from disk.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
+++ b/src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
@@ -90,6 +90,7 @@ public class ForgeConfigSpec extends UnmodifiableConfigWrapper<UnmodifiableConfi
                 ((FileConfig) config).save();
             }
         }
+        this.afterReload();
     }
 
     public boolean isCorrecting() {
--- a/src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
+++ b/src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
@@ -108,6 +108,22 @@ public class ForgeConfigSpec extends UnmodifiableConfigWrapper<UnmodifiableConfi
         return this.values;
     }
 
+    public void afterReload() {
+        this.resetCaches(getValues().valueMap().values());
+    }
+
+    private void resetCaches(final Iterable<Object> configValues) {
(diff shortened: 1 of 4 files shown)


marchermans:
Added a set of three unit tests which serve as a benchmark for the implementation.
Full log: https://gist.github.com/OrionDevelopment/2f6b1397d9c92888a362a94a1c5224ef

Testcases added:
- Simple string lookup, one key deep.
- Complex object lookup, List of List of String.
- Deep object lookup, key depth plays a large factor in the lookup strategy of nightconfig. Deeper keys take a lot longer.

The tests were chosen cause they represent an aspect config access mods perform during rendering.
The config access during rendering was especially chosen, since access to a config value during normal ticking and/or individual access of a config value incidentally does not have a large enough impact on performance, however accessing the value over and over again in a single frame (or multiple for that matter) can cause a performance impact.

The direct results:
```
Computed test results for: deepKeySpeedTest:
  > Enabled:  ~ 247372.60 ns.   +-  198559.60 ns.
  > Disabled: ~2777916.60 ns.   +- 1191883.40 ns.

Computed test results for: objectSpeedTest:
  > Enabled:  ~ 139451.80 ns.   +-  607665.20 ns.
  > Disabled: ~ 431878.05 ns.   +-  604732.95 ns.

Computed test results for: simpleSpeedTest:
  > Enabled:  ~  38551.05 ns.   +-   13802.05 ns.
  > Disabled: ~ 369100.30 ns.   +-   82096.30 ns.
```

In 2 out of 3 cases the cached version is an order of magnitude faster, and in the object speed test the implementation is 75% faster.

pupnewfster (approved):

Issue #7610, opened 2021-01-19
Race condition in LazyOptional


malte0811:
Version: 19f8d2a7937dc7968ecbef2f9785687193fcd210/36.0.1.

This probably isn't a major issue in practice (though we did get reports of it for IE, see BluSunrize/ImmersiveEngineering#4547), I still think that it should at least be documented somewhere:

`LazyOptional` appears to be designed for multithreaded use (otherwise using an `AtomicReference` would be a very unusual choice). However [this](https://gist.github.com/malte0811/217cf7b7cd5e190958df44e65c56b0fe) code does not produce the expected result: Expected behavior would be for both threads to run without any exceptions, in practice one of them crashes with [this](https://github.com/MinecraftForge/MinecraftForge/blob/19f8d2a7937dc7968ecbef2f9785687193fcd210/src/main/java/net/minecraftforge/common/util/LazyOptional.java#L133) exception.  
(Note: The call to sleep isn't realistic, but the race condition can still occur with reasonable getters. Sleep just makes it trivial to reproduce)  
The problem is that `resolved` is [set to a reference to `null` in one thread](https://github.com/MinecraftForge/MinecraftForge/blob/19f8d2a7937dc7968ecbef2f9785687193fcd210/src/main/java/net/minecraftforge/common/util/LazyOptional.java#L115), causing the second thread to [return that `null`](https://github.com/MinecraftForge/MinecraftForge/blob/19f8d2a7937dc7968ecbef2f9785687193fcd210/src/main/java/net/minecraftforge/common/util/LazyOptional.java#L110-L111).  
The easiest fix would be to add a lock, similar to `Lazy.concurrentOf`. I'm not sure about the performance implications of that lock though, and the original code seems to be an attempt at keeping that function as fast as possible. On the other hand the "inner" part of `getValue` should only run rarely (outside of bugs), so it may not be necessary to keep that section super fast.

LexManos:
The inner part of getValue, as in everything inside the `if (supplier != null)` block should only ever be run once per lazy optional. So throwing a sync lock internally on that would probably be the best idea.
Would need to be a proper double checked sync lock. But should be simple enough.
LazyOptional was designed to work around costly value construction, so having the supplier take some time isn't unreasonable.

malte0811:
[...]

PR #7611, opened 2021-01-20
FIx race condition in LazyOptional

malte0811:
See #7610 

Also adds a test for both corner cases discussed in the issue (parallel access and contract-violating suppliers).

### DIFF
--- a/src/main/java/net/minecraftforge/common/util/LazyOptional.java
+++ b/src/main/java/net/minecraftforge/common/util/LazyOptional.java
@@ -57,7 +57,8 @@ import net.minecraftforge.common.capabilities.Capability;
 public class LazyOptional<T>
 {
     private final NonNullSupplier<T> supplier;
-    private AtomicReference<T> resolved;
+    private final Object lock = new Object();
+    private T resolved;
     private Set<NonNullConsumer<LazyOptional<T>>> listeners = new HashSet<>();
     private boolean isValid = true;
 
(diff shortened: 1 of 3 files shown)


src/main/java/net/minecraftforge/common/util/LazyOptional.java
```diff
@@ -23,12 +23,13 @@
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.atomic.AtomicReference;
```
TheCurle, line 26: What's the purpose of moving away from the standard library functions?
The Apache tools are much less useful for this, as we wanted to keep it Atomic, which Apache's Mutable has notable shortcomings with...

malte0811: What use would keeping it atomic have in the new implementation? It would most likely be slower (no idea how much), and the multithreading is correct without it (the lock deals with synchronization).

LexManos: As I said on the issue, keep the atomic and don't make weird random state value things.
Any claims of it being 'slower' need to actually have statistics to back it up.
Currently your doing a 36 liner change for something that could be 2.

LexManos:
Anyways im tired of arguing with you when i've explicitly told you how to do it.
However this isn't that important of a change do I don't care.
Next time if you want to submit fixes for minor things, don't include random-redesigns along side it.

PR #7613, opened 2021-01-21
Dimension with modded chunkgenerator when updating .4 -> .5 nukes end and nether testmod

Cyborgmas:
Made a testmod for a current bug that happens when upgrading the mc version and one of the dimensions used has a custom chunk generator.
Here is a world created in 1.16.4 with the dimension, loading it with this branch will nuke end/nether. 
Loading with this branch before the merge commit will load them both. (or can see they exist in the level.dat with a nbt viewer).
[TestWorldCopy.zip](https://github.com/MinecraftForge/MinecraftForge/files/5846656/TestWorldCopy.zip)


Cyborgmas:
The bug is the same as #7527 but instead on the CompoundListCodec that the datafixer for the WorldGenSettings in the level.dat uses.
The fix is the same, except that the codec is only given through the TypeTemplate/Type so I had to copy that as well.

Cyborgmas:
So #7629 fixes this, unintentionally. As DFU will be deleting the dim, but then it gets re-injected.
Might still log errors like "Unknown key: some_modded_chunk_generator".

PR #7617, opened 2021-01-24
Add 2 new events related to Villagers

sguest:
Adds 2 new events
- `VillagerProfessionChangeEvent` fired when a villager's profession changes
- `AgeableEntityGrowingAdultEvent` fired when an `AgeableEntity` grows from baby to adult.

Note: I combined these into a single pull request so the Forge maintainers only have 1 PR to review instead of 2, but they're in separate commits and I'd be happy to make 2 PRs instead if desired.

Reason for change:
Working on a mod where I am altering Villager behavior in order to have villagers interact with a custom block. I accomplish this by creating a new `Task` and adding it to the villager's brain during `EntityJoinWorld`. Sample code [here](https://github.com/sguest/village-life/blob/master/src/main/java/sguest/villagelife/entity/ai/VillagerBehavior.java). This is functional and seems to work well. However, when a villager changes profession or grows into an adult the villager's brain is re-initialized. This wipes out my added task and the villager no longer performs it until they trigger `EntityJoinWorld` again. These 2 new events would allow me to rerun the task init code in these cases.

Without these events I'm not sure how to accomplish persistence of this task beyond using a tick handler or similar construct to check periodically if things have changed, but that seems highly inefficient. Technically an event that simply fires on an entity's brain being reset would meet my personal needs even better, but these 2 events seem like they'd have much wider applicability.

### DIFF
--- a/patches/minecraft/net/minecraft/entity/AgeableEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/AgeableEntity.java.patch
@@ -1,0 +1,10 @@
+--- a/net/minecraft/entity/AgeableEntity.java
++++ b/net/minecraft/entity/AgeableEntity.java
+@@ -135,6 +135,7 @@
+    }
+ 
+    protected void func_175500_n() {
++      net.minecraftforge.event.ForgeEventFactory.onAgeableEntityAgeChange(this);
+    }
+ 
+    public boolean func_70631_g_() {
--- a/patches/minecraft/net/minecraft/entity/merchant/villager/VillagerEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/merchant/villager/VillagerEntity.java.patch
@@ -9,7 +9,16 @@
           if (this.func_70631_g_()) {
              this.func_213756_es();
              return ActionResultType.func_233537_a_(this.field_70170_p.field_72995_K);
-@@ -634,7 +634,8 @@
+@@ -489,6 +489,8 @@
+       }
+ 
+       this.field_70180_af.func_187227_b(field_213775_bC, p_213753_1_);
++      if (villagerdata.func_221130_b() != p_213753_1_.func_221130_b())
++         net.minecraftforge.event.ForgeEventFactory.onVillagerProfessionChange(this, villagerdata.func_221130_b());
+    }
+ 
+    public VillagerData func_213700_eh() {
+@@ -634,7 +636,8 @@
     }
  
     protected ITextComponent func_225513_by_() {
@@ -19,7 +28,7 @@
     }
  
     @OnlyIn(Dist.CLIENT)
-@@ -687,7 +688,7 @@
+@@ -687,7 +690,7 @@
     }
  
     public void func_241841_a(ServerWorld p_241841_1_, LightningBoltEntity p_241841_2_) {
@@ -28,7 +37,7 @@
           field_184243_a.info("Villager {} was struck by lightning {}.", this, p_241841_2_);
           WitchEntity witchentity = EntityType.field_200759_ay.func_200721_a(p_241841_1_);
           witchentity.func_70012_b(this.func_226277_ct_(), this.func_226278_cu_(), this.func_226281_cx_(), this.field_70177_z, this.field_70125_A);
-@@ -699,6 +700,7 @@
+@@ -699,6 +702,7 @@
           }
  
           witchentity.func_110163_bv();
--- a/patches/minecraft/net/minecraft/entity/monster/HoglinEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/monster/HoglinEntity.java.patch
@@ -9,7 +9,15 @@
              this.func_241412_a_(SoundEvents.field_232715_fE_);
              this.func_234360_a_((ServerWorld)this.field_70170_p);
           }
-@@ -212,6 +212,7 @@
+@@ -144,6 +144,7 @@
+          this.func_110148_a(Attributes.field_233823_f_).func_111128_a(6.0D);
+       }
+ 
++      super.func_175500_n();
+    }
+ 
+    public static boolean func_234361_c_(EntityType<HoglinEntity> p_234361_0_, IWorld p_234361_1_, SpawnReason p_234361_2_, BlockPos p_234361_3_, Random p_234361_4_) {
+@@ -212,6 +213,7 @@
        ZoglinEntity zoglinentity = this.func_233656_b_(EntityType.field_233590_aW_, true);
        if (zoglinentity != null) {
           zoglinentity.func_195064_c(new EffectInstance(Effects.field_76431_k, 200, 0));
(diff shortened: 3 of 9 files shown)


Mysterious-Dev (approved):

patches/minecraft/net/minecraft/entity/AgeableEntity.java.patch
```diff
@@ -0,0 +1,10 @@
+--- a/net/minecraft/entity/AgeableEntity.java
++++ b/net/minecraft/entity/AgeableEntity.java
+@@ -135,6 +135,7 @@
+    }
+ 
+    protected void func_175500_n() {
++      net.minecraftforge.event.ForgeEventFactory.onAgeableEntityGrowingAdult(this);
```
LexManos: Hooglins don't call super so this wont fire for them.

src/main/java/net/minecraftforge/event/entity/living/AgeableEntityGrowingAdultEvent.java
```diff
@@ -0,0 +1,28 @@
...
+ * This event is fired when an {@link AgeableEntity} becomes an adult. <br>
+ * It is fired via {@link ForgeEventFactory#onAgeableEntityGrowingAdult(AgeableEntity)}.
+ * <br>
+ * This event not is {@link net.minecraftforge.eventbus.api.Cancelable}.
+ * <br>
+ * This event does not have a result. {@link HasResult} <br>
+ * <br>
+ * This event is fired on the {@link MinecraftForge#EVENT_BUS}.
+ */
+public class AgeableEntityGrowingAdultEvent extends LivingEvent
```
LexManos: This actually isn't fired when a entity becomes an adult.
It's fired when it's age changes. So it COULD be fired when an entity turns into a child FROM an adult.
Also the name needs work to reflect that.

patches/minecraft/net/minecraft/entity/ai/brain/task/ChangeJobTask.java.patch
```diff
@@ -0,0 +1,12 @@
+--- a/net/minecraft/entity/ai/brain/task/ChangeJobTask.java
++++ b/net/minecraft/entity/ai/brain/task/ChangeJobTask.java
+@@ -19,7 +19,9 @@
+    }
+ 
+    protected void func_212831_a_(ServerWorld p_212831_1_, VillagerEntity p_212831_2_, long p_212831_3_) {
++      VillagerProfession oldProfession = p_212831_2_.func_213700_eh().func_221130_b();
+       p_212831_2_.func_213753_a(p_212831_2_.func_213700_eh().func_221126_a(VillagerProfession.field_221151_a));
+       p_212831_2_.func_213770_a(p_212831_1_);
++      net.minecraftforge.event.ForgeEventFactory.onVillagerProfessionChange(p_212831_2_, oldProfession);
```
LexManos: Why are these called from the tasks? There are other paths that change the profession.
Why not just call it from setVillagerData, which both these hooks call?

LexManos (changes requested):

patches/minecraft/net/minecraft/entity/AgeableEntity.java.patch
sguest: Added event to `HoglinEntity.onGrowingAdult()` also

src/main/java/net/minecraftforge/event/entity/living/AgeableEntityGrowingAdultEvent.java
sguest: Renamed event to reflect that it can be child->adult or vice-versa, and updated header comments. Now called `AgeableEntityAgeChange`.

patches/minecraft/net/minecraft/entity/ai/brain/task/ChangeJobTask.java.patch
sguest: The idea behind adding them to the tasks was so they could fire after the brain reset, at which point the profession change feels more "complete". But I see your point about other code paths potentially changing things, so the event is now in `setVillagerData`.

patches/minecraft/net/minecraft/entity/merchant/villager/VillagerEntity.java.patch
```diff
@@ -9,7 +9,15 @@
           if (this.func_70631_g_()) {
              this.func_213756_es();
              return ActionResultType.func_233537_a_(this.field_70170_p.field_72995_K);
-@@ -634,7 +634,8 @@
+@@ -489,6 +489,7 @@
+       }
+ 
+       this.field_70180_af.func_187227_b(field_213775_bC, p_213753_1_);
++      if(villagerdata.func_221130_b() != p_213753_1_.func_221130_b()) net.minecraftforge.event.ForgeEventFactory.onVillagerProfessionChange(this, villagerdata.func_221130_b());
```
LexManos: Spaces before (s
And a second line after the if for the event
Having things run off the side of the screen is a PITA to read

patches/minecraft/net/minecraft/entity/monster/HoglinEntity.java.patch
```diff
@@ -9,7 +9,15 @@
              this.func_241412_a_(SoundEvents.field_232715_fE_);
              this.func_234360_a_((ServerWorld)this.field_70170_p);
           }
-@@ -212,6 +212,7 @@
+@@ -144,6 +144,7 @@
+          this.func_110148_a(Attributes.field_233823_f_).func_111128_a(6.0D);
+       }
+
```
LexManos, line 15: This should just call super. There is no reason not to and it makes the patch easier.

sguest: Done, now just calls super

patches/minecraft/net/minecraft/entity/merchant/villager/VillagerEntity.java.patch
sguest: My understanding of the PR guidelines was to keep things in patches to 1 line if humanly possible. But I've now split it to a 2nd line.

TheCurle: we prefer not having *large* patches, but splitting on long lines helps readability a ton.

sguest: Fair enough, makes sense

ChampionAsh5357:
This PR should be updated to 1.17 as that is now the latest version. This could then be backported to 1.16 depending on if the PR gets merged.

sciwhiz12:
Hello! @sguest, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

sguest:
I think this would still be applicable to 1.18, but in the almost-year since I created this PR, I no longer have a functioning build for Forge locally locally (and certainly not one for 1.18), and I'm not sure when I'll have a chance to rebase and test this.

PR #7618, opened 2021-01-24
Fix IAngerable entities crashing when reading nbt on client

Cyborgmas:
Their readAdditional method has an unchecked cast to ServerWorld, even though vanilla itself uses this to render entities in the mob spawner (these entities all don't render in the mob spawner)

See #7619


pupnewfster:
Should reference [MC-189565](https://bugs.mojang.com/browse/MC-189565) in the comments

TelepathicGrunt (approved):

PR #7612, opened 2021-01-20
[Omnibus] Fix typos and javadocs across code base

TheCurle:
# This Is An Omnibus PR
[Open PRs](https://github.com/TheCurle/MinecraftForge/pulls?q=is%3Apr+is%3Aopen+base%3Anoci-omnibus%2Fcleanup) | [Included PRs](https://github.com/MinecraftForge/MinecraftForge/compare/1.16.x...TheCurle:noci-omnibus/cleanup) | [Submit PR](https://github.com/TheCurle/MinecraftForge/pulls)

### What does this Omnibus target?
This PR is targeting the many typos across the Forge code base; in javadocs, other comments, variables, and other miscellaneous files, along with inaccuracies in javadocs. As long as it does not break binary compatibility, it's a valid fix for this PR.

### What does this mean?
An omnibus PR is our new way of combining common, small contributions into a single PR so that they can be merged in a single build, while maintaining the simplicity of contributing small fixes piecewise.

### How can I contribute?
To contribute to this omnibus, you must make your PR at the source branch. You'll need to switch the source repository/branch when using that link. Once your PR is approved it will be merged to the omnibus branch for later merging into the main trunk.

PR #7621, opened 2021-01-25
[1.16.5] Fix SpawnEggItem requiring problematic workarounds or code duplication

XFactHD:
This PR proposes a solution to the problem of SpawnEggItems requiring you to either pass in an EntityType that was initialized too early and therefore breaks registry contracts or pass in null and pollute the map with useless null keys.

### Proposed Solution
The proposed solution is to store all SpawnEggItems added by mods in a list and transfer them over to the EntityType->SpawnEggItem map after EntityTypes have been registered.
The map in question is used in four places:

1. **`IForgeEntity#getPicketResult(RayTraceResult target)` via `SpawnEggItem.getEgg(EntityType type)`**
This is only called during gameplay which means the transfer to the map has already happened
2. **`ItemColors.init()` via `SpawnEggItem.getEggs()`**
Even though this happens before EntityTypes are registered, it doesn't create any problems related to them because the EntityType is not used there. The include the mod eggs, the value collection from the map is concatenated with the MOD_EGGS list
3. **`BlockModelProvider#func_239863_a_()` via `SpawnEggItem.getEggs()`**
Doesn't matter to us because modders should not be using the vanilla BlockModelProvider in datagen
4. **`IDispenseItemBehaviour.init()` via `SpawnEggItems.getEggs()`**
This can be ignored for this purpose because I have found in my testing with breakpoints that the call to `IDispenseItemBehaviour.init()` happense before the Item RegistryEvent has fired which means that mod eggs are not included anyway. To make sure mod eggs get `IDispenseItemBehaviour`s, they are registered after EntityType registry when the mod eggs are transferred to the type->egg map.

### Test
To test this I have included a test mod that adds a test entity and a spawn egg for that entity. In my testing spawning by hand, the coloring of the item, the dispense behaviour and "picking" the entity to get the egg for it worked flawlessly. The entity renderer doesn't render anything custom to keep the test mod as small as possible. This means that you need to enable debug entity bounding box to "see" the test entity.

### Conclusion
Please let me know if this solution is any good or if something needs to be changed.

~~I will keep this as a draft for now because I really don't like the fact that the `RegistryObject<EntityType>` has to be declared with a wildcard and therefore forces me to use an unchecked cast when registering the renderer. The only solution I could come up with is to generify the `SpawnEggItem` but I don't know if that would be accepted. If that is a viable solution or if anyone has another idea for that, please let me know.~~ Fixed


Cyborgmas:
you should be able to remove the wildcard by making the `.create()` function's generic args explicit, 
`.<EntityType<MyType>>create()` should work.

XFactHD:
I should have been more clear about that. I can specifify the type for the `RegistryObject<EntityType>` for registering the `EntityType` but if I do that, the new SpawnEggItem constructor doesn't accept the `RegistryObject<EntityType<TestEntity>>` as a `Supplier<EntityType<?>>`.

patches/minecraft/net/minecraft/item/SpawnEggItem.java.patch
```diff
@@ -0,0 +1,55 @@
...
+ 
+ public class SpawnEggItem extends Item {
+    public static final Map<EntityType<?>, SpawnEggItem> field_195987_b = Maps.newIdentityHashMap();
++   public static final java.util.List<SpawnEggItem> MOD_EGGS = new java.util.ArrayList<>();
+    private final int field_195988_c;
+    private final int field_195989_d;
+-   private final EntityType<?> field_200890_d;
++   private final java.util.function.Supplier<? extends EntityType<?>> field_200890_d;
+ 
++   @Deprecated //FORGE: use constructor with Supplier parameter instead
```
mrp-v2: Please edit your comment to match the future Omnibus PR. See https://github.com/MinecraftForge/MinecraftForge/pull/7612/commits/dae2fb7e20fdf46d1baee2ee2d9212d78daeb69b

mrp-v2 (changes requested):

patches/minecraft/net/minecraft/item/SpawnEggItem.java.patch
```diff
@@ -0,0 +1,55 @@
...
+    private final int field_195988_c;
+    private final int field_195989_d;
+-   private final EntityType<?> field_200890_d;
++   private final java.util.function.Supplier<? extends EntityType<?>> field_200890_d;
+ 
++   /**@deprecated Use {@link SpawnEggItem#SpawnEggItem(java.util.function.Supplier, int, int, Item.Properties)}*/@Deprecated
+    public SpawnEggItem(EntityType<?> p_i48465_1_, int p_i48465_2_, int p_i48465_3_, Item.Properties p_i48465_4_) {
+       super(p_i48465_4_);
+-      this.field_200890_d = p_i48465_1_;
++      this.field_200890_d = () -> p_i48465_1_;
```
pupnewfster: This should be using the delegate, though because a bunch of mods currently pass null as the entity type it probably will need to end up as:
```java
this.typeIn = typeIn == null ? () -> null : typeIn.delegate;//TODO - 1.17: Remove check for null and enforce it being nonnull
```

pupnewfster (commented):
One thing this PR does not address is the middle click ("pick block") on the custom entities, and still requires the modder to overwrite `IForgeEntity#getPickedResult`. I am not sure if there is a good solution to this or not, but at the very least it may make sense to add to the SpawnEggItemTest mod this override so that when people look at the test mod to see how to implement a spawn egg they see they need to also override that method.

XFactHD:
> 
> 
> One thing this PR does not address is the middle click ("pick block") on the custom entities, and still requires the modder to overwrite `IForgeEntity#getPickedResult`. I am not sure if there is a good solution to this or not, but at the very least it may make sense to add to the SpawnEggItemTest mod this override so that when people look at the test mod to see how to implement a spawn egg they see they need to also override that method.

That's incorrect, I just verified in my Forge workspace that the "pick block" functionality works flawlessly, as I already mentioned in the initial message

pupnewfster:
> That's incorrect, I just verified in my Forge workspace that the "pick block" functionality works flawlessly, as I already mentioned in the initial message

Ah I overlooked the change in EntityTypeCallbacks, though we really shouldn't be exposes `SpawnEggItem.EGGS` or realistically even `SpawnEggItem.MOD_EGGS` as API, and it probably would be better as a method in SpawnEggItems that transfers mod eggs to the normal eggs map.

XFactHD:
I thought about that but did not do it because I wasn't sure if the patch would be too large that way. I will move it into SpawnEggItem and lock the collections down

XFactHD:
Moved the list to map transfer and dispenser behavior registration to SpawnEggItem and changed the old constructor to use the registry delegate of the EntityType passed in.
I also noticed that the dispenser behaviour would crash when passing null to the old constructor. This null check is also to be removed in 1.17 when non-nullability of the type is enforced.

src/test/java/net/minecraftforge/debug/item/SpawnEggItemTest.java
```diff
@@ -0,0 +1,100 @@
...
+        protected void readAdditional(CompoundNBT compound) { }
+
+        @Override
+        protected void writeAdditional(CompoundNBT compound) { }
+
+        @Override
+        public IPacket<?> createSpawnPacket() { return NetworkHooks.getEntitySpawningPacket(this); }
+    }
+
+    @Mod.EventBusSubscriber(modid = MODID, value = Dist.CLIENT, bus = Mod.EventBusSubscriber.Bus.MOD)
```
aimiebell, line 86: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

BlueAgent: But it does start with a lowercase letter. Is this a bot?

aimiebell, line 86: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

aimiebell, line 86: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -575,6 +576,17 @@ public void onCreate(IForgeRegistryInternal<Item> owner, RegistryManager stage)
...
     }
 
+    private static class EntityTypeCallbacks implements IForgeRegistry.BakeCallback<EntityType<?>>
+    {
+        static final EntityTypeCallbacks INSTANCE = new EntityTypeCallbacks();
+
+        @Override
+        public void onBake(IForgeRegistryInternal<EntityType<?>> owner, RegistryManager stage)
+        {
+            SpawnEggItem.finalizeModEggs();
```
diesieben07: Why is this here and not in say `FMLCommonSetupEvent`?

patches/minecraft/net/minecraft/item/SpawnEggItem.java.patch
```diff
@@ -0,0 +1,82 @@
...
++         }
++      };
++
++      MOD_EGGS.forEach(egg -> {
++         field_195987_b.put(egg.func_208076_b(null), egg);
++         net.minecraft.block.DispenserBlock.func_199774_a(egg, eggDispenseBehaviour);
++      });
++      MOD_EGGS.clear();
++   }
++   /* ========================================= FORGE END ====================================== */
```
diesieben07: Couldn't all this (including `MOD_EGGS`) be in say `ForgeHooks` and make the whole patch a lot smaller? Additionally, how would a mod override the default dispense behavior you are adding here?

patches/minecraft/net/minecraft/item/SpawnEggItem.java.patch
```diff
@@ -0,0 +1,82 @@
...
+-      return Iterables.unmodifiableIterable(field_195987_b.values());
++      return Iterables.unmodifiableIterable(Iterables.concat(field_195987_b.values(), MOD_EGGS));
+    }
+ 
+    public EntityType<?> func_208076_b(@Nullable CompoundNBT p_208076_1_) {
+       if (p_208076_1_ != null && p_208076_1_.func_150297_b("EntityTag", 10)) {
+          CompoundNBT compoundnbt = p_208076_1_.func_74775_l("EntityTag");
+          if (compoundnbt.func_150297_b("id", 8)) {
+-            return EntityType.func_220327_a(compoundnbt.func_74779_i("id")).orElse(this.field_200890_d);
++            return EntityType.func_220327_a(compoundnbt.func_74779_i("id")).orElse(this.field_200890_d.get());
```
diesieben07: You could use `orElseGet` here.

patches/minecraft/net/minecraft/item/SpawnEggItem.java.patch
```diff
@@ -0,0 +1,82 @@
...
+       super(p_i48465_4_);
+-      this.field_200890_d = p_i48465_1_;
++      this.field_200890_d = p_i48465_1_ == null ? () -> null : p_i48465_1_.delegate; //TODO - 1.17: Remove check for null and enforce it being nonnull
+       this.field_195988_c = p_i48465_2_;
+       this.field_195989_d = p_i48465_3_;
+       field_195987_b.put(p_i48465_1_, this);
+    }
+ 
++   public SpawnEggItem(java.util.function.Supplier<? extends EntityType<?>> typeIn, int primaryColorIn, int secondaryColorIn, Item.Properties builder) {
++      super(builder);
```
diesieben07, line 26: It might be better/more compatible to delegate to the normal constructor here.

patches/minecraft/net/minecraft/item/SpawnEggItem.java.patch
```diff
@@ -0,0 +1,82 @@
...
++++ b/net/minecraft/item/SpawnEggItem.java
+@@ -35,18 +_,28 @@
+ 
+ public class SpawnEggItem extends Item {
+    private static final Map<EntityType<?>, SpawnEggItem> field_195987_b = Maps.newIdentityHashMap();
++   private static final java.util.List<SpawnEggItem> MOD_EGGS = new java.util.ArrayList<>();
+    private final int field_195988_c;
+    private final int field_195989_d;
+-   private final EntityType<?> field_200890_d;
++   private final java.util.function.Supplier<? extends EntityType<?>> field_200890_d;
```
diesieben07: Probably should not remove the field (even if its final) and there is no reason to name the supplier field like that.

diesieben07 (changes requested):

patches/minecraft/net/minecraft/item/SpawnEggItem.java.patch
XFactHD: The code in this method was originally in the registry bake callback and @pupnewfster mentioned that neither `MOD_EGGS` nor the original spawn eggs map should be public API which is why I moved it into `SpawnEggItem` itself.
If everything worked as you would expect from the implementation of the vanilla spawn eggs, a mod would have absolutely no reason to register a dispense behaviour for their spawn eggs. Apart from that, if I haven't missed anything, a mod registering a dispense behaviour in `FMLCommonSetupEvent` would overwrite this anyway due to the fact that the dispense behaviour is registered in a bake callback and the dispense behaviour "registry" is just a item->behaviour map.

src/main/java/net/minecraftforge/registries/GameData.java
XFactHD: Because it seemed appropriate considering what the other bake callbacks are doing.
It makes even more sense with your concerns of who wins when registering the dispense behaviour.

patches/minecraft/net/minecraft/item/SpawnEggItem.java.patch
XFactHD: That would poison the `BY_ID` map with an entry with a null key and necessitate the null check in the original constructor because I would need to pass null as the type.

patches/minecraft/net/minecraft/item/SpawnEggItem.java.patch
XFactHD: What value should I assign to the original field if I make a new field for the supplier?
The supplier field is called this because I didn't change the mapped name of the field.

patches/minecraft/net/minecraft/item/SpawnEggItem.java.patch
XFactHD: Implemented

patches/minecraft/net/minecraft/item/SpawnEggItem.java.patch
diesieben07: Registry bake will occur multiple times, for example it will run whenever you load a world. So no, you cannot override it with `FMLCommonSetupEvent`.

patches/minecraft/net/minecraft/item/SpawnEggItem.java.patch
diesieben07: You probably should assign it the `EntityType` once its available.

patches/minecraft/net/minecraft/item/SpawnEggItem.java.patch
diesieben07: Also, you can make fields private in any class, so just because you move `MOD_EGGS` to e.g. `ForgeHooks` doesn't mean it suddenly can no longer be private.

XFactHD: That is true in this case for `MOD_EGGS` but not for the `BY_ID` map.

XFactHD:
I have moved the call to `SpawnEggItem#finalizeModEggs` to the `FMLCommonSetupEvent` handler in `ForgeMod`, added a method to allow mods registering a `SpawnEggItem` to override the dispense behaviour and added a seperate field for the `Supplier<EntityType>`. The original field gets the `EntityType` assigned to it in `SpawnEggItem#finalizeModEggs`.

patches/minecraft/net/minecraft/item/SpawnEggItem.java.patch
XFactHD: I decided to keep this code in the patch for simplicity and encapsulation.

XFactHD:
Superceded by #8044

PR #7304, opened 2020-09-04
Add support for custom/namespaced WoodTypes

kaydogz:
`WoodType`'s are needed to make sign blocks, and currently, they are bound to the Minecraft namespace (see `Atlases#getSignMaterial`). The only way around this is an AT-filled hack that requires several classes and event subscribers to override GUIs and rendering. This PR will allow namespaced `WoodType`'s, as well as a proper system for creating a modded one, using `WoodType#create`. I also created a test that adds a simple sign using the new method of doing so.

PR #7623, opened 2021-01-25
[1.16.5] Add support for custom WoodTypes

kaydogz:
This is a newer version of #7304, due to my fork breaking. As per the original description:

`WoodType`'s are needed to make sign blocks, and currently, they are bound to the Minecraft namespace (see `Atlases#getSignMaterial`). The only way around this is an AT-filled hack that requires several classes and event subscribers to override GUIs and rendering. This PR will allow namespaced `WoodType`'s, as well as a proper system for creating a modded one, using `WoodType#create`. I also created a test that adds a simple sign using the new method of doing so.


LexManos:
Since this hasn't been relayed yet.
I'm having trouble seeing why any of this is necessary at all.
The map is public, you can insert your own values whenever you want.
The getName() function is only called in getSignMaterial, which is just a helper function to build the initial map.
So we don't need to introduce a big set of changes. People can just add their own WoodType/RenderMaterial to that map at the appropriate time with the existing code.

kaydogz:
getSignMaterial is called at line 93 of SignTileEntityRenderer, so I guess all that should be done is to return the map's value?

LexManos:
Humm, I missed that in my reference search. But yes, if the map has a value it should probably return that value.
You could also have a 2 line change to getSignMaterial that turns the name into a RL and splits it out.
The point is to make the changes as small as possible.

kaydogz:
Ok, should be good now. The additional line in ForgeHooksClient is needed to render the sign.

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -109,11 +109,7 @@
 import java.lang.reflect.Field;
 import java.nio.Buffer;
 import java.nio.ByteBuffer;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Random;
-import java.util.Set;
+import java.util.*;
```
covers1624: Don't collapse imports to *'s

covers1624 (changes requested):
Minor style changes

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
kaydogz: Done.

covers1624 (approved):

PR #7624, opened 2021-01-26
[1.16.5] Fix shaders breaking almost every overlay that uses textures, fixes #7575

### LINKED ISSUE #7575: Having a screen shader applied removes texturing on the hotbar
**Minecraft Version:** 1.16.4

**Forge Version:** 35.1.28

**Logs:** https://pastebin.com/96eMAMrx

**Steps to Reproduce:**
 1. Go to spectator mode
 2. Left click a creeper or enderman
 3. Press the number keys to open up the hotbar

**Description of issue:**
When a screen shader is applied, texturing on the hotbar is removed. I noticed this when adding a colorblind filter option, but it can be seen with vanilla by using the shaders provided by possessing a creeper or enderman in spectator mode.

XFactHD:
This PR fixes post processing shaders causing the hotbar, health, armor, XP bar, helmet overlay, portal overlay, cooldown indicator and others drawing in white.
See #7575 and the issue referencing it aswell as [these](https://imgur.com/a/lcigeIQ) screenshots for examples.

**MOJMAP WARNING: code snippets with official mappings in the Mojira issue comments**
The Mojang bug report can be found here: https://bugs.mojang.com/browse/MC-194675

The bug is most likely "caused" by `ShaderInstance#func_216544_e()` calling `GlStateManager.disableTexture()` and not enabling the texture bound after that call.
It would be great if someone with more GL knowledge could confirm if I am on the right track with this.

The bug is fixed by re-enabling textures after the shaders are drawn.
The bug can also be "hidden" by mods that explicitly enable textures in a RenderOverlayEvent.Pre handler:

- If a ROE.Pre handler for a specific element is used, the handled type and all following types are "fixed".
- If a handler for ROE.Post for a specific element is used, all types following the handled type are "fixed".
- If a handler for ROE.Pre for type all is used, the bug is completely "hidden".

I have included a test mod that binds the desaturate shader to the player. To "activate" the shader, you have to reload resources after entering a world.


ChampionAsh5357:
Seems to be correct. Textures can be enabled and disabled using the symbolic constants either `TEXTURE_1D`, `TEXTURE_2D`, `TEXTURE_3D`, or `TEXTURE_CUBE_MAP` as specified by [3.8.16](https://www.khronos.org/registry/OpenGL/specs/gl/glspec21.pdf). If we trace `GlStateManager::disableTexture`, we can see it grabs a `BooleanState` for `3553` which corresponds to `TEXTURE_2D`, disabling all textures in the game. I would believe this would also have implications in the call within `WorldRenderer#updateCameraAndRender` if the next render call wasn't to a `RenderType` with a `TextureState` which will always call `RenderSystem::enableTexture` during its startup phase.

PR #7625, opened 2021-01-28
AT DataGenerator in LootTableProvider to protected

Parker8283:
Prevent the need for those who extend the LootTableProvider and want to access the DataGenerator from creating another variable within their subclass. I've seen quite a few do this and figured this should just be something that Forge could just take care of.


Cyborgmas:
Not sure this is going to pass, you are saying there is just a slight inconvenience of having a redundant field... not worth an AT I would say.

ChampionAsh5357:
I see no real reason of ATing the loot table provider field as it is unnecessary for the purpose of using the provider as intended.

Parker8283:
Interesting, I was following McJty's tutorials and he required it, and there was no documentation in the Forge docs on how best to do it. If that's the case, I'll close this, but ask if some could point me in the right direction.

TheCurle:
Quite so. It's entirely unnecessary

PR #7628, opened 2021-01-28
Fix wrong eye height when entity size changes (esp. baby animals)

diesieben07:
The size event was fired with the eye height for the old size labelled as the "new eye height", hence the eye height is never changed.


mrp-v2:
Does this fix #7399? Also, #7501 is related to this.

diesieben07:
Yes, it should fix #7399.
I wasn't aware of your PR, sorry.

mrp-v2:
That's fine, now we just have to decide which solution is best for Forge. #6858 may be a useful reference, as it created the `EntityEvent.Size` event.

The advantage to your PR is that it fixes things without making many changes to the existing system.

My PR, however, splits the size event into a 2, one event specifically for the entity size, and one specifically for the entity height. This allows the eye height passed to the `EntityEvent.EyeHeight` event to be based on the result of the `EntityEvent.Size` event.

Feel free to correct me if I've gotten something wrong.

diesieben07:
I'm not quite sure what the purpose of splitting the event would be. Is there a situation where the eye height changes independent of the entity size?

mrp-v2:
The eye height given to the size event is based on the entity's pose and size. If a mod changes the size during a size event, the eye height is then no longer correct based on the size. Splitting the events allows the eye height event to adjust to changes to the entity size.

diesieben07:
If a mod changes the height with the event they obviously also need to change the eye height

mrp-v2:
Yes, but they won't be able to use the entity method that takes a pose and a size to recalculate the eye height. For example, the slime overrides the eye height method to make the eye height lower than default. If a mod changes the slimes size, it won't be able to correctly adjust the eye height because the method to calculate the eye height based on size is protected.

diesieben07:
We could just make the method not protected then.

diesieben07:
Added a parameter for `setNewSize` to automatically update the eye height (`false` by default for backwards compatibility).
Added an accessor method to `Entity` to allow calling `getEyeHeight(Pose, EntitySize)`.

gigaherz (approved):

PR #7630, opened 2021-01-28
[1.16.x] Capability network synchronisation

josephtarbit:
The follow up to #6954.

This PR provides an API for capabilities to sync their data across the network. This resolves the problem with `ItemStack`s not syncing capability data, as well as entire `ItemStack`s being replaced upon sync; while also providing a method for any `ICapabilityProvider` to synchronise data, not just items. Capabilities also have the option of specifying whether their data is 'dirty' and requires syncing. Root `ICapabilityProvider`s have the responsibility of checking whether their child capabilities are dirty or not and to sync them at the appropriate time.

This PR implements `INetworkCapability` support for `ItemStack`s, `Entity`s and `TileEntity`s.


pupnewfster:
At a very brief glance one thing that stands out is the changes to `SEntityMetadataPacket` and to an extent also the `PacketBuffer` changes. One thing forge strives to do is maintain vanilla compatibility (vanilla client connecting to a forge server and a forge client connecting to a vanilla server). I am pretty sure your changes to how these objects/packets get written/read breaks this.

In terms of the entity capability syncing I am guessing forge's [VanillaConnectionNetworkFilter](https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/src/main/java/net/minecraftforge/network/VanillaConnectionNetworkFilter.java) might be able to be used in some way, or at least a similar based thing to send a different packet, or potentially even better maybe send an additional custom packet to sync the extra data. In regards to the item stack encoding/decoding I am not sure of a great way to do it as while I know the server knows if it is a vanilla or forge client, I am not sure offhand if a forge client knows if it is a vanilla or forge server.

One additional complexity is that this because we are passed the breaking changes stage, this also will need to make sure that forge clients *without* this can properly connect to forge servers *with* this, and vice versa.

josephtarbit:
I completely forgot about maintaining vanilla compatibility.  I'll go and amend that now.

Also, does anyone know how I can remove the commit above that didn't have the correct author info which is now messing up the CLA?

gigaherz:
Squash the commits so that you don't have commits authored by BuildTools.

EDIT: If you want to keep multiple commits you can use an interactive rebase to pick which ones to squash and which to keep.

josephtarbit:
> At a very brief glance one thing that stands out is the changes to `SEntityMetadataPacket` and to an extent also the `PacketBuffer` changes. One thing forge strives to do is maintain vanilla compatibility (vanilla client connecting to a forge server and a forge client connecting to a vanilla server). I am pretty sure your changes to how these objects/packets get written/read breaks this.
> 
> In terms of the entity capability syncing I am guessing forge's [VanillaConnectionNetworkFilter](https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/src/main/java/net/minecraftforge/network/VanillaConnectionNetworkFilter.java) might be able to be used in some way, or at least a similar based thing to send a different packet, or potentially even better maybe send an additional custom packet to sync the extra data. In regards to the item stack encoding/decoding I am not sure of a great way to do it as while I know the server knows if it is a vanilla or forge client, I am not sure offhand if a forge client knows if it is a vanilla or forge server.
> 
> One additional complexity is that this because we are passed the breaking changes stage, this also will need to make sure that forge clients _without_ this can properly connect to forge servers _with_ this, and vice versa.

Surely if a mod is registering a network capability, their mod will be required client side anyway so a vanilla client won't be able to connect?

gigaherz:
Forge's connection compatibility checks are not that complex: they only care whether or not the network channels advertise themselves as compatible with the other side's network channels, so, if a network channel doesn't allow `ABSENT`, the mod will be required to be on both sides, and if a network channel doesn't allow `ACCEPTVANILLA` (or whatever the constant name is), then vanilla clients/servers won't be allowed.

Also, just because a mod has set up capability synchronization, doesn't mean it expects the other side to have the mod present. Imagine a mod that sends data to clients, but the client-side mod is optional, some clients may have it, while others won't, and it would be way too convoluted to keep track of which capabilities of which objects need to be synchronized to which clients.

Making sure vanilla keeps working is the best idea here.

josephtarbit:
Any suggestions on how I make the writing of `ItemStack` capability data vanilla compatible because there's no access to any context in those `PacketBuffer` methods.

The only thing I can think of is adding a field to `PacketBuffer` which dictates whether it is vanilla or not. Any other method would require patching every place that an `ItemStack` is encoded which is terrible.

alcatrazEscapee:
Attach the capability NBT data as a special tag (e.g. `__FORGE_SYNC_CAPS__`) onto the normal item stack tag, then intercept + remove on the other end?

josephtarbit:
I'm completely incompetent with git, I can't for the life of me squash the commits to make CLA stop complaining.

gigaherz:
You'll want to read up on "interactive rebase". And then remember that in order to get the squashed commits into your fork, you will need to force-push. Make sure to check the logs after the rebase, before force-pushing anything, to be certain the changes look as intended.

That aside: the build failed because your patches are not clean:
```
> Task :forge:checkPatches FAILED

Patch contains white space change in valid hunk at line 45 (cannot auto fix), file: patches/minecraft/net/minecraft/network/PacketBuffer.java.patch
```
You should run `gradlew checkPatches` before pushing, to make sure it passes.

josephtarbit:
Thanks, I’ll give that a go.

And yes, I saw the patch error before I pushed but I can’t seem to locate the issue, I’ll try replacing chunks of the file.

I’m going to rework how `ItemStack` synchronisation happens as I don’t like the current implementation. Every time a capability is dirty the entire item gets resent which isn’t  very pleasant. I think I’ll sync dirty `ItemStack` capabilities via the entity meta data packet instead.

josephtarbit:
Draft 2.

- Resolves vanilla incompatibilities.
- Avoids sending full `ItemStack`s for item capability syncing.
- Use custom packets instead of patching vanilla packets.

josephtarbit:
Word of warning, this is a breaking change in terms of Forge compatibility as it introduces a new network channel* so will either need to be merged before 1.16.5 RB or for 1.17.

*if there's a way of selectively sending packets please let me know so this can be made backwards compatible with earlier versions of Forge. Right now it's only compatible with vanilla by filtering out all packets from the forge network channel.

loordgek:
i am not a fan of having even more tick events
is this a idea?
https://gist.github.com/1dc039961a1f6e4fbb146cad4c799391

josephtarbit:
> i am not a fan of having even more tick events
> is this a idea?
> https://gist.github.com/1dc039961a1f6e4fbb146cad4c799391

What's the purpose of that?

josephtarbit:
The reason a `TileEntity` tick event is necessary is because it's used to detect capability changes. The alternative method without adding the event is using the world tick event however, that would be far less efficient as you would then have to iterate over every single `TileEntity` in order to check if it needs updating. A `TileEntity` tick event is by far the quickest and cleanest option. Furthermore, a tick event for `TileEntity`s doesn't exactly have a niche use case, I'm sure a lot of modders would utilise it.

alcatrazEscapee:
Okay, I would like to present the (possibly disagreeing with the opinion of Forge based on the way prior PRs have gone) opinion that this should be much more targeted specifically for `ItemStack`s, based on the use cases I am aware of.

1. So, why do we even need to synchronize capability data at all? Many possible reasons - tooltips, rendering, etc. This applies to `ItemStack`s, tile entities, worlds, etc. There are many valid use cases for this. So let's assume there is a need to sync data, from all capability-attachable objects.
2. What tools or mechanisms are **already available** to synchronize capability data? The obvious question being "if there already is an existing mechanism(s) that works for this use case, why do we need a bespoke (and in the PR's current state, rather bloated and sizable overhead layer imo) on top of it? So, what do we have:
    - For `World` or `Chunk` capabilities: These have load and unload events, we have chunk watch and unwatch events, both are uniquely identifiable (e.g. you can identify a chunk to sync via it's position) and can already be synced via means of custom packets. **Most importantly,** these objects will not be overwritten on client. This means if you change your capability data, and synchronize it yourself, it will persist until the object is unloaded or removed (something you can detect).
    - For `TileEntity` capabilities. The same argument applies when the tile entity is owned by your mod. For example, you have `onLoad` methods, `ITickableTileEntity`, and various other mechanisms for syncing the tile entity data, again via the means of custom packets. Now, I can see more issues here with *external* capabilities attached to tile entities - for example, you cannot hook into the `tick` function of a tickable tile entity. However, you still have the same basic guarantee's as before: if you sync some data, *it will persist* until you either sync it again or the tile entity itself is removed on client. (Note: I am not experienced with the use cases in this instance, so if there's someone that has a better explanation of the possibilities of syncing attached capabilities to tile entities, please correct me if there are things that I've missed.)

And now we get to `ItemStack`s. The fundamental difference that I see here, is that vanilla will sync *and override* item stacks, entirely out of the modder's control, which will *overwrite* your capability changes. (Note: item stacks do have a method for allowing sync on item stacks you control - namely, `getShareTag`. However, this does not work for items outside your mod which have capabilities attached to them - my most common use case). This is why item stack capability sync has been such a pain in the rear - if you want to sync capabilities right now in a consistent way, you need to:

- Assume that basically every tick, vanilla may have overwritten your capability data, and so you need to proactively sync it.
- You cannot only sync changes, you cannot assume the client has the same data the last time you synced.
- You cannot even assume that if you and vanilla both sync an item stack (as it so happens containers will do, as they will not automatically send capability-only changes), that it will work correctly (this leads to "flickering" of an item stack in a slot as two things are syncing the same stack.)

And how does one intercept this? i.e. is there a comparable mechanism to how load/unload events, or tickable tile entities allow you to monitor and sync your own data? No. Vanilla can and will sync item stacks completely outside of modder control. This is the primary area where a forge hook *of some sort* is required.

So what do *I* think should be done? Expand `getShareTag` to be usable for externally attached capabilities on item stacks. For item stack attached capabilities, expose the share tag to each capability instance (if required) to write syncable data.


Addendum: a tick event for tile entities sounds like a horrible idea imo. That is a performance sinkhole waiting to happen.

> I'm sure a lot of modders would utilise it.

This strikes fear into my heart.

josephtarbit:
Everything you said about `ItemStack`s was the exact motivation for my original PR but as you said many people wanted this to be applicable to use cases outside of `ItemStack`s so I wrote something that catered for that.

As stated, there are many ways to sync data for other types of capabilities but I think a unified API is much nicer. It removes all lot of the boiler plate code and provides a single place to sync data no matter the type of capability. This means modders write less code and don't need research when their capabilities might need syncing or need data being sent because it's wrapped up in a nice API that hides all of that mess. The API can also take care of different eventualities such as capabilities missing on the client etc. again it boils down to the basic concept of abstraction and not having to reinvent the wheel every time you implement a certain functionality.

It's the same principle as `ICapabilitySerializable`, you can serialize capabilities without that API however it's much simpler having a single interface that provides that functionality.


> Addendum: a tick event for tile entities sounds like a horrible idea imo. That is a performance sinkhole waiting to happen.

The same could be said for any tick event. It should be used sparingly and only when appropriate. A lot of modders already implement ways of handling `TileEntity` ticks which mostly require iterating over the `World`s tile entity list to filter out what they're looking for which is highly inefficient, especially when Minecraft already does this and you could just hook into that. As I said before, this is no different to an `Entity` tick event or any other tick event for that matter. I agree that adding extra tick events should be avoided for circumstances where there's already an appropriate event available however there is no event or anything close to that in order to handle `TileEntity` ticks.

josephtarbit:
In terms of people “thumbs downing” this PR - it would be helpful (and much appreciated) if you provided some actual feedback because it's impossible to improve something if you're not told what's wrong.

src/main/java/net/minecraftforge/event/TickEvent.java
```diff
@@ -84,4 +86,19 @@ public RenderTickEvent(Phase phase, float renderTickTime)
...
         }
     }
+    
+    public static class BlockEntityTickEvent extends TickEvent {
+
+        private final TileEntity blockEntity;
+
+        public BlockEntityTickEvent(Phase phase, TileEntity blockEntity)
+        {
+            super(Type.TILE_ENTITY, blockEntity.getLevel() instanceof ServerWorld ? LogicalSide.SERVER : LogicalSide.CLIENT, phase);
```
Sirttas: Should be Type.BLOCK_ENTITY

patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
```diff
@@ -368,14 +368,25 @@
...
        this.func_184608_ct();
        this.func_205014_p();
-@@ -2232,6 +_,7 @@
+@@ -2231,7 +_,8 @@
+          }
  
           ItemStack itemstack1 = this.func_184582_a(equipmentslottype);
-          if (!ItemStack.func_77989_b(itemstack1, itemstack)) {
+-         if (!ItemStack.func_77989_b(itemstack1, itemstack)) {
++         if (!itemstack1.equals(itemstack, true)) {
```
Sirttas, line 378: What does this change do

patches/minecraft/net/minecraft/inventory/container/Container.java.patch
```diff
@@ -1,17 +1,28 @@
...
+@@ -120,12 +_,21 @@
           ItemStack itemstack = this.field_75151_b.get(i).func_75211_c();
           ItemStack itemstack1 = this.field_75153_a.get(i);
           if (!ItemStack.func_77989_b(itemstack1, itemstack)) {
 +            boolean clientStackChanged = !itemstack1.equals(itemstack, true);
              ItemStack itemstack2 = itemstack.func_77946_l();
              this.field_75153_a.set(i, itemstack2);
- 
+-
++
```
Sirttas: useless patch

patches/minecraft/net/minecraft/network/PacketBuffer.java.patch
```diff
@@ -34,14 +34,20 @@
 +         if (item.isDamageable(p_150788_1_) || item.func_77651_p()) {
 +            compoundnbt = limitedTag ? p_150788_1_.getShareTag() : p_150788_1_.func_77978_p();
           }
- 
+-
```
Sirttas: useless patch

src/main/java/net/minecraftforge/common/capabilities/CapabilityDispatcher.java
```diff
@@ -152,4 +162,43 @@ public void invalidate()
...
-}
+
+    @Override
+    public void encode(PacketBuffer out, boolean writeAll)
+    {
+        for (Map.Entry<String, INetworkCapability> entry : networkCapabilities.entrySet())
+        {
+            if (!writeAll && !entry.getValue().requiresSync())
+                continue;
+            out.writeUtf(entry.getKey(), 0x100);
```
Sirttas, line 173: shouldn't use an constant for the 0x100 ?

src/main/java/net/minecraftforge/fml/hooks/BasicEventHooks.java
```diff
@@ -119,4 +121,16 @@ public static void onPostServerTick()
...
+    {
+        MinecraftForge.EVENT_BUS.post(new TickEvent.BlockEntityTickEvent(TickEvent.Phase.START, blockEntity));
+    }
+
+    public static void onPostBlockEntityTick(TileEntity blockEntity)
+    {
+        if (blockEntity.getLevel() != null)
+            ForgeNetwork.sendBlockEntityCapabilities(blockEntity, false);
+        MinecraftForge.EVENT_BUS.post(new TickEvent.BlockEntityTickEvent(TickEvent.Phase.END, blockEntity));
+    }
```
Sirttas: So this can be the main concern this pr bring, is this event a good or a bad thing it should be discussed.
Maybe it could be better to comment out the event and create a separate pr for it .

src/main/java/net/minecraftforge/network/ForgeConnectionNetworkFilter.java
```diff
@@ -28,7 +28,6 @@
 import net.minecraft.network.ProtocolType;
 import net.minecraft.network.play.server.STagsListPacket;
 import net.minecraft.network.play.server.SUpdateRecipesPacket;
-import net.minecraftforge.fml.network.NetworkHooks;
```
Sirttas: why this import was removed ?

src/main/java/net/minecraftforge/network/ForgeNetwork.java
```diff
@@ -0,0 +1,263 @@
...
+            .decoder(SyncBlockEntityCapabilities::decode)
+            .consumer(SyncBlockEntityCapabilities::handle)
+            .add();
+
+        CHANNEL.messageBuilder(SyncEquipmentSlotCapabilities.class, 0x04, NetworkDirection.PLAY_TO_CLIENT)
+            .encoder(SyncEquipmentSlotCapabilities::encode)
+            .decoder(SyncEquipmentSlotCapabilities::decode)
+            .consumer(SyncEquipmentSlotCapabilities::handle)
+            .add();
+    }
```
Sirttas, line 66: This is also a point to discuss, if there is a better solution to do this or not.

Sirttas:
also a discussion on the subject can be found starting here: https://discord.com/channels/313125603924639766/852298000042164244/855054007991271454

patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
josephtarbit: Technically this isn't necessary for this PR however it essentially means that an item's share tag will be considered when syncing items instead of it's full NBT (items without a share tag will default to full NBT so doesn't break existing behaviour). This allows equipment items with a custom share tag to be considered when syncing.

src/main/java/net/minecraftforge/common/capabilities/CapabilityDispatcher.java
josephtarbit: It's the constant used by vanilla in `PacketBuffer::writeUtf(String)`, `PacketBuffer::readResourceLocation` and `PacketBuffer::readUtf`. I'm not using the 'lengthless' write methods as `PacketBuffer::readUtf` is stripped from the server. I'm not sure what I'd call the constant if I pulled it out into a variable because I don't know what its significance is.

Sirttas: I meant declare a static final for it.

josephtarbit: Yeah I know, not sure what I'd call it though.

src/main/java/net/minecraftforge/fml/hooks/BasicEventHooks.java
shartte: Erm yes, I am getting performance anxiety by just reading this

Sirttas: I'm not a big fan of it either but currently some mods found a way to do it and it's ugly and performance hungry.

Technici4n: Well it's up to them to make sure they sync lazily, Forge probably shouldn't introduce such broken behavior.

Sirttas: > I'm not a big fan of it either but currently some mods found a way to do it and it's ugly and performance hungry.

I was talking about the event some mods managed to do TE ticking events from world ticking event.

josephtarbit: I think I'm gonna scrap the event; if somebody needs it a new PR can be opened. 

Also, in terms of the performance of Forge syncing caps, only caps that ask to be actively synced will be synced and if no caps request a sync then no packet will be sent. Moreover, only dirty data will serialised - I've tried to make it as slimline as possible. A flag is passed into the encode method of the cap to distinguish between syncing (only if requested) and serialisation of the parent object such as if an `ItemStack` is written to a `PacketBuffer`, `Entity` starts getting tracked by a player etc.

josephtarbit:
The `TileEntity` tick event is no more as it wasn't necessary anymore (hurray).

src/main/java/net/minecraftforge/fml/hooks/BasicEventHooks.java
shartte: Yeah i was only really concerned about the two event dispatches in the tight loop. We've been burned by the dispatch in the ItemStack ctor and this just gave me PTSD :)

Sirttas (approved):

josephtarbit:
Closing in favour of #7952 which I will backport if merged.

PR #7636, opened 2021-01-30
[1.16.5] Shutup datafixer warning for modded entity types

XFactHD:
This PR proposes a solution to make DFU shutup about modded entities not having a datafixer.
The only caveat is that in case a modder does register a datafixer, it will probably not work anymore with this fix.

### DIFF
--- a/patches/minecraft/net/minecraft/util/Util.java.patch
+++ b/patches/minecraft/net/minecraft/util/Util.java.patch
@@ -1,0 +1,11 @@
+--- a/net/minecraft/util/Util.java
++++ b/net/minecraft/util/Util.java
+@@ -200,7 +200,7 @@
+       try {
+          type = DataFixesManager.func_210901_a().getSchema(DataFixUtils.makeKey(SharedConstants.func_215069_a().getWorldVersion())).getChoiceType(p_240990_0_, p_240990_1_);
+       } catch (IllegalArgumentException illegalargumentexception) {
+-         field_195650_a.error("No data fixer registered for {}", (Object)p_240990_1_);
++         field_195650_a.debug("No data fixer registered for {}", (Object)p_240990_1_);
+          if (SharedConstants.field_206244_b) {
+             throw illegalargumentexception;
+          }


TheCurle:
This does not seem like a good tradeoff. 
There are mods that benefit from using a Datafixer, and there are people (read: Forge support) that benefit from not having to scroll through 14 pages of "no datafixer registered" in the logs..

Can you verify that datafixers *do* work, given that you gave a "probably"?

TheCurle:
Instead of adding an extra check, could you not just... remove the line that logs to the console?

pupnewfster:
Or lower it to a debug level warning.

XFactHD:
I made sure that the call to `Util.func_240976_a_()` still runs for vanilla. The reason I said "probably" is that I have not looked deep enough into DFU to know if a datafixer still works if that call is missing but I assume that it wouldn't.

I would not really have expected that there are mods that use the datafixers.
I will remove the logging, I think lowering it to debug is useless because most of the time people will be looking at a debug log anyway which would contain the lines. Should I comment the line out or remove it outright?

TheCurle:
Can you gate it behind a config option?

XFactHD:
Will look into it

XFactHD:
There we go, logger is disabled by a config flag in the fml config (Forge config doesn't work because it has not been loaded when RegistryEvents fire). I have verified with the test mod from #7621 that it does silence the "error" and can be toggled.

patches/minecraft/net/minecraft/util/Util.java.patch
```diff
@@ -0,0 +1,11 @@
+--- a/net/minecraft/util/Util.java
++++ b/net/minecraft/util/Util.java
+@@ -200,7 +200,7 @@
+       try {
+          type = DataFixesManager.func_210901_a().getSchema(DataFixUtils.makeKey(SharedConstants.func_215069_a().getWorldVersion())).getChoiceType(p_240990_0_, p_240990_1_);
+       } catch (IllegalArgumentException illegalargumentexception) {
+-         field_195650_a.error("No data fixer registered for {}", (Object)p_240990_1_);
++         if (!net.minecraftforge.fml.loading.FMLConfig.disableDatafixerWarning()) field_195650_a.error("No data fixer registered for {}", (Object)p_240990_1_);
```
pupnewfster: If you put the if statement on its own line above (without changing indentation of the log line the patch will be slightly smaller.

XFactHD: Done

patches/minecraft/net/minecraft/util/Util.java.patch
```diff
@@ -0,0 +1,10 @@
+--- a/net/minecraft/util/Util.java
++++ b/net/minecraft/util/Util.java
+@@ -200,6 +200,7 @@
+       try {
+          type = DataFixesManager.func_210901_a().getSchema(DataFixUtils.makeKey(SharedConstants.func_215069_a().getWorldVersion())).getChoiceType(p_240990_0_, p_240990_1_);
+       } catch (IllegalArgumentException illegalargumentexception) {
++         if (!net.minecraftforge.fml.loading.FMLConfig.disableDatafixerWarning())
+          field_195650_a.error("No data fixer registered for {}", (Object)p_240990_1_);
```
LexManos: better fix, as logging already has a config file {the log4j config} change the 'error' to 'debug' and call it good. None of the other things are needed.

XFactHD: Sure, I can do that, though I don't see how this is useful. Most of the time people will be looking at the debug log which would still contain those lines and the log4j config is not exactly "user servicable", especially considering the fact that certain "launchers" still fuck around with the logging config.

LexManos: Yes, but these things DO need to be logged. Modders should be addressing them.
We're not caring about users in the debug log. Users don't give a crap if thats spammy or not. Just look at all the popular mods that cause dozens of stacks to be dumped every run.
Honestly, I dont think this message is that big of a deal. Especially not worth having a custom config for.

XFactHD: Fair enough

XFactHD:
Changed error to debug, removed config.

PR #7638, opened 2021-02-02
Add stack_sensitive transformer

Bricktricker:
This PR adds a stack_sensitive coremod transformer to change calls to non stack sensitive methods to stack sensitive methods,
similar to the field_to_instanceof transformer added in f169c993487609954c78103de3091d32c1d9cacc
Currently all calls to Item#isDamageable() (func_77645_m) are replaced with calls to IForgeItem#isDamageable(ItemStack), but I can add more if needed.

The script does the following transformations:
1. `itemstack.getItem().foo()` => `itemstack.getItem().bar(itemstack)`
2. `item.foo() => item.foo(this)`, if 'this' is an ItemStack
3. `void bizz(ItemStack stack){ item.foo() }` => `void bizz(ItemStack stack){ item.bar(stack) }`

The script assumes the following things:
1. The non stack sensitive method takes no arguments
2. The stack sensitive method only takes only one argument, an ItemStack
3. The non stack sensitive method and the stack sensitve method have the same return type
4. The stack sensitive method is a interface method

The added `findStackSensitiveChecks` task in the build.gradle uses the output from `reobfJar` as input, while the `findFieldInstanceChecks` uses the output from the `createJoinedSRG` task.
It looks like createJoinedSRG jar does not have the forge patches applied. Because forge changes some method signatures
(e.g. here: <https://github.com/MinecraftForge/MinecraftForge/blob/83fee0b3d43c8541be2b5e6f40f53e3ccf97c4ef/patches/minecraft/net/minecraft/item/ItemStack.java.patch#L26L28>) these changes are not present in the createJoinedSRG created jar.
Therefore the generated JSON file would contain the old method signature (`ItemStack(IItemProvider, int)`) not the new one (`ItemStack(IItemProvider, int, CompoundNBT)`).
Is this correct or am I missing something?


Cyborgmas:
My issue with these is that you have to remember about these coremods when looking at the code, but it's been decided that it's a worthwile tradeoff. However, could you add some javadocs to the forge methods here (so isDamageable for now), that states at least what function it replaces and even perhaps all the places where it gets transformed (might be overkill).

Also you have some minor discrepancies with tabs vs space in the transformer and the task scripts

LexManos:
That's a good idea, anything that we magically deprecate/move via a coremod should be @Deprecated and commented properly. 

Yes, these things make the source code a bit more of an issue to follow. But the trade off is maintenance cost. As long as we point modders in the right direction in the right spots. They don't need to know random caller of that function was changed in some random class.

Bricktricker:
Fixed the tabs/spaces and added some documentation. Is that enough or should I list all transformed places?

Cyborgmas:
I think that's fine, I would just add to the deprecation comment that vanilla calls are overriden using the coremod when applicable.
So that you have a message both on the replaced and replacement function.

TheCurle:
This is still a highly useful feature. Are you able to rebase this onto latest changes to move it onto further stages?

Bricktricker:
Rebased the PR, hope I haven't screwed something up.

LexManos:
As it sits, with all the changes happeing in 1.17 I think this will be a better spot to address these type of cleanups. So i've taged this for 1.17 and will keep it in mind during the update process.

LexManos:
Going to close this, yes it's a good idea. But we've been discussing other methods of making the patches more manageable on our discord. It's on our todo, just taking time.

PR #7639, opened 2021-02-03
Disable syncing of Feature and Biome registries 

### LINKED ISSUE #7451: Biome and Feature registries are being forced synced by Forge when it shouldn't
With the rise of datapack worldgen in 1.16.2, people are now able to make datapacks that have custom biomes or configured features and so on entirely serverside with vanilla clients able to connect. The server/client interactions are performed safely by vanilla. 

However, my structure mod which contains a few features won't allow any Forge client without my mod to connect to the server with my mod. Vanilla clients are able to connect just fine however. It seems that the biome and feature registry are missing the .disableSync() method which means Forge client will attempt to sync these two worldgen registries and will fail when the server has a serverside mod that adds to those registries. Picture is from a friend who looked into Forge and found why:
image

Bug report I had gotten on my mod about this issue and how we narrowed it down to something with only Forge clients without my mod:
https://github.com/TelepathicGrunt/RepurposedStructures/issues/33

To be clear, vanilla will handle the syncing of Biomes and generating ConfiguredFeatures safely. The client doesn't actually need the Feature registry at all as they are only exposed to the blocks the feature places. This is why worldgen mods are able to work completely serverside with vanilla clients as long as they place only vanilla blocks. Even biome colors are correctly sent to the vanilla client from the server.

TelepathicGrunt:
Resolves https://github.com/MinecraftForge/MinecraftForge/issues/7451

Basically, Forge clients without a biome/feature mod cannot connect to a server with the mod. Vanilla clients can connect to the server with the mod.

The reason for this is Minecraft will send the necessary worldgen data from the server to the client which is how worldgen datapack with custom biomes and such will work perfectly fine entirely server side. Forge, however, forces the biome and feature registry to be exactly the same when a Forge client tries to connect to a server. This is not need and this PR turns off the syncing for those two registries to better match vanilla behavior.

I tested 36.0.14 client connecting to this PR's server and tested this PR's client connecting to 36.0.14 server. Both without worldgen stuff and it works both ways across the version difference safely.

I tested a 36.0.14 client with worldgen mod connecting to this PR's server and it allows the connection properly which makes sense as worldgen mods are useless on client.

Then lastly, I connected the 36.0.14 client to this PR's server with the test mod on and register worldgen stuff on the server. Client was able to connect just fine.

Of course, the 36.0.14 server with worldgen mods rejected this PR's client as the server is the Forge without this PR.

No major issues came up that I can tell from mismatch Forge versions with the testing I done above


LexManos:
This doesn't look bad, my only concern is the codecs used to encode the biomes/features dynamic registries when sending them to the client.
Do those use the registry names from the seed registry or are the sending the full object specs?
I legitimately do not know, as I haven't looked into it. If you can do so and writeup what data is used to sync the registries for these, that'd be great.

More for documentation reasons then me not trusting that it works.

TelepathicGrunt:
The feature registry does not have a packet codec and does not get sent to the client at all. The client is entirely unaware of what features are on the server. 

The biome does have a packet codec except when sending it to the client, it strips out all configuredfeature, configuredcarvers, configuredstructures, and all mob spawn info. The info the client receives about the biome are the following:

-precipitation
-temperature
-temperature_modifier
-downfall
-category
-depth
-scale
-effects (sound, plant color, water color, sky, particles, music, etc)

The generation settings and mob spawn info sections are filled with `BiomeGenerationSettings.DEFAULT_SETTINGS, MobSpawnInfo.EMPTY` which both are empty. 

Basically, the client only receives info about the biome that does not rely on the other worldgen registries or entities. Most of it is for rendering such as particles or color. Only issue I could see is if a mod on the server has custom music or particles along with their biome, only then would a client without the mod might default to the wrong settings for the biome visually or sound wise. But at that point, the mod is already one that should be on both sides anyway. Vanilla datapacks can only use what vanilla has so they do not have this issue with particles/sounds.

PR #7641, opened 2021-02-04
Add IExtensibleEnum to GrassColorModifier

Drullkus:
Makes `GrassColorModifier` implement `IExtensibleEnum` so modders can create their own `GrassColorModifier` entries for their biomes to custom-modify the biome color based on sampling position. This can be used to create special color patterns like noise or a swirly pattern with the grass tinting.

### DIFF
--- a/patches/minecraft/net/minecraft/world/biome/BiomeAmbience.java.patch
+++ b/patches/minecraft/net/minecraft/world/biome/BiomeAmbience.java.patch
@@ -9,7 +9,7 @@
        NONE("none") {
           @OnlyIn(Dist.CLIENT)
           public int func_241853_a(double p_241853_1_, double p_241853_3_, int p_241853_5_) {
-@@ -226,18 +226,32 @@
+@@ -226,18 +226,36 @@
        };
  
        private final String field_242543_e;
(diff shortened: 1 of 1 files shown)


patches/minecraft/net/minecraft/world/biome/BiomeAmbience.java.patch
```diff
@@ -0,0 +1,46 @@
...
++         return delegate.modifyGrassColor(p_241853_1_, p_241853_3_, p_241853_5_);
++      }
+ 
+       private GrassColorModifier(String p_i241940_3_) {
+          this.field_242543_e = p_i241940_3_;
+       }
+ 
++      private ColorModifier delegate;
++      private GrassColorModifier(String name, ColorModifier delegate) {
++         this.field_242543_e = name;
```
pupnewfster: This line should probably instead be `this(name);` just to make it so that less of a patch is needed if vanilla changes the base version.

pupnewfster (commented):
I am not 100% positive (so if someone else could confirm that would be great), but looking at other extensible enums, I believe this is missing something along the lines of:
```java
@Override
@Deprecated
public void init() {
   NAME_TO_MODIFIER_MAP.put(this.getName(), this);
}
```

patches/minecraft/net/minecraft/world/biome/BiomeAmbience.java.patch
Drullkus: Done 👍

Drullkus:
> 
> 
> I am not 100% positive (so if someone else could confirm that would be great), but looking at other extensible enums, I believe this is missing something along the lines of:
> 
> ```java
> @Override
> @Deprecated
> public void init() {
>    NAME_TO_MODIFIER_MAP.put(this.getName(), this);
> }
> ```

Oops, I missed this reply. Resolved now too 👍

PR #7525, opened 2020-11-29
Allow Chunk loading Tickets to opt-into forcing chunk ticks.

covers1624:
By default, vanilla only ticks chunks when players are nearby, regardless of what ticket is loading the chunk (this is true even with [/forceload](https://minecraft.gamepedia.com/Commands/forceload#Note)).

This patch simply adds an extra flag a ChunkTicket can opt-into, in order to force chunks to tick regardless of nearby players.

Tested with a modified version of [ChickenChunks](https://github.com/TheCBProject/ChickenChunks), made some crops, chunk loaded them, teleported away, set random tick speed to insane, reset random tick speed, teleported back, crops are grown.

PR #7642, opened 2021-02-04
Reinstate's Forge's Chunk Manager

pupnewfster:
This PR supersedes #7517, by reinstating forge's chunk manager to provide modders a way to keep track of chunks they want forced in a persistent way and in one that allows for multiple mods (or even the same mod) to register tickets for the same chunk unlike vanilla's forced chunk system which if two mods try to use (with default forced chunk tickets) and then one removes the ticket then the chunk will be unloaded even though the other mod may still want it loaded. This PR currently adds support for managing two types of ticket owners for modders: BLOCK (BlockPos) and ENTITY (UUID).

One thing to note about this PR is that it does not override `ServerWorld#getForcedChunks` as that method is primarily used for the `ForceLoadCommand` which Forge's Chunk Manager does not adjust. Instead we patch one of the calls to it to a method we add to check if any chunks are forced in the world (that is used to see if a world should keep ticking) for purposes of doing a quick lookup check without having to merge a potentially decent number of sets. We do not override the call in DebugOverlayGui that shows how many chunks are forced as in my opinion that is a more useful debug thing for purposes of the force load command and merging a bunch of sets to get the total count seems rather pointless to me.

### DIFF
--- a/patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
+++ b/patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
@@ -61,6 +61,15 @@
           }
  
           this.func_229856_ab_();
+@@ -338,7 +343,7 @@
+       this.func_147488_Z();
+       this.field_211159_Q = false;
+       iprofiler.func_219895_b("entities");
+-      boolean flag3 = !this.field_217491_A.isEmpty() || !this.func_217469_z().isEmpty();
++      boolean flag3 = !this.field_217491_A.isEmpty() || net.minecraftforge.common.world.ForgeChunkManager.hasForcedChunks(this); //Forge: Replace vanilla's has forced chunk check with forge's that checks both the vanilla and forge added ones
+       if (flag3) {
+          this.func_82742_i();
+       }
 @@ -395,7 +400,7 @@
              }
  
--- a/patches/minecraft/net/minecraft/server/MinecraftServer.java.patch
+++ b/patches/minecraft/net/minecraft/server/MinecraftServer.java.patch
@@ -33,7 +33,15 @@
        ServerWorld serverworld = this.func_241755_D_();
        field_147145_h.info("Preparing start region for dimension {}", (Object)serverworld.func_234923_W_().func_240901_a_());
        BlockPos blockpos = serverworld.func_241135_u_();
-@@ -566,6 +569,7 @@
+@@ -479,6 +482,7 @@
+                ChunkPos chunkpos = new ChunkPos(i);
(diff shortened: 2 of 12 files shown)


patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
```diff
@@ -61,6 +61,15 @@
           }
  
           this.func_229856_ab_();
+@@ -338,7 +343,7 @@
+       this.func_147488_Z();
+       this.field_211159_Q = false;
+       iprofiler.func_219895_b("entities");
+-      boolean flag3 = !this.field_217491_A.isEmpty() || !this.func_217469_z().isEmpty();
++      boolean flag3 = !this.field_217491_A.isEmpty() || net.minecraftforge.common.world.ForgeChunkManager.hasForcedChunks(this);
```
marchermans: Why does this not matter anymore: `!this.func_217469_z().isEmpty()`, I more orless expected this PR to just extend vanilla behaviour, and append another or case. Instead it also removes one here.

Not saying this is wrong, but a bit of documentation on why this happens here is needed I think, since it feels counter intuitive.

marchermans (changes requested):
Generally a good PR.

Could you add a comment on why we replace that second is empty call, instead of just extending the logic.

marchermans (approved):

covers1624:
This currently has some limitations making it unsuitable for my uses in ChickenChunks as Chunk ticks(weather, mob spawning, random block ticks, see #7525) aren't fired. This is something I believe should be a built-in feature of this system as, from my perspective, the main uses for this are Chunk Loading mods which require these ticks to be fired, with a few exceptions such as mods keeping tiles/entities loaded without the need for chunk ticks.

pupnewfster:
> This currently has some limitations making it unsuitable for my uses in ChickenChunks as Chunk ticks(weather, mob spawning, random block ticks, see #7525) aren't fired. This is something I believe should be a built-in feature of this system as, from my perspective, the main uses for this are Chunk Loading mods which require these ticks to be fired, with a few exceptions such as mods keeping tiles/entities loaded without the need for chunk ticks.

I believe this is addressed now as of the latest commit. I also fixed the following issues relating to #7525:
- Extra constructor was never added to exc file so params and stuff would not be properly mapped if mappings get defined
- There was no way to actually add tickets of the new type via the ticket manager and server chunk provider, and instead was requiring people to AT multiple things to actually add the tickets of said types.

covers1624:
Changes look good and should properly cover ticking ticket use cases.

PR #7645, opened 2021-02-06
Add option for linear filtering of text texture

Nightenom:
When scaling text in unrounded scales (eg. 1.5x) the nn (nearest neighbor) filtering makes the text look weird and in small scales unreadable. This PR adds the ability to turn on linear filtering for text texture making a bit more readable but might seem a bit more transparent.
note: I don't know enough about mipmaps and if they would bring improvement so they are kept disabled

Before:
image

After:
image

### DIFF
--- a/src/main/java/net/minecraftforge/client/ForgeRenderTypes.java
+++ b/src/main/java/net/minecraftforge/client/ForgeRenderTypes.java
@@ -19,18 +19,21 @@
 
 package net.minecraftforge.client;
 
+import com.mojang.blaze3d.systems.RenderSystem;
+import net.minecraft.client.Minecraft;
 import net.minecraft.client.renderer.RenderState;
+import net.minecraft.client.renderer.RenderState.TextureState;
 import net.minecraft.client.renderer.RenderType;
+import net.minecraft.client.renderer.RenderType.State;
 import net.minecraft.client.renderer.texture.AtlasTexture;
+import net.minecraft.client.renderer.texture.TextureManager;
 import net.minecraft.client.renderer.vertex.DefaultVertexFormats;
 import net.minecraft.client.renderer.vertex.VertexFormat;
 import net.minecraft.util.ResourceLocation;
 import net.minecraftforge.common.util.NonNullLazy;
 import net.minecraftforge.common.util.NonNullSupplier;
 import org.lwjgl.opengl.GL11;
-
-import net.minecraft.client.renderer.RenderState.TextureState;
-import net.minecraft.client.renderer.RenderType.State;
+import java.util.function.Supplier;
 
 @SuppressWarnings("deprecation")
 public enum ForgeRenderTypes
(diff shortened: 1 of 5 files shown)


src/main/java/net/minecraftforge/client/ForgeRenderTypes.java
```diff
@@ -206,5 +230,43 @@ public static RenderType layeredItemTranslucent(ResourceLocation locationIn) {
...
         }
+
+        public static RenderType getText(ResourceLocation locationIn) {
+            RenderType.State rendertype$state = RenderType.State.getBuilder()
+                    .texture(new CustomizableTextureState(locationIn, () -> ForgeRenderTypes.enableTextTextureLinearFiltering, () -> false))
+                    .alpha(DEFAULT_ALPHA)
+                    .transparency(TRANSLUCENT_TRANSPARENCY)
+                    .lightmap(LIGHTMAP_ENABLED)
+                    .build(false);
+            return makeType("forge_text", DefaultVertexFormats.POSITION_COLOR_TEX_LIGHTMAP, 7, 256, false, true, rendertype$state);
```
Nightenom: Not exactly sure if should keep the vanilla name or add the "forge_" prefix

src/test/java/net/minecraftforge/debug/client/rendering/LinearTextTextureFilteringTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+import net.minecraft.client.gui.screen.MainMenuScreen;
+import net.minecraftforge.client.ForgeRenderTypes;
+import net.minecraftforge.client.event.GuiScreenEvent.DrawScreenEvent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.api.distmarker.Dist;
+
+@Mod(LinearTextTextureFilteringTest.MODID)
+@Mod.EventBusSubscriber(value = Dist.CLIENT)
```
aimiebell, line 30: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

OrionDevelopment: @aimiebell These method names are fine....

src/main/java/net/minecraftforge/client/ForgeRenderTypes.java
marchermans: Forge prefix is good.

marchermans (approved):

src/main/java/net/minecraftforge/client/ForgeRenderTypes.java
Nightenom: It's replacing the vanilla ones, so I was like it should have the same key

marchermans: Yeah that is a valid point. But this works as well. Especially since the rendering types are more or less named instances. And this makes it clear that it is not the original text rendering type. But an adapted one by forge.

gigaherz (changes requested):
Can we make it so if the config is disabled, the **original** vanilla render types are still used? I'm concerned existing stuff that messes with rendering could possibly rely on the original names...

something like say
```
if (ForgeRenderTypes.enableTextTextureLinearFiltering)
  return <forge call>;
<vanilla return>
```

Nightenom:
> Can we make it so if the config is disabled, the **original** vanilla render types are still used? I'm concerned existing stuff that messes with rendering could possibly rely on the original names...

This is rather problematic because of calls that are run once per client session

Nightenom:
Rebased, can I get any progress info on this?

gigaherz (approved):

PR #7646, opened 2021-02-06
Fix updateItemStackNBT not properly being called using a field transformer

KnightMiner:
When an item is read from NBT and it contains a NBT tag, it calls `getItem().updateItemStackNBT()` within the NBT constructor. However, Forge does not initialize the item delegate until the end of the consructor, meaning at that time `getItem()` always returns `Items.AIR`.

This fixes the issue by setting the delegate earlier. Since it was just pulled out of the registry, the field will never returns null so its null safe. In addition, since only one other constructor needed the delegate set, it now directly sets it as well.

A few implementation notes:
* As per request on Discord, this PR includes a field transformer to replace all references to `this.item` with `this.getItem()`
* Since only the two constructors need to directly access `this.item` now, `getRawItem()` was removed.
     * In the NBT constructor, a local variable `rawItem` was created instead to set the delegate without it getting transformed to `this.getItem().delegate`.
     * In the `IItemProvider` constructor, it uses the `IItemProvider` parameter directly meaning its an extra call to `IItemProvider#asItem`. That method is trivial or cached in its implementations so I figured it won't be a hit, though I could swap it for a similar local variable using a slightly larger patch
* `forgeInit` no longer sets the delegate. The null check on `this.item` was replaced with a null check to `this.delegate`, which will only be null when creating `ItemStack#EMPTY` using a null item provider (meaning the empty item stack will not get capabilities). Item stacks that are empty via stack size will receive capabilities as before.
* Removed the patch on `ItemStack#isEmpty` as `getItem()` can be accurately checked against air. The `this.getItem() != null` check is redundant as its never null, but shouldn't impact performance

----

To test that this PR does not break item stack delegates, simply load a world that contains item stacks in some inventory. They should be non-empty.

To test this PR fixes`updateItemStackNBT`, place a breakpoint in `SkullItem::updateItemStackNBT`, then cheat in a player head containing NBT. Grabbing the item in creative and renaming it in an anvil is a sufficient method. After adding the NBT, exit and rejoin the world and the breakpoint should be hit.

### DIFF
--- a/patches/minecraft/net/minecraft/item/ItemStack.java.patch
+++ b/patches/minecraft/net/minecraft/item/ItemStack.java.patch
@@ -13,13 +13,13 @@
           return Optional.ofNullable(p_234704_0_.field_77990_d);
        })).apply(p_234698_0_, ItemStack::new);
     });
-+   private net.minecraftforge.registries.IRegistryDelegate<Item> delegate;
++   private final net.minecraftforge.registries.IRegistryDelegate<Item> delegate;
 +   private CompoundNBT capNBT;
 +
     private static final Logger field_199558_c = LogManager.getLogger();
     public static final ItemStack field_190927_a = new ItemStack((Item)null);
     public static final DecimalFormat field_111284_a = Util.func_200696_a(new DecimalFormat("#.##"), (p_234699_0_) -> {
-@@ -109,14 +112,18 @@
+@@ -109,14 +112,19 @@
        p_i231596_3_.ifPresent(this::func_77982_d);
     }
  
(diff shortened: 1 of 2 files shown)


KnightMiner:
Updated to instead of swapping `this.getItem()` for `item`, to set `this.delegate = this.item.delegate`. Means its set twice, but thats a trivial operation and shouldn't matter. Note if calling `this.item` in the constructor directly was a problem as I was told over Discord, some of the other ItemStack constructors will have the same problem.

KnightMiner:
Updated PR with a field transformer from `this.item` to `this.getItem()`, moved delegates to the constructor, and removing the now unneeded `getItemRaw()`. Main post updated with a few implementation notes for those interested.

PR #7649, opened 2021-02-07
Make EntityDamageSource actually check "difficultyScaled" field

MonsterPoops:
Hello friends. I apologize in advance if I've done something wrong. I am really new to this.

I am working on a mod that heavily rebalances the core game and (hopefully) makes the gameplay experience feel more consistent, focused, and the different gameplay features more coherent with each other.

One thing I need to do for this is disable vanilla's "damage difficulty scaling". It seemed easy enough at first to just use an AT to always set "DamageSource#difficultyScaled" to false during the "LivingAttack" Event. Unfortunately, many damage sources completely ignore this field for some reason. Including all "EntityDamageSource"-s.

What these patches do is just change EntityDamageSource to first check it's super, which does use said field, before using it's default logic. It also updates DamageSource so that all the methods creating an EntityDamageSource set the "difficultyScaled" flag to true, so they work as intended.

I'm sure you get the idea already, but for the sake of doing this right, here is my use example:

https://pastebin.com/cu1cPmdW


LexManos:
No, Mojang does not want this damage scaled, so just telling it to be would cause way to many side effects/changes to vanilla.
File a bug report with Mojang and hope they fix it.

MonsterPoops:
I think you have it backwards my guy. The damage source _**is**_ scaled by default.

The issue was that when it checks for whether or not it should scale it via the "isDifficultyScaled" method, it doesn't use the "isDifficultyScaled" field, it checks whether the entity is a living entity and not a player entity.

All the patch did was make it also consider the "difficultyScaled" field, so that it would not scale if a mod set that to false. It could not force it to be scaled if it would not be by default.

I disagree there would be any side effects. I looked into it thoroughly, there's nothing the change would have interacted with otherwise.

Either way, I've taken care of the issue for myself by just replacing the class.

PR #7651, opened 2021-02-10
Improve compatibility between world-gen datapacks and mods that add structures.

Won-Ton:
## About

This PR adds a custom worldgen datapack loader which applies some targeted manipulation of each registry entry's  json data to do things like merge content from duplicate files across datapacks, and restore content that mods might have registered in the WorldGenRegistries but have been discarded by the presence of a datapack entry.

In particular this PR targets the `DimensionSettings` registry, because mods need to add their structures and separation settings to a map stored in `DimensionStructureSettings` (which is a field on the DimSettings). Some mods do this via datapack which conflicts with those that do it programatically (the latter group's content essentially gets discarded and isn't used at all).

This system could be used to address other similar conflicts (jigsaw pools might be another candidate) and could serve as a useful hook point for other world-gen modification apis.

The system is entirely optional and is disabled by default in the forge common config. It delegates to the vanilla loader when disabled or when there are no injections for a given registry entry type.

Cheers!

> _Can also confirm: "I'm here live. I'm not a cat"_

## Testing

Attached are two datapacks which alter the overworld `DimensionSettings` to only generate villages in one, and ruined portals in the other. They also both slightly decrease the spacing of each structure type to help make it visible when they're active.

1. Test each pack individually to confirm they both exclude all but the one structure type (use `/locate` to try find other structures.
2. Test both packs loaded at once. Whichever is ordered on top in the world datapacks menu should end up having an effect on the world and not the other one.
3. Test the datapack merging function of this PR:
  3.1 In the forge-common config set `mergeDataPackWorldGenData = true` and `worldGenDataMergeStrategy = "ALL_NAMESPACES"`.
  3.2 Create a new world with both datapacks added in any order.
  3.3 Join the world. You should be able to locate both ruined portals and villages and no other structure types.
4. Test the registry injection function of the PR:
  4.1 In the forge-common config set `injectMissingWorldGenData = true`  and `worldGenDataMergeStrategy = "ALL_NAMESPACES"`.
  4.2 Create a world with the 'Just_Portals' datapack enabled.
  4.3 Join the world. You should be able to locate all overworld structures normally with the ruined portals being noticeably closer together than normal.

**Test Datapacks:**
[Just_Portals.zip](https://github.com/MinecraftForge/MinecraftForge/files/5983368/Just_Portals.zip)
[Just_Villages.zip](https://github.com/MinecraftForge/MinecraftForge/files/5983370/Just_Villages.zip)


src/main/java/net/minecraftforge/registries/injection/MergeStrategy.java
```diff
@@ -0,0 +1,23 @@
+package net.minecraftforge.registries.injection;
+
+public enum MergeStrategy {
```
50ap5ud5: Please be consistent with how you add curly brackets. It seems you use a new line for each curl bracket in other Forge related classes, but this class appears to be an exception. Just a minor nitpick :P.

src/main/java/net/minecraftforge/registries/injection/RegistryEntryModifier.java
```diff
@@ -0,0 +1,147 @@
...
+    {
+        RegistryKey<E> typedEntryKey = getTypedKey(entryKey);
+
+        for (Injector<E> injector : injectors)
+        {
+            injector.inject(typedEntryKey, entryData);
+        }
+    }
+
+    // Returns the type-bound instance of the registry key.
```
50ap5ud5: Might I suggest using a Javadoc comment on this method? For consistency purposes, perhaps you can use `/** Returns the type-bound instance of the registry key. */ ` on a single line, so that modders can see this comment without going into the class definition.

50ap5ud5 (changes requested):
Thanks for making this! I have uses cases for this PR in my mods. The patches are clean, specific and work well. There's just a few minor code style consistency issues you could fix but this is mostly well done overall.

src/main/java/net/minecraftforge/registries/injection/MergeStrategy.java
Won-Ton: Ahah thanks good spot, it's fighting my instincts _and_ my IDE to newline opening brackets 😅 I'll do another pass for consistency, think I noticed a try/catch somewhere too that I was meaning to touch up. Nitpicks very much welcome :D

TheCurle:
Okay, this is a big one. 

First, fix the merge conflict (it looks like a rebase should do), then fix the license headers that are completely missing on some classes.

Second, the biggest problem we need to solve here is compatibility between datapack structures, and those added in BiomeLoadingEvent.
@TelepathicGrunt can weigh in there, if there is a reasonable way to alleviate the issues with this PR.

We can go from there.

Won-Ton:
Tidied up.

The BiomeLoadingEvent is only used to add/remove the StructureFeatures to Biomes. The separation settings _for_ those structures are a global thing and part of the ChunkGenerator's domain (configured within the DimensionSettings).

TelepathicGrunt:
This PR does not allow modders to do dimensional blacklisting of their structures through code as the WorldGenRegistry noise setting entries can be reused for multiple dimension. Thus making this PR not a suitable replacement for the widespread WorldEvent.Load hack that people are using to make sure their structures only spawn in the exact dimension that they want. And pushing said customization onto users to have to make their own datapacks from scratch to replicate what modders want will not work well. And modders have no way of using this system to keep their structure from say, JungleLand dimension that uses the overworld noise setting file since this PR is additive and is firing before the worlds are even constructed. 

A better hook would have to be where dimensions themselves are constructed, slip in an event to intercept the spacing map, make a new deep copy of the map to prevent messing with other dimension's instances, and have structure mods use that event to add, remove, or adjust their structure's spacing. And after that event, fires, Forge could hook in as well to provide an external way for users to fine turn the spacing even more similar to this PR. And then that finalized map is set into the chunk generator. This theoretical event would cover the actual use-cases and problems that structure mods are facing. Until then, it will be a struggle to get people to drop WorldEvent.Load in favor of this current PR.

Won-Ton:
This PR is intended to address a fundamental issue that arises when you combine mod-added settings and datapack added settings.

I'd consider dynamic "blacklisting" or trying to accommodate dimensions using overworld's DimensionSettings when they could/should be registering & using their own (if they indeed need independent configuration from overworld's), as out of scope of the PR.

This maybe lays the groundwork for you to build those extra features in another PR. With mojang moving towards a more statically configured system I don't see those sorts of things as easily maintainable, personally.

PR #7653, opened 2021-02-11
Adding Tags for Ices

OneLemonyBoi:
This PR adds tags for vanilla's different ices. Many mods are adding different kinds of ices, and it would be great to unify them.


MuteTiefling:
🍿

DaemonUmbra:
This doesn't compile

OneLemonyBoi:
> This doesn't compile

Fixed

OneLemonyBoi:
Everything here should be final.

TheCurle:
Okay... But why? Why is this needed?

Articulate yourself.

OneLemonyBoi:
> Okay... But why? Why is this needed?
> 
> Articulate yourself.

Many new exploration and content mods are adding new types of ice, whether it is for decoration, or other uses. A couple of good examples are BetterEnd's Plain, Dense, and Ancient Emerald Ice, as well a Powah's Dry Ice. Along with that, it would help with Mod Compatibility, such as the PR for Immersive Engineering that will add Powah compat. This will help unify all the different kinds of ice that mod makers may use under one branch of tags and make recipes more universal with ice.

OneLemonyBoi:
bump

LexManos:
Don't bump your posts it's annoying.
Adding this tag can be done by any data pack and doesn't need a Forge PR.
However, this doesn't really hurt anything, Except you forgot about frosted ice. Where does that sit?
And before you say it yes, I know that it doesn't have a item form. But you add block tags as well.
Also not a fan of "ices/ice" it should be "ices/normal" or something. As all the ice are ice.. hence why they are in the group.

OneLemonyBoi:
This PR should be ready.

TheCurle:
This really isn't needed. Any mod that wants to use these tags can simply make them in their own mod. As long as it's not set to re/overwrite it'll work fine. 

Combined with the fact that you seem to be making this PR in notepad or something.. the quality isn't quite up to standards.

PR #7655, opened 2021-02-11
Block pathnode entity parameter

flowtender:
I saw there is a method `getAiPathNodeType` in `IForgeBlockState` with a `MobEntity` parameter but it is only ever called with `null` because it is accessed from a static context, even though there is always a relevant `MobEntity` 1-2 invocations away. So I threaded it through. This way blocks can have entity-sensitive pathing to go with their entity-sensitive collision boxes. I also created deprecated mirror methods in case any mods are still using the unaltered methods.

I will use this in my mod to create a block which undead cannot path through.

Also attached: 1 commit with updated licenses (required to build & test this change).

### DIFF
--- a/patches/minecraft/net/minecraft/pathfinding/WalkNodeProcessor.java.patch
+++ b/patches/minecraft/net/minecraft/pathfinding/WalkNodeProcessor.java.patch
@@ -1,10 +1,42 @@
 --- a/net/minecraft/pathfinding/WalkNodeProcessor.java
 +++ b/net/minecraft/pathfinding/WalkNodeProcessor.java
-@@ -463,9 +463,11 @@
+@@ -392,16 +392,21 @@
+    }
  
+    public PathNodeType func_186330_a(IBlockReader p_186330_1_, int p_186330_2_, int p_186330_3_, int p_186330_4_) {
+-      return func_237231_a_(p_186330_1_, new BlockPos.Mutable(p_186330_2_, p_186330_3_, p_186330_4_));
++      return func_237231_a_(p_186330_1_, new BlockPos.Mutable(p_186330_2_, p_186330_3_, p_186330_4_), this.field_186326_b);
+    }
+ 
++   @Deprecated
+    public static PathNodeType func_237231_a_(IBlockReader p_237231_0_, BlockPos.Mutable p_237231_1_) {
++      return func_237231_a_(p_237231_0_, p_237231_1_, null);
++   }
++
++   public static PathNodeType func_237231_a_(IBlockReader p_237231_0_, BlockPos.Mutable p_237231_1_, @Nullable MobEntity mobEntity) {
+       int i = p_237231_1_.func_177958_n();
+       int j = p_237231_1_.func_177956_o();
+       int k = p_237231_1_.func_177952_p();
+-      PathNodeType pathnodetype = func_237238_b_(p_237231_0_, p_237231_1_);
++      PathNodeType pathnodetype = func_237238_b_(p_237231_0_, p_237231_1_, mobEntity);
+       if (pathnodetype == PathNodeType.OPEN && j >= 1) {
+-         PathNodeType pathnodetype1 = func_237238_b_(p_237231_0_, p_237231_1_.func_181079_c(i, j - 1, k));
++         PathNodeType pathnodetype1 = func_237238_b_(p_237231_0_, p_237231_1_.func_181079_c(i, j - 1, k), mobEntity);
+          pathnodetype = pathnodetype1 != PathNodeType.WALKABLE && pathnodetype1 != PathNodeType.OPEN && pathnodetype1 != PathNodeType.WATER && pathnodetype1 != PathNodeType.LAVA ? PathNodeType.WALKABLE : PathNodeType.OPEN;
+          if (pathnodetype1 == PathNodeType.DAMAGE_FIRE) {
+             pathnodetype = PathNodeType.DAMAGE_FIRE;
+@@ -461,11 +466,18 @@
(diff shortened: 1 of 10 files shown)


patches/minecraft/net/minecraft/pathfinding/WalkNodeProcessor.java.patch
```diff
@@ -1,10 +1,42 @@
...
+-      return func_237231_a_(p_186330_1_, new BlockPos.Mutable(p_186330_2_, p_186330_3_, p_186330_4_));
++      return func_237231_a_(p_186330_1_, new BlockPos.Mutable(p_186330_2_, p_186330_3_, p_186330_4_), this.field_186326_b);
+    }
+ 
++   @Deprecated
+    public static PathNodeType func_237231_a_(IBlockReader p_237231_0_, BlockPos.Mutable p_237231_1_) {
++      return func_237231_a_(p_237231_0_, p_237231_1_, null);
++   }
++
++   public static PathNodeType func_237231_a_(IBlockReader p_237231_0_, BlockPos.Mutable p_237231_1_, @Nullable MobEntity mobEntity) {
```
LexManos, line 16: This and the other method you add an argument to need exec/srg lines so that it actually generates properly if people ever name the field/parameters.

flowtender: ok, I hope I did it right. 🙏

src/main/java/net/minecraftforge/event/ItemAttributeModifierEvent.java
```diff
@@ -16,6 +16,7 @@
  * License along with this library; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
  */
+
```
mrp-v2, line 19: Don't mess with whitespace in a file unrelated to your PR.

src/main/java/net/minecraftforge/common/world/ForgeChunkManager.java
```diff
@@ -1,6 +1,6 @@
 /*
  * Minecraft Forge
- * Copyright (c) 2016-2021.
+ * Copyright (c) 2016-2020.
```
mrp-v2, line 3: Why are you changing this?

src/test/java/net/minecraftforge/debug/item/ItemAttributeModifierTest.java
```diff
@@ -16,6 +16,7 @@
  * License along with this library; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
  */
+
```
mrp-v2, line 19: Again with the whitespace change.

src/test/java/net/minecraftforge/debug/world/ForgeChunkManagerTest.java
```diff
@@ -1,6 +1,6 @@
 /*
  * Minecraft Forge
- * Copyright (c) 2016-2021.
+ * Copyright (c) 2016-2020.
```
mrp-v2, line 3: Why are you changing the copyright?

mrp-v2 (changes requested):
Some changes that seem to have been included in this PR by accident.

src/main/java/net/minecraftforge/common/world/ForgeChunkManager.java
sciwhiz12: This is the work of the `updateLicenses` task, which checks that license headers conforms to the one specified in `LICENSE-headers.txt` and if not, replaces them. Without this, the `checkLicenses` task will complain.

I'd say that this change be okayed for now, until someone can update the license header file and all of the license headers via `updateLicenses` in another PR, to maintain consistency and avoid `checkLicenses` from erroring.

src/main/java/net/minecraftforge/event/ItemAttributeModifierEvent.java
sciwhiz12: Again from my [previous comment](https://github.com/MinecraftForge/MinecraftForge/pull/7655/commits/13d20fd3e44ef2b575663bd455ff39898b0ad905#r580661234), this is the work of the `updateLicenses` task.

src/test/java/net/minecraftforge/debug/item/ItemAttributeModifierTest.java
sciwhiz12: Again from my [previous comment](https://github.com/MinecraftForge/MinecraftForge/pull/7655/commits/13d20fd3e44ef2b575663bd455ff39898b0ad905#r580661234), this is the work of the `updateLicenses` task.

src/test/java/net/minecraftforge/debug/world/ForgeChunkManagerTest.java
sciwhiz12: Again from my [previous comment](https://github.com/MinecraftForge/MinecraftForge/pull/7655/commits/13d20fd3e44ef2b575663bd455ff39898b0ad905#r580661234), this is the work of the `updateLicenses` task.

TheCurle:
Unfortunately this PR is far too invasive for what it does. Combined with the fact that it's patching together something that should have been abandoned early in the 1.16 update cycle, and that all of the licenses and logic have to be redone, it's best to let this go.

The functionality that would have been offered by this (preventing vanilla mobs from pathfinding over it) can be replicated by other means, so it is not a great loss.

PR #7657, opened 2021-02-11
Small patch to deprecate .isAir more gracefully

flowtender:
As explained in my post on issue #7409

I am less confident in this one but maybe it will help.


LexManos:
There is no sane way to deprecate it gracefully without causing more issues.
This is Minecraft modding, shit changes there isn't much we can do about it.
Only *possible* think is that we add that it's known to be going away in the javadocs of the existing function. But honestly everyting else is going to break in 1.17 so that doesn't really matter.

PR #7659, opened 2021-02-16
Clean History Entity Attribute Event PR (Redo of #7475)

Kovak:
Hi Everyone,

This PR is a clean version of #7475 (addresses #6928) . The rebase got real ugly for me with the 1.16.5 changes and the amount of reworking that had happened in that PR and I made it pretty much unsalvageable. 

To summarize the remaining requests from the previous thread:

1. The post registration logic should only be called if we do not revert to vanilla registries. This has been added and I made it an explicit separate function called postRegistrationWork so that hopefully it is more clear than just shoving it into checkAndRevertVanilla with no context.

2. It was desired that we move the FORGE_ATTRIBUTES into ForgeHooks instead of having it inline in GlobalEntityTypeAttributes. This involved introducing a getAttribute and hasAttribute to replace the original inline stuff in GlobalEntityTypeAttributes#doesEntityHaveAttribute and GlobalEntityTypeAttributes#getAttributesForEntity. In addition a special putAttributesOld hook was added that logs a warning if GlobalEntityTypeAttributes#put is called. There are also a ForgeHooks#putAttributes and ForgeHooks#combineAttributes that are used by the new attribute events.


Kovak:
@LexManos this is ready for you again, sorry for disrupting the review thread

patches/minecraft/net/minecraft/entity/ai/attributes/GlobalEntityTypeAttributes.java.patch
```diff
@@ -1,24 +1,23 @@
...
-+   private static final Map<EntityType<? extends LivingEntity>, AttributeModifierMap> FORGE_ATTRIBUTES = new java.util.HashMap<EntityType<? extends LivingEntity>, AttributeModifierMap>();
```
LexManos, line 7: `@Deprecated private static final Map<EntityType<? extends LivingEntity>, AttributeModifierMap> FORGE_ATTRIBUTES = net.minecraftforge.common.ForgeHooks.getAttributeView()`

Said function should be `@Deprecated //INTERNAL USE ONLY Remove: 1.17` and return a [unmodifiableMap](https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#unmodifiableMap(java.util.Map))

This will remove the need for your get/has functions. And marks it as something we need to remove in 1.17

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1402,4 +1405,51 @@ public static void fixNullStructureReferences(IChunk chunk, Map<Structure<?>, Lo
         }).result().orElse(dymData);
         return data.set(DIMENSIONS_KEY, withInjected);
     }
+
+    private static final Map<EntityType<? extends LivingEntity>, AttributeModifierMap> FORGE_ATTRIBUTES = new HashMap<>();
+
+    /**  FOR INTERNAL USE ONLY, DO NOT CALL DIRECTLY
+     * ONLY EXISTS FOR LEGACY REASONS SHOULD BE REMOVED IN 1.17
+     */
+    public static AttributeModifierMap putAttributesOld(EntityType<? extends LivingEntity> type, AttributeModifierMap map)
```
LexManos, line 1414: @Deprecated //Internal Use Only, Remove: 1.17

src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -385,10 +391,20 @@ public static void revert(RegistryManager state, ResourceLocation registry, bool
...
                 revertTo(RegistryManager.VANILLA, false);
                 LOGGER.fatal("Detected errors during registry event dispatch, roll back to VANILLA complete");
+            } else {
+                postRegistrationWork();
             }
         }, executor);
     }
 
+    public static void postRegistrationWork()
+    {
```
LexManos, line 401: Move this to ForgeHooks and you can make all the accessors private. Passing in the private map to the Event constructor means you don't have to expose a public put function which is kinda the entire thing we're trying to avoid.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1402,4 +1405,51 @@ public static void fixNullStructureReferences(IChunk chunk, Map<Structure<?>, Lo
...
+    {
+        LOGGER.warn("Called deprecated GlobalEntityTypeAttributes#put for {}, use EntityAttributeCreationEvent instead.", type.getRegistryName());
+        return putAttributes(type, map);
+    }
+
+    /**  FOR INTERNAL USE ONLY, DO NOT CALL DIRECTLY.
+     * Still use GlobalEntityTypeAttributes#getAttributesForEntity
+     */
+    @Nullable
+    public static AttributeModifierMap getAttributes(EntityType<? extends LivingEntity> livingEntity)
```
LexManos, line 1424: This goes away

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1402,4 +1405,51 @@ public static void fixNullStructureReferences(IChunk chunk, Map<Structure<?>, Lo
...
+    {
+        return FORGE_ATTRIBUTES.get(livingEntity);
+    }
+
+    /**  FOR INTERNAL USE ONLY, DO NOT CALL DIRECTLY.
+     * Still use GlobalEntityTypeAttributes#doesEntityHaveAttributes
+     */
+    public static boolean hasAttributes(EntityType<?> entityType)
```
LexManos, line 1432: This goes away

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1402,4 +1405,51 @@ public static void fixNullStructureReferences(IChunk chunk, Map<Structure<?>, Lo
...
+    {
+        return FORGE_ATTRIBUTES.containsKey(entityType);
+    }
+
+    /**  FOR INTERNAL USE ONLY, DO NOT CALL DIRECTLY.
+     * Make use of the interface provided in net.minecraftforge.event.entity.EntityAttributeModificationEvent
+     */
+    public static AttributeModifierMap combineAttributes(EntityType<? extends LivingEntity> type, AttributeModifierMap.MutableAttribute toCombine)
```
LexManos, line 1440: This becomes private/inside the new fireGatherAttributesEvent function

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1402,4 +1405,51 @@ public static void fixNullStructureReferences(IChunk chunk, Map<Structure<?>, Lo
...
+        AttributeModifierMap modifiers = GlobalEntityTypeAttributes.getAttributesForEntity(type);
+        AttributeModifierMap.MutableAttribute newMutable = modifiers != null ? new AttributeModifierMap.MutableAttribute(modifiers) : new AttributeModifierMap.MutableAttribute();
+        newMutable.combine(toCombine);
+        return putAttributes(type, newMutable.create());
+    }
+
+    /**  FOR INTERNAL USE ONLY, DO NOT CALL DIRECTLY.
+     * Make use of the interface provided in net.minecraftforge.event.entity.EntityAttributeCreationEvent
+     */
+    public static AttributeModifierMap putAttributes(EntityType<? extends LivingEntity> type, AttributeModifierMap map)
```
LexManos, line 1451: This goes away as it can be private

src/main/java/net/minecraftforge/event/entity/EntityAttributeCreationEvent.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public EntityAttributeCreationEvent() {
+        super();
+    }
+
+    public void put(EntityType<? extends LivingEntity> entity, AttributeModifierMap map){
+        if (GlobalEntityTypeAttributes.doesEntityHaveAttributes(entity)){
+            throw new IllegalStateException("Duplicate GlobalEntityTypeAttributes entry: " + entity);
+        }
+        ForgeHooks.putAttributes(entity, map);
```
LexManos, line 47: This changes to use an internal map reference so you don't leak put access

patches/minecraft/net/minecraft/entity/ai/attributes/AttributeModifierMap.java.patch
```diff
@@ -0,0 +1,23 @@
...
+       private final Map<Attribute, ModifiableAttributeInstance> field_233811_a_ = Maps.newHashMap();
+       private boolean field_233812_b_;
+ 
++      public MutableAttribute() { }
++
++      public MutableAttribute(AttributeModifierMap attributeMap) {
++         this.field_233811_a_.putAll(attributeMap.field_233802_a_);
++      }
++
++      public void combine(MutableAttribute other) {
```
LexManos, line 13: We need to keep track of the other mutable attribute, so that when we call create() we can set the 'edited' {should be 'baked'} flag. So that those added attributes will be locked as well.

LexManos (changes requested):

LexManos:
I got tired of trying to hold your hand https://github.com/MinecraftForge/MinecraftForge/commit/ba15737fe3d3c6286cfb90c0dca96e2381914245
Notice how the exposed API for internal things is SOO much smaller then it is for your attempts.
And the things to be removed to continue to lower that API are clearly marked.
This is some of the basic ideas that need to be kept in mind when writing PRs to Forge.

PR #7660, opened 2021-02-17
Add Opacity to ForgeHooksClient#renderMainMenu

FireController1847:
It's minor, but I was working on a title screen modification mod and found that this wasn't passing the alpha to this (though all of the other text was fading), so I figured "eh, what the heck, I'll make a PR" and now here we are.

This does not break the color formatting, the 0xFFFFFF does _not_ override the TextFormatting style and is required for it to work correctly.

### DIFF
--- a/patches/minecraft/net/minecraft/client/gui/screen/MainMenuScreen.java.patch
+++ b/patches/minecraft/net/minecraft/client/gui/screen/MainMenuScreen.java.patch
@@ -38,7 +38,7 @@
  
           this.field_230706_i_.func_110434_K().func_110577_a(field_194400_H);
           func_238463_a_(p_230430_1_, j + 88, 67, 0.0F, 0.0F, 98, 14, 128, 16);
-+         net.minecraftforge.client.ForgeHooksClient.renderMainMenu(this, p_230430_1_, this.field_230712_o_, this.field_230708_k_, this.field_230709_l_);
++         net.minecraftforge.client.ForgeHooksClient.renderMainMenu(this, p_230430_1_, this.field_230712_o_, this.field_230708_k_, this.field_230709_l_, l);
           if (this.field_73975_c != null) {
              RenderSystem.pushMatrix();
              RenderSystem.translatef((float)(this.field_230708_k_ / 2 + 90), 70.0F, 0.0F);
--- a/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
+++ b/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
@@ -242,6 +242,12 @@ public class ForgeHooksClient
         //RenderingRegistry.registerBlockHandler(RenderBlockFluid.instance);
     }
 
+    @Deprecated // TODO: Remove in 1.17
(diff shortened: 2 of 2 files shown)


src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -242,16 +242,16 @@ public static double getFOVModifier(GameRenderer renderer, ActiveRenderInfo info
         //RenderingRegistry.registerBlockHandler(RenderBlockFluid.instance);
     }
 
-    public static void renderMainMenu(MainMenuScreen gui, MatrixStack mStack, FontRenderer font, int width, int height)
+    public static void renderMainMenu(MainMenuScreen gui, MatrixStack mStack, FontRenderer font, int width, int height, int alpha)
```
covers1624, line 251: Binary compatibility must be maintained here.

covers1624 (changes requested):

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
FireController1847: Sorry I'm still a little new to contributing, what do you mean by this? ([Solved in the Discord server](https://discord.com/channels/313125603924639766/725850371834118214/811869300104626188))

patches/minecraft/net/minecraft/client/gui/screen/MainMenuScreen.java.patch
```diff
@@ -38,7 +38,7 @@
  
           this.field_230706_i_.func_110434_K().func_110577_a(field_194400_H);
           func_238463_a_(p_230430_1_, j + 88, 67, 0.0F, 0.0F, 98, 14, 128, 16);
-+         net.minecraftforge.client.ForgeHooksClient.renderMainMenu(this, p_230430_1_, this.field_230712_o_, this.field_230708_k_, this.field_230709_l_);
++         net.minecraftforge.client.ForgeHooksClient.renderMainMenu(this, p_230430_1_, this.field_230712_o_, this.field_230708_k_, this.field_230709_l_, l);
```
marchermans, line 41: Why is here 1 supplied but in the default overload later -1?

This does not make sense.

marchermans:
@covers1624 Why was Binary Compat needed here?

The method in question is a hook entry point. Those are not considered directly part of the API itself, but only for keeping the patch size small. Am I mistaken in the fact that binary compat needs to be maintained for those as well or not?

patches/minecraft/net/minecraft/client/gui/screen/MainMenuScreen.java.patch
BlueAgent: It's `l` not `1`. They do look quite similar. XD

marchermans: aaaaah
it is an l okey.
Perfect thank you.

LexManos:
Yes, anything we expose as public is considered for our Binary breaking rules.
Even if modders arn't meant to call them because modders will cuz that's what they do.
Private fields/methods are the only potential exception.
If you're reflecting/breaking into our internals then you're doing it wrong, you should be using our public facing methods/talking to us to get one.

PR #7661, opened 2021-02-17
[1.16] Rework RenderWorldLastEvent (different approach)

Nightenom:
_Based on [tterrag's comments](https://discord.com/channels/313125603924639766/725850371834118214/807735836749201438)_
Supersedes and closes #7225

Fires events per each renderType finish and per each blockLayer, also fires terrainUpdate event, doesn't fire another event when already firing (prevents loop-firing from mods)

[Sample](<https://gist.github.com/Nightenom/4ece03ee1907f86ceb54cb98dab92612>) from new world without TEs and small amount of entities
Event count per frame (roughly, vanilla): min 15 | average 25 | max 50
Time took by one empty event posting (based on #7225): +- 1000 ns, at worst case should far less than 100 μs (0.1 ms) per frame, as mentioned by tterrag timing might be an issue for this approach

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@
 **/.classpath
 *.project
 
+#vscode
+/.vscode/
+
 #In case people make their workspace in the repo
 /eclipse/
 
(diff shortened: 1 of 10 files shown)


sciwhiz12:
Hello! @Nightenom, please leave a comment whether you wish to continue this PR for 1.16. Note that there is already an updated version of the feature intended by this PR at #7924, so it is your decision whether to discontinue updating this PR in favor of that or to continue with your own implementation. (We suggest the former so there is a concentration of effort in one PR rather than fragmented across multiple.) If we do not hear back from you after some time, this PR will be closed for inactivity.

Nightenom:
@sciwhiz12 we can close this now, I would like to get back to this during February if there won't be any activity/resolution by then

PR #7663, opened 2021-02-21
Chunk tick event

flowtender:
After seeing multiple people trying to update capabilities on world tick and running into possible concurrency problems with the threaded loading/unloading of chunks, I thought it might be nice to have an event that only fires when the chunk is actively loaded and ticking, instead of wasting cpu on chunks that are unloaded and awaiting garbage collection or something like that.

Inserting it at the randomTickSpeed check also allows mods to adjust the randomTickSpeed for individual chunks if they want, which could be useful for testing and server administration.

People besides me whom this would be useful for:
1) https://forums.minecraftforge.net/topic/94589-116-how-to-get-loaded-chunks/
2) https://discord.com/channels/313125603924639766/725850371834118214/812119574707306526
3) https://discord.com/channels/313125603924639766/313125603924639766/812840611551510528

Personally I'm saving metadata that relies on randomly ticking blocks I don't own, so that's why I need something like this. Gigaherz's solution is workable but will result in extra passes on chunks that are slated for GC and requires use of threadsafe collections which impact parallelization. Diesieben's solution requires reflection (or I guess an AT).

### DIFF
--- a/patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
+++ b/patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
@@ -96,7 +96,24 @@
           if (biome.func_201848_a(this, blockpos3)) {
              this.func_175656_a(blockpos3, Blocks.field_150432_aD.func_176223_P());
           }
-@@ -598,9 +604,10 @@
+@@ -489,7 +495,7 @@
+       }
+ 
+       iprofiler.func_219895_b("tickBlocks");
+-      if (p_217441_2_ > 0) {
++      if (net.minecraftforge.common.ForgeHooks.onChunkTickPre(p_217441_1_, p_217441_2_) > 0) {
+          for(ChunkSection chunksection : p_217441_1_.func_76587_i()) {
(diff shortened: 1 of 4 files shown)


ChampionAsh5357:
So, this needs a more defined use case since everything as you have defined can be obtained with management between two or three existing events. As for the performance of the event, it will make thousands of event calls per tick per player, so there needs to be some benchmarks done to verify that the event calls are negligible or very minor as otherwise the event is not worthwhile. The benchmark should be done on a server with some number of people playing to get a better sense of the impact.

If you can provide these two things, then there will probably be more discussion on whether the event is needed and feasible. Currently, I'm under the opinion that it is unnecessary. However, depending on the response given to the above requirements, that may change.

sciwhiz12:
Hello! @MachineMuse, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

sciwhiz12:

PR #7665, opened 2021-02-24
Added a way for modders to modify the players turn speed

seymourimadeit:
This PR adds a new event called `UpdatePlayerLookEvent` which allows modders to modify the player's turn speed via `UpdatePlayerLookEvent#setMouseTurnSpeed`. The event is fired in the `updatePlayerLook` method in the `MouseHelper` class.

This PR was made so modders could modify the players turn speed, for example if a modder wanted had an item that slowed the player's turn speed, they'd use this event.

### DIFF
--- a/patches/minecraft/net/minecraft/client/MouseHelper.java.patch
+++ b/patches/minecraft/net/minecraft/client/MouseHelper.java.patch
@@ -65,7 +65,15 @@
                 }, "mouseDragged event handler", iguieventlistener.getClass().getCanonicalName());
              }
           }
-@@ -267,6 +277,10 @@
+@@ -226,6 +236,7 @@
+       this.field_198050_o = d0;
+       if (this.func_198035_h() && this.field_198036_a.func_195544_aj()) {
+          double d4 = this.field_198036_a.field_71474_y.field_74341_c * (double)0.6F + (double)0.2F;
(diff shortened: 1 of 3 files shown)


TheCurle:
Can we get some why?
What purpose does this serve? What problems do it solve?

seymourimadeit:
> Can we get some why?
> What purpose does this serve? What problems do it solve?

The purpose of this was so modders could change the player's turn speed if their mod needed it, for example if a modder wanted to add an item that slowed the players turn speed down they'd just subscribe to this event and check if the player had the item.

KnightMiner:
Why are you changing a player's turn speed? That is basically changing their mouse speed, huge accessibility problem in general. Use a more concrete example, say you want to implement a telescope and don't want dramatic view shifts when the mouse moves. (thats just what I can think of, what you actually need this for would be a lot more helpful then me spitballing ideas)

Anyways, I feel this is run too often to be an event. My intuition is a field (might cause issues if multiple mods change it) or an attribute (might not work great as attributes are serverside)

TheCurle:
Indeed, this seems to be called a couple times per frame.
The issue here is that multiple adjustments by multiple mods are up to random chance.
If you can find a way to cleanly resolve conflicts, that'd be brilliant.

We'd also like you to join the team, because that's an incredibly complicated problem that nobody has been able to solve yet.

In short, maybe with some small adjustments and solving the P=NP problem.

PR #3922, opened 2017-05-26
Add a hook to allow world providers to change the lightmap colours.

bs2609:
This PR patches in a hook to `EntityRenderer.updateLightmap()` to allow custom `WorldProvider` implementations to change the lightmap texture colouring.

Currently it's possible to alter the base brightness table used, but not change the colouring, or treat sky/block lighting differently. Vanilla already has a hardcoded check for the End dimension here to achieve this.

The method call is inserted just after the End dimension check, and before night vision and gamma adjustments.

As an example, here is the same world; first as vanilla, then with this hook used to apply something similar to the End's lighting.

Vanilla:

Custom:

PR #7667, opened 2021-02-27
[1.16.x] Readd the ability to modify Dimension Lightmap Colours

50ap5ud5:
### Superceded by #7678
=================

### About
This PR adds a new event that restores the functionality originally provided by #3922. It adds the ability for modders to adjust the light map colours of a specific dimension.
### What are LightMap Colours
The Lightmap colours are the colours encoded into the brightness of surfaces. By modifying these colours the "tint" of a surface can be changed.
### Why this is needed
From 1.16 onwards, the ability to adjust the lightmap colours of a dimension was removed following the sweeping changes to Dimensions and the Forge Dimension API.
Currently in vanilla Minecraft, these lightmap colours are hardcoded.
### Use Cases
There is evidence of a number of valid usecases for this PR.
For example, RFTools uses it to darken the lighting of their dimensions if the dimension’s “power” is low.
(See the lighting of the dimension darken from 16:41 onwards in the below video)
- https://youtu.be/h53ADXTuNwM?t=1001 

Another use case is in one of my mods, where I interpolate the world’s gameTime with the lighting to make the dimension “flash red” as part of an “alarm” system as shown in the below video.
- https://www.youtube.com/watch?v=gsZH5SC3nbI

I also have use cases to modify the "mood lighting" of my custom dimensions, which is needed for users to have a visual indication to determine the state of a core feature in the mod. The below images show a "warm" and "cold" coloured lighting I use. (This was done in versions below 1.16)

- image
- image

### Showcase of this Pull Request's Changes
In the provided example test mod, I have included a custom dimension and set its lightmap colours to “flash red” in a similar fashion to a use case in my mod. 
https://gyazo.com/8c3caaeb8cfc97c12fec478ed9d7794b 

### Implementation and Discussion
This PR is based off the 1.15 version of #3922 through patching the LightTexture class in a similar location as before. 

There are a number of reasons to add an event instead of modifying an existing object like the DimensionRenderInfo:

- DimensionRenderInfo registration involves encouraging poor modding practices. To add a new DimensionRenderInfo you need to use non-api methods to access its private immutable static map.

- DimensionRenderInfo uses the OnlyIn annotation which is known to cause a number of other side effects. We do not want to encourage modders to adopt this practice either.

- DimensionRenderInfo only applies to a DimensionType. One DimensionType can be used for multiple Dimensions. This means the functionality of using DimensionRenderInfo is limited, as modders may not want all Dimensions using that DimensionType to have the same lightmap colours. 

- Keep in mind that the original PR allowed for colour modification per dimension, so using something like DimensionRenderInfo makes it a regression of functionality compared to before.

There are also other reasons I believe this event is a good solution:
- Using an event could reduce future maintenance requirements when Forge is updated to future versions of Minecraft. There are less patches needed to fire this event compared to adding many hooks to existing Minecraft objects.

- There aren't many other suitable places to add this hook. The ClientWorld cannot be easily used by modders, and adding a new object specifically for this hook would require alot more work to allow it to sync with the vanilla Dimension system, which is data driven.

Thank you for considering this pull request. If you have any questions or suggestions feel free to discuss it!


patches/minecraft/net/minecraft/client/renderer/LightTexture.java.patch
```diff
@@ -1,5 +1,14 @@
 --- a/net/minecraft/client/renderer/LightTexture.java
 +++ b/net/minecraft/client/renderer/LightTexture.java
+@@ -122,7 +122,7 @@
+                         vector3f1.func_229190_a_(vector3f3, f9);
+                      }
+                   }
+-
++                  net.minecraftforge.client.ForgeHooksClient.onDimensionLightMapUpdate(clientworld, p_205106_1_, f, f4, f5, vector3f1);
```
sciwhiz12: Minor nitpick: Could you add a blank line above your change, so the `-` diff is removed? (just a minor nitpick tho)

src/main/java/net/minecraftforge/client/event/DimensionLightMapModificationEvent.java
```diff
@@ -0,0 +1,103 @@
...
+ * Allows modders to adjust the lightmap colours for a specific Dimension
+ * <br>
+ * While DimensionRenderInfo could look like the closest solution for this task, its entries are stored in private, immutable map that is specific to a DimensionType
+ * <br>
+ * There can be many Dimension(s) per DimensionType, which means DimensionRenderInfo is limited in both functionality and encourages poor practices.
+ * <br>
+ * Hence, this event is the preferred solution over DimensionRenderInfo.
+ * <br>
+ * This event fired in {@link LightMapTexture#updateLightMap}
+ */
```
sciwhiz12: A suggestion for the javadoc, just to tidy it up a bit more (and use less `<br>` tags):
```java
/**
 * <p>Allows modders to adjust the lightmap colours for a specific {@code Dimension}.</p>
 * 
 * <p>While {@code DimensionRenderInfo} could look like the closest solution for this task, 
 * its entries are stored in private, immutable map that is specific to a {@code DimensionType}. 
 * There can be many {@code Dimension}(s) per {@code DimensionType}, which means 
 * {@code DimensionRenderInfo} is limited in both functionality and encourages poor practices.</p>
 * 
 * <p>Hence, this event is the preferred solution over DimensionRenderInfo.</p>
 * 
 * <p>This event is not {@linkplain net.minecraftforge.eventbus.api.Cancelable cancellable}, 
 * and does not {@linkplain HasResult have a result}.</p>
 * 
 * <p>This event fired in {@link LightMapTexture#updateLightMap}, on the 
 * {@link net.minecraftforge.common.MinecraftForge#EVENT_BUS} only on the 
 * {@linkplain net.minecraftforge.fml.LogicalSide#CLIENT client-side}.
 */
```

src/test/java/net/minecraftforge/debug/world/DimensionLightMapTest.java
```diff
@@ -0,0 +1,70 @@
...
+    public static final String MODID = "dim_lightmap_test";
+
+    public static ResourceLocation DIMENSION_LOC = new ResourceLocation(MODID,"custom_dimension_1");
+    public static RegistryKey<World> TEST_WORLD;
+    private static final boolean ENABLED = true;
+    public DimensionLightMapTest()
+    {
+        FMLJavaModLoadingContext.get().getModEventBus().register(this);
+        FMLJavaModLoadingContext.get().getModEventBus().addListener(this::commonSetup);
+        if(ENABLED)
```
sciwhiz12: Minor style nitpick:
```suggestion
        if (ENABLED)
```

src/test/java/net/minecraftforge/debug/world/DimensionLightMapTest.java
```diff
@@ -0,0 +1,70 @@
...
+{
+
+    public static final String MODID = "dim_lightmap_test";
+
+    public static ResourceLocation DIMENSION_LOC = new ResourceLocation(MODID,"custom_dimension_1");
+    public static RegistryKey<World> TEST_WORLD;
+    private static final boolean ENABLED = true;
+    public DimensionLightMapTest()
+    {
+        FMLJavaModLoadingContext.get().getModEventBus().register(this);
```
sciwhiz12: Remove this line, since it registers nothing (the `commonSetup` and `onDimensionLightMapModification` listeners are registered manually).

sciwhiz12 (commented):
Just a few minor nitpicks and an improvement of sorts to the javadoc. I have too little knowledge of rendering to comment on this PR completely, but it seems to be in order.

src/main/java/net/minecraftforge/client/event/DimensionLightMapModificationEvent.java
50ap5ud5: Thanks! I like your suggestion, it covers a few things I did not consider. I will include it in my next commit.

patches/minecraft/net/minecraft/client/renderer/LightTexture.java.patch
50ap5ud5: Sure, thanks!

src/test/java/net/minecraftforge/debug/world/DimensionLightMapTest.java
50ap5ud5: Thanks for pointing that out, that's been fixed.

50ap5ud5:
Thanks for the review Sci, I've taken your suggestions into account and fixed the issues you pointed out.

50ap5ud5:
Closing this PR to allow for a more clean git history.

This PR is superceded by #7678.

PR #7668, opened 2021-02-27
Add getCodec method in ForgeRegistry

Frinn38:
This is mostly the same PR as #7190 a bit rewritten and simplified.
The main change is that this PR add a getCodec() method  in IForgeRegistry instead of having the registry implements the whole Codec interface (asked by LexManos on discord in modder-support-116, 03/08/2020)

This PR add a getCodec() method in IForgeRegistry interface and implement it's methods encode & decode in ForgeRegistry class.
It also add a test mod to show a possible usage of this registry codec and prove that this implementation of encode and decode works.

The Codec returned by the method will be a useful tool to serialize/deserialize registry entries in Json or NBT.
It is compatible with all Forge Registries including custom ones.


src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -101,6 +106,42 @@
     private final ResourceLocation name;
     private final RegistryKey<Registry<V>> key;
     private final RegistryBuilder<V> builder;
+    
+    private final Codec<V> CODEC = new Codec<V>() {
```
kashike: This is not a `static` field.
```suggestion
    private final Codec<V> codec = new Codec<V>() {
```

kashike (changes requested):

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -101,6 +106,42 @@
...
+        }
+
+        @Override
+        public <T> DataResult<T> encode(V input, DynamicOps<T> ops, T prefix) {
+            ResourceLocation key = getKey(input);
+            if(key == null)
+            {
+                return DataResult.error("Unknown registry element " + input);
+            }
+            T toMerge = ops.compressMaps() ? ops.createInt(getID(input)) : ops.createString(key.toString());
```
sciwhiz12: Question: why is the `ResourceLocation` converted to a string via `.toString()` instead of its `Codec` (from `ResourceLocation.CODEC`)? (aside from this being a formatted copy from vanilla's `Registry#encode`)
(Yes, I understand this works as vanilla uses it, but I'm curious as to why not the codec too; disclaimer: I have only a passing familiarity with codecs)

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
Frinn38: Yes, changed it to lowercase.

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
Frinn38: As you sait it's a copy of the vanilla registry implementation. But just look at the ResourceLocation.CODEC and you will see that it's encode method is basically ResourceLocation::toString.
So there is no need of the ResoureceLocation codec in this case.
Also the codec would return a DataResult<T> with T being the String so I would need to add severals checks to handle the case the codec failed to encode the ResourceLocation, this should not happen because the encode is just the toString() method but still the code for the checks would be needed to retrieve the encoded String from the DataResult.

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -839,6 +850,42 @@ public Snapshot makeSnapshot()
         return ret;
     }
 
+    public class RegistryCodec implements Codec<V> {
```
pupnewfster: Is there a reason this implementation detail is public instead of private?

Frinn38: Not at all, it can (and should probably) be private.

Frinn38: done

TheCurle:
Some preliminary testing shows it works. Codecs are a known nightmare, so i'll test it more thoroughly before recommending anything.

Your test class is missing a license header. That seems to be the only thing holding this back for the moment, though.

Frinn38:
Completely forgot the License header, now fixed.
I agree that Codecs can be a pain (at least they are relatively easy to use but still a nightmare to understand) but I believe that in the future Mojang will use them more and more and modders will have to follow, and when that time come, having a codec for all ForgeRegistryEntry will be necessary (it's just my guess, maybe Mojang will just void them and turn back to good old serializers but I doubt)

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -185,6 +191,11 @@ public String getTagFolder()
         return tagFolder;
     }
 
+    @Override
+    public Codec<V> getCodec() {
```
ChampionAsh5357: Bracket on new line.

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -839,6 +850,42 @@ public Snapshot makeSnapshot()
         return ret;
     }
 
+    private class RegistryCodec implements Codec<V> {
```
ChampionAsh5357: Bracket on new line.

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -839,6 +850,42 @@ public Snapshot makeSnapshot()
...
+            else
+            {
+                return ResourceLocation.CODEC.decode(ops, input).flatMap(keyValuePair -> !ForgeRegistry.this.containsKey(keyValuePair.getFirst()) ?
+                        DataResult.error("Unknown registry key: " + keyValuePair.getFirst()) :
+                        DataResult.success(keyValuePair.mapFirst(ForgeRegistry.this::getValue)));
+            }
+        }
+
+        @Override
+        public <T> DataResult<T> encode(V input, DynamicOps<T> ops, T prefix) {
```
ChampionAsh5357: Bracket on new line.

src/test/java/net/minecraftforge/debug/misc/CodecRegistryTest.java
```diff
@@ -0,0 +1,86 @@
...
+import net.minecraft.nbt.NBTDynamicOps;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.event.lifecycle.FMLCommonSetupEvent;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.registries.ForgeRegistries;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+@Mod("codec_registry_test")
+public class CodecRegistryTest {
```
ChampionAsh5357: Bracket on new line.

src/test/java/net/minecraftforge/debug/misc/CodecRegistryTest.java
```diff
@@ -0,0 +1,86 @@
...
+     *         "block": "block_registry_name",
+     *         "item": "item_registry_name"
+     *     }
+     */
+    private static final Codec<Pair<Block, Item>> CODEC = RecordCodecBuilder.create(codecInstance -> codecInstance.group(
+        ForgeRegistries.BLOCKS.getCodec().fieldOf("block").forGetter(Pair::getFirst),
+        ForgeRegistries.ITEMS.getCodec().fieldOf("item").forGetter(Pair::getSecond)
+    ).apply(codecInstance, Pair::of));
+
+    public CodecRegistryTest() {
```
ChampionAsh5357: Bracket on new line.

src/test/java/net/minecraftforge/debug/misc/CodecRegistryTest.java
```diff
@@ -0,0 +1,86 @@
...
+        FMLJavaModLoadingContext.get().getModEventBus().addListener(this::commonSetup);
+    }
+
+    public void commonSetup(final FMLCommonSetupEvent event) {
```
ChampionAsh5357: Bracket on new line.

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -839,6 +850,42 @@ public Snapshot makeSnapshot()
         return ret;
     }
 
+    private class RegistryCodec implements Codec<V> {
+        @Override
+        public <T> DataResult<Pair<V, T>> decode(DynamicOps<T> ops, T input) {
```
ChampionAsh5357: Bracket on new line.

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -839,6 +850,42 @@ public Snapshot makeSnapshot()
...
+                return ops.getNumberValue(input).flatMap(n ->
+                {
+                    int id = n.intValue();
+                    if(ids.get(id) == null)
+                    {
+                        return DataResult.error("Unknown registry id " + n);
+                    }
+                    V val = ForgeRegistry.this.getValue(id);
+                    return DataResult.success(val);
+                }).map(v -> Pair.of(v, input));
```
ChampionAsh5357: The `Pair` returned by `#decode` represents the data that was successfully decoded to the object and the data that failed to decode. This says that the output was both fully encoded and decoded like schrodinger's cat.

ChampionAsh5357 (changes requested):
Leaving some general fixes that's needed. From my testing, it seems to be fine besides from the misunderstanding involving `#decode`. Will wait on @TheCurle to confirm whenever she has the time. You should also update this to 1.17 since that is now latest.

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
Frinn38: You're right, vanilla registry codec return ops.empty() here...

Frinn38:
I will add the same PR for 1.17 now that it's the latest version.

Frinn38:
Closed as it's finally merged in 1.18

ChampionAsh5357:
Why wouldn't this be an LTS PR for merging into 1.16?

PR #7671, opened 2021-03-01
[1.16] Attempt to fix issues with configurations deemed invalid for no valid reason

TheCurle:
Currently, there are a few oversights with regards to how config files are validated.  

Firstly, comments are checked bit-for-bit against the specification, which goes against platform standardization - that meaning, it could cause issues if the spec contains `\r\n` but the file is written with `\n`.  

Secondly, lists by default are deemed invalid if they are empty. This is a little silly, so i added an optional `defineListAllowEmpty` builder that skips that empty check. This is useful for eg. blacklists.  

Thirdly, strings are not checked for nullity when passed through the comment function. Since null strings is an invalid state, i added a wrapper to ensure this does not get saved to the file. Due to some limitations in NightConfig - mainly, writing an empty string ("") or a space (" ") will get parsed as null when the file is read. This means there has to be *some* text there as a default. I chose "No comment" as a safe default, but this can be changed on request.

Lastly, the log files do not say exactly why some things are deemed to need correction. To this end, i also added a bunch of logging to the preset configs, and to the validator (when it fails).

This solves all of the log spam (and as a bonus, the backup creation) with Repurposed Structures, Aquaculture and AE2, which are the only mods i'm aware of where this issue is easily reproducible.

This is not finalised, and there are more places where logging is needed.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
+++ b/src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
@@ -86,7 +86,7 @@ public class ForgeConfigSpec extends UnmodifiableConfigWrapper<UnmodifiableConfi
             String configName = config instanceof FileConfig ? ((FileConfig) config).getNioPath().toString() : config.toString();
             LogManager.getLogger().warn(CORE, "Configuration file {} is not correct. Correcting", configName);
             correct(config, (action, path, incorrectValue, correctedValue) ->
-                    LogManager.getLogger().warn(CORE, "Incorrect key {} was corrected from {} to {}", DOT_JOINER.join( path ), incorrectValue, correctedValue));
+                    LogManager.getLogger().warn(CORE, "Incorrect key {} was corrected from {} to its default, {}", DOT_JOINER.join( path ), incorrectValue, correctedValue));
             if (config instanceof FileConfig) {
                 ((FileConfig) config).save();
             }
@@ -222,6 +222,7 @@ public class ForgeConfigSpec extends UnmodifiableConfigWrapper<UnmodifiableConfi
                     if (dryRun)
                         return 1;
 
+                    LogManager.getLogger().warn(CORE, "The comment on key {} does not match the spec.", key);
                     //TODO: Comment correction listener?
                     config.setComment(key, valueSpec.getComment());
                 }
@@ -354,11 +355,13 @@ public class ForgeConfigSpec extends UnmodifiableConfigWrapper<UnmodifiableConfi
                 @Override
                 public Object correct(Object value) {
                     if (value == null || !(value instanceof List) || ((List<?>)value).isEmpty()) {
+                        LogManager.getLogger().debug(CORE, "List on key {} is deemed to need correction. It is null, not a list, or an empty list. Modders, consider defineListAllowEmpty?", path.get(path.size() - 1));
                         return getDefault();
                     }
                     List<?> list = Lists.newArrayList((List<?>) value);
                     list.removeIf(elementValidator.negate());
                     if (list.isEmpty()) {
+                        LogManager.getLogger().debug(CORE, "List on key {} is deemed to need correction. It failed validation.", path.get(path.size() - 1));
                         return getDefault();
                     }
                     return list;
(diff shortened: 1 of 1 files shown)


TelepathicGrunt (approved):

src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
```diff
@@ -249,6 +249,24 @@ else if (dryRun)
         return count;
     }
 
+    private boolean stringsMatchIgnoringNewlines(Object obj1, Object obj2)
```
TheCurle: I'll just make a note here for future reference:
This function can be passed null, for eg. when the config is first being generated.
Calling replaceAll on null causes an immediate hard crash, so i had to add instanceof and length checks to ensure it's even valid for replacement. Without it, this just causes bigger issues.

Someone asked, so i thought i'd make a note here.

TheCurle:
One last issue to figure out and i'm happy to submit this.
Quark used to print this to the log:
```Incorrect key world.new_stone_types.marble.oregen.Max Height was corrected from 256 to its default, 256
Incorrect key world.new_stone_types.limestone.oregen.Max Height was corrected from 256 to its default, 256
Incorrect key world.new_stone_types.jasper.oregen.Max Height was corrected from 256 to its default, 256
Incorrect key world.new_stone_types.slate.oregen.Max Height was corrected from 256 to its default, 256
Incorrect key world.new_stone_types.voidstone.oregen.Max Height was corrected from 256 to its default, 256
Incorrect key world.new_stone_types.myalite.oregen.Max Height was corrected from 256 to its default, 256
```
This happened because they defined the valid range as 0-255, and set the default to 256. Ergo, every time the config was reloaded it caused a backup and this error.

I'm not sure how to say that the default may be invalid, so i added a check and `This seems to be an error.` to the end of the line if the correction and the invalid item match. 
Adding a case for this is extremely hard, as they hardcode the predicate checker to `o -> true` despite the range being present. Therefore, my log listeners in the ForgeConfigSpec aren't triggered for this case.

Maybe this is a lost cause.

TheCurle:
This is now ready for review, but i have one last feature to implement.

src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
```diff
@@ -527,11 +528,20 @@ public LongValue defineInRange(List<String> path, Supplier<Long> defaultSupplier
...
+            if(comment == null || comment.length() < 1)
+            {
+                comment = "No comment";
+            }
             context.setComment(comment);
             return this;
         }
         public Builder comment(String... comment)
         {
+            if(comment == null || comment.length < 1 || comment[comment.length - 1].length() < 1)
```
KnightMiner: I don't like the last condition here. You are checking if the last element in the array is an empty string. That would match a comment such as `comment("Line 1", "", "Line 2", "")` that has extra forced line breaks. I think that will cause many false positives.

A better check would be if the length is equal to 1 and the only comment in the array is empty. Or if you think the effort is worth it, that all comments in the array are empty.

Also, `String::isEmpty()` for the record

TheCurle: Good point. I did briefly debate myself over this, but i didn't reach a conclusion. One moment.

LexManos:
> This happened because they defined the valid range as 0-255, and set the default to 256. Ergo, every time the config was reloaded it caused a backup and this error.

If someone goes out of their way to bypass our validator, and put default values that are outside of the range they explicitly define is acceptible.. then that's on them. I'd be fine with hard crashing and calling them out.
But that would hurt end users so we can't do that. Don't change anything on our end. Bad modders is bad.

The other changes here are somewhat acceptable as its just better tracking what things are needing correcting. Specifically parts where loggers wernt showing it.
And the empty string conversations to null is a somewhat obscure behavior of upstream so warning about that is fine. But still modders should be making sure that the values they put into their configs are representable in the configs.

src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
KnightMiner: Since it may be passed null, why not mark the parameters `@Nullable`?

TheCurle: Good point! One moment.

PR #7676, opened 2021-03-08
Add barrels and barrels/wooden block and item tags

Trikzon:
This pr adds two tags: `forge:barrels` and `forge:barrels/wooden`.

Reasoning:
Mods, such as my own Crumbs, may add their own barrel blocks. If a mod wants to use the `minecraft:barrel` in a recipe, it will not be compatible with the modded barrels. A modpack author could add their own tag and change the recipes with datapacks, but I think it should be standardized in forge as a forge tag so that it doesn't require extra manual work by a modpack author.


TheCurle:
We tend to dislike adding tags for common modded things, because that's a very slippery slope that ends up in us standardizing every metal, real or fictional, to ever exist...

As it stands, this one has been requested a lot more often than you'd think, so i'm willing to move this forward..

PR #7677, opened 2021-03-08
[1.16] Add missing Multipart Blockstate Builder feature: Nested condition groups

mrp-v2:
Multipart blockstate models support having nested condition groups in the 'when' section of each part. For an example of this, see the redstone dust blockstate file. Forge's multipart builder does not currently have a way to make nested condition groups, which is what this PR adds. I tested the changes by publishing to my local maven and trying them out with one of my mods that would benefit from this feature.

Feel free to critique my formatting and naming.

### DIFF
--- a/src/main/java/net/minecraftforge/client/model/generators/MultiPartBlockStateBuilder.java
+++ b/src/main/java/net/minecraftforge/client/model/generators/MultiPartBlockStateBuilder.java
@@ -191,7 +191,7 @@ public final class MultiPartBlockStateBuilder implements IGeneratedBlockstate {
              */
             public ConditionGroup nestedGroup()
             {
-                Preconditions.checkState(conditions.size() == 0, "Can't have nested condition groups if there are already normal conditions");
+                Preconditions.checkState(conditions.isEmpty(), "Can't have nested condition groups if there are already normal conditions");
                 ConditionGroup group = new ConditionGroup();
                 group.parent = this;
                 this.nestedConditionGroups.add(group);
--- a/src/main/java/net/minecraftforge/client/model/generators/MultiPartBlockStateBuilder.java
+++ b/src/main/java/net/minecraftforge/client/model/generators/MultiPartBlockStateBuilder.java
@@ -72,7 +72,7 @@ public final class MultiPartBlockStateBuilder implements IGeneratedBlockstate {
         return main;
     }
 
-    public class PartBuilder implements IConditionGroup {
+    public class PartBuilder {
         public BlockStateProvider.ConfiguredModelList models;
         public boolean useOr;
         public final Multimap<Property<?>, Comparable<?>> conditions = MultimapBuilder.linkedHashKeys().arrayListValues().build();
@@ -82,6 +82,9 @@ public final class MultiPartBlockStateBuilder implements IGeneratedBlockstate {
             this.models = models;
         }
 
+        /**
+         * Makes this part get applied if any of the conditions/condition groups are true, instead of all of them needing to be true.
+         */
         public PartBuilder useOr() {
             this.useOr = true;
             return this;
@@ -110,7 +113,7 @@ public final class MultiPartBlockStateBuilder implements IGeneratedBlockstate {
             Preconditions.checkArgument(values.length > 0, "Value list must not be empty");
             Preconditions.checkArgument(!conditions.containsKey(prop), "Cannot set condition for property \"%s\" more than once", prop.getName());
             Preconditions.checkArgument(canApplyTo(owner), "IProperty %s is not valid for the block %s", prop, owner);
-            Preconditions.checkState(nestedConditionGroups.size() == 0, "Can't have normal conditions if there are already nested condition groups");
+            Preconditions.checkState(nestedConditionGroups.isEmpty(), "Can't have normal conditions if there are already nested condition groups");
             this.conditions.putAll(prop, Arrays.asList(values));
             return this;
         }
@@ -121,7 +124,7 @@ public final class MultiPartBlockStateBuilder implements IGeneratedBlockstate {
          */
         public final ConditionGroup nestedGroup()
         {
-            Preconditions.checkState(conditions.size() == 0, "Can't have nested condition groups if there are already normal conditions");
+            Preconditions.checkState(conditions.isEmpty(), "Can't have nested condition groups if there are already normal conditions");
             ConditionGroup group = new ConditionGroup();
             this.nestedConditionGroups.add(group);
             return group;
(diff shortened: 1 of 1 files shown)


mrp-v2:
I've done some digging and it seems that multipart models can have infinite nesting of OR and AND groups. I am currently working on adjusting the multipart builder to be able to do this, so I am making this a draft PR until I finish that.

mrp-v2:
Now supports infinite nesting of condition groups. Also it turns out the `useOr` method didn't actually work so that is fixed now as well.

mrp-v2:
Example usage of nested condition groups:
-Datagen: https://github.com/mrp-v2/MoreWires/blob/affe1fa526a892dd2d99b95a6575ffc9ed293797/src/main/java/mrp_v2/morewires/datagen/BlockStateGenerator.java#L79
-Result: https://github.com/mrp-v2/MoreWires/blob/affe1fa526a892dd2d99b95a6575ffc9ed293797/src/generated/resources/assets/morewires/blockstates/red_infiniwire.json.

src/main/java/net/minecraftforge/client/model/generators/MultiPartBlockStateBuilder.java
```diff
@@ -107,12 +110,27 @@ public PartBuilder useOr() {
             Preconditions.checkArgument(values.length > 0, "Value list must not be empty");
             Preconditions.checkArgument(!conditions.containsKey(prop), "Cannot set condition for property \"%s\" more than once", prop.getName());
             Preconditions.checkArgument(canApplyTo(owner), "IProperty %s is not valid for the block %s", prop, owner);
+            Preconditions.checkState(nestedConditionGroups.size() == 0, "Can't have normal conditions if there are already nested condition groups");
```
pupnewfster: `nestedConditionGroups.size() == 0` -> `nestedConditionGroups.isEmpty()`

src/main/java/net/minecraftforge/client/model/generators/MultiPartBlockStateBuilder.java
```diff
@@ -107,12 +110,27 @@ public PartBuilder useOr() {
...
             return this;
         }
 
+        /**
+         * Allows having nested groups of conditions if the current condition group is empty.
+         * @throws IllegalStateException if there are any normal conditions in the current condition group
+         */
+        public final ConditionGroup nestedGroup()
+        {
+            Preconditions.checkState(conditions.size() == 0, "Can't have nested condition groups if there are already normal conditions");
```
pupnewfster: `conditions.size() == 0` -> `conditions.isEmpty()`

src/main/java/net/minecraftforge/client/model/generators/MultiPartBlockStateBuilder.java
```diff
@@ -127,18 +145,167 @@ JsonObject toJson() {
...
         public boolean canApplyTo(Block b) {
             return b.getStateDefinition().getProperties().containsAll(conditions.keySet());
         }
+
+        public class ConditionGroup implements IConditionGroup
+        {
+            public boolean useOr;
+            private ConditionGroup parent = null;
+            public final Multimap<Property<?>, Comparable<?>> conditions = MultimapBuilder.linkedHashKeys().arrayListValues().build();
+            public final List<ConditionGroup> nestedConditionGroups = new ArrayList<>();
```
pupnewfster: This would be a bit more readable if it went conditions, nestedConditionGroups, parent, useOr. Given then the final fields are at the top and also as they are the longest lines it is easier to read.

src/main/java/net/minecraftforge/client/model/generators/MultiPartBlockStateBuilder.java
```diff
@@ -127,18 +145,167 @@ JsonObject toJson() {
...
+             */
+            @SafeVarargs
+            public final <T extends Comparable<T>> ConditionGroup condition(Property<T> prop, T... values)
+            {
+                Preconditions.checkNotNull(prop, "Property must not be null");
+                Preconditions.checkNotNull(values, "Value list must not be null");
+                Preconditions.checkArgument(values.length > 0, "Value list must not be empty");
+                Preconditions.checkArgument(!conditions.containsKey(prop), "Cannot set condition for property \"%s\" more than once", prop.getName());
+                Preconditions.checkArgument(canApplyTo(owner), "IProperty %s is not valid for the block %s", prop, owner);
+                Preconditions.checkState(nestedConditionGroups.size() == 0, "Can't have normal conditions if there are already nested condition groups");
```
pupnewfster: `nestedConditionGroups.size() == 0` -> `nestedConditionGroups.isEmpty()`

src/main/java/net/minecraftforge/client/model/generators/MultiPartBlockStateBuilder.java
```diff
@@ -127,18 +145,167 @@ JsonObject toJson() {
...
+                return this;
+            }
+
+            /**
+             * Allows having nested groups of conditions if the current condition group is empty.
+             * @throws IllegalStateException if there are any normal conditions in the current condition group
+             */
+            public final ConditionGroup nestedGroup()
+            {
+                Preconditions.checkState(conditions.size() == 0, "Can't have nested condition groups if there are already normal conditions");
```
pupnewfster: `conditions.size() == 0` -> `conditions.isEmpty()`

src/main/java/net/minecraftforge/client/model/generators/MultiPartBlockStateBuilder.java
```diff
@@ -127,18 +145,167 @@ JsonObject toJson() {
...
+        }
+    }
+
+    public interface IConditionGroup
+    {
+        IConditionGroup nestedGroup();
+        /**
+         * If this is a {@link PartBuilder}, returns itself.
+         */
+        IConditionGroup endGroup();
```
pupnewfster: I am unsure about having endGroup be exposed for the PartBuilder. Is there some benefit to having the PartBuilder implement IConditionGroup as I don't really see any in terms of it needing to then be passed as a parameter to something that expects an IConditionGroup.

src/main/java/net/minecraftforge/client/model/generators/MultiPartBlockStateBuilder.java
```diff
@@ -127,18 +145,167 @@ JsonObject toJson() {
...
+                        innerGroupJson.add(group.toJson());
+                    }
+                    return groupJson;
+                }
+                return new JsonObject();
+            }
+        }
+    }
+
+    public interface IConditionGroup
```
pupnewfster: The methods here should be javadoc'd (can be moved from documenting them in ConditionGroup as they will still show the docs due to this being the root for it. It may also be worth throwing on a TODO here for 1.17 about converting this to being abstract or something given it seems like there is a lot of duplicate code between ConditionGroup and PartBuilder which could be cleaned up using an abstract class, except I think that would potentially break binary compat to do now due to useOr and conditions (and now nestedConditionGroups) being public.

src/main/java/net/minecraftforge/client/model/generators/MultiPartBlockStateBuilder.java
```diff
@@ -127,18 +145,167 @@ JsonObject toJson() {
...
+                return group;
+            }
+
+            public IConditionGroup endGroup() 
+            {
+                if (parent == null)
+                {
+                    return MultiPartBlockStateBuilder.PartBuilder.this; 
+                }
+                else {
```
pupnewfster: This bracket should be on the next line, though it may just be cleaner as they are single line things for this entire thing to be:
```java
return parent == null ? MultiPartBlockStateBuilder.PartBuilder.this : parent;
```

src/main/java/net/minecraftforge/client/model/generators/MultiPartBlockStateBuilder.java
```diff
@@ -127,18 +145,167 @@ JsonObject toJson() {
...
+            public MultiPartBlockStateBuilder end()
+            {
+                if (this.parent != null)
+                {
+                    return this.parent.end();
+                }
+                else
+                {
+                    return MultiPartBlockStateBuilder.PartBuilder.this.end();
+                }
```
pupnewfster: This potentially should also be a ternary or at least may be more readable without the brackets so then it would be like:
```java
if (this.parent != null)
    return this.parent.end();
return MultiPartBlockStateBuilder.PartBuilder.this.end();
```

mrp-v2:
Removed the `IConditionGroup` interface and made `ConditionGroup` not allow modders to mess up ending groups and things.

src/main/java/net/minecraftforge/client/model/generators/MultiPartBlockStateBuilder.java
```diff
@@ -107,30 +113,34 @@ public PartBuilder useOr() {
...
-                    }
-                    when.addProperty(e.getKey().getName(), activeString.toString());
-                }
-                if (useOr) {
-                    JsonObject innerWhen = when;
-                    when = new JsonObject();
-                    when.add("OR", innerWhen);
-                }
+                JsonObject when = MultiPartBlockStateBuilder.toJson(this.conditions, this.useOr);
+                out.add("when", when);
```
pupnewfster: This can just be inlined to `out.add("when", MultiPartBlockStateBuilder....` same with the one in the nestedConditionsGroup branch.

mrp-v2: Ah, I didn't catch that.

PR #7678, opened 2021-03-10
[1.16.x] Readd the ability to modify Dimension Lightmap Colours

50ap5ud5:
(This is a copy of #7667 remade to allow for a cleaner git history.)

### About
This PR adds a new event that restores the functionality originally provided by #3922. It adds the ability for modders to adjust the light map colours of a specific dimension.
### What are LightMap Colours
The Lightmap colours are the colours encoded into the brightness of surfaces. By modifying these colours the "tint" of a surface can be changed.
### Why this is needed
From 1.16 onwards, the ability to adjust the lightmap colours of a dimension was removed following the sweeping changes to Dimensions and the Forge Dimension API.
Currently in vanilla Minecraft, these lightmap colours are hardcoded.
### Use Cases
There is evidence of a number of valid usecases for this PR.
For example, RFTools uses it to darken the lighting of their dimensions if the dimension’s “power” is low.
(See the lighting of the dimension darken from 16:41 onwards in the below video)
- https://youtu.be/h53ADXTuNwM?t=1001 

Another use case is in one of my mods, where I interpolate the world’s gameTime with the lighting to make the dimension “flash red” as part of an “alarm” system as shown in the below video.
- https://www.youtube.com/watch?v=gsZH5SC3nbI

I also have use cases to modify the "mood lighting" of my custom dimensions, which is needed for users to have a visual indication to determine the state of a core feature in the mod. The below images show a "warm" and "cold" coloured lighting I use. (This was done in versions below 1.16)

- https://user-images.githubusercontent.com/22887654/109372806-a4b12b00-78ff-11eb-85d2-1296758b5fb5.png
- https://user-images.githubusercontent.com/22887654/109372836-c5798080-78ff-11eb-9681-ba10f2701892.png

### Event Parameters
- `world`. The world object we are modifying. 
- `partialTicks`. Allows for interpolation of colours and other rendering usecases.
- `blockBrightness`. Allows for the darkening or brightening of the block colours in the dimension
- `lightMapColors`. Allows for modification of the colours of the Dimension's LightMap

### Usage of this PR's Event
- Subscribe to the DimensionLightMapModificationEvent event
- Check if the event's world's RegistryKey is that of the world you are modifying
- Use the event's getters to access parameters such as the lightMapColours and blockBrightness
- Call `Vector3f#set` to modify the lightMap colours and call `DimensionLightMapModificationEvent#setBlockBrightness` to set the blockBrightness.

See the the provided example test mod for more details.

### Showcase of this Pull Request's Changes
In the provided example test mod, I have included a custom dimension and set its lightmap colours to flash in different colours based on the partialTicks provided by the PR along with the world's game time.

**Alternating RGB Colours in Lightmap**

_Animated Gif_
https://gyazo.com/5820cf40ba55895ea3f8c526d30c23ff

_Example Screenshot_
LightmapColourExample

Additionally, the PR also includes a parameter for "Block Brightness", which allows modders to adjust how bright or dark the blocks are. See the example below:

**Block Brightness Darkened**
_Before_
BlockBrightnessBefore

_After_
BlockBrightnessAfter

**Lightmap Colours flashing Red**
This is similar to my personal usecase.
https://gyazo.com/8c3caaeb8cfc97c12fec478ed9d7794b 

### Implementation and Discussion
This PR is based off the 1.15 version of #3922 through patching the LightTexture class. 

This patch location has been moved to be fired earlier in the class compared to the 1.15 version, in order to allow for some of its parameters such as BlockBrightness to be properly used.

There are a number of reasons to add an event instead of modifying an existing object like the DimensionRenderInfo:

- DimensionRenderInfo registration involves encouraging poor modding practices. To add a new DimensionRenderInfo you need to use non-api methods to access its private immutable static map.

- DimensionRenderInfo uses the OnlyIn annotation which is known to cause a number of other side effects. We do not want to encourage modders to adopt this practice either.

- DimensionRenderInfo only applies to a DimensionType. One DimensionType can be used for multiple Dimensions. This means the functionality of using DimensionRenderInfo is limited, as modders may not want all Dimensions using that DimensionType to have the same lightmap colours. 

- Keep in mind that the original PR allowed for colour modification per dimension, so using something like DimensionRenderInfo makes it a regression of functionality compared to before.

There are also other reasons I believe this event is a good solution:
- Using an event could reduce future maintenance requirements when Forge is updated to future versions of Minecraft. There are less patches needed to fire this event compared to adding many hooks to existing Minecraft objects.

- There aren't many other suitable places to add this hook. The ClientWorld cannot be easily used by modders, and adding a new object specifically for this hook would require alot more work to allow it to sync with the vanilla Dimension system, which is data driven.

Thank you for considering this pull request. If you have any questions or suggestions feel free to discuss it!


50ap5ud5:
This PR is now ready for review. I have removed variables in the event which were redundant and fixed variables in the event not passing the modified values back to the LightTexture class patch site.

50ap5ud5 (commented):
I wasn't sure about this, so I'd appreciate some feedback; should Modders be setting the values for the colours by allowing them to call Vector3f#set (a public method) manually, or should the LightTexture patch add a line that calls Vector3f#set for them?

I can see cases where Modders may be confused on how to set the colours as the event only provides a getter for the Lightmap colours via DimensionLightMapModificationEvent#getMightMapColors, even though there is a java doc on the getter telling them about calling Vector3f#set

TheCurle:
This one slipped through our radar. 
My initial testing proved this difficult to program for, so we need to add some javadoc, in the most likely case.
If you fix the merge conflict here, I'll be in touch with you for how we move forward with this.

50ap5ud5:
Alright, thank you for the feedback Curle. I've fixed the merge conflict and am happy to discuss this PR further. :)

src/main/java/net/minecraftforge/client/event/DimensionLightMapModificationEvent.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * <p>Allows modders to adjust the lightmap colors for a specific {@code Dimension}.</p>
+ *
+ * <p>While {@code DimensionRenderInfo} could look like the closest solution for this task,
+ * its entries are stored in private, immutable map that is specific to a {@code DimensionType}.
+ * There can be many {@code Dimension}(s) per {@code DimensionType}, which means
+ * {@code DimensionRenderInfo} is limited in both functionality and encourages poor practices.</p>
+ *
+ * <p>Hence, this event is the preferred solution over DimensionRenderInfo.</p>
```
TheCurle, line 33: You can just say "DimensionRenderInfo is similar but uses private, singleton maps that are not Dimension specific.".
Much less wordy.

src/main/java/net/minecraftforge/client/event/DimensionLightMapModificationEvent.java
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * The {@code World} which we're modifying the lightmap colors for. Remember that each {@code Dimension} has one {@code World} instance, so a {@code Dimension} has a one to one relationship with a {@code World}.
+     */
+    public World getWorld()
+    {
+        return this.world;
+    }
+
+    /**
+     * Get the progress between ticks.
```
TheCurle, line 66:
```
/**
 * Get Variable.
 */
public Variable getVariable() {
```

Explain why this is useful, if at all. If it isn't useful, why do we have it?

src/main/java/net/minecraftforge/client/event/DimensionLightMapModificationEvent.java
```diff
@@ -0,0 +1,96 @@
...
+
+    public void setBlockBrightness(float blockBrightness)
+    {
+        this.blockBrightness = blockBrightness;
+    }
+
+    /**
+     * The color values that can be used to adjust lighting in the dimension.
+     * <p> Comprised of {@linkplain Vector3f#x() Red}, {@linkplain Vector3f#y() Green}, {@linkplain Vector3f#z() Blue}.
+     * <br> Call {@link Vector3f#set(float, float, float)} to set the values you modified
```
TheCurle, line 89: You have a setBlockBrightness, why ask them to set the color manually?

TheCurle (changes requested):
Some minor nitpicks.
We need to go over the documentation for modders more in depth at a later date, as it seems you've designed this with your usecase in mind (as is expected) but haven't taken into consideration how others would interact with it.

As an example, what happens if multiple mods change a lightmap in a row?

sciwhiz12:
Hello! @50ap5ud5, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

sciwhiz12:

PR #7684, opened 2021-03-13
Fix reflective access that was broken by Mojmap update

diesieben07:
The SRG name in this reflection code was not correctly split to prevent automatic replacement and thus was incorrectly replaced during the Mojmap update.


gigaherz:
Pushed as e33f794298dff0ebafce89f401e66d9cf6b2bc1e

PR #7686, opened 2021-03-14
[1.16] Fix crash related to EnumArgument and add test

SizableShrimp:
This PR fixes a crash caused by `EnumArgument` if it used with sibling argument nodes that provide examples that are not a valid enum constant. Now, a `CommandSyntaxException` is correctly thrown in the event that an invalid enum constant is provided. This also has the benefit of telling players in-game the possible set of arguments they can provide that are valid enum constants.

### DIFF
--- a/src/main/java/net/minecraftforge/server/command/EnumArgument.java
+++ b/src/main/java/net/minecraftforge/server/command/EnumArgument.java
@@ -24,18 +24,23 @@ import com.mojang.brigadier.StringReader;
 import com.mojang.brigadier.arguments.ArgumentType;
 import com.mojang.brigadier.context.CommandContext;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
+import com.mojang.brigadier.exceptions.Dynamic2CommandExceptionType;
 import com.mojang.brigadier.suggestion.Suggestions;
 import com.mojang.brigadier.suggestion.SuggestionsBuilder;
 import net.minecraft.command.ISuggestionProvider;
 import net.minecraft.command.arguments.IArgumentSerializer;
 import net.minecraft.network.PacketBuffer;
+import net.minecraft.util.text.TranslationTextComponent;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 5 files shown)


SizableShrimp:
Updated to remove unnecessary formatting. Tested with forge server and vanilla client and confirmed to work without issues using the example outlined in `EnumArgumentTest`. The only issue on vanilla clients is that the translation key for an invalid enum argument `CommandSyntaxException` does not exist, which is not fixable.

The uncaught `IllegalArgumentException` does not actually cause a "crash" as I initially labeled it, but rather is caught by Minecraft when loading the world and prevents further world loading. A reproducible example of this issue can be found by using `EnumArgumentTest` and modifying `EnumArgument` to instead re-throw the `IllegalArgumentException` instead of catching it. This causes the world to not load. On the client after attempting to load a world, an option is provided to load only the vanilla datapack with a button labeled "Safe Mode". Since commands are loaded regardless of what datapacks are enabled, the world still cannot load. This means that affected commands prevent world loading until the problematic mod is removed. A stack trace for this issue can be found [here](https://gist.github.com/SizableShrimp/8259259768fa26777d7969bdde965dcf) using the modification outlined above.

The issue lies in the fact that `ArgumentCommandNode#parse(StringReader, CommandContextBuilder)` catches `RuntimeException`s, which an `IllegalArgumentException` falls under, but `ArgumentCommandNode#isValidInput(String)` does not. When checking for ambiguities in `CommandNode#findAmbiguities(AmbiguityConsumer)`, this `isValidInput` method is called. In any scenario where an enum argument has a sibling command node which provides examples to test on that aren't a valid enum constant, it will inherently fail during this ambiguity check as the `IllegalArgumentException` is not caught and propagated up until it is caught by `Minecraft#loadWorld` (mojmaps). More info can be found on the Forge discord [here](https://discord.com/channels/313125603924639766/725850371834118214/820567534264385537) and [here](https://discord.com/channels/313125603924639766/725850371834118214/818697103542976532).

diesieben07:
Looks good, but please fix the formatting in EnumArgument (braces on new lines).

LexManos:
I told him not to so he could keep his diff small.

PR #7693, opened 2021-03-19
Add RenderItemEvent

exokem:
This pull request intends to afford mods greater control over how items are rendered by the ItemRenderer. This includes what is drawn beneath rendered items, as well as the item counts, durabilities, and cooldowns rendered above them. Some possible use cases are as follows:

- Any mod that renders additional durability bars over items, for example a mod with tools that have both energy and durability.
- A mod adjusting how the durability, item count, or item cooldown are rendered over items, including the colors used and the specific visual methods used to represent each property.
- Visual adjustments to certain itemstacks (e.g. overlaid markings or graphics used to further categorize items outside of the creative inventory) rendered above the default overlay or beneath the item rendering itself.

I have encountered at least the first listed use case in one of my own mods, and am reasonably certain that the events introduced by this pull request would simplify what would otherwise require a lot of awkward workarounds. While a custom durability bar would have previously required multiple event handlers for the display of items in different contexts (while being moved in the gui, when rendered in the hotbar, and when rendered in the gui) the RenderItemEvent added by this pull request combines the same adjustments into a single event. 

The event centralizes the possibility for item overlay customization about the item itself, rather than the context in which it is displayed.

---

This pull request includes a RenderItemEventTest class that makes use of each new event to achieve a certain result. An overview of the sample uses is as follows:

- A replacement of the default durability rendering that uses colored rectangles (behind damaged tools) that scale horizontally with the tool's durability (using RenderItemEvent.Model to render and the DURABILITY variant of RenderItemEvent.Overlay to cancel the default overlay).
- A replacement of the default item count rendering that turns the numerical display into a progress bar instead (using the COUNT variant of the RenderItemEvent.Overlay).
- A replacement of the default item cooldown rendering that turns the cooldown highlight into a red progress bar that ticks down as the cooldown time does (using the COOLDOWN variant of the RenderItemEvent.Overlay).
- A gold square rendered above all edible items (using the EXTRA variant of the RenderItemEvent.Overlay).

The documentation for RenderItemEvent contains specific instructions for how each event should be used, which can be canceled, and when they are fired on the forge event bus.

### DIFF
--- a/patches/minecraft/net/minecraft/client/renderer/ItemRenderer.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/ItemRenderer.java.patch
@@ -9,7 +9,14 @@
  
        for(Item item : Registry.field_212630_s) {
           if (!field_195411_c.contains(item)) {
-@@ -95,7 +_,7 @@
+@@ -89,13 +_,14 @@
+ 
+    public void func_229111_a_(ItemStack p_229111_1_, ItemCameraTransforms.TransformType p_229111_2_, boolean p_229111_3_, MatrixStack p_229111_4_, IRenderTypeBuffer p_229111_5_, int p_229111_6_, int p_229111_7_, IBakedModel p_229111_8_) {
+       if (!p_229111_1_.func_190926_b()) {
++         net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.client.event.RenderItemEvent.Model(p_229111_4_, p_229111_1_, p_229111_5_, p_229111_6_, p_229111_7_, p_229111_8_));
+          p_229111_4_.func_227860_a_();
+          boolean flag = p_229111_2_ == ItemCameraTransforms.TransformType.GUI || p_229111_2_ == ItemCameraTransforms.TransformType.GROUND || p_229111_2_ == ItemCameraTransforms.TransformType.FIXED;
+          if (p_229111_1_.func_77973_b() == Items.field_203184_eO && flag) {
              p_229111_8_ = this.field_175059_m.func_178083_a().func_174953_a(new ModelResourceLocation("minecraft:trident#inventory"));
           }
  
@@ -56,12 +63,21 @@
              crashreportcategory.func_189529_a("Item Damage", () -> {
                 return String.valueOf(p_239387_2_.func_77952_i());
              });
+@@ -290,7 +_,7 @@
+    public void func_180453_a(FontRenderer p_180453_1_, ItemStack p_180453_2_, int p_180453_3_, int p_180453_4_, @Nullable String p_180453_5_) {
+       if (!p_180453_2_.func_190926_b()) {
+          MatrixStack matrixstack = new MatrixStack();
+-         if (p_180453_2_.func_190916_E() != 1 || p_180453_5_ != null) {
++         if ((p_180453_2_.func_190916_E() != 1 || p_180453_5_ != null) && !net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.client.event.RenderItemEvent.Overlay(p_180453_1_, matrixstack, p_180453_2_, p_180453_3_, p_180453_4_, net.minecraftforge.client.event.RenderItemEvent.Overlay.OverlayPart.COUNT))) {
+             String s = p_180453_5_ == null ? String.valueOf(p_180453_2_.func_190916_E()) : p_180453_5_;
+             matrixstack.func_227861_a_(0.0D, 0.0D, (double)(this.field_77023_b + 200.0F));
+             IRenderTypeBuffer.Impl irendertypebuffer$impl = IRenderTypeBuffer.func_228455_a_(Tessellator.func_178181_a().func_178180_c());
 @@ -298,18 +_,16 @@
              irendertypebuffer$impl.func_228461_a_();
           }
  
 -         if (p_180453_2_.func_77951_h()) {
-+         if (p_180453_2_.func_77973_b().showDurabilityBar(p_180453_2_)) {
++         if (p_180453_2_.func_77973_b().showDurabilityBar(p_180453_2_) && !net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.client.event.RenderItemEvent.Overlay(p_180453_1_, matrixstack, p_180453_2_, p_180453_3_, p_180453_4_, net.minecraftforge.client.event.RenderItemEvent.Overlay.OverlayPart.DURABILITY))) {
              RenderSystem.disableDepthTest();
              RenderSystem.disableTexture();
              RenderSystem.disableAlphaTest();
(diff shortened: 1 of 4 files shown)


KingLemming:
A concern I would have with this is just how OFTEN renderItem gets called.

I don't disagree with your use cases, but I definitely worry about how much lag this could cause.

exokem:
> A concern I would have with this is just how OFTEN renderItem gets called.
> 
> I don't disagree with your use cases, but I definitely worry about how much lag this could cause.

If the number of events fired in renderItem is problematic a couple of the events are removable without much loss to the functionality. The last fired event would be removable, and this would only be troublesome when the desired result is to render above item cooldowns. Even so, the removal of that last event would not make it impossible to do so. Further, the first event (fired before the item model is rendered) could also be removed, but this would decentralize the process of rendering beneath items. 

Alternatively, the ability of the events fired to cancel default overlay renderings could be combined into a single event. In this case, a single event would be fired (before any parts of the overlays of non-empty items are rendered) with states that would subsequently determine whether certain parts of the overlay are drawn. A handler for this event could simultaneously allow rendering beneath all parts of the default overlay while also allowing the removal of specific parts of it. A final event could be fired at the end of the overlay rendering to allow rendering on top of the entire overlay, and what is now four events would become two. 

With that said, such a reduction would eliminate the ability for handlers to render between specific parts of the item overlay, though I can't myself think of use cases where such specificity would be required.

Otherwise, I would only wonder the maximum impact all of the events could actually have, presuming that the number of calls to renderItem is limited to the number of items currently displayed. I will absolutely look into testing the projected lag if there is a viable means to do so.

KingLemming:
Make a chest with 100+ slots visible at once. Add...1 item that actually uses these events and put it in the inventory, along with non-empty stacks in every other slot. But add 40 mods that subscribe to it and have to check every single call for the item they care about. That's how I'd start with stress-testing.

Then make it 10 items and 200 mods. This is the scale of modded Minecraft and we have to code for it, unfortunately.

exokem:
> Make a chest with 100+ slots visible at once. Add...1 item that actually uses these events and put it in the inventory, along with non-empty stacks in every other slot. But add 40 mods that subscribe to it and have to check every single call for the item they care about. That's how I'd start with stress-testing.
> 
> Then make it 10 items and 200 mods. This is the scale of modded Minecraft and we have to code for it, unfortunately.

I've spent some more time stress testing as you suggested - I added a 256 slot container and generated 100 mods. Each mod implements its own item and refers to it in a subscriber to the event I have added (the subscriber catches all four variants of the RenderItemEvent.Overlay event, so it runs four times for each item), and even when all 100 items are visible inside of the 256 slot container with non-empty stacks in between them, there does not appear to be a noticeable impact on the game performance.

In anticipation of greater impact though, I could reduce the number of fired events, if that would be beneficial, by combining their effects.

covers1624:
`there does not appear to be a noticeable impact on the game performance` Provide numbers, this means nothing.

exokem:
> `there does not appear to be a noticeable impact on the game performance` Provide numbers, this means nothing.

What should I be looking for? I'm not sure which figures are meaningful.

Shadows-of-Fire:
What does this solve that is not already possible through the ISTER system?

Shadows-of-Fire:
And as far as numbers are concerned, you would need to be looking at frame times with and without the patch applied, then with and without a test mod applied, and then again with and without several test mods applied (as that is a very possible scenario)

exokem:
> What does this solve that is not already possible through the ISTER system?

Functionally, I cannot suggest what this makes possible that is not already, in part because I am not familiar enough with the ISTER system to know everything it enables, and also because in all of my efforts to find specific, intuitive, and well documented usages of the ISTER system I came to nothing that brought me nearer to my intended result.

I can, however, suggest that to someone such as myself who is not inherently expert in all of the tools available to them, this change provides a more intuitive alternative. It also seems that the complete rendering adjustments that the ISTER system offers might be more than is needed for simpler use cases - why would I create a completely individualized rendering function for an item for which I only want to render an underlay or overlay? I have attempted to use an ISTER for my own use case (which is very simply to draw a rectangular durability bar over some items) and found, perhaps ignorantly, that using an ISTER might require me to manually re-render the item model itself, just so I can draw my extra rectangle. 

The last I can offer is that this allows customized rendering of _all_ items, regardless of the mod that owns them (assuming that it is possible to retrieve the items of other mods for comparison). I could not, to my understanding, use an ISTER to change how the counts for all items are rendered without attempting to modify the default behavior directly (e.g. using a mixin). This series of events allows adjusting and removing specific parts of universally displayed details, which I do not see to be possible with individual implementations of ISTERs.

XFactHD:
Everything this does is definitely possible with an ISTER (in 1.16 atleast, due to ISTER#render() getting the ItemTransformType argument) but in a lot of cases where these events are useful (such as the overlays), an ISTER would be completely overkill, mostly because you need to render the item model completely yourself (see [here](https://github.com/XFactHD/AdvancedTechnology/blob/1.15.2/src/main/java/xfacthd/advtech/client/render/ister/RenderItemFluidTank.java) for an example of that - version doesn't matter in this case).
Changing the rendering of items from vanilla or other mods can not be done with an ISTER because the ISTER is set in the item properties.

TheCosmicNebula:
Anyone know when this is likely to be committed? I have literally started today on trying to implement this exact functionality.

LexManos:
There is no plans to accept this, we're still waiting on justification, and quantifiable numbers.
This type of catch-all event is not good for the reasons specified by the others.
Combined with not having a good justification as to why our existing hooks will not work. This is stalled until they can explain themselves and return with proper performance numbers.

TheCosmicNebula:
Okay many thanks for the reply. I am unsure as to how to get this working with an ISTER is all.

ferriarnus:
I have/had the same issue. I am trying to render a second durability bar for items. The issue is that the item can be any item with durability (so also vanilla and other mods), so using ISTER is not possible for those items. With the current events, you can not render the bar when you're holding the item with the mouse. Currently I used a Mixin, but it would be nice if I didn't need to.

ChampionAsh5357:
> I have/had the same issue. I am trying to render a second durability bar for items. The issue is that the item can be any item with durability (so also vanilla and other mods), so using ISTER is not possible for those items. With the current events, you can not render the bar when you're holding the item with the mouse. Currently I used a Mixin, but it would be nice if I didn't need to.

This doesn't really need a new event. While it would be more convenient, you could use a combination of RenderGameOverlayEvent and DrawScreenEvent. This event is more of a helper to do less math calculations more than anything.

ferriarnus:
It seems you're wright aboud using a combination of RenderGameOverlayEvent and DrawScreenEvent. I tried GuiContainerEvent before, but it doesn't allow you to draw over a "dragged" item. DrawScreenEvent does, so now I am doing the "math calculations" without my mixin. Thanks for the help and clarification.

sciwhiz12:
Hello! @Exokem, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

exokem:
> Hello! @Exokem, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity. If anyone else wishes to continue the work of this PR, please create a PR with a link to this PR for reference. Thank you for your contributions. 👋

As much as I would like to see a change to simplify the process of rendering simple graphics over items, I am not actively planning to pursue this PR to any further extent myself. I am not aware of whether or not it remains to be applicable to the latest version. I did attempt to run some stress tests as suggested (with seemingly fine results) but it seems more thorough testing would be necessary before this PR could be accepted. I am not presently interested in this process and I have adapted a solution to my own individual case so I have no issue with the closure of this PR.

sciwhiz12:
Sure, I'll be closing this PR since you've found a solution for your individual usecase. If anyone wishes to look into this, please feel free (and heed the comments from this PR as well).

TheCosmicNebula:
For anyone interested, you can use the vanilla system (for your own items) by using these methods:
`public boolean isBarVisible(ItemStack stackIn);`
`public int getBarColor(ItemStack stackIn);`
`public int getBarWidth(ItemStack stackIn);`


As it gives you access to the ItemStack, and therefore the NBT. 

I came up with an extremely convoluted system using a BEWLR or once known as an ISTER. I have discovered this method and for my use case at least it is so much better. It wont give you access to other mod items information, or allow you to render a bar for them, but it may help.

PR #7694, opened 2021-03-19
Add generic EntityTeleportEvent

SizableShrimp:
Forge currently does not have a generic `EntityTeleportEvent` that can be used for catching any and all teleportation related to an `Entity`. This PR adds the functionality along with hooks into vanilla, enabling the events.

Currently, `TeleportCommand`, `SpreadPlayersCommand`, `EnderEntity`, `EnderPearl`, and `ChorusFruit` are all supported subclasses. Documentation has also been added to the generic event and all subclasses. `EnderTeleportEvent` has been deprecated for removal in 1.17 in favor of `EntityTeleportEvent.EnderEntity` and `EntityTeleportEvent.EnderPearl`.

A test mod is also provided that logs whenever an entity teleport event takes place. I also did local testing and confirmed that cancelling the event works in all supported subclasses.

I tried to keep my patch changes minimal, but please let me know if there is anything else I can do to minimize them further.

### DIFF
--- a/patches/minecraft/net/minecraft/command/impl/SpreadPlayersCommand.java.patch
+++ b/patches/minecraft/net/minecraft/command/impl/SpreadPlayersCommand.java.patch
@@ -1,0 +1,12 @@
+--- a/net/minecraft/command/impl/SpreadPlayersCommand.java
++++ b/net/minecraft/command/impl/SpreadPlayersCommand.java
+@@ -162,7 +_,8 @@
+             spreadplayerscommand$position = p_241072_2_[i++];
+          }
+ 
+-         entity.func_223102_j((double)MathHelper.func_76128_c(spreadplayerscommand$position.field_198713_a) + 0.5D, (double)spreadplayerscommand$position.func_198710_a(p_241072_1_, p_241072_3_), (double)MathHelper.func_76128_c(spreadplayerscommand$position.field_198714_b) + 0.5D);
++         net.minecraft.util.math.vector.Vector3d target = net.minecraftforge.event.ForgeEventFactory.onEntityTeleportSpreadPlayersCommand(entity, (double)MathHelper.func_76128_c(spreadplayerscommand$position.field_198713_a) + 0.5D, (double)spreadplayerscommand$position.func_198710_a(p_241072_1_, p_241072_3_), (double)MathHelper.func_76128_c(spreadplayerscommand$position.field_198714_b) + 0.5D);
++         if (target != null) entity.func_223102_j(target.field_72450_a, target.field_72448_b, target.field_72449_c);
+          double d2 = Double.MAX_VALUE;
+ 
+          for(SpreadPlayersCommand.Position spreadplayerscommand$position1 : p_241072_2_) {
--- a/patches/minecraft/net/minecraft/command/impl/SpreadPlayersCommand.java.patch
+++ b/patches/minecraft/net/minecraft/command/impl/SpreadPlayersCommand.java.patch
@@ -5,8 +5,8 @@
           }
  
 -         entity.func_223102_j((double)MathHelper.func_76128_c(spreadplayerscommand$position.field_198713_a) + 0.5D, (double)spreadplayerscommand$position.func_198710_a(p_241072_1_, p_241072_3_), (double)MathHelper.func_76128_c(spreadplayerscommand$position.field_198714_b) + 0.5D);
-+         net.minecraft.util.math.vector.Vector3d target = net.minecraftforge.event.ForgeEventFactory.onEntityTeleportSpreadPlayersCommand(entity, (double)MathHelper.func_76128_c(spreadplayerscommand$position.field_198713_a) + 0.5D, (double)spreadplayerscommand$position.func_198710_a(p_241072_1_, p_241072_3_), (double)MathHelper.func_76128_c(spreadplayerscommand$position.field_198714_b) + 0.5D);
-+         if (target != null) entity.func_223102_j(target.field_72450_a, target.field_72448_b, target.field_72449_c);
++         net.minecraftforge.event.entity.living.EntityTeleportEvent.SpreadPlayersCommand event = net.minecraftforge.event.ForgeEventFactory.onEntityTeleportSpreadPlayersCommand(entity, (double)MathHelper.func_76128_c(spreadplayerscommand$position.field_198713_a) + 0.5D, (double)spreadplayerscommand$position.func_198710_a(p_241072_1_, p_241072_3_), (double)MathHelper.func_76128_c(spreadplayerscommand$position.field_198714_b) + 0.5D);
++         if (!event.isCanceled()) entity.func_223102_j(event.getTargetX(), event.getTargetY(), event.getTargetZ());
           double d2 = Double.MAX_VALUE;
  
           for(SpreadPlayersCommand.Position spreadplayerscommand$position1 : p_241072_2_) {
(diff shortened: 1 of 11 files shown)


patches/minecraft/net/minecraft/command/impl/TeleportCommand.java.patch
```diff
@@ -1,5 +1,17 @@
...
 +++ b/net/minecraft/command/impl/TeleportCommand.java
+@@ -121,6 +_,11 @@
+    }
+ 
+    private static void func_201127_a(CommandSource p_201127_0_, Entity p_201127_1_, ServerWorld p_201127_2_, double p_201127_3_, double p_201127_5_, double p_201127_7_, Set<SPlayerPositionLookPacket.Flags> p_201127_9_, float p_201127_10_, float p_201127_11_, @Nullable TeleportCommand.Facing p_201127_12_) throws CommandSyntaxException {
++      net.minecraft.util.math.vector.Vector3d target = net.minecraftforge.event.ForgeEventFactory.onEntityTeleportCommand(p_201127_1_, p_201127_3_, p_201127_5_, p_201127_7_);
++      if (target == null) return;
++      p_201127_3_ = target.field_72450_a;
++      p_201127_5_ = target.field_72448_b;
++      p_201127_7_ = target.field_72449_c;
```
robotgryphon: Can this be traded for a cancelled event and a one-line version for the x-y-z values?
```java
if(!event) return;
p_201127_3_ = event.target.field_72450_a; p_201127_5_ = event.target.field_72448_b; p_201127_7_ = event.target.field_72449_c;
```

patches/minecraft/net/minecraft/command/impl/TeleportCommand.java.patch
```diff
@@ -1,5 +1,17 @@
...
+       BlockPos blockpos = new BlockPos(p_201127_3_, p_201127_5_, p_201127_7_);
+       if (!World.func_234935_k_(blockpos)) {
+          throw field_241077_a_.create();
```
robotgryphon, line 12: Checking the position is loaded? Explain why this is added to the patch.

SizableShrimp: I didn't add this, so I'm not sure how it appeared.

patches/minecraft/net/minecraft/command/impl/TeleportCommand.java.patch
SizableShrimp: I'll refactor this soon

SizableShrimp: Fixed with 8b629ad.

patches/minecraft/net/minecraft/command/impl/TeleportCommand.java.patch
SizableShrimp: Actually, this is automatically generated to show the relevant lines around the patch. This is automatic. Notice how there is no `+` or `-` at the start of the line, meaning this was not changed and is just being displayed.

patches/minecraft/net/minecraft/command/impl/SpreadPlayersCommand.java.patch
```diff
@@ -0,0 +1,12 @@
+--- a/net/minecraft/command/impl/SpreadPlayersCommand.java
++++ b/net/minecraft/command/impl/SpreadPlayersCommand.java
+@@ -162,7 +_,8 @@
+             spreadplayerscommand$position = p_241072_2_[i++];
+          }
+ 
+-         entity.func_223102_j((double)MathHelper.func_76128_c(spreadplayerscommand$position.field_198713_a) + 0.5D, (double)spreadplayerscommand$position.func_198710_a(p_241072_1_, p_241072_3_), (double)MathHelper.func_76128_c(spreadplayerscommand$position.field_198714_b) + 0.5D);
++         net.minecraftforge.event.entity.living.EntityTeleportEvent.SpreadPlayersCommand event = net.minecraftforge.event.ForgeEventFactory.onEntityTeleportSpreadPlayersCommand(entity, (double)MathHelper.func_76128_c(spreadplayerscommand$position.field_198713_a) + 0.5D, (double)spreadplayerscommand$position.func_198710_a(p_241072_1_, p_241072_3_), (double)MathHelper.func_76128_c(spreadplayerscommand$position.field_198714_b) + 0.5D);
++         if (!event.isCanceled()) entity.func_223102_j(event.getTargetX(), event.getTargetY(), event.getTargetZ());
```
robotgryphon, line 9: Else condition for the event here? Shouldn't this perform an early exit or continue?

SizableShrimp: The patch does not fully show the scope of this method execution. The code in question is actually executed in a loop for all entities selected by the selector or w/e in `/spreadplayers`. Exiting early would cause all entities after the one that is canceled from teleportation to also be canceled, which would be an unintended side effect. This way, entities can be selectively blocked from teleporting instead of all of them.

actuallyasmartname (approved):

patches/minecraft/net/minecraft/command/impl/TeleportCommand.java.patch
TheCurle: Ah yes, a patch of a patch. Truly the bane of our collective existence.

patches/minecraft/net/minecraft/entity/item/EnderPearlEntity.java.patch
```diff
@@ -1,23 +1,26 @@
...
                 if (this.field_70146_Z.nextFloat() < 0.05F && this.field_70170_p.func_82736_K().func_223586_b(GameRules.field_223601_d)) {
                    EndermiteEntity endermiteentity = EntityType.field_200804_r.func_200721_a(this.field_70170_p);
                    endermiteentity.func_175496_a(true);
-@@ -66,9 +_,10 @@
+@@ -66,9 +_,11 @@
                    entity.func_184210_p();
                 }
  
 -               entity.func_70634_a(this.func_226277_ct_(), this.func_226278_cu_(), this.func_226281_cx_());
-+               entity.func_70634_a(event.getTargetX(), event.getTargetY(), event.getTargetZ());
```
TheCurle: Is there any reason you can't use the event's positions here, just so the patch is smaller? And when the Ender event is scrapped in 1.17, it should be obvious to use the new "evt"'s position..

src/test/java/net/minecraftforge/debug/entity/EntityTeleportEventTest.java
```diff
@@ -0,0 +1,28 @@
...
+    public static final Logger LOGGER = LogManager.getLogger();
+
+    public EntityTeleportEventTest()
+    {
+        if (ENABLE)
+            MinecraftForge.EVENT_BUS.register(this);
+    }
+
+    @SubscribeEvent
+    public void registerCommands(EntityTeleportEvent event)
```
TheCurle: Bizarre name for a teleport listener..

TheCurle (approved):
Seems good. Some small optimisations are possible, but nothing world shattering.

robotgryphon (approved):

SizableShrimp:
Updated to reduce the patch size. `EnderPearlEntity`, `EndermanEntity`, and `ShulkerEntity` have all been refactored to instead use a similar call to the original `EnderTeleportEvent` to reduce changes. `ForgeEventFactory` has also been modified to post an `EnderTeleportEvent` first for the 2 related events (`EnderEntity` & `EnderPearl`), and then modify the event with the values returned from the post of this event on the event bus. Removing this in 1.17 should now be simpler needing only to have the event class itself deleted, and the relevant method + 2 lines in `ForgeEventFactory` removed as well. Comments have been added detailing which lines need to be removed in 1.17.

TheCurle (approved):
That's now a lot easier to read. The code seems good to me, and some testing shows it works.

PR #7695, opened 2021-03-22
Added flag to distinguish spawning and loading from disk

tmvkrpxl0:
This commit adds flag that distinguishes spawning entity and loading from disk
Example Mod: https://del.dog/ywulybocis.go
After applying the mod, every newly spawned entities will launched into sky, but once being loaded from disk won't

### DIFF
--- a/patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
+++ b/patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
@@ -88,15 +88,21 @@
              }
  
              iprofiler.func_76319_b();
-@@ -475,6 +_,7 @@
+@@ -475,9 +_,10 @@
           BlockPos blockpos2 = this.func_205770_a(Heightmap.Type.MOTION_BLOCKING, this.func_217383_a(i, 0, j, 15));
           BlockPos blockpos3 = blockpos2.func_177977_b();
           Biome biome = this.func_226691_t_(blockpos2);
+-         if (biome.func_201848_a(this, blockpos3)) {
+-            this.func_175656_a(blockpos3, Blocks.field_150432_aD.func_176223_P());
+-         }
 +         if (this.isAreaLoaded(blockpos2, 1)) // Forge: check area to avoid loading neighbors in unloaded chunks
-          if (biome.func_201848_a(this, blockpos3)) {
-             this.func_175656_a(blockpos3, Blocks.field_150432_aD.func_176223_P());
-          }
-@@ -598,9 +_,10 @@
++            if (biome.func_201848_a(this, blockpos3)) {
++               this.func_175656_a(blockpos3, Blocks.field_150432_aD.func_176223_P());
++            }
+ 
+          if (flag && biome.func_201850_b(this, blockpos2)) {
+             this.func_175656_a(blockpos2, Blocks.field_150433_aE.func_176223_P());
+@@ -598,10 +_,11 @@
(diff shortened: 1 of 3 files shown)


tmvkrpxl0:
oh sorry, i accidentally uploaded commit as BuildTool and not me

patches/minecraft/net/minecraft/world/server/ServerWorld.java.patch
```diff
@@ -88,26 +88,34 @@
...
-             this.func_175656_a(blockpos3, Blocks.field_150432_aD.func_176223_P());
-          }
-@@ -598,9 +_,10 @@
++            if (biome.func_201848_a(this, blockpos3)) {
++               this.func_175656_a(blockpos3, Blocks.field_150432_aD.func_176223_P());
++            }
+ 
+          if (flag && biome.func_201850_b(this, blockpos2)) {
+             this.func_175656_a(blockpos2, Blocks.field_150433_aE.func_176223_P());
+@@ -598,10 +_,11 @@
```
tmvkrpxl0, line 105: I honestly don't know why those lines are changed, I only edited lines where they create new EntityJoinWorldEvent

tmvkrpxl0:
Reopening pull request with correct account

diesieben07 (changes requested):
- Please fix the patches.
- You're setting the `isSpawn` value incorrectly, `ServerWorld#addEntity(Entity)` is used for newly spawned entities. I am not sure if it's even possible to truly distinguish the two.
- Your change to `EntityJoinWorldEvent` is a breaking change. You can't just add parameters.

That said, what is this even useful for (given that it works)?

tmvkrpxl0:
> * Please fix the patches.
> * You're setting the `isSpawn` value incorrectly, `ServerWorld#addEntity(Entity)` is used for newly spawned entities. I am not sure if it's even possible to truly distinguish the two.
> * Your change to `EntityJoinWorldEvent` is a breaking change. You can't just add parameters.
> 
> That said, what is this even useful for (given that it works)?

I see, I fixed `ServerWorld#addEntity(Entity)` so that it sets isSpawn to true. About adding parameter to EntityJoinWorldEvent, i don't understand why it is a breaking change
The flag `isSpawn` will tell if entity is being spawned or being loaded from disk so that mods that modify vanila entities or other mod's entity only when they initially spawn and don't have access to their original constructor can work easier and less hacky
"modify" i meant above is not modifying class, rather changing velocity or setting inventory, etc..

diesieben07:
> About adding parameter to EntityJoinWorldEvent, i don't understand why it is a breaking change

Because code that calls this constructor will no longer compile and also no longer link (i.e. `NoSuchMethodError` at runtime).

> The flag isSpawn will tell if entity is being spawned or being loaded from disk so that mods that modify vanila entities or other mod's entity only when they initially spawn and don't have access to their original constructor can work easier and less hacky
"modify" i meant above is not modifying class, rather changing velocity or setting inventory, etc..

There are much better ways to detect this. It also does not cover the case when a mod is added after the fact. Overall I don't think this is needed just for this use-case.

tmvkrpxl0:
I see, i should have looked into forge events deeper.
Could you tell me what is "much better ways"?
I'm looking for way to detect when arrow is spawned

diesieben07:
> Could you tell me what is "much better ways"?

Put a flag in the entity data that tells you whether or not you've seen this entity before. For further questions make a topic on the forums.

PR #7697, opened 2021-03-24
Bypass chunk future chain when processing loads and getChunk called.

covers1624:
The Problem:
World interaction from places such as `TileEntity.onLoad()`, `ChunkEvent.Load`, and `EntityJoinWorldEvent` (possibly others), sometimes causes calls to `World.getChunk` and by extension `ServerChunkProvider.getChunk`. As these methods/events are called from within the Chunk CompletableFututre chain, it is incredibly easy to cause a loading deadlock, as the current thread starts blocking waiting for the CompletableFuture chain to exit, whilst it's already executing.

Current workarounds for this would be to delay any of your interaction with the world from any of the above, until the end of the tick or start of the next. However, this itself is still not a perfect solution either as some interactions might not be possible to delay, (replacing tiles without skipping ticks, setting up the internal state of things that are time-sensitive).

The solution:
There have been recent attempts at resolving this problem such as [this PR](https://github.com/MinecraftForge/MinecraftForge/pull/7532), ~~however, this doesn't really solve the root issue, instead opting for a delayed workaround as mentioned above for Item custom ItemEntities. With this PR. these changes become irrelevant and can likely be reverted, however, this is not verified.~~

This PR attempts to resolve the root issue outright by simply bypassing calls to the future chain in `ServerChunkProvider.getChunk` whilst the chunk is being post-processed (tiles/entities added and `ChunkEvent.Load`).

Is this safe?:
Yes, I believe so. From all the testing I have done with these patches, the bypass is not hit during normal gameplay, instead, _only_ when a deadlock would have occurred. Additionally, at the time the bypass is active, the chunk has already been marked as `FULL` and as loaded, the future chain just needs to exit out. Whilst this solution may not be perfect or the cleanest, there does not appear to be any cleaner way.

Other considerations:
At first, I considered moving the content post-processing blob to a separate detached future to be automatically executed once the future containing the initial post-processing had finished. However, it was impossible to assert that the post-process section of a given chunk was executed before any other events/tasks were processed by the executor.

### DIFF
--- a/patches/minecraft/net/minecraft/world/server/ChunkHolder.java.patch
+++ b/patches/minecraft/net/minecraft/world/server/ChunkHolder.java.patch
@@ -1,5 +1,13 @@
 --- a/net/minecraft/world/server/ChunkHolder.java
 +++ b/net/minecraft/world/server/ChunkHolder.java
+@@ -61,6 +_,7 @@
+    private final ChunkHolder.IPlayerProvider field_219328_w;
+    private boolean field_219329_x;
+    private boolean field_244384_x;
++   Chunk currentlyLoading; //Forge - Used to bypass future chain when loading chunks.
+ 
+    public ChunkHolder(ChunkPos p_i50716_1_, int p_i50716_2_, WorldLightManager p_i50716_3_, ChunkHolder.IListener p_i50716_4_, ChunkHolder.IPlayerProvider p_i50716_5_) {
+       this.field_219319_n = p_i50716_1_;
 @@ -204,7 +_,7 @@
     }
  
--- a/patches/minecraft/net/minecraft/world/server/ChunkManager.java.patch
+++ b/patches/minecraft/net/minecraft/world/server/ChunkManager.java.patch
@@ -8,11 +8,23 @@
                 }
  
                 this.func_219229_a(p_219185_5_);
-@@ -601,6 +_,7 @@
+@@ -581,6 +_,8 @@
+             chunk.func_217318_w();
+             if (this.field_219254_h.add(chunkpos.func_201841_a())) {
+                chunk.func_177417_c(true);
++               try {
++               p_219200_1_.currentlyLoading = chunk; //Forge - bypass the future chain when getChunk is called, this prevents deadlocks.
+                this.field_219255_i.func_147448_a(chunk.func_177434_r().values());
+                List<Entity> list = null;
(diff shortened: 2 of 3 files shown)


AterAnimAvis:
:+1: Seems like a much cleaner/better implementation of my first pass for a fix before I pivoted https://github.com/MinecraftForge/MinecraftForge/issues/7519#issuecomment-735245822 .

Actually does this handle the ConcurrentModificationException I was getting from adding Entities to the chunk? [relevant patch](https://gist.github.com/AterAnimAvis/800eb413eac59434592387e5305bb5db#file-forge7519-patch-L43)

covers1624:
> Actually does this handle the ConcurrentModificationException I was getting from adding Entities to the chunk?

Not from what I can see, this would only solve the case of re-entry to `getChunk` causing a deadlock. However, that issue could be solved by switching the for-each in the post-process to iterate over the `getAllInstances` result of `ClassInheritenceMultiMap`, albeit at the cost of copying the list. Or switching it to use a for-i over the underlying `allInstances` field of `ClassInheritenceMultiMap`.

TelepathicGrunt:
Quick question, will this PR fix deadlocks caused by people calling serverWorld.addEntity(...) that calls .getChunk() in any entity spawn events while the chunk is currently being generated? It seems to be a very common issues as I found two mods today with that deadlocking. There many be a ton more other mods with unsafe addEntity code in mob spawning events. 
<https://github.com/Alex-the-666/Rats/issues/671>
<https://github.com/ricksouth/serilum-mc-mods/issues/526>

covers1624:
I don't believe so, is there a minimal test mod to reproduce this issue?

AterAnimAvis:
Sorry this is what I meant by my question earlier. It solves the deadlock but then causes a crash due to Concurrent Modification Exception. 

A fairly dirty test event (you can’t just use the mod because it would break your fix). Load a world throw an item. Save and exit. Reload the world and it should crash from what I remember.

<https://github.com/AterAnimAvis/ForgeBugs/blob/2fe158947531967a2b4df4c145cd2b427f576504/src/main/java/f7519/F7519.java#L42>

TelepathicGrunt:
@covers1624 Fastest way to reproduce the deadlock I mention is to put The Bumblezone mod on and Realistic Bees on. Then enter The Bumblezone dimension and it should deadlock right away as I spawn bees as part of worldgen which Realistic Bees deadlocks with their spawn event code.

I should make a minimum example mod to reproduce it alone

PR #7707, opened 2021-04-03
Add JVM options for Java 16.

ZekerZhayard:
>https://github.com/cpw/modlauncher/issues/65#issuecomment-807578410

`-XX:+IgnoreUnrecognizedVMOptions` should work with Java 8u51.


LexManos:
It is our current stance that if someone wants to test out J16, they have to manually change their JRE in the launcher anyways so they can add their own arguments.
We have work that will be going into targeting J16 better. When we have the time/reason to do it.

PR #7712, opened 2021-04-10
bump gradle to 6.8.3, update deps

actuallyasmartname:


DaemonUmbra:
Can you justify this change?

kashike:
We cannot bump dependency versions for most things that are provided by vanilla - there are incompatible changes in some libraries, and the game itself, along with mods, would break in some way.

actuallyasmartname:
Ok

PR #7715, opened 2021-04-11
Add an event when a shootable item tries to find a projectile item.

noeppi-noeppi:
This allows for example for bows to find arrows from quivers.

### DIFF
--- a/patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
@@ -329,7 +329,7 @@
        return this.func_208016_c(iformattabletextcomponent);
     }
  
-@@ -1901,18 +_,19 @@
+@@ -1901,18 +_,18 @@
           Predicate<ItemStack> predicate = ((ShootableItem)p_213356_1_.func_77973_b()).func_220006_d();
           ItemStack itemstack = ShootableItem.func_220005_a(this, predicate);
           if (!itemstack.func_190926_b()) {
@@ -347,8 +347,7 @@
              }
  
 -            return this.field_71075_bZ.field_75098_d ? new ItemStack(Items.field_151032_g) : ItemStack.field_190927_a;
-+            return net.minecraftforge.common.ForgeHooks.findProjectile(this, p_213356_1_,
-+                    this.field_71075_bZ.field_75098_d ? new ItemStack(Items.field_151032_g) : ItemStack.field_190927_a);
++            return net.minecraftforge.common.ForgeHooks.findProjectile(this, p_213356_1_, this.field_71075_bZ.field_75098_d ? new ItemStack(Items.field_151032_g) : ItemStack.field_190927_a);
           }
        }
     }
--- a/patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
@@ -329,6 +329,29 @@
        return this.func_208016_c(iformattabletextcomponent);
     }
  
+@@ -1901,18 +_,19 @@
+          Predicate<ItemStack> predicate = ((ShootableItem)p_213356_1_.func_77973_b()).func_220006_d();
+          ItemStack itemstack = ShootableItem.func_220005_a(this, predicate);
+          if (!itemstack.func_190926_b()) {
+-            return itemstack;
++            return net.minecraftforge.common.ForgeHooks.findProjectile(this, p_213356_1_, itemstack);
+          } else {
+             predicate = ((ShootableItem)p_213356_1_.func_77973_b()).func_220004_b();
+ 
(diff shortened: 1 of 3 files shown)


src/main/java/net/minecraftforge/event/entity/player/PlayerFindProjectileEvent.java
```diff
@@ -0,0 +1,87 @@
...
+ * This event is {@link net.minecraftforge.eventbus.api.Cancelable}. If the event is canceled, in survival mode
+ * no projectile will be found and in creative mode a plain arrow will be returned. This equals vanilla behaviour
+ * when no arrow is in the players inventory.
+ * <br>
+ * This event does not have a result. {@link HasResult}<br>
+ * <br>
+ * This event is fired on the {@link MinecraftForge#EVENT_BUS}.
+ */
+@Cancelable
+public class PlayerFindProjectileEvent extends Event
```
mrp-v2: Why doesn't this extend `PlayerEvent`?

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1068,6 +1070,19 @@ public static void onAdvancement(ServerPlayerEntity player, Advancement advancem
...
+    /**
+     * Hook to fire {@link PlayerFindProjectileEvent}. Returns the ammo to be used.
+     */
+    public static ItemStack findProjectile(PlayerEntity player, ItemStack shootable, ItemStack ammo)
+    {
+        PlayerFindProjectileEvent event = new PlayerFindProjectileEvent(player, shootable, ammo);
+        if (MinecraftForge.EVENT_BUS.post(event)) {
+            return player.isCreative() ? new ItemStack(Items.ARROW) : ItemStack.EMPTY;
+        } else {
+            return event.getProjectile();
```
MWall541, line 1086: The code style needs to stay consistent and match the style with the rest of the API (check your braces here)

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
```diff
@@ -329,6 +329,29 @@
...
+                ItemStack itemstack1 = this.field_71071_by.func_70301_a(i);
+                if (predicate.test(itemstack1)) {
+-                  return itemstack1;
++                  return net.minecraftforge.common.ForgeHooks.findProjectile(this, p_213356_1_, itemstack1);
+                }
+             }
+ 
+-            return this.field_71075_bZ.field_75098_d ? new ItemStack(Items.field_151032_g) : ItemStack.field_190927_a;
++            return net.minecraftforge.common.ForgeHooks.findProjectile(this, p_213356_1_,
++                    this.field_71075_bZ.field_75098_d ? new ItemStack(Items.field_151032_g) : ItemStack.field_190927_a);
```
MWall541: Shouldn't this return statement be all on one line?

MWall541 (commented):
Fix up these small issues and you'll be good from what I see. Other things I gave thought to as to why it should be cancellable (couldn't you just return what the original item was gonna be?) etc. but those were design changes and I'm not sure if you had them for a different reason.

TheCurle:
The ArrowNockEvent exists for this purpose. What does this solve that Nock doesn't?

noeppi-noeppi:
1. It also works for crossbows and any other projectile item added by a mod that uses `PlayerEntity#findProjectile`
2. I have not found a way in `ArrowNockEvent` to set the ammo. The only thing I get there is the `hasAmmo` parameter. I can set an `ActionResult<ItemStack>` but this does not set the ammo.
It would be possible to return a `CONSUME` action result from there in which case it would be possible to draw the bow but when the arrow is about to be shot, the projectile stack will be empty and I only get an `ArrowLooseEvent` with no ammo. So what I would have to do is copy the 49 lines of code that summon the arrow from `BowItem` into my event handler to manually spawn the arrow and apply all the enchantments and such.
3. If I went the way to manually summon the arrow in an event handler for `ArrowLooseEvent`, and there was another mod doing the same, there would either be 2 arrows summoned if I don't cancel the event. Because then the other mod will receive the event as well and summon it's arrow. If I cancel the event there will only be one arrow but the play won't be able to stop using the bow. This event however will always only find one ItemStack as projectile and if multiple mods add their own ways for finding projectiles there will always be only one projectile that is shot.

MelanX:
I could really need it. Would be great if one of the triage team could review it when there is enough time.

sciwhiz12:
Hello! @noeppi-noeppi, please leave a comment whether you wish to continue this PR for 1.16. Note that there is already an updated version of the feature intended by this PR at #8322, so it is your decision whether to discontinue updating this PR in favor of that or to continue with your own implementation. (We suggest the former so there is a concentration of effort in one PR rather than fragmented across multiple.) If we do not hear back from you after some time, this PR will be closed for inactivity.

noeppi-noeppi:
I'll close this in favour of https://github.com/MinecraftForge/MinecraftForge/pull/8322 then.

Issue #7714, opened 2021-04-10
getDamageSource() can return null in LootingLevelEvent due to global loot modifiers with RandomChanceWithLooting condition


vemerion:
**Minecraft Version:** 1.16.5

**Forge Version:** 36.1.0

**Logs:** Here is an issue with a crash report: https://github.com/Beethoven92/BetterEndForge/issues/96#issuecomment-815442034

**Steps to Reproduce:**
 1. Create a global loot modifier that uses the RandomChanceWithLooting loot condition.
 2. Subscribe to the LootingLevelEvent.
 3. Print getDamageSource() in the LootingLevelEvent.
 4. Destroy a block in-game to see that getDamageSource() returns null.

**Description of issue:**
It turns out that there is a corner case with using the RandomChanceWithLooting LootCondition with global loot modifiers, since then the DamageSource will be null if there is no damage done (and this might happen because every loot modifier is run after every block destroyed etc). Normally this is not a problem, but if another mod subscribes to the LootingLevelEvent and assumes that the DamageSource is not null, there will be a null pointer exception. An example of a mod where this happens is the Unique Enchantments mod, on the exact source code line [here](https://github.com/TinyModularThings/UniqueEnchantments/blob/aa225de8cbbb2192b22c086adbbaf16df7ba5042/src/main/java/uniquee/handler/EntityEvents.java#L918).

I am unsure of exactly what the best fix for this would be. Just adding a Nullable annotation would mean that it would still be an issue until the problematic mods update. Maybe getDamageSource() could return some kind of 'dummy' DamageSource instead of null?

diesieben07:
> but if another mod subscribes to the LootingLevelEvent and assumes that the DamageSource is not null, there will be a null pointer exception

Then it is a bug in said mod. If there isn't a DamageSource available, we can't exactly make one up.
However we might not want to fire the looting level event at all if there isn't a killer entity, because looting without killing an entity doesn't really make sense. That still would not mean there is always a DamageSource necessarily though.

vemerion:
> Then it is a bug in said mod.

I agree with you that the fault lies partly on the mod makers, but I have stumbled upon at least two mods that assume it is never null, so it seems like a somewhat common misconception. I feel like at least a Nullable annotation and maybe a comment in LootingLevelEvent would be a good thing.

> However we might not want to fire the looting level event at all if there isn't a killer entity, because looting without killing an entity doesn't really make sense.

[...]

PR #7717, opened 2021-04-11
Only fire LootingLevelEvent if there is a killer entity

vemerion:
This fix makes it so that the LootingLevelEvent only fires if there is a
killer entity, since looting without a killer makes little sense. Also
adds the Nullable annotation to LootingLevelEvent.getDamageSource() to
more clearly communicate to modders that the DamageSource can be null in
LootingLevelEvent.

See issue #7714 for a more in depth discussion about the origin of this
pull request.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -390,7 +390,7 @@ public class ForgeHooks
         int looting = 0;
         if (killer instanceof LivingEntity)
             looting = EnchantmentHelper.getMobLooting((LivingEntity)killer);
-        if (target instanceof LivingEntity)
+        if (target instanceof LivingEntity && killer != null)
             looting = getLootingLevel((LivingEntity)target, cause, looting);
         return looting;
     }
--- a/src/main/java/net/minecraftforge/event/entity/living/LootingLevelEvent.java
+++ b/src/main/java/net/minecraftforge/event/entity/living/LootingLevelEvent.java
@@ -19,6 +19,8 @@
 
 package net.minecraftforge.event.entity.living;
 
+import javax.annotation.Nullable;
+
 import net.minecraft.entity.LivingEntity;
 import net.minecraft.util.DamageSource;
 
@@ -34,6 +36,7 @@ public class LootingLevelEvent extends LivingEvent {
         this.lootingLevel = lootingLevel;
     }
 
+    @Nullable
     public DamageSource getDamageSource() {
         return damageSource;
     }


TheCurle:
Fixes crashes, alerts modders that something can potentially be wrong. What's not to love?
I'd like for there to be a small comment here mentioning that the edge case exists, but that's not a necessity.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -390,7 +390,7 @@ public static int getLootingLevel(Entity target, @Nullable Entity killer, Damage
         int looting = 0;
         if (killer instanceof LivingEntity)
             looting = EnchantmentHelper.getMobLooting((LivingEntity)killer);
-        if (target instanceof LivingEntity)
+        if (target instanceof LivingEntity && killer != null)
```
LexManos, line 393: The killer can be explicitly null. Firing the looting level event is explicitly intended to allow handeling when the killer is null so things can have a base looting level, or damage sources can have a looting level.

src/main/java/net/minecraftforge/event/entity/living/LootingLevelEvent.java
```diff
@@ -34,6 +36,7 @@ public LootingLevelEvent(LivingEntity entity, DamageSource damageSource, int loo
         this.lootingLevel = lootingLevel;
     }
 
+    @Nullable
     public DamageSource getDamageSource() {
```
LexManos, line 40: This 1) 1/2 does a @Nullable you'd need it on the field, and constructor argument
Edit: Reading further into the context, this being nullable seems like it is needed then. Considering damage sources in loot tables can be null. So this needs a proper @Nullable all the way up the stack, not just the getter.

src/main/java/net/minecraftforge/common/ForgeHooks.java
vemerion: Thank you for the review! So you think the predicate should be removed again? My reason for adding it was that vanilla requires the KILLER_ENTITY loot parameter for all looting related things, so it seemed reasonable to assume that looting is only relevant if there is a killer. But of course that is debatable, as you pointed out.

src/main/java/net/minecraftforge/event/entity/living/LootingLevelEvent.java
vemerion: I agree with you that it would have been nicer if the damage source was never null. The original reason for why I created this pull request was that several people complained that my mod was crashing, when in fact it was other mods causing NPE because of getDamageSource() being null. I'll add the nullable annotation all the way up the chain then!

PR #7718, opened 2021-04-13
Fix shears not properly breaking tripwire (MC-129055)

KnightMiner:
Shears when breaking tripwire are supposed to deactivate the tripwire to prevent tripwire hooks from triggering. However, they randomly do not properly breaking the tripwire, instead immediately replacing after breaking ([MC-129055](https://bugs.mojang.com/browse/MC-129055))

Apparently Mojang had the condition to fix this problem, but the problematic code was not inside the condition. Essentially part of the tripwire update logic iterates all relevant tripwire blocks in the area, and attaches them if relevant.

----

As far as replicating the bug, I typically place tripwire hooks at ground level and connect them with string. Then I just break and replace a piece of tripwire a few times until its triggered. I get the issue a bit less than 50% of the time. One you are confident that the issue is in fact happening, rebuild with this patch and you shouldn't see the bug happening.

### DIFF
--- a/patches/minecraft/net/minecraft/block/TripWireHookBlock.java.patch
+++ b/patches/minecraft/net/minecraft/block/TripWireHookBlock.java.patch
@@ -1,0 +1,12 @@
+--- a/net/minecraft/block/TripWireHookBlock.java
++++ b/net/minecraft/block/TripWireHookBlock.java
+@@ -150,8 +150,8 @@
+             BlockPos blockpos2 = p_176260_2_.func_177967_a(direction, k);
+             BlockState blockstate2 = ablockstate[k];
+             if (blockstate2 != null) {
++               if (!p_176260_1_.func_180495_p(blockpos2).func_196958_f()) { // FORGE: fix MC-129055
+                p_176260_1_.func_180501_a(blockpos2, blockstate2.func_206870_a(field_176265_M, Boolean.valueOf(flag2)), 3);
+-               if (!p_176260_1_.func_180495_p(blockpos2).func_196958_f()) {
+                }
+             }
+          }


TheCurle:
Did they seriously put the if below the offending line? I find that hard to believe. 
This fix works, i'm happy to assign it.

KnightMiner:
I honestly have no idea why it showed up there. I at first assumed failed forge patch, but there was no patch. More likely mojang did a git merge or something and it merged improperly

Ynaught:
Please undo this. It breaks parody with vanilla, which is the job for individual mods and not the responsibility of the mod-loader itself.

This specific glitch is what allows for string dupers to work. They work in vanilla, on fabric, pretty much everywhere else, just not here. . . because of a bug fix was made here that isn't in the base-game (which is anti-parody). If users want a vanilla bug patched, then they should be looking for individual mods to patch it.

KnightMiner:
> Please undo this. It breaks parody with vanilla, which is the job for individual mods and not the responsibility of the mod-loader itself.

This is a vanilla bug, and Forge often fixes vanilla bugs. Especially when they break mods. But in this case it also breaks vanilla because tripwire disarming just randomly does not work.

> This specific glitch is what allows for string dupers to work. They work in vanilla, on fabric, pretty much everywhere else that's java-based, just not here. . . because of a bug fix was made here that isn't in the base-game (which is anti-parody). If users or developers want a vanilla bug-exploit patched, then they should be looking for individual mods to patch it.

So you are telling me this behavior that breaks mods should be kept because it also breaks vanilla? thats the dumbest argument I ever heard.

Dupe bugs are not gameplay. They are exploits used by technical players. You play mods, go download an intentional string dupe mod like a cotton plant.

> No developer has reported this issue to break their mod. This report was done just for the sake of reporting it. Meaning the fix was implemented for no other reason other than to break parody out of a potentiality that never came to be at the dissatisfaction of many users.

This issue broke my mod. That why I PRed a fix for it. AOE shears are completely impossible without this fix.

This issue also broke vanilla, 50% of the time tripwire disarming just does not work, and when it does it leaves you with a nearly unbreakable block. This is a stupid thing to try to keep around. Its clearly broken as the behavior on the bedrock edition shows.

Ynaught:
> This issue broke my mod.

Apologies then, I retract my statement. Though, I would still argue that parody is still parody, which includes the originals flaws regardless of how we feel about said flaws. But if parody must be sacrificed because of issues preventing mods to opperate, then that much cannot be helped and is completely understandable for that context exclusively.

I am curious though, does the previously claimed mod not have a fabric port (even an unofficial one)? I ask because this bug is not patched on fabric and it's useful application is still present on that loader and still within parody with vanilla. If it can work perfectly on fabric, then the presumption is that it is not a fault of the vanilla bug since such a patch is not present in that environment.

KnightMiner:
> Apologies then, I retract my statement. Though, I would still argue that parody is still parody, which includes the originals flaws regardless of how we feel about said flaws. But if parody must be sacrificed because of issues preventing mods to opperate, then that much cannot be helped and is completely understandable for that context exclusively.

Again, forge fixes many bugs in vanilla. This is a bug, not a feature. It is really, really, really broken behavior that is really, really clearly broken if you spend any time interacting with it. Ignore string dupers, those are a buggy exploit that is not relevant here. Just try breaking tripwire with shears. Then try removing the wire it left behind. Do this 10-15 times and then tell me that behavior should be kept.

Compare it to the same thing with Forge or on Minecraft Bedrock and then tell me the vanilla buggy behavior is intended.

> I am curious though, does the previously claimed mod not have a fabric port (even an unofficial one)? I ask because this bug is not patched on fabric and it's useful application is still present on that loader and still within parody with vanilla. If it can work perfectly on fabric, then the presumption is that it is not a fault of the vanilla bug since such a patch is not present in that environment.

This is a silly argument. Not every mod has a fabric port. And even if it did, why does it matter? Fabric can choose to keep around vanilla bugs that break mods without forge having to do so. This bug has been fixed in forge for over 3 years and this year is the first time people started to complain. ***This is a bug***, and its really, really breaks the proper interaction between tripwire and shears in vanilla. Forge agrees its a bug, NeoForge agrees its a bug, Minecraft Bedrock agrees its a bug, even vanilla Java agrees its a bug as they never closed this issue (they just never fixed it either). Fabric is the one that lacks parity here.

That said, my mod does have a fabric port. If you just actually read this issue and the other issues that link it you would be able to find this out pretty easily. This issue either exists on the unofficial Fabric port or they use a mixin to prevent it, you can talk to that dev.

PR #7722, opened 2021-04-19
Fix cats, phantoms and pillagers saying they're at BlockPos.ZERO during LivingSpawnEvent.CheckSpawn

ShroomAgent27:
# The Problem

Phantoms, Cats, and Pillagers from a Pillager Patrol do not set their positions until after `ForgeHooks.canEntitySpawn` is called.

# The Cause

Let's take a look at the phantom spawner.

As a result of my Forge PR, `net.minecraft.world.spawner.PhantomSpawner`,

the relevant method section look like this (lines 59 to 63 shown).

```java
PhantomEntity phantomentity = EntityType.PHANTOM.create(p_230253_1_);
if(net.minecraftforge.common.ForgeHooks.canEntitySpawn(phantomentity, p_230253_1_, blockpos1.getX(), blockpos1.getY(), blockpos1.getZ(), null, SpawnReason.NATURAL) == -1) return 0;
phantomentity.moveTo(blockpos1, 0.0F, 0.0F);
ilivingentitydata = phantomentity.finalizeSpawn(p_230253_1_, difficultyinstance, SpawnReason.NATURAL, ilivingentitydata, (CompoundNBT)null);
p_230253_1_.addFreshEntityWithPassengers(phantomentity);
```

First, the phantom is created. This sets the default location to `BlockPos.ZERO`.

This happens in `net.minecraft.entity.Entity` on line 205.

Second, the ForgeHook to fire the `LivingSpawnEvent.CheckSpawn` event.

Third, the phantom is moved to the desired `blockpos1`.

The relevant method section looks like this (lines 199 to 207 shown).

```java
public Entity(EntityType<?> p_i48580_1_, World p_i48580_2_) {
  super(Entity.class);
  this.type = p_i48580_1_;
  this.level = p_i48580_2_;
  this.dimensions = p_i48580_1_.getDimensions();
  this.position = Vector3d.ZERO;
  this.blockPosition = BlockPos.ZERO;
  this.packetCoordinates = Vector3d.ZERO;
  this.setPos(0.0D, 0.0D, 0.0D);
```

`BlockPos.ZERO` is just `new BlockPos(0, 0, 0)`. 

Finally, the phantom's spawn is finalized.

Note that the event is fired before the phantom is moved to `blockpos`!

I should have correctly made sure that the event was fired after the entity's position was moved to the correct location.

However I failed to do so because I didn't think it would cause that big of a deal at the time. However, mods only really check the Entity's location, which is set to `BlockPos.ZERO`.

This value is technically correct at the time, as the entity hasn't been moved yet, but extremely unhelpful, since it doesn't show where the entity is going to be.

# The Solution

The solution is to just make sure that the event is always fired after the entity is moved to its correct location.

This is exactly what this patch does.

I do understand that this is an extremely minor patch, but I wanted to open it anyway to make awareness of the issue that I introduced.

### DIFF
--- a/patches/minecraft/net/minecraft/world/spawner/CatSpawner.java.patch
+++ b/patches/minecraft/net/minecraft/world/spawner/CatSpawner.java.patch
@@ -4,10 +4,10 @@
        if (catentity == null) {
           return 0;
        } else {
--         catentity.func_213386_a(p_221122_2_, p_221122_2_.func_175649_E(p_221122_1_), SpawnReason.NATURAL, (ILivingEntityData)null, (CompoundNBT)null);
-          catentity.func_174828_a(p_221122_1_, 0.0F, 0.0F);
++         catentity.func_174828_a(p_221122_1_, 0.0F, 0.0F); // Fix MC-147659: Some witch huts spawn the incorrect cat
 +         if(net.minecraftforge.common.ForgeHooks.canEntitySpawn(catentity, p_221122_2_, p_221122_1_.func_177958_n(), p_221122_1_.func_177956_o(), p_221122_1_.func_177952_p(), null, SpawnReason.NATURAL) == -1) return 0;
-+         catentity.func_213386_a(p_221122_2_, p_221122_2_.func_175649_E(p_221122_1_), SpawnReason.NATURAL, (ILivingEntityData)null, (CompoundNBT)null);
+          catentity.func_213386_a(p_221122_2_, p_221122_2_.func_175649_E(p_221122_1_), SpawnReason.NATURAL, (ILivingEntityData)null, (CompoundNBT)null);
+-         catentity.func_174828_a(p_221122_1_, 0.0F, 0.0F);
           p_221122_2_.func_242417_l(catentity);
           return 1;
        }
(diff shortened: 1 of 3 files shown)


LexManos:
Interesting my first thought was not to care considering the event has the real location passed in just fine.
However, looking deeper into the car code, it looks like you may of uncovered the cause for a vanilla bug.
If you look into the cat's finalizeSpawn, it checks the current  position for a witch's hut to set the cat type.
If i'm reading this right the position will always be zero and thus that code never runs.
This should be filed as a bug with Mojang and the associated JIRA linked in the code.

sciwhiz12:
It seems that the 'cats may spawn as non-black in witch huts' bug has been filed as [MC-147659](https://bugs.mojang.com/browse/MC-147659) already, all the way from 1.14. A commenter has proposed the same fix done here: moving the `CatEntity#finalizeSpawn` call to be after the position of the entity is set.

From my and @ChampionAsh5357's musing of the code, it seems that because this bug is only present in `CatSpawner`, this only affects when a new cat is spawned within a witch's hut; the cat that spawns with the hut on world generation is handled by `SwampHutPiece`, which does set the position before the `#finalizeSpawn` call.

PR #7723, opened 2021-04-20
Update for Forge's new maven

AterAnimAvis:
Didn't touch the promotions_slim in `mods.toml` as it doesn't seem to be present at the new maven url


LexManos:
promotions_slim.json, and other metadata files that our PageGen repo creates are under / now instead of /maven.

Issue #7658, opened 2021-02-12
Furnace minecart speed increased when using Forge


addmix:
**Minecraft Version:** {Minecraft version}
1.16.4
**Forge Version:** {Forge version. *Version number, not latest/rb*}
35.1.5
**Steps to Reproduce:**
Compare furnace minecart speed in vanilla to forge.

**Description of issue:**
furnace minecart runs much faster when using forge, compared to vanilla.

JoJoDeveloping:
I can not reproduce this. I ran the powered minecart on an identical loop in vanilla Minecraft, forge dev and forge-dev-clean (which is decompiled but without forge's patches). This is the loop (12x6):
![2021-04-30_00 58 52](https://user-images.githubusercontent.com/7198985/116628402-5647f900-a94f-11eb-935f-1c7949754d85.png)

The minecarts took 38.41s in vanilla, 38.57 in forge, and 38.51 in forge-clean.

I did use Minecraft 1.16.5 and Forge latest, though.

Cyborgmas:
Perhaps it is becaused the linked PR was merged but the issue was not closed :p

TheCurle:
Indeed. This should have been closed ages ago.

PR #7725, opened 2021-04-22
Fix for increased Furnace minecart speed issue #7658

atkarimarton:
This PR overrides the FurnaceMinecartEntity getMaxCartSpeedOnRail method because the cart was too fast compared to the vanilla version like it was mentioned in issue: https://github.com/MinecraftForge/MinecraftForge/issues/7658

Here's a gif for comparison:

Tumc


LexManos:
Minecart hooks really need to be removed.
We need to figure out how many people actually use this invasive patches.

PR #7733, opened 2021-04-26
Fix FakePlayers having null `connection` and add FakePlayer test mod

### LINKED ISSUE #7604: FakePlayer's "connection" field
Maybe it's just me who's experienced this but whenever a modder does something like this:

```
if (player instanceof ServerPlayerEntity) {
    ((ServerPlayerEntity) player).connection.sendPacket(packet);
}
```

They are risking a null pointer exception by not checking if the ServerPlayerEntity is actually a FakePlayer, which has the `connection` field set to null.

I have solved this in my own mod by creating a `SafeFakePlayer` which has the connection field set to a `FakePlayNetHandler extends ServerPlayNetHandler` where all the methods are overridden to do nothing. 

I feel like something like this should be part of the default FakePlayer, because the `.connection.sendPacket()` syntax is heavily used by vanilla and has the potential to cause lots of crashes if the wrong path gets executed with a FakePlayer.

williambl:
This PR adds a `FakePlayerNetHandler`, a dummy net handler which stands in for the player's `ServerPlayNetHandler`. It gives the `FakePlayer` a `FakePlayerNetHandler` for its `connection` field, thus solving #7604.

It also adds a fake player test mod, which adds commands to test fake player functionality.

Closes #7604.


nikita488:
I saw that some mods (like Mekanism and ModularRouters) also have a fake NetworkManager that overrides all methods to do nothing. I'm not sure if this really needed or not, but can be considered.
[Mekanism](https://github.com/mekanism/Mekanism/blob/b0a96e1eb7b5f5741cb1ca2f4ce656084ae923d5/src/main/java/mekanism/common/base/MekFakeNetHandler.java#L290)
[ModularRouters](https://github.com/desht/ModularRouters/blob/e001dea43ccfaee477fd82002f47e03bfe7bb113/src/main/java/me/desht/modularrouters/util/fake_player/FakeNetHandlerPlayerServer.java#L22)

TheCurle:
If you fix the missing license on FakePlayerNetHandler, and possibly fix the import scramble in FakePlayer, this should be good to go. My initial testing seems to indicate it works.

pupnewfster:
For a bit of reference on why Mekanism has a fake NetworkManager as well:
1. I didn't feel like figuring out if we need one or if the normal one *could* potentially work fine.
2. The connection field in the ServerPlayNetHandler so I wanted to guard against any mods potentially calling that and then interacting with the network manager directly in such a way that would NO-OP the methods (and also ensure things like isConnected return false)

Not sure if it is really needed or not for this PR though. Either way one thing I can see Lex potentially not fully liking in regards to this PR is the maintenance headache of each MC version seeing if there are any more methods that have to be NO-OP'd, so I am sort of wondering how manageable it would be to make another coremod to basically just NO-OP all the methods in a class, and then pass it the FakePlayerNetHandler as then in theory there shouldn't be any extra maintenance cost between MC versions as the coremod would catch any new methods and just NO-OP them.

williambl:
> If you fix the missing license on FakePlayerNetHandler, and possibly fix the import scramble in FakePlayer, this should be good to go. My initial testing seems to indicate it works.

Done.

LexManos:
The FakePlayerNetHandler needs to be a private inner class so that it isn't exposed as API to modders.
Also, not a real issue, but parameter names in non-mc code shouldn't be using srg names.

williambl:
Done.

PR #7740, opened 2021-04-30
Add an event similar to PlayerEvent.NameFormat but for the name shown in the tab list.

noeppi-noeppi:


patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
```diff
@@ -150,6 +150,20 @@
     }
  
     public ChatVisibility func_147096_v() {
+@@ -1276,7 +_,12 @@
+ 
+    @Nullable
+    public ITextComponent func_175396_E() {
+-      return null;
++      if (!this.hastTablistName)
```
josephtarbit: Can't you just do a null check?

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -180,6 +182,43 @@ public void setDisplayname(ITextComponent displayname)
...
+     * This event is not {@link net.minecraftforge.eventbus.api.Cancelable}.
+     * <br>
+     * This event does not have a result. {@link HasResult}
+     * <br>
+     * This event is fired on the {@link MinecraftForge#EVENT_BUS}.
+     **/
+    public static class TabListNameFormat extends PlayerEvent
+    {
+        @Nullable
+        private ITextComponent displayname;
```
josephtarbit: Use camel case, e.g. `displayName` (applies to other areas in this class too)

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
```diff
@@ -206,6 +220,16 @@
 +    */
 +   public String getLanguage() {
 +      return this.language;
++   }
++
++   // We need this as tablistDisplayname may be null even if the the event was fired.
++   private boolean hastTablistName = false;
++   private ITextComponent tablistDisplayname = null;
```
josephtarbit: Use camel case, e.g `tabListDisplayName`

josephtarbit (changes requested):

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
noeppi-noeppi: I copied that name from the `NameFormat` event but I can change this

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
noeppi-noeppi: No. As explained [here](https://github.com/noeppi-noeppi/MinecraftForge/blob/feature/tablist_name/patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch#L225) this field is null when the client should determine the tab list name itself by checking the scoreboard team. So this variable will be null in most cases and cacheing it would no longer make sense as the event would be posted every time the method is called.

josephtarbit: Ahh yes, moment of madness, I understand now.

diesieben07 (changes requested):
This is missing a way (maybe automatic, maybe manually triggered) to resend the packet to update the name when the name changes.

noeppi-noeppi:
Would this fit into the `refreshTabListName` method or would it be better to create a new method for this?

diesieben07:
Yes you could send it at the end of that method after the event has fired, if the value has changed.

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
```diff
@@ -193,10 +207,11 @@
           this.field_70170_p.func_217376_c(itementity);
           ItemStack itemstack = itementity.func_92059_d();
           if (p_146097_3_) {
-@@ -1403,6 +_,15 @@
+@@ -1402,6 +_,29 @@
+          }
  
           return itementity;
-       }
++      }
```
diesieben07, line 213: What is this change?

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -180,6 +182,43 @@ public void setDisplayname(ITextComponent displayname)
...
     }
 
+    /**
+     * TabListNameFormat is fired when a player's display name for the tablist is retrieved.<br>
+     * This event is fired whenever a player's display name for the tablist is retrieved in
+     * {@link ServerPlayerEntity#getTabListDisplayName()} or {@link ServerPlayerEntity#refreshTabListName()}.<br>
+     * <br>
+     * This event is fired via the {@link ForgeEventFactory#getPlayerTabListDisplayName(PlayerEntity)}.<br>
+     * <br>
+     * {@link #displayName} contains the display name of the player or null if the client should determine the display name itself.
```
diesieben07: Should use the getter name here, as the field is private.

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
```diff
@@ -150,6 +150,20 @@
     }
  
     public ChatVisibility func_147096_v() {
+@@ -1276,7 +_,12 @@
+ 
+    @Nullable
+    public ITextComponent func_175396_E() {
+-      return null;
++      if (!this.hasTabListName)
++      {
```
diesieben07: Formatting does not match the normal formatting in vanilla files.

diesieben07 (changes requested):

patches/minecraft/net/minecraft/entity/player/ServerPlayerEntity.java.patch
noeppi-noeppi: The `genPatches` task generated it like this. It thought I added the bracket here and interpreted the bracket at the end of the `refreshTabListName` method as if it has been there before. But I don't think this should cause any problems.

TheCurle: There's a whitespace difference here. There should be a space after the bracket.

noeppi-noeppi: Sorry but I don't seem to be getting the point. When I add a space after that bracket there will just be an additional pointless space in the patch file that was not there before and the change will still be there.

TheCurle: Your change is removing the space. That's why this line in the patch exists.
image

noeppi-noeppi: Isn't there an extra space in patch files for unmodified lines?

TheCurle: If a line is unmodified it won't show up as a modified line..

noeppi-noeppi: image
When I add a spce in the patch file and run `setup` `ServerPlayerEntity.java` looks like this. As you can see here, the bracket is offset by one to the vertical line. So I can change this but for me it gives wrong formatting.

TheCurle: Uhhh, super brainfart. ignore me, i completely missed the + +.
Diffs of diffs, man. Can never keep them straight in your mind

diesieben07: Yeah, apologize my ineptitude. This is fine.

diesieben07 (approved):

PR #7741, opened 2021-04-30
[1.16.x] Add additional context to block side render check

XFactHD:
This PR adds additional context to the block side render check by introducing an alternate method to `AbstractBlock#skipRendering()` in `IForgeBlock`.
This is useful for mods adding blocks that can change their appearance and need data stored in the `TileEntity` to decide if a side should be hidden or not.

### DIFF
--- a/patches/minecraft/net/minecraft/block/AbstractBlock.java.patch
+++ b/patches/minecraft/net/minecraft/block/AbstractBlock.java.patch
@@ -19,6 +19,14 @@
     }
  
     @Deprecated
+@@ -112,6 +_,7 @@
+       return p_196271_1_;
+    }
+ 
++   /** Override {@link net.minecraftforge.common.extensions.IForgeBlock#skipRendering(IBlockReader, BlockPos, BlockState, BlockState, Direction)} instead */
+    @Deprecated
+    @OnlyIn(Dist.CLIENT)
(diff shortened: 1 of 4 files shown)


LexManos:
This needs more context, yes the general concept of 'i need my tile entity!' is a thing. However Minecraft, and Forge are heading more twards block states being the defining things for rendering, physics, etc... Soo I'm reluctant to ADD more tile entity reliance on things without actual use cases or actual justifications.

XFactHD:
My use case for this is very simple: I have a block that can take the appearance of any "basic" block. That "camo" data is stored in the tile entity and needs to compared to the neighboring block (or the neighboring blocks "camo" data if that block is also my special block) in order to decide if the given side should be hidden.

The part of my current implementation that would use this is [here](https://github.com/XFactHD/FramedBlocks/blob/1.16.x/src/main/java/xfacthd/framedblocks/common/util/SideSkipPredicate.java#L49-L58) which is called from [here](https://github.com/XFactHD/FramedBlocks/blob/1.16.x/src/main/java/xfacthd/framedblocks/common/block/FramedBlock.java#L87) through an [abstraction layer](https://github.com/XFactHD/FramedBlocks/blob/1.16.x/src/main/java/xfacthd/framedblocks/common/block/IFramedBlock.java#L145-L150).

XFactHD:
I am closing this because a superior solution that doesn't require any changes to vanilla code has been worked out in a conversation on the Forge Discord.

PR #7743, opened 2021-05-01
Add a small event to retrieve the dynamic registries as soon as they're ready.

TheCurle:
Full credit to Commoble for locating the proper place to inject this, and for providing the base of the test mod.

The new DynamicRegistriesLoadedEvent (name pending rethink) serves a similar purpose to the BiomeLoadingEvent.  
The main difference is that it allows direct access to the baked representation of JSON datapack objects.  
It provides the DynamicRegistries.Impl object that contains all seeded objects from the Forge registries, and all JSON objects from datapacks, rather than just the Builder for a specific biome with no other objects available.

DynamicRegistries contains baked objects, which means that directly editing any object involved requires some reflection or AT, but having one centralized place to acquire the full list of these objects is more conducive to having these improvements in future..

Ideally, we would pass the Dynamic Registries impl to the BiomeLoadingEvent to get the best of both worlds, but the timings of both events would have to be drastically altered in order to achieve that.

Whereas before, in order to add a ConfiguredFeature to a vanilla biome, you had to either:  
a) create and register the feature in code, add it to a biome via the BiomeLoadingEvent, or
b) create the feature in JSON, add it to a biome by overwriting the biome with json,  
this PR allows mods to create the feature in JSON and add it to a biome's world generation settings directly, which makes customization and compatibility automatic due to the DynReg containing only the ultimate object.  
(pending another PR to make that easier)

To reiterate, this PR allows mods to both a) use JSON worldgen customization, and b) allow _other mods_ (or just ordinary user datapacks) to change those JSON worldgen customizations. Both A and B are not possible with BiomeLoadingEvent, because it fires after the DynReg is discarded.

A small test mod is included that covers vanilla deserts in gold blocks using this feature:
image

As compatibility between worldgen mods is currently shaky, I plan to test this in various places before opening this PR for full scrutiny.

Tests to run: 
☑Datapack feature added to code biome  
☑Datapack feature overwriting BiomeLoadingEvent feature  
 ☐ Datapack feature overwriting other datapack feature between mods


Commoble:
Since this fires after most datapack stuff is loaded, this would also enable the existence of a data loader for tag-like things for biomes and e.g. allow mods to add features to all biomes in a given tag, which is something that's been attempted before and would help worldgen mods do their things in a modpack-friendlier manner than the clunky defining-which-biomes-features-should-be-added-to-in-config-files pattern that's typically done.

50ap5ud5:
This is amazing!

One thing to note is that this PR does not cover certain use cases such as modifying structure spacing settings for a dimension. In this case we need to access the dimension registry, then access its chunk generator, then access its structure spacing settings.

The problem here is that the Dimension registry is not available after this event fires (after dynamic registries have loaded). Mojang seems to handle dimensions differently from other data driven registry so this is likely why it's such a special snowflake. 

For special cases like inner dimension settings, seperate hooks may be needed.

TheCurle:
Dimensions are indeed a whole other beast.

I quickly looked into extending the life cycle of that registry to encompass the firing of this event, but it's way too much work and very invasive for not a very consistent result.

Another event is probably the best way to go about this.

Won-Ton:
> One thing to note is that this PR does not cover certain use cases such as modifying structure spacing settings for a dimension. In this case we need to access the dimension registry, then access its chunk generator, then access its structure spacing settings.

There is a registry for DimensionSettings in the DynamicRegistries so you'd be able to add/remove/replace structure settings dynamically with some reflection or w/e (don't love it but gets the job done).

Doesn't help the situation where two different datapacks add different settings in overworld's dim-settings but in an ideal world mods would stop doing it that way and use something like this.

TheCurle:
Yes, two mods changing vanilla datapacks is an unfortunate edge case, and the vanilla systems are really not enough for this. 

That was the main impetus of making this event.

scrouthtv:
is there any way for me to test this patchset before it is merged?

TheCurle:
Clone my fork.

Commoble:
Some relevant observations on the loading of worldgen data that should be mentioned:

A) This event fires whenever minecraft loads data from jsons into the dynamic registries (overwriting any of the builtin vanilla biomes, features, etc), regardless of why the dynamic registries were loaded. If they get loaded, the event fires.

B) Creating singleplayer worlds causes the dynamic registries to be loaded twice (the data is loaded, thrown out, and then loaded again). Joining an existing singleplayer world causes the dynamic registries to be loaded four times (causing the event to fire each time). This is partially related to the experimental-worldgen-warning screen (pushing the "I'm okay with this" button causes the data to be loaded an additional time).

For dedicated server worlds, data is only loaded once at server startup in all cases, as far as I can tell.

As far as I can tell, **at no point is any data from a previous iteration fed into a subsequent event**, so e.g. adding features to biomes via the event *does not* cause the feature to be added twice -- Mojang is just loading data and throwing it out multiple times, for reasons I can't fathom.

I don't personally believe it's worth trying to account for this or optimizing the event to only fire when it's actually needed -- the goal of this event is to allow the worldgen data to be modified after all relevant data is loaded from jsons, to make modifications to worldgen more datapack-friendly; the event should fire after the data is loaded (whether the data ends up being used for anything or not) for maximum consistency.

Optimizing world-loading via forge patches to not load and throw out the worldgen data multiple times when singleplayer worlds are loaded could theoretically be done. **I don't believe such a thing would be within the scope of this PR**, but it could potentially be done in a separate PR at a later time if needed.

(Of course, since the extra loads are at least partially due to worldgen datapacks being "experimental", this could all end up being moot after another major minecraft update or two.)

Commoble:
Okay, further tests invalidate a lot of my previous comment.

If a vanilla biome is overridden via JSON, then each time the event fires, a fresh biome instance is created (so if somebody adds a  feature to the biome in the event and the event fires multiple times, the feature won't be added multiple times).

If a vanilla biome is NOT overridden via json, then each time the event fires, the *same* biome instance is used (meaning attempts to add features to the biome will cause those features to be added to the biome instance multiple times).

Looking into how this problem can be addressed now.

Commoble:
I looked into moving the patch into some of the places that *call* WorldSettingsImport#create, but recreate world does things in a manner that might make this whole concept infeasible without substantial patching. We might need to find a later spot to move the hook to, maybe somewhere in the server-starting lifecycle

Commoble:
As far as I can tell, just modifying biomes in FMLServerAboutToStartEvent seems to work quite nicely -- it only runs once per server start, it's early enough that syncable modifications to the server's dynamicregistries (e.g. biome colors) will be properly synced to clients that join later, and it's late enough that the vanilla biomes available are the redeserialized copies of the vanilla biomes (so starting multiple singleplayer servers in a single session won't cause the addition of a feature to a biome to happen multiple times).

If there's not some weird edge case I haven't encountered yet, the biome load event could be conceivably moved to the server-starting lifecycle in 1.17 to allow datapack-sensitive modifications to biomes as described earlier.

Won-Ton:
That wouldn't be suitable as a general DynamicRegistries load event as some of those registries (DimensionType/DimensionSetiings off the top of my head) will have already been consumed by dimensions and chunkgenerators, so modifications there would not be effective or reliable.

You should def test this idea for biome modification with modded chunk generators & biome-providers before fully committing as it relies on them deferring any interactions with the registry objects already passed to them. TerraForged should be fine but others may have (reasonably) already used the biomes and unpacked features/structures/carvers etc into whateveer data structure they need so may not 'see' the modifications.

Should also be careful about exposing the DynamicRegistries as that in turn exposes MutableRegistry#registerOrOverride which mods could use to entirely replace the biome instance which'd cause issues when the biome provider looks up the id of the now unregistered biome that they hold reference to (thinking SingleBiomeProvider would probably break if you want a vanilla example).

Appreciate you soldiering on with this. It _is_ needed for sure and a difficult one to find the right hook point for.

TheCurle:
That was my initial reservation with using the final DynamicRegistries field stored in the server thread - it is final, will not be read any more, and modifications are pointless.

I'm waiting for Commoble to get back to me to see if the ServerStarting event fires early enough to seed Dimension settings and the such. I personally doubt it, but if it is not, we must reopen this jar of worms.

PR #7744, opened 2021-05-01
Replace hardcoded check for dirt in AnimalEntity#checkAnimalSpawnRules.

### LINKED ISSUE #7685: Getting creatures mobs to spawn on custom blocks is difficult to do with mod compatibility in mind
In EntitySpawnPlacementRegistry, several vanilla mobs will register AnimalEntity's canAnimalSpawn method as their predicate for spawning. The issue is, this method is specifically hardcoded to vanilla's grass block:
`return worldIn.getBlockState(pos.down()).isIn(Blocks.GRASS_BLOCK) && worldIn.getLightSubtracted(pos, 0) > 8;` 

Furthermore, in WorldEntitySpawner's func_234966_a_ method (used for spawning mobs every tick), it calls func_234975_a_ which checks the registered spawn placement but does it before Forge's canEntitySpawn event can fire. Thus rendering the canEntitySpawn event useless to bypassing this hardcoded grass block check. 

The only workaround at the moment is to try gathering all creature mobs, get their spawn predicate, wrap it in your own that checks for your block and defers to the original behavior if your condition is false, and register that new spawn predicate to that mob. The issue is, other mods may also use this method for their mobs which may not always be a creature category. And checking if the original spawn is using canAnimalSpawn will be impossible to do if multiple mods wraps the predicate. 

A solution offers by someone was to replace canAnimalSpawn's `.isIn(Blocks.GRASS_BLOCK)` with a block tag for allowing mobs using that method to spawn on custom blocks in a more mod compatible way. `forge:creatures_can_spawn_on` or something along those lines might work as a name. Thoughts?

TheCurle:
Changes the check for Blocks.GRASS_BLOCK to a new Forge tag, creatures_can_spawn, which contains grass block by default.
Mods can add to this tag to make creatures spawnable on their custom blocks.

Closes #7685

### DIFF
--- a/patches/minecraft/net/minecraft/entity/passive/AnimalEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/passive/AnimalEntity.java.patch
@@ -1,5 +1,14 @@
 --- a/net/minecraft/entity/passive/AnimalEntity.java
 +++ b/net/minecraft/entity/passive/AnimalEntity.java
+@@ -98,7 +_,7 @@
+    }
+ 
+    public static boolean func_223316_b(EntityType<? extends AnimalEntity> p_223316_0_, IWorld p_223316_1_, SpawnReason p_223316_2_, BlockPos p_223316_3_, Random p_223316_4_) {
+-      return p_223316_1_.func_180495_p(p_223316_3_.func_177977_b()).func_203425_a(Blocks.field_196658_i) && p_223316_1_.func_226659_b_(p_223316_3_, 0) > 8;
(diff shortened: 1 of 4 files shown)


pupnewfster:
I don't have my IDE open currently but is this only used on the server if not it needs to check grass and then this so that it doesn't change functionality/desyncs when connecting to vanilla servers. Also it seems like the cache file for datagen didn't get updated (or does forge only commit that for the test module)?

sciwhiz12:
Hello! @TheCurle, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

sciwhiz12:

PR #7747, opened 2021-05-01
Comply with PlayerEvent.BreakSpeed's pos being nullable.

### LINKED ISSUE #7615: Inconsistent nullability of position for dig speed checks
**Minecraft Version:** 1.16.5

**Forge Version:** Code is present in 19f8d2a7937dc7968ecbef2f9785687193fcd210

**Description of issue:**
The position for the dig speed check is `@Nullable` here:
https://github.com/MinecraftForge/MinecraftForge/blob/19f8d2a7937dc7968ecbef2f9785687193fcd210/patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch#L100-L108

`getBreakSpeed` passes the (potentially `null`) position directly to the `BreakSpeed` event, which does not mark the position as `@Nullable` and instead specifies y=-1 for unknown locations
https://github.com/MinecraftForge/MinecraftForge/blob/19f8d2a7937dc7968ecbef2f9785687193fcd210/src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java#L106
`getBreakSpeed` should either replace `null` positions with something like `(0, -1, 0)` or the annotation and Javadoc on the event should be updated.

See also BluSunrize/ImmersiveEngineering#4587. @desht suggested that this might be something to put in #7612. @TheCurle, what do you think about that?

TheCurle:
As @malte0811 noted in #7615, the BreakSpeed event is fired with Nullable position:
https://github.com/MinecraftForge/MinecraftForge/blob/19f8d2a7937dc7968ecbef2f9785687193fcd210/patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch#L100-L108

but the event itself just passes it through with no double checking.
The javadoc mentions that a y position of -1 indicates unknown position, so I added a null check:
```
this.pos = pos != null ? pos : new BlockPos(0, -1, 0);
```

This retains compatibility with mods that were written according to the javadoc, and avoids an accidental breaking change.

Closes #7615.

### DIFF
--- a/src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
+++ b/src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
@@ -126,7 +126,7 @@ public class PlayerEvent extends LivingEvent
             this.state = state;
             this.originalSpeed = original;
             this.setNewSpeed(original);
-            this.pos = pos;
+            this.pos = pos != null ? pos : new BlockPos(0, -1, 0);
         }
 
         public BlockState getState() { return state; }


LexManos:
The better option would either be to add @Nullable to the  constructor args, or move this check outside. But either way this works. Also we need to keep this function in mind for 1.17's update with y-1 being valid now.

TheCurle:
I've put it on the list.

Issue #7506, opened 2020-11-22
[1.16.4-35.0.17] - EntityMinecart.registerCollisionHandler does not exist, contrary to documentation


SnapperTT:
The documentation of `IMinecartCollisionHandler` (see the java file) instructs you to register it with `EntityMinecart.registerCollisionHandler`, however `EntityMinecart` no longer exists (replaced by `AbstractMinecartEntity`), and it has no such method.

Forcing a registration by directly assigning `IForgeEntityMinecart.COLLISIONS = ...` is not permitted by the compiler despite the static member being declared public.

Forge version: 1.16.4-35.0.17

Relevant files:
net.minecraftforge.common.IMinecartCollisionHandler
net.minecraftforge.common.extensions.IForgeEntityMinecart
net.minecraft.entity.item.AbstractMinecartEntity

EDIT: I have made a workaround a week ago on my end by using a mixin. If this is a nofix or something that is halfway through a feature rewrite I would appreciate it if the comments in the files were updated instructing the modder to not use this for the time being.

sciwhiz12:
The move of the global `IMinecartCollisionHandler` from the `EntityMinecart` class to `IForgeEntityMinecart` was done in b6ddb0a7cf2b1ff353ff450c895c09f55753c7e9 (which created the aforementioned interface), which was in _2018_. It seems that nobody used that field, and therefore didn't find that error about it, until you did.

The fix here would be to move the field and getter back into `AbstractMinecartEntity`, since it cannot exist in the interface due to all fields in interfaces being implicitly `public static final`.

TheCurle:
I made #7748 to address these concerns. The current minecart system is a little bit of a mess, but we can work it out...

PR #7748, opened 2021-05-01
Reinstate the MinecartCollisionHandler field to AbstractMinecartEntity

TheCurle:
Some time during the 1.12 update cycle, the `IMinecartCollisionHandler` functionality was moved out of EntityMinecart and into a new interface `IForgeEntityMinecart`.
Unfortunately the static field containing the current handler was moved with it, and due to interface fields being implicitly `public static final`, and initialized to `null`, it (and the getter) became totally unusable.

This PR moves the field and getter (plus the `registerCollisionHandler` function that was dropped in the move to the interface) back to `AbstractMinecartEntity`, making this field usable for vanilla minecarts again.

Since `AbstractMinecartEntity` is such a usable and prominent template used in almost every mod that uses Minecarts nowadays, I also added comments calling for the removal of the `IForgeEntityMinecart` field and getter, since it's (normally) impossible for them to return anything except null.

### DIFF
--- a/patches/minecraft/net/minecraft/entity/item/minecart/AbstractMinecartEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/item/minecart/AbstractMinecartEntity.java.patch
@@ -25,22 +25,21 @@
  
     protected AbstractMinecartEntity(EntityType<?> p_i48538_1_, World p_i48538_2_) {
        super(p_i48538_1_, p_i48538_2_);
-@@ -104,6 +_,15 @@
+@@ -103,6 +_,14 @@
+       this.field_70167_r = p_i48539_5_;
        this.field_70166_s = p_i48539_7_;
     }
- 
++   
 +   public net.minecraftforge.common.IMinecartCollisionHandler getCollisionHandler() {
 +      return COLLISIONS;
 +   }
 +
-+   // IMinecartCollisionHandler redirects here via comment. We ensure it exists and it works.
 +   public static void registerCollisionHandler(@Nullable net.minecraftforge.common.IMinecartCollisionHandler handler) {
 +      COLLISIONS = handler;
 +   }
-+
+ 
     public static AbstractMinecartEntity func_184263_a(World p_184263_0_, double p_184263_1_, double p_184263_3_, double p_184263_5_, AbstractMinecartEntity.Type p_184263_7_) {
        if (p_184263_7_ == AbstractMinecartEntity.Type.CHEST) {
-          return new ChestMinecartEntity(p_184263_0_, p_184263_1_, p_184263_3_, p_184263_5_);
 @@ -138,7 +_,7 @@
     }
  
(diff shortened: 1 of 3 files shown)


pupnewfster:
Correct me if I am wrong but wouldn't the current implementation of this PR be a breaking change to any mods that set collisions currently using `IForgeEntityMinecart.COLLISIONS =`

sciwhiz12:
As said in the PR description, fields in `interface`s are implicitly (and required by the JVM to be) `public static final`, which makes assigning a value to be impossible.

pupnewfster:
Ah skimmed over that part by accident thanks.

patches/minecraft/net/minecraft/entity/item/minecart/AbstractMinecartEntity.java.patch
```diff
@@ -17,6 +25,22 @@
...
        super(p_i48538_1_, p_i48538_2_);
+@@ -104,6 +_,15 @@
+       this.field_70166_s = p_i48539_7_;
+    }
+ 
++   public net.minecraftforge.common.IMinecartCollisionHandler getCollisionHandler() {
++      return COLLISIONS;
++   }
++
++   // IMinecartCollisionHandler redirects here via comment. We ensure it exists and it works.
```
pupnewfster: This comment doesn't make much sense as it would seem "we ensure it exists and it works" is talking about the IMinecartCollisionHandler, but it doesn't seem that any verification actually happens here (in registerCollisionHandler)

TheCurle: its' existence in this class means it works.
It currently does not work.
I agree the wording is terrible :P

TheCurle: Note that also, the handler's default state is null, and thus null must remain a valid option for the setter, which means any checks we would reasonably need to do are done implicitly by the JVM when the setter is called.

pupnewfster: Ya I understand not checking anything here makes sense, my more point is that second part of the comment potentially shouldn't be there at all. The first part should make it clear enough why it exists/is unused but is needed

TheCurle: True enough. I'll remove it. Give me one moment.

LexManos:
This is still one of the most invasive and painful patches to manage every Minecraft version.
How many people still use all of these minecart hooks. They were originally created for CovertJaguire and rail craft. 
Are they still actively used? I know some are like the extension to allow custom rails. But physics handlers and things like that...

SkySom:
I can confirm that I do use all of the IAbstractRailBlock hooks. As for the Minecart hooks, I don't touch the Physics/Collison stuff, I do make some use of the speed/drag getters, as well as isPoweredCart/canBeRidden/getComparatorLevel methods.

TheCurle:
This PR was created in response to #7506. It seems this feature is relatively widely used, there are a bunch of hits on Google for people asking why this isn't usable any more.

How well that justifies the maintenance headache is a different matter, but all of the rest of the patches we need are there, it just needs this field to be somewhere the JVM will let you change it.

PR #7752, opened 2021-05-02
Fix "userProperties" errors for 1.12.x users. (Backport of #6740)

magicus:
Hi!

This is a fix for the problem where users with properties on their account cannot use the forge launcher. I have verified locally that it solves the problem for me. The fix is a straightforward backport of #6740; it's not a trivial cherry-pick since the code was in i different file, but the logic behind the fix is the same.

I understand if you guys just want to get rid of 1.12 already (trust me, I do too!) but due to Circumstances™️, it's not always that easy, and me (and certainly others!) are still stuck in the history. Please accept this PR, for our sake.


magicus:
I don't understand why the CI build fails. There seems to be some infrastructure issue; it did not even come near compiling the code I have changed.

PR #7753, opened 2021-05-03
Inconsistent Attribute Modifier Order

DavidQF555:
Currently, attribute modifiers added through the ItemAttributeModifierEvent are unreadable in the item tooltips because the order of the attribute modifiers is inconsistent in a HashMultimap and the attribute modifiers just keep on changing position on the tooltip. However, changing the attribute modifier HashMultimap to a LinkedHashMultimap causes the order of the attribute modifiers to be consistent and readable.

### DIFF
--- a/src/main/java/net/minecraftforge/event/ItemAttributeModifierEvent.java
+++ b/src/main/java/net/minecraftforge/event/ItemAttributeModifierEvent.java
@@ -19,7 +19,7 @@
 
 package net.minecraftforge.event;
 
-import com.google.common.collect.HashMultimap;
+import com.google.common.collect.LinkedHashMultimap;
 import com.google.common.collect.Multimaps;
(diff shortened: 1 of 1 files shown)


LexManos:
~~This is a quick fix, it still relies on the ordering of the handlers and all that. Better would be to have some actually defined sorting order do event handlers didnt have a issue. Also, this would be better if you included the gif and other explanations you had on discord, so that we have the extra context.~~

Ya, looking deeper into it, the issue is that you're creating a new random UUID every call for your Attribute, which is now how things should be done.
The random call is used in vanilla for the potion effect randomizations. Everything else that adds a consistent attribute uses a consistent UUID. So that it can be uniquly addressed and modified/removed.
So I would consider this working as intended, you just need to stabilize the value you're supplying.

PR #7754, opened 2021-05-04
Add Client Commands

coehlrich:
Updated version of #6952 which didn't have much happening since October.

There are a couple changes:

- Renamed the event to `RegisterClientCommandsEvent`.
- When running a client command add the original message to the sent messages history.


marchermans (approved):

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -61,10 +65,7 @@
 import net.minecraft.inventory.EquipmentSlotType;
 import net.minecraft.item.ItemStack;
 import net.minecraft.item.crafting.RecipeManager;
-import net.minecraft.util.Direction;
-import net.minecraft.util.Hand;
-import net.minecraft.util.MovementInput;
-import net.minecraft.util.ResourceLocation;
+import net.minecraft.util.*;
```
pupnewfster: Forge *tries* to stay away from wildcard imports (you have another being added lower down in this file as well)

coehlrich:
I just tried suggestions on both a single player world and on an offline mode vanilla server run on the same PC and both times suggestions worked fine.

On single player world:
image

On offline mode vanilla server:
image

patches/minecraft/net/minecraft/command/Commands.java.patch
```diff
@@ -25,3 +25,11 @@
           } catch (CommandException commandexception) {
              p_197059_1_.func_197021_a(commandexception.func_197003_a());
              return 0;
+@@ -276,6 +_,7 @@
+                argumentbuilder.executes((p_197053_0_) -> {
+                   return 0;
+                });
++               argumentbuilder.executes(null);
```
kashike: Why?

patches/minecraft/net/minecraft/network/play/server/SCommandListPacket.java.patch
```diff
@@ -0,0 +1,10 @@
+--- a/net/minecraft/network/play/server/SCommandListPacket.java
++++ b/net/minecraft/network/play/server/SCommandListPacket.java
+@@ -221,6 +_,7 @@
+                   this.field_197726_a.executes((p_197724_0_) -> {
+                      return 0;
+                   });
++                  this.field_197726_a.executes(null);
```
kashike: Why?

sciwhiz12: I'm not the author of the PR, but after reading the original PR and the justification given by the original PR's author (https://github.com/MinecraftForge/MinecraftForge/pull/6952#discussion_r458315650), I believe I understand the purpose of this.

**My suggestion**: As said in the linked review comment from the original PR, please add a comment briefly explaining why Forge sets the `executes` lambda of commands to be `null`.

***semi TL;DR: This allows the client hook to know that there is no client-side command, so it sends the command over to the server for final processing.***

### Background

This PR allows clients to register their custom commands to the client-side `CommandDispatcher<ISuggestionsProvider>` (as opposed to the server-side `CommandDispatcher<CommandSource>`). But, the commands are usually sent over to the server for final processing. 

### PR Logic

To allow the client-side custom commands to run, the PR manually executes the client-side dispatcher with the command. If there is no valid command found, then it delegates to the server for final execution. This happens when the parsed command tree does not end in a valid `executes` lambda. When no valid command is found, a `CommandSyntaxException` with the [`dispatcherUnknownCommand()`](https://github.com/Mojang/brigadier/blob/cf754c4ef654160dca946889c11941634c5db3d5/src/main/java/com/mojang/brigadier/exceptions/BuiltInExceptionProvider.java#L59) type is thrown. 

The hook captures command syntax exceptions, and if it captures one [with this specific type](https://github.com/coehlrich/MinecraftForge/blob/4b0d0484a6c6dbb5e293f2413b21909466d63ec8/src/main/java/net/minecraftforge/client/ForgeHooksClient.java#L854-L857), it assumes that there is no client-side command (otherwise execution would have succeeded, or an exception without that specific type is thrown), and thus throws it over to the server for processing.

(See [`ForgeHooksClient#sendMessage`](https://github.com/coehlrich/MinecraftForge/blob/4b0d0484a6c6dbb5e293f2413b21909466d63ec8/src/main/java/net/minecraftforge/client/ForgeHooksClient.java#L841) for the logic of the hook.)

### Modifying Vanilla's Behavior

In the original vanilla logic, when the client receives the server-side command information to construct a replica of the command tree (stored in the `CommandDispatcher<ISuggestionsProvider>`), it adds in empty lambdas for the command trees. This means that the explained logic would never work: since the command is valid due to the existence of the `executes` lambda, it gets executed _on the client_ (as the logic assumes if a command executes, it must be a custom client-side command) and it is never sent to the server for actual execution.

To allow the logic of the PR to work, this patch sets the `executes` parts of the replica command tree to be `null` instead of an empty (executing) lambda. This causes the command dispatcher to throw a command syntax exception with the aforementioned type, which allows the hook's logic to intercept it and send the command over to the server.

### Alternative

Possibly, this PR could create its own command syntax exception type, and instead of `null`-ing the lambdas, throws the exception with that custom type, and the hook would recognize that type specifically as the marker to send the command over to the server for real processing.

This has the added benefit of only sending valid server-listed commands (as defined in the replica command tree) over to the server for processing, while swallowing and presenting to the user if there was a genuine error with the client-side custom command (such as accidentally forgetting to add an `executes`).

patches/minecraft/net/minecraft/command/Commands.java.patch
sciwhiz12: See my previous lengthy explanation. As explained in the comment in the original PR, this  applies to the integrated server. The `SCommandListPacket` patch applies to the dedicated server (over the network syncing).

coehlrich:
I have separated the suggestions commands dispatcher from the client commands dispatcher (which allowed for removing the executes call when reading and writing packets), implemented a different form of the alternative that sciwhiz mentioned where instead of it throwing a custom exception when a server command get's executed the command is instead set to send the message to the server, I also created a ClientCommandSource with which most argument types that use a command source can be used apart from argument types that also use the world (since the return type of the method that the vanilla command source has is the server world) or the server, I also added a comment to the sendMessage method which was written prior to the suggestions commands dispatcher and the client commands dispatcher being split.

TheCurle:
This is a beast of a PR that none of us really feel comfortable triaging. 

We'll defer this until someone that knows what they're doing (like fry, preferably) takes a look at it.

This may just be something that has to be integrated upstream...

coehlrich:
Updated to the 1.17.x branch and the failed build's seems to be because of Jenkin's and not this PR.

coehlrich:
I want to override the `getServer` and `getLevel` methods in `ClientCommandSource` so that they throw an exception when trying to run in a client command but I'm not currently sure which exception to throw. I also want to add a separate method that can be used to get the un-sided Level.

coehlrich:
I have made a few more changes

Some things about the last change I mentioned in the commit message:

- The getWritableObjective for ObjectiveArgument will always be the same as getObjective when no objective's have been added client side since the server seems to always send objectives with the dummy type.
- The check that BlockPosArgument does to check if a block pos is in a loaded chunk always returns true on the client but the in world bounds check still works.
- Just remembered about the predicate argument's but I think I will wait for a review first.

src/main/java/net/minecraftforge/client/ClientCommandSource.java
```diff
@@ -0,0 +1,126 @@
...
+import net.minecraft.world.phys.Vec2;
+import net.minecraft.world.phys.Vec3;
+import net.minecraft.world.scores.Scoreboard;
+
+import java.util.Collection;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class ClientCommandSource extends CommandSourceStack
```
sciwhiz12: This should be named `ClientCommandSourceStack`, to follow the naming scheme of its superclass and to clarify it is different from `CommandSource`. Additionally, javadocs should be created for this class and each of its methods (along with `IForgeCommandSourceStack`).

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,226 @@
...
+
+        mergeCommandNode(serverCommandsCopy, commands.getRoot(), new HashMap<>(), Minecraft.getInstance().getConnection().getSuggestionsProvider(),
+                (source) -> {
+                    Minecraft.getInstance().player.chat((source.getInput().startsWith("/") ? "" : "/") + source.getInput());
+                    return 0;
+                }, (suggestions) -> null);
+    }
+
+    private static <S, T> void mergeCommandNode(CommandNode<S> sourceNode, CommandNode<T> resultNode, Map<CommandNode<S>, CommandNode<T>> sourceToResult,
+            S canUse, Command<T> execute, Function<SuggestionProvider<S>, SuggestionProvider<T>> sourceToResultSuggestion)
```
sciwhiz12, line 174: This and the `toResult` methods should have informative javadocs describing what they do and their parameters, for easier maintenance in the future.

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,226 @@
...
+
+    private static <S, T> CommandNode<T> toResult(CommandNode<S> sourceNode, Map<CommandNode<S>, CommandNode<T>> sourceToResult, S canUse, Command<T> execute,
+            Function<SuggestionProvider<S>, SuggestionProvider<T>> sourceToResultSuggestion)
+    {
+        if (!sourceToResult.containsKey(sourceNode))
+        {
+            ArgumentBuilder<T, ?> resultBuilder = null;
+            if (sourceNode instanceof ArgumentCommandNode<?, ?>)
+            {
+                ArgumentCommandNode<S, ?> sourceArgument = (ArgumentCommandNode<S, ?>) sourceNode;
```
sciwhiz12: This could possibly use pattern matching `instanceof` -- my IDE reports that `ArgumentCommandNode<S, ?> sourceArgument` doesn't cause an error, but I suggest a bit more testing to test and confirm it.

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,226 @@
...
+                RequiredArgumentBuilder<T, ?> resultArgumentBuilder = RequiredArgumentBuilder.argument(sourceArgument.getName(), sourceArgument.getType());
+                if (sourceArgument.getCustomSuggestions() != null)
+                {
+                    resultArgumentBuilder.suggests(sourceToResultSuggestion.apply(sourceArgument.getCustomSuggestions()));
+                }
+                resultBuilder = resultArgumentBuilder;
+            }
+            else if (sourceNode instanceof LiteralCommandNode<?>)
+            {
+                LiteralCommandNode<S> sourceLiteral = (LiteralCommandNode<S>) sourceNode;
```
sciwhiz12: This could possibly use pattern matching `instanceof` -- my IDE reports that `LiteralCommandNode<S> sourceLiteral` doesn't cause an error, but I suggest a bit more testing to test and confirm it.

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,226 @@
...
+                resultBuilder = LiteralArgumentBuilder.literal(sourceLiteral.getLiteral());
+            }
+
```
sciwhiz12: What happens if a node is neither a `LiteralCommandNode` or an `ArgumentCommandNode`? Currently, this will cause a `NullPointerException`, but I'd like for the behavior to be defined explicitly.

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,226 @@
...
+            {
+                resultNode.addChild(toResult(sourceChild, sourceToResult, canUse, execute, sourceToResultSuggestion));
+            }
+        }
+    }
+
+    private static <S, T> CommandNode<T> toResult(CommandNode<S> sourceNode, Map<CommandNode<S>, CommandNode<T>> sourceToResult, S canUse, Command<T> execute,
+            Function<SuggestionProvider<S>, SuggestionProvider<T>> sourceToResultSuggestion)
+    {
+        if (!sourceToResult.containsKey(sourceNode))
```
sciwhiz12: You could avoid a level of indentation by inverting this check and returning early the result of `sourceToResult.get(sourceNode)` instead.

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,226 @@
...
+
+    /**
+     * Always try to execute the cached parsing of client message as a command. Requires that the execute field of the commands to be set to send to server so that they aren't
+     * treated as client command's that do nothing.
+     * 
+     * {@link net.minecraft.command.Commands#handleCommand(net.minecraft.command.CommandSource,String)} for reference
+     * 
+     * @param currentParse
+     *            current state of the parser for the message
+     * @return false leaves the message to be sent to the server, true means it should be caught before {@link net.minecraft.client.gui.screen.ChatScreen#func_231161_c_(String)}
```
sciwhiz12: These javadocs need to be updated for the new names.

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,226 @@
...
+                Minecraft.getInstance().player.sendMessage(new TextComponent("").append(details).withStyle(ChatFormatting.RED), Util.NIL_UUID);
+            }
+        }
+        catch (Exception generic)// Probably thrown by the command
+        {
+            TextComponent message = new TextComponent(generic.getMessage() == null ? generic.getClass().getName() : generic.getMessage());
+            Component component = new TranslatableComponent("command.failed");
+            component.getStyle().withHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, message));
+            Minecraft.getInstance().player.sendMessage(new TextComponent("").append(component).withStyle(ChatFormatting.RED), Util.NIL_UUID);
+            generic.printStackTrace();
```
sciwhiz12: This should be logged with a message instead of printed into `System.out`.

src/test/java/net/minecraftforge/debug/chat/ClientCommandTest.java
```diff
@@ -0,0 +1,120 @@
...
+        event.getDispatcher().register(
+                Commands.literal("clientcommandtest")
+                        .then(Commands.literal("rawsuggest")
+                                .then(Commands.argument("block", ResourceLocationArgument.id())
+                                        .suggests((c, b) -> SharedSuggestionProvider.suggestResource(ForgeRegistries.BLOCKS.getKeys(), b))
+                                        .executes(new TestCommand())))
+                        .then(Commands.literal("registeredsuggest").then(
+                                Commands.argument("block", ResourceLocationArgument.id())
+                                        .suggests(SuggestionProviders.AVAILABLE_BIOMES)
+                                        .executes(new TestCommand())))
```
sciwhiz12: Instead of creating an object, this could a method reference to a static method instead.

src/test/java/net/minecraftforge/debug/chat/ClientCommandTest.java
```diff
@@ -0,0 +1,120 @@
...
+{
+    public ClientCommandTest()
+    {
+        MinecraftForge.EVENT_BUS.addListener(this::init);
+    }
+
+    private void init(RegisterClientCommandsEvent event)
+    {
+        event.getDispatcher().register(
+                Commands.literal("clientcommandtest")
```
sciwhiz12, line 51: Each of these subcommands should have a small comment explaining what they are testing for.

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,226 @@
...
+     * 
+     * Merges command dispatcher use for suggestions to the command dispatcher used for client commands so they can be sent to the server, and vice versa so client commands appear
+     * with server commands in suggestions
+     */
+    public static void mergeServerCommands(RootCommandNode<SharedSuggestionProvider> serverCommands)
+    {
+        commands = new CommandDispatcher<>();
+        MinecraftForge.EVENT_BUS.post(new RegisterClientCommandsEvent(commands));
+        RootCommandNode<SharedSuggestionProvider> serverCommandsCopy = new RootCommandNode<>();
+        mergeCommandNode(serverCommands, serverCommandsCopy, new HashMap<>(), Minecraft.getInstance().getConnection().getSuggestionsProvider(),
```
sciwhiz12: These three invocations of `mergeCommandNode` should be documented with comments explaining what they do and why they do so.

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,226 @@
...
+import net.minecraftforge.client.event.RegisterClientCommandsEvent;
+import net.minecraftforge.common.MinecraftForge;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+
+public class ClientCommandHandler
+{
+    private static CommandDispatcher<CommandSourceStack> commands = null;
```
sciwhiz12, line 61: Should the client-side command dispatcher be exposed too, like how `Commands#getDispatcher()` exposes the server-side command dispatcher?

sciwhiz12 (changes requested):
First off, please update this PR to 1.18.1. 🚀
Second, I've done a bit of testing for this PR, using the test client and both a (forgedev) Forge server and a vanilla server, and it seems to work fine without problems. However, aside from my review comments and a general call to add more javadocs, I'd like a few points to be addressed (with test mods to confirm the behavior when possible):

- How does this handle command requirements, via `requires`?
- How does this handle forking commands (`ArgumentBuilder#isFork()`), or those with custom redirect modifiers (`ArgumentBuilder#getRedirectModifier()`)?
- There should be documentation somewhere on the behaviors of at least `ResourceLocationArgument.getAdvancement`, `ResourceLocationArgument.getRecipe`, and `ObjectiveArgument.getObjective` when called from a client command. Specifically, from my testing, they can only return those objectives, recipes, and advancements visible to the player -- advancements shown on the advancements screen, recipes shown in the recipe book, and objectives set to a display location (`/scoreboard objectives setdisplay`) that are shown to the player.

I encourage people to try out this PR locally and make their own client commands to see if there are any changes or fixes which needs to be made to this PR. 

I'm personally looking to have this merged in soon:tm:, since this is a good feature that's been requested for a good amount of time and it would be a shame to have this keep sitting here without any progress. For the record, I am willing to take over maintenance of this feature once this gets into Forge.

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
coehlrich: The first type parameter for `ArgumentCommandNode` is used for the type parameter for `CommandNode` so the cast should be safe and likewise for `LiteralCommandNode` . Pattern matching `instanceof` can't have type parameters while this requires the type parameter.

coehlrich:
Recipes get sent to the client and stored in the RecipeManager even when they aren't visible in the recipe book. The recipe book instead uses `ClientRecipeBook` for recipes to show.

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    /**
+     * Always try to execute the cached parsing of client message as a command. Requires that the execute field of the commands to be set to send to server so that they aren't
+     * treated as client command's that do nothing.
+     * 
+     * {@link net.minecraft.commands.Commands#performCommand(CommandSourceStack, String)} for reference
+     * 
+     * @param currentParse
+     *            current state of the parser for the message
```
sciwhiz12: This needs to match the parameter name.

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,312 @@
...
+     * @return false leaves the message to be sent to the server, true means it should be caught before {@link net.minecraft.client.gui.screen.ChatScreen#func_231161_c_(String)}
```
sciwhiz12: The reference to the `ChatScreen` method needs to be fixed.

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,312 @@
...
+
+        // Copies the server commands into another RootCommandNode so that redirects can't be used with client commands
+        RootCommandNode<SharedSuggestionProvider> serverCommandsRoot = serverCommands.getRoot();
+        CommandDispatcher<SharedSuggestionProvider> newServerCommands = new CommandDispatcher<>();
+        copy(serverCommandsRoot, newServerCommands.getRoot());
+
+        // Copies the server side commands into a temporary server side commands root node to be used later without the client commands
+        RootCommandNode<SharedSuggestionProvider> serverCommandsCopy = new RootCommandNode<>();
+        mergeCommandNode(newServerCommands.getRoot(), serverCommandsCopy, new IdentityHashMap<>(),
+                Minecraft.getInstance().getConnection().getSuggestionsProvider(), (suggestions) -> 0, (suggestions) -> null);
```
sciwhiz12: Lambda parameters should be better named:
```suggestion
                Minecraft.getInstance().getConnection().getSuggestionsProvider(), (context) -> 0, (suggestions) -> null);
```

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,312 @@
...
+                    return commands.getCompletionSuggestions(parse);
+                };
+            }
+            return suggestionProvider;
+        });
+
+        // Copies the server side commands into the client side commands so that they can be sent to the server as a chat message
+        mergeCommandNode(serverCommandsCopy, commands.getRoot(), new IdentityHashMap<>(), Minecraft.getInstance().getConnection().getSuggestionsProvider(),
+                (source) -> {
+                    Minecraft.getInstance().player.chat((source.getInput().startsWith("/") ? "" : "/") + source.getInput());
```
sciwhiz12:
```suggestion
                (context) -> {
                    Minecraft.getInstance().player.chat((context.getInput().startsWith("/") ? "" : "/") + context.getInput());
```

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,312 @@
...
+
+        // Copies the server side commands into a temporary server side commands root node to be used later without the client commands
+        RootCommandNode<SharedSuggestionProvider> serverCommandsCopy = new RootCommandNode<>();
+        mergeCommandNode(newServerCommands.getRoot(), serverCommandsCopy, new IdentityHashMap<>(),
+                Minecraft.getInstance().getConnection().getSuggestionsProvider(), (suggestions) -> 0, (suggestions) -> null);
+
+        // Copies the client side commands into the server side commands to be used for suggestions
+        mergeCommandNode(commands.getRoot(), newServerCommands.getRoot(), new IdentityHashMap<>(), getSource(), (suggestions) -> 0, (client) -> {
+            SuggestionProvider<SharedSuggestionProvider> suggestionProvider = SuggestionProviders
+                    .safelySwap((SuggestionProvider<SharedSuggestionProvider>) (SuggestionProvider<?>) client);
```
sciwhiz12:
```suggestion
        mergeCommandNode(commands.getRoot(), newServerCommands.getRoot(), new IdentityHashMap<>(), getSource(), (context) -> 0, (suggestions) -> {
            SuggestionProvider<SharedSuggestionProvider> suggestionProvider = SuggestionProviders
                    .safelySwap((SuggestionProvider<SharedSuggestionProvider>) (SuggestionProvider<?>) suggestions);
```

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,312 @@
...
+                Minecraft.getInstance().player.sendMessage(new TextComponent("").append(details).withStyle(ChatFormatting.RED), Util.NIL_UUID);
+            }
+        }
+        catch (Exception generic)// Probably thrown by the command
+        {
+            TextComponent message = new TextComponent(generic.getMessage() == null ? generic.getClass().getName() : generic.getMessage());
+            Component component = new TranslatableComponent("command.failed");
+            component.getStyle().withHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, message));
+            Minecraft.getInstance().player.sendMessage(new TextComponent("").append(component).withStyle(ChatFormatting.RED), Util.NIL_UUID);
+            LOGGER.error("Error executing client command", generic);
```
sciwhiz12: This logging message should include the failing command.

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,312 @@
...
+     *            a function to convert from the {@link SuggestionProvider} with the original source stack to the {@link SuggestionProvider} with the result source stack
+     * @return the deep copied command node with the new source stack
+     */
+    private static <S, T> CommandNode<T> toResult(CommandNode<S> sourceNode, Map<CommandNode<S>, CommandNode<T>> sourceToResult, S canUse, Command<T> execute,
+            Function<SuggestionProvider<S>, SuggestionProvider<T>> sourceToResultSuggestion)
+    {
+        if (sourceToResult.containsKey(sourceNode))
+            return sourceToResult.get(sourceNode);
+
+        ArgumentBuilder<T, ?> resultBuilder = null;
```
sciwhiz12: The initializer to `null` is unneeded, as variables for reference objects are by default set to `null` and this variable will always have a value (or an exception is thrown).

sciwhiz12 (changes requested):
A few more small changes, and I think this looks fine to me.

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,312 @@
...
+                }
+                details.append(new TranslatableComponent("command.context.here").withStyle(ChatFormatting.RED, ChatFormatting.ITALIC));
+                Minecraft.getInstance().player.sendMessage(new TextComponent("").append(details).withStyle(ChatFormatting.RED), Util.NIL_UUID);
+            }
+        }
+        catch (Exception generic)// Probably thrown by the command
+        {
+            TextComponent message = new TextComponent(generic.getMessage() == null ? generic.getClass().getName() : generic.getMessage());
+            Component component = new TranslatableComponent("command.failed");
+            component.getStyle().withHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, message));
```
sciwhiz12: This won't work, as `Style#withHoverEvent` returns a new `Style` (as they are immutable objects).

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,312 @@
...
+        }
+        catch (CommandSyntaxException syntax)// Usually thrown by the CommandDispatcher
+        {
+            Minecraft.getInstance().player.sendMessage(
+                    new TextComponent("").append(ComponentUtils.fromMessage(syntax.getRawMessage())).withStyle(ChatFormatting.RED), Util.NIL_UUID);
+            if (syntax.getInput() != null && syntax.getCursor() >= 0)
+            {
+                int position = Math.min(syntax.getInput().length(), syntax.getCursor());
+                MutableComponent details = new TextComponent("").withStyle(ChatFormatting.GRAY);
+                details.getStyle().withClickEvent(new ClickEvent(ClickEvent.Action.SUGGEST_COMMAND, reader.getString()));
```
sciwhiz12: This won't work, as `Style#withClickEvent` returns a new `Style` (as they are immutable objects).

src/main/java/net/minecraftforge/client/ClientCommandHandler.java
```diff
@@ -0,0 +1,312 @@
...
+     * 
+     * {@link net.minecraft.commands.Commands#performCommand(CommandSourceStack, String)} for reference
+     * 
+     * @param sendMessage
+     *            the chat message
+     * @return false leaves the message to be sent to the server, true means it should be caught before {@link LocalPlayer#chat(String)}
+     */
+    public static boolean sendMessage(String sendMessage)
+    {
+        ClientCommandSourceStack source = getSource();
```
sciwhiz12: A small nitpick: This should be moved to below the `if`-check, so the source isn't constructed if not needed.

sciwhiz12 (changes requested):
A bit more changes that _almost_ got by me. 😅

sciwhiz12 (approved):
:shipit:

PR #7763, opened 2021-05-09
[1.16.x] Use wither as griefing entity for roses

Judge40:
A MobGriefingEvent is fired when a LivingEntity is killed by a WitherEntity, the outcome of the event is used to determine whether to create a WitherRoseBlock or WitherRose item.

The entity passed to the MobGriefingEvent is currently the LivingEntity being killed, but I feel that it would be more intuitive if it was the WitherEntity itself.
All other mobGriefing events are based on the entity taking an action (movement, attacking etc.) so it feels like the attacking Wither should be the owner of this particular behaviour, rather than the entity who is passively dying.

I realise this deviates slightly from how the other MobGriefingEvents are done, so I'll accept this being rejected if its not something that warrants consideration/discussion.

### DIFF
--- a/patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
@@ -160,7 +160,7 @@
           boolean flag = false;
           if (p_226298_1_ instanceof WitherEntity) {
 -            if (this.field_70170_p.func_82736_K().func_223586_b(GameRules.field_223599_b)) {
-+            if (net.minecraftforge.event.ForgeEventFactory.getMobGriefingEvent(this.field_70170_p, this)) {
++            if (net.minecraftforge.event.ForgeEventFactory.getMobGriefingEvent(this.field_70170_p, p_226298_1_)) {
                 BlockPos blockpos = this.func_233580_cy_();
                 BlockState blockstate = Blocks.field_222388_bz.func_176223_P();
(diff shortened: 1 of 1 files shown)


Judge40:
A little more background for why I wanted to make this change...

I have a mod that toggles mobGriefing for individual entity types separately.
The Wither killing entities with different mobGriefing values feels awkward, because some will turn in to blocks and others will turn in to items.

Switching the griefing entity to the Wither for this event allows that behaviour to be consistent again.

mcenderdragon:
I think the correct solution would be a sub event, fired from the wither with the target entity. so no information is lost.

Judge40:
> I think the correct solution would be a sub event, fired from the wither with the target entity. so no information is lost.

I'm not sure what you mean by that, could you give an example?

From the context of the Wither there is no sight of the WitherRose behaviour, so any change would need to be in LivingEntity called during the death event.

I suppose it could combine both checks? 
Like so:
```
if (net.minecraftforge.event.ForgeEventFactory.getMobGriefingEvent(this.field_70170_p, p_226298_1_) &&
    net.minecraftforge.event.ForgeEventFactory.getMobGriefingEvent(this.field_70170_p, this)) {
  ...
}
```

That does feel a bit awkward though, since the ability to force a `true` state is lost.

sciwhiz12:
Hello! @Judge40, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

Judge40:
Hi @sciwhiz12, I'll update it to the latest once I get some time 👍

PR #7767, opened 2021-05-14
Make DoorBlock.isWoodenDoor use the wooden doors tag

KnightMiner:
Previously, villagers closing doors required the tag, but opening the door was based on a material check, meaning that doors that do not have `Materials.WOOD` will be unable to be opened by villagers. In addition, some mob path finding is based on this method, to determine if a door block is openable without redstone. As a result of this patch, mods can add doors which can be controlled by hand but are not wood to the tag `minecraft:wooden_doors` and villagers will be able to fully control the door. Since this is all serverside only logic, it seems safe to use a tag in patches.

To test this patch, use [this attached datapack](https://github.com/MinecraftForge/MinecraftForge/files/6476415/DoorMadness.zip). It will tag iron doors as `minecraft:wooden_doors`, allowing villagers to open iron doors. You can use a bell to force a villager to run inside.

Actual use cases:
* I have a "slimewood" door which is effectively wood, but not broken by an axe (so it uses a different material). This would allow treating that door as wooden doors for villagers
* I have a glass door that similarly has the wrong material for villagers to use it

----

It is worth noting to get the desired behavior, doors that are not logically wooden will be added to the wooden doors tag by mods. Semantically this is a little odd if mods are expecting that tag to contain only doors that are logically wood. However, recipes would be unaffected as only the block tag needs wooden doors for this functionality, and not the item tag.

If this is still considered a problem, there are a few potential changes:
* Forge could add separate wooden tags for mods that want to check if something is logically wood
* Forge could add a new tag (e.g. `forge:interactable_doors`) for the desired functionality for this pull request. I did not do that innitially as it would increase the patch number quite a bit, especially as I did not feel it was right to replace the logic inside `DoorBlock.isWoodenDoor` to use a tag that is not labeled wooden, so I would have to instead update places that call that method (essentially the same problem before adding the new tag)

If either of those solutions is preferred I can change my patch to match.

### DIFF
--- a/patches/minecraft/net/minecraft/block/DoorBlock.java.patch
+++ b/patches/minecraft/net/minecraft/block/DoorBlock.java.patch
@@ -1,0 +1,10 @@
+--- a/net/minecraft/block/DoorBlock.java
++++ b/net/minecraft/block/DoorBlock.java
+@@ -224,6 +_,6 @@
+    }
+ 
+    public static boolean func_235492_h_(BlockState p_235492_0_) {
+-      return p_235492_0_.func_177230_c() instanceof DoorBlock && (p_235492_0_.func_185904_a() == Material.field_151575_d || p_235492_0_.func_185904_a() == Material.field_237214_y_);
++      return p_235492_0_.func_177230_c() instanceof DoorBlock && p_235492_0_.func_177230_c().func_203417_a(net.minecraft.tags.BlockTags.field_200152_g);
+    }
+ }


marchermans:
Personally I find the use of the tag here a weird solution.

If I want my none wooden door to be used by villagers i have to add it to the tag. But that makes it also useable in recipes, which i do not want....

So not sure how this should be treated. Maybe a different tag altogether? Or some additional logic?

XFactHD:
The recipe uses the item tag and the villager interaction uses the block tag, which means this won't be a problem.

marchermans:
Okey. That makes sense.

Then yeah this is okey for me.

marchermans (approved):

marchermans:
After internal feedback, i need made a mistake here assigning this.

As it stands right now this is a breaking change. You will need to add the material check back in and add a new tag.

KnightMiner:
How is this a breaking change? As described in my post, doors with the tag have half the functionality, and doors with the material the other half. It is impossible that someone has a working door with the material and not the tag, it will be able to open but not close. Likewise, if someone has the tag but not the material it will close but not open 

I am fine with making a new tag if that's desired, I just see no backwards compatibility benefit to the material check. And if I'm going to add the new tag you need to answer the other question in the original post

LexManos:
Your fundamental understanding of how entity tasks/ai works seems to be wrong.
A mod's door that is the correct material is not 'half the functionality'
It will just work with the Goal and not the Task.
Both the Goal and the Task do the same thing, opening AND closing doors.
It's just a matter of which one you want to work with.
But yes it looks like Mojang 1/2 did the task, it *should* be it's own dedicated tag if they were gunna use tags as the logic here. As Orion and the others on the discord are correct having non-wooden doors in the 'wooden' tag is a flat out no. It would have to be some stupid name like 'entity_pathing_openable_door' or better something that incorporates the entity name so that we could make doors openable by cats.. but not zombies. {If they were to have the goal} I would advise asking Mojang to rename that tag to something more expandable.

However, as it sits, it seems that this is working as intended. The difference is Vindicator/Zombies breaking down doors, and Vindicators opening the doors during raids use the material {The children of InteractDoorGoal} vs Everything that uses the new brain system {Villagers, Brutes, and Piglins} using the tag based door check for interaction. 

The only other potential issue is things that use the WalkNodeProcessor to gather their path. Not properly getting the DOOR_WOOD_CLOSED note type. Which can be addressed by you using our getAiPathNodeType hook correctly. 

Now considering this is a breaking change, because it breaks the fully functional Goal for anyone who has an existing wooden door, you can achieve full functionality using existing hooks, and the ickyness of throwing non-wood things into a tag labeled wood. This is denied.

Issue #7396, opened 2020-10-05
Two single quotes are ignored in translation entries (other than en_us)


AldanTanneo:
**Minecraft Version:** all minecraft 1.15 & 1.16 versions

**Forge Version:** all forge 1.15 & 1.16 versions

**Steps to Reproduce:**
*The easiest I could find*
1. Switch language to "Français (Canada)"
2. Get the "Kill the Ender Dragon" advancement (`/advancement grant @s from minecraft:end/kill_dragon`)
3. Check the "Bonne chance!" (aka Free the End) advancement description; in Vanilla Minecraft it should be `Libérez l'End en tuant l'Ender Dragon` but here appears as `Libérez lEnd en tuant lEnder Dragon`

![2020-10-05_09 53 19](https://user-images.githubusercontent.com/64717124/95064749-789ece80-0700-11eb-8840-63334f5273c2.png)

**Description of issue:**
Single quotes seem to be interpreted as string delimiters even *inside* translation entries and are ignored when they come in pairs. The issue does not appear in Vanilla Minecraft, but does with Forge even without any mods installed. I tried with a bunch of custom lang files, the issue appears in block names, advancements, etc.

~~I noticed however that the issue doesn't happen with the en_us translation...~~ _(Edit: false)_

diesieben07:
This is caused by Forge's extended message format for the language files. In particular [this method in ForgeI18n](https://github.com/MinecraftForge/MinecraftForge/blob/9f3141ea1630221b254c71bee82b95d723991d22/src/main/java/net/minecraftforge/fml/ForgeI18n.java#L109-L112).

I don't know if anyone actually uses this or if it is still necessary.

diesieben07:
Yes it is necessary, it is needed for Forge's translations.
An ugly workaround would be to check for `{` and `}` in the message, but that's not exactly foolproof.

PR #7769, opened 2021-05-15
Fix for ignored quotes in translation entries #7396

peterpalfi:
This PR fixes issue #7396 by escaping single apostrophes if a format string contains an even number of them.

Here's an example of this fix for the problem detailed in the original issue:


LexManos:
This wont actually fix the issue it'd just hide it better.
What should probably done, If anyone cares, is to either add a flag to the language file that enables the extended system. Or do it on a per-translation bases by trying to detect groups `{}`'s and enabling it there...

sciwhiz12:
Closing as there exists an LTS backport PR (#8074) which backports a fix for the same issue that's present in 1.17 from another PR (#8050). Thank you for your submission!

PR #7770, opened 2021-05-16
RFC for 1.17: New API for HUD overlays

gigaherz:
## What this does

1. Introduces a registry-based overlay list, with a basic dependency ordering option (**not** topological sorting).
2. Gives all current render elements their own overlay entry.
3. Reworks the RenderGameOverlayEvent to accomodate this new system: overlay layers become RenderGameOverlayEvent.PreLayer/PostLayer.

## What does doesn't do

1. It does not remove RenderGameOverlayEvent.
2. It does not change which layers exist in the rendering.
3. It does not change the order in which vanilla layers render.

### To add a new layer

Call `OverlayRegistry.registerOverlayBottom`, `OverlayRegistry.registerOverlayBelow`, `OverlayRegistry.registerOverlayAbove`, or `OverlayRegistry.registerOverlayTop`, during FMLClientSetupEvent. This method is synchronized so there's no need for enqueueWork,

To know what priority value you want to use, look at `ForgeIngameGui` for which priorities are assigned to the vanilla layers.

### To hide or un-hide an existing layer

Call `OverlayRegistry.enableOverlay` with the appropriate enable value.

### DIFF
--- a/src/main/java/net/minecraftforge/client/event/RenderGameOverlayEvent.java
+++ b/src/main/java/net/minecraftforge/client/event/RenderGameOverlayEvent.java
@@ -22,6 +22,7 @@ package net.minecraftforge.client.event;
 import java.util.ArrayList;
 
 import com.mojang.blaze3d.matrix.MatrixStack;
+import net.minecraftforge.client.gui.IIngameOverlay;
 import net.minecraftforge.eventbus.api.Cancelable;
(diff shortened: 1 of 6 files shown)


Sarinsa:
What does doesn't do

gigaherz:
Ported into the internal 1.17 branch. Will be included on release.

Issue #7622, opened 2021-01-25
[RFC][1.17] Removal of Capability Storage/Default Factory


LexManos:
So, the entire point of the Storage/Default system in capabilities, was to hopefully force modders to document and give an example of their capability. That didn't seem to ever happen. So I'm proposing in 1.17, we just strip both those functions out. Unless there is a compelling reason to keep it. 

Basically it would be the removal of:
Capability.IStorage
Capability.getDefaultInstance
Capability.getStorage
Capability.readNBT
Capability.writeNBT
And the change to CapabilityManager.register to just take in Class<T> type.

Additional question: 
Should we force capabilities to be Interfaces. Right now it could be either as we don't validate the type. But we could easily add a  `if (!type.isInterface()) throw IAE` to CapabilityManager.register.
This would somewhat point people in the correct dimension that these capabilities SHOULD be interfaces {Ideally with good javadocs}, not concrete classes.

Alpvax:
I always found the default unhelpful, because I needed to pass in additional data (usually whatever I'm attaching the capability to).
As for storage, the only advantage that could have given is for different implementations to use the same storage, but generally that is done with inheritance.
Personally, I'm in favour of removal.

gigaherz:
I'm all in for it. I understand the intention, but I haven't really encountered the situation where I want to expose an API that would truly benefit from default instances. All my use cases always require at least an instance of the parent object (usually a PlayerEntity) passed into the constructor (to be able to make the owner a final field), and making the capability API use Function<Object, T> would be rather untypesafe.

josephtarbit:
Forcing interfaces seems like a good idea to me. I see a lot of people miss-using capabilities and using concrete classes.

gigaherz:
> Forcing interfaces seems like a good idea to me. I see a lot of people miss-using capabilities and using concrete classes.

Oh I missed that last bit from Lex's message. I would have to disagree with that part. In many cases (at least many of *my* use cases) a capability isn't a public API, but rather the only way we have to attach a concrete data object to a thirdparty object, and that data is never meant to be accessed by other mods because it contains *only* implementation details.

I may be overthinking this, but if we require that all capabilities are designed as public APIs, then it feels like we should also have some parallel API, which is designed for attaching "plain data"?

[...]

PR #7772, opened 2021-05-16
Add RegisterCapabilitiesEvent

diesieben07:
Adds a new mod event bus event which is fired when it is time to register capabilities.
### Why is this necessary?
Currently the place to register capabilities is in `FMLCommonSetupEvent`. This event however is _too late_, because Minecraft will initialize its search tree earlier than that and items will have their `initCapabilities` method called. That means any `initCapabilities` method will need to have a null-check to see if the capabilities have been initialized yet. Capabilities cannot be registered in the mod class constructor, because the capability manager is not ready yet at that point. In tradition with other registrations, a dedicated event should be added for this.

### Future
In 1.17 we should make `CapabilityManager.register` error-out if it is called outside of `RegisterCapabilitiesEvent`.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeMod.java
+++ b/src/main/java/net/minecraftforge/common/ForgeMod.java
@@ -166,10 +166,10 @@ public class ForgeMod implements WorldPersistenceHooks.WorldPersistenceHook
 
     public void registerCapabilities(RegisterCapabilitiesEvent evt)
     {
-        CapabilityItemHandler.register();
-        CapabilityFluidHandler.register();
-        CapabilityAnimation.register();
-        CapabilityEnergy.register();
+        CapabilityItemHandler.register(evt);
+        CapabilityFluidHandler.register(evt);
+        CapabilityAnimation.register(evt);
+        CapabilityEnergy.register(evt);
     }
 
     public void preInit(FMLCommonSetupEvent evt)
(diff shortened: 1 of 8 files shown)


pupnewfster:
Part of me wonders if a slightly better way of doing this would be to:
1. Add a todo to rename `injectCapabilities` to say `injectAndRegisterCapabilities` (given while it is internal it still is public so probably shouldn't be changed yet)
2. Move the event to being fired from that method
3. Have the event have an `register(Class<T>, Capability.IStorage<T> storage, Callable<? extends T> factory)` method that then adds things to a list, this may be cleaner in 1.17 due to #7622 when it would be able to just be `register(Class<T>)` (at which point potentially a set may be better than a list not sure).
4. Have the method grab from the event the list and add it to the manager (as the manager would be what is firing it.
5. Deprecate the current register method and have it print a warning to the log that says the event should be used.

The main benefit of this would be that then in 1.17 we could just remove the current register method to make sure people register it from the correct place/time as there will only be one place it can be called.

diesieben07:
Yes, that makes sense. I will improve it.

diesieben07:
I've added a registration method in the event and deprecated the old one.

diesieben07:
Superseded by https://github.com/MinecraftForge/MinecraftForge/pull/8021.

PR #7773, opened 2021-05-16
RFC for 1.17: Move OnlyIn(CLIENT) forge additions out of common code

gigaherz:
Currently forge has quite a few patches with OnlyIn(CLIENT). This makes no sense.

This introduces 3 mixins into forge, for Item, Block, and Effect. Along with setting up the basis of building forge with mixins. These mixins are only applied to the client, and add a field that holds the client rendering object for each object. Along with those mixins, there's 3 `initializeClient` methods patched into Item, Block, and Effect respectively.

Overriding the `initializeClient` method in a mod class allows providing the client-specific data in a way that will never be loaded in the dedicated server unless you get out of your way to break it. 

That info can be obtained via the methods in `RenderProperties`.


diesieben07:
Shouldn't these maps use `IRegistryDelegate` to support registry replacement?

gigaherz:
Now that you mention it.................... Yes.

gigaherz:
This has already been included in the internal 1.17 branch. Except for the mixins, but that will come at a later date.

PR #7777, opened 2021-05-18
Make deep copy of DimensionStructuresSettings for all chunkgenerator and made structure spacing map mutable

TelepathicGrunt:
Ok bear with me here. This PR's reason for existing is a bit tricky to understand but I'll start from the top.

## The desired result we want:
Many structure mods want to do a dimensional based whitelist/blacklist to add or remove their structures per dimension. This kind of config option is in extremely high demand by users and they won't stop asking us for it lol. We also want to be able to get our structures into custom dimensions from datapacks or mods and also still have our structure in the Overworld even after an Overworld overhaul datapack is put on. 

## The first problem:
At first, it may seem like we can just slap a dimensional into `isFeatureChunk` or into the Structure's `IStartFactory$create`. However, if you call locate command or treasure map tries to locate the structure, the game checks the Biome Source's list of biomes first to see if it has the structure in any biome that can spawn. If that passes, which it will, the locate method will now check up to 40000 chunks for the structure by calling the method to get the structure start locations, run `isFeatureChunk` on them, and then run the Structure's `IStartFactory$create` method until it finds a single chunk that returned at least 1 piece in its children list. If the dimension is blacklisted (or not whitelisted) within any of these 3 methods, this check never passes and the 40000 chunk check takes a massive amount of time that the server hangs for like 10+ minutes which is more than enough for it to kick all players off of it due to no response. This is why the dimensional blacklist/whitelist cannot ever be part of the structure itself.

Mojang realized this and their solution was to make every dimension's Chunk Generator take a noise setting and in there, are the structure's spacing configs. If a dimension doesn't have the spacing config, then the structure does not spawn at all in the dimension. Seems good right? We just need to add/remove from the noise setting and problem solved. Not quite...

## The second problem:
If we loop through the `WorldGenRegistries.NOISE_SETTINGS` and add/remove our structure spacing from there at say FMLCommonSetupEvent, there's several issues.
- Multiple dimensions can share the same noise setting which is a perfectly valid use case if the dimension is similar to the Overworld. For example, Lost Cities dimension uses `minecraft:overworld` which means, if we edit that noise setting file, treating it as a dimensional blacklist, it actually isn't and any change we do will propagate to both dimensions at the same time. 
- JSON noise setting files can override the `WorldGenRegistries.NOISE_SETTINGS` entries. This means users putting on a Overworld Overhaul datapack will be strictly incompatible with all structure mods using this way of injecting their structure spacing config. Users would then have to go into the datapack every single time, find all the other mod's Structure registry names, and add it manually to the new JSON `minecraft:overworld` noise setting file. Extremely bad for datapack compat and mod compat.
- Mods and datapacks that define their own dimensions with their own noise setting JSON file will not get the structure mod's structure spacing configs as `WorldGenRegistries.NOISE_SETTINGS` only works for code based noise settings entries registered beforehand.
- Structure mods will forget to add to `minecraft:amplified` noise settings. I do see a few mods trying to use  `WorldGenRegistries.NOISE_SETTINGS` loop to add their structures and most of the time, they keep accidentally skipping over `minecraft:amplified` which means their structures will not spawn in Amplified worldtype.
- Explaining to users the difference between a noise setting whitelist/blacklist and a dimensional whitelist/blacklist is futile. Users only know the biome registry names and the dimension registry names. They know nothing about noise settings as it is a buried technical thing that backs the dimension and finding what noise setting a dimension has is extremely difficult for a user. They would need to open the mod's jar to look for what noise setting a JSON dimension file specifies to use or look directly at code to see what noise setting was passed into the code defined dimension. And furthermore, I believe noise settings can be embedded into the dimension's JSON file or coded dimension as well which would mean the noise setting has no registry entry at all and thus, breaking the whitelist/blacklist noise setting config options

So what is the solution for this? Well, obvious due to the worldgen datapack stuff and our desire for dimensional whitelisting/blacklisting, we would need to do our changes after the world is first created. Which leads us to the `WorldEvent.Load` workaround that the majority of structure mods are currently doing. The event fires for every dimension, we can see what the dimension's registry name is for our whitelist/blacklist, make a new structure spacing map copy of the Chunk Generator's structure configs immutable map, do our changes to this new mutable map, then we can use an AT or accessor mixin to assign the mutable map into the Chunk Generator. Like this https://github.com/TelepathicGrunt/StructureTutorialMod/blob/3d22fe6884c4d596e7749096a95afedd98a88eda/src/main/java/com/telepathicgrunt/structuretutorial/StructureTutorialMain.java#L100
Problem solved right? Right?...

## The final problem which this PR address:
I did make a grave mistake when I made this workaround. It's not that the structure spacing map is shared between dimensions because we make a new map to replace it so that isn't the problem. It's the class that holds that map. DimensionStructuresSettings. All Chunk Generators with the same noise setting entry being used is using the exact same DimensionStructuresSettings instance. That means if the Overworld and LostCities both use `minecraft:overworld` noise setting and we only whitelist the Overworld, that means we add our structure into a new map and assign it to the Overworld and then make a new map without our structure and assigned it to LostCities. But since they both use the same DimensionStructuresSettings instance, we just overrode the Overworld's map with LostCities's map. That means there is no way to assign structure spacings configs per dimension as they share that one class instance between dimensions of the same noise settings.

## The solution:
This is where Forge comes in to save the day. The solution I came up with is to patch the `ChunkGenerator` constructor which all chunk generators uses and then create a new deep copy of the DimensionStructuresSettings passed in for that specific chunk generator. Thus making all dimensions now have their own unique DimensionStructuresSettings. Thus allowing everyone to do proper and working dimensional whitelisting/blacklisting for their structures in `WorldEvent.Load`. Part of the deep copying also makes a new copy for the Stronghold spacing config so now if a stronghold tweaking mod wants to allow users to change stronghold spacing and count per dimension with the same noise settings entry used, they can now do so. I also threw in a patch into DimensionStructuresSettings that makes its most used constructor now make a new mutable map copy of the structure spacing configs. Ironically, the second constructor in DimensionStructuresSettings does make a new mutable map but Mojang decided not to do that for the above constructor for wack reason lol. Thus this eliminates the need for ATs or accessor mixins as the map in DimensionStructuresSettings should now always be mutable unless someone hacks in an immutable map bypassing Forge's patch but we will deal with those mods on a case-by-case basis so that they cease doing that and give it a mutable map.

With the patch, this reduces the linked `WorldEvent.Load` workaround to this much simplier code. No ATs. No accessor mixin. No need for reflection due to a certain mod choosing to store the structure spacing config map in a new field to deliberately break the old workaround on purpose for... reasons I yet to understand...
```java
public void addDimensionalSpacing(final WorldEvent.Load event) {
	if(event.getWorld() instanceof ServerWorld){
		ServerWorld serverWorld = (ServerWorld)event.getWorld();

		if(serverWorld.getChunkSource().getGenerator() instanceof FlatChunkGenerator &&
			serverWorld.dimension().equals(World.OVERWORLD)){
			return;
		}
		
		serverWorld.getChunkSource().generator.getSettings().structureConfig
			.putIfAbsent(STStructures.RUN_DOWN_HOUSE.get(), DimensionStructuresSettings.DEFAULTS.get(STStructures.RUN_DOWN_HOUSE.get()));
	}
}
```

Now why did I patch the constructors instead of say the codec? Well, the codec is much, much harder to patch and maintain and would not cover cases where a mod would create the chunk generator or DimensionStructuresSettings directly and thus, bypass the codec. If a mod chooses to bypass the extremely important constructors of these two classes, then they have to deal with the side-effects of that choice as they are now doing unexpected behavior that no other mod can anticipate. Thus, the best practice would be to always call these vanilla constructors and use their fields properly. And by making the structure spacing config map mutable, it will become safe in the future for mods to store the spacing config map in a field and use that when all current mods finally stop replacing that map (although mods storing the structure spacing map in a new field really should always be grabbing the spacings from the DimensionStructuresSettings's method itself instead. Less redundant code and much better protection against whatever changes other mods do)

I also heard from several people who said, "just patch DimensionStructuresSettings's method for getting the spacings!" and the issue with that is, there is no world. Chunk Generators have no concept of a world. The only worlds they ever get is passed in by method params but is never stored in any field. Thus the DimensionStructuresSettings in the Chunk Generator will never be able to do dimensional whitelisting/blacklisting inside of itself since the world context is never passed to it when the spacing is grabbed. Hence why it is best for Forge to make every dimension have a unique DimensionStructuresSettings instance and unique mutable structure spacing map to make it a million times easier for all structure mods to add/remove their spacings on a per-dimension basis in `WorldEvent.Load`.

A proper and safe dimensional whitelisting/blacklisting config option for structure mods!

Let me know if there's any part of this explanation that is unclear or if there's any issues with this patch. I ran the license task which found that three other files in Forge didn't have licences but I was told to just slip it with this PR to get it out of the way for Forge. 

Hopefully this help! Ask any question or say any concerns about this PR now.


50ap5ud5:
This PR is exactly what I'm looking for. I came a cross the same issue where the DimensionStructureSettings persisted across all worlds which used the same noise setting.

Looking forward to seeing this merged!

TelepathicGrunt:
I should probably state the actual use cases for who this PRs benefits. My mod [Repurposed Structures](https://www.curseforge.com/minecraft/mc-mods/repurposed-structures#c596) and Yung’s mod called [Better Strongholds](https://www.curseforge.com/minecraft/mc-mods/yungs-better-strongholds) are two that tries to use dimensional whitelisting/blacklisting. In fact, Better Strongholds was having problems with some dimensional mods which is how I found out about this issue originally. 

There are most likely many other structure mods that I do not know of right now that also attempted dimensional whitelisting/blacklisting which is why this Pr is probably best to get into 1.16.5 sooner than later. I can make a new Pr for 1.17.0 when it releases without this pr’s code ported.

ModernGamingWorld:
Yeah I hope this gets consider for 1.16.5 before 1.17

Won-Ton:
If a mod has set up its dimension to use overworld's DimensionSettings they should be able to rely on _getting_ overworld's DimensionSettings. Contrived example: you're making an upside down version of the overworld - you'd want all settings to be exactly the same including structure placement.

If a mod just wants their dimension to be similar to overworld but to have independently configurable settings they should register and use their own DimensionSettings instance that copies over the settings they want from the overworld's - 
Forge could make that process easier by making the DimensionSettings constructor public and providing a registration event.

"Whitelist/blacklisting" a dimension then just becomes normal configuration of its DimensionSettings. #7743 and/or #7651 open up the opportunity for it to work with datapack'd stuff too.

imo this^ approach would be more in keeping _with_ vanilla's worldgen system, rather than working against/around it. Not a major stakeholder in this either way, just sharing some thoughts.

tldr; could be fixed by mods registering their own DimensionSettings for their dimensions

TelepathicGrunt:
No wonton. Forge's goal here is to not force extra work onto all dimension mod or datapacks just to set a non-agreed upon standard of every dimension must have a custom DimensionSettings defined.

Re-using the DimensionSettings instead of making a new custom copy of it for every overworld-shaped world is a valid use case for dimension mods and datapacks but the expectation is that other mods such as structures mods, can be customized to work with these dimensions out of the box. Mods do not and should not need to have arbitrary rules just to work with each other and explaining this distinction between DimensionSettings and dimensions to players is next to impossible.

If the vanilla system is problematic in a way for dynamic customization of mods, then it's fine to slightly adjust it. And in this case, just making the DimensionSettings unique for every dimension is the best, simple solution that automatically allows structure mods to configured per-dimension with zero extra work on the part of dimension mods. And it is far, far more intuitive for the players to understand and utilize mod config to customize. Forge isn't here to increase everyone's workload. And your PR is far from intuitive, redundant in a lot of ways (defining an entire noise setting files just to add a structure), and specific behavior is hidden from users in forge configs. Users will not be able to understand how to limit only certain structures with your system. And the PR Curle did which was helped by Commoble has a flaw that it breaks with recreate world to the point that they are giving up on that hook and waiting for 1.17 Forge to see if there's an alternative.

All this uncertainty, confusion, and extra work can be avoided... by making deep copies of DimensionSettings per dimension so all mods and datapacks work out of the box and structure mods can have a nice, simple, easy to understand config for users that can turn off/on specific structures from that mod per dimension.

Won-Ton:
Yes, there are valid usecases for mods to use overworld settings. That was the first thing I covered.

Configuring each dimension is exactly what DimensionSettings are for (clue's in the name). It's mojang's standard (not mine) just nobody is using it which is what's causing you pain. If it can be made easier to use through forge you'd see more adoption.

Curle's PR might not work atm for editing biomes but there's an event for that already. It'd work perfectly for modifying DimensionSettings and structure spacings because there's nowhere you'd be duplicating stuff if the event ran twice.

It's far less complicated to work with the system that vanilla uses than the coremod/AT work around you're telling people to do. Registering and using registered stuff has been a fundamental of modding for years. It's only because it's new to world-gen that it has caused some confusion and misuse. + You can slap whatever configuration abstractions you like on top still.

🤷‍♂️

Commoble:
This would break the ability to add structures to dimensions by modifying the noise settings themselves, I have several separate PRs I'm working on that address this

PR #7783, opened 2021-05-23
Fixed Entity Renderer Validator

DaemonUmbra:
The validator now runs on an executor that will log its failure exception.


diesieben07:
There is more to this. If the renderer factory throws an exception it is swallowed as well.

sciwhiz12:
Hello! @DaemonUmbra, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

PR #7786, opened 2021-05-24
Move UpdateRecipesPacket after TagsList on reload

uecasm:
This makes the reload packets be sent in the same order as when the client initially joins the server.  It's annoying to have these execute in different orders.

Not sure if this patch is still syntactically valid (it might need regenerating to get more prefix context).  But the idea of the desired change is here at least.


uecasm:
Whoops, wrong one, apologies for the noise.

PR #7787, opened 2021-05-24
Send recipe packets in the same order at reload as at world join.

uecasm:
This makes the recipe/tag reload packets be sent in the same order as when the client initially joins the server. It's annoying to have these execute in different orders.

### DIFF
--- a/patches/minecraft/net/minecraft/server/management/PlayerList.java.patch
+++ b/patches/minecraft/net/minecraft/server/management/PlayerList.java.patch
@@ -124,14 +124,23 @@
     }
  
     @Nullable
-@@ -804,6 +_,7 @@
+@@ -803,11 +_,15 @@
+          playeradvancements.func_240918_a_(this.field_72400_f.func_191949_aK());
        }
  
-       this.func_148540_a(new STagsListPacket(this.field_72400_f.func_244266_aF()));
-+      net.minecraftforge.fml.network.NetworkHooks.syncCustomTagTypes(this.field_72400_f.func_244266_aF());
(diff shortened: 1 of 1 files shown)


uecasm:
While it probably would be more logical to send the tags before the recipes, I figure that it's better to preserve the existing order of packets at client connect, since that happens more frequently than datapack reload and it would have better compatibility with existing mods.  So this will sync the tags after the recipes in both cases.

(Also, the whitespace layout is not ideal, but the patch checker got upset when I tried to delete a superfluous blank line.)

LexManos:
We are not in the habit of changing the ordering or format of vanilla packets. Especially for things designed to be intercepted/expanded by external projects.
You might file this as a 'bug' on Mojang's issue tracker and they can decide if its intentional that the order is different.
However, Tags and recipes are in theory independent systems that should be able to be reloaded/processed completely separate from each other {Yes recipes need to know about tags but that can be done without the contents of the tag}.
So if you're doing something weird that requires a specific order, that's more a issue with your code then a issue with the game.

Also, "It's annoying" is not valid justification for any changes like this.

uecasm:
The way things currently are, JEI does not work.  In order to be able to load recipes and tags on startup, it initialises itself on the CustomTagType event (which is the last one sent during startup).

However, on reload, that is the second of three events, and in particular the recipes themselves have not yet been synched, which means that it loads the recipes as they existed _before_ the datapack update.  It has to be reloaded a second time (with no further changes) to actually successfully populate the recipes.

If it loaded itself on the recipe population instead... well, that's the first event on connect, which means that none of the tags would be correct whenever you connected to a world until something triggered a reload (which is even worse).

Issue #7708, opened 2021-04-04
AnvilUpdateEvent has mismatched comment and actual behaviour


Rongmario:
**Minecraft Version:** 1.16.5

**Forge Version:** 36.1.4

**Description of issue:**
Since #7442, AnvilUpdateEvent has the comment: `AnvilUpdateEvent is fired when the inputs (either input stack, or the name) to an anvil are changed.`  This is currently untrue/unclear as it is only fired when both left, right input ItemStacks are populated. The name field can change and not fire this event if the right ItemStack is empty.

diesieben07:
This has been the case as far back as 1.12.2 at least (I couldn't check 1.7.10 as I couldn't get the workspace to work). Question is if we should change the behavior.

Rongmario:
The behaviour before did fit the comment (just wasn't explicated enough) prior to the aforementioned PR: `AnvilUpdateEvent is fired when a player places items in both the left and right slots of a anvil.`

TheCurle:
This is probably an issue for #7612, which aims to fix this type of comment mismatch. When i get around to rebasing that monster, i'll update you here.

Matyrobbrt:
This issue has been fixed by https://github.com/MinecraftForge/MinecraftForge/commit/130953e519e0c7f4a32b4d6c3af3740a09f07f30.

PR #7791, opened 2021-05-26
even call onAnvilChange if itemslot2 is air

pdummy00:
This PR implements the functionality what described in the event documentation.

I have seen that the issue (#7708) is marked as Work in Progress.
Currently event only called if there are items on both slots, with this patch it's even called if slot2 is empty (air).

Based on my tests, this patch will not cause any additional bugs, but it will provides more flexibility for modders in the future.


sciwhiz12:
Closing as it seems the user has a clear history without a fork of the Forge repository, and the commit is made using the GitHub web editor. If anyone wishes to pick up this PR, please feel free.

PR #7792, opened 2021-05-27
[1.16.x] Add method to get the number of elements in a model builder

mrp-v2:
Currently, model builders allow you to retrieve elements based on index, but do not provide a way to know how many elements there are. Because of this, iterating over all the elements in a builder requires using a try-catch block to increment the element index until a `IndexOutOfBoundsException` occurs. This PR fixes that by providing a method that simply returns the number of elements in the builder.


diesieben07:
Looks straightforward enough to me

diesieben07 (approved):

PR #7802, opened 2021-06-03
Fix vanilla packet splitter remote detection and add advancement packet splitting

diesieben07:
The "does the remote know about packet splitting" check was broken, because it always ran before the client event sent its channel list. This was fixed by moving the injection and check to just after the switch to the PLAY protocol (after the handshake).

Additionally this removes the client-side whitelist from the packet splitting and adds splitting for the advancement packet (prompted by [this thread](https://forums.minecraftforge.net/topic/101170-player-cannot-join-serverbadly-compressed-package/), thanks @SizableShrimp for the idea of making it compatible with older clients).

### DIFF
--- a/src/main/java/net/minecraftforge/fml/network/NetworkHooks.java
+++ b/src/main/java/net/minecraftforge/fml/network/NetworkHooks.java
@@ -52,6 +52,7 @@ import net.minecraft.util.math.BlockPos;
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.event.entity.player.PlayerContainerEvent;
 import net.minecraftforge.fml.config.ConfigTracker;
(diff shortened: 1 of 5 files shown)


marchermans (approved):

diesieben07:
Here are my (limited) test results (could not test advancements, because apparently you need so many of them that the server literally spends minutes just building the packet? I gave up):

- Old forge client with new forge server: Works fine. 
- New forge client with old forge server: works fine, if packets are small enough (splitter was broken before this PR)
- old forge client with old forge server: as before
- new forge client with new forge server: all is okay.
- vanilla client with new forge server: works fine if packets are small enough
- new forge client with vanilla server: works fine if packets are small enough

PR #7807, opened 2021-06-06
Add attack reach attribute to players

Thelnfamous1:
This PR adds a new Forge Attack Reach attribute and gives it to players, in addition to making a few patches to replace references to either the hardcoded vanilla attack reach of 3.0D or that same value squared (9.0D) with the equivalent using the new attribute.

This makes it easier for modders to add new items and effects that influence the default attack reach of players, without having to resort to using coremods that can cause incompatibilities with other mods.

A test mod is added, "player_attack_reach_test", that adds a "Longsword" which modifies the attack reach of the user when held in the mainhand.


Thelnfamous1:
After discussion with ChampionAsh, I came to the understanding that the existing reach attribute is the one that decides whether or not the cursor is over an entity, meaning that it would be possible for an increased attack reach to do nothing if the reach distance attribute is less than it. I've now made a different PR, #7808, which may be more suitable, so for now, I will close this.

PR #4736, opened 2018-02-11
Fixed the reach distance to work with entities as well as blocks.

Tmtravlr:
**Problem**

The new reach distance attribute works well with blocks, but doesn't work at all with entities.

In a mod I just updated, I used to add my own reach attribute, so seeing an official forge one is good. However, for the mod, I need it to work with entities. 

It would also generally be useful, for instance, for weapons that have a shorter and longer attack range, like daggers and spears.

**Solution**

I fixed the reach distance by fixing how the vanilla moused over entity is found. 

For client-server security, I also made sure the packet sent checks that the distance is valid server-side, to prevent reach hacks.

I did run into a small issue with that server-side check, since it checks the distance from the player to the entity from their positions (not from the player's eye height and the point on the entity's bounding box the player is looking at). So for large entities like ghasts, it didn't register that the ghast was close enough even though it was. 

To fix this, I made the check's distance the maximum it could be by adding the player's eye height and the biggest side of the entity's bounding box. If you can think of a better solution, let me know.

**Testing**

To test, you can give yourself an item like 
```/give @p diamond_chestplate 1 0 {AttributeModifiers:[{Name:reach, AttributeName:generic.reachDistance, Amount:10, Operation:0, Slot:chest, UUIDLeast:1, UUIDMost:1}]}```
and test that the reach works at the same distance for blocks and entities.

I also created a test world set up for testing it. If you want to see it, let me know the best way to share it.

**Compatability**

For forge clients connecting to a vanilla server, the packets will still work if the entity is within 6 blocks (or 3 without line of sight), as it normally works in vanilla.

The packets would also work for vanilla clients connecting to a forge server, but I'm not even sure if that's possible.

PR #7808, opened 2021-06-06
Add attack reach attribute and expand usage of reach distance attribute

Thelnfamous1:
Based on my previous #7807 PR

This PR adds a new ForgeMod.ATTACK_REACH attribute that replaces references to the hardcoded vanilla attack reach of 3.0D (and its square, 9.0D) with the equivalent using the new attribute. It also expands upon the existing ForgeMod.REACH_DISTANCE attribute by making a few patches to replace references to the ~~hardcoded vanilla attack reach of 3.0D (and its square, 9.0D) with the equivalent using the reach attribute. It also replaces references to the~~ creative reach of 6.0D (and its square, 36.0D) with the equivalent using the reach attribute.

This makes it easier for modders to add new items and effects that influence the default attack reach of players, without having to use click event handlers + packets or resort to using coremods that can cause incompatibilities with other mods.

A test mod is added, "player_attack_reach_test", that adds a "Longsword" which modifies the ~~reach distance~~ attack reach of the user when held in the mainhand ~~, and most importantly, extends the user's attack reach.~~


Kovak:
Is it actually desirable to collapse melee attack distance into block reach distance - some constant?

It seems to me it would be better to have separate attributes as they are sort of totally different mechanics. For instance lets say you implement a gun as a raytracing weapon with like a 20 block range. Now the player's reach distance for other block placing activities is like 21 blocks. If you hold a torch in your offhand you would be able to place it out a ridiculous amount while holding the gun.

ChampionAsh5357:
So the entity reach distance (the thing being implemented here) is dependent on the global reach distance (the current reach distance attribute already implemented). The entity reach distance cannot be larger than the global reach distance. So, there are two methods of tackling this. The first is to make a fixed distance as done above. The second would be to make a percentage attribute that scales between the global reach distance. You could just strip out everything and make them completely separate, but that's very much invasive and doesn't serve any purpose. The reach distance is not specifically for a block, its the maximum distance the player can reach. The entity check just happens to be a portion of it.

The example you provide makes no sense in context anyways. The gun would be shooting a projectile to which it wouldn't need any distance larger than what is currently provided by default Minecraft.

Thelnfamous1:
@Kovak , I imagine you meant something more like maybe an armor that for example is designed to enhance block breaking reach, and this PR resulting in that armor also extending the entity reach. That case was made in the "ancestor" of this PR, #4736.

There certainly is a case to be made for keeping it separate, but the entity reach would still dependent on the global reach, like what @ChampionAsh5357 said above. A situation shouldn't arise where the entity reach is made greater than the global reach, and the global reach makes the additional entity reach useless.

I'm open to the alternate scaling approach, but what concerns me is that there are other places already patched that use the global reach +/- a constant, and my desire was to retain consistency with those.

Kovak:
Guns dont have to shoot projectiles many mods across the years have already implemented certain type of gun items as basically a sword item with a long reach.

Many weapon mods that introduce their own reach attribute are literally already separating these 2 things and not making their item attack calls dependent on what ya'll are calling 'global reach' because it doesn't make sense. 

The block reach distance and the item attack distance need to be fully separate or you end up with weird scenarios where either block interaction becomes too long or attack reach is artificially capped at below block reach as ya'll are proposing.

As someone who literally has one of the mods that introduces our own reach distance for item attacks, the fix proposed here would be considered undesirable and I would end up having to work around it because what if you DO want an attack reach greater than block reach.

Thelnfamous1:
Then what do you propose should be changed specifically in GameRenderer#pick to allow for the two separate cases to work properly? Currently, the global reach determines the "picked" hit result (line 229), and then the entity reach determines if the entity that may have resulted from it can be attacked (lines 238 and 261). And note that line 229 currently remains unchanged in this PR.

I understand why this matters to you, and probably other modders as well who desire separate implementations of reach, but as-is, that's not going to fly without breaking some vanilla logic. I myself have a mod that adds attack reach, and I would prefer to not have to do janky hacks to make it work, and this PR works for me.

ChampionAsh5357:
> Guns dont have to shoot projectiles many mods across the years have already implemented certain type of gun items as basically a sword item with a long reach.
> 
> Many weapon mods that introduce their own reach attribute are literally already separating these 2 things and not making their item attack calls dependent on what ya'll are calling 'global reach' because it doesn't make sense.
> 
> The block reach distance and the item attack distance need to be fully separate or you end up with weird scenarios where either block interaction becomes too long or attack reach is artificially capped at below block reach as ya'll are proposing.
> 
> As someone who literally has one of the mods that introduces our own reach distance for item attacks, the fix proposed here would be considered undesirable and I would end up having to work around it because what if you DO want an attack reach greater than block reach.

Once again, the entity reach is literally capped by the block reach in default Minecraft. The proposed fix would be to maintain that behavior as Minecraft has defined it. Patching all the places where the game hardcodes this would mean creating two independent handlers for each distance.

What you're suggesting is that the reach distance of a player is different from how far they can use a long item in a very specific use case. If I have a longsword, I could still hit a block from some distance away, the same as an entity. The usecase you're suggesting is what if the user just raytraced itself, and even then you wouldn't be using the reach distance of an item, it would be hardcoded into the check of when the item is right-clicked, not reliant on the reach distance.

Kovak:
Ya'll there's a reason the combat test snapshots introduce a separate attack reach stat instead of trying to make them piggyback off each other, and it's not nearly as difficult to do this as you're suggesting.  It is also possible to implement a longer attack reach than block reach without even doing any coremodding. You do have to do your own raytrace in the ClickInputEvent, but it might just fundamentally be that you need to do 2 raytraces to properly determine whether a block is within block reach or an entity is within attack reach. It also really doesn't matter as these ray traces are done on the client and distances simply confirmed on the server, this is never going to be some giant performance concern.

What I am suggesting is fully within the bounds of the realm of possibility for mods, been experimented with in combat snapshots by mojang themselves, and been historically used in modding for many versions (the first time I can recall seeing a recommendation about implementing a separate attack reach from block reach for longer range weapons is from mc 1.8) seems like you are just wanting to be dismissive because your current solution actually closes off possibilities rather than expanding them. 

From where I stand it seems like it would be better to just do nothing than to bind block reach and attack reach together. A proper solution IMO would perform a clean separation, but also this is literally possible to achieve without coremodding so it's probably fine for forge to just not do it either.

ChampionAsh5357:
>  seems like you are just wanting to be dismissive because your current solution actually closes of possibilities rather than expanding them.

Apologies, I'm not trying to be dismissive. My whole point is to maintain vanilla behavior since, as you said, its just as possible to do without anything by running your own ray trace. In my opinion, it makes more sense to have it such that you can modify the entity reach distance within the current attribute either by fixing it or using a percentage of that value and if modders don't want to do that, they can implement the ray trace themselves. That logically makes sense to me because the reach distance of the player, even with any item, should be bound together.

I understand you disagree and as such think they should just be two separate independent attribute which is just as possible. The difficulty is about the same as doing it this way with just a bit more local variable replacement in certain sections. However, and this is once again my opinion, the notion that your entity range of the player can be larger than the block range makes no sense. This way, even if the gun were to increase the attack range, if I were to punch with the gun, the attack range would still apply, same if I were to right click to interact with an entity. It brings up more cases where player interactions can be accessed from farther away just by holding a single item.

Those are my reasonings for either a fixed or percentage on the reach distance attribute. Ultimately its up to infamous and then eventually the reviewers to decide what is the best implementation. These are just my thoughts on the matter.

Thelnfamous1:
I'm well aware of Mojang separating the two reaches in the combat snapshots, but given that I have no idea how much they reworked their code to allow for that, I can only either offer this solution or (a slightly updated version of) #7807.

I'm also well aware of the click handler approach (mentioned in the first post), which is something I used and ultimately discarded for my mod because I didn't like how janky it was (also, the attack indicator doesn't naturally work with that approach). I'd much rather have an attribute that directly affects the entity reach (whether separate or based on block reach doesn't matter to me) and doesn't require me as a user of Forge to write a recreation of vanilla code myself for my mod.

The current vanilla code makes it so that block reach hardcaps entity reach, and unless someone wants to propose how to rework that without excessive patching, I can only leave this PR as is.

Thelnfamous1:
Okay, I'm not at the rig I use to handle the Forge repo right now so I can't test it yet or make a new commit, but upon further inspection in a regular mdk environment, I noticed something.

It seems like starting from line 251 in GameRenderer#pick, you can replace references to d0 or d1 with Math.max(d0, attack reach value) or Math.max(d1, attack reach value squared). If "attack reach value" is indeed a separate attribute and it is made greater than block reach, this should theoretically allow for attacking entities further away than the distance you can break blocks from, meaning that a separate attack reach attribute should actually work without being handicapped by block reach.

KnightMiner:
If you are going to go with a separate attribute for attack reach, I would suggest renaming the existing reach attribute. Changing the registry name is a breaking change of course, but you can change the lang name.

Thelnfamous1:
Sure. How about "Block Reach"?

patches/minecraft/net/minecraft/client/renderer/GameRenderer.java.patch
```diff
@@ -9,6 +9,46 @@
...
+@@ -246,19 +_,21 @@
+             }
+ 
+             Vector3d vector3d1 = entity.func_70676_i(1.0F);
+-            Vector3d vector3d2 = vector3d.func_72441_c(vector3d1.field_72450_a * d0, vector3d1.field_72448_b * d0, vector3d1.field_72449_c * d0);
++            double reach = java.lang.Math.max(d0, this.field_78531_r.field_71439_g.func_110148_a(net.minecraftforge.common.ForgeMod.ATTACK_REACH.get()).func_111126_e());
++            Vector3d vector3d2 = vector3d.func_72441_c(vector3d1.field_72450_a * reach, vector3d1.field_72448_b * reach, vector3d1.field_72449_c * reach);
+             float f = 1.0F;
+-            AxisAlignedBB axisalignedbb = entity.func_174813_aQ().func_216361_a(vector3d1.func_186678_a(d0)).func_72314_b(1.0D, 1.0D, 1.0D);
++            double reachSq = java.lang.Math.max(d1, this.field_78531_r.field_71439_g.func_110148_a(net.minecraftforge.common.ForgeMod.ATTACK_REACH.get()).func_111126_e() * this.field_78531_r.field_71439_g.func_110148_a(net.minecraftforge.common.ForgeMod.ATTACK_REACH.get()).func_111126_e());
```
ChampionAsh5357: You can just use `reach * reach` instead of grabbing that attribute field again.

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
```diff
@@ -210,6 +210,16 @@
                 }
  
                 f = f + f1;
+@@ -1081,7 +_,8 @@
+                      float f3 = 1.0F + EnchantmentHelper.func_191527_a(this) * f;
+ 
+                      for(LivingEntity livingentity : this.field_70170_p.func_217357_a(LivingEntity.class, p_71059_1_.func_174813_aQ().func_72314_b(1.0D, 0.25D, 1.0D))) {
+-                        if (livingentity != this && livingentity != p_71059_1_ && !this.func_184191_r(livingentity) && (!(livingentity instanceof ArmorStandEntity) || !((ArmorStandEntity)livingentity).func_181026_s()) && this.func_70068_e(livingentity) < 9.0D) {
++                        double attackReachSq = this.func_110148_a(net.minecraftforge.common.ForgeMod.ATTACK_REACH.get()).func_111126_e() * this.func_110148_a(net.minecraftforge.common.ForgeMod.ATTACK_REACH.get()).func_111126_e();
++                        if (livingentity != this && livingentity != p_71059_1_ && !this.func_184191_r(livingentity) && (!(livingentity instanceof ArmorStandEntity) || !((ArmorStandEntity)livingentity).func_181026_s()) && this.func_70068_e(livingentity) < attackReachSq) {
```
ChampionAsh5357, line 238: This can be combined into one line, a long one yes but it reduces the number of lines to add.

patches/minecraft/net/minecraft/network/play/ServerPlayNetHandler.java.patch
```diff
@@ -43,7 +43,17 @@
...
  
-@@ -1187,6 +_,7 @@
+@@ -1180,13 +_,16 @@
+       this.field_147369_b.func_226284_e_(p_147340_1_.func_241792_e_());
+       if (entity != null) {
+          double d0 = 36.0D;
+-         if (this.field_147369_b.func_70068_e(entity) < 36.0D) {
++         double attackReachSq = this.field_147369_b.func_110148_a(net.minecraftforge.common.ForgeMod.ATTACK_REACH.get()).func_111126_e() * this.field_147369_b.func_110148_a(net.minecraftforge.common.ForgeMod.ATTACK_REACH.get()).func_111126_e();
++         double creativeReachSq = (this.field_147369_b.func_110148_a(net.minecraftforge.common.ForgeMod.REACH_DISTANCE.get()).func_111126_e() + 1.0D) * (this.field_147369_b.func_110148_a(net.minecraftforge.common.ForgeMod.REACH_DISTANCE.get()).func_111126_e() + 1.0D);
++         if (this.field_147369_b.func_70068_e(entity) < (p_147340_1_.func_149565_c() == CUseEntityPacket.Action.ATTACK ? attackReachSq : creativeReachSq)) {
```
ChampionAsh5357, line 53: Once again, one line patch, it does not need to be three lines.

src/test/java/net/minecraftforge/debug/entity/player/PlayerAttackReachTest.java
```diff
@@ -0,0 +1,66 @@
+package net.minecraftforge.debug.entity.player;
```
ChampionAsh5357, line 20: Add license header.

src/test/java/net/minecraftforge/debug/entity/player/PlayerAttackReachTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.UUID;
+/**
+ * Tests if the various patches made to utilize values associated with ForgeMod.REACH_DISTANCE && ForgeMod.ATTACK_REACH works
+ * The Longsword adds a modifier to the user's attack reach, which is by default 3.0D
+ * The final attack reach can in fact be made greater than reach distance,
+ * and will work to attack from a distance further than the one you can break blocks from
+ */
+@Mod(PlayerAttackReachTest.MODID)
+public class PlayerAttackReachTest {
```
ChampionAsh5357: Forge classes use new line opening brackets.

ChampionAsh5357 (commented):
I don't think you add the lang and model files. They are not needed to verify that the logic works.

patches/minecraft/net/minecraft/client/renderer/GameRenderer.java.patch
Thelnfamous1: reach uses d0, which is a different value from just being the root of d1 which is used by reachSq. Although d1 at this point can be the square of d0, it can also become the distance squared from the camera to the block hit result

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
Thelnfamous1: Okay, I was just under the impression that I should've avoided making a really long line like that

patches/minecraft/net/minecraft/client/renderer/GameRenderer.java.patch
ChampionAsh5357: Yeah, but they're set to the same values aren't they?

src/test/java/net/minecraftforge/debug/entity/player/PlayerAttackReachTest.java
Thelnfamous1: Will do

src/test/java/net/minecraftforge/debug/entity/player/PlayerAttackReachTest.java
Thelnfamous1: Will do

patches/minecraft/net/minecraft/network/play/ServerPlayNetHandler.java.patch
Thelnfamous1: That's going to be a hell of a line

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
ChampionAsh5357: I don't believe that's the case, but I'm not 100% sure on my statement. I'm just remembering previous PRs.

patches/minecraft/net/minecraft/client/renderer/GameRenderer.java.patch
Thelnfamous1: d0 is either the pick reach (5.0) or the creative reach (6.0). d1 is equal to d0 until d1 gets squared, and then potentially changed to the distance sq from the camera to the block hit result if it's not null. 

Do you think that the distance sq from the camera to the block hit result is going to be the same as d0 squared? I'm under the impression that they can be different.

ChampionAsh5357: I'm using an incorrect given for my claim since d0 shouldn't be a part of the attack reach check. Apologies.

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
TheCurle: This is a gigantic mess! Split it up for readability, jesus.

patches/minecraft/net/minecraft/network/play/ServerPlayNetHandler.java.patch
TheCurle: Mess! Split it up.

Thelnfamous1: Done.

patches/minecraft/net/minecraft/entity/player/PlayerEntity.java.patch
Thelnfamous1: Done.

TheCurle:
```
* What went wrong:

Execution failed for task ':forge:applyPatches'.

> Patches failed to apply.
```

Seems you didn't genPatches properly.

sciwhiz12:
Hello! @Thelnfamous1, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

sciwhiz12:

As a note, PR #8478 seems to implement a similar feature to this PR.

PR #7809, opened 2021-06-07
[1.17] Add ability to change mouse cursor image

Nightenom:
Add methods for managing cursor image. Uses GLFW cursor api to change cursor image to one of the standard shapes or to ResLoc, also adds a method for resetting to default value (usually an arrow).
Exposes address of NativeImage byte buffer.
No screenshots provided as Windows doesn't capture cursor in screenshots.
Example usages: hand cursor over buttons, drag cursor in map elements


diesieben07:
Question is, why does this need to be in Forge? Forge rarely adds "utility" things like this, which can easily be done in a library mod without compatibility concerns.

Nightenom:
> Question is, why does this need to be in Forge? Forge rarely adds "utility" things like this, which can easily be done in a library mod without compatibility concerns.

True, although I would say Forge should provide utility things for native APIs that aren't straightforward. If more of lead says no to this, I can just reduce this to the patch only

KnightMiner:
Why is the patch needed at all? You could accomplish the exact same thing with an access transformer.

TheCurle:
This feels like an *extremely* niche change. 

I have an issue with the large manager class that touches opengl/glfw classes directly, when it's clear that Mojang want to go a more compatible route.

We'll defer this change to the 1.17 cycle, when we have more time to mull it over. As it stands, i don't think this'll be accepted as-is.

sciwhiz12:
Closing this PR as the above users, the PR author (communicated via Discord), and I agree that this feature is too niche to be present in Forge.

PR #7810, opened 2021-06-07
LivingBreatheEvent and LivingDrownEvent

Meldexun:
This PR adds three new events to allow modders to control if a living entity can breathe and how air is consumed/refilled.
With the LivingBreatheEvent modders could easily prevent players from breathing in something like a moon dimension or allow players to breathe everywhere while wearing a specific item.
Withe the LivingConsumeAirEvent and LivingRefillAirEvent modders could change how air is consumed/refilled or add an item which can hold air which will be refilled when the players air is full.

Currently a problem could be something like this:
When a modder adds a moon dimension where the player can't breathe then he has to revert the refilling of the air supply and the make the player consume air. This would be a lot cleaner and compatibility between mods would be easier to achieve with the LivingBreatheEvent.

I also added two test mods which each add an item with make usage of the events.


patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
```diff
@@ -33,15 +33,65 @@
...
++
++            if (this.func_70086_ai() == -20) {
++               this.func_70050_g(0);
++               Vector3d vector3d = this.func_213322_ci();
++
++               for (int i = 0; i < 8; ++i) {
++                  double d2 = this.field_70146_Z.nextDouble() - this.field_70146_Z.nextDouble();
++                  double d3 = this.field_70146_Z.nextDouble() - this.field_70146_Z.nextDouble();
++                  double d4 = this.field_70146_Z.nextDouble() - this.field_70146_Z.nextDouble();
++                  this.field_70170_p.func_195594_a(ParticleTypes.field_197612_e, this.func_226277_ct_() + d2, this.func_226278_cu_() + d3, this.func_226281_cx_() + d4, vector3d.field_72450_a, vector3d.field_72448_b, vector3d.field_72449_c);
```
ChampionAsh5357: This is really invasive. You do not need this much of a rewrite to handle this logic. You can simply add the event below #isAlive, post it, and then store the result in the flag to pass into multiple places. You also reverse the logic of when the entity stops riding which in vanilla's case doesn't break behavior but could in other mods. You can also use `ForgeHooks` to handle the actual event creation bring this down to just three lines added with a few more modified.

src/test/java/net/minecraftforge/debug/entity/living/LivingAirEventTest.java
```diff
@@ -0,0 +1,96 @@
+package net.minecraftforge.debug.entity.living;
```
ChampionAsh5357: Same applies here for license header and spacing.

src/main/java/net/minecraftforge/event/entity/living/LivingBreatheEvent.java
```diff
@@ -0,0 +1,23 @@
+package net.minecraftforge.event.entity.living;
```
ChampionAsh5357, line 20: Add license header.

src/main/java/net/minecraftforge/event/entity/living/LivingBreatheEvent.java
```diff
@@ -0,0 +1,23 @@
...
+
+import net.minecraft.entity.LivingEntity;
+
+public class LivingBreatheEvent extends LivingEvent
+{
+   private boolean canBreathe;
```
ChampionAsh5357: Forge classes are four spaces per tab, not three.

src/main/java/net/minecraftforge/event/entity/living/LivingConsumeAirEvent.java
```diff
@@ -0,0 +1,23 @@
+package net.minecraftforge.event.entity.living;
```
ChampionAsh5357: Same applies here for license header and spacing.

src/main/java/net/minecraftforge/event/entity/living/LivingRefillAirEvent.java
```diff
@@ -0,0 +1,23 @@
+package net.minecraftforge.event.entity.living;
```
ChampionAsh5357: Same applies here for license header and spacing.

src/test/java/net/minecraftforge/debug/entity/living/LivingAirEventTest.java
```diff
@@ -0,0 +1,96 @@
...
+import net.minecraftforge.api.distmarker.Dist;
+import net.minecraftforge.api.distmarker.OnlyIn;
+import net.minecraftforge.event.RegistryEvent;
+import net.minecraftforge.event.entity.living.LivingConsumeAirEvent;
+import net.minecraftforge.event.entity.living.LivingRefillAirEvent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
+
+@Mod(LivingAirEventTest.MODID)
+public class LivingAirEventTest {
```
ChampionAsh5357: Opening brackets are on new line for Forge classes.

src/test/java/net/minecraftforge/debug/entity/living/LivingBreatheEventTest.java
```diff
@@ -0,0 +1,46 @@
+package net.minecraftforge.debug.entity.living;
```
ChampionAsh5357, line 20: Same applies here for license header, spacing, and open bracket new line.

patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
```diff
@@ -33,15 +33,65 @@
...
+-               this.func_184210_p();
++               this.func_70097_a(DamageSource.field_76369_e, 2.0F);
              }
-          } else if (this.func_70086_ai() < this.func_205010_bg()) {
+-         } else if (this.func_70086_ai() < this.func_205010_bg()) {
+-            this.func_70050_g(this.func_207300_l(this.func_70086_ai()));
++         } else {
++            net.minecraftforge.event.entity.living.LivingRefillAirEvent event1 = new net.minecraftforge.event.entity.living.LivingRefillAirEvent(this, this.func_207300_l(0));
++            net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(event1);
++            this.func_70050_g(MathHelper.func_76125_a(this.func_70086_ai() + event1.getAmount(), -20, this.func_205010_bg()));
```
ChampionAsh5357: Same applies here that this could all go in `ForgeHooks` to reduce the amount of calls.

patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
```diff
@@ -33,15 +33,65 @@
...
 -            if (!this.field_70170_p.field_72995_K && this.func_184218_aH() && this.func_184187_bx() != null && !this.func_184187_bx().func_205710_ba()) {
-+            if (!this.field_70170_p.field_72995_K && this.func_184218_aH() && this.func_184187_bx() != null && !this.func_184187_bx().canBeRiddenInWater(this)) {
-                this.func_184210_p();
+-               this.func_184210_p();
++               this.func_70097_a(DamageSource.field_76369_e, 2.0F);
              }
-          } else if (this.func_70086_ai() < this.func_205010_bg()) {
+-         } else if (this.func_70086_ai() < this.func_205010_bg()) {
+-            this.func_70050_g(this.func_207300_l(this.func_70086_ai()));
++         } else {
```
ChampionAsh5357: Fires event unnecessarily. The point of the event is to refill. If you are not or your air meter is full, you should have no reason to fire the event.

ChampionAsh5357 (commented):
Question for you: why do you need three events to handle this logic? I'm pretty sure this could be done in a single cancelable event. For example, if the event is canceled, execute Forge logic where it checks if it can or can not breathe and then sets the amount. Calling three events every tick for something that could be done in one seems resource consuming.

Meldexun:
So I formatted the forge classes and added the missing license headers.

<br>

>This is really invasive. You do not need this much of a rewrite to handle this logic. You can simply add the event below #isAlive, post it, and then store the result in the flag to pass into multiple places.

This is basically the vanilla code:
```Java
if (isUnderWater && !notInBubbleColumn) {
    if (!canBreatheUnderwater && !hasWaterBreathing && !damageIsDisabled) {
        // consume air
    }
    // check if vehicle can be ridden underwater
} else if (air < maxAir) {
    // refill air
}
```
So to insert the events the only way I could think of was to:
- move the vehicle check out of that statement
- merge the first two if statements

You are free to help me making this less invasive though I didn't see a way to keep the original structure in tact. Which is the reason why it is how it is now.

<br>

>Question for you: why do you need three events to handle this logic? I'm pretty sure this could be done in a single cancelable event. For example, if the event is canceled, execute Forge logic where it checks if it can or can not breathe and then sets the amount. Calling three events every tick for something that could be done in one seems resource consuming.

(two events every tick not three)
For example maybe I want to add an item which can store air.
I could use the consume event to first extract air out of my item and only when the item is empty the players air supply is used.
And the refill event could be used to fill the item once the players air supply is full (which is also the reason why the `air < maxAir` check was removed).

>Fires event unnecessarily. The point of the event is to refill. If you are not or your air meter is full, you should have no reason to fire the event.

As described above

ChampionAsh5357:
> You are free to help me making this less invasive though I didn't see a way to keep the original structure in tact. Which is the reason why it is how it is now.

You can just create a method within `ForgeHooks` to handle the logic for you and then just return what you need. That reduces those three line patches to a single line or even less in some cases.

> And the refill event could be used to fill the item once the players air supply is full (which is also the reason why the air < maxAir check was removed).

But...that would never do anything. `#setAirSupply` is always hardcoded to choose the minimum between the added value and max air. Sure, you could add logic within the event to consume the item if used, but it's quite pointless to remove the check if literally nothing will happen when its called. 

This is why I said the three event calls could be reduced down to one. That one event would be cancelable and set if the user can breathe and the amount to add/remove. When canceled, the event will execute the forge data retrieved from the event.

Meldexun:
Hm I guess this could work.
Should I rather copy the `if (air == -20) then DROWN` into the ForgeHooks method or do more changes to the vanilla class?

ChampionAsh5357:
You should always do the least amount of lines added/modified as possible within the vanilla class.

Meldexun:
So I update this again. Now it is only one event which is fired inside a ForgeHooks method and thus the changes required to the vanilla class are very small now.

sciwhiz12:
Hello! @Meldexun, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

Meldexun:
Hi, yes I would like to continue this PR. I have created a new branch in my fork for MC 1.18 https://github.com/Meldexun/MinecraftForge/tree/living-breathe-event-1.18. So how should I proceed now? Should I create a new pull request?

sciwhiz12:
@Meldexun, you have two options here:
 - Force-push the commits from that new branch to the branch of this PR, then [change the base branch of this PR][base-branch] to the `1.18.x` branch. I personally prefer this, as this preserves the history of the PR.
 - Create a new PR with that new branch, and close this PR as superseded by the new pull request.

[base-branch]: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

Meldexun:
I updated everything to MC 1.18 now. Also I moved pretty much all changes into the `ForgeHooks#onLivingBreathe(LivingEntity, int, int)` method and added a living drown event which allows modders to control how fast an entity takes damage while drowning or cancel the event when the entity is suffocating in a space dimension to not spawn the bubble particles and to not play the underwater drown sound.

TheCurle:
This may be easier to implement in 1.19 with the new fluid API. 

I'm going to close this old one for inactivity (which is entirely our fault!) and we can continue if you'd like in a new PR.

PR #7813, opened 2021-06-09
Remove ModelResourceLocation patch in favor of SAS entry

SizableShrimp:
Seems like a patch file that was never turned into a SAS entry. Added SAS entry and confirmed it works in setup.


LexManos:
You didnt run checkSAS but that's not a big deal.

PR #7816, opened 2021-06-13
Let Trapdoors be climbable for custom ladders

### LINKED ISSUE #7775: Custom ladders don't work properly in combination with trapdoors
**Minecraft Version:** 1.16.5

**Forge Version:** 1.16.5-36.1.0

**Description of issue:**
Modded ladder blocks don't allow climbing through a trapdoor placed at the top. This is despite having the ladder block added to the `blocks/climbable` tag, so not a duplicate of [7692](https://github.com/MinecraftForge/MinecraftForge/issues/7692).

image

With a ladder and trapdoor placed in the above configuration, when the trapdoor is opened, vanilla ladders allow you to climb up through the trapdoor. But modded ladders don't.

This appears to happen because the trapdoor block considers itself to be ladder-climbable if a ladder block is placed below - but it only checks the vanilla ladder. The relevant patched code in `TrapDoorBlock.java`:

```
   //Forge Start
   @Override
   public boolean isLadder(BlockState state, net.minecraft.world.IWorldReader world, BlockPos pos, net.minecraft.entity.LivingEntity entity) {
      if (state.get(OPEN)) {
         BlockState down = world.getBlockState(pos.down());
         if (down.matchesBlock(net.minecraft.block.Blocks.LADDER))
            return down.get(LadderBlock.FACING) == state.get(HORIZONTAL_FACING);
      }
      return false;
   }
   //Forge End
```

It would be possible to "fix" this by replacing the `matchesBlock(net.minecraft.block.Blocks.LADDER)` call with one that checks if the block is in the `climbable` tag instead. But that would also include other blocks in the tag, vines & Nether vines etc. and a quick check on a vanilla-only installation tells me that wouldn't match vanilla functionality.

cech12:
This PR fixes #7775 and changes the `isLadder` method of a Trapdoor to check, if the block below it is an instance of `LadderBlock` instead of being the vanilla Ladder block to support modded ladders.

Using the `climbable` tag is not useful here, because vines and other climbable blocks are also listed there.

Another possibility to fix the issue would be to invent a new tag called `makes_trapdoor_climbable` (or something similar), where blocks for this functionality are listed. But the additional check, that the facing of the ladder and the trapdoor must be equal, convinced me to do the `instanceof LadderBlock` check.

### DIFF
--- a/patches/minecraft/net/minecraft/block/TrapDoorBlock.java.patch
+++ b/patches/minecraft/net/minecraft/block/TrapDoorBlock.java.patch
@@ -10,7 +10,7 @@
 +   public boolean isLadder(BlockState state, net.minecraft.world.IWorldReader world, BlockPos pos, net.minecraft.entity.LivingEntity entity) {
 +      if (state.func_177229_b(field_176283_b)) {
 +         BlockState down = world.func_180495_p(pos.func_177977_b());
-+         if (down.func_203425_a(net.minecraft.block.Blocks.field_150468_ap))
++         if (down.func_177230_c() instanceof LadderBlock)
 +            return down.func_177229_b(LadderBlock.field_176382_a) == state.func_177229_b(field_185512_D);
 +      }
 +      return false;
--- a/patches/minecraft/net/minecraft/block/TrapDoorBlock.java.patch
+++ b/patches/minecraft/net/minecraft/block/TrapDoorBlock.java.patch
@@ -9,9 +9,9 @@
 +   @Override
 +   public boolean isLadder(BlockState state, net.minecraft.world.IWorldReader world, BlockPos pos, net.minecraft.entity.LivingEntity entity) {
 +      if (state.func_177229_b(field_176283_b)) {
-+         BlockState down = world.func_180495_p(pos.func_177977_b());
-+         if (down.func_177230_c() instanceof LadderBlock)
-+            return down.func_177229_b(LadderBlock.field_176382_a) == state.func_177229_b(field_185512_D);
++         BlockPos downPos = pos.func_177977_b();
++         BlockState down = world.func_180495_p(downPos);
++         return down.func_177230_c().makesOpenTrapdoorAboveClimbable(down, world, downPos, state);
 +      }
 +      return false;
 +   }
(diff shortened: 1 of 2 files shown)


XFactHD:
Adding a method to IForgeBlock that by default does this instanceof check and can be overriden by mod blocks that want to be a ladder might be a better option.

cech12:
There is already an `isLadder` method which can be overridden to be a ladder. This PR changes this method for the Trapdoor block to be compatible with custom ladders.

XFactHD:
Right, I completely missed that.
Though I would argue that you should call isLadder on the block below and check that it is not a trapdoor instead of only checking for instanceof Ladder because IMO your current implementation makes isLadder completely useless for custom ladders in conjunction with trapdoors.

cech12:
No, I only changed **one** line in this PR to support the climbing mechanism of trapdoors if a custom ladder is below. 
The standard implementation of the isLadder method checks if the block is part of the climbable tag. So, calling the isLadder method is not the way to go here, because there are other blocks like vines which return also true there. 
The isLadder method is used in other place to check if a block is climbable. So, it is not useless. Maybe the name is a bit confusing. 
Please read the discussion of the related issue.

XFactHD:
I still don't see how that would fix this issue for custom ladders that do not inherit from LadderBlock and instead use isLadder to communicate that they are a ladder.
Maybe isLadder should be changed in 1.17 to only reflect that something is a ladder and not that it is a generic climbable thing.

KnightMiner:
That would not be ideal, I make use of `isLadder` on a block to denote its conditionally climbable in a similar way to the climbable trapdoors, so that would be a pretty big loss of functionality if that is required to be equivalent to vanilla ladders. At that point, make a second method.

XFactHD:
Making a second method is basically what I am talking about

cech12:
@XFactHD I understand your point, that blocks that not extend `LadderBlock` were not supported by my first commit. I added a new method to `IForgeBlock` that can be overridden by such blocks. 

@KnightMiner If you have a block that should act similar like a Trapdoor, you can check everthing you need inside the `isLadder` method.

XFactHD:
Looks good to me

PR #7817, opened 2021-06-14
Make DimensionSettings easier to use

Won-Ton:
### What?
This PR makes it easier for mods to create & register `DimensionSettings` for their dimension.

### Why?
`DimensionSettings` hold the settings for the dimension (surprise!) including the structure spacing settings specific to that dimension. By registering these settings in the `WorldGenRegistries` it makes their settings visible for other mods to add, remove, or change things like structure settings on a per-dimension basis.

Registering the settings this way allows for nicer, non-dependent compatability between mods, and for end-users/modpack creators are still overridable via datapack.

### How?
A test mod is provided showing how a datapack'd dimension can reference `DimensionSettings` that it has registered to the `WorldGenRegistries`. It also shows how those settings can be configured by other mods to add structure settings that are specific for that dimension.


Mysterious-Dev (approved):

Won-Ton:
hi, could this be labelled for triage please :]

PR #7822, opened 2021-06-18
Add custom TextureAtlasSprite loaders

diesieben07:
Adds a system for mods to specify custom texture loaders which can be requested by the resource pack through a metadata section in the `.mcmeta` file.
A test mod is included which creates a randomly flashing texture.


src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -822,4 +825,17 @@ public static boolean hasBiomeGeneratorSettingsOptionsScreen(Optional<BiomeGener
     {
         return Optional.of(ForgeWorldTypeScreens.getDefaultGenerator());
     }
+
+    public static TextureAtlasSprite loadTextureAtlasSprite(
```
nikita488, line 830: Should have Nullable annotation

diesieben07: Fixed

KnightMiner:
What is an example usecase for this? I don't expect a randomly flashing texture is your end goal here.

Also, is it possible with this PR to have a texture mcmeta without having the .png, say if you wanted to redirect to another location or generate the texture? I suppose in either case you could ignore the .png contents, but would be nice to not need to provide one in a few cases (not sure if that is handled by this PR)

nikita488:
The example use case for this is procedurally generated textures (animation) like this from my mod:
image
The pixels are recalculated every frame:
[CloudSprite.java](https://github.com/nikita488/ZYCraft/blob/1.16.x/src/main/java/nikita488/zycraft/client/texture/CloudSprite.java#L87)

diesieben07:
> Also, is it possible with this PR to have a texture mcmeta without having the .png

Unfortunately not, however that is a deliberate decision to keep the patch size from being excessive. The whole loading of textures is tied to looking up the `.png` resource and bailing out with a "not found" exception if it does not exist. The loading of `.mcmeta` files is tied to their corresponding resource. All this would have to be short-circuited, leading to basically the entire texture loading code being changed or repeated in Forge code, just to first check if a .mcmeta file exists without the corresponding png.

diesieben07:
> What is an example usecase for this?

Another use-case is a texture which changes appearance based on certain conditions that are not animation. E.g a texture which looks different at night and day or a block which looks different based on dimension, etc. etc.

PR #7824, opened 2021-06-19
[1.16] Allow blocks other than beds or respawn anchors to act as respawn points 

diesieben07:
This adds a new method `getRespawnPosition` to blocks which supersedes `getBedSpawnPosition`. `getRespawnPosition` is a more general hook that is always called when a player is attempting to respawn at a block, exempt from respawn anchor or bed rules.
A fallback for `getBedSpawnPosition` is included if `isBed` returns true and `getBedSpawnPosition` is deprecated.
A test mod is included which adds a simple respawn-anchor-like block that allows the player to set their respawn point anywhere without condition.


50ap5ud5 (approved):
Thanks diesie, this is a great usecase, I have usecases for using this hook too. Changes are well documented and cleanly executed.

PR #7825, opened 2021-06-20
Expand Elytra Flight, add IForgeAttribute and BooleanAttribute

Thelnfamous1:
This PR aims to expand upon Forge's patching for custom elytra flight logic by adding a new attribute called **FALL_FLYING** which is by default 0 and can be set to virtually any positive or negative value. When the attribute value is greater than 0, the entity can engage in elytra flight. The actual value is irrelevant when above 0, as greater than 0 is simply treated as "This entity can elytra fly". 

In addition, a new attribute is added called **FALL_FLIGHT_SPEED** which is by default 0 and can be set up to 1024. When made greater than 0, this attribute is used to allow for intrinsic flight boosting akin to using a Firework Rocket for each tick while elytra flying. The boost amount is scaled by the attribute's value, and the boost logic for a value of 1 is the exact same as for being boosted by a firework rocket.

A test mod is added that adds various effects, potions, enchantments and items that either allow, deny, or disable fall flight via the attribute.

**Update:**
This PR now adds **IForgeAttribute**, implements it into Attribute, and has patched ModifiableAttributeInstance to be capable of handling custom operation logic for an Attribute implementation. **BooleanAttribute** now serves as an example of this, with **FALL_FLYING** being an instance of it. This decision was made in order to simplify logic for the **FALL_FLIGHT** attribute, by making it work like a boolean. 

Because of this, if two ADDITION or MULTIPLY_BASE modifiers are applied at once for **FALL_FLIGHT** with opposing values (1 and -1), they will cancel each other out, and a MULTIPLY_TOTAL modifier with a value less than or equal to -1 will disable it completely. 

In addition, if an Attribute is designated as a custom one via IForgeAttribute#isCustom, then ItemStack#getTooltipList and PotionUtils#getPotionTooltip will add customized tooltip info for that Attribute to the tooltips sent to the client in place of vanilla's handling. For a **BooleanAttribute**, the client sees that the item/potion allows, denies, or disables whatever action is involved with the attribute.

_An end-user will be able to determine whether or not they can fall fly simply by visually discerning if they have more items/effects active that allow fall flight than ones that deny it, or if they have any items/effects that disable it completely._

**Update 2:**
**Without breaking any existing functionality**, ItemStack#canElytraFly and ItemStack#elytraFlightTick have been folded into the new Forge Hook **canFallFly**. Any ItemStack in the chestplate slot that allows fall flight "adds" 1 to the attribute value in order to potentially let you fly, unless the attribute value was already multiplied by 0, in which case it won't.


Thelnfamous1:
Since fall flight can show up as displayed attribute on an item, I figure it may be worth looking into patching how it appears in the client, so it doesn't say something strange like "+1 Fall Flight" and instead says "Enables Fall Flight"

Thelnfamous1:
Would like feedback on whether or not the chestplate checks and the Elytra itself should be folded into this new system, were it to be merged.

src/test/java/net/minecraftforge/debug/entity/living/ExpandedElytraFlightTest.java
```diff
@@ -0,0 +1,221 @@
...
+            EFFECTS.register(modEventBus);
+            POTIONS.register(modEventBus);
+            ENCHANTMENTS.register(modEventBus);
+            ITEMS.register(modEventBus);
+            MinecraftForge.EVENT_BUS.register(this);
+        }
+    }
+
+    @SubscribeEvent
+    public void onLivingUpdate(LivingEvent.LivingUpdateEvent event)
```
KnightMiner, line 129: You could use the item attribute event here for the record, cleaner than manually changing the attributes. Also means it will show up in the tooltip, to be a good test case for how these boolean attributes display.

Thelnfamous1: The modifiers added here during this event are for the enchantments, using the same logic as how vanilla adds the Soul Speed enchantment's modifier to the player. Everything else shows up in the tooltip.

src/main/java/net/minecraftforge/entity/attributes/BooleanAttribute.java
```diff
@@ -0,0 +1,71 @@
...
+        return valueIn > 0 ? 1 : 0;
+    }
+
+    @Override
+    public boolean isBooleanAttribute() {
+        return true;
+    }
+
+    /*
+    To reduce headache, every one of these calculations is treated as addition.
```
TheIllusiveC4: I think the nuance of having multiplicative values should be preserved. Otherwise, you could never add an effect that disables fall flying (or any other boolean attribute) without knowing that, with enough sources given to them, a player could just overcome that behavior. Being able to simply multiply the attribute by 0 would ensure that the intended behavior is carried through.

Thelnfamous1: That’s a valid point, and an idea I had for being able to customize how an attribute is displayed in a tooltip would be effective in conveying what exactly is being done by the attribute modifier in that case.

Functionally though, I imagine there would be no significant difference between a MULTIPLY_BASE and MULTIPLY_TOTAL for a Boolean Attribute. And multiplying by a value greater than 0 simply wouldn’t (shouldn't?) accomplish anything.

Thelnfamous1: Should be resolved with latest commit

brandon3055:
I have not looked too closely at the code but I have been testing out your entity attributes. So far the `fall_flight` attribute seems to work exactly as expected but I do have some concerns regarding the `fall_flying_speed` attribute.

*<Disclaimer: The following is my personal opinion and I would really like others to pitch in with their own opinions on this>*

I understand you were trying to mimic the effect of the firework rocket and you have succeeded with that.
The problem is when using a rocket your no longer "elytra flying" you're basically just traveling in a straight line at a fixed speed. This is fine if you just want a quick boost in altitude but your very limited in how you can implement it. 
For example say you want "less boost" something that gives you enough of a push to keep flying but without instantaneously breaking the sound barrier every time you press your "boost key"
You can't really do that with this because setting a lower speed just slows you down so you're now just flying in a straight line in slow motion.

There are a lot of ways to implement a speed / boost mechanic but my suggestion would either be to just let mods handle it themselves in their own way or rather than using a "fixed" velocity you could boost / add to the players existing velocity. 
For example this is how i handle this in my own mod.
```java
double flightSpeed = 1; //flightModule == null ? 0 : flightModule.getElytraBoost();
if (flightSpeed > 0 && player.isFallFlying() && entity.isSprinting()) {
    double speed = 1.5D * flightSpeed;
    double accel = 0.01 * flightSpeed;
    Vector3d look = entity.getLookAngle();
    Vector3d motion = entity.getDeltaMovement();
    entity.setDeltaMovement(motion.add(
            look.x * accel + (look.x * speed - motion.x) * accel,
            look.y * accel + (look.y * speed - motion.y) * accel,
            look.z * accel + (look.z * speed - motion.z) * accel
    ));
}
```
So with this `speed` is a *theoretical* max velocity assuming zero drag and `accel` is the rate at which you increase speed. The math works out so that the closer you get to your target velocity the less acceleration is applied. In this example, I am calculating both speed and acceleration from a single "flightSpeed" input but there are a lot of ways this can be handled. 
The effect given by this just feels a lot more natural it's generally a lot more enjoyable. 

Perhaps the biggest issue here is that there are so many ways to do this and no matter what you go with there will probably be someone who wants it a different way. 
But at the end of the day so long as the `fall_flight` attribute continues to work the way it does I'm personally not too concerned about how the `fall_flying_speed` attribute works because I can and probably will just use my own implementation so I can have better control over how it works. The only real concern I have is other mods getting a little too excited, throwing around the `fall_flying_speed` attribute, and ultimately breaking my own elytra functionality. But that's just an example of an inter-mod compatibility issue and there's not much you can really do to avoid those.

Thelnfamous1:
Anymore thoughts about the fall flight speed attribute? I'm not too concerned about keeping it in, I just thought it might be useful for some people to manipulate the velocity via the attribute at the right time.

brandon3055:
> Anymore thoughts about the fall flight speed attribute? I'm not too concerned about keeping it in, I just thought it might be useful for some people to manipulate the velocity via the attribute at the right time.

My opinion: It might be best to just leave it out. That would simplify the PR and there are plenty of ways people can implement their own speed mechanic including using their own custom attributes.

Thelnfamous1:
Removed the fall flying speed attribute. If someone from Triage can restart the jenkins task, it should be ready to merge based off the previous successful checks (the only thing I changed in the failed commit was removing the fall flying speed attribute localization in the lang file).

sciwhiz12:
Hello! @Thelnfamous1, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

sciwhiz12:

PR #7827, opened 2021-06-20
[1.16] Fix WeightedBakedModel not delegating getModelData call

diesieben07:
`WeightedBakedModel` did not delegate `getModelData` to the wrapped model, making it so custom models cannot have custom model data.

### DIFF
--- a/patches/minecraft/net/minecraft/client/renderer/model/WeightedBakedModel.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/model/WeightedBakedModel.java.patch
@@ -32,7 +32,7 @@
     public boolean func_177556_c() {
        return this.field_177566_c.func_177556_c();
     }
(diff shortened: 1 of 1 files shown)


KnightMiner:
Me and tterrag talked about this about a year ago. The concern is getQuads is random sensitive, while the model data is not. Means no matter which model is randomly chosen, model data from the first will be used.

In my opinion, it's less of a problem than having no model data access, but it's worth documenting.

diesieben07:
Yeah, in theory we could actually choose the correct model, because it's based on block position. Question is, is the additional call to setSeed and nextLong on the Random worth it or not.
If not - where would we document this? On `IModelData`? On `WeightedBakedModel`?

Additionally: the same is true for things like `handlePerspective`, which already have this same behavior in an equally undocumented fashion.

PR #7828, opened 2021-06-20
Add the ability to pass along IRecipe hint in ItemCraftedEvent

shartte:
This allows listeners to avoid having to scan the recipe manager to find the recipe that was potentially used to craft the item.

This has been a major performance concern in mods such as CraftTweaker and previously ~~Witchery~~Wizardry, since they scanned the entire list of available recipes against the crafting inventory to figure out which recipe was being crafted. That resulted in performance degradation for fast autocrafting mods that still fire this event (i.e. AE2).

By passing along the recipe that was used - even if optional - such mods now have a fast-path available and only need to scan the registry as a fallback.

This is more useful for 1.17 since most mods are probably not going to consider this new hint for 1.16. I've commented on the existing ctor that now forwards with a null IRecipe that it should be removed in 1.17 since mod authors who fire this event should at least consider the problem.

### DIFF
--- a/patches/minecraft/net/minecraft/inventory/container/CraftingResultSlot.java.patch
+++ b/patches/minecraft/net/minecraft/inventory/container/CraftingResultSlot.java.patch
@@ -1,10 +1,14 @@
 --- a/net/minecraft/inventory/container/CraftingResultSlot.java
 +++ b/net/minecraft/inventory/container/CraftingResultSlot.java
-@@ -43,6 +_,7 @@
+@@ -43,6 +_,11 @@
     protected void func_75208_c(ItemStack p_75208_1_) {
        if (this.field_75237_g > 0) {
           p_75208_1_.func_77980_a(this.field_75238_b.field_70170_p, this.field_75238_b, this.field_75237_g);
-+         net.minecraftforge.fml.hooks.BasicEventHooks.firePlayerCraftingEvent(this.field_75238_b, p_75208_1_, this.field_75239_a);
++         net.minecraft.item.crafting.IRecipe<?> recipe = null;
++         if (this.field_75224_c instanceof IRecipeHolder) {
++            recipe = ((IRecipeHolder)this.field_75224_c).func_193055_i();
++         }
++         net.minecraftforge.fml.hooks.BasicEventHooks.firePlayerCraftingEvent(this.field_75238_b, p_75208_1_, this.field_75239_a, recipe);
        }
  
        if (this.field_75224_c instanceof IRecipeHolder) {
--- a/patches/minecraft/net/minecraft/inventory/container/CraftingResultSlot.java.patch
+++ b/patches/minecraft/net/minecraft/inventory/container/CraftingResultSlot.java.patch
@@ -1,14 +1,10 @@
 --- a/net/minecraft/inventory/container/CraftingResultSlot.java
 +++ b/net/minecraft/inventory/container/CraftingResultSlot.java
-@@ -43,6 +_,11 @@
+@@ -43,6 +_,7 @@
     protected void func_75208_c(ItemStack p_75208_1_) {
        if (this.field_75237_g > 0) {
           p_75208_1_.func_77980_a(this.field_75238_b.field_70170_p, this.field_75238_b, this.field_75237_g);
-+         net.minecraft.item.crafting.IRecipe<?> recipe = null;
-+         if (this.field_75224_c instanceof IRecipeHolder) {
-+            recipe = ((IRecipeHolder)this.field_75224_c).func_193055_i();
-+         }
-+         net.minecraftforge.fml.hooks.BasicEventHooks.firePlayerCraftingEvent(this.field_75238_b, p_75208_1_, this.field_75239_a, recipe);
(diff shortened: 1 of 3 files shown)


src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -479,6 +497,15 @@ public IInventory getInventory()
         {
             return this.craftMatrix;
         }
+
+        /**
+         * @return Teturns the recipe that was used to craft the item. This can be null in case no
```
sciwhiz12: Nitpicky:
```suggestion
         * @return Returns the recipe that was used to craft the item. This can be null in case no
```

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -479,6 +497,15 @@ public IInventory getInventory()
...
+         * particular recipe was used, or the recipe is unknown.
+         */
+        @Nullable
+        public IRecipe<?> getRecipe() {
```
sciwhiz12: Braces on newlines, please.

src/main/java/net/minecraftforge/fml/hooks/BasicEventHooks.java
```diff
@@ -59,9 +60,15 @@ public static void firePlayerItemPickupEvent(PlayerEntity player, ItemEntity ite
         MinecraftForge.EVENT_BUS.post(new PlayerEvent.ItemPickupEvent(player, item, clone));
     }
 
+    // TODO 1.17 REMOVE
     public static void firePlayerCraftingEvent(PlayerEntity player, ItemStack crafted, IInventory craftMatrix)
```
sciwhiz12: Same reasoning as the constructor, mark as `@Deprecated` and optionally add a small remark to ues the new method instead.

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -462,11 +463,28 @@ public ItemEntity getOriginalEntity() {
         @Nonnull
         private final ItemStack crafting;
         private final IInventory craftMatrix;
+        @Nullable
+        private final IRecipe<?> recipe;
+
+        // TODO 1.17 REMOVE
         public ItemCraftedEvent(PlayerEntity player, @Nonnull ItemStack crafting, IInventory craftMatrix)
```
sciwhiz12: Possibly mark this constructor as `@Deprecated`, to further discourage any current callers and alert them to change to the other constructor. Also, might be worth adding a small note that just remarks "see the other constructor".

sciwhiz12 (changes requested):

patches/minecraft/net/minecraft/inventory/container/CraftingResultSlot.java.patch
```diff
@@ -1,10 +1,14 @@
...
+@@ -43,6 +_,11 @@
     protected void func_75208_c(ItemStack p_75208_1_) {
        if (this.field_75237_g > 0) {
           p_75208_1_.func_77980_a(this.field_75238_b.field_70170_p, this.field_75238_b, this.field_75237_g);
-+         net.minecraftforge.fml.hooks.BasicEventHooks.firePlayerCraftingEvent(this.field_75238_b, p_75208_1_, this.field_75239_a);
++         net.minecraft.item.crafting.IRecipe<?> recipe = null;
++         if (this.field_75224_c instanceof IRecipeHolder) {
++            recipe = ((IRecipeHolder)this.field_75224_c).func_193055_i();
++         }
++         net.minecraftforge.fml.hooks.BasicEventHooks.firePlayerCraftingEvent(this.field_75238_b, p_75208_1_, this.field_75239_a, recipe);
```
ChampionAsh5357: Couldn't this patch be mitigated by moving the check if it is a recipe holder into the event hook and just adding the inventory container itself as a parameter?

shartte: Is BasicEventHooks intended to be used by modders or just internal infrastructure to move code out of the patches?
If it's not intended to be used by modders, moving the cast there would make sense.

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
shartte: Done

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
shartte: Done.

src/main/java/net/minecraftforge/fml/hooks/BasicEventHooks.java
shartte: Dne.

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
shartte: Done

patches/minecraft/net/minecraft/inventory/container/CraftingResultSlot.java.patch
shartte: Alright. Done. The patch is now a one-liner again. I went with the assumption that BasicEventHooks is not for modders to use.

PR #7830, opened 2021-06-21
Lazily initialize ItemStack capabilities

Technici4n:
This PR aims to fix the major bottleneck that item stack capabilities are - specifically, the attach capabilities event. I have seen countless profiles that indicate this is a major issue on worlds that make heavy use of item transfer. Lazily attaching capabilities means that the event will only be fired when accessing a capability. This should appropriately fix the performance issue.

~~I added a boolean field to `ItemStack` to track whether capabilities have already been initialized. The rest should be quite straightforward.~~ This PR allows `CapabilityProvider`s to be lazy. In that case, `capNBT` is used instead of the capabilities until the first `getCapability` call.

~~I will add a few unit tests to make sure this behaves as expected, but I'd like some feedback on the approach first.~~ Added a testmod that automatically runs a test on server start.

### DIFF
--- a/patches/minecraft/net/minecraft/item/ItemStack.java.patch
+++ b/patches/minecraft/net/minecraft/item/ItemStack.java.patch
@@ -9,25 +9,23 @@
     public static final Codec<ItemStack> field_234691_a_ = RecordCodecBuilder.create((p_234698_0_) -> {
        return p_234698_0_.group(Registry.field_212630_s.fieldOf("id").forGetter((p_234706_0_) -> {
           return p_234706_0_.field_151002_e;
-@@ -82,6 +_,9 @@
+@@ -82,6 +_,8 @@
           return Optional.ofNullable(p_234704_0_.field_77990_d);
        })).apply(p_234698_0_, ItemStack::new);
     });
 +   private final net.minecraftforge.registries.IRegistryDelegate<Item> delegate;
-+   private CompoundNBT capNBT;
 +
     private static final Logger field_199558_c = LogManager.getLogger();
     public static final ItemStack field_190927_a = new ItemStack((Item)null);
     public static final DecimalFormat field_111284_a = Util.func_200696_a(new DecimalFormat("#.##"), (p_234699_0_) -> {
-@@ -109,14 +_,19 @@
+@@ -109,10 +_,13 @@
        p_i231596_3_.ifPresent(this::func_77982_d);
     }
  
 -   public ItemStack(IItemProvider p_i48204_1_, int p_i48204_2_) {
 +   public ItemStack(IItemProvider p_i48204_1_, int p_i48204_2_) { this(p_i48204_1_, p_i48204_2_, (CompoundNBT) null); }
 +   public ItemStack(IItemProvider p_i48204_1_, int p_i48204_2_, @Nullable CompoundNBT capNBT) {
-+      super(ItemStack.class);
-+      this.capNBT = capNBT;
++      super(ItemStack.class, capNBT);
        this.field_151002_e = p_i48204_1_ == null ? null : p_i48204_1_.func_199767_j();
 +      this.delegate = p_i48204_1_ == null ? null : p_i48204_1_.func_199767_j().delegate;
        this.field_77994_a = p_i48204_2_;
@@ -36,17 +34,11 @@
           this.func_196085_b(this.func_77952_i());
        }
  
-       this.func_190923_F();
-+      this.forgeInit();
-    }
- 
-    private void func_190923_F() {
-@@ -125,18 +_,23 @@
+@@ -125,14 +_,17 @@
     }
  
     private ItemStack(CompoundNBT p_i47263_1_) {
-+      super(ItemStack.class);
-+      this.capNBT = p_i47263_1_.func_74764_b("ForgeCaps") ? p_i47263_1_.func_74775_l("ForgeCaps") : null;
++      super(ItemStack.class, p_i47263_1_.func_74764_b("ForgeCaps") ? p_i47263_1_.func_74775_l("ForgeCaps") : null);
 +      Item rawItem =
        this.field_151002_e = Registry.field_212630_s.func_82594_a(new ResourceLocation(p_i47263_1_.func_74779_i("id")));
 +      this.delegate = rawItem.delegate;
@@ -61,11 +53,6 @@
           this.func_196085_b(this.func_77952_i());
        }
  
-       this.func_190923_F();
-+      this.forgeInit();
-    }
- 
-    public static ItemStack func_199557_a(CompoundNBT p_199557_0_) {
 @@ -167,10 +_,19 @@
     }
  
@@ -220,7 +207,7 @@
        return multimap;
     }
  
-@@ -952,6 +_,24 @@
+@@ -952,6 +_,21 @@
  
     public boolean func_222117_E() {
        return this.func_77973_b().func_219971_r();
@@ -230,17 +217,14 @@
 +   public void deserializeNBT(CompoundNBT nbt) {
 +      final ItemStack itemStack = ItemStack.func_199557_a(nbt);
 +      getStack().func_77982_d(itemStack.func_77978_p());
-+      if (itemStack.capNBT != null) deserializeCaps(itemStack.capNBT);
 +   }
 +
-+   /**
-+    * Set up forge's ItemStack additions.
-+    */
-+   private void forgeInit() {
++   @Override
++   protected void initializeCaps(@Nullable CompoundNBT capNBT) {
 +      if (this.delegate != null) {
-+         net.minecraftforge.common.capabilities.ICapabilityProvider provider = field_151002_e.initCapabilities(this, this.capNBT);
++         net.minecraftforge.common.capabilities.ICapabilityProvider provider = field_151002_e.initCapabilities(this, capNBT);
 +         this.gatherCapabilities(provider);
-+         if (this.capNBT != null) deserializeCaps(this.capNBT);
++         if (capNBT != null) deserializeCaps(capNBT);
 +      }
     }
  
--- a/src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
+++ b/src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
@@ -19,6 +19,7 @@
 
 package net.minecraftforge.common.capabilities;
 
+import java.util.Objects;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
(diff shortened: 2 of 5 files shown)


patches/minecraft/net/minecraft/item/ItemStack.java.patch
```diff
@@ -169,7 +160,7 @@
           return field_190927_a;
        } else {
 -         ItemStack itemstack = new ItemStack(this.func_77973_b(), this.field_77994_a);
-+         ItemStack itemstack = new ItemStack(this.func_77973_b(), this.field_77994_a, this.serializeCaps());
++         ItemStack itemstack = new ItemStack(this.func_77973_b(), this.field_77994_a, capsInitialized ? this.serializeCaps() : this.capNBT == null ? null : this.capNBT.func_74737_b());
```
Technici4n: In theory that copy isn't even necessary - `capNBT` should:tm: not be mutated.

Technici4n:
Updated with a testmod. It seems to be working fine, so I am marking this as ready for review!

marchermans:
@Technici4n This PR in general looks pretty good.

Two things need fixing/adding though;
1) Your branch is out of date.
2) We will need performance numbers for this. In general it is sufficient to write a command which you can run ingame (via a testmod or in the test mod common init or something) that proves your claimed efficiency improvements. I think it is obvious that it should help here. But it would be nice if you could actually provide some numbers, that would make the process a lot faster.

Thanks orion.

patches/minecraft/net/minecraft/item/ItemStack.java.patch
```diff
@@ -230,17 +221,51 @@
...
 +   }
 +
 +   /**
-+    * Set up forge's ItemStack additions.
++    * Initialize caps if not already initialized.
 +    */
-+   private void forgeInit() {
-+      if (this.delegate != null) {
++   private void forgeInitCaps() {
++      if (!this.capsInitialized && this.delegate != null) {
```
malte0811: Just a quick remark: I'm not sure what kind of threading guarantee (if any) this is supposed to have. But before it was safe to access caps from multiple threads (AFAICT), now that would be a race condition.

KnightMiner: What is the usecase for accessing item stacks on multiple threads? In my understanding, most item stack accesses are on the main thread. Is it something you could synchronize in your own code?

Technici4n: Good point, in theory someone could be copying a stack on the render thread, while someone is querying `getCapability` on the client thread. Will add a double-checked lock.

shartte: I don't think any of the gameplay objects in Minecraft are thread-safe.

Technici4n: After thinking more about it, the race condition already exists currently. You can mutate the stack's capabilities on one thread and copy it on another thread. That would cause the caps to be modified and serialized at the same time.

Technici4n:
Before: https://spark.lucko.me/l4rTRWfv69
AE2 import bus time (`doBusWork`): 14.01% (most of it due to `ItemStack.<init>`).
image
(lots of other constructor calls in other places)

After: https://spark.lucko.me/ZRS1Syh3tv
AE2 import bus time: 1.25% (a lot of it due to storage drawers's sync).
image
(there are still a few `ItemStack.<init>` calls, but no event bottleneck anymore)

The setup is the same: 8 max speed import busses on storage drawers, leading through a few subnetworks to an array of 64 storage busses, each placed on top of a single storage drawer. The modpack that I used is ATM6: Skyblock.

Technici4n:
I have I believe sufficiently addressed thread-safety. This should ensure that capability reads are safe across threads.

I added quite a few comments in case someone needs to update this in the future, hopefully that makes understanding what I did easier.

Merge conflicts have been addressed as well.

patches/minecraft/net/minecraft/item/ItemStack.java.patch
Technici4n: Should be fine now.

src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
```diff
@@ -48,17 +48,17 @@ protected final void gatherCapabilities(@Nullable ICapabilityProvider parent)
         this.capabilities = ForgeEventFactory.gatherCapabilities(baseClass, this, parent);
     }
 
-    protected final @Nullable CapabilityDispatcher getCapabilities()
+    protected @Nullable CapabilityDispatcher getCapabilities()
```
pupnewfster: Not sure if this should be changed or not but part of me wonders from an API standpoint and modders potentially doing weird things in their tiles among other things if these methods should stay final and then have helper package private methods be defined these proxy via and then those get overrode in `LazyCapabilityProvider`. That way modders can't start overriding them and potentially breaking them.

src/main/java/net/minecraftforge/common/capabilities/LazyCapabilityProvider.java
```diff
@@ -0,0 +1,112 @@
...
+        // Detect calls that don't check for capsInitialized, or calls from external code.
+        if (!this.capsInitialized) throw new IllegalStateException("Capabilities are not yet initialized. This is a bug!");
+        return super.getCapabilities();
+    }
+
+    // Reimplementation of comparison that does not fire any capability event, relying on capNBT instead.
+    @Override
+    public boolean areCapsCompatible(@Nonnull CapabilityProvider<B> other)
+    {
+        LazyCapabilityProvider<B> o = (LazyCapabilityProvider<B>) other;
```
pupnewfster: I am not sure this is a safe cast to do as at least API wise another CapabilityProvider may be passed, so this should probably be instanced checked, and then for when not lazy compare it as if the other one is initialized already or maybe some other handling such as just returning false.

src/main/java/net/minecraftforge/common/capabilities/LazyCapabilityProvider.java
```diff
@@ -0,0 +1,112 @@
...
+    private volatile boolean capsInitialized = false; // must be volatile due for double-checked locking to be safe
+
+    protected LazyCapabilityProvider(Class<B> baseClass, @Nullable CompoundNBT capNBT)
+    {
+        super(baseClass);
+        this.capNBT = capNBT;
+    }
+
+    /**
+     * Called right before getCapability when necessary, init your capabilities there.
```
pupnewfster: I think this meant to be `init your capabilities here`

src/main/java/net/minecraftforge/common/capabilities/LazyCapabilityProvider.java
Technici4n: I think this is fine with the `B extends LazyCapabilityProvider<B>` bound.

src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
Technici4n: Not sure tbh, seems quite involved, and modders can still mess it up in other ways. For example they can spam `gatherCapabilities`, or do weird stuff in the event.

Technici4n:
Rebased onto `1.16.x` to fix the conflict.

src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
Technici4n: Changed the approach, now everything is final.

src/main/java/net/minecraftforge/common/capabilities/LazyCapabilityProvider.java
Technici4n: No need for a cast anymore.

Technici4n:
Updated without a `CapabilityProvider` subclass, this is now much nicer.

src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
```diff
@@ -50,16 +79,34 @@ protected final void gatherCapabilities(@Nullable ICapabilityProvider parent)
 
     protected final @Nullable CapabilityDispatcher getCapabilities()
     {
+        // Detect calls that don't check for capsInitialized, or calls from external code.
+        if (!this.capsInitialized)
+        {
+            throw new IllegalStateException("Lazy capabilities are not yet initialized. This is a bug!");
```
LexManos, line 85: This is a no-go, it changes the contract of this method during a RB cycle.
This should init the caps if they are not init.
What is your reasoning for not simply forcing the caps to init here if this is called?
Seems like you could just add a initializeCapsIfNeeded() call instead of throwing an exception...

Technici4n: It's a contract change in a protected method that's only used for ItemStack. Nothing else can trigger it besides the ItemStack patches.

Technici4n: I mean that anything else than ItemStack is not lazy, thus capsInitialized is always true.

LexManos: I understand that, but any modder could be consuming this.
I have yet to hear a compelling reason as to why this can't force the init instead of throwing an error...

Technici4n: I added it as a sanity check mainly, although I guess the restriction could be lifted later.

Forcing init would have an issue with recursive initialization here I think if someone mistakenly calls `getCapabilities` inside their `initializeCaps`. We couldn't force an init as one is already happening, so we'd probably have to throw an error anyway.

KnightMiner: Wouldn't that just give a stack overflow? Obvious enough error IMO, use the javadocs to say don't do that.

Technici4n: I mean let's be practical... When would you want to call this function directly?

LexManos: In any of the cases where people call it now, currently we use it in Forge in the serialization functions.
But since you refuse to do what I am explicitly telling you to do, that being to not breaking existing contracts outside our breaking window, I guess I'll just close this now.

shartte: Uhm, why are you not just retargeting this at 1.17 if you're concerned about breaking changes in 1.16? The massive performance improvement is still valid there.

TheCurle: 1.16 is going to be around for *a while*, and we'd rather this performance gain be merged.

Avoiding the breaking change is simple. It has been directly requested of the PR author twice and both times were ignored.

If the author doesn't wish to cooperate, we're unwilling to waste time with them.

Technici4n: Now a one-time warning.

PR #7832, opened 2021-06-23
Allows for NBT input and output for RecipeBuilders

OneLemonyBoi:
Does what the title says, also changes Ingredient to save NBT

### DIFF
--- a/patches/minecraft/net/minecraft/item/crafting/Ingredient.java.patch
+++ b/patches/minecraft/net/minecraft/item/crafting/Ingredient.java.patch
@@ -25,7 +25,29 @@
     }
  
     public ItemStack[] func_193365_a() {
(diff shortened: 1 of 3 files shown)


diesieben07 (changes requested):
There are two concepts here, one of them straightforward, one of them not.

1. Adding NBT output to ShapedRecipeBuilder. This could be done way simpler than just copy-pasting the whole builders with simple one-line patches to the recipe builders:
    - Add a new field `ItemStack output`
    - Add a setter for this field
    - Patch `serializeRecipeData` to add it to the JSON output.
2. Adding NBT matching to `Ingredient`. A custom ingredient system already exists and `NBTIngredient` is also already a thing (usable via `"forge:nbt"`). As such this is not needed.

PR #7837, opened 2021-06-24
Allows pet death message to be cancelled along with death

### LINKED ISSUE #7274: [1.15.2]Pet death message still sent in chat if LivingDeathEvent cancelled
In TameableEntity#onDeath the death message to the owner is sent before super.onDeath is called, which is where the forge hook is, so you can't stop the death message from being sent even if you cancel the event. :(

OneLemonyBoi:
Allows pet death message to be cancelled along with death and closes #7274, albeit for 1.16. I had to copy paste from LivingEntity.die because if i didn't, ForgeHooks.onLivingDeath would be called twice.


TheCurle:
We should prefer to be able to tell ForgeHooks.onLivingDeath that the name is involved, rather than making the patch unnecessarily huge.

PR #7838, opened 2021-06-24
Prevents Tile Entity data leak by copying CompoundNBT before writing

### LINKED ISSUE #7760: Custom tile entity data sided leak on integrated servers
**Minecraft Version:** 1.16.5

**Forge Version:** 36.1.15

**Description of issue:**

Forge currently writes to NBT by just setting the current instance of its custom tile data into the [NBT `writeInternal`/`saveMetadata` method](https://github.com/MinecraftForge/MinecraftForge/blob/a95b968c1397a1597d4a6228d8e1abce6495570e/patches/minecraft/net/minecraft/tileentity/TileEntity.java.patch#L38). Additionally, it reads the NBT by just setting the custom tile data equal to the current NBT tag. While this works, this means the same NBT compound stored in the tile entity is now also stored in that tag, so updating one will update both.

**Steps to Reproduce:**

One case where this became a problem is when syncing tile data using `getUpdateTag()`. Most implementations will ultimately call `writeInternal`, which includes the implementation in vanilla tile entities. On a ~~dedicated~~ integrated server, the packet is sent directly between threads without serializing it to a packet buffer, so when its received again on the client side and read using `read`/`load`, the compound is still the same tag from the server side. This can lead to some very hard to trace bugs where setting NBT on one side updates it on the other without syncing.

It is possible there are other cases affected by this leak, though since the primary use of that method other than writing to disk is syncing, the only case I know of is syncing on a ~~dedicated~~ integrated server.

**Solution:**

Easiest fix is to call `copy()` on the data before writing and after reading. This will ensure the NBT in the packet is a unique instance so no other objects are modifying the internal tag. At minimum just adding a copy in one of the two places will fix the problem.

OneLemonyBoi:
Prevents Tile Entity data leak by copying CompoundNBT before writing and closes #7760.

### DIFF
--- a/patches/minecraft/net/minecraft/tileentity/TileEntity.java.patch
+++ b/patches/minecraft/net/minecraft/tileentity/TileEntity.java.patch
@@ -31,15 +31,30 @@
     }
  
     public CompoundNBT func_189515_b(CompoundNBT p_189515_1_) {
-@@ -62,6 +_,8 @@
-          p_189516_1_.func_74768_a("x", this.field_174879_c.func_177958_n());
-          p_189516_1_.func_74768_a("y", this.field_174879_c.func_177956_o());
-          p_189516_1_.func_74768_a("z", this.field_174879_c.func_177952_p());
-+         if (this.customTileData != null) p_189516_1_.func_218657_a("ForgeData", this.customTileData);
-+         if (getCapabilities() != null) p_189516_1_.func_218657_a("ForgeCaps", serializeCaps());
-          return p_189516_1_;
+@@ -54,15 +_,18 @@
+    }
+ 
+    private CompoundNBT func_189516_d(CompoundNBT p_189516_1_) {
++      CompoundNBT compoundNBT = p_189516_1_.func_74737_b();
+       ResourceLocation resourcelocation = TileEntityType.func_200969_a(this.func_200662_C());
+       if (resourcelocation == null) {
+          throw new RuntimeException(this.getClass() + " is missing a mapping! This is a bug!");
+       } else {
+-         p_189516_1_.func_74778_a("id", resourcelocation.toString());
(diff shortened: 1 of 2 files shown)


patches/minecraft/net/minecraft/tileentity/TileEntity.java.patch
```diff
@@ -31,15 +31,30 @@
...
-          p_189516_1_.func_74768_a("y", this.field_174879_c.func_177956_o());
-          p_189516_1_.func_74768_a("z", this.field_174879_c.func_177952_p());
-+         if (this.customTileData != null) p_189516_1_.func_218657_a("ForgeData", this.customTileData);
-+         if (getCapabilities() != null) p_189516_1_.func_218657_a("ForgeCaps", serializeCaps());
-          return p_189516_1_;
+@@ -54,15 +_,18 @@
+    }
+ 
+    private CompoundNBT func_189516_d(CompoundNBT p_189516_1_) {
++      CompoundNBT compoundNBT = p_189516_1_.func_74737_b();
```
pupnewfster: Not sure about this PR overall (didn't really review the functionality or things) but wouldn't this be a lot smaller of a patch if it was just `p_189516_1_ = p_189516_1_.func_74737_b()`

covers1624:
This change at current will do nothing. The tag passed into `save` is always a new tag. The root issue states that the `customTileData` field for tiles and entities is not copied when on an integrated server, _because_ vanilla does not read/write the tag to the network. Thus the `customTileData` tag written _into_ the tile's NBT is the same instance on the integrated server and client.

To resolve this issue, you simply need to copy the `customTileData` tag, _before_ writing it to the tile's NBT.

This PR should also be expanded to include the same fix for Entities and its `persistentData` field.

OneLemonyBoi:
Fixed covers

covers1624 (approved):

PR #7842, opened 2021-06-24
Fishing bobber projectile impact

AwesomeSyd239:
Adds a ProjectileImpactEvent for the FishingBobberEntity, since it extends from ProjectileEntity. 
Includes appropriate hook and post from within FishingBobberEntity class.

Also includes a simple mod showing some use cases: sticking the bobber to a block and putting the punch enchantment on a fishing rod.

I think I've done this right


src/main/resources/META-INF/mods.toml
```diff
@@ -3,7 +3,6 @@ loaderVersion="[24,]"
 issueTrackerURL="http://www.minecraftforge.net/"
 logoFile="forge_logo.png"
 license="LGPL v2.1"
-
```
SizableShrimp: You only changed this file with whitespace.

src/test/java/awesomesyd/fishingexample/BarbedEnchantment.java
```diff
@@ -0,0 +1,31 @@
+package awesomesyd.fishingexample;
```
SizableShrimp: No, you cannot have your own name and package for the tests. And a single test mod should all be in one file. Also, you need to run `licenseFormat` so that the license is applied.

src/main/java/net/minecraftforge/event/entity/ProjectileImpactEvent.java
```diff
@@ -125,4 +126,23 @@ public FireworkRocketEntity getFireworkRocket()
             return fireworkRocket;
         }
     }
+    /**
```
SizableShrimp, line 130: Please update these javadocs and remove the TODO. If the event is cancelled, then the raytrace hit for the fishing hook will not be processed and it will continue moving, similar to the other projectile impact events.

.gitignore
```diff
@@ -44,3 +44,4 @@ changelog.txt
 # Srg2Source causes import bugs on these patches. If you intentionnally change these, remove the ignore.
 /patches/minecraft/net/minecraft/client/renderer/ViewFrustum.java.patch
 /patches/minecraft/net/minecraft/data/BlockModelDefinition.java.patch
+*.xml
```
SizableShrimp: Don't change the `.gitignore` file for no reason. If you need to ignore something in git, use `.git/info/exclude` so that is it not committed publicly.

SizableShrimp (changes requested):
There a few files you changed for no good reason, and your test mod is overly complex and not in the right package. Test mods are just used to simply demonstrate stuff and/or make sure it works. They should not need a bunch of classes and enchantments.

.gitignore
AwesomeSyd239: I was trying to not include eclipseformatter.xml in my commit. Am now going with 'discard changes' on that file.

src/main/java/net/minecraftforge/event/entity/ProjectileImpactEvent.java
AwesomeSyd239: Is it preferable to remove the doc line, as in most other ImpactEvents, or have it be the same as firework?

SizableShrimp: Rewording the existing one for fireworks should be fine.

src/main/resources/META-INF/mods.toml
AwesomeSyd239: I previously had fishing_example in this toml. When I moved it to src/test, there appears to be a leftover whitespace. Is there a way to completely remove the mods.toml from the PR here?

src/test/java/awesomesyd/fishingexample/BarbedEnchantment.java
AwesomeSyd239: cleaned up package stuff.
ran gradlew licenseFormat, it also put the forge license on a few things that I never touched.

src/main/resources/META-INF/mods.toml
SizableShrimp: The changes to this file still need to be discarded.

src/test/java/FishingExample.java
```diff
@@ -0,0 +1,38 @@
+
```
SizableShrimp: Do not put the test mod in the default package. Please follow the existing examples on how to structure the package.

src/test/java/FishingExample.java
```diff
@@ -0,0 +1,38 @@
...
+import net.minecraftforge.fml.common.Mod;
+
+@Mod("fishing_example")
+public class FishingExample
+{
+    public FishingExample()
+    {
+        MinecraftForge.EVENT_BUS.register(this);
+        MinecraftForge.EVENT_BUS.addListener(FishingExample::handleImpact);
+    }
```
SizableShrimp: Please add an ENABLED boolean which if true runs everything in the constructor. See other test mods for an example.

SizableShrimp (changes requested):
Looks pretty good now. Just a few comments from reading the code.

ScLohaifg:
merge?

src/main/java/net/minecraftforge/event/entity/ProjectileImpactEvent.java
```diff
@@ -125,4 +126,24 @@ public FireworkRocketEntity getFireworkRocket()
             return fireworkRocket;
         }
     }
+
+    /**
+     * Event is cancellable, causes bobber to ignore the current hit and continue on its journey.
+     */
+    @Cancelable
+    public static class FishingBobber extends ProjectileImpactEvent//TODO figure out what happens on cancel
```
diesieben07: Why is this `TODO` still here?

src/main/java/net/minecraftforge/event/entity/living/EntityTeleportEvent.java
```diff
@@ -1,3 +1,22 @@
+/*
```
diesieben07: Don't change files unrelated to your PR.

src/main/resources/META-INF/mods.toml
```diff
@@ -15,4 +15,3 @@ license="LGPL v2.1"
     description='''
     Forge, a broad compatibility API.
     '''
-
```
diesieben07: Don't change files unrelated to your PR.

src/test/java/net/minecraftforge/debug/block/CustomSignsTest.java
```diff
@@ -1,3 +1,22 @@
+/*
```
diesieben07: Don't change files unrelated to your PR.

src/test/java/net/minecraftforge/debug/client/rendering/LinearTextTextureFilteringTest.java
```diff
@@ -1,6 +1,6 @@
 /*
  * Minecraft Forge
- * Copyright (c) 2016-2020.
+ * Copyright (c) 2016-2021.
```
diesieben07: Don't change files unrelated to your PR.

src/test/java/net/minecraftforge/debug/entity/EntityTeleportEventTest.java
```diff
@@ -1,3 +1,22 @@
...
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
```
diesieben07: Don't change files unrelated to your PR.

diesieben07 (changes requested):

src/main/resources/META-INF/mods.toml
AwesomeSyd239: I ended up editting the file in hex to make it the same as the one from the main github. If it isn't good now, it isn't going to be.

src/main/java/net/minecraftforge/event/entity/ProjectileImpactEvent.java
AwesomeSyd239: because I was apparently blind. fixed now.

patches/minecraft/net/minecraft/entity/projectile/FishingBobberEntity.java.patch
```diff
@@ -1,5 +1,14 @@
 --- a/net/minecraft/entity/projectile/FishingBobberEntity.java
 +++ b/net/minecraft/entity/projectile/FishingBobberEntity.java
+@@ -245,7 +_,7 @@
+ 
+    private void func_190624_r() {
+       RayTraceResult raytraceresult = ProjectileHelper.func_234618_a_(this, this::func_230298_a_);
+-      this.func_70227_a(raytraceresult);
++      if (!net.minecraftforge.event.ForgeEventFactory.onProjectileImpact(this, raytraceresult)) this.func_70227_a(raytraceresult);
```
diesieben07: All other projectiles only fire this event if it isn't a `MISS`. This fires the event even if it is a miss.

diesieben07 (changes requested):

patches/minecraft/net/minecraft/entity/projectile/FishingBobberEntity.java.patch
AwesomeSyd239: I think I left that out originally because most other projectiles fire the event from inside their onHit, and I don't think I realized the non-miss check was a forge insert. I was worried about accidentally messing with/cancelling things too early or accidentally.
Fixed now though.

diesieben07: Well, you are right, `func_70227_a` should still be called on `MISS`, which your patch now changes. You need an `||`, not `&&`. You want to call `func_70227_a` if it's a `MISS` _or_ the forge event was not cancelled:
`if (type == MISS || !onProjectileImpact) func_70227_a(...)`. Due to short-circuiting of `||` this will only fire the event for non-miss events still.

AwesomeSyd239: Oh, that's a good use of the short-circuit, thanks! Patched now.

diesieben07 (approved):

patches/minecraft/net/minecraft/entity/projectile/FishingBobberEntity.java.patch
```diff
@@ -1,5 +1,14 @@
 --- a/net/minecraft/entity/projectile/FishingBobberEntity.java
 +++ b/net/minecraft/entity/projectile/FishingBobberEntity.java
+@@ -245,7 +_,7 @@
+ 
+    private void func_190624_r() {
+       RayTraceResult raytraceresult = ProjectileHelper.func_234618_a_(this, this::func_230298_a_);
+-      this.func_70227_a(raytraceresult);
++      if (raytraceresult.func_216346_c() == RayTraceResult.Type.MISS || !net.minecraftforge.event.ForgeEventFactory.onProjectileImpact(this, raytraceresult)) this.func_70227_a(raytraceresult);
```
LexManos, line 8: This should be above the function call, so its more obvious what the patch is doing. And the line is less wide.

PR #7843, opened 2021-06-25
Moved issue templates to the new issue forms templates

TheOnlyTails:
With the new Issue Forms Templates, users will see a clearer menu, with dropdown menus for versions, Markdown controls for each text input area.

By the way, is it worth it to add a dropdown for Forge versions too? I've added one for Minecraft versions because they're pretty rare, and there aren't that many supported by Forge, while Forge versions happen all the time.

Note: I've only converted the issue report template, considering the feature request one doesn't contain anything.


ChampionAsh5357:
As 1.17 is now out, this PR should be updated to that version.
However, it would be better suited as part of the current omnibus PR: #8144. You can find what the omnibus is and how to submit a pull request on the PR description.
Thank you for your contributions!

Issue #7829, opened 2021-06-21
[RFC][1.17] Refactoring/removing the `PlantType` system


SizableShrimp:
In its current state, the `PlantType` system leads to more bugs and hidden issues for modders than it seems to warrant keeping.

## The problem
`BushBlock` has a patch to make it extend `IPlantable`. However, this means that any modded plant blocks that extend `BushBlock` or any subclass automatically have this interface applied to their block as well. This means, in most cases, a modder gets the default implementation of `IPlantable#getPlantType`. If a modder does not override this implementation and instead uses `mayPlaceOn`, they will have the `PlantType` still applied to their block and may cause unintended side-effects. In some cases, this `PlantType` system breaks vanilla behavior. One example of this for 1.16 is being able to place saplings and other plants with the `PlantType.PLAINS` type on mycelium, which is not possible in Vanilla 1.16. *This has since been changed in 1.17, but the point still stands.*

### Modding example
Let's say a modder extends from `CropBlock`. They don't know of `IPlantable` and are going off of vanilla examples, so they override `mayPlaceOn` to only let the block be placed on cobblestone. However, this means the default implementation of `getPlantType` will be applied, and since their block is an instance of `CropBlock`, the plant type for their custom block by default will be `PlantType.PLAINS`. This type is then checked in `Block#canSustainPlant` when a plant is attempted to be placed on a block. Modders do not have control over this method for vanilla blocks or blocks from other mods, since it is called on the block which is attempting to have a plant placed on it. If the plant block is an instance of `BushBlock` and `mayPlaceOn` returns true, then `canSustainPlant` also returns true. However, if `mayPlaceOn` returns false, the rest of the `canSustainPlant` method is executed, which checks the `PlantType`. With a default return of `PlantType.PLAINS` for a `CropBlock`, this means it will return true when checking `PlantType.PLAINS.equals(type)`, and allow the block to be placed on a grass block, blocks in the `forge:dirt` tag, and farmland without the modder even intending it. The modder is then forced to override `getPlantType` to return null so that only `mayPlaceOn` is used when checking if a block can sustain a plant, if they even know of the method's existence at all. This seems like unintended behavior. I believe `Block#canSustainPlant` is still a useful method to be able to override, however the way it interacts with the `PlantType` system and `IPlantable` seems suboptimal and easy to create unintended bugs both in keeping vanilla behavior and in mod development.

## Feedback
[...]

PR #7844, opened 2021-06-25
[RFC][1.17] Overhaul of PlantType system

OneLemonyBoi:
If there is anything I should add, please LMK.
Relates to #7829.


OneLemonyBoi:
Continued by #7846

PR #7846, opened 2021-06-27
[RFC][1.17] Overhaul of PlantType system

OneLemonyBoi:
If there is anything I should add, please LMK.
Relates to #7829 and continuation of #7844.

### DIFF
--- a/patches/minecraft/net/minecraft/block/Block.java.patch
+++ b/patches/minecraft/net/minecraft/block/Block.java.patch
@@ -90,7 +90,7 @@
     }
  
     @OnlyIn(Dist.CLIENT)
-@@ -409,6 +_,78 @@
+@@ -409,6 +_,43 @@
     protected Block func_230328_p_() {
        return this;
     }
@@ -114,44 +114,9 @@
 +
 +   @Override
 +   public boolean canSustainPlant(BlockState state, IBlockReader world, BlockPos pos, Direction facing, net.minecraftforge.common.IPlantable plantable) {
-+      BlockState plant = plantable.getPlant(world, pos.func_177972_a(facing));
 +      net.minecraftforge.common.PlantType type = plantable.getPlantType(world, pos.func_177972_a(facing));
 +
-+      if (plant.func_177230_c() == Blocks.field_150434_aF)
-+         return state.func_203425_a(Blocks.field_150434_aF) || state.func_203425_a(Blocks.field_150354_m) || state.func_203425_a(Blocks.field_196611_F);
-+
-+      if (plant.func_177230_c() == Blocks.field_196608_cF && this == Blocks.field_196608_cF)
-+         return true;
-+
-+      if (plantable instanceof BushBlock && ((BushBlock)plantable).func_200014_a_(state, world, pos))
-+         return true;
-+
-+      if (net.minecraftforge.common.PlantType.DESERT.equals(type)) {
-+         return this.getBlock() == Blocks.field_150354_m || this.getBlock() == Blocks.field_150405_ch || this.getBlock() instanceof GlazedTerracottaBlock;
-+      } else if (net.minecraftforge.common.PlantType.NETHER.equals(type)) {
-+         return this.getBlock() == Blocks.field_150425_aM;
-+      } else if (net.minecraftforge.common.PlantType.CROP.equals(type)) {
-+         return state.func_203425_a(Blocks.field_150458_ak);
-+      } else if (net.minecraftforge.common.PlantType.CAVE.equals(type)) {
-+         return state.func_224755_d(world, pos, Direction.UP);
-+      } else if (net.minecraftforge.common.PlantType.PLAINS.equals(type)) {
-+         return this.getBlock() == Blocks.field_196658_i || net.minecraftforge.common.Tags.Blocks.DIRT.func_230235_a_(this) || this.getBlock() == Blocks.field_150458_ak;
-+      } else if (net.minecraftforge.common.PlantType.WATER.equals(type)) {
-+         return state.func_185904_a() == net.minecraft.block.material.Material.field_151586_h; //&& state.getValue(BlockLiquidWrapper)
-+      } else if (net.minecraftforge.common.PlantType.BEACH.equals(type)) {
-+         boolean isBeach = state.func_203425_a(Blocks.field_196658_i) || net.minecraftforge.common.Tags.Blocks.DIRT.func_230235_a_(this) || state.func_203425_a(Blocks.field_150354_m) || state.func_203425_a(Blocks.field_196611_F);
-+         boolean hasWater = false;
-+         for (Direction face : Direction.Plane.HORIZONTAL) {
-+             BlockState blockState = world.func_180495_p(pos.func_177972_a(face));
-+             net.minecraft.fluid.FluidState fluidState = world.func_204610_c(pos.func_177972_a(face));
-+             hasWater |= blockState.func_203425_a(Blocks.field_185778_de);
-+             hasWater |= fluidState.func_206884_a(net.minecraft.tags.FluidTags.field_206959_a);
-+             if (hasWater)
-+                break; //No point continuing.
-+         }
-+         return isBeach && hasWater;
-+      }
-+      return false;
++      return type.test(world, pos, state, facing);
 +  }
 +
 +  @Override
--- a/src/main/java/net/minecraftforge/common/IPlantable.java
+++ b/src/main/java/net/minecraftforge/common/IPlantable.java
@@ -19,27 +19,30 @@
 
 package net.minecraftforge.common;
 
-import net.minecraft.block.CropsBlock;
-import net.minecraft.block.FlowerBlock;
-import net.minecraft.block.SaplingBlock;
-import net.minecraft.block.BlockState;
-import net.minecraft.block.Blocks;
+import net.minecraft.block.*;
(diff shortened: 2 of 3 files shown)


patches/minecraft/net/minecraft/block/Block.java.patch
```diff
@@ -114,44 +114,9 @@
...
-+      BlockState plant = plantable.getPlant(world, pos.func_177972_a(facing));
 +      net.minecraftforge.common.PlantType type = plantable.getPlantType(world, pos.func_177972_a(facing));
 +
-+      if (plant.func_177230_c() == Blocks.field_150434_aF)
-+         return state.func_203425_a(Blocks.field_150434_aF) || state.func_203425_a(Blocks.field_150354_m) || state.func_203425_a(Blocks.field_196611_F);
-+
-+      if (plant.func_177230_c() == Blocks.field_196608_cF && this == Blocks.field_196608_cF)
-+         return true;
-+
-+      if (plantable instanceof BushBlock && ((BushBlock)plantable).func_200014_a_(state, world, pos))
```
SizableShrimp, line 126: This check should be preserved.

patches/minecraft/net/minecraft/block/Block.java.patch
```diff
@@ -114,44 +114,9 @@
...
-+             net.minecraft.fluid.FluidState fluidState = world.func_204610_c(pos.func_177972_a(face));
-+             hasWater |= blockState.func_203425_a(Blocks.field_185778_de);
-+             hasWater |= fluidState.func_206884_a(net.minecraft.tags.FluidTags.field_206959_a);
-+             if (hasWater)
-+                break; //No point continuing.
-+         }
-+         return isBeach && hasWater;
-+      }
-+      return false;
++      return type.test(world, pos, state, facing);
```
SizableShrimp, line 119: getPlantType should be allowed to be null

src/main/java/net/minecraftforge/common/IPlantable.java
```diff
@@ -19,28 +19,31 @@
...
+        if (this instanceof NetherSproutsBlock) return PlantType.NYLIUM_PLANTS;
+        if (this instanceof NetherWartBlock) return PlantType.NETHER_WART;
+        if (this instanceof SeaGrassBlock) return PlantType.SEA_GRASS;
+        if (this instanceof SeaPickleBlock) return PlantType.SEA_PICKLE;
+        if (this instanceof StemBlock) return PlantType.FARMLAND;
+        if (this instanceof TallSeaGrassBlock) return PlantType.SEA_GRASS;
+        if (this instanceof WitherRoseBlock) return PlantType.WITHER_ROSE;
+
+        if (this instanceof BushBlock) return PlantType.PLAINS;
+        throw new IllegalStateException("The class implementing IPlantable has no PlantType!");
```
SizableShrimp: This should return null instead and the method be marked as `@Nullable`.

src/main/java/net/minecraftforge/common/PlantType.java
```diff
@@ -19,27 +19,71 @@
...
     private static final Map<String, PlantType> VALUES = new ConcurrentHashMap<>();
 
-    public static final PlantType PLAINS = get("plains");
-    public static final PlantType DESERT = get("desert");
-    public static final PlantType BEACH = get("beach");
-    public static final PlantType CAVE = get("cave");
-    public static final PlantType WATER = get("water");
-    public static final PlantType NETHER = get("nether");
-    public static final PlantType CROP = get("crop");
+    public static final PlantType PLAINS = get("plains", x -> x.getState().is(Blocks.GRASS_BLOCK) || Tags.Blocks.DIRT.contains(x.getState().getBlock()) || x.getState().is(Blocks.FARMLAND));
```
SizableShrimp, line 50: `x` is confusing here. Generally, lambda parameters are named as abbreviated versions of their otherwise normal parameters. I would prefer something like `pi` for `plantInfo`.

src/main/java/net/minecraftforge/common/PlantType.java
```diff
@@ -19,27 +19,71 @@
...
+    public static final PlantType SUGAR_CANE = get("sugar_cane", BEACH.getBlockPredicate().or(x -> x.getState().is(Blocks.SUGAR_CANE)));
+    public static final PlantType NYLIUM_PLANTS = get("nylium_plants", PLAINS.getBlockPredicate().or(x -> x.getState().is(BlockTags.NYLIUM) || x.getState().is(Blocks.SOUL_SOIL)));
+    public static final PlantType SEA_PICKLE = get("sea_pickle", x -> {
+        BlockState below = x.getWorld().getBlockState(x.getPos().below());
+        return !below.getCollisionShape(x.getWorld(), x.getPos()).getFaceShape(Direction.UP).isEmpty() || below.isFaceSturdy(x.getWorld(), x.getPos(), Direction.UP);
+    });
+    public static final PlantType WITHER_ROSE = get("wither_rose", PLAINS.getBlockPredicate().or(x -> x.getState().is(Blocks.NETHERRACK) || x.getState().is(Blocks.SOUL_SAND) || x.getState().is(Blocks.SOUL_SOIL)));
+    public static final PlantType FUNGUS = get("fungus", PLAINS.getBlockPredicate().or(x -> x.getState().is(BlockTags.NYLIUM) || x.getState().is(Blocks.MYCELIUM) || x.getState().is(Blocks.SOUL_SOIL)));
+    public static final PlantType DEAD_BUSH = get("dead_bush", x ->
+    {
```
SizableShrimp, line 83: Please choose a consistent formatting for the lambda brackets. I personally think they are more confusing on new lines, but at least do the same for all of them.

src/main/java/net/minecraftforge/common/PlantType.java
```diff
@@ -19,27 +19,71 @@
...
+    });
+    public static final PlantType SEA_GRASS = get("sea_grass", x -> x.getState().is(Blocks.FARMLAND));
+    public static final PlantType FARMLAND = get("farmland", x -> x.getState().is(Blocks.FARMLAND));
+    public static final PlantType MUSHROOM = get("mushroom", x -> x.getState().isSolidRender(x.getWorld(), x.getPos()));
+    public static final PlantType LILY_PAD = get("lily_pad", x -> {
+        FluidState fluidstate = x.getWorld().getFluidState(x.getPos());
+        FluidState fluidstate1 = x.getWorld().getFluidState(x.getPos().above());
+        return (fluidstate.getType() == Fluids.WATER || x.getState().getMaterial() == Material.ICE) && fluidstate1.getType() == Fluids.EMPTY;
```
SizableShrimp, line 93: These should have better variable names

src/main/java/net/minecraftforge/common/PlantType.java
```diff
@@ -54,26 +98,83 @@
...
+        {
+            return facing;
+        }
+
+        public IBlockReader getWorld()
+        {
+            return world;
+        }
+    }
+}
```
SizableShrimp, line 225: Should probably have an extra newline  character here.

SizableShrimp (changes requested):
I think this is promising. My main concern is modders have a default implementation of getPlantType put upon them that defaults their PlantType based on whatever vanilla block they extend. I think attaching a set of allowed blocks to a PlantType might solve this issue. Then, VALUES could be iterated for any PlantTypes that either have any empty set of allowed blocks (meaning any block) or the block is contained in the set AND the predicate returns true. Additionally, I would stray away from allowing tags to work with this system and purely use a Set of Blocks, because of possible desync issues when dealing with tags and vanilla clients/servers. I'm thinking all the vanilla plant types would have this set hardcoded to whatever vanilla blocks match the PlantTypes. However, mods should be allowed to access a given allowed block set for a PlantType to either modify it or add their own blocks, etc. This set should not be allowed to be null but can be empty. That's just my thoughts on how we could go about making sure modders don't have a default PlantType put on them and making the system more powerful overall

src/main/java/net/minecraftforge/common/IPlantable.java
```diff
@@ -23,26 +23,22 @@
...
-        if (this instanceof SeaGrassBlock) return PlantType.SEA_GRASS;
-        if (this instanceof SeaPickleBlock) return PlantType.SEA_PICKLE;
-        if (this instanceof StemBlock) return PlantType.FARMLAND;
-        if (this instanceof TallSeaGrassBlock) return PlantType.SEA_GRASS;
-        if (this instanceof WitherRoseBlock) return PlantType.WITHER_ROSE;
-
-        if (this instanceof BushBlock) return PlantType.PLAINS;
-        throw new IllegalStateException("The class implementing IPlantable has no PlantType!");
+        for (Map.Entry<String, PlantType> plantType : PlantType.VALUES.entrySet()) {
+            if (plantType.getValue().getAllowedPlants().size() == 0) {
```
KnightMiner, line 34: This looks wrong to me, it will cause the first plant type with no allowed blocks to match for all plants. Means all I have to do is add a plant type with no allowed values and all other plant types fail.

I would also suggest instead of this messy looking conditonal, add something like `PlantType::contains(BlockState)`

SizableShrimp: Yeah, the default implementation should only return plant types if they are in a set of allowed blocks. Otherwise, if the set of allowed blocks is empty, then it should be up to implementers of this method to return that plant type instead.

PR #7849, opened 2021-06-29
Adds a damage float to LightningEntity to enable dynamic lightning damage

baileyholl:
A small patch that moves the hardcoded 5.0f from Entity#thunderHit to a field in LightningBoltEntity, with appropriate setter and getter. 

This will allow modders to adjust lightning damage during the onEntityStruckByLightning event, as well as create lightning bolts that deal differing amounts of damage. 

**Why a new field?**
The damage source passed by the thunderHit method does not provide the damage source an instance of the lightning bolt entity, it is impossible to know if a mod using this damage source has actually spawned a bolt of lighting, or just wishes to use 'lightning damage' as a typing. 

Additionally, if a mod summons a bolt of lightning, but only wants that particular instance of lightning to deal additional damage, LivingHurtEvent does not provide the context necessary to increase that exclusive instance of damage without extra workarounds.

### DIFF
--- a/patches/minecraft/net/minecraft/entity/effect/LightningBoltEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/effect/LightningBoltEntity.java.patch
@@ -1,5 +1,28 @@
 --- a/net/minecraft/entity/effect/LightningBoltEntity.java
 +++ b/net/minecraft/entity/effect/LightningBoltEntity.java
+@@ -30,6 +_,7 @@
+    private boolean field_184529_d;
+    @Nullable
+    private ServerPlayerEntity field_204810_e;
++   private float damage = 5.0F;
+ 
+    public LightningBoltEntity(EntityType<? extends LightningBoltEntity> p_i231491_1_, World p_i231491_2_) {
+       super(p_i231491_1_, p_i231491_2_);
+@@ -51,6 +_,14 @@
+       this.field_204810_e = p_204809_1_;
+    }
+ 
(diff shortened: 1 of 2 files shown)


Thelnfamous1:
Can attest that this is a helpful change for one of my mods.

patches/minecraft/net/minecraft/entity/effect/LightningBoltEntity.java.patch
```diff
@@ -1,5 +1,28 @@
 --- a/net/minecraft/entity/effect/LightningBoltEntity.java
 +++ b/net/minecraft/entity/effect/LightningBoltEntity.java
+@@ -30,6 +_,7 @@
+    private boolean field_184529_d;
+    @Nullable
+    private ServerPlayerEntity field_204810_e;
++   private float damage = 5.0F;
```
KnightMiner, line 7: Does this need to be saved to NBT? I get the feeling most people will call the setter right after the constructor. On the odd chance the bolt lives long enough for the world to close, its damage will be wrong.

Edit: I am told lightning bolts do not serialize to NBT at all, so this is probably not an issue

baileyholl: It appears that LightningBoltEntity EntityType is marked with .noSave(), and from my testing, does not save into the world. So the entity will be gone anyway on world restart.

marchermans (approved):

PR #7850, opened 2021-06-30
Added NBT field for stonecutter result stack.

ferriarnus:
Added a "nbt" field to the SingleItemRecipe Serializer (stonecutter), making it possible to add nbt data to the result stack. The code is a slightly modifiedversion of the shapedcrafting method. Also a small test mod, which adds one item with 2 different tags to the gui (BuildingBlocks) and 3 crafting recipes with stone in the stone cutter. Two of those have the same tag as the items in the gui, the third one is different.

### DIFF
--- a/patches/minecraft/net/minecraft/item/crafting/SingleItemRecipe.java.patch
+++ b/patches/minecraft/net/minecraft/item/crafting/SingleItemRecipe.java.patch
@@ -1,22 +1,8 @@
 --- a/net/minecraft/item/crafting/SingleItemRecipe.java
 +++ b/net/minecraft/item/crafting/SingleItemRecipe.java
-@@ -59,8 +_,21 @@
-    public ItemStack func_77572_b(IInventory p_77572_1_) {
+@@ -60,7 +_,7 @@
        return this.field_222132_b.func_77946_l();
     }
-+   
-+   public static ItemStack itemFromJson(JsonObject p_199798_0_) {
-+      String s = JSONUtils.func_151200_h(p_199798_0_, "result");
-+      net.minecraft.item.Item item = Registry.field_212630_s.func_241873_b(new ResourceLocation(s)).orElseThrow(() -> {
-+          return new com.google.gson.JsonSyntaxException("Unknown item '" + s + "'");
-+      });
-+      if (p_199798_0_.has("data")) {
-+         throw new com.google.gson.JsonParseException("Disallowed data tag found");
-+      } else {
-+         int i = JSONUtils.func_151203_m(p_199798_0_, "count");
-+         return net.minecraftforge.common.crafting.CraftingHelper.getResultItemStack(p_199798_0_, true);
-+      }
-+   }
  
 -   public static class Serializer<T extends SingleItemRecipe> implements IRecipeSerializer<T> {
 +   public static class Serializer<T extends SingleItemRecipe> extends net.minecraftforge.registries.ForgeRegistryEntry<IRecipeSerializer<?>> implements IRecipeSerializer<T> {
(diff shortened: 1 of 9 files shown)


pupnewfster:
I feel like it would be a lot less of a maintenance overhead if this was done in a similar way as how forge [patches](https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/patches/minecraft/net/minecraft/item/crafting/CookingRecipeSerializer.java.patch#L16-L20) cooking recipes to have the same support.

src/main/java/net/minecraftforge/common/crafting/CraftingHelper.java
```diff
@@ -197,6 +197,49 @@ public static ItemStack getItemStack(JsonObject json, boolean readNBT)
 
         return new ItemStack(item, JSONUtils.getAsInt(json, "count", 1));
     }
+    
+    public static ItemStack getResultItemStack(JsonObject json, boolean readNBT)
+    {
+        String itemName = JSONUtils.getAsString(json, "result");
```
KnightMiner: Instead of duplicating the entire method to change one string, why not make a method `getItemStack(JsonObject json, boolean readNBT, String key)` then have the current header call that with `"item"`

ferriarnus: I also taught about doing that, but it would only really be useful if others would also call it. But making it so it _can_ be used is a good idea. I’m currently adding a method to the craftinghelper, so it can be used by others as well (like maybe cooking?). Or would it be better to not do that and just patch it for the stonecutter?

patches/minecraft/net/minecraft/item/crafting/SingleItemRecipe.java.patch
```diff
@@ -9,3 +9,14 @@
        final SingleItemRecipe.Serializer.IRecipeFactory<T> field_222180_t;
  
        protected Serializer(SingleItemRecipe.Serializer.IRecipeFactory<T> p_i50146_1_) {
+@@ -76,9 +_,7 @@
+             ingredient = Ingredient.func_199802_a(JSONUtils.func_152754_s(p_199425_2_, "ingredient"));
+          }
+ 
+-         String s1 = JSONUtils.func_151200_h(p_199425_2_, "result");
```
ferriarnus: Would it be better to not remove "s1" and "i" and just not use them?

src/main/java/net/minecraftforge/common/crafting/CraftingHelper.java
KnightMiner: Almost any time you are not duplicating code its useful, whether its a private helper or a public one. Currently if any change is made to one copy, you have to make it to both. So whether this method will be used in more places or not, don't duplicate a large chunk of code.

It is probably useful to consistently add NBT support to all vanilla recipes if you are planning to do so, but that part is not my call.

ferriarnus:
It seems like there is an inconsistency in the json formats used by vanilla. As far as I can tell uses the workbench a json object as "result", while the others just use a json primitives. The cooking recipe has been patch to also use an object for modded crafting (like adding nbt data). I don't think this is necessary, the nbt field can be added next to the count field. But changing this now would be a breaking change, maybe something for 1.17? It would decrease the patch size.

sciwhiz12:
Hello! @ferriarnus, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

ferriarnus:
Yeah go ahead and close it. A custom recipe can handle everything this does. If I run into anything it can't deal with, I'll open a PR on 1.18.1.

PR #7852, opened 2021-06-30
Exposes getExperienceReward and patches an experience dupe in ZombieEntity

baileyholl:
This PR adds a new method to LivingEntity that calls the protected getExperienceRewards method and fires the forge getExperienceDrop event to return the 'true' exp value of a mob. The forge event in dropExperience has been replaced with this new method call that provides the same functionality. If shouldDropExperience returns false, this method returns 0.

One additional patch is made to ZombieEntity to remove the mutation of xpReward for babies, otherwise the experience value of the mob would increase each time getExperienceDrop is called. This change does not modify the experience value in any way in the super call, MobEntity#getExperienceReward where xpReward is simply checked for 0 and assigned to another int.

**Why a new method?**

Currently, there is no way to get the experience value of a mob until the forge event fires, when it has already died. getExpierenceDrop, shouldDropExperience, and xpReward are all set to protected in LivingEntity and MobEntity. Additionally, the xpReward  field is unreliable for getting the experience value of the mob, as it is not used at all for AnimalEntity and WaterMobEntity.

Changing getExperienceDrop and shouldDropExperience to public is a breaking change for 1.16, and should perhaps be considered for 1.17. Furthermore, even if these two methods were public, this method also combines the experience reward and the forge event for the modder without requiring them to manually fire the forge event each time.

### DIFF
--- a/patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
+++ b/patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
@@ -195,14 +195,15 @@
     }
  
     protected void func_213337_cE() {
-@@ -1240,6 +_,7 @@
-       if (!this.field_70170_p.field_72995_K && (this.func_70684_aJ() || this.field_70718_bc > 0 && this.func_146066_aG() && this.field_70170_p.func_82736_K().func_223586_b(GameRules.field_223602_e))) {
-          int i = this.func_70693_a(this.field_70717_bb);
+@@ -1238,7 +_,7 @@
  
-+         i = net.minecraftforge.event.ForgeEventFactory.getExperienceDrop(this, this.field_70717_bb, i);
+    protected void func_226294_cV_() {
+       if (!this.field_70170_p.field_72995_K && (this.func_70684_aJ() || this.field_70718_bc > 0 && this.func_146066_aG() && this.field_70170_p.func_82736_K().func_223586_b(GameRules.field_223602_e))) {
+-         int i = this.func_70693_a(this.field_70717_bb);
++         int i = this.getTotalExperienceValue(this.field_70717_bb);
+ 
           while(i > 0) {
              int j = ExperienceOrbEntity.func_70527_a(i);
-             i -= j;
 @@ -1260,7 +_,8 @@
        ResourceLocation resourcelocation = this.func_213346_cF();
        LootTable loottable = this.field_70170_p.func_73046_m().func_200249_aQ().func_186521_a(resourcelocation);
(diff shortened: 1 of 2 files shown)


marchermans (changes requested):
This is backwards breaking in it's current state and won't be merged into 1.16 as it states right now. 

Since an RB is out you need to maintain backwards compatibility and that means that the old behaviour is still needed. You just need to mark it as depreciated.

baileyholl:
I have reverted the patch to dropExperience in order to maintain the old behavior. What should be marked as depreciated? The other methods are inaccessible without AT/Mixin at the moment.

patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
```diff
@@ -195,9 +195,12 @@
...
  
     protected void func_213337_cE() {
-@@ -1240,6 +_,7 @@
+@@ -1238,8 +_,9 @@
+ 
+    protected void func_226294_cV_() {
        if (!this.field_70170_p.field_72995_K && (this.func_70684_aJ() || this.field_70718_bc > 0 && this.func_146066_aG() && this.field_70170_p.func_82736_K().func_223586_b(GameRules.field_223602_e))) {
-          int i = this.func_70693_a(this.field_70717_bb);
+-         int i = this.func_70693_a(this.field_70717_bb);
++         int i = func_70693_a(this.field_70717_bb);
```
KnightMiner: This patch change is a bit redundant now, its just removing the `this`

marchermans (approved):
Ignore my stuff....
I misread the patch file.... twice.....

patches/minecraft/net/minecraft/entity/LivingEntity.java.patch
baileyholl: I have readded 'this' to match the previous patch.

TheCurle:
I fail to see how "Changing getExperienceDrop and shouldDropExperience to public is a breaking change for 1.16". 
Things that relied on them being private would have no change, as a wider visibility is a superset rather than a subset.

Mods compiled against the new build would error on older versions of Forge, yes, but the same is true for every new feature we add in the release cycle. It's the reason we bump the RB every so often. 

Honestly, this could all just be one AT line and it'd be fine. But i'll defer to the actual maintainers of the project for final opinion.

baileyholl:
Are existing binaries affected by changing this method from protected to public? Any mod that overrides that method will be doing so with a stricter access than the parent class. Is this a compile time check only? 

Additionally, while I originally planned this PR with only changing the access in mind, I think this method providing the event call for the modder and respecting canDropExperience is a nice quality of life addition.

sciwhiz12:

PR #7857, opened 2021-07-08
Update README.md

SollyW:
Since other modloaders exist I think you should have this edited


OrionDevelopment:
This is as of this moment not something we currently care about.
Maybe in the future this is something we can take a look at, but at the moment we are currently focussing on getting the mappings into a fully useable up to date state. And pull requests like these take time away from that goal.

SollyW:
Leaving this in is spreading misinformation
You shouldn't be closing it

diesieben07:
Even if we change this text, it is not something that happens via a pull request like this.

immibis:
This is literally what pull requests are for

PR #7862, opened 2021-07-09
[1.16.x] Add an API for dynamic dimensions

Commoble:
## What's this add?
* An API for adding and removing dimensions during server runtime

## Why's this add that?
* This API allows mods to generate indefinitely many concurrently-existing worlds whose properties and quantity are not necessarily known ahead of time
* Mods would be able to use this API to (for example) add random dungeon dimensions, per-player personal spaces, or dimensions whose properties are determined via player input.
* The ability to remove or unregister dimensions is needed by mods that add many dynamic dimensions, as registered worlds are kept in memory; this allows mods to completely unload worlds while they are not in use.

## Public APIs
* The new DynamicDimensionManager class contains all public APIs for this feature:
  * `getOrCreateWorld` returns a world for a specified world key, creating and registering the world and dimension if the world doesn't already exist.
    * This also fires a world load event once the world has been registered into the server.
  * `markDimensionForUnregistration` schedules a world and dimension to be unregistered and unloaded at the end of the current server tick, just after the post-server-tick-event fires.
    * World unload events are fired for each world in sequence, before any worlds and dimensions are removed from the registries.
    * If any players remain in a world when the world is unloaded, they will be ejected to their respawn point (or to the overworld if their respawn point was in a world-being-removed). Mods whose dynamic dimensions require players to be ejected elsewhere should manually teleport them before scheduling dimension removal.
  * `getWorldsPendingUnregistration` returns an immutable view of the world keys scheduled for unregistration.

## Other Features
* When dimensions are added or removed, a syncing packet is sent to clients to update their dimension ID list.
  * This list is normally only used for suggesting dimension IDs for commands; minecraft does not normally sync any information about dimensions other than their IDs.
  * This syncing feature is only available to clients if the client also has an up-to-date build of forge with the features added in this PR, as vanilla clients are only able to update their dimension ID lists on login. Vanilla clients and older forge clients (using forge builds without this PR's features) will be able to connect to forge servers, but are unable to handle the syncing packet when the servers' dimensions update during runtime, so these clients' dimension ID lists will remain outdated until they leave and rejoin the server.

## Effects on Existing Mods
* This PR adds no patches to vanilla sources, and does not directly cause any binary-breaking changes; if a server updates to a new build of forge with this PR without updating any existing mods installed on the server, then the server will experience no problems.
* However, the existence of a mod *using* the features added in this PR does change the assumptions that other mods are allowed to make (meaning that mods that start using the dynamic dimensions API may lose compatibility with other mods whose assumptions have been broken, until the other mods update to account for these changes):
  * This PR definalizes the dimension registry field of the server's DimensionGeneratorSettings, so that when dimensions are unregistered, the server's dimension registry can be replaced with a new registry containing only the keys and dimensions that were not removed (which is substantially simpler than removing dimensions from the existing registry). Any mod that is caching the server's dimension registry would need to update to not cache the server's dimension registry in order to be compatible with mods that use the dimension removal API. The probability of such a mod existing is not expected to be high.
  * This PR breaks the assumption that a world or dimension present in the server's world registry or dimension registry at any given point in time will continue to be present in the server's world or dimension registry at any point in the future (or that a world not being present will continue to not be present). Mods that assume this would need to update to be compatible with mods that use the dynamic dimension API. I have seen at least one existing mod that makes this assumption.

## Other Concerns
* While the dimension removal API does prevent the vanilla dimensions from being unregistered, currently there are no safeguards to prevent mods from using the dynamic dimension API to remove json dimensions. This is because A) this would be difficult to implement (json dimensions are loaded in the same manner as previously-registered dynamic dimensions), and B) unregistering a json dimension generally doesn't cause significant problems (the server will automatically reregister the dimension the next time the server starts).
* If a player leaves the server, then a dimension is removed from the server, then the player rejoins the server, they will be placed into the overworld instead (this is a vanilla failsafe).
* This PR also fixes the test mods for forge world types and player game mode change events so that this PR could be tested on a dedicated server.
* This API is not intended to replace the JSON dimension system; static dimensions should still be defined via JSONs.

### DIFF
--- a/src/main/java/net/minecraftforge/common/DynamicDimensionManager.java
+++ b/src/main/java/net/minecraftforge/common/DynamicDimensionManager.java
@@ -1,6 +1,5 @@
 package net.minecraftforge.common;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
@@ -92,7 +91,7 @@ public class DynamicDimensionManager
      * Marks a world and its dimension for unregistration. Unregistered worlds will stop ticking,
      * unregistered dimensions will not be loaded on server startup unless and until they are reregistered again.
      * 
-     * Unregistration is delayed until the end of the server tick (just before the post-server-tick-event fires).
+     * Unregistration is delayed until the end of the server tick (just after the post-server-tick-event fires).
      * 
      * Players who are still in the given world at that time will be ejected to their respawn points.
      * Players who have respawn points in worlds being unloaded will have their spawn points reset to the overworld and respawned there.
--- a/src/main/java/net/minecraftforge/common/DynamicDimensionManager.java
+++ b/src/main/java/net/minecraftforge/common/DynamicDimensionManager.java
@@ -47,6 +47,7 @@ import net.minecraftforge.fml.network.NetworkHooks;
 @ParametersAreNonnullByDefault
 public class DynamicDimensionManager
 {
+    private static final Set<RegistryKey<World>> VANILLA_WORLDS = ImmutableSet.of(World.OVERWORLD, World.NETHER, World.END); 
     private static Set<RegistryKey<World>> pendingWorldsToUnregister = new HashSet<>();
     
     /**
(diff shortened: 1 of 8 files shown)


Commoble:
Closing this as the branch for later versions was closed

PR #7864, opened 2021-07-11
Add a default for AbstractRailBlock.getShapeProperty

AwesomeSyd239:
Gives AbstractRailBlock a default SHAPE property, and uses that for a default implementation of getShapeProperty(). 
Also replaced a few Deprecated getShapeProperty() function calls with the forge-recommended getRailDirection().

### DIFF
--- a/patches/minecraft/net/minecraft/block/AbstractRailBlock.java.patch
+++ b/patches/minecraft/net/minecraft/block/AbstractRailBlock.java.patch
@@ -1,6 +1,6 @@
 --- a/net/minecraft/block/AbstractRailBlock.java
 +++ b/net/minecraft/block/AbstractRailBlock.java
-@@ -13,7 +_,7 @@
+@@ -13,9 +_,10 @@
  import net.minecraft.world.IWorldReader;
  import net.minecraft.world.World;
  
@@ -8,12 +8,17 @@
 +public abstract class AbstractRailBlock extends Block implements net.minecraftforge.common.extensions.IAbstractRailBlock {
     protected static final VoxelShape field_185590_a = Block.func_208617_a(0.0D, 0.0D, 0.0D, 16.0D, 2.0D, 16.0D);
     protected static final VoxelShape field_190959_b = Block.func_208617_a(0.0D, 0.0D, 0.0D, 16.0D, 8.0D, 16.0D);
++   public static net.minecraft.state.EnumProperty<RailShape> SHAPE = net.minecraft.state.properties.BlockStateProperties.field_208165_R;
     private final boolean field_196277_c;
-@@ -37,6 +_,7 @@
+ 
+    public static boolean func_208488_a(World p_208488_0_, BlockPos p_208488_1_) {
+@@ -36,7 +_,8 @@
+    }
  
     public VoxelShape func_220053_a(BlockState p_220053_1_, IBlockReader p_220053_2_, BlockPos p_220053_3_, ISelectionContext p_220053_4_) {
-       RailShape railshape = p_220053_1_.func_203425_a(this) ? p_220053_1_.func_177229_b(this.func_176560_l()) : null;
-+      RailShape railShape2 = p_220053_1_.func_203425_a(this) ? getRailDirection(p_220053_1_, p_220053_2_, p_220053_3_, null) : null;
+-      RailShape railshape = p_220053_1_.func_203425_a(this) ? p_220053_1_.func_177229_b(this.func_176560_l()) : null;
++      //RailShape railshape = p_220053_1_.is(this) ? p_220053_1_.getValue(this.getShapeProperty()) : null;
++      RailShape railshape = p_220053_1_.func_203425_a(this) ? getRailDirection(p_220053_1_, p_220053_2_, p_220053_3_, null) : null;
        return railshape != null && railshape.func_208092_c() ? field_190959_b : field_185590_a;
     }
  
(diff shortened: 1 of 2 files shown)


patches/minecraft/net/minecraft/block/AbstractRailBlock.java.patch
```diff
@@ -1,19 +1,24 @@
...
+    public static boolean func_208488_a(World p_208488_0_, BlockPos p_208488_1_) {
+@@ -36,7 +_,8 @@
+    }
  
     public VoxelShape func_220053_a(BlockState p_220053_1_, IBlockReader p_220053_2_, BlockPos p_220053_3_, ISelectionContext p_220053_4_) {
-       RailShape railshape = p_220053_1_.func_203425_a(this) ? p_220053_1_.func_177229_b(this.func_176560_l()) : null;
-+      RailShape railShape2 = p_220053_1_.func_203425_a(this) ? getRailDirection(p_220053_1_, p_220053_2_, p_220053_3_, null) : null;
+-      RailShape railshape = p_220053_1_.func_203425_a(this) ? p_220053_1_.func_177229_b(this.func_176560_l()) : null;
++      //RailShape railshape = p_220053_1_.is(this) ? p_220053_1_.getValue(this.getShapeProperty()) : null;
++      RailShape railshape = p_220053_1_.func_203425_a(this) ? getRailDirection(p_220053_1_, p_220053_2_, p_220053_3_, null) : null;
```
diesieben07, line 21: This is a strange patch. Why is there a comment here?

patches/minecraft/net/minecraft/block/AbstractRailBlock.java.patch
```diff
@@ -1,19 +1,24 @@
...
-@@ -13,7 +_,7 @@
+@@ -13,9 +_,10 @@
  import net.minecraft.world.IWorldReader;
  import net.minecraft.world.World;
  
 -public abstract class AbstractRailBlock extends Block {
 +public abstract class AbstractRailBlock extends Block implements net.minecraftforge.common.extensions.IAbstractRailBlock {
     protected static final VoxelShape field_185590_a = Block.func_208617_a(0.0D, 0.0D, 0.0D, 16.0D, 2.0D, 16.0D);
     protected static final VoxelShape field_190959_b = Block.func_208617_a(0.0D, 0.0D, 0.0D, 16.0D, 8.0D, 16.0D);
++   public static net.minecraft.state.EnumProperty<RailShape> SHAPE = net.minecraft.state.properties.BlockStateProperties.field_208165_R;
```
diesieben07, line 11: Why?

patches/minecraft/net/minecraft/block/AbstractRailBlock.java.patch
```diff
@@ -35,12 +49,16 @@
...
        return blockstate.func_206870_a(this.func_176560_l(), flag ? RailShape.EAST_WEST : RailShape.NORTH_SOUTH);
     }
  
+-   public abstract Property<RailShape> func_176560_l();
 +   @Deprecated //Forge: Use getRailDirection(IBlockAccess, BlockPos, IBlockState, EntityMinecart) for enhanced ability
-    public abstract Property<RailShape> func_176560_l();
++   public Property<RailShape> func_176560_l()
++   {
++       return SHAPE;
++   };
```
diesieben07, line 61: Why?

diesieben07 (changes requested):

patches/minecraft/net/minecraft/block/AbstractRailBlock.java.patch
AwesomeSyd239: The existing code had both versions (getShapeProperty and getRailDirection) but assigned from the property.
I assume there was some reason why both function calls are here, so rather than outright deleting the first, I just commented it and changed the assignment to use the existing getRailDirection call.

patches/minecraft/net/minecraft/block/AbstractRailBlock.java.patch
AwesomeSyd239: This is the default RailShape Property.

patches/minecraft/net/minecraft/block/AbstractRailBlock.java.patch
AwesomeSyd239: Currently, any extension of AbstractRailBlock must override this function, but is not strictly required to have a RailShape Property. The code will compile, and then crash
Adding a default allows safe extension of AbstractRailBlock without having to deal with RailShape.

patches/minecraft/net/minecraft/block/AbstractRailBlock.java.patch
KnightMiner: Yes, but it's making the patch larger for no functional gain, just directly reference BlockStatePropeeties where you use it, and let the modder add this field if they need it.

patches/minecraft/net/minecraft/block/AbstractRailBlock.java.patch
diesieben07: It seems this patch was broken in the 1.16 update. Look at the Git history of the patch file.

patches/minecraft/net/minecraft/block/AbstractRailBlock.java.patch
diesieben07: No it doesn't, the subclass still needs to register this property into its block state. By making this not abstract you are even making this requirement harder to discover. Plus this is adding patch size for not much gain.

AwesomeSyd239: That is unfortunate; the entire reason I started this PR was to avoid registering (an unwanted) RailShape into a subclass.

diesieben07: You don't need a RailShape property if everything calls getRailDirection, that is the entire reason it exists and this method is deprecated.

AwesomeSyd239: RailState.place (invoked by a vanilla rail, acting on my custom AbstractRailBlock) uses this.state = this.state.setValue(this.block.getShapeProperty(), railshape);
I don't see a way to use getRailDirection in there.

patches/minecraft/net/minecraft/block/AbstractRailBlock.java.patch
AwesomeSyd239: Placing a vanilla rail adjacent to any AbstractRailBlock will try to setValue the shape property of the AbstractRailBlock, without ever checking for the existence of that property. (This results in a NPE for any AbstractRailBlock that does not have a RailShape Property)
The point of this PR is to patch that.

patches/minecraft/net/minecraft/block/AbstractRailBlock.java.patch
diesieben07: A hook is probably needed then to provide a custom RailState implementation.

sciwhiz12:
Hello! @AwesomeSyd239, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

sciwhiz12:

PR #7865, opened 2021-07-11
[1.16.x] Made Grindstone and Anvil Recipe Type and Serializer. 

ferriarnus:
Added a Grindstone recipe type and (default) serializer, making it possible to use json recipes and custom serializers for the grindstone. The grindstone first checks for a recipe, after which it does its default behaviour.  The grindstone now also accepts any item as input, not just damageable or book-items. The implemented serializer is a copy of the smithing table, but I will add nbt support. An example mod which adds a recipe is also included. If you have some comments, suggestions or concerns please tell them.


Thelnfamous1:
I like the idea. If you can figure out the NBT stuff for grinding, perhaps add a recipe type for the anvil as well if the logic is similar enough. It’d be extremely nice to use both of these containers for some data-driven NBT crafting.

patches/minecraft/net/minecraft/inventory/container/GrindstoneContainer.java.patch
```diff
@@ -1,5 +1,46 @@
...
+    private final IWorldPosCallable field_217015_e;
++   private final World level;
+ 
+    public GrindstoneContainer(int p_i50080_1_, PlayerInventory p_i50080_2_) {
+       this(p_i50080_1_, p_i50080_2_, IWorldPosCallable.field_221489_a);
+@@ -35,14 +_,15 @@
+    public GrindstoneContainer(int p_i50081_1_, PlayerInventory p_i50081_2_, final IWorldPosCallable p_i50081_3_) {
+       super(ContainerType.field_221521_o, p_i50081_1_);
+       this.field_217015_e = p_i50081_3_;
++      this.level = p_i50081_2_.field_70458_d.field_70170_p;
```
ChampionAsh5357: You have an `IWorldPosCallable`, this addition is unnecessary.

patches/minecraft/net/minecraft/inventory/container/GrindstoneContainer.java.patch
```diff
@@ -1,5 +1,46 @@
...
+       this.func_75146_a(new Slot(this.field_217014_d, 0, 49, 19) {
+          public boolean func_75214_a(ItemStack p_75214_1_) {
+-            return p_75214_1_.func_77984_f() || p_75214_1_.func_77973_b() == Items.field_151134_bR || p_75214_1_.func_77948_v();
++            return true;
```
ChampionAsh5357, line 8: Breaks vanilla behavior. This is my opinion, but you should check whether the stack is valid rather than just saying all stacks are valid. This could be justified in either direction however.

patches/minecraft/net/minecraft/inventory/container/GrindstoneContainer.java.patch
```diff
@@ -1,5 +1,46 @@
...
+          }
+       });
+       this.func_75146_a(new Slot(this.field_217014_d, 1, 49, 40) {
+          public boolean func_75214_a(ItemStack p_75214_1_) {
+-            return p_75214_1_.func_77984_f() || p_75214_1_.func_77973_b() == Items.field_151134_bR || p_75214_1_.func_77948_v();
++            return true;
```
ChampionAsh5357, line 14: Breaks vanilla behavior. This is my opinion, but you should check whether the stack is valid rather than just saying all stacks are valid. This could be justified in either direction however.

patches/minecraft/net/minecraft/inventory/container/GrindstoneContainer.java.patch
```diff
@@ -1,5 +1,46 @@
...
+       boolean flag = !itemstack.func_190926_b() || !itemstack1.func_190926_b();
+       boolean flag1 = !itemstack.func_190926_b() && !itemstack1.func_190926_b();
++      java.util.Optional<net.minecraftforge.common.crafting.GrindStoneRecipe> optional = this.level.func_199532_z().func_215371_a(net.minecraft.item.crafting.IRecipeType.GRINDING, this.field_217014_d, this.level);
++      if (optional.isPresent()) {
++        net.minecraftforge.common.crafting.GrindStoneRecipe grindStoneRecipe = optional.get();
++        ItemStack result = grindStoneRecipe.func_77572_b(this.field_217013_c);
++        this.field_217013_c.func_70299_a(0, result);
++        this.func_75142_b();
++        return;
++      }
```
ChampionAsh5357: This can be a one line patch with an external method reference. Also, it's located in the wrong place. The flag after your patch checks to see if both are empty so this call executes unnecessarily. You should either move it where it needs to go or check flag1 and execute the method check. That way no unnecessary evaluations are done.

patches/minecraft/net/minecraft/item/crafting/IRecipeSerializer.java.patch
```diff
@@ -9,7 +9,11 @@
     IRecipeSerializer<ShapedRecipe> field_222157_a = func_222156_a("crafting_shaped", new ShapedRecipe.Serializer());
     IRecipeSerializer<ShapelessRecipe> field_222158_b = func_222156_a("crafting_shapeless", new ShapelessRecipe.Serializer());
     SpecialRecipeSerializer<ArmorDyeRecipe> field_222159_c = func_222156_a("crafting_special_armordye", new SpecialRecipeSerializer<>(ArmorDyeRecipe::new));
-@@ -30,6 +_,7 @@
+@@ -27,9 +_,11 @@
+    CookingRecipeSerializer<CampfireCookingRecipe> field_222174_r = func_222156_a("campfire_cooking", new CookingRecipeSerializer<>(CampfireCookingRecipe::new, 100));
+    IRecipeSerializer<StonecuttingRecipe> field_222175_s = func_222156_a("stonecutting", new SingleItemRecipe.Serializer<>(StonecuttingRecipe::new));
+    IRecipeSerializer<SmithingRecipe> field_234826_u_ = func_222156_a("smithing", new SmithingRecipe.Serializer());
++   IRecipeSerializer<net.minecraftforge.common.crafting.GrindStoneRecipe> GRINDSTONE = func_222156_a("grinding", new net.minecraftforge.common.crafting.GrindStoneRecipe.Serializer());
```
ChampionAsh5357: This is not how you register within forge. See `ForgeMod` and how attributes are registered. You'll find its the same as the documentation.

patches/minecraft/net/minecraft/item/crafting/IRecipeType.java.patch
```diff
@@ -0,0 +1,10 @@
+--- a/net/minecraft/item/crafting/IRecipeType.java
++++ b/net/minecraft/item/crafting/IRecipeType.java
+@@ -14,6 +_,7 @@
+    IRecipeType<CampfireCookingRecipe> field_222153_e = func_222147_a("campfire_cooking");
+    IRecipeType<StonecuttingRecipe> field_222154_f = func_222147_a("stonecutting");
+    IRecipeType<SmithingRecipe> field_234827_g_ = func_222147_a("smithing");
++   IRecipeType<net.minecraftforge.common.crafting.GrindStoneRecipe> GRINDING = func_222147_a("grinding");
```
ChampionAsh5357: Unnecessary patch, move to `ForgeMod`.

src/main/java/net/minecraftforge/common/crafting/GrindStoneRecipe.java
```diff
@@ -0,0 +1,110 @@
+package net.minecraftforge.common.crafting;
```
ChampionAsh5357: Missing license

src/main/java/net/minecraftforge/common/crafting/GrindStoneRecipe.java
```diff
@@ -0,0 +1,110 @@
...
+import net.minecraft.item.crafting.ShapedRecipe;
+import net.minecraft.network.PacketBuffer;
+import net.minecraft.util.JSONUtils;
+import net.minecraft.util.ResourceLocation;
+import net.minecraft.world.World;
+
+public class GrindStoneRecipe implements IRecipe<IInventory>
+{
+
+	private ResourceLocation id;
```
ChampionAsh5357: Forge uses four spaces, not tabs.

src/main/java/net/minecraftforge/common/crafting/GrindStoneRecipe.java
```diff
@@ -0,0 +1,110 @@
...
+import net.minecraft.item.crafting.IRecipeSerializer;
+import net.minecraft.item.crafting.IRecipeType;
+import net.minecraft.item.crafting.Ingredient;
+import net.minecraft.item.crafting.ShapedRecipe;
+import net.minecraft.network.PacketBuffer;
+import net.minecraft.util.JSONUtils;
+import net.minecraft.util.ResourceLocation;
+import net.minecraft.world.World;
+
+public class GrindStoneRecipe implements IRecipe<IInventory>
```
ChampionAsh5357: No documentation and classname does not match vanilla naming scheme.

src/main/java/net/minecraftforge/common/crafting/GrindStoneRecipe.java
```diff
@@ -0,0 +1,110 @@
...
+{
+
+	private ResourceLocation id;
+	private Ingredient ingredient;
+	private Ingredient ingredient1;
+	private ItemStack result;
+	
+
+	public GrindStoneRecipe(ResourceLocation id,Ingredient ingredient, Ingredient ingredient1, ItemStack result)
```
ChampionAsh5357: Inconsistent parameter formatting

src/main/java/net/minecraftforge/common/crafting/GrindStoneRecipe.java
```diff
@@ -0,0 +1,110 @@
...
+	public GrindStoneRecipe(ResourceLocation id,Ingredient ingredient, Ingredient ingredient1, ItemStack result) 
+	{
+		this.id = id;
+		this.ingredient = ingredient;
+		this.ingredient1 = ingredient1;
+		this.result = result;
+	}
+	
+	@Override
+	public boolean matches(IInventory p_77569_1_, World p_77569_2_)
```
ChampionAsh5357: Param names in your own class, rename to something understandable.

src/main/java/net/minecraftforge/common/crafting/GrindStoneRecipe.java
```diff
@@ -0,0 +1,110 @@
...
+		return IRecipeSerializer.GRINDSTONE;
+	}
+
+	@Override
+	public IRecipeType<?> getType() 
+	{
+		return IRecipeType.GRINDING;
+	}
+	
+	public static class Serializer extends net.minecraftforge.registries.ForgeRegistryEntry<IRecipeSerializer<?>> implements IRecipeSerializer<GrindStoneRecipe>
```
ChampionAsh5357: Does this really need to be fully qualified if it's your own class?

src/main/java/net/minecraftforge/common/crafting/GrindStoneRecipe.java
```diff
@@ -0,0 +1,110 @@
...
+	        Ingredient ingredient1 = Ingredient.fromJson(JSONUtils.getAsJsonObject(p_199425_2_, "addition"));
+	        ItemStack result = ShapedRecipe.itemFromJson(JSONUtils.getAsJsonObject(p_199425_2_, "result"));
+			return new GrindStoneRecipe(p_199425_1_, ingredient, ingredient1, result);
+		}
+
+		@Override
+		public GrindStoneRecipe fromNetwork(ResourceLocation p_199426_1_, PacketBuffer p_199426_2_) 
+		{
+			Ingredient ingredient = Ingredient.fromNetwork(p_199426_2_);
+	        Ingredient ingredient1 = Ingredient.fromNetwork(p_199426_2_);
```
ChampionAsh5357: Mixture of spaces and tabs everywhere.

src/main/java/net/minecraftforge/common/crafting/GrindStoneRecipe.java
```diff
@@ -0,0 +1,110 @@
...
+			p_199427_2_.ingredient.toNetwork(p_199427_1_);
+	        p_199427_2_.ingredient1.toNetwork(p_199427_1_);
+	        p_199427_1_.writeItem(p_199427_2_.result);
+		}
+		
+	}
+
+	
+	
+
```
ChampionAsh5357: Weird blank space.

src/test/java/net/minecraftforge/debug/item/crafting/GrindStoneRecipeTest.java
```diff
@@ -0,0 +1,35 @@
...
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.debug.item.crafting;
+
+import net.minecraftforge.fml.common.Mod;
+
+@Mod(GrindStoneRecipeTest.MOD_ID)
+@Mod.EventBusSubscriber(bus = Mod.EventBusSubscriber.Bus.MOD, modid = GrindStoneRecipeTest.MOD_ID)
```
ChampionAsh5357: But there's no events?

src/main/java/net/minecraftforge/common/crafting/GrindStoneRecipe.java
```diff
@@ -0,0 +1,110 @@
...
+	private Ingredient ingredient;
+	private Ingredient ingredient1;
+	private ItemStack result;
+	
+
+	public GrindStoneRecipe(ResourceLocation id,Ingredient ingredient, Ingredient ingredient1, ItemStack result) 
+	{
+		this.id = id;
+		this.ingredient = ingredient;
+		this.ingredient1 = ingredient1;
```
ChampionAsh5357: What is ingredient1?

src/main/java/net/minecraftforge/common/crafting/GrindStoneRecipe.java
```diff
@@ -0,0 +1,110 @@
...
+	private ResourceLocation id;
+	private Ingredient ingredient;
+	private Ingredient ingredient1;
+	private ItemStack result;
+	
+
+	public GrindStoneRecipe(ResourceLocation id,Ingredient ingredient, Ingredient ingredient1, ItemStack result) 
+	{
+		this.id = id;
+		this.ingredient = ingredient;
```
ChampionAsh5357: What is ingredient?

src/test/java/net/minecraftforge/debug/item/crafting/GrindStoneRecipeTest.java
```diff
@@ -0,0 +1,35 @@
...
+public class GrindStoneRecipeTest
+{
+    public static final String MOD_ID = "grindstonerecipe_test";
```
ChampionAsh5357: Nitpick, but why one word except for `test`?

src/test/resources/data/grindstonerecipes_test/recipes/test.json
```diff
@@ -0,0 +1,12 @@
...
+  "base": {
+    "item": "minecraft:diamond_axe"
+  },
+  "addition": {
+    "item": "minecraft:netherite_ingot"
+  },
+  "result": {
+    "item": "minecraft:netherite_axe"
+  }
+}
```
ChampionAsh5357: Inconsistent line ending.

src/test/resources/data/grindstonerecipes_test/recipes/test.json
```diff
@@ -0,0 +1,12 @@
+{
+  "type": "minecraft:grinding",
```
ChampionAsh5357: This is not a minecraft added recipe. Should be under the forge namespace.

ChampionAsh5357 (commented):
This is missing some tools that would be expected of a new recipe type. Specifically, the data providers for creating this recipe and a test of it working. That would also need to have the forge nbt support as is provided within the grindstone recipe deserializer. There are also naming schemes which indicate no verbosity as to what the field or parameter is for.

I'm not sure how valid of a PR this is. One of the major points that comes to mind is the recipe book. A good portion of vanilla recipes use the recipe book. And for those that don't, the recipes are available to the user in a different format. Only case this does not hold true is for the brewing recipes but they use a different recipe system. However, this PR forces the grindstone to conform to the same behavior and logic that vanilla would follow. So would there be a recipe book implementation? Would it have some other way to show recipes? Or would it just completely lack this feature since vanilla did not intend this to be data driven (yet?). 

Another opinion is on the anvil which is essentially the same as a grindstone. Why not do something similar with an event system? Should the anvil also conform to this data driven behavior? If so, will the above concerns also be addressed.

This system more or less just seems like a patchwork feature rather than a fully implemented patchwork feature. This is my opinion on the matter anyways. It might be completely valid and my concerns are just overly thought. But it does bring up the point of how much should Forge patch to implement data driven features and which other recipe systems should be patched to implement as such.

patches/minecraft/net/minecraft/item/crafting/IRecipeSerializer.java.patch
ferriarnus: Thank you, I was also wondering about that. This makes it a "vanilla" method, which it actually isn't. Will take a look at `ForgeMod`

src/main/java/net/minecraftforge/common/crafting/GrindStoneRecipe.java
ferriarnus: The grindstone has 2 input slots, so their are 2 ingredients. I will probably change it to a list, similar to Shapeless, unless someone thinks its better to keep them separate? Than I will find some better names.

ferriarnus:
It's still a draft cause i knew there would be quiet a few issue I couldn't solve without feedback. Adding the anvil is a good idea and I was planning to add it to this PR if the way I did it here worked. So the main problem (outside of misformatted code and poor naming) is if the grindstone (and anvil) should get a "vanilla" based recipe type. What do you mean with "event"? Add an event when the grindstone or anvil is used? This works, but would make them completely different from "normal". This may be the right move, because they are different, but I think it would confuse people. It is also easier to use and add a lot of recipes if it uses the json files (when I a add a data provider). But this is not my decision to make. I will clean up the code and hopefully Forge can make a decision on if it's valid (or what would be needed to make it valid). The recipe book is a good point, but honestly, not sure if it's used enough by the people to warrant a problem. JEI is widely used and mostly unmissable in any modded minecraft world. The grindstone could be added to JEI (if the owner thinks it's good) to solve the recipe issue. But again, not sure what Forge's stance is on "essential" mods like this.

patches/minecraft/net/minecraft/inventory/container/GrindstoneContainer.java.patch
ferriarnus: How would I go about checking? I can't call the recipe if I don't have both items to check. Are the recipe items stored somewhere (like a list)?

ChampionAsh5357: No, but you can construct and invalidate a map for that. But again, this starts being overengineered and questions whether Forge should have it since this checking mechanic could be justified for or against depending on how much is implemented.

src/main/java/net/minecraftforge/common/crafting/GrindStoneRecipe.java
ChampionAsh5357: Yes, but those are not the name of the slots you gave it, you called it `base` and `addition`, which does not match the naming scheme here. Plus, they're just two generic ingredients as the grindstone could be considered a shapeless recipe with only two slots. It doesn't matter which goes in which. So, actually you'll probably need to reimplement the checking logic as well.

ChampionAsh5357:
> What do you mean with "event"?

Look at `AnvilUpdateEvent`. It  is Forge's current addition for a recipe system within anvils.

> JEI is widely used and mostly unmissable in any modded minecraft world. The grindstone could be added to JEI (if the owner thinks it's good) to solve the recipe issue.

This doesn't really solve the recipe issue. All it does is says to view recipes, a feature which vanilla adds natively, ForgeMod must rely on JEI to which JEI relies on ForgeMod. It's a circular dependency construct even though Forge suggests using JEI.

src/main/java/net/minecraftforge/common/crafting/GrindStoneRecipe.java
ferriarnus: Yes I agree, like I said, I copied the smithing table for it, which uses those tags. I will change it to match what it is for the grindstone.

ferriarnus:
The issue with an event as well is that the slots don't accept items. You said it before (and that its even a breaking change, though I'm not certain it's that severe, as vanilla still works), the slots in the gui will need to be patched to take other items. And deciding which is not easy, if you want only items used in recipes.

ChampionAsh5357:
> And deciding which is not easy, if you want only items used in recipes.

It doesn't break vanilla results, it just breaks expected behavior where you expect the grindstone to only allow items that it can actually grind. But again, this is really nuance so the justification for just allowing any item or only specific items is either or.

ferriarnus:
I think everything got fixed now (or at least all I could see). Is everything acceptable? Can/should I start on the anvil, or is it better to wait a bit for others to pick into the discussion of adding recipe types vs events?

ferriarnus:
So it's still a breaking change rigth? Should I port it to 1.17.1? Are there still issues?

patches/minecraft/net/minecraft/network/play/server/SUpdateRecipesPacket.java.patch
```diff
@@ -0,0 +1,11 @@
+--- a/net/minecraft/network/play/server/SUpdateRecipesPacket.java
++++ b/net/minecraft/network/play/server/SUpdateRecipesPacket.java
+@@ -46,7 +_,7 @@
+ 
+    }
+ 
+-   @OnlyIn(Dist.CLIENT)
++   //@OnlyIn(Dist.CLIENT)
```
ChampionAsh5357: I'm pretty sure you can't do this.

src/main/java/net/minecraftforge/client/gui/GuiHelper.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * A Class containing functions to help with gui's
+ */
+public class GuiHelper
+{
+
+    /**
+     * Adds Grinding recipes to the grinder. Takes the container, the IWorldPosCallable, the input and output inventory.
+     */
+    public static void grindingHelper(GrindstoneContainer container, IWorldPosCallable access, IInventory input, IInventory output)
```
ChampionAsh5357: Why couldn't the method go within `ForgeHooksClient` or some existing gui helper (pretty sure there is one somewhere)?

src/main/java/net/minecraftforge/client/gui/GuiHelper.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * A Class containing functions to help with gui's
+ */
+public class GuiHelper
+{
+
+    /**
+     * Adds Grinding recipes to the grinder. Takes the container, the IWorldPosCallable, the input and output inventory.
+     */
```
ChampionAsh5357: Javadoc is incorrect, missing params

src/main/java/net/minecraftforge/client/gui/GuiHelper.java
```diff
@@ -0,0 +1,56 @@
...
+    public static void grindingHelper(GrindstoneContainer container, IWorldPosCallable access, IInventory input, IInventory output)
+    {
+        if (input.getItem(0).isEmpty() && input.getItem(1).isEmpty())
```
ChampionAsh5357: This would be an or, both need to be present for a grindstone recipe.

src/main/java/net/minecraftforge/client/gui/GuiHelper.java
```diff
@@ -0,0 +1,56 @@
...
+            return;
+        }
+        access.execute((level, pos) -> {
+            Optional<GrindingRecipe> optional = level.getRecipeManager().getRecipeFor(ForgeMod.GRINDING, input, level);
+            if (optional.isPresent()) {
+                GrindingRecipe grindingRecipe = optional.get();
+                ItemStack result = grindingRecipe.assemble(input);
+                output.setItem(0, result);
+                container.broadcastChanges();
+            }
```
ChampionAsh5357: I would use `#ifPresent` over this.

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -112,7 +114,9 @@
     private static boolean enableMilkFluid = false;
     public static final RegistryObject<Fluid> MILK = RegistryObject.of(new ResourceLocation("milk"), ForgeRegistries.FLUIDS);
     public static final RegistryObject<Fluid> FLOWING_MILK = RegistryObject.of(new ResourceLocation("flowing_milk"), ForgeRegistries.FLUIDS);
-
+    
+    public static final IRecipeType<GrindingRecipe> GRINDING = new IRecipeType<GrindingRecipe>() {};
```
ChampionAsh5357, line 127: Since Forge is against static initialization, I would probably leave this null to be set during that particular phase, but this would need to be confirmed by someone on the triage or forge team since most of them usually just static init recipe types.

src/main/java/net/minecraftforge/common/crafting/GrindingRecipe.java
```diff
@@ -0,0 +1,120 @@
...
+public class GrindingRecipe implements IRecipe<IInventory>
+{
+    public static final IRecipeSerializer<GrindingRecipe> SERIALIZER = new Serializer();
+    
+    private ResourceLocation id;
+    private Ingredient base;
+    private Ingredient addition;
+    private ItemStack result;
+    
+    public GrindingRecipe(ResourceLocation id,Ingredient ingredient, Ingredient ingredient1, ItemStack result)
```
ChampionAsh5357: Once again with the mismatched param and field names. Make a decision to rename one or the other.

src/main/java/net/minecraftforge/network/VanillaConnectionNetworkFilter.java
```diff
@@ -100,4 +104,21 @@ private static SCommandListPacket filterCommandList(SCommandListPacket packet)
         });
         return new SCommandListPacket(newRoot);
     }
+    
+    /**
+     * Filter for SUpdateRecipesPacket. Filters out any ArgumentTypes that are not in the "minecraft" namespace.
+     * A vanilla client would fail to deserialize the packet and disconnect with an error message if these were sent.
+     */
+    @Nonnull
+    private static SUpdateRecipesPacket filterRecipes(SUpdateRecipesPacket packet)
```
ChampionAsh5357, line 122: You'll probably need to filter out advancements as well.

src/main/java/net/minecraftforge/common/data/GrindingRecipeBuilder.java
```diff
@@ -0,0 +1,141 @@
...
+ */
+public class GrindingRecipeBuilder
+{
+    private final Ingredient base;
+    private final Ingredient addition;
+    private final Item result;
+    private final Advancement.Builder advancement = Advancement.Builder.advancement();
+    private final IRecipeSerializer<?> type;
+    
+    public GrindingRecipeBuilder(IRecipeSerializer<?> type, Ingredient base, Ingredient addition, Item result)
```
ChampionAsh5357: Shouldn't be `Item` as data type, should serialize to `ItemStack` since that is supported.

src/main/java/net/minecraftforge/common/data/GrindingRecipeBuilder.java
```diff
@@ -0,0 +1,141 @@
...
+           this.advancementId = advancementId;
+        }
+
+        public void serializeRecipeData(JsonObject tag) 
+        {
+            tag.add("base", this.base.toJson());
+            tag.add("addition", this.addition.toJson());
+            JsonObject jsonobject = new JsonObject();
+            jsonobject.addProperty("item", Registry.ITEM.getKey(this.result).toString());
+            tag.add("result", jsonobject);
```
ChampionAsh5357: Again, serialize to `ItemStack`.

ChampionAsh5357 (commented):
This is not a breaking change afaik. You're not modifying anything pre-existing, just adding new material. The patches don't change anything of consequence either besides where to inherit behavior from. This should be ported to 1.17 though and kept on 1.16 as well. Otherwise, here are my current comments.

patches/minecraft/net/minecraft/network/play/server/SUpdateRecipesPacket.java.patch
KnightMiner: Forge has a [side only annotation stripper](https://github.com/MinecraftForge/MinecraftForge/blob/1.17.x/src/main/resources/forge.sas), I don't think commenting out only in is the way they do it

ferriarnus: Oh, didn't know it existed. I will change it.

src/main/java/net/minecraftforge/client/gui/GuiHelper.java
ferriarnus: I wasn't sure where it needed to go. Will move it

src/main/java/net/minecraftforge/client/gui/GuiHelper.java
ferriarnus: But there could be recipes with only one item (like gemcutting or something)

src/main/java/net/minecraftforge/common/ForgeMod.java
alcatrazEscapee: ~~Wait a minute this shouldn't be `new IRecipeType<>` in any case, it should be registering it (`IRecipeType#register`). (Notably, the registry in question is barely used, but still, there's a register method *right there*.~~ Nevermind, I found the registration~~

ferriarnus:
I've started adding the anvil (blacksmithing) recipe type. Since there was an event I simply used it to add it without a complicated patch. I still needed to change the experience cost of crafting, making it so you can actually craft items with no cost. I will start  on the data generators tomorrow. Any questions/remarks?

ferriarnus:
Everything should be in place right now, I think it's ready for review again.

ChampionAsh5357:
Closing this since it has been determined as out of scope for Forge. See Curle's explanation on the [brewing datapack PR](https://github.com/MinecraftForge/MinecraftForge/pull/7949#issuecomment-944211848) for more details.

PR #7871, opened 2021-07-17
[1.16 Hotfix] Fix missing userdev config for datagen runs

TheSilkMiner:
This PR fixes missing run configuration environment variables that cause data generation to break in a userdev environment after #6982.


marchermans:
This works on my end.
@TheCurle Can you verify this so we have 4 eyes looking over it

marchermans (approved):

Mysterious-Dev (approved):

PR #7872, opened 2021-07-17
[1.16.x] Adding TimeStamp to setCraftingPlayer

mcenderdragon:
Right now it is hard to check if a mod is not using setCraftingPlayer correctly, this PR adds a timestamp so a mod can easily check if the current player is up to date or made in the last tick (the time delta would be over 50ms so I would say it save to assume that the value is outdated)

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -31,6 +31,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.System;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -737,14 +738,21 @@ public class ForgeHooks
     }
 
     private static ThreadLocal<PlayerEntity> craftingPlayer = new ThreadLocal<PlayerEntity>();
+    private static ThreadLocal<Long> craftingPlayerTime = new ThreadLocal<Long>();
     public static void setCraftingPlayer(PlayerEntity player)
     {
         craftingPlayer.set(player);
+        craftingPlayerTime.set(System.currentTimeMillis());
     }
     public static PlayerEntity getCraftingPlayer()
     {
         return craftingPlayer.get();
     }
+    public static Long getCraftingPlayerUpdateTime()
+    {
+        return craftingPlayerTime.get();
+    }
+    
     @Nonnull
     public static ItemStack getContainerItem(@Nonnull ItemStack stack)
     {
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -19,6 +19,7 @@
 
 package net.minecraftforge.common;
 
+import java.lang.System;
 import java.lang.reflect.Field;
 import java.net.URI;
 import java.net.URISyntaxException;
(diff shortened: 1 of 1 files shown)


ChampionAsh5357:
So, how are modders using this outside of Forge? Currently, Forge only uses it to supply a player context for vanilla crafting recipes via the `ResultSlot` when taken. It is unset immediately after. So, I'm not sure how a mod would be using this if its implementation is to only give some player context to get the remaining results of a recipe.

If a mod is choosing to not unset the player after done crafting, that would be a bug in their code and should probably be addressed as an issue to them instead of providing a workaround to deal with those people who choose not to deal with it imo.

mcenderdragon:
> 
> 
> So, how are modders using this outside of Forge? Currently, Forge only uses it to supply a player context for vanilla crafting recipes via the `ResultSlot` when taken. It is unset immediately after. So, I'm not sure how a mod would be using this if its implementation is to only give some player context to get the remaining results of a recipe.
> 
> If a mod is choosing to not unset the player after done crafting, that would be a bug in their code and should probably be addressed as an issue to them instead of providing a workaround to deal with those people who choose not to deal with it imo.

absolutlty right, but how do you detect this? Detecting that the player set is wrong is currently not resly possible as far as I can tell. This is to find otu when this is happening and then blaming the correct mod.

KnightMiner:
Do you have an example of bugs caused by this issue that were fixed? For context about how common this issue is?

Also, I feel it would be better if `getCraftingPlayer` automatically performed the time check. However, I feel a condition like that is very unstable, for instance if a recipe has complicated code its possible 50ms would pass before fetching the player.

With the usage in forge, its impossible that the crating player would be wrong in any case, is there an example use in a mod where it might be wrong and cause this issue?

mcenderdragon:
> 
> 
> So, how are modders using this outside of Forge? Currently, Forge only uses it to supply a player context for vanilla crafting recipes via the `ResultSlot` when taken. It is unset immediately after. So, I'm not sure how a mod would be using this if its implementation is to only give some player context to get the remaining results of a recipe.
> 
> If a mod is choosing to not unset the player after done crafting, that would be a bug in their code and should probably be addressed as an issue to them instead of providing a workaround to deal with those people who choose not to deal with it imo.

absolutlty right, but how do you detect this? Detecting that the player set is wrong is currently not resly possible as far as I can tell. This is to find otu when this is happening and then blaming the correct mod.

pupnewfster:
Isn't the entire point that the player that is set is supposed to be the correct one? So if there are cases some mod is doing something absolutely stupid it should be reported to them, whereas otherwise as knight said you can't really make any assumptions related to time for if the data is correct or not (even if ideally mods won't have recipe logic that takes large amounts of time). And even if this was trying to get around mods being dumb by providing a rather inaccurate guess for if it is still valid or not, there still isn't an accurate way of knowing *which* mod is the one that never unset it.

Shadows-of-Fire:
I agree that some sort of sanity check is ideal, as this is pretty easy to break, but this timing check is not sufficient.  There are too many ways it could fail unintentionally.  This check also provides no accountability in the first place, so I'm not really sure what benefit it would provide (if the desired goal is to be able to blame mods for breaking it).

sciwhiz12:
Hello! @mcenderdragon, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

PR #7874, opened 2021-07-18
[1.16.x] Expand the DimensionSettings/NoiseGeneratorSettings codec to allow noise_settings jsons to specify that they should deep-copy another noise setting

Commoble:
## First, Some Background Info

The terms here are very similar, so the names of things will be clarified first:

| MCP Name | Mojang Name | Description
| --- | - | -
| DimensionSettings | NoiseGeneratorSettings | Objects created from jsons in `worldgen/noise_settings`, defining noise parameters and structure spacing entries.
| NoiseSettings | NoiseSettings | A set of noise parameters in a DimensionSettings/NoiseGeneratorSettings instance (*not* the top-level noise_settings json objects, confusingly)
| DimensionStructuresSettings | StructureSettings | An object containing the structure spacing entries, as well as some extra data specific to stronghold generation.

The rest of this document will use "noise settings" to refer to the objects generated from the top-level worldgen/noise_settings json objects unless otherwise stated.

### What's a Noise Setting?
Noise settings are one of the nine types of DynamicRegistries data, making them one of the fundamental parts of data-driven worldgen. Noise settings are loaded from the `worldgen/noise_settings` data folder and specify many parameters used by chunk generators to generate fancy terrain; each noise setting json causes one noise setting instance to be created

Noise settings are used by the Noise Chunk Generator, one of several types of chunk generators. Each Dimension has its own *unique* chunk generator instance; not all dimensions use the noise chunk generator, but most do (the overworld, nether, and end all use noise chunk generators, as well as most mods' dimensions that would use the standard terrain generation system). Noise Chunk Generators have a *reference* to a noise setting instance, so chunk generators that use the same noise setting share that same noise setting.

More importantly, a noise setting also specifies the separation settings for each structure, which determines the farapartness for structures of a given type of structure. A side effect of this is that a noise setting *must* have an entry for a given structure type for that structure to be able to generate in a dimension that uses that noise setting. Structure features are also part of biomes, so a structure must exist in both a biome and in a dimension's noise setting for structures to generate in biomes in that dimension.

This has caused a great deal of anguish for worldgen mods, especially when it is desired that structures should generate in biomes of one dimension but not in the same biomes used in a different dimension.

### Existing PRs Related to Noise Settings
#7817 includes a test mod as an example of registering noise settings to the "builtin" (vanilla) registries and modifying the structure entries in builtin noise settings, but these operations wouldn't be suitable for modifying noise settings jsons added by datapacks, which is something that we should be moving toward, rather than requiring mods' noise settings be created in java.

#7777 causes *every* dimension's chunk generator to have a deep copy of their noise settings' structure entries, whether needed or not. While this does help structure mods more easily cause their structures to only generate in specific dimensions, it removes the ability for structures to only be added to specific noise settings themselves, and breaks the assumption that two dimensions specifying the same noise setting are getting the same instance of data.

## What's This PR Do?
This PR first patches DimensionSetting's codec to be type-dispatched, allowing custom noise setting parsers to be used (similarly to how forge patches the model loader to allow custom model loaders). We then add a new forge registry for these custom parsers, and register A) the original vanilla codec, and B) a `forge:copy` codec. These patches are binary-safe; all references to the original vanilla noise setting codec will now point to the new type-dispatching codec, which is still a valid DimensionSettings codec and can still parse vanilla noise_settings jsons. To implement the type dispatcher, we also patch an extension interface into DimensionSettings with some default methods.

A noise setting json that uses the copy codec would use this format:
```json
{
    "type": "forge:copy",
    "parent": "minecraft:overworld"
}
```
In this example, we specify the ID of the overworld's noise setting json as the parent.

When the copy codec is used for the noise setting json, when the json is loaded via the worldgen json importer, a deep copy of all of the original's parameters and structure entries are made. If a dimension uses this noise setting, then modifications made to that noise setting after dynamic registries load will affect the structures in that dimension, but will not affect dimensions that use the "parent" noise setting. This is similar to the deep-copying done by #7777 but permits dimensions to share noise settings without deep-copying them if the copying is not desired (it also ensures that any dimensions that specify the same noise settings instance are indeed all getting the same noise settings).

This is only necessary for e.g. dimensions that have overworld biomes but don't necessarily want overworld structures for whatever reason.

## What's This PR Good For?
While this isn't tremendously useful by itself, it's one of (many) steps toward making modification of worldgen data more data-driven; e.g. a mod could add a structure entry to all noise settings in a specified set (which itself could be defined via some kind of tag-like data). These would themselves require separate PRs:
* If structure separation entries are not automatically deep-copied when dimensions/chunkgenerators are constructed, then noise settings can safely be modified when the server is about to start. This is, however, clunky, and an API for doing so could be created in a later PR.
* Alternatively, noise settings, biomes, and other dynamicregistry objects *could* be modified after dynamic registries load but before dimensions load; that would work even if chunk generators' structure separations are automatically deep-copied via #7777 but I haven't found a way to implement it without ensuring that binaries can't be broken, so such a hook probably wouldn't be viable for 1.16
* Adding tag-like data for noise settings and other dynamic registries objects can easily be added in 1.16+ and I have a separate PR I'm working on for that, though it would likely require one of the above two hooks to be useful

## Other things this PR adds:
* A new utility codec builder: `FallbackKeyDispatchCodec`. This is similar to standard DFU dispatch codecs, but accepts a fallback codec to use if no type field is found. This is important for this PR as A) the standard dispatch codec *requires* a type field, and would be unable to read vanilla noise setting jsons, and B) while we could implement a fallback with an Either codec, Either codecs have a flaw where if both the primary and fallback codec fail, then only the fallback codec's parse failure is reported. The fallback dispatch codec correctly reports parse errors when both codecs fail; if the json has a type field, then the dispatch codec's error is reported, and if the json does not have a type field, then the fallback codec's parse error is reported.
* Other mods can register new noise settings parsers to ForgeRegistries.NOISE_SETTING_CODECS in the manner in which they're accustomed. Classes that extend DimensionSettings should override the getType method in the IForgeNoiseGeneratorSettings to interface to return their registered codec type (this is not currently required).


sciwhiz12:
Hello! @Commoble, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

Commoble:
Noise generator settings got a bit weirder in 1.18 and I think I'd do what this is attempting to do substantially differently, yeah

PR #7876, opened 2021-07-19
[1.16.x][LTS] Add Resource Key Tags for biomes, dimensions, dimensiontypes, and noise settings; add an event that allows data-driven modification of biomes and noise settings

Commoble:
## Background Information
* The BiomeLoadingEvent cannot be used to read loaded datapack data to decide how to modify biomes or what to add to them, because vanilla biomes are constructed before datapacks are loaded.
* In addition to not being sensitive to datapack overrides, the BiomeLoadingEvent cannot use any kind of data-driven tags; the best we have at the moment is the BiomeDictionary, whose contents must be defined entirely in java.

This makes modifying biomes in a data-driven manner difficult as the only part of the forge API where the behavior of the BiomeLoadingEvent can be modified outside of java is in common configs, which aren't great for this purpose and can't really be used for mods to add compat to each other to the extent that datapacks can.

This PR adds two features to resolve these problems.

## Resource Key Tags
Since datapacks can create and register entirely new biomes, making a tag of biomes or other dynamic registry objects has different requirements and ramifications than making a tag of blocks. The Resource Key Tags added by this PR differ from standard (static registry) tags in important ways.

Resource Key Tags are a special kind of tag that represents a set of RegistryKeys (mcp name) or ResourceKeys (mojang name). Resource Keys (for those who aren't familiar with them) server a similar purpose to ResourceLocations; they're two ResourceLocations stapled to each other and a generic type parameter representing what kind of thing the key represents (this parameter is not used by the resource key itself; a RegistryKey<Biome> has no Biome fields in it). Resource Keys also have canonical instances, so only one ResourceKey for the PLAINS biome will ever exist.

Because resource key tags, like function tags, do not represent tags of static registry objects, we implement them differently than standard tags (as the standard tags assume that they are tags of static registry objects in several places), and have a separate loader for them.

Resource Key Tags are not resolvable during the BiomeLoadingEvent due to the event running before any kind of data is loaded, but they *are* usable and resolvable in the new DynamicRegistriesLoadedEvent.

Resource Key Tags are not synced and should only be queried for server purposes.

### Public APIs for Resource Key Tags
  * This allows mods to notify the tag loader to load tags from custom directories. Forge automatically marks vanilla's nine dynamic registries for loading resource key tags for (see ForgeMod for an example of doing this). This is safe to call during parallel modloading and can be called from a mod constructor. Mods that add their own loading directories should namespace them with their modid to avoid collisions. If two mods register two different directories for the same master registry key, an error will be logged (only one directory can be loaded per master registry key as the key is used to determine where to load key tags from).
 
It should be emphasized that ResourceKeyTags' values are `100% optional`. Resource Key Tags load before dynamic registry objects do, so a biome tag *cannot* verify whether any of the biomes their keys represent actually exist. Users of resource key tags should be wary of this when iterating over their contents, e.g. if using a biome tag in the DynamicRegistriesLoadedEvent below to add features to all biomes in the tag, then they will need to verify whether a given biome exists.

## DynamicRegistriesLoadedEvent
This event fires after dynamic registries (biomes, configuredfeatures, etc) load, but before dimensions/chunkgenerators/biomeproviders load from json, providing access to the dynamic registries (containing biomes and similar objects loaded from worldgen datapacks). It also provides an API for modifying biomes at this point, as well as adding structure spacings to noise settings (allowing structures to spawn in dimensions that use that noise setting). Resource Key Tags can be used to decide what biomes or noise settings to modify in this event.

This event also allows mods to retrieve configuredfeatures and similar objects from the dynamic registries and add them to biomes, meaning that unlike the BiomeLoadingEvent, this event allows mods to define their custom worldgen features entirely in json without needing to create or register those features in java first.

A test mod is provided with this PR; this test mod acts as an example of how to use resource key tags and this new event to add configuredfeature and configured structure feature jsons to existing biomes, without needing to create the configuredfeature or configuredstructure in java or register them to the builtin worldgen registries (it does create them in java for datagen purposes but this is optional).

## Other Things Added
* In order to implement the biome modification API, some extra biome builders were made that create deep copies of existing biomes to use as a base. These can also be used by mods for datagen purposes.
* A generic data provider that can be used to generate json files for configuredfeatures and other codec-based objects is added to assist developers with making their features more data-driven.

## Explanation of Patches
In order to fire the DynamicRegistriesLoadedEvent before dimensions load, we must fire it immediately after WorldSettingsImport imports datapacks into a dynamic registries instance. This works perfectly fine in dedicated servers, but during the singeplayer create world screen, datapacks can be can be imported multiple times into the same registries instance. This causes problems with modifications via the event happening to non-json biomes multiple times.

* Firstly, we add a WorldSettingsImport field to DynamicRegistries instances, allowing us to cache a WSI instance in the registries instance.
* Secondly, we have the WorldSettingsImport.create() static methods redirect to a new static method with context about why those methods are being called. This lets us determine whether a fresh dynamicregistries instance is being created, or datapacks are being imported into an existing set of registries.
* Thirdly, in this redirected WorldSettingsImport.create method, we check if we are importing datapacks AND datapacks have already been imported into the provided dynamicregistries instance. If so, we return a previously cached WorldSettingsImport instance.
  * Since WSI is a wrapper around a DynamicOps for parsing either json or nbt, and we must provide a WSI wrapped around the correct type when returning early, we also patch a map into WorldSettingsImport to be able to cache and return a wrapper of the correct dynamicops (vanilla already does this but just for jsonops).
* Finally, if we allowed a new WorldSettingsImport to be created and we are importing datapacks into the dynamic registries, we fire the dynamic registries loaded event. Since we've previously ensured that datapacks aren't imported into this registries instance if they already have been, the event is not fired more than once for the same set of registry imports. (changing or reorganizing datapacks in the create world gui doesn't cause issues here because vanilla creates a new registries instance with freshly recreated builtin objects before it imports these datapacks here)

## Other Concerns
* This PR doesn't cause any binary-breaking changes. This does definalize biome parameters, but the provided event fires before dimensions load, so mods adjusting biome parameters via that event shouldn't cause any problems for mods that assume that biomes won't change their state during or after dimensions load (which is well before the server starts).

## Relations to Existing PRs
#7777 should be generally be compatible with this PR as the dynamic registry event's modifications to noise settings occur before chunk generators are loaded from json.


Won-Ton:
Ah, shame you didn't keep the DRLoad hook in the place we talked about.

We've already discussed this on discord but just so the point's documented here:

ChunkGenerators are initialized before the world load event that this PR uses. By allowing modification to the DimensionSettings/Biomes _after_ the ChunkGenerators have been created you're basically dictating a contract where the ChunkGenerators can't actually rely on the information they've been instantiated with because it's all subject to being changed later in this new event.

Chunk-gen is async so the generators are typically immutable and therefore may need to, in their constructors, unpack data like biomes/features/structures/structure-settings into efficient internal data structures, process/filter/sort biome content, etc. If they want to support this event, they become very much restricted as to what information they can use during init.

If the plan is to get the DRLoadEvent into long-term 1.16 support I'd really urge for it be fired before chunkgenerators are init'd. Otherwise, good job and thanks!

Commoble:
I agree that it would be better if we were modifying biomes before chunk generators load, I wasn't able to find a good way to implement this event at the earlier timing without breaking (crashing on world load) existing mods in certain rare edge cases (curle saw my patch and veto'd it for 1.16.x). I'll be PRing the earlier timing to the 1.17 branch in any case once that enters public beta.

Commoble:
Spoke too soon, had a brain thing and found a way to patch WorldSettingsImports without having the breaky edge case. Updating the PR with the new timing now, will edit PR description shortly afterward

50ap5ud5:
Upvoted! This will be great for so many world gen mods, considering 1.16 will be in LTS for a while.

Commoble:
Closing this as the 1.18 PR has been split up into two PRs, so new backport PRs would be needed if those ever go in

PR #7877, opened 2021-07-20
ForgeGradle 5 and Gradle 7

SizableShrimp:
This PR updates the buildscript and MDK to use Gradle 7.1.1 and ForgeGradle 5. The buildSrc tasks have also been updated to utilize properties, and gradleutils is now used. ASM has been updated to 9.1 and the ASM API opcode has been bumped to ASM9 everywhere.


build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'net.minecraftforge.gradle:ForgeGradle:4.1.+'
+        classpath 'net.minecraftforge.gradle:ForgeGradle:5.+'
```
pupnewfster: With how breaking changes get handled for FG, I feel like it may almost make sense to just hold off on this slightly longer until this can be 5.1.+

mdk/build.gradle
```diff
@@ -4,7 +4,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath group: 'net.minecraftforge.gradle', name: 'ForgeGradle', version: '4.1.+', changing: true
+        classpath group: 'net.minecraftforge.gradle', name: 'ForgeGradle', version: '5.+', changing: true
```
pupnewfster: With how breaking changes get handled for FG, I feel like it may almost make sense to just hold off on this slightly longer until this can be 5.1.+ especially for purposes of the mdk to make it more clear about the .odd = stable, .even = breaking

PR #7878, opened 2021-07-20
add SChunkDataPacket to splitting list

RogueLogix:
In cases where a single chunk contains tens of thousands of tile entities this packet can be overflowed.

### DIFF
--- a/src/main/java/net/minecraftforge/network/ForgeConnectionNetworkFilter.java
+++ b/src/main/java/net/minecraftforge/network/ForgeConnectionNetworkFilter.java
@@ -33,6 +33,7 @@ import net.minecraft.network.ProtocolType;
 import net.minecraft.network.play.server.SAdvancementInfoPacket;
 import net.minecraft.network.play.server.STagsListPacket;
 import net.minecraft.network.play.server.SUpdateRecipesPacket;
+import net.minecraft.network.play.server.SChunkDataPacket;
 
 import com.google.common.collect.ImmutableMap;
 
@@ -69,6 +70,7 @@ public class ForgeConnectionNetworkFilter extends VanillaPacketFilter
         if (compatibility == VanillaPacketSplitter.RemoteCompatibility.V11)
         {
             builder.put(SAdvancementInfoPacket.class, ForgeConnectionNetworkFilter::splitPacket);
+            builder.put(SChunkDataPacket.class, ForgeConnectionNetworkFilter::splitPacket);
         }
         return builder.build();
     }


marchermans:
This will need more documentation, justification or at least a way to test this.

RogueLogix:
With my mod (Bigger Reactors), its feasible to have up to 50k-ish tile entities in a single chunk when building a max size reactor as they span about 3.2 million blocks (by default), and all reactor parts are tile entities due to how my multiblock system works. In this case, the overflow happens at about 32k tiles (I believe). 

The way I tested it was the same way I discovered it, which is with a reactor that has more than 32k tiles/parts in a single chunk. As soon as the server tries to send over that chunk, it will fail and disconnect the client.

marchermans:
> With my mod (Bigger Reactors), its feasible to have up to 50k-ish tile entities in a single chunk when building a max size reactor as they span about 3.2 million blocks (by default), and all reactor parts are tile entities due to how my multiblock system works. In this case, the overflow happens at about 32k tiles (I believe).
> 
> The way I tested it was the same way I discovered it, which is with a reactor that has more than 32k tiles/parts in a single chunk. As soon as the server tries to send over that chunk, it will fail and disconnect the client.

I need a way to test this without adding mods....
So anything in vanilla else this becomes a massive corner case we would need to discuss internally.

LexManos:
This sounds more like an issue with how your multiblock system works.
There are times where you have to work within the Minecraft engine's designs.
There are many things you can do that could make your multi-block structures not need tile entities.
Just telling the game to split packets is a hack. The only reason we do it for some existing vanilla packets is because their existing data driven systems are in place, and Mojang is taking their sweet time fixing the issues. 

It would be better to re-think your multi-block structure design.

RogueLogix:
So, I realized, this is a normal Vanilla issue too, commonly exploited on 2b2t, chunkbanning.
Put enough furnaces in a single chunk, same result.

Lex:
I had that discussion with someone else immediately when i asked about this issue on the discord, but not having them be tile entities isn't really an option, as its simply more complex than what non-entity blocks allows without there being other significant issues. If you would like to discuss it with me on discord, id be more than happy to.

RogueLogix:
_GitHub doesn't ping for label additions, so I thought this had just been sitting the entire time_

I see that the Needs Update tag was added months ago, that said I'm not sure if this does actually need updated. Mojang mitigated this themselves in 1.17 by increasing the packet size. It's still an issue if you go over that size, but its quite a bit harder to hit, even with the increased world size

sciwhiz12:
Hello! @RogueLogix, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

sciwhiz12:
@RogueLogix and I discussed this PR a bit in the Discord: this PR doesn't need to be updated to 1.18, as Mojang increased the packet limit in 1.17 from ~2MB to ~8MB, a 4× increase which will likely be enough even for mods on 1.18. However, this leaves this PR still open on 1.16.

I'm a bit split on this. On one hand, Lex brings up the point that packet splitting is a hackish workaround for large amounts of recipes, tags, and advancements that may be present in modpacks, and not meant for all packets (which is why it specifically lists which packets are split). 
However, there is also the point brought up that this PR would resolve an issue where, to my knowledge, malicious actors can flood a chunk with enough data that clients which receive the chunk data packet drop the connection due to the abnormal size -- this is colloquially called _chunkbanning_.

While I personally lean towards merging this PR to resolve that chunkbanning issue, I don't feel confident that this will be accepted by the maintainers as seen in the comment above by Lex. I ask other Triagers and interested parties, especially the @MinecraftForge/lts-legacy team, to discuss this PR in the discord, whether the benefit of solving chunkbanning here outweighs the use of the packet splitting workaround.

TheCurle:
Since chunkbanning isnt an immediate and present issue for modloaders (in fact, it's more of a server moderation issue than something we should be bothered with), this PR will not be puilled on that ground.

If it eventually does become a problem, we have other, less intrusive solutions for that problem. Until then, I see no need for this change.

As such, i'm closing this issue / PR / whatever.

PR #7879, opened 2021-07-21
Implements Lazy ItemStack Capabilities

marchermans:
### General:
Picks up where the original author of: #7830 left of.

Simple support for lazy capabilities by delaying the call to `ForgeEventFactory.gatherCapabilities(baseClass, this, parent);` to the first moment when `CapabilityDispatcher getCapabilities()` is called.

In the meanwhile this caches both the parent dispatcher and the potential cap data.

#### Simple oak log (Has no intrinsic capabilities of its own):
| Test type                   | Total time | 
|:--------------------------- |:----------:| 
| Lazy: Disabled / Caps: None |  8414 ms.  | 
| Lazy: Disabled / Caps: One  |  5680 ms.  | 
| Lazy: Enabled  / Caps: None |  3616 ms.  | 
| Lazy: Enabled  / Caps: One  |  3668 ms.  | 

#### Buckets (Have the fluid handler capability):
| Test type                   | Total time | 
|:--------------------------- |:----------:| 
| Lazy: Disabled / Caps: None | 20166 ms.  | 
| Lazy: Disabled / Caps: One  | 21638 ms.  | 
| Lazy: Enabled  / Caps: None |  6865 ms.  | 
| Lazy: Enabled  / Caps: One  |  3873 ms.  |


src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
```diff
@@ -22,35 +22,77 @@
...
+    {
+        if (isLazy && !initialized) {
+            lazyParentSupplier = parent == null ? () -> null : parent;
+            return;
+        }
+
+        doGatherCapabilities(parent == null ? null : parent.get());
+    }
+
+    private void doGatherCapabilities(@Nullable ICapabilityProvider parent) {
```
TheCurle: Brace on newline, please

patches/minecraft/net/minecraft/item/ItemStack.java.patch
```diff
@@ -45,7 +45,7 @@
     }
  
     private ItemStack(CompoundNBT p_i47263_1_) {
-+      super(ItemStack.class);
++      super(ItemStack.class, true);
 +      this.capNBT = p_i47263_1_.func_74764_b("ForgeCaps") ? p_i47263_1_.func_74775_l("ForgeCaps") : null;
 +      Item rawItem =
```
TheCurle, line 46: It's not clear whether this is an unfinished line or you're setting it to the field below.

src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
```diff
@@ -22,35 +22,77 @@
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
+
```
TheCurle: What's up with these import spaces?

src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
```diff
@@ -22,35 +22,77 @@
...
+    }
+
+    protected CapabilityProvider(final Class<B> baseClass, final boolean isLazy)
     {
         this.baseClass = baseClass;
+        this.isLazy = SUPPORTS_LAZY_CAPABILITIES && isLazy;
     }
 
-    protected final void gatherCapabilities() { gatherCapabilities(null); }
+    protected final void gatherCapabilities() {
```
TheCurle: Brace on newline, please

src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
```diff
@@ -22,35 +22,77 @@
...
+
+    private void doGatherCapabilities(@Nullable ICapabilityProvider parent) {
         this.capabilities = ForgeEventFactory.gatherCapabilities(baseClass, this, parent);
+        this.initialized = true;
     }
 
     protected final @Nullable CapabilityDispatcher getCapabilities()
     {
-        return this.capabilities;
+        if(isLazy && !initialized) {
```
TheCurle: Brace on newline, please

src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
```diff
@@ -22,35 +22,77 @@
...
+            doGatherCapabilities(lazyParentSupplier.get());
+            if (lazyData != null) {
```
TheCurle: Brace on newline, please

src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
```diff
@@ -90,6 +132,11 @@ public final boolean areCapsCompatible(@Nullable CapabilityDispatcher other)
 
     protected final void deserializeCaps(CompoundNBT tag)
     {
+        if (isLazy && !initialized) {
```
TheCurle: Brace on newline, please

TheCurle (changes requested):
Code seems good, just has some minor formatting issues.

patches/minecraft/net/minecraft/item/ItemStack.java.patch
TheCurle: Scratch that, i just can't read diffs.

marchermans:
@TheCurle Done.

Technici4n (changes requested):
This only delays the event dispatch until the first comparison, which usually happens instantly with `IItemHandler`. You need to override `areCapsCompatible` as well to make sure it doesn't dispatch events, otherwise there will be no actual performance improvement.

pupnewfster (approved):
This looks good, though I am going to hold off on assigning it until #7945 is merged as that is latest.

Though in relation to Technici4n's comment about `areCapsCompatible` I am going to leave this comment here to reduce the likely hood some future PR tries to override it, as doing so as suggested would invalid results in a few cases; most notably:
- If the two ItemStacks were saved with a non-matching set of mods installed that add capabilities, then they would have different `lazyData`, though initializing them would properly allow the two stacks to be compared.
- We are trying to compare a new stack with the current set of mods against a previously saved stack that either had more or less mods that contributed to the `lazyData`

Tslat:
This has caused a problem with tooltips for items with attribute modifiers.

Previously you could jump in on initCapabilities (yes, we shouldn't be, but that's all we had) and add a hideFlags tag to the stack, but you can no longer do this because initCapabilities isn't called until the item is in your hand.

This causes any items with attribute modifiers (particularly those with multiple equipment slots) to have massive tooltips as they can no longer hide their attribute modifiers.

I can't see any way to get around this besides hooking the tooltip events globally and listening in, which is rather inefficient.

This also breaks any capability for items that rely on inventory-ticking, since the capability isn't initialised until the player actually holds the item for the first time.

diesieben07:
> I can't see any way to get around this besides hooking the tooltip events globally and listening in, which is rather inefficient.

Forge could add `IForgeItem#getHideFlags`, which would allow an Item to provide additional hide flags in addition to what the stack tag has set. Alternatively more fine grained control could be added (`IForgeItem#shouldShowAttributeModifierInTooltip` or something) that allows hiding individual modifiers. Both should be fairly straightforward hooks to add.

> This also breaks any capability for items that rely on inventory-ticking, since the capability isn't initialised until the player actually holds the item for the first time.

I don't see how this is true. As soon as getCapability is called it will initialize the capability. i.e. it is initialized as soon as it is needed. If it "relies on inventory ticking" then whatever inventory tick hook must call getCapability and cause the capability to initialize immediately. Please elaborate on this.

sciwhiz12:
Noting that discussion regarding Tslat's comment was held [in the Discord](https://discord.com/channels/313125603924639766/852298000042164244/928543006536138802). From what I can tell, the conclusions were that:
 - The current behavior implemented by this PR is sufficient and working as intended;
 - @Tslat's current logic of adding the `hideFlags` NBT flag to the itemstack during capability initialization (`initCapabilities`) was called by @OrionDevelopment as being outside the intended behavior of the capability initialization method, as the method should _not_ be modifying the itemstack it is called on; and
 - A hook to control the tooltip elements of the itemstack -- including attribute modifiers, or any other element controllable via the `hideFlags` NBT flag -- would be preferrable instead to avoid needing to use the `hideFlags` NBT flag, which is usually intended for use by mapmakers for their custom items. _(I will note that Tslat declined making such a PR, so the idea is free to be made into a PR by anyone who wishes to contribute.)_

PR #7880, opened 2021-07-21
[1.16.x] Fix typo in FluidBlockWrapper

iTitus:
A little oversight in the FluidBlockWrapper class that was never detected.


iTitus:
If you want to put this into an omnibus PR instead, go ahead!

PR #7882, opened 2021-07-22
[1.16.x][LTS] Add a registry-friendly spawn egg constructor and an event for adding spawn eggs to the vanilla entitytype->egg map

Commoble:
## Background Information
While vanilla minecraft static-inits everything at once, forge registers things in a specific order. This occasionally requires improvements to vanilla types by adding supplier-based constructors to them.

While this generally works, spawn egg items are particularly problematic due to the eggs' constructor adding the egg's entity type to an entitytype->eggitem lookup map, which is used for the following things:

* When a creative-mode player middle-click-picks an entity, if that entity's entitytype is present in the egg lookup map, the associated spawn egg is added to the player's inventory.
* When vanilla item colors init (just before the item color registration event that mods use), every egg *value* in the lookup map has a color tinter assigned to it.
* When vanilla dispenser behaviors are created during vanilla registry bootstrap (before mods load), a vanilla dispenser behavior is added for each entry in the lookup map.
* Vanilla datagen uses the lookup map as well, but this is mostly ignorable.

This causes problems for mods that add spawn eggs, as the vanilla spawn egg map is difficult to reconcile with the supplier-in-constructor patterns. Generally, mods follow one of three patterns when creating spawn eggs:

* Extend SpawnEggItem to use a supplier, pass in a null entitytype to the base class, and override the type getter to use the supplier.
  * Entity picking will not work unless the mod manually adds their egg item to the lookup map in common setup.
  * A color tinter *must* be manually added during the color handler event.
    * The last mod to fail to do this will have an automatic color tinter registered by vanilla due to the item being assigned to the null key; this is considered inconsistent behavior.
  * Dispenser spawning behaviors must be manually registered, otherwise dispensers will dispense the egg like a normal item.
* Preinit the entity type before registration and construct a spawn egg in the vanilla manner. This pattern tends to be tolerated more then pre-registration-init of other things because it results in a relatively clean way to achieve consistency with vanilla spawn eggs (though it makes registry-replacement of the entity type more difficult).
  * Entity picking will automatically work
  * A color tinter will be automatically registered
  * Dispenser spawning behaviors must be manually registered, otherwise dispensers will dispense the egg like a normal item.
* Reimplement spawning-of-entity behavior in a separate item that does not extend SpawnEggItem.
  * The mod must use the middle-click event to manually implement entitytype picking.

All three of these patterns have flaws; this PR aims to provide an alternative method.

## Goals of This PR
* Patch sizes should be minimal.
* Binary compatibility with existing mods must be retained.
* Mods that add spawn eggs via existing standard patterns should continue to work; this PR should not cause unwanted changes to users of existing mods.
  * It is not a goal of this PR that existing mods whose spawn eggs have inconsistent or unstable behavior remains inconsistent or unstable.
* With the new APIs made available in this PR, mods should be able to add spawn eggs without preinitializing entitytypes before registration and without causing inconsistent behavior.
* It is not a goal of this PR to add additional APIs for registering dispenser-spawning behavior for spawn eggs, as dispenser behavior of mod's spawn eggs is already consistent (either they add a dispenser behavior for their spawn eggs and dispensers spawn mobs, or they don't add a behavior and dispensers eject eggs like a standard item (or they add a custom behavior)).
  * i.e. consider the potential existence of a server where a player's dispenser contraption is relying on the fact that a mod's spawn egg ejects like a regular item and doesn't spawn their mob

## What This PR Adds
* A very small patch to SpawnEggItem, which prevents spawn eggs from registering null entity types to the lookup map. This also deprecates the vanilla SpawnEggItem constructor.
* A supplier-based constructor for a new class that extends SpawnEggItem, which mods can use to construct spawn egg instances in a registry-friendly way.
* An event for adding mods' spawn egg items to the vanilla lookup map. If a mod specifies that their egg should be added to the lookup map via this event, that egg's associated entities will have the standard middle-click-pick behavior; on the client distribution, the standard item color tinting function will also become assigned to the egg. This provides a standard timing for adding spawn egg items to this map.

## Alternative PRs that add this feature
#7621 provides similar features; it patches SpawnEggItem more substantially, and adds automatic registration of dispenser spawning behavior.

### DIFF
--- a/src/main/java/net/minecraftforge/event/RegisterSpawnEggsEvent.java
+++ b/src/main/java/net/minecraftforge/event/RegisterSpawnEggsEvent.java
@@ -1,3 +1,22 @@
+/*
+ * Minecraft Forge
+ * Copyright (c) 2016-2021.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation version 2.1
+ * of the License.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
 package net.minecraftforge.event;
 
 import java.util.Map;
--- a/src/main/java/net/minecraftforge/event/RegisterSpawnEggsEvent.java
+++ b/src/main/java/net/minecraftforge/event/RegisterSpawnEggsEvent.java
@@ -1,0 +1,51 @@
+package net.minecraftforge.event;
+
+import java.util.Map;
+import java.util.function.Supplier;
+
+import net.minecraft.util.ResourceLocation;
+import net.minecraftforge.client.event.ColorHandlerEvent;
+import net.minecraftforge.common.ForgeSpawnEggItem;
+import net.minecraftforge.eventbus.api.Event;
+import net.minecraftforge.fml.event.lifecycle.IModBusEvent;
+
+/**
+ * Mod bus event that allows mods to mark their spawn eggs as behaving like vanilla spawn eggs.
+ * Marking an egg as a "standard mod egg" by adding it to this event's map will cause these things to happen:<br>
+ * 
+ * -- The egg item will be added to vanilla's entitytype->eggitem lookup map.<br>
+ * -- Creative players will be able to obtain the egg item by middle-click-picking associated entities<br>
+ * -- On client distributions, a standard color tinter will be assigned to the egg item. This will occur before the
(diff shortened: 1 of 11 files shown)


src/main/java/net/minecraftforge/event/RegisterSpawnEggsEvent.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Gets the (mutable) map of standard mod spawn egg suppliers.
+     * Egg suppliers added to this map will be considered "standard mod spawn eggs" -- they will be added to the vanilla
+     * EntityType->SpawnEggItem lookup map, will be obtainable in creative mode by middle-click-picking their associated entities,
+     * and will automatically have a color tinter assigned.
+     * 
+     * This does *not* automatically register the egg item to the dispenser behavior registry.
+     * @return map
+     */
+    public Map<ResourceLocation, Supplier<? extends ForgeSpawnEggItem>> getStandardModEggs()
```
KnightMiner, line 66: This seems to be exposing a bit too much data. Nothing would stop a modder from clearing the map here, or removing other eggs. All you really need is a method to add data to the map in my opinion.

Commoble: Being able to remove other eggs is the point (it allows addon mods that are associating non-vanilla behavior with the other mod's spawn egg to unmark them)

Commoble:
Closing this as an alternative solution was merged

PR #7884, opened 2021-07-22
1.17 minor pack related fixes

Cyborgmas:
Renamed classes to be in line with the mojang naming
Fixed an issue where UnionPath#toFile errors when the `ResourcePackFileNotFoundException` is thrown, but it is silenced later down the line for the pack.png, and throwing FileNotFoundException is silenced as well.
Bumped the pack format in the pack.mc meta for forge & the test mods
Filted out mod:minecraft from the pack names added to the DataPackConfig. This was causing the datapack screen to go through validation, even when no packs had been changed.

### DIFF
--- a/src/main/java/net/minecraftforge/fmlclient/ClientModLoader.java
+++ b/src/main/java/net/minecraftforge/fmlclient/ClientModLoader.java
@@ -62,8 +62,8 @@ import net.minecraftforge.client.model.ModelLoaderRegistry;
 import net.minecraftforge.common.ForgeConfig;
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.fmlclient.gui.screen.LoadingErrorScreen;
-import net.minecraftforge.fmllegacy.packs.DelegatingPackResources;
-import net.minecraftforge.fmllegacy.packs.ModFilePackResources;
+import net.minecraftforge.fmllegacy.packs.DelegatingResourcePack;
+import net.minecraftforge.fmllegacy.packs.ModFileResourcePack;
 import net.minecraftforge.fmllegacy.packs.ResourcePackLoader;
 import net.minecraftforge.fmllegacy.server.LanguageHook;
 import net.minecraftforge.forgespi.language.IModInfo;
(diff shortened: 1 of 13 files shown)


srs-bsns:
The pack.mcmeta version in the MDK could also be updated.

gigaherz:
This is one of those cases where it should ideally be 2 separate commits: one with actual fixes, and one with the style fixes/renaming.
As it is, it's very hard to find the actual change among the renames.

Cyborgmas:
great tip, re-did the commits

src/main/java/net/minecraftforge/fmlclient/ClientModLoader.java
```diff
@@ -62,8 +62,8 @@
 import net.minecraftforge.common.ForgeConfig;
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.fmlclient.gui.screen.LoadingErrorScreen;
-import net.minecraftforge.fmllegacy.packs.DelegatingResourcePack;
-import net.minecraftforge.fmllegacy.packs.ModFileResourcePack;
+import net.minecraftforge.fmllegacy.packs.DelegatingPackResources;
+import net.minecraftforge.fmllegacy.packs.ModFilePackResources;
```
LexManos: No to both of these.

src/main/java/net/minecraftforge/fmllegacy/packs/ResourcePackLoader.java
```diff
@@ -60,7 +60,7 @@ public static void loadResourcePacks(PackRepository resourcePacks, BiFunction<Ma
     }
 
     public static List<String> getPackNames() {
-        return ModList.get().applyForEachModFile(mf->"mod:"+mf.getModInfos().get(0).getModId()).collect(Collectors.toList());
+        return ModList.get().applyForEachModFile(mf->"mod:"+mf.getModInfos().get(0).getModId()).filter(n->!n.equals("mod:minecraft")).collect(Collectors.toList());
```
LexManos, line 64: What does this fix?

src/main/java/net/minecraftforge/fmllegacy/packs/ModFilePackResources.java
```diff
@@ -66,7 +66,7 @@ protected InputStream getResource(String name) throws IOException
     {
         final Path path = modFile.findResource(name);
         if(!Files.exists(path))
-            throw new ResourcePackFileNotFoundException(modFile.getFilePath().toFile(), name);
+            throw new FileNotFoundException("Can't find resource " + name + " at " + modFile.getFilePath().toString());
```
LexManos: Why change the exception type?

Cyborgmas: Because the vanilla exception needs a file, which can't be obtained from what i can tell.

src/main/java/net/minecraftforge/fmllegacy/packs/ResourcePackLoader.java
Cyborgmas: A logged error saying the datapack mod:minecraft is missing, this filtering was removed during the port to 1.17, not sure why.
vanilla resources are already loaded separately from this pack.
Edit: it also caused the datapackscreen to reload datapacks even when no changes to the enabled packs were made.

src/main/java/net/minecraftforge/fmllegacy/packs/ModFilePackResources.java
gigaherz: Yep, you can't call .toFile() on anything that isn't convertible to a native path string (or URL?)

PR #7887, opened 2021-07-23
[1.17.x] Extract fmlcommon ; remove gross symlink

ShroomAgent27:
This PR extracts FMLCommon as its own subproject and adds the proper dependencies to the projects that depend on it. 

As such, the symlink that linked the same code to two places has also been removed, as it is no longer needed.

[command line output of `./gradlew setup --info` and `./gradlew publish --info`](https://gist.github.com/shroomdog27/c49750e0c1f01ce4a88b03f467858214)
[intellij output of `./gradlew setup --info` and `./gradlew publish --info`](https://gist.github.com/shroomdog27/98354a90857b91d590231fa07b0cb4ba)
[eclipse output of `./gradlew setup` and `./gradlew publish`](https://gist.github.com/shroomdog27/95a529af39162f49c60297a9fef87d0d)

All of these outputs are from a fresh clone of my branch.


LexManos:
We've discussed this internally, and the symlink is the best option we have come up with because intellij doesn't allow two projects to reference the same folder.
Extracting this to an external project gets all kinds of screwey when it comes to dependencies on the Minecraft code, and classloaders.
Thats the point tho, this is code that is common to Forge and FMLOnly that can work with the Minecraft code.

We've decided that running the initial clone of git as admin {or making the symlink yourself} to work on FMLOnly {the symlink is not nessasary for working on Forge itself} is not a great ask.

PR #7890, opened 2021-07-23
[1.17] Add stack_sensitive transformer 

Bricktricker:
A port of #7638 to 1.17

This PR adds a stack_sensitive coremod transformer to change calls to non stack sensitive methods to stack sensitive methods,
similar to the field_to_instanceof transformer added in f169c99
Currently all calls to Item#canBeDepleted() (m_41465_) are replaced with calls to IForgeItem#isDamageable(ItemStack), but I can add more if needed.

The script does the following transformations:

    itemstack.getItem().foo() => itemstack.getItem().bar(itemstack)
    item.foo() => item.foo(this), if 'this' is an ItemStack
    void bizz(ItemStack stack){ item.foo() } => void bizz(ItemStack stack){ item.bar(stack) }

The script assumes the following things:

    The non stack sensitive method takes no arguments
    The stack sensitive method only takes only one argument, an ItemStack
    The non stack sensitive method and the stack sensitve method have the same return type
    The stack sensitive method is a interface method

The added findStackSensitiveChecks task in the build.gradle uses the output from reobfJar as input, while the findFieldInstanceChecks uses the output from the createJoinedSRG task.
It looks like createJoinedSRG jar does not have the forge patches applied. Because forge changes some method signatures
(e.g. here: https://github.com/MinecraftForge/MinecraftForge/blob/b027a92dd287d6810a9fdae4d4b1e1432d7dc9cc/patches/minecraft/net/minecraft/world/item/ItemStack.java.patch#L26-L31) these changes are not present in the createJoinedSRG created jar.
Therefore the generated JSON file would contain the old method signature (ItemStack(IItemProvider, int)) not the new one (ItemStack(IItemProvider, int, CompoundNBT)).
Is this correct or am I missing something?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1024,6 +1024,19 @@ project(':forge') {
             */
         }
     }
+    
+    task findStackSensitiveChecks(type: StackSensitiveFinder, dependsOn: ['reobfJar']) {
+        jar = tasks.getByName('reobfJar').output
+        output = rootProject.file('src/main/resources/coremods/stack_sensitive.json')
+        methods {
+            canBeDepleted {
+                name = 'm_41465_'
+                cls = 'net/minecraftforge/common/extensions/IForgeItem'
+                replacement = 'isDamageable'
+                blacklist 'net/minecraft/world/item/enchantment/EnchantmentCategory$10' //no stack available
+            }
+        }
+    }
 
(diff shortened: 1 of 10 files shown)


sciwhiz12:
Hello! @Bricktricker, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable. If we do not hear back from you after some time, this PR will be closed for inactivity.

PR #7891, opened 2021-07-23
[1.18.x] Update tags for new 1.17 and 1.18 content

### LINKED ISSUE #7982: [1.17] Vanilla Copper isn't registered in forge tags
Found this out when trying to make a copper chest from iron chests.

### LINKED ISSUE #8319: Forge Unintentionally Overrides Some Vanilla Block Replacement
**Minecraft Version:** 1.18.1

**Forge Version:** 39.0.8

**Steps to Reproduce:**
 1. Lay cluster of moss blocks with a single line of dirt blocks as indicator
 2. Place 2x2 spruce sapling cluster
 3. Maturate naturally or with bone meal

**Description of issue:**
Forge overrides podzol replacement when 2x2 spruce sapling cluster matures on a moss blocks. Other blocks, such as dirt, are replaced as normal. Issue occurs with or without modes installed, issue is not reproducible on clean Minecraft install. Blocks directly under tree is replaced, other block replacement issues have not been notice, but exhaustive testing was not carried out.

SuperMartijn642:
## Description:
This pull request updates the Forge data providers to include the new 1.17 content in tags and recipes.
#### New block tags:
- ores/copper
- tinted_glass
- storage_blocks/copper
- storage_blocks/amethyst
- raw_storage_blocks
- raw_storage_blocks/gold
- raw_storage_blocks/iron
- raw_storage_blocks/copper
- candles
#### New item tags:
- gems/amethyst
- ingots/copper
- raw_ores
- raw_ores/gold
- raw_ores/iron
- raw_ores/copper
#### Existing tags:
- added Cobbled Deepslate to `cobblestone`
- added Rooted Dirt to `dirt`
- added deepslate ore variants to the corresponding ore tags
- added Deepslate, Polished Deepslate, Infested Deepslate, and Tuff to `stone`
#### Recipes:
- replaced Copper Ingot with `ingots/copper`

## Remarks:
The tags `raw_storage_blocks` (Block of Raw Gold, Block of Raw Iron, and Block of Raw Copper) and `raw_ores` (Raw Gold, Raw Iron, Raw Copper) could have better names, but I can't think of anything better 😅
If there are new items or blocks I missed please let me know.


Mysterious-Dev (approved):

KnightMiner:
We should probably have a discussion on a few of these tags before just adding them, as a couple violate previous assumptions about ores.

Notably, `ores/copper` does not follow the previous rule that 1 block = 1 ingot (or raw ore more accurately), as copper drops roughly 3x the amount of raw ores as the other two. Should it still be tagged as `ores` or should we make a new prefix for that behavior?

`tinted_glass` also stands out, as we have the `glass` tag which is subdivided into colors. Giving it a separate tag sets a precedence that custom glass types (e.g. soul sand glass or reinforced glass) should be tagged separately from regular glass. I don't disagree with this, but just wanted to point out that adding that tag sets that precedent that was not there previously.

SuperMartijn642:
> Notably, `ores/copper` does not follow the previous rule that 1 block = 1 ingot (or raw ore more accurately), as copper drops roughly 3x the amount of raw ores as the other two. Should it still be tagged as `ores` or should we make a new prefix for that behavior?

That's a good point. I did not realize gold and iron ore still only dropped 1 raw ore.
I do think copper ore should be in the `ores` tag, as other ores such as redstone and diamond are also in there. Maybe adding two additional tags, like `rich_ores` and `poor_ores` for example, would be better.

> `tinted_glass` also stands out, as we have the `glass` tag which is subdivided into colors. Giving it a separate tag sets a precedence that custom glass types (e.g. soul sand glass or reinforced glass) should be tagged separately from regular glass. I don't disagree with this, but just wanted to point out that adding that tag sets that precedent that was not there previously.

I think tinted glass should also be part of the regular `glass` tag, I just forget to add it 😅. I think a separate tag, `tinted_glass` (or maybe `glass/tinted`), should still be added, as tinted glass' behavior is different from regular glass and mods might add their own variants of tinted glass.

I also forget about Block of Amethyst which should probably be under `storage_blocks/amethyst`.

KnightMiner:
> That's a good point. I did not realize gold and iron ore still only dropped 1 raw ore.
I do think copper ore should be in the ores tag, as other ores such as redstone and diamond are also in there. Maybe adding two additional tags, like rich_ores and poor_ores for example, would be better.

Iron and gold behave just like diamond, they drop 1, but drop more with fortune. Its a good point that copper basically behaves like redstone which is also in the tag. Main difference is copper ore sets a precedent for metals which did not exist before, as for example a tin ore block may drop 1 or may drop multiple, so having them in the same tag leaves a question of how to process them.

Adding in rich and poor ores might lead to clutter of tags and make it complicated to identify if a block is an ore in the first place (as you would likely need `rich_ores/copper`, so now there are several tags to check just to check if something is an ore, or a more complicated tag tree for them). One alternate suggestion here would be adding in `ores/rich`, and use set intersections to determine behavior (its tagged as `ores/rich` and `ores/copper`, so its worth 3 copper ingots). I have a custom ingredient in mantle that allows the intersection of two tags in a crafting recipe pretty easily if this approach is chosen, would be happy to PR it to Forge.

> I think tinted glass should also be part of the regular glass tag, I just forget to add it 😅. I think a separate tag, tinted_glass (or maybe glass/tinted), should still be added, as tinted glass' behavior is different from regular glass and mods might add their own variants of tinted glass.

Putting it as `glass/tinted` means we are tagging tinted glass as `glass` as well (or its inconsistent usage of nested tags). I am a bit more hesitant to do that as the composition of tinted glass is not just sand/glass, and some mods use the `glass` tag expecting just glass (e.g. tinkers melts glass blocks into molten glass). I could always change my recipe to accept colorless and any of the 16 dyed tags, or consider amethyst as basically a dye of course. However, if we do this we should describe what it means to be in the root `glass` tag if its not "made from exactly 1 glass block". I think using either `tinted_glass` and `glass/tinted` should be discussed a bit more.

----

For both of these, might be better to discuss on discord as I don't want to spam github.

src/main/java/net/minecraftforge/common/data/ForgeItemTagsProvider.java
```diff
@@ -111,9 +113,14 @@ public void addTags()
         copy(Tags.Blocks.ORES_QUARTZ, Tags.Items.ORES_QUARTZ);
         copy(Tags.Blocks.ORES_REDSTONE, Tags.Items.ORES_REDSTONE);
         copy(Tags.Blocks.ORES_NETHERITE_SCRAP, Tags.Items.ORES_NETHERITE_SCRAP);
+        copy(Tags.Blocks.ORES_COPPER, Tags.Items.ORES_COPPER);
         tag(Tags.Items.RODS).addTags(Tags.Items.RODS_BLAZE, Tags.Items.RODS_WOODEN);
         tag(Tags.Items.RODS_BLAZE).add(Items.BLAZE_ROD);
         tag(Tags.Items.RODS_WOODEN).add(Items.STICK);
+        tag(Tags.Items.RAW_ORES).addTags();
```
MelanX: This should include `RAW_ORES_GOLD`, `RAW_ORES_IRON`, and `RAW_ORES_COPPER` I guess.

SuperMartijn642: Jup, missed that 🤦‍♂️

marchermans:
Is this ready?

SuperMartijn642:
I think the changes in this PR, as of right now, are fine.

The only thing left is whether to add some additional tags to:
1. distinguish between ores that result in only one resource worth (iron, gold, diamond), more than one (copper, redstone), and less than one (nether gold ore)
2. distinguish between glass made from just sand (regular and stained glass) and glass made with other materials (tinted glass)

Proposed solutions:
1. add additional tags for ores, namely:
    - `ores/sparse` for ores that drop less than one resource worth,
    - `ores/singular` for ores that drop one resource worth, and
    - `ores/rich` for ores that drop more than one resource worth
2. add a `glass/silica` tag for glass made with mainly sand

I will bring it up on Discord again and update the PR based on that.

KnightMiner:
As another thought for tags for 1.17 content, since we now have both deepslate and stone ores, a tag catching what the ore is in would also be useful for mod compatibility. (e.g. `ores/deepslate` or `ore/in_deepslate`, and likewise for stone and netherack). Some mods like to output cobblestone as a byproduct for ores, and using that tag they could determine whether an ore should output cobblestone or cobbled deepslate for 1.17.

Is there a better spot to be discussing tags for 1.17 content, or Forge tags in general? We discussed it on discord earlier but no one on the forge team stated opinions, which makes it harder to gauge what will be accepted into forge.

SuperMartijn642:
I didn't really get any objections yesterday when I asked on Discord again, so I added the `ores/sparse`, `ores/singular`, `ores/dense`, and `ores/silica`.

> As another thought for tags for 1.17 content, since we now have both deepslate and stone ores, a tag catching what the ore is in would also be useful for mod compatibility. (e.g. ores/deepslate or ore/in_deepslate, and likewise for stone and netherack). Some mods like to output cobblestone as a byproduct for ores, and using that tag they could determine whether an ore should output cobblestone or cobbled deepslate for 1.17.

Although I can't find that many examples for 1.13+, most mods that add ores not in stone seem to just add them to the regular `ores/...` tags. This already causes some weird recipe outputs in Thermal Expansion, so I agree adding additional tags for it would be an improvement. Since the `ores/sparse`, `ores/singular`, `ores/dense`, already require some intersection, I don't think adding the `ores/in_...` tags would be an issue.
I do think the tags should contain the 'in_' prefix to avoid confusing them with the actual ores.

src/generated/resources/data/forge/tags/blocks/ores/dense.json
```diff
@@ -0,0 +1,9 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:coal_ore",
```
KnightMiner: Why is coal ore marked as dense? It drops a single item just like diamond

SuperMartijn642: Jup indeed 🤦‍♂️, thank you

src/main/java/net/minecraftforge/common/Tags.java
```diff
@@ -110,7 +114,12 @@ private static void init(){}
         public static final IOptionalNamedTag<Block> ORES_NETHERITE_SCRAP = tag("ores/netherite_scrap");
         public static final IOptionalNamedTag<Block> ORES_QUARTZ = tag("ores/quartz");
         public static final IOptionalNamedTag<Block> ORES_REDSTONE = tag("ores/redstone");
-        public static final IOptionalNamedTag<Block> ORES_COPPER = tag("ores/copper");
+        public static final IOptionalNamedTag<Block> ORES_SINGULAR = tag("ores/singular");
```
KnightMiner: Singular, dense, and sparse really could used javadocs to explain them. The user should not be required to have read this PR to understand the distinction

SuperMartijn642: Indeed, I hadn't added them because I wasn't completely sure were to put them as none of the other tags had javadocs.

KnightMiner: I was thinking normal field javadocs. There are probably older tags that could use docs, but that's out of scope here

src/generated/resources/data/forge/tags/blocks/candles.json
```diff
@@ -0,0 +1,22 @@
...
+    "minecraft:lime_candle",
+    "minecraft:magenta_candle",
+    "minecraft:orange_candle",
+    "minecraft:pink_candle",
+    "minecraft:purple_candle",
+    "minecraft:red_candle",
+    "minecraft:white_candle",
+    "minecraft:yellow_candle"
+  ]
+}
```
coolsimulations: Why does forge need to add it's own item & block tags for candles when the tag `minecraft:candles` already exists? Shouldn't mods that add their own candles simply add to the vanilla tag like they do with `minecraft:logs`?

src/generated/resources/data/forge/tags/blocks/ores/dense.json
```diff
@@ -0,0 +1,8 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:copper_ore",
+    "minecraft:lapis_ore",
+    "minecraft:redstone_ore"
```
coolsimulations: If the purpose of this tag is to identify all ores that have a 100% chance of more than 1 drop, then is there a reason why deepslate variants aren't included?

src/generated/resources/data/forge/tags/blocks/ores/singular.json
```diff
@@ -0,0 +1,12 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:ancient_debris",
+    "minecraft:coal_ore",
+    "minecraft:diamond_ore",
+    "minecraft:emerald_ore",
+    "minecraft:gold_ore",
+    "minecraft:iron_ore",
```
coolsimulations: Likewise with dense ores, should deepslate variants be included? Or does it not matter since there's going to be a `ores/in_` tag anyway that would separate this?

coolsimulations (changes requested):

src/generated/resources/data/forge/tags/blocks/candles.json
SuperMartijn642: I didn't see `minecraft:candles`, I must have overlooked it.

src/generated/resources/data/forge/tags/blocks/ores/dense.json
SuperMartijn642: Jup, you are correct 🤦‍♂️

Sinhika:
Before you get too locked into this, shouldn't "raw_ores" actually be "raw_metals"?

Another thought, why a "raw_storage_blocks" tag? The mechanics are the same as with non-raw metal ingots: 9 to 1. Why not just "storage_blocks/raw_copper", "storage_blocks/raw_gold", etc?  Let's not multiply entities unnecessarily.

Drullkus:
For Ore Quantity mechanics, I don't fully agree with the such like the `ores/singular` - It's no longer singular if you Fortune-mine, say, Iron Ore

pupnewfster:
Personally I also think that amount things probably shouldn't go directly in the ores tag so that it is easier for people to do more generalized assumptions if they want that everything in ores is an ore of a given type, and it should be something more along the lines of say `ore_rates/singular` or some other name

CAS-ual-TY:
>     
>     
>     1. distinguish between ores that result in only one resource worth (iron, gold, diamond),
>     more than one (copper, redstone), and less than one (nether gold ore)

Since you asked before **wether or not** this is needed, personally I would say this is not needed. To do this accurately you would probably have to overcomplicate things. An example for this is the raised problem when it comes to Fortune and singular ores then dropping multiple raw materials.

> 
> 
> Before you get too locked into this, shouldn't "raw_ores" actually be "raw_metals"?
> 
> Another thought, why a "raw_storage_blocks" tag? The mechanics are the same as with non-raw metal ingots: 9 to 1. Why not just "storage_blocks/raw_copper", "storage_blocks/raw_gold", etc? Let's not multiply entities unnecessarily.

I think "raw_materials" would be even better. Sometimes ores arent nececarilly containing metals, like Silicon Ore which is also relatively commonly added.

With the 2nd point I definitely agree.vIt should be "storage_blocks/raw_XX".

KnightMiner:
Fortune does not matter for the distinction between sizes of ores. Yes, fortune means it drops more than one, but the point is to tell the difference between copper style metal ores and iron style metal ores, as knowing mods some will add custom metal ores based on both styles. If the name singular is bad because of fortune, use a different name, but don't throw out the whole concept.

SuperMartijn642:
I added `ores_in_ground/deepslate`, `ores_in_ground/netherrack`, and `ores_in_ground/stone` so mods know which blocks could be used as, for example, recipe output for ores.
The javadoc explanation for `ores_in_ground/deepslate` is "Ores in deepslate which could logically result in deepslate as recipe output".

Since mods will also add for example `ores_in_ground/marble` there's a need to know which block 'marble' refers to. For this reason, I added `ore_bearing_ground/*` tags. Mods can find the blocks corresponding to the `ores_in_ground/x` tag in `ore_bearing_ground/x`.
The javadoc explanation for `ore_bearing_ground/deepslate` is "Blocks which the ores in the tag 'ores_in_ground/deepslate' are stored in".

Any feedback or better names and javadoc explanations for the tags would be appreciated 🙂

JaisDK:
Should there be an `ore_bearing_ground/end_stone` tag?

Sirttas:
> Should there be an `ore_bearing_ground/end_stone` tag?

I don't think so. This would be the first empty tag provided by forge AFAIK. Modders can always add it if they add ore in end stone.

KnightMiner:
The tag would not be empty, it would contain end stone (this is ore bearing ground, not ores in endstone that was suggested). That said, if someone wants to add endstone ore, they can define the tag and add end stone to it themselves

Sirttas:
> The tag would not be empty, it would contain end stone.

In that case all `ore_in_ground` should also contain the ground type? Or am I missing something special with end stone?

KnightMiner:
No, `ore_bearing_ground/endstone` contains end stone. `ore_in_ground/endstone` contains ores embeded in endstone. End stone is not special here.

In the same way, `ore_bearing_ground/stone` contains stone. `ore_in_ground/stone` contains all ores embedded in stone. `ore_bearing_ground/marble` would contain all generic marble blocks that would accept ores spawning in them. `ore_in_ground/marble` would contain all ore blocks embedded in marble.

----

I might suggest tweaking the java doc descriptions for the two. `ore_bearing_ground/netherrack` is really saying "this block could be replaced by a netherack ore during worldgen" in addition to having usage in recipes that take a block and want to craft an ore.

Sirttas:
Thanks my mistake I confused `ore_in_ground` and `ore_bearing_ground`.

Also some modpacks add tags like `forge:[tool]` like `forge:swords` should we add them in this PR?

JaisDK:
I know there's no use for the `ore_bearing_ground/end_stone`, it would just be for completeness sake and to prevent a tired modder from implementing `ore_bearing_ground/endstone`

SuperMartijn642:
> Also some modpacks add tags like `forge:[tool]` like `forge:swords` should we add them in this PR?

We would have to see if there are enough people that would benefit from those tags.

> I know there's no use for the `ore_bearing_ground/end_stone`, it would just be for completeness sake and to prevent a tired modder from implementing `ore_bearing_ground/endstone`.

The same argument could be made for, for example, basalt, since some mods add ores in basalt. I think just having mods add the tag when they use it would be better.

KnightMiner:
I would avoid adding tags that are unrelated to new content in 1.17 to keep the size of this PR down. The `ore_<X>_ground` tags were added because one of the 1.17 changes is overworld ores in two types of blocks, so might as well distinguish netherrack ores for completion as well.

Plus, tool tags are in most cases handled by tool actions, so I would discourage adding tags to avoid a mod using that to detect if a tool meets certain behaviors. If a mod wants to add a recipe using them they can add the tags as needed.

Sirttas:
Fair enough.

Should `ore_bearing_groud` tags be used for datagen? This would allow custom stones to be replaced by vanilla ore features. Or maybe it's better to do it in another pr.

JaisDK:
> might as well distinguish netherrack ores for completion as well.

Looks to me that all netherrack stuff is already in the PR

> Should `ore_bearing_groud` tags be used for datagen? This would allow custom stones to be replaced by vanilla ore features. Or maybe it's better to do it in another pr.

Do you mean oregen?

Sirttas:
> Do you mean oregen?

Yes worldgen sorry.

src/main/java/net/minecraftforge/common/data/ForgeRecipeProvider.java
```diff
@@ -77,6 +71,7 @@ protected void buildCraftingRecipes(Consumer<FinishedRecipe> consumer)
         replace(Items.GOLD_INGOT,   Tags.Items.INGOTS_GOLD);
         replace(Items.IRON_INGOT,   Tags.Items.INGOTS_IRON);
         replace(Items.NETHERITE_INGOT, Tags.Items.INGOTS_NETHERITE);
+        replace(Items.COPPER_INGOT, Tags.Items.INGOTS_COPPER);
```
ChampionAsh5357, line 74: What about those involving amethyst? Should they also be replaced/excluded as needed?

uecasm:
So, can we get this merged soon?  Does it just need a rebase, or is there another holdup?

src/main/java/net/minecraftforge/common/data/ForgeRecipeProvider.java
SuperMartijn642: As discussed on Discord, another PR should properly look into which items need to be replaced by tags.
For now, I did add the replacement for Amethyst as that seems to fall most in line with the other materials.

ChampionAsh5357 (approved):

src/generated/resources/data/forge/tags/items/ores/coal.json
```diff
@@ -1,6 +1,7 @@
 {
   "replace": false,
   "values": [
-    "minecraft:coal_ore"
+    "minecraft:coal_ore",
```
bageldotjpg: Shouldn't all the `ores/x` tags use the `#minecraft:x_ores` tags in them, like `#forge:ores/gold` does?

src/main/java/net/minecraftforge/common/Tags.java
```diff
@@ -104,8 +108,33 @@ private static void init(){}
         public static final IOptionalNamedTag<Block> GRAVEL = tag("gravel");
         public static final IOptionalNamedTag<Block> NETHERRACK = tag("netherrack");
         public static final IOptionalNamedTag<Block> OBSIDIAN = tag("obsidian");
+        /**
+         * Blocks which are often replaced by deepslate ores, i.e. the ores in the tag {@link #ORES_IN_GROUND_DEEPSLATE}, during world generation
+         */
+        public static final IOptionalNamedTag<Block> ORE_BEARING_GROUND_DEEPSLATE = tag("ore_bearing_ground/deepslate");
```
bageldotjpg, line 113: How is this tag different from `#minecraft:deepslate_ore_replaceables`?

src/main/java/net/minecraftforge/common/Tags.java
```diff
@@ -124,14 +165,19 @@ private static void init(){}
         public static final IOptionalNamedTag<Block> STAINED_GLASS_PANES = tag("stained_glass_panes");
         public static final IOptionalNamedTag<Block> STONE = tag("stone");
         public static final IOptionalNamedTag<Block> STORAGE_BLOCKS = tag("storage_blocks");
+        public static final IOptionalNamedTag<Block> STORAGE_BLOCKS_AMETHYST = tag("storage_blocks/amethyst");
```
bageldotjpg, line 167: Amethysts blocks aren't a storage block, they are crafted with 2x2 Amethysts and do not craft back into Amethyst

src/main/java/net/minecraftforge/common/data/ForgeBlockTagsProvider.java
```diff
@@ -52,47 +52,64 @@ public void addTags()
...
-        tag(STORAGE_BLOCKS).addTags(STORAGE_BLOCKS_COAL, STORAGE_BLOCKS_DIAMOND, STORAGE_BLOCKS_EMERALD, STORAGE_BLOCKS_GOLD, STORAGE_BLOCKS_IRON, STORAGE_BLOCKS_LAPIS, STORAGE_BLOCKS_QUARTZ, STORAGE_BLOCKS_REDSTONE, STORAGE_BLOCKS_NETHERITE);
+        tag(STONE).add(Blocks.ANDESITE, Blocks.DIORITE, Blocks.GRANITE, Blocks.INFESTED_STONE, Blocks.STONE, Blocks.POLISHED_ANDESITE, Blocks.POLISHED_DIORITE, Blocks.POLISHED_GRANITE, Blocks.DEEPSLATE, Blocks.POLISHED_DEEPSLATE, Blocks.INFESTED_DEEPSLATE, Blocks.TUFF);
+        tag(STORAGE_BLOCKS).addTags(STORAGE_BLOCKS_AMETHYST, STORAGE_BLOCKS_COAL, STORAGE_BLOCKS_COPPER, STORAGE_BLOCKS_DIAMOND, STORAGE_BLOCKS_EMERALD, STORAGE_BLOCKS_GOLD, STORAGE_BLOCKS_IRON, STORAGE_BLOCKS_LAPIS, STORAGE_BLOCKS_QUARTZ, STORAGE_BLOCKS_RAW_COPPER, STORAGE_BLOCKS_RAW_GOLD, STORAGE_BLOCKS_RAW_IRON, STORAGE_BLOCKS_REDSTONE, STORAGE_BLOCKS_NETHERITE);
+        tag(STORAGE_BLOCKS_AMETHYST).add(Blocks.AMETHYST_BLOCK);
         tag(STORAGE_BLOCKS_COAL).add(Blocks.COAL_BLOCK);
+        tag(STORAGE_BLOCKS_COPPER).add(Blocks.COPPER_BLOCK, Blocks.CUT_COPPER);
```
bageldotjpg, line 102: Cut Copper also isn't a storage block, it can't be crafted back into Copper Ingots.
Waxed Copper Blocks can be crafted back into Copper Ingots, but I'm not sure if it would count because there's also wax involved in the recipe.

src/generated/resources/data/forge/tags/items/ores/coal.json
SuperMartijn642: Yes, you are correct. I mixed up the 1.16 and 1.17 folders, so I didn't even realize a bunch of these new block tags existed 🤦‍♂️

src/main/java/net/minecraftforge/common/Tags.java
SuperMartijn642: Does not being able to craft amethyst blocks back into amethyst exclude it from being a storage block? Quartz blocks, for example, can also not be crafted back into quartz, yet it is a storage block for quartz.

src/main/java/net/minecraftforge/common/data/ForgeBlockTagsProvider.java
KnightMiner: Quartz is marked as a storage block in forge, and it's not reversible. My main concern with the tags is crafting equivalency, which copper and cut copper have (both worth 9 ingots)

If needed we could always consider more tags to distinguish "can be crafted into 9 ingots" from "can be crafted put of 9 ingots", but if we do so quartz should also be updated. I'm not convinced it's needed.

src/main/java/net/minecraftforge/common/Tags.java
bageldotjpg: Hmm, this complicates things. I'd argue Quartz Blocks aren't storage blocks since they don't "store it" because you can't get it back. Storing it implies getting it back for later use.

src/main/java/net/minecraftforge/common/data/ForgeBlockTagsProvider.java
bageldotjpg: I wasn't aware of this but as I said earlier Quartz I don't think should be considered a storage block - you can't 'store' the quartz, its used up by crafting the block with no way to get it back. Storing it implies being able to get it back

KnightMiner: I have expressed concerned about quartz in the past being under the storage blocks tag (my concern is the 3x3 vs 2x2 distinction), but as it stands that is the current standard, and I don't think this PR should try to change it. A separate RFC could discuss the issue, at this point I think its more consistent for cut copper to be in the storage blocks tag (I dont think recipes using copper blocks will object to cut copper blocks)

Raycoms:
Any progress on this?

TheLMiffy1111:
Personally I think the raw storage block tags should be in a separate directory

sciwhiz12:
@SuperMartijn642: Please re-target this PR to 1.18. I believe there aren't any new blocks/items which need to be added, though a quick pass may be good to make sure nothing falls through, and possibly a quick pass on existing Minecraft tags if there are any Forge tags that can also be removed (one that I saw: `minecraft:dirt` block tag replacing the existing `forge:dirt` block tag)

ChampionAsh5357 (approved):
I see no more issues on this, so I'll assign it now.

bageldotjpg:
I am still confused on the difference between Vanilla's `minecraft:deepslate_ore_replaceables` and the `forge:ore_bearing_ground/deepslate` tag here.

SuperMartijn642:
`forge:ore_bearing_ground/deepslate` is meant for the blocks which the deepslate ores are visually embedded in. This can be used in for example grinding recipes to give the correct block as output. For vanilla this doesn't matter as much, however if multiple mods add marble ore, the `forge:ore_bearing_ground/marble` tag can be used to obtain the corresponding marble blocks the ores are embedded in.
`minecraft:deepslate_ore_replaceables` also includes other blocks like tuff, which would not make sense in recipes using deepslate ores.

SuperMartijn642:
https://github.com/MinecraftForge/MinecraftForge/pull/7891/commits/38c91607ecdac1e43c88db7790a0a00e49a6d60f fixes #8319. Podzol can once again replace moss blocks just like vanilla.

TheLMiffy1111:
I still think raw storage blocks and storage block tags should be in separate directories for consistency with the existing tags

SuperMartijn642:
Raw ores can be crafted into blocks, just as other materials can be. I don't see a reason why these would be in a separate tag just for raw ores.

ChampionAsh5357 (approved):

TheLMiffy1111:
My reasoning is that the current format seems to have tags with material names directly under a directory, so making it `storage_blocks/raw_iron` would break this consistency. Perhaps `raw_storage_blocks/iron` or `storage_blocks/raw/iron` would be more consistent.

Eelt:
The format `storage_blocks/raw/iron` would be best because it would create `raw` as a subdirectory under `storage_blocks` in the event there's a use case where a modder would want only all the _raw_ storage blocks. 

Otherwise, the current `storage_blocks/raw_iron` would be better than `raw_storage_blocks/iron` as at least in the former the `raw_iron` block would still count as a storage block, while with `raw_storage_block` it wouldn't.

uecasm:
Deleting the `dirt` tag has been causing some backwards-compatibility grief.  Any chance that could be re-added to just forward to the vanilla tag, instead of removing it entirely?

pupnewfster:
Backwards-compatibility is not guaranteed before an RB

PR #7898, opened 2021-07-23
1.17 Cleanup Tag Loading

Cyborgmas:
There was some refactors of how tags were loaded, with LoaderInfo being a wrapper for a CompletableFuture, so instead of making the custom CF, the patch is simplified to create a LoaderInfo instead.


src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -398,4 +398,6 @@ public net.minecraft.world.level.newbiome.layer.Layers m_76729_(JLnet/minecraft/
 protected net.minecraft.world.level.portal.PortalForcer f_77648_ # level
 public net.minecraft.world.level.storage.LevelResource <init>(Ljava/lang/String;)V # constructor
 private-f net.minecraft.world.level.storage.loot.LootPool f_79028_ # rolls
-private-f net.minecraft.world.level.storage.loot.LootPool f_79029_ # bonusRolls
\ No newline at end of file
+private-f net.minecraft.world.level.storage.loot.LootPool f_79029_ # bonusRolls
+public net.minecraft.tags.TagManager$LoaderInfo
+public net.minecraft.tags.TagManager$LoaderInfo <init>(Lnet/minecraft/tags/StaticTagHelper;Ljava/util/concurrent/CompletableFuture;)V
```
SizableShrimp: These should be alphabetized to the right spot in the file with the empty line removed, then run checkAts.

patches/minecraft/net/minecraft/tags/TagManager.java.patch
```diff
@@ -39,3 +17,11 @@
              SerializationTags.m_13202_(tagcontainer);
              this.f_13478_ = tagcontainer;
           }
+@@ -73,6 +_,7 @@
+          }, p_144577_);
+          return new TagManager.LoaderInfo<>(p_144578_, completablefuture);
+       } else {
++         if (true) return net.minecraftforge.common.ForgeTagHandler.makeCustomTagTypeLoader(p_144578_, p_144576_, p_144577_, customTagTypeReaders);
```
pupnewfster: This seems to be silencing a vanilla warning. While I agree if we are able to make a custom tag type loader we shouldn't warn that it isn't found, but we probably should still have a warning if it failed (be it that it couldn't find the registry or that the registry doesn't support custom tag types).

src/main/java/net/minecraftforge/common/ForgeTagHandler.java
```diff
@@ -289,21 +282,6 @@ public static void populateTagCollectionManager()
...
-     * @apiNote Internal: For use by NetworkTagManager
-     */
-    public static void updateCustomTagTypes(List<TagCollectionReaderInfo> tagCollectionReaders)
-    {
-        ImmutableMap.Builder<ResourceLocation, TagCollection<?>> builder = ImmutableMap.builder();
-        for (TagCollectionReaderInfo info : tagCollectionReaders)
-        {
-            builder.put(info.tagType, info.reader.build(info.tagBuilders));
-        }
-        customTagTypes = builder.build();
```
pupnewfster, line 304: I haven't looked in great detail yet how tags work/changed in 1.17 but it seems like `ForgeTagHandler.customTagTypes` may not being updated anymore post reload due to the removal of this line?

Cyborgmas:
So small issue, I am adding the custom tags to the vanilla construct (TagContainer) so it gets serialized/deserialized by vanilla, i can patch in deserialization, but it wont work for vanilla clients however it only logs an error, it doesnt crash (at least up to the point i checked).
if the vanilla error logging is fine, ill try and continue with this, because I think it will reduce the overall code needed for custom tags to work.

pupnewfster:
For vanilla clients we could always use the to vanilla packet filter forge has to remove the "invalid" types from it. But if vanilla's tag changes make a cleaner/easier solution to custom tag types possible then it is probably worth doing so.

Cyborgmas:
Vanilla stopped hardcoding to 4 types of tags, allowing for a much easier handling of custom tag types. 
It's also completely vanilla compatible, the error i mentioned will only happen for non-vanilla registries.

patches/minecraft/net/minecraft/client/multiplayer/ClientPacketListener.java.patch
```diff
@@ -77,18 +77,17 @@
...
        PacketUtils.m_131363_(p_105134_, this, this.f_104888_);
        TagContainer tagcontainer = TagContainer.m_144449_(this.f_104903_, p_105134_.m_179482_());
 -      Multimap<ResourceKey<? extends Registry<?>>, ResourceLocation> multimap = StaticTags.m_13283_(tagcontainer);
 +      boolean vanillaConnection = net.minecraftforge.fmllegacy.network.NetworkHooks.isVanillaConnection(f_104885_);
-+      Multimap<ResourceKey<? extends Registry<?>>, ResourceLocation> multimap = vanillaConnection ? StaticTags.m_13283_(net.minecraftforge.common.ForgeTagHandler.withNoCustom(tagcontainer)) : StaticTags.validateVanillaTags(tagcontainer);//Forge: If we are connecting to vanilla validate all tags to properly validate custom tags the client may "require", and if we are connecting to forge only validate the vanilla tag types as the custom tag types get synced in a separate packet so may still arrive
++      Multimap<ResourceKey<? extends Registry<?>>, ResourceLocation> multimap = StaticTags.validateTags(tagcontainer, vanillaConnection);//Forge: If we are connecting to vanilla, custom tag types will be skipped over in validation as the server can't have them
```
pupnewfster: This comment (and I presume behavior based on what the comment is), is not actually quite right. The idea is that even for custom tags the connection is terminated if the tags are declared as being required. For custom tag types that the client doesn't necessarily require to be present the idea was to use forge's optional named tag system the same as how you can declare vanilla tag types on clients in a way that won't crash when connecting to a vanilla server/server without the mod.

patches/minecraft/net/minecraft/server/ServerResources.java.patch
```diff
@@ -12,7 +12,7 @@
  
     public void m_136179_() {
        this.f_136148_.m_13480_().m_13431_();
-+      net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.TagsUpdatedEvent.CustomTagTypes(f_136148_.m_13480_()));
++      net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.TagsUpdatedEvent(f_136148_.m_13480_()));
```
pupnewfster: This line/patch can just be removed as the above line fires the event in `bindToGlobal`

patches/minecraft/net/minecraft/tags/StaticTags.java.patch
```diff
@@ -21,42 +21,30 @@
...
- 
++   }
++
 +   @javax.annotation.Nullable
 +   public static StaticTagHelper<?> get(ResourceLocation rl) {
 +      return HELPER_MAP.get(ResourceKey.m_135788_(rl));
 +   }
 +
-+   public static Multimap<ResourceKey<? extends Registry<?>>, ResourceLocation> validateVanillaTags(TagContainer tagCollectionSupplier) {
++   public static Multimap<ResourceKey<? extends Registry<?>>, ResourceLocation> validateTags(TagContainer tagContainer, boolean isVanilla) {
```
pupnewfster: See my comment that I put where this is called, but also maybe this should be called `getAllMissingTags` the same as vanilla's method is (maybe even using an exec for letting the tag container param get mapped)

src/main/java/net/minecraftforge/common/ForgeTagHandler.java
```diff
@@ -21,54 +21,42 @@
...
 import com.google.common.collect.ImmutableSet;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Map.Entry;
-import java.util.Optional;
-import java.util.Set;
+
+import java.util.*;
```
pupnewfster: Forge tries to avoid wildcard imports (there are a couple other ones being added into this class as well)

patches/minecraft/net/minecraft/tags/TagManager.java.patch
```diff
@@ -8,34 +8,20 @@
...
-+            net.minecraftforge.common.ForgeTagHandler.reinjectOptionalTagsCustomTypes();
              SerializationTags.m_13202_(tagcontainer);
              this.f_13478_ = tagcontainer;
           }
+@@ -73,6 +_,8 @@
+          }, p_144577_);
+          return new TagManager.LoaderInfo<>(p_144578_, completablefuture);
+       } else {
++         if (net.minecraftforge.common.ForgeTagHandler.getCustomTagTypeNames().contains(p_144578_.m_144338_().m_135782_()))
++            return net.minecraftforge.common.ForgeTagHandler.makeCustomTagTypeLoader(p_144578_, p_144576_, p_144577_, customTagTypeReaders);
```
pupnewfster: Is there a reason this calls a one line helper method that doesn't even query anything native to ForgeTagHandler instead of just inlining the single line here? Doing so would allow for removing the ATs this PR is adding exposing LoaderInfo.

src/main/java/net/minecraftforge/common/ForgeTagHandler.java
```diff
@@ -246,166 +224,44 @@ public static void setCustomTagTypes(Set<ResourceLocation> customTagTypes)
...
-                return results;
-            });
-        }
-        return customResults;
-    }
-
-    /**
-     * Add all the missing optional tags back into the custom tag types tag collections
+     * Creates a map for serialization, copied from {@link TagCollection#serializeToNetwork} but adapted it since it requires
+     * a normal registry.
```
pupnewfster: I am not sure which is better/less maintenance, but in theory this could make use of `TagCollection#serializeToNetwork` by making use of the backing vanilla registries all the custom types have and passing it to that. See `ForgeRegistyTagProvider#wrapRegistry`.

src/main/java/net/minecraftforge/common/ForgeTagHandler.java
```diff
@@ -246,166 +224,44 @@ public static void setCustomTagTypes(Set<ResourceLocation> customTagTypes)
...
-
-        private TagCollectionReaderInfo(ResourceLocation tagType, TagLoader<?> reader, Map<ResourceLocation, Tag.Builder> tagBuilders)
-        {
-            this.tagType = tagType;
-            this.reader = reader;
-            this.tagBuilders = tagBuilders;
-        }
+            idMap.put(name, idList);
+        });
+        return new TagCollection.NetworkPayload(idMap);
```
pupnewfster: Given this method is just used in one place, if we end up going with this way of doing it over using the backing vanilla registry, this should probably just return the idMap so that `TagCollection.NetworkPayload` constructor does not have to be AT'd to being visible/public. (Either way the AT should be able to be removed as if we do switch to the `TagCollection#serializeToNetwork` method then we won't need access to the constructor either)

pupnewfster (commented):
I haven't done any testing yet, but there are some things that potentially should be addressed. In addition to my comments on specific parts of the code, the following things potentially should be changed as well due to having a similar concept/target as this PR:
- `ForgeRegistyTagProvider` should potentially have fallback handling for the new vanilla tag type of `game_event`. Though that may(?) be slightly outside the scope of this PR due to the fact the GAME_EVENT registry would need to be taken over by forge, which may be better suited to being in its own separate PR that primarily is targetting taking over various new 1.17 registries that forge can easily take over.
- The two `createOptional` methods we patch into all the vanilla tag types that existed in 1.16 should also be patched into the new type vanilla supports (GameEventTags)
- Similarly to how forge patches, `BlockTagsProvider`, `EntityTypeTagsProvider`, `FluidTagsProvider`, and `ItemTagsProvider`, vanilla's new `GameEventTagsProvider` should probably be patched with the extra constructor.

Cyborgmas:
So did not add GAME_EVENT to forge registries, that will be cleaner in another PR I think.
Renamed a bunch of methods, hopefully it's fine. Not sure about the ones in GameData though, thought it made more sense, but i can revert if needed

patches/minecraft/net/minecraft/data/tags/GameEventTagsProvider.java.patch
```diff
@@ -0,0 +1,12 @@
+--- a/net/minecraft/data/tags/GameEventTagsProvider.java
++++ b/net/minecraft/data/tags/GameEventTagsProvider.java
+@@ -11,6 +_,9 @@
+    public GameEventTagsProvider(DataGenerator p_176826_) {
```
pupnewfster, line 8: This constructor should be deprecated

src/main/resources/forge.exc
```diff
@@ -16,6 +16,7 @@ net/minecraft/data/tags/BlockTagsProvider.<init>(Lnet/minecraft/data/DataGenerat
 net/minecraft/data/tags/EntityTypeTagsProvider.<init>(Lnet/minecraft/data/DataGenerator;Ljava/lang/String;Lnet/minecraftforge/common/data/ExistingFileHelper;)V=|p_126517_,modId,existingFileHelper
 net/minecraft/data/tags/FluidTagsProvider.<init>(Lnet/minecraft/data/DataGenerator;Ljava/lang/String;Lnet/minecraftforge/common/data/ExistingFileHelper;)V=|p_126523_,modId,existingFileHelper
 net/minecraft/data/tags/ItemTagsProvider.<init>(Lnet/minecraft/data/DataGenerator;Lnet/minecraft/data/tags/BlockTagsProvider;Ljava/lang/String;Lnet/minecraftforge/common/data/ExistingFileHelper;)V=|p_126530_,p_126531_,modId,existingFileHelper
+net/minecraft/data/tags/GameEventTagsProvider.<init>(Lnet/minecraft/data/DataGenerator;Ljava/lang/String;Lnet/minecraftforge/common/data/ExistingFileHelper;)V=|p_126517_,modId,existingFileHelper
```
pupnewfster, line 19: This should be above ItemTagsProvider as G comes between F and I

patches/minecraft/net/minecraft/tags/TagContainer.java.patch
```diff
@@ -1,22 +1,21 @@
...
-+      net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.TagsUpdatedEvent.VanillaTagTypes(this));
++      net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.TagsUpdatedEvent(this));
     }
  
     public Map<ResourceKey<? extends Registry<?>>, TagCollection.NetworkPayload> m_144442_(final RegistryAccess p_144443_) {
+@@ -81,6 +_,7 @@
+       this.m_144436_(new TagContainer.CollectionConsumer() {
+          public <T> void m_142574_(ResourceKey<? extends Registry<T>> p_144481_, TagCollection<T> p_144482_) {
+             Optional<? extends Registry<T>> optional = p_144443_.m_6632_(p_144481_);
++            optional = optional.isPresent() ? optional : net.minecraftforge.common.ForgeTagHandler.getWrapperRegistry(p_144481_);
```
pupnewfster: I feel like this may be slightly cleaner moving the optional.isPresent() -> optional return into the getWrapperRegistry method

pupnewfster:
I believe this PR is ready to be assigned but as I made some changes myself to fix some issues I ran into I am going to label this as triage for now instead so that another triage member can review it/at least my changes.

A more complete summary of changes in this PR now that it is complete is as follows:
- Implement patches related to tags for GameEvent that already existed for the other 4 vanilla tag types
- Rewrite hooks for custom tags to be more inline with and take advantage of how vanilla does tag types due to vanilla not hard coding tags quite as much
- Fixed bug 1.17 porting bug causing custom tag types at least for forge patched in ones of vanilla registry types to not actually load due to not specifying the `tags/` sub directory when creating the tag type. Vanilla no longer implicitly includes the tags folder
- Breaking: Removed sub even types of TagsUpdatedEvent for Vanilla and Custom as they are all loaded in one place now with the above change instead of being slightly offset in timing
- Breaking: Removed no longer needed packets for syncing custom tag types as it makes use of vanilla's system now
- ~Breaking: Renamed methods for creating optional tags to be more in line with vanilla's naming scheme instead of still being based on mcp's naming scheme~
- ~Breaking: Removed patched in methods for BlockEntityType, Enchantment, and Potion for checking if they are in a tag. This follows the change vanilla made to Item and EntityType as there is not much benefit to having a `x#in(tag)` helper method instead of just directly calling `tag#contains(x)`~
- ~Breaking: Renamed/changed signatures slightly of the methods in GameData for getting the wrapper registries to not have a bunch of unused params (these methods shouldn't be getting called by modders in the first place, but mentioning it as a breaking change regardless)~
- Rerun test datagen as it never got reran after the tag folder for block entities got changed from when it was for tile entities
- Minor cleanup to the custom tag type test mod

Edit: Forgot to include the cases I tested:
- Forge client + Forge server
- Forge client + Vanilla server
- Vanilla client + Forge server

I did not test client/server pre this change with post this change iven the network sync is changing from custom to vanilla I suspect it may behave somewhat oddly and as breaking changes are still allowed (and this PR has a good number), it is not worth struggling with as I am unsure if there even is a way to make sure it is handled properly.

patches/minecraft/net/minecraft/tags/BlockTags.java.patch
```diff
@@ -4,12 +4,12 @@
        return f_13053_.m_13244_(p_13117_);
     }
  
-+   public static net.minecraftforge.common.Tags.IOptionalNamedTag<Block> createOptional(net.minecraft.resources.ResourceLocation name) {
-+       return createOptional(name, null);
++   public static net.minecraftforge.common.Tags.IOptionalNamedTag<Block> bindOptional(net.minecraft.resources.ResourceLocation name) {
++       return bindOptional(name, null);
 +   }
 +
-+   public static net.minecraftforge.common.Tags.IOptionalNamedTag<Block> createOptional(net.minecraft.resources.ResourceLocation name, @javax.annotation.Nullable java.util.Set<java.util.function.Supplier<Block>> defaults) {
```
LexManos: Honestly these changes are unneeded
Create, bind, of, could just called it 'optional'
We don't need to break even more things for no real reason.

src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -199,7 +199,7 @@ public static void init()
         return new RegistryBuilder<T>().setName(key.location()).setType(type).setMaxID(MAX_VARINT).hasWrapper().setDefaultKey(new ResourceLocation(_default));
     }
 
-    public static <T extends IForgeRegistryEntry<T>> MappedRegistry<T> getWrapper(ResourceKey<? extends Registry<T>> key, Lifecycle lifecycle)
+    public static <T extends IForgeRegistryEntry<T>> MappedRegistry<T> getWrapper(ResourceKey<? extends Registry<T>> key)
```
LexManos: Honestly, these changes are not needed. Yes they arguments are unused, but these are ease of use internal functions for our patches. Not something modders should be calling. Also, it seems completely unrelated to this PR.

marchermans (approved):

Issue #7558, opened 2020-12-20
Apple Silicon Mac GLFW error on startup


panther03:
**Minecraft Version:** 1.16.4

**Forge Version:** all versions from 35.1.4 to 35.1.13, as well as 35.0.0

**Logs:** https://paste.ee/p/lIYZ6

**Steps to Reproduce:**
 1. Simply install Forge with the vanilla minecraft launcher through Rosetta, or create a new blank forge instance in MultiMC, running either natively (with [this](https://gist.github.com/vortck/4ef2d8f1dd2a63689b82a5b05f7b1b37) guide) or also through Rosetta.
 2. Launch the game.

**Description of issue:**

Upon launch of any 1.16.4 version with Forge installed, using native or Intel LWJGL, a GLFW error causes the game to crash: 
```
java.lang.IllegalStateException: GLFW error before init:
[0x10008]Cocoa: Failed to find service port for display
```

[Some](https://gist.github.com/vortck/4ef2d8f1dd2a63689b82a5b05f7b1b37) say they have fixed this error using the argument `-Dfml.earlyprogresswindow=false`, but for me and several [others](https://forums.minecraftforge.net/topic/94803-javalangillegalstateexception-glfw-error-before-init-0x10008cocoa-failed-to-find-service-port-for-display/?do=findComment&comment=435153), adding the argument just causes the GLFW error to display in a popup window, and the error in the log is mostly the same:
![image](https://user-images.githubusercontent.com/9328518/102726774-f7aaa880-42e6-11eb-97f1-fb09ba99cd9b.png)
```
[17:14:59] [Render thread/ERROR] [mojang/GLX]: GLFW error collected during initialization: GLFW error during init: [0x10008]6123248856
[17:15:29] [Render thread/WARN] [minecraft/Main]: Failed to create window: 
net.minecraft.client.MainWindow$GlException: GLFW error 65544: Cocoa: Failed to find service port for display
```

This does not happen in Vanilla, and I have also observed it in many forge 1.15.2 versions. I have also observed that the instance will very rarely launch successfully. I have no good explanation for this, and it's extremely inconsistent, but for some reason, sometimes it just decides it will launch fine, and in these cases the game will be perfectly playable. 

[...]

PR #7908, opened 2021-07-24
[1.18.x] Update EarlyLoaderGUI to GL3.2 core

RogueLogix:
Updates rendering code used by the EarlyLoaderGUI (note: not the FMLEarlyLoadingWindow, this overlays when the Mojang logo appears) to a separated OpenGL 3.2 core compliant renderer. Also added support for dark mode colors (that was actually shrimp).

The renderer is separated out to allow the ClientVisualization (FMLEarlyLoadingWindow) to use the same rendering code if/when that gets re-added, file may need to be moved to provide that functionality.


src/main/java/net/minecraftforge/fmlclient/StartupMessageRenderer.java
```diff
@@ -0,0 +1,279 @@
...
+            glAttachShader(program, fragmentShader);
+            glLinkProgram(program);
+            
+            if (glGetProgrami(program, GL_LINK_STATUS) != GL_TRUE) {
+                System.err.println("Program linking failed");
+                System.err.println("Program link log\n" + glGetProgramInfoLog(program));
+                throw new IllegalStateException("Program link failed");
+            }
+            
+            glDeleteShader(vertexShader);
```
TheSilkMiner: Shaders need to be detached (`glDetachShader`) before they can be deleted

src/main/java/net/minecraftforge/fmlclient/StartupMessageRenderer.java
```diff
@@ -0,0 +1,279 @@
...
+            int blendSrcMode = glGetInteger(GL_BLEND_SRC_ALPHA);
+            int blendDstMode = glGetInteger(GL_BLEND_DST_ALPHA);
+            
+            glEnable(GL_BLEND);
+            glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
+            
+            glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, elementBuffer);
+            glBindBuffer(GL_ARRAY_BUFFER, dataBuffer);
+            
+            glBindVertexArray(VAO);
```
TheSilkMiner: Binding a vertex array already binds the given array and element array buffers, so why are you binding them again before?

src/main/java/net/minecraftforge/fmlclient/StartupMessageRenderer.java
```diff
@@ -0,0 +1,279 @@
...
+                glUniform2f(posMultiplierLocation, 2.0f / width, 2.0f / height);
+                // TODO: 21/06/12: handle text resizing better, this is ok, but its not perfect
+                float size = Math.min(width / 854, height / 480) * textSize;
+                glUniform1f(sizeLocation, size);
+                
+                List<Pair<Integer, StartupMessageManager.Message>> messages = StartupMessageManager.getMessages();
+                for (int i = 0; i < messages.size(); i++) {
+                    boolean nofade = i == 0;
+                    final Pair<Integer, StartupMessageManager.Message> pair = messages.get(i);
+                    final float fade = clamp((4000.0f - (float) pair.getLeft() - (i - 4) * 1000.0f) / 5000.0f, 0.0f, 1.0f);
```
TheSilkMiner: I would avoid computing this if `nofade` is true. It's not much of a time save, but at this point, might as well

src/main/java/net/minecraftforge/fmlclient/StartupMessageRenderer.java
```diff
@@ -0,0 +1,279 @@
...
+            
+            float[] memorycolour = hsvToRGB((1.0f - (float) Math.pow(pctmemory, 1.5f)) / 3f, 1.0f, 0.5f);
+            renderMessage(memory, memorycolour, 2, 1.5f, size, false, 1.0f);
+        }
+        
+        private void renderMessage(final String message, final float[] colour, int startingColumn, float row, float size, boolean fromBottom, float alpha) {
+            if (bufferLength < message.length() * 270) {
+                bufferLength = (int) (message.length() * 1.2 * 270);
+                MemoryUtil.memFree(charBuffer);
+                charBuffer = MemoryUtil.memAlloc(bufferLength);
```
TheSilkMiner: `memRealloc`

TheSilkMiner (approved):

src/main/java/net/minecraftforge/fmlclient/StartupMessageRenderer.java
RogueLogix: [https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glDeleteShader.xhtml](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glDeleteShader.xhtml)

> If a shader object to be deleted is attached to a program object, it will be flagged for deletion, but it will not be deleted until it is no longer attached to any program object, for any rendering context (i.e., it must be detached from wherever it was attached before it will be deleted). A value of 0 for shader will be silently ignored.

[https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glDeleteProgram.xhtml](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glDeleteProgram.xhtml)

> If a program object is in use as part of current rendering state, it will be flagged for deletion, but it will not be deleted until it is no longer part of current state for any rendering context. If a program object to be deleted has shader objects attached to it, those shader objects will be automatically detached but not deleted unless they have already been flagged for deletion by a previous call to glDeleteShader. A value of 0 for program will be silently ignored.

This is valid and defined to do.

src/main/java/net/minecraftforge/fmlclient/StartupMessageRenderer.java
RogueLogix: A VAO bind does not bind buffers to the GL_ARRAY_BUFFER or GL_ELEMENT_BUFFER binding points. The only time these binding points matter is at VAO specification. The GL 4.5 Direct State Access API makes whats actually going on much clearer. The buffers are bound for when they are updated later, avoiding multiple bind calls.

src/main/java/net/minecraftforge/fmlclient/StartupMessageRenderer.java
RogueLogix: Comment was actually regarding why i dont detach the objects after linking, so OpenGL can immediately delete the objects instead of doing so when the program is later deleted. _I forgot you can do that._

OroArmor:
You should use the GlStateManager methods and not the raw OpenGL calls. This provides compatibility if someone had the crazy idea of changing the rendering engine away from OpenGL.

RogueLogix:
This is designed to work before Minecraft's or Forge's classes are loaded in the FMLEarlyLoaderWindow, that part just doesn't exist (yet?) because of the peel.

OroArmor:
Ah ok. I'll need to take a closer look at how the early loading screen works. From a quick pass it looked more tied to minecraft than it is.

SizableShrimp:
FMLLoader loads forge and minecraft classes/jars. Loading those now would first of all not work second of all probably load it on the wrong classloader?

RogueLogix:
That's how its designed, not how its currently used. With the way its used here, it could use GlStateManager, but if i did so here, i would need basically duplicate rendering code anyway for the FMLEarlyLoadingWindow, so instead i just made this compatible in both instances.

The FMLEarlyLoadingWindow was removed in the peel, so that capability/second use doesn't currently exist, but if/when it does exist, it would be needed before Forge itself is loaded. I don't know how that's going to be done, hence me saying that the file may need to be moved.

marchermans (approved):
This looks good.
That said, I would like others to run some tests over this, in particular people with macs are an interesting case which should test this.

Which is also why I won't assign it at the moment.

RogueLogix:
I can say that I have personally tested this on Nvidia's windows driver, and AMD's Linux driver (open source stack). It has also been run through Renderdoc and Nvidia NSight, neither of which found any errors. However, I do not have a mac handy to test on, so someone else would need to test that.

marchermans:
> I can say that I have personally tested this on Nvidia's windows driver, and AMD's Linux driver (open source stack). It has also been run through Renderdoc and Nvidia NSight, neither of which found any errors. However, I do not have a mac handy to test on, so someone else would need to test that.

Yeah it is fully understandable that you can't test every configuration, and I ran it on NVIDIA Hardware on Windows, which means I was able to verify the basic construct and that is basically it. I still want others to test the implementation though.

One question I have though:
- Is it possible to render the normal splashscreen (mojang logo) instead of the white/dark background? So that it is more inline with the default screen.

sciwhiz12:
Hello! @RogueLogix, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable. If we do not hear back from you after some time, this PR will be closed for inactivity.

RogueLogix:
~~Appears no changes needed for 1.18, changed target branch~~
nope, github was slow to update, and didnt tell me there were merge conflicts, ill update my branch

RogueLogix:
Updated to 1.18.1

3TUSK:
> I do not have a mac handy to test on

I tested this PR several times with the following setup:

  - a 2020 MacBook Air w/ M1 chip
  - macOS Big Sur 11.6
  - (Adoptium) Eclipse Termurin JDK 17.0.1_2, the x86-64/amd64 build running via Rosetta
  - This PR in a Forge dev setup

Here are results:

  - The game client will run into that "65544: Cocoa: Failed to find service port for display“ error with this PR verbatim.
    - The game will start normally if I switched to another full-screen window immediately after I see the game window opens up. I can see the familiar FML progress tooltips on the bottom left. Not sure if this leads to inconsistent results; in my tests this hasn't failed yet.
    - If I force vanilla game ignore the error (in `Window#bootCrash`, remove the `throw` statement), the game actually launches just fine. 
    - Sometimes I can launch the game by putting a breakpoint in `EarlyLoaderGUI#renderTick` then manually resume the process, but this approach has inconsistent results: most of times I would still see "Failed to find service port for display".
  - The same environment can launch a Forge 1.18.1-39.0.10 userdev client just fine, without modifying the run configuration.
  - The same environment can launch a vanilla 1.18.1 client just fine, without modifying the installation profile (that is, use the launcher-bundled JRE instead).
  - The same environment can launch a 1.18.1 client with Forge 39.0.14 installed just fine, without modifying the installation profile (that is, use the launcher-bundled JRE instead).

@RogueLogix Let me know if you need further tests on Mac.

RogueLogix:
That appears to be an MacOS11/M1 specific error that will need MC shipping GLFW 3.3.3+ which is in LWJGL 3.3.0 to fix.

<https://github.com/MinecraftForge/MinecraftForge/issues/7558>
<https://github.com/glfw/glfw/pull/1833>
<https://github.com/LWJGL/lwjgl3/issues/601>

What I find more odd than that is that this PR causes it as no direct GLFW calls are made, any are done through Vanilla code used elsewhere in a normal vanilla run too. In regards to #7558 I wonder if the cause is the same as the 1.16 version interacts with GLFW in the same way this does.

I removed the only call that made any GLFW calls from this code, as it seems to be unnecessary anyway. That may fix this m1 issue. If it does, ill also open a PR to fix #7558 for 1.16.

3TUSK:
Tested ~10 times w/ the 5d26037, and so far the game works perfectly fine with the said setup above. I would call it as a working fix.

RogueLogix:
@OrionDevelopment that unblocks this with Mac tested.
This was developed with Nvidia on Windows (you also tested that), I've tested with AMD on Linux (mesa stack), and Mac (and M1) was just tested above.
Configs that could be tested to ensure complete compatibility would be Intel on both Windows and Linux, and AMD on Windows. Nvidia on Linux could potentially be skipped as the proprietary driver is the same stack as on windows, unlike Intel and AMD who have two driver stacks.

TheCurle:
After much pausing and waiting, we've decided that it's probably not worth the effort to maintain all of this legacy jank and the rather ugly UI of the old progress visualization.

I have some other plans upcoming for a replacement for all of this.

PR #7870, opened 2021-07-17
[1.16.x] Add support for forge's "remove" list to tag datagenerators

Commoble:
## Background

Forge expands the tag format to support a "remove" list, allowing tags to specify IDs of individual elements or IDs of tags to remove from that tag rather than adding to it. However, the current tag generators do not support this; tag generators cannot currently generate tag jsons that use the "remove" list. This PR expands the tag datagenerators to support this.

For clarification, there are two kinds of "tag builders" that this PR touches:
* Firstly is the ITag.Builder (sometimes called the "internal" or "raw" builder, which is used for both parsing tag jsons when data is loaded, and for writing tag jsons during datagen. This builder only works with the IDs of things and doesn't care about what the tags actually represent.
* The other one is the TagsProvider.Builder, which is exclusively used for datagenerating tags of registry objects. TagsProviders provide TagsProviders.Builders to callers; these are wrappers around ITag.Builders but with methods for making tags of registry objects.

## What's this PR do?
We patch ITag.Builder in the following ways:
* We implement a new IForgeRawTagBuilder forge extension interface to minimize the size of the patch
* We add a list field to ITag.Builder that items to add to the "remove" list in the json can be added to
* Where the builder serializes its data to a json, we add a line to serialize our extra removal data

IForgeRawTagBuilder has methods for adding to the removal list in a similar manner to the way ITag.Builder adds values to the value list.

We also add extra methods to IForgeTagBuilder (the extension interface of TagsProvider.Builder) to assist with adding registry objects and tags to the removal list in the same way it can add things to the standard values list.

The reason for adding these methods to both builders is that A) data generation of tags of registry objects should use TagsProviders and their builders rather than working with ITag.Builders directly, while B) tags of things that aren't registry objects can easily exist (see: function tags) and would necessarily need their own data providers (not TagsProviders) backed by ITag.Builders in a similar way.

PR #7911, opened 2021-07-24
[1.17] Add support for forge's "remove" list to tag datagenerators

Commoble:
This is the same as #7870 but ported to 1.17

## Background

Forge expands the tag format to support a "remove" list, allowing tags to specify IDs of individual elements or IDs of tags to remove from that tag rather than adding to it. However, the current tag generators do not support this; tag generators cannot currently generate tag jsons that use the "remove" list. This PR expands the tag datagenerators to support this.

For clarification, there are two kinds of "tag builders" that this PR touches:
* Firstly is the Tag.Builder, which is used for both parsing tag jsons when data is loaded, and for writing tag jsons during datagen. This builder only works with the IDs of things and doesn't care about what the tags actually represent.
* The other one is the TagsProvider.TagAppender, which is exclusively used for datagenerating tags. TagsProviders provide TagAppenders to callers; these are wrappers around Tag.Builders but with extra methods for making it easier to build the tag by specifying which objects the tag should represent.

## What's this PR do?
We patch Tag.Builder in the following ways:
* We implement a new IForgeRawTagBuilder forge extension interface to minimize the size of the patch
* We add a list field to Tag.Builder that items to add to the "remove" list in the json can be added to
* Where the builder serializes its data to a json, we add a line to serialize our extra removal data

IForgeRawTagBuilder has methods for adding to the removal list in a similar manner to the way ITag.Builder adds values to the value list.

We also add extra methods to IForgeTagAppender (the extension interface of TagsProvider.TagAppender) to assist with adding single entries and tag entries to the removal list in the same way it can add things to the standard values list.

The reason for adding these methods to both builders is that A) data generation of tags of registry objects should use TagsProviders and their appenders rather than working with Tag.Builders directly, while B) tags of things that aren't registry objects can easily exist (see: function tags or the resource key tags I've been working on) and would necessarily need their own data providers (not TagsProviders) backed by Tag.Builders in a similar way.


marchermans (approved):

src/main/java/net/minecraftforge/common/extensions/IForgeRawTagBuilder.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+    
+    /**
+     * Adds a tag proxy to the remove list to generate a json with
+     * @param proxy Tag proxy (either an item or another tag, usually)
+     * @return builder for chaining
+     * @apiNote internal; the other methods are preferable to call
+     */
+    default Tag.Builder removeProxy(final Tag.BuilderEntry proxy)
```
LexManos: All of these should just be called remove.
Just like all the add functions are named add.
The only changes are the conflicts in parameters.
So remove, removeElement, removeTag

Commoble: Fair enough, renamed the methods here as suggested. Also removed usages of apiNote per orion's recommendations.

patches/minecraft/net/minecraft/tags/Tag.java.patch
```diff
@@ -1,8 +1,13 @@
...
+@@ -47,8 +_,11 @@
+       return SetTag.m_13222_(p_13301_);
+    }
  
-    public static class Builder {
+-   public static class Builder {
++   public static class Builder implements net.minecraftforge.common.extensions.IForgeRawTagBuilder {
++      private final List<Tag.BuilderEntry> removeEntries = Lists.newArrayList(); /** Forge: Get the list of remove entries for datagen **/
++      public List<Tag.BuilderEntry> getRemoveEntries() { return this.removeEntries; } // also forge
        private final List<Tag.BuilderEntry> f_13302_ = Lists.newArrayList();
```
LexManos, line 16: Missed this in my last pass, this should return either a read only view, or like the vanilla getEntries, a stream. So that user's can't modify the value of the data objects without us knowing about it. Yes I know this means you'll need to move the implementation of one of the remove() functions to the patch. This is fine.

src/main/java/net/minecraftforge/common/extensions/IForgeTagAppender.java
```diff
@@ -66,4 +66,82 @@
...
+        }
+        return self();
+    }
+    
+    /**
+     * Adds a tag to the tag json's remove list. Callable during datageneration.
+     * @param tag The ID of the tag to remove
+     * @return The builder for chaining
+     */
+    default TagsProvider.TagAppender<T> removeTag(Tag.Named<T> tag)
```
LexManos: Again, all of these should just be 'remove'
The only reason you need two functions {or the name changes} is because of the conflict with removeRegistryEntry
However:
`remove(T, T...)` should compile to `remove(Object, Object[])`
and
`remove(Tag.Named, Tag.Named...)` should compile to `remove(Tag.Named, Tag.Named[])` saving the type issues.
And still be able to be called with a single argument.
So you can simplify all the methods to just 'remove'.

Commoble:
Renamed the rest of the remove methods to remove and rearranged the patch to provide a stream of the remove entries instead of a list per reviews

PR #7912, opened 2021-07-24
[1.17.x] Add a send datapack data to client event

SilverDavidMC:
This PR introduces the SendDatapacksToClientEvent to add a time with which modders can have their custom datapack data sent to a client player.

This new event fires server-side when a player logs in and when the /reload command is ran. Network handling and datapack registry data are expected to be handled by the modder using it.

The main purposes:
1) This event provides an easy centralized time with which modders can safely send datapack information to the client player. Existing ways of doing this may be unreliable at times as existing event hooks happen too early or too late.
2) Fixes issues with the `/reload` command. By allowing data to be properly synced during reloading, debugging and server management become much easier.
3) Maintains compatibility with mods that utilize data being sent to the client, such as JEI. Most mods that check for data being sent to the client wait for crafting recipes to be received. By firing this event before crafting recipes are sent, mods that do that are still able to function without any changes.


marchermans (approved):
This is useful and I wish we had this earlier.

LexManos:
What uses would this have?

KnightMiner:
This pull request is solving the same problem as #7410, which was targeting 1.16.3 (and needs an update).

The problem mods have is twofold:
* First, if a mod adds custom content to datapacks, the only way they can sync it is using the `PlayerLoggedIn` event, which means datapack changes are not synced on datapack reload. While reload is rarely done outside of data pack development, it can make it really hard to test datapacks for modded datapack elements.
* Second, recipe viewers such as JustEnoughItems currently load after the recipes/tag packets are received on the client. However, the `PlayerLoggedIn` event syncs long after the recipes load, which means recipes based on custom datapack elements do not properly show (and no event is fired for reload so that does not sync at all).

To solve the first problem, we need some event that fires when datapacks are reloaded, such as by the command. This pull request does so by adding an event to the reload hook.

To solve the second problem, we either need an event other than recipes loaded to alert recipe viewers that datapacks are fully synced (which would have to run after the player logged in event), or we need a way to sync data before recipes sync. This pull request solved the problem by adding a single event before recipes loaded,  #7410 solved the problem by adding an event before and after recipes loaded, with the intention that JEI would reload on the after event.

----

For an example of a mod that would use this event, Tinkers Construct in 1.16 defines its materials via JSON in datapacks. I currently have a ton of issues displaying material based items and recipes in JEI as there is no event I can sync my materials in on servers soon enough for JEI to show the recipes, and there is no event JEI can use that runs late enough to ensure my materials are synced. I also cannot sync material changes to datapacks when the reload command is run, requiring a full world reload just to reload datapack changes to materials.

gigaherz:
Hmm, the way I sync data in at least one of my mods, is by using the "apply" phase of the reload listeners:

1. In PlayerLoggedIn, you get the already-loaded data, and send it to the loggin-in player.
2. In the reload listener's apply phase, after you are done updating your data registries (maps or whatever), you send a packet to all players (there's a PacketDistributor for it), with the data.

So... I'm not sure what problem this PR solves exactly.

EDIT: After reading the above comment more carefully, it feels to me like the issue here isn't the act of sending the data itself, but rather being able to tell when all the mods have processed their data, so mods in both the server and client can update themselves.
If this is what this PR is intended for, I can see the usefulness of it, otherwise I'm not really seeing it.

KnightMiner:
I recall having issues with trying to sync in the apply method, as datapack reload in single player first happens on the client thread, and it caused a crash trying to access the player packet distributor. I don't recall exact details as that was several months ago (might even have been a 1.15 issue fixed in 1.16), just that it was a crash that happened when trying to load a integrated server. Might also have been a threading issue, and I needed to find the right place to defer the event, but I don't think datapacks are designed for task deferral. We can talk about this aspect more on Discord if you want

----

In response to the edit, I agree that the biggest benefit is just syncing data at a predictable time for the sake of intermod compatibility. My comment was describing the two benefits I saw to this: consistent timing and a consistent place to sync datapack data (notably consistent with vanilla timings, which should minimizes edge cases I keep hitting), the consistency in timing for syncing is definitely the one that will solve more problems of the two.

50ap5ud5:
Having a centralised time to sync data would be indeed be beneficial. 

I'm not sure if I am experiencing the same packet distributor accessing issue as KnightMiner's, but when you try to send sync packets to players in the apply phase of a reload listener, a NPE crash issue can occur during the initial loading of a world. 

During this time, datapacks are being reloaded multiple times to discover new world gen objects. The server has not started and the player packet distributor will be null. 

I believe the reloading of datapacks before the server has started is a factor in the packet distributor access issue.

I've seen many mods use ``ServerLifeCycleHooks#getCurrentServer`` for a null check before sending the sync packet in the apply phase of reload listeners. This seems to work as a temporary workaround, but having a proper location to send sync datapack information to clients is definitely beneficial for mod compatibility.

patches/minecraft/net/minecraft/server/players/PlayerList.java.patch
```diff
@@ -138,14 +141,18 @@
...
-@@ -786,6 +_,7 @@
+@@ -786,9 +_,11 @@
        }
  
        this.m_11268_(new ClientboundUpdateTagsPacket(this.f_11195_.m_129895_().m_144442_(this.f_11206_)));
 +      net.minecraftforge.fmllegacy.network.NetworkHooks.syncCustomTagTypes(this.f_11195_.m_129895_());
        ClientboundUpdateRecipesPacket clientboundupdaterecipespacket = new ClientboundUpdateRecipesPacket(this.f_11195_.m_129894_().m_44051_());
  
        for(ServerPlayer serverplayer : this.f_11196_) {
++         net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.SendDatapacksToClientEvent(serverplayer));
```
LexManos: So, two things. One are we wanting to garente this happens before or after tag sync? The two hooks make it happen at different times. I would vote before, I would also say that instead of firing this for every player on the server. I'd just make player @Nullable and pass in the PlayerList as well. So if the player == null send to all.
This will allow consumers to do optimizations like the one in this function where it shares the same packet across players. Also the name, it should be more 'OnDataConfigSentToClient" or something less stupid then what I just said. It's a generic 'this just happened' event not a 'you should do this thing.' event.

PR #7914, opened 2021-07-24
add AT entry for availableGoals

eerussianguy:
Modders need a way to remove goals from the goal list when creating entity AI to replace it with their own logic. To use `removeGoal` you need the actual goal instance, which is impossible a lot of the time. So we should just make this public and let modders remove stuff in easier ways.


ghost:
Is this really necessary when there is a getter for that set?
```java
public Set<WrappedGoal> getAvailableGoals() {
    return this.availableGoals;
}
```

eerussianguy:
OH. In 1.16 it used to be private, for some reason it slipped my mind to check if it turned public in 1.17. That's exciting then.

PR #7917, opened 2021-07-24
[1.17.x] Add Resource Key Tags for biomes, dimensions, dimensiontypes, and noise settings; add an event that allows data-driven modification of biomes and noise settings

Commoble:
This is essentially the same PR as #7876 but for 1.17; the description of that PR has been pasted below.

## Background Information
* The BiomeLoadingEvent cannot be used to read loaded datapack data to decide how to modify biomes or what to add to them, because vanilla biomes are constructed before datapacks are loaded.
* In addition to not being sensitive to datapack overrides, the BiomeLoadingEvent cannot use any kind of data-driven tags; the best we have at the moment is the BiomeDictionary, whose contents must be defined entirely in java.

This makes modifying biomes in a data-driven manner difficult as the only part of the forge API where the behavior of the BiomeLoadingEvent can be modified outside of java is in common configs, which aren't great for this purpose and can't really be used for mods to add compat to each other to the extent that datapacks can.

This PR adds two features to resolve these problems.

## Resource Key Tags
Since datapacks can create and register entirely new biomes, making a tag of biomes or other dynamic registry objects has different requirements and ramifications than making a tag of blocks. The Resource Key Tags added by this PR differ from standard (static registry) tags in important ways.

Resource Key Tags are a special kind of tag that represents a set of ResourceKeys. Resource Keys (for those who aren't familiar with them) server a similar purpose to ResourceLocations; they're two ResourceLocations stapled to each other and a generic type parameter representing what kind of thing the key represents (this parameter is not used by the resource key itself; a ResourceKey<Biome> has no Biome fields in it). Resource Keys also have canonical instances, so only one ResourceKey for the PLAINS biome will ever exist.

Because resource key tags, like function tags, do not represent tags of static registry objects, we implement them differently than standard tags (as the standard tags assume that their entries are entries in some Registry and make several assumptions that cannot hold for tags of resource keys), and have a separate loader for them.

Resource Key Tags are not resolvable during the BiomeLoadingEvent due to the event running before any kind of data is loaded, but they *are* usable and resolvable in the new DynamicRegistriesLoadedEvent.

Resource Key Tags are not synced and should only be queried for server purposes.

### Public APIs for Resource Key Tags
  * This allows mods to notify the tag loader to load tags from custom directories. Forge automatically marks vanilla's nine dynamic registries for loading resource key tags for (see ForgeMod for an example of doing this). This is safe to call during parallel modloading and can be called from a mod constructor. Mods that add their own loading directories should namespace them with their modid to avoid collisions. If two mods register two different directories for the same master registry key, an error will be logged (only one directory can be loaded per master registry key as the key is used to determine where to load key tags from).
 
It should be emphasized that ResourceKeyTags' values are `100% optional`. Resource Key Tags load before dynamic registry objects do, so a biome tag *cannot* verify whether any of the biomes their keys represent actually exist. Users of resource key tags should be wary of this when iterating over their contents, e.g. if using a biome tag in the DynamicRegistriesLoadedEvent below to add features to all biomes in the tag, then they will need to verify whether a given biome exists.

## DynamicRegistriesLoadedEvent
This event fires after dynamic registries (biomes, configuredfeatures, etc) load, but before levelstems/chunkgenerators/biomeproviders load from json, providing access to the dynamic registries (containing biomes and similar objects loaded from worldgen datapacks). It also provides an API for modifying biomes at this point, as well as adding structure spacings to noise settings (allowing structures to spawn in dimensions that use that noise setting). Resource Key Tags can be used to decide what biomes or noise settings to modify in this event.

This event also allows mods to retrieve configuredfeatures and similar objects from the dynamic registries and add them to biomes, meaning that unlike the BiomeLoadingEvent, this event allows mods to define their custom worldgen features entirely in json without needing to create or register those features in java first.

A test mod is provided with this PR; this test mod acts as an example of how to use resource key tags and this new event to add configuredfeature and configured structure feature jsons to existing biomes, without needing to create the configuredfeature or configuredstructure in java or register them to the BuiltinRegistries (it does create them in java for datagen purposes but this is optional).

## Other Things Added
* In order to implement the biome modification API, some extra biome builders were made that create deep copies of existing biomes to use as a base. These can also be used by mods for datagen purposes.
* A generic data provider that can be used to generate json files for configuredfeatures and other codec-based objects is added to assist developers with making their features more data-driven.

## Explanation of Patches
In order to fire the DynamicRegistriesLoadedEvent before dimensions load, we must fire it immediately after RegistryReadOps imports datapacks into a RegistryAccess instance. This works perfectly fine in dedicated servers, but during the singeplayer create world screen, datapacks can be can be imported multiple times into the same registries instance. This causes problems with modifications via the event happening to non-json biomes multiple times.

* Firstly, we add a RegistryReadOps field to RegistryAccess instances, allowing us to cache a RegistryReadOps instance in the registries instance.
* Secondly, we have the RegistryReadOps.createAndLoad() static methods redirect to a new static method with context about why those methods are being called. This lets us determine whether a fresh RegistryAccess instance is being created, or datapacks are being imported into an existing set of registries.
* Thirdly, in this redirected RegistryReadOps.forgeCreateAndLoad method, we check if we are importing datapacks AND datapacks have already been imported into the provided RegistryAccess instance. If so, we return a previously cached RegistryReadOps instance.
  * Since RegistryAccess is a wrapper around a DynamicOps for parsing either json or nbt, and we must provide a WSI wrapped around the correct type when returning early, we also patch a map into RegistryReadOps to be able to cache and return a wrapper of the correct dynamicops (vanilla already does this but just for jsonops).
* Finally, if we allowed a new RegistryReadOps to be created and we are importing datapacks into the dynamic registries, we fire the dynamic registries loaded event. Since we've previously ensured that datapacks aren't imported into this registries instance if they already have been, the event is not fired more than once for the same set of registry imports. (changing or reorganizing datapacks in the create world gui doesn't cause issues here because vanilla creates a new registries instance with freshly recreated builtin objects before it imports these datapacks here)

## Deprecation of Existing APIs
* BiomeDictionary has been deprecated by this PR as biome key tags serve a similar purpose and can be defined in datapacks.
* BiomeLoadingEvent has been deprecated by this PR as the dynamic registries loaded event also allows modification of biomes but is sensitive to datapacks.

## Other Concerns
* This PR doesn't cause any binary-breaking changes. This does definalize biome parameters, but the provided event fires before dimensions load, so mods adjusting biome parameters via that event shouldn't cause any problems for mods that assume that biomes won't change their state during or after dimensions load (which is well before the server starts).
* Though the thing that used to be called DynamicRegistries in the old class names is now called RegistryAccess in the mojang class names, the naming of the DynamicRegistriesLoadedEvent is kept as it's less generic than RegistryAccessLoadedEvent and helps distinguish it from loading of static registries, or from loading of unrelated datapack data such as loot tables and recipes.

## Relations to Existing PRs
#7777 should be generally be compatible with this PR as the dynamic registry event's modifications to noise settings occur before chunk generators are loaded from json.


patches/minecraft/net/minecraft/core/RegistryAccess.java.patch
```diff
@@ -9,3 +9,9 @@
        WritableRegistry<E> writableregistry = p_123072_.m_175512_(resourcekey);
  
        for(Entry<ResourceKey<E>, E> entry : registry.m_6579_()) {
+@@ -227,4 +_,5 @@
+          });
+       }
+    }
++   /** @apiNote forge internal use only */ @Deprecated @Nullable public net.minecraft.resources.RegistryReadOps<?> datapackImports = null;
```
marchermans: @apiNote is not supported as far as I know.

marchermans (approved):
This works remarkably well.

marchermans:
@Commoble Change the `@apiNote` thingies to something the default Javadoc compiler understands and I think this is good to go.

Commoble:
My bad, I saw apiNote in a few other places in forge and thought it was normal, but after doing more research they are indeed not part of the official javadoc spec (they're specifically part of OpenJDK, if I'm reading google right). Removed the apiNotes from this PR and marked a few more of this PR's internal-forge-use methods deprecated.

Angular-Angel:
Going to comment to say I think that biomes should have more tags and/or types - by specific tree type would be nice, for example. Also, the absence of a tag should NEVER be used to communicate information other than the absence of the tag. For example, with the tree-based tags, 'Specifying no tag implies a biome has temperate trees (Such as Oak)', but what if someone adds a new category, such as arctic, desert, or volcanic trees, or wants to tag their biome as having no trees at all? Now any biome that has those types/tags is going to also scan as temperate.

Edit: Also, checking !SAVANNA and !CONIFEROUS and !JUNGLE is way more annoying than just checking TEMPERATE.

Commoble:
> Going to comment to say I think that biomes should have more tags and/or types - by specific tree type would be nice, for example. Also, the absence of a tag should NEVER be used to communicate information other than the absence of the tag. For example, with the tree-based tags, 'Specifying no tag implies a biome has temperate trees (Such as Oak)', but what if someone adds a new category, such as arctic, desert, or volcanic trees, or wants to tag their biome as having no trees at all? Now any biome that has those types/tags is going to also scan as temperate.
> 
> Edit: Also, checking !SAVANNA and !CONIFEROUS and !JUNGLE is way more annoying than just checking TEMPERATE.

I included oak and birch forest tags in the forge biome tags in this PR for this reason specifically

patches/minecraft/net/minecraft/resources/RegistryReadOps.java.patch
```diff
@@ -1,14 +1,51 @@
...
++   }
++   /** @deprecated forge internal field -- ensures we can provide the correct RegistryReadOps instance **/ @Deprecated public final Map<DynamicOps<?>, RegistryReadOps<?>> extraOps;
++   public static <T> RegistryReadOps<T> forgeCreateAndLoad(DynamicOps<T> p_179871_, RegistryReadOps.ResourceAccess p_179872_, RegistryAccess p_179873_, boolean importingDataPacks) {
++      if (importingDataPacks && p_179873_.datapackImports != null) return net.minecraftforge.common.ForgeHooks.getCachedRegistryOps(p_179871_, p_179873_.datapackImports, p_179872_, p_179873_); // if we've already imported datapacks into these registries, return a previously cached registries instance instead of importing again
+       RegistryReadOps<T> registryreadops = new RegistryReadOps<>(p_179871_, p_179872_, p_179873_, Maps.newIdentityHashMap());
+       RegistryAccess.m_175500_(p_179873_, registryreadops);
++      if (importingDataPacks) { net.minecraftforge.common.ForgeHooks.onDynamicRegistriesLoaded(registryreadops, p_179873_); } // forge -- if we've imported datapacks into a fresh registries instance, fire the load event
+       return registryreadops;
+    }
```
alcatrazEscapee: This seems like an unnecessarily convoluted patch. The only hook here is actually in `createAndLoad(.. ResourceManager ..)`. `createAndLoad(.. ResourceAccess .. )` is redirected to `forgeCreateAndLoad` but effectively does nothing. Consider the following:

```java
// patch this method only
createAndLoad(.. ResourceManager ...) {
    if (importingDataPacks) // ...
    RegistryReadOps<T> ops = createAndLoad(RegistryReadOps.ResourceAccess.forResourceManager(manager);
    if (importingDataPacks) // ...
}

// no need to hook  createAndLoad(.. ResourceAccess .. ) because it's either called through a path that is intercepted above, or through ResourceAccess#builtin, which should be untouched.
// no need to add `forgeCreateAndLoad` since it's identical to `createAndLoad(ResourceManager)`, and not called anywhere else with a ResourceAccess
```

50ap5ud5:
Hi @Commoble, are there any plans to continue updating this PR? This would be amazing to have in the new versions and I'm sure many others would agree! :)

Commoble:
> Hi @Commoble, are there any plans to continue updating this PR? This would be amazing to have in the new versions and I'm sure many others would agree! :)

Been super busy with real life stuff. Eventually I need to look at the concern alcatraz raises; the code made sense to me at the time, but it's been a while. They're probably right but I really don't trust the world import lasagna.

patches/minecraft/net/minecraft/resources/RegistryReadOps.java.patch
Commoble: I agree with this analysis, refactoring this patch in the next push

Commoble:
The most recent push cleans up the RegistryReadOps patch re: alcatraz's suggestion, refactors the related forge hook accordingly, and adds a few more comments to clarify what the hook is doing.

With 1.18 around the corner, I'm not expecting this to be merged into 1.17.x, but I'll continue listening to feedback on this and I plan to make a similar PR for 1.18.x. I'll *probably* split this into two PRs for 1.18 (one for the tags and one for the dynamic registries hook) as a few people have suggested that

sciwhiz12:
Closing as 1.17 is now unsupported and this seems to have been superseded by two PRs for 1.18: #8251 for the resource key tags, and #8316 for an event to add biomes to multi-noise biome sources.

sciwhiz12:
I've just been informed that PR #8316 is unrelated to the changes of this PR, as that PR is for a 1.18-specific problem while this event is essentially a BiomeLoadingEvent replacement that is sensitive to datapack replacements.
However, with the agreement of the PR author, this PR will remain closed.

PR #7918, opened 2021-07-25
[1.17.x] Re-added RegisterDimensionsEvent

SilverDavidMC:
This PR brings back the RegisterDimensionsEvent to allow modders to register dimensions without needing to use datapacks to do so. 

It works by firing this event when the server starts, during `MinecraftServer.createLevels` and allows for modders to add to the current server's dimension registry. With this event, access to the overworld's seed is now available, allowing for modded dimensions to have different seeds, unlike with datapacks.


Commoble:
Wouldn't a PR to make it easier for datapack dimensions to be based on the overworld seed be better?

(I think you can already do that with a custom chunk generator but I'm not super familiar with the subject)

SilverDavidMC:
The primary point of this is to allow for dimensions to be easily created in code, just like how you currently can with biomes, configured features, and other parts of world gen datapack registry stuff. As of now, datapacks are required for creating dimensions, but they aren't required for anything else in worldgen. 

I'm sure the seed from a datapack dimension could be modified, but I just haven't looked into it fully for a good way to do it. That would likely be better off as a separate PR.

ChampionAsh5357:
> just like how you currently can with biomes, configured features, and other parts of world gen datapack registry stuff.

The reason why we create certain world gen features in code is because mojang decided to hardcode some features to use the static registry, so they need to exist within code. As for custom dimensions, they don't need to be, hence it makes no sense to declare them in code.

imo, all world generation should be done as much as possible through data gen as that is the direction Mojang is heading. There are very specific usecases where you would need to register a dimension like dynamic adding/removing for Hyberboxes or the Tardis mod, but not for anything else.

Commoble:
> The primary point of this is to allow for dimensions to be easily created in code, just like how you currently can with biomes

Biomes needing to be registered in code is a workaround due to mojang's half-implemented worldgen data system, not a feature. The best practice for biomes is already "register an empty dummy biome and override it with a json biome." If we had a way to add json biomes to the vanilla biome providers without needing to register them first then we'd do that.

> configured features, and other parts of world gen datapack registry stuff

Again, workaround -- the current BiomeLoadingEvent fires on vanilla biomes before datapacks load, and unregistered features can't be added to biomes, so worldgen features have to be created and registered and added to biomes all in java. This is a worse situation than the biomes themselves because the worldgen features added this way *can't* be overridden via datapacks, though I have a proposal to address this issue and allow json features to be added to biomes in #7917.

> As of now, datapacks are required for creating dimensions, but they aren't required for anything else in worldgen.

We're trying to move away from defining things in java and toward defining things in datapacks, not the other way around -- the idea being that behavior gets defined in java (like how ore veins generate) but sets of data (how big the veins are and what ore blocks they are) get defined in datapacks.

All that being said -- *if* this PR were to allow things that weren't possible with datapacks alone (e.g. adding indefinitely many dimensions at runtime ala mystcraft; I have a PR for an API for this that I'm working on for 1.17), I'd consider supporting it, but the things you can do with static datapack dimensions are very very large since any mod can define a custom chunk generator serializer.

marchermans:
So I don't think this will get pulled.

As a project we decided that we follow Mojangs lead when it comes to feature design, well as much as possible.
And this completely circumvents mojangs ideas of dimensions (or levels or what ever you want to call them) which are datadriven.

If you want to you talk to @Commoble and his open PR #7929 . Which completely supersedes this PR in my eyes, both on a functional and on a practical level.

marchermans (changes requested):
This PR completely circumvents mojangs ideas of how to deal with dimensions.

ChampionAsh5357:
This implementation breaks the contract of how vanilla now deals with world generation now as expressed by Orion. While I wouldn't say that Commoble's PR supersedes this one, their implementation provides an API for an edge case regarding dynamic dimension loading and unloading compared to alllowing static dimensions to be registered in code as well. For those reasons, this PR will be closed.

PR #7921, opened 2021-07-25
[1.17.x] Change `fluid` field in `LiquidBlock` from protected to private to match 1.16

Eelt:
Change `fluid` field from protected to private in `LiquidBlock` as per discussion on [Discord](https://discord.com/channels/313125603924639766/867851603468615740/868704757018153010) about crashes caused by `fluid` being protected in 1.17, while in 1.16 the field was private.

### DIFF
--- a/patches/minecraft/net/minecraft/world/level/block/LiquidBlock.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/block/LiquidBlock.java.patch
@@ -1,6 +1,12 @@
 --- a/net/minecraft/world/level/block/LiquidBlock.java
 +++ b/net/minecraft/world/level/block/LiquidBlock.java
-@@ -35,6 +_,7 @@
+@@ -30,11 +_,12 @@
+ 
+ public class LiquidBlock extends Block implements BucketPickup {
+    public static final IntegerProperty f_54688_ = BlockStateProperties.f_61422_;
+-   protected final FlowingFluid f_54689_;
++   private final FlowingFluid f_54689_;
+    private final List<FluidState> f_54691_;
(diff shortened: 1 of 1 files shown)


ChampionAsh5357:
1. This is not how you change the visibility of a method, use an access transformer.
2. Why does it matter whether the field is protected or private? It doesn't change anything about the issue of the problem. If you look within `field_to_method.js`, the coremod that usually addresses this has been commented out. So, the fix would be to properly reimplement this.

Eelt:
Yes, I know what an AT is, however, this PR is mostly a misunderstanding on my part. My bad.

Though, commenting on the discussion from Discord there's an ongoing issue with `LiquidBlock#fluid` in that anything that calls this using the non-deprecated constructor will have the fluid left as null, and anything that calls `LiquidBlock#fluid` (eg. `LiquidBlock#shouldSpreadLiquid`, `LiquidBlock#neighborChanged`, etc). Which causes various crashes relating to the fluid being null, preventing me from porting my fluids over to 1.17.

I guess this PR can be closed since it's mostly a misunderstanding on how to report this issue, [though, there might be value in keeping this open until this issue is fixed (under a different title?)](https://discord.com/channels/313125603924639766/867851603468615740/868737547214389269) .

ChampionAsh5357:
I did just explain where the error stems from between versions in point 2. As shrimp said, Lex was having problems with it so it was commented out. The proper thing to do here would be to reimplement correctly yourself or close the PR and make an issue instead with the associated error.

Eelt:
Thanks for the explanation, I'll convert this to an issue then

PR #7923, opened 2021-07-25
[1.16.5] Backport MC-209819

brandon3055:
Backport fix for mc bug https://bugs.mojang.com/browse/MC-209819

Under certain rare circumstances (Potentially more common in moded) if a player teleports away while an entity is tracking them the entity may still try to do a raytrace to the player's new position which may be thousands of blocks away. This can freeze the server long enough for the server watchdog thread to kill the instance. 
This PR simply backports the fix mojang applied in 1.17.


marchermans (approved):

phit:
any chance this can be looked at? this is a really common crash on servers

PR #7924, opened 2021-07-25
[1.17.x] Render events rework

RogueLogix:
Based off #7661

These events expand on and replace the functionality of RenderWorldLastEvent to allow rendering extensions to do their rendering at the correct stage of rendering.

Draft as I'm not sure if these should be the exact events used.


marchermans:
The technique of using events here is internally still not cleared up.
We understand why this kind of functionality is needed, but these events are fired 1000s of times per second, and we are not sure what the performance impact of this is.

Would it be possible for you to add a test mod that adds several of these handlers and then computes the additional time taken?
See here how such a testmod could work: <https://github.com/MinecraftForge/MinecraftForge/blob/69b9d1e37b56521559b3694e586656de5c2641a4/src/test/java/net/minecraftforge/common/capabilities/LazyCapabilitiesOnItemsTest.java> in principle

RogueLogix:
Ran some rudimentary tests using the new test mod on my machine.
numbers given are in us per frame to render the world, as an average over 20k frames, lower is better
```
Void world with 2 handlers:   174us
Void world with 400 handlers: 339us
Difference of 165us

Normal world with 2 handlers:     3331us
Normal world with 1000+ handlers: 2947us
Difference of -384us
```
Based on these tests, the performance impact of these events is negligible at most.

for more details on the test runs: https://discordapp.com/channels/313125603924639766/852298000042164244/869139580828418088

patches/minecraft/net/minecraft/client/renderer/LevelRenderer.java.patch
```diff
@@ -33,8 +33,12 @@
                 this.f_109466_.add(chunkrenderdispatcher$renderchunk1);
              } else {
                 this.f_109461_.m_91307_().m_6180_("build near");
-@@ -980,13 +_,14 @@
+@@ -978,15 +_,18 @@
+       FogRenderer.m_109018_(p_109604_, p_109601_, this.f_109461_.f_91073_, this.f_109461_.f_91066_.f_92106_, p_109605_.m_109131_(p_109601_));
+       FogRenderer.m_109036_();
        RenderSystem.m_69421_(16640, Minecraft.f_91002_);
++      profilerfiller.m_6182_("forge_render_level_begin");
```
Nightenom: Might want to move profiler calls to eventhandler.

Nightenom (commented):
The problem with #7661 approach is that it doesn't fire for rendertypes that have empty buffers and also it doesn't allow usage of custom rendertypes. I'm more in favour of the first approach in #7225

patches/minecraft/net/minecraft/client/renderer/LevelRenderer.java.patch
RogueLogix: https://github.com/MinecraftForge/MinecraftForge/pull/7924/commits/12bb5fdfa59c1d18cf362ce0bc3da22b838fb47a

RogueLogix:
A single event with the full RenderBuffers object does seem better than the events used here, but where should that event be fired from? Right before or right after block rendering seems like it would be fine, though I'm not sure how that would cover custom translucent render types.

robotgryphon:
@RogueLogix could definitely use this on 1.18. Would you be opposed to a port?

sciwhiz12:
Hello! @RogueLogix , please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable (or possibly agreeing to handing off the 1.18 PR to @robotgryphon). If we do not hear back from you after some time, this PR will be closed for inactivity.

RogueLogix:
I'd prefer a handoff as I'm not familiar enough with how B3D is used to decide on how some of these events should be done.
@robotgryphon or @Nightenom seem like the best two too hand off to.

sciwhiz12:
Closing as effectively superseded by #8603.

PR #7925, opened 2021-07-25
[1.15.x] Fix latest version not being changed for promotion file when updating

MelanX:
This is the missing part from the Jenkinsfile which is present in `1.16.x` and `1.17.x`.
This *should* fix the problem when a new 1.15 version is released, that it's not changed in https://files.minecraftforge.net/net/minecraftforge/forge/promotions_slim.json
Obviously, I couldn't test it.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -72,6 +72,11 @@ pipeline {
                     }
                 }
             }
+            post {
+                success {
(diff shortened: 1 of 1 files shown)


marchermans (approved):
Looks identical to the other branches now.....
Can not really test it, and also I am not sure how long this will still be relevant for.

LexManos:
Ya, this could be pulled, its not that big of a deal tho, It should be moved to use the correct utility: https://github.com/MinecraftForge/forgesharedlibrary/blob/master/vars/gradleVersion.groovy#L4
Which would fix things if we changed packages or whatever. But it's not that big of a deal for 1.15.

Issue #7915, opened 2021-07-24
[1.17.1] Hook for side-specific redstone connections


Da-Technomancer:
In MC1.16 (and before), overriding `Block::canConnectRedstone` allowed modded blocks to specify that redstone connections based on side.
An example use of this would be something like a repeater, where redstone connects on only two sides.
As of MC1.17, `Block::canConnectRedstone` is gone, and redstone dust connections are all-or-nothing based on `BlockBehaviour::isSignalSource`. Vanilla repeaters are special cased in `RedstoneWireBlock::shouldConnectTo`, which mods don't have access to.

As of 1.17, mods cannot cleanly add redstone components similar to repeaters.

I suggest that Forge re-implements `Block::canConnectRedstone` or something similar in 1.17.

rikka0w0:
#8014 adds the fully functional `canConnectRedstone` hook, the 1.17.x version will behave as it was back in 1.16.5.

ChampionAsh5357:
Closed as of #8014.

PR #7927, opened 2021-07-25
[1.17.x] Redstone dust now defers to the blockstate for determining connectivity

Da-Technomancer:
Implements #7915 

Makes redstone dust defer to a new method in IForgeBlockState (which calls a new method in IForgeBlock) to determine whether it should connect to a block on a side. 
The new method in IForgeBlock defaults to the original vanilla logic, so this has no gameplay impact on vanilla features without mods.


marchermans (changes requested):
Please add a test mod showing this feature in action.

patches/minecraft/net/minecraft/world/level/block/RedStoneWireBlock.java.patch
```diff
@@ -0,0 +1,11 @@
+--- a/net/minecraft/world/level/block/RedStoneWireBlock.java
++++ b/net/minecraft/world/level/block/RedStoneWireBlock.java
+@@ -363,7 +_,7 @@
+       } else if (p_55595_.m_60713_(Blocks.f_50455_)) {
+          return p_55596_ == p_55595_.m_61143_(ObserverBlock.f_52588_);
+       } else {
+-         return p_55595_.m_60803_() && p_55596_ != null;
++         return p_55595_.canConnectRedstone(p_55596_);
```
marchermans: A better name would be `canRedstoneConnectTo`

marchermans (changes requested):
I would choose a different name for the new method.

Da-Technomancer:
I've applied the suggested changes.

Da-Technomancer:
Did a quick commit to fix the merge conflict on mods.toml. 
This PR should be merge-ready at this point.

rikka0w0:
Is it possible to make this hook be aware of world and blockpos (The old `boolean canConnectRedstone(BlockState state, IBlockReader world, BlockPos pos, @Nullable Direction side)`)? A lot of mod rely on this feature to access the data stored in BlockEntity.

https://github.com/RoyalAliceAcademyOfSciences/SimElectricity/blob/34eb82f68ac2b7c19b45922307cc5e6cac60feae/src/main/java/simelectricity/essential/cable/BlockCable.java#L562

Thanks!

KnightMiner:
I would be concerned that making it world sensitive would cause some people to allow redstone to connect but not trigger the proper events to ensure redstone updates. If world/pos access is granted I would add a statement to the javadoc saying the modder is responsible for triggering block updates if the redstone connectivity changes due to something other than the block state properties

That said, I do see little reason not to include redstone connectivity in your blockstate as you probably want to render it in the model as well. The TE can always change the block state as needed, it will persist as long as the block stays the same.

rikka0w0:
> I would be concerned that making it world sensitive would cause some people to allow redstone to connect but not trigger the proper events to ensure redstone updates. If world/pos access is granted I would add a statement to the javadoc saying the modder is responsible for triggering block updates if the redstone connectivity changes due to something other than the block state properties
> 
> That said, I do see little reason not to include redstone connectivity in your blockstate as you probably want to render it in the model as well. The TE can always change the block state as needed, it will persist as long as the block stays the same.

Imagine this scenario, there is a cable block which may have addon panels installed on any sides, there are two catogories of addon panels, one may emit redstone signal, the other does not. Using blockstate, there will be 2^6 combinations for each cable...

Looking at the old `canConnectRedstone`, it provides the world and pos in addition to blockstate and direction. In most case, blockstate should be good enough. But for those really dynamic stuff (like the cable block example I mentioned above), world and blockpos access is still necessary. I would agree that it is the modder's responsiblity to trigger block updates in the latter case.

My proposed solution is presented in #8014.

Da-Technomancer:
I agree with the changes by @rikka0w0 in PR #8014, which is based off this PR. 
As a matter of recordkeeping, should I close this PR to avoid having two competing PRs?

ChampionAsh5357:
To address your question, yes it should if you believe the other implementation is better. However, since that has been merged now, I'll close the PR.

PR #7931, opened 2021-07-25
[1.17.x] Fix typo in FluidBlockWrapper

iTitus:
Port of #7880 to 1.17


LexManos:
Guess this shows how little the fluid wrapper is used... How has nobody seen this before?

PR #7932, opened 2021-07-26
[1.17.x] Add OpenGL version search to use latest available OpenGL version

RogueLogix:
One of the many quirks of OpenGL implementations, they don't all act the same, particularly with context creation. When creating a context, the driver is free to give any context that meets the requirements laid out by the GLFW window hints, this includes Nvidia giving an OpenGL 3.2 core profile and AMD giving an OpenGL 4.6 core profile when both are asked for a GL 3.2 core compliant profile with 4.6 capable hardware. Unfortunately there is no way to query the latest GL version supported by a given implementation, so your a forced to search for it, by attempting to create a context and then try the next lower version if creation failed.

The point of doing this search is to allow mods to utilize the latest OpenGL features available, though they would need to check what is available and adapt accordingly.

### DIFF
--- a/src/main/java/net/minecraftforge/fmllegacy/NoVizFallback.java
+++ b/src/main/java/net/minecraftforge/fmllegacy/NoVizFallback.java
@@ -19,12 +19,124 @@
 
 package net.minecraftforge.fmllegacy;
 
+import org.lwjgl.PointerBuffer;
+import org.lwjgl.glfw.GLFWErrorCallback;
+import org.lwjgl.system.MemoryStack;
+
+import java.util.EnumSet;
 import java.util.function.IntSupplier;
 import java.util.function.LongSupplier;
 import java.util.function.Supplier;
 
+import static org.lwjgl.glfw.GLFW.*;
+import static org.lwjgl.system.MemoryUtil.NULL;
+import static org.lwjgl.system.MemoryUtil.memUTF8;
+
 public final class NoVizFallback {
     public static LongSupplier fallback(IntSupplier width, IntSupplier height, Supplier<String> title, LongSupplier monitor) {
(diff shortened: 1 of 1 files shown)


Jozufozu:
Flywheel uses instanced arrays (GL33) and I've plans to use multidraw indirect (GL43), so having guaranteed access to the newest GL context possible (for the user's system) would be very nice. While those features are also exposed through extensions to OpenGL, having them available through the core profile improves the QoL for us devs.

RogueLogix:
for the purposes of continuity, beginning of convo about this PR on Discord
https://discord.com/channels/313125603924639766/852298000042164244/875968451049570304

src/main/java/net/minecraftforge/fmllegacy/NoVizFallback.java
```diff
@@ -19,12 +19,124 @@
...
+        static Iterable<GLVersion> allowedFromVersionInt(int versionInt) {
+            GLVersion lowerBound = fromInt(versionInt / 100 % 100);
+            GLVersion upperBound = fromInt(versionInt % 100);
+            
+            if (upperBound == ANY) {
+                return EnumSet.range(GL46, GL32);
+            }
+            // only one bound set, we want that specific version
+            if(lowerBound == ANY){
+                EnumSet.of(upperBound);
```
RogueLogix: bug here, that should be returning that enumset

sciwhiz12:
Hello! @RogueLogix, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18. If we do not hear back from you after some time, this PR will be closed for inactivity.

RogueLogix:
From talks with Gigaherz and Lex, this PR wont be accepted without it being disabled by default, which makes it basically useless. Given that, I don't plan on updating it.

PR #7939, opened 2021-07-27
Rename potion registry fields to match the mojang names

HyCraftHD:
Rename potion registry fields to match the mojang names (and the vanilla Registry Keys)


marchermans (approved):

LexManos:
This change is good as it seems that this is the official Mojang names.
However its super dirty as it's making one field change to another of the same name. 
But we're in the early breaking changes phase of 1.17, so it should be fine.
I'm sorry to the guys on discord who will have to deal with the fallout of this...

PR #7941, opened 2021-07-27
[1.17.1] Allow to check spawner block entity in events

strikerrocker:
In 1.16, Spawner's block entity can be obtained via getting the block pos from BaseSpawner/AbstractSpawner but in 1.17 there is no way to get the block pos of the spawner block(SpawnerEntity can be obtained tho). ~~This PR adds BlockPos field to CheckSpawn and SpecialSpawn event.~~ This PR changes the anonymous inner class instance of Base Spawner to inner class and adds the ability to get SpawnerBlockEntity from it.


src/main/java/net/minecraftforge/event/entity/living/LivingSpawnEvent.java
```diff
@@ -86,11 +90,12 @@ public LivingSpawnEvent(Mob entity, LevelAccessor world, double x, double y, dou
          * @param spawner position of the MobSpawner
          *                  null if it this spawn is coming from a WorldSpawner
          */
-        public CheckSpawn(Mob entity, LevelAccessor world, double x, double y, double z, @Nullable BaseSpawner spawner, MobSpawnType spawnReason)
+        public CheckSpawn(Mob entity, LevelAccessor world, double x, double y, double z, @Nullable BaseSpawner spawner, MobSpawnType spawnReason,BlockPos spawnerPos)
```
SuperMartijn642: The `spawnerPos` argument is missing the `Nullable` annotation

src/main/java/net/minecraftforge/event/entity/living/LivingSpawnEvent.java
```diff
@@ -108,6 +113,8 @@ public MobSpawnType getSpawnReason()
         {
             return spawnReason;
         }
+
+        public BlockPos getSpawnerPos(){return spawnerPos;}
```
SuperMartijn642: This is not formatted like Forge's coding style

src/main/java/net/minecraftforge/event/entity/living/LivingSpawnEvent.java
```diff
@@ -128,16 +135,19 @@ public MobSpawnType getSpawnReason()
...
         /**
          * @param spawner the position of a tileentity or approximate position of an entity that initiated the spawn if any
          */
-        public SpecialSpawn(Mob entity, Level world, double x, double y, double z, @Nullable BaseSpawner spawner, MobSpawnType spawnReason)
+        public SpecialSpawn(Mob entity, Level world, double x, double y, double z, @Nullable BaseSpawner spawner, MobSpawnType spawnReason,BlockPos spawnerPos)
         {
             super(entity, world, x, y, z);
             this.spawner = spawner;
             this.spawnReason = spawnReason;
+            this.spawnerPos=spawnerPos;
```
SuperMartijn642: You seem to be missing spaces between things

strikerrocker:
Fixed all the styling issues.

marchermans:
I understand why you would want this. But have you considered making the anonymous inner class, not anonymous and pass in the spawner instance there?

That way you can get away with a two line patch, and don't need to hassle with the blockposes which could be null.
In the end a simple instanceof against the base spawner, cast and for example a field retrieval would do the job, and would likely be much cleaner and easier to maintain.

marchermans (changes requested):
I like this change, but can you do the same thing for the MinecartSpawner that you did for the SpawnerBlockEntity?

strikerrocker:
> 
> 
> I like this change, but can you do the same thing for the MinecartSpawner that you did for the SpawnerBlockEntity?

You can already get the MinecartSpawner entity using BaseSpawner.getSpawnerEntity() so no need to modify MinecartSpawner Spawner instance. Test mod now also now blocks minecart spawner spawns.

marchermans (approved):

PR #7942, opened 2021-07-27
[1.16.5] add possibility to add brewing recipes via datapacks

Minecraftschurli:
Adds the brewing potion type and serializer for creation of brewing recipes with datapacks
Based on #7865 adding the same for the grindstone

### DIFF
--- a/patches/minecraft/net/minecraft/inventory/container/BrewingStandContainer.java.patch
+++ b/patches/minecraft/net/minecraft/inventory/container/BrewingStandContainer.java.patch
@@ -5,7 +5,7 @@
  
        public boolean func_75214_a(ItemStack p_75214_1_) {
 -         return PotionBrewing.func_185205_a(p_75214_1_);
-+         return net.minecraftforge.common.brewing.BrewingRecipeRegistry.isValidIngredient(p_75214_1_);
++         return true;
        }
  
        public int func_75219_a() {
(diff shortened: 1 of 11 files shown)


ChampionAsh5357:
This is a breaking change for binary compatibility with forge and probably won't be accepted. This needs to be redesigned in a non-breaking way if you want 1.16.5 to support this or just push it up to 1.17 since an RB hasn't occurred yet for that.

Minecraftschurli:
Ok I'll port and submit a new PR

KnightMiner:
As another comment on this, if you could move the vanilla logic into a `IBrewingRecipe` that is loaded from the datapack (via the serializer for single instance recipes), that would make it easier for datapacks to remove brewing recipes (pretty sure the vanilla logic is already an `IBrewingRecipe`, just need to change where its loaded). They would of course have to reimplement all the recipes they want to keep in order to remove one, but in my opinion if you are going to put it in JSON, you should give them full control.

That part would definitely be a breaking change, probably best done by replacing the brewing recipe registry with a JSON recipe type.

ferriarnus:
Does this also alow you to change the fuel? I remember some talk about that on the discord but I'm not sure it made it into a PR.

IchHabeHunger54:
The fuel check is literally `itemstack.is(Items.BLAZE_POWDER)` (BrewingStandBlockEntity#serverTick and canPlaceItem). It has nothing to do with the brewing stand recipes themselves.

ferriarnus:
I mean I remeber some talk about not making it only be blaze powder, making it more like furnace fuel. It doesn't seem like it got made into a PR. But if it does, it would clash with this PR. I personaly would like it if not only blaze powder could be used.

Minecraftschurli:
I ported it to 1.17.1 (#7949)

PR #7943, opened 2021-07-28
[1.16.x] Add Custom Mirror Method Analogous to Rotate method to IForgeBlockstare and block

Raycoms:
As in the description. For certain blocks it is interesting to have the option to also do the mirror with the full world context.

### DIFF
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
@@ -47,6 +47,7 @@ import net.minecraft.pathfinding.PathNodeType;
 import net.minecraft.tags.BlockTags;
 import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.Direction;
+import net.minecraft.util.Mirror;
 import net.minecraft.util.ResourceLocation;
 import net.minecraft.util.Rotation;
 import net.minecraft.util.math.BlockPos;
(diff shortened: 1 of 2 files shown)


LexManos:
This needs more justification then 'its interesting'.
It mirroring and rotation is used for schematics and the like which may not have world context.
What uses do you have, and why does it need the world?

Raycoms:
We are having certain tileEntities with position dependent variables (offsets/directions).
And we use Schematic templates with those blocks, as such, we're rotating and mirror them.
Especially in 1.17, the mirror/rotate function in tileEntities will be gone, so this is the next best way to get access to this.

LexManos:
The point is that Mojang is trying to move away from tile entity based systems. We are no longer limited by block ids. So honestly this calls into question why we even have the rotate method still.
You're not providing any concrete examples besides "I have a TE that needs the TE because it's a TE". 
Please provide actual justification.

Raycoms:
A) Our TileEntity, is a piston that can pull from any direction to any direction. We store the direction is is pulling from, and the direction it's pushing to. Depending on rotation and mirror this direction changes.

B) Another use-case are containers we're using that have a "managing container" in a certain offset. When rotated or mirrored, this offset changes.

Using this in the context of schematic like systems require mirror/rotate functions on the TileEntities. Handling this with blockstates is either impossible (offset based containers) or impractical (piston that push/pull from several directions).

marchermans:
Again this is missing the connection with the current callers of the original override method.
Please check if it is possible to redirect the callers of the current system to these new methods by default.

sciwhiz12:
Hello! @Raycoms, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

PR #7944, opened 2021-07-28
[1.17.x] Add Custom Mirror Method Analogous to Rotate method to IForgeBlockstare and block

Raycoms:
As in the description. For certain blocks it is interesting to have the option to also do the mirror with the full world context.

### DIFF
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
@@ -32,8 +32,7 @@ import net.minecraft.world.entity.boss.enderdragon.EnderDragon;
 import net.minecraft.world.entity.boss.wither.WitherBoss;
 import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.entity.projectile.WitherSkull;
-import net.minecraft.world.level.block.LadderBlock;
-import net.minecraft.world.level.block.TrapDoorBlock;
+import net.minecraft.world.level.block.*;
 import net.minecraft.world.level.material.FluidState;
 import net.minecraft.world.item.AxeItem;
 import net.minecraft.world.item.ItemStack;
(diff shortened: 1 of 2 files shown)


marchermans:
Good addition, but I think you might want to check if the call to the current method can be redirected.

Commoble:
I agree that block entities need their own standardized way of rotating their data, but this isn't the right way to do it -- the biggest place where this is needed is when that data is being deserialized into a different rotation than what it was serialized into (e.g. structure pieces), which this doesn't help as there's no world context there.

Raycoms:
I think the onlt alternative I can think of, is patching rotate and mirror methods back into the blockentities for 1.17, and then call it from the vanilla structure handler placement - pos placement.

ChampionAsh5357:
To better elaborate, all places where `#mirror` is used has the necessary contexts as provided by the method parameters. The only place in `#rotate` would be for `StructureTemplate#filterBlocks`. From what I can tell, if you backtrace all the calls, you will eventually get the context needed for the level. However, patching all those is an insane idea and way to invasive for what this should be.

I think this may be more useful if this was split into two separate contexts: one for use after world gen when you have the world context fully present, and one for use during world gen where you can get the necessary structure info as required.

SkullGamer205:
В письме от понедельник, 11 октября 2021 г. 00:29:52 +10 пользователь 
ChampionAsh5357 написал:
> To better elaborate, all places where `#mirror` is used has the necessary
> contexts as provided by the method parameters. The only place in `#rotate`
> would be for `StructureTemplate#filterBlocks`. From what I can tell, if you
> backtrace all the calls, you will eventually get the context needed for the
> level. However, patching all those is an insane idea and way to invasive
> for what this should be.
> 
> I think this may be more useful if this was split into two separate
> contexts: one for use after world gen when you have the world context fully
> present, and one for use during world gen where you can get the necessary
> structure info as required.

sciwhiz12:
Hello! @Raycoms, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable. If we do not hear back from you after some time, this PR will be closed for inactivity.

PR #7945, opened 2021-07-28
[1.17] Lazy capabilities on itemstacks.

marchermans:
1.17 version of #7879.


src/test/java/net/minecraftforge/debug/world/item/LazyCapabilitiesOnItemsTest.java
```diff
@@ -0,0 +1,165 @@
+package net.minecraftforge.debug.world.item;
```
pupnewfster, line 20: License header is missing

src/test/resources/META-INF/mods.toml
```diff
@@ -142,5 +142,6 @@ license="LGPL v2.1"
     modId="overlay_layers_test"
 [[mods]]
     modId="entity_renderer_events_test"
-
```
pupnewfster, line 155: This blank line should probably continue to exist after the lines adding the test mod. Also this file has a merge conflict.

src/test/java/net/minecraftforge/debug/world/item/LazyCapabilitiesOnItemsTest.java
marchermans: What was the way to add those again?
checkLincenses?

sciwhiz12: IIRC, the `licenseFormat` task.

gigaherz: updateLicenses

src/test/resources/META-INF/mods.toml
```diff
@@ -148,5 +150,3 @@ license="LGPL v2.1"
     modId="tag_based_tool_types"
 [[mods]]
     modId="check_spawn_event_test"
-
-# ADD ABOVE THIS LINE
```
pupnewfster: This should not be being removed...

gigaherz: I added that to avoid having every single PR swap this between "line break at end of file" and "no line break at end of file".
It's technically not necessary, just removes that annoyance.

src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
```diff
@@ -22,35 +22,81 @@
...
+    private void doGatherCapabilities(@Nullable ICapabilityProvider parent)
     {
         this.capabilities = ForgeEventFactory.gatherCapabilities(baseClass, this, parent);
+        this.initialized = true;
     }
 
     protected final @Nullable CapabilityDispatcher getCapabilities()
     {
-        return this.capabilities;
+        if(isLazy && !initialized)
```
josephtarbit: Inconsistencies in spacing after the `if`.

src/test/resources/META-INF/mods.toml
pupnewfster: I mean yes it isn't technically necessary but it is somewhat useful so there is no real point in removing it (especially as a side effect of an arbitrary PR)

PR #7946, opened 2021-07-28
Reimplement IForgeBlockEntity#onLoad() hook

### LINKED ISSUE #7926: [1.17.1] BlockEntity#onLoad is never called
**Minecraft Version:** 1.17.1

**Forge Version:** Tested on 37.0.2, confirmed in code in 6ddf56ef158af0282696f33d10ebed452a751375 (current latest 1.17 commit)

**Logs:** Not applicable

**Steps to Reproduce:**
 1. Implement `onLoad` in a block entity
 2. Observe that it is not called when the block entity is loaded

**Description of issue:**
The patch calling `BlockEntity#onLoad` seems to have been missed in the 1.17 update, there are no references to it in Forge besides the declaration in `IForgeBlockEntity`.

XFactHD:
This PR adds back the `IForgeBlockEntity#onLoad()` hook that was removed in the initial port. Fixes #7926

There are four cases where this method is called:

- Server side when placing a block: handled by `LevelChunk#addAndRegisterBlockEntity()`
- Client side when placing a block: handled by `LevelChunk#addAndRegisterBlockEntity()`
- Server side when generating or loading a chunk: handled by `LevelChunk#registerAllBlockEntitiesAfterLevelLoad()`
- Client side when generating or loading a chunk: handled by `LevelChunk#addAndRegisterBlockEntity()`

Due to the way the chunk loading works now, the call to `onLoad()` can't happen in `LevelChunk#registerAllBlockEntitiesAfterLevelLoad()` because at this point the Chunk has not been fully added to the Level yet and accessing the world in `onLoad()` would call into the same `CompletableFuture` that is already running and even circumvent the deadlock avoidance added by Forge in `ChunkMap#protoChunkToFullChunk()`.
To get around this, the server side call to `onLoad()` after chunk generation/loading is deffered to right before the first tick for that BE through the help of a `TickEvent.WorldTickEvent` handler.

### DIFF
--- a/patches/minecraft/net/minecraft/world/level/Level.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/Level.java.patch
@@ -17,13 +17,15 @@
     private final Thread f_46423_;
     private final boolean f_46424_;
     private int f_46425_;
-@@ -99,8 +_,12 @@
+@@ -99,8 +_,14 @@
     private final WorldBorder f_46447_;
     private final BiomeManager f_46420_;
     private final ResourceKey<Level> f_46421_;
 +   public boolean restoringBlockSnapshots = false;
 +   public boolean captureBlockSnapshots = false;
 +   public java.util.ArrayList<net.minecraftforge.common.util.BlockSnapshot> capturedBlockSnapshots = new java.util.ArrayList<>();
++   private final java.util.ArrayList<BlockEntity> freshBlockEntities = new java.util.ArrayList<>();
++   private final java.util.ArrayList<BlockEntity> pendingFreshBlockEntities = new java.util.ArrayList<>();
  
     protected Level(WritableLevelData p_46450_, ResourceKey<Level> p_46451_, final DimensionType p_46452_, Supplier<ProfilerFiller> p_46453_, boolean p_46454_, boolean p_46455_, long p_46456_) {
 +      super(Level.class);
@@ -119,12 +121,30 @@
                 }
              });
              CrashReportCategory.m_178950_(crashreportcategory, this, p_46587_, blockstate);
-@@ -404,8 +_,9 @@
+@@ -401,11 +_,28 @@
+       (this.f_151504_ ? this.f_151503_ : this.f_151512_).add(p_151526_);
+    }
+ 
++   public void addFreshBlockEntities(java.util.Collection<BlockEntity> beList) {
++      if (this.f_151504_) {
++         this.pendingFreshBlockEntities.addAll(beList);
++      } else {
++         this.freshBlockEntities.addAll(beList);
++      }
++   }
++
     protected void m_46463_() {
        ProfilerFiller profilerfiller = this.m_46473_();
        profilerfiller.m_6180_("blockEntities");
--      this.f_151504_ = true;
-+      this.f_151504_ = true;// Forge: Move above remove to prevent CMEs
++      if (!this.pendingFreshBlockEntities.isEmpty()) {
++         this.freshBlockEntities.addAll(this.pendingFreshBlockEntities);
++         this.pendingFreshBlockEntities.clear();
++      }
+       this.f_151504_ = true;
++      if (!this.freshBlockEntities.isEmpty()) {
++         this.freshBlockEntities.forEach(BlockEntity::onLoad);
++         this.freshBlockEntities.clear();
++      }
        if (!this.f_151503_.isEmpty()) {
 +         this.f_46434_.forEach(e -> e.onChunkUnloaded());
           this.f_151512_.addAll(this.f_151503_);
--- a/src/test/java/net/minecraftforge/debug/block/BlockEntityOnLoadTest.java
+++ b/src/test/java/net/minecraftforge/debug/block/BlockEntityOnLoadTest.java
@@ -1,0 +1,109 @@
+package net.minecraftforge.debug.block;
+
+import net.minecraft.core.BlockPos;
+import net.minecraft.world.InteractionHand;
+import net.minecraft.world.InteractionResult;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.*;
+import net.minecraft.world.level.Level;
+import net.minecraft.world.level.block.*;
+import net.minecraft.world.level.block.entity.*;
+import net.minecraft.world.level.block.state.BlockBehaviour.Properties;
+import net.minecraft.world.level.block.state.BlockState;
+import net.minecraft.world.level.material.Material;
+import net.minecraft.world.level.material.MaterialColor;
+import net.minecraft.world.phys.BlockHitResult;
+import net.minecraftforge.eventbus.api.IEventBus;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.fmllegacy.RegistryObject;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import javax.annotation.Nullable;
+
+@Mod("be_onload_test")
+public class BlockEntityOnLoadTest
+{
+    private static final boolean ENABLED = true;
+
+    private static final DeferredRegister<Block> BLOCKS = DeferredRegister.create(ForgeRegistries.BLOCKS, "be_onload_test");
+    private static final DeferredRegister<Item> ITEMS = DeferredRegister.create(ForgeRegistries.ITEMS, "be_onload_test");
+    private static final DeferredRegister<BlockEntityType<?>> BE_TYPES = DeferredRegister.create(ForgeRegistries.BLOCK_ENTITIES, "be_onload_test");
+
+    private static final RegistryObject<Block> TEST_BLOCK = BLOCKS.register("be_onload_testblock", () -> new TestBlock(Properties.of(Material.DECORATION, MaterialColor.SAND)));
+    private static final RegistryObject<Item> TEST_BLOCK_ITEM = ITEMS.register("be_onload_testblock", () -> new BlockItem(TEST_BLOCK.get(), new Item.Properties().tab(CreativeModeTab.TAB_DECORATIONS)));
+    private static final RegistryObject<BlockEntityType<TestBlockEntity>> TEST_BE_TYPE = BE_TYPES.register("be_onload_testbe", () -> BlockEntityType.Builder.of(TestBlockEntity::new, TEST_BLOCK.get()).build(null));
+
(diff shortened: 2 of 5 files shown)


src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1398,4 +1399,30 @@ public static void modifyAttributes()
             FORGE_ATTRIBUTES.put(k, newBuilder.build());
         });
     }
+
+    private static final Set<BlockEntity> LOADING_BLOCK_ENTITIES = Sets.newConcurrentHashSet();
```
TBoshoven: I don't think maintaining a global static set is an acceptable way of doing this.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1398,4 +1399,30 @@ public static void modifyAttributes()
...
+
+    @Deprecated
+    public static void enqueueLoadingBlockEntity(BlockEntity be)
+    {
+        LOADING_BLOCK_ENTITIES.add(be);
+    }
+
+    @Deprecated
+    public static void onLevelTick(Level level)
```
TBoshoven: Waiting all the way for the first tick just to make sure you can access the chunk from its level will probably work, but it seems a bit drastic.

Don't we just need to wait for the future to resolve?
Then we don't even need to keep track of the "loading" block entities. We can just iterate over the ones in the chunk.

TBoshoven (commented):
I've been waiting for this to be fixed so my model data loads properly.
However, I think there's a cleaner way of doing this than proposed here.

I'd experiment a bit on my end, but unfortunately I'm running into [this weird issue](https://forums.minecraftforge.net/topic/102976-117x-crash-when-trying-to-run-forge-forge-client/) so I can't at the moment, at least until I figure that one out.

src/main/java/net/minecraftforge/common/ForgeHooks.java
XFactHD: This really depends on wether we want to allow full access to the chunk in `IForgeBlockEntity#onLoad()`. In 1.16 for example this is not possible.
If we want to go the same route as 1.16, then the necessary change is incredibly simple and no defferal to the first tick is necessary.
If full access to the chunk is desired then there is (as far as I can see) no way around this whole thing. In theory I can guard against the deadlock created by accessing the chunk while it is loading. In practice this only works as long as the only chunk that is accessed by whatever you do in `onLoad()` is the one that is loading. As soon as you access another chunk (either by your `BlockEntity` being placed on a chunk border and doing something that causes a block update (like the test mod) or your code explicitly reaching into other chunks) you get another deadlock that can't be guarded against because that other chunk doesn't start loading until the current one is done.

Simply waiting for the future to resolve is also not an option because Mojang are not only using the futures for loading the chunk but also for retrieving a loaded chunk when it is not contained in their "cache array" holding the 4 last used chunks.

TBoshoven: > This really depends on wether we want to allow full access to the chunk in IForgeBlockEntity#onLoad(). In 1.16 for example this is not possible.

I think there is a discussion to be had on what the intention here is because clearly this is impacting simplicity somewhat. If it never worked like that in 1.16, I don't know if there are good reasons to change that. Not saying there aren't, but I'm not aware of what people typically do in their `onLoad` calls. All I'm personally looking for is for my model data to be loaded, which doesn't require the level. In fact, I don't even override the `onLoad` call at all.

---

I was typing up a whole story about a good place to add a hook when I decided to check about existing hooks. Would the `ChunkEvent.Load` event (or a hook in the same place) not suffice? That one gets called during promotion on the `ChunkMap` itself, so the mentioned cache doesn't play a role.

50ap5ud5: If you need example of how other mods use onLoad, my use of onLoad function includes deferring the spawning of a fresh entity around my tile entity, through decreasing a counter in onLoad to allow some entities in my tick method to be spawned once that counter reaches 0.

I also send packets to update data for models in the onLoad method for another tile.

Not sure if this method is how we are meant to use it, but that is how i currently do so.

XFactHD: > I was typing up a whole story about a good place to add a hook when I decided to check about existing hooks. Would the `ChunkEvent.Load` event (or a hook in the same place) not suffice? That one gets called during promotion on the `ChunkMap` itself, so the mentioned cache doesn't play a role.

This event is called in the same place as the Chunk method that calls my current "enqueue" hook. This still happens in the running future which means accessing the Chunk through the Level in the BlockEntity will cause a deadlock.
The main problem here is that the futures are blocking the main thread and the one spot where that blocking is done has no idea if the requested chunk is has to be loaded/generated or not. Writing this gave me an idea though...

> All I'm personally looking for is for my model data to be loaded, which doesn't require the level.

The ModelData request in onLoad() requires a Level but not a Chunk. Accessing the Level in onLoad() is completely fine in 1.16 as long as you don't do anything that accesses Chunks

TBoshoven: My impression was that at this point, the promotion has been done and actual future resolves to the chunk, but I am probably missing something. It's not easy to navigate this code.

> Writing this gave me an idea though...

I'd love to hear/see it.

XFactHD: > My impression was that at this point, the promotion has been done and actual future resolves to the chunk, but I am probably missing something. It's not easy to navigate this code.

The problem is that anything that accesses a chunk needs to happen outside of the future because accessing a chunk creates a future that then is enqueued in the `BlockableEventLoop` because that is passed as the `Executor` to the future. Because that new future is enqueued after the running future, it will not complete before the one currently running and therefore block the running future.

> I'd love to hear/see it.

```
CompletableFuture<Either<ChunkAccess, ChunkHolder.ChunkLoadingFailure>> completablefuture = this.getChunkFutureMainThread(p_8360_, p_8361_, p_8362_, p_8363_);
boolean loaded = completablefuture.isDone(); //Check if the chunk was already loaded
this.mainThreadProcessor.managedBlock(completablefuture::isDone);
ChunkAccess chunkaccess1 = completablefuture.join().map((p_8406_) -> {...});
this.storeInCache(i, chunkaccess1, p_8362_);
if (!loaded) {
    //Iterate BEs to call onLoad()
}
return chunkaccess1;
```
This is the piece of code that I mentioned that blocks the main thread. The `CompletableFuture` gets the `mainThreadProcessor` (which is a `BlockableEventLoop`) as its `Executor`. The CF calls `execute()` on that which leads to enqueueing the CFs `Runnable` in the event loops queue. The call to `managedBlock()` then works through queued tasks until the `Predicate` it gets passed (`completablefuture::isDone` in this case) returns true. One of the queued tasks will be the chunk load task. Because of that queueing I am able to check if the CF is already done (which would indicate that this Chunk was already loaded earlier). This works perfectly fine in theory but in testing I found that there is another path where chunk loading can happen (which probably has to do with either the spawn chunks or the chunks the player is in, I have not found the specific code path yet) that completely bypasses the method this code block is in and therefore makes this approach useless.

rikka0w0:
Currently I'm using `clearRemoved` and `setRemoved` as the replacement of `onLoad` and `onUnload`, respectively, to perform some server-only logic (register the blockEntity to a graph, no block access). What are possible negative impacts? Thanks!

src/main/java/net/minecraftforge/common/ForgeHooks.java
XFactHD: Moved to `Level`

patches/minecraft/net/minecraft/world/level/Level.java.patch
```diff
@@ -17,13 +17,14 @@
...
     private int f_46425_;
-@@ -99,8 +_,12 @@
+@@ -99,8 +_,13 @@
     private final WorldBorder f_46447_;
     private final BiomeManager f_46420_;
     private final ResourceKey<Level> f_46421_;
 +   public boolean restoringBlockSnapshots = false;
 +   public boolean captureBlockSnapshots = false;
 +   public java.util.ArrayList<net.minecraftforge.common.util.BlockSnapshot> capturedBlockSnapshots = new java.util.ArrayList<>();
++   public java.util.Set<BlockEntity> freshBlockEntities = com.google.common.collect.Sets.newHashSet();
```
RogueLogix: Is a HashSet the right collection type here? It doesn't seem like the best choice based on how its being used, as it being a set vs a list doesn't seem to be necessary.

XFactHD: The Set was just the first thing I thought off. I could make it an ArrayList instead if that is better (my knowledge about data structures is severely lacking to be completely honest).

RogueLogix: When I implemented a very similar system for my mod, I settled on an ArrayList for the iteration speed. In this use case, an ArrayList would probably use less memory and be faster. With an ArrayList an ensureCapacity call could also be made just ahead of adding everything to avoid reallocs in the add loop, though that would probably have less of an impact than just using an ArrayList in the first place.

XFactHD: Changed it to an ArrayList and included your suggestion to call `ensureCapacity()` before adding the `BlockEntities` to it

rehwinkel (approved):
I'm no expert on this, but it looks alright to me :+1:

src/test/java/net/minecraftforge/debug/block/BlockEntityOnLoadTest.java
```diff
@@ -0,0 +1,109 @@
...
+                first = false;
+                LOGGER.info("[BE_ONLOAD] TestBlockEntity#tick at pos {} for {}", worldPosition, this);
+                if (!loaded)
+                {
+                    throw new IllegalStateException(String.format("BlockEntity at %s ticked before onLoad()!", getBlockPos()));
+                }
+            }
+        }
+    }
+}
```
TBoshoven: Missing newline

src/test/resources/META-INF/mods.toml
```diff
@@ -148,5 +148,7 @@ license="LGPL v2.1"
     modId="tag_based_tool_types"
 [[mods]]
     modId="check_spawn_event_test"
+[[mods]]
+	modId="be_onload_test"
 
-# ADD ABOVE THIS LINE
+# ADD ABOVE THIS LINE
```
TBoshoven: Looks like you accidentally removed the newline here

TBoshoven (approved):
I like the changes you made.
Things are a lot more compact now at least, and I don't expect this will break easily.

Unfortunately I'm still suffering from this weird segfault when attempting to run a Forge client, but I ended up testing this the long way around (building an installer, installing it, and running it with my mod). It looks like things are working exactly as expected :+1:

PR #7949, opened 2021-07-28
[1.17.1] add possibility to add brewing recipes via datapacks

Minecraftschurli:
adds the possibility to add brewing recipes via datapacks and moves vanilla brewing recipes to json files in the minecraft namespace

#### Goals:
- make it possible to add brewing recipes via datapacks
- make it possible to "disable" vanilla brewing recipes
- not break non vanilla brewing stands possibility via BrewingRecipeRegistry#canBrew/brewPotions

#### Tested:
- [x] Vanilla Client + Forge Server
- [x] Forge Client + Vanilla Server
- [x] Forge Client + Forge Server

#### Test mod includes:
- add brewing: recipe bowl + brown mushroom => mushroom stew
- "disable" brewing of lingering potions

##### for discussion ping me on discord `@Minecraftschurli#0150`


patches/minecraft/net/minecraft/world/inventory/BrewingStandMenu.java.patch
```diff
@@ -5,7 +5,7 @@
  
        public boolean m_5857_(ItemStack p_39121_) {
 -         return PotionBrewing.m_43506_(p_39121_);
-+         return net.minecraftforge.common.brewing.BrewingRecipeRegistry.isValidIngredient(p_39121_);
++         return true;
```
ChampionAsh5357: Again, I'm pretty sure this should not just be true always and check the valid potion ingredients. But as for the other datagen pr, it could be argued either way.

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -294,11 +299,16 @@ public void registerRecipeSerialziers(RegistryEvent.Register<RecipeSerializer<?>
         CraftingHelper.register(TrueCondition.Serializer.INSTANCE);
         CraftingHelper.register(TagEmptyCondition.Serializer.INSTANCE);
 
+        Registry.register(Registry.RECIPE_TYPE, new ResourceLocation("forge", "brewing"), BrewingRecipeRegistry.BREWING);
```
ChampionAsh5357: Keep the recipe type in this class, not somewhere else.

src/main/java/net/minecraftforge/common/brewing/BrewingRecipe.java
```diff
@@ -19,23 +19,26 @@
...
-    @Nonnull private final ItemStack output;
-
-    public BrewingRecipe(Ingredient input, Ingredient ingredient, ItemStack output)
-    {
-        this.input = input;
-        this.ingredient = ingredient;
-        this.output = output;
-    }
+    @ObjectHolder("forge:brewing")
+    public static final RecipeSerializer<BrewingRecipe> SERIALIZER = null;
```
ChampionAsh5357: Implement using `RegistryObject` in `ForgeMod`, not here.

src/main/java/net/minecraftforge/common/brewing/IBrewingRecipe.java
```diff
@@ -42,4 +48,31 @@
...
+        if (container instanceof BrewingStandBlockEntity)
+            return (isInput(container.getItem(0)) || isInput(container.getItem(1)) || isInput(container.getItem(2))) && isIngredient(container.getItem(3));
+        else
+            return isInput(container.getItem(0)) && isInput(container.getItem(1));
+    }
+
+    @Override
+    default ItemStack assemble(Container container)
+    {
+        return getOutput(container.getItem(0), container.getItem(1));
```
ChampionAsh5357: If the brewing stand has a special matching scenario, why doesn't assemble? Also, what is the point of the deferred method call?

src/main/java/net/minecraftforge/common/brewing/IBrewingRecipe.java
```diff
@@ -19,9 +19,15 @@
...
 
+import net.minecraft.world.Container;
 import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.crafting.Recipe;
+import net.minecraft.world.item.crafting.RecipeType;
+import net.minecraft.world.level.Level;
+import net.minecraft.world.level.block.entity.BrewingStandBlockEntity;
 
-public interface IBrewingRecipe {
+public interface IBrewingRecipe extends Recipe<Container>
```
ChampionAsh5357: Shouldn't just base this off of the existing brewing recipe. Should be implemented using the recipe system.

src/main/java/net/minecraftforge/common/brewing/ContainerBrewingRecipe.java
```diff
@@ -0,0 +1,112 @@
...
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.ForgeRegistryEntry;
+import net.minecraftforge.registries.ObjectHolder;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public record ContainerBrewingRecipe(@Nonnull ResourceLocation id,
+                                     @Nonnull Item input,
+                                     @Nonnull Ingredient ingredient,
```
ChampionAsh5357: What is `ingredient`?

src/main/java/net/minecraftforge/common/brewing/ContainerBrewingRecipe.java
```diff
@@ -0,0 +1,112 @@
...
+                                     @Nonnull Item output) implements IBrewingRecipe
+{
+    @ObjectHolder("forge:container_brewing")
+    public static final RecipeSerializer<ContainerBrewingRecipe> SERIALIZER = null;
```
ChampionAsh5357: Move to `ForgeMod` using `RegistryObject`.

src/main/java/net/minecraftforge/common/brewing/BrewingRecipeRegistry.java
```diff
@@ -19,60 +19,29 @@
...
+import net.minecraft.world.level.Level;
+
+import javax.annotation.Nullable;
 import java.util.Collections;
 import java.util.List;
-import net.minecraft.world.item.ItemStack;
-import net.minecraft.world.item.crafting.Ingredient;
-import net.minecraft.core.NonNullList;
 
 public class BrewingRecipeRegistry {
```
ChampionAsh5357: Does this class need to exist at all? It can just be grabbed through the recipe manager.

src/main/java/net/minecraftforge/common/brewing/BrewingRecipe.java
```diff
@@ -19,23 +19,26 @@
...
+import net.minecraftforge.registries.ForgeRegistryEntry;
+import net.minecraftforge.registries.ObjectHolder;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
-public class BrewingRecipe implements IBrewingRecipe
+public record BrewingRecipe(@Nonnull ResourceLocation id,
+                            @Nonnull Ingredient input,
+                            @Nonnull Ingredient ingredient,
```
ChampionAsh5357: What is `ingredient`? It's very nonspecific for what it's supposed to do.

src/main/java/net/minecraftforge/common/brewing/MixingBrewingRecipe.java
```diff
@@ -0,0 +1,115 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public record MixingBrewingRecipe(@Nonnull ResourceLocation id,
+                                  @Nonnull Potion input,
+                                  @Nonnull Ingredient ingredient,
+                                  @Nonnull Potion output) implements IBrewingRecipe
+{
+    @ObjectHolder("forge:mixing_brewing")
+    public static final RecipeSerializer<MixingBrewingRecipe> SERIALIZER = null;
```
ChampionAsh5357: Move to `ForgeMod` as `RegistryObject`.

src/main/java/net/minecraftforge/common/brewing/MixingBrewingRecipe.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @Override
+    public boolean isInput(@Nonnull ItemStack stack)
+    {
+        final Item item = stack.getItem();
+        return (item == Items.POTION || item == Items.SPLASH_POTION || item == Items.LINGERING_POTION || item == Items.GLASS_BOTTLE) && PotionUtils.getPotion(stack) == input();
```
ChampionAsh5357: Hardcoded entries makes it impossible to add new potion contains..

src/main/java/net/minecraftforge/common/brewing/MixingBrewingRecipe.java
```diff
@@ -0,0 +1,115 @@
...
+
+    public static class Serializer extends ForgeRegistryEntry<RecipeSerializer<?>> implements RecipeSerializer<MixingBrewingRecipe>
+    {
+
+        @Override
+        public MixingBrewingRecipe fromJson(final ResourceLocation p_199425_1_, final JsonObject p_199425_2_)
+        {
+            Potion input = ForgeRegistries.POTION_TYPES.getValue(new ResourceLocation(p_199425_2_.get("input_potion").getAsString()));
+            Ingredient ingredient = Ingredient.fromJson(p_199425_2_.getAsJsonObject("ingredient"));
+            Potion output = ForgeRegistries.POTION_TYPES.getValue(new ResourceLocation(p_199425_2_.get("output_potion").getAsString()));
```
ChampionAsh5357: Unchecked NPE.

src/main/java/net/minecraftforge/common/brewing/MixingBrewingRecipe.java
```diff
@@ -0,0 +1,115 @@
...
+            return new MixingBrewingRecipe(p_199425_1_, input, ingredient, output);
+        }
+
+        @Nullable
+        @Override
+        public MixingBrewingRecipe fromNetwork(final ResourceLocation p_199426_1_, final FriendlyByteBuf p_199426_2_)
+        {
+            Potion input = ForgeRegistries.POTION_TYPES.getValue(p_199426_2_.readResourceLocation());
```
ChampionAsh5357: More unchecked NPEs.

src/main/java/net/minecraftforge/common/brewing/ContainerBrewingRecipe.java
```diff
@@ -0,0 +1,112 @@
...
+        return SERIALIZER;
+    }
+
+    public static class Serializer extends ForgeRegistryEntry<RecipeSerializer<?>> implements RecipeSerializer<ContainerBrewingRecipe>
+    {
+
+        @Override
+        public ContainerBrewingRecipe fromJson(final ResourceLocation p_199425_1_, final JsonObject p_199425_2_)
+        {
+            Item input = ForgeRegistries.ITEMS.getValue(new ResourceLocation(p_199425_2_.get("input_container").getAsString()));
```
ChampionAsh5357: Declaring the parameter in the constructor as non null does really solve all NPE and Json exceptions here. They should be caught properly and use the `GsonHelper`.

src/main/java/net/minecraftforge/common/data/BrewingRecipeBuilder.java
```diff
@@ -0,0 +1,313 @@
...
+import net.minecraftforge.registries.ForgeRegistries;
+
+import javax.annotation.Nullable;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+/**
+ * Class for building a brewing recipe. Takes an ingredients as input, an ingredient as ingredient and an itemstack as output.
+ */
+public abstract class BrewingRecipeBuilder
```
ChampionAsh5357, line 42: Needs more documentation.

src/test/java/net/minecraftforge/debug/PotionEventTest.java
```diff
@@ -41,12 +41,6 @@
 
     public PotionEventTest()
     {
-    	FMLJavaModLoadingContext.get().getModEventBus().addListener(this::onSetup);
-    }
-    
-    private void onSetup(FMLCommonSetupEvent event)
-    {
-    	BrewingRecipeRegistry.addRecipe(Ingredient.of(Items.ICE), Ingredient.of(Items.LAVA_BUCKET), new ItemStack(Items.OBSIDIAN));
```
ChampionAsh5357, line 49: I'm pretty sure this breaks the test mod.

src/test/resources/data/minecraft/recipes/brewing/splash_potion_dragon_breath_lingering_potion.json
```diff
@@ -0,0 +1,2 @@
+{
```
ChampionAsh5357, line 1: Should there be something here?

ChampionAsh5357 (commented):
Questions:
- Why for the brewing type parameters is the input and output suffixed with their type (e.g. `input_container`), it's unnecessary and could just be called `input` and `output`, or more appropriate, `base`, `reagent`, and `result`.
- What happens in the situation where I have three different potions that can apply the same reagent? Based on what I see, it'll just only consider the first input.
- The classes itself should be documented more.

src/test/java/net/minecraftforge/debug/PotionEventTest.java
Minecraftschurli: Well the method BrewingRecipeRegistry.addRecipe no longer exists so the call needs to be removed

src/main/java/net/minecraftforge/common/brewing/IBrewingRecipe.java
Minecraftschurli: the method needs to be implemented but is actually never used so it can work in non vanilla brewing stands too

src/main/java/net/minecraftforge/common/brewing/IBrewingRecipe.java
Minecraftschurli: that's not really possible because my goal is to still support non vanilla brewing stands (via BrewingRecipeRegistry) which is not possible with only the Container as a parameter

src/main/java/net/minecraftforge/common/brewing/MixingBrewingRecipe.java
Minecraftschurli: MixingBrewingRecipe is intended to handle just the vanilla recipes
besides I don't know how else to check it because an instanceof check would be too broad (maybe a tag?)

src/main/java/net/minecraftforge/common/brewing/MixingBrewingRecipe.java
Minecraftschurli: what should I throw in the network methods?

patches/minecraft/net/minecraft/world/inventory/BrewingStandMenu.java.patch
Minecraftschurli: to check it I would need the level but this is a Slot which has no itea of the level

src/main/java/net/minecraftforge/common/brewing/BrewingRecipeRegistry.java
Minecraftschurli: No because one of my goals is to still be able to use non vanilla brewing stands which is only possible with the canBrew and brewPotions methods

Minecraftschurli:
> What happens in the situation where I have three different potions that can apply the same reagent? Based on what I see, it'll just only consider the first input.

It works for parallel brewing because it checks each slot and brews each slot by itself (see BrewingRecipeRegistry#canBrew)

patches/minecraft/net/minecraft/world/inventory/BrewingStandMenu.java.patch
ChampionAsh5357: The slot is part of the menu which can have a context of the level, so this is untrue.

src/main/java/net/minecraftforge/common/brewing/BrewingRecipeRegistry.java
ChampionAsh5357: Then this wouldn't be registry but a helper, and should be documented as such. This class isn't needed whatsoever since you can recreate the method yourself via the `RecipeManager`, but it's just nice to have around.

src/main/java/net/minecraftforge/common/brewing/IBrewingRecipe.java
ChampionAsh5357: Sure it is? A container can be created based off of any inventory. The container is only used to check the data, which could be the user's inventory, a block's inventory, etc. These can be wrapped.

src/main/java/net/minecraftforge/common/brewing/IBrewingRecipe.java
```diff
@@ -19,27 +19,58 @@
...
-     * ingredient. Empty if invalid input or ingredient.
+     * Returns the result when the passed base is brewed with the passed
+     * reagent. Empty if invalid base or reagent.
      */
     ItemStack getOutput(ItemStack input, ItemStack ingredient);
+
+    @Override
+    default boolean matches(Container container, Level level)
+    {
+        throw new NotImplementedException("Use the methods of IBrewingRecipe instead");
```
ChampionAsh5357: Problematic code implementation, should not use `RecipeManager` if you're gonna go the half separated route. Since it is a data driven implementation though, you need to implement these properly.

src/main/java/net/minecraftforge/common/brewing/MixingBrewingRecipe.java
```diff
@@ -0,0 +1,116 @@
...
+public record MixingBrewingRecipe(@Nonnull ResourceLocation id,
+                                  @Nonnull Potion base,
+                                  @Nonnull Ingredient reagent,
+                                  @Nonnull Potion result) implements IBrewingRecipe
+{
+    @Override
+    public boolean isInput(@Nonnull ItemStack stack)
+    {
+        final Item item = stack.getItem();
+        return (item == Items.POTION || item == Items.SPLASH_POTION || item == Items.LINGERING_POTION || item == Items.GLASS_BOTTLE) && PotionUtils.getPotion(stack) == base();
```
ChampionAsh5357: Again, aren't these containers? So they should be obtained from valid container inputs.

src/main/java/net/minecraftforge/common/brewing/MixingBrewingRecipe.java
ChampionAsh5357: Disregard this, I forgot that forge does infact default to `Potion#EMPTY` if the value is nonexistent which is vanilla behavior.

src/main/java/net/minecraftforge/common/data/BrewingRecipeBuilder.java
```diff
@@ -0,0 +1,407 @@
...
+                             ItemStack result,
+                             Advancement.Builder advancement,
+                             ResourceLocation advancementId) implements FinishedRecipe
+        {
+
+            @Override
+            public void serializeRecipeData(JsonObject tag)
+            {
+                final ResourceLocation result = ForgeRegistries.ITEMS.getKey(this.result.getItem());
+                if (result == null)
```
ChampionAsh5357: Validation should not be happening this late.  It should be given before the `IFinishedRecipe` is constructed.

src/main/java/net/minecraftforge/common/data/ForgeRecipeProvider.java
```diff
@@ -99,6 +96,100 @@ protected void buildCraftingRecipes(Consumer<FinishedRecipe> consumer)
...
+        addMixingRecipe(consumer, Potions.STRENGTH, Tags.Items.DUSTS_REDSTONE, Potions.LONG_STRENGTH);
+        addMixingRecipe(consumer, Potions.STRENGTH, Tags.Items.DUSTS_GLOWSTONE, Potions.STRONG_STRENGTH);
+        addMixingRecipe(consumer, Potions.WATER, Items.FERMENTED_SPIDER_EYE, Potions.WEAKNESS);
+        addMixingRecipe(consumer, Potions.WEAKNESS, Tags.Items.DUSTS_REDSTONE, Potions.LONG_WEAKNESS);
+        addMixingRecipe(consumer, Potions.AWKWARD, Items.PHANTOM_MEMBRANE, Potions.SLOW_FALLING);
+        addMixingRecipe(consumer, Potions.SLOW_FALLING, Tags.Items.DUSTS_REDSTONE, Potions.LONG_SLOW_FALLING);
+
+    }
+
+    private void addMixingRecipe(final Consumer<FinishedRecipe> consumer, final Potion input, final Item ingredient, final Potion output)
```
ChampionAsh5357, line 164: Could make this protected for extension use.

src/main/java/net/minecraftforge/common/data/ForgeRecipeProvider.java
```diff
@@ -99,6 +96,100 @@ protected void buildCraftingRecipes(Consumer<FinishedRecipe> consumer)
...
+    {
+        String name = "brewing/" + input.getRegistryName().getPath() + "_" +
+                ingredient.getRegistryName().getPath() + "_" +
+                output.getRegistryName().getPath();
+        BrewingRecipeBuilder.mixing(input, ingredient, output)
+                .unlocks("hasBrewingStand", InventoryChangeTrigger.TriggerInstance.hasItems(Blocks.BREWING_STAND))
+                .save(consumer, new ResourceLocation(output.getRegistryName().getNamespace(), name));
```
ChampionAsh5357: Nitpick: could generalize code fragment with other methods.

ChampionAsh5357 (commented):
Looks better, but the main issue is that if you're going to neglect the `RecipeManager` system, then this should be a completely separate reload listener instead of forcing it to integrate. Everything described by you can be done just by knowing the container, so there should be no reason not to implement like vanilla.

patches/minecraft/net/minecraft/world/inventory/BrewingStandMenu.java.patch
Minecraftschurli: the slot has no context of level in the mayPlace method

src/main/java/net/minecraftforge/common/brewing/IBrewingRecipe.java
Minecraftschurli: actually a container is what was the inventory before the change to mojmaps

src/main/java/net/minecraftforge/common/data/ForgeRecipeProvider.java
Minecraftschurli: why ForgeRecipeProvider is final

ChampionAsh5357: In that case then, this should probably be moved to some helper interface then for users to extend similar to `IConditionBuilder`. Makes relatively no sense to make these methods unobtainable to an average modder since they will be creating these jsons too.

src/main/java/net/minecraftforge/common/brewing/ContainerBrewingRecipe.java
```diff
@@ -0,0 +1,122 @@
...
+import net.minecraft.world.level.Level;
+import net.minecraftforge.common.ForgeMod;
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.ForgeRegistryEntry;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.function.Predicate;
+
+public record ContainerBrewingRecipe(@Nonnull ResourceLocation id,
```
KnightMiner, line 42: What is the reason behind making this a record? Doesn't that prevent extension? Extending vanilla recipes is a quite common practice in previous versions, would rather not see it more difficult than it needs to be.

Minecraftschurli: well as the implementations themselves are so simple you could just implement IBrewingRecipe yourself (and I couldn't see a usecase where you would need to extend one of the implementations)

src/main/java/net/minecraftforge/common/brewing/IBrewingRecipe.java
```diff
@@ -20,26 +20,27 @@
...
+    {
+        return ForgeMod.BREWING;
+    }
 
-    /**
-     * Returns the output when the passed input is brewed with the passed
-     * ingredient. Empty if invalid input or ingredient.
-     */
-    ItemStack getOutput(ItemStack input, ItemStack ingredient);
+    Predicate<ItemStack> getReagent();
```
KnightMiner: A method returning a predicate feels awkward to me, wouldn't it be cleaner to just have an `is` method as before? Alternatively, do these still need to exist if `matches` does all the checking logic anyways? Would in theory let you have a recipe that matches depending on both the NBT of the base and a reagent if the two are separated.

If these are kept, a default implementation of `matches` would make sense as both your implementations have the same logic. I could also see these being kept in an abstract class but not in the base interface.

src/main/java/net/minecraftforge/common/brewing/ContainerBrewingRecipe.java
```diff
@@ -0,0 +1,122 @@
...
+        return Ingredient.of(base());
+    }
+
+    public static class Serializer extends ForgeRegistryEntry<RecipeSerializer<?>> implements RecipeSerializer<ContainerBrewingRecipe>
+    {
+
+        @Override
+        public ContainerBrewingRecipe fromJson(final ResourceLocation p_199425_1_, final JsonObject p_199425_2_)
+        {
+            Item input = ForgeRegistries.ITEMS.getValue(new ResourceLocation(GsonHelper.getAsString(p_199425_2_, "base")));
```
ChampionAsh5357: Could use `getValueOptional` and then `getOrThrow`. Seems odd to have the checks after everything has been deserialized.

src/main/java/net/minecraftforge/common/brewing/IBrewingRecipe.java
```diff
@@ -38,7 +40,7 @@ default boolean canCraftInDimensions(int p_43999_, int p_44000_)
         return ForgeMod.BREWING;
     }
 
-    Ingredient getReagent();
+    Predicate<ItemStack> getReagent();
```
ChampionAsh5357: Nitpick: doesn't need a predicate, they can be methods themselves to be called. Seems weird to pass a predicate somewhere if it's never stored.

src/main/java/net/minecraftforge/common/brewing/MixingBrewingRecipe.java
```diff
@@ -0,0 +1,122 @@
...
+        return is -> PotionUtils.getPotion(is) == base() && is.is(Tags.Items.POTION_CONTAINERS);
+    }
+
+    public static class Serializer extends ForgeRegistryEntry<RecipeSerializer<?>> implements RecipeSerializer<MixingBrewingRecipe>
+    {
+
+        @Override
+        public MixingBrewingRecipe fromJson(final ResourceLocation p_199425_1_, final JsonObject p_199425_2_)
+        {
+            Potion input = ForgeRegistries.POTION_TYPES.getValue(new ResourceLocation(p_199425_2_.get("base").getAsString()));
```
ChampionAsh5357: I would use `getValueOptional` and then `getOrThrow`.

src/main/java/net/minecraftforge/common/brewing/BrewingContainerWrapper.java
```diff
@@ -0,0 +1,85 @@
...
+package net.minecraftforge.common.brewing;
+
+import net.minecraft.world.Container;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
+
+/**
+ * Wraps a container representing a brewing stand
+ */
+public record BrewingContainerWrapper(ItemStack base, ItemStack reagent) implements Container
```
KnightMiner: In my mods, I tend to implement container wrappers as wrapping my inventory capability instance, as then I don't have to create a new instance for every recipe call. Making this a record means a lot more objects are generated.

In my opinion, using an interface in the recipe is cleaner, one that extends container, defaults unused methods, and includes the base and reagent. You could still have a record implement the interface for the forge implementation like this, or create an implementation that is specifically wrapping the brewing stand tile entity.

----

If this method is preferred (instance per recipe call), I think the java doc/name is misleading, as this does not wrap a container, it wraps two item stacks. Its really a `BrewingContainerRecord` to reuse the java term, a record acting as a container.

src/main/java/net/minecraftforge/common/brewing/IBrewingRecipe.java
Minecraftschurli: well
1. it is a Predicate because there is no Ingredient that checks for a potion
2. they are needed for checking the validity of a reagent/base
3. yes the default implementation wold work
4. no it has to be in the base interface because it is checked for when checking item validity for reagent/base

src/test/java/net/minecraftforge/debug/PotionEventTest.java
ChampionAsh5357: Then the test case itself should be added as a json.

src/main/java/net/minecraftforge/common/brewing/BrewingContainerWrapper.java
Minecraftschurli: 1. it has to be like that to be able to brew different recipes with the same reagent at the same time
2. I forgot to change the javadoc when I changed the class

src/main/java/net/minecraftforge/common/brewing/IBrewingRecipe.java
KnightMiner: You did not answer my first question. What benefit is a predicate over `isBase(ItemStack)`? You already need an instance of `IBrewingRecipe`, so might as well just call a method on it over generating another lambda.

patches/minecraft/net/minecraft/world/inventory/BrewingStandMenu.java.patch
ChampionAsh5357: I looked within the menu, and it does seem there is nothing for the level. Though most menu containers take in the level, so it could be adapted. But, it doesn't matter either way so just leave it for now.

ChampionAsh5357 (commented):
Only a few comments:
- Fix the other potion tests.
- Is there any use for the `potion_containers` tag or is it just there?

src/main/java/net/minecraftforge/common/brewing/MixingBrewingRecipe.java
Minecraftschurli: there is no `getValueOptional` (at least I can't find one)

Minecraftschurli:
the `potion_containers` tag is for adding the ability to add more potion containers (it is used instead of the hard coded `== Items.POTION`

src/main/java/net/minecraftforge/common/brewing/BrewingContainerWrapper.java
KnightMiner: No? You can make more than one wrapper for a single TE. One of my mods has what I will summarize as a multi-furnace, I create a wrapper per slot.

In any case, that is an implementation detail, you could use this design just fine, I think its just not a great idea to force other mods implementing brewing stands to this design. An interface that you implement would give them full control.

src/main/java/net/minecraftforge/common/brewing/IBrewingRecipe.java
Minecraftschurli: I do with my next commit xD

Minecraftschurli:
@KnightMiner @ChampionAsh5357 I hope I have all of your questions/suggestions answered/implemented now

src/main/java/net/minecraftforge/common/brewing/MixingBrewingRecipe.java
ChampionAsh5357: I'm thinking of vanilla registries, my bad.

patches/minecraft/net/minecraft/world/level/block/entity/BrewingStandBlockEntity.java.patch
```diff
@@ -1,29 +1,42 @@
...
-+      net.minecraftforge.event.ForgeEventFactory.onPotionBrewed(p_155293_);
++   private static void doBrew(Level p_155291_, BlockPos p_155292_, BrewingStandBlockEntity p_155293_) {
++      if (net.minecraftforge.event.ForgeEventFactory.onPotionAttemptBrew(p_155293_.f_58975_)) return;
++      ItemStack itemstack = p_155293_.f_58975_.get(3);
++
++      net.minecraftforge.common.brewing.BrewingRecipeHelper.brewPotions(p_155291_, p_155293_, f_155280_, new int[]{0, 1, 2});
++      net.minecraftforge.event.ForgeEventFactory.onPotionBrewed(p_155293_.f_58975_);
 +      if (itemstack.hasContainerItem()) {
 +         ItemStack itemstack1 = itemstack.getContainerItem();
 +         itemstack.m_41774_(1);
```
ChampionAsh5357: You could simply redirect this call to some external helper to reduce the patch size here. Also, we shouldn't be replacing method signatures with our own.

src/main/java/net/minecraftforge/common/brewing/BrewingRecipe.java
```diff
@@ -19,54 +19,89 @@
...
     }
 
-    @Override
-    public boolean isIngredient(ItemStack ingredient)
+    public static class Serializer extends ForgeRegistryEntry<RecipeSerializer<?>> implements RecipeSerializer<BrewingRecipe>
     {
-        return this.ingredient.test(ingredient);
+
+        @Override
+        public BrewingRecipe fromJson(final ResourceLocation p_199425_1_, final JsonObject p_199425_2_)
```
ChampionAsh5357: Readable parameter names.

src/main/java/net/minecraftforge/common/brewing/BrewingRecipe.java
```diff
@@ -19,54 +19,89 @@
...
+        {
+            Ingredient base = Ingredient.fromJson(GsonHelper.getAsJsonObject(p_199425_2_, "base"));
+            Ingredient reagent = Ingredient.fromJson(GsonHelper.getAsJsonObject(p_199425_2_, "reagent"));
+            ItemStack result = ShapedRecipe.itemStackFromJson(GsonHelper.getAsJsonObject(p_199425_2_, "result"));
+            return new BrewingRecipe(p_199425_1_, base, reagent, result);
+        }
+
+        @Nullable
+        @Override
+        public BrewingRecipe fromNetwork(final ResourceLocation p_199426_1_, final FriendlyByteBuf p_199426_2_)
```
ChampionAsh5357: Readable parameter names.

src/main/java/net/minecraftforge/common/brewing/BrewingRecipe.java
```diff
@@ -19,54 +19,89 @@
...
+        {
+            Ingredient base = Ingredient.fromNetwork(p_199426_2_);
+            Ingredient reagent = Ingredient.fromNetwork(p_199426_2_);
+            ItemStack result = p_199426_2_.readItem();
+            return new BrewingRecipe(p_199426_1_, base, reagent, result);
+        }
+
+        @Override
+        public void toNetwork(final FriendlyByteBuf p_199427_1_, final BrewingRecipe p_199427_2_)
```
ChampionAsh5357: Readable parameter names.

src/main/java/net/minecraftforge/common/brewing/BrewingRecipeHelper.java
```diff
@@ -0,0 +1,103 @@
...
+     * Used by the brewing stand to determine if its contents can be brewed.
+     * Extra parameters exist to allow modders to create bigger brewing stands
+     * without much hassle
+     */
+    public static boolean canBrew(Level level, Container container, int reagentSlot, int[] baseSlots)
+    {
+        final ItemStack reagent = container.getItem(reagentSlot);
+        if (reagent.isEmpty()) return false;
+        final RecipeManager recipeManager = level.getRecipeManager();
+        for (int baseSlot : baseSlots) {
```
ChampionAsh5357: New line brace.

src/main/java/net/minecraftforge/common/brewing/BrewingRecipeHelper.java
```diff
@@ -0,0 +1,103 @@
...
+            ItemStack base = container.getItem(baseSlot);
+            final IBrewingContainer c = new IBrewingContainer.Impl(base, reagent);
+            Optional<IBrewingRecipe> recipeFor = recipeManager.getRecipeFor(ForgeMod.BREWING, c, level);
+            if (recipeFor.isPresent()) {
```
ChampionAsh5357: New line brace.

src/main/java/net/minecraftforge/common/brewing/BrewingRecipeHelper.java
```diff
@@ -0,0 +1,103 @@
...
+     * without much hassle
+     */
+    public static boolean canBrew(Level level, Container container, int reagentSlot, int[] baseSlots)
+    {
+        final ItemStack reagent = container.getItem(reagentSlot);
+        if (reagent.isEmpty()) return false;
+        final RecipeManager recipeManager = level.getRecipeManager();
+        for (int baseSlot : baseSlots) {
+            ItemStack base = container.getItem(baseSlot);
+            final IBrewingContainer c = new IBrewingContainer.Impl(base, reagent);
```
ChampionAsh5357: Inconsistent parameter names across usages: `c` vs `wrapper`.

src/main/java/net/minecraftforge/common/brewing/BrewingRecipeHelper.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Used by the brewing stand to brew its inventory Extra parameters exist to
+     * allow modders to create bigger brewing stands without much hassle
+     */
+    public static void brewPotions(Level level, Container container, int reagentSlot, int[] baseSlots)
+    {
+        final ItemStack reagent = container.getItem(reagentSlot);
+        final RecipeManager recipeManager = level.getRecipeManager();
+        for (int baseSlot : baseSlots) {
```
ChampionAsh5357: New line brace.

src/main/java/net/minecraftforge/common/brewing/ContainerBrewingRecipe.java
```diff
@@ -0,0 +1,114 @@
...
+    public boolean isBase(final ItemStack base)
+    {
+        return base.is(base());
+    }
+
+    public static class Serializer extends ForgeRegistryEntry<RecipeSerializer<?>> implements RecipeSerializer<ContainerBrewingRecipe>
+    {
+
+        @Override
+        public ContainerBrewingRecipe fromJson(final ResourceLocation p_199425_1_, final JsonObject p_199425_2_)
```
ChampionAsh5357: Readable parameter names.

src/main/java/net/minecraftforge/common/brewing/ContainerBrewingRecipe.java
```diff
@@ -0,0 +1,114 @@
...
+            Ingredient reagent = Ingredient.fromJson(GsonHelper.getAsJsonObject(p_199425_2_, "reagent"));
+            Item result = ForgeRegistries.ITEMS.getValue(new ResourceLocation(GsonHelper.getAsString(p_199425_2_, "result")));
+            if (result == null)
+                throw new JsonParseException("Invalid item supplied for result in " + p_199425_1_);
+            return new ContainerBrewingRecipe(p_199425_1_, base, reagent, result);
+        }
+
+        @Nullable
+        @Override
+        public ContainerBrewingRecipe fromNetwork(final ResourceLocation p_199426_1_, final FriendlyByteBuf p_199426_2_)
```
ChampionAsh5357: Readable parameter names.

src/main/java/net/minecraftforge/common/brewing/ContainerBrewingRecipe.java
```diff
@@ -0,0 +1,114 @@
...
+        {
+            Item base = ForgeRegistries.ITEMS.getValue(p_199426_2_.readResourceLocation());
+            Ingredient reagent = Ingredient.fromNetwork(p_199426_2_);
+            Item result = ForgeRegistries.ITEMS.getValue(p_199426_2_.readResourceLocation());
+            return new ContainerBrewingRecipe(p_199426_1_, base, reagent, result);
+        }
+
+        @Override
+        public void toNetwork(final FriendlyByteBuf p_199427_1_, final ContainerBrewingRecipe p_199427_2_)
```
ChampionAsh5357: Readable parameter names.

src/main/java/net/minecraftforge/common/brewing/IBrewingRecipe.java
```diff
@@ -20,26 +20,32 @@
...
+    }
 
-    /**
-     * Returns true is the passed ItemStack is an input for this recipe. "Input"
-     * being the item that goes in one of the three bottom slots of the brewing
-     * stand (e.g: water bottle)
-     */
-    boolean isInput(ItemStack input);
+    @Override
+    default boolean canCraftInDimensions(int p_43999_, int p_44000_)
```
ChampionAsh5357: Readable parameter names.

Not needed, but still an interesting thought. Since you mentioned that this could be expanded to larger containers, the width and height determined here could be used to calculate if there are enough spaces for more than the specified number of brewing recipes.

src/main/java/net/minecraftforge/common/brewing/MixingBrewingRecipe.java
```diff
@@ -0,0 +1,114 @@
...
+    public boolean isBase(final ItemStack base)
+    {
+        return PotionUtils.getPotion(base) == base() && base.is(Tags.Items.POTION_CONTAINERS);
+    }
+
+    public static class Serializer extends ForgeRegistryEntry<RecipeSerializer<?>> implements RecipeSerializer<MixingBrewingRecipe>
+    {
+
+        @Override
+        public MixingBrewingRecipe fromJson(final ResourceLocation p_199425_1_, final JsonObject p_199425_2_)
```
ChampionAsh5357: Readable parameter names.

src/main/java/net/minecraftforge/common/brewing/MixingBrewingRecipe.java
```diff
@@ -0,0 +1,114 @@
...
+            Ingredient reagent = Ingredient.fromJson(p_199425_2_.getAsJsonObject("reagent"));
+            Potion result = ForgeRegistries.POTIONS.getValue(new ResourceLocation(p_199425_2_.get("result").getAsString()));
+            if (result == null)
+                throw new JsonParseException("Invalid item supplied for result in " + p_199425_1_);
+            return new MixingBrewingRecipe(p_199425_1_, base, reagent, result);
+        }
+
+        @Nullable
+        @Override
+        public MixingBrewingRecipe fromNetwork(final ResourceLocation p_199426_1_, final FriendlyByteBuf p_199426_2_)
```
ChampionAsh5357: Readable parameter names.

src/main/java/net/minecraftforge/common/brewing/MixingBrewingRecipe.java
```diff
@@ -0,0 +1,114 @@
...
+        {
+            Potion base = ForgeRegistries.POTIONS.getValue(p_199426_2_.readResourceLocation());
+            Ingredient reagent = Ingredient.fromNetwork(p_199426_2_);
+            Potion result = ForgeRegistries.POTIONS.getValue(p_199426_2_.readResourceLocation());
+            return new MixingBrewingRecipe(p_199426_1_, base, reagent, result);
+        }
+
+        @Override
+        public void toNetwork(final FriendlyByteBuf p_199427_1_, final MixingBrewingRecipe p_199427_2_)
```
ChampionAsh5357: Readable parameter names.

src/main/java/net/minecraftforge/common/data/BrewingRecipeBuilder.java
```diff
@@ -0,0 +1,395 @@
...
+            @Override
+            public void serializeRecipeData(JsonObject tag)
+            {
+                tag.add("base", this.base.toJson());
+                tag.add("reagent", this.reagent.toJson());
+                JsonObject o = new JsonObject();
+                final ResourceLocation result = ForgeRegistries.ITEMS.getKey(this.result.getItem());
+                o.addProperty("item", result.toString());
+                o.addProperty("count", this.result.getCount());
+                if (this.result.hasTag()) {
```
ChampionAsh5357: New line brace.

ChampionAsh5357 (changes requested):
Mainly just formatting fixes and patch size reductions.

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -116,6 +125,16 @@
 
     public static final RegistryObject<Attribute> REACH_DISTANCE = ATTRIBUTES.register("reach_distance", () -> new RangedAttribute("generic.reachDistance", 5.0D, 0.0D, 1024.0D).setSyncable(true));
 
+    public static final Supplier<RecipeType<IBrewingRecipe>> BREWING = Lazy.of(() -> RecipeType.register("forge:brewing"));
```
ChampionAsh5357, line 128: You should still resolve the registry entry during the mod event bus either at registry event for serializers or common setup.

ChampionAsh5357 (changes requested):
Main thing of concern is what I mentioned prior about what can go in what slots. That becomes majorly apparent when trying to shift click in anything. It tries to go to the reagent slot and then does not allow any other item to shift click anymore. At least blaze powder should check if it can enter the extraneous slot before going to the reagent. Also, the potion slots should also be able to shift click into instead of just succeeding and not merging into the main reagent slot.

Minecraftschurli:
> Main thing of concern is what I mentioned prior about what can go in what slots. That becomes majorly apparent when trying to shift click in anything. It tries to go to the reagent slot and then does not allow any other item to shift click anymore. At least blaze powder should check if it can enter the extraneous slot before going to the reagent. Also, the potion slots should also be able to shift click into instead of just succeeding and not merging into the main reagent slot.

The problem here is that checking the validity would need a check for the recipes but to check the recipes I would need access to the world which is not available in the BrewingStandMenu. So I would have to do a check if the Container is the BrewingStandBlockEntity (which is true on the server) and then use its world and on the client use net.minecraft.client.Minecraft.getInstance().level . That's the only way I can think of to accomplish what you are requesting.

src/main/java/net/minecraftforge/common/brewing/BrewingRecipeHelper.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Proxy method for BrewingStandMenu.PotionSlot#mayPlaceItem
+     */
+    public static boolean mayPlacePotion(Container container, ItemStack stack)
+    {
+        if (container instanceof BrewingStandBlockEntity stand)
+            return isValidInput(stand.getLevel(), stack);
+        else
+            return isValidInput(Minecraft.getInstance().level, stack);
+    }
```
ChampionAsh5357: So this crashes the dedicated server whenever the brewing stand is loaded. I think it may be better to just add another field to the container that grabs the level from the player instance and use that instead for checking here.

Minecraftschurli: The field is only a solution for the check in quickMoveStack because the Slot classes are static (so an additional parameter on the Slot constructor and a corresponding field would be needed)

ChampionAsh5357: Thought about it some more. Probably the best thing to do in this case is use `LogicalSidedProvider` to get the client world safely. Or you could replace all level logic with `LogicalSidedProvider` if you use the forge interface for getting the current logical side.

src/main/java/net/minecraftforge/network/VanillaConnectionNetworkFilter.java
```diff
@@ -108,10 +112,28 @@ private static ClientboundCommandsPacket filterCommandList(ClientboundCommandsPa
         return new ClientboundCommandsPacket(newRoot);
     }
 
+    /**
+     * Filter for SUpdateRecipesPacket. Filters out any ArgumentTypes that are not in the "minecraft" namespace.
```
ChampionAsh5357: Old classname.

src/main/java/net/minecraftforge/common/brewing/BrewingRecipe.java
```diff
@@ -19,54 +19,89 @@
...
+import net.minecraft.world.item.crafting.RecipeSerializer;
+import net.minecraft.world.item.crafting.ShapedRecipe;
+import net.minecraftforge.common.ForgeMod;
+import net.minecraftforge.registries.ForgeRegistryEntry;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
-public class BrewingRecipe implements IBrewingRecipe
+public record BrewingRecipe(@Nonnull ResourceLocation id,
```
ChampionAsh5357, line 39: Add class javadoc explaining what this does.

src/main/java/net/minecraftforge/common/brewing/ContainerBrewingRecipe.java
```diff
@@ -0,0 +1,114 @@
...
+import net.minecraft.world.item.crafting.Ingredient;
+import net.minecraft.world.item.crafting.RecipeSerializer;
+import net.minecraftforge.common.ForgeMod;
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.ForgeRegistryEntry;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public record ContainerBrewingRecipe(@Nonnull ResourceLocation id,
```
ChampionAsh5357, line 42: Add class javadoc explaining what this does.

src/main/java/net/minecraftforge/common/brewing/MixingBrewingRecipe.java
```diff
@@ -0,0 +1,114 @@
...
+import net.minecraft.world.item.crafting.RecipeSerializer;
+import net.minecraftforge.common.ForgeMod;
+import net.minecraftforge.common.Tags;
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.ForgeRegistryEntry;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public record MixingBrewingRecipe(@Nonnull ResourceLocation id,
```
ChampionAsh5357, line 42: Add class javadoc explaining what this does.

src/main/java/net/minecraftforge/common/brewing/IBrewingRecipe.java
```diff
@@ -20,26 +20,32 @@
 package net.minecraftforge.common.brewing;
 
 import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.crafting.Recipe;
+import net.minecraft.world.item.crafting.RecipeType;
+import net.minecraft.world.level.Level;
+import net.minecraftforge.common.ForgeMod;
 
-public interface IBrewingRecipe {
```
ChampionAsh5357, line 24: Needs to be documented for anyone else who wants to create their own.

src/main/java/net/minecraftforge/common/brewing/IBrewingContainer.java
```diff
@@ -0,0 +1,93 @@
...
+    default boolean stillValid(final Player p_18946_)
+    {
+        return true;
+    }
+
+    ItemStack base();
+
+    ItemStack reagent();
+
+    record Impl(ItemStack base, ItemStack reagent) implements IBrewingContainer
```
ChampionAsh5357, line 103: I know this is wrapping around `#base` and `#reagent` because of the method overrides in a record, but that should probably be documented here as part of the spec.

src/main/java/net/minecraftforge/common/brewing/BrewingRecipeHelper.java
```diff
@@ -0,0 +1,155 @@
...
+            ItemStack base = container.getItem(baseSlot);
+            final IBrewingContainer wrapper = new IBrewingContainer.Impl(base, reagent);
+            Optional<IBrewingRecipe> recipeFor = recipeManager.getRecipeFor(ForgeMod.BREWING.get(), wrapper, level);
+            if (recipeFor.isPresent()) return true;
+        }
+        return false;
+    }
+
+    /**
+     * Used by the brewing stand to brew its inventory Extra parameters exist to
```
ChampionAsh5357: What are extra parameters? You are referring to the fact that `baseSlots` takes in an integer array to get all base slots to construct. Javadoc should be clearer.

src/main/java/net/minecraftforge/common/brewing/BrewingRecipeHelper.java
```diff
@@ -0,0 +1,155 @@
...
+import net.minecraftforge.fml.util.thread.EffectiveSide;
+import net.minecraftforge.fmllegacy.LogicalSidedProvider;
+
+import java.util.Optional;
+
+public class BrewingRecipeHelper
+{
+    /**
+     * Used by the brewing stand to determine if its contents can be brewed.
+     * Extra parameters exist to allow modders to create bigger brewing stands
```
ChampionAsh5357: What are extra parameters? You are referring to the fact that `baseSlots` takes in an integer array to get all base slots to construct. Javadoc should be clearer.

ChampionAsh5357 (changes requested):
Logic-wise, it's fine now. Just needs clearer and more documentation.

ChampionAsh5357 (approved):
This should be fine now imo.

TheCurle:
On further inspection, we've decided that this change is out of scope for Forge.

It *is* a good idea and somewhat well implemented, but there are some issues. 

Primarily, moving an existing vanilla system to datapacks and adding data generators for them is the task of Mojang.

Additionally, there seems to be a fundamental systemic flaw with this system relating to synchronisation against the server, because of the way the potions system was designed beforehand. 

We'll get in touch and ask *Mojang* to implement this at the earliest convenience, but until then this isn't something we're willing to accept in Forge.

Apologies.

PR #7950, opened 2021-07-29
Expose the world's default seed for modded ChunkGenerators and BiomeSources

TelepathicGrunt:
ChunkGenerators and BiomeSources require a seed in their constructor that basically controls the terrain and biome layout. By default, vanilla dimension json files only allow modders to have a fixed seed field which means every world will have the same exact dimension layout in terms of terrain and biome layout. Getting the default seed (what goes into the seed text box and all), is not possible in ChunkGenerators and BiomeSources constructors as all it gets is the biome registry.

This PR creates a DefaultWorldSeed class with a public static getSeed method for any modder to get the seed anywhere they need it. Especially for ChunkGenerators and BiomeSources constructors but other places can also use it in case I miss any other use case. Only issue is that the setSeedInternal has to be public static as well since it has to grab the seed from WorldGenSettings.


TelepathicGrunt:
Closing this as this forge pr is more flexible and allows users to make most dimension no longer have fixed seeds too https://github.com/MinecraftForge/MinecraftForge/pull/7955

Issue #7883, opened 2021-07-22
1.17.1 Forge should pause the game in dev if it detects unregistered worldgen elements


TelepathicGrunt:
This is more of an idea that I would like to put on the table so there is some discussion around it. The issue seen in 1.16.2 and onwards is that while the new worldgen system is both a blessing and a curse, it does require modders to register their ConfiguredFeatures, ConfiguredStructures, and ConfiguredCarvers. Failure to register these can cause rare hard-to-debug issues where the unregistered worldgen element nukes the registered ones from the same biome and generation stage.

How prevalent was this issue in 1.16? Very. I even made a mod called Blame to try and detect unregistered worldgen elements and it caught many mods. So many that I cannot keep up with reporting the issues and at one point, it found I forgot to register a ConfiguredFeature myself! Here is a picture of a tiny slice of all the mods with unregistered worldgen stuff that I reported to. A small slice.
![unregistered](https://user-images.githubusercontent.com/40846040/126711053-577a7c61-637b-4afb-a647-a67ef1255d32.png)

Now that Forge is on 1.17.1, I think this may be a good time to think about having Forge annoy modders in some way to register their worldgen elements but ONLY in development environment. An idea is to use Util.pauseInIde and just pause the IDE in dev to force modders to read the comment at the spot about Forge finding an unregistered worldgen element (would only pause once max per run). The reason I propose pausing the IDE is because I worry that logs may be ignored and a pause makes absolutely certain the modder sees it. 

Of course, how and where to check for unregistered worldgen elements may be tougher without iterating over all biomes and their contents manually. But that may be needed.

Thoughts on this? Is this something worth looking into and for someone to make a PR for? (I might be able to do it when I get some free time later if it is something people really want added to Forge)

Commoble:
I'm okay with this

williambl:
Why not just crash the game with a clear message in the log to register them?

TelepathicGrunt:
@williambl ngl, I actually would love that but several people expressed displeasure when I said for Forge to just crash the game for unregistered worldgen stuff lol. Pausing the IDE is less harsh and should get the same point across hopefully.

Pechayland:
That happened when I tested Twilight Forest mod "The game crashed whilst colliding entity with block"
Log: https://gist.github.com/Pechayland/6bd112690ff688e26c1d97bfc376861f

LexManos:
[...]

PR #7951, opened 2021-07-29
Validate for unregistered or broken worldgen elements

TelepathicGrunt:
Here's an implementation of the system I described in https://github.com/MinecraftForge/MinecraftForge/issues/7883
However, this PR will crash the game in ~~both production and~~ development environments ~~as it turned out there does not seem to be any good way to tell when the game is running in an IDE with using Java.~~ Edit: FMLEnvironment#production allows this code to only run during development environments now

But since 1.17 does not have a recommended build, crashing may be alright to do. The crash will provide a prettified JSON output of which worldgen element is unregistered and what biome it was added to. The JSON alone should make it very easy to find which worldgen element is unregistered. Example: https://paste.ee/p/nGSx8

Part of making the JSON output involved putting the worldgen element through its codec which means any broken configuredfeature or whatnot, will trigger a different error message explaining that the modder may had messed up the values for their confiuredfeature. For example, setting the trunk placer for a custom tree to an invalid value can break a lot of datapacks or mods in nasty ways like it did here: https://github.com/OrderedChaos-Dev/ProjectVibrantJourneys/issues/150

So having Forge now check for both unregistered or broken worldgen elements, this should greatly help modders make sure they are doing things correctly in a more mod/datapack compatible way.

It would've been ideal to only run this code in development environments ~~but since it now has to run in production,~~ I tried to so some caching to speed up the checking and validating. 

If the crashing is not ideal, I can change this to just logging out to do logspam instead. However, the amount of logspam can be insanely large but maybe that's a good thing to annoy modders to registering their stuff. 

Feedback would be great. Otherwise, this PR is ready for review/triage


src/main/java/net/minecraftforge/common/world/WorldGenValidator.java
```diff
@@ -0,0 +1,122 @@
...
+     * Handles checking if the worldgen element is broken (unable to be turned into JSON) or unregistered.
+     * If an issue is found, this will pause the IDE so the modder can come here and see the error and the worldgenElement for which element is the issue.
+     * The console output will have the Forge message as last thing when paused too.
+     */
+    private static <T> void validate(T worldgenElement, Registry<T> registry, String worldgenElementType, Codec<T> codec, ResourceLocation biomeName) {
+
+        // Checks to make sure the element can be turned into JSON safely or else Minecraft will explode with vague errors.
+        JsonElement worldgenElementJSON;
+        try {
+            worldgenElementJSON = codec.encode(worldgenElement, JsonOps.INSTANCE, JsonOps.INSTANCE.empty()).get().left().orElseThrow(UnsupportedOperationException::new);
```
Cyborgmas: Is this try/catch necessary? encode won't throw, so the only thing that can is your own `#orElseThrow` so at that point just throw the exception from there. I would also consider using `#getOrThrow` on the dataResult instead to be able to output the parsing error message as well.

Cyborgmas (commented):
Don't think its a great idea to crash in production, but luckily you can check `FMLEnvironment#production` to avoid that. It also check for a system property allowing you to simulate prod in dev

src/main/java/net/minecraftforge/common/world/WorldGenValidator.java
TelepathicGrunt: I thought I had moments when it crash but you're probably right. I'll make the change but can you check out the change to make sure it looks right? I haven't used getOrThrow before

Won-Ton:
Appreciate the intent but this crashes even with the testmod disabled due to a flaw in vanilla:

`TagMatchTest` is used in things like ore-features as a blockstate predicate and is backed by a `Tag<Block>`. Tags have a lifecycle linked to their `TagCollection` which reload at various points causing the `Tag` instances held in the `TagMatchTest`s become dissociated from their collection and fail to decode with the `"Unknown tag ..."` message.

You can confirm this by simply trying to open or create a world for a second time during the test-client session.

TelepathicGrunt:
The validation hook was moved to RegistryAccess$builtin at the end of that method where it does not run into the tag issue. This is based on the mod Blame's hook as no issue has been reported over months of having the hook at that spot for parsing worldgen elements with tags into json. The downside of this new spot means the validation is ran multiple times due to the dynamic registry mess backend.

Here are video evidence of the testing of the new spot.

With no test mod running and me creating a world, exiting it, and re-entering it. No crash: https://streamable.com/mbzaap

With test mods running and I add a broken ConfiguredFeature by BiomeLoadingEvent. Crashes with info about it: https://streamable.com/gyf8za

With test mods running and I add an unregistered ConfiguredFeature by BiomeLoadingEvent. Crashes with info saying to register it and has a JSON output to make it easier to know what ConfiguredFeature is unregistered: https://streamable.com/folfma

Won-Ton:
Looks like that's fixed it 👍

It's taking sub 20ms each validation for me so I reckon you'd have to be adding a _lot_ of biomes/content for it to become noticeably slow.

It wouldn't now catch unregistered stuff added via the DynamicRegistriesEvent (assuming that, in some form, gets pulled) but presumably at that point it doesn't matter so much since those biomes aren't en/de-coded in full iirc

src/main/java/net/minecraftforge/common/world/WorldGenValidator.java
```diff
@@ -0,0 +1,121 @@
...
+            BiomeGenerationSettings biomeGenerationSettings = biome.getGenerationSettings();
+            ResourceLocation biomeName = biome.getRegistryName();
+
+            biomeGenerationSettings.features().forEach(generationStageFeatureList ->
+                    generationStageFeatureList.forEach(configuredFeatureSupplier -> {
+                        ConfiguredFeature<?, ?> feature = configuredFeatureSupplier.get();
+                        if (!checkedElements.contains(feature)) {
+                            validate(feature, configuredFeatureRegistry, "ConfiguredFeature", ConfiguredFeature.DIRECT_CODEC, biomeName);
+                            checkedElements.add(feature);
+                        }
```
sciwhiz12: This could possibly be a `if (checkedElements.add(feature))` instead, as `Set#add` returns `true` if the the element was added to the set (and therefore, the element was not previously within the set). This would have the implication of the element being marked as 'checked' before it is actually checked, though.
_(same comment with the rest of the similar `contains` then `add` patterns in the method)_

src/main/java/net/minecraftforge/common/world/WorldGenValidator.java
```diff
@@ -0,0 +1,121 @@
...
+                    }));
+
+            for (GenerationStep.Carving carvingStage : GenerationStep.Carving.values()) {
+                biomeGenerationSettings.getCarvers(carvingStage).forEach(configuredWorldCarverSupplier -> {
+                    ConfiguredWorldCarver<?> carver = configuredWorldCarverSupplier.get();
+                    if (!checkedElements.contains(carver)) {
+                        validate(carver, configuredCarverRegistry, "ConfiguredCarver", ConfiguredWorldCarver.DIRECT_CODEC, biomeName);
```
sciwhiz12: The `"ConfiguredCarver"` within the string literal would best match the name of the class, `ConfiguredWorldCarver`.

src/main/java/net/minecraftforge/common/world/WorldGenValidator.java
```diff
@@ -0,0 +1,121 @@
...
+                        checkedElements.add(carver);
+                    }
+                });
+            }
+
+            biomeGenerationSettings.structures().forEach(structureFeatureSupplier -> {
+                ConfiguredStructureFeature<?, ?> structureFeature = structureFeatureSupplier.get();
+                if (!checkedElements.contains(structureFeature)) {
+                    validate(structureFeature, configuredStructureRegistry, "ConfiguredStructure", ConfiguredStructureFeature.DIRECT_CODEC, biomeName);
```
sciwhiz12: The `"ConfiguredStructure"` within the string literal would best match the name of the class, `ConfiguredStructureFeature`.

src/test/java/net/minecraftforge/debug/world/WorldGenValidatorTest.java
```diff
@@ -0,0 +1,83 @@
...
+import org.apache.logging.log4j.Logger;
+
+@Mod(WorldGenValidatorTest.MODID)
+public class WorldGenValidatorTest {
+    static final String MODID = "worldgen_validator_test";
+    private static final Logger LOGGER = LogManager.getLogger(MODID);
+    private static final boolean ENABLED = false;
+
+    public WorldGenValidatorTest(){
+        if(ENABLED) {
```
sciwhiz12: Nitpick: spaces before and after parentheses in declarations and conditionals.
```suggestion
    public WorldGenValidatorTest() {
        if (ENABLED) {
```

src/test/java/net/minecraftforge/debug/world/WorldGenValidatorTest.java
```diff
@@ -0,0 +1,83 @@
...
+            MinecraftForge.EVENT_BUS.addListener(this::onBiomeLoading);
+        }
+    }
+
+    public void onBiomeLoading(BiomeLoadingEvent event){
+        if(event.getName().equals(Biomes.SUNFLOWER_PLAINS.location())){
```
sciwhiz12: Nitpick: spaces before and after parentheses in declarations and conditionals.
```suggestion
    public void onBiomeLoading(BiomeLoadingEvent event) {
        if (event.getName().equals(Biomes.SUNFLOWER_PLAINS.location())) {
```

src/main/java/net/minecraftforge/common/world/WorldGenValidator.java
```diff
@@ -0,0 +1,121 @@
...
+                                "Check to make sure your codecs are correct or it and that the values you give it does not exceed any limits in the object's codecs. Error: " + errorMsg));
+
+        // Checks to make sure the element is registered.
+        if (!registry.getResourceKey(worldgenElement).isPresent()) {
+            throw new UnsupportedOperationException(
+                    "Unregistered " + worldgenElementType + " found in " + biomeName + " biome. " +
+                            "Please register your " + worldgenElementType + " using Minecraft's Registry class within FMLCommonSetupEvent's enqueueWork. " +
+                            "Failure to do so may cause unregistered worldgen elements to nuke registered ones from the same biome which is difficult for players to debug. " +
+                            "See this json of the unregistered worldgen element to know which one is unregister: \n" + (worldgenElementJSON != null ? gson.toJson(worldgenElementJSON) : "Unable to be parsed into JSON"));
```
sciwhiz12: This can possibly be moved to use text blocks and `String#formatted(Object...)`: `"""Unregisterd %s found in %s ...""".formatted(worldgenElementType, ...)`.

src/test/java/net/minecraftforge/debug/world/WorldGenValidatorTest.java
```diff
@@ -0,0 +1,83 @@
...
+import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.event.world.BiomeLoadingEvent;
+import net.minecraftforge.fml.common.Mod;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+@Mod(WorldGenValidatorTest.MODID)
+public class WorldGenValidatorTest {
+    static final String MODID = "worldgen_validator_test";
+    private static final Logger LOGGER = LogManager.getLogger(MODID);
```
sciwhiz12, line 53: This `Logger` is unused. Either remove it, or use it in the constructor to leave an informational/warning message that this test is enabled and will cause a crash on validation.

sciwhiz12 (changes requested):
I've not tested this PR yet, but I do have some comments. Also, please fix the formatting of your files, specifically that opening braces should be on newlines.

sciwhiz12:
Hello! @TelepathicGrunt, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable. If we do not hear back from you after some time, this PR will be closed for inactivity.

PR #7952, opened 2021-07-29
[1.17] Network capabilities

josephtarbit:
This is a reformed version of #7630 for 1.17.

The aim of this PR is to provide a simple contract for capabilities to synchronize data along side vanilla's synchronization of it's own data.

### `ItemStack` Capabilities

### `INetworkCapability`

```java
void write(FriendlyByteBuf out, boolean writeAll);

void read(FriendlyByteBuf in);

boolean requiresSync();
```


patches/minecraft/net/minecraft/network/FriendlyByteBuf.java.patch
```diff
@@ -25,23 +25,30 @@
...
-          CompoundTag compoundtag = null;
+-         CompoundTag compoundtag = null;
 -         if (item.m_41465_() || item.m_41468_()) {
 -            compoundtag = p_130056_.m_41783_();
++         CompoundTag compoundtag = new CompoundTag();
 +         if (item.isDamageable(p_130056_) || item.m_41468_()) {
 +            compoundtag = limitedTag ? p_130056_.getShareTag() : p_130056_.m_41783_();
           }
- 
+-
```
Sirttas: Remove this useless patch

patches/minecraft/net/minecraft/server/level/ChunkMap.java.patch
```diff
@@ -47,3 +47,17 @@
...
+    void m_143075_(ChunkPos p_143076_, ChunkHolder.FullChunkStatus p_143077_) {
+       this.f_143031_.m_156794_(p_143076_, p_143077_);
++   }
++
++   // Forge: expose TrackedEntity#seenBy
++   @Nullable
++   public Set<ServerPlayerConnection> getSeenBy(Entity entity) { 
++      TrackedEntity trackedEntity = this.f_140150_.get(entity.m_142049_());
++      return trackedEntity == null ? null : java.util.Collections.unmodifiableSet(trackedEntity.f_140475_);
+    }
```
Sirttas, line 61: Would it be better to return `java.util.Collections.emptySet()` ?

patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
```diff
@@ -384,14 +384,26 @@
...
              if (map == null) {
                 map = Maps.newEnumMap(EquipmentSlot.class);
              }
+@@ -2327,7 +_,8 @@
+             if (!itemstack1.m_41619_()) {
+                this.m_21204_().m_22178_(itemstack1.m_41638_(equipmentslot));
+             }
+-         }
++         } else
++             net.minecraftforge.network.ForgeNetwork.sendEquipmentSlotCapabilities(this, equipmentslot, itemstack1, false);
```
Sirttas: You can do this in a single line

patches/minecraft/net/minecraft/world/inventory/AbstractContainerMenu.java.patch
```diff
@@ -1,17 +1,32 @@
...
+-         if (!ItemStack.m_41728_(itemstack, p_150437_)) {
++         if (!itemstack.equals(p_150437_, true)) {
+             ItemStack itemstack1 = p_150438_.get();
+             this.f_150394_.set(p_150436_, itemstack1);
+             if (this.f_150397_ != null) {
+                this.f_150397_.m_142074_(this, p_150436_, itemstack1);
+             }
+-         }
++         } else
++             if (this.f_150397_ != null)
```
Sirttas: why is this `else if` on 2 lines did you mean to do an `else if` or an `else { if () {} }` ?

src/main/java/net/minecraftforge/common/capabilities/CapabilityDispatcher.java
```diff
@@ -141,6 +153,47 @@ public void deserializeNBT(CompoundTag nbt)
...
 
+    @Override
+    public void encode(FriendlyByteBuf out, boolean writeAll)
+    {
+        for (Map.Entry<String, INetworkCapability> entry : networkCapabilities.entrySet())
+        {
+            if (!writeAll && !entry.getValue().requiresSync())
+                continue;
+            out.writeBoolean(true);
+            out.writeUtf(entry.getKey(), 0x100);
```
Sirttas: better use a constant for `0x100`, I know it's what vanilla uses but it would increase readability especialy if you ad a javadoc on the constant to explain it.

src/main/java/net/minecraftforge/network/ForgeNetwork.java
```diff
@@ -0,0 +1,283 @@
...
+import net.minecraftforge.common.ForgeMod;
+import net.minecraftforge.fmllegacy.network.NetworkDirection;
+import net.minecraftforge.fmllegacy.network.NetworkRegistry;
+import net.minecraftforge.fmllegacy.network.simple.SimpleChannel;
+import net.minecraftforge.network.message.BlockEntityCapabilitiesMessage;
+import net.minecraftforge.network.message.EntityCapabilitiesMessage;
+import net.minecraftforge.network.message.EquipmentSlotCapabilitiesMessage;
+import net.minecraftforge.network.message.SlotCapabilitiesMessage;
+
+public class ForgeNetwork {
```
Sirttas, line 31: As I said for the 1.16 PR this is the main concern for this PR, do we want to go with forge own network channel. It's aggressive but it also offers a performance gain during capabilities sync.

I really want to see a discussion around it with forge team/triage team.

Sirttas (approved):
Overall this is the best version I have seen of this PR so far and I thing it would be great if it could be merged early on.

patches/minecraft/net/minecraft/server/level/ChunkMap.java.patch
josephtarbit: I used null to distinguish between an entity having nobody tracking it and it not existing.

src/main/java/net/minecraftforge/network/ForgeNetwork.java
```diff
@@ -0,0 +1,283 @@
...
+
+    public static Stream<Connection> getTrackingConnections(Entity entity)
+    {
+        if (entity.level instanceof ServerLevel level)
+        {
+            ServerChunkCache chunkSource = level.getChunkSource();
+            Set<ServerPlayerConnection> players = chunkSource.chunkMap.getSeenBy(entity);
+            if (players != null)
+            {
+                return chunkSource.chunkMap.getSeenBy(entity).stream()
```
Sirttas: You can directly use players here, it would increase perfs and readability.

patches/minecraft/net/minecraft/server/level/ChunkMap.java.patch
Sirttas: It's a choice, I would have returned an empty list as I think that I wouldn't have make an huge difference, maybe if forge team want it changeg.

src/main/java/net/minecraftforge/network/ForgeNetwork.java
josephtarbit: Whoops, didn't see that

PR #7953, opened 2021-07-29
Re-add call to `Ingredient#invalidateAll()`

alcatrazEscapee:
`invalidateAll()` is never called, in 1.17.x or 1.16.x. This seems to have no point anymore. So out it goes!

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeMod.java
+++ b/src/main/java/net/minecraftforge/common/ForgeMod.java
@@ -24,6 +24,7 @@ import net.minecraft.commands.synchronization.ArgumentTypes;
 import net.minecraft.commands.synchronization.ArgumentSerializer;
 import net.minecraft.world.entity.ai.attributes.Attribute;
 import net.minecraft.world.entity.ai.attributes.RangedAttribute;
+import net.minecraft.world.item.crafting.Ingredient;
 import net.minecraft.world.level.material.Fluid;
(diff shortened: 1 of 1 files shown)


LexManos:
Okay, actually looked into this. And it was a casualty of the [1.13 update](https://github.com/MinecraftForge/MinecraftForge/commit/b50b76885256f85cd277124f4309728ac1f94d51#diff-7aadec2bd21bcedb71830e99756fad3c45d2222c2ddb40da034e43a6a7f3fe31R439).
This invalidate call actually needs to be re-added. Because the mechanics that vanilla uses that needed it are still in place.
Namely the fact that the recipe book still uses item ids for searching/auto fill.
And those IDs can change based on worlds.
Honestly it's a small edge case that requires you to load two worlds with different ID maps in the same instance of the game.
But technically its still possible so the call should be readded.

alcatrazEscapee:
Wizard. Well that seems straightforward enough for me to put back in the same place.

LexManos:
Please include the other invalidations that were around it. As they are also the same vein. Where vanilla caches int ID maps.

alcatrazEscapee:
Ore dictionary is pooched.
Stats list is either gone or changed, in either case there's no matching forge patch I can see.
Search tree doesn't use int ID maps, and is also a resource listener already meaning it'll already get invalidated?

PR #7955, opened 2021-07-29
[1.18.x] Expand the LevelStem codec to allow dimension jsons to specify that the dimension's chunk generator should use the overworld/server's seed

### LINKED ISSUE #8396: Seeds for Custom Dimensions does not work (Small MC Bug)
**Minecraft Version:** 1.18.1

**Forge Version:** 1.18.1-39.0.20

**Steps to Reproduce:**
This is a Minecraft issue but a small one i heared Forge does fix small bugs of minecraft, so i hope you can fix it :D

The Link to the issue:
https://bugs.mojang.com/browse/MC-195717

Commoble:
This is an alternative to #7950, serving the same purpose but with a different implementation and usage.

As 7950 states:

> ChunkGenerators and BiomeSources require a seed in their constructor that basically controls the terrain and biome layout. By default, vanilla dimension json files only allow modders to have a fixed seed field which means every world will have the same exact dimension layout in terms of terrain and biome layout. Getting the default seed (what goes into the seed text box and all), is not possible in ChunkGenerators and BiomeSources constructors as all it gets is the biome registry.

## Public APIs
* Dimension jsons may now specify a `"forge:use_server_seed"` boolean field; this field is optional and defaults to false if not present (see this PR's included test resources for an example of a dimension json that includes this). Dimension jsons must still specify the fixed seeds for the chunk generators and biome sources but these seeds can be 0 if the use_server_seed override is present.
* ChunkGenerator and BiomeSource extensions should implement the vanilla withSeed method with the expectation that the seed passed in is the server/overworld seed, and that a chunk generator or biome source appropriate for that seed should be returned from withSeed. withSeed is a vanilla method and this is likely already the intended usage of the method, so this should be compatible with mods' existing custom chunk generators and biome sources, though vanilla generally doesn't use withSeed itself.

## Explanation of Patches

* A patch to LevelStem expands the dimension json format to allow an optional `"forge:use_server_seed"` boolean field (an example of a json using this is included in this PR's test resources).
* A patch to WorldGenSettings's codec causes its dimension registry to be copied and recreated when dimensions are loaded from datapacks; LevelStem instances whose dimension jsons specified that the server seed should be used are reconstructed, using ChunkGenerator.withSeed(serverSeed) to reconstruct their chunk generators.

What the chunk generator does with the provided seed depends on the chunk generator's implementation; the vanilla noise chunk generator takes the provided seed to construct a new chunk generator instance that uses that seed, and also provides the seed to its biome source to allow the biome source to reconstruct itself if necessary.

Compared to #7950, this PR is probably more stable as it doesn't rely on the timing of setting static fields from WorldGenSettings's constructor (WorldGenSettings are constructed in a couple dozen different places), though it's probably less performant due to chunk generators being copied and reconstructed during dimension parsing. This PR also has a larger patch footprint than #7950.

### Procedure
* create two dimension jsons that use the vanilla noise chunk generator and vanilla layered biome source
  * "fixed_seed_dimension" without server seed field and with fixed seeds 12345
  * "server_seed_test" with "forge:use_server_seed" = true and with fixed seeds 12345
* create two worlds with seeds "argent" and "boron", respectively, twice each
### Expected Results
* "fixed_seed_dimension" has the same geography in both argent and boron in both iterations (control dimension)
* "server_seed_dimension"'s geography differs between argent and boron, but remains the same between two iterations of argent and between two iterations of boron
### Actual Results
* argent 1
  * fixed seed dimension -- taiga mountains at xz = 0,0
  * server seed dimension -- savanna at 0,0
* argent 2
  * fixed seed dimension -- taiga mountains at 0,0 (same geography as argent 1)
  * server seed dimension -- savanna at 0,0 (same geography as argent 1)
* boron 1
  * fixed seed dimension -- taiga mountains at 0,0 (same geography as argent)
  * server seed dimension -- snowy tundra at 0,0
* boron 2
  * fixed seed dimension -- taiga mountains at 0,0 (same geography as argent and boron 1)
  * server seed dimension -- snowy tundra at 0,0 (same geography as boron 1)

Actual results match expected results.

## Other Concerns
* 1.16.x also has a withSeed method for chunk generators and biome sources, but strips this method from dedicated server jars due to unused methods being stripped in minecraft versions prior to 1.17, and vanilla only uses this method during part of the singleplayer create world guis. This means that backporting this feature to 1.16.x would require additional forge patches and extensions to make this viable on 1.16 server distributions.

## Relevant Mojang Bug Reports
* [[MC-195717] Custom Dimensions JSON Requires Seed](https://bugs.mojang.com/browse/MC-195717)


TelepathicGrunt (approved):

patches/minecraft/net/minecraft/world/level/dimension/LevelStem.java.patch
```diff
@@ -0,0 +1,26 @@
...
+    public static final ResourceKey<LevelStem> f_63971_ = ResourceKey.m_135785_(Registry.f_122820_, new ResourceLocation("overworld"));
+    public static final ResourceKey<LevelStem> f_63972_ = ResourceKey.m_135785_(Registry.f_122820_, new ResourceLocation("the_nether"));
+@@ -33,8 +_,13 @@
+    private final ChunkGenerator f_63976_;
+ 
+    public LevelStem(Supplier<DimensionType> p_63979_, ChunkGenerator p_63980_) {
++      this(p_63979_, p_63980_, false); // forge: by default, dimension jsons use the fixed seed required by the vanilla format
++   }
++   public final boolean useServerSeed; // forge: whether the dimension parser should use the server seed instead of the fixed seed
++   public LevelStem(Supplier<DimensionType> p_63979_, ChunkGenerator p_63980_, boolean useServerSeed) { // forge: allow dimension jsons to specify that they should use the server seed instead of the fixed seed field
```
ChampionAsh5357, line 21: There should be an `exc` entry for this constructor.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1406,4 +1411,48 @@ public static void modifyAttributes()
...
+        // make a copy of the dimension registry; for dimensions that specify that they should use the server seed instead
+        // of the hardcoded json seed, recreate them with the correct seed
+        MappedRegistry<LevelStem> seededRegistry = new MappedRegistry<>(Registry.LEVEL_STEM_REGISTRY, Lifecycle.experimental());
+        for (Entry<ResourceKey<LevelStem>, LevelStem> entry : originalRegistry.entrySet())
+        {
+            ResourceKey<LevelStem> key = entry.getKey();
+            LevelStem dimension = entry.getValue();
+            if (dimension.useServerSeed)
+            {
+                seededRegistry.register(key, new LevelStem(dimension.typeSupplier(), dimension.generator().withSeed(seed)), originalRegistry.lifecycle(entry.getValue()));
```
ChampionAsh5357: Shouldn't this continue to pass in the parameter into `LevelStem` that it is using the server seed?

ChampionAsh5357 (changes requested):
I think this is fine besides from the two changes mentioned. This also should mention somewhere in a comment that this fixes MC-195717, most likely on `WorldGenSettings#loadDimensionsWithServerSeed`.

patches/minecraft/net/minecraft/world/level/dimension/LevelStem.java.patch
Commoble: Added in latest push

src/main/java/net/minecraftforge/common/ForgeHooks.java
Commoble: Good catch, added this in latest push

ChampionAsh5357 (approved):
This should be fine now, though it'll most likely be deferred until 1.18 (Forge v38). Since the issue is still unresolved by Mojang, it will be still be needed.

TelepathicGrunt (approved):

Commoble:
I've rebased and retargeted this to 1.18 instead of making a separate PR; the test dimensions still behave as expected (albeit with different biomes due to the worldgen changes).

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -433,6 +433,9 @@ protected net.minecraft.world.level.levelgen.NoiseSampler f_158643_ # noiseSetti
 public net.minecraft.world.level.levelgen.NoiseSampler f_158644_ # blendedNoise
 public net.minecraft.world.level.levelgen.NoiseSampler f_158645_ # islandNoise
 protected net.minecraft.world.level.levelgen.NoiseSampler m_158667_(DI)D # applySlide
+public net.minecraft.world.level.levelgen.WorldGenSettings f_64606_ # legacyCustomOptions
+public net.minecraft.world.level.levelgen.WorldGenSettings <init>(JZZLnet/minecraft/core/MappedRegistry;Ljava/util/Optional;)V # constructor
+public net.minecraft.world.level.levelgen.feature.blockplacers.BlockPlacerType <init>(Lcom/mojang/serialization/Codec;)V # constructor
```
ChampionAsh5357: Was this added as a comment from someone on discord? Otherwise, it's a bit out of scope for this PR.

Commoble: Think that was a rebasing flub, I'll clean that up

ChampionAsh5357 (approved):
Should be good to go.

patches/minecraft/net/minecraft/world/level/dimension/LevelStem.java.patch
```diff
@@ -0,0 +1,24 @@
...
+    });
+    public static final ResourceKey<LevelStem> f_63971_ = ResourceKey.m_135785_(Registry.f_122820_, new ResourceLocation("overworld"));
+    public static final ResourceKey<LevelStem> f_63972_ = ResourceKey.m_135785_(Registry.f_122820_, new ResourceLocation("the_nether"));
+@@ -34,6 +_,11 @@
+    private final ChunkGenerator f_63976_;
+ 
+    public LevelStem(Supplier<DimensionType> p_63979_, ChunkGenerator p_63980_) {
++      this(p_63979_, p_63980_, false); // forge: by default, dimension jsons use the fixed seed required by the vanilla format
++   } // forge: reroute the vanilla constructor to our expanded constructor with default args
++   public final boolean useServerSeed; // forge: whether the dimension parser should use the server seed instead of the fixed seed
```
TheCurle: This should be private with a getter?

src/test/resources/data/dimension_seed_test/dimension/fixed_seed_dimension.json
```diff
@@ -0,0 +1,12 @@
+{
+  "type": "minecraft:overworld",
+  "generator": {
+    "biome_source": {
+      "preset": "minecraft:overworld",
+      "type": "minecraft:multi_noise"
+    },
+    "seed": 12345,
```
TheCurle, line 8: Is it feasible to allow this flag to be datagenerated with other dimension jsons?

TheCurle (changes requested):
Two small nitpicks. Seems good to go otherwise.

src/test/resources/data/dimension_seed_test/dimension/fixed_seed_dimension.json
Commoble: * Correct me if I'm wrong here, but as far as I can tell I can't find any builtin datagenerators for dimensions/levelstems in vanilla or forge
* Adding a dimension datagenerator to forge wouldn't be complicated but it might be outside the scope of this PR
  * Anyone can make a datagenerator for any worldgen json by skooshing together RegistryWriteOps, a worldgen codec, and DataProvider.save(), not sure whether adding a helper method for that would be within the scope of forge in general or not
* Any dimension/levelstem datagenerator that *does* leverages codecs would automatically support this flag since the flag is being added to the standard LevelStem codec

TheCurle: I couldn't confirm the existence of a dimension datagenerator, so thanks for the clarification.

patches/minecraft/net/minecraft/world/level/dimension/LevelStem.java.patch
Commoble: Fair enough, updated this in most recent push

TheCurle (approved):

PR #7958, opened 2021-07-30
[1.17.x] Add `PickaxeItem` tooltype patch

SuperMartijn642:
The `PickaxeItem` tooltype [patch from 1.16](https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/patches/minecraft/net/minecraft/item/PickaxeItem.java.patch) is currently missing.
This means any blocks that use `BlockBehaviour.Properties#setHarvestTool` with `ToolType.PICKAXE` currently don't work with vanilla pickaxes.

This PR adds back the tooltype patch.

### DIFF
--- a/patches/minecraft/net/minecraft/world/item/PickaxeItem.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/PickaxeItem.java.patch
@@ -1,0 +1,10 @@
+--- a/net/minecraft/world/item/PickaxeItem.java
++++ b/net/minecraft/world/item/PickaxeItem.java
+@@ -4,6 +4,6 @@
+ 
+   public class PickaxeItem extends DiggerItem {
+      public PickaxeItem(Tier p_42961_, int p_42962_, float p_42963_, Item.Properties p_42964_) {
+-        super((float)p_42962_, p_42963_, p_42961_, BlockTags.MINEABLE_WITH_PICKAXE, p_42964_);
++        super((float)p_42962_, p_42963_, p_42961_, BlockTags.MINEABLE_WITH_PICKAXE, p_42964_.addToolType(net.minecraftforge.common.ToolType.PICKAXE, p_42961_.m_6604_()));
+      }
+   }
--- a/patches/minecraft/net/minecraft/world/item/PickaxeItem.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/PickaxeItem.java.patch
@@ -1,10 +1,10 @@
 --- a/net/minecraft/world/item/PickaxeItem.java
 +++ b/net/minecraft/world/item/PickaxeItem.java
-@@ -4,6 +4,6 @@
+@@ -4,6 +_,6 @@
  
-   public class PickaxeItem extends DiggerItem {
-      public PickaxeItem(Tier p_42961_, int p_42962_, float p_42963_, Item.Properties p_42964_) {
--        super((float)p_42962_, p_42963_, p_42961_, BlockTags.MINEABLE_WITH_PICKAXE, p_42964_);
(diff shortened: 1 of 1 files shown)


TheCurle:
This is the second time this patch has had to be reinstated.
I find it mildly hilarious.

However, 1.17 now uses the item tags.
`minecraft:minable/pickaxe` and `minecraft:needs_<material>_tool`

So this shouldn't be as impactful to miss for now.

gigaherz:
This is outside the scope of this PR, but sooner or later we'll need to consider how to deal with cases that have both tags and tool types. If we want to have mods that work exclusively from tags be represented in the harvestTool/harvestLevel, we'll need to have some system that runs after tag reload to refresh the tool and level for each block.

SuperMartijn642:
Yup, I did mess up the patch 🤦‍♂️. It works now though.

LexManos:
Honestly, because Mojang has moved to a tag based system for determining tool levels, technically it can be an arbitrary number of levels.
So i'm more inclined to remove these hooks entirely. And use the vanilla system.
Is there any compelling reason not to?

gigaherz:
As I was discussing with others earlier, I believe we should remove the Item/Block properties values, and the respective getters, but we need to leave a means for items to customize which tags they use, in a stack sensitive way (upgradeable multi-tools, and NBT-based tools such as Tinker's Construct). 

~~Since getDestroySpeed is already stack sensitive, it would just mean adding a stack sensitive version of `isCorrectToolForDrops`, which currently only has a BlockState param.~~ Forgot about canHarvestBlock, so this point is meaningless, we already have a stack sensitive method.

We could then agree on a standard for tier tags. My suggestion would be `forge:needs_tierN_tool` to match the vanilla naming, but with a generic "tier4" instead of the name of a material, since there could be multiple mods having equivalent tiers.

EDIT: Err I pressed sendwhen I didn't mean to. Fixed the message.

LexManos:
Ya a quick skim it looks like we have all the hooks in place to allow stack sensitive tool type tag {canHarvestBlock, should probably rename to sync with vanilla} 

What we may need to do, is instead of caring about Tier#getLevel {which is only used in isCorrectToolForDrops} we may need to design a sorting system for Tier. Maybe something similiar to the mod sorter with optional/required deps?

```java
minecraft:wood ( 0,  59, 2.0F, 0.0F,   15, () -> Ingredient.of(ItemTags.PLANKS), Tags.NEEDS_WOOD_TOOL,  null)
minecraft:stone( 1, 131, 4.0F, 1.0F,    5, () -> Ingredient.of(ItemTags.STONE ), Tags.NEEDS_STONE_TOOL, "required-after: minecraft:wood")
noobcraft:clay (-1,  10, 5.0F, 5.0F, 9001, () -> Ingredient.of(ItemTags.CLAY  ), Tags.NEEDS_CLAY_TOOL,  "required-after: minecraft:wood", "required-before: minecraft:stone")
```

Would result in:
`Tier[] sorted = [minecraft:wood, noobcraft:clay, minecraft:stone]`

Then we could change  isCorrectToolForDrops to:

```java
for (int x = sorted.indexOf(this.tier) + 1; x < sorted.length; x++) {
  if (block.isIn(sorted[x].getTag())
    return false;
}
return block.isIn(this.blocks);
```

Something like that might work... Thoughts?

gigaherz:
I'd probably design a builder-style API for it, but the concept seems good to me:
```java
public static Supplier<Tier> UNOBTANIUM = ForgeItemTier.builder("mymod:unobtanium", tier values....).after(Tiers.NETHERITE).before("mod2:overpowerium").defer();
```
My thinking would be to support both explicit references and by-name references. Strings would be assumed to be optional. The returned Supplier<Tier> would become valid shortly before the registry events are fired.

SuperMartijn642:
This is obsolete as #7970 is now the way to go

PR #7960, opened 2021-07-30
Changed Monster Requirement to Mob in RangedBowAttackGoal

DavidQF555:
RangedBowAttackGoal requires the goal owner to be an instance of Monster even though none of the required methods for the goal owner are from Monster. The methods of the goal owner are all from Mob or below, so the goal owner's type can be broadened to Mob without issues in order to allow for easier use of RangedBowAttackGoal.


LexManos:
Unfortunately this is a vanilla binary break, in  a fairly commonly used class.
So you'll have to re-add the removed constructor  for this to be feasible.
Luckily tho, it's only one method.

DavidQF555:
The latest commit fixes the binary break in the RangedBowAttackGoal class.

PR #7961, opened 2021-07-30
Fixed wrong SRG name in ForgeHooks used with ObfuscationReflectionHelper

noeppi-noeppi:
Currently this throws an `ObfuscationReflectionHelper$UnableToFindFieldException` as the used SRG name existed in 1.16.5 but does no longer exist in 1.17.1.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -1336,7 +1336,7 @@ public class ForgeHooks
             if (!currentDimNames.containsAll(VANILLA_DIMS))
             {
                 LOGGER.warn("Detected missing vanilla dimensions from the world!");
-                RegistryAccess regs = ObfuscationReflectionHelper.getPrivateValue(RegistryReadOps.class, ops, "f_179860_");
+                RegistryAccess regs = ObfuscationReflectionHelper.getPrivateValue(RegistryReadOps.class, ops, "f_17986" + "0_");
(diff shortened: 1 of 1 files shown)


src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1336,7 +1336,7 @@ public static void fixNullStructureReferences(ChunkAccess chunk, Map<StructureFe
             if (!currentDimNames.containsAll(VANILLA_DIMS))
             {
                 LOGGER.warn("Detected missing vanilla dimensions from the world!");
-                RegistryAccess regs = ObfuscationReflectionHelper.getPrivateValue(RegistryReadOps.class, ops, "f_13563" + "6_");
+                RegistryAccess regs = ObfuscationReflectionHelper.getPrivateValue(RegistryReadOps.class, ops, "f_179860_");
```
coolsimulations: I would recommend you follow the established convention used throughout Forge for `ObfuscationReflectionHelper` which would mean `"f_179860_"` becomes `"f_17986" + "0_"`.

coolsimulations (changes requested):

coolsimulations (approved):

LexManos:
I just switched it to an AT considering the field is final.
I also added a timebomb to throw an error if we use this code in future MC versions as this only fixes a bug that was present for a small window of 1.16.5. So it shouldn't be necessary any more. 
(Also fixed other warnings in the ForgeHooks class since I was in there. Unrelated but simple enough to do}

PR #7964, opened 2021-07-30
[1.17] [Draft] JarInJar and Dependency management system.

marchermans:
### Introduces 
- Dependency Locator
- Changes to how a mod file determines if it is a library or a mod:
  - Anything with a set `FMLModType` attribute uses that attributes value.
  - Anything with a mods.toml is a `MOD`
  - Anything else is considered a `LIBRARY`.

### Dependency Locator functionality
Loads the `deps.txt` file from the `META-INF/dependencies` directory and parses each line for 4 pieces of information:
- The maven coordinate of the dependency.
- The supported version range of the dependency
- The included version of the dependency.
- The path in the jar where the dependency jar can be found.

The locator extracts this data recursively until all dependencies are transitively found, and then uses the specified coordinates and version ranges to determine what dependency to load. If multiple are available the first one found is loaded. Since this is based of a Hashtable, the order is not fixed and as such in practice a random variant of the available dependencies is selected and loaded.

### Requirements not included in this PR
-  Updates to the ForgeSPI, which will also get their own PR once this leaves draft stage.


marchermans:
This DraftPR supersedes #7947

PR #7967, opened 2021-07-31
[1.18.x] Move .is check with ItemStack to the Item and expand on BookItem and EnchantedBookItem functionality

xElementzx:
Move the .is(Item) Check to the Item you are checking against allowing this to be modified on a per-item basis.

Implement the .is check in SetEnchantmentsFunction where it was currently an == check seems to have been missed by Mojang

Implement a method on BookItem and EnchantedBookItem for createEnchantedBook and createUnenchantedBook for mods to be able to expand on enchanted books and have compatibility with other mods

### DIFF
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
@@ -782,4 +782,20 @@ public interface IForgeItem
         return 0;
     }
 
+    /**
+     * Used to test if this item is "sufficiently equivalent" to another item.<br>
+     * Overriding this can be used to "impersonate" a particular item for certain comparison operations.<br>
+     * <br>
+     * "Sufficiently Equivalent" implies that this item is a drop-in replacement for the "other" item, and should be able to handle
+     * all operations the "other" item would need to. This requires being a subclass or instance of the "other" item's class.<br>
+     * <br>
+     * Contractually, returning true to this method implies <pre>other.getClass().isInstance(this) == true</pre>
+     *
+     * @param stack An ItemStack containing this item.
+     * @param other The Item being checked against.
+     * @return If this item is "sufficiently equivalent" to the other item.
+     */
+    default boolean is(ItemStack stack, Item other) {
+        return self() == other;
+    }
 }
--- a/patches/minecraft/net/minecraft/world/inventory/AnvilMenu.java.patch
+++ b/patches/minecraft/net/minecraft/world/inventory/AnvilMenu.java.patch
@@ -27,7 +27,7 @@
           if (!itemstack2.m_41619_()) {
 -            boolean flag = itemstack2.m_150930_(Items.f_42690_) && !EnchantedBookItem.m_41163_(itemstack2).isEmpty();
 +            if (!net.minecraftforge.common.ForgeHooks.onAnvilChange(this, itemstack, itemstack2, f_39768_, f_39001_, j, this.f_39771_)) return;
-+            flag = itemstack2.m_41720_() == Items.f_42690_ && !EnchantedBookItem.m_41163_(itemstack2).isEmpty();
++            flag = itemstack2.m_150930_(Items.f_42690_) && !EnchantedBookItem.m_41163_(itemstack2).isEmpty();
              if (itemstack1.m_41763_() && itemstack1.m_41720_().m_6832_(itemstack, itemstack2)) {
                 int l2 = Math.min(itemstack1.m_41773_(), itemstack1.m_41776_() / 4);
                 if (l2 <= 0) {
--- a/patches/minecraft/net/minecraft/world/inventory/EnchantmentMenu.java.patch
+++ b/patches/minecraft/net/minecraft/world/inventory/EnchantmentMenu.java.patch
@@ -37,6 +37,15 @@
                 }
  
                 for(int l = 0; l < 3; ++l) {
+@@ -146,7 +_,7 @@
+                   p_39465_.m_7408_(itemstack, i);
+                   boolean flag = itemstack.m_150930_(Items.f_42517_);
+                   if (flag) {
+-                     itemstack2 = new ItemStack(Items.f_42690_);
++                     itemstack2 = ((net.minecraft.world.item.BookItem) itemstack.m_41720_()).createEnchantedBook(itemstack);
+                      CompoundTag compoundtag = itemstack.m_41783_();
+                      if (compoundtag != null) {
+                         itemstack2.m_41751_(compoundtag.m_6426_());
 @@ -235,7 +_,7 @@
              if (!this.m_38903_(itemstack1, 2, 38, true)) {
                 return ItemStack.f_41583_;
--- a/patches/minecraft/net/minecraft/world/inventory/GrindstoneMenu.java.patch
+++ b/patches/minecraft/net/minecraft/world/inventory/GrindstoneMenu.java.patch
@@ -19,3 +19,12 @@
                 if (!ItemStack.m_41728_(itemstack, itemstack1)) {
                    this.f_39559_.m_6836_(0, ItemStack.f_41583_);
                    this.m_38946_();
+@@ -201,7 +_,7 @@
+       EnchantmentHelper.m_44865_(map, itemstack);
+       itemstack.m_41742_(0);
+       if (itemstack.m_150930_(Items.f_42690_) && map.size() == 0) {
+-         itemstack = new ItemStack(Items.f_42517_);
++         itemstack = ((net.minecraft.world.item.EnchantedBookItem) itemstack.m_41720_()).createUnenchantedBook(itemstack);
+          if (p_39580_.m_41788_()) {
+             itemstack.m_41714_(p_39580_.m_41786_());
+          }
--- a/patches/minecraft/net/minecraft/world/item/BookItem.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/BookItem.java.patch
@@ -1,0 +1,11 @@
+--- a/net/minecraft/world/item/BookItem.java
++++ b/net/minecraft/world/item/BookItem.java
+@@ -12,4 +_,8 @@
+    public int m_6473_() {
+       return 1;
+    }
++
++   public ItemStack createEnchantedBook(ItemStack stack) {
++      return new ItemStack(Items.f_42690_);
++   }
+ }
--- a/patches/minecraft/net/minecraft/world/item/EnchantedBookItem.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/EnchantedBookItem.java.patch
@@ -1,0 +1,11 @@
+--- a/net/minecraft/world/item/EnchantedBookItem.java
++++ b/net/minecraft/world/item/EnchantedBookItem.java
+@@ -87,4 +_,8 @@
+       }
+ 
+    }
++
++   public ItemStack createUnenchantedBook(ItemStack stack) {
++      return new ItemStack(Items.f_42517_);
++   }
+ }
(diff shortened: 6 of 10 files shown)


ChampionAsh5357:
Giving an item the ability to say its any other item in the game will basically break all vanilla logic. If an item wants to look or be like another item, they should have their own modded item which can mimic that behavior. Another thing is that Mojang probably didn't miss that check. It affects nothing within the game so it would just be left alone.

If you could provide some clarification on the use cases of this PR, that would be great.

xElementzx:
Hi Ash,

An example is actually a mod I created in 1.16.4 that adds more enchantable books all of which have higher levels of enchantablility. Currently, I mixin into vanilla enchantment logic to replace the existing == checks with is an instance of enchantedbook or bookitem. This changes in 1.17 as the checks have moved from an == Items.BOOK to .is(Items.BOOK) , requiring a mixin also caused issues in which other mods in which enchanting tables existed were not compatible and simply wouldn't work. This is also an issue with the grindstone and anvil in similar fashions.

Vanilla functionality checks if the item is an Items.BOOK, which depending on whether the existing method is passing a tag or simply an item depends on whether it will check for a specific tag or ask the item if it is the item specified in the method. this will allow you to overwrite it on a per-item basis if you wish but shouldn't break existing vanilla functionality.

My use case for the functionality could be accomplished if there were tags for Books and Enchanted books separately and the methods were updated to check for tags, but this seemed like the more functionally useful way to achieve the result.

ChampionAsh5357:
So, why couldn't this be achieved without having to interact with the book and enchanted book system at all? It seems as though vanilla does support just adding multi-enchantable items and there are hooks to place those from your custom item to another. There isn't anything special about a book or an enchanted book. One's an item, the other is an item with a tag. Forge has hooks all relating to enchantments that would allow this behavior to work fine.

xElementzx:
Enchanting books in vanilla have all enchantments contained within a single storage tag, simply enchanting a book would give you a book with knockback for example, however vanilla stores all the enchants within a single storage tag so that you don't get the effect of the enchant on the book itself. Vanilla enchanting has logic in place specifically for handling books due to this. This is also not limited to the enchanting table, but all handling of enchants including generating dungeon chest books, Anvils and Grindstones.

ChampionAsh5357:
If that's the case then, would it not be better to add an additional method allowing a modder to specify that the enchantments should be treated as stored data and not as part of the actual item itself? I believe that would address the usecase and reduce the invasiveness of this.

brandon3055:
Literally, all xElementzx is trying to do (the end goal) is effectively extend the vanilla book. To create a custom book that can be enchanted in the same way as a vanilla book (with modified enchantability) resulting in an enchanted book that functions exactly the same as the vanilla enchanted book.

There are only two valid ways I see to do this. Switch books over to use tags which would require significant changes to vanilla and every mod that implements enchanting mechanics. Or do what xElementzx has done with this PR. 
@ChampionAsh5357 Your last suggestion could potentially work but like the tags option, it would require mods to go out of their way to implement it in their custom enchanting logic.

xElementzx:
It may solve the issue with the enchanting table, however then more issues arise from the grindstone, the anvil, and anything else that checks to see if an item is enchanted before putting it in or checks if it is specifically an enchantedbook. as far as I can tell, this shouldn't affect anything specifically in vanilla unless a mod developer does something incorrectly, I'm doing the exact same check as vanilla, just doing it elsewhere that can be overwritten by a mod developer wishing to do so on their own specific item alone . extending functionality but not breaking the existing mechanics

ChampionAsh5357:
> Your last suggestion could potentially work but like the tags option, it would require mods to go out of their way to implement it in their custom enchanting logic.

Sure, but it's a much more extensible system. Currently, this would require that all implementations must have a separate item holding the stored enchantments. It would also force the logic to require a `BookItem` and `EnchantedBookItem`. My reasoning for just another method is that it provides a simple way of declaring the enchantments on an item as stored instead of applied. Any item regardless of type could apply it, and it would solve the issue as above.

The above use case is only for a single individual for their single mod and only covers one implementation. The PR itself should be able to span multiple use cases and be beneficial to multiple users who may want a similar system yet slightly different. Currently, I think that the PR itself is limited in scope to this one case, and I want to address that to make it available for others since I do see a usage for what this PR is trying to do.

brandon3055:
The problem is everything is special-cased for `BookItem` and `EnchantedBookItem`.  with your method you could create an item with "stored enchantements" but then you would have to mixin / PR the anvil the grindstone and then convince every mod dev that does anything with enchanted books to switch over to the new system. 

Sure this is a very special case PR that not many people will use. But that does not mean it won't be useful. I see this as something like the isAir check on Block. At first glance, it does not seem all that useful and sounds like it could cause lots of issues. But when used properly it lets you do some cool things that wouldnt normally be possible. 

Anyway, that pretty much concludes my thoughts on this matter.

ChampionAsh5357:
My final thoughts on this PR after a bit of discussion with some others.

I do not see a reason to force a hard requirement on the existing book system through this current PR's implementation. Convincing mod devs is sorta a non issue since they can do whatever they want. This can be seen with people still reflecting into the flammability map instead of using the `IForgeBlock` method. They don't have to support our current method, it's just highly recommended. The point of the API forge exposes is to be extensible such that other users can create mods without the need for reflection or coremodding. Yours is not the only use case we need to consider, but others who may want something similar to yours and how they would address it. As such, the question becomes what provides the greatest extensibility which is minimally invasive and simple to maintain?

The alternative I was proposing patches about 10 or 11 places with the rest being handled through existing forge hooks compared to the above 9 patches required. While it does require a bit more work to get it similar to books, it also provides a larger extensible system to which other modders may choose to implement in the future to suffice a similar usecase. This idea probably needs a bit more analysis since it can probably be brought down even more depending on further scoping.

xElementzx:
I'm not sure if it's worth pointing out, but the functionality that we're specifically discussing here (redirecting the .is check to the item instead of the itemstack) is actually just one patch to ItemStack and 11 lines in IForgeItem, most of which are actually just documenting what it's doing.  The rest of the patches included here are for the extra functionality on BookItem and EnchantedBookItem to allow you to create an enchanted or unenchanted version of these items (Swap from minecraft:book to minecraft:enchanted_book) and actually implementing that functionality in existing places where it currently just replaces an itemstack which has a book with an itemstack that has an enchanted_book or vice versa.  But I understand that it may need to be discussed more with the team and that's perfectly okay, I just wanted to clarify that this PR is actually 3 changes, however, they do accomplish one goal.

Shadows-of-Fire:
Having implemented this feature myself in apotheosis, the quick solution would be to disallow custom EnchantedBookItems, yet provide a way for other items to transform into enchanted books when they are enchanted.  This could be done by an event.

I get the use case for new books, and they do need to be turned into enchanted books to work right.  I do not, however, see anything relevant allowing for custom enchanted books, which would likely cause all kinds of headaches across the modscape.

xElementzx:
Custom enchanted books would be used to handle grindstones appropriately, to turn a 'Diamond Enchanted Book' into a 'Diamond Book' instead of just a regular Book. that's one use case that I can think of right now. All that modders would have to do to support custom EnchantedBookItems, would be simply use the methods and changes in this PR.

xElementzx:
I'd love to rebase it but it seems like there is little-to-no reason to if a decision can't actually be made on how to implement it?

sciwhiz12:
Hello! @xElementzx, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable. If we do not hear back from you after some time, this PR will be closed for inactivity.

xElementzx:
As i said on discord,
I'd happily update my PR but I'd like to know where to go ahead with it, the discussion sort of ended from the forge team's end on how they want it implemented so until such a time that it can be decided on how it should be implemented I don't really wish to spend the time to update something for it to just go stale again.

xElementzx:
@sciwhiz12 Any update?
Again, am happy to update the PR if it can be decided on the best way to go about it. I'd still like the feature

xElementzx:
Not sure what's up with the build failure on Jenkins can someone poke it?

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -740,4 +740,15 @@ default AABB getSweepHitBox(@Nonnull ItemStack stack, @Nonnull Player player, @N
     {
         return target.getBoundingBox().inflate(1.0D, 0.25D, 1.0D);
     }
+
+    /**
+     * Used to test if this ItemStack contains the item in question.
+     *
+     * @param stack ItemStack in the Chest slot of the entity.
```
Shadows-of-Fire: Pretty sure that `param` doc is incorrect.

Shadows-of-Fire:
AnvilMenu has a hard reference to `Items.ENCHANTED_BOOK` that this PR missed `            flag = itemstack2.getItem() == Items.ENCHANTED_BOOK && !EnchantedBookItem.getEnchantments(itemstack2).isEmpty();`

Documentation on `IForgeItem#is` needs to clearly establish a contract with what returning true means - You are assuming that returning true to `is(Items.BOOK)` implies `instanceof BookItem`.  This is fine, since the previous implementation that can now be overridden was `this.getItem() == item`, but that MUST be noted on an overridable api method.

Shadows-of-Fire:
Alright, I wrote up a more comprehensive javadoc for `IForgeItem#is` that I think covers all details needed
```
   /**
    * Used to test if this item is "sufficiently equivalent" to another item.<br>
    * Overriding this can be used to "impersonate" a particular item for certain comparison operations.<br>
    * <br>
    * "Sufficiently Equivalent" implies that this item is a drop-in replacement for the "other" item, and should be able to handle
    * all operations the "other" item would need to. This requires being a subclass or instance of the "other" item's class.<br>
    * <br>
    * Contractually, returning true to this method implies <pre>other.getClass().isInstance(this) == true</pre>
    * 
    * @param stack An ItemStack containing this item.
    * @param other The Item being checked against.
    * @return If this item is "sufficiently equivalent" to the other item.
    */
```
This leaves you with three tasks to allow this to be moved forward:
- Update the JavaDoc on `IForgeItem#is` to the provided one above
- Fix the `==` check on AnvilMenu to be `ItemStack#is`
- Resolve merge conflicts

xElementzx:
Those changes have been made and pushed. Thanks for the feedback!

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
xElementzx: Updated the doc in IForgeItem

Shadows-of-Fire:
I'm going to go ahead and assign this as it is done from my standpoint.  This is not a guarantee this is accepted, as there has been some resistance to this from a design standpoint, mostly relating to the overridable `Item#is` function, which has quite a lot of side effects.

Since overriding that check implies total equivalence, it also means that items "saying" they are a book are valid as books in recipes, same for enchanted books.  It might not really be that much of a problem, but there is always the possibility for alternate design and even a more generified design in which we have the following functions:

`IForgeItem#transformsWhenEnchanted` -> Replaces `ItemStack.is(Items.BOOK)` checks, for if the item transforms when enchanted.
`IForgeItem#getEnchantedForm` -> replaces `new ItemStack(Items.ENCHANTED_BOOK)` calls.
`IForgeItem#transformsWhenDisenchanted` -> replaces `ItemStack.is(Items.ENCHANTED_BOOK)` calls for when enchanted books are disenchanted (grindstone). 
`IForgeItem#isEnchantedBook` -> replaces `ItemStack.is(Items.ENCHANTED_BOOK)` calls when transferring enchantments off of books (anvil)
`IForgeItem#getDisenchantedForm` -> replaces `new ItemStack(Items.BOOK)` when books are disenchanted.

In such a scenario `transformsWhenEnchanted` and `transformsWhenDisenchanted` could simply be omitted as well, and always calling `getEnchantedForm` / `getDisenchantedForm` would be sufficient, as they could return the original stack if no changes were made.  Such a generified design would allow for any items to be transformed on enchantment/disenchantment.

marchermans:
Currently we are internally discussing the API design here.
Untill further notice: This PR is undergoing discussion as to how to solve this.

PR #7970, opened 2021-07-31
Redesign the tool system to rely on the new vanilla tags.

gigaherz:
Introduces a TierSortingRegistry where tiers can be submitted along with before/after dependencies.

Vanilla Tiers are added automatically and available for modders to sort before or between them.

To register your tiers just call
```java
TierSortingRegistry.registerTier(yourTier, NAME, List.of(THINGS TO SORT BEFORE YOUR TIER), List.of(TIERS TO SORT AFTER YOUR TIER));
```

The before/after list can accept multiple types: a String or ResourceLocation containing the name of a tier, or a Tier object that has bee previously registered.

Example:
```java
       TierSortingRegistry.registerTier(MY_SANDSTONE_TIER, makeLoc("sandstone"), List.of(Tiers.GOLD), List.of(iron));
```

## Datapack-based loading

This redesign can now read the sorting list from a json file in datapack, from a file `data/forg/item_tier_ordering.json`.

This file is intended for modpack authors and always overwrites the full sorting list.

The format of the file is as follows:
```js
{
  "order": [ "wood", "stone", "gold", "iron", "diamond", "netherite" ]
}
```


Commoble:
String-comparison-based tool types for wood stripping etc is a bit weird, a registry of canonical ToolType instances (not enums) might make more sense

src/main/java/net/minecraftforge/common/extensions/IForgeBlockState.java
```diff
@@ -648,10 +624,10 @@ default boolean shouldDisplayFluidOverlay(BlockAndTintGetter world, BlockPos pos
      * @return The resulting state after the action has been performed
      */
     @Nullable
-    default BlockState getToolModifiedState(Level world, BlockPos pos, Player player, ItemStack stack, ToolType toolType)
+    default BlockState getToolModifiedState(Level world, BlockPos pos, Player player, ItemStack stack)
```
KnightMiner: I don't like removing the tool type argument from this. I have some tools that support performing tilling or other interactions without being a hoe for instance, and have even successfully made a tool that can both till and strip wood (if the player makes weird choices). Removing tool type also means I am now stuck relying on the order defined by blocks instead of declaring which interaction I am interested in.

src/main/java/net/minecraftforge/common/ToolType.java
```diff
@@ -1,61 +0,0 @@
...
- * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
- */
-
-package net.minecraftforge.common;
-
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.regex.Pattern;
-
-public final class ToolType
```
KnightMiner, line 26: Personally, I would rather keep using `ToolType` for the methods that you switched to a string tool type, makes comparisons a lot cleaner and gives the modder a clear idea of what vanilla tool types are supported

src/main/java/net/minecraftforge/common/ToolActions.java
```diff
@@ -0,0 +1,66 @@
...
+     */
+    public static final String WAX_OFF = "wax_off";
+
+    /**
+     *  Queried when a shovel wants to turn dirt into path
+     */
+    public static final String PATH = "path";
+
+    /**
+     *  Queried when a shovel wants to turn dirt into path
```
pupnewfster: Copy paste error in this javadoc?

pupnewfster (commented):
Would it make sense to make some of the shears interactions have a "tool action" as well instead of the semi weird state shearing currently is in?

src/main/java/net/minecraftforge/common/ToolActions.java
gigaherz: Yes, although that action isn't implemented so it probably should be removed instead.

src/main/java/net/minecraftforge/common/TierSortingRegistry.java
```diff
@@ -0,0 +1,157 @@
...
+    private static ResourceLocation processTier(Tier tier, ResourceLocation name, List<Object> afters, List<Object> befores)
+    {
+        tiers.put(name, tier);
+        for(Object after : afters)
+        {
+            if (after instanceof String s)
+                edges.put(new ResourceLocation(s), name);
+            else if (after instanceof ResourceLocation rl)
+                edges.put(rl, name);
+            else if (after instanceof Tiers t)
```
TheSilkMiner: Why is this check against `Tiers` rather than `Tier`?

src/main/java/net/minecraftforge/common/TierSortingRegistry.java
```diff
@@ -0,0 +1,157 @@
...
+                edges.put(other, name);
+            }
+        }
+        for(Object before : befores)
+        {
+            if (before instanceof String s)
+                edges.put(name, new ResourceLocation(s));
+            else if (before instanceof ResourceLocation rl)
+                edges.put(name, rl);
+            else if (before instanceof Tiers t)
```
TheSilkMiner: Why is this check against `Tiers` rather than `Tier`?

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -729,14 +687,20 @@ default boolean shouldDisplayFluidOverlay(BlockState state, BlockAndTintGetter w
...
+    default BlockState getToolModifiedState(BlockState state, Level world, BlockPos pos, Player player, ItemStack stack, String toolAction)
+    {
+        if (!stack.canPerformAction(toolAction)) return null;
+        if ("strip".equals(toolAction)) return AxeItem.getAxeStrippingState(state);
+        else if("scrape".equals(toolAction)) return WeatheringCopper.getPrevious(state).orElse(null);
+        else if("wax_off".equals(toolAction)) return Optional.ofNullable(HoneycombItem.WAX_OFF_BY_BLOCK.get().get(state.getBlock())).map((p_150694_) -> {
+            return p_150694_.withPropertiesOf(state);
+        }).orElse(null);
+        //else if("till".equals(toolAction)) return HoeItem.getHoeTillingState(state); //TODO HoeItem bork
+        else return "path".equals(toolAction) ? ShovelItem.getShovelPathingState(state) : null;
```
TheSilkMiner: Shouldn't these be constants from `ToolActions`?

src/main/java/net/minecraftforge/common/TierSortingRegistry.java
```diff
@@ -0,0 +1,157 @@
...
+{
+    public static synchronized void registerTier(Tier tier, ResourceLocation name, List<Object> after, List<Object> before)
+    {
+        if (tiers.containsKey(name))
+            throw new IllegalStateException("Duplicate tier name " + name);
+
+        processTier(tier, name, after, before);
+    }
+
+    public static List<Tier> getSortedTiers()
```
TheSilkMiner, line 95: Can this list be cached by modders? If so, when is it a safe time to do so?

src/main/java/net/minecraftforge/common/TierSortingRegistry.java
gigaherz: I wrote that logic before the redesign, when only vanilla Tiers objects were supported by instance. This slipped through. :P

src/main/java/net/minecraftforge/common/TierSortingRegistry.java
gigaherz: Ah good reminder. I meant to change it to be a final list that gets refilled as needed.

src/main/java/net/minecraftforge/common/ForgeTier.java
```diff
@@ -0,0 +1,114 @@
...
+import net.minecraft.world.item.Tier;
+import net.minecraft.world.item.crafting.Ingredient;
+import net.minecraft.world.level.block.Block;
+
+import javax.annotation.Nonnull;
+import java.util.Objects;
+import java.util.function.Supplier;
+
+/**
+ * Record containing the values for a tier
```
pupnewfster: This isn't a record (as the suppression and comment mentions)

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -906,6 +861,19 @@ public static String getDefaultWorldType()
...
+    public static Tag<Block> getTagFromVanillaTier(Tiers tier)
+    {
+        return switch(tier)
+                {
+                    case WOOD -> Tags.Blocks.NEEDS_WOOD_TOOL;
+                    case STONE -> BlockTags.NEEDS_STONE_TOOL;
+                    case IRON -> BlockTags.NEEDS_IRON_TOOL;
+                    case DIAMOND -> BlockTags.NEEDS_DIAMOND_TOOL;
+                    case NETHERITE -> Tags.Blocks.NEEDS_NETHERITE_TOOL;
+                    case GOLD -> Tags.Blocks.NEEDS_GOLD_TOOL;
```
pupnewfster: This feels a bit odd having gold at the end just given it almost makes it seem like the default is that it is a higher tier than netherite, though not really sure where it should go.

src/main/java/net/minecraftforge/common/ToolActions.java
```diff
@@ -0,0 +1,64 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.common;
+
+import java.util.List;
+
+public class ToolActions
```
pupnewfster, line 29: Should this class be similar to how ToolType was in earlier versions for if modders want to define custom actions that then there is a way to get all the actions and also could use instance equality for checking the actions.

src/main/java/net/minecraftforge/common/ForgeTier.java
gigaherz: Oops, it was! Will fix comment.

src/main/java/net/minecraftforge/common/ForgeHooks.java
gigaherz: Yeah, although vanilla itself has it in a random spot: 
image
But I'll move it.

src/main/java/net/minecraftforge/common/ToolActions.java
gigaherz: I'm open for it. I see both options as valid.

LexManos: Yes, I've already told him he should model this after tool types.
Basically making it a manual enum. Static final ToolAction fields, and a get factory.
It's basically the same as string.intern() but having defined fields is more likely to cause compile errors if you typo the action name. Or if you're making a custom one, it's only one place you need to fix not 100.

gigaherz: It is done. :P

src/main/java/net/minecraftforge/common/TierSortingRegistry.java
gigaherz: As of the latest push, the list updates happen always in either the server thread (loading/reloading), or the client thread (sync received), so it should be safe to keep a reference to the list, so long as you don't keep an iterator. :P

desht:
Will there still be a way in code to tell if a given item is of a specific tool type?  E.g. I have this (1.16.5) code here for a custom "pickaxe ingredient": https://github.com/desht/ModularRouters/blob/MC1.16.4-master/src/main/java/me/desht/modularrouters/recipe/PickaxeModuleRecipe.java#L118-L121

The intention here is that any pickaxe can be used to craft this breaker module (but the crafted item takes on the harvest level of the pickaxe that was used to craft it).

gigaherz:
That use case wasn't considered in this PR.
It could be solved by using a series of tags instead: `forge:tools/<type>` with matching string as `minecraft:minable/<type>`, the tier can be obtained from TieredItem.
It wouldn't work for NBT-based (multi)tools though. Will have to be discussed if it's important enough to warrant adding (back) a way to query tool types, just for this.

mickelus:
> It wouldn't work for NBT-based (multi)tools though. Will have to be discussed if it's important enough to warrant adding (back) a way to query tool types, just for this.

So it'll no longer be possible to have an item be able to harvest/break different blocks based on the item's nbt?

gigaherz:
Yes it will. In this implementation the tool chooses what tool it wants to be, when being used. An item that can have different modes/upgrades can choose to use different block tags / tier tags when calculating the return value of `getBreakSpeed` and `isCorrectToolForDrops`.
The only thing that currently isn't possible, is to query what kind(s) of tool a given itemstack is.

Crimix:
My 2 cents here is that someway to query what tool(s) a given itemstack is from code would be great, be it enums or tags

src/main/java/net/minecraftforge/common/ToolActions.java
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     *  Queried by blocks that can be insta-broken with a sword
+     */
+    public static final ToolAction SLICE = ToolAction.get("slice");
+
+
+    // Default actions supported by each tool type
+    public static final Set<ToolAction> DEFAULT_AXE_ACTIONS =  Stream.of(STRIP, SCRAPE, WAX_OFF).collect(Collectors.toCollection(Sets::newIdentityHashSet));
+    public static final Set<ToolAction> DEFAULT_HOE_ACTIONS = Stream.of(STRIP, SCRAPE, WAX_OFF).collect(Collectors.toCollection(Sets::newIdentityHashSet));;
+    public static final Set<ToolAction> DEFAULT_SHOVEL_ACTIONS = Stream.of(STRIP, SCRAPE, WAX_OFF).collect(Collectors.toCollection(Sets::newIdentityHashSet));;
```
pupnewfster: Copy paste error?

gigaherz: ...yes.

gigaherz:
~~As of the newest commit, there's now 2 new methods:~~

~~1. A new method that applies to all items, `Item#getToolTypeBlockTags`, returns the list of block tags the item uses to check for tool type compatibility~~
~~2. A new method that applies to TieredItems, `getTier(ItemStack, Tag<Block>)`, returns the tier for the given stack, when used to mine a block in the given tag.~~

~~The combination of both methods allows `DiggerItem#getDestroySpeed` and `IForgeTieredItem#isCorrectToolForDrops` to iterate through all the tags from the stack automatically.~~

EDIT: Obsolete. It's now just getDestroySpeed and isCorrectToolForDrops again.

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -729,14 +689,21 @@ default boolean shouldDisplayFluidOverlay(BlockState state, BlockAndTintGetter w
...
-        else return toolType == ToolType.SHOVEL ? ShovelItem.getShovelPathingState(state) : null;
+    default BlockState getToolModifiedState(BlockState state, Level world, BlockPos pos, Player player, ItemStack stack, ToolAction toolAction)
+    {
+        if (!stack.canPerformAction(toolAction)) return null;
+        if (ToolActions.STRIP.equals(toolAction)) return AxeItem.getAxeStrippingState(state);
+        else if(ToolActions.SCRAPE.equals(toolAction)) return WeatheringCopper.getPrevious(state).orElse(null);
+        else if(ToolActions.WAX_OFF.equals(toolAction)) return Optional.ofNullable(HoneycombItem.WAX_OFF_BY_BLOCK.get().get(state.getBlock())).map((p_150694_) -> {
+            return p_150694_.withPropertiesOf(state);
+        }).orElse(null);
+        //else if(ToolActions.STRIP.equals(toolAction)) return HoeItem.getHoeTillingState(state); //TODO HoeItem bork
```
KnightMiner: I know its commented out, but that says STRIP instead of TILL

gigaherz: Will be fixed on the next commit -- whenever that happens.

Crimix:
Have I missed anything in the code but is `getToolTypeBlockTags` never overriden and thus always empty list?
Also would that cover a sword?

gigaherz:
Err... See DiggerItem patch.

https://github.com/MinecraftForge/MinecraftForge/pull/7970/files#diff-a7654144f1da20b805ebf16f96e3572bca674ee0ff0b085a2dc7d4a058980ee8R33-R35

Also no it's not overriden in swords, forgot about them being able to break things. Will look into that later after work.

Crimix:
My bad, keep up the good work 👍

src/main/java/net/minecraftforge/common/ToolActions.java
```diff
@@ -0,0 +1,66 @@
...
+     *  Queried by blocks that can be insta-broken with a sword
+     */
+    public static final ToolAction SLICE = ToolAction.get("slice");
+
+
+    // Default actions supported by each tool type
+    public static final Set<ToolAction> DEFAULT_AXE_ACTIONS =  Stream.of(STRIP, SCRAPE, WAX_OFF).collect(Collectors.toCollection(Sets::newIdentityHashSet));
+    public static final Set<ToolAction> DEFAULT_HOE_ACTIONS = Set.of(); // TODO: Stream.of(TILL).collect(Collectors.toCollection(Sets::newIdentityHashSet));;
+    public static final Set<ToolAction> DEFAULT_SHOVEL_ACTIONS = Stream.of(STRIP, SCRAPE, WAX_OFF).collect(Collectors.toCollection(Sets::newIdentityHashSet));;
+    public static final Set<ToolAction> DEFAULT_SWORD_ACTIONS = Stream.of(SLICE).collect(Collectors.toCollection(Sets::newIdentityHashSet));;
```
TheSilkMiner: Are the double semicolons intended?

src/main/java/net/minecraftforge/common/ToolAction.java
```diff
@@ -0,0 +1,61 @@
...
+
+package net.minecraftforge.common;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+@SuppressWarnings("ClassCanBeRecord")
+public class ToolAction
+{
+    private static final Map<String, ToolAction> actions = new ConcurrentHashMap<>();
```
TheSilkMiner, line 30: Could it be possible to obtain a view of this map?
I'm thinking of tweaking-type mods or multitool mods that may want to provide a list of all available actions in a certain game instance so that users and pack makers can have a clearer picture and maybe reassign or change actions on items.

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -729,14 +689,21 @@ default boolean shouldDisplayFluidOverlay(BlockState state, BlockAndTintGetter w
      * @param pos The block position in world
      * @param player The player clicking the block
      * @param stack The stack being used by the player
+     * @param toolAction
```
TheSilkMiner: `@param toolAction The action that will be applied to the given state`?

src/main/java/net/minecraftforge/common/ToolActions.java
gigaherz: Nope 🤦‍♂️ copypasta.

src/main/java/net/minecraftforge/common/ToolAction.java
gigaherz: Done.

src/main/java/net/minecraftforge/common/TierSortingRegistry.java
```diff
@@ -0,0 +1,387 @@
...
+    private static final Logger LOGGER = LogManager.getLogger();
+
+    /**
+     * Registers a tier into the tier sorting registry.
+     * @param tier The tier to register
+     * @param name The name to use internally for dependency resolution
+     * @param after List of tiers to place this tier after (the tiers in the list will be considered lesser tiers)
+     * @param before List of tiers to place this tier before (the tiers in the list will be considered better tiers)
+     */
+    public static synchronized void registerTier(Tier tier, ResourceLocation name, List<Object> after, List<Object> before)
```
TheSilkMiner: Seeing the usage of this method in the test mod, would it be possible to return the passed in `Tier`? Or is that a potential problem with initialization order?

src/test/resources/data/forge/item_tier_ordering.json
```diff
@@ -0,0 +1,3 @@
+{
+  "order": [ "iron" ]
```
TheSilkMiner: Doesn't having only one of the various tiers cause issues when reading the file?

gigaherz: Err yes I tested the error condition last things before I made the commit. will fix later.

src/main/java/net/minecraftforge/common/TierSortingRegistry.java
gigaherz: 🤔 I suppose. just a `return tier;` at the end...

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -219,50 +212,46 @@ public static boolean canContinueUsing(@Nonnull ItemStack from, @Nonnull ItemSta
...
-        int toolLevel = stack.getHarvestLevel(tool, player, state);
-        if (toolLevel < 0)
-            return player.hasCorrectToolForDrops(state);
-
-        return ForgeEventFactory.doPlayerHarvestCheck(player, state, toolLevel >= state.getHarvestLevel());
+        return player.hasCorrectToolForDrops(state);
     }
 
-    public static boolean isToolEffective(LevelReader world, BlockPos pos, @Nonnull ItemStack stack)
+    public static boolean isCorrectToolForDrops(TieredItem tieredItem, ItemStack stack, BlockState state)
```
KnightMiner: Can this be changed to `IForgeTieredItem` instead of `TieredItem`? Would be easier for extension as you could implement that interface on a different class (and the one method called is from `IForgeTieredItem`)

The interface has just two methods, the one that assumes its a `TieredItem` could easily be overridden, and it would be nice to not have to reimplement this hook.

gigaherz: Is there a real use case for having getTier() in a class that isn't TieredItem? If so I'd move the method to IForgeItem so any item can implement it.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -220,37 +220,40 @@ public static boolean isCorrectToolForDrops(@Nonnull BlockState state, @Nonnull
         return player.hasCorrectToolForDrops(state);
     }
 
-    public static boolean isCorrectToolForDrops(TieredItem tieredItem, ItemStack stack, BlockState state)
+    public static boolean isCorrectToolForDrops(ItemStack stack, BlockState state)
```
KnightMiner: I think this should be named `isCorrectTierForDrops` as its just doing tier checks. Gives the modder a clue that this alone is an incomplete check (also need tag check)

gigaherz:
As of the newest commit, the system works like this:
* ItemStack#isCorrectToolForDrops(BlockState): decides if the item is allowed to get loot by breaking the block. if false, the block will still break slowly.
* ItemStack#getDestroySpeed(BlockState): returns how fast the block is mined. note that if the tool is not correct, it will mine slower, regardless of the returned value from this method.
* ItemStack#canPerformAction(ToolAction): queries if a tool is capable of performing a certain action, such as "dig like an axe", see ToolActions for the stock actions. This is informational, and the returned value is not used to decide if the block can break.

Overriding these methods is necessary for custom tools that support more than one tool "type" at the same time.

KnightMiner (approved):

TheSilkMiner (approved):

PR #7973, opened 2021-08-01
Track the active mod within DeferredWorkQueue

SquidDev:
When working with non-`ForgeRegistry` registries (such as `ArgumentType`), one needs to be careful to only interact with them on the main thread. To do this, mods either piggyback on another Registry event or use the `DeferredWorkQueue`.

My mod exposes an API which allows other mods to register upgrades. We use `ModLoadingContext.get().getActiveContainer()` to determine which mod registered an upgrade.

Unfortunately, using `DeferredWorkQueue` means you lose information about which mod tried to register something. This patch changes the queue to set the current mod container when running each task.

I'm aware this patch is probably a little contentious, and does involve several minor breaking changes (hence targetting 1.17). I just felt it was sufficiently useful that it was worth submitting.

### DIFF
--- a/fmlcore/src/main/java/net/minecraftforge/fml/DeferredWorkQueue.java
+++ b/fmlcore/src/main/java/net/minecraftforge/fml/DeferredWorkQueue.java
@@ -34,8 +34,6 @@ import com.google.common.base.Stopwatch;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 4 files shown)


LexManos:
This isn't bad, its a breaking change which is never good. But the vast majority of people should be using event.enqueWork and not the  queue directly. So that should be fine.
Ideally we'd make that utility class package private to  lower our  API  footprint. But that is a job for the general cleanup.

PR #7976, opened 2021-08-02
Buildscript fixups and ClasspathLocator-related fixes

### LINKED ISSUE #7957: [1.17.1] ClasspathLocator fails to validate manifests
**Minecraft Version:** 1.17.1

**Forge Version:** 1.17.1-37.0.13

**Logs:** [latest.log](https://paste.labymod.net/totimocuhi.sql) (Note, see below for identified issue, the log is only added for completeness)

**Related to:** https://github.com/MinecraftForge/MinecraftForge/commit/7d4c23816590087e0baa995425a24051b98a1fae

**Steps to Reproduce:**
 1. Create an empty gradle project
 2. Add a forge mod as a gradle subproject
 3. Add a java library as a gradle subproject
 4. Configure gradle to add `FMLModType: LIBRARY` to the manifest of the java library
 5. Add an `implementation` dependency
 6. Add any class to the library
 7. Reference the class from your mod in some start event handler (or statically, doesn't matter)
 8. Observe a `ClassNotFoundException`  

**Description of issue:**
This seems to be an issue which occurred while porting over the `ClasspathLocator` from Forge 1.16. 
In Line `97` `sj.getRootPath()` is called, returning `/` on Unix systems (and I would guess `<drive letter>:\` on NT systems). This ends up in an `IOException`, as the root directory is not a jar file.

https://github.com/MinecraftForge/MinecraftForge/blob/7d4c23816590087e0baa995425a24051b98a1fae/fmlloader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ClasspathLocator.java#L97

**Proposed fix:**
From what I can tell, the wanted method seems to be `SecureJar#getPrimaryPath()`. This (correctly) returns the full path of the Jar file.

SizableShrimp:
This PR utilizes the new `inheritArgs` property in run configs that simplifies the `forge_test_data` run config.
ClasspathLocator's isValidManifest function has been fixed to use the right path from SJH, this fixes #7957. MOD_CLASSES is now reversed when parsed to deal with SJH currently only checking the last base path for a manifest. Since MOD_CLASSES always passes in with `resources;classes`, this is necessary so that manifests can be loaded in dev.


fmlloader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ClasspathLocator.java
```diff
@@ -94,7 +94,7 @@ public void initArguments(Map<String, ?> arguments) {
     }
 
     private static boolean isValidManifest(SecureJar sj) {
-        try (var jis = new JarInputStream(Files.newInputStream(sj.getRootPath()))) {
+        try (var jis = new JarInputStream(Files.newInputStream(sj.getPrimaryPath()))) {
```
marchermans: This won't suffice for normal libraries.
The way CPL creates ModFiles requires a mods.toml to be available, so that they are loaded in the correct module.

See my other two PRs on this subject on how to fix that for the libraries.

marchermans (changes requested):

fmlloader/src/main/java/net/minecraftforge/fml/loading/targets/CommonLaunchHandler.java
```diff
@@ -72,6 +73,12 @@ record ExplodedModPath(String modid, Path path) {}
                 .map(splitString -> new ExplodedModPath(splitString.length == 1 ? "defaultmodid" : splitString[0], Paths.get(splitString[splitString.length - 1])))
                 .collect(Collectors.groupingBy(ExplodedModPath::modid, Collectors.mapping(ExplodedModPath::path, Collectors.toList())));
 
+        // We need to REVERSE the passed in mod classes so that SJH takes it properly.
+        // The last path passed into SecureJar#from is the "main" path used to grab the manifest.
+        // ForgeGradle always passed in resources,classes with the expectation in 1.16 or lower that the resources would be checked first.
+        // So in 1.17+ FML/SJH, we need to reverse it.
+        modClassPaths.forEach((k, v) -> Collections.reverse(v));
```
LexManos: So the problem isn't that things are in the wrong order its that things can quite often be in ANY order.
So the best option is the trust the order that exists, but put the one that specified the manifest files at the first one.
So instead of reverse. Sort.
Would a possible option be to copy over the manifest to every stream thats opened?
Or, since this is a read only file system. What we could do is make a fake input stream provider that sorts the entries/lists all files as if they were a single zip {merge together the zip headers}.

SizableShrimp: This is removed/fixed as of [`6d9030b` (#7976)](https://github.com/MinecraftForge/MinecraftForge/pull/7976/commits/6d9030be92921e297c3859cf7145ed99f904fd4a) because SJH 0.9.51 and up properly checks ALL paths to find a manifest instead of just the last one.

LexManos:
As stated on discord, I am not a fan of you packaging gradle  spam and 'fixups' that arnt really fixups. It  makes reviewing your PRs a lot more annoying. This however is pending the fixes for the ignore list as discussed on discord.

build.gradle
```diff
@@ -392,7 +391,7 @@ def sharedFmlonlyForge = { Project prj ->
         }
 
         // SecureJarHandler bootstrap values.
-        run.property 'ignoreList', prj.configurations.moduleonly.files.collect {it.name.replaceAll(/([-_]([.\d]*\d+)|\.jar$)/, '') }.join(',') + ",client-extra,fmlcore,javafmllanguage,mclanguage,${prj.name}-"
+        run.property 'ignoreList', prj.configurations.moduleonly.files.collect {it.name.replaceAll(/([-_]([.\d]*\d+)|\.jar$)/, '') }.join(',') + ",client-extra,fmlcore,javafmllanguage,mclanguage,${prj.name}-${prj.version},${prj.rootProject.ext.MC_VERSION}-${prj.name}${prj.version.substring(prj.rootProject.ext.MC_VERSION.length())}"
```
LexManos, line 394: This is still a no.
Nothing in our environment is ever named MC_VERSION-forge-VERSION. So its just a complex filter for no reason.

LexManos:
Got tired of waiting on shrimp to finish this, so implemented a quick pass myself.
I do not have any test mods for this, so anyone who was running into this issue please report back with your results.
Note the introduction of GAMELIBRARY which adds things to the same classloader as mods {can access MC code} vs LIBRARY which adds them to the PLUGIN layer, which can't.

PR #7978, opened 2021-08-03
Ensure that all render targets use the same format

malte0811:
The problem was that in Fabulous mode an additional `TextureTarget` target is used, which was not affected by enabling stencil. Calling `copyDepthFrom` with that buffer then caused the errors from the issue. With this change all `RenderTarget`s always use the same format, so this situation can no longer occur.

### DIFF
--- a/patches/minecraft/com/mojang/blaze3d/pipeline/RenderTarget.java.patch
+++ b/patches/minecraft/com/mojang/blaze3d/pipeline/RenderTarget.java.patch
@@ -1,5 +1,29 @@
 --- a/com/mojang/blaze3d/pipeline/RenderTarget.java
 +++ b/com/mojang/blaze3d/pipeline/RenderTarget.java
+@@ -18,6 +_,7 @@
+ 
+ @OnlyIn(Dist.CLIENT)
+ public abstract class RenderTarget {
++   private static final java.util.Set<RenderTarget> ACTIVE_TARGETS = new java.util.HashSet<>();
+    private static final int f_166194_ = 0;
+    private static final int f_166195_ = 1;
+    private static final int f_166196_ = 2;
+@@ -66,6 +_,7 @@
+ 
+    public void m_83930_() {
+       RenderSystem.m_69393_(RenderSystem::m_69587_);
++      ACTIVE_TARGETS.remove(this);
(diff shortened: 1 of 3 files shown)


gigaherz:
I'm not a fan of the implication that all render targets should use the same exact depth/stencil format. We should probably consider some way to link together render targets that should have the same settings. Maybe adding a RenderTargetGroup API, where all vanilla RTs default to the same one, or maybe allowing RTs to have children, with the vanilla RTs defaulting to be children of the main RT.

malte0811:
Do you have any specific use-case in mind with that? I considered making it less global, but decided that there wasn't any real advantage to it. I'm not an OpenGL expert by any stretch, but given that there are only two formats available in the first place (depth and depth+stencil) I don't think there would be that much use in having "mixed" formats.

gigaherz:
No, I do not myself have an use case, but 1.17 has greatly expanded the use of shaders in the game, and as such, a lot of people will be interested in trying to do custom rendering things. In the game, there's only 3 places where the game calls copyDepthFrom, and it's the only places where the game calls _glBlitFrameBuffer. So it seems unnecessary to me to have every single RT ever created use the same format, when we can do just a little bit more work and do it right.

After looking at the code, all these calls are run from objects created by a PostChain, which already has a Map<String, RenderTarget>, and a reference to the main RT of the game. Since it creates its own RTs in `addTempTarget`, it seems like the right place to set the format.
```patch
    public void addTempTarget(String p_110039_, int p_110040_, int p_110041_) {
       RenderTarget rendertarget = new TextureTarget(p_110040_, p_110041_, true, Minecraft.ON_OSX);
       rendertarget.setClearColor(0.0F, 0.0F, 0.0F, 0.0F);
+      if (screenTarget.isStencilEnabled()) rendertarget.enableStencil();
```

That alone should fix the issues, and not affect any other RT. No turning things static, no global list.

EDIT: Forgot to say that, since all those post chains are created either from resource reload, or later, there's places already that have the right kind of timing to enable the stencil without running into issues, such as the mod construct event (via enqueueWork), RegisterClientReloadListenersEvent, or EntityRenderersEvent.RegisterRenderers.

gigaherz:
LGTM now. :)

PR #7979, opened 2021-08-03
Added event to register menu screens with appropriate timing

HyCraftHD:
Adds ``RegisterMenuScreensEvent`` event which registers a screen for a menu type with appropriate timing.

Usage:
```java
	@SubscribeEvent
	public static void registerMenuScreen(RegisterMenuScreensEvent event)
	{
		event.register(MY_MENU_TYPE, MyMenuScreen::new);
	}
```


ChampionAsh5357:
So, I'm not sure whether a timing needs to be present for this event. Unlike, BERs or ERs, there's not really a specific timing that's needed for screens as they can be registered at any point. They can't be used until the server and a player exists which is any time during the mod loading phase or even slightly after. BERs and ERs on the other hand are loaded on initial resource reload which is before the mod loading phase has finished constructing.

If we wanted to do something for this, it would be better to move all non-thread safe register methods to a generic mod loading event and register it all in there. Whether that would be valid would leave up on what should be registered during what phase of mod loading.

HyCraftHD:
You're right, have not thought about that.

Issue #7933, opened 2021-07-26
[1.17.1] forge:nbt ingredient fails to sync to the client


Mrbysco:
**Minecraft Version:** 1.17.1

**Forge Version:** 37.0.10

**Logs:** 
[latest.log](https://gist.github.com/Mrbysco/4e227fe2ea59f6d055b3d9b1c132131d)
[nbt_datapack.zip](https://github.com/MinecraftForge/MinecraftForge/files/6878764/nbt_datapack.zip)

**Steps to Reproduce:**
 1. Add a recipe to a mod that requires the `forge:nbt` ingredient
 2. Start a server and a client
 3. Connect the client to the server
 4. Observe how the client gets denied with a DecoderException

**Description of issue:**
The `forge:nbt` ingredient fails to sync to the client when connecting to a server.

I've provided a datapack to reproduce the issue.

GeheimagentNr1:
I had the same issue and found, that there is a problem in the fromNetwork/toNetwork methods of the Ingredient class:
This is the 1.17.1 version of the Ingredient#toNetwork methods:
```java
   public final void toNetwork(FriendlyByteBuf p_43924_) {
      this.dissolve();
      if (!this.isVanilla()) {
         net.minecraftforge.common.crafting.CraftingHelper.write(p_43924_, this);
         return;
      }
      p_43924_.writeCollection(Arrays.asList(this.itemStacks), FriendlyByteBuf::writeItem);
   }
```
This is the 1.17.1 version of the Ingredient#fromNetwork method:
```java
   public static Ingredient fromNetwork(FriendlyByteBuf p_43941_) {
      return fromValues(p_43941_.readList(FriendlyByteBuf::readItem).stream().map(Ingredient.ItemValue::new));
   }
```
In version 1.16.5 the Ingredient#fromNetwork method looked like that:
```java
   public static Ingredient fromNetwork(PacketBuffer p_199566_0_) {
      int i = p_199566_0_.readVarInt();
      if (i == -1) {
         return net.minecraftforge.common.crafting.CraftingHelper.getIngredient(p_199566_0_.readResourceLocation(), p_199566_0_);
      }
      return fromValues(Stream.generate(() -> {
         return new Ingredient.SingleItemList(p_199566_0_.readItem());
      }).limit((long)i));
   }
```
The handling for custom Ingredient is missing in the 1.17.1 version of the Ingredient#fromNetwork method.

[...]

PR #7980, opened 2021-08-05
[1.17.1] fix forge:nbt ingredient fails to sync to the client

Sirttas:
This PR aim at fixing deserialization of forge custom ingredients (#7933) while keeping network compatibility with vanilla.

### DIFF
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeFriendlyByteBuf.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeFriendlyByteBuf.java
@@ -19,14 +19,6 @@
 
 package net.minecraftforge.common.extensions;
 
-import java.util.Collection;
-import java.util.List;
-import java.util.Objects;
-import java.util.function.Function;
-import java.util.function.IntFunction;
-
-import javax.annotation.Nonnull;
-
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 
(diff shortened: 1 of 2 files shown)


src/main/java/net/minecraftforge/common/extensions/IForgeFriendlyByteBuf.java
```diff
@@ -152,4 +158,25 @@ default FluidStack readFluidStack()
...
+        }
+        return c;
+    }
+
+    /**
+     * Read a List from this buffer with set size.
+     */
+    default <T> List<T> readListWithSize(int size, Function<FriendlyByteBuf, T> builder)
+    {
+        return this.readCollectionWithSize(size, Lists::newArrayListWithCapacity, builder);
```
alcatrazEscapee: Can I point out, the following javadoc on `Lists#newArrayListWithCapacity`
<b>Note for Java 7 and later:</b> this method is now unnecessary and should be treated as deprecated. Instead, use `new ArrayList(int)` directly, taking advantage of the new <a href="http://goo.gl/iz2Wi">"diamond" syntax</a>.

*so... why write `Lists::newArrayListWithCapacity` when you could write `ArrayList::new`...*

Sirttas: This is what vanilla used for `FriendlyByteBuf#readList` I kept it the same for consistency.

LexManos: Whats wrong with doing the limit that we did in the old patch and not needing any of this at all?

Sirttas: TBH I didn't checked what the 1.16 patch was. I thought this patch was omitted because of vanilla changes.

LexManos: Always check, the people who update this are, unfortunately, human. And things sometimes get missed.

Sirttas: So I checked and the limit was vanilla behavior that was changed to use the new `FriendlyByteBuf#readList`. It should still works though.

Sirttas:
I revert back to the old 1.16.5 behavior with a limit.

PR #7981, opened 2021-08-05
[1.17.1] Add custom attack sweep

Sirttas:
This PR adds `ItemStack#getSweepHitBox` and  `Item#getSweepHitBox`. This enable mods to create custom sweep box for their items.
theses methods return a `net.minecraft.world.phys.AABB` to offers more customisation over the current sword only sweep behavior.

It also makes sure swords keeps their own sweep.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
@@ -209,7 +209,7 @@
                 i = i + EnchantmentHelper.m_44894_(this);
                 if (this.m_20142_() && flag) {
                    this.f_19853_.m_6263_((Player)null, this.m_20185_(), this.m_20186_(), this.m_20189_(), SoundEvents.f_12314_, this.m_5720_(), 1.0F, 1.0F);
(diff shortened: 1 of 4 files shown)


loordgek:
you can omit the SwordItem patch by doing 
`@Nullable
    default AABB getSwingAABB(@Nonnull ItemStack stack, @Nonnull Player player, @Nonnull Entity target)
    {
        if (stack.getItem() instanceof SwordItem)
            return target.m_142469_().m_82377_(1.0D, 0.25D, 1.0D);
        return null;
    }`

Cyborgmas:
You can also reduce the patch size by keeping flag3 and setting it to `swordAABB != null`

Sirttas:
I called it swing in first commit because of dislexia, I renamed the new method to `getSweepHitBox` I think it will be easier to understand what it does.

Also I reduced patch size

Sirttas:
I made the changes to use tool actions. I added a new toll action called `sword_sweep`. I also made the sword sweep hitbox the default one since the tool action now take care of determining if the hit must be a sweep.

PR #7983, opened 2021-08-06
It's intended!

BalaM314:
Replace
`This could be an intended override, but in most cases indicates a broken mod.`
with
`This could be a broken mod, but in most cases indicates an intended override.`
I've seen this message over ten thousand times, and it has never actually been a sign of a broken mod. This has also confused some of my friends who were new to modding.
You may also want to consider suppressing this warning, as on most modpacks(tested on Omnifactory) this one message takes up a whopping **37%** of the log. I couldn't find where that would be handled.


50ap5ud5:
1.12 has been past its support period for quite some time, please retarget this PR for the 1.17.x branch if you would like it to be considered for merging.

When you are retargeting for 1.17.x, please take the time to find the root cause of the large log file size and prove your solution fixes the issue.

If you are simply looking to change the text for some of the logging messages, these changes would be best done in an omnibus PR that bulk updates similar text issues.

diesieben07:
"Please reduce the spam caused by broken mods" is not a valid argument.
No, it is not intended. If this message happens, the mod is indeed broken in most cases.

PR #7994, opened 2021-08-09
Use longs instead of ints for IEnergyStorage

RogueLogix:
Switches IEnergyStorage to using longs instead of ints.

While it is still somewhat an edge case, there are cases of more than int max being used, and there are multiple extensions/implementations (or additional interfaces) to IEnergyStorage to allow the use of longs (or something very similar) in 1.16.x

Known mods with extensions:
[Phosphophyllite](https://github.com/BiggerSeries/Phosphophyllite/blob/8e037ed9260569ad22cce5c8466fa909da250753/src/main/java/net/roguelogix/phosphophyllite/energy/IPhosphophylliteEnergyStorage.java) (Bigger Reactors)
[BrandonsCore](https://github.com/brandon3055/BrandonsCore/blob/9c464f1dc0ea587bb2625f15c4bb659df2d8ad9b/src/main/java/com/brandon3055/brandonscore/api/power/IOPStorage.java) (Draconic Evolution)
[Industrial Foregoing](https://github.com/InnovativeOnlineIndustries/Industrial-Foregoing/blob/bdee4b2aba5c327b676d837933113e7297b9e437/src/main/java/com/buuz135/industrial/item/infinity/InfinityEnergyStorage.java)
[Mekanism](https://github.com/mekanism/Mekanism/blob/c5a752ff7d122799459ef53a8b6a6a263e212ec5/src/api/java/mekanism/api/energy/IStrictEnergyHandler.java)
[Flux Networks](https://github.com/SonarSonic/Flux-Networks/blob/45379d14358d95a46bc4ead6b903488686605ff2/src/main/java/sonar/fluxnetworks/api/energy/IFNEnergyStorage.java)
More, probably

### DIFF
--- a/src/main/java/net/minecraftforge/energy/EnergyStorage.java
+++ b/src/main/java/net/minecraftforge/energy/EnergyStorage.java
@@ -19,7 +19,7 @@
 
 package net.minecraftforge.energy;
 
-import net.minecraft.nbt.IntTag;
+import net.minecraft.nbt.LongTag;
 import net.minecraft.nbt.Tag;
 import net.minecraftforge.common.util.INBTSerializable;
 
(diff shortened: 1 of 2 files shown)


LexManos:
This has been discussed before. It's not gunna happen. There is no sane reason to use longs.
Anyone who does just wants to feel special with large numbers.
They can have their own API.
Nothing stops the two from working fine together.

PR #7997, opened 2021-08-11
[1.17] Add Shears ToolActions

cech12:
This PR does the same as https://github.com/MinecraftForge/MinecraftForge/pull/7544 but for 1.17.

"Makes it so less code has to be replicated in onItemUse checking the block used against to make custom shears function properly with some of the custom interactions vanilla has."


KnightMiner:
Instead of instance of checks, it might be good to use the new `ToolAction`s from Gigas tool PR. Means you don't have to extend the shears class to get full shearing behavior, e.g. you want to add shearing to a `DiggerItem`

pupnewfster:
Personally I agree it probably makes more sense to use the new ToolAction stuff, but if that doesn't end up being the way things go, these changes should be done with forge's field_to_instanceof transformer that the shears instance stuff got changed to at some point after #7544 in 1.16 rather than with how this PR is doing it.

cech12:
Ok I will look into the new ToolAction stuff. It looks interesting :)

cech12:
I've adapted the `instanceof` checks to the new `ToolAction` approach.
The check in the `DiggingEnchantment` is not a `ToolAction` in my opinion, so I left it that way .

src/main/java/net/minecraftforge/common/ToolActions.java
```diff
@@ -53,6 +53,11 @@
      */
     public static final ToolAction SWORD_DIG = ToolAction.get("sword_dig");
 
+    /**
+     *  Exposed by swords to allow querying tool behaviours
+     */
+    public static final ToolAction SHEARS_DIG = ToolAction.get("shears_dig");
```
gigaherz, line 60: copypasta in the comment

pupnewfster:
It would also probably be wise to make a custom loot condition and use that instead of the shears tag in the [vanilla loot table replacer](https://github.com/MinecraftForge/MinecraftForge/blob/1.17.x/src/main/java/net/minecraftforge/common/data/ForgeLootTableProvider.java) so that the custom shears can also be used to harvest blocks properly even if they may be NBT sensitive.

src/main/java/net/minecraftforge/common/ToolActions.java
cech12: Thanks!^^

cech12:
Perhaps in this context it also makes sense to replace the `onSheard` method of entities (Sheep, Mooshroom, Snow Golem) with a new `ToolAction`?

This would be a breaking change but would enhance the consistency.

What do you think?

cech12:
> It would also probably be wise to make a custom loot condition and use that instead of the shears tag in the [vanilla loot table replacer](https://github.com/MinecraftForge/MinecraftForge/blob/1.17.x/src/main/java/net/minecraftforge/common/data/ForgeLootTableProvider.java) so that the custom shears can also be used to harvest blocks properly even if they may be NBT sensitive.

Then the `shears` tag could be removed because if a custom loot condition related to the new `shears_dig` `ToolAction` is added it would no longer be in use.

gigaherz:
If a loot condition is added, it would probably be nice to make it a generic `"forge:can_perform_action"` where you can query `"action":"shears_dig"` but also any other action. (I don't really have an opinion on whether it should be added, I leave that up to the interested parties to decide)

cech12:
I added a generic `"forge:can_tool_perform_action"` loot condition, where you can define a tool action via `"action":"shears_dig"`.

It replaces the `"forge:shears"` tag and is added to the vanilla loot tables by the ForgeLootTableProvider.

patches/minecraft/net/minecraft/world/item/enchantment/DiggingEnchantment.java.patch
```diff
@@ -0,0 +1,10 @@
+--- a/net/minecraft/world/item/enchantment/DiggingEnchantment.java
++++ b/net/minecraft/world/item/enchantment/DiggingEnchantment.java
+@@ -22,6 +_,6 @@
+    }
+ 
+    public boolean m_6081_(ItemStack p_44668_) {
+-      return p_44668_.m_150930_(Items.f_42574_) ? true : super.m_6081_(p_44668_);
++      return p_44668_.m_41720_() instanceof net.minecraft.world.item.ShearsItem ? true : super.m_6081_(p_44668_);
```
KnightMiner, line 8: Is this intentionally still `instanceof`? I suppose a non-shears class could always override the can be enchanted method

cech12: This change is carried over from the original PR 7544. But actually you are right that the "can be enchanted" method could be overridden by every modded shears item. I can reverse this change here.

KnightMiner: I assume most modders would expect making their own instance of the shears class would be enchantable automatically. Might be nice if having the tool action allowed it as well. That said, we could save a patch by doing neither, so its not really my call.

src/main/java/net/minecraftforge/common/loot/CanToolPerformAction.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public static class Serializer implements net.minecraft.world.level.storage.loot.Serializer<CanToolPerformAction> {
+        public void serialize(JsonObject json, CanToolPerformAction itemCondition, @Nonnull JsonSerializationContext context) {
+            json.addProperty("action", itemCondition.action.name());
+        }
+
+        @Nonnull
+        public CanToolPerformAction deserialize(JsonObject json, @Nonnull JsonDeserializationContext context) {
+            return new CanToolPerformAction(ToolAction.get(json.get("action").getAsString()));
```
KnightMiner, line 57: Is it worth having an JSON parsing error here for "ToolAction does not exist"? If the tool action is only "defined" via parsing the JSON, then the condition will never match, so an error might be nice to help with typos. I have implemented a similar system in one of my mods, in tool action terms it would work out to either `ToolAction.isDefined` or a nullable `ToolAction.getIfPresent`.

Only downside I can imagine to such a check is it prevents detecting a tool action that is dynamically defined after loot tables are parsed (e.g. defined via datapacks), but I cannot imagine why you would need that.

patches/minecraft/net/minecraft/world/item/enchantment/DiggingEnchantment.java.patch
cech12: I also think modders would expect that. So I left the patch in there. A ToolAction would be kind of strange in this case, since it is not really an action. It would be called "shears_digging_echantable" (or similar) and it would not be good for anything other than this case.

src/main/java/net/minecraftforge/common/loot/CanToolPerformAction.java
cech12: I've already tried something like this. Unfortunately the ToolActions are not yet defined when the LootTables are loaded. So, I didn't include this here. 

I used the `Toolaction.getActions()` method, but the result was an empty list. :/

But an alternative could be that the name of the ToolAction instead of the ToolAction itself is saved in the LootContition. During the test method it could be compared against the existing ToolActions. If none exist, a warning can be logged.

KnightMiner: Aren't tool actions statically initialized? Means that they are not initialized because nothing referenced the `ToolActions` class. I suppose we could initialize it earlier, but there is no reason a mod would have to do so for its tools, meaning an error from undefined actions might cause more false positives than useful errors.

cech12: Yes they are statically initialized. Maybe a registry could fix that? But including an extra registry just to display errors... Isn't that a bit exaggerated? ^^

KnightMiner: Making it a registry is not needed, just ensuring it's initialized before datapack load. That kinda violates the simple design of the tool actions as you now have to tell people when to get them

So, probably best to rely on datagen to minimize errors

cech12: Do you have an idea how to ensure that it's initialized before datapack load?

gigaherz: You can't, not really. It would require turning it into a registry-like setup with an event where mods are required to register the ToolActions. 
IMO, the best thing we can do is to just not have a warning. While it would be nice, it would make the system a lot more cumbersome.

cech12: Ok, then I would just leave the part as it is.

pupnewfster:
I will do a bit of testing later to confirm things work as expected given codewise I think this looks rather good now, though there is one file that is conflicting and will need to be fixed so that it will be able to be assigned and then merged.

cech12:
Merge conflict is fixed. :)

src/main/java/net/minecraftforge/common/Tags.java
```diff
@@ -276,7 +276,6 @@ private static void init(){}
         public static final IOptionalNamedTag<Item> SEEDS_MELON = tag("seeds/melon");
         public static final IOptionalNamedTag<Item> SEEDS_PUMPKIN = tag("seeds/pumpkin");
         public static final IOptionalNamedTag<Item> SEEDS_WHEAT = tag("seeds/wheat");
-        public static final IOptionalNamedTag<Item> SHEARS = tag("shears");
```
LexManos: This shouldn't be removed, as it is still useful for recipes. Also, binary breaks are bad mmkay.

src/main/java/net/minecraftforge/common/ToolActions.java
```diff
@@ -80,7 +85,22 @@
      *  @see {@link IForgeItem#getSweepHitBox}
      */
     public static final ToolAction SWORD_SWEEP = ToolAction.get("sword_sweep");
-    
+
+    /**
+     *  Exposed by shears to allow harvesting tool behaviours (example: harvesting beehives)
+     */
+    public static final ToolAction SHEARS_HARVEST = ToolAction.get("shears_harvest");
+
```
LexManos, line 94: This is a little confusing, you have this named harvest but use 'dig' in all the loot tables.
These need better docs or something to describe that is what

src/main/java/net/minecraftforge/common/ToolActions.java
```diff
@@ -93,4 +113,5 @@
     public static final Set<ToolAction> DEFAULT_SHOVEL_ACTIONS = Stream.of(SHOVEL_DIG, SHOVEL_FLATTEN).collect(Collectors.toCollection(Sets::newIdentityHashSet));
     public static final Set<ToolAction> DEFAULT_PICKAXE_ACTIONS = Stream.of(PICKAXE_DIG).collect(Collectors.toCollection(Sets::newIdentityHashSet));
     public static final Set<ToolAction> DEFAULT_SWORD_ACTIONS = Stream.of(SWORD_DIG, SWORD_SWEEP).collect(Collectors.toCollection(Sets::newIdentityHashSet));
+    public static final Set<ToolAction> DEFAULT_SHEARS_ACTIONS = Stream.of(SHEARS_DIG, SHEARS_HARVEST, SHEARS_CARVE, SHEARS_DISARM).collect(Collectors.toCollection(Sets::newIdentityHashSet));
```
LexManos: you know we should probaby make a helper:
```
private static Set<ToolAction> of(ToolAction... actions) {
    return Stream.of(actions).collect(Collectors.toCollection(Sets:newIdentityHashSet));
  }
```
  To make these inits less spammy.

src/main/java/net/minecraftforge/common/Tags.java
KnightMiner: Would a tool action ingredient be useful? Be trivial to make an ingredient that matches if the stack has a tool action

LexManos: I don't want to add the concept of using tool actions in recipes.

src/main/java/net/minecraftforge/common/ToolActions.java
cech12: That is a good idea. :)

src/main/java/net/minecraftforge/common/ToolActions.java
cech12: The "dig" action is used for breaking blocks. The "harvest" action is used for blocks with a right click behaviour that allows a harvest operation for shears (like beehives with honey level 5)
I will extend the desciption. :)

src/main/java/net/minecraftforge/common/Tags.java
cech12: Okay, I revert the shears tag removement

src/main/java/net/minecraftforge/common/ToolActions.java
cech12: I hope it's a little clearer now :)

gigaherz: Maybe we should be more specific and say "right-click harvesting" instead of just "harvesting"

cech12: > This action is exposed by shears and corresponds to a harvest action that is triggered with a right click on a block that supports such behaviour. Example: Right click with shears on a beehive with honey level 5 to harvest it

This is the new text. Or do you mean to rername the action?

gigaherz: Oh, ignore me. I was looking idly while working and didn't notice I was seeing the *old* code.

PR #8003, opened 2021-08-13
Add Z-rotation to model variant

RobMayer:


SuperMartijn642:
Z-rotation can already be accomplished by rotating 90 degrees around the y-axis and then rotation around the x-axis.

RobMayer:
That's not strictly true. You can get approximates, but a full descriptive set of orientations requires 4 models and use of X and Y rotations, under current conditions where a single block has 6 different textures (one for each face).

here is an example where that is required - and cannot be reduced to a single Model json.
https://pastebin.com/tuv5JAKd

I had this conversation with Gigahertz a year ago: https://discord.com/channels/313125603924639766/454376090362970122/714598978611707965

SuperMartijn642:
Ah yes, you are correct.
I was thinking something like rotate around the y-axis, then x -axis, then y-axis again, but of course that's not possible 🤦‍♂️

ChampionAsh5357:
If the goal is to support z rotation within the model variant, I would say it would also be of interest to support it within at least the forge data provider if not also the vanilla one.

RobMayer:
Yup - I realized this morning that I should probably patch the DataGen while I'm at it - I will put this into Draft until I do so (if I can figure out how)

RobMayer:
> If the goal is to support z rotation within the model variant, I would say it would also be of interest to support it within at least the forge data provider if not also the vanilla one.

I'm curious as to what you think here. I added the property to VariantProperties, but do you think it worth adding BlockModelGenerator's methods? those seem very specifically structured in a way that I'm hesitant to touch.

ChampionAsh5357:
I would, the easiest way to deal with them is to have overloads that take in the extra param that pass in 0 by default and make sure that values with no rotation do not get written to json to maintain vanilla compat. Forge already does this with a number of methods such as those within `FogRenderer` if you need a reference.

Something like:
```java
// Original Function
foo(int x, int y) {
    foo(x, y, 0);
}

// New Function with original behavior
foo(int x, int y, int z) {
    // Execute code
}
```

Also, I would suggest adding this to the forge model generators as well.

RobMayer:
As discussed in the discord, this is exactly what I did with methods with additional parameters.

RobMayer:
is there any chance that this will / can get reviewed?

ChampionAsh5357:
This still needs the additions to the forge model providers. The vanilla one I considered optional as all though we should support it, forge has its own provider which it supports. Also, you should write a test case to demo the new feature being added.

sciwhiz12:
Hello! @RobMayer, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable. If we do not hear back from you after some time, this PR will be closed for inactivity.

PR #8004, opened 2021-08-14
Add new Ingredient Type that allows NBT Soft-Matching

Drullkus:
Currently, NBTIngredient `forge:nbt` checks for exact-matching in ingredients' NBT for recipes. Sometimes ItemStacks may be littered with extra data or a modder wants to leave some values up for wildcarding, and in these cases the `forge:nbt` will not allow that item to pass. An easily notable example of this are named items.

Normally in Vanilla, recipes are tolerant if any of the stacks are named, but `forge:nbt` doesn't allow named ingredient to pass because its name exists as part of the NBT.

With this change, a new Ingredient type `forge:nbt_fuzzy` is added that checks the NBT against a defined predicate instead of checking for hard equality. This is at Lex's recommendation to avoid having to fuss around with packets.

`forge:nbt`'s ctor is made `public` for DataGen to be consistent with this new `Ingredient`.

----

Example: I've made a Datapack for my testing world adding a recipe that specifically asks for a named Block of Amethyst "`I've got NBT!`". The unnamed Block of Amethyst is not accepted, the named Block of Amethyst has an exact NBT match, and the similarly named Block of Amethyst with added description is accepted anyway because this `NBTIngredient` change ignores the `Lore` tag,

Example of tolerant NBT predicate behavior

Example recipe json for the gif seen above:
```java
{
	"type": "minecraft:crafting_shapeless",
	"ingredients": [{
		"item": "minecraft:amethyst_block",
		"type": "forge:nbt_fuzzy",
		"nbt": {
			"display": {"Name":"{\"text\":\"I've got NBT!\"}"}
		}
	}],
	"result": {
		"item": "minecraft:diamond",
		"count": 1,
		"nbt": {
			"display": {
				"Name": "{\"text\":\"Tada!\"}",
				"Lore": [
					"{\"text\":\"This diamond was crafted from a\"}",
					"{\"text\":\"named Block of Amethyst, neat!\"}"
				]
			}
		}
	}
}
```


src/main/java/net/minecraftforge/common/crafting/NBTIngredient.java
```diff
@@ -78,17 +94,17 @@ public JsonElement toJson()
...
         @Override
         public NBTIngredient parse(JsonObject json) {
-            return new NBTIngredient(CraftingHelper.getItemStack(json, true));
+            return new NBTIngredient(CraftingHelper.getItemStack(json, true), GsonHelper.getAsBoolean(json, "exact_match", false));
         }
 
         @Override
         public void write(FriendlyByteBuf buffer, NBTIngredient ingredient) {
-            buffer.writeItem(ingredient.stack);
+            buffer.writeItem(ingredient.stack).writeBoolean(ingredient.exactMatch);
```
pupnewfster: There is no real reason to chain these calls compared to adding buffer.writeBoolean on the next line which is a bit more readable.

src/main/java/net/minecraftforge/common/crafting/NBTIngredient.java
```diff
@@ -26,17 +26,32 @@
...
+    private final boolean exactMatch;
+    private final NbtPredicate predicate;
+    protected NBTIngredient(ItemStack stack, boolean exactMatch)
     {
         super(Stream.of(new Ingredient.ItemValue(stack)));
         this.stack = stack;
+        this.exactMatch = exactMatch;
+
+        if (this.exactMatch)
+            this.predicate = NbtPredicate.ANY;
```
pupnewfster: I understand that this predicate is never actually used if exact match is true, but this does feel weird setting it to any when it is a very specific thing it expects.

KnightMiner: One option would be setting it to null, and using a null check on the predicate instead of the boolean field to determine whether to use it.

Drullkus: I do agree it is weird, I didn't feel comforting putting it as `null`. If you have any other suggestions or think it should be null, I'm all ears,

Drullkus: Ah, good point Knight. I'm a little `null`-phobic these days but that works

src/main/java/net/minecraftforge/common/crafting/NBTIngredient.java
```diff
@@ -78,17 +93,18 @@ public JsonElement toJson()
...
         @Override
         public NBTIngredient parse(FriendlyByteBuf buffer) {
-            return new NBTIngredient(buffer.readItem());
+            return new NBTIngredient(buffer.readItem(), buffer.readBoolean());
         }
 
         @Override
         public NBTIngredient parse(JsonObject json) {
-            return new NBTIngredient(CraftingHelper.getItemStack(json, true));
+            return new NBTIngredient(CraftingHelper.getItemStack(json, true), GsonHelper.getAsBoolean(json, "exact_match", false));
```
LexManos: This needs to default to true, to preserve existing functionality.

src/main/java/net/minecraftforge/common/crafting/NBTIngredient.java
```diff
@@ -78,17 +93,18 @@ public JsonElement toJson()
...
         }
 
         @Override
         public void write(FriendlyByteBuf buffer, NBTIngredient ingredient) {
             buffer.writeItem(ingredient.stack);
+            buffer.writeBoolean(ingredient.needsExactMatch());
```
LexManos: This is also a network break. So that's a no go.

Drullkus: How do you recommend that I address this? There has to be a boolean, unless you recommend me making a duplicate `forge:nbt_predicate`, or putting the boolean inside the Stack's NBT, which sounds like a terrible solution

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -33,6 +33,7 @@
 import net.minecraft.world.level.storage.LevelStorageSource;
 import net.minecraftforge.api.distmarker.Dist;
 import net.minecraftforge.client.ForgeHooksClient;
+import net.minecraftforge.common.crafting.*;
```
pupnewfster: Forge tries to avoid introducing wildcard imports

src/main/java/net/minecraftforge/common/crafting/NBTIngredientFuzzy.java
```diff
@@ -0,0 +1,97 @@
...
+
+public class NBTIngredientFuzzy extends Ingredient
+{
+    private final ItemStack stack;
+    private final NbtPredicate predicate;
+    protected NBTIngredientFuzzy(ItemStack stack)
+    {
+        super(Stream.of(new ItemValue(stack)));
+        this.stack = stack;
+        this.predicate = stack.hasTag() ? NbtPredicate.ANY : new NbtPredicate(stack.getShareTag());
```
pupnewfster: I think the ternary here is backwards.

src/main/java/net/minecraftforge/common/crafting/NBTIngredientFuzzy.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    public boolean test(@Nullable ItemStack input)
+    {
+        if (input == null)
+            return false;
+        //Can't use areItemStacksEqualUsingNBTShareTag because it compares stack size as well
```
KnightMiner: Comment is not super relevant to this new separate ingredient.

I'm also convinced damage value should be left up to the predicate to decide if its important (though vanilla forces the damage value to be set for item stacks, meaning the constructor would need to take a `NBTPredicate` and item instead for that to work.

Edit: For a little more information on above, on creating an item stack, vanilla calls `setDamage(getDamage())` on item construct (unless that changed in 1.17), which means if the damage NBT tag is missing, it will be set to 0. Used with the current constructor, all usages of this ingredient will have a strict match on damage for any damagable item as that tag is set, which is undesirable behavior (I might want to match golden swords with a particular tag but any damage). I cannot think of a way to avoid this problem without setting the tag without that tag being saved in the itemstack.

Perhaps this could be solved by entirely separating the stack from the NBT Predicate, which could also in theory let you use a NBT Predicate with a tag.

pupnewfster: This comment potentially should be removed given there are other reasons it can't be used for one it is using the predicate which is not checked in the method the comment references.

src/main/java/net/minecraftforge/common/crafting/NBTIngredientFuzzy.java
```diff
@@ -0,0 +1,97 @@
...
+import net.minecraft.advancements.critereon.NbtPredicate;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.crafting.Ingredient;
+import net.minecraft.network.FriendlyByteBuf;
+
+public class NBTIngredientFuzzy extends Ingredient
+{
+    private final ItemStack stack;
+    private final NbtPredicate predicate;
+    protected NBTIngredientFuzzy(ItemStack stack)
```
KnightMiner: Its really annoying for datagen having this constructor be protected. Same issue with the existing nbt ingredient. Any reason not to make it public? Alternatively, could make a static method to get a NBT fuzzy ingredient like exists in the base Ingredient class.

src/main/java/net/minecraftforge/common/crafting/NBTIngredientFuzzy.java
Drullkus: I... kinda see how I missed that in testing. Thanks!

src/main/java/net/minecraftforge/common/crafting/NBTIngredientFuzzy.java
Drullkus: I already separated the NBT from the stack once in the [initial version of my mod](https://www.curseforge.com/minecraft/mc-mods/nbt-ingredient-predicate/) and it created bugs inside the Smithing Table, JEI, Botania, and who knows how many other mods. Unfortunately it's necessary to keep the NBT in the stack.

src/main/java/net/minecraftforge/common/crafting/NBTIngredientFuzzy.java
Drullkus: I'm risking making this PR out of scope but I'll entertain the request

src/main/java/net/minecraftforge/common/crafting/NBTIngredientFuzzy.java
```diff
@@ -0,0 +1,96 @@
...
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+import net.minecraft.advancements.critereon.NbtPredicate;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.crafting.Ingredient;
+import net.minecraft.network.FriendlyByteBuf;
+
+public class NBTIngredientFuzzy extends Ingredient
```
pupnewfster, line 34: It may make more sense for this class to be called `FuzzyNBTIngredient`

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -297,6 +298,7 @@ public void registerRecipeSerialziers(RegistryEvent.Register<RecipeSerializer<?>
 
         CraftingHelper.register(new ResourceLocation("forge", "compound"), CompoundIngredient.Serializer.INSTANCE);
         CraftingHelper.register(new ResourceLocation("forge", "nbt"), NBTIngredient.Serializer.INSTANCE);
+        CraftingHelper.register(new ResourceLocation("forge", "nbt_fuzzy"), NBTIngredientFuzzy.Serializer.INSTANCE);
```
pupnewfster, line 306: I am not sure if this or `fuzzy_nbt` would be better as a name.

src/main/java/net/minecraftforge/common/crafting/NBTIngredientFuzzy.java
```diff
@@ -0,0 +1,96 @@
...
+        this.stack = stack;
+        this.predicate = stack.hasTag() ? new NbtPredicate(stack.getShareTag()) : NbtPredicate.ANY;
+    }
+
+    @Override
+    public boolean test(@Nullable ItemStack input)
+    {
+        if (input == null)
+            return false;
+        return this.stack.getItem() == input.getItem() && this.stack.getDamageValue() == input.getDamageValue() && this.predicate.matches(input);
```
pupnewfster, line 50: I am not sure this even really matters given the partial matching but I think this potentially should end with `this.predicate.matches(input.getShareTag())` so that it is comparing the share tag of the stack similar to how the predicate is built from the share tag? ~(Though I am not quite sure that matches method then short circuits the `NbtPredicate.ANY` to being true so then it potentially would need that check as well)~

Edit: Actually on second look it the `matches(Tag)` method does short circuit for `NbtPredicate.ANY`, but my original point of this comment still stands.

sciwhiz12:
Hello! @Drullkus, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 (or leave an explanation why this PR's changes are not applicable to 1.18). If we do not hear back from you after some time, this PR will be closed for inactivity.

sciwhiz12:

PR #8014, opened 2021-08-17
[1.17.x] implements canConnectRedstone(), Redstone dust now defers to the blockstate for determining connectivity

rikka0w0:
Addressing #7915
#7927 partially re-introduces the `canConnectRedstone` hook, but it is only direction-sensitive. 

A lot of advanced mods require world and blockpos sensitive version of `canConnectRedstone`:
`https://github.com/RoyalAliceAcademyOfSciences/SimElectricity/blob/a6c5c9e5d5d3603b4260de1ad69f4adc38607f59/src/main/java/simelectricity/essential/cable/BlockCable.java#L562`

This PR adds the fully functional `canConnectRedstone` hook, `canConnectRedstone` on 1.17.x will behave as it was back in 1.16.5.


Da-Technomancer:
I agree with what you're trying to do, but two nitpicks:
-The patch in RedstoneWireBlock can be made smaller (in terms of line count), which makes it easier to maintain. 
-As KnightMiner suggested in the other thread (on #7925), you should mention in the javadocs that if the position and world parameters are used, then the block is responsible for creating the necessary updates.

rikka0w0:
> I agree with what you're trying to do, but two nitpicks:
> -The patch in RedstoneWireBlock can be made smaller (in terms of line count), which makes it easier to maintain.
> -As KnightMiner suggested in the other thread (on #7925), you should mention in the javadocs that if the position and world parameters are used, then the block is responsible for creating the necessary updates.

1. A super long line of code may not be easy to understand, especially when it contains multiple pair of brackets.
2. Comments were added, see https://github.com/MinecraftForge/MinecraftForge/blob/d9eb4d1184001702542613a40dc2a2d989d88db7/src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java#L734

TheCurle:
Concur'd. Having complex code in one line makes it *harder* to maintain. There's a fine line between readable and compact.

What we avoid is having lots of injections at various, separate points in a function. As long as the reason / function is clear, moving things into one line isn't typically a worry.

desht:
It looks like the assets for the test case have been added with dot-separated paths, e.g. ` src/test/resources/assets.redstone_sided_connectivity_test.lang/en_us.json`.  IntelliJ strikes again?

rikka0w0:
> It looks like the assets for the test case have been added with dot-separated paths, e.g. ` src/test/resources/assets.redstone_sided_connectivity_test.lang/en_us.json`. IntelliJ strikes again?

The problem has been fixed, in addition, the side that may accept redstone connection now has a new texture, making it easier to be identified.

patches/minecraft/net/minecraft/world/level/block/RedStoneWireBlock.java.patch
```diff
@@ -0,0 +1,39 @@
+--- a/net/minecraft/world/level/block/RedStoneWireBlock.java
++++ b/net/minecraft/world/level/block/RedStoneWireBlock.java
+@@ -199,7 +_,8 @@
+       BlockState blockstate = p_55523_.m_8055_(blockpos);
+       if (p_55526_) {
+          boolean flag = this.m_55612_(p_55523_, blockpos, blockstate);
+-         if (flag && m_55640_(p_55523_.m_8055_(blockpos.m_7494_()))) {
++         flag &= p_55523_.m_8055_(blockpos.m_7494_()).canRedstoneConnectTo(p_55523_, blockpos.m_7494_(), null);
```
SizableShrimp: This is using a bit wise operator, meaning `canRedstoneConnectTo` will always be run regardless of if `canSurviveOn` returns false. This should be amended to be `flag = flag &&` to use the proper boolean AND operator, which means the jvm will not execute the second half if the first half is already false.

patches/minecraft/net/minecraft/world/level/block/RedStoneWireBlock.java.patch
```diff
@@ -0,0 +1,39 @@
...
++         if (flag) {
+             if (blockstate.m_60783_(p_55523_, blockpos, p_55525_.m_122424_())) {
+                return RedstoneSide.UP;
+             }
+@@ -207,8 +_,9 @@
+             return RedstoneSide.SIDE;
+          }
+       }
+-
```
SizableShrimp: Doesn't really matter, but you delete an empty line here for no reason which means the hunk is one line larger.

patches/minecraft/net/minecraft/world/level/block/RedStoneWireBlock.java.patch
```diff
@@ -0,0 +1,39 @@
...
+-      return !m_55594_(blockstate, p_55525_) && (blockstate.m_60796_(p_55523_, blockpos) || !m_55640_(p_55523_.m_8055_(blockpos.m_7495_()))) ? RedstoneSide.NONE : RedstoneSide.SIDE;
++      boolean flag1 = blockstate.canRedstoneConnectTo(p_55523_, blockpos, p_55525_);
++      boolean flag2 = p_55523_.m_8055_(blockpos.m_7495_()).canRedstoneConnectTo(p_55523_, blockpos.m_7495_(), null);
++      return !flag1 && (blockstate.m_60796_(p_55523_, blockpos) || !flag2) ? RedstoneSide.NONE : RedstoneSide.SIDE;
```
SizableShrimp: You store the output of `canRedstoneConnectTo` for the blockpos below in a `flag2` variable, which means any expensive operations could potentially be calculated for no reason if the boolean condition can be exited early because either `flag1` is true or `isRedstoneConductor` is true. I would also stray away from choosing variable names that forgeflower will automatically choose, as you've done here. This means if another boolean variable were to be added to the method, it would conflict with the one you declare. It's always better to choose a readable variable name over 2 variables only differing by a number, especially when that conflicts with what the decompiler generates for local variable names.

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -719,4 +719,26 @@ default boolean isScaffolding(BlockState state, LevelReader world, BlockPos pos,
...
+     * @param pos The block position in world
+     * @param direction The coming direction of the redstone dust connection (with respect to the block at pos)
+     * @return True if redstone dust should visually connect on the side passed
+     * <p> If the return value is evaluated based on world and pos (e.g. from BlockEntity), then the implementation of
+     *  this block should notify the its neighbors to update their shapes when necessary. Consider using
+     *  "yourBlockState.updateNeighbourShapes(world, yourBlockPos, UPDATE_ALL);" or
+     *  "neighborState.updateShape(fromDirection, stateOfYourBlock, world, neighborBlockPos, yourBlockPos)",
+     *  fromDirection is defined from the neighbor block's point of view.
+     */
+    default boolean canConnectRedstone(BlockState state, BlockGetter world, BlockPos pos, @Nullable Direction direction)
```
SizableShrimp, line 752: This javadoc could use some work. First off, you already import `RedStoneWireBlock`, so you can use the imported classname in the link. Second off, using code/links in quote is a bad idea. You should instead use `{@link}` or `{@code}`. You also do not need to indent an extra space for extra paragraphs. I think a format accepted by forge is to denote new paragraphs with a single `<p>` on a javadoc line and then a new paragraph below that. Also, "the its" is just wrong.

src/test/java/net/minecraftforge/debug/block/RedstoneSidedConnectivityTest.java
```diff
@@ -0,0 +1,95 @@
...
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.fmllegacy.RegistryObject;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+
+import javax.annotation.Nullable;
+
+@Mod(RedstoneSidedConnectivityTest.MODID)
+@Mod.EventBusSubscriber(modid = RedstoneSidedConnectivityTest.MODID, bus = Bus.MOD)
+public class RedstoneSidedConnectivityTest
```
SizableShrimp, line 43: This test mod should preferably have a public static final ENABLED boolean that is set to true. And it should register everything in the mod constructor only if ENABLED is set to true, along with this `@EventBusSubscriber` annotation being unnecessary. This makes it easy to disable test mods for testing / incompatibility purposes, especially when porting if something breaks. Your code style is inconsistent, please choose something instead of mixing two.

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -382,6 +382,7 @@ public net.minecraft.world.level.block.WoodButtonBlock <init>(Lnet/minecraft/wor
 public net.minecraft.world.level.block.WoolCarpetBlock <init>(Lnet/minecraft/world/item/DyeColor;Lnet/minecraft/world/level/block/state/BlockBehaviour$Properties;)V # constructor
 #endgroup
 public net.minecraft.world.level.block.Block m_49805_(Lnet/minecraft/server/level/ServerLevel;Lnet/minecraft/core/BlockPos;I)V # popExperience
+public net.minecraft.world.level.block.RedStoneWireBlock m_55594_(Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/core/Direction;)Z # shouldConnectTo
```
SizableShrimp: Please run `checkATs` as this line is in the wrong spot.

SizableShrimp (changes requested):
A few things that still need some work. Did not run the test mod, just looked through the code.

patches/minecraft/net/minecraft/world/level/block/RedStoneWireBlock.java.patch
rikka0w0: Fixed in the latest commit.

patches/minecraft/net/minecraft/world/level/block/RedStoneWireBlock.java.patch
rikka0w0: In the latest commit, that long line of return has been replaced with if statements, making it more readable. Unnecessary function calls are avoided.

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
rikka0w0: The javadoc has been formatted according to the suggestions.

src/test/java/net/minecraftforge/debug/block/RedstoneSidedConnectivityTest.java
rikka0w0: `public static final boolean ENABLE = true;` has been added, inconsistent naming convention has been fixed.

rikka0w0:
Test mod:

src/main/resources/META-INF/accesstransformer.cfg
rikka0w0: I ran `./gradlew checkATs` in the root folder and it didnt fail. What convention is used in this file? Should I simply append the new line at the end of this file?

rikka0w0: Should I place it in between `#group` and `#endgroup`?

rikka0w0: Should I place it in between `#group` and `#endgroup`?

rikka0w0 (commented):
problems are fixed.

src/main/resources/META-INF/accesstransformer.cfg
SizableShrimp: Running checkATs will automatically resort the line.

SizableShrimp (approved):

src/test/java/net/minecraftforge/debug/block/RedstoneSidedConnectivityTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+    private static final DeferredRegister<Block> BLOCKS = DeferredRegister.create(ForgeRegistries.BLOCKS, MODID);
+    private static final DeferredRegister<Item> ITEMS = DeferredRegister.create(ForgeRegistries.ITEMS, MODID);
+
+    private static final RegistryObject<Block> TEST_REDSTONE_BLOCK = BLOCKS.register(BLOCK_ID, EastRedstoneBlock::new);
+
+    static
+    {
+        ITEMS.register(BLOCK_ID, () -> new BlockItem(TEST_REDSTONE_BLOCK.get(), new Item.Properties().tab(CreativeModeTab.TAB_BUILDING_BLOCKS)));
+    }
```
pupnewfster: This would be cleaner if you just assigned it to a RegistryObject variable rather than using a static init block, even if you never are making use of the variable.

patches/minecraft/net/minecraft/world/level/block/RedStoneWireBlock.java.patch
```diff
@@ -0,0 +1,43 @@
+--- a/net/minecraft/world/level/block/RedStoneWireBlock.java
++++ b/net/minecraft/world/level/block/RedStoneWireBlock.java
+@@ -199,7 +_,8 @@
+       BlockState blockstate = p_55523_.m_8055_(blockpos);
+       if (p_55526_) {
+          boolean flag = this.m_55612_(p_55523_, blockpos, blockstate);
+-         if (flag && m_55640_(p_55523_.m_8055_(blockpos.m_7494_()))) {
++         flag = flag && p_55523_.m_8055_(blockpos.m_7494_()).canRedstoneConnectTo(p_55523_, blockpos.m_7494_(), null);
++         if (flag) {
```
pupnewfster: This should just be patching the if statement from `shouldConnectTo` to `canRedstoneConnectTo`, I don't see any reason that the value of the flag boolean should be getting changed here.

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -387,6 +387,7 @@ public net.minecraft.world.level.block.WoolCarpetBlock <init>(Lnet/minecraft/wor
 public net.minecraft.world.level.block.Block m_49805_(Lnet/minecraft/server/level/ServerLevel;Lnet/minecraft/core/BlockPos;I)V # popExperience
 public net.minecraft.world.level.block.FireBlock m_53492_(Lnet/minecraft/world/level/block/state/BlockState;)I # getBurnOdd
 public net.minecraft.world.level.block.FireBlock m_53494_(Lnet/minecraft/world/level/block/state/BlockState;)I # getFlameOdds
+public net.minecraft.world.level.block.RedStoneWireBlock m_55594_(Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/core/Direction;)Z # shouldConnectTo
```
pupnewfster: Given how straightforward and short the `shouldConnectTo` helper method is and that it doesn't need to access any private fields or things.. It would make more sense to just copy the few line body to the default implementation for `canRedstoneConnectTo` rather than using an AT to make it be "public API" and then just deprecating it with a comment saying not to use it.

patches/minecraft/net/minecraft/world/level/block/RedStoneWireBlock.java.patch
```diff
@@ -0,0 +1,43 @@
...
++      }
+    }
+ 
+    public boolean m_7898_(BlockState p_55585_, LevelReader p_55586_, BlockPos p_55587_) {
+@@ -350,10 +_,14 @@
+       }
+    }
+ 
++   //Forge: Do not use, use IForgeBlockState::canRedstoneConnectTo
++   @Deprecated
```
pupnewfster: It doesn't really matter but given my other comments and that you will need to be regening patches, this might as well just all be on one line: `@Deprecated //Forge: ...`

src/test/java/net/minecraftforge/debug/block/RedstoneSidedConnectivityTest.java
rikka0w0: This has been addressed in the latest commit

patches/minecraft/net/minecraft/world/level/block/RedStoneWireBlock.java.patch
rikka0w0: Fixed

src/main/resources/META-INF/accesstransformer.cfg
rikka0w0: In the latest commit, the AT has been removed, the default impl has been copied to `IForgeBlock::canConnectRedstone`.

patches/minecraft/net/minecraft/world/level/block/RedStoneWireBlock.java.patch
rikka0w0: Fixed

rikka0w0 (commented):
Problems have been addressed

PR #8023, opened 2021-08-21
[1.17.x]: Added SendPlayerPermissionLevelEvent

mysticdrew:
This PR adds and event on the server for when a players permission level is changed via command or another mod. 

This is a useful event for me because I have UI that is only available for OPs and I don't want to poll the server constantly to see if the current user is an OP. With this event, I can be notified is a player is OPed via command or if another mod sets it through code, then tell the ui that the user has permissions.

### DIFF
--- a/patches/minecraft/net/minecraft/server/players/PlayerList.java.patch
+++ b/patches/minecraft/net/minecraft/server/players/PlayerList.java.patch
@@ -81,6 +81,14 @@
        if (flag2) {
           serverplayer.f_8906_.m_141995_(new ClientboundSoundPacket(SoundEvents.f_12377_, SoundSource.BLOCKS, (double)blockpos.m_123341_(), (double)blockpos.m_123342_(), (double)blockpos.m_123343_(), 1.0F, 1.0F));
        }
+@@ -563,6 +_,7 @@
+             b0 = (byte)(24 + p_11228_);
+          }
+ 
++         net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.entity.player.PlayerPermissionChangedEvent(p_11227_, p_11228_));
+          p_11227_.f_8906_.m_141995_(new ClientboundEntityEventPacket(p_11227_, b0));
+       }
+ 
 @@ -726,13 +_,6 @@
        if (serverstatscounter == null) {
           File file1 = this.f_11195_.m_129843_(LevelResource.f_78175_).toFile();
--- a/patches/minecraft/net/minecraft/server/players/PlayerList.java.patch
+++ b/patches/minecraft/net/minecraft/server/players/PlayerList.java.patch
@@ -81,14 +81,38 @@
        if (flag2) {
           serverplayer.f_8906_.m_141995_(new ClientboundSoundPacket(SoundEvents.f_12377_, SoundSource.BLOCKS, (double)blockpos.m_123341_(), (double)blockpos.m_123342_(), (double)blockpos.m_123343_(), 1.0F, 1.0F));
        }
-@@ -563,6 +_,7 @@
-             b0 = (byte)(24 + p_11228_);
-          }
+@@ -535,19 +_,23 @@
+    }
  
-+         net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.entity.player.SendPlayerPermissionLevelEvent(p_11227_, p_11228_));
-          p_11227_.f_8906_.m_141995_(new ClientboundEntityEventPacket(p_11227_, b0));
+    public void m_5749_(GameProfile p_11254_) {
+-      this.f_11200_.m_11381_(new ServerOpListEntry(p_11254_, this.f_11195_.m_7022_(), this.f_11200_.m_11351_(p_11254_)));
+-      ServerPlayer serverplayer = this.m_11259_(p_11254_.getId());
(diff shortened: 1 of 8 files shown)


diesieben07:
The client is already notified of its permission level through an entity event (sent from the same method you're patching), so `LocalPlayer#permissionLevel` is always correct and thus `hasPermissions` will work fine for the client player - does this not solve your problem of not having to poll the server?

mysticdrew:
It does for one instance, yes. 
But, we also enable/disable client features if a user is opped or not. We send a packet with the features to the client on login and dimension change as they can change per dimension. They can also be enabled for ops and disabled for non ops. This event would be beneficial so we can send the features to the client when the permission level changes. I also understand that in this case the features are enabled/disabled on the client as it is primarily a client mod and it can be modded/hacked, but it is enough for most server admins. Just like locking your car door wont stop all thieves, just most. 

I could use the command event and check for op/deop but it would not catch the case if a mod sets a users permission level through code.

diesieben07:
Well then your initial description is wrong, since that says you don't want to poll the server from the client for the OP status, which is not necessary even without this event.

If it is added to do things on the server, the javadoc and potentially the name must be changed, because in your current implementation it fires also when the permission level has not changed at all. It fires also for example on dimension change or teleport or resource reload.

mysticdrew:
I do have several uses for this yes, I only gave one use case I had, which I don't need this anymore for that use case. Thanks for that by the way. 
You are correct, I will update the names and javadoc.

mysticdrew:
@diesieben07 I made the changes.

src/main/java/net/minecraftforge/event/entity/player/SendPlayerPermissionLevelEvent.java
```diff
@@ -0,0 +1,47 @@
...
+
+package net.minecraftforge.event.entity.player;
+
+import net.minecraft.server.level.ServerPlayer;
+
+/**
+ * This event is fired on the server when a player's permission level is sent to the client.
+ *
+ * This event will fire on resource reload, when the player's dimension changes, when the player teleports, on respawn,
+ * when the new player is placed in world, and when the player is opped or deopped.
```
diesieben07: I really wouldn't specify this that clearly here. 
> This event will fire when the player's permission level changes, but will also fire at other times when the game syncs the permission level to the client.

diesieben07 (changes requested):

src/main/java/net/minecraftforge/event/entity/player/SendPlayerPermissionLevelEvent.java
mysticdrew: 👍 
Sounds good. Addressed the comment, thanks! 
Pushed

diesieben07 (approved):

LexManos:
Wouldn't this be better as a 'PermissionsChangedEvent' with a old, new and player params?
So it has nothing to do with networking and is more of a 'things changed' notification.

mysticdrew:
> Wouldn't this be better as a 'PermissionsChangedEvent' with a old, new and player params?
> So it has nothing to do with networking and is more of a 'things changed' notification.

Sure, I will update shortly.

mysticdrew:
I updated this now only fire on op and deop and made the event cancelable.

diesieben07 (changes requested):
Please use early returns in those patches to reduce the patch size.

mysticdrew:
@diesieben07 ahh yea ok good point.

patches/minecraft/net/minecraft/server/players/PlayerList.java.patch
```diff
@@ -81,6 +81,24 @@
...
++         return;
+       this.f_11200_.m_11381_(new ServerOpListEntry(p_11254_, this.f_11195_.m_7022_(), this.f_11200_.m_11351_(p_11254_)));
+       ServerPlayer serverplayer = this.m_11259_(p_11254_.getId());
+       if (serverplayer != null) {
+@@ -544,6 +_,8 @@
+    }
+ 
+    public void m_5750_(GameProfile p_11281_) {
++      if (!net.minecraftforge.event.ForgeEventFactory.onPermissionChanged(p_11281_, 0, this))
++         return;
```
diesieben07: Put the return on the same line please.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -249,7 +252,7 @@ public static Component getPlayerDisplayName(Player player, Component username)
         MinecraftForge.EVENT_BUS.post(event);
         return event.getDisplayname();
     }
-    
+
```
diesieben07: Unnecessary whitespace change.

diesieben07 (changes requested):

patches/minecraft/net/minecraft/server/players/PlayerList.java.patch
mysticdrew: 👍

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
mysticdrew: 👍

mysticdrew:
Updated based on your comments @diesieben07

diesieben07 (approved):

PR #8024, opened 2021-08-22
Integrate the Gametest framework with Forge

SizableShrimp:
This PR modifies the Gametest framework to be more compatible with Forge and more accessible to mods. Notable changes are:
* Providing a `GameTestMain` launcher class for the gametest server and gametest server launch handlers for forge dev and user dev, along with run configurations for forge dev
* * One question is whether a default gametest configuration should be included in the MDK's build.gradle. At the moment, the gametest run configuration is included in the userdev JSON, but not included in the MDK's build.gradle.
* Adding a new SharedConstant field called `ENABLE_GAMETEST` which can be enabled using the system property `forge.enablegametest`. This boolean supplier has been patched to be used in vanilla, to provide more granular control to enable the gametest system without imposing the silly requirements of `IS_RUNNING_IN_IDE` like forcing all the class names of Blocks and Items to end in `Block` and `Item`, respectively.
* Adding a few new properties to the `Gametest` annotation and patching `StructureUtils` to allow for more control over the template name and its resource namespace.
* Allowing `@Gametest` methods to be static

While it is not a goal of this PR, in the future, CI could be setup to run `forge forge test gametest` for each PR to validate whether all tests pass successfully. Test mods could register game tests in their mod constructors to the `GameTestRegistry` class to validate behavior is correct. Given enough time and enough tests, this could help find bugs in Forge when porting to newer Minecraft versions using automated CI and powerful reporting from Gametest versus user testing and reporting. The system exit code of the `GameTestServer` signifies whether the tests passed, where the exit code is the number of required tests that failed. For a success, this means the exit code would be 0.


src/main/java/net/minecraftforge/gametest/GameTestMain.java
```diff
@@ -0,0 +1,192 @@
...
+            if (worldData == null) {
+                DedicatedServerProperties dedicatedserverproperties = serverSettings.getProperties();
+                LevelSettings levelsettings = new LevelSettings(dedicatedserverproperties.levelName, dedicatedserverproperties.gamemode, dedicatedserverproperties.hardcore,
+                        dedicatedserverproperties.difficulty, false, new GameRules(), datapacks);
+                WorldGenSettings worldgensettings = dedicatedserverproperties.getWorldGenSettings(registryHolder);
+
+                // Forge: Deserialize the DimensionGeneratorSettings to ensure modded dims are loaded on first server load (see SimpleRegistryCodec#decode). Vanilla behaviour only loads from the
+                // server.properties and deserializes only after the 2nd server load.
+                worldgensettings =
+                        WorldGenSettings.CODEC.encodeStart(net.minecraft.resources.RegistryWriteOps.create(NbtOps.INSTANCE, registryHolder), worldgensettings).flatMap(nbt -> WorldGenSettings.CODEC.parse(registryOps, nbt)).getOrThrow(false, errorMsg -> {});
```
Cyborgmas: not sure why this was split?

Cyborgmas:
I think a testmod would be of use to show how the system works

patches/minecraft/net/minecraft/gametest/framework/StructureUtils.java.patch
```diff
@@ -0,0 +1,17 @@
+--- a/net/minecraft/gametest/framework/StructureUtils.java
++++ b/net/minecraft/gametest/framework/StructureUtils.java
+@@ -234,11 +_,12 @@
+ 
+    private static StructureTemplate m_127880_(String p_127881_, ServerLevel p_127882_) {
+       StructureManager structuremanager = p_127882_.m_8875_();
+-      Optional<StructureTemplate> optional = structuremanager.m_163774_(new ResourceLocation(p_127881_));
++      ResourceLocation location = new ResourceLocation(p_127881_);
++      Optional<StructureTemplate> optional = structuremanager.m_163774_(location);
```
pupnewfster: I feel like splitting this even if it makes it so that it has to switch from string to resource location one extra time below it isn't an expensive enough call to justify the extra hunk in the patch up here.

fmlloader/src/main/java/net/minecraftforge/fml/loading/targets/ForgeGametestDevLaunchHandler.java
```diff
@@ -0,0 +1,20 @@
+package net.minecraftforge.fml.loading.targets;
+
+import net.minecraftforge.api.distmarker.Dist;
+
+import java.util.concurrent.Callable;
+
+public class ForgeGametestDevLaunchHandler extends CommonDevLaunchHandler {
+    @Override public String name() { return "forgegametestdev"; }
+    @Override public Dist getDist() { return Dist.DEDICATED_SERVER; }
```
pupnewfster, line 28: Two questions:
1. What is the purpose of having `ForgeGametestDevLaunchHandler` and `ForgeGametestUserdevLaunchHandler` as there does not seem to really be a difference between them
2. Why is the dist for them dedicated server and not client, as if memory serves I thought vanilla's gametest framework somehow has support for testing the client and some rendering things as well.

pupnewfster (commented):
To answer your question I think it probably does make sense to add a default run config in the MDK for gametest as similar to datagen it probably makes sense to provide it easily to people rather than them having to find and copy paste a decent chunk of the build.gradle

fmlloader/src/main/java/net/minecraftforge/fml/loading/targets/ForgeGametestDevLaunchHandler.java
SizableShrimp: To answer number 1, the inheritance of both classes is different. This is how all the launch handler works. It could be potentially refactored in another PR, but isn't something urgent as it's just a few lines of code duplication.
For number 2, I have no idea if the gametest **server** can test client stuff. `GameTestServer` is just a subclass of `MinecraftServer`, but I don't think it has access to any client stuff. However, one could theoretically write a game test and register it only on Dist.CLIENT, assuming they have `forge.gametest` set to `true`, then they can run that test manually with `/test runall` in their client instance.

SizableShrimp:
I've added a gametest test mod with detailed comments which should help with people trying to learn the framework, along with some other general cleanup.

patches/minecraft/net/minecraft/SharedConstants.java.patch
```diff
@@ -1,5 +1,13 @@
 --- a/net/minecraft/SharedConstants.java
 +++ b/net/minecraft/SharedConstants.java
+@@ -102,6 +_,7 @@
+    public static final long f_136181_ = Duration.ofMillis(300L).toNanos();
+    public static boolean f_136182_ = true;
+    public static boolean f_136183_;
++   public static final java.util.function.BooleanSupplier ENABLE_GAMETEST = () -> f_136183_ || Boolean.getBoolean("forge.enablegametest");
```
alcatrazEscapee: Just out of total curiosity, is there any reason this needs to be a patch at all? It could totally be located in forge code somewhere. Like, yes, vanilla has all it's constants here but since forge is adding stuff, there's no reason forge needs to put it's constants here as well.

You could also then make it a method rather than this odd "field which is a boolean supplier yet only used as a method" i.e. `ForgeMod.areGameTestsEnabled();`

alcatrazEscapee (commented):
Yay! :D

src/main/java/net/minecraftforge/gametest/GameTestMain.java
```diff
@@ -0,0 +1,195 @@
...
+            if (!optionSet.has(acceptEulaOpt) && !eula.hasAgreedToEULA()) {
+                LOGGER.info("You need to agree to the EULA in order to run the server. Go to eula.txt for more info.");
+                return;
+            }
+
+            CrashReport.preload();
+            Bootstrap.bootStrap();
+            Bootstrap.validate();
+            Util.startTimerHackThread();
+            net.minecraftforge.fmllegacy.server.ServerModLoader.load(); // Load mods before we load almost anything else anymore. Single spot now.
```
pupnewfster: This may as well just import this rather than using the fully qualified name given it isn't in a patch.

src/main/java/net/minecraftforge/gametest/GameTestMain.java
```diff
@@ -0,0 +1,195 @@
...
+import java.io.File;
+import java.nio.file.Paths;
+import java.util.Collection;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+public class GameTestMain {
+    private static final Logger LOGGER = LogManager.getLogger();
+
+    public static void main(String[] args) {
```
pupnewfster, line 81: Given this is a bit on the complex side, it is probably worth a comment either on the method or even on the class that most of this is based on/copied from `net.minecraft.server.Main` so that if something gets out of sync when porting between versions it is easier to know what this was based on.

src/main/java/net/minecraftforge/gametest/GameTestMain.java
```diff
@@ -0,0 +1,195 @@
...
+        System.setProperty("forge.enablegametest", "true");
+        SharedConstants.tryDetectVersion();
+        OptionParser optionParser = new OptionParser();
+        OptionSpec<Void> helpOpt = optionParser.accepts("help").forHelp();
+        OptionSpec<Void> safeModeOpt = optionParser.accepts("safeMode", "Loads level with vanilla datapack only");
+        OptionSpec<String> savesDirOpt = optionParser.accepts("universe").withRequiredArg().defaultsTo(".");
+        OptionSpec<String> saveNameOpt = optionParser.accepts("world").withRequiredArg();
+        OptionSpec<BlockPos> spawnPosOpt = optionParser.accepts("spawnPos").withRequiredArg().withValuesConvertedBy(new BlockPosValueConverter()).defaultsTo(new BlockPos(0, 60, 0));
+        OptionSpec<Void> acceptEulaOpt = optionParser.accepts("acceptEula");
```
pupnewfster: Honestly this option strikes me as a bit odd given the fact that with how things are setup currently, this option is effectively a "require manually accepting the eula", and defaults to "false". So I am not really sure what benefit this option serves given I highly doubt someone would add it just to have to accept the eula as well when without it they could just run it without having to bother accepting it. Part of me feels like because of that this option should just be removed, or if not at the very least a description should be added that basically says this makes it require manually accepting it.

src/main/java/net/minecraftforge/gametest/GameTestMain.java
```diff
@@ -0,0 +1,195 @@
...
+                @Override
+                public void run() {
+                    // We don't need to halt here as GameTestServer always calls System#exit on both crash and normal exit
+                    LogManager.shutdown(); // we're manually managing the logging shutdown on the server. Make sure we do it here at the end.
+                }
+            };
+            thread.setUncaughtExceptionHandler(new DefaultUncaughtExceptionHandler(LOGGER));
+            Runtime.getRuntime().addShutdownHook(thread);
+        } catch (Exception e) {
+            LOGGER.fatal("Failed to start the minecraft server", e);
```
pupnewfster: This logger message probably should mention something about game test minecraft server rather than just minecraft server.

src/main/java/net/minecraftforge/gametest/GameTestMain.java
```diff
@@ -0,0 +1,195 @@
...
+
+    private static final class BlockPosValueConverter implements ValueConverter<BlockPos> {
+        @Override
+        public BlockPos convert(String value) {
+            String[] split = value.split(",");
+            return new BlockPos(Double.parseDouble(split[0]), Double.parseDouble(split[1]), Double.parseDouble(split[2]));
+        }
+
+        @Override
+        public Class<BlockPos> valueType() {return BlockPos.class;}
```
pupnewfster: Is there a reason this method and the valuePattern method are all shoved onto a single line?

src/test/java/net/minecraftforge/debug/misc/GameTestTest.java
```diff
@@ -0,0 +1,72 @@
...
+import net.minecraft.gametest.framework.TestFunction;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.level.block.Blocks;
+import net.minecraft.world.level.block.Rotation;
+import net.minecraftforge.fml.common.Mod;
+
+import java.util.List;
+
+@Mod(GameTestTest.MODID)
+public class GameTestTest {
```
pupnewfster: Brackets in this class should be on new lines just like how it is done in other test mods and forge code. (Note: that **_does not_** mean **just** this _specific_ bracket I am commenting on).

src/test/java/net/minecraftforge/debug/misc/GameTestTest.java
```diff
@@ -0,0 +1,72 @@
...
+    public static final boolean ENABLE = true;
+
+    public GameTestTest() {
+        if (ENABLE) {
+            // Registers all methods annotated with @GameTest or @GameTestGenerator to the GameTestRegistry
+            GameTestRegistry.register(this.getClass());
+        }
+    }
+
+    // An example game test
```
pupnewfster: Is there a reason the comments on this method and on `generateTests` are formatted as regular comments rather than javadocs?

pupnewfster (commented):
Finally got a chance to do some more testing on this and review it again, and it overall it looks pretty good. One overarching question I have though is how does it handle tests dependencies may have? For example if one modder makes a library and adds *a lot* of testing for it, and then I were to use that library and add a few tests I want to have run, does that mean I would be forced to let it rerun all of the tests of the library (which in theory should be stable and already tested by the modder who wrote it) each time I want to run my tests?

alcatrazEscapee:
> One overarching question I have though is how does it handle tests dependencies may have?

@pupnewfster If said library ships tests, both test classes and structures, as part of their mod/library, then yes. But what's stopping you (or said mod/library) from putting their tests in the, ya know, *test* source set? Then they won't be shipped with said library and you won't have their tests in your development environment.

SizableShrimp:
GameTestServer will run all loaded tests, however if you use the `/test` command you can choose which tests to run. However, all tests are run concurrently and as long as the test isn't more than a couple seconds it really shouldn't matter that much. Using a test sourceset for gametests makes sense, however, it means other mods cannot use your gametests for compat testing.

fmlloader/src/main/java/net/minecraftforge/fml/loading/targets/ForgeGametestDevLaunchHandler.java
pupnewfster: After having had a chance to actually do some testing and take a better/more in depth look at how game tests work I better understand that the reason it is dedicated server is that is just what the environment that the runs that are being tested are for in that it launches a server and runs the test, and that for testing the client launching the client and using the command works fine.

Part of me wishes there was a better solution for how to register tests that are "client side only" rather than having the modders have to do the dist checks themselves, but that is not that big of a deal and if someone comes up with a better way of doing that then they can always make a follow up PR. Thinking about it maybe the "best" solution from an API standpoint might be to allow for a dist value to be specified in the `GameTest` and `GameTestGenerator` annotations (and maybe even BeforeBatch and AfterBatch) and then patch the if statements that do things if the annotation was found to also require the dist to be correct... But given how forge has to patch things I feel like that would probably be overkill from extra maintenance standpoint, compared to just telling devs to put the client only tests in a different class and then manually check the dist and only add the client only test classes if it is the client.

src/main/java/net/minecraftforge/gametest/GameTestMain.java
SizableShrimp: I'm not sure what you mean here. If the option is enabled, the eula will be considered accepted, and the check on `eula.txt` will be skipped. If the option is not present, then `eula.txt` must have `eula=true` or the server will exist.

pupnewfster:
> @pupnewfster If said library ships tests, both test classes and structures, as part of their mod/library, then yes. But what's stopping you (or said mod/library) from putting their tests in the, ya know, _test_ source set? Then they won't be shipped with said library and you won't have their tests in your development environment.

That is a good point, though given how many mods ship all their datagen related code, I wouldn't be surprised if most also ship all their tests as well.


> GameTestServer will run all loaded tests, however if you use the `/test` command you can choose which tests to run. However, all tests are run concurrently and as long as the test isn't more than a couple seconds it really shouldn't matter that much. Using a test sourceset for gametests makes sense, however, it means other mods cannot use your gametests for compat testing.

Fair enough, and yes I wouldn't expect them to necessarily add lots of time to the testing, though with some mods who knows what people may do (though then again if someone runs into that then they can always make a PR to try and adjust that to add a way to exclude them running in the GameTestServer).

KnightMiner:
I will say that for the record, Tinkers Construct is currently using a JUnit based test framework, and we use a separate source set. It was overall easier to set up tests as a separate source set, while datagen was easier to set up as part of the main sourceset (due to needing to subscribe to certain forge events). I suspect other mods will view things the same way and be less likely to ship tests in their libraries. If they do, can always open an issue with the library as that would be a bug in my book.

src/main/java/net/minecraftforge/gametest/GameTestMain.java
pupnewfster: On second look I had read the if statement that makes use of this backwards. Though what I had meant when I said with the description, I meant as the second parameter to `accepts` similar to how `safeMode` has, rather than just as a comment above it.

fmlloader/src/main/java/net/minecraftforge/fml/loading/targets/ForgeGametestDevLaunchHandler.java
```diff
@@ -0,0 +1,20 @@
+package net.minecraftforge.fml.loading.targets;
```
sciwhiz12, line 20: Missing license header.

fmlloader/src/main/java/net/minecraftforge/fml/loading/targets/ForgeGametestDevLaunchHandler.java
```diff
@@ -0,0 +1,20 @@
...
+
+import net.minecraftforge.api.distmarker.Dist;
+
+import java.util.concurrent.Callable;
+
+public class ForgeGametestDevLaunchHandler extends CommonDevLaunchHandler {
```
sciwhiz12, line 26: Braces on newlines, please.

fmlloader/src/main/java/net/minecraftforge/fml/loading/targets/ForgeGametestUserdevLaunchHandler.java
```diff
@@ -0,0 +1,20 @@
+package net.minecraftforge.fml.loading.targets;
```
sciwhiz12, line 20: Missing license header.

fmlloader/src/main/java/net/minecraftforge/fml/loading/targets/ForgeGametestUserdevLaunchHandler.java
```diff
@@ -0,0 +1,20 @@
...
+
+import net.minecraftforge.api.distmarker.Dist;
+
+import java.util.concurrent.Callable;
+
+public class ForgeGametestUserdevLaunchHandler extends ForgeUserdevLaunchHandler {
```
sciwhiz12, line 26: Braces on newlines, please.

src/main/java/net/minecraftforge/gametest/GameTestMain.java
```diff
@@ -0,0 +1,210 @@
...
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
```
sciwhiz12, line 18: As the javadocs of this class says,
> ```
> /**
> * A copy of {@link net.minecraft.server.Main server main} with parts removed that are not relevant to {@link GameTestServer}.
> * Care should be taken to make sure this stays relatively close to server main during updating.
> */
> ```

This makes it a bit iffy on whether we can apply Forge's LGPL license header on a class which purports to be a copy of `net.minecraftforge.gametest.GameTestMain` with parts removed. I'm leaning towards the train of thought that this should not be headed by Forge's LGPL notice, though I don't know what license text notice to be added as replacement, if one is appropriate.

src/test/java/net/minecraftforge/debug/misc/GameTestTest.java
```diff
@@ -0,0 +1,95 @@
+package net.minecraftforge.debug.misc;
```
sciwhiz12, line 20: Missing license header.

src/test/java/net/minecraftforge/debug/misc/GameTestTest.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    @GameTest(templateNamespace = MODID)
+    public static void testWood(GameTestHelper helper)
+    {
+        // The woodPos is in the bottom center of the 3x3x3 structure
+        BlockPos woodPos = new BlockPos(1, 1, 1);
+
+        // assertBlockState will convert the relative woodPos into a real world block position and check it with the predicate
+        // If the predicate fails, the String supplier will be used to construct an exception message, which is thrown
+        helper.assertBlockState(woodPos, b -> b.is(Blocks.OAK_LOG), () -> "Block was not an oak log");
```
sciwhiz12: It is unclear as to from where the relative positions passed to the methods of `GameTestHelper` are based from, when converting them to absolute world coordinates.

src/main/java/net/minecraftforge/gametest/GameTestMain.java
```diff
@@ -0,0 +1,210 @@
...
+        try {
+            OptionSet optionSet = optionParser.parse(args);
+            if (optionSet.has(helpOpt)) {
+                optionParser.printHelpOn(System.err);
+                return;
+            }
+            Eula eula = new Eula(Paths.get("eula.txt"));
+
+            if (!optionSet.has(acceptEulaOpt) && !eula.hasAgreedToEULA()) {
+                LOGGER.info("You need to agree to the EULA in order to run the server. Go to eula.txt for more info.");
```
sciwhiz12: Should this be modified to mention the new `--acceptEula` argument?

src/main/java/net/minecraftforge/fmllegacy/server/ServerLifecycleHooks.java
```diff
@@ -97,7 +100,11 @@ public static boolean handleServerAboutToStart(final MinecraftServer server)
 
     public static boolean handleServerStarting(final MinecraftServer server)
     {
-        DistExecutor.runWhenOn(Dist.DEDICATED_SERVER, ()->()->LanguageHook.loadLanguagesOnServer(server));
+        DistExecutor.runWhenOn(Dist.DEDICATED_SERVER, ()->()->{
+            LanguageHook.loadLanguagesOnServer(server);
+            if (!(server instanceof GameTestServer))
```
pupnewfster, line 106: This could probably use a comment at the end to explain this. I understand why it is happening but to someone in the future reading the code it probably looks rather backwards.

patches/minecraft/net/minecraft/gametest/framework/GameTestRegistry.java.patch
```diff
@@ -1,5 +1,56 @@
...
+    }
+ 
++   /** @deprecated Forge: Use {@link net.minecraftforge.event.RegisterGameTestsEvent RegisterGameTestsEvent} to register game tests */
++   @Deprecated
+    public static void m_177503_(Method p_177504_) {
++      register(p_177504_, java.util.Set.of());
++   }
++   /** @deprecated Forge: Use {@link net.minecraftforge.event.RegisterGameTestsEvent RegisterGameTestsEvent} to register game tests */
++   @Deprecated
++   public static void register(Method p_177504_, java.util.Set<String> allowedNamespaces) {
```
pupnewfster, line 20: This method should be added to the forge.exc file

patches/minecraft/net/minecraft/gametest/framework/GameTestRegistry.java.patch
```diff
@@ -1,5 +1,56 @@
...
++   /** @deprecated Forge: Use {@link net.minecraftforge.event.RegisterGameTestsEvent RegisterGameTestsEvent} to register game tests */
++   @Deprecated
+    public static void m_177501_(Class<?> p_177502_) {
+-      Arrays.stream(p_177502_.getDeclaredMethods()).forEach(GameTestRegistry::m_177503_);
++      register(p_177502_, java.util.Set.of());
++   }
++   /** @deprecated Forge: Use {@link net.minecraftforge.event.RegisterGameTestsEvent RegisterGameTestsEvent} to register game tests */
++   @Deprecated
++   public static void register(Class<?> p_177502_, java.util.Set<String> allowedNamespaces) {
++      Arrays.stream(p_177502_.getDeclaredMethods()).forEach(m -> GameTestRegistry.register(m, allowedNamespaces));
```
pupnewfster: If the body of this helper gets moved to the event, then this method that is just deprecated does not need to be added via a patch.

src/main/java/net/minecraftforge/event/RegisterGameTestsEvent.java
```diff
@@ -0,0 +1,90 @@
...
+     * All methods annotated with {@link GameTest} or {@link GameTestGenerator} will be registered.
+     * If {@link #getEnabledNamespaces()} is non-empty,
+     * a method will only be registered if its {@link GameTest#templateNamespace() template namespace} is in an enabled namespace.
+     *
+     * @param testClass the test class to register to the game test registry
+     */
+    @SuppressWarnings("deprecation")
+    public void register(Class<?> testClass)
+    {
+        Arrays.stream(testClass.getDeclaredMethods()).forEach(m -> GameTestRegistry.register(m, this.getEnabledNamespaces()));
```
pupnewfster: If this just proxies the methods via `RegisterGameTestEvent#register(Method)` then you can remove the deprecation warning suppression (and even just use a method reference in the for each)

sciwhiz12 (changes requested):

fmlloader/src/main/java/net/minecraftforge/fml/loading/targets/ForgeGametestDevLaunchHandler.java
SizableShrimp: Other launch handlers follow the same format shown here.

Mysterious-Dev (approved):

SizableShrimp:
Opened on 1.18 with #8225

PR #8025, opened 2021-08-22
Remove UseHoeEvent and fix hoes to use BlockInteractEvent

SizableShrimp:
This PR gets rid of `UseHoeEvent` in favor of moving hoes to use `BlockToolInteractEvent` and calling `IForgeBlock#getToolModifiedState` like other tools.


src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -220,6 +222,30 @@ public static boolean isCorrectToolForDrops(@Nonnull BlockState state, @Nonnull
         return player.hasCorrectToolForDrops(state);
     }
 
+    public static Pair<Predicate<UseOnContext>, Consumer<UseOnContext>> getHoeTillable(UseOnContext context, Pair<Predicate<UseOnContext>, Consumer<UseOnContext>> pair)
```
diesieben07, line 225: I really don't like this.
In my opinion there are two cleaner solutions to this:

1. Add more parameters to `getToolModifiedState` so that it can construct the `UseOnContext` and use the native methods in `HoeItem` like it does for all other actions.
2. Change `getToolModifiedState` to take a `UseOnContext` instead of separate parameters (might be limiting, but a `UseOnContext` is just a POJO...) and then do the same as in 1.
3. Change the event from being cancellable to having a result so we can distinguish "cancel action" and "use vanilla". `getToolModifiedState` would need to have its return type altered as well then.

In my opinion 1 or 2 are the best options here.

diesieben07 (changes requested):

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -220,6 +222,30 @@ public static boolean isCorrectToolForDrops(@Nonnull BlockState state, @Nonnull
...
+        // Since we have a pair variable to default back to, we need to know whether the event was canceled or not.
+        // To do this, we directly call ForgeEventFactory#onToolUse and IForgeBlock#getToolModifiedState.
+        // This is so we can make sure to return null if the event was canceled.
+        BlockState eventState = ForgeEventFactory.onToolUse(state, level, pos, context.getPlayer(), context.getItemInHand(), ToolActions.HOE_DIG);
+        if (eventState == null) // It was canceled
+            return null;
+        BlockState modified;
+        // If the eventState is equal to state, then the event did not change the state, and we should get the tool modified state from the block
+        if (eventState == state) {
+            modified = state.getBlock().getToolModifiedState(state, level, pos, context.getPlayer(), context.getItemInHand(), ToolActions.HOE_DIG);
```
gigaherz: This should be its own HOE_TILL ToolAction, not HOE_DIG.
Tool right-click uses are different from their primary "left-click" dig use.

EDIT: To clarify, there's the commented out HOE_TILL in the ToolActions class that I left ready to be used for this.

gigaherz (changes requested):
I don't currently have a strong opinion wrt. what @diesieben07 said, but it probably also needs to be considered.

src/main/java/net/minecraftforge/common/ForgeHooks.java
diesieben07: Definitely: https://github.com/MinecraftForge/MinecraftForge/blob/bec9d6f53423af875c6f723f261065cffab8c81c/src/main/java/net/minecraftforge/common/ToolActions.java#L84-L87

sciwhiz12:
Hello! @SizableShrimp, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable. If we do not hear back from you after some time, this PR will be closed for inactivity.

SizableShrimp:
I do plan on updating this however the implementation might need an overhaul to clean up how it works.

SizableShrimp:
Superseded by #8503.

PR #8027, opened 2021-08-24
[1.17.1] add possibility to add recipes to the anvil via datapacks

Minecraftschurli:
based on #7865 
does currently not include a data generator


ChampionAsh5357:
Doesn't that PR you're basing it off of already implement anvil recipe types? Can you give a clearer definition of what makes yours different compared to their implementation?

Minecraftschurli:
the PR i'm basing on is for 1.16

ChampionAsh5357:
If that's the case, why not ask the original writer to port to 1.17 or wait until the 1.16 has been merged so there are no conflicts being porting versions. If you want to do it yourself, why not port both grindstone and anvil recipes instead of a partial port of an existing PR which may be merged?

ferriarnus:
I was/am planning on porting the PR to 1.17 once 1.16 is finished and approved. Currently there are still some small issues I need to deal with. If you have some remarks or suggestions, feel free to post them.

Minecraftschurli:
@ChampionAsh5357 
> If that's the case, why not ask the original writer to port to 1.17 or wait until the 1.16 has been merged so there are no conflicts being porting versions. If you want to do it yourself, why not port both grindstone and anvil recipes instead of a partial port of an existing PR which may be merged?

I actually did the grindstone too just not on the same branch and I haven't made the pr yet

@ferriarnus 
> I was/am planning on porting the PR to 1.17 once 1.16 is finished and approved. Currently there are still some small issues I need to deal with. If you have some remarks or suggestions, feel free to post them.

Ok I just did it because I am doing the same for the brewing stand and thought I'll just do it

I have some remarks on your pr and things that you could port over from my implementation
1. using the an intermediary interface for the recipe type instead of the implementing class
2. providing the name string for the anvil recipe
3. defining the xp reward for the grindstone recipe
4. defining the number of input items
5. moving the serializers to ForgeMod and use a DeferredRegister for them
6. using records for the recipes instead of simple classes (in 1.17)

ChampionAsh5357:
> 1. using the an intermediary interface for the recipe type instead of the implementing class
> 2. providing the name string for the anvil recipe
> 3. defining the xp reward for the grindstone recipe
> 4. defining the number of input items
> 5. moving the serializers to ForgeMod and use a DeferredRegister for them
> 6. using records for the recipes instead of simple classes

1. Why not just use `IRecipeType#register`? creating a new instance and then deferring to later is currently unnecessary. If so, you could use a `Lazy` for that.
2. Why call the field blacksmithing if it's name is anvil?
4. Can already be defined by an `NBTIngredient`.
6. From what I understand, implementations can't be records since the method names will be obfuscated.

Minecraftschurli:
1. the thing with `IRecipeType#register` idk
2. a mistake when changing the name
3. if you are referring to the number of items with NBTIngredient you are mistaken I checked the code and in the test method it checks for everything but count
4. if the records' fields are not what implement the methods from minecraft it should not be a problem

ChampionAsh5357:
1. It makes the most sense since creating the object separately is a bit pointless when the method already does that for you and registers it as well.
3. Then why not just implement a new ingredient instead that takes in a count? I don't see why the recipe itself has to specify anything in relation to count for input or output.
4. I believe you just described what you are doing.

Minecraftschurli:
2. because it needs to handle the decreasing of the stack not just the checking if it matches
3. yes basically

Minecraftschurli:
is this `Lazy<RecipeType<IBlacksmithingRecipe>> BLACKSMITHING = Lazy.of(() -> RecipeType.register("forge:blacksmithing"))` what you meant by using `Lazy` for the RecipeType

ChampionAsh5357:
> is this Lazy<RecipeType<IBlacksmithingRecipe>> BLACKSMITHING = Lazy.of(() -> RecipeType.register("forge:blacksmithing")) what you meant by using Lazy for the RecipeType

Yeah since that will defer registration until you want it and then afterwards be essentially treated the same as a registry object. You could also just leave it as static init for right now instead, doesn't matter.

> because it needs to handle the decreasing of the stack not just the checking if it matches

I'm aware. ~~Though I feel like I remember crafting recipes being able to take in a count somehow and then properly decrementing them with the ingredient. But I could be mistaken so I'm looking for it.~~ So `ResultSlot#onTake` does hardcode the shrinking to 1. Since anvils don't use a result slot, I would think it would be better in converting it there and then handling the logic from removing the data from the other two.

ferriarnus:
Since I'm here might as well give a reaction :

> I have some remarks on your pr and things that you could port over from my implementation
> 
> 1. using the an intermediary interface for the recipe type instead of the implementing class
> 2. providing the name string for the anvil recipe
> 3. defining the xp reward for the grindstone recipe
> 4. defining the number of input items
> 5. moving the serializers to ForgeMod and use a DeferredRegister for them
> 6. using records for the recipes instead of simple classes (in 1.17)

1) Was looking add the vanilla code and only the crafting table uses an interface. If you want a new recipe you could extend the class itself, but an interface may be more elegant.
2) What do you mean exactly? 
3) Didn't think about that, good idea!
4) That's the "issue" I was/am still struggling with. Still looking for an "elegant solution"
5) Didn't think about that, but also not really necessary, cause i can access them with "SERIALIZER". But a DeferredRegister might be better

PS: I'm also on discord if you want a discussion/point something out.

Minecraftschurli:
2 and 4 are answered by my code but for 2 i am using a wrapper around the container/inventory which stores the string currently present in the anvils text field which recipes can then use to check against or assemble the result. And 4 i am doing in my code i can send you the snippet if you want to (for my grindstone code look at the branch on my fork as i havent opened a pr yet)

Minecraftschurli:
For further things you can dm me on discord (same name as here)

ferriarnus:
> ... 2 i am using a wrapper around the container/inventory which stores the string currently present in the anvils text field which recipes can then use to check against or assemble the result...

Hmm, so the item name can be checked. A good idea, I do find it somewhat complex for the goal (but I also don't see another solution right away). I'm also using the anvil event to reduce the patches I need, but honestly, the event might not be necessary anymore with the new recipe type.

patches/minecraft/net/minecraft/world/inventory/AnvilMenu.java.patch
```diff
@@ -1,14 +1,63 @@
 --- a/net/minecraft/world/inventory/AnvilMenu.java
 +++ b/net/minecraft/world/inventory/AnvilMenu.java
-@@ -53,6 +_,8 @@
+@@ -31,6 +_,10 @@
+    private static final int f_150466_ = 1;
+    private static final int f_150467_ = 1;
+ 
++   @javax.annotation.Nullable
++   private net.minecraftforge.common.crafting.IBlacksmithingRecipe selectedRecipe;
++   private final net.minecraft.world.level.Level level;
```
ferriarnus, line 9: I don't think you need this. You can uses access instead to get the level.

Minecraftschurli: I actually just copied what the smithing table does

ferriarnus: Yeah I did that too, but than I got a comment to use this.

Minecraftschurli: well access is sometimes "empty" but the level field always contains a level

ChampionAsh5357: If access is empty, you are on the client as defined by vanilla, and you shouldn't be doing any logic on the client for containers. So that situation should be fine.

Minecraftschurli: well the way I am doing it is what vanilla does for the smithing table (where it also does it on the client)

ChampionAsh5357:
Closing this since it has been determined as out of scope for Forge. See Curle's explanation on the [brewing datapack PR](https://github.com/MinecraftForge/MinecraftForge/pull/7949#issuecomment-944211848) for more details.

PR #8029, opened 2021-08-27
Fix broken fmlonly installer and promote it to page gen

SizableShrimp:
This PR fixes the fmlonly installer being broken, and promotes fmlonly to page generation now.

Previously, attempting to use the fmlonly installer resulted in the following error:
```
Considering library net.minecraftforge:fmlonly:1.17.1-37.0.44:universal
  Downloading library from https://maven.creeperhost.net/net/minecraftforge/fmlonly/1.17.1-37.0.44/fmlonly-1.17.1-37.0.44-universal.jar
    Download failed: Checksum invalid, deleting file:
      Expected: 1fdc3cb626372bf39801caa7733fdbfafbe69089
      Actual:   d655d641d2ae545073ae8347b3f4be536a853bc1
```

This was due to the fact that the universal jar was being signed *after* the installerJson task queried the SHA1 of the output file. This issue did not appear in the forge subproject because it explicitly specified `installerJar.dependsOn 'signUniversalJar'`. However, this is technically still incorrect as `installerJson` itself needs to depend on the sign task to ensure the correct SHA1, and the sign jar task being run before installerJson is not
explicitly defined. Gradle actually warns about inexplicit task dependencies now in Gradle 7, which can be seen [in the Jenkins output here](https://jenkins.minecraftforge.net/job/forge/job/forge/job/1.17.x/44/console#gradle-task-99).

The Gradle wrapper has also been updated from the previous snapshot version to Gradle 7.2.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -41,6 +41,7 @@ pipeline {
                 }
                 script {
                     env.MYVERSION = sh(returnStdout: true, script: './gradlew :forge:properties -q | grep "^version:" | cut -d" " -f2').trim()
+                    env.FMLONLY_VERSION = sh(returnStdout: true, script: './gradlew :fmlonly:properties -q | grep "^version:" | cut -d" " -f2').trim()
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -76,6 +77,7 @@ pipeline {
             post {
                 success {
                     build job: 'filegenerator', parameters: [string(name: 'COMMAND', value: "promote net.minecraftforge:forge ${env.MYVERSION} latest")], propagate: false, wait: false
+                    build job: 'filegenerator', parameters: [string(name: 'COMMAND', value: "promote net.minecraftforge:fmlonly ${env.FMLONLY_VERSION} latest")], propagate: false, wait: false
```
LexManos, line 80: The proper way to do this would be to use our new helper, which pulls out the group/artifact as well. But this works for now.

PR #8038, opened 2021-08-29
Tooltip event rewrite

### LINKED ISSUE #7995: [1.17] Wrapped Item tooltips loses styling on split
**Minecraft Version:** 1.17.1

**Forge Version:** 37.0.22

**Steps to Reproduce:**
 1. Add a long tooltip to any item with some styling like color.
 2. Hover over that item
 3. See that it got broken into multiple lines and only the last line kept the styling

**Description of issue:**
If a tooltip componet gets split because of wrapping, it loses it style on all but the last line
image
The first line below the name of the item should also have been gray.

I am not sure if this is a issue in Forge or Minecraft itself

### LINKED ISSUE #8037: Extended bundle tooltip not rendered
**Minecraft Version:** 1.17.1
**Forge Version:** 37.0.45

**Steps to Reproduce:**
 1. `/give Dev minecraft:bundle`
 2. Add an item to the bundle
 3. Hover over the bundle

**Description of issue:**
No extended rendering for the bundle.
Wrong: 
image

Right: 
image

### LINKED ISSUE #8124: RenderTooltipEvent objects have an empty ItemStack when fired by recipe book tooltips
**Minecraft Version:** 1.17.1

**Forge Version:** 37.0.59

**Steps to Reproduce:**
 1. Subscribe to RenderTooltipEvent.PostBackground events in mod code
 2. Set a breakpoint in the subscribed event handler
 3. Launch the game in debug mode
 4. Open the crafting table's recipe book
 5. Mouse over a recipe button, triggering the breakpoint
 6. Observe that the result of the event's getStack() method is an empty item stack

**Description of issue:**
Mods are currently not able to rely on the presence of an ItemStack object when handling tooltip code for the recipe book.  This can prevent various tooltip-augmenting features from working as intended.

The ultimate cause of the issue appears to be in the RecipeBookPage#renderTooltip method.  Before calling renderComponentTooltip, the page must call GuiUtils#preItemToolTip with the ItemStack object returned by the hovered button's current recipe.  The recipe book page must then call GuiUtils#postItemToolTip after the call to renderComponentTooltip is complete.

diesieben07:
Fixes #8037. Fixes #7995. Fixes #8124.

Currently the Post events are removed, because I do not think they make sense with the changes to the vanilla system.


diesieben07:
I've added a pluggable system for custom ClientTooltipComponents as well, hooking into the hardcoded bundle tooltip code.

Mysterious-Dev (approved):

diesieben07:
Why does the RenderTooltipEvent.Pre allow changing the screen size? How is that helpful? If that functionality is removed and the changing of the font renderer is merged with the Colors event then we only need that one event, not both.

Crimix:
Is it possible this could fix the styling issue with wrapped tooltips from #7995?

diesieben07:
Unlikely, as the same underlying method is used. But I can look into that, if line wrapping is reimplemented.

diesieben07:
- GatherComponents event implement with ability to set maximum width
- Reimplement text splitting
- Pre event reimplemented with removed ability to set maximum width 
- Color event reimplemented

diesieben07:
@Crimix Consider yourself lucky, #7995 is fixed by this now:

Crimix:
Nice, good job @diesieben07 thank you 😄

JTK222:
I'd appreciate if it would be possible to add non text components, here's an example usage:
image

Possibly just a render component that allows you to specify a size which is considered when rendering the BG?

diesieben07:
@JTK222 A pluggable system for that is included with this PR, check the example mod.

JTK222:
Oh, perfect sorry I assumed this only works with Text Component, then this entire thing looks perfect from my point of view ^^

Mysterious-Dev (approved):

Mysterious-Dev:
Look good

patches/minecraft/net/minecraft/client/gui/screens/Screen.java.patch
```diff
@@ -1,68 +1,104 @@
...
  
     public void m_96617_(PoseStack p_96618_, List<? extends FormattedCharSequence> p_96619_, int p_96620_, int p_96621_) {
--      this.m_169383_(p_96618_, p_96619_.stream().map(ClientTooltipComponent::m_169948_).collect(Collectors.toList()), p_96620_, p_96621_);
-+      this.renderToolTip(p_96618_, p_96619_, p_96620_, p_96621_, f_96547_);
-+   }
-+   public void renderToolTip(PoseStack p_96618_, List<? extends FormattedCharSequence> p_96619_, int p_96620_, int p_96621_, Font font) {
-+      this.renderTooltipInternal(p_96618_, p_96619_.stream().map(ClientTooltipComponent::m_169948_).collect(Collectors.toList()), p_96620_, p_96621_, font);
+       this.m_169383_(p_96618_, p_96619_.stream().map(ClientTooltipComponent::m_169948_).collect(Collectors.toList()), p_96620_, p_96621_);
     }
++   public void renderToolTip(PoseStack p_96618_, List<? extends FormattedCharSequence> p_96619_, int p_96620_, int p_96621_, Font font) {
```
pupnewfster: This method should have proper param names instead of srg param names, as the entire body of it is forge added. Also this should probably be named `renderTooltip` to have consistent capitalization in the word `Tooltip` with all the other tooltip methods.

patches/minecraft/net/minecraft/client/gui/screens/Screen.java.patch
```diff
@@ -1,68 +1,104 @@
...
-+      this.renderComponentToolTip(p_96566_, this.m_96555_(p_96567_), p_96568_, p_96569_, (font == null ? this.f_96547_ : font));
-+      net.minecraftforge.fmlclient.gui.GuiUtils.postItemToolTip();
++      tooltipFont = net.minecraftforge.client.RenderProperties.get(p_96567_).getFont(p_96567_);
++      tooltipStack = p_96567_;
+       this.m_169388_(p_96566_, this.m_96555_(p_96567_), p_96567_.m_150921_(), p_96568_, p_96569_);
++      tooltipFont = null;
++      tooltipStack = ItemStack.f_41583_;
     }
  
++   public void renderTooltip(PoseStack p_169389_, List<Component> p_169390_, Optional<TooltipComponent> p_169391_, int p_169392_, int p_169393_, Font font) {
```
pupnewfster: This method should have proper param names instead of srg param names, as the entire body of it is forge added.

patches/minecraft/net/minecraft/client/gui/screens/Screen.java.patch
```diff
@@ -1,68 +1,104 @@
...
--      this.m_96617_(p_96603_, Arrays.asList(p_96604_.m_7532_()), p_96605_, p_96606_);
-+      this.m_96597_(p_96603_, Arrays.asList(p_96604_), p_96605_, p_96606_);
+@@ -183,20 +_,34 @@
     }
  
     public void m_96597_(PoseStack p_96598_, List<Component> p_96599_, int p_96600_, int p_96601_) {
 -      this.m_96617_(p_96598_, Lists.transform(p_96599_, Component::m_7532_), p_96600_, p_96601_);
-+      this.renderComponentToolTip(p_96598_, p_96599_, p_96600_, p_96601_, f_96547_);
++      List<ClientTooltipComponent> components = net.minecraftforge.client.MinecraftForgeClient.gatherTooltipComponents(this.tooltipStack, p_96599_, p_96600_, p_96601_, f_96543_, f_96544_, this.tooltipFont == null ? this.f_96547_ : this.tooltipFont);
++      this.m_169383_(p_96598_, components, p_96600_, p_96601_);
```
pupnewfster: Is there a reason these two lines are being patched in instead, rather than just having one line that makes use of the new renderComponentTooltip method that takes a Font?

patches/minecraft/net/minecraft/client/gui/screens/Screen.java.patch
```diff
@@ -1,68 +1,104 @@
...
 +   }
-+   public void renderComponentToolTip(PoseStack matrixStack, List<? extends net.minecraft.network.chat.FormattedText> tooltips, int mouseX, int mouseY, Font font) {
-+      net.minecraftforge.fmlclient.gui.GuiUtils.drawHoveringText(matrixStack, tooltips, mouseX, mouseY, f_96543_, f_96544_, -1, font);
++   public void renderComponentToolTip(PoseStack poseStack, List<? extends net.minecraft.network.chat.FormattedText> tooltips, int mouseX, int mouseY, Font font) {
```
pupnewfster: This should probably be named `renderComponentTooltip` to have consistent capitalization in the word `Tooltip` with all the other tooltip methods.

patches/minecraft/net/minecraft/client/gui/screens/Screen.java.patch
```diff
@@ -1,68 +1,104 @@
...
        if (!p_169385_.isEmpty()) {
++         net.minecraftforge.client.event.RenderTooltipEvent.Pre preEvent = net.minecraftforge.client.MinecraftForgeClient.preTooltipEvent(this.tooltipStack, p_169384_, p_169386_, p_169387_, f_96543_, f_96544_, this.tooltipFont == null ? this.f_96547_ : this.tooltipFont, p_169385_);
++         if (preEvent.isCanceled()) return;
           int i = 0;
           int j = p_169385_.size() == 1 ? -2 : 0;
  
           for(ClientTooltipComponent clienttooltipcomponent : p_169385_) {
 -            int k = clienttooltipcomponent.m_142069_(this.f_96547_);
-+            int k = clienttooltipcomponent.m_142069_(font);
++            int k = clienttooltipcomponent.m_142069_(f_96547_);
```
pupnewfster: Is this meant to be `event#getFontRenderer()`? If not, then adding back the `this.` before the passed variable would allow for this line of the patch to be removed, though my suspicion is it is meant to get it from the event.

patches/minecraft/net/minecraft/client/gui/screens/Screen.java.patch
```diff
@@ -1,68 +1,104 @@
...
+-         m_93123_(matrix4f, bufferbuilder, j2 - 3, k2 - 3, j2 + i + 3, k2 + j + 3, 400, -267386864, -267386864);
+-         m_93123_(matrix4f, bufferbuilder, j2 - 4, k2 - 3, j2 - 3, k2 + j + 3, 400, -267386864, -267386864);
+-         m_93123_(matrix4f, bufferbuilder, j2 + i + 3, k2 - 3, j2 + i + 4, k2 + j + 3, 400, -267386864, -267386864);
+-         m_93123_(matrix4f, bufferbuilder, j2 - 3, k2 - 3 + 1, j2 - 3 + 1, k2 + j + 3 - 1, 400, 1347420415, 1344798847);
+-         m_93123_(matrix4f, bufferbuilder, j2 + i + 2, k2 - 3 + 1, j2 + i + 3, k2 + j + 3 - 1, 400, 1347420415, 1344798847);
+-         m_93123_(matrix4f, bufferbuilder, j2 - 3, k2 - 3, j2 + i + 3, k2 - 3 + 1, 400, 1347420415, 1347420415);
+-         m_93123_(matrix4f, bufferbuilder, j2 - 3, k2 + j + 2, j2 + i + 3, k2 + j + 3, 400, 1344798847, 1344798847);
++         net.minecraftforge.client.event.RenderTooltipEvent.Color colorEvent = net.minecraftforge.client.MinecraftForgeClient.colorTooltipEvent(this.tooltipStack, p_169384_, p_169386_, p_169387_, preEvent.getFontRenderer(), p_169385_);
++         j2 = preEvent.getX();
++         k2 = preEvent.getY();
```
pupnewfster: I think moving these two lines above the firing of the color event would be a bit easier to read, in that it is more clear that it isn't grabbing them from the color event, and is mostly just here for reducing patch size as they are not needed earlier. Though part of me is wondering if the actual values should also be passed into the color tooltip event as well in case someone wants to make use of those values for calculating things but they ended up being changed by someone else in the pre event.

patches/minecraft/net/minecraft/client/gui/screens/Screen.java.patch
```diff
@@ -1,68 +1,104 @@
...
++         m_93123_(matrix4f, bufferbuilder, j2 - 3, k2 - 3, j2 + i + 3, k2 + j + 3, 400, colorEvent.getBackground(), colorEvent.getBackground());
++         m_93123_(matrix4f, bufferbuilder, j2 - 4, k2 - 3, j2 - 3, k2 + j + 3, 400, colorEvent.getBackground(), colorEvent.getBackground());
++         m_93123_(matrix4f, bufferbuilder, j2 + i + 3, k2 - 3, j2 + i + 4, k2 + j + 3, 400, colorEvent.getBackground(), colorEvent.getBackground());
++         m_93123_(matrix4f, bufferbuilder, j2 - 3, k2 - 3 + 1, j2 - 3 + 1, k2 + j + 3 - 1, 400, colorEvent.getBorderStart(), colorEvent.getBorderStart());
++         m_93123_(matrix4f, bufferbuilder, j2 + i + 2, k2 - 3 + 1, j2 + i + 3, k2 + j + 3 - 1, 400, colorEvent.getBorderStart(), colorEvent.getBorderStart());
++         m_93123_(matrix4f, bufferbuilder, j2 - 3, k2 - 3, j2 + i + 3, k2 - 3 + 1, 400, colorEvent.getBorderStart(), colorEvent.getBorderStart());
++         m_93123_(matrix4f, bufferbuilder, j2 - 3, k2 + j + 2, j2 + i + 3, k2 + j + 3, 400, colorEvent.getBorderStart(), colorEvent.getBorderStart());
```
pupnewfster: Are some of these meant to go between `getBorderStart` and `getBorderEnd`? Also given it is a gradient, would there be any benefit to allowing there to be two background colors so that people can make the background itself go between multiple colors.

src/main/java/net/minecraftforge/fmlclient/gui/GuiUtils.java
```diff
@@ -237,6 +237,7 @@ public static void drawTexturedModalRect(PoseStack matrixStack, int x, int y, in
      *
      * @param stack The stack for which a tooltip is about to be drawn.
      */
+    @Deprecated
```
pupnewfster: All these methods that are being deprecated potentially should have a TODO for 1.18 to remove them.

src/main/java/net/minecraftforge/client/event/RenderTooltipEvent.java
```diff
@@ -45,20 +50,29 @@
 {
     @Nonnull
     protected final ItemStack stack;
+    @Deprecated
```
pupnewfster: I feel like the things that are being deprecated in this class probably should have TODO's for 1.18 to remove them.

src/main/java/net/minecraftforge/client/event/RenderTooltipEvent.java
```diff
@@ -73,14 +87,21 @@ public ItemStack getStack()
     /**
      * The lines to be drawn. May change between {@link RenderTooltipEvent.Pre} and {@link RenderTooltipEvent.Post}.
      * 
-     * @return An <i>unmodifiable</i> list of strings. Use {@link ItemTooltipEvent} to modify tooltip text.
+     * @return An <i>unmodifiable</i> list of strings. Use {@link net.minecraftforge.event.entity.player.ItemTooltipEvent} to modify tooltip text.
+     * @deprecated use {@link #getComponents()}
```
pupnewfster, line 97: If this method is deprecated as indicated by the change in javadocs the method should probably actually be annotated with `@Deprecated` as well

src/main/java/net/minecraftforge/client/event/RenderTooltipEvent.java
```diff
@@ -73,14 +87,21 @@ public ItemStack getStack()
...
      */
     @Nonnull
     public List<? extends FormattedText> getLines()
     {
         return lines;
     }
 
+    @Nonnull
+    public List<ClientTooltipComponent> getComponents()
```
pupnewfster, line 112: It seems like this method should have a javadoc similar to what getLines had?

src/main/java/net/minecraftforge/client/event/RenderTooltipEvent.java
```diff
@@ -106,39 +127,108 @@ public int getY()
...
+        public GatherComponents(ItemStack stack, int screenWidth, int screenHeight, List<Either<FormattedText, TooltipComponent>> tooltipElements, int maxWidth)
+        {
+            this.stack = stack;
+            this.screenWidth = screenWidth;
+            this.screenHeight = screenHeight;
+            this.tooltipElements = tooltipElements;
+            this.maxWidth = maxWidth;
+        }
+
+        public ItemStack getStack()
```
pupnewfster, line 177: While this method and a bunch of the other methods being added are relatively self explanatory, it still may be worthwhile to have brief javadocs for them.

src/main/java/net/minecraftforge/client/event/RenderTooltipEvent.java
```diff
@@ -45,20 +50,29 @@
 {
     @Nonnull
     protected final ItemStack stack;
+    @Deprecated
     protected final List<? extends FormattedText> lines;
     protected final PoseStack matrixStack;
```
pupnewfster: Probably worth a TODO on this for 1.18 to rename it to be more in line with vanilla, as well as the getter for it.

src/main/java/net/minecraftforge/client/MinecraftForgeClient.java
```diff
@@ -176,4 +192,103 @@ public static ITextureAtlasSpriteLoader getTextureAtlasSpriteLoader(ResourceLoca
...
+        MinecraftForge.EVENT_BUS.post(colorEvent);
+        return colorEvent;
+    }
+
+    public static List<ClientTooltipComponent> gatherTooltipComponents(ItemStack stack, List<? extends FormattedText> textElements, int mouseX, int mouseY, int screenWidth, int screenHeight, Font font)
+    {
+        return gatherTooltipComponents(stack, textElements, Optional.empty(), mouseX, mouseY, screenWidth, screenHeight, font);
+    }
+
+    public static List<ClientTooltipComponent> gatherTooltipComponents(ItemStack stack, List<? extends FormattedText> textElements, Optional<TooltipComponent> itemComponent, int mouseX, int mouseY, int screenWidth, int screenHeight, Font font)
```
pupnewfster: Is it intentional that the `mouseY` parameter is not being used?

src/main/java/net/minecraftforge/client/MinecraftForgeClient.java
```diff
@@ -176,4 +192,103 @@ public static ITextureAtlasSpriteLoader getTextureAtlasSpriteLoader(ResourceLoca
...
+        {
+            return event.getTooltipElements().stream()
+                    .flatMap(either -> either.map(
+                            text -> font.split(text, tooltipTextWidthF).stream().map(ClientTooltipComponent::create),
+                            component -> Stream.of(ClientTooltipComponent.create(component))
+                    ))
+                    .toList();
+        }
+        else
+        {
```
pupnewfster: Given it is sort of hard to read already due to the multiple lines from streams and things, I feel like it would be easier to read if the else was just removed given the if above always returns so there isn't any case the that the if can be fallen through, and it would then allow for there to be slightly less indentation.

pupnewfster (commented):
Note: I haven't gotten a chance to actually test it yet, so I may end up having a couple more comments once I do, but for now added a bunch of comments related to what I saw while reviewing the code.

patches/minecraft/net/minecraft/client/gui/screens/Screen.java.patch
diesieben07: That would be a binary breaking change

patches/minecraft/net/minecraft/client/gui/screens/Screen.java.patch
diesieben07: Because that method takes a `List<FormattedText>` because of legacy reasons.

patches/minecraft/net/minecraft/client/gui/screens/Screen.java.patch
pupnewfster: Ah, good catch, reading patches of patches is a pain, and missed the fact that it was really just changing the body of the method.

patches/minecraft/net/minecraft/client/gui/screens/Screen.java.patch
pupnewfster: But it takes a list of `? extends FormattedText` so you could replace these two lines with I was pointing out with:
```java
renderComponentToolTip(p_96598_, p_96599_, p_96600_, p_96601_, this.tooltipFont == null ? this.font : this.tooltipFont);
```
and it should work just fine? Though on second thought maybe that doesn't quite make as much sense with how the font param would end up being/setting things.

diesieben07: Yeah, that'd be setting the font again which is why I didn't do this.

src/main/java/net/minecraftforge/client/MinecraftForgeClient.java
diesieben07: Yes, it is not needed for text wrapping. I'll remove it.

diesieben07:
Addressed your review @pupnewfster.

src/main/java/net/minecraftforge/client/event/RenderTooltipEvent.java
```diff
@@ -106,39 +133,117 @@ public int getY()
...
+        /**
+         * The elements to be rendered. These can be either formatted text or custom tooltip components.
+         * This list is modifiable.
+         */
+        public List<Either<FormattedText, TooltipComponent>> getTooltipElements()
+        {
+            return tooltipElements;
+        }
+
+        public int getMaxWidth()
```
pupnewfster, line 210: Probably worth a JavaDoc

src/main/java/net/minecraftforge/client/event/RenderTooltipEvent.java
```diff
@@ -106,39 +133,117 @@ public int getY()
...
      */
-    @net.minecraftforge.eventbus.api.Cancelable
+    @Cancelable
     public static class Pre extends RenderTooltipEvent
     {
         private int screenWidth;
         private int screenHeight;
         private int maxWidth;
 
+        @Deprecated
```
pupnewfster: These deprecated methods (and other nested event classes) should also probably have a TODO for removal in 1.18 given they aren't used any more and don't serve a purpose.

src/main/java/net/minecraftforge/client/event/RenderTooltipEvent.java
```diff
@@ -259,30 +372,61 @@ public PostText(@Nonnull ItemStack stack, @Nonnull List<? extends FormattedText>
...
+         * @deprecated use {@link #getBackgroundStart()} and {@link #getBackgroundEnd}
+         */
+        @Deprecated
         public int getBackground()
         {
-            return background;
+            return backgroundStart;
+        }
+
+        public int getBackgroundStart()
```
pupnewfster, line 426: I am not really sure if it serves much purpose but would it make sense to deprecate the `getOriginalBackground` method and add two corresponding ones for the original values? This may not make sense given currently the event only really allows for one original background so the values are assumed the same. (Putting the comment here as I can't add it down below where the existing method is)

pupnewfster (commented):
Couple more comments on the code, and also a few from when I was doing some testing.
1. In the test mod, is the changes to sticks, clocks, and apples supposed to be noticeable? Given I am not really seeing anything, but I may just be blind if it is a subtle change.
2. The text wrapping seems to have some issues in regards to if it properly stops itself from going off the screen or not.

src/main/java/net/minecraftforge/client/event/RenderTooltipEvent.java
sciwhiz12: Along with a `// TODO: Remove in 1.18`, it should also use the `forRemoval` value to more clearly indicate that it is deprecated for removal.

diesieben07:
> In the test mod, is the changes to sticks, clocks, and apples supposed to be noticeable? Given I am not really seeing anything, but I may just be blind if it is a subtle change.

The apple demonstrates custom background and border colors:


The stick demonstrates adding custom tooltips via the event:


The clock demonstrates a custom maximum width:

diesieben07:
Fixed the text wrapping and position bugs, fixed the missing deprecation comments and javadocs.

src/main/java/net/minecraftforge/client/event/RenderTooltipEvent.java
```diff
@@ -106,39 +133,120 @@ public int getY()
...
     {
         private int screenWidth;
         private int screenHeight;
         private int maxWidth;
 
+        @Deprecated(forRemoval = true) // TODO: remove in 1.18
         public Pre(@Nonnull ItemStack stack, @Nonnull List<? extends FormattedText> lines, PoseStack matrixStack, int x, int y, int screenWidth, int screenHeight, int maxWidth, @Nonnull Font fr)
         {
-            super(stack, lines, matrixStack, x, y, fr);
+            this(stack, matrixStack, x, y, screenWidth, screenHeight, fr, List.of());
```
pupnewfster: While it is a bit more of a mess, I feel like this should more or less be a copy of the new Pre constructor instead of calling this, so that it can then pass `lines` to the now deprecated `RenderTooltipEvent` super constructor, in case any mods are currently listening to the events and any other mods are still using/decide to use the GuiUtils methods for some reason.

src/main/java/net/minecraftforge/client/event/RenderTooltipEvent.java
```diff
@@ -259,30 +375,61 @@ public PostText(@Nonnull ItemStack stack, @Nonnull List<? extends FormattedText>
...
+        private int backgroundEnd;
         private int borderStart;
         private int borderEnd;
 
-        public Color(@Nonnull ItemStack stack, @Nonnull List<? extends FormattedText> textLines, PoseStack matrixStack, int x, int y, @Nonnull Font fr, int background, int borderStart,
-                int borderEnd)
+        public Color(@Nonnull ItemStack stack, @Nonnull List<? extends FormattedText> textLines, PoseStack matrixStack, int x, int y, @Nonnull Font fr, int background, int borderStart, int borderEnd)
         {
-            super(stack, textLines, matrixStack, x, y, fr);
+            this(stack, matrixStack, x, y, fr, background, borderStart, borderEnd, List.of());
```
pupnewfster: While it is a bit more of a mess, I feel like this should more or less be a copy of the new Color constructor instead of calling this, so that it can then pass `lines` to the now deprecated `RenderTooltipEvent` super constructor, in case any mods are currently listening to the events and any other mods are still using/decide to use the GuiUtils methods for some reason.

src/main/java/net/minecraftforge/client/event/RenderTooltipEvent.java
```diff
@@ -45,20 +50,29 @@
...
+    protected final PoseStack matrixStack; // TODO: rename in 1.18
     protected int x;
     protected int y;
     protected Font fr;
+    protected final List<ClientTooltipComponent> components;
 
+    @Deprecated(forRemoval = true) // TODO: remove in 1.18
     public RenderTooltipEvent(@Nonnull ItemStack stack, @Nonnull List<? extends FormattedText> lines, PoseStack matrixStack, int x, int y, @Nonnull Font fr)
+    {
+        this(stack, matrixStack, x, y, fr, List.of());
```
pupnewfster: While it is a bit more of a mess, I feel like this should more or less be a copy of the new RenderTooltipEvent constructor instead of calling this, which would allow the lines parameter to be properly assigned to the lines field. This way while usage of this is deprecated it at least has proper values for the deprecated variant.

pupnewfster (commented):
Looks good overall, but when I was double checking things noticed that there are currently a couple tiny binary breaks in regards to the now deprecated `RenderTooltipEvent.lines` field not being assigned when using the deprecated constructors that takes a list of lines instead of components.

tterrag1098:
Haven't had time to dig into details, but the broad strokes looks like a faithful port of the original structure. LGTM.

Mysterious-Dev:
This PR is still relevant ?

diesieben07:
Updated again:

- Resolved merge conflicts
- Moved event call hooks to `ForgeHooksClient`
- Fixed tooltip events firing without `ItemStack` info on creative menu search tab
- Streamlined font overriding handling via helper method in `ForgeHooksClient`

diesieben07:
Resolved merge conflicts again

PR #8039, opened 2021-08-31
Fix all layers of ItemLayerModel being fullbright

### LINKED ISSUE #8007: ItemLayerModel  renders all layer as fullbright
**Minecraft Version:** 1.17.1
**Forge Version:** 37.0.33 (also present in 36.2.0)

**Logs:** No logs

**Steps to Reproduce:**
 1. Create an item with a model using the `forge:item-layers` loader and multiple layers. Define one of the layers as fullbrught. The test item in Forge works for this case
 
https://github.com/MinecraftForge/MinecraftForge/blob/1.17.x/src/generated_test/resources/assets/new_model_loader_test/models/item/item_layers.json

 2. Launch and set the time to night

All layers are full brightness while only one layer should be. The image shows the item on the left and normal coal as reference on the right.

image

As far as I can tell the problem is in this line https://github.com/MinecraftForge/MinecraftForge/blob/1.17.x/src/main/java/net/minecraftforge/client/model/ItemLayerModel.java#L116

the last boolean parameter should be a check to see if the layer is fullbright or not instead of just always true

sciwhiz12:
This PR fixes #8007 by properly passing on the fullbrightness of a layer to the method which creates the quads for the sprite, as noted in the linked issue.

The issue seems to have been introduced in commit https://github.com/MinecraftForge/MinecraftForge/commit/ce3d8b40cf37924caf1708cdde6842ae6fdcee31#diff-faf15b5b7caaccd4c0f4a3c1aa1ca361092e2e0bb83577188760148de7f8bf5eR108 by gigaherz on Jul 3, 2020. This means that (as noted in the issue) this bug also occurs in 1.16.x. A separate PR will be made to address that once this one has been reviewed.

**Screenshot of the implemented fix:**

(_as compared with the photo from the linked issue:_)

### DIFF
--- a/src/main/java/net/minecraftforge/client/model/ItemLayerModel.java
+++ b/src/main/java/net/minecraftforge/client/model/ItemLayerModel.java
@@ -112,8 +112,9 @@ public final class ItemLayerModel implements IModelGeometry<ItemLayerModel>
         for(int i = 0; i < textures.size(); i++)
         {
             TextureAtlasSprite tas = spriteGetter.apply(textures.get(i));
-            RenderType rt = getLayerRenderType(fullbrightLayers.contains(i));
-            builder.addQuads(rt, getQuadsForSprite(i, tas, transform, true));
+            boolean fullbright = fullbrightLayers.contains(i);
+            RenderType rt = getLayerRenderType(fullbright);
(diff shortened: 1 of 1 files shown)


gigaherz:
Oh... 🤦‍♂️

PR #8042, opened 2021-08-31
Re-add patch to release buffer in custom payload packet

### LINKED ISSUE #7486: SCustomPayloadPlayPacket Buffer Leak Regression
**Minecraft Version:** 1.13 - 1.16

**Forge Version:** 25.0 - 35.0.15

**Logs:** [Output of Netty Leak Detection](https://gist.github.com/pau101/e5af99da06886e336e80ee83083aadb6)

**Steps to Reproduce:**

1. Launch with JVM arg `-Dio.netty.leakDetection.level=advanced`
2. Join a multiplayer server

**Description:**

In #4512 two patches were added to resolve [MC-121884](https://bugs.mojang.com/browse/MC-121884), one in `SCustomPayloadPlayPacket` (then `SPacketCustomPayload`) to add the buffer release, and one in `ClientPlayNetHandler` (then `NetHandlerPlayClient`) to remove an arbitrary release for "MC|TrList". In https://github.com/MinecraftForge/MinecraftForge/commit/bf9158172fbb5123cd4eae59a5c65c2ed56ed30c the second patch became misapplied to new closeable of copied buffer. While the former patch was never reinstated leading to the memory leak still persisting.

To resolve this issue the packet buffer release patch would need to be reinstated while the current erroneous patch removed.

sciwhiz12:
This PR fixes #7486 by readding the patch from #4512 to release the buffer in `SCustomPayloadPacket` -- see https://github.com/MinecraftForge/MinecraftForge/commit/baaa6c65056ba71fe5e3d85e60ebb6ddbfc5370c#diff-0f241acf691432d38feb5d53220be71e25f194858ac81a938263fc107c59eae4R17-R18.
A separate commit also removes another patch which prevented the copy of the custom payload packet's data buffer from being released properly.

To paraphrase @gigaherz's [analysis in a conversation we had](https://discord.com/channels/313125603924639766/852298000042164244/882232990464372738) in the discord about this:
> So it seem that deep deep down, Netty uses `alloc.ioBuffer`, which can either use a direct buffer or a heap buffer depending on how that shit is set up.
> On LocalChannels (singleplayer), it seems to use a prefer-heap allocator. On multiplayer, it probably uses direct buffers though.
> So it could be the case that in multiplayer/LAN connections we have a real memory leak but in singleplayer it's only a counting leak.

This issue also exists in 1.16, and a backport PR will be made to address that once this one has been reviewed.

### DIFF
--- a/patches/minecraft/net/minecraft/client/multiplayer/ClientPacketListener.java.patch
+++ b/patches/minecraft/net/minecraft/client/multiplayer/ClientPacketListener.java.patch
@@ -85,7 +85,7 @@
           this.f_104895_ = tagcontainer;
           if (!this.f_104885_.m_129531_()) {
              tagcontainer.m_13431_();
-@@ -1780,10 +_,12 @@
+@@ -1780,7 +_,8 @@
              int i5 = friendlybytebuf.m_130242_();
              this.f_104888_.f_91064_.f_173815_.m_173830_(positionsource, i5);
           } else {
@@ -95,7 +95,3 @@
           }
        } finally {
           if (friendlybytebuf != null) {
-+            if (false) // Forge: let packet handle releasing buffer
-             friendlybytebuf.release();
-          }
- 
--- a/patches/minecraft/net/minecraft/network/protocol/game/ClientboundCustomPayloadPacket.java.patch
+++ b/patches/minecraft/net/minecraft/network/protocol/game/ClientboundCustomPayloadPacket.java.patch
@@ -9,3 +9,11 @@
     private static final int f_178834_ = 1048576;
     public static final ResourceLocation f_132012_ = new ResourceLocation("brand");
     public static final ResourceLocation f_132013_ = new ResourceLocation("debug/path");
+@@ -52,6 +_,7 @@
+ 
+    public void m_5797_(ClientGamePacketListener p_132041_) {
+       p_132041_.m_7413_(this);
++      this.f_132030_.release(); // FORGE: Fix network buffer leaks (MC-121884)
+    }
+ 
+    public ResourceLocation m_132042_() {


gigaherz:
I can confirm after testing that in a multiplayer connection it does in fact use a DirectByteBuf:

image

Single-player uses Heap buffer but netty still does reference counting and heap size counting so while those leaks wouldn't be real leaks, they would still cause netty to erroneously think those buffers still exist, and never subtract the heap usage from the counter.

ichttt:
This looks wrong. The bug has been fixed in vanilla, our patch is just wrong. It was added in https://github.com/MinecraftForge/MinecraftForge/pull/4512/files, as in 1.12.2, only one branch release the packet while all the others didn't release it. So it made sense to dummy out the one release and release them in a global spot. But now there is a giant try finally around the whole thing, so the vanilla release is fine, the problem is just that we kept the patch that removes the release and removed the patch that adds the release in a different place.
Long story short: The second commit is fine, but the first is wrong as it would release the buffer twice if I see everything correctly. Didn't test it though.

sciwhiz12:
@ichttt: I've looked at the 1.12.2 source code, and I've found out the following differences and observations:

* In 1.12, the method which returns the buffer containing the packet's data returned the actual buffer stored in the packet object, while in 1.17 that method now returns a new buffer containing a copy of the data from the actual packet object's buffer.
* As noted by you and linked the issue, the patch in question removes in 1.12 a buffer release done when the packet channel is `MC|TrList`, which is not done for any other packet channel. However, in 1.17:
  * The patch in 1.17 now prevents the copied buffer from being released in _any_ case (as the `try`-`catch` block covers all of the channel switching and handling, instead of only one specific channel); and
  * Because the buffer passed to the client packet play listener (and the one which is released within the packet listener) is only the copied buffer, the original buffer in the actual packet object remains unreleased.

This means that both commits are correct: the first commit patches the packet buffer to release the actual buffer stored in the packet object, while the second commit removes a patch which prevents the copied buffer from being released within the client packet listener. The buffers in both places are different, the one in the packet listener having a copy of the data within the former.

ichttt:
Whoops, my bad, you seem to be correct. I didn't assume getData would copy the data...

PR #8044, opened 2021-08-31
[1.17.1] Add ForgeSpawnEggItem to lazily handle EntityTypes

XFactHD:
This PR supercedes #7621.

This PR proposes a solution to the problem of SpawnEggItems requiring you to either pass in an EntityType that was initialized too early and therefore breaks registry contracts or pass in null and pollute the map with a useless null key which indirectly leads to problems later on, for example in the `ItemColors` registration where only the last modded egg registered this way would get tinting.

### Proposed Solution

The proposed solution is to store all ForgeSpawnEggItems added by mods in a list and transfer them over to the EntityType->SpawnEggItem map after EntityTypes have been registered. The original `SpawnEggItem` is patched to deprecate the constructor and guard against adding a null key to the `BY_ID` map.
The `BY_ID` map in question is used in four places:

1. **`IForgeEntity#getPicketResult(HitResult target)` via `SpawnEggItem.byId(EntityType type)`**
This is only called during gameplay which means the transfer to the map has already happened

2. **`ItemColors.createDefault()` via `SpawnEggItem.eggs()`**
Even though this happens before EntityTypes are registered, it doesn't create any problems related to them because the EntityType is not used there. To include the mod eggs, the call site is patched to instead call `ForgeSpawnEggItem#eggs()` instead which concatenates the `Iterable` from `SpawnEggItem.eggs()` with the List of modded eggs in `ForgeSpawnEggItem#MOD_EGGS`. Due to the timing of this, a mod can easily override the `ItemColors` in `ColorHandlerEvent.Item` if they wish to do so.
    
3. **`BlockModelGenerators#run()` via `SpawnEggItem.eggs()`**
Doesn't matter to us because modders should not be using the vanilla `BlockModelGenerators` in datagen

4. **`DispenseItemBehaviour.bootStrap()` via `SpawnEggItems.eggs()`**
This can be ignored for this purpose because I have found in my testing with breakpoints that the call to DispenseItemBehaviour.bootStrap() happens before `RegistryEvent<Item>` has fired which means that mod eggs are not included anyway. To make sure mod eggs get `DispenseItemBehaviour`s, they are registered in `FMLCommonSetup` at the same time the mod eggs are transferred to the type->egg map. If the modder wishes they can provide their own `DispenseItemBehaviour` or none via `ForgeSpawnEggItem#createDispenseBehavior()`.

Not adding any more ways to deviate from the default behaviour of the vanilla `SpawnEggItem` is very much intentional.
If you want your spawn eggs to behave differently to the vanilla spawn eggs, then you should IMO implement them yourself.

### Test

To test this I have included a test mod that adds a test entity and a spawn egg for that entity. In my testing spawning by hand, the coloring of the item, the dispense behaviour and "picking" the entity to get the egg for it worked flawlessly.


AnonymousHacker1279:
This would certainly fix the whole spawn egg frustration. It would also keep [code like this](https://github.com/AnonymousHacker1279/ImmersiveWeapons/blob/master/src/main/java/com/anonymoushacker1279/immersiveweapons/item/utility/CustomSpawnEggItem.java) from existing in mods.

ShroomAgent27 (approved):
I wrote an [example mod](https://github.com/FoalFreePress/TestingForgeMods/tree/d6d4bc1e205f0bd6f72c515f7885380dbb216563/8044/src/main/java/com/example/examplemod) that included several different new entities (based on 4 existing ones.) All the eggs were added correctly and they all behaved as I would expect spawn eggs to behave.  Pick block works, dispensers work, and there are no errors in the debug log that suggest any issues going on in terms of registration of items or entities from what I can tell. DataGen works as expected when using items of `ForgeSpawnEggItem`.

All of the patch and new files look to be formatted correctly as well.

src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
```diff
@@ -0,0 +1,104 @@
...
+                    LOGGER.error("Error while dispensing spawn egg from dispenser at {}", source.getPos(), exception);
+                    return ItemStack.EMPTY;
+                }
+
+                stack.shrink(1);
+                source.getLevel().gameEvent(GameEvent.ENTITY_PLACE, source.getPos());
+                return stack;
+            }
+        };
+    }
```
ShroomAgent27: This could be a Singleton instance of `DefaultDispenseItemBehavior` unless a mod specifies a different `DispenseItemBehavior`.

XFactHD: Moved the default behavior to a static field

src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @Nullable
+    protected DispenseItemBehavior createDispenseBehavior()
+    {
+        return DEFAULT_DISPENSE_BEHAVIOR;
+    }
+
+    public static Iterable<SpawnEggItem> eggs()
+    {
+        return Iterables.concat(SpawnEggItem.eggs(), MOD_EGGS);
```
LexManos: This is going to result in many duplicates, depending on if called before or after finalize. Better to call distinct.
Or, we could not use our own map, and just BY_ID as our map. Or, we could not merge things at all.
There would be two places where we would care about it.
IForgeEntity.getPickedResult and Mob.getPickedResult, both of which are fine with null, so you should change it to checking both maps.

patches/minecraft/net/minecraft/client/color/item/ItemColors.java.patch
```diff
@@ -10,6 +10,15 @@
  
     public static ItemColors m_92683_(BlockColors p_92684_) {
        ItemColors itemcolors = new ItemColors();
+@@ -68,7 +_,7 @@
+          return p_92700_ > 0 ? -1 : PotionUtils.m_43575_(p_92699_);
+       }, Items.f_42589_, Items.f_42736_, Items.f_42739_);
+ 
+-      for(SpawnEggItem spawneggitem : SpawnEggItem.m_43233_()) {
++      for(SpawnEggItem spawneggitem : net.minecraftforge.common.ForgeSpawnEggItem.eggs()) {
+          itemcolors.m_92689_((p_92681_, p_92682_) -> {
```
LexManos: Instead of patching this, you should just use the event we have already, and leave the vanilla path to itself.

LexManos (commented):
I am not seeing why anything needs a patch for this to function.
It's a simple utility class that handles registering a color and a dispense behavior.
All of which can be done at the correct time using the events we have.
The only thing I can see that this adds is a sane injection to the BY_ID map. Which is only used in getPickedEntity. Which can be overridden by your custom entity.

So, what does this add that needs to be added?

patches/minecraft/net/minecraft/client/color/item/ItemColors.java.patch
XFactHD: Should I move this to a Forge internal handler with `HIGHEST` priority to make sure mods can still override the color handler with their own or leave it to the mod entirely?

LexManos: I am not a fan of helper methods for no good reason. So I'm more leaning twards let the modders do it.
However I know people are lazy about it. So yes, move it to an event handler.
Honestly this can all be wrapped in one class, with no access or patch changes.
And one line edited in IForgeEntity.getPickedResult.

src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
XFactHD: The picked results don't care about this specific method at all.
If I move the color registration to an event, then I could remove the concatanation and just return the `MOD_EGGS` list because the only use for that list outside of the `ForgeSpawnEgg` class is the color handler registration.

patches/minecraft/net/minecraft/client/color/item/ItemColors.java.patch
XFactHD: Using our own map, not moving stuff to the `BY_ID` map and changing `IForgeEntity.getPickedResult()` would still result in having to use the additional list and moving stuff over at a certain point.

Removing the deprecation patch to `SpawnEggItem` would be counterproductive IMO because no one would care about the replacement because there is no IDE warning that annoys them.
The other patch line in `SpawnEggItem` is intended to avoid adding a null key to the `BY_ID` map which would lead to a random spawn egg from a random mod being duplicated and getting a color handler registered for it twice.

LexManos: >Using our own map, not moving stuff to the BY_ID map and changing IForgeEntity.getPickedResult() would still result in having to use the additional list and moving stuff over at a certain point.

Why?

XFactHD: Because we can't build a type->item map when the items are constructed, that must always happen after entity type registration. Building a reverse map instead is pointless because we would need to iterate all items until we find the correct one, might as well only use a list in that case.

XFactHD: Moved the color handler registration to an event handler

src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
XFactHD: This method now only returns eggs registered by mods and is used by the color handler registration

XFactHD:
> I am not seeing why anything needs a patch for this to function.
> It's a simple utility class that handles registering a color and a dispense behavior.
> All of which can be done at the correct time using the events we have.
> The only thing I can see that this adds is a sane injection to the BY_ID map. Which is only used in getPickedEntity. Which can be overridden by your custom entity.
> 
> So, what does this add that needs to be added?

As you said it is simply a utility class for convenience to get people to stop complaining about spawn eggs being shit.
I moved the color registry part from a patch to an event handler as suggested. The `SpawnEggItem` patch is mostly there to make people aware of the replacement and also to prevent adding a `null` key to the map.

src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
```diff
@@ -0,0 +1,104 @@
...
+    {
+        MOD_EGGS.forEach(egg ->
+        {
+            DispenseItemBehavior dispenseBehavior = egg.createDispenseBehavior();
+            if (dispenseBehavior != null)
+            {
+                DispenserBlock.registerBehavior(egg, dispenseBehavior);
+            }
+
+            BY_ID.put(egg.typeSupplier.get(), egg);
```
LexManos: Remove

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -251,6 +251,7 @@ public net.minecraft.world.item.HoeItem <init>(Lnet/minecraft/world/item/Tier;IF
 public net.minecraft.world.item.PickaxeItem <init>(Lnet/minecraft/world/item/Tier;IFLnet/minecraft/world/item/Item$Properties;)V # constructor
 public net.minecraft.world.item.RecordItem <init>(ILnet/minecraft/sounds/SoundEvent;Lnet/minecraft/world/item/Item$Properties;)V # constructor
 #endgroup
+protected net.minecraft.world.item.SpawnEggItem f_43201_ # BY_ID
```
LexManos: Remove

Shadows-of-Fire:
Why is it desirable to have two separate maps for this information?

src/main/java/net/minecraftforge/common/extensions/IForgeEntity.java
```diff
@@ -88,6 +89,9 @@ default ItemStack getPickedResult(HitResult target)
         ItemStack result = self().getPickResult();
         if (result == null) {
             SpawnEggItem egg = SpawnEggItem.byId(self().getType());
+            if (egg == null) {
+                egg = ForgeSpawnEggItem.fromEntityType(self().getType());
+            }
```
diesieben07: Personally I think this should be patched into `SpawnEggItem.byId` so that mods that call that method continue to function.

src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Nullable
+    public static ForgeSpawnEggItem fromEntityType(@Nullable EntityType<?> type)
+    {
+        return TYPE_MAP.get(type);
+    }
+
+    public static Iterable<ForgeSpawnEggItem> getModEggs()
+    {
+        return MOD_EGGS;
```
diesieben07: This should return an unmodifiable view, if anything. Why is this method necessary?

src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public static void finalizeModEggs()
+    {
+        MOD_EGGS.forEach(egg ->
```
diesieben07: Why is this field necessary at all? Can't we use the registry?

diesieben07 (changes requested):

src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
XFactHD: It is needed for the color handler registration

src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
XFactHD: In theory I could stream and filter the item registry for all forge eggs and remove the field entirely

src/main/java/net/minecraftforge/common/extensions/IForgeEntity.java
XFactHD: Lex wanted this to be done with the least possible amount of patches which is why I did it this way. In terms of only needing to call one method instead of two to cover all eggs, I suggested calling `SpawnEggItem.byId()` from `ForgeSpawnEggItem.fromEntityType()` on Discord in #issues-and-prs.

diesieben07: Then we should at least have a utility method somewhere that does this and SpawnEggItem.byId should be deprecated - because it is not accurate.

XFactHD:
> 
> 
> Why is it desirable to have two separate maps for this information?

Because Lex requested it to be done that way. See the resolved conversations below his first comment and the conversation about an hour ago in #issues-and-prs on Discord

src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
diesieben07: Should be an unmodifiable view then.

src/main/java/net/minecraftforge/common/extensions/IForgeEntity.java
XFactHD: I would argue that `ForgeSpawnEggItem.fromEntityType()` is that utility method when the additional call to `SpawnEggItem.byId()` is added to it.

src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
LexManos: This wouldnt need to be public at all of the event handler was in this class.

src/main/java/net/minecraftforge/common/extensions/IForgeEntity.java
diesieben07: Add a utility method that does calls vanilla's byId and falls back on Forge's method if that is not empty, so we have a "entity type -> spawn egg" conversion. You can use this method from here.

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -1001,5 +1003,13 @@ public static void registerShaders(RegisterShadersEvent event) throws IOExceptio
                 rendertypeEntityTranslucentUnlitShader = p_172645_;
             });
         }
+
```
LexManos: Move to ForgeSpawnEggItem, private method, so that getModEggs() does not have to be public.

src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
```diff
@@ -0,0 +1,113 @@
...
+    @Nullable
+    protected DispenseItemBehavior createDispenseBehavior()
+    {
+        return DEFAULT_DISPENSE_BEHAVIOR;
+    }
+
+    @Nullable
+    public static ForgeSpawnEggItem fromEntityType(@Nullable EntityType<?> type)
+    {
+        return TYPE_MAP.get(type);
```
LexManos: To make people shut up about having to use two methods.
```
SpawnEggItem ret = TYPE_MAP.get(type);
return ret != null ? ret : SpawnEggItem.getByType(type);
```

src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public static Iterable<ForgeSpawnEggItem> getModEggs()
+    {
```
LexManos: Remove, as you can access the field in the private event handler I told you to move here.
This way there is no public API to argue. There is no good use for itterating all the eggs.

src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
```diff
@@ -0,0 +1,113 @@
...
+        return MOD_EGGS;
+    }
+
+    public static void finalizeModEggs()
```
LexManos: This method should either be commented as 'For internal use only'
Or moved to a private method that is auto resisted using @EventBusSubscriber.
So that it doesn't exist from a API standpoint.

src/main/java/net/minecraftforge/common/extensions/IForgeEntity.java
```diff
@@ -88,6 +89,9 @@ default ItemStack getPickedResult(HitResult target)
         ItemStack result = self().getPickResult();
         if (result == null) {
             SpawnEggItem egg = SpawnEggItem.byId(self().getType());
+            if (egg == null) {
```
LexManos: Honestly I would prefer this didn't exist at all, because modders *should* be implementing getPickResult themselves and this making this moot. However, once you change ForgeSpawnEggItem.get to call vanilla's then this can be changed to a single call.

LexManos (changes requested):
The entire point im trying to get at, is that this is a simple utility class. It can be contained in a single file.
There is no logical reason, besides to make people stop complaints, for this to be added to Forge.
Does not need to have any patches {yes having a deprecated marker is nice, not nessasary, but i'll accept it}.
The only reason you need a get(Type) function is for the default implementation of entity.getPicked.
But that can be done by the modder implementing getPicked themselves in their entity.
The colors are also a non-issue, as they can register their own colors using our color event.
The dispenser handler again, can be done themselves by registering the handler in their registry event.

XFactHD:
> The entire point im trying to get at, is that this is a simple utility class. It can be contained in a single file. There is no logical reason, besides to make people stop complaints, for this to be added to Forge. Does not need to have any patches {yes having a deprecated marker is nice, not nessasary, but i'll accept it}. The only reason you need a get(Type) function is for the default implementation of entity.getPicked. But that can be done by the modder implementing getPicked themselves in their entity. The colors are also a non-issue, as they can register their own colors using our color event. The dispenser handler again, can be done themselves by registering the handler in their registry event.

I am aware that this whole thing can be implemented by a mod without changes in Forge but people prefer to complain that the existing `SpawnEggItem` class is not viable instead of implementing this themselves. The same would be true if this is added but the picking change, dispenser behaviour registration and color handler registration where left to be implemented by the modder.

diesieben07 (approved):

LexManos (approved):

PR #8046, opened 2021-09-01
[1.16.5] Update gradle wrapper

jriwanek:
As requested by Curle on Discord.

PR #8047, opened 2021-09-01
[1.17.1] Update gradle wrapper

jriwanek:
As requested by Curle on Discord.


marchermans:
Why?
And without description this is never getting merged.

dpeter99:
This was indeed suggested by Curle on discord: [Message](https://discord.com/channels/313125603924639766/852298000042164244/882425229312422039)
Same goes for the LTS PR #8046

TheCurle:
As some context- Jri was working on a docker image with a stable version of Gradle inside - and to make it easier by way of allowing the current latest version to be used without chance of issue, i suggested these wrappers be updated.

However, it may be better to include this bump in some other unrelated PR so that we don't have a build without significant change to modders or end-users.

SizableShrimp:
I currently have this included alongside #7976

LexManos:
For note: No we do not care about random 3rd parties making docker containers. The wrapper species the version of gradle that we use.
If you wish to include a gradle install in their docker, they can simply not call the wrapper scripts. Us updating this script should have no effect on whatever they do.
Also, if this is indeed the output of `gradlew wrapper` then I have no idea why it wasnt included in the commit that actually updated the wrapper version. As this does not do that, it just edits the linux script. Either way not important until next time we update the wrapper version.

PR #8048, opened 2021-09-01
RenderGuiItemDecorations event

lobbienjonsji:
This PR adds two new hooks to the renderGuiItemDecorations method in the ItemRenderer class.
One before all the Decoration Elements have been rendered, which is cancellable and one after all Decoration Elements have been rendered, giving read access to all parameter of the method call. 

A test mod for both hooks is provided.

The idea of the event is that it allows modders to add additional overlays to items in a non item specific way.
For example, an enchantment could add a counter to the enchanted item and that counter would appear on all
items with said enchantment.


patches/minecraft/net/minecraft/client/renderer/entity/ItemRenderer.java.patch
```diff
@@ -57,6 +57,14 @@
              crashreportcategory.m_128165_("Item Damage", () -> {
                 return String.valueOf(p_174237_.m_41773_());
              });
+@@ -323,6 +_,7 @@
+ 
+    public void m_115174_(Font p_115175_, ItemStack p_115176_, int p_115177_, int p_115178_, @Nullable String p_115179_) {
+       if (!p_115176_.m_41619_()) {
++         if(net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.client.event.RenderGuiItemDecorationsEvent.Pre(p_115175_, p_115176_, p_115177_, p_115178_, p_115179_))) return;
```
marchermans: Create a method in ForgeHooksClient, instead of constructing the event and firing it yourself.

patches/minecraft/net/minecraft/client/renderer/entity/ItemRenderer.java.patch
```diff
@@ -76,6 +84,15 @@
              this.m_115152_(bufferbuilder, p_115177_ + 2, p_115178_ + 13, 13, 2, 0, 0, 0, 255);
              this.m_115152_(bufferbuilder, p_115177_ + 2, p_115178_ + 13, i, 1, j >> 16 & 255, j >> 8 & 255, j & 255, 255);
              RenderSystem.m_69478_();
+@@ -360,7 +_,7 @@
+             RenderSystem.m_69493_();
+             RenderSystem.m_69482_();
+          }
+-
++         net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.client.event.RenderGuiItemDecorationsEvent.Post(p_115175_, p_115176_, p_115177_, p_115178_, p_115179_));
```
marchermans: Also here. Use a ForgeHooksClient callback,

marchermans (changes requested):

patches/minecraft/net/minecraft/client/renderer/entity/ItemRenderer.java.patch
sciwhiz12: Also, just to shorten the patch a bit, instead of adding content to the empty line, add the new code line before or after the empty line.

lobbienjonsji:
Alright, I'll do that

ChampionAsh5357:
This seems to be subject to the same concerns as #7693. While although less prevalent, can you provide some numbers on the impact of game performance this may cause?

diesieben07:
As an alternative to the event, we could add a capability that mods can add to either their own or other ItemStacks which would then be called to do this rendering. That way no event dispatching needs to happen but the system is still pluggable even for items from other mods or vanilla.

lobbienjonsji:
I understand the concerns, I'll try implementing it in a way where the render methods only get called if it is requested for a specific itemStack by a specific mod, so probably with capabilities.

lobbienjonsji:
Will make a new Pull request since I took a different approach thanks to diesiebens suggestion

PR #8055, opened 2021-09-01
Add new tool action for shield blocking, replacing `IForgeItem#isShield`

### LINKED ISSUE #8054: Custom shields held by zombies in mainhand instead of offhand
**Minecraft Version:** 1.17.1
**Forge Version:** 37.0.50 (commit 19f470ee8cb1907a9b7e19edf4c548ec2fb3afbd, issue reproduction done in forgedev)
**Logs:** Not applicable

**Steps to Reproduce:**
 1. Install or create a mod which adds a custom shield item, that which overrides `IForgeItem#isShield` (for convenience, a test mod is provided below)
 1. Summon a zombie (`/summon` or a spawn egg) 
 1. Use the following command to modify the (nearest) zombie such that their `CanPickUpLoot` NBT tag is set to `1b`:
    `/data merge entity @e[type=zombie,sort=nearest,limit=1] {CanPickUpLoot:1b}`
 1. Throw the custom shield itme near the zombie, such that it picks it up
    _Observe as the zombie holds the custom shield item in the main hand instead of the offhand._
 1. Repeat the above steps, but this time with a regular, vanilla shield item.
    _Observe as the second zombie holds the vanilla shield in the offhand._
 1. Repeat the above steps in 1.16.5, both with a custom shield and a vanilla shield
    _Observe as the two zombies in 1.16.5 both hold the shield in the offhand._

**Description of issue:**
When a zombie (or any other entity which may pick up items off of the ground) picks up a custom shield item -- an item which overrides `IForgeItem#isShield` to return `true` -- the zombie holds the custom shield in the main hand instead of the offhand.

From my investigation, this seems to be this missing patch for `LivingEntity#getEquipmentSlotForItem` (which in 1.16.5 was `MobEntity#getEquipmentSlotForItem`):
https://github.com/MinecraftForge/MinecraftForge/blob/a95b968c1397a1597d4a6228d8e1abce6495570e/patches/minecraft/net/minecraft/entity/MobEntity.java.patch#L44-L52

**Screenshot of the issue**: (the middle zombie holds the custom shield item)

**Test mod**: _(place at `src/test/java/net/minecraftforge/debug/item`, add appropriate entry to `mods.toml`)_
<details>
<summary>For 1.17</summary>

```java
package net.minecraftforge.debug.item;

import net.minecraft.world.InteractionHand;
import net.minecraft.world.InteractionResultHolder;
import net.minecraft.world.entity.LivingEntity;
import net.minecraft.world.entity.player.Player;
import net.minecraft.world.item.CreativeModeTab;
import net.minecraft.world.item.Item;
import net.minecraft.world.item.ItemStack;
import net.minecraft.world.item.UseAnim;
import net.minecraft.world.level.Level;
import net.minecraftforge.fml.common.Mod;
import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
import net.minecraftforge.fmllegacy.RegistryObject;
import net.minecraftforge.registries.DeferredRegister;

@Mod("custom_shield_test")
public class CustomShieldTest {
    private static final DeferredRegister<Item> ITEMS = DeferredRegister.create(Item.class, "custom_shield_test");

    private static final RegistryObject<CustomShieldItem> CUSTOM_SHIELD_ITEM = ITEMS.register("custom_shield",
            () -> new CustomShieldItem((new Item.Properties()).durability(336).tab(CreativeModeTab.TAB_COMBAT)));

    public CustomShieldTest() {
        ITEMS.register(FMLJavaModLoadingContext.get().getModEventBus());
    }

    private static class CustomShieldItem extends Item {
        public CustomShieldItem(Properties properties) {
            super(properties);
        }

        @Override
        public UseAnim getUseAnimation(ItemStack stack) {
            return UseAnim.BLOCK;
        }

        @Override
        public int getUseDuration(ItemStack stack) {
            return 72000;
        }

        @Override
        public InteractionResultHolder<ItemStack> use(Level world, Player player, InteractionHand hand) {
            ItemStack itemstack = player.getItemInHand(hand);
            player.startUsingItem(hand);
            return InteractionResultHolder.consume(itemstack);
        }

        @Override
        public boolean isShield(ItemStack stack, @Nullable LivingEntity entity) {
            return true;
        }
    }
}

```
</details>

<details>
<summary>For 1.16.5</summary>

```java
package net.minecraftforge.debug.item;

import net.minecraft.entity.LivingEntity;
import net.minecraft.entity.player.PlayerEntity;
import net.minecraft.item.*;
import net.minecraft.util.ActionResult;
import net.minecraft.util.Hand;
import net.minecraftforge.fml.RegistryObject;
import net.minecraftforge.fml.common.Mod;
import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
import net.minecraftforge.registries.DeferredRegister;

@Mod("custom_shield_test")
public class CustomShieldTest {
    private static final DeferredRegister<Item> ITEMS = DeferredRegister.create(Item.class, "custom_shield_test");

    private static final RegistryObject<CustomShieldItem> CUSTOM_SHIELD_ITEM = ITEMS.register("custom_shield",
            () -> new CustomShieldItem((new Item.Properties()).durability(336).tab(ItemGroup.TAB_COMBAT)));

    public CustomShieldTest() {
        ITEMS.register(FMLJavaModLoadingContext.get().getModEventBus());
    }

    private static class CustomShieldItem extends Item {
        public CustomShieldItem(Properties properties) {
            super(properties);
        }

        @Override
        public UseAction getUseAnimation(ItemStack stack) {
            return UseAction.BLOCK;
        }

        @Override
        public int getUseDuration(ItemStack stack) {
            return 72000;
        }

        @Override
        public ActionResult<ItemStack> use(World world, PlayerEntity player, Hand hand) {
            ItemStack itemstack = player.getItemInHand(hand);
            player.startUsingItem(hand);
            return ActionResult.consume(itemstack);
        }

        @Override
        public boolean isShield(ItemStack stack, @Nullable LivingEntity entity) {
            return true;
        }
    }
}

```
</details>

sciwhiz12:
~~This PR fixes #8054 by re-adding the missing patch for `IForgeItem#isShield` back to `LivingEntiy#getEquipmentSlotForItem`, as noted in the linked issue. This also adds a test mod which adds a custom shield item, to allow testing this PR specifically and the `isShield` feature in general.~~

This PR fixes #8054 by adding a new tool action for shield blocking, `ToolActions#SHIELD_BLOCK`, and adding the necessary patches to make it work correctly. The `isShield` methods are kept for backwards compatibility, deprecated as to be removed in the next major version, and preserved in mods that already use this by redirecting calls of `canPerformAction` with the shield block action to this method.

**Screenshot of the implemented fix using the steps from the linked issue**:

### DIFF
--- a/patches/minecraft/net/minecraft/world/item/ShieldItem.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/ShieldItem.java.patch
@@ -1,0 +1,14 @@
+--- a/net/minecraft/world/item/ShieldItem.java
++++ b/net/minecraft/world/item/ShieldItem.java
+@@ -49,4 +_,11 @@
+    public static DyeColor m_43102_(ItemStack p_43103_) {
+       return DyeColor.m_41053_(p_43103_.m_41698_("BlockEntityTag").m_128451_("Base"));
+    }
++
++   /* ******************** FORGE START ******************** */
++
++   @Override
++   public boolean canPerformAction(ItemStack stack, net.minecraftforge.common.ToolAction toolAction) {
++      return net.minecraftforge.common.ToolActions.DEFAULT_SHIELD_ACTIONS.contains(toolAction);
++   }
+ }
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
@@ -695,7 +695,7 @@ public interface IForgeItem
      * @param stack  the item stack
      * @param entity the entity holding the item stack
      * @deprecated To be removed in 1.18. Override {@link #canPerformAction(ItemStack, ToolAction)} and return
-     * {@code true} if the passed in tool action is contained in {@link ToolActions#DEFAULT_SHOVEL_ACTIONS} or is
+     * {@code true} if the passed in tool action is contained in {@link ToolActions#DEFAULT_SHIELD_ACTIONS} or is
      * equals to {@link ToolActions#SHIELD_BLOCK}.
      */
     @Deprecated(since = "1.17.1", forRemoval = true)
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
@@ -526,7 +526,7 @@ public interface IForgeItem
     default boolean canPerformAction(ItemStack stack, ToolAction toolAction)
     {
         // Temporary patch to keep #isShield working until its removed; change to `return false` once removed
-        return ToolActions.DEFAULT_SHIELD_ACTIONS.contains(toolAction) && isShield(stack, null);
+        return isShield(stack, null) && ToolActions.DEFAULT_SHIELD_ACTIONS.contains(toolAction);
     }
 
     /**
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
@@ -51,6 +51,7 @@ import net.minecraft.util.Mth;
 import net.minecraft.network.chat.Component;
 import net.minecraft.world.level.Level;
 import net.minecraftforge.common.ToolAction;
+import net.minecraftforge.common.ToolActions;
 
 // TODO review most of the methods in this "patch"
 public interface IForgeItem
@@ -524,7 +525,8 @@ public interface IForgeItem
      */
     default boolean canPerformAction(ItemStack stack, ToolAction toolAction)
     {
-        return false;
+        // Temporary patch to keep #isShield working until its removed; change to `return false` once removed
+        return ToolActions.DEFAULT_SHIELD_ACTIONS.contains(toolAction) && isShield(stack, null);
     }
 
     /**
@@ -685,6 +687,21 @@ public interface IForgeItem
     default boolean canDisableShield(ItemStack stack, ItemStack shield, LivingEntity entity, LivingEntity attacker)
     {
         return this instanceof AxeItem;
+    }
+
+    /**
+     * {@return {@code true} if this item is considered a shield}
+     *
+     * @param stack  the item stack
+     * @param entity the entity holding the item stack
+     * @deprecated To be removed in 1.18. Override {@link #canPerformAction(ItemStack, ToolAction)} and return
+     * {@code true} if the passed in tool action is contained in {@link ToolActions#DEFAULT_SHOVEL_ACTIONS} or is
+     * equals to {@link ToolActions#SHIELD_BLOCK}.
+     */
+    @Deprecated(since = "1.17.1", forRemoval = true)
+    default boolean isShield(ItemStack stack, @Nullable LivingEntity entity)
+    {
+        return stack.getItem() == Items.SHIELD;
     }
 
     /**
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
@@ -688,18 +688,6 @@ public interface IForgeItem
     }
 
     /**
-     * Is this Item a shield
-     *
-     * @param stack  The ItemStack
-     * @param entity The Entity holding the ItemStack
-     * @return True if the ItemStack is considered a shield
-     */
-    default boolean isShield(ItemStack stack, @Nullable LivingEntity entity)
-    {
-        return stack.getItem() == Items.SHIELD;
-    }
-
-    /**
      * @return the fuel burn time for this itemStack in a furnace. Return 0 to make
      *         it not act as a fuel. Return -1 to let the default vanilla logic
      *         decide.
--- a/src/test/java/net/minecraftforge/debug/item/CustomShieldTest.java
+++ b/src/test/java/net/minecraftforge/debug/item/CustomShieldTest.java
@@ -1,0 +1,57 @@
+package net.minecraftforge.debug.item;
+
+import net.minecraft.world.InteractionHand;
+import net.minecraft.world.InteractionResultHolder;
+import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.CreativeModeTab;
+import net.minecraft.world.item.Item;
+import net.minecraft.world.item.ItemStack;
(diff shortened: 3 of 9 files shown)


src/test/java/net/minecraftforge/debug/item/CustomShieldTest.java
```diff
@@ -0,0 +1,57 @@
+package net.minecraftforge.debug.item;
```
diesieben07, line 20: Missing license header.

src/test/java/net/minecraftforge/debug/item/CustomShieldTest.java
```diff
@@ -0,0 +1,57 @@
...
+import net.minecraft.world.level.Level;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.fmllegacy.RegistryObject;
+import net.minecraftforge.registries.DeferredRegister;
+
+import javax.annotation.Nullable;
+
+@Mod("custom_shield_test")
+public class CustomShieldTest {
```
diesieben07: Code formatting is wrong for this file.

diesieben07 (changes requested):

Sirttas:
I wold suggest moving away from `isShield` and use a tool action.

sciwhiz12:
Changed the PR and patches to use a new `ToolAction`, `ToolActions.SHIELD_BLOCK`, instead of checking `isShield(LivingEntity)`. The `isShield` methods are deprecated, and `canPerformAction` is modified to return `true` if the action queried is `SHIELD_BLOCK` and the item is a shield through `isShield`, to preserve backwards compatibility. 

I do wonder whether `IForgeItemStack#isShield` should be redirected to `IForgeItem#canPerformAction` too, to preserve compatibility for items which override `canPerformAction` to act as a shield with mods that call `isShield`. Doing this would mean that the `LivingEntity` parameter of `isShield` when passed to the item will always be `null`, though.

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -524,7 +525,8 @@ default void setDamage(ItemStack stack, int damage)
      */
     default boolean canPerformAction(ItemStack stack, ToolAction toolAction)
     {
-        return false;
+        // Temporary patch to keep #isShield working until its removed; change to `return false` once removed
+        return ToolActions.DEFAULT_SHIELD_ACTIONS.contains(toolAction) && isShield(stack, null);
```
KnightMiner: Minor, but I would invert the order. An instance comparison/boolean is faster than a hash set lookup.

This fallback backwards compat does feel a bit messy still, but I suppose not much to do about that.

sciwhiz12: Done.

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -688,12 +690,15 @@ default boolean canDisableShield(ItemStack stack, ItemStack shield, LivingEntity
...
-     * Is this Item a shield
+     * {@return {@code true} if this item is considered a shield}
      *
-     * @param stack  The ItemStack
-     * @param entity The Entity holding the ItemStack
-     * @return True if the ItemStack is considered a shield
+     * @param stack  the item stack
+     * @param entity the entity holding the item stack
+     * @deprecated To be removed in 1.18. Override {@link #canPerformAction(ItemStack, ToolAction)} and return
+     * {@code true} if the passed in tool action is contained in {@link ToolActions#DEFAULT_SHOVEL_ACTIONS} or is
```
Sirttas: Should be default shield actions instead of default shovle actions.

sciwhiz12: Done. _(how did I miss this?!)_

patches/minecraft/net/minecraft/world/item/ShieldItem.java.patch
```diff
@@ -0,0 +1,14 @@
...
+    public static DyeColor m_43102_(ItemStack p_43103_) {
+       return DyeColor.m_41053_(p_43103_.m_41698_("BlockEntityTag").m_128451_("Base"));
+    }
++
++   /* ******************** FORGE START ******************** */
++
++   @Override
++   public boolean canPerformAction(ItemStack stack, net.minecraftforge.common.ToolAction toolAction) {
++      return net.minecraftforge.common.ToolActions.DEFAULT_SHIELD_ACTIONS.contains(toolAction);
++   }
```
diesieben07, line 13: Why add this patch here instead of doing what `isShield` used to do in `canPerformAction` (or rather, just call `isShield` which already checks this...)?

diesieben07 (changes requested):

patches/minecraft/net/minecraft/world/item/ShieldItem.java.patch
sciwhiz12: First, it is consistent with the pre-existing implementation of `ToolAction`. If I look into `SwordItem`, I see `canPerformAction` overriden. If I look into `PickaxeItem`, I see `canPerformAction` overriden. Same with `ShovelItem`, `AxeItem`, `HoeItem`, and `ShearsItem`. Leaving `ShieldItem` out of this consistency will make it the odd one out, and it will seem odd why shields are given that specific treatment.

Second, it will be easier for the next person who comes to remove `isShield` to do so. They only have to do two things now: remove `isShield` from both `IForgeItem` and `IForgeItemStack`, and change the implementation of `IForgeItem#canPerformAction` to `return false`. Doing the former will cause a compile error in the latter, and the comment in the method body will steer them to change it to `return false`. This also avoids changing the patch later.

Third, if we move the check which exists currently in `isShield` to `IForgeItem#canPerformAction`, then any mod currently made which uses `isShield` in their custom shield item will not work, as then `isShield` would never be called. As I've seen Lex state that Forge 1.17 is effectively API-stable, I'd like to avoid breaking `isShield`. Furthermore, `IForgeItem#canPerformAction` already calls `IForgeItem#isShield`.

diesieben07: Fair enough

diesieben07 (approved):

PR #8058, opened 2021-09-02
Add shebang to run.sh in server files

noeppi-noeppi:


TheCurle:
I knew we forgot something.

Thanks for the PR!

PR #8059, opened 2021-09-02
Json-based modifier system to manipulate vanilla StructureTemplatePools

rehwinkel:
For adding custom templates to existing Minecraft structures, such as Villages, Pillager Outposts or Bastions, one would have to add these into the List that is created (e.g. in `PlainVillagePools`). To make that possible this PR adds an event that is fired from within the constructor of the StructureTemplatePool to be created, allowing users to add their own templates to the structure. The event being fired carries the name of the pool (for selectively adding templates to specific pools) and underlying objects necessary to add a custom template. Because these pools are usually bootstrapped statically as soon as the registry is created, I had to remove the static call to delay bootstrapping to a point after forge init. This PR also features a test mod to use the new event.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
@@ -84,7 +84,7 @@
           return false;
        } else {
           MobEffectInstance mobeffectinstance = this.f_20945_.get(p_147208_.m_19544_());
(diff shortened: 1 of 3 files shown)


ChampionAsh5357:
If these are for what I think they are, then this is very undeveloped. These are defined in json, so I would think a better way to inject them would be through json similar to how global loot modifiers are done.

marchermans:
Yeah I looked into this, for the review, and I personally think this needs a redesign.

The events here are not in line with other existing systems to modify the JSON loaded from disk. Please create something similar to the GLM for this as well.

rehwinkel:
Just saw how IntelliJ obliterated the imports on ForgeHooks. Will fix ASAP.

rehwinkel:
Bump

PR #8060, opened 2021-09-03
Propagate all arguments passed to the server start scripts to the game

noeppi-noeppi:


server_files/run.bat
```diff
@@ -1,5 +1,5 @@
 REM Forge requires a configured set of both JVM and program arguments.
 REM Add custom JVM arguments to the user_jvm_args.txt
 REM Add custom program arguments {such as nogui} to this file at the end of the next line
```
ichttt: (Same for the comment in the run.sh): Does this hold true for with the new propagation logic as well? Also, it might be good to mention that you can call the script with params and it will pass them on.

noeppi-noeppi: I'll fix that

PR #8061, opened 2021-09-04
[1.17] Add an event when a shootable item tries to find a projectile item.

MelanX:
Look at #7715 for more information about it.
This is just the 1.17 version for it.


ChampionAsh5357:
So, if this aims to expand upon the behavior of getting stack information out of the held item, I think this would make more sense to add a method to `IForgeItem` which takes in that information and then can return the necessary result. An event here doesn't make as much sense to me.

sciwhiz12:
Closing as 1.17 is now an unsupported version, and there is an updated version of the feature intended by this PR at #8322.

PR #8062, opened 2021-09-04
Add system property for test command

Mysterious-Dev:
This PR add a system property for enable test command in-game.

This avoids to set IS_RUNNING_IN_IDE variable for dev and unset for build.

### DIFF
--- a/patches/minecraft/net/minecraft/commands/Commands.java.patch
+++ b/patches/minecraft/net/minecraft/commands/Commands.java.patch
@@ -1,5 +1,14 @@
 --- a/net/minecraft/commands/Commands.java
 +++ b/net/minecraft/commands/Commands.java
+@@ -169,7 +_,7 @@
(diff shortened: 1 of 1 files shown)


sciwhiz12:
See #8024 for a PR which integrates the gametest framework into Forge in a more mod-friendly way. If you do pursue this PR, I'd suggest following the approach of that PR which adds a hook to `ForgeHooks` that checks if either `IS_RUNNING_IN_IDE` is enabled or a specified system property is enabled (or possibly other logic, like checking if `FMLEnvironment.production` is `false`).

Mysterious-Dev:
Intresting. 8024 is good, I close this PR.

sciwhiz12:
In the meantime while that PR's not done, one can still manually register the `TestCommand` in their own mod in the `RegisterCommandsEvent` if they wish to try out the gametest framework as-is.

PR #8065, opened 2021-09-05
[1.17] Fix breaking block particles not respecting IModelData

percivalalb:
The sprite used for the block particles needs updating after the constructor otherwise it defaults to the texture sprite returned from passing `ModelData.INSTANCE` to get `IForgeBakedModel#getParticleIcon`. Calling `updateSprite` fetches the real IModelData instance for that blockstate/pos and sets the texture to the desired one

### DIFF
--- a/patches/minecraft/net/minecraft/client/particle/TerrainParticle.java.patch
+++ b/patches/minecraft/net/minecraft/client/particle/TerrainParticle.java.patch
@@ -9,7 +9,7 @@
        }
 +   }
 +
-+   private Particle updateSprite(BlockState state, BlockPos pos) { //FORGE: we cannot assume that the x y z of the particles match the block pos of the block.
++   public Particle updateSprite(BlockState state, BlockPos pos) { //FORGE: we cannot assume that the x y z of the particles match the block pos of the block.
 +      if (pos != null) // There are cases where we are not able to obtain the correct source pos, and need to fallback to the non-model data version
 +         this.m_108337_(Minecraft.m_91087_().m_91289_().m_110907_().getTexture(state, f_107208_, pos));
 +      return this;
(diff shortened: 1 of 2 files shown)


percivalalb:
My might change vanilla behaviour... maybe this was a better solution https://github.com/MinecraftForge/MinecraftForge/commit/b466dc2c997a2872f4efec52da906900e05428f0

patches/minecraft/net/minecraft/client/particle/TerrainParticle.java.patch
```diff
@@ -1,17 +1,29 @@
...
+    public TerrainParticle(ClientLevel p_108282_, double p_108283_, double p_108284_, double p_108285_, double p_108286_, double p_108287_, double p_108288_, BlockState p_108289_) {
+       this(p_108282_, p_108283_, p_108284_, p_108285_, p_108286_, p_108287_, p_108288_, p_108289_, new BlockPos(p_108283_, p_108284_, p_108285_));
+    }
+@@ -23,7 +_,8 @@
+    public TerrainParticle(ClientLevel p_172451_, double p_172452_, double p_172453_, double p_172454_, double p_172455_, double p_172456_, double p_172457_, BlockState p_172458_, BlockPos p_172459_) {
+       super(p_172451_, p_172452_, p_172453_, p_172454_, p_172455_, p_172456_, p_172457_);
+       this.f_108280_ = p_172459_;
+-      this.m_108337_(Minecraft.m_91087_().m_91289_().m_110907_().m_110882_(p_172458_));
++      if (p_172459_ == null) this.m_108337_(Minecraft.m_91087_().m_91289_().m_110907_().m_110882_(p_172458_));
++      else this.m_108337_(Minecraft.m_91087_().m_91289_().m_110907_().getTexture(p_172458_, p_172451_, p_172459_));
```
ichttt: We can't do this. As you can see above, the BlockPos is instantiated with the x y z pos if it is not passed in, which is not always the position of the block, but the position where the particles should render. Block pos may and will be != particle pos in some cases and ause wrong lookups, which we definetly don't want. The other commit you mentioned looks better, as it passes the correct BlockPos.

percivalalb: Yeah thanks, you're right, realised this after. I've switched back to my first solution which works and is correct the only issue being if it was ok for `updateSprite` to be public facing

PR #8068, opened 2021-09-06
Remove old Log4j thread cache invalidation

ichttt:
As minecraft now requires java 16, the code wasn't even running anymore due to the java version check.


LexManos:
Someone remind me in 1.18, to migrate the hooks classes to a .impl. package to remove them from our binary break rules.

PR #8070, opened 2021-09-07
Custom Item decorations using capabilities 

lobbienjonsji:
As suggested in #8048 this PR
adds a Capability that can be attached to Itemstacks to add CustomItemDecorators objects with render methods that get executed in the RenderGuiItemDecorations method within the item stack renderer class. This is preferable to event dispatching, as no event dispatching needs to happen for every single item that is being rendered for every single subscriber (see #7693  for more information). There are a few things that still might need changing:

- Notably, that currently the capability writes to the Itemstacks NBT directly, since the capability is supposed to work with all Items, which means, that the getShareTag method wouldn't work. The alternative would be to ask modders to send a synchronization package themselves whenever they change the ItemStacks decorations which could be unneccesarily inconvenient. 

- Another issue would be the case, that addDecorations is called server side with a key of a decorator, that doesn't exist. The current implementation will simply ignore faulty keys client side, which can result in Itemstacks having unneeded NBT. The alternative would be to throw an exception when the decorator does not exist, but that could potentially lead to render world saves unable to load without first disabling the mod that caused the issue. A safer alternative would be to register the keys both client side and server side and not write to NBT if the key does not exist in the first place. Ultimately these issues would only result from wrong mod code and likely be noticed before mod release.

- Another issue would be, that uninstalling certain mods or them changing their registry keys for the decorators would result in items having unneeded NBT data attached to them.

The test mod shows the capability working in theory, but the implementation still has many issues.


lobbienjonsji:
I'm sorry for this commit history, next time I'll squash my commits before I push

Issue #5837, opened 2019-06-19
[1.14.2] ISelectiveResourceReloadListener is still not server-safe


Shadows-of-Fire:
The majority of the classes relating to selective reloading have had their `@OnlyIn` markers removed, except ReloadRequirements, which still has it, and causes the entire class to be unusable on a server.

LexManos:
ISelecxtiveResourceReloadListener is deprecated until we fully grok the new threaded resource management system introduced in 1.14.

williewillus:
IMO the benefit is not that clear anymore to me. It was introduced to prevent laggy reloads but since the threaded system lets you do a bunch of stuff in parallel it's not clear how much of a performance win it is anymore.

At a high level, the threaded reload system is simple, the `ReloadListener` class has two methods, one for parallel work (reading files, expensive local computation, etc.), and one for serial work (uploading textures, inserting into global maps, etc.).

PR #8072, opened 2021-09-08
Remove the remnants of the selective reload listener

ichttt:
The selective reload logic has been dead since 1.14 (see https://github.com/MinecraftForge/MinecraftForge/issues/5837), and with each update, more and more patches for the logic have been dropped.

With 1.17.1, the ISelectiveResourceReloadListener got deleted as well, but the deprecation Marker still exists in the ResourceManagerReloadListener.

This PR removes all that is left of the original logic.
Nobody felt like reimplementing this since 1.14 introduced threaded resource loading, so I belive we should just kill it completly. It would probably need a rewrite anyways if somebody would try to reimplement the feature.


pupnewfster (commented):
This looks relatively good and it will definitely be nice to not deprecation warnings when implementing `ResourceManagerReloadListener`. One other patch that should also probably be removed as part of this that I ran into when doing a brief search is [this](https://github.com/MinecraftForge/MinecraftForge/blob/b7379953d50a04f4cc7f7c9912969959ed6268d2/patches/minecraft/net/minecraft/client/Minecraft.java.patch#L150).

LexManos:
Honestly, even tho 1.17 hasn't officially had a RB yet, this is just a lot of binary breaks for breaks sake.
Better to mark things as deprecated/remove the patches then to break everything as is.
Now with J16 we can have since/forRemoval in @Deprecated, probably worth adding a gradle build verifier to make sure we strip all those out in 1.18.

pupnewfster (commented):
I think it looks good now, though the deprecated annotations should probably have a `since = "1.18"` as well given I know lex would like to eventually have a check get ran that ensures things are removed in the next MC version port when they are marked as deprecated/for removal.

Edit: 1.18 not 1.17, as the idea is to have it be the version it will get removed instead of actually when it was deprecated.

patches/minecraft/net/minecraft/server/packs/resources/ResourceManagerReloadListener.java.patch
```diff
@@ -1,22 +1,12 @@
...
-    default CompletableFuture<Void> m_5540_(PreparableReloadListener.PreparationBarrier p_10752_, ResourceManager p_10753_, ProfilerFiller p_10754_, ProfilerFiller p_10755_, Executor p_10756_, Executor p_10757_) {
-       return p_10752_.m_6769_(Unit.INSTANCE).thenRunAsync(() -> {
-@@ -17,4 +_,9 @@
+@@ -17,4 +_,11 @@
     }
  
     void m_6213_(ResourceManager p_10758_);
 +
++   /** @deprecated Selective reloading is no longer supported due to vanilla now reloading resources in parallel, so this is no longer needed */
++   @Deprecated(forRemoval = true)
```
pupnewfster: This is missing a `since="1.17.1"`

ichttt: Shows how much out of the loop I am, I found this as well when checking what I needed to update but forgot to gen patches...

PR #8073, opened 2021-09-09
[1.17.x] make RecipeManager#byType public

Minecraftschurli:
make RecipeManager#byType public so not every mod that needs access to the recipe by type map needs to use reflection/mixins/coremods/at


ChampionAsh5357:
This can be done by using `RecipeManager#getAllRecipesFor` where the id is stored on the recipes themselves iirc.

PR #8075, opened 2021-09-11
Add Mixin 0.8.4 to Forge 1.17

SizableShrimp:
This PR adds back Mixin to Forge for 1.17, based on the amazing work from Mumfrey. Some basic testing shows all userdev, forgedev, and installed version working correctly with Mixin, however more testing would be appreciated.

Support has now been added for Forge to use its own mixins, although this PR by itself does not move any forge patches/code to mixins. Userdev mods can now use mixin without having to depend on external projects or shade Mixin.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,10 +2,12 @@ buildscript {
     repositories {
         mavenLocal()
         maven { url = 'https://maven.minecraftforge.net/' }
+        maven { url = 'https://repo.spongepowered.org/repository/maven-public/' }
         mavenCentral()
     }
     dependencies {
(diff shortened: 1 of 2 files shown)


LexManos:
Just posting some of the things we've been discussing interanlly.
Now that Mumfrey has released a stable version of Mixin we can ship it.
We've been wanting to look into using it on our end for a while, to try and cleanup some of our code.
However, there is some work that needs to be done on our maven {namely I want to mirror sponge's so we dont get broken builds IF theirs goes down/they remove things} 
If you want to split this off into just shipping mixin we can pull that real quick  {its literally just the installer dep line}
The rest of the stuff needs more work.

PR #8078, opened 2021-09-14
Added EntityEvent.EnteringSection

diesieben07:
Intended as a successor to `EntityEvent.EnteringChunk`


marchermans (approved):
Perfect this was indeed missing.

LexManos:
It'd be worth caching if we should fire the event based on the class passed into the constructor.
So that we don't have to do a instanceof that will always be true.

diesieben07:
Added caching for the actual `Entity` instance to avoid instanceof / cast.

diesieben07:
Updated to fire the event after the entity has moved to the new section, not before.

PR #8079, opened 2021-09-15
Add a capability for custom Item decorations

lobbienjonsji:
### Item Decorator Handler Capability:
A capability that can be attached to any item stack. It allows adding item decorators to and removing them from any itemstack.
Decorators need to be added server side, since they should persist between world saves.
Currently the capability data gets saved on the item stacks NBT directly, since this is supposed to work with any item, making overwriting getShareTag not a possible solution. The alternative would be, to ask modders to sync the capability manually or to modify the capability on both sides. I am not sure which would be the preferrable option.

### Item Decorator:
Simple registry object with a render method that gets called during renderGuiItemDecorations and passed the same parameters when added to an item with the capability.

### New Registry:
The item decorators have a common registry, since registering them client side only resulted in the following issue:
Previously, when decorators were only registered client side, the addDecorator method had to be called server side
with the key of the respective decorator. Since there was no way of  checking whether a decorator with that key actually exists, this could result in useless NBT data getting added to the item, that referenced a decorator that does not exist.

### Test Mod:
Adds the capability and a decorator that renders a smiley face to any DiggerItem.
When the Item is dropped, the decorator is removed.

### Issue:
Since capability data gets saved on the items NBT data directly,
unloading mods that add decorators results in stacks having leftover NBT with no use.
This would also be the case if a mod changed the name of its decorators.
Technically a way could be implemented for the capability to check the items decorator NBT for
no longer existing decorator keys.
I am not sure how to implement it in a way, that's both consistent and not checking for it every tick though.
If instead the Capability does not write to the items NBT and gets manually synched by the modders using it, this would not be an issue, but it would make the Capability significantly less convenient to use.


ChampionAsh5357:
What's the point of a capability if you need to register the capability in a different fashion? You can have client only capabilities. And if you need to send something to the server, you can still do so without any issues.

lobbienjonsji:
The capability does not need to be registered in a different fashion, only the renderers/decorators that get added to the capability.
The capability mainly serves the function of easily attaching/removing them from ItemStacks. 
The idea of the current implementation is, to add renderers to stacks on a per-stack basis. 
This is supposed to be more efficient than #8048, where an event would get fired whenever an Items decorations a are rendered.
Now, instead of firing events whenver the method is called, the ItemStack gets checked whether it has the Item Decorator Handler Capability, and if it does, the render method of all decorators within the capability gets called.
Regarding client only capabilities, can they be made persistent?
The decorators being saved for the stack allows them to be treated like a property that can be applied to an item via things like crafting. E.g. an Item gets enchanted with a specific enchantment, the enchantment is supposed to add a secondary bar similar to a durability bar. Now a decorator, that renders the required bar, can simply be added to the items Item Decorator Handler Capability.

src/main/java/net/minecraftforge/client/MinecraftForgeClient.java
```diff
@@ -123,6 +129,20 @@ public static void onRebuildChunk(Level world, BlockPos position, RenderChunkReg
...
+        LazyOptional<IItemDecoratorHandler> targetCapability = customItemDecorationCache.get(stack);
+        if (targetCapability == null)
+        {
+            ICapabilityProvider provider = stack;
+            targetCapability = provider.getCapability(CapabilityItemDecoratorHandler.ITEM_DECORATOR_HANDLER_CAPABILITY, null);
+            customItemDecorationCache.put(stack, targetCapability);
+            targetCapability.addListener(self -> customItemDecorationCache.put(stack, null));
+        }
+        targetCapability.ifPresent(decoration -> decoration.render(font, stack, xOffset, yOffset, stackSizeLabel));
+    }
```
diesieben07, line 144: This cache is a huge memory leak.

src/main/java/net/minecraftforge/common/ItemDecorator.java
```diff
@@ -0,0 +1,29 @@
...
+import net.minecraftforge.registries.ForgeRegistryEntry;
+import net.minecraftforge.registries.IForgeRegistryEntry;
+
+import javax.annotation.Nullable;
+
+/**
+ * An Object that is used by the {@Link: net.minecraftforge.items.ItemDecoratorHandler} capability.
+ * Please try to add only one decorator per mod, to keep NBT size to a minimum.
+ */
+public class ItemDecorator extends ForgeRegistryEntry<ItemDecorator> {
```
diesieben07: Why is this a registry entry?

src/main/java/net/minecraftforge/common/ItemDecorator.java
```diff
@@ -0,0 +1,29 @@
...
+    public ItemDecorator()
+    {}
+
+    /**
+     * Is passed the parameters from {@link ItemRenderer#renderGuiItemDecorations(net.minecraft.client.gui.Font, net.minecraft.world.item.ItemStack, int, int, java.lang.String)}
+     * Is only called client side
+     */
+    @OnlyIn(Dist.CLIENT)
```
diesieben07: Don't use `@OnlyIn`. This system as a whole should be client-only anyways.

src/main/java/net/minecraftforge/items/IItemDecoratorHandler.java
```diff
@@ -0,0 +1,20 @@
...
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.api.distmarker.Dist;
+import net.minecraftforge.api.distmarker.OnlyIn;
+import net.minecraftforge.common.ItemDecorator;
+import net.minecraftforge.common.util.INBTSerializable;
+
+import javax.annotation.Nullable;
+
+public interface IItemDecoratorHandler extends INBTSerializable<CompoundTag>
```
diesieben07: Why does this implement `INBTSerializable`?

diesieben07 (changes requested):
I have several problems with this:

- Why does this save to NBT? KISS should be applied, this system should do _one_ thing, i.e. allow rendering custom item decorations. Leave data storage to existing systems, such as capabilities.
- Why is this not a purely client-side system?
- Why is there a registry being added?
- Who has the responsibility to add the decorator capability? How are mods supposed to add decorators to "foreign" items? If they add the capability using `AttachCapabilitiesEvent` this can cause issues if the capability is already added by the item itself or other mods.

src/main/java/net/minecraftforge/client/ItemDecorator.java
```diff
@@ -0,0 +1,34 @@
...
+
+import net.minecraft.client.gui.Font;
+import net.minecraft.client.renderer.entity.ItemRenderer;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.api.distmarker.Dist;
+import net.minecraftforge.api.distmarker.OnlyIn;
+import net.minecraftforge.registries.ForgeRegistryEntry;
+import net.minecraftforge.registries.IForgeRegistryEntry;
+import org.stringtemplate.v4.ST;
```
TheCurle: Why is this here? It's not used.

src/main/java/net/minecraftforge/items/IItemDecoratorHandler.java
```diff
@@ -0,0 +1,17 @@
...
+import net.minecraftforge.api.distmarker.OnlyIn;
+import net.minecraftforge.client.ItemDecorator;
+
+import javax.annotation.Nullable;
+
+public interface IItemDecoratorHandler
+{
+    void addDecorator(ItemDecorator decorator);
+    void removeDecorator(ItemDecorator decorator);
+    @OnlyIn(Dist.CLIENT)
```
TheCurle: Remove this OnlyIn.

src/main/java/net/minecraftforge/items/ItemDecoratorHandler.java
```diff
@@ -0,0 +1,34 @@
+package net.minecraftforge.items;
+
+import net.minecraft.client.gui.Font;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.client.ItemDecorator;
+
+import javax.annotation.Nullable;
+import java.util.*;
```
TheCurle: Try to avoid wildcard imports like this.

src/main/java/net/minecraftforge/items/ItemDecoratorHandler.java
```diff
@@ -0,0 +1,34 @@
+package net.minecraftforge.items;
```
TheCurle, line 20: This class needs a license header. Run the licenseFormat task.

src/main/java/net/minecraftforge/items/ItemDecoratorHandler.java
```diff
@@ -0,0 +1,34 @@
...
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.client.ItemDecorator;
+
+import javax.annotation.Nullable;
+import java.util.*;
+
+public class ItemDecoratorHandler implements IItemDecoratorHandler
+{
+    private Map<String, ItemDecorator> ItemDecoratorCache = new HashMap();
+    private static final String DECORATORS = "CapabilityItemDecoratorHandler";
```
TheCurle: What's this string for?

TheCurle (changes requested):
Seems like it fulfils its promise overall, but it needs some adjustments.

lobbienjonsji:
fixed

src/main/java/net/minecraftforge/client/MinecraftForgeClient.java
desht: Agreed.  It should be some form of LRU cache, with a fixed upper bound on the size and auto-ejection of old entries.  Guava `LoadingCache`/`CacheBuilder` could work here.

Another problem is that ItemStack does not override `equals()` or `hashCode()`, so every separate instance of an ItemStack is a different map key, even if the two stacks are otherwise identical.  In other words:
```java
ItemStack stack = new ItemStack(Items.STONE);
cache.put(stack, handler);
LazyOptional<IItemDecoratorHandler> handler2 = cache.get(stack.copy());
```
doesn't work.  Unless that's your intention?

src/main/java/net/minecraftforge/client/ItemDecorator.java
```diff
@@ -0,0 +1,48 @@
...
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.item.ItemStack;
+
+import javax.annotation.Nullable;
+
+/**
+ * An Object that is used by the {@Link: net.minecraftforge.items.ItemDecoratorHandler} capability.
+ */
+public abstract class ItemDecorator{
+    private final ResourceLocation key;
```
ChampionAsh5357, line 33: If they key is only used as an identifier in the map, why does it need to be attached to the decorator itself? Couldn't I just pass in the resource key for the map without having to deal with knowing what the decorator is? Otherwise I would always need to keep an instance of the decorator if I ever wanted to remove it based on the current implementation.

src/main/java/net/minecraftforge/items/IItemDecoratorHandler.java
```diff
@@ -0,0 +1,35 @@
...
+import net.minecraftforge.api.distmarker.Dist;
+import net.minecraftforge.api.distmarker.OnlyIn;
+import net.minecraftforge.client.ItemDecorator;
+
+import javax.annotation.Nullable;
+
+public interface IItemDecoratorHandler
+{
+    void addDecorator(ItemDecorator decorator);
+    void removeDecorator(ItemDecorator decorator);
```
ChampionAsh5357, line 33: This is limiting in design. I would suggest implementing a positioning system similar to #7770.

src/test/java/net/minecraftforge/debug/client/rendering/CustomItemDecorationsTest.java
```diff
@@ -0,0 +1,92 @@
...
+            bufferbuilder.vertex(x + 0, y + 0, 0.0D).color(1.0f, 1.0f, 1.0f, 1.0f).uv(0, 0).endVertex();
+            bufferbuilder.vertex(x + 0, y + 5, 0.0D).color(1.0f, 1.0f, 1.0f, 1.0f).uv(0, 1).endVertex();
+            bufferbuilder.vertex(x + 5, y + 5, 0.0D).color(1.0f, 1.0f, 1.0f, 1.0f).uv(1, 1).endVertex();
+            bufferbuilder.vertex(x + 5, y + 0, 0.0D).color(1.0f, 1.0f, 1.0f, 1.0f).uv(1, 0).endVertex();
+            bufferbuilder.end();
+            BufferUploader.end(bufferbuilder);
+        }
+    };
+
+    @Mod.EventBusSubscriber(modid = MODID)
```
ChampionAsh5357, line 67: Since this only exists on the client, it should probably be sided off here as well.

ChampionAsh5357 (commented):
Mainly a few design comments. I agree with dies so I won't repeat his points here.

lobbienjonsji:
Regarding the topic of who has the responsibility of attaching the capability: 
Which would be preferrable: exposing the capability in the ItemStack class or attaching the capability within forge itself?
Or is there maybe a different solution?
Exposing the Capability would also make the cache in MinecraftForgeClient redundant.

lobbienjonsji:
Both solutions would come with the downside of giving the capability to all ItemStacks, even those that don't need it, making the capability kind of pointless, because at that point it might as well just be a hard implementation in the forge itemstack interface.

sciwhiz12:
Hello! @lobbienjonsji, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable. If we do not hear back from you after some time, this PR will be closed for inactivity.

PR #8081, opened 2021-09-16
Add a new CapabilityReference class

sciwhiz12:
This PR adds the new `CapabilityReference` class, which holds a possibly-empty reference to the `Capability` object for a capability. This is intended to be an alternative to the use of `@CapabilityInject`, akin to the use of `RegistryObject` as an alternative to `@ObjectHolder`.

There are multiple reasons why such an object would be useful to mod developers:
 - **It does not carry the same side-effect as `@CapabilityInject` with regards to `final` fields.**

   `@CapabilityInject` has the semi-documented side-effect of removing the `final` modifier from fields where it is present. This finality removal is done by the [`CapabilityInjectDefinalize`][cap-definalize] transformer, whose javadocs notes that the removal is done in order to prevent the JIT (just-in-time) compiler (from the HotSpot VM, the most popular JVM implementation) from inlining references to the field. 

   Additionally, the `final` modifier removal also allows reflective operations to set the value of a `static final` field to succeed, something which is otherwise not possible even with the use of [`AccessibleObject#setAccessible(boolean)`][setAccessible].<sup><a id="ret-fn-1" href="#footnote-1">1</a></sup>

   Additionally, the definalization also causes hotswapping of classes which contain a `@CapabilityInject`-annotated field to fail due to the changed class structure (specifically, the modifiers of the annoatated field), as at runtime the field is non-`final` but the hotswapping agent receives the to-be-hotswapped class bytes with the field as `final`.

 - **It allows storing a capability reference in an interface field.**
   
   Due to the above side-effect of de-finalizing the field, the `@CapabilityInject` must not be applied to interface fields. Doing so will lead to an error, as the Java Virtual Machine Specification mandates that the fields of interfaces must be `public static final`.<sup><a id="ret-fn-2" href="#footnote-2">2</a></sup>

 - **It is an alternative to the use of `@CapabilityInject`, akin to `RegistryObject`.**

   A developer may wish to not use in their `@CapabilityInject` code, whether due to the above noted side-effects or due to their own coding preferences as to the use of annotations. This class provides an API-provided way to retrieve `Capability` objects for their capabilities without the need of the annotation.

   Furthermore, various libraries may be interested in an API to programmatically obtain references to `Capability` objects of any arbitrary capability class or interface, such as those which reduce developer boilerplate by providing convenience classes or helpers which implement various functionality, without requiring the manual intervention of the developer to pass in the `Capability` object themselves from a `@CapabilityInject`-annotated field or method.

### Remaining tasks before ready for review
 - [ ] Add a new test mod / modify existing test mod to use and test `CapabilityReference`
 - [ ] Test that injection works as expected
   - [ ] Test the same but with multiple mods creating references to the same capability (to ensure the use of `Set` doesn't cause any problems)
 - [ ] Test that creating an instance during static initialization does not cause any issues
 - [ ] Test the warning when an instance is created after capabilities are injected
 - [ ] Test that no regression occurs by using in a medium-large modpack
 - [ ] Review javadocs to add missing information and clarify as needed

---

<span id="footnote-1"></span><sup><a href="#ret-fn-1">^</a> 1: Java SE Documentation, Java SE 16 & JDK 16, [`AccessibleObject#setAccessible(boolean)`][setAccessible]: "This method cannot be used to enable _write_ access to a _non-modifiable_ final field. The following fields are non-modifiable: - static final fields declared in any class or interface ..."</sup>
<span id="footnote-2"></span><sup><a href="#ret-fn-2">^</a> 2: The Java® Virtual Machine Specification, _Java SE 16 Edition_, [4.5. Fields][jvms-4.5]: "Fields of interfaces must have their `ACC_PUBLIC`, `ACC_STATIC`, and `ACC_FINAL` flags set; ..."</sup>

[cap-definalize]: https://github.com/MinecraftForge/MinecraftForge/blob/ca2023fa80b943cf595de53434a3c67f974e15f6/fmlloader/src/main/java/net/minecraftforge/common/asm/CapabilityInjectDefinalize.java
[setAccessible]: https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/lang/reflect/AccessibleObject.html#setAccessible(boolean)
[jvms-4.5]: https://docs.oracle.com/javase/specs/jvms/se16/html/jvms-4.html#jvms-4.5

### DIFF
--- a/src/main/java/net/minecraftforge/common/capabilities/CapabilityManager.java
+++ b/src/main/java/net/minecraftforge/common/capabilities/CapabilityManager.java
@@ -35,7 +35,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.objectweb.asm.Type;
 
-import java.util.function.Function;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 import static net.minecraftforge.fml.Logging.CAPABILITIES;
@@ -78,7 +78,7 @@ public enum CapabilityManager
 
     // INTERNAL
     private final IdentityHashMap<String, Capability<?>> providers = new IdentityHashMap<>();
-    private volatile IdentityHashMap<String, List<Function<Capability<?>, Object>>> callbacks;
+    private volatile IdentityHashMap<String, List<Consumer<Capability<?>>>> callbacks;
     public void injectCapabilities(List<ModFileScanData> data)
     {
         final List<ModFileScanData.AnnotationData> elementsToInject = data.stream()
@@ -86,7 +86,7 @@ public enum CapabilityManager
             .flatMap(Collection::stream)
             .filter(a -> CAP_INJECT.equals(a.annotationType()))
             .collect(Collectors.toList());
-        final IdentityHashMap<String, List<Function<Capability<?>, Object>>> callbacks = new IdentityHashMap<>();
+        final IdentityHashMap<String, List<Consumer<Capability<?>>>> callbacks = new IdentityHashMap<>();
         elementsToInject.forEach(entry -> gatherCallbacks(callbacks, entry));
         CapabilityReference.addCallbacks(callbacks);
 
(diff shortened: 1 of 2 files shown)


diesieben07:
One thing that `@CapabilityInject` solves that this doesn't is optionality. `@CapabilityInject` won't cause any runtime errors if the capability class is not found due to a soft dependency. `CapabilityReference` would throw in that case, due to the hard reference to the `Class`.

marchermans:
I don't like the way the CapabilityReference is implement. This exposes internal systems to the modder, when using a different pattern like a factory, akin to `DeferredRegister`, the implementation of the way the reference is injected can be hidden from the user by simply splitting it into an interface and a package private implementation together with a factory.

Besides that, this CapabilityReference should just be a supplier like the RegistryObject is. And throw if any modder tries to access the cap before it is injected / initialized. Since this is an illegal state.

sciwhiz12:
Superseded by #8104 and #8116.

PR #8084, opened 2021-09-17
New raid related events

PellucidDice:
This pr adds a total of 4 events.

1. RaidEvent.GetOrCreate is triggered when a raid is requested in the given position, users are allowed to assign custom ones.

2. RaidEvent.Load is triggered when Raids trying to load, users are able to retrieve custom raids here

3. RaidEvent.SpawnGroup is triggered when a raid is trying to spawn a new wave. Users are allowed to change the spawn position and force spawn a bonus group. This event is Cancelable.

4. RaidEvent.Stop is triggered when the raid attempts to end. Users are allowed to prevent the raid for being cancelled. This event is Cancelable.

Also added [two getters](https://github.com/MinecraftForge/MinecraftForge/compare/1.17.x...PellucidDice:1.17.x#diff-2248f92eb2620ef4d5165f810f50671e16747c6a04aa95c4d2d5977adc9cfa61R98-R106) in Raid.java that allows users to access more about raids, for example players and raiders groups.

[Test Mod](https://github.com/MinecraftForge/MinecraftForge/compare/1.17.x...PellucidDice:1.17.x#diff-537d2852e3d01f22f103763bc5831df0f9dec7b665691ee6414ce618bd86ca46):
RaidEventTest ("raid_event_test")

Players receive slowness effect during raid:
image

When remaining raider count less than 2, they glows and are poisoned:
image
image

Player receives message when raid is ended:
image


patches/minecraft/net/minecraft/world/entity/raid/Raid.java.patch
```diff
@@ -1,5 +1,113 @@
...
++                     }
++                  }
++
++                  ++k;
++                  if (raider1 != null)
++                  {
++                     this.m_37713_(i, raider1, p_37756_, false);
++                     raider1.m_20035_(p_37756_, 0.0F, 0.0F);
++                     raider1.m_20329_(raider);
++                  }
```
dpeter99, line 90: This is a lot of changes. Either don't indent, or better try to do an early return.

patches/minecraft/net/minecraft/world/entity/raid/Raids.java.patch
```diff
@@ -0,0 +1,24 @@
...
++++ b/net/minecraft/world/entity/raid/Raids.java
+@@ -135,8 +_,9 @@
+    }
+ 
+    private Raid m_37960_(ServerLevel p_37961_, BlockPos p_37962_) {
+-      Raid raid = p_37961_.m_8832_(p_37962_);
+-      return raid != null ? raid : new Raid(this.m_37977_(), p_37961_, p_37962_);
++      net.minecraftforge.event.world.RaidEvent.GetOrCreate event = net.minecraftforge.event.ForgeEventFactory.getOrCreateRaid(p_37961_.m_8832_(p_37962_), p_37961_, p_37962_);
++      // Custom Raid -> Existing Raid -> New Vanilla Raid
++      return event.getCustomRaid() != null ? event.getCustomRaid().apply(m_37977_()) : event.getRaid() != null ? event.getRaid() : new Raid(m_37977_(), p_37961_, p_37962_);
```
dpeter99, line 11: This is a really complex few lines. It could be a single line between the two vanilla lines if you make the ``getOrCreateRaid`` return the current one if no new one was created.

patches/minecraft/net/minecraft/world/entity/raid/Raid.java.patch
```diff
@@ -1,5 +1,113 @@
...
+-                  raider1 = EntityType.f_20513_.m_20615_(this.f_37675_);
+-               } else if (i >= this.m_37724_(Difficulty.HARD)) {
+-                  if (k == 0) {
+-                     raider1 = EntityType.f_20568_.m_20615_(this.f_37675_);
+-                  } else {
+-                     raider1 = EntityType.f_20493_.m_20615_(this.f_37675_);
+-                  }
++      net.minecraftforge.event.world.RaidEvent.SpawnGroup event = net.minecraftforge.event.ForgeEventFactory.spawnGroupForRaid(this, p_37756_, flag1);
++      if (!event.isCanceled())
++      {
```
dpeter99, line 41: This shoud also be on the same line.

src/main/java/net/minecraftforge/event/world/RaidEvent.java
```diff
@@ -0,0 +1,219 @@
...
+     * attempts to end
+     *
+     * Users are allowed to prevent the raid
+     * for being cancelled
+     *
+     * This event is {@link Cancelable}
+     * If cancelled, the raid is not stopped
+     */
+    @Cancelable
+    public static class Stop extends RaidEvent
```
dpeter99, line 212: A better name for this would be ``Ending`` as it would be more obvious what is happening and what canceling it causes.

src/test/java/net/minecraftforge/debug/world/RaidEventTest.java
```diff
@@ -0,0 +1,96 @@
...
+        MinecraftForge.EVENT_BUS.addListener(this::loadCustomRaid);
+        MinecraftForge.EVENT_BUS.addListener(this::onRaidSpawnGroup);
+        MinecraftForge.EVENT_BUS.addListener(this::onRaidStop);
+    }
+
+    private static final Random RAND = new Random();
+    public void getCustomRaid(RaidEvent.GetOrCreate event)
+    {
+        if (RAND.nextBoolean())
+            event.setCustomRaid((id) -> new TestModRaid(id, event.getLevel(), event.getPos()));
```
dpeter99, line 35: For a test mod it is not a really good idea to make the effect change based. You coud be unlucky and not see anything for quite a few raids.

src/test/java/net/minecraftforge/debug/world/RaidEventTest.java
```diff
@@ -0,0 +1,96 @@
...
+import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.event.world.RaidEvent;
+import net.minecraftforge.fml.common.Mod;
+
+import java.util.ArrayList;
+import java.util.Random;
+import java.util.Set;
+
+@Mod("raid_event_test")
+public class RaidEventTest
```
dpeter99, line 21: If you look in other test mods there is ususally a ``private static final boolean ENABLE = false;`` this is so that they can be disabled more easily.

src/test/java/net/minecraftforge/debug/world/RaidEventTest.java
```diff
@@ -0,0 +1,96 @@
...
+        {
+            super.tick();
+            if (!isStopped())
+            {
+                if (getLevel().getGameTime() % 60 == 0)
+                {
+                    new ArrayList<>(getRaidEvent().getPlayers()).forEach((player) ->
+                    {
+                        player.addEffect(new MobEffectInstance(MobEffects.MOVEMENT_SLOWDOWN, 60));
+                    });
```
dpeter99, line 78: You could just use a normal for loop for this.

dpeter99 (changes requested):
This looks quite promissing. I added a few smaller notes on the code.
The main concern I have is the fact that this leaves the detection of the raid type to the modder. This could cause mods colliding, as different modders might decide to save the type differently. (One modder might call it ``type`` and other might call it ``raidType``) Also the current solution separates the loading and saving code. I would suggest introducing a registry for a class that handles saving and loading. And move the detection into forge.

Thelnfamous1:
Can you include an event to detect and even modify specific raiders that are to be spawned as a part of the raid?

Example use case: I want all pillagers that spawn with the raid at the final wave to spawn with multishot piercing crossbows

Another example use case: I want to all evokers in the raid to spawn riding a Ravager

sciwhiz12:
Hello! @PellucidDice, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable. If we do not hear back from you after some time, this PR will be closed for inactivity.

PellucidDice:
Hi, sorry I've been a bit bust in the past month.
I'll redo all the stuff on 1.18.

PellucidDice:
Will make another PR for modern versions of changes, thanks for the feedback and comments.

PR #8085, opened 2021-09-17
[HotFix][1.17.x] Enable the Lazy Capabilities technology on ItemStacks

marchermans:
This adds the missing patch to enable the lazy capabilities on itemstacks.

This was noticed by @pupnewfster when she compared #7945 to its 1.16.x counterpart.
The cause of the missing file is unknown, but I am not the greatest gitter, and I had problems updating the branch of the previous PR, which likely caused git to somehow miss the changes in the file.

Again sorry.


LexManos:
This is why we can't have nice things...

PR #8087, opened 2021-09-18
ATs for world generation extensibility, focusing on `NoiseBasedChunkGenerator`, `NoiseSampler`, `NoiseBasedAquifer`, and `Beardifier`

alcatrazEscapee:
### Motivation

Say you want to create a custom chunk generator, that is *similar* to vanilla's. So your starting point is "I want to re-implement the vanilla chunk generator, and then change various things. Now, the issue is, there are so many private, package private, or final classes in your way, you need to rip out huge sections of vanilla code, copy paste it, and generally write something that has no inter-operability with existing vanilla code. This is awful.

### Justifications

I am going to go over all the ATs I've added, and justify why they're included, why their access level is either `protected` or `public`, and why notable ones are omitted.

1. `public-f NoiseBasedChunkGenerator`
    - In order to do any similar chunk generator operations to vanilla, this is an ideal class to extend, as it has (in very nicely separated methods) all of the relevant implementation details of a "vanilla like" chunk generator.
2. `protected NoiseBasedChunkGenerator` fields and methods
    - If extending this chunk generator, almost all fields and methods are good candidates for overriding, or using in said overriden methods. As otherwise, it ends in a situation where to override method A, you need to call private method B, which needs method C, etc.
    - There are a couple field omissions, (such as one `public static final BlockState AIR`) which are ignored as they're either unnecessary or easy to re-implement in one or two lines of code, because they have no dependencies on any other implementation details of `NoiseBasedChunkGenerator`.
    - These methods and fields are all `protected`, as the intended purpose of the ATs is for extensibility. All the public APIs of `NoiseBasedChunkGenerator` are already public and inherited from `ChunkGenerator`.
3. `public NoodleCaveNoiseModifier` and `public OreVeinNoiseSource`
    - Both of these are non-static inner classes which implement `NoiseModifier` and `BaseStoneSource` respectively. They are good base implementations of both of these interfaces.
    - I have neglected to AT the fields of these classes, as the intended purpose here is not extensibility (as these are both short helper classes), but usage, by other parts of the chunk generator (for instance, in a custom `createBaseStoneSource` override.)
4. `protected Beardifier <init>`
    - This is to allow extensibility of custom `Beardifier`s. As the public constructor basically hardcodes an implementation that may not be desired, but the no-op constructor does not.
5. `public Beardifier beardifyOrBury`
    - This is the only method in `Beardifier` that is technically part of it's public API, and should be public as it's called from `NoiseBasedChunkGenerator`, which would now include any mod subclasses that want to use the beardifier.
6. `protected Beardifier` fields and methods
   - As with `NoiseBasedChunkGenerator`, the rest of the `Beardifier` field and method ATs are intended for extending or implementing a custom `Beardifier`. There is one notable omission, `computeBeardContribution`, because it is a static helper method used to initialize the static `BEARD_KERNEL` field, which has been made protected. So the method itself is not required.
7. `protected NoiseBasedAquifer` constructor, fields and methods
    - All useful fields, and all methods have been AT'd for extensibility here. The constructor has been made protected, as there's a factory method in `Aquifer` for creating default implementations, although extensions should still be considered. All public APIs of `NoiseBasedAquifer` are already `public` as they implement the `Aquifer` interface.
    - This additionally includes the now `public` class `AquiferStatus`, which was used extensively by `NoiseBasedAquifer`.
8. `protected NoiseSampler` fields and methods
    - Again, `NoiseSampler` is a class used throughout `NoiseBasedChunkGenerator`, but since 1.17, a fair amount of logic has migrated from the chunk generator to the sampler ([example in a 1.16 mod of logic that now requires an override in `NoiseSampler`](https://github.com/alcatrazEscapee/hexlands/blob/5f57e1a6f7929d9d87791381f391ad1ba88377ff/src/main/java/com/alcatrazescapee/hexlands/world/HexChunkGenerator.java#L488-L556)).
9. Selected `public NoiseSampler` fields.
    - These are a couple fields which have been chosen to be made `public final` in `NoiseSampler`. These are fields which are computed in the `NoiseBasedChunkGenerator` constructor, and then passed onto `NoiseSampler` (thus, totally inaccessible to mods).
    - Since `NoiseBasedChunkGenerator` instantiates a `NoiseSampler`, the provided path here is to allow mods to use a custom `NoiseSampler`, that just references the same fields as the existing `NoiseSampler`. ([A bit like this](https://github.com/alcatrazEscapee/hexlands/blob/0dac12f5306c59e78dac303fd9e6a2358f98d519/src/main/java/com/alcatrazescapee/hexlands/world/HexChunkGenerator.java#L75)).
    - These fields are still `final`, and don't have any notable accessible or mutable state, so exposing them as `public` is fine for this purpose.


src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -392,9 +392,67 @@ public net.minecraft.world.level.block.entity.HopperBlockEntity m_59395_(I)V # s
...
+protected net.minecraft.world.level.levelgen.NoiseBasedChunkGenerator f_158380_ # waterLevelNoise
+protected net.minecraft.world.level.levelgen.NoiseBasedChunkGenerator f_158381_ # lavaNoise
+protected net.minecraft.world.level.levelgen.NoiseBasedChunkGenerator f_158382_ # sampler
+protected net.minecraft.world.level.levelgen.NoiseBasedChunkGenerator f_158383_ # baseStoneSource
+protected net.minecraft.world.level.levelgen.NoiseBasedChunkGenerator f_158384_ # oreVeinifier
+protected net.minecraft.world.level.levelgen.NoiseBasedChunkGenerator f_158385_ # noodleCavifier
+protected net.minecraft.world.level.levelgen.NoiseBasedChunkGenerator f_64330_ # surfaceNoise
+protected net.minecraft.world.level.levelgen.NoiseBasedChunkGenerator f_64333_ # seed
+protected net.minecraft.world.level.levelgen.NoiseBasedChunkGenerator f_64334_ # height
+protected net.minecraft.world.level.levelgen.NoiseBasedChunkGenerator m_158391_(IIII)[D # makeAndFillNoiseColumn
```
ChampionAsh5357: This is just a passthrough method, so it doesn't need to be made protected.

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -392,9 +392,67 @@ public net.minecraft.world.level.block.entity.HopperBlockEntity m_59395_(I)V # s
...
+protected net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer m_158024_(II)D # similarity
+protected net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer m_158027_(III)I # getIndex
+protected net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer m_158031_(J)Lnet/minecraft/world/level/levelgen/Aquifer$NoiseBasedAquifer$AquiferStatus; # getAquiferStatus
+protected net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer m_158039_(I)I # gridX
+protected net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer m_158041_(III)Lnet/minecraft/world/level/levelgen/Aquifer$NoiseBasedAquifer$AquiferStatus; # computeAquifer
+protected net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer m_158045_(I)I # gridY
+protected net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer m_158047_(I)I # gridZ
+public net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer$AquiferStatus
+protected net.minecraft.world.level.levelgen.Beardifier <init>()V # constructor
+protected net.minecraft.world.level.levelgen.Beardifier f_158062_ # BEARD_KERNEL
```
ChampionAsh5357: This one seems strange as its only used in `#getBeardContribution` which is also not made protected like `#getBuryContribution`. Maybe a missed access transformer?

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -392,9 +392,67 @@ public net.minecraft.world.level.block.entity.HopperBlockEntity m_59395_(I)V # s
...
+protected net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer m_158019_(IDLnet/minecraft/world/level/levelgen/Aquifer$NoiseBasedAquifer$AquiferStatus;Lnet/minecraft/world/level/levelgen/Aquifer$NoiseBasedAquifer$AquiferStatus;)D # calculatePressure
+protected net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer m_158024_(II)D # similarity
+protected net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer m_158027_(III)I # getIndex
+protected net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer m_158031_(J)Lnet/minecraft/world/level/levelgen/Aquifer$NoiseBasedAquifer$AquiferStatus; # getAquiferStatus
+protected net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer m_158039_(I)I # gridX
+protected net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer m_158041_(III)Lnet/minecraft/world/level/levelgen/Aquifer$NoiseBasedAquifer$AquiferStatus; # computeAquifer
+protected net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer m_158045_(I)I # gridY
+protected net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer m_158047_(I)I # gridZ
+public net.minecraft.world.level.levelgen.Aquifer$NoiseBasedAquifer$AquiferStatus
```
ChampionAsh5357, line 418: This would not expose the `fluidLevel` and `fluidType` data held within making this patch unusable.

ChampionAsh5357 (changes requested):
Generally, I approve of the extensibility of the noise chunk generator since it would be much more useful. However, there are a few comments/concerns related to what is currently ATed.

src/main/resources/META-INF/accesstransformer.cfg
alcatrazEscapee: True. The justification was it was called from other methods which are now protected, so overrides would basically have to reimplement them as well, but it is a very simple method and isn't hard to do so.

src/main/resources/META-INF/accesstransformer.cfg
alcatrazEscapee: Yup, I think I had intended to AT both `getBeardContribution` and `getBuryContribution`, the only one I meant to leave out was `computeBeardContribution`.

src/main/resources/META-INF/accesstransformer.cfg
alcatrazEscapee: Oops, those two should've been made public (final).

ChampionAsh5357 (approved):

LexManos:
I don't see any bad side effects besides the one I mentioned, so if you can remove that one, we should be good.

src/main/resources/META-INF/accesstransformer.cfg
LexManos: This one I don't think is a good idea because it is static, and thus making it accessible would allow people to modify it's values and effect things that are not itself. It has a getter that you've made protected {getBeardContribution} so that should satisfy it.

alcatrazEscapee: Makes sense.

PR #8096, opened 2021-09-19
Move ItemStack#forgeInit call earlier to allow earlier capability queries

### LINKED ISSUE #8093: Querying capabilities from Item#getDamage causes NPE during loading
**Minecraft Version:** 1.17.1
**Forge Version:** 1.17.1-37.0.68
**Logs:** https://gist.github.com/sciwhiz12/02cb5a5b26f7dce76137a5a78c05d262
Note that the `latest.log` contains the exception and stacktrace, not the `debug.log`

**Steps to Reproduce:**
 1. Download the following archive containing a minimal reproducible example: [`examplemod.zip`](https://github.com/MinecraftForge/MinecraftForge/files/7190907/examplemod.zip)

    Source is available at [`examplemod-source.zip`](https://github.com/MinecraftForge/MinecraftForge/files/7190908/examplemod-source.zip). For quick overview, the source file and recipe (`data/examplemod/recipes/damaged_item.json`) contained with the mod is available at https://gist.github.com/sciwhiz12/40a3f98d43b2fd43b56bc46c0ac01f63.

 1. Install the mod into a dedicated server with the above given Forge version.
 1. Run the server (if error does not occur, accept the EULA and try again).
 1. __Observe as the server fails to load with a datapack loading error.__

## History

This issue [was first reported](first-report) to The Forge Project discord by user `The Magic Chad#9337`, having tried to run the [All the Mods 7 modpack][atm7] with Forge version 37.0.68 and failing because of an error during server loading.

After the report and confirming provisionally through code review of the mod involved ([ElementalCraft][elementalcraft] by @Sirttas) and Forge that the issue seems to be a Forge issue, I first confirmed that the issue exists through an installation of a Forge server with the above version and the aforementioned mod (along with its two dependencies).

Having confirmed the issue with the above setup, Iproceeded to create the minimal reproducible example mod attached above, and tested that it crashes when ran in the user development environment and in the Forge server installation. The log given above is from the Forge server installation with the MRE mod.

## Description of issue

The core issue lies with the following lines in the method:

https://github.com/MinecraftForge/MinecraftForge/blob/30a33f5c3eee95337179c1df276c372eea8cea8c/src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java#L89-L93

The problem is that when the call to `doGatherCapabilities` is reached, `lazyParentSupplier` is almost certain to be `null` (as the code which sets its value is in `doGatherCapabilities` itself). As `lazyParentSupplier` is `null`, this cause an NPE when `get` is called on the supplier.

This error situation can be reached if a mod's custom item overrides `getDamage` (or any method called from `ItemStack`'s constructor, chiefly prominent being `getDamange`) and from that method queries any capability, and an item stack is created with that item -- in this case, a recipe triggers the creation of the item stack with the custom item, which triggers the above situation.

[first-report]: https://discord.com/channels/313125603924639766/437001959950778368/888947502164889680
[atm7]: https://www.curseforge.com/minecraft/modpacks/all-the-mods-7
[elementalcraft]: https://www.curseforge.com/minecraft/mc-mods/elemental-craft

sciwhiz12:
This PR fixes #8093 by moving the initialization/gathering of `ItemStack` capabilities (and deserialization of their NBT data) to be after the item delegate (and other properties like the NBT tag and count of the stack) is initialized, and before any methods are called within the constructor (such as `ItemStack#getDamage(ItemStack)`). This ensures that the `lazyParentSupplier` field is initialized before it is needed for any capabilitiy queries (such as can be done from `getDamage` by mods' custom items), preventing the NPE from happening.

Originally, the fix as seen in the first comment was to guard against `lazyParentSupplier` being `null`, but upon comment by @OrionDevelopment (https://github.com/MinecraftForge/MinecraftForge/issues/8093#issuecomment-922518201) based on internal team discussions, the new fix is as suggested: moving `forgeInit` earlier so capabilities are ready for querying before `getDamage` is called.

### DIFF
--- a/patches/minecraft/net/minecraft/world/item/ItemStack.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/ItemStack.java.patch
@@ -19,7 +19,7 @@
     private static final Logger f_41585_ = LogManager.getLogger();
     public static final ItemStack f_41583_ = new ItemStack((Item)null);
     public static final DecimalFormat f_41584_ = Util.m_137469_(new DecimalFormat("#.##"), (p_41704_) -> {
-@@ -127,14 +_,19 @@
+@@ -127,10 +_,15 @@
        p_41606_.ifPresent(this::m_41751_);
     }
  
@@ -33,17 +33,13 @@
 +      this.capNBT = p_41606_;
 +      this.f_41589_ = p_41604_ == null ? null : p_41604_.m_5456_();
 +      this.delegate = p_41604_ == null ? null : p_41604_.m_5456_().delegate;
++      this.forgeInit();
 +      this.f_41587_ = p_41605_;
 +      if (this.f_41589_ != null && this.f_41589_.isDamageable(this)) {
           this.m_41721_(this.m_41773_());
        }
  
-       this.m_41617_();
-+      this.forgeInit();
-    }
- 
-    private void m_41617_() {
-@@ -143,18 +_,23 @@
+@@ -143,14 +_,19 @@
     }
  
     private ItemStack(CompoundTag p_41608_) {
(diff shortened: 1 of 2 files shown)


marchermans:
Leave the null check in.
Just in case.

sciwhiz12:
@OrionDevelopment: done, re-added the `null` check for `lazyParentSupplier`.

marchermans (approved):

PR #8097, opened 2021-09-19
Prevent call to Item#initializeClient when in datagen

sciwhiz12:
When running under data-generation, the `Minecraft` singleton is unavailable even though the operating env is under the `CLIENT` dist. Along with the fact that most uses of IItemRenderProperties (usually to provide a custom BEWLR) make use of the `Minecraft` singleton, this means that developers unaware of `Minecraft`'s unavailability will encounter a null pointer exception when they try to access its singleton.

This PR adds a check to prevent the `Item#initializeClient` method from being called during data-generation, to prevent the situation described before from causing an NPE. As `IItemRenderProperties` has nothing which might be used by datageneration, this is safe to do.

This PR is inspired by a [Discord conversation][discord] on The Forge Project discord server wherein @gigaherz expressed that this method shouldn't be called in datagen but this check was forgotten, and that it should be PR-ed.

[discord]: https://discord.com/channels/313125603924639766/867851603468615740/888932184122069036

### DIFF
--- a/patches/minecraft/net/minecraft/world/effect/MobEffect.java.patch
+++ b/patches/minecraft/net/minecraft/world/effect/MobEffect.java.patch
@@ -17,7 +17,7 @@
     }
  
     public void m_6742_(LivingEntity p_19467_, int p_19468_) {
-@@ -179,4 +_,28 @@
+@@ -179,4 +_,29 @@
     public boolean m_19486_() {
        return this.f_19447_ == MobEffectCategory.BENEFICIAL;
     }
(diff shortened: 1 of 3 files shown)


marchermans (approved):

gigaherz (changes requested):
There's 3 initializeClient methods in total (Item, Block, and MobEffect). Would be nice to get the same fix done to the other 2.

marchermans:
> initializeClient

I noticed them, but I am not sure they are related to issues in Datagen.
How are these other two involved?

gigaherz:
> How are these other two involved?

They are all equally involved. When datagen runs, forge runs its initialization logic, during which it registers all the things (items, blocks, and mobeffects included).
Since the initializeClient calls are done in the constructor of their respective object, it means all 3 will be called during datagen, and all 3 will run into the same issue if the logic in those classes references Minecraft.getInstance().

sciwhiz12:
Updated to also prevent `Block#initializeClient` and `Item#initializeClient` from being called during datagen.

gigaherz (approved):

marchermans (approved):

gigaherz:
Oops I forgot about this. Meant to merge sometime I wasn't busy.... and I guess I'm not busy now.

PR #8101, opened 2021-09-21
PlayerDamageItemEvent

Luis-St:
The `PlayerDamageItemEvent` allows the modder to run some code when an ItemStack is damaged via `ItemStack#hurt`, 
it's an easier way instead of using the code explained in [this](https://forums.minecraftforge.net/topic/104577-event-triggered-on-item-durability-loss/?tab=comments#comment-468296) forum thread. it also allows to prevent a `ItemStack` for being damaged, 
by canceling the event. for an example how to use the Event look at `PlayerDamageItemEventTest `


src/main/java/net/minecraftforge/event/entity/player/PlayerDamageItemEvent.java
```diff
@@ -0,0 +1,86 @@
...
+ * <br>
+ * This event does not have a result. {@link HasResult}<br>
+ * <br>
+ * This event is fired from {@link ForgeEventFactory#onPlayerDamageItem(ServerPlayer, ItemStack, int, int)}.<br>
+ * This event is fired on the {@link MinecraftForge#EVENT_BUS}.
+ **/
+@Cancelable
+public class PlayerDamageItemEvent extends Event
+{
+	@Nullable // copy vanilla logic
```
diesieben07: Wrong indentation.
What does this comment mean?

src/main/java/net/minecraftforge/event/entity/player/PlayerDamageItemEvent.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * PlayerDamageItemEvent is fired when a player damage an item.<br>
+ * This event is fired whenever a player damage an item in
+ * {@link ItemStack#hurt(int, java.util.Random, ServerPlayer)}.<br>
+ * <br>
+ * {@link #player} contains the player which is damage the item. <br>
+ * {@link #itemStack} contains the ItemStack which is damaged.<br>
+ * {@link #durability} contains the durability of the ItemStack before it will damaged. <br>
+ * {@link #newDurability} contains the durability of the ItemStack after it was damaged.<br>
```
diesieben07: These should be on the getters instead, not here.

src/main/java/net/minecraftforge/event/entity/player/PlayerDamageItemEvent.java
```diff
@@ -0,0 +1,86 @@
...
+ * <br>
+ * This event is {@link Cancelable}.<br>
+ * <br>
+ * This event does not have a result. {@link HasResult}<br>
+ * <br>
+ * This event is fired from {@link ForgeEventFactory#onPlayerDamageItem(ServerPlayer, ItemStack, int, int)}.<br>
```
diesieben07: This is entirely irrelevant to the event's function.

src/main/java/net/minecraftforge/event/entity/player/PlayerDamageItemEvent.java
```diff
@@ -0,0 +1,86 @@
...
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.event.ForgeEventFactory;
+import net.minecraftforge.eventbus.api.Cancelable;
+import net.minecraftforge.eventbus.api.Event;
+import net.minecraftforge.eventbus.api.Event.HasResult;
+
+/**
+ * PlayerDamageItemEvent is fired when a player damage an item.<br>
+ * This event is fired whenever a player damage an item in
```
diesieben07: Grammer ("damages", not "damage")

patches/minecraft/net/minecraft/world/item/ItemStack.java.patch
```diff
@@ -144,6 +144,10 @@
     }
  
     public boolean m_41629_(int p_41630_, Random p_41631_, @Nullable ServerPlayer p_41632_) {
++      if (net.minecraftforge.event.ForgeEventFactory.onPlayerDamageItem(p_41632_, this, this.m_41776_() - this.m_41773_(), this.m_41776_() - this.m_41773_() - p_41630_)) return false;
```
diesieben07: Why not pass in the damage value and the new damage value (which is what everything in Minecraft uses) instead of the inverted one ("durability")?

diesieben07 (changes requested):
Might also be nice to allow changing the amount of damage via the event.

Luis-St:
the event is now finished and should work
all errors/proposed changes noted by diesieben07 have been fixed

diesieben07:
> all errors/proposed changes noted by diesieben07 have been fixed

Incorrect. The wrong formatting as well as my feature request regarding changing the amount of damage have not been addressed.

Luis-St:
- the stupid comment has been removed
- the comments has been moved to the getter
- the javadoc has been updated (removed result comment and where it fired from)
- grammar has been fixed
- and the event use now the damage value

what's missing, btw look at both commits (add changes and add missing comment)

src/test/java/net/minecraftforge/debug/entity/player/PlayerDamageItemEventTest.java
```diff
@@ -0,0 +1,33 @@
+package net.minecraftforge.debug.entity.player;
```
sciwhiz12, line 20: Missing license header.

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/event/entity/player/PlayerDamageItemEvent.java
```diff
@@ -0,0 +1,89 @@
...
+ * <br>
+ * This event is {@link Cancelable}.<br>
+ * If the event is canceled, it will prevent the item from being damaged.<br>
+ * <br>
+ * This event is fired on the {@link MinecraftForge#EVENT_BUS}.
+ **/
+@Cancelable
+public class PlayerDamageItemEvent extends Event
+{
+	@Nullable
```
diesieben07: Indentation still wrong.

src/main/java/net/minecraftforge/event/entity/player/PlayerDamageItemEvent.java
```diff
@@ -0,0 +1,89 @@
...
+    public PlayerDamageItemEvent(@Nullable ServerPlayer player, ItemStack itemStack, int damageValue, int newDamageValue)
+    {
+        this.player = player;
+        this.itemStack = itemStack;
+        this.damageValue = damageValue;
+        this.newDamageValue = newDamageValue;
+    }
+    
+    /**
+     * @return the player which is damage the item
```
diesieben07:
```suggestion
     * @return the player damaging the item
```

Should also probably add Javadocs explaining again that this can be null (even though `@Nullable` is already present)

src/main/java/net/minecraftforge/event/entity/player/PlayerDamageItemEvent.java
```diff
@@ -0,0 +1,89 @@
...
+     **/
+    @Nullable
+    public ServerPlayer getPlayer() 
+    {
+        return this.player;
+    }
+
+    /**
+     * @return the ItemStack which is damaged
```
diesieben07:
```suggestion
     * @return the ItemStack being damaged
```

src/main/java/net/minecraftforge/event/entity/player/PlayerDamageItemEvent.java
```diff
@@ -0,0 +1,89 @@
...
+     **/
+    public ItemStack getStack() 
+    {
+        return this.itemStack;
+    }
+
+    /**
+     * @return the damage value of the ItemStack before it will damaged
```
diesieben07:
```suggestion
     * @return the damage value of the ItemStack before it will be damaged
```

diesieben07 (changes requested):
My previous review comment has also not been addressed:

> Might also be nice to allow changing the amount of damage via the event.

src/main/java/net/minecraftforge/event/entity/player/PlayerDamageItemEvent.java
Luis-St: why whats wrong, this is how the other events look too

sciwhiz12: The indentation on this specific line is a tab character, while the indentation of Forge classes is 4 spaces. The rest of the class seems to use 4 spaces fine, this line is the odd one out.

Luis-St:
so again, the event is now finished and should work
all errors/proposed changes have been fixed/were added

diesieben07 (approved):

LexManos:
Why? Also the event is in the wrong place, it should be inside the else, and potentially after the unbreaking check.
Because the durability you're passing to the event is not actually correct.
But ya, the main reason is why? Its not a big patch but everything comes with a cost and everything needs to be justified.

Luis-St:
> Also the event is in the wrong place, it should be inside the else, and potentially after the unbreaking check.

That makes sense, I changed that

> But ya, the main reason is why? Its not a big patch but everything comes with a cost and everything needs to be justified.

The `PlayerDamageItemEvent` allows the modder to run some code when an `ItemStack` is damaged via `ItemStack#hurt,`
it's an easier way instead of using the code explained in [this](https://forums.minecraftforge.net/topic/104577-event-triggered-on-item-durability-loss/?tab=comments#comment-468296) forum thread. It also allows to prevent a `ItemStack` for being damaged, by canceling the event or you set the damage of the `ItemStack` via `PlayerDamageItemEvent#setNewDamage.` For an example how to use the Event look at `PlayerDamageItemEventTest`.
Is that not enough?

LexManos:
> The `PlayerDamageItemEvent` allows the modder to run some code when an `ItemStack` is damaged via `ItemStack#hurt,` it's an easier way instead of using the code explained in [this](https://forums.minecraftforge.net/topic/104577-event-triggered-on-item-durability-loss/?tab=comments#comment-468296) forum thread. It also allows to prevent a `ItemStack` for being damaged, by canceling the event or you set the damage of the `ItemStack` via `PlayerDamageItemEvent#setNewDamage.` For an example how to use the Event look at `PlayerDamageItemEventTest`. Is that not enough?

`Putting a screen door in a submarine allows there to be a screen door in a submarine.`
Stating that adding something allows that thing to be added is not a justification for adding that thing.

Luis-St:
> `Putting a screen door in a submarine allows there to be a screen door in a submarine.` Stating that adding something allows that thing to be added is not a justification for adding that thing.

I mean the part about the forum thread:
> it's an easier way instead of using the code explained in this forum thread.

diesieben07:
The question is which use-case does this solve. Which mod can be written using this even that cannot be written without it?

Luis-St:
There are many use cases:
1. custom `Enchantment` or `Effect` (prevent an Item for being broken, ...)
2. editing the `Unbreaking Enchantment` behavior
3. modify the damage of vanilla Tools (`ShieldItem` will be damaged with 1 instead of 6, `CrossbowItem` will get more damage when shooting a rocket, random damage, ...)
4. custom `MobEffect` which work like the `Unbreaking Enchantment`

enough use cases?

Luis-St:
never mind case 2 is not possible, but it could be possible with a simple move of the patch line. 
I will change this tomorrow

Luis-St:
all cases stated by me, are now possible

Luis-St:
@diesieben07 whats happen now (sorry for ping)?
the Event is finished from my side

Luis-St:
since I am assuming that the event is not needed, I will close this pull request

PR #8103, opened 2021-09-21
Fix onChunkUnloaded not gettting called correctly

ichttt:
Fix onChunkUnloaded not gettting called correctly. Currently, the patch is still a wrong position, which is most likely caused by a patch-porting error from 1.16.x to 1.17.x as the vanilla code changed quite a bit.
This PR restores the call and removes old patches to the Level class that are no longer needed after the 1.17.x update


TheCurle:
Whoops. Think this one is my fault. Good catch on the fix.

ChampionAsh5357 (approved):
lgtm. Just need to keep track to make sure only `Level#unload(LevelChunk)` only uses that method.

PR #8104, opened 2021-09-21
Add CapabilityReference, alternative to @CapabilityInject

diesieben07:
This is a safer alternative to `@CapabilityInject`, working without definalizing or "magic field injection".
WIP and open for comments, requires Forge SPI update (and associated pull request) before merge.

See https://github.com/MinecraftForge/ForgeSPI/compare/master...diesieben07:feature/generic-field-data for the necessary changes in ForgeSPI.

### DIFF
--- a/fmlloader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModClassVisitor.java
+++ b/fmlloader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModClassVisitor.java
@@ -28,6 +28,8 @@ import org.objectweb.asm.Opcodes;
 import org.objectweb.asm.Type;
 
 import java.lang.annotation.ElementType;
+import java.util.ArrayList;
+import java.util.HashSet;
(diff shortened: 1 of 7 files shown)


LexManos:
https://github.com/MinecraftForge/MinecraftForge/compare/1.17.x...LexManos:1.17.x-CapRework
That was my thoughts.

diesieben07:
If I understand that correctly that queries the capability immediately in static init, which most likely wouldn't work.

Edit: Ah, no, it just always creates the `Capability` instead of using `null` which would be a breaking change.

LexManos:
Actually, it's not a breaking change, because it isn't calling the injects until the thing is registered.
So the only things that would get the instance before registration is new code that calls the new get method.
Which should know about isRegistered();

diesieben07:
Interesting. I think I might like your solution better then :D

diesieben07:
Closed in favor of #8116.

Issue #8086, opened 2021-09-18
CraftingHelper.getItemStack does not throw a JsonSyntaxException for unknown items


ochotonida:
**Minecraft Version:** 1.17.1

**Forge Version:** 37.0.67

**Logs:** -

**Steps to Reproduce:**
 1. Pass a JsonObject with an unknown item to `getItemStack` in `CraftingHelper`
 2. The method returns an empty stack instead of throwing an exception

**Description of issue:**
`getItemStack` throws an exception when the result of `ForgeRegistries.ITEMS.getValue` is null. Because `getValue` returns the default value (`minecraft:air`) for invalid keys, the exception is never thrown

ChampionAsh5357:
Closed by #8105. Air will be properly disallowed in vanilla recipes in 1.18 unless otherwise stated by Mojang's implementation with the changes made in the PR.

PR #8105, opened 2021-09-21
Fix Behavior With Unknown Items in Recipes

ChampionAsh5357:
As mentioned by #8086, `ForgeRegistry#getValue` will only return null if there is no default value. However, the `ITEM` registry will always default to `#AIR` if no item is available. With normal vanilla behavior (as defined by `ShapedRecipe#itemFromJson`, this returns a `JsonSyntaxException` mentioning that the air item is invalid.

This PR readds that check for `Items#AIR` and returns the same error message as vanilla would.

### DIFF
--- a/patches/minecraft/net/minecraft/world/item/crafting/ShapedRecipe.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/crafting/ShapedRecipe.java.patch
@@ -78,7 +78,7 @@
 -            return new ItemStack(item, i);
 -         }
 -      }
-+      return net.minecraftforge.common.crafting.CraftingHelper.getItemStack(p_151275_, true, true);
++      return net.minecraftforge.common.crafting.CraftingHelper.getItemStack(p_151275_, true, false); // TODO: Swap false to true in 1.18 to mimic vanilla behavior
     }
  
     public static Item m_151278_(JsonObject p_151279_) {
--- a/patches/minecraft/net/minecraft/world/item/crafting/ShapedRecipe.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/crafting/ShapedRecipe.java.patch
@@ -78,7 +78,7 @@
 -            return new ItemStack(item, i);
 -         }
 -      }
-+      return net.minecraftforge.common.crafting.CraftingHelper.getItemStack(p_151275_, true);
++      return net.minecraftforge.common.crafting.CraftingHelper.getItemStack(p_151275_, true, true);
     }
  
     public static Item m_151278_(JsonObject p_151279_) {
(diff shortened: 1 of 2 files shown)


Shadows-of-Fire:
I would ideally change this to only raise an exception if itemName is not `"minecraft:air"`.

KnightMiner:
The way I typically do the above is using `containsValue` instead of an air check. Lets us easily distinguish the result between defaulted and intentionally the default without hardcoding the registry name of the default.

ChampionAsh5357:
Yeah, my first thought is usually to go with vanilla logic. Probably the best way to do it is to check if it contains the key first and then check if the key is equal to the default key.

KnightMiner:
I think shadows was suggesting that the default item be allowed, so if I specifically request `minecraft:air` (empty item stack) that is still valid. While vanilla recipes may not want that behavior, its possible some mod is relying on the fact that you can make an empty item stack using air for their serializer

Shadows-of-Fire:
Yeah that's what I was suggesting, I have some recipes where air is a valid recipe component that has to be explicitly declared.   This should be able to return air when specifically asked for "minecraft:air" and error otherwise.

ChampionAsh5357:
For that reason, I feel as though it shouldn't be allowed as specifying an item component as air is the same as providing nothing. You could essentially specify the same logic by checking if a particular position in a recipe is empty.

KnightMiner:
In any case, disallowing `minecraft:air` is a breaking change, as you are requiring mods to change their code to fix this bug. If you think it shouldn't be allowed, I would just have it log a warning that "parsing an empty itemstack using `minecraft:air` is deprecated", and mark it to throw an exception for 1.18

pupnewfster:
This is a somewhat complicated case in regards to being a breaking change or not, as the vanilla code that this method is used in place of changed to disallow air in 1.17. Though I agree it would be a break to existing mods to change it... so at least for now the best solution may be to add another parameter for if air is allowed or not, and then an overload with the existing signature that passes that it allows air as to not break any mods that may be using that method, and change the places forge patches to call it that shouldn't allow air anymore to not allow it. This technically would still break any mod recipes where air is defined explicitly in the json... but given the fact that it is before an RB an edge case and is restoring vanilla functionality of not allowing it so otherwise would and should have caused those to error when mods port to 1.17 who specify air in their recipes it probably isn't worth worrying about compared to the fixing vanilla functionality while also not breaking any mods who call the method for their own use cases.

patches/minecraft/net/minecraft/world/item/crafting/ShapedRecipe.java.patch
```diff
@@ -78,7 +78,7 @@
 -            return new ItemStack(item, i);
 -         }
 -      }
-+      return net.minecraftforge.common.crafting.CraftingHelper.getItemStack(p_151275_, true);
++      return net.minecraftforge.common.crafting.CraftingHelper.getItemStack(p_151275_, true, true);
```
Shadows-of-Fire: I think this needs to default to false for now.  While true is in-line with vanilla behavior, it is still a breaking change to crash on invalid compared to current behavior.

ChampionAsh5357: Since it's been a month from the PRs start date and as we get closer to a potential RB, I agree that this should probably change to false for now and then left with a todo comment to change to true in 1.18.

PR #8106, opened 2021-09-21
[1.17] Pass IModelData through IForgeBakedModel#getModelData() before using it to get the particle texture

XFactHD:
This PR modifies the `IModelData`-aware particle texture getter in `BlockModelShaper` to pass the `IModelData` through `IForgeBakedModel#getModelData()` to allow custom baked models that either do not use the `ModelDataManager` from the `BlockEntity` or are used for a `Block` that doesn't even have a `BlockEntity` to alter the particle texture based on the `IModelData`.


ChampionAsh5357:
~~So whenever `ModelManager#getModelData` is called, it refers to `#refreshModelData` which only adds data to the cache if there is a `BlockEntity` at the specified position. Unless I am missing something, the premise of this PR would be incorrect.~~ I can't read patch/diffs, my bad.

ChampionAsh5357 (approved):

SkullGamer205:
В письме от воскресенье, 10 октября 2021 г. 03:43:54 +10 пользователь 
ChampionAsh5357 написал:
> So whenever `ModelManager#getModelData` is called, it refers to
> `#refreshModelData` which only adds data to the cache if there is a
> `BlockEntity` at the specified position. Unless I am missing something, the
> premise of this PR would be incorrect.

PR #8108, opened 2021-09-22
[1.17] Add accessor to an unmodifiable view of the fullPots map in FlowerPotBlock

XFactHD:
This PR adds a method to `FlowerPotBlock` that provides an unmodifiable view of the `fullPots` map added by Forge.

If a mod wants to add a plant pot that supports all plants that can be placed in the vanilla pot without requiring explicit registration of a plant-specific filled pot by the plant "owner", no matter how the mod does it (hope it is the last mod to get the block registry event and iterate the map there, check the map in `Block#use()` and store the inserted plant in a `BlockEntity` or some other way I can't think of) it has to access the aforementioned map.
While reflection works in both of those examples because the speed penalty doesn't matter, it is a really ugly way to access this map. Adding an accessor to an unmodifiable view of the map allows mods to access this map while making sure that they can't break anything.


patches/minecraft/net/minecraft/world/level/block/FlowerPotBlock.java.patch
```diff
@@ -82,5 +82,12 @@
 +       }
 +       fullPots.put(flower, fullPot);
 +   }
++
++   public Map<net.minecraft.resources.ResourceLocation, java.util.function.Supplier<? extends Block>> getFullPotsView() {
++      if (fullPots == null) {
```
Shadows-of-Fire: fullPots is never null, it is either a new HashMap or a Collections.emptyMap()
This condition should probably be `if(this.emptyPot != null)` which implies that `this` is a full pot.
There is another question to be asked if this should even raise an exception instead of just returning the empty map.

Shadows-of-Fire (changes requested):

patches/minecraft/net/minecraft/world/level/block/FlowerPotBlock.java.patch
XFactHD: Will fix the condition, no idea how I overlooked that.

I decided to throw an exception to match the behaviour of `addPlant()` but I can remove the exception if that is deemed to be more appropriate.

Shadows-of-Fire (approved):

XFactHD:
According to the feedback on Discord I have removed the exception from the getter and added a test mod that demonstrates my use case.

The current implementation of my use case in the target mod can be found [here](https://github.com/XFactHD/FramedBlocks/blob/1.17.x/src/main/java/xfacthd/framedblocks/common/block/FramedFlowerPotBlock.java#L41) and [here](https://github.com/XFactHD/FramedBlocks/blob/1.17.x/src/main/java/xfacthd/framedblocks/client/model/FramedFlowerPotModel.java#L152-L163).

rehwinkel:
The `FullPotsAccessorDemo` class is missing the license header! Currently, it's popping up in my uncommitted files because of that (after running updateLicenses).

PR #8109, opened 2021-09-22
[1.16] Add accessor to an unmodifiable view of the fullPots map in FlowerPotBlock

XFactHD:
This is the LTS backport of #8108.

This PR adds a method to `FlowerPotBlock` that provides an unmodifiable view of the `fullPots` map added by Forge.

If a mod wants to add a plant pot that supports all plants that can be placed in the vanilla pot without requiring explicit registration of a plant-specific filled pot by the plant "owner", no matter how the mod does it (hope it is the last mod to get the block registry event and iterate the map there, check the map in `Block#use()` and store the inserted plant in a `BlockEntity` or some other way I can't think of) it has to access the aforementioned map.
While reflection works in both of those examples because the speed penalty doesn't matter, it is a really ugly way to access this map. Adding an accessor to an unmodifiable view of the map allows mods to access this map while making sure that they can't break anything.

### DIFF
--- a/patches/minecraft/net/minecraft/block/FlowerPotBlock.java.patch
+++ b/patches/minecraft/net/minecraft/block/FlowerPotBlock.java.patch
@@ -55,7 +55,7 @@
           }
  
           return ActionResultType.func_233537_a_(p_225533_2_.field_72995_K);
-@@ -77,10 +_,34 @@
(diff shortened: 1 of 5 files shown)


ChampionAsh5357:
Just a reminder that since #8108 is now merged that the changes should be backported here.

PR #8110, opened 2021-09-22
Allows pet death message to be cancelled along with death

### LINKED ISSUE #7274: [1.15.2]Pet death message still sent in chat if LivingDeathEvent cancelled

OneLemonyBoi:
Allows pet death message to be cancelled along with death and closes #7274.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/TamableAnimal.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/TamableAnimal.java.patch
@@ -1,10 +1,8 @@
 --- a/net/minecraft/world/entity/TamableAnimal.java
 +++ b/net/minecraft/world/entity/TamableAnimal.java
-@@ -196,12 +_,13 @@
-       return super.m_7307_(p_21833_);
+@@ -197,11 +_,11 @@
     }
  
-+   @Override
     public void m_6667_(DamageSource p_21809_) {
 -      if (!this.f_19853_.f_46443_ && this.f_19853_.m_46469_().m_46207_(GameRules.f_46142_) && this.m_142480_() instanceof ServerPlayer) {
 +      super.m_6667_(p_21809_);
--- a/patches/minecraft/net/minecraft/world/entity/TamableAnimal.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/TamableAnimal.java.patch
@@ -1,16 +1,17 @@
 --- a/net/minecraft/world/entity/TamableAnimal.java
 +++ b/net/minecraft/world/entity/TamableAnimal.java
-@@ -197,11 +_,11 @@
+@@ -197,11 +_,14 @@
     }
  
     public void m_6667_(DamageSource p_21809_) {
--      if (!this.f_19853_.f_46443_ && this.f_19853_.m_46469_().m_46207_(GameRules.f_46142_) && this.m_142480_() instanceof ServerPlayer) {
++      // FORGE: Super moved to top so that death message would be cancelled properly
 +      super.m_6667_(p_21809_);
 +
-+      if (!this.f_19853_.f_46443_ && this.f_19853_.m_46469_().m_46207_(GameRules.f_46142_) && this.m_142480_() instanceof ServerPlayer && this.f_20890_) {
++      if (this.f_20890_)
+       if (!this.f_19853_.f_46443_ && this.f_19853_.m_46469_().m_46207_(GameRules.f_46142_) && this.m_142480_() instanceof ServerPlayer) {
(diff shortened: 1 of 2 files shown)


patches/minecraft/net/minecraft/world/entity/TamableAnimal.java.patch
```diff
@@ -0,0 +1,16 @@
...
+    }
+ 
+-   public void m_6667_(DamageSource p_21809_) {
++   public void sendDeathMessage() {
+       if (!this.f_19853_.f_46443_ && this.f_19853_.m_46469_().m_46207_(GameRules.f_46142_) && this.m_142480_() instanceof ServerPlayer) {
+          this.m_142480_().m_6352_(this.m_21231_().m_19293_(), Util.f_137441_);
+       }
+-
+-      super.m_6667_(p_21809_);
+    }
```
sciwhiz12: I'm a bit reluctant on the removal and replacement an override of a method here with a custom method that's called from the superclass. Is there no better option here? I can only think of possibly making an overload for `m_6667_`/`die` which accepts a boolean parameter to skip the `LivingDeathEvent` hook, then calling the hook here ourselves?

OneLemonyBoi: could we flip the order for the super and the message send code, and check if the entity is dead?

sciwhiz12: That sounds like a better option, please do so.

patches/minecraft/net/minecraft/world/entity/TamableAnimal.java.patch
```diff
@@ -0,0 +1,17 @@
+--- a/net/minecraft/world/entity/TamableAnimal.java
++++ b/net/minecraft/world/entity/TamableAnimal.java
+@@ -197,11 +_,11 @@
+    }
+ 
+    public void m_6667_(DamageSource p_21809_) {
+-      if (!this.f_19853_.f_46443_ && this.f_19853_.m_46469_().m_46207_(GameRules.f_46142_) && this.m_142480_() instanceof ServerPlayer) {
++      super.m_6667_(p_21809_);
++
++      if (!this.f_19853_.f_46443_ && this.f_19853_.m_46469_().m_46207_(GameRules.f_46142_) && this.m_142480_() instanceof ServerPlayer && this.f_20890_) {
```
sciwhiz12: Instead of adding the `this.f_20890_` condition to end of the `if` (which causes the line to be removed and then readded with the modifications), keep the `if` unchanged and add a new `if` (without braces) before that one which checks the same field:
```patch
+      if (this.f_20890_)
       if (!this.f_19853_.f_46443_ && this.f_19853_.m_46469_().m_46207_(GameRules.f_46142_) && this.m_142480_() instanceof ServerPlayer) {
```

patches/minecraft/net/minecraft/world/entity/TamableAnimal.java.patch
```diff
@@ -0,0 +1,17 @@
+--- a/net/minecraft/world/entity/TamableAnimal.java
++++ b/net/minecraft/world/entity/TamableAnimal.java
+@@ -197,11 +_,11 @@
+    }
+ 
+    public void m_6667_(DamageSource p_21809_) {
+-      if (!this.f_19853_.f_46443_ && this.f_19853_.m_46469_().m_46207_(GameRules.f_46142_) && this.m_142480_() instanceof ServerPlayer) {
++      super.m_6667_(p_21809_);
```
sciwhiz12, line 8: Add a comment near this moved super call to explain that it was moved to allow the pet death message to not be sent out if the death event for the entity was cancelled.

patches/minecraft/net/minecraft/world/entity/TamableAnimal.java.patch
```diff
@@ -0,0 +1,17 @@
...
++
++      if (!this.f_19853_.f_46443_ && this.f_19853_.m_46469_().m_46207_(GameRules.f_46142_) && this.m_142480_() instanceof ServerPlayer && this.f_20890_) {
+          this.m_142480_().m_6352_(this.m_21231_().m_19293_(), Util.f_137441_);
+       }
+-
```
sciwhiz12: Just for compactness' sake of the patch, readd this empty line after the `if`.

sciwhiz12 (changes requested):

OneLemonyBoi:
^ Misleading Commit Name

sciwhiz12 (approved):

PR #8111, opened 2021-09-22
Prevents Tile Entity data leak by copying CompoundNBT before writing

### LINKED ISSUE #7760: Custom tile entity data sided leak on integrated servers

OneLemonyBoi:
Prevents Tile Entity data leak by copying CompoundNBT before writing and closes #7760.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/Entity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/Entity.java.patch
@@ -151,7 +151,7 @@
  
 +         CompoundTag caps = serializeCaps();
 +         if (caps != null) p_20241_.m_128365_("ForgeCaps", caps);
-+         if (persistentData != null) p_20241_.m_128365_("ForgeData", persistentData);
++         if (persistentData != null) p_20241_.m_128365_("ForgeData", persistentData.m_6426_());
 +
           this.m_7380_(p_20241_);
           if (this.m_20160_()) {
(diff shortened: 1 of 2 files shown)


KnightMiner:
For the record, I'm of the opinion it's better to just remove the system in favor of capabilities, so deprecating it and tagging for removal in 1.18 might be a better approach. All the uses I had for it could be handled via data capabilities or via the native read/write methods in tile entities and entities.

sciwhiz12:
This is an updated version of #7838 for 1.17.x.
@KnightMiner: While I do agree that the custom tile entity data NBT should be revisited and likely removed in favor of custom capabilities (as is also said by Lex in the linked issue), that is a separate topic for an RFC issue/PR to tackle.

sciwhiz12 (approved):

LexManos:
Honestly should go through and mark everything deprecated for removal in 1.18.

PR #8112, opened 2021-09-22
Adds toString method to Tag to make debugging easier when tags are involved

### LINKED ISSUE #7570: Add toString method to vanilla Tag class
Because this error is not verbose is suggest adding some information to the Tag instance:

```
java.lang.IllegalStateException: Duplicate key net.minecraft.tags.Tag@35be503c
	at java.util.stream.Collectors.lambda$throwingMerger$0(Unknown Source) ~[?:1.8.0_211] {}
	at java.util.HashMap.merge(Unknown Source) ~[?:1.8.0_211] {}
	at java.util.stream.Collectors.lambda$toMap$58(Unknown Source) ~[?:1.8.0_211] {}
	at java.util.stream.ReduceOps$3ReducingSink.accept(Unknown Source) ~[?:1.8.0_211] {}
	at java.util.stream.ReferencePipeline$2$1.accept(Unknown Source) ~[?:1.8.0_211] {}
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(Unknown Source) ~[?:1.8.0_211] {}
	at java.util.stream.AbstractPipeline.copyInto(Unknown Source) ~[?:1.8.0_211] {}
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(Unknown Source) ~[?:1.8.0_211] {}
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(Unknown Source) ~[?:1.8.0_211] {}
	at java.util.stream.AbstractPipeline.evaluate(Unknown Source) ~[?:1.8.0_211] {}
	at java.util.stream.ReferencePipeline.collect(Unknown Source) ~[?:1.8.0_211] {}
	at net.minecraft.tags.TagRegistry.reinjectOptionalTags(TagRegistry.java:89) ~[forge-1.16.4-35.1.7_mapped_snapshot_20201028-1.16.3-recomp.jar:?] {re:classloading}
	at net.minecraft.tags.ITagCollectionSupplier.reinjectOptionalTags(ITagCollectionSupplier.java:66) ~[forge-1.16.4-35.1.7_mapped_snapshot_20201028-1.16.3-recomp.jar:?] {re:classloading}
```

OneLemonyBoi:
Adds toString method to Tag to make debugging easier when tags are involved and closes #7570.

### DIFF
--- a/patches/minecraft/net/minecraft/tags/SetTag.java.patch
+++ b/patches/minecraft/net/minecraft/tags/SetTag.java.patch
@@ -1,0 +1,12 @@
+--- a/net/minecraft/tags/SetTag.java
++++ b/net/minecraft/tags/SetTag.java
+@@ -59,4 +_,9 @@
+ 
+       return p_13218_;
+    }
++
++   @Override
++   public String toString() {
++      return "SetTag{" + "valuesList=" + f_13211_ + ", values=" + f_13212_ + ", closestCommonSuperType=" + f_13210_ + '}';
++   }
+ }


ChampionAsh5357:
So, `SetTag` holds three fields:
- `values`: an immutable set in all vanilla cases
- `valuesList`: a copy of `values`
- `closetCommonSuperType`: The tag type

Displaying all three of these parameters is fine, but I'm not sure how it addresses the linked issue. This does not display the tag name (since its not stored) which would be much more beneficial to know. Given the value list and tag type, you could eventually find the tag the values were part of, but if there are tags with similar values or a large number of tags, this information, in my opinion wouldn't be enough to figure out the exception thrown in the issue.

mcenderdragon:
the problem is, in most cases I had the issue/crash the Tag was empty and the error was caused ebcuase the same empty tag instance was used and thus it crashed.

KnightMiner:
There are methods to look up an ID from a tag, why not use the one that does not throw an exception if the tag has no ID? Printing that alongside a class type and maybe a value list be good debug.

I suppose the challenge is knowing which method to call though, as the method is per tag type.

sciwhiz12:
@KnightMiner: The problem with that is `SetTag` is exclusively a value holder: it has no sense of name or identity, as it is freely creatable with any set of tag values (see `SetTag#create`). So, the only way for this to work is if we patch in a field for the tag name, and patch in _each_ situation where this is created to set that field with the name of the tag that the `SetTag` 'belongs' to.

---
I've written up a more detailed explanation in a comment on the linked PR (https://github.com/MinecraftForge/MinecraftForge/issues/7570#issuecomment-978902329) on the likely cause for the stacktrace shown in the issue, and proposed a solution which seems to fix the problem in my local testing.

As it is, this PR does not provide an effective solution towards the problem presented in the linked issue, and a better solution of the same vein of this PR (as described in my reply to KnightMiner) would be more cumbersome to create, maintain, and troubleshoot in the future. A solution which directly treats the cause of the problem in the linked issue without resorting to patching this class and is (in my opinion) better in that regard has been found.

I do acknowledge that, as @ChampionAsh5357 says, there are certain situations in which knowing the contents of the `SetTag` can lead to identifying the original tag which it belongs to, but it also has its limitations as also stated.

For the above reasons, I am closing this PR. Thank you for your contributions!

PR #8114, opened 2021-09-22
[1.17.x] EntityMountEvent Infinite Client Loop Fix

ChampionAsh5357:
Port of #7125 to 1.17.

> Whenever EntityMountEvent was cancelled to dismount the player, there would be an infinite loop triggered on the client side. This is because the client player overrides the standard sneaking methods to use movement inputs instead, so the checks used to see if the player is dismounting a passenger results in something different. Movement inputs are only updated when the client player is ticked. However, since the client player is always riding a passenger, the player gets stuck in an infinite loop on the client.


GizmoTheMoonPig:
Wanted to poke this PR to see if it was ready to be pulled or not. I haven't seen any activity on it in a couple months and I do kinda need this fix for something

PR #8116, opened 2021-09-22
[RFC] Cap Rework

LexManos:
There has been a request too move away from reflection {which makes  sense because of J16} and synthetic final fields that can be null.

So here is is. The current system of @CapabilityInject in deprecated for removal in 1.18.
It still works as before, only being injected when the capability is registered. 

You now always get a non-null Capability. Using CapabiltyToken to specify the type.

```
public static Capability<IDataHolder> DATA_HOLDER_CAPABILITY = CapabilityManager.get(new CapabilityToken<>(){});
```
Note the fact that it's a anon class. This forces the compiler to include the generic information we need.
We are not going to pass in a Class object because we need to keep the soft dependency system.

The general concept is Capability now has a 'isRegistered()' function instead of testing != null.
It also has a addListener(Consumer<Capability<T>>) function, which is analogous too the old @CapabilityInject on a method.

A note about my use of @Deprecated.
Recently they added a 'since' and 'forRemoval' field to this annotation, I have gotten in the habit of setting the 'since' to the MC version in which it should be removed. Ideally in the future we'll add a build step that verifies that no @Deprecations exist for the <= the current MC version.

Re: https://github.com/MinecraftForge/MinecraftForge/pull/8104

### DIFF
--- a/fmlloader/src/main/java/net/minecraftforge/common/asm/CapabilityTokenSubclass.java
+++ b/fmlloader/src/main/java/net/minecraftforge/common/asm/CapabilityTokenSubclass.java
@@ -1,0 +1,121 @@
+/*
+ * Minecraft Forge
+ * Copyright (c) 2016-2021.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation version 2.1
+ * of the License.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.common.asm;
+
+import java.util.ArrayDeque;
+import java.util.Deque;
+import java.util.EnumSet;
+
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.Type;
+import org.objectweb.asm.signature.SignatureReader;
+import org.objectweb.asm.signature.SignatureVisitor;
+import org.objectweb.asm.tree.ClassNode;
+import org.objectweb.asm.tree.MethodNode;
+
+import cpw.mods.modlauncher.serviceapi.ILaunchPluginService;
+
+/**
+ * Removes the final modifier from fields with the @CapabilityInject annotation, prevents the JITer from in lining them so our runtime replacements can work.
+ */
+public class CapabilityTokenSubclass implements ILaunchPluginService {
+
+	private final String FUNC_NAME = "getType";
+	private final String FUNC_DESC = "()Ljava/lang/String;";
+    private final String CAP_INJECT = "net/minecraftforge/common/capabilities/CapabilityToken"; //Don't directly reference this to prevent class loading.
+
+    @Override
+    public String name() {
+        return "capability_token_subclass";
+    }
+
+    private static final EnumSet<Phase> YAY = EnumSet.of(Phase.AFTER);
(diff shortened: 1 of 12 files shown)


fmlloader/src/main/java/net/minecraftforge/common/asm/CapabilityTokenSubclass.java
```diff
@@ -0,0 +1,121 @@
...
+import org.objectweb.asm.signature.SignatureReader;
+import org.objectweb.asm.signature.SignatureVisitor;
+import org.objectweb.asm.tree.ClassNode;
+import org.objectweb.asm.tree.MethodNode;
+
+import cpw.mods.modlauncher.serviceapi.ILaunchPluginService;
+
+/**
+ * Removes the final modifier from fields with the @CapabilityInject annotation, prevents the JITer from in lining them so our runtime replacements can work.
+ */
```
diesieben07: This is still the old javadoc from the definalize transformer

src/main/java/net/minecraftforge/common/capabilities/CapabilityManager.java
```diff
@@ -55,30 +54,56 @@
      * @param type The class type to be registered
      * @deprecated use {@link RegisterCapabilitiesEvent}
      */
-    @Deprecated(forRemoval = true)
+    @Deprecated(since = "1.18", forRemoval = true)
     public <T> void register(Class<T> type)
     {
         Objects.requireNonNull(type,"Attempted to register a capability with invalid type");
-        String realName = type.getName().intern();
+        get(Type.getType(type).toString(), true);
```
diesieben07: I don't think this line is correct. `Type#toString` says it returns the descriptor, which would be `Lfoo/bar/Baz;`.
It should be
```suggestion
        get(type.getName(), true);
```
like in `RegisterCapabilitiesEvent`

src/main/java/net/minecraftforge/common/capabilities/CapabilityManager.java
```diff
@@ -55,30 +54,56 @@
...
+    }
+
+
+    public static <T> Capability<T> get(CapabilityToken<T> type)
+    {
+        return INSTANCE.get(type.getType(), false);
```
diesieben07, line 67: `CapabilityToken#getType` returns the internal name (using slashes), all other callers of `get(String, boolean)` pass the binary name (using dots) though.

src/main/java/net/minecraftforge/common/capabilities/CapabilityManager.java
```diff
@@ -55,30 +54,56 @@
...
         synchronized (providers)
         {
-            if (providers.containsKey(realName)) {
+            realName = realName.intern();
+            cap = (Capability<T>)providers.get(realName);
+            if (cap == null)
+            {
+                cap = new Capability<>(realName);
+                providers.put(realName, cap);
+            }
```
diesieben07:
```suggestion
            cap = (Capability<T>)providers.computeIfAbsent(realName, Capability::new);
```

src/main/java/net/minecraftforge/common/capabilities/CapabilityManager.java
```diff
@@ -55,30 +54,56 @@
...
+
+            if (cap.isRegistered() && registering)
```
diesieben07: This call to `isRegistered` can technically fail, because it's not guarded by a `synchronized` on the `Capability`.

diesieben07 (changes requested):
Overall I approve, just some small oversights

diesieben07 (approved):
lgtm!

fmlloader/src/main/java/net/minecraftforge/common/asm/CapabilityTokenSubclass.java
```diff
@@ -33,7 +33,20 @@
 import cpw.mods.modlauncher.serviceapi.ILaunchPluginService;
 
 /**
- * Removes the final modifier from fields with the @CapabilityInject annotation, prevents the JITer from in lining them so our runtime replacements can work.
+ * Implements getType() in CapabilityTokeen subclasses.
```
josephtarbit: Typo on CapabilityToken

LexManos:
A week of no bad comments, and nothing but good on here and discord. It should be fully bin compat. So no downside, and gives prep for future.

PR #8119, opened 2021-09-25
Add BaseSpawnerActivationEvent

TheMuso:
This event allows a mod to control the activation conditions of
net.minecraft.world.level.BaseSpawner. The event class contains documentation on
how to use the event.

This commit also contains a test mod, disabled bye default, that changes a
vanilla mob spawner activation range based on the world difficulty setting.

Rationale:
I am writing a mod where I want to control the range at which a spawner activates based on attributes that nearby players have. Whilst it is possible to modify requiredPlayerRange through NBT, that does not feel nearly as clean a solution to me, particularly since I then want to make sure that change is not written to disk, in case the user decides to remove my mod, and wants spawners to revert to vanilla behavior.

I am happy to enable the test mod by default if desired, and am also happy to split things out into separate commits if preferred.

Thank you in advance for your time and consideration.

### DIFF
--- a/patches/minecraft/net/minecraft/world/level/BaseSpawner.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/BaseSpawner.java.patch
@@ -1,5 +1,16 @@
 --- a/net/minecraft/world/level/BaseSpawner.java
 +++ b/net/minecraft/world/level/BaseSpawner.java
+@@ -62,7 +_,9 @@
+    }
+ 
+    private boolean m_151343_(Level p_151344_, BlockPos p_151345_) {
+-      return p_151344_.m_45914_((double)p_151345_.m_123341_() + 0.5D, (double)p_151345_.m_123342_() + 0.5D, (double)p_151345_.m_123343_() + 0.5D, (double)this.f_45452_);
++      net.minecraftforge.eventbus.api.Event.Result result = net.minecraftforge.event.ForgeEventFactory.canSpawnerActivate(this, p_151344_, p_151345_, this.f_45452_);
++      if (result == net.minecraftforge.eventbus.api.Event.Result.DENY) return false;
++      return result == net.minecraftforge.eventbus.api.Event.Result.ALLOW || p_151344_.m_45914_(p_151345_.m_123341_() + 0.5D, p_151345_.m_123342_() + 0.5D, p_151345_.m_123343_() + 0.5D, this.f_45452_);
+    }
+ 
+    public void m_151319_(Level p_151320_, BlockPos p_151321_) {
 @@ -127,11 +_,12 @@
                    entity.m_7678_(entity.m_20185_(), entity.m_20186_(), entity.m_20189_(), p_151312_.f_46441_.nextFloat() * 360.0F, 0.0F);
                    if (entity instanceof Mob) {
--- a/src/main/java/net/minecraftforge/event/ForgeEventFactory.java
+++ b/src/main/java/net/minecraftforge/event/ForgeEventFactory.java
@@ -134,6 +134,7 @@ import net.minecraftforge.event.entity.player.SleepingLocationCheckEvent;
 import net.minecraftforge.event.entity.player.SleepingTimeCheckEvent;
 import net.minecraftforge.event.entity.player.UseHoeEvent;
 import net.minecraftforge.event.furnace.FurnaceFuelBurnTimeEvent;
+import net.minecraftforge.event.world.SpawnerActivationEvent;
 import net.minecraftforge.event.world.BlockEvent;
 import net.minecraftforge.event.world.BlockEvent.BlockToolInteractEvent;
 import net.minecraftforge.event.world.BlockEvent.CreateFluidSourceEvent;
@@ -795,5 +796,12 @@ public class ForgeEventFactory
             return MinecraftForge.EVENT_BUS.post(new PermissionsChangedEvent(player, newLevel, oldLevel));
         }
         return false;
+    }
+
+    public static Result canSpawnerActivate(BaseSpawner spawner, Level level, BlockPos pos, int playerRange)
+    {
+        SpawnerActivationEvent event = new SpawnerActivationEvent(level, spawner, pos);
+        MinecraftForge.EVENT_BUS.post(event);
+        return event.getResult();
     }
 }
--- a/src/main/java/net/minecraftforge/event/world/SpawnerActivationEvent.java
+++ b/src/main/java/net/minecraftforge/event/world/SpawnerActivationEvent.java
@@ -1,0 +1,66 @@
+/*
+ * Minecraft Forge
+ * Copyright (c) 2016-2021.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation version 2.1
+ * of the License.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.event.world;
+
+import net.minecraft.core.BlockPos;
+import net.minecraft.world.level.BaseSpawner;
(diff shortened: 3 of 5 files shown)


ShroomAgent27 (commented):
By "activates" do you just mean the client side animations that start when a player is near a spawner block? Because right now you just overwrite the method that checks if a player is near the spawner block. On the server side, that appears to only be used to determine (somewhat randomly) how much time until the spawner block attempts to spawn a mob.

TheMuso:
On Tue, Sep 28, 2021 at 09:58:29AM AEST, Sydney wrote:
> @shroomdog27 commented on this pull request.
> 
> Just some quick comments I get around to testing this.
> 
> > +-      return p_151344_.m_45914_((double)p_151345_.m_123341_() + 0.5D, (double)p_151345_.m_123342_() + 0.5D, (double)p_151345_.m_123343_() + 0.5D, (double)this.f_45452_);
> ++      return net.minecraftforge.event.ForgeEventFactory.canBaseSpawnerActivate(this, p_151344_, p_151345_, this.f_45452_);
> 
> Vanilla code should be kept in the vanilla class, in case the algorithm changes in a future release.

Fair enough, would using the contents of net.minecraftforge.event.ForgeEventFactory.canBaseSpawnerActivate here, and eliminating the ForgeEventFactory method be appropriate?
> 
> > + * This library is free software; you can redistribute it and/or
> + * modify it under the terms of the GNU Lesser General Public
> + * License as published by the Free Software Foundation version 2.1
> + * of the License.
> + *
> + * This library is distributed in the hope that it will be useful,
> + * but WITHOUT ANY WARRANTY; without even the implied warranty of
> + * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
> + * Lesser General Public License for more details.
> + *
> + * You should have received a copy of the GNU Lesser General Public
> + * License along with this library; if not, write to the Free Software
> + * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
> + */
> +
> +package net.minecraftforge.event;
> 
> This should be in the `world` subpackage, since it has to do with blocks.

Yep, was not quite sure where to put it, will change.

TheMuso:
Activation is the best term I could think of to refer to the behavior of spawners starting to function when the player is within 16 blocks. I know that this is relevant for both client and server side.

patches/minecraft/net/minecraft/world/level/BaseSpawner.java.patch
```diff
@@ -1,5 +1,14 @@
 --- a/net/minecraft/world/level/BaseSpawner.java
 +++ b/net/minecraft/world/level/BaseSpawner.java
+@@ -62,7 +_,7 @@
+    }
+ 
+    private boolean m_151343_(Level p_151344_, BlockPos p_151345_) {
+-      return p_151344_.m_45914_((double)p_151345_.m_123341_() + 0.5D, (double)p_151345_.m_123342_() + 0.5D, (double)p_151345_.m_123343_() + 0.5D, (double)this.f_45452_);
++      return net.minecraftforge.event.ForgeEventFactory.canBaseSpawnerActivate(this, p_151344_, p_151345_, this.f_45452_);
```
ShroomAgent27: > Fair enough, would using the contents of net.minecraftforge.event.ForgeEventFactory.canBaseSpawnerActivate here, and eliminating the ForgeEventFactory method be appropriate?

Personally I would do something along the lines of

```java
Result result = ForgeEventFactory.canBaseSpawnerActivate(params);
if (result == Result.DENY) return false;
return vanilla code || result == Result.ALLOW;
```

Formatted to be in compliance with the patch formatting, of course.

src/main/java/net/minecraftforge/event/BaseSpawnerActivationEvent.java
```diff
@@ -0,0 +1,66 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.event;
```
ShroomAgent27: Posted this before, but yeah, this needs to be in the ` net.minecraftforge.event.world` package.

src/test/java/net/minecraftforge/debug/BaseSpawnerActivationEventTest.java
```diff
@@ -0,0 +1,73 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.debug;
```
ShroomAgent27: Posted this before, but yeah, this needs to be in the ` net.minecraftforge.debug.world` package.

ShroomAgent27 (commented):
Here, I'll add my comments that because I had a dumb moment and deleted them. If I could restore them, I would. My bad!

In addition, note that this new event would be fired every tick (as the method you are adding an event too is called every tick by both the server and client).

This has the potential to be pretty resource demanding if mods are doing a lot of calculations during the check.

TheMuso:
I am well aware that this runs every tick, and could be a performance issue with calculations going on etc. If it is ultimately decided that this is not desirable, I will have to reconsider what I do in my mod.

patches/minecraft/net/minecraft/world/level/BaseSpawner.java.patch
```diff
@@ -1,5 +1,16 @@
 --- a/net/minecraft/world/level/BaseSpawner.java
 +++ b/net/minecraft/world/level/BaseSpawner.java
+@@ -62,7 +_,9 @@
+    }
+ 
+    private boolean m_151343_(Level p_151344_, BlockPos p_151345_) {
+-      return p_151344_.m_45914_((double)p_151345_.m_123341_() + 0.5D, (double)p_151345_.m_123342_() + 0.5D, (double)p_151345_.m_123343_() + 0.5D, (double)this.f_45452_);
++      net.minecraftforge.eventbus.api.Event.Result result = net.minecraftforge.event.ForgeEventFactory.canBaseSpawnerActivate(this, p_151344_, p_151345_, this.f_45452_);
++      if (result == net.minecraftforge.eventbus.api.Event.Result.DENY) return false;
++      return p_151344_.m_45914_(p_151345_.m_123341_() + 0.5D, p_151345_.m_123342_() + 0.5D, p_151345_.m_123343_() + 0.5D, this.f_45452_) || result == net.minecraftforge.eventbus.api.Event.Result.ALLOW;
```
Shadows-of-Fire: This should be short-circuiting to check the result first, not after, as the least expensive calls should proceed the more expensive ones.

ShroomAgent27:
I wrote another test mod as of commit `e163a07a9435ecd5a9c5ebb53161cde3bfc3db62`.

https://github.com/FoalFreePress/TestingForgeMods/blob/main/8119/src/main/java/com/example/examplemod/ExampleMod.java

This PR works as advertised.

Shadows-of-Fire (changes requested):
Aside from changing the order of the conditional, the only other thing I think would make sense is having the name changed to `SpawnerActivationEvent`, since the event refers to the activation status of spawners in-world.  The "Base" prefix is on `BaseSpawner` since it is an abstract class.

TheMuso:
All suggested changes made in the latest push.

LexManos:
I am concerned about the fact that this is yet another tick event.
At the very least it should be hidden behind the vanilla check first.
If you want to extend the range, then do so in the spawner.
Allowing things to lower the range sounds more like the usecases you want.

TheMuso:
Yes, for my use case, I am only interested in lowering the range of the spawner. I initially started working on a solution that only took the lowering of the range into account. However, I thought that there may be mods in the future that would wish to increase the range.

In light of the multiple concerns raised in this discussion, I will rework the event, also taking Lex's request into consideration, that being allowing the vanilla check to evaluate as true first.

Thanks again to all for your input.

sciwhiz12:
Hello! @TheMuso, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable. If we do not hear back from you after some time, this PR will be closed for inactivity.

TheMuso:
Will close for now, my mod in on the backburner, and am also unlikely to follow up on this any time soon.

PR #8121, opened 2021-09-26
Add ability for people to add custom pack finders (RepositorySources)

gigaherz:
* Adds AddPackFindersEvent which lets you add additional resource/data pack sources to the game.
* Adds PathResourcePack as a utility to allow easy providing of a subfolder in a mod as a resource pack.

I realized after writing the code, when I was preparing the PR, that there was a 1.16 PR implementing an equivalent feature in #6919


src/main/java/net/minecraftforge/fmllegacy/packs/PathResourcePack.java
```diff
@@ -0,0 +1,52 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.fmllegacy.packs;
+
+import java.nio.file.Path;
+
+public class PathResourcePack extends BasePathResourcePack
```
ChampionAsh5357: This seems good from my testing. Only comment I'll add here is that this should probably have some documentation that the source passed in references the root folder of the associated nested resource/data pack.

gigaherz: Oops yes I forgot to document this class.

ChampionAsh5357 (approved):
Looks good to me.

src/main/java/net/minecraftforge/fmlclient/ClientHooks.java
```diff
@@ -266,6 +267,10 @@ public static void logMissingTextureErrors()
                         List<IModInfo> mods = modRP.getModFile().getModInfos();
                         logger.error("      mod(s) {} resources at {}", mods.stream().map(IModInfo::getDisplayName).collect(Collectors.toList()), modRP.getModFile().getFilePath());
                     }
+                    else if (resPack instanceof PathResourcePack)
+                    {
+                        logger.error("      resource pack at path {}", ((PathResourcePack)resPack).getSource());
```
LexManos: As I said in discord, we should move this to just resPack.toString() because these type checks are exposing internal classes for no good reason.

gigaherz: Forgot that bit. Will change in the morning.

PR #8127, opened 2021-09-29
Bump SecureJarHandler to 0.9.54

DaemonUmbra:


malte0811:
Did you test this with multi-source-set mods? I'm getting [this](https://gist.github.com/malte0811/d9b5e2a5e4f99c1214f1d4bf285131e6) error when trying to run `./gradlew runClient` with a slightly modified MDK (see [diff](https://gist.github.com/malte0811/392f2d0ebab4c66d22fc964fcc443281)).

DaemonUmbra:
> Did you test this with multi-source-set mods? I'm getting [this](https://gist.github.com/malte0811/d9b5e2a5e4f99c1214f1d4bf285131e6) error when trying to run `./gradlew runClient` with a slightly modified MDK (see [diff](https://gist.github.com/malte0811/392f2d0ebab4c66d22fc964fcc443281)).

I've seen issues like this before, it doesn't check if the resources folder exists (which it doesn't because you don't have any resources in that sourceSet) before trying to search it for a mods.toml and failing

TL;DR: Issue is unrelated to this PR

Issue #7253, opened 2020-08-14
Breaking waterlogged blocks occasionally displays air for a frame


zachman42:
**Minecraft Version:** 1.16.2
**Forge Version:** 33.0.3

**Steps to Reproduce:**
 1. place a waterloggable block underwater
 2. break it
 3. you may have to do this a couple times to see the effect

**Description of issue:**
Title says it all.
![ezgif-4-425c9733913c](https://user-images.githubusercontent.com/68206966/90282923-d287c400-de34-11ea-82a3-aba2813d4b36.gif)

ChampionAsh5357:
This is a result of `ForgeHooks#onBlockBreakEvent` not taking into account the `FluidState` before setting the block to air for non-tile entities. Currently, it just sets sends a packet to the client setting the block to air. So, this probably just needs to send over a `BlockState` of any associated `FluidState` the original block had.

PR #8128, opened 2021-09-29
[1.17.x] Make Broken Blocks Display Associated FluidState

ChampionAsh5357:
This addresses #7253 for 1.17. When breaking a block within `ForgeHooks#onBlockBreakEvent`, Forge tells the client the block is gone for non-block entities. However, this just sets the block to air through the use of a dummy world. This would be fine normally; however, the addition of waterlogged blocks means that the `FluidState` of the block must also be taken into consideration.

This PR fixes that by returning the `BlockState` of the current fluid at the position. This also removes the `DUMMY_WORLD` as it seems it was only implemented for this express purpose (see [initial commit](https://github.com/MinecraftForge/MinecraftForge/commit/cdfa7caaeb85283f33e0a7d70840fa2e53568cbd#diff-a0412d7ac689ab1c48becee0dd0e7d0c2aea7fb6f82d2985cb124268ff7b1909L844-R803)).

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -536,7 +536,7 @@ public class ForgeHooks
         // Tell client the block is gone immediately then process events
         if (world.getBlockEntity(pos) == null)
         {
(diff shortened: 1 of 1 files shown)


ShroomAgent27:
This fix works and doesn't cause any issues with existing worlds.

ShroomAgent27 (approved):

PR #8129, opened 2021-09-29
[1.16.x] Make Broken Blocks Display Associated FluidState

ChampionAsh5357:
Backported version of #8128.

> When breaking a block within ForgeHooks#onBlockBreakEvent, Forge tells the client the block is gone for non-block entities. However, this just sets the block to air through the use of a dummy world. This would be fine normally; however, the addition of waterlogged blocks means that the FluidState of the block must also be taken into consideration.


ShroomAgent27:
It's the same as the 1.17 PR. This works as well. 

Jenkins PR test failed because another build was running at the same time. Unlucky!

ShroomAgent27 (approved):

PR #8132, opened 2021-10-01
Fix Signs breaking with custom `WoodType`s

### LINKED ISSUE #8131: [1.17] New issue with creating Custom Sign WoodTypes
**Minecraft Version:** 1.17.1

**Forge Version:** 37.0.0+

**Description:** Due to the new model system implemented by Mojang in 1.17.1, a new issue has popped up with trying to add custom signs using the features added in this PR: https://github.com/MinecraftForge/MinecraftForge/pull/7623
The method `ModelLayerLocation#createSignModelName()` is now hardcoded to use `minecraft` as the namespace when creating a new `ResourceLocation` for a given `WoodType` causing a `ResourceLocationException` to be thrown when initializing the `ResourceLocation`.

**Solution:**
Patch `ModelLayerLocation#createSignModelName()` to use of the namespace and path of a given `WoodType`.
```Java
public static ModelLayerLocation createSignModelName(WoodType type) {
      ResourceLocation location = new ResourceLocation(type.name());
      return new ModelLayerLocation(new ResourceLocation(location.getNamespace(), "entity/signs/" + location.getPath()), "main"));
}
```

ShroomAgent27:
This PR Fixes #8131 by implementing the solution that @SmellyModder suggested. This has been tested using the [CustomSignsTest](https://github.com/MinecraftForge/MinecraftForge/blob/166feefc023c7e153072a18e6b49982d7246fb0a/src/test/java/net/minecraftforge/debug/block/CustomSignsTest.java)

And the game loads just fine.
3iGlKbR

This was the only mod I was able to test, because BiomesOPlenty doesn't add custom signs, Biomes You'll Go doesn't have a Forge version for 1.17.x, and TwilightForest uses their own sign renderer.

### DIFF
--- a/patches/minecraft/net/minecraft/client/model/geom/ModelLayers.java.patch
+++ b/patches/minecraft/net/minecraft/client/model/geom/ModelLayers.java.patch
@@ -6,7 +6,7 @@
     public static ModelLayerLocation m_171291_(WoodType p_171292_) {
 -      return m_171300_("sign/" + p_171292_.m_61846_(), "main");
 +      ResourceLocation location = new ResourceLocation(p_171292_.m_61846_());
-+      return new ModelLayerLocation(new ResourceLocation(location.m_135827_(), "signs/" + location.m_135815_()), "main");
++      return new ModelLayerLocation(new ResourceLocation(location.m_135827_(), "sign/" + location.m_135815_()), "main");
     }
  
     public static Stream<ModelLayerLocation> m_171288_() {
--- a/patches/minecraft/net/minecraft/client/model/geom/ModelLayers.java.patch
+++ b/patches/minecraft/net/minecraft/client/model/geom/ModelLayers.java.patch
@@ -6,7 +6,7 @@
     public static ModelLayerLocation m_171291_(WoodType p_171292_) {
 -      return m_171300_("sign/" + p_171292_.m_61846_(), "main");
 +      ResourceLocation location = new ResourceLocation(p_171292_.m_61846_());
-+      return new ModelLayerLocation(new ResourceLocation(location.m_135827_(), "entity/signs/" + location.m_135815_()), "main");
++      return new ModelLayerLocation(new ResourceLocation(location.m_135827_(), "signs/" + location.m_135815_()), "main");
     }
  
     public static Stream<ModelLayerLocation> m_171288_() {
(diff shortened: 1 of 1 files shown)


SmellyModder (approved):

patches/minecraft/net/minecraft/client/model/geom/ModelLayers.java.patch
```diff
@@ -0,0 +1,12 @@
+--- a/net/minecraft/client/model/geom/ModelLayers.java
++++ b/net/minecraft/client/model/geom/ModelLayers.java
+@@ -199,7 +_,8 @@
+    }
+ 
+    public static ModelLayerLocation m_171291_(WoodType p_171292_) {
+-      return m_171300_("sign/" + p_171292_.m_61846_(), "main");
++      ResourceLocation location = new ResourceLocation(p_171292_.m_61846_());
++      return new ModelLayerLocation(new ResourceLocation(location.m_135827_(), "entity/signs/" + location.m_135815_()), "main");
```
ChampionAsh5357: So I took a better look at this now as a reviewer. The solution in issue in #8131 while although valid is for the wrong reason. The error being displayed is a result of the `WoodType` having a `:` in the string passed into the path. Normally this wouldn't be an issue as vanilla would parse out the `:`. However, the additional prefix and delegation make the path invalid. If the entire string was considered, the namespace would instead error from `/`.

The `ModelLayerLocation` just represents a unique name for the root model. As such, this could have just replaced all `:` with `_` and worked as well. However, removing the hardcoded instance is always a better idea. As such, the initial prefix should be instead `sign/` instead of `entity/signs/` as that is indicative of the texture location and changes what the model name is otherwise.

ShroomAgent27: Done. Tests still work as well. Even loaded the first world I used.

patches/minecraft/net/minecraft/client/model/geom/ModelLayers.java.patch
```diff
@@ -0,0 +1,12 @@
+--- a/net/minecraft/client/model/geom/ModelLayers.java
++++ b/net/minecraft/client/model/geom/ModelLayers.java
+@@ -199,7 +_,8 @@
+    }
+ 
+    public static ModelLayerLocation m_171291_(WoodType p_171292_) {
+-      return m_171300_("sign/" + p_171292_.m_61846_(), "main");
++      ResourceLocation location = new ResourceLocation(p_171292_.m_61846_());
++      return new ModelLayerLocation(new ResourceLocation(location.m_135827_(), "signs/" + location.m_135815_()), "main");
```
ChampionAsh5357: Again, `sign/` not `signs/`. It shouldn't change any model layer name from vanilla.

ShroomAgent27: Typos be like

ChampionAsh5357 (approved):

PR #8133, opened 2021-10-01
Changed control flow of post mouse events

### LINKED ISSUE #6060: [1.15.2]GuiScreenEvent.MouseClickedEvent.Post not fired for ContainerScreen
**Minecraft Version:** 1.15.2

**Forge Version:** 31.1.35

**Steps to Reproduce:**
 1. Record a listener for this specific case:
`private void clickInContainer(GuiScreenEvent.MouseClickedEvent.Post event){
System.out.println("Mouse click");
}`
 2. Open a workbench or furnace or whatever block inventory
 3. Click anywhere

**Description of issue:**
I used this event in 1.12 for handling shift clicks on slots for various crafting screens (vanilla and modded ones). I prefer the post action event to have an accurate state of the inventory contents, and thus minimize conflict.
From what i can tell ContainerScreen#mouseClicked(double,double,int) always returns true, meaning the patch in MouseHelper is without effect.

Harleyoc1:
Changes control flow of `GuiScreenEvent$MouseClickedEvent$Post` and `GuiScreenEvent$MouseReleasedEvent$Post` as suggested by @Shadows-of-Fire to fix #6060. Overview is this:

- Fire pre event, set `aboolean[0]` (`handled`) to return value.
- If pre event was not cancelled (`!aboolean[0]`): 
  - Call regular `mouseClicked` method, set `aboolean[0]` to return value.
  - Fire post event (now sending `aboolean[0]` as a new `handled` field for the event). 
  - If post event result was default: 
    - Set `aboolean[0]` to use handled from `mouseClicked` return. 
  - Else if event result was allow:
    - Force `aboolean[0]` to `true`. 
  - Else if event result was deny:
    - Force `aboolean[0]` to `false`. 

The reason for this change is that it gives the post event a wider variety of use cases, such as that described by the OP of #6060, rather than only firing if the mouse event wasn't handled by vanilla. 

This PR does not attempt to fully clean up the events since there is already a PR for this which would need to be merged at some point on either end anyway. Only necessary Javadoc changes for accuracy of how the event now works are included.

### DIFF
--- a/patches/minecraft/net/minecraft/client/MouseHandler.java.patch
+++ b/patches/minecraft/net/minecraft/client/MouseHandler.java.patch
@@ -8,21 +8,25 @@
           boolean[] aboolean = new boolean[]{false};
           if (this.f_91503_.m_91265_() == null) {
              if (this.f_91503_.f_91080_ == null) {
-@@ -82,11 +_,15 @@
+@@ -82,11 +_,19 @@
                 if (flag) {
                    screen.m_169415_();
                    Screen.m_96579_(() -> {
 -                     aboolean[0] = screen.m_6375_(d0, d1, i);
 +                     aboolean[0] = net.minecraftforge.client.ForgeHooksClient.onGuiMouseClickedPre(this.f_91503_.f_91080_, d0, d1, i);
-+                     if (!aboolean[0]) aboolean[0] = this.f_91503_.f_91080_.m_6375_(d0, d1, i);
-+                     if (!aboolean[0]) aboolean[0] = net.minecraftforge.client.ForgeHooksClient.onGuiMouseClickedPost(this.f_91503_.f_91080_, d0, d1, i);
++                     if (!aboolean[0]) {
++                        aboolean[0] = this.f_91503_.f_91080_.m_6375_(d0, d1, i);
++                        aboolean[0] = net.minecraftforge.client.ForgeHooksClient.onGuiMouseClickedPost(this.f_91503_.f_91080_, d0, d1, i, aboolean[0]);
++                     }
                    }, "mouseClicked event handler", screen.getClass().getCanonicalName());
                 } else {
                    Screen.m_96579_(() -> {
 -                     aboolean[0] = screen.m_6348_(d0, d1, i);
 +                     aboolean[0] = net.minecraftforge.client.ForgeHooksClient.onGuiMouseReleasedPre(this.f_91503_.f_91080_, d0, d1, i);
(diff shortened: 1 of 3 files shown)


Shadows-of-Fire:
Should we maybe give the Post events `@HasResult` instead of `@Cancellable` and allow them to consume (`ALLOW`), pass (`DEFAULT`) or force-unconsume (`DENY`) instead of the current cancel to force consume or do nothing?

Shadows-of-Fire:
Do note that any such changes must retain binary compat (The event must stay `@Cancellable` until 1.18), but it can be edited such that `Event#setCancelled` simply manipulates the result instead of removing the annotation for now.

Shadows-of-Fire:
Looks good to me

LexManos:
I am not fully understanding why you guys want to override the vanilla handeling after the fact.
Why does post need to be called when the event has already been processed?

Shadows-of-Fire:
I don't have the answer to that myself.  Author of #6060 might.
@GotoLink What was your previous usecase for the way the events fired in 1.12?

Harleyoc1:
Yeah OP would have to comment on specific use case. My understanding is that the OP wanted to listen for shift clicks on container slots (regardless of whether or not vanilla handled it, which it presumably always will have if successful), and the post event was needed because they wanted the state of an inventory's contents after it had been changed by the default `moseClicked` behaviour.

GotoLink:
Pretty much what  Harleyoc1 said.
My use case is not to override but to track the action and get the inventory contents right after the action has been handled. 
To be exact, a shift-click at the output slot in a crafting inventory.

LexManos:
My concern is that this will trigger the event handlers for things that it shouldn't that already exist.
Causing weird behavior in existing mods.
The best option would be to delay this until a breaking change in 1.18.
The second best option would be to keep canceleable and pre-cancel the event.
So that existing handlers that don't have receiveCanceled wont get the event when they dont expect it.
I am still not convinced of the usecase either. You have inventory change and other events to get the state of inventories. Why do you need to do it in the mouse handler? What functionality are you trying to acomplish?

Shadows-of-Fire:
Pre-Cancellation works, as far as binary compat is concerned.  The mod in question is likely a clientside utility mod that needs to track inventory state for some reason or another (Link would have to clarify).
What inventory events even exist to track that behavior?

Shadows-of-Fire:
@Harleyoc1 Can you update this PR to include the pre-cancellation changes as mentioned earlier?

PR #8134, opened 2021-10-02
Add missing call to Entity#onAddedToWorld for non-player entities server-side

maxanier:
Currently `IForgeEntity#onAddedToWorld` is not called on server side (`ServerLevel`) for non-player entities. As far as I can tell this is unintentional and differs from 1.16.

The vanilla `ServerLevel` now handles `addEntity` and `addPlayer` completely separately. Previously `addPlayer` also called `addEntity` (which included the `onAddedToWorld` patch). In 1.17 two separate patches are needed and currently only `addPlayer` is patched.

On client-side (`ClientLevel`) nothing has changed. Same for 'onRemovedFromWorld`.


ShroomAgent27 (approved):
Looks good to me.

PR #8135, opened 2021-10-03
1.17.x client chat received on client

coehlrich:
Adds an event for messages that were either received from the server or from messages that are put in chat from client-side mods or debug messages while also allowing the change the id of the message so that if a message with the same id as another id get's received then the older message get's removed.

The vanilla method that accept's the message id is currently private.

I feel like there would be a better name for the event than `ClientChatReceivedBothSidesEvent` but `ClientChatReceivedEvent` is already used which is the event that fires only for messages from the server.


diesieben07:
What is the use case for this?

src/main/java/net/minecraftforge/client/event/ClientChatReceivedBothSidesEvent.java
```diff
@@ -0,0 +1,66 @@
...
+
+import net.minecraft.network.chat.Component;
+import net.minecraftforge.eventbus.api.Cancelable;
+import net.minecraftforge.eventbus.api.Event;
+
+/**
+ * Fired when a message is received on the client from either the server or the same client
+ */
+@Cancelable
+public class ClientChatReceivedBothSidesEvent extends Event
```
diesieben07, line 30: The javadoc as well as the name of this event do not at all accurately describe what it does.

src/main/java/net/minecraftforge/client/event/ClientChatReceivedBothSidesEvent.java
```diff
@@ -0,0 +1,66 @@
...
+     */
+    public int getId()
+    {
+        return id;
+    }
+
+    /*
+     * Sets the id of the message that is being received. If there is another message with the same id that was already received then that message will be removed
+     */
+    public void setId(int id)
```
diesieben07, line 62: What is the use-case for setting the ID on a message via this event?

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -149,6 +139,17 @@
...
+
+import java.io.File;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Random;
+import java.util.UUID;
+import java.util.function.Consumer;
+
 public class ForgeEventFactory
```
diesieben07: Do not reorganize the imports in this file.

diesieben07 (changes requested):
I am really not sure what the use-case for this event would be. Please elaborate.

src/main/java/net/minecraftforge/client/event/ClientChatReceivedBothSidesEvent.java
coehlrich: I was trying to think of a better name for the event (which would probably also help with the javadoc) but couldn't think of anything other than `ClientChatReceived` which is already used for the event that only receives messages from the server.

coehlrich:
For the use-case I was thinking it would be useful for chat message modification's that would fit for both message's coming from the server, debug keys, and other client-side mods like adding timestamps to the start of chat messages.

About the message id the use-case I had in mind was where it could be used for if a message from a server would better fit replacing the previous message of it's kind which since server's can't use the message id system it can't do that itself.

sciwhiz12:
Hello! @coehlrich, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable. If we do not hear back from you after some time, this PR will be closed for inactivity.

sciwhiz12:

PR #8136, opened 2021-10-03
[1.17.x] crude fix for gamelibraries

Minecraftschurli:
Currently libraries with the FMLModType of GAMELIBRARY do not load from legacyClasspath and otherwise error because they don't have a mods.toml . This PR proposes a fix that creates a ModFileInfo from a manifest for gamelibraries. It also makes it so the fields FMLModLoader, FMLLoaderVersion and LICENSE are required to be set in the manifest for gamelibraries.

To test this I published to my mavenLocal and used it in a mod that depends on a gamelibrary (which loads without any problems with these changes).

### DIFF
--- a/fmlloader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
+++ b/fmlloader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
@@ -21,16 +21,13 @@ package net.minecraftforge.fml.loading.moddiscovery;
 
 import com.google.common.base.Strings;
 import cpw.mods.modlauncher.api.LamdbaExceptionUtils;
-import cpw.mods.util.LambdaExceptionUtils;
 import net.minecraftforge.fml.loading.StringUtils;
 import net.minecraftforge.forgespi.language.IConfigurable;
 import net.minecraftforge.forgespi.language.IModFileInfo;
 import net.minecraftforge.forgespi.language.IModInfo;
 import net.minecraftforge.forgespi.language.MavenVersionAdapter;
-import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.apache.maven.artifact.versioning.VersionRange;
 
 import javax.security.auth.x500.X500Principal;
 import java.net.URL;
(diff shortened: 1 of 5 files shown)


ChampionAsh5357:
So, this is going to need a much larger explanation. There's already awareness of the bug here, but a few things need to be added:
- What is the problem you are trying to solve in specific detail?
- How can we verify and test that problem and your solution?
- Why should this crude solution be accepted over a proper one?
It needs a much larger explanation for those who are trying to review and verify your work; otherwise, we have to spend more time piecing together on assumptions your understanding and reasoning.

Minecraftschurli:
1. Updated the description
2. I called it crude because it is small and changes just as much as absolutely needed

fmlloader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
```diff
@@ -62,6 +60,40 @@
...
+                return List.of();
+            }
+        };
+        var modLoader = this.config.<String>getConfigElement("FMLModLoader").orElseThrow(() -> new InvalidModFileException("Missing ModLoader in file", this));
+        var modLoaderVersion = this.config.<String>getConfigElement("FMLLoaderVersion").map(MavenVersionAdapter::createFromVersionSpec).orElseThrow(() -> new InvalidModFileException("Missing ModLoader version in file", this));
+        this.languageSpecs = new ArrayList<>(List.of(new LanguageSpec(modLoader, modLoaderVersion)));
+        this.license = this.config.<String>getConfigElement("LICENSE").orElse("");
+        this.issueURL = this.config.<String>getConfigElement("ISSUES-URL").map(StringUtils::toURL).orElse(null);
+        this.moduleName = this.config.<String>getConfigElement(Attributes.Name.IMPLEMENTATION_TITLE.toString()).orElseThrow(() -> new InvalidModFileException("Missing title in file", this));
+        this.versionString = this.config.<String>getConfigElement(Attributes.Name.IMPLEMENTATION_VERSION.toString()).orElseThrow(() -> new InvalidModFileException("Missing version in file", this));
```
LexManos, line 91: My main concern is the requirements added to the meta file.
This also doesn't solve the normal LIBRARY requirements.
Also, moduleName needs to take Automatic-module-name or module-info into account.
Mainly I want to get a proper spec written up about what data is needed to be in the META-INF vs the mods.toml.

PR #5969, opened 2019-07-24
Modded River biomes

valoeghese:
Simple change that lets modders specify river biomes through overriding a method in your Biome's class
Being able to specify river biomes is very useful for biome mods. It means the said mods can:

- Set the rivers in their snowy biomes to be frozen
- Remove rivers in biomes in which they look unusual
- Add custom rivers to themed biomes

The code has been tested and it works.

Please note any changes I should make to this pr. Thanks.

PR #8137, opened 2021-10-03
Reintroduce modded biome variants

maxanier:
This PR reintroduces a way to register/provide certain biome variants for your own biomes (or for vanilla or other modded biomes).
This e.g. allows having a hills variant of your biome seamlessly integrate into the world.

## Biome layer and variants in vanilla
MC first generates a "biome layer" based on the biomes with a biome entry/weight which specifies the biome for each chunk. Then it applies several "layers" that can modify the biome layer. For example the `RegionHillsLayer` can replace some biome chunks within a larger biome with a hills variant of that biome (`desert` -> `desert_hills`).
These modifications work on hardcoded biome ids in the respective layer class.

## Previous work
In 1.15 three of these layers were patched to allow mods to introduce their own biome variants
Hills: https://github.com/MinecraftForge/MinecraftForge/pull/6571
River: https://github.com/MinecraftForge/MinecraftForge/pull/5969
Shore: https://github.com/MinecraftForge/MinecraftForge/pull/7001

They all rely on new methods in the `Biome` class which can be overridden to return the respective variant for the modded biome.
Since with 1.16.2 modded biomes do not use a subclass of `Biome`, a new method is needed.

## How this works
The patches to the biome modification layers remain functionally unchanged as nothing has changed here. Only analog patch to `BiomeEdgeLayer` is added.
Instead of querying the respective Biome, Forge's `BiomeManager` class is now used to get the biome variants.
If a modder wants to provide biome variants to their own (or other biomes), they can register a `IBiomeLayerModifier` and override the respective method to provide the desired variant.
The integer biome ids that are used in the layer classes are translated to the `ResourceKey` id of the biome.

## Test
This PR brings a (disabled-by-default) test mod, that introduces a new biome (added with a `BiomeEntry` to be generated) and hills, edge, shore and river variant.

## Discussion
Things to think about:
- Use `null` instead of `Optional` in `IBiomeLayerModifier` return
- Instead of registering a single `IBiomeLayerModifier` that is queryied for any biome, register biome specific `IBiomeLayerModifier`

### DIFF
--- a/patches/minecraft/net/minecraft/world/level/newbiome/layer/BiomeEdgeLayer.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/newbiome/layer/BiomeEdgeLayer.java.patch
@@ -1,11 +1,10 @@
 --- a/net/minecraft/world/level/newbiome/layer/BiomeEdgeLayer.java
 +++ b/net/minecraft/world/level/newbiome/layer/BiomeEdgeLayer.java
-@@ -19,7 +_,11 @@
-                   return 23;
+@@ -20,6 +_,11 @@
                 }
              }
--
-+            net.minecraftforge.registries.ForgeRegistry<net.minecraft.world.level.biome.Biome> registry = (net.minecraftforge.registries.ForgeRegistry<net.minecraft.world.level.biome.Biome>)net.minecraftforge.registries.ForgeRegistries.BIOMES;
+ 
++            net.minecraftforge.registries.ForgeRegistry<net.minecraft.world.level.biome.Biome> registry = (net.minecraftforge.registries.ForgeRegistry<net.minecraft.world.level.biome.Biome>) net.minecraftforge.registries.ForgeRegistries.BIOMES;
 +            java.util.Optional<net.minecraft.resources.ResourceKey<net.minecraft.world.level.biome.Biome>> edge = net.minecraftforge.common.BiomeManager.getEdgeBiome(registry.getKey(p_76669_), registry.getKey(p_76665_), registry.getKey(p_76666_), registry.getKey(p_76667_), registry.getKey(p_76668_));
 +            if(edge.isPresent()) {
 +               return net.minecraft.data.BuiltinRegistries.f_123865_.m_7447_(net.minecraft.data.BuiltinRegistries.f_123865_.m_6246_(edge.get()));
--- a/patches/minecraft/net/minecraft/world/level/newbiome/layer/BiomeEdgeLayer.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/newbiome/layer/BiomeEdgeLayer.java.patch
@@ -1,0 +1,15 @@
+--- a/net/minecraft/world/level/newbiome/layer/BiomeEdgeLayer.java
++++ b/net/minecraft/world/level/newbiome/layer/BiomeEdgeLayer.java
+@@ -19,7 +_,11 @@
+                   return 23;
+                }
+             }
+-
++            net.minecraftforge.registries.ForgeRegistry<net.minecraft.world.level.biome.Biome> registry = (net.minecraftforge.registries.ForgeRegistry<net.minecraft.world.level.biome.Biome>)net.minecraftforge.registries.ForgeRegistries.BIOMES;
++            java.util.Optional<net.minecraft.resources.ResourceKey<net.minecraft.world.level.biome.Biome>> edge = net.minecraftforge.common.BiomeManager.getEdgeBiome(registry.getKey(p_76669_), registry.getKey(p_76665_), registry.getKey(p_76666_), registry.getKey(p_76667_), registry.getKey(p_76668_));
++            if(edge.isPresent()) {
++               return net.minecraft.data.BuiltinRegistries.f_123865_.m_7447_(net.minecraft.data.BuiltinRegistries.f_123865_.m_6246_(edge.get()));
++            }
+             return p_76669_;
+          } else {
+             return 34;
--- a/patches/minecraft/net/minecraft/world/level/newbiome/layer/RegionHillsLayer.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/newbiome/layer/RegionHillsLayer.java.patch
@@ -1,15 +1,13 @@
 --- a/net/minecraft/world/level/newbiome/layer/RegionHillsLayer.java
 +++ b/net/minecraft/world/level/newbiome/layer/RegionHillsLayer.java
-@@ -51,7 +_,11 @@
+@@ -51,6 +_,10 @@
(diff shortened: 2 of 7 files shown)


patches/minecraft/net/minecraft/world/level/newbiome/layer/BiomeEdgeLayer.java.patch
```diff
@@ -0,0 +1,15 @@
+--- a/net/minecraft/world/level/newbiome/layer/BiomeEdgeLayer.java
++++ b/net/minecraft/world/level/newbiome/layer/BiomeEdgeLayer.java
+@@ -19,7 +_,11 @@
+                   return 23;
+                }
+             }
+-
```
MelanX: I think you should re-add this empty line to make the patch smaller

patches/minecraft/net/minecraft/world/level/newbiome/layer/BiomeEdgeLayer.java.patch
```diff
@@ -0,0 +1,15 @@
...
++            net.minecraftforge.registries.ForgeRegistry<net.minecraft.world.level.biome.Biome> registry = (net.minecraftforge.registries.ForgeRegistry<net.minecraft.world.level.biome.Biome>)net.minecraftforge.registries.ForgeRegistries.BIOMES;
++            java.util.Optional<net.minecraft.resources.ResourceKey<net.minecraft.world.level.biome.Biome>> edge = net.minecraftforge.common.BiomeManager.getEdgeBiome(registry.getKey(p_76669_), registry.getKey(p_76665_), registry.getKey(p_76666_), registry.getKey(p_76667_), registry.getKey(p_76668_));
```
pupnewfster: Is there a reason you are getting the registry here rather than just passing the parameters to the `getEdgeBiome` method and then grabbing the registry and looking up parts of the registry based on those passed in parameters there?

patches/minecraft/net/minecraft/world/level/newbiome/layer/RegionHillsLayer.java.patch
```diff
@@ -0,0 +1,15 @@
...
+@@ -51,7 +_,11 @@
+       } else {
+          if (p_76838_.m_5826_(3) == 0 || k == 0) {
+             int l = i;
+-            if (i == 2) {
++            java.util.Optional<net.minecraft.resources.ResourceKey<net.minecraft.world.level.biome.Biome>> hills = net.minecraftforge.common.BiomeManager.getHillsBiome(((net.minecraftforge.registries.ForgeRegistry<net.minecraft.world.level.biome.Biome>)net.minecraftforge.registries.ForgeRegistries.BIOMES).getKey(i));
++            if(hills.isPresent()) {
++               l = net.minecraft.data.BuiltinRegistries.f_123865_.m_7447_(net.minecraft.data.BuiltinRegistries.f_123865_.m_6246_(hills.get()));
++            }
++            else if (i == 2) {
```
pupnewfster: This probably should be: 
```java
} else
if (i == 2) {
```
So as to not have this patch having to remove any lines and just add some.

patches/minecraft/net/minecraft/world/level/newbiome/layer/ShoreLayer.java.patch
```diff
@@ -0,0 +1,14 @@
+--- a/net/minecraft/world/level/newbiome/layer/ShoreLayer.java
++++ b/net/minecraft/world/level/newbiome/layer/ShoreLayer.java
+@@ -12,6 +_,11 @@
+    private static final IntSet f_76911_ = new IntOpenHashSet(new int[]{168, 169, 21, 22, 23, 149, 151});
+ 
+    public int m_6949_(Context p_76918_, int p_76919_, int p_76920_, int p_76921_, int p_76922_, int p_76923_) {
++      net.minecraftforge.registries.ForgeRegistry<net.minecraft.world.level.biome.Biome> registry = (net.minecraftforge.registries.ForgeRegistry<net.minecraft.world.level.biome.Biome>)net.minecraftforge.registries.ForgeRegistries.BIOMES;
++      java.util.Optional<net.minecraft.resources.ResourceKey<net.minecraft.world.level.biome.Biome>> shore = net.minecraftforge.common.BiomeManager.getShoreBiome(registry.getKey(p_76923_), registry.getKey(p_76919_), registry.getKey(p_76920_), registry.getKey(p_76921_), registry.getKey(p_76922_));
```
pupnewfster: Same question here as I raised with the `getEdgeBiome`.

patches/minecraft/net/minecraft/world/level/newbiome/layer/BiomeEdgeLayer.java.patch
maxanier: Not really, I just felt like it was cleaner to keep the integer ids away from the BiomeManager. But I can change that.

ChampionAsh5357:
Based on your parameters of choice within `IBiomeLayerModifier`, couldn't this be completely implemented as a data generated feature?

Cyborgmas:
This PR might actually be better suited to target 1.16, as it will very probably continue to be LTS after 1.18 drops, whereas 1.18 will be removing hills biomes and iirc edge biomes as well (can't quite remember the source though).
So could be worth checking that option first.

maxanier:
Oh, Cyborgmas is right, I checked 21w39a and it seems the biome selection system is completely changed.

Therefore, there isn't really a point of targeting 1.17. However, I am not sure if 1.16 PR are accepted if they are not merged to 1.17 first and it doesn't hurt to have this in 1.17 as well. I can do a 1.16 variant later on, if this PR is merged.

@ChampionAsh5357 
I am not really familiar with the data based creation of world (gen), therefore I do not know if that makes sense or how to do that.
Since 1.18 changes things again, I also don't think it is worth the effort to look into that, unless someone really wants that.

thedarkcolour:
Why use Optional?

dpeter99:
Hi!
This is a large PR, so with the internal deadling for 1.17 approaching, we've decided to close this PR.
It's not an outright rejection - this is a good change that we'd like to see in 1.18 or 1.16, but for either of those to be viable, the PR would need to be retargetted. This involves closing the PR anyway.

If you'd like to port this change to one of the aforementioned versions, let us know. Otherwise, one of our team will take care of it for you, with full credit to you.

alcatrazEscapee:
> this is a good change that we'd like to see in 1.18

That'll be a bit difficult, seeing that in 1.18 Mojang has completely removed every *single* class that this patches.

maxanier:
Alright, thank you for the update.
I have created a 1.16 version #8219

tmvkrpxl0:
So Where is 1.18 version of this PR?

PR #8140, opened 2021-10-04
Deprecate net.minecraftforge.common.util.Constants and inner classes to encourage usage of vanilla constants

PlatinPython:
Because in 1.17 constants are no longer stripped from the vanilla source code the Constant class provided by Forge is no longer necessary.
This PR deprecates said classes, replaces the existing initializations with references to the vanilla equivalents and replaces existing references in forge code/patches with the vanilla equivalents.

### DIFF
--- a/patches/minecraft/net/minecraft/world/effect/MobEffectInstance.java.patch
+++ b/patches/minecraft/net/minecraft/world/effect/MobEffectInstance.java.patch
@@ -72,9 +72,9 @@
 +      this.curativeItems = curativeItems;
 +   }
 +   private static MobEffectInstance readCurativeItems(MobEffectInstance effect, CompoundTag nbt) {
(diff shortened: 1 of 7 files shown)


marchermans:
If you depreciate something like this, please also add documentation with regards to which classes / fields / methods are supposed to be used as a replacement.
In particular here, it is a good idea that for every one of these there is a replacement available

PlatinPython:
I have added javadocs to the inner classes linking to the vanilla classes that contain the replacements.
But if wished for I can add javadocs to every field.

marchermans (approved):

Shadows-of-Fire:
Looks good, probably should be done.  Only concern is maybe the loss of those javadocs in the forge constants class that would be useful to retain.

sciwhiz12 (approved):

PR #8142, opened 2021-10-04
Little change in README

tmvkrpxl0:
We moved on to 1.17 and made 1.16 LTS, so I think we should change this


dpeter99:
Hi! This pull request seems to be a better fit as part of the current **omnibus PR**: https://github.com/MinecraftForge/MinecraftForge/pull/8144.
The PR description includes what an omnibus PR is, and how to submit a pull request to the omnibus PR.
Thank you for your contributions!

PR #8145, opened 2021-10-06
added toString function to Tag

apetschenka:
There are no Properties in Tag so i used the default implementations.


ChampionAsh5357:
This is a duplicate of #8112 trying to solve #7570. However, this does not provide any additional information relating to what the tag currently has. The information provided is what is already generated by default. Finally, this patch does not address the associated issue at all since it is in the incorrect class.

Please reread the issue and rethink your design. You can also look at the comments on the other PR for guidance.

PR #8148, opened 2021-10-10
[1.17.x] Use HttpClient for VersionChecker

PaintNinja:
This PR updates VersionChecker to use modern Java's `HttpClient` instead of `HttpUrlConnection` and implements GZip support, a timeout feature and performance improvements. All existing functionality is preserved.


fmlcore/src/main/java/net/minecraftforge/fml/VersionChecker.java
```diff
@@ -116,46 +119,47 @@ public void run()
...
+                            continue;
+                        }
+                        catch (NullPointerException e)
+                        {
+                            throw new IOException("Got a 3xx response code but Location header was null while trying to fetch " + url);
                         }
                     }
 
-                    return c.getInputStream();
+                    return response.body();
```
ChampionAsh5357: Shouldn't we shutdown the client executor thread and connections if it is no longer referenced?

fmlcore/src/main/java/net/minecraftforge/fml/VersionChecker.java
```diff
@@ -116,46 +119,47 @@ public void run()
...
-                            {
-                                huc.disconnect();
-                            }
+                            String newLocation = response.headers().map().get("Location").get(0);
+                            currentUrl = new URL(currentUrl, newLocation);
+                            continue;
+                        }
+                        catch (NullPointerException e)
+                        {
+                            throw new IOException("Got a 3xx response code but Location header was null while trying to fetch " + url);
```
diesieben07: This should use a null check, not a NPE catch.
Or really this should just be
`String newLocation = response.headers().firstValue("Location").orElseThrow(...)`

diesieben07 (changes requested):

fmlcore/src/main/java/net/minecraftforge/fml/VersionChecker.java
diesieben07: HttpClient doesn't provide a way to do so.

fmlcore/src/main/java/net/minecraftforge/fml/VersionChecker.java
PaintNinja: `firstValue("Location")` cannot be used as the returned response might not always have `Location` as the first header.

I've implemented an alternative approach that takes this into account but still uses `Optional`, thanks for your feedback.

diesieben07: Headers might be repeated. `firstValue` simply returns the first Location header, not the first header should it be the `Location`. `firstValue` uses code equivalent to what you have currently.

PaintNinja: My mistake, I misunderstood the JavaDoc. Tested it now and you're right, using `firstValue()` instead...

fmlcore/src/main/java/net/minecraftforge/fml/VersionChecker.java
PaintNinja: If I remember correctly, connections are automatically closed either once a response has been received, it times out or when the server has closed it.

The HttpClient object is reused for each mod's update check to improve performance and reduce memory usage.

PR #8149, opened 2021-10-10
[1.17.x] Add "forge:enchanting_materials" tag in place of Lapis in the Enchanting Table

SilverDavidMC:
This PR simply adds an item tag for materials that can substitute lapis in the enchanting table called `forge:enchanting_materials`.

The main reason for having a new tag specifically for enchanting table substitutes is that doing so allows for items to be used for enchanting while not being tagged as lapis in crafting recipes where they wouldn't be applicable.

Since the new tag incorporates `#forge:gems/lapis` within it, this produces no breaking changes with mods that used the lapis tag for enchanting material substitutions.

This PR also fixes an issue where shift-clicking a lapis substitute into the enchanting table didn't work.


patches/minecraft/net/minecraft/world/inventory/EnchantmentMenu.java.patch
```diff
@@ -76,3 +76,12 @@
                 }
  
                 for(int j1 = 0; j1 < 3; ++j1) {
+@@ -253,7 +_,7 @@
+             if (!this.m_38903_(itemstack1, 2, 38, true)) {
+                return ItemStack.f_41583_;
+             }
+-         } else if (itemstack1.m_150930_(Items.f_42534_)) {
++         } else if (itemstack1.m_150922_(net.minecraftforge.common.Tags.Items.ENCHANTING_MATERIALS)) {
```
ChampionAsh5357: You are using both `Tag#contains` and `ItemStack#is` here between two different checks. `ItemStack#is` defers to `Tag#contains`, but I would make these conform to one method of checking if the stack is in the tag.

ChampionAsh5357 (changes requested):
In general, this looks good from both FC -> FS and VC -> FS. Just one minor comment on programming choices from what I see.

patches/minecraft/net/minecraft/world/inventory/EnchantmentMenu.java.patch
SilverDavidMC: Good catch. Using `ItemStack#is` for both since that's what the original code did in both places anyway

ChampionAsh5357 (approved):

LexManos:
The concept of this PR isn't bad, its more arcuately defining what this is used for.
But the name is bad, 'materials' is vague, wouldn't it be better as 'fuel' or 'catalyst' something denoting that it'll be consumed. 
Also, have you tested shift clicking while being connected to a vanilla server? This seems like it would break that.

SilverDavidMC:
I renamed it to "enchanting_fuels" since that seems like the simplest option.

As for testing, I tried the three things listed below. All of them produced the same results where the item was ultimately placed in the correct slot. The only issue that occurs is a brief desync when shift-clicking where an item attempts to go into the wrong slot on the client, but is then snapped back to where it belongs since the server ultimately takes priority.

- FC -> VS 
- VC -> FS 
- VC -> FS + With a datapack that replaces the tag with one that only includes an iron ingot

SilverDavidMC:
I was able to fix the desync issue when a forge client connects to a vanilla server. If the tag's `isDefaulted` returns true, it uses the vanilla behavior of only allowing lapis.

PR #8150, opened 2021-10-10
[1.16.x] Add ForgeSpawnEggItem to lazily handle EntityTypes

XFactHD:
This is the LTS backport of #8044.

This PR proposes a solution to the problem of SpawnEggItems requiring you to either pass in an EntityType that was initialized too early and therefore breaks registry contracts or pass in null and pollute the map with a useless null key which indirectly leads to problems later on, for example in the `ItemColors` registration where only the last modded egg registered this way would get tinting.

### Proposed Solution

The proposed solution is to store all ForgeSpawnEggItems added by mods in a list and transfer them over to the EntityType->SpawnEggItem map after EntityTypes have been registered. The original `SpawnEggItem` is patched to deprecate the constructor and guard against adding a null key to the `BY_ID` map.
The `BY_ID` map in question is used in four places:

1. **`IForgeEntity#getPickedResult(HitResult target)` via `SpawnEggItem.byId(EntityType type)`**
The call to `SpawnEggItem.byId(EntityType type)` is replaced with a call to `ForgeSpawnEggItem#fromEntityType(EntityType type)` which checks both `SpawnEggItem.byId(EntityType type)` and the `ForgeSpawnEggItem#TYPE_MAP`.

2. **`ItemColors.createDefault()` via `SpawnEggItem.eggs()`**
Even though this happens before EntityTypes are registered, it doesn't create any problems related to them because the EntityType is not used there. An event handler on `HIGHEST` priority registers the color handlers for mod eggs. A mod can easily override the `ItemColors` in `ColorHandlerEvent.Item` with standard priority if they wish to do so.
    
3. **`BlockModelProvider#run()` via `SpawnEggItem.eggs()`**
Doesn't matter to us because modders should not be using the vanilla `BlockModelGenerators` in datagen

4. **`IDispenseItemBehaviour.bootStrap()` via `SpawnEggItems.eggs()`**
This can be ignored for this purpose because I have found in my testing with breakpoints that the call to DispenseItemBehaviour.bootStrap() happens before `RegistryEvent<Item>` has fired which means that mod eggs are not included anyway. To make sure mod eggs get `DispenseItemBehaviour`s, they are registered in `FMLCommonSetup` at the same time the mod eggs are transferred to the type->egg map. If the modder wishes they can provide their own `DispenseItemBehaviour` or none via `ForgeSpawnEggItem#createDispenseBehavior()`.

Not adding any more ways to deviate from the default behaviour of the vanilla `SpawnEggItem` is very much intentional.
If you want your spawn eggs to behave differently to the vanilla spawn eggs, then you should IMO implement them yourself.

### Test

To test this I have included a test mod that adds a test entity and a spawn egg for that entity. In my testing spawning by hand, the coloring of the item, the dispense behaviour and "picking" the entity to get the egg for it worked flawlessly.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
+++ b/src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
@@ -36,7 +36,10 @@ import net.minecraftforge.fml.common.Mod;
 import net.minecraftforge.fml.event.lifecycle.FMLCommonSetupEvent;
 
 import javax.annotation.Nullable;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.IdentityHashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.function.Supplier;
 
 public class ForgeSpawnEggItem extends SpawnEggItem
@@ -82,7 +85,16 @@ public class ForgeSpawnEggItem extends SpawnEggItem
             Direction face = source.getBlockState().getValue(DispenserBlock.FACING);
             EntityType<?> type = ((SpawnEggItem) stack.getItem()).getType(stack.getTag());
 
-            type.spawn(source.getLevel(), stack, null, source.getPos().relative(face), SpawnReason.DISPENSER, face != Direction.UP, false);
+            // FORGE: fix potential crash
+            try
+            {
+                type.spawn(source.getLevel(), stack, null, source.getPos().relative(face), SpawnReason.DISPENSER, face != Direction.UP, false);
+            }
+            catch (Exception exception)
+            {
+                DefaultDispenseItemBehavior.LOGGER.error("Error while dispensing spawn egg from dispenser at {}", source.getPos(), exception);
+                return ItemStack.EMPTY;
+            }
 
             stack.shrink(1);
             return stack;
--- a/src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
+++ b/src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
@@ -1,0 +1,125 @@
+/*
+ * Minecraft Forge
+ * Copyright (c) 2016-2021.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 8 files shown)


src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Nullable
+    protected DefaultDispenseItemBehavior createDispenseBehavior()
+    {
+        return DEFAULT_DISPENSE_BEHAVIOR;
+    }
+
+    @Nullable
+    @OnlyIn(Dist.CLIENT)
```
pupnewfster: Why does this have an only in annotation?

XFactHD: Because I am an idiot and forgot to remove it after your response on Discord.

ShroomAgent27 (approved):

src/main/java/net/minecraftforge/common/ForgeSpawnEggItem.java
```diff
@@ -0,0 +1,122 @@
...
+
+
+    private static final DefaultDispenseItemBehavior DEFAULT_DISPENSE_BEHAVIOR = new DefaultDispenseItemBehavior()
+    {
+        protected ItemStack execute(IBlockSource source, ItemStack stack)
+        {
+            Direction face = source.getBlockState().getValue(DispenserBlock.FACING);
+            EntityType<?> type = ((SpawnEggItem) stack.getItem()).getType(stack.getTag());
+
+            type.spawn(source.getLevel(), stack, null, source.getPos().relative(face), SpawnReason.DISPENSER, face != Direction.UP, false);
```
gigaherz: the 1.17 version has a try/catch here, is it relevant in the backport?

XFactHD: The dispense behaviors are a copy of the vanilla behaviors used by the normal spawn egg and Mojang added a try/catch there in 1.17.
If you want, I can add the try/catch here anyway.

PR #8151, opened 2021-10-11
[1.17.x] Add three new AT entries to make adding structures easier for mods

thedarkcolour:
This PR adds three access transformers to allow mods to add structures properly without needing to add an access transformer. Currently, in order for a mod to add a structure, it must use an access transformer, usually with these three AT entries that get copied throughout a lot of the mods that add overworld structures:

This access transformer is needed to make the ImmutableMap non-final to add the structure's default spawning settings:
```cfg
public-f net.minecraft.world.level.levelgen.StructureSettings f_64580_ # DEFAULTS
```
This access transformer is needed to make the ImmutableList non-final to add ground around the base of the structure in Beardifier
```cfg
public-f net.minecraft.world.level.levelgen.feature.StructureFeature f_67031_ # NOISE_AFFECTING_FEATURES
```
This access transformer is used by some mods ([here](https://github.com/TelepathicGrunt/StructureTutorialMod/blob/b550b51f3046846c2dc396dde3a23183cebdea9c/src/main/java/com/telepathicgrunt/structuretutorial/STStructures.java#L128), [here](https://github.com/ValhelsiaTeam/Valhelsia-Structures/blob/cb3f4ab4741ad51ee8045587609220439602c2d2/src/main/java/com/stal111/valhelsia_structures/core/init/ModStructures.java#L68), [here](https://github.com/xyroc/DungeonCrawl/blob/72915495bd563d419ec7e70e799f9c80fa879799/src/main/java/xiroc/dungeoncrawl/init/ModStructures.java#L96)) to re-add their structure's placement settings to `BuiltinRegistries.NOISE_GENERATOR_SETTINGS`:
```cfg
public-f net.minecraft.world.level.levelgen.StructureSettings f_64582_ # structureConfig
```

### DIFF
--- a/src/main/resources/META-INF/accesstransformer.cfg
+++ b/src/main/resources/META-INF/accesstransformer.cfg
@@ -418,3 +418,6 @@ protected net.minecraft.world.level.portal.PortalForcer f_77648_ # level
 public net.minecraft.world.level.storage.LevelResource <init>(Ljava/lang/String;)V # constructor
 private-f net.minecraft.world.level.storage.loot.LootPool f_79028_ # rolls
 private-f net.minecraft.world.level.storage.loot.LootPool f_79029_ # bonusRolls
+public-f net.minecraft.world.level.levelgen.feature.StructureFeature f_67031_ # NOISE_AFFECTING_FEATURES
+public-f net.minecraft.world.level.levelgen.StructureSettings f_64582_ # structureConfig
+public-f net.minecraft.world.level.levelgen.StructureSettings f_64580_ # DEFAULTS


TheCurle:
Can you more thoroughly justify these changes for us? "A lot of mods do this already" isn't justification, before we get that far..

We tend to dislike un-finalizing things that Mojang have made it obvious should not be changed.

With most of these cases there is a known workaround, such as for the structure placement maps.

thedarkcolour:
I think it's a beneficial change because currently any mod that adds a structure needs to also have an access transformer, which I don't think should be necessary for something simple like a structure to spawn in the world. `NOISE_AFFECTING_FEATURES` is non-final because I didn't think a patch making it an arraylist would be a great idea, because existing mods could replace it with an ImmutableList again. I think the first two AT entries are necessary but I've gotten structures working without the third one. What is the workaround for the structure placement maps?

ChampionAsh5357:
I disagree with the ATs used here. From what I can tell, this just allows people to replace the associated maps with whatever the user wants for any generation settings which could cause major mod compatibility issues.

Overriding `DEFAULTS` doesn't really affect anything. It gets passed into `StructureSettings#structureConfig` which is always mutable in vanilla and as such could be modified easily. For a better solution, as these are generated similar to biomes, I would say that something similar to `BiomeLoadingEvent` would probably be much more preferable.

 As for `NOISE_AFFECTING_FEATURES`, this would need to be replaced in some manner, but a better way would be to redirect the call to a concatenation of vanilla and some forge extension which allows you to register your features to the beardifier.

thedarkcolour:
Would the `NOISE_AFFECTING_FEATURES` be good as a new event where mods can add structures, similar to `ColorHandlerEvent`?

ChampionAsh5357:
I would say it would be better as some API method to which can be called within `FMLCommonSetupEvent` since that's where we usually end up dealing with deferred vanilla stuff. Adding that as an event I think would be complicated when it gets to the dedicated server since its not the staging environment.

thedarkcolour:
Looks like the only one left is the NOISE_AFFECTING_FEATURES

thedarkcolour:
Turns out `structureConfig` is not always mutable in vanilla. It's immutable for superflat worlds

ChampionAsh5357:
> Turns out `structureConfig` is not always mutable in vanilla. It's immutable for superflat worlds

Where do you see this? Also, in any case, the map could be made mutable during the associated proposed event alternative.

thedarkcolour:
Looks like it happens in the StructureSettings.CODEC because Codec.simpleMap always returns a SingletonImmutableBiMap
image

ChampionAsh5357:
@thedarkcolour will you be implementing the above alternatives within this PR? If not, it would probably be better to close this and open as an issue for someone else to take over at some point.

thedarkcolour:
I was hoping to try making it but I'm not really sure if I'll follow through
If I end up making it though, I'll open a separate PR

PR #8153, opened 2021-10-13
Fix the crash caused by #8038

diesieben07:
A broken patch I introduced crashes the game when hovering in the recipe book.
Sorry.


marchermans (approved):
Tested it with my crash report.
This fixes it.

Shipit

PR #8157, opened 2021-10-16
Add `BiomeDictionary#hasType(String)`

alcatrazEscapee:
Add a way of checking if a type exists by name. This is useful in terms of error reporting when provided by biome dictionary names from untrusted sources (e.g. a config or codec). As an example, currently all that's possible is the following code:

```java
Codec<BiomeDictionary.Type> CODEC = Codec.STRING.xmap(
    BiomeDictionary.Type::getType,
    BiomeDictionary.Type::getName
);
```

Whereas, adding `hasType` allows us to report errors and typos in the user's configuration:

```java
Codec<BiomeDictionary.Type> CODEC = Codec.STRING.comapFlatMap(
    t -> BiomeDictionary.Type.hasType(t) ?
        DataResult.success(BiomeDictionary.Type.getType(t)) :
        DataResult.error("No biome dictionary type with name: " + t),
    BiomeDictionary.Type::getName
);
```

### DIFF
--- a/src/main/java/net/minecraftforge/common/BiomeDictionary.java
+++ b/src/main/java/net/minecraftforge/common/BiomeDictionary.java
@@ -157,6 +157,19 @@ public class BiomeDictionary
         }
 
         /**
+         * Checks if a type instance exists for a given name. Does not have any side effects if a type does not already exist.
+         * This can be used for checking if a user-defined type is valid, for example, in a codec which accepts biome dictionary names.
+         * @param name The name.
+         * @return {@code true} if a type exists with this name.
+         *
+         * @see #getType(String, Type...) for type naming conventions.
+         */
+        public static boolean hasType(String name)
+        {
+            return byName.containsKey(name.toUpperCase());
+        }
+
+        /**
          * @return An unmodifiable collection of all current biome types.
          */
         public static Collection<Type> getAll()
--- a/src/main/java/net/minecraftforge/common/BiomeDictionary.java
+++ b/src/main/java/net/minecraftforge/common/BiomeDictionary.java
@@ -162,7 +162,7 @@ public class BiomeDictionary
          * @param name The name.
          * @return {@code true} if a type exists with this name.
          *
-         * @see #getType(String, Type...) for type naming conventions.
+         * @see #getType(String, Type...) #getType for type naming conventions.
          */
         public static boolean hasType(String name)
         {
--- a/src/test/java/net/minecraftforge/debug/block/BlockEntityOnLoadTest.java
+++ b/src/test/java/net/minecraftforge/debug/block/BlockEntityOnLoadTest.java
@@ -1,3 +1,22 @@
+/*
+ * Minecraft Forge
+ * Copyright (c) 2016-2021.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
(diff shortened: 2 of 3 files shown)


src/main/java/net/minecraftforge/common/BiomeDictionary.java
```diff
@@ -156,6 +156,19 @@ public static Type getType(String name, Type... subTypes)
             return t;
         }
 
+        /**
+         * Checks if a type instance exists for a given name. Does not have any side effects if a type does not already exist.
+         * This can be used for checking if a user-defined type is valid, for example, in a codec which accepts biome dictionary names.
+         * @param name The name.
+         * @return {@code true} if a type exists with this name.
+         *
+         * @see #getType(String, Type...) for type naming conventions.
```
Harleyoc1: I suggest adding `#getType` after the `#getType(String, Type...)` link or the Javadoc shows up as `See Also:
for type naming conventions.` when rendered.

Harleyoc1 (approved):

ChampionAsh5357:
I understand the use of this method when it comes from a config and codec. Though why not also just supply a codec for the `BiomeDictionary$Type` as well? It seems like a common enough use case and it would make it such that modders do not need to write the same codec multiple times.

alcatrazEscapee:
Mostly in the interest of making a PR small and targeted, and broadly useful. Would you like me to add a codec as part of this PR? I could also add a `getTypeIfPresent` or any other number of "probably also useful" things but I went with the simplest viable solution here.

src/main/java/net/minecraftforge/common/BiomeDictionary.java
```diff
@@ -156,6 +156,19 @@ public static Type getType(String name, Type... subTypes)
...
+         * Checks if a type instance exists for a given name. Does not have any side effects if a type does not already exist.
+         * This can be used for checking if a user-defined type is valid, for example, in a codec which accepts biome dictionary names.
+         * @param name The name.
+         * @return {@code true} if a type exists with this name.
+         *
+         * @see #getType(String, Type...) #getType for type naming conventions.
+         */
+        public static boolean hasType(String name)
+        {
+            return byName.containsKey(name.toUpperCase());
```
ChampionAsh5357, line 169: Should use `Locale#ROOT` in upper case to avoid any casing issues in other languages.

ChampionAsh5357 (changes requested):
As part of this PR, probably not since it seems out of scope. But I would definitely say in another a common codec would probably be nice, though it would require a bit more extensive testing on our part.

src/main/java/net/minecraftforge/common/BiomeDictionary.java
alcatrazEscapee: If I do this I'd also need to change the existing `BiomeDictionary.Type#getType` as otherwise they'd have inconsistent results in said languages with casing issues (since I mirrored that functionality here). Which, again, in theme, was something I decided not to change in the initial draft, would you like me to change that one as well?

ChampionAsh5357: Hmm...after taking a closer look at the code, it seems to be unnecessary. Since `BiomeDictionary$Type#getType` is always used to generate the data, the comparisons will be the same regardless. For this case, I have no issues otherwise.

ChampionAsh5357 (approved):

PR #8158, opened 2021-10-17
Fix render pipeline lighting issue with small cubes

sekwah41:
This will also allow other mods which need to calculate small normalised vectors to be able to use the default classes without needing to write their own, while not affecting any side effects created by how the base game uses this function.

image

It fixes the bug shown above and is useful for other mods where any problems with the resulting vector are probably not as bad as it not being calculated in the first place.

After changes
image

### DIFF
--- a/patches/minecraft/com/mojang/math/Vector3f.java.patch
+++ b/patches/minecraft/com/mojang/math/Vector3f.java.patch
@@ -1,16 +1,11 @@
 --- a/com/mojang/math/Vector3f.java
 +++ b/com/mojang/math/Vector3f.java
-@@ -137,8 +_,12 @@
-    }
+@@ -138,7 +_,7 @@
  
     public boolean m_122278_() {
-+      return normalize(false);
-+   }
-+
-+   public boolean normalize(boolean force) {
        float f = this.f_122228_ * this.f_122228_ + this.f_122229_ * this.f_122229_ + this.f_122230_ * this.f_122230_;
 -      if ((double)f < 1.0E-5D) {
-+      if (!force && (double)f < 1.0E-5D) {
++      if ((double)f == 0) {
           return false;
        } else {
           float f1 = Mth.m_14195_(f);
--- a/patches/minecraft/com/mojang/math/Vector3f.java.patch
+++ b/patches/minecraft/com/mojang/math/Vector3f.java.patch
@@ -1,5 +1,19 @@
 --- a/com/mojang/math/Vector3f.java
 +++ b/com/mojang/math/Vector3f.java
+@@ -137,8 +_,12 @@
+    }
+ 
+    public boolean m_122278_() {
++      return normalize(false);
++   }
++
++   public boolean normalize(boolean force) {
+       float f = this.f_122228_ * this.f_122228_ + this.f_122229_ * this.f_122229_ + this.f_122230_ * this.f_122230_;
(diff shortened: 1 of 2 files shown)


marchermans (approved):

marchermans:
So I know that this fixes it, since this is how I fixed this in C&B.
But would this not also require a Mojang Minecraft Issue ticket?

sekwah41:
I'm don't know if it creates a problem in vanilla. So far I've only found it only causes a problem in the experimental forge render pipeline.

I'm not sure how quick or if mojang would fix it unless a vanilla problem is found.

gigaherz:
If the bug is in vanilla code we request that the issues are reported to mojang. Even if it's unlikely they will fix them, we want them to be reported.

We can then mark forge's fix as `// Forge: fix MC-xxxxx`

sekwah41:
> If the bug is in vanilla code we request that the issues are reported to mojang. Even if it's unlikely they will fix them, we want them to be reported.
> 
> We can then mark forge's fix as `// Forge: fix MC-xxxxx`

I'll report it when I'm back later then. Though I'll take a quick look at where this function is used in vanilla to see if I can find anything it might break.

sekwah41:
I've created the issue https://bugs.mojang.com/browse/MC-239212

I will also change the code to just avoid it when the length is 0 rather than avoiding it entirely, seeing as normalising doesn't work in that case anyway.

gigaherz:
I'm wondering if it would be worth doing instead, `if ((double)f < Float.MIN_NORMAL)` which would mean the value is either denormalized or zero.

pupnewfster:
> I'm wondering if it would be worth doing instead, `if ((double)f < Float.MIN_NORMAL)` which would mean the value is either denormalized or zero.

At that point shouldn't it be `if (f < Float.MIN_NORMAL)` so as to not potentially have issues due to float to double floating point precision? At least given the fact it seems to use it as a float still lower down so wouldn't have that precision issue show up there instead.

patches/minecraft/com/mojang/math/Vector3f.java.patch
```diff
@@ -1,16 +1,11 @@
...
  
     public boolean m_122278_() {
-+      return normalize(false);
-+   }
-+
-+   public boolean normalize(boolean force) {
        float f = this.f_122228_ * this.f_122228_ + this.f_122229_ * this.f_122229_ + this.f_122230_ * this.f_122230_;
 -      if ((double)f < 1.0E-5D) {
-+      if (!force && (double)f < 1.0E-5D) {
++      if ((double)f == 0) {
```
pupnewfster: At the end of this line you should include a comment like: `//Forge: Fix MC-239212`

sekwah41:
> I'm wondering if it would be worth doing instead, `if ((double)f < Float.MIN_NORMAL)` which would mean the value is either denormalized or zero.

Forgot that existed, that would probably be the best

sekwah41:
After taking a deeper look into the code, the only time Mojang ever calls it, they actually perform another function on it which doesn't seem to matter if it's normalised or not.

Other than that they normalise some vectors which are pre-defined.

I hope they fix it but with that in mind, altering this and making it part of forge should not cause any problems in mc and should only benefit any mods using these classes rather than implementing their own.

ChampionAsh5357 (approved):

Issue #7969, opened 2021-07-31
RFC - Remove ForgeRegistryEntry and ForgeRegistries being Type bound


Cyborgmas:
This would allow to make any registry a forge registry.
- Right now functional interfaces can't be forge registries
- Can't have more than one registry per class, while mojang for instance has 2 Codec registries (`BIOME_SOURCE` and `CHUNK_GENERATOR`).

This would however mean registry objects won't be aware about being in a registry, so `#setRegistryName` and `#getRegistryName` would be no more.

However, there would still be a need to handle delegates so a `DelegatedRegistryEntry` might be needed where possible, but would not be a requirment.

Is this desired by forge/modders ? Have I missed any constraints about this change?

LexManos:
Yes, sometime in the future we need to re-write the Forge registry system to not be IForgeRegistryEntry based. This will be a fundemental shift in the way that registry entries work tho. They will no longer know what registry they belong to. All related functions, ObjectHolders, Delegates, Register events, etc... would need to be redesigned.
Its a really large project that has no real need right now. So it hasn't been worked on.

Cyborgmas:
Worked on this a while back, but did not have time to continue. 

If anyone wants to continue working on this, feel free to use [what I've done](https://github.com/Cyborgmas/MinecraftForge/tree/rewrite-registries-118) and to ping me on discord for any questions about my work.

The main thing that is missing is a way to handle re-registration, i.e. data-driven objects, like Biomes. 
Also needs a lot of testing eventually, porting worlds could cause issues.

userofbricks:
It would be really nice if something like the deferredRegister system was kept. It is my primary way of registering things. However that said it could be really nice if it worked off of the vanilla classes instead of forge ones. This my already be the case as I’m not at my computer right now and it has been a while since I had to start afresh with a mod.

SizableShrimp:
As discussed on #8538, this change will be included in Forge 1.19.
To see (most of) the upcoming changes, please view #8538 (based on 1.18.2).
IForgeRegistryEntry/ForgeRegistryEntry will no longer exist, and IForgeRegistry/ForgeRegistry no longer has a type bound.

sciwhiz12:
The registry rework has been published in 1.19, which means this RFC is complete. Goodbye `IForgeRegistryEntry`, you had a good run.

Issue #7574, opened 2020-12-30
[1.6.4-35.1.4] ObjectHolder annotation missing from EntityTypes.class


MonsterPoops:
There isn't an @ObjectHolder annotation over the EntityTypes.class and it seems like there should be, like there are over Items.class, Blocks.class, etc

I found this out trying to override the vanilla Skeleton with my own Skeleton from a private mod. Because the "EntityTypes.SKELETON" field doesn't actually get replaced, parts of the game that use the field directly instead of using the registry like the "skeleton horse trap" become broken.

ChampionAsh5357:
There are quite a number of vanilla registries represented within Forge that do not have this label. Specifically, fluids, potions, entity types, tile entity types, container types, painting types, recipe serializers, attributes, stat types, villager professions, point of interest types, memory module types, sensor types, schedules, activities, world carvers, surface builders, features, placements, chunk statuses, structures, block state provider types, block placer types, foliage placer types, and tree decorator types. Biomes are a special case as they are now data driven.

PR #8159, opened 2021-10-17
[1.17.x] Made several vanilla registries forge registries, added missing ObjectHolder annotations 

rehwinkel:
Adds TrunkPlacerTypes as a forge registry.


pupnewfster:
A couple things:
1. I believe you also need create the registry in [GameData](https://github.com/MinecraftForge/MinecraftForge/blob/1.17.x/src/main/java/net/minecraftforge/registries/GameData.java#L129)
2. It might be wise to broaden this PR slightly to capture other new vanilla registries that can easily be captured as well. I know there is at least GameEvent, and I wouldn't be surprised if there were others as well. (Special note on game events the registry should also be added [here](https://github.com/MinecraftForge/MinecraftForge/blob/1.17.x/src/main/java/net/minecraftforge/common/data/ForgeRegistryTagsProvider.java#L39-L46))

thedarkcolour:
The other ones that I see are `FEATURE_SIZE_TYPES`, `BIOME_SOURCE`, `CHUNK_GENERATOR`, `STRUCTURE_PROCESSOR`, and `STRUCTURE_POOL_ELEMENT` from Registry

alcatrazEscapee:
Biome source and chunk generators cannot be made into a Forge registry as their elements have the same generic type, and notably are of a `Codec<?>`, which would be a huge breaking change, along with a big pile o patches if you were to change those to a forge registry type. See #7969

ChampionAsh5357:
Structure processors and structure pool elements are also a no go since those are also functional interfaces around a codec similar to how `RecipeType`s are constructed anonymously.

rehwinkel:
I added the missing lines to make the registry actually work like @pupnewfster mentioned in #1. Adding the registry for FeatureSizeType would make sense here too.

ChampionAsh5357:
There are many more types as pup said including `GameEvent`, any of the loot types, and `PositionSourceType`. Also, it would probably be good to AT any constructors to be public as the two you have implemented are currently inaccessible to modders due to their private constructors.

ChampionAsh5357:
Since this is already a registry addition, wondering if it would also be appropriate to deal with #7574 as well since all those would need is `ObjectHolder` annotations on most vanilla registry object classes (excluding `Biome` and `RecipeSerializer` currently).

rehwinkel:
@ChampionAsh5357 How about interfaces like RecipeType ~~or RecipeSerializer~~?

ChampionAsh5357:
So, for registries you are still missing the loot types and `PositionSourceType`. I'm not sure whether the builtin registries should be forge registries currently so I'm omitting them from this explanation for now.

As for the annotation, you seemed to have put them on every class regardless of if it was a forge wrapped registry or not. So you need to remove those that aren't.

> How about interfaces like RecipeType ~~or RecipeSerializer~~?

I've already mentioned not to add the annotation to `RecipeSerializer` (or any interfaces for that matter). What the definalizer does violates the contract of fields in interfaces and throws an exception (see [JVMS 4.5](https://docs.oracle.com/javase/specs/jvms/se16/html/jvms-4.html#jvms-4.5)). As for `RecipeType`, it's not a forge wrapped registry.

src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -173,11 +176,16 @@ public static void init()
...
         makeRegistry(FOLIAGE_PLACER_TYPES, c(FoliagePlacerType.class)).disableSaving().disableSync().create();
+        makeRegistry(FEATURE_SIZE_TYPES, c(FeatureSizeType.class)).disableSaving().disableSync().create();
+        makeRegistry(TRUNK_PLACER_TYPES, c(TrunkPlacerType.class)).disableSaving().disableSync().create();
         makeRegistry(TREE_DECORATOR_TYPES, c(TreeDecoratorType.class)).disableSaving().disableSync().create();
 
         // Dynamic Worldgen
         makeRegistry(BIOMES, Biome.class).disableSync().create();
 
+        // Other
+        makeRegistry(GAME_EVENT, c(GameEvent.class)).disableSaving().disableSync().create();
```
pupnewfster: I am not sure if these are the correct flags for Game Events, in terms of disable saving and sync... Given the fact vanilla has game event tags they definitely should sync and honestly they probably should save as well? Also Can you address this TODO: https://github.com/MinecraftForge/MinecraftForge/blob/1.17.x/src/main/java/net/minecraftforge/common/data/ForgeRegistryTagsProvider.java#L39-L46 and add the reference to the forge game event registry and tag folder there?

rehwinkel: I'll admit I wasn't at all sure what to put there, but I figured since the Registry would remain static after the game has started, syncing wouldn't be required. Am I right in thinking that "sync" in this context refers to the server giving the client(s) the contents of the registry? And "saving" meaning the registry is saved on a per-world basis?

rehwinkel:
I've fixed the ObjectHolder annotations to only apply to classes that hold forge registry objects.

> So, for registries you are still missing the loot types and `PositionSourceType`.

What is the standard/preferred way for making classes like `LootPoolEntryType` a `ForgeRegistryEntry`? Do I just implement `IForgeRegistryEntry`? Do I do that in the base class? I'd naively just implement it in the `LootPoolEntryType` type directly, but if there is a better way to do that, let me know.

ChampionAsh5357:
> What is the standard/preferred way for making classes like `LootPoolEntryType` a `ForgeRegistryEntry`?

See `RecipeSerializer` and its subtypes. That tells you exactly how to do it.

rehwinkel:
`RecipeSerializer` is an interface, `SerializerType` is not. I'll just implement the `IForgeRegistryEntry` interface manually.

ChampionAsh5357:
Apologies, misread because of the quote. Just implement ForgeRegistryEntry on the class type. Do not implement anything on SerializerType.

src/main/java/net/minecraftforge/registries/GameData.java
pupnewfster: I believe that is correct, at least the syncing making sure the client and server have the same entries (which given vanilla has tags for game events is important).

rehwinkel: Yeah! In my most recent commits I enabled saving and sync again.

patches/minecraft/net/minecraft/world/level/storage/loot/entries/LootPoolEntryType.java.patch
```diff
@@ -0,0 +1,37 @@
...
++   public final LootPoolEntryType setRegistryName(String modID, String name){ return setRegistryName(modID + ":" + name); }
++
++   public final net.minecraft.resources.ResourceLocation getRegistryName()
++   {
++      return registryName;
++   }
++
++   public final Class<LootPoolEntryType> getRegistryType() { return LootPoolEntryType.class; }
++
++   net.minecraft.resources.ResourceLocation checkRegistryName(String name) { return net.minecraftforge.registries.GameData.checkPrefix(name, true); }
```
pupnewfster: I am not sure if this (and the other loot types can be properly taken over by forge, I haven't looked at it yet), but as a lot of them extend `SerializerType` and are adding these massive copy paste blocks that `ForgeRegistryEntry` is used to reduce the stuff that needs patching... I feel like creating a `ForgeSerializerTypeRegistryEntry` intermediary class would be less messy and certainly reduce the size of a bunch of these patches

src/main/java/net/minecraftforge/registries/ForgeSerializerTypeRegistryEntry.java
```diff
@@ -0,0 +1,57 @@
+package net.minecraftforge.registries;
```
ChampionAsh5357, line 20: Missing license.

src/main/java/net/minecraftforge/registries/ForgeSerializerTypeRegistryEntry.java
```diff
@@ -0,0 +1,57 @@
...
+
+import com.google.common.reflect.TypeToken;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.level.storage.loot.Serializer;
+import net.minecraft.world.level.storage.loot.SerializerType;
+
+import javax.annotation.Nullable;
+
+public class ForgeSerializerTypeRegistryEntry<V extends IForgeRegistryEntry<V>, T> extends SerializerType<T> implements IForgeRegistryEntry<V>
```
ChampionAsh5357, line 36: Needs javadoc on what this is for.

src/main/java/net/minecraftforge/registries/ForgeSerializerTypeRegistryEntry.java
```diff
@@ -0,0 +1,57 @@
...
+{
+
+    public ForgeSerializerTypeRegistryEntry(Serializer<? extends T> p_79330_)
```
ChampionAsh5357: Name the parameter.

ChampionAsh5357 (changes requested):
I think just a few formatting and javadoc concerns left.

ChampionAsh5357 (approved):
Update the PR name so Lex and others know what this is for without having to guess. For small context in case you don't before anyone can look at it, this PR just adds in the missing forge wrapped registries and adds object holders to those that already exist.

sciwhiz12:
@rehwinkel: Please retarget this PR to 1.18 and check for any additional vanilla registries which may need to have new Forge registry counterparts.

sciwhiz12:
Hello! @rehwinkel, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable. If we do not hear back from you after some time, this PR will be closed for inactivity.

PR #8160, opened 2021-10-18
[1.17.x] Fix blocks with flammable material catching fire from lava when they are marked as non-flammable

XFactHD:
Currently only the `FireBlock` asks the block itself wether it is flammable through the use of Forge extensions. This leads to issues like this: https://github.com/XFactHD/FramedBlocks/issues/31 where a block that uses `Material.WOOD` but returns `false` from `IForgeBlock#isFlammable()` catches fire from the lava but is never destroyed by the `FireBlock` because it thinks the block is not flammable.

This PR patches the `LavaFluid` to use the Forge extensions for flammability instead of checking the material for flammability. This way the behaviour of the `LavaFluid` is equivalent to the `FireBlock` which is already using these Forge extensions.

I can add a test mod for the flammability hooks if that is desired.


patches/minecraft/net/minecraft/world/level/material/LavaFluid.java.patch
```diff
@@ -9,15 +9,37 @@
...
 +                  p_76239_.m_46597_(blockpos1.m_7494_(), net.minecraftforge.event.ForgeEventFactory.fireFluidPlaceBlockEvent(p_76239_, blockpos1.m_7494_(), p_76240_, Blocks.f_50083_.m_49966_()));
                 }
              }
           }
+@@ -97,7 +_,7 @@
+ 
+    private boolean m_76227_(LevelReader p_76228_, BlockPos p_76229_) {
+       for(Direction direction : Direction.values()) {
+-         if (this.m_76245_(p_76228_, p_76229_.m_142300_(direction))) {
++         if (this.isFlammable(p_76228_, p_76229_.m_142300_(direction), direction)) {
```
ChampionAsh5357: The third parameter needs to get the opposite direction. The flame is trying to spawn at `p_76229_`. The positions around it are being checked which means the direction the flame will spawn will be the opposite of the direction moved.

ChampionAsh5357 (changes requested):
I'm not sure about the method replacement since I don't know when its valid, but otherwise just one thing.

XFactHD:
> I'm not sure about the method replacement since I don't know when its valid

The `FireBlock` [patch](https://github.com/MinecraftForge/MinecraftForge/blob/1.17.x/patches/minecraft/net/minecraft/world/level/block/FireBlock.java.patch#L70-L73) does the exact same thing with the additional parameter and the related EXC entry. I would assume it's valid in this case due to the method being private.

patches/minecraft/net/minecraft/world/level/material/LavaFluid.java.patch
XFactHD: Good catch

Shadows-of-Fire:
Are there any other places checking flammability incorrectly that could be rolled into this PR or is this the only instance of the incorrect check?

ChampionAsh5357:
> The `FireBlock` [patch](https://github.com/MinecraftForge/MinecraftForge/blob/1.17.x/patches/minecraft/net/minecraft/world/level/block/FireBlock.java.patch#L70-L73) does the exact same thing with the additional parameter and the related EXC entry. I would assume it's valid in this case due to the method being private.

After asking, it seems as though there is no real benefit to removing a private method, so I would just deprecate and redirect instead.

It was originally done in this fashion before patches were removed and regenned in [this commit](https://github.com/MinecraftForge/MinecraftForge/commit/7c0d94ce9c7481fe4a122aa5731aa3becbaf317f#diff-7c2d5454fc6a358a6a578202e552e460d1e02cfd143bff15bde425aa91110e83) to which the [1.13.2 update](https://github.com/MinecraftForge/MinecraftForge/commit/ca72710937a88ffbcccb9f748ae1890ea9887514#diff-7c2d5454fc6a358a6a578202e552e460d1e02cfd143bff15bde425aa91110e83) seemed to fix with the `exc` system. The fire patch also only replaced a part of the body + signature while your patch replaces the entire method itself. The only thing `exc` seems to do is specify what each parameter name is so mapping sets can properly reference them. I don't think it should be used as an ad hoc system whenever something is being completely removed and replaced with another.

XFactHD:
> Are there any other places checking flammability incorrectly that could be rolled into this PR or is this the only instance of the incorrect check?

Definitely none that directly check `Material#isFlammable()`, the method is unused after this change. 

> After asking, it seems as though there is no real benefit to removing a private method, so I would just deprecate and redirect instead.

If it is preferable to copy the method for this change, then I can do that. Should I add a proper deprecation notice that refers to the new method even though these are both private methods?

ChampionAsh5357:
Out of convention, I would

ChampionAsh5357 (approved):

PR #8161, opened 2021-10-18
[1.16.x] Fix blocks with flammable material catching fire from lava when they are marked as non-flammable

XFactHD:
This is the LTS backport of #8160

Currently only the `FireBlock` asks the block itself wether it is flammable through the use of Forge extensions. This leads to issues like this: https://github.com/XFactHD/FramedBlocks/issues/31 where a block that uses `Material.WOOD` but returns `false` from `IForgeBlock#isFlammable()` catches fire from the lava but is never destroyed by the `FireBlock` because it thinks the block is not flammable.

This PR patches the `LavaFluid` to use the Forge extensions for flammability instead of checking the material for flammability. This way the behaviour of the `LavaFluid` is equivalent to the `FireBlock` which is already using these Forge extensions.

I can add a test mod for the flammability hooks if that is desired.

### DIFF
--- a/patches/minecraft/net/minecraft/fluid/LavaFluid.java.patch
+++ b/patches/minecraft/net/minecraft/fluid/LavaFluid.java.patch
@@ -29,17 +29,22 @@
              return true;
           }
        }
-@@ -104,8 +_,8 @@
+@@ -104,10 +_,15 @@
        return false;
     }
  
--   private boolean func_176368_m(IWorldReader p_176368_1_, BlockPos p_176368_2_) {
--      return p_176368_2_.func_177956_o() >= 0 && p_176368_2_.func_177956_o() < 256 && !p_176368_1_.func_175667_e(p_176368_2_) ? false : p_176368_1_.func_180495_p(p_176368_2_).func_185904_a().func_76217_h();
-+   private boolean isFlammable(IWorldReader p_176368_1_, BlockPos p_176368_2_, Direction face) {
-+      return p_176368_2_.func_177956_o() >= 0 && p_176368_2_.func_177956_o() < 256 && !p_176368_1_.func_175667_e(p_176368_2_) ? false : p_176368_1_.func_180495_p(p_176368_2_).isFlammable(p_176368_1_, p_176368_2_, face);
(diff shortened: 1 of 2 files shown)


patches/minecraft/net/minecraft/fluid/LavaFluid.java.patch
```diff
@@ -9,15 +9,42 @@
...
+@@ -104,10 +_,15 @@
+       return false;
+    }
+ 
++   /** @deprecated Forge: use {@link LavaFluid#isFlammable(IWorldReader,BlockPos,Direction)} instead */
+    private boolean func_176368_m(IWorldReader p_176368_1_, BlockPos p_176368_2_) {
+       return p_176368_2_.func_177956_o() >= 0 && p_176368_2_.func_177956_o() < 256 && !p_176368_1_.func_175667_e(p_176368_2_) ? false : p_176368_1_.func_180495_p(p_176368_2_).func_185904_a().func_76217_h();
+    }
+ 
++   private boolean isFlammable(IWorldReader p_176368_1_, BlockPos p_176368_2_, Direction face) {
```
sciwhiz12: Please give these parameters non-obfuscated names, to match the 1.17.x PR.

sciwhiz12 (changes requested):

ChampionAsh5357 (approved):

patches/minecraft/net/minecraft/fluid/LavaFluid.java.patch
```diff
@@ -9,15 +9,42 @@
...
++         if (this.isFlammable(p_176369_1_, p_176369_2_.func_177972_a(direction), direction.func_176734_d())) {
+             return true;
+          }
+       }
+@@ -104,10 +_,15 @@
+       return false;
+    }
+ 
++   /** @deprecated Forge: use {@link LavaFluid#isFlammable(IWorldReader,BlockPos,Direction)} instead */
+    private boolean func_176368_m(IWorldReader p_176368_1_, BlockPos p_176368_2_) {
```
gigaherz, line 38: Forgot Deprecated annotation here, which the 1.17 version has

XFactHD: Fixed

PR #8162, opened 2021-10-18
Fix the hopefully last issue with GAMELIBRARY loading

Minecraftschurli:
Currently using GAMELIBRARY mod files does not work as it throws a IndexOutOfBoundsException on startup because the mod list of a gamelibrary is empty.


ShroomAgent27:
[My test shows its okay.](https://github.com/FoalFreePress/TestingForgeMods/tree/main/8162)

ShroomAgent27 (approved):

PepperCode1 (approved):

PR #8163, opened 2021-10-19
Don't reload resources twice when creating worlds, Fix #8125 for 1.17

### LINKED ISSUE #8125: Loading an existing world with forge triggers two data reloads instead of one
**Minecraft Version:** 1.16.5

**Forge Version:** 36.2.5

**Logs:** https://gist.github.com/Shadows-of-Fire/87b74bcecca041ee3a7e6c060dae7062
[World Creation on Forge](https://gist.github.com/Shadows-of-Fire/87b74bcecca041ee3a7e6c060dae7062#file-forge-debug-log-L332)
[World _Load_ on Forge (two reloads)](https://gist.github.com/Shadows-of-Fire/87b74bcecca041ee3a7e6c060dae7062#file-forge-debug-log-L332)
[World Creation on Vanilla](https://gist.github.com/Shadows-of-Fire/87b74bcecca041ee3a7e6c060dae7062#file-forge-debug-log-L332)
[World _Load_ on Vanilla (one reload)](https://gist.github.com/Shadows-of-Fire/87b74bcecca041ee3a7e6c060dae7062#file-forge-debug-log-L332)

**Steps to Reproduce:**
 1. Create a new instance with just forge
 2. Make a world
 3. Exit said world
 4. Load into said already existing world
 5. Observe that two data reloads happen.
 6. Repeat on vanilla, observing that only a single reload happens in step (5)

**Description of issue:**
Something forge is doing causes data resource to be reloaded twice instead of once when loading into an already generated world in singleplayer (does not happen when making the world for the first time).  I cannot determine why or how this is happening, but it is a clear deviation from vanilla and it drastically increases startup time in larger packs.

alcatrazEscapee:
Background is in https://github.com/MinecraftForge/MinecraftForge/issues/8125#issuecomment-946683773

The solution?

- Notice that `WorldPersistenceHook` takes a `WorldData` parameter which is unused in Forge's implementations.
- Notice that we have all the context available to call directly into the world persistence hooks, from where we create the registries if we just pass in a `null` for this `WorldData`, with a simple added method in `LevelStorageSource`
- Delay a call to `RegistryAccess#builtin` until after the world persistence hooks have been run, in the case where we load an existing world.

Additionally
- The `WorldPersistenceHook` used by `ForgeMod` is... totally useless. It's from pre 1.16 `DimensionManager` code and should be ripped out and removed at one's earliest convienience.
- This doesn't build because of missing license headers, so I took the liberty to include those as well.

Draft because I'd like feedback on this while I (figure out how to) test this.

### DIFF
--- a/patches/minecraft/net/minecraft/client/Minecraft.java.patch
+++ b/patches/minecraft/net/minecraft/client/Minecraft.java.patch
@@ -310,12 +310,13 @@
     }
  
     private boolean m_91278_() {
-@@ -1814,11 +_,11 @@
+@@ -1814,11 +_,12 @@
     }
  
     public void m_91200_(String p_91201_) {
 -      this.m_91219_(p_91201_, RegistryAccess.m_123086_(), Minecraft::m_91133_, Minecraft::m_91135_, false, Minecraft.ExperimentalDialogType.BACKUP);
-+      this.doLoadLevel(p_91201_, RegistryAccess.m_123086_(), Minecraft::m_91133_, Minecraft::m_91135_, false, Minecraft.ExperimentalDialogType.BACKUP, false);
++      // We need to delay the call to RegistryAccess.builtin() as long as possible, in order to allow our world persistence hooks to run. So we pass in null here, and fix it later, only in the case where creating = false
++      this.doLoadLevel(p_91201_, null, Minecraft::m_91133_, Minecraft::m_91135_, false, Minecraft.ExperimentalDialogType.BACKUP, false);
     }
  
     public void m_91202_(String p_91203_, LevelSettings p_91204_, RegistryAccess.RegistryHolder p_91205_, WorldGenSettings p_91206_) {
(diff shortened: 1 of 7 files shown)


ChampionAsh5357:
This seems to address the issue about loading resources twice after a lot of double checking, but it introduces the fact that now all `WorldPersistenceHook` are loaded twice, once with a null world data and once with the loaded world data.

alcatrazEscapee:
> it introduces the fact that now all WorldPersistenceHook are loaded twice, once with a null world data and once with the loaded world data.

They are loaded twice anyway at the current time, once with a dummy world data that is immediately thrown away, and once with the actual world data. This PR doesn't introduce that.

I'm open to suggestions on how to work `WorldPersistenceHook` into a better location, but unsure about how to do that without breaking changes.

ChampionAsh5357:
> They are loaded twice anyway at the current time, once with a dummy world data that is immediately thrown away, and once with the actual world data.

Apologies, I was being unclear. While there is a dummy world data, it still contains the exact same information as the actual world data causing no really issues. This would force it to `null` which would be a contract change since the variable is expected to be non-null when passed in.

I think we should still generate a dummy world data to pass in there such that any existing programs that do use it can still have the same contract, just that the data itself will be invalid. Alternatively, deprecate the old method and have it default to one without the `WorldData` parameter. 

'd rather have something more that just clarifies the contract as nullable rather than just an annotation which wouldn't be known to the already implemented without looking at the superimplementation.

alcatrazEscapee:
Well, either creating a dummy, or removing the `WorldData` parameter are both valid solutions (I would suggest removing it and deprecating it, because I think the use cases of it are low).

I could also go ahead and remove the original injection point for `WorldPersistenceHook` if it was desirable to not do a duplicate injection anyway, but that would require finding a different injection point for server loads (and potentially, in `McRegionUpgrader` but really, there's no need to care about that).

alcatrazEscapee:
I've been busy and haven't had time to test this (and my forgedev env. has been bustificated) so with 1.18 looming I'll see if I can reimplement this there (actually breaking the world persistence hooks to fit in the new system)

PR #8164, opened 2021-10-21
[RFC] Implement JarInJar Dependency Loading on top of the LayeredZipFS

marchermans:
### Prelude
This implements the loading of in Jar dependencies using the LayeredZipFS implementation that is now available in the new SecureJarHandler version.

The implementation also leverages changes to ForgeSPI to allow for a second stage loading of the mods.
This second stage is known as dependency loading and takes in the mods loaded by stage 1 (which is the current system of mod loading using `IModLocators`) and scans for additional `IModFiles` to load.

### Implementation
As stated already, this new system leverages two additions to other libraries, part one being the LayeredZipFS and part two being ForgeSPI's addition of `IDependencyLocators`.
Some abstractions where introduced to allow for easy reuse of code (there is not much difference between loading a dependency and loading a mod in the first place). Besides this we modified the ModDiscoverer to takes the new additions into account.

#### Specifications:
The new JarInJarDependencyLocator internally leverages a file known as deps.txt in the dependencies subdirectory of the META-INF directory. Allthough initially I considered using a json file for this, or following some kind of maven directory structure, al of these seemed rather overkill especially since I just needed to support a list of dependencies to be loaded.
The deps.txt file is a CSV file where each line is made up out of 4 sections, seperated by a `;`. Lines starting with `#` are considered comments and excluded from processing. The sections of the CSV lines are interpretted as follows:
1) The maven coordinate of the artifact without version information (So basically in the form: `group:artifactId:classifier:versiondata` with classifier and versiondata being optional, however recommended to be supplied).
2) Supported version range in Apache Maven Version Specification (So for example: `[1.0,2.0)`, eg following the dependency specification format in the mods.toml)
3) Exact supplied version, has to be part of the supported version range.
4) The path to the file in the jar.

Once the dependency data entries from all dependencies from all mods are loaded from the deps.txt file from each file, we determine which jar to load. If multiple jars match the specifications we take an arbitrary jar.
If multiple none overlapping version ranges are specified we throw an exception, since this means that two mods would have a conflicting dependency.

### Related to:
- MinecraftForge/ForgeSPI#10 (**Needs to be merged before this, so that artifact versions can be updated accurately**)
- MinecraftForge/securejarhandler#14 (**Needs to be merged before this, so that artifact versions can be updated accurately**)


marchermans:
This needs retargetting for 1.18.
Note from myself.

@LexManos Note: Can we get the dependency PRs for this one (so the one for ForgeSPI and SJH) get a review and possibly a merge, they require a bump in spec potentially and I think having that ready for 1.18 makes a lot of sense.

PR #8167, opened 2021-10-24
Make WorldEvent.Load run for Overworld before chunk creation like other dimensions

TelepathicGrunt:
I was doing some testing by removing structures from a dimension by removing them from the world's chunk generate in WorldEvent.Load. What I noticed is while doing that worked perfectly for any dimension, Overworld was an exception where the structure still spawned around the world origin but my change seemed to apply past origin area (200 block radius around origin).

Looking at the patch for WorldEvent.Load, I noticed the event is added to a for loop that deliberately skips over the overworld inside but the hook is after the check. Meaning the Overworld's stuff was defined further above and when looking higher, I saw setInitialSpawn method being called after the Overworld is setup.

I believe setInitialSpawn is creating chunks around the spawn of the Overworld which causes WorldEvent.Load to fire when the overworld has already genned some chunks as it is in a different state now than the rest of the dimensions.

So this pr is to move the old WorldEvent.Load event into the loop's if statement to exclude the overworld there. Then make a second hook for the WorldEvent.Load event that is placed before setInitialSpawn method. That way when the event fires, modders know the overworld is not at a different stage or loading than other dimensions. Just something for better consistency.

Testing quickly with a temporary datapack and some small code seems like this PR works and now modders can be certain that the dimension is consistent each time the event fires now.

Also, someone forgot to put a license header on their file so that's slipped into this pr to help. but I can remove this extra license file thing if preferred.

Once this is merged, I'll make pr for 1.16.5

### DIFF
--- a/patches/minecraft/net/minecraft/server/MinecraftServer.java.patch
+++ b/patches/minecraft/net/minecraft/server/MinecraftServer.java.patch
@@ -9,14 +9,22 @@
           atomicreference.get().m_130011_();
        }, "Server thread");
        thread.setUncaughtExceptionHandler((p_177909_, p_177910_) -> {
-@@ -402,6 +_,7 @@
+@@ -365,6 +_,7 @@
+       this.f_129732_ = new CommandStorage(dimensiondatastorage);
+       WorldBorder worldborder = serverlevel.m_6857_();
+       worldborder.m_61931_(serverleveldata.m_5813_());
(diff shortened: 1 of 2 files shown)


Romaindu35 (approved):

ChampionAsh5357 (approved):

LexManos:
Altho your usecase is bad, as there are other better ways to deal with world generation data. This doesn't seem to hurt anyting that I can see. They manually put the overworld in the world map before your hook. So handlers should still be able to look up things. The only issues would come with people assuming that the player list is in a valid state, which doesn't seem like a big userbase.

PR #8168, opened 2021-10-24
[1.17.x] Changing hardcoded Items.FISHING_ROD to allow for custom fishing rods

rehwinkel:
With this PR, modders can add their custom fishing rods to the game. Essentially, I just replaced `Items.FISHING_ROD` mentions with `instanceof` checks, which allow for custom `FishingRodItem`s to throw the bobber too.


TheCurle:
Would it not be better to use an item tag here?

src/test/java/net/minecraftforge/debug/item/CustomFishingRodTest.java
```diff
@@ -0,0 +1,53 @@
...
+                        flag1 = false;
+                    }
+
+                    return (flag || flag1) && entity instanceof Player && ((Player) entity).fishing != null ? 1.0F : 0.0F;
+                }
+            });
+        });
+    }
+
+    private static class CustomFishingRodItem extends FishingRodItem {
```
TheCurle: Why does this class exist? Just use `new FishingRodItem(...)`...

src/test/java/net/minecraftforge/debug/item/CustomFishingRodTest.java
```diff
@@ -0,0 +1,53 @@
...
+                if (entity == null) {
+                    return 0.0F;
+                } else {
+                    boolean flag = entity.getMainHandItem() == stack;
+                    boolean flag1 = entity.getOffhandItem() == stack;
+                    if (entity.getMainHandItem().getItem() instanceof FishingRodItem) {
+                        flag1 = false;
+                    }
+
+                    return (flag || flag1) && entity instanceof Player && ((Player) entity).fishing != null ? 1.0F : 0.0F;
```
TheCurle: Letter soup. What is all this? Where's it from? Add a comment with at least a LITTLE explanation.
If this test fails in a future version, we'd at least like to know where it originated.

rehwinkel: I'll make it more clear

src/test/java/net/minecraftforge/debug/item/CustomFishingRodTest.java
rehwinkel: Removed it!

rehwinkel:
I'll make this a tag then, that does seem a little better than instanceof

rehwinkel:
Where would I put the Tag? In `net.minecraftforge.common.Tags`? That seems more for crafting tags than for functional ones.

SizableShrimp:
Crafting tags are functional ones? And other mods can use those standardized tags for more than just crafting. As far as I know, that's the only place for forge tags.

patches/minecraft/net/minecraft/client/renderer/item/ItemProperties.java.patch
```diff
@@ -96,7 +96,13 @@
           if (p_174597_ == null) {
              return 0.0F;
           } else {
-@@ -267,13 +_,13 @@
+             boolean flag = p_174597_.m_21205_() == p_174595_;
+             boolean flag1 = p_174597_.m_21206_() == p_174595_;
+-            if (p_174597_.m_21205_().m_41720_() instanceof FishingRodItem) {
++            if (p_174597_.m_21205_().m_150930_(Items.f_42523_)) {
```
marchermans: I am pretty sure you missed a check here.

rehwinkel: Nope, left that one there intentionally. This is the item property code for the fishing rod. Since the property is only registered for that specific item (Items.FISHING_ROD), generifying that check is pointless.

ShroomAgent27:
[Looks good to me. ](https://github.com/FoalFreePress/TestingForgeMods/tree/0947d6dc978e9d92bf18f262b400b8c657a4db21/8168/src/main/java/com/example/examplemod)

ShroomAgent27 (approved):

pupnewfster:
Is there any actual benefit to this using tag checks instead of instanceof checks? Given it seems like this would be a lot cleaner especially from a maintenance perspective if it just used forge's [field to instanceof transformer](https://github.com/MinecraftForge/MinecraftForge/blob/1.17.x/build.gradle#L1014-L1027).

KnightMiner:
Alternatively, why not tool actions? I don't think forge has a field to tool action transformer, but one would not be hard to create. Tool actions means the behavior could be conditional. Also means you don't have to extend the vanilla class to get the behavior, sorta like a tag would let you do.

rehwinkel:
@pupnewfster replacing the field `Items.FISHING_ROD` with instanceof everywhere (using the transformer) would be undesirable. The item property of the fishing rod depends on the Item being held, and I don't think vanilla fishing rods should display as "cast" when ANY instance of `FishingRodItem` is cast & held. Other than that case, manually replacing the field with instanceof instead of a tag would be possible, but @TheCurle suggested using a tag, so I went with that.

PR #8169, opened 2021-10-25
RFC [1.18] Ping data compression

diesieben07:
This gets rid of the need for truncating the ping data by using a compact binary format instead of plain JSON.
Requires a bump in the FML Network Version obviously, so this should be done once 1.18 drops.

A drawback of course is that the data is now not at all human readable, however I would argue that that is worth it considering this enables the ping data to work properly no matter the size of the mod and channel list.

Some numbers:
For the "All The Mods 7" modpack (which is the largest one I could find for 1.17, but granted I did not search much) the old format would use 11120 characters of JSON data (not truncated). The new format uses 2269 characters of JSON data.


alcatrazEscapee:
> Requires a bump in the FML Network Version obviously, so this should be done once 1.18 drops.

Then why is being made backwards compatible? Is there actually a use case for connecting to different FML network versions across different Minecraft versions? Seems like that would be an excellent time to strip out all the deprecated previous version code, no?

diesieben07:
> Then why is being made backwards compatible?

There are two ways in which it is kept backwards compatible:

1. The serializer writes out empty dummy arrays for the `mods` and `channel` properties. This is necessary, because previous versions have a hard requirement on these properties. Not having them will log a warning message (at debug level) and the server will assumed to be vanilla.
2. The deserializer accepts the old format still, so that data from old servers can still be displayed, even if a connection is not possible. This one is not technically necessary.

alcatrazEscapee:
> Not having them will log a warning message (at debug level) and the server will assumed to be vanilla.

On any connection that isn't forge-forge, this would just be ignored, no? Because there's no data present on either side. But I think the thing I still don't understand is, if this is targeting 1.18:

> This is necessary, because previous versions have a hard requirement on these properties

what use case do you have for connecting a previous **minecraft** version, to a newer one, where both are modded Forge? Isn't that an ideal time to introduce this breaking network change?

diesieben07:
This change is about the server ping data, not for actually connecting to the server.

alcatrazEscapee:
Ahhhh, so you can absolutely have servers listed for incompatible Minecraft versions that then also need to avoid crashing due to incompatible pings - gotcha. Thanks for clarifying that :)

diesieben07:
Updated to 1.18 and now ready for review.

diesieben07:
- Merged current 1.18.x
- Added possible truncation for even the compressed data, even though this should never happen.
- Revert bump of FMLNETVERSION

The behavior is now as follows:
- old client / old server: as before
- old client / new server: will show "incompatible" and "truncated" (but connection will succeed if the data is actually compatible)
- new client / old server: as before (code can still read the old format)
- new client / new server: proper checks without truncation

The encoded data is now truncated if it reaches around 60,000 bytes (before optimized encoding into UTF-16). Testing with All The Mods 7 shows that that modpack uses 5676 bytes out of the possible 60,000 - with over 200 mods.

src/main/java/net/minecraftforge/network/ServerStatusPing.java
```diff
@@ -107,60 +191,232 @@ public static ServerStatusPing deserialize(JsonObject forgeData, JsonDeserializa
...
+        /**
+         * Decode binary data encoded by {@link #encodeOptimized}
+         */
+        private static ByteBuf decodeOptimized(String s)
+        {
+            var size0 = ((int) s.charAt(0));
+            var size1 = ((int) s.charAt(1));
+            var size = size0 | (size1 << 15);
+
+            var buf = Unpooled.buffer(size);
```
sciwhiz12, line 381: Reading through https://github.com/netty/netty/issues/9991 and the code for the buffer returned by `Unpooled.buffer`, specifically how it implements `release()`, it may be good to ensure this buffer is released promptly in this class rather than letting the GC handle it (because the buffer is a heap buffer) and letting Netty report it as a leaked buffer.

src/main/java/net/minecraftforge/network/ServerStatusPing.java
```diff
@@ -58,21 +66,50 @@
...
+ * <code>
+ * <pre>
+ * {
+ *     "fmlNetworkVersion": FMLNETVERSION,
+ *     "channels": [],
+ *     "mods": [],
+ *     "d": "&lt;binary data&gt;"
+ * }
+ * </pre>
+ * </code>
```
sciwhiz12: This should use a `{@code}` block tag wrapped in `<pre>`:
```suggestion
 * <pre>{@code
 * {
 *     "fmlNetworkVersion": FMLNETVERSION,
 *     "channels": [],
 *     "mods": [],
 *     "d": "&lt;binary data&gt;"
 * }
 * }</pre>
```

src/main/java/net/minecraftforge/network/ServerStatusPing.java
```diff
@@ -35,13 +37,19 @@
...
 import java.util.function.Supplier;
 
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
 /**
+ * Represents additional data sent by FML when a server is pinged.
+ * Previous versions used the following format:
+ * <code>
+ * <pre>
```
sciwhiz12: This should use a `{@code}` block tag wrapped in `<pre>`:
```suggestion
 * <pre>{@code
```

src/main/java/net/minecraftforge/network/ServerStatusPing.java
```diff
@@ -58,21 +66,50 @@
  *          }
  *     ]
  * }
+ * </pre>
+ * </code>
```
sciwhiz12: This should use a `{@code}` block tag wrapped in `<pre>`: (this is the closing for the preceding review comment)
```suggestion
 * }</pre>
```

src/main/java/net/minecraftforge/network/ServerStatusPing.java
```diff
@@ -35,13 +37,19 @@
...
 import java.util.stream.StreamSupport;
 
 /**
+ * Represents additional data sent by FML when a server is pinged.
+ * Previous versions used the following format:
+ * <code>
+ * <pre>
  * {
  *      "fmlNetworkVersion" : FMLNETVERSION,
  *      "channels": [
```
sciwhiz12: Since this is class is already being modified, remove an extra space from these two to be lined up with the `"mods"` part:
```suggestion
 *     "fmlNetworkVersion" : FMLNETVERSION,
 *     "channels": [
```

sciwhiz12 (changes requested):
The code seems solid to me, and testing the PR seems to show that it works as inteded:
boop

I only have a few small comments regarding javadocs and the manual release of an allocated heap buffer, otherwise this looks ready for assignment.

sciwhiz12 (approved):

LexManos:
This data is always going to be on a limited buffer, so moving to a binary format just means we buy more time before mod packs grow too large for the buffer to be accurate. However, this is purely a metadata/informative aspect so this change does not break anything, and having the truncated data didn't break anything. It's just a 'best attempt' to see if the server is compatible. So going to accept this.

PR #8170, opened 2021-10-25
Add IExtensibleEnum to Biome.TemperatureModifier

Luis-St:
This PR add the `IExtensibleEnum` to the `Biome.TemperatureModifier`

Why is PR needed?
This PR allows Modders to create their own `TemperatureModifier`,
for example they can be used for:
if you are create a `Dimension` which looks similar to the `Overworld`, 
but the base height of the World is moved to the top/bottom (e.g. `SeaLevel` at 128), 
you can use the custom `TemperatureModifier` to rework the vanilla temperature-system,
since vanilla reduces the temperature the higher you go (it ensures that it snows in the mountains).
Than you can use the custom `TemperatureModifier` to make this logic working for your `Dimension`/the base height you use

### DIFF
--- a/patches/minecraft/net/minecraft/world/level/biome/Biome.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/biome/Biome.java.patch
@@ -20,3 +20,41 @@
     });
     public static final Codec<Biome> f_47430_ = RecordCodecBuilder.create((p_47504_) -> {
        return p_47504_.group(Biome.ClimateSettings.f_47679_.forGetter((p_151703_) -> {
+@@ -607,7 +_,7 @@
+       }
+    }
+ 
+-   public static enum TemperatureModifier implements StringRepresentable {
++   public static enum TemperatureModifier implements StringRepresentable, net.minecraftforge.common.IExtensibleEnum {
+       NONE("none") {
+          public float m_8117_(BlockPos p_47767_, float p_47768_) {
+             return p_47768_;
+@@ -635,10 +_,27 @@
+          return p_47753_;
+       }));
+ 
+-      public abstract float m_8117_(BlockPos p_47754_, float p_47755_);
++      public float m_8117_(BlockPos p_47754_, float p_47755_) { return this.temperatureModifier.modifyTemperature(p_47754_, p_47755_); }
+ 
+       TemperatureModifier(String p_47745_) {
+          this.f_47738_ = p_47745_;
++      }
++
++      private BiomeTemperatureModifier temperatureModifier;
++      private TemperatureModifier(String modifierName, BiomeTemperatureModifier temperatureModifier) {
++         this(modifierName);
++         this.temperatureModifier = temperatureModifier;
++      }
++      public static TemperatureModifier create(String name, String modifierName, BiomeTemperatureModifier temperatureModifier) {
++    	  throw new IllegalStateException("Enum not extended");
++      }
++      @java.lang.Override
++      public void init() {
++         f_47739_.put(this.m_47758_(), this);
++      }
++      @java.lang.FunctionalInterface
++      public interface BiomeTemperatureModifier {
++         float modifyTemperature(BlockPos pos, float noise);
+       }
+ 
+       public String m_47758_() {
--- a/patches/minecraft/net/minecraft/world/level/biome/Biome.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/biome/Biome.java.patch
@@ -29,7 +29,7 @@
        NONE("none") {
           public float m_8117_(BlockPos p_47767_, float p_47768_) {
              return p_47768_;
-@@ -635,10 +_,27 @@
+@@ -635,11 +_,30 @@
           return p_47753_;
        }));
  
@@ -38,23 +38,26 @@
  
        TemperatureModifier(String p_47745_) {
           this.f_47738_ = p_47745_;
-+      }
+       }
 +
++      /* ***** FORGE START ***** */
 +      private BiomeTemperatureModifier temperatureModifier;
 +      private TemperatureModifier(String modifierName, BiomeTemperatureModifier temperatureModifier) {
 +         this(modifierName);
 +         this.temperatureModifier = temperatureModifier;
 +      }
 +      public static TemperatureModifier create(String name, String modifierName, BiomeTemperatureModifier temperatureModifier) {
-+    	  throw new IllegalStateException("Enum not extended");
++          throw new IllegalStateException("Enum not extended");
 +      }
-+      @java.lang.Override
++      @Override
 +      public void init() {
 +         f_47739_.put(this.m_47758_(), this);
 +      }
-+      @java.lang.FunctionalInterface
++      @FunctionalInterface
 +      public interface BiomeTemperatureModifier {
 +         float modifyTemperature(BlockPos pos, float noise);
-       }
++      }
++      /* ***** FORGE END ***** */
  
        public String m_47758_() {
+          return this.f_47738_;
--- a/src/test/java/net/minecraftforge/debug/world/BiomeTemperatureModifierTest.java
+++ b/src/test/java/net/minecraftforge/debug/world/BiomeTemperatureModifierTest.java
@@ -40,34 +40,51 @@ import net.minecraftforge.fmllegacy.RegistryObject;
 import net.minecraftforge.registries.DeferredRegister;
 import net.minecraftforge.registries.ForgeRegistries;
 
-@Mod("biome_temperature_modifier_test")
+@Mod(BiomeTemperatureModifierTest.MOD_ID)
 public class BiomeTemperatureModifierTest 
 {
+	static final String MOD_ID = "biome_temperature_modifier_test";
     private static final boolean ENABLED = false;
-    public static final DeferredRegister<Biome> BIOMES = DeferredRegister.create(ForgeRegistries.BIOMES, "biome_temperature_modifier_test");
-    public static final ResourceKey<Biome> TEST_BIOME_KEY = ResourceKey.create(Registry.BIOME_REGISTRY, new ResourceLocation("biome_temperature_modifier_test", "test_biome"));
-    public static final RegistryObject<Biome> TEST_BIOME = BIOMES.register("test_biome", BiomeTemperatureModifierTest::makeTestBiome);
+    public static final DeferredRegister<Biome> BIOMES = DeferredRegister.create(ForgeRegistries.BIOMES, MOD_ID);
+    public static final ResourceKey<Biome> TEST_BIOME_KEY = ResourceKey.create(Registry.BIOME_REGISTRY, new ResourceLocation(MOD_ID, "test_biome"));
+    public static final RegistryObject<Biome> TEST_BIOME = BIOMES.register("test_biome", () -> BiomeTemperatureModifierTest.TEST_BIOME_BIOME);
 
     public BiomeTemperatureModifierTest() 
     {
+    	if (!ENABLED) return;
         BIOMES.register(FMLJavaModLoadingContext.get().getModEventBus());
         FMLJavaModLoadingContext.get().getModEventBus().addListener(this::commonSetup);
     }
 
     private void commonSetup(FMLCommonSetupEvent event) 
     {
-        if (!ENABLED) return;
         event.enqueueWork(() -> {
             BiomeDictionary.addTypes(TEST_BIOME_KEY, BiomeDictionary.Type.MODIFIED);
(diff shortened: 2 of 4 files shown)


ChampionAsh5357:
This needs a test case to verify that the codec does work and can deserialize the injected enum entries.

Luis-St:
How do I test if the Codec works?
Via a Test Mod which creats custom Biome which use a new `Biome.TemperatureModifier` ?

ChampionAsh5357:
Yep, that should be more than fine.

Luis-St:
Okay so I have now added a Test Mod that adds a Biome that uses a custom Biome.TemperatureModifier, 
I've test it and it works

patches/minecraft/net/minecraft/world/level/biome/Biome.java.patch
```diff
@@ -20,3 +20,41 @@
...
++
++      private BiomeTemperatureModifier temperatureModifier;
++      private TemperatureModifier(String modifierName, BiomeTemperatureModifier temperatureModifier) {
++         this(modifierName);
++         this.temperatureModifier = temperatureModifier;
++      }
++      public static TemperatureModifier create(String name, String modifierName, BiomeTemperatureModifier temperatureModifier) {
++    	  throw new IllegalStateException("Enum not extended");
++      }
++      @java.lang.Override
```
sciwhiz12: The fully-qualified reference to `java.lang.Override` is not needed, as all classes in `java.lang` are automatically imported:

> Every compilation unit implicitly imports every `public` type name declared in the predefined package `java.lang`, as if the declaration `import java.lang.*;` appeared at the beginning of each compilation unit immediately after any `package` statement. As a result, the names of all those types are available as simple names in every compilation unit. 
> ~ _[The Java™ Language Specification, Java SE 16 Edition, § 7.3. Compilation Units](https://docs.oracle.com/javase/specs/jls/se7/html/jls-7.html#jls-7.3)_

patches/minecraft/net/minecraft/world/level/biome/Biome.java.patch
```diff
@@ -20,3 +20,41 @@
...
++      public void init() {
++         f_47739_.put(this.m_47758_(), this);
++      }
++      @java.lang.FunctionalInterface
```
sciwhiz12: Same comment here on the unnecessary full qualification to `java.lang.FunctionalInterface`.

patches/minecraft/net/minecraft/world/level/biome/Biome.java.patch
```diff
@@ -20,3 +20,41 @@
...
+       }));
+ 
+-      public abstract float m_8117_(BlockPos p_47754_, float p_47755_);
++      public float m_8117_(BlockPos p_47754_, float p_47755_) { return this.temperatureModifier.modifyTemperature(p_47754_, p_47755_); }
+ 
+       TemperatureModifier(String p_47745_) {
+          this.f_47738_ = p_47745_;
++      }
++
++      private BiomeTemperatureModifier temperatureModifier;
```
sciwhiz12, line 44: Since this whole block is added by a Forge patch, it may be good to add marker comments in the form of `/* ***** FORGE [START/END] ***** */` before and after this block.

src/test/java/net/minecraftforge/debug/world/BiomeTemperatureModifierTest.java
```diff
@@ -0,0 +1,73 @@
...
+        FMLJavaModLoadingContext.get().getModEventBus().addListener(this::commonSetup);
+    }
+
+    private void commonSetup(FMLCommonSetupEvent event) 
+    {
+        if (!ENABLED) return;
+        event.enqueueWork(() -> {
+            BiomeDictionary.addTypes(TEST_BIOME_KEY, BiomeDictionary.Type.MODIFIED);
+            BiomeManager.addBiome(BiomeManager.BiomeType.WARM, new BiomeEntry(TEST_BIOME_KEY, 100));
+        });
```
sciwhiz12: Are there any side-effects of registering a biome but not registering it with the `BiomeDictionary` or `BiomeManager`? It may be better to put the enabled check in the constructor instead.

patches/minecraft/net/minecraft/world/level/biome/Biome.java.patch
```diff
@@ -20,3 +20,41 @@
...
++      private TemperatureModifier(String modifierName, BiomeTemperatureModifier temperatureModifier) {
++         this(modifierName);
++         this.temperatureModifier = temperatureModifier;
++      }
++      public static TemperatureModifier create(String name, String modifierName, BiomeTemperatureModifier temperatureModifier) {
++    	  throw new IllegalStateException("Enum not extended");
```
sciwhiz12: There is a tab mixed with the spaces here.
tab mixed with spaces

src/test/java/net/minecraftforge/debug/world/BiomeTemperatureModifierTest.java
```diff
@@ -0,0 +1,73 @@
...
+import net.minecraftforge.fmllegacy.RegistryObject;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+
+@Mod("biome_temperature_modifier_test")
+public class BiomeTemperatureModifierTest 
+{
+    private static final boolean ENABLED = false;
+    public static final DeferredRegister<Biome> BIOMES = DeferredRegister.create(ForgeRegistries.BIOMES, "biome_temperature_modifier_test");
+    public static final ResourceKey<Biome> TEST_BIOME_KEY = ResourceKey.create(Registry.BIOME_REGISTRY, new ResourceLocation("biome_temperature_modifier_test", "test_biome"));
```
sciwhiz12: Since the mod ID is used in multiple places here, it should be stored in a `static final` field.

src/test/java/net/minecraftforge/debug/world/BiomeTemperatureModifierTest.java
```diff
@@ -0,0 +1,73 @@
...
+        if (!ENABLED) return;
+        event.enqueueWork(() -> {
+            BiomeDictionary.addTypes(TEST_BIOME_KEY, BiomeDictionary.Type.MODIFIED);
+            BiomeManager.addBiome(BiomeManager.BiomeType.WARM, new BiomeEntry(TEST_BIOME_KEY, 100));
+        });
+    }
+
+    private static Biome makeTestBiome() 
+    {
+        Biome.TemperatureModifier temperatureModifier = Biome.TemperatureModifier.create("TEST_MODIFIER", "test_modifier", (pos, noise) -> 2.0F);
```
sciwhiz12: Shouldn't this be created and stored in a `static final` field?

src/test/java/net/minecraftforge/debug/world/BiomeTemperatureModifierTest.java
```diff
@@ -0,0 +1,73 @@
...
+        return new Biome.BiomeBuilder().biomeCategory(BiomeCategory.NONE).depth(0.1F).scale(0.1F).downfall(0.0F).generationSettings(new BiomeGenerationSettings.Builder().surfaceBuilder(SurfaceBuilders.GRASS).build())
+        .mobSpawnSettings(MobSpawnSettings.EMPTY).specialEffects(new BiomeSpecialEffects.Builder().waterColor(4159204).waterFogColor(329011).fogColor(12638463).skyColor(128)
+        .ambientMoodSound(AmbientMoodSettings.LEGACY_CAVE_SETTINGS).build()).precipitation(Precipitation.NONE).temperature(2.0F).temperatureAdjustment(temperatureModifier).build();
```
sciwhiz12: Please cleanup the formatting for the biome code, and also please explain where these settings came from (whether you made them up on-the-fly or taken from another existing biome).

sciwhiz12 (changes requested):

src/test/java/net/minecraftforge/debug/world/BiomeTemperatureModifierTest.java
Luis-st: it could be, but when i clean up the code of the Biome,
i want to adapt it to the vanilla code style (for example refer to the `VanillaBiome` class)

Luis-St:
the code has been cleaned up

src/test/java/net/minecraftforge/debug/world/BiomeTemperatureModifierTest.java
```diff
@@ -0,0 +1,90 @@
...
+import net.minecraftforge.fml.event.lifecycle.FMLCommonSetupEvent;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.fmllegacy.RegistryObject;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+
+@Mod(BiomeTemperatureModifierTest.MOD_ID)
+public class BiomeTemperatureModifierTest 
+{
+	static final String MOD_ID = "biome_temperature_modifier_test";
```
ChampionAsh5357: Tab

src/test/java/net/minecraftforge/debug/world/BiomeTemperatureModifierTest.java
```diff
@@ -0,0 +1,90 @@
...
+        event.enqueueWork(() -> {
+            BiomeDictionary.addTypes(TEST_BIOME_KEY, BiomeDictionary.Type.MODIFIED);
+            BiomeManager.addBiome(BiomeManager.BiomeType.WARM, new BiomeEntry(TEST_BIOME_KEY, 100));
+        });
+    }
+
+    /*
+     * the Biome is only made for testing, the parameters which are used are test parameters
+     */
+	private static final Biome TEST_BIOME_BIOME = new Biome.BiomeBuilder()
```
ChampionAsh5357: Tab

src/test/java/net/minecraftforge/debug/world/BiomeTemperatureModifierTest.java
```diff
@@ -0,0 +1,90 @@
...
+            .specialEffects(new BiomeSpecialEffects.Builder()
+                    .waterColor(255)
+                    .waterFogColor(80)
+                    .fogColor(16777215)
+                    .skyColor(65535)
+                    .ambientMoodSound(AmbientMoodSettings.LEGACY_CAVE_SETTINGS)
+                    .build())
+            .temperatureAdjustment(Biome.TemperatureModifier.create("TEST_MODIFIER", "test_modifier", (pos, noise) -> 2.0F))
+            .build();
+}
```
ChampionAsh5357, line 90: New line at end of file.

src/test/java/net/minecraftforge/debug/world/BiomeTemperatureModifierTest.java
ChampionAsh5357: This should be registered properly, not as a static initialized entry.

ChampionAsh5357 (changes requested):
The biome should be in json anyways with a dummy registered that will be replaced. That would be a more proper test case than just referring to it within code.

src/test/java/net/minecraftforge/debug/world/BiomeTemperatureModifierTest.java
Luis-st: t/fill

Luis-St: i will move this directly into the Supplier of the RegistryObject,
and about the json Biome, is it possible to generate it via the Data generator?
and if yes, where do i need to set the mod id for the generator?

ChampionAsh5357: Currently there is no data generator for world gen objects, there is a PR for it though: #7257

Luis-St: okay good to know

Luis-St:
@ChampionAsh5357 I have accepted your proposed changes, 
the json Biome use other settings than the "code" Biome to make it clear that the "code" Biome is being replaced

src/test/java/net/minecraftforge/debug/world/BiomeTemperatureModifierTest.java
ChampionAsh5357: Other than this, you should be good.

Luis-St:
I close this PR, since vanilla change a few things in Biome generation. 
And i don't know if i have the time to update this PR.
if someone wants to add a IExtensibleEnum to Biome.TemperatureModifier feel free to use my code in this PR.

PR #8171, opened 2021-10-25
[1.17.x] Added a hook to activate spyglass-style scoping

Da-Technomancer:
Added a hook to trigger the 'scoping' effect the spyglass uses (zoom in, visual overlay, adjusted movement).

Mods can register a predicate which will be tested on the client side to see whether the scoping effect is being externally triggered. If any registered predicate returns true, scoping will be active. 

Also wrote a test mod.


src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -21,16 +21,9 @@
...
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Deque;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
diesieben07: Don't use * imports

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1321,6 +1314,22 @@ public static void modifyAttributes()
         });
     }
 
+    private static final List<Predicate<Player>> SPYGLASS_PREDICATES = new ArrayList<>();
+
+    /**
+     * Registers a predicate which will be checked for whether the player is considered to be 'scoping' with a spyglass on the client
+     * Only one registered predicate needs to be true for scoping to occur
```
diesieben07: Missing periods at the ends of the sentences.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1321,6 +1314,22 @@ public static void modifyAttributes()
...
+     * @param spyglassCondition A predicate which will be checked on the client side. Might be tested very frequently
+     */
+    public static void registerSpyglassCondition(Predicate<Player> spyglassCondition)
+    {
+        SPYGLASS_PREDICATES.add(spyglassCondition);
```
diesieben07: This method should be made threadsafe, to enable use from `FMLCommonSetupEvent`.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1321,6 +1314,22 @@ public static void modifyAttributes()
...
+    }
+
+    public static boolean isPlayerScoping(Player player){
```
diesieben07: Formatting (braces on new lines).

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1321,6 +1314,22 @@ public static void modifyAttributes()
...
+        return SPYGLASS_PREDICATES.stream().anyMatch(pred -> pred.test(player));
```
diesieben07: Considering this is called frequently, you should use a normal foreach loop for performance reasons.

src/test/java/net/minecraftforge/debug/entity/player/PlayerSpyglassScopeTest.java
```diff
@@ -0,0 +1,43 @@
...
+@Mod(PlayerSpyglassScopeTest.MODID)
+public class PlayerSpyglassScopeTest
+{
+
+    static final String MODID = "player_spyglass_scope_test";
+    static final boolean ENABLE = false;
+
+    public PlayerSpyglassScopeTest()
+    {
+        if(ENABLE){
```
diesieben07: Formatting (whitespace and braces on newlines)

diesieben07 (changes requested):

src/test/java/net/minecraftforge/debug/entity/player/PlayerSpyglassScopeTest.java
```diff
@@ -0,0 +1,44 @@
...
+@Mod(PlayerSpyglassScopeTest.MODID)
+public class PlayerSpyglassScopeTest
+{
+
+    static final String MODID = "player_spyglass_scope_test";
+    static final boolean ENABLE = false;
+
+    public PlayerSpyglassScopeTest()
+    {
+        if(ENABLE)
```
diesieben07: Missing space after if

src/main/java/net/minecraftforge/common/SpyglassHooks.java
```diff
@@ -0,0 +1,35 @@
...
+    public static synchronized void registerSpyglassCondition(Predicate<Player> spyglassCondition)
+    {
+        SPYGLASS_PREDICATES.add(spyglassCondition);
+    }
+
+    public static boolean isPlayerScoping(Player player)
+    {
+        for(Predicate<Player> spyglassPredicate : SPYGLASS_PREDICATES)
+        {
+            if(spyglassPredicate.test(player))
```
diesieben07: Missing space after if and for

diesieben07 (changes requested):

src/main/java/net/minecraftforge/common/SpyglassHooks.java
```diff
@@ -0,0 +1,35 @@
+package net.minecraftforge.common;
```
sciwhiz12, line 20: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/test/java/net/minecraftforge/debug/entity/player/PlayerSpyglassScopeTest.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Tests the patch to Player to allow alternate triggers for the spyglass scoping effect
+ * While enabled, sneaking will cause the scoping effect
+ */
+@Mod(PlayerSpyglassScopeTest.MODID)
+public class PlayerSpyglassScopeTest
+{
+
+    static final String MODID = "player_spyglass_scope_test";
+    static final boolean ENABLE = false;
```
sciwhiz12: The convention is for the field controlling the activation status of a test mod to be named `ENABLED`.

src/main/java/net/minecraftforge/common/SpyglassHooks.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class SpyglassHooks
+{
+
+    private static final List<Predicate<Player>> SPYGLASS_PREDICATES = new ArrayList<>();
+
+    /**
+     * Registers a predicate which will be checked for whether the player is considered to be 'scoping' with a spyglass on the client.
+     * Only one registered predicate needs to be true for scoping to occur.
+     * @param spyglassCondition A predicate which will be checked on the client side. Might be tested very frequently.
```
sciwhiz12: The comment regarding the given predicate being likely to be called very frequently (how frequently? every tick?) should be moved to the javadoc body proper and **marked in bold** (placed between `<strong>` and `</strong>`).

Also, the phrase _"on the client"_ suggests that this predicate is called only on the client-side. Is this true, or is it also called on the server-side (as the existence of the check method on the `Player` class would likely indicate)?

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/common/SpyglassHooks.java
Da-Technomancer: Done in the latest commit. Note that the task also added a header to an unrelated class.

src/main/java/net/minecraftforge/common/SpyglassHooks.java
Da-Technomancer: "Also, the phrase "on the client" suggests that this predicate is called only on the client-side. Is this true, or is it also called on the server-side (as the existence of the check method on the Player class would likely indicate)?"
The method in Player is only ever called on the client side in vanilla usage.

Da-Technomancer:
I would like to note that running the `updateLicenses` task as suggested also added a license header to an unrelated class- which was presumably missing a header before.
I've included that change in the commit. 

Edit: Reverted that change in a later commit.

patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
```diff
@@ -346,6 +346,15 @@
        return this.m_36218_(mutablecomponent);
     }
  
+@@ -1948,7 +_,7 @@
+    }
+ 
+    public boolean m_150108_() {
+-      return this.m_6117_() && this.m_21211_().m_150930_(Items.f_151059_);
++      return this.m_6117_() && this.m_21211_().m_150930_(Items.f_151059_) || net.minecraftforge.common.SpyglassHooks.isPlayerScoping(this);
```
rehwinkel, line 354: Could it be desirable to override the default spyglass behavior, like disabling vanilla spyglasses from working to add custom behavior? With this current method, default spyglasses would always "work". Just a some food for thought.

src/main/java/net/minecraftforge/common/SpyglassHooks.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param spyglassCondition A predicate which will be checked on the client side.
+     */
+    public static synchronized void registerSpyglassCondition(Predicate<Player> spyglassCondition)
+    {
+        SPYGLASS_PREDICATES.add(spyglassCondition);
+    }
+
+    public static boolean isPlayerScoping(Player player)
+    {
+        for (Predicate<Player> spyglassPredicate : SPYGLASS_PREDICATES)
```
rehwinkel, line 48: This code looks fine, however I think having an opposite check would be useful too. Like "if this condition matches, DO NOT show the scoping effect". Something like a paralysis potion effect, or blindness might wish to disable scoping, just an idea though.

src/test/java/net/minecraftforge/debug/block/FullPotsAccessorDemo.java
```diff
@@ -1,3 +1,22 @@
...
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
```
rehwinkel: Does this belong in this PR? On a related note, I've also seen that some datagen-assets are missing in git, maybe making an extra-PR for adding this license header and the mentioned assets would be better.

rehwinkel (commented):
The intended changes look good, I added a little suggestion for your consideration though.

patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
Da-Technomancer: There are existing ways for mods to disable the vanilla spyglass behavior (by changing the item or preventing it being set to active).

src/main/java/net/minecraftforge/common/SpyglassHooks.java
sciwhiz12: That would either mean holding another list of predicates which deny the scoping behavior and run before the allowing predicates, or something akin to a `Function<Player, Result>` where the result encodes some status like 'defer', 'allow (if not blocked)', or 'block'; am I correct?

sciwhiz12:
Hello! @Da-Technomancer, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable. If we do not hear back from you after some time, this PR will be closed for inactivity.

Da-Technomancer:
I've now rebased this PR onto 1.18, in a separate PR (here: https://github.com/MinecraftForge/MinecraftForge/pull/8325).
Closing this PR, as it is stale and for an outdated version.

PR #8172, opened 2021-10-26
New events for Minecarts to handle things like making trains

rehwinkel:
This PR provides 4 events to the `AbstractMinecart` entity:
* `MinecartEvent.Steered` for when a player moves in a minecart
* `MinecartEvent.Pushed` for when a minecart is pushed by another entity
* `MinecartEvent.PreMove` called before the code for moving along the track is executed, cancellable
* `MinecartEvent.PostMove` to adjust movements after the fact

Theses events aim to replace the old Minecart events (such as `MinecartUpdateEvent`). I've also written a little test mod that makes linking minecarts together possible, just like RailCraft does it.


patches/minecraft/net/minecraft/world/entity/vehicle/AbstractMinecart.java.patch
```diff
@@ -148,7 +156,15 @@
...
  
-@@ -445,7 +_,7 @@
+@@ -440,12 +_,14 @@
+          double d9 = vec32.m_165925_();
+          double d11 = this.m_20184_().m_165925_();
+          if (d9 > 1.0E-4D && d11 < 0.01D) {
+-            this.m_20256_(this.m_20184_().m_82520_(vec32.f_82479_ * 0.1D, 0.0D, vec32.f_82481_ * 0.1D));
++            net.minecraftforge.event.entity.MinecartEvent.Steered event = new net.minecraftforge.event.entity.MinecartEvent.Steered(this, (Player) entity, vec32.f_82479_ * 0.1D, vec32.f_82481_ * 0.1D);
++            net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(event);
++            this.m_20256_(this.m_20184_().m_82520_(event.getX(), 0.0D, event.getX()));
```
ChampionAsh5357: Gets the same coordinate for both directions instead of one for each.

patches/minecraft/net/minecraft/world/entity/vehicle/AbstractMinecart.java.patch
```diff
@@ -181,6 +197,15 @@
           Vec3 vec36 = this.m_20184_();
           double d27 = vec36.m_165924_();
           if (d27 > 0.01D) {
+@@ -537,6 +_,8 @@
+          }
+       }
+ 
++      net.minecraftforge.event.entity.MinecartEvent.PostMove postMoveEvent = new net.minecraftforge.event.entity.MinecartEvent.PostMove(this, p_38156_, p_38157_);
++      net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(postMoveEvent);
```
ChampionAsh5357: Can be condensed to one line.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -133,10 +133,7 @@
 import net.minecraftforge.event.ForgeEventFactory;
 import net.minecraftforge.event.ItemAttributeModifierEvent;
 import net.minecraftforge.event.ServerChatEvent;
-import net.minecraftforge.event.entity.EntityAttributeCreationEvent;
-import net.minecraftforge.event.entity.EntityAttributeModificationEvent;
-import net.minecraftforge.event.entity.EntityEvent;
-import net.minecraftforge.event.entity.EntityTravelToDimensionEvent;
+import net.minecraftforge.event.entity.*;
```
ChampionAsh5357: No wildcard imports.

src/main/java/net/minecraftforge/event/entity/MinecartEvent.java
```diff
@@ -0,0 +1,135 @@
...
+
+import net.minecraft.core.BlockPos;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.entity.vehicle.AbstractMinecart;
+import net.minecraft.world.level.block.state.BlockState;
+import net.minecraftforge.eventbus.api.Cancelable;
+import net.minecraftforge.eventbus.api.Event;
+
+public class MinecartEvent extends Event {
```
ChampionAsh5357: License header missing, brackets on new lines, documentation.

src/main/java/net/minecraftforge/event/entity/MinecartEvent.java
```diff
@@ -0,0 +1,135 @@
...
+            this.pushZ = pushZ;
+        }
+
+        public Entity getPusher() {
+            return pusher;
+        }
+
+    }
+
+    @net.minecraftforge.eventbus.api.Cancelable
```
ChampionAsh5357: Just import, this is a forge class.

src/test/java/net/minecraftforge/debug/entity/MinecartLinkTest.java
```diff
@@ -0,0 +1,427 @@
...
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+
+@Mod("minecart_link_test")
+public class MinecartLinkTest {
```
ChampionAsh5357: Missing license header, brackets on new lines, and remove wildcard imports.

ChampionAsh5357 (changes requested):
I'm not sure how I feel about this particular event set since its called three events on almost every tick for any minecart present. Though, I currently cannot think of another way to do so.

src/main/java/net/minecraftforge/event/entity/MinecartEvent.java
rehwinkel: I did it that way because I saw other forge events do it that way, e.g.:
```java
@net.minecraftforge.eventbus.api.Cancelable
public class EntityJoinWorldEvent extends EntityEvent { ... }
```

ChampionAsh5357: I don't really know why other Forge events did it that way, but in any case you should just import it.

src/main/java/net/minecraftforge/event/entity/MinecartEvent.java
```diff
@@ -0,0 +1,194 @@
...
+        {
+            return railState;
+        }
+
+    }
+
+    /**
+     * This event is called at the end of {@link AbstractMinecart#moveAlongTrack(BlockPos, BlockState)} to apply changes after movement has commenced.
+     */
+    @net.minecraftforge.eventbus.api.Cancelable
```
ChampionAsh5357: Import.

src/test/java/net/minecraftforge/debug/entity/MinecartLinkTest.java
```diff
@@ -0,0 +1,497 @@
...
+                if (cartB != null) {
+                    adjustVelocity(cart, cartB);
+                }
+            }
+        }
+
+        protected static void adjustVelocity(AbstractMinecart cart1, AbstractMinecart cart2)
+        {
+            double dist = cart1.position().distanceTo(cart2.position());
+            if (dist > 8F) {
```
ChampionAsh5357: Opening brace on new line.

ChampionAsh5357 (changes requested):
I'll give this a proper testing when I can, but currently just fixing up formatting and clarity.

Shadows-of-Fire:
We had hooks like these previously, and they were removed due to their limited usage and high cost.  I'm not sure readding them in any form is a good idea.

rehwinkel:
The main use for this would be minecart linking the way railcraft does it, which was my primary motivation for making this PR. I agree, the use cases are quite limited, however I am not sure how one would best achieve functionality like that without an event like this (without replacing vanilla minecarts entirely, which is very bad for compatbility).

sciwhiz12:
Hello! @rehwinkel, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable. If we do not hear back from you after some time, this PR will be closed for inactivity.

PR #8173, opened 2021-10-26
Json-based modifier system to manipulate vanilla StructureTemplatePools

rehwinkel:
This PR adds structure pool modifiers, which are run when the worldgen code asks for the list of elements of the pool. Then, a modifier can add/remove structure pool elements there. This makes it possible to dynamically add custom houses to villages (and more).


rehwinkel:
Related to #8059

ShroomAgent27 (changes requested):
Here's a few formatting changes that need to happen:

* Braces always go on a new line.
* Do not use wildcard imports, like in `src/main/java/net/minecraftforge/common/world/StructurePoolModifierManager.java `
* Included license headers in all new files. You can easily do this by running `./gradlew updateLicenses`

patches/minecraft/net/minecraft/world/level/levelgen/feature/structures/StructureTemplatePool.java.patch
```diff
@@ -1,7 +1,44 @@
...
+-      return ImmutableList.copyOf(ObjectArrays.shuffle(this.f_69250_.toArray(new StructurePoolElement[0]), p_69277_));
++      return ImmutableList.copyOf(ObjectArrays.shuffle(this.getTemplates().toArray(new StructurePoolElement[0]), p_69277_));
+    }
+ 
+    public ResourceLocation m_69275_() {
+       return this.f_69248_;
+    }
+ 
+-   public int m_69278_() {
+-      return this.f_69250_.size();
```
sciwhiz12: Please do not remove existing methods in vanilla classes.

patches/minecraft/net/minecraft/world/level/levelgen/feature/structures/StructureTemplatePool.java.patch
```diff
@@ -1,7 +1,44 @@
...
-       return this.f_69250_.size();
+@@ -74,9 +_,15 @@
+       this.f_69251_ = p_69260_;
+    }
+ 
++   public List<StructurePoolElement> getTemplates() {
++      List<StructurePoolElement> templatesCopy = Lists.newArrayList(f_69250_);
++      net.minecraftforge.common.ForgeHooks.modifyStructurePool(f_69248_, templatesCopy);
++      return templatesCopy;
++   }
```
sciwhiz12, line 12: Is there a specific reason why this method is positioned here? If not, please move this method to beside the `isNotEmpty` method, and bound the area of those two methods with `/* *************** FORGE [START/END] *************** */` markers.

src/main/java/net/minecraftforge/common/world/StructurePoolModifierSerializer.java
```diff
@@ -0,0 +1,87 @@
...
+package net.minecraftforge.common.world;
+
+import com.google.gson.JsonObject;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraftforge.registries.GameData;
+import net.minecraftforge.registries.IForgeRegistryEntry;
+
+import javax.annotation.Nullable;
+
+public abstract class StructurePoolModifierSerializer<T extends IStructurePoolModifier> implements IForgeRegistryEntry<StructurePoolModifierSerializer<?>>
```
sciwhiz12: Why not extend `ForgeRegistryEntry<StructurePoolModifierSerializer<?>>`?

src/main/java/net/minecraftforge/registries/ForgeRegistries.java
```diff
@@ -153,6 +155,7 @@
         //Forge
         public static final ResourceKey<Registry<DataSerializerEntry>> DATA_SERIALIZERS = key("data_serializers");
         public static final ResourceKey<Registry<GlobalLootModifierSerializer<?>>> LOOT_MODIFIER_SERIALIZERS = key("forge:loot_modifier_serializers");
+        public static final ResourceKey<Registry<StructurePoolModifierSerializer<?>>> POOL_MODIFIER_SERIALIZERS = key("forge:pool_modifier_serializers");
```
sciwhiz12: Instead of `POOL_MODIFIER_SERIALIZERS` (which could be confused with loot _pools_), why not something a bit more long and descriptive like `STRUCTURE_MODIFIER_SERIALIZERS` or `STRUCTURE_POOL_MODIFIER_SERIALIZERS`?

sciwhiz12 (changes requested):

patches/minecraft/net/minecraft/world/level/levelgen/feature/structures/StructureTemplatePool.java.patch
```diff
@@ -1,5 +1,48 @@
...
++   /* *************** FORGE START ************ */
++   public List<StructurePoolElement> getTemplates() {
++      List<StructurePoolElement> templatesCopy = Lists.newArrayList(f_69250_);
++      net.minecraftforge.common.ForgeHooks.modifyStructurePool(f_69248_, templatesCopy);
++      return templatesCopy;
++   }
++
++   public boolean isNotEmpty() {
++       if (!this.f_69250_.isEmpty()) return true;
++       return !this.getTemplates().isEmpty();
```
ChampionAsh5357: If we replace all reference of `f_69250_` with `#getTemplates`, why would we ever check if the field itself is empty? It would only matter what `#getTemplates` returns as that is what is used.

ChampionAsh5357 (changes requested):
I probably want a few more triage members to test this, but I think this is good otherwise.

patches/minecraft/net/minecraft/world/level/levelgen/feature/structures/StructureTemplatePool.java.patch
rehwinkel: `getTemplates` needs to call all the modifiers, which takes time. If the list itself isn't empty, I would expect the modified list to also not be empty. Therefore, I can skip running the modifiers by checking the original list individually first.
I will admit, when writing this, I figured modifiers would be used for adding only. If someone should decide to remove things using the modifiers, this code would produce incorrect results. Considering that, I think its better to remove that line.

dpeter99:
Hi!
This PR still needs more testing and making sure it covers a large enougth set of use cases. With the internal deadline for 1.17 approaching, we've decided to close this PR.
It's not an outright rejection - this is something that would be useful for modding. We would like to see this in 1.18, but for that, the PR would need to be retargetted. This involves closing the PR anyway.

If you'd like to port this change to 1.18, let us know, and make sure you open up a PR when the new version is out.

PR #8174, opened 2021-10-26
Create worldgen.json

UnseenAcademical:
I propose a tag that can be used to query for "raw" resources. This tag would contain only resources/blocks in the world, not derivative ones (so bamboo yes, but sticks no). There will be a set of sub-tags that specify which dimension the resources are generated in. Mods that add dimensions can add their own sub-tags.

This will primarily be used to get a list of "all resources generated in the world" which will help with mod integration.

### DIFF
--- a/src/generated/resources/data/forge/tags/blocks/worldgen.json
+++ b/src/generated/resources/data/forge/tags/blocks/worldgen.json
@@ -1,0 +1,8 @@
+{
+  "replace": false,
+  "values": [
+    "forge:worldgen/overworld",
+    "forge:worldgen/the_nether",
+    "forge:worldgen/the_end"
+  ]
+}


ChampionAsh5357:
Could you elaborate on the use case? Why would I need to know every block that is generated within a particular level? And if not level, why world? All blocks are always present in all worlds. How would this even be tracked? This PR doesn't keep track of the current vanilla blocks and any mod that doesn't add their own would cause this tag to be desynchronous.

As for the actual data itself, what are you trying to store? It seems you are trying to store some world gen registry object in a block tag. If this was tested, the game would surely be unable to load the tag since those blocks do not exist.

UnseenAcademical:
Sorry, I'll attempt to clarify.

I'm working with a number of mods that each add non-vanilla world generated blocks (such as Astral Sorcery's marble and collector crystals). I would like to be able to write integrations with mods such as OneBlockPlus or Dimensional Dungeons that include giving mod-specific resources. However, the current lucky block mechanics I've seen give a random item from the entire item database, leading to weird "skipping" problems as you are given items that don't match your current position in the tech tree of a mod.

Theoretically, I could make something that only pulled blocks and not items, but that still would lead to problems when it would pull blocks that aren't world present (such as storage blocks) or wouldn't pull blocks that are present but not in the blocks tag (such as AS collector crystals).

While its possible to write out a list of every block per mod that would be findable "in the world" this list would need to be kept manually updated for each integration with each mod. Since Forge's tag system plus vanilla allows for flexible querying, this seemed like the simplest / most resilient way to meet the requirement of "query blocks from this mod OR from this dimension, where they are generated but not necessarily in the blocks tag"

It seemed like a good idea to create this tag at the Forge level so that everyone will have a central tag to use for this purpose, standardizing across mods. If there's a better way to go about this, I'd be happy to hear alternatives. My post on the Forge forums didn't receive any comment after several weeks, so I assumed that it was an unobjectionable addition.

Not keeping track of vanilla blocks in this PR is my own mistake.

ChampionAsh5357:
Correct me if I'm wrong, you want to standardize a tag such that you can select some random item that correctly meshes with the ability of other mods? If that's the case, that brings up two issues. The first is that the use case seems rather limited to be added into forge as a standardized tag. You can create your mod to have a custom tag in the forge namespace without the need for a PR. If you wanted other compatibility with mods, you would simply have a datapack that added the associated items from the other mods to the tag. Secondly, this would prevent "tech trees" from unlocking new items within the tag itself making them reload the entire datapack every time one is unlocked. This would probably be better put as an API within your own mod where mods can give yours a supplier containing items that can be obtained at any time which could be invalidated either via a callback or computated on the fly if too complex.

All in all, I don't see a reason why the tag should be in Forge itself, rather just a tag in your mod under the forge namespace to which can be appended. Is there another use case besides random item pulling that you may be able to provide to justify this?

UnseenAcademical:
No other use case in particular, the API suggestion makes a lot more sense anyway. Thanks for the help! Closing this PR.

PR #8176, opened 2021-10-28
Prospective BackgroundScanHandler Fix

DaemonUmbra:
My attempt at fixing a race condition in BackgroundScanHandler


GizmoTheMoonPig:
Hey there, I've been having the "failed to complete mod scan" issue for a while now, and I was told to test this PR probably close to 2 weeks ago. I've been running it and I haven't once encountered the issue since then, so I think this works just fine. Just wanted to leave my feedback in case anyone was skeptical about if it does anything or not.

Shadows-of-Fire:
Given the overall difficulty at reproducing, that should be good enough evidence to send it along

LexManos:
I did a few local tests with some simulated lag, and it seems to work fine. At the very least I don't see it opening any extra issues. And none of the people who reported this issue have any bad words for this, so accepting.

PR #8178, opened 2021-10-30
Add copper and deepslate ores to ore tags

uecasm:


sciwhiz12:
Duplicate of #7891

uecasm:
Weird, that didn't come up in github search when I tried.  Agreed, that one's more comprehensive.

PR #8179, opened 2021-10-30
[1.17.x] make ScrollPanel better usable by modders

Minecraftschurli:
This PR adds:
- Parameters to the constructor of ScrollPanel for 
  - border
  - bar width
  - background color (from, to)
  - bar color (background, handle, handle border)
- A drawBackground method that takes the PoseStack, the Tesselator and the partialTicks and by default renders the background with the colors set in the constructor or the dirt background if the gui is outside of a world context (i.e. the main menu)

It also moves from the GL11 calls to calls to the methods provided by RenderSystem.

For any requested changes/additions please ping `@Minecraftschurli` in `#issues-and-prs` on the Forge Discord

### DIFF
--- a/src/main/java/net/minecraftforge/client/gui/ScrollPanel.java
+++ b/src/main/java/net/minecraftforge/client/gui/ScrollPanel.java
@@ -22,7 +22,11 @@ package net.minecraftforge.client.gui;
 import java.util.Collections;
 import java.util.List;
 
-import com.mojang.blaze3d.vertex.*;
+import com.mojang.blaze3d.vertex.BufferBuilder;
+import com.mojang.blaze3d.vertex.DefaultVertexFormat;
+import com.mojang.blaze3d.vertex.PoseStack;
+import com.mojang.blaze3d.vertex.Tesselator;
+import com.mojang.blaze3d.vertex.VertexFormat;
 import net.minecraft.client.gui.narration.NarratableEntry;
 import net.minecraft.client.renderer.GameRenderer;
 
--- a/src/main/java/net/minecraftforge/client/gui/ScrollPanel.java
+++ b/src/main/java/net/minecraftforge/client/gui/ScrollPanel.java
@@ -54,13 +54,16 @@ public abstract class ScrollPanel extends AbstractContainerEventHandler implemen
     private final int barLeft;
     private final int bgColorFrom;
     private final int bgColorTo;
+    private final int barBgColor;
+    private final int barColor;
+    private final int barBorderColor;
 
     public ScrollPanel(Minecraft client, int width, int height, int top, int left)
     {
-        this(client, width, height, top, left, 4, 0xC0101010, 0xD0101010);
+        this(client, width, height, top, left, 4, 0xC0101010, 0xD0101010, 0xFF000000, 0xFF808080, 0xFFC0C0C0);
     }
 
-    public ScrollPanel(Minecraft client, int width, int height, int top, int left, int border, int bgColorFrom, int bgColorTo)
+    public ScrollPanel(Minecraft client, int width, int height, int top, int left, int border, int bgColorFrom, int bgColorTo, int barBgColor, int barColor, int barBorderColor)
     {
         this.client = client;
         this.width = width;
@@ -73,6 +76,9 @@ public abstract class ScrollPanel extends AbstractContainerEventHandler implemen
         this.border = border;
         this.bgColorFrom = bgColorFrom;
         this.bgColorTo = bgColorTo;
+        this.barBgColor = barBgColor;
+        this.barColor = barColor;
+        this.barBorderColor = barBorderColor;
     }
 
     protected abstract int getContentHeight();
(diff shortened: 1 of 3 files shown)


src/main/java/net/minecraftforge/client/gui/ScrollPanel.java
```diff
@@ -48,12 +48,12 @@
...
     protected float scrollDistance;
     protected boolean captureMouse = true;
-    protected final int border = 4;
+    protected final int border;
 
     private final int barWidth = 6;
     private final int barLeft;
 
-    public ScrollPanel(Minecraft client, int width, int height, int top, int left)
+    public ScrollPanel(Minecraft client, int width, int height, int top, int left, int border)
```
sciwhiz12, line 87: This is a breaking change, as any mod which links to the constructor without the new parameter will fail. Create an overload of this constructor without the new `border` parmaeter which passes in a default value for it (such as the previously present value, `4`).

sciwhiz12 (changes requested):

bl4ckscor3:
It might be worth also making the color customizable. Similar to this <https://github.com/Geforce132/SecurityCraft/blob/8b871156adf5f453d6493587ee4af2248de49d29/src/main/java/net/geforcemods/securitycraft/screen/components/ColorableScrollPanel.java>

src/main/java/net/minecraftforge/client/gui/ScrollPanel.java
```diff
@@ -48,12 +47,22 @@
...
+    private final int barBgColor;
+    private final int barColor;
+    private final int barBorderColor;
 
     public ScrollPanel(Minecraft client, int width, int height, int top, int left)
+    {
+        this(client, width, height, top, left, 4, 6, 0xC0101010, 0xD0101010, 0xFF000000, 0xFF808080, 0xFFC0C0C0);
+    }
+
+    public ScrollPanel(Minecraft client, int width, int height, int top, int left, int border, int barWidth, int bgColorFrom, int bgColorTo, int barBgColor, int barColor, int barBorderColor)
```
ChampionAsh5357, line 153: This constructor seems a bit too redundant if seven new parameters are being added. You may want to create other constructors that just modify a single section (such as change background color, change bar info, etc.)

src/main/java/net/minecraftforge/client/gui/ScrollPanel.java
```diff
@@ -63,12 +72,41 @@ public ScrollPanel(Minecraft client, int width, int height, int top, int left)
...
+        this.barBgColor = barBgColor;
+        this.barColor = barColor;
+        this.barBorderColor = barBorderColor;
     }
 
     protected abstract int getContentHeight();
 
     protected void drawBackground() {}
 
+    protected void drawBackground(PoseStack matrix, Tesselator tess, float partialTicks)
```
ChampionAsh5357: These two are badly named now. The one with no parameters is the first thing to be drawn while the second just renders the actual background. It would probably be good to mark one of these to have their name changed in 1.18 to better exemplify what they are supposed to be used for.

src/main/java/net/minecraftforge/client/gui/ScrollPanel.java
```diff
@@ -63,12 +72,41 @@ public ScrollPanel(Minecraft client, int width, int height, int top, int left)
...
+            worldr.vertex(this.right, this.bottom, 0.0D).uv(this.right / texScale, (this.bottom + (int)this.scrollDistance) / texScale).color(0x20, 0x20, 0x20, 0xFF).endVertex();
+            worldr.vertex(this.right, this.top,    0.0D).uv(this.right / texScale, (this.top    + (int)this.scrollDistance) / texScale).color(0x20, 0x20, 0x20, 0xFF).endVertex();
+            worldr.vertex(this.left,  this.top,    0.0D).uv(this.left  / texScale, (this.top    + (int)this.scrollDistance) / texScale).color(0x20, 0x20, 0x20, 0xFF).endVertex();
+            tess.end();
+        }
+    }
+
     /**
      * Draw anything special on the screen. GL_SCISSOR is enabled for anything that
      * is rendered outside of the view box. Do not mess with SCISSOR unless you support this.
```
ChampionAsh5357: Outdated javadoc.

ChampionAsh5357 (changes requested):
Since you're already adding new information to this class, it would probably be good to javadoc it properly as well to make it more understandable on how to properly use. That falls under the scope of the PR imo.

src/main/java/net/minecraftforge/client/gui/ScrollPanel.java
Minecraftschurli: I think the parameterless one is not useful at all so I would either 
- deprecate it or
- make a comment to rename it and add at least the PoseStack

src/main/java/net/minecraftforge/client/gui/ScrollPanel.java
Minecraftschurli: no because I just replaced the GL11 things with the RenderSystem wrappers (it still does the same thing)

src/main/java/net/minecraftforge/client/gui/ScrollPanel.java
ChampionAsh5357: Unsure since it is drawn before the scissor is active. So it probably is better to comment and then rename instead.

src/main/java/net/minecraftforge/client/gui/ScrollPanel.java
ChampionAsh5357: It still is outdated. GL_SCISSOR is internally called, but we don't want people to reference GL calls directly so it should mention the method at least instead.

src/main/java/net/minecraftforge/client/gui/ScrollPanel.java
```diff
@@ -63,17 +160,45 @@ public ScrollPanel(Minecraft client, int width, int height, int top, int left)
...
+        this.barColor = barColor;
+        this.barBorderColor = barBorderColor;
     }
 
     protected abstract int getContentHeight();
 
+    // TODO: 1.18 rename and add PoseStack parameter
     protected void drawBackground() {}
 
+    protected void drawBackground(PoseStack matrix, Tesselator tess, float partialTicks)
```
ChampionAsh5357, line 180: Needs javadoc still.

src/main/java/net/minecraftforge/client/gui/ScrollPanel.java
```diff
@@ -147,7 +273,8 @@ public boolean mouseClicked(double mouseX, double mouseY, int button) {
     }
 
     @Override
-    public boolean mouseReleased(double p_mouseReleased_1_, double p_mouseReleased_3_, int p_mouseReleased_5_) {
+    public boolean mouseReleased(double p_mouseReleased_1_, double p_mouseReleased_3_, int p_mouseReleased_5_)
```
ChampionAsh5357: Rename parameters to something useful.

ChampionAsh5357 (changes requested):
Nothing breaks from this so it should be fine, just another javadoc and parameter change and it'll be good (unless I catch something after this again, sorry).

ChampionAsh5357 (approved):

PR #8183, opened 2021-11-01
[1.17.x] Custom advancement rewards

Minecraftschurli:
This PR adds custom advancement rewards.

Benefits to using an mcfunction:
- not limited to the mcfunction format
- not limited by the availability of commands
- reusable and configurable (if you have multiple values for multiple advancements you don't need a function for each)

Test mod:
use an ender eye and you get a message in chat


src/test/java/net/minecraftforge/debug/block/FullPotsAccessorDemo.java
```diff
@@ -1,3 +1,22 @@
...
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
```
rehwinkel: Does this belong in this PR?

src/test/java/net/minecraftforge/debug/CustomAdvancementRewardsTest.java
```diff
@@ -0,0 +1,69 @@
...
+import net.minecraftforge.common.advancements.ICustomAdvancementReward;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.fmllegacy.RegistryObject;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+
+@Mod("custom_advancement_rewards_test")
+public class CustomAdvancementRewardsTest {
+    private static final DeferredRegister<ICustomAdvancementReward.Serializer<?>> DEFERRED_REGISTER = DeferredRegister.create(ForgeRegistries.CUSTOM_ADVANCEMENT_REWARD_SERIALIZERS, "custom_advancement_rewards_test");
```
rehwinkel: I wouldn't call this field DEFERRED_REGISTER, call it CUSTOM_ADVANCEMENT_REWARD_SERIALIZERS or something.

patches/minecraft/net/minecraft/advancements/AdvancementRewards.java.patch
```diff
@@ -1,5 +1,28 @@
...
++
++   public AdvancementRewards(int experience, ResourceLocation[] loot, ResourceLocation[] recipes, CommandFunction.CacheableFunction function, net.minecraftforge.common.advancements.ICustomAdvancementReward[] customRewards) {
++      this.f_9979_ = experience;
++      this.f_9980_ = loot;
++      this.f_9981_ = recipes;
++      this.f_9982_ = function;
++      this.customRewards = customRewards;
++   }
+ 
+    public AdvancementRewards(int p_9985_, ResourceLocation[] p_9986_, ResourceLocation[] p_9987_, CommandFunction.CacheableFunction p_9988_) {
```
rehwinkel, line 9: The old constructor should probably be deprecated

patches/minecraft/net/minecraft/advancements/AdvancementRewards.java.patch
```diff
@@ -9,3 +32,57 @@
        boolean flag = false;
  
        for(ResourceLocation resourcelocation : this.f_9980_) {
+@@ -68,7 +_,7 @@
+       if (this.f_9981_.length > 0) {
+          p_9990_.m_7902_(this.f_9981_);
+       }
+-
++      net.minecraftforge.common.advancements.CustomAdvancementRewards.grant(p_9990_, this.customRewards);
```
rehwinkel: More context than the player might be nice, such as the lootcontext.

src/main/java/net/minecraftforge/common/advancements/CustomAdvancementRewards.java
```diff
@@ -0,0 +1,83 @@
...
+import com.google.gson.JsonParseException;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.util.GsonHelper;
+import net.minecraftforge.registries.ForgeRegistries;
+
+/**
+ * Class holding the hooks and registry for custom advancement rewards
+ */
+public final class CustomAdvancementRewards {
```
rehwinkel: Since these are hooks, maybe this should be called AdvancementRewardHooks (?)

src/main/java/net/minecraftforge/common/advancements/CustomAdvancementRewards.java
```diff
@@ -0,0 +1,83 @@
...
+     * Hook for AdvancementRewards#serialize
+     */
+    public static void serialize(JsonObject jsonObject, ICustomAdvancementReward[] customRewards)
+    {
+        if (customRewards.length == 0) return;
+        JsonArray arr = new JsonArray();
+        for (ICustomAdvancementReward customReward : customRewards)
+        {
+            @SuppressWarnings("unchecked")
+            ICustomAdvancementReward.Serializer<ICustomAdvancementReward> serializer = (ICustomAdvancementReward.Serializer<ICustomAdvancementReward>) customReward.getSerializer();
```
rehwinkel: Remove this cast by making the signature of getSerializer: `<T extends ICustomAdvancementReward> Serializer<T> getSerializer();`

src/main/java/net/minecraftforge/common/advancements/ICustomAdvancementReward.java
```diff
@@ -0,0 +1,51 @@
...
+     *
+     * @param player the player the advancement is granted to
+     */
+    void grant(ServerPlayer player);
+
+    /**
+     * Get the serializer for this custom advancement reward
+     * @return the serializer for this custom advancement reward
+     */
+    Serializer<?> getSerializer();
```
rehwinkel: Change this signature as noted in my other comment.

src/test/java/net/minecraftforge/debug/CustomAdvancementRewardsTest.java
```diff
@@ -0,0 +1,69 @@
...
+        public void grant(ServerPlayer player) {
+            player.sendMessage(new TextComponent(this.message()), Util.NIL_UUID);
+        }
+
+        @Override
+        public Serializer getSerializer() {
+            return SEND_MESSAGE.get();
+        }
+
+        public static class Serializer extends ICustomAdvancementReward.Serializer<SendMessageReward> {
```
rehwinkel: Braces on newline

src/main/java/net/minecraftforge/common/advancements/CustomAdvancementRewards.java
rehwinkel: Braces on newlines

rehwinkel (changes requested):

patches/minecraft/net/minecraft/advancements/AdvancementRewards.java.patch
ChampionAsh5357: Adding a new line is less patchwork than removing and then adding the line.

patches/minecraft/net/minecraft/advancements/AdvancementRewards.java.patch
```diff
@@ -9,3 +32,57 @@
...
+       MinecraftServer minecraftserver = p_9990_.f_8924_;
+       this.f_9982_.m_78002_(minecraftserver.m_129890_()).ifPresent((p_9996_) -> {
+          minecraftserver.m_129890_().m_136112_(p_9996_, p_9990_.m_20203_().m_81324_().m_81325_(2));
+@@ -107,7 +_,7 @@
+ 
+             jsonobject.add("recipes", jsonarray1);
+          }
+-
++         net.minecraftforge.common.advancements.CustomAdvancementRewards.serialize(jsonobject, this.customRewards);
```
ChampionAsh5357: Adding a new line is less patchwork than removing and then adding the line.

src/main/java/net/minecraftforge/common/advancements/ICustomAdvancementReward.java
```diff
@@ -0,0 +1,51 @@
...
+
+    abstract class Serializer<T extends ICustomAdvancementReward> extends ForgeRegistryEntry<Serializer<?>>
+    {
+        public abstract T deserialize(JsonObject json);
+        public abstract JsonObject serialize(T reward);
+    }
```
ChampionAsh5357: While I do agree with the concept, I think we should be moving towards codecs for these types of solutions. Though, this is purely an opinion of mine.

src/test/java/net/minecraftforge/debug/CustomAdvancementRewardsTest.java
```diff
@@ -0,0 +1,69 @@
...
+import net.minecraftforge.fmllegacy.RegistryObject;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+
+@Mod("custom_advancement_rewards_test")
+public class CustomAdvancementRewardsTest {
+    private static final DeferredRegister<ICustomAdvancementReward.Serializer<?>> DEFERRED_REGISTER = DeferredRegister.create(ForgeRegistries.CUSTOM_ADVANCEMENT_REWARD_SERIALIZERS, "custom_advancement_rewards_test");
+    private static final RegistryObject<SendMessageReward.Serializer> SEND_MESSAGE = DEFERRED_REGISTER.register("send_message", SendMessageReward.Serializer::new);
+
+    public CustomAdvancementRewardsTest() {
```
ChampionAsh5357: New line brace.

src/test/java/net/minecraftforge/debug/CustomAdvancementRewardsTest.java
```diff
@@ -0,0 +1,69 @@
...
+import net.minecraft.util.GsonHelper;
+import net.minecraftforge.common.advancements.ICustomAdvancementReward;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.fmllegacy.RegistryObject;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+
+@Mod("custom_advancement_rewards_test")
+public class CustomAdvancementRewardsTest {
```
ChampionAsh5357: New line brace.

src/test/java/net/minecraftforge/debug/CustomAdvancementRewardsTest.java
```diff
@@ -0,0 +1,69 @@
...
+    private static final DeferredRegister<ICustomAdvancementReward.Serializer<?>> DEFERRED_REGISTER = DeferredRegister.create(ForgeRegistries.CUSTOM_ADVANCEMENT_REWARD_SERIALIZERS, "custom_advancement_rewards_test");
+    private static final RegistryObject<SendMessageReward.Serializer> SEND_MESSAGE = DEFERRED_REGISTER.register("send_message", SendMessageReward.Serializer::new);
+
+    public CustomAdvancementRewardsTest() {
+        DEFERRED_REGISTER.register(FMLJavaModLoadingContext.get().getModEventBus());
+    }
+
+    public record SendMessageReward(String message) implements ICustomAdvancementReward {
```
ChampionAsh5357: New line brace.

src/test/java/net/minecraftforge/debug/CustomAdvancementRewardsTest.java
```diff
@@ -0,0 +1,69 @@
...
+        @Override
+        public void grant(ServerPlayer player) {
```
ChampionAsh5357: New line brace.

src/test/java/net/minecraftforge/debug/CustomAdvancementRewardsTest.java
```diff
@@ -0,0 +1,69 @@
...
+            player.sendMessage(new TextComponent(this.message()), Util.NIL_UUID);
+        }
+
+        @Override
+        public Serializer getSerializer() {
```
ChampionAsh5357: New line brace.

src/test/java/net/minecraftforge/debug/CustomAdvancementRewardsTest.java
```diff
@@ -0,0 +1,69 @@
...
+            return SEND_MESSAGE.get();
+        }
+
+        public static class Serializer extends ICustomAdvancementReward.Serializer<SendMessageReward> {
+
+            @Override
+            public SendMessageReward deserialize(JsonObject json) {
```
ChampionAsh5357: New line brace.

src/test/java/net/minecraftforge/debug/CustomAdvancementRewardsTest.java
```diff
@@ -0,0 +1,69 @@
...
+                return new SendMessageReward(GsonHelper.getAsString(json, "message"));
+            }
+
+            @Override
+            public JsonObject serialize(SendMessageReward reward) {
```
ChampionAsh5357: New line brace.

src/test/java/net/minecraftforge/debug/block/FullPotsAccessorDemo.java
ChampionAsh5357: This is fine provided the the change didn't get merged in with another. The PR is small enough imo for this to be here.

src/test/resources/data/custom_advancement_rewards_test/advancements/test/root.json
```diff
@@ -0,0 +1,37 @@
+{
```
ChampionAsh5357: As AdvancementProviders now work, you may want to make this a generated test to make sure the serialization works as expected.

ChampionAsh5357 (changes requested):
I would update the description of the PR to probably include why this system should be included over creating an mcfunction that is executed at permission level 2 instead. While this seems useful, most of the behavior could be defined via a single mcfunction, so I do believe there needs to be some convincing here.

src/main/java/net/minecraftforge/common/advancements/ICustomAdvancementReward.java
Minecraftschurli: I would prefer Codecs too but the problem is a Codec can't be registered in a Registry.

ChampionAsh5357: True, but you can wrap the codec in another object. Since there is an RFC to remove the `IForgeRegistryEntry` limitation in the future, I would say that would cover that issue and provide for a swift transition out of the wrapping object.

Minecraftschurli: So I should just convert it to a class holding a codec

ChampionAsh5357: That's my opinion on the matter on it yeah. Not sure if anyone else has a differing opinion though, so it may be worthwhile to hold off on it for a bit until someone else chimes in on the idea.

src/test/java/net/minecraftforge/debug/CustomAdvancementRewardsTest.java
Minecraftschurli: I think as it is just a test mod the name is irrelevant

patches/minecraft/net/minecraft/advancements/AdvancementRewards.java.patch
Minecraftschurli: Idk as it still works just the same as passing an empty array for the custom rewards

patches/minecraft/net/minecraft/advancements/AdvancementRewards.java.patch
Minecraftschurli: I don't think this is needed as that information comes from the player anyway

src/main/java/net/minecraftforge/common/advancements/CustomAdvancementRewardHooks.java
```diff
@@ -0,0 +1,82 @@
...
+import com.google.gson.JsonParseException;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.util.GsonHelper;
+import net.minecraftforge.registries.ForgeRegistries;
+
+/**
+ * Class holding the hooks for custom advancement rewards
+ */
+public final class CustomAdvancementRewardHooks {
```
ChampionAsh5357: Brace on new line.

patches/minecraft/net/minecraft/advancements/AdvancementRewards.java.patch
```diff
@@ -9,3 +32,55 @@
...
+-      return new AdvancementRewards(i, aresourcelocation, aresourcelocation1, commandfunction$cacheablefunction);
++      net.minecraftforge.common.advancements.ICustomAdvancementReward[] customRewards = net.minecraftforge.common.advancements.CustomAdvancementRewardHooks.deserialize(p_9992_);
++      return new AdvancementRewards(i, aresourcelocation, aresourcelocation1, commandfunction$cacheablefunction, customRewards);
+    }
+ 
+    public static class Builder {
+       private int f_9999_;
+       private final List<ResourceLocation> f_10000_ = Lists.newArrayList();
+       private final List<ResourceLocation> f_10001_ = Lists.newArrayList();
++      private final List<net.minecraftforge.common.advancements.ICustomAdvancementReward> customRewards = Lists.newArrayList();
```
ChampionAsh5357: `Lists#newArrayList` is considered deprecated by guava as of Java 7+. Use the `ArrayList` constructor instead.

ChampionAsh5357 (changes requested):
Just a couple more syntax and usage changes.

patches/minecraft/net/minecraft/advancements/AdvancementRewards.java.patch
rehwinkel: I was thinking for convenience, but idk how often you'd even need the lootcontext.

src/main/java/net/minecraftforge/common/advancements/ICustomAdvancementReward.java
sciwhiz12: The status quo is that these kinds of customizable data-driven objects have serializers which are registered with an existing Forge registry. While it may be worth looking into the matter for possible changes across all similar objects in the future, I'm for keeping the status quo on this for now.

ChampionAsh5357 (approved):

patches/minecraft/net/minecraft/advancements/AdvancementRewards.java.patch
pupnewfster: This patch should more or less be:
```
+ this(p_9985_, p_9986_, p_9987_, p_9988_, new net.minecraftforge.common.advancements.ICustomAdvancementReward[0]);
+ }
+ public AdvancementRewards(int p_9985_, ResourceLocation[] p_9986_, ResourceLocation[] p_9987_, CommandFunction.CacheableFunction p_9988_, net.minecraftforge.common.advancements.ICustomAdvancementReward[] customRewards) {
```
And then add the new constructor to forge's exc file. Though thinking about it, it may as well be varargs for the rewards as well to make calling the constructor a bit easier?

src/main/java/net/minecraftforge/common/advancements/ICustomAdvancementReward.java
```diff
@@ -0,0 +1,51 @@
...
+package net.minecraftforge.common.advancements;
+
+import com.google.gson.JsonObject;
+import net.minecraft.server.level.ServerPlayer;
+import net.minecraftforge.registries.ForgeRegistryEntry;
+
+/**
+ * Interface for a custom advancement reward
+ */
+public interface ICustomAdvancementReward
```
pupnewfster: Basically anything being added to the game is "custom" so this name (and other references to this system) would make a bit more sense as just `IAdvancementReward`.

pupnewfster (commented):
Couple comments, but also what does this system expose that isn't currently possible by using vanilla's function system to implement the custom reward via that?

Minecraftschurli:
@pupnewfster 
1. The function system is limited to executing commands
2. You need a function for each advancement if you want to use different values

LexManos:
Commands can do anything, and arguments can modify functions/commands.
The problem I have is with us changing vanilla data formats for no compelling reason.
On top of you creating an entirely new modder facing package for one class, because the hooks could be done in either extends or ForgeHooks.

Minecraftschurli:
the package thing can be changed but afaik functions cant take arguments from the advancement json. Also I am not changing the format as the PR is just adding an optional field

LexManos:
`change (v) make (someone or something) different; alter or modify.`
Says nothing about it having to be a removal, the point is this *change* causes future compatibility issues that we have to support in perpetuity. 
And even if functions can not take arguments, you're using data generators which means you can easily make as many functions as you want.

I am not seeing the benefit of this system reach anywhere near the cost of future maintainability. Perhaps you should approach Mojang and request a proper extension to their data format that support arbitrary rewards. So that when they change the format again, it will be in a way that doesn't break anything.

Minecraftschurli:
I don't think functions are datagenable
also in some usecases creating a command just for the usage in a function is unnecessary work
but I do understand the concern with the possibility of an upstream change

patches/minecraft/net/minecraft/advancements/AdvancementRewards.java.patch
```diff
@@ -1,5 +1,28 @@
...
++      this.rewards = rewards;
++   }
+ 
+    public AdvancementRewards(int p_9985_, ResourceLocation[] p_9986_, ResourceLocation[] p_9987_, CommandFunction.CacheableFunction p_9988_) {
+-      this.f_9979_ = p_9985_;
+-      this.f_9980_ = p_9986_;
+-      this.f_9981_ = p_9987_;
+-      this.f_9982_ = p_9988_;
++      this(p_9985_, p_9986_, p_9987_, p_9988_, new net.minecraftforge.common.advancements.IAdvancementReward[0]);
+    }
```
pupnewfster: I meant make this entire patch be what I showed, not make the patch even bigger... The line below the existing constructor's signature you add a this call, then you end the constructor, then you start the new constructor that you are adding. Use srg names in it for all but the parameter you are adding and then add the new constructor's signature to the exc file so that the params can be properly mapped.

patches/minecraft/net/minecraft/advancements/AdvancementRewards.java.patch
```diff
@@ -9,3 +32,55 @@
...
+       private final List<ResourceLocation> f_10001_ = Lists.newArrayList();
++      private final List<net.minecraftforge.common.advancements.IAdvancementReward> customRewards = new java.util.ArrayList<>();
+       @Nullable
+       private ResourceLocation f_10002_;
+ 
+@@ -185,8 +_,17 @@
+          return this;
+       }
+ 
++      public static AdvancementRewards.Builder custom(net.minecraftforge.common.advancements.IAdvancementReward customReward) {
```
pupnewfster: This should probably just be called `reward` instead of `custom`.

patches/minecraft/net/minecraft/advancements/AdvancementRewards.java.patch
```diff
@@ -9,3 +32,55 @@
...
++         return (new AdvancementRewards.Builder()).addCustomReward(customReward);
++      }
++
++      public AdvancementRewards.Builder addCustomReward(net.minecraftforge.common.advancements.IAdvancementReward customReward) {
```
pupnewfster: `addCustomReward` -> `addReward`

patches/minecraft/net/minecraft/advancements/AdvancementRewards.java.patch
Minecraftschurli: I have been told that signature change in patches is not wanted

pupnewfster: I never said to change the signature?

Minecraftschurli: Then I don't understand what you want me to change

pupnewfster: Exactly what my first comment with the example patch was that you marked as resolved earlier:
```java
//Existing constructor
public AdvancementRewards(int p_9985_, ResourceLocation[] p_9986_, ResourceLocation[] p_9987_, CommandFunction.CacheableFunction p_9988_) {
//added code
   this(p_9985_, p_9986_, p_9987_, p_9988_, new net.minecraftforge.common.advancements.IAdvancementReward[0]);
}
public AdvancementRewards(int p_9985_, ResourceLocation[] p_9986_, ResourceLocation[] p_9987_, CommandFunction.CacheableFunction p_9988_, net.minecraftforge.common.advancements.IAdvancementReward... rewards) {
//Existing constructor body
...
//line you are adding to the body
this.rewards = rewards;
```

TheCurle:
Since the primary function of this PR can be replicated by other means, and it extends some vanilla formats, we've decided that this likely isn't worth the hassle of including.

Mainly, because the additions to vanilla formats mean that we're obligated to support the addition ad infinitum, which may eventually turn out to be more difficult than we bargained for.  

If you want to take another attempt at this with a cleaner and smaller solution, i suggest waiting for us to finish with 1.18.

Apologies.

Issue #8182, opened 2021-11-01
[1.17.1] LootPool serialization doesn't save functions when no conditions are defined


P3pp3rF1y:
I have noticed that functions got removed from my json loot tables when rerunning `runData` in 1.17.1.

Tracked it down to serialization in LootPool

https://github.com/MinecraftForge/MinecraftForge/blob/1.17.x/patches/minecraft/net/minecraft/world/level/storage/loot/LootPool.java.patch#L120-L121

```
         if (!p_79094_.conditions.isEmpty()) {
            jsonobject.add("functions", p_79096_.serialize(p_79094_.functions));
         }
```

It incorrectly checks for whether there are conditions present before saving functions. In the patch it can be seen that vanilla properly checks the functions.
(I may PR the fix for this, but haven't done a forge PR in quite a while and don't have time to figure it out now so logging this to be sure it's fixed at some point)

sciwhiz12:
Yep, a clear bug as it looks at the `f_79024_` (`conditions`) field instead of the `f_79026_` (`functions`) field before serializing the `functions` part of the loot pool JSON.

ChampionAsh5357:
Closed by 8a38416e0fe78f169a24c544dfda4a4ba22aad37

PR #8184, opened 2021-11-02
[1.17.x] Remove Old LootPool Patches

ChampionAsh5357:
This fixes a bug tracked down by #8182 where conditions were used to check whether to serialize functions.

This PR is specifically for removing the patches that are no longer necessary due to the move to GLMs. See sci's comment below for more details.

### DIFF
--- a/patches/minecraft/net/minecraft/world/level/storage/loot/LootPool.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/storage/loot/LootPool.java.patch
@@ -117,7 +117,7 @@
           }
  
 -         if (!ArrayUtils.isEmpty((Object[])p_79094_.f_79026_)) {
-+         if (!p_79094_.f_79024_.isEmpty()) {
++         if (p_79094_.f_79026_.length > 0) {
              jsonobject.add("functions", p_79096_.serialize(p_79094_.f_79026_));
           }
  
--- a/patches/minecraft/net/minecraft/world/level/storage/loot/LootPool.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/storage/loot/LootPool.java.patch
@@ -47,7 +47,7 @@
  
        public LootPool.Builder m_165133_(NumberProvider p_165134_) {
           this.f_79070_ = p_165134_;
-@@ -153,11 +_,21 @@
+@@ -153,11 +_,22 @@
           return this;
        }
  
@@ -56,6 +56,7 @@
 +         return this;
 +      }
 +
++      @Deprecated(forRemoval = true, since = "1.17.1") // FORGE: Use #setBonusRolls instead with UniformGenerator
 +      public LootPool.Builder bonusRolls(float min, float max) {
 +         this.f_79071_ = net.minecraft.world.level.storage.loot.providers.number.UniformGenerator.m_165780_(min, max);
 +         return this;
--- a/patches/minecraft/net/minecraft/world/level/storage/loot/LootPool.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/storage/loot/LootPool.java.patch
@@ -96,7 +96,7 @@
           }
        }
     }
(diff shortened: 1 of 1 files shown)


sciwhiz12 (commented):
I've dug down the `git blame` rabbit hole to figure out why the arrays of loot entries and conditions were changed to be `List`s instead, since nothing which references the fields has any particular use that would not be served by remaining as arrays (all references only read from the array/list). 

It seems to lead to this commit from May 2016: https://github.com/MinecraftForge/MinecraftForge/commit/aaa146c10da9869f59c97105411aa767d4712785. This seems to be the precursor to the modern global loot modifiers mechanism, in which it allows mods to modify loot tables while loading through the `LootTableLoadEvent` (and only in that event). That commit exposed methods to get, insert, and remove loot entries from the loot pool, hence why the arrays were converted into `List`s.

However, Forge has moved on from harcoded modifications through the event to data-driven modification through the global loot modifiers system. This means the previously-existing methods to modify the loot pool were removed, which means the reason for the array fields being converted to `List`s is now gone. I've confirmed in-IDE that nothing accesses these fields to modify their contents, only to read them.

Now, this leads us to a decision which needs to be made: should we continue on with the current PR's changes, which removes the incorrect patch, _or_ should we expand the PR to also revert the changes of the fields, moving them back to being arrays? Keep in mind that all the parts touched by the patch to convert the array fields to `List`s are not accessible to the outside (non-public API), so this would not constitute a breaking change (unless a mod were reflecting into these classes, but in my opinion that's on them since GLMs are available).

ChampionAsh5357:
In my opinion in that case would be to revert all patches. It's unnecessary to have a redundant patch from a system that was previously removed. I do have some reservations when it comes to vanilla code instead of changing private forge code, but if it's considered problematic, it can just be marked as a todo for 1.18.

sciwhiz12 (changes requested):
Since this PR is now targeted to removing old LootPool patches, I'd like to expand it a small bit to deprecate for removal the patched-in `LootPool.Builder#bonusRolls` method, with a comment redirecting users to the `setBonusRolls` method instead (with a note to use `UniformGenerator` as the number provider).

sciwhiz12 (approved):

patches/minecraft/net/minecraft/world/level/storage/loot/LootPool.java.patch
```diff
@@ -1,50 +1,24 @@
...
-    final LootItemFunction[] f_79026_;
-    private final BiFunction<ItemStack, LootContext, ItemStack> f_79027_;
+@@ -39,7 +_,8 @@
     NumberProvider f_79028_;
     NumberProvider f_79029_;
  
 -   LootPool(LootPoolEntryContainer[] p_165128_, LootItemCondition[] p_165129_, LootItemFunction[] p_165130_, NumberProvider p_165131_, NumberProvider p_165132_) {
--      this.f_79023_ = p_165128_;
--      this.f_79024_ = p_165129_;
 +   private LootPool(LootPoolEntryContainer[] p_165128_, LootItemCondition[] p_165129_, LootItemFunction[] p_165130_, NumberProvider p_165131_, NumberProvider p_165132_, String name) {
```
LexManos: Why is this private? And it changes signature, so add a bouncer.

sciwhiz12: That constructor signature change is part of the original patch and untouched by this PR, so a bouncer is unnecessary (as nothing calls this outside of this class as it is `private`). Furthermore, the change in modifier is because in 1.16.x, the unmodified constructor was `private`, but in 1.17.x, it is package-private, yet the patch was not changed during the update process to not add `private`.

LexManos: So it was a oversight during the update, and should be fixed properly in this pr...

PR #8185, opened 2021-11-02
Hotfix: Prevent a crash on a getCapabilities() call when the ItemStack is EMPTY

marchermans:
If a `getCapabilities()` call is made on the `EMPTY` ItemStack (so the actual object, as in the ones used in none populated slots in Nonnullists like in inventories) then a crash occurs.

This is due to the fact that this particular object does not have a delegate associated and and as such is the only object for which `gatherCapabilities(...)` is not called in the `forgeInit()` method of the itemstack.

In general this now prevents a crash from occurring on lazily initialized capability providers which don't call the `gatherCapabilities(...)` method to initialize the provider.

### DIFF
--- a/src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
+++ b/src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
@@ -57,6 +57,11 @@ public abstract class CapabilityProvider<B extends CapabilityProvider<B>> implem
     {
         this.baseClass = baseClass;
         this.isLazy = SUPPORTS_LAZY_CAPABILITIES && isLazy;
+
+        if (this.isLazy)
+        {
+            this.lazyParentSupplier = () -> null; //Items have a null delegate so this needs to be set to prevent a crash
+        }
     }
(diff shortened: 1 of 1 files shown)


marchermans:
I also incorporated the same fix in the 1.16 port #7879

sciwhiz12:
For transparency, the reason this PR was closed is because the one place in the 1.17.x codebase where `lazyParentSupplier` is referenced is already guarded by a `null` check, which was added by PR [#8096](https://github.com/MinecraftForge/MinecraftForge/pull/8096/files#diff-e8af24081992abb731132ffe8f9261746a689fe3653614377cf3ae6b16cb878f):
https://github.com/MinecraftForge/MinecraftForge/blob/394e76eed3ee54476528a0150090714fa9ba672b/src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java#L93

See [this Discord conversation](https://discord.com/channels/313125603924639766/852298000042164244/905215137621889054) about this PR in the official Forge discord server, which led to this PR's closing.

Issue #8188, opened 2021-11-04
[1.16.5] Client side PartEntity id doesn't match with server side


XenonDB:
**Minecraft Version:** 1.16.5

**Forge Version:** 1.16.5-36.1.0 & 1.16.5-36.2.9

**Description of issue:**

I'm making a system to let most entity adapted with PartEntity system of forge.
However, logs show the id of PartEntity from ender dragon in client side doesn't match with server side (see logs below)
These are generated by mixin in forge-1.16.5-36.1.0.

I think the reason is here:
https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/patches/minecraft/net/minecraft/client/network/play/ClientPlayNetHandler.java.patch#L49

i starts from 0 and causes id shifting (first PartEntity's id is same as its parent. However each entity should have different id)

This seems still exist in forge-1.16.5-36.2.9:
![未命名](https://user-images.githubusercontent.com/58476103/140300106-d7d861af-f693-475c-baeb-a2f5b736d21b.png)

**Logs:** 
![unknown](https://user-images.githubusercontent.com/58476103/140298465-392003ad-9782-42e6-8dd9-62f1c06ca12d.png)

is this a vanilla or forge bug?

uSkizzik:
This is more of a vanilla bug since the linked line isn't changed by Forge.
Forge Part Entities also **do not** touch IDs, so they can't be causing the issue.

ChampionAsh5357:
This is already confirmed and still unresolved in the Jira [MC-158205](https://bugs.mojang.com/browse/MC-158205).

uSkizzik:
Actually, I experienced this exact thing too! One of my entity's parts wasn't taking any damage. I tried fixing it but I gave up and left it like it was.

XenonDB:
ok...maybe I need to fix it by myself

uSkizzik:
> ok...maybe I need to fix it by myself

PR?

XenonDB:
> > ok...maybe I need to fix it by myself
> 
> PR?

Not sure if forge want to fix vanilla bug.
Maybe I will use mixin or prevent my custom PartEntity to be set a wrong id

ChampionAsh5357:
Why would you need a mixin for your own part entity? Just add 1 to the incoming client part entities and that should suffice.

uSkizzik:
> Not sure if forge want to fix vanilla bug.
[...]

PR #8190, opened 2021-11-06
[1.17.1]Fixes client side PartEntity id doesn't matches with server side

XenonDB:
#8188 
Fix shifting first despite other issue causing id mismatch that still exists.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java.patch
@@ -9,12 +9,10 @@
                       flag1 = this.f_19853_.m_7471_(blockpos, false) || flag1;
                    } else {
                       flag = true;
-@@ -847,6 +_,16 @@
- 
-    public boolean m_6072_() {
+@@ -849,12 +_,22 @@
        return false;
-+   }
-+
+    }
+ 
 +   @Override
 +   public boolean isMultipartEntity() {
 +      return true;
@@ -23,6 +21,15 @@
 +   @Override
 +   public net.minecraftforge.entity.PartEntity<?>[] getParts() {
 +      return this.f_31089_;
-    }
- 
++   }
++
     public void m_142223_(ClientboundAddMobPacket p_149572_) {
+       super.m_142223_(p_149572_);
+       EnderDragonPart[] aenderdragonpart = this.m_31156_();
+ 
+       for(int i = 0; i < aenderdragonpart.length; ++i) {
+-         aenderdragonpart[i].m_20234_(i + p_149572_.m_131552_());
++         aenderdragonpart[i].m_20234_((i + 1) + p_149572_.m_131552_()); //Forge: Fixes MC-158205
+       }
+ 
+    }


ChampionAsh5357:
I'm currently unsure whether this would be valuable if the [edge case](https://github.com/MinecraftForge/MinecraftForge/issues/8188#issuecomment-962162488) presented in the issue is present. Currently I'm unable to replicate this bug by checking the ids on world join so I'm unsure whether the provided test case is valid.

If the deferred ids are a true issue, then this would not be a valid fix. Please replicate your error along with set up and hardware specs so we can have a decent idea on what is happening since I don't really see anything within the code that may cause this.

XenonDB:
> I'm currently unsure whether this would be valuable if the [edge case](https://github.com/MinecraftForge/MinecraftForge/issues/8188#issuecomment-962162488) presented in the issue is present. Currently I'm unable to replicate this bug by checking the ids on world join so I'm unsure whether the provided test case is valid.
> 
> If the deferred ids are a true issue, then this would not be a valid fix. Please replicate your error along with set up and hardware specs so we can have a decent idea on what is happening since I don't really see anything within the code that may cause this.

Since logger will show which thread it current is in, I made a test below:
未命名
```java
if(!this.level.isClientSide) {
    	  LOGGER.info(this.getClass());
          LOGGER.info(this.id);
}
```

And I found there are other threads creating entity indeed: (these threads are called "Worker-Main-x")
https://paste.ee/p/zeKnD
https://paste.ee/p/0PKSq

These threads seem relative to chunk generation because if I back to screen and join the world again, these threads don't exist.
They are also relative to ender crystal generation.

However, parts id preempted may be a rare situation since I got into the end and let game generate it over 30 times but still don't encounter this situation again.

But, if I make many other entities also compatible with PartEntity system, such as enderman, this issue may be everywhere.

ChampionAsh5357:
So the edge case mentioned above comes from a fully constructed entity being added to a `ProtoChunk`. In these cases, the entity is not initially spawned. Instead, the entity is saved to nbt, and then rebuilt when the chunk becomes loaded. Since chunks are initially generated asynchronously, you could add a number of entities at the same time. They will then be added synchronously to the level once the chunk has finished loading. Since the entity id is an `AtomicInteger`. This wouldn't normally be a problem. However, in the case of multipart entities, this is a different story. In a very specific case when a `ProtoChunk` begins constructing which spawns an entity at the same time a multipart is loaded on the server thread, the entity id could be incremented by the worker causing the entity ids on the server to shift slightly. This could cause a desynchronization from the server to client ids as it is expected that the multipart ids would be in the same increments after the main entity id.

This problem is mostly mitigated by how Mojang handles their ticking structure, but in some extremely specific circumstances this would be very present. I would imagine you could replicate this buggy behavior by replacing every loading entity with that of a multipart entity with 10 parts such that the chance of the issue occurring is higher.

The simple solution to this would be to send all the ids for the multiparts to the client to avoid this issue. Another solution may check the current status of the loaded area and only get and increment the id if the entity will be actually joining the level instead of serialized. Since the entity id is never serialized, we don't need to worry about any collisions.

Since this is fixing part entity desynchronization, this should also be addressed in the PR.

XenonDB:
I think I need to do some research on how server send entity id packet to client.
Because the connection between server side and client side entity is only a entity id, client must set a correct id for a new created entity immediately or server will lost control of it.

sciwhiz12:
Hello! @XenonDB, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable. If we do not hear back from you after some time, this PR will be closed for inactivity.

XenonDB:
> Hello! @XenonDB, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.18 if applicable. If we do not hear back from you after some time, this PR will be closed for inactivity. If anyone else wishes to continue the work of this PR, please create a PR with a link to this PR for reference. Thank you for your contributions. 👋

I think I should leave this work to other since I am busy recently.

PR #8193, opened 2021-11-11
Added JVM Version Checker

DaemonUmbra:
This should hopefully allow us to get warnings out to users on old or incompatible JVMs


marchermans (approved):

diesieben07:
Running `.java` files directly from the command line (without compilation) using the `java` executable was added in Java 11, so this feature can't be used here.

server_files/run.sh
```diff
@@ -1,6 +1,12 @@
 #!/usr/bin/env sh
+
+# Some versions of Java lack features necessary for Forge to function.
+# CheckVersion.java will be invoked to check for these versions.
+# To skip this check, simply comment out the following line:
+if ! java CheckVersion.java ; then exit 0; fi
```
desht: This could be more concisely written as `java CheckVersion.java || exit 1` (diesieben's comment about running .java files directly notwithstanding)

Also note exit status 1, not 0 - non-zero exit code is Unix convention for non-successful script termination.

DaemonUmbra: That was supposed to be a 1, oops

DaemonUmbra:
This got too complex for what little it does to be accepted

PR #8194, opened 2021-11-11
[1.17.x] Change expected type of LootTable fields to arrays

SuperMartijn642:
#8184 reverted some LootTable patches, including the types of fields `f_79023_` and `f_79024_`. However, `ForgeLootTableProvider` expects a list for these fields, however the type of the fields is now an array.
This PR changes the expected type in `ForgeLootTableProvider` to an array as well.


ChampionAsh5357 (approved):
Looks good and works fine. Apologies for not catching this originally.

PR #8196, opened 2021-11-13
Fix death message for tamed animals

### LINKED ISSUE #8195: [1.17.1] Wrong death message sent to chat for pets
**Minecraft Version:** 1.17.1

**Forge Version:** 37.0.109

**Description of issue:**
Mobs that extend `TamableAnimal` do not send the correct death message to their owner. No matter how the pet died, the message is always "[Pet name] died" as a result of [this](https://github.com/MinecraftForge/MinecraftForge/pull/8110) pull request that moved when the death message was sent to after the`LivingDeathEvent` is fired. The problem is that the death message is taken from the `CombatTracker` class which is reset in the method `LivingEntity#die`, which is why `TamableAnimal` originally sent the death message before the super call to `LivingEntity#die`.

diesieben07:
Fixes #8195

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/TamableAnimal.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/TamableAnimal.java.patch
@@ -1,15 +1,17 @@
 --- a/net/minecraft/world/entity/TamableAnimal.java
 +++ b/net/minecraft/world/entity/TamableAnimal.java
-@@ -197,11 +_,14 @@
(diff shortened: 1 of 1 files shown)


Mysterious-Dev (approved):
LGTM (I don't know if code is formated properly but the bug is fix for me).

sciwhiz12 (approved):
Confirmed via in-game test with a wolf exploded by a block of TNT.

PR #8205, opened 2021-11-20
Add DamageBonus event to allow more sophisticated damage enchantments

### LINKED ISSUE #6564: Allow enchantments to modify damage based on the entity hit
Enchantments can only modify the dealt damage based on the entity's `CreatureAttribute`, not on the entity object itself. Therefore it's not possible to e.g. change the damage based some capability of a player. More information in the linked PR below.

I once made a PR https://github.com/MinecraftForge/MinecraftForge/pull/4052, which was eventually declined due to some thread safety concerns by Lex.
However, the issue is still present. I don't assume Lex got around talking to grum about this, but I noticed that in `PlayerEntity#attackTargetEntityWithCurrentItem` the entity object is passed to a `CriticalHitEvent` just a few lines after the enchantment damage calculation, which means it is also available to modders.

I do not exactly know what's the thread safety issue here (as I don't see any difference from other parts of MC), but if there is a potential issue, then it is already exposed by said event. So why not add the enchantment thing as well.

If you agree, I would update the PR for 1.15 and make sure to add a comment that the entity should not be modified in that method.

maxanier:
Closes #6564

Enchantments can only modify the dealt damage based on the entity's CreatureAttribute, not on the entity object itself. Therefore it's not possible to e.g. change the damage based a capability of an attacked player/entity.

## Details
When a player attacks an entity, the dealt damage is calculated in `Player#attack` as follows:
`Total = (0.8*Att * Scale^2 + 0.2 * Att) * Critical + Ench * Scale`
Where
`Att` AttackDamage attribute
`Scale` multiplier from the attack cooldown
`Critical` multiplier based on critical hit
`Ench` Additional damage caused by enchantment

During `LivingAttackEvent` only `Total` is available. The attack cooldown has been reset already, so there is no way to accurately calculate the damage caused by a potential enchantment during that event. Only workaround I can see is to cache the attack cooldown during `AttackEntityEvent` event. But that will get messy, because two different entities (attacker and victim) are involved and the event (or the firing method) is called from many different places.

## Solution
As suggested by @ChampionAsh5357, this PR introduces a new event. This allows modifying the base damage before the scale is applied.
`Total = (0.8*Att * Scale^2 + 0.2 * Att) * Critical + (Ench + Bonus) * Scale`

For Mobs attacking with enchanted weapons `LivingAttackEvent` is sufficient, because they do not have the attack cooldown functionality.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
@@ -200,6 +200,15 @@
        if (p_36347_.m_6097_()) {
           if (!p_36347_.m_7313_(this)) {
              float f = (float)this.m_21133_(Attributes.f_22281_);
+@@ -1037,7 +_,7 @@
+             } else {
+                f1 = EnchantmentHelper.m_44833_(this.m_21205_(), MobType.f_21640_);
+             }
+-
++            f1 = net.minecraftforge.common.ForgeHooks.getDamageBonus(this, this.m_21205_(), p_36347_, f1);
+             float f2 = this.m_36403_(0.5F);
+             f = f * (0.2F + f2 * f2 * 0.8F);
+             f1 = f1 * f2;
 @@ -1045,7 +_,7 @@
              if (f > 0.0F || f1 > 0.0F) {
                 boolean flag = f2 > 0.9F;
--- a/patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
@@ -200,11 +200,10 @@
        if (p_36347_.m_6097_()) {
           if (!p_36347_.m_7313_(this)) {
              float f = (float)this.m_21133_(Attributes.f_22281_);
-@@ -1037,7 +_,7 @@
-             } else {
+@@ -1038,6 +_,7 @@
                 f1 = EnchantmentHelper.m_44833_(this.m_21205_(), MobType.f_21640_);
              }
--
+ 
 +            f1 = net.minecraftforge.common.ForgeHooks.getDamageBonus(this, this.m_21205_(), p_36347_, f1);
              float f2 = this.m_36403_(0.5F);
              f = f * (0.2F + f2 * f2 * 0.8F);
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -154,6 +154,7 @@ import net.minecraftforge.event.entity.player.AdvancementEvent;
 import net.minecraftforge.event.entity.player.AnvilRepairEvent;
 import net.minecraftforge.event.entity.player.AttackEntityEvent;
 import net.minecraftforge.event.entity.player.CriticalHitEvent;
+import net.minecraftforge.event.entity.player.DamageBonusEvent;
 import net.minecraftforge.event.entity.player.PlayerEvent;
 import net.minecraftforge.event.entity.player.PlayerInteractEvent;
 import net.minecraftforge.event.world.BiomeLoadingEvent;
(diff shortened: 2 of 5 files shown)


patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
```diff
@@ -200,6 +200,15 @@
        if (p_36347_.m_6097_()) {
           if (!p_36347_.m_7313_(this)) {
              float f = (float)this.m_21133_(Attributes.f_22281_);
+@@ -1037,7 +_,7 @@
+             } else {
+                f1 = EnchantmentHelper.m_44833_(this.m_21205_(), MobType.f_21640_);
+             }
+-
```
ChampionAsh5357: You can just add an extra line here instead of replacing to reduce the patch size to one line instead of two.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1326,4 +1327,11 @@ public static void onEntityEnterSection(Entity entity, long packedOldPos, long p
         MinecraftForge.EVENT_BUS.post(new EntityEvent.EnteringSection(entity, packedOldPos, packedNewPos));
     }
 
+    public static float getDamageBonus(Player player, ItemStack weapon, Entity target, float enchantmentBonus)
```
ChampionAsh5357, line 1330: I would probably reduce the scope type to `LivingEntity` for reasons I'll explain in the main comment.

src/main/java/net/minecraftforge/event/entity/player/DamageBonusEvent.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * This event is fired whenever a player attacks an Entity in
+ * Player#attack(Entity).<br>
+ * <br>
+ * This event is not {@link Cancelable}.<br>
+ * <br>
+ * This allows adding to the base damage like DamageEnchantments do, but also considering the attacked Entity object including capabilities, not just the Entity's MobType
+ * <br>
+ * This event is fired on the {@link MinecraftForge#EVENT_BUS}.
```
ChampionAsh5357: I'd use standard javadoc formatting with trailing `<p>` tags. Also one of those lines are just really long and should be spaced into multiple lines. Javadoc formatting will handle that with no issues.

src/main/java/net/minecraftforge/event/entity/player/DamageBonusEvent.java
```diff
@@ -0,0 +1,81 @@
...
+        return weapon;
+    }
+
+    /**
+     * @param additionalDamage Added to the existing bonus damage
+     */
+    public void addBonus(float additionalDamage)
+    {
+        this.bonus += additionalDamage;
+    }
```
ChampionAsh5357, line 74: I would probably recommend making this a `setBonus` method instead of add as most Forge methods uses the `set` syntax so reclaculation may be different. This could be an additional method, but there should be something to just set raw.

ChampionAsh5357 (changes requested):
Is the scope of this event only supposed to pertain to players? It seems that the idea provided for damage enchantments makes it much more feasible to add this change to other entity attack methods as well.

Also, you should add a test mod to verify that the new event works as intended.

diesieben07:
To avoid adding another event, couldn't this be done via an attribute? Either via the existing attack damage or via a new to be introduced one ("forge:attack_damage_bonus" or similar). Then either `ItemAttributeModifierEvent` could be used to apply the modifier to all stacks with the enchantment, or an `IForgeEnchantment` could be introduced to add the modifiers without an event.

KnightMiner:
I don't know that an attribute will work if the goal is recreating enchantments like smite. Such enchantments conditionally boost the player's damage based on the targeted mob, so unless you make an attribute for each possible targeted mob you lack the proper context.

The other option is adding an event to add attributes before attack (with context) and remove extra attributes after attack, but how is that different from this proposed event?

maxanier:
> Is the scope of this event only supposed to pertain to players? It seems that the idea provided for damage enchantments makes it much more feasible to add this change to other entity attack methods as well.

Yes, it would probably be nicer to have the event for mob attacks as well. However, it is not strictly necessary for functionality. As mobs do not have the attackStrengthScale (cooldown) that players have, one can just use `LivingAttackEvent` to manipulate the damage dealt. 
To minimize the number of patches, I decided to keep this player only. But I can extend this to all LivingEntities if you want

> Also, you should add a test mod to verify that the new event works as intended.

I wasn't sure if it is necessary to add a dedicated test mod for such a simple patch, so I just tested it manually. I will add one

> To avoid adding another event, couldn't this be done via an attribute? Either via the existing attack damage or via a new to be introduced one ("forge:attack_damage_bonus" or similar). Then either ItemAttributeModifierEvent could be used to apply the modifier to all stacks with the enchantment, or an IForgeEnchantment could be introduced to add the modifiers without an event.

As KnightMiner explained the goal is to boost the damage based on the targeted mob. The enchantment could for example consider an active potion effect on the targeted mob, or on an attribute stored in a capability of the targeted mob/player.

maxanier:
1.18: https://github.com/MinecraftForge/MinecraftForge/pull/8243

PR #8206, opened 2021-11-20
[1.17] Fix Spectating Part Entity (MC-46486)

uSkizzik:
This PR fixes MC-46486 which affects all Part Entities by adding a check in ServerPlayer#setCamera to check if the entity the player  is trying to spectate is an instance of `PartEntity`. If it is, the player starts spectating the parent entity instead.

### DIFF
--- a/patches/minecraft/net/minecraft/server/level/ServerPlayer.java.patch
+++ b/patches/minecraft/net/minecraft/server/level/ServerPlayer.java.patch
@@ -150,12 +150,11 @@
     }
  
     public boolean m_143432_() {
-@@ -1264,7 +_,7 @@
- 
+@@ -1265,6 +_,7 @@
     public void m_9213_(Entity p_9214_) {
        Entity entity = this.m_8954_();
--      this.f_8926_ = (Entity)(p_9214_ == null ? this : p_9214_);
-+      this.f_8926_ = (Entity)(p_9214_ == null ? this : p_9214_ instanceof net.minecraftforge.entity.PartEntity ? ((net.minecraftforge.entity.PartEntity<?>) p_9214_).getParent() : p_9214_); // FORGE: fix MC-46486
+       this.f_8926_ = (Entity)(p_9214_ == null ? this : p_9214_);
++      if (this.f_8926_ instanceof net.minecraftforge.entity.PartEntity<?>) this.f_8926_ = ((net.minecraftforge.entity.PartEntity<?>) f_8926_).getParent(); // FORGE: fix MC-46486
        if (entity != this.f_8926_) {
           this.f_8906_.m_141995_(new ClientboundSetCameraPacket(this.f_8926_));
           this.m_6021_(this.f_8926_.m_20185_(), this.f_8926_.m_20186_(), this.f_8926_.m_20189_());
--- a/patches/minecraft/net/minecraft/server/level/ServerPlayer.java.patch
+++ b/patches/minecraft/net/minecraft/server/level/ServerPlayer.java.patch
@@ -154,7 +154,7 @@
     public void m_9213_(Entity p_9214_) {
        Entity entity = this.m_8954_();
        this.f_8926_ = (Entity)(p_9214_ == null ? this : p_9214_);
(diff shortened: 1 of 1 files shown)


patches/minecraft/net/minecraft/server/level/ServerPlayer.java.patch
```diff
@@ -150,6 +150,15 @@
     }
  
     public boolean m_143432_() {
+@@ -1264,7 +_,7 @@
+ 
+    public void m_9213_(Entity p_9214_) {
+       Entity entity = this.m_8954_();
+-      this.f_8926_ = (Entity)(p_9214_ == null ? this : p_9214_);
++      this.f_8926_ = (Entity)(p_9214_ == null ? this : p_9214_ instanceof net.minecraftforge.entity.PartEntity ? ((net.minecraftforge.entity.PartEntity<?>) p_9214_).getParent() : p_9214_); // FORGE: fix MC-46486
```
pupnewfster: I wonder if this patch might be a bit cleaner by instead of doing this, on the next line adding:
```java
if (this.f_8926_ instanceof net.minecraftforge.entity.PartEntity<?> partEntity) this.f_8926_ = partEntity.getParent(); // FORGE: fix MC-46486
```
Though I forget if forge's patching system supports the pattern matching instanceof so if it doesn't then the extra cast would be needed, but either way I think it would slightly reduce the complexity of the patch and improve the readability of the code.

uSkizzik: There ya go. I decided to go with a cast instead, just in case.

patches/minecraft/net/minecraft/server/level/ServerPlayer.java.patch
```diff
@@ -150,6 +150,14 @@
     }
  
     public boolean m_143432_() {
+@@ -1265,6 +_,7 @@
+    public void m_9213_(Entity p_9214_) {
+       Entity entity = this.m_8954_();
+       this.f_8926_ = (Entity)(p_9214_ == null ? this : p_9214_);
++      if (this.f_8926_ instanceof net.minecraftforge.entity.PartEntity<?> partEntity) this.f_8926_ = partEntity.getParent(); // FORGE: fix MC-46486
```
ChampionAsh5357: So while this could be fine, is there any reason why the parent of the part entity is not another part entity? Or is this just an assumed thing with the PartEntity API that the parent will always be the actual entity?

uSkizzik: I feel like part entities can't actually have their own part entities. From my experience they're really buggy and can't even act like a normal entity. I should probably still test it out though.

ChampionAsh5357: Well give it a test today so it can get in for the 1.17 RB. I don't see any reason why it couldn't be true, but if you determine otherwise, just give a reason for it.

uSkizzik: I'll test it out tomorrow after school since I gtg now.

uSkizzik: @ChampionAsh5357 
I have an idea. Maybe I can check for the entity. If it's an instance of parent entity, I'll rerun the setCamera method with the parent and if it isn't I'll continue executing it.

ChampionAsh5357: That's not a good idea. You could just as simply make it a while loop that just says while the entity is a part, set the entity to the parent.

uSkizzik: Yea that's what I thought and that's why I commented before actually doing it.
Anyways, I changed it.

ChampionAsh5357 (approved):
Scopewise, this does fix mc-46486 and all subsequent part entities. However, I will leave a note that the direction the ender dragon is looking should be another mc issue if it is not already.

PR #8207, opened 2021-11-21
Apply window resizing to all gui layers

MelanX:
This will cause *all* screens in the guiLayers Stack to re-init. 
Fixes widgets being dislocated on bottom layers when re-sizing the window.

### DIFF
--- a/patches/minecraft/net/minecraft/client/Minecraft.java.patch
+++ b/patches/minecraft/net/minecraft/client/Minecraft.java.patch
@@ -185,7 +185,13 @@
        }
  
        if (this.f_91056_ != null) {
-@@ -1136,6 +_,7 @@
+@@ -1131,11 +_,12 @@
+       int i = this.f_90990_.m_85385_(this.f_91066_.f_92072_, this.m_91390_());
+       this.f_90990_.m_85378_((double)i);
+       if (this.f_91080_ != null) {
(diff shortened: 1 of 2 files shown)


patches/minecraft/net/minecraft/client/Minecraft.java.patch
```diff
@@ -185,7 +185,13 @@
        }
  
        if (this.f_91056_ != null) {
-@@ -1136,6 +_,7 @@
+@@ -1131,11 +_,12 @@
+       int i = this.f_90990_.m_85385_(this.f_91066_.f_92072_, this.m_91390_());
+       this.f_90990_.m_85378_((double)i);
+       if (this.f_91080_ != null) {
+-         this.f_91080_.m_6574_(this, this.f_90990_.m_85445_(), this.f_90990_.m_85446_());
```
gigaherz: You should leave this in. The guiLayers list in ForgeHooksClient contains only the layers behind, so by removing this call you'd be causing the top layer to not have resize called on it.

gigaherz (changes requested):

patches/minecraft/net/minecraft/client/Minecraft.java.patch
MelanX: Fixed

gigaherz (approved):

PR #8210, opened 2021-11-24
1.16.x Add option to ITeleporter to disable creation of an Obsidian Platform when travelling to the End dimension.

TheCosmicNebula:
Im running into an issue with my mod, where whenever I changeDimension the player to the End, it always spawns the End platform, no matter what BlockPos I teleport them to. I can't find a way to disable this with the current system.


pupnewfster:
Just pass false for the spawn portal input of the reposition entity function?

PR #8211, opened 2021-11-24
Fix potential issue with BackgroundScanHandler for users with slow disc IO

DaemonUmbra:
1.16.x Cherry-Pick of #8176 for LTS purposes


ChampionAsh5357 (approved):
So, this wasn't a bug back in 1.16 as it was never fully implemented. But since this fixes the issue anyways, I see no reason why this can't be merged in.

Issue #8201, opened 2021-11-18
Unnecessary durability bar methods in item extension


PepperCode1:
`IForgeItem` contains the following three methods to define how and when a durability bar should be rendered for a particular `ItemStack`: `showDurabilityBar`, `getDurabilityForDisplay`, and `getRGBDurabilityForDisplay`. However, these methods are no longer necessary and can in fact create confusion because 1.17 added exact replacements for all three. Because the `ItemRenderer` patch removes usage of the vanilla methods and replaces it with the Forge ones, the vanilla ones become completely useless.
Replacements (Forge -> Mojmap):
`showDurabilityBar` -> `isBarVisible`
`getDurabilityForDisplay` -> `getBarWidth`
`getRGBDurabilityForDisplay` -> `getBarColor`
The best solution I ended up at, after discussion in the Forge Discord, was to deprecate the `IForgeItem` methods for removal in 1.18, remove the patched lines in `ItemRenderer`, and patch the vanilla methods to redirect to the Forge methods. I may make a PR for these changes if time permits.

TheCurle:
You're totally right. I have no idea how this slipped past us, thanks for the heads up.

PR #8213, opened 2021-11-25
Reorder image tooltip component to match vanilla

### LINKED ISSUE #8212: [1.17.1] Bundle is missing progress bar and item slots are misplaced
**Minecraft Version:** 1.17.1

**Forge Version:** 37.0.112

**Steps to Reproduce:**
 1. Give yourself `minecraft:bundle`
 2. Put some items in

**Description of issue:**
Vanilla 1.17.1 - as you can see, the durability/progress bar is visible on top of the item and the item slots are shown above the ID and item counter.
image
Forge 37.0.112 - the progress bar is not visible and the slots are below the item ID and item counter.
image

sciwhiz12:
This PR fixes #8212 by moving the image tooltip component (used most prominently by the bundle item) to index `1`, after the component for the item name and before the rest of the components, matching vanilla rendering. 

This seems to have been an oversight by the tooltip rewrite PR #8038. The PR patched `Screen` to remove the code which adds the image component to the second position within the list (the index argument of `1`):

https://github.com/MinecraftForge/MinecraftForge/blob/cdd38270e9080a4579003c5138afbe300e00693e/patches/minecraft/net/minecraft/client/gui/screens/Screen.java.patch#L39-L41

That piece of code was translocated to the Forge hook in `ForgeHooksClient` instead, but the index argument was forgotten, which leads the image component to be added to the end of the list and causes the linked issue:

https://github.com/MinecraftForge/MinecraftForge/blob/cdd38270e9080a4579003c5138afbe300e00693e/src/main/java/net/minecraftforge/client/ForgeHooksClient.java#L1126

This PR simply re-adds the index argument of `1` to the `List#add` call, mimicing the original vanilla code to insert the image component to the second position within the list rather than at the end.

**In-game screenshot of the PR in action:**
image

### DIFF
--- a/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
+++ b/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
@@ -1123,7 +1123,7 @@ public class ForgeHooksClient
         List<Either<FormattedText, TooltipComponent>> elements = textElements.stream()
                 .map((Function<FormattedText, Either<FormattedText, TooltipComponent>>) Either::left)
                 .collect(Collectors.toCollection(ArrayList::new));
-        itemComponent.ifPresent(c -> elements.add(Either.right(c)));
+        itemComponent.ifPresent(c -> elements.add(1, Either.right(c)));
 
         var event = new RenderTooltipEvent.GatherComponents(stack, screenWidth, screenHeight, elements, -1);
         MinecraftForge.EVENT_BUS.post(event);


ChampionAsh5357:
So, this only fixes half the problem. It does fix the rendering with the tooltip, but not the progress bar. That relates to #8201 since `BundleItem` overrides `Item#isBarVisible`. For the time being until the #8201 is implemented, `#showDurabilityBar` should be implemented on the `BundleItem` to account for this oversight. Alternatively, this could also patch #8201 quickly.

PR #8219, opened 2021-11-27
[1.16] Reintroduce modded biome variants

maxanier:
1.16 version of #8137 
In 1.18 the biome layer system is removed/overhauled, so this PR is only interesting for 1.16 LTS

This PR reintroduces a way to register/provide certain biome variants for your own biomes (or for vanilla or other modded biomes).
This e.g. allows having a hills variant of your biome seamlessly integrate into the world.

## Biome layer and variants in vanilla
MC first generates a "biome layer" based on the biomes with a biome entry/weight which specifies the biome for each chunk. Then it applies several "layers" that can modify the biome layer. For example the `HillsLayer` can replace some biome chunks within a larger biome with a hills variant of that biome (`desert` -> `desert_hills`).
These modifications work on hardcoded biome ids in the respective layer class.

## Previous work
In 1.15 three of these layers were patched to allow mods to introduce their own biome variants
Hills: https://github.com/MinecraftForge/MinecraftForge/pull/6571
River: https://github.com/MinecraftForge/MinecraftForge/pull/5969
Shore: https://github.com/MinecraftForge/MinecraftForge/pull/7001

They all rely on new methods in the `Biome` class which can be overridden to return the respective variant for the modded biome.
Since with 1.16.2 modded biomes do not use a subclass of `Biome`, a new method is needed.

## How this works
The patches to the biome modification layers remain functionally unchanged as nothing has changed here. Only analog patch to `EdgeBiomeLayer` is added.
Instead of querying the respective Biome, Forge's `BiomeManager` class is now used to get the biome variants.
If a modder wants to provide biome variants to their own (or other biomes), they can register a `IBiomeLayerModifier` and override the respective method to provide the desired variant.
The integer biome ids that are used in the layer classes are translated to the `RegistryKey` id of the biome.

## Test
This PR brings a (disabled-by-default) test mod, that introduces a new biome (added with a `BiomeEntry` to be generated) and hills, edge, shore and river variant.
Must be manually activated.

## Discussion
Things to think about:
- Use `null` instead of `Optional` in `IBiomeLayerModifier` return
- Instead of registering a single `IBiomeLayerModifier` that is queried for any biome, register biome specific `IBiomeLayerModifier`

### DIFF
--- a/patches/minecraft/net/minecraft/world/gen/layer/EdgeBiomeLayer.java.patch
+++ b/patches/minecraft/net/minecraft/world/gen/layer/EdgeBiomeLayer.java.patch
@@ -1,0 +1,14 @@
+--- a/net/minecraft/world/gen/layer/EdgeBiomeLayer.java
++++ b/net/minecraft/world/gen/layer/EdgeBiomeLayer.java
+@@ -20,6 +_,11 @@
+                }
+             }
+ 
++            net.minecraftforge.registries.ForgeRegistry<net.minecraft.world.biome.Biome> registry = (net.minecraftforge.registries.ForgeRegistry<net.minecraft.world.biome.Biome>) net.minecraftforge.registries.ForgeRegistries.BIOMES;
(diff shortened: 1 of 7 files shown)


sciwhiz12:
Hello! This PR will now be closed as 1.16 is out of support. Thank you for your contributions. :wave:

PR #8220, opened 2021-11-28
[1.17.x] Added missing copper ingot item tag.

MarkusBordihn:
The copper ingot is currently missing in the item tags.
This PR adds the copper ingot to the item tags.

Run a test instance and confirmed that the copper ingot recipes works as expected:
Screenshot 2021-11-28 160256


ChampionAsh5357:
This is a subset of #7891 which already encapsulates this change and much more. For this reason, I'm closing the PR.

PR #8224, opened 2021-11-30
Network Capabilities

josephtarbit:
1.18 port of #7952 

The aim of this PR is to provide a simple system for capabilities to synchronize data along side vanilla's synchronization of it's own data.

I have also changed a few `ItemStack::matches` checks to use `ItemStack::equals` in order to compare share tags instead of full tags. This is unrelated to the implementation of `INetworkCapability` however I thought I might as well fix it.

### `ItemStack` Capabilities

*Note: it's also now impossible to send your own custom sync packets for arbitrary container menus as there is no way to access a player from a `ContainerSynchronizer` instance.*

### `INetworkCapability`

```java
void writeCapabilities(FriendlyByteBuf out, boolean writeAll);

void readCapabilities(FriendlyByteBuf in);

boolean requiresSync();
```


src/main/java/net/minecraftforge/common/capabilities/INetworkCapability.java
```diff
@@ -0,0 +1,65 @@
...
+     * generally {@link #writeCapabilities(FriendlyByteBuf, boolean)} should be used
+     * to avoid unnecessarily creating a {@link byte[]}.
+     * 
+     * @param writeAll - {@code true} if all data should be written, {@code false}
+     *                 if only dirty data should be written
+     * @return a {@code byte[]} containing the serialized capability data
+     */
+    default byte[] writeCapabilities(boolean writeAll)
+    {
+        var buf = new FriendlyByteBuf(Unpooled.buffer());
```
ChampionAsh5357, line 64: Probably shouldn't be just an unpooled buffer. Packets do have a maximum size limit and these are not being split.

src/main/java/net/minecraftforge/common/capabilities/INetworkCapability.java
```diff
@@ -0,0 +1,65 @@
+package net.minecraftforge.common.capabilities;
```
ChampionAsh5357, line 20: Missing license header.

src/main/java/net/minecraftforge/common/capabilities/INetworkCapability.java
```diff
@@ -0,0 +1,65 @@
...
+
+import io.netty.buffer.Unpooled;
+import net.minecraft.network.FriendlyByteBuf;
+
+/**
+ * Provides the ability for {@link ICapabilityProvider}s to synchronize
+ * capabilities over the network.
+ */
+public interface INetworkCapability {
```
ChampionAsh5357: New line braces.

src/main/java/net/minecraftforge/common/capabilities/INetworkCapability.java
```diff
@@ -0,0 +1,65 @@
...
+     * 
+     * @param bytes - the {@link byte[]} to read from
+     */
+    default void readCapabilities(byte[] bytes)
+    {
+        var buf = new FriendlyByteBuf(Unpooled.wrappedBuffer(bytes));
+        this.readCapabilities(buf);
+        buf.release();
+    }
+}
```
ChampionAsh5357: New line at EOF.

src/main/java/net/minecraftforge/network/ForgeNetwork.java
```diff
@@ -0,0 +1,269 @@
...
+import net.minecraft.world.level.block.entity.BlockEntity;
+import net.minecraftforge.network.message.BlockEntityCapabilitiesMessage;
+import net.minecraftforge.network.message.EntityCapabilitiesMessage;
+import net.minecraftforge.network.message.EquipmentSlotCapabilitiesMessage;
+import net.minecraftforge.network.message.SlotCapabilitiesMessage;
+import net.minecraftforge.network.simple.SimpleChannel;
+
+public class ForgeNetwork {
+
+    public static final String PROTOCOL_VERSION = "0.0.1";
```
ChampionAsh5357: Why `0.0.1`? Seems arbitrary just not to use `1` for this.

src/main/java/net/minecraftforge/network/ForgeNetwork.java
```diff
@@ -0,0 +1,269 @@
+package net.minecraftforge.network;
```
ChampionAsh5357, line 20: Missing license header.

src/main/java/net/minecraftforge/network/ForgeNetwork.java
```diff
@@ -0,0 +1,269 @@
...
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.level.ChunkPos;
+import net.minecraft.world.level.block.entity.BlockEntity;
+import net.minecraftforge.network.message.BlockEntityCapabilitiesMessage;
+import net.minecraftforge.network.message.EntityCapabilitiesMessage;
+import net.minecraftforge.network.message.EquipmentSlotCapabilitiesMessage;
+import net.minecraftforge.network.message.SlotCapabilitiesMessage;
+import net.minecraftforge.network.simple.SimpleChannel;
+
+public class ForgeNetwork {
```
ChampionAsh5357: New line braces.

src/main/java/net/minecraftforge/network/message/BlockEntityCapabilitiesMessage.java
```diff
@@ -0,0 +1,35 @@
+package net.minecraftforge.network.message;
```
ChampionAsh5357, line 20: Missing license header.

src/main/java/net/minecraftforge/network/ForgeNetwork.java
```diff
@@ -0,0 +1,269 @@
...
+            if (players != null)
+            {
+                return players.stream()
+                    .map(ServerPlayerConnection::getPlayer)
+                    .map(player -> player.connection.getConnection());
+            }
+        }
+        return Stream.empty();
+    }
+}
```
ChampionAsh5357: New line at EOF.

src/main/java/net/minecraftforge/network/message/BlockEntityCapabilitiesMessage.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.function.Supplier;
+
+import net.minecraft.core.BlockPos;
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraftforge.common.util.LogicalSidedProvider;
+import net.minecraftforge.network.NetworkEvent;
+
+public record BlockEntityCapabilitiesMessage(BlockPos blockPos, FriendlyByteBuf capabilityData) {
```
ChampionAsh5357: New line braces.

src/main/java/net/minecraftforge/network/message/BlockEntityCapabilitiesMessage.java
```diff
@@ -0,0 +1,35 @@
...
+    public boolean handle(Supplier<NetworkEvent.Context> ctx)
+    {
+        ctx.get()
+            .enqueueWork(() -> LogicalSidedProvider.CLIENTWORLD.get(ctx.get().getDirection().getReceptionSide())
+                .map(level -> level.getBlockEntity(this.blockPos))
+                .ifPresent(blockEntity -> blockEntity.readCapabilities(this.capabilityData)))
+            .thenRun(this.capabilityData::release);
+        return true;
+    }
+}
```
ChampionAsh5357: New line at EOF.

src/main/java/net/minecraftforge/network/message/EntityCapabilitiesMessage.java
```diff
@@ -0,0 +1,34 @@
+package net.minecraftforge.network.message;
```
ChampionAsh5357, line 20: Missing license header.

src/main/java/net/minecraftforge/network/message/EntityCapabilitiesMessage.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.function.Supplier;
+
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraftforge.common.util.LogicalSidedProvider;
+import net.minecraftforge.network.NetworkEvent;
+
+public record EntityCapabilitiesMessage(int entityId, FriendlyByteBuf capabilityData) {
```
ChampionAsh5357: New line braces.

src/main/java/net/minecraftforge/network/message/EntityCapabilitiesMessage.java
```diff
@@ -0,0 +1,34 @@
...
+    public boolean handle(Supplier<NetworkEvent.Context> ctx)
+    {
+        ctx.get()
+            .enqueueWork(() -> LogicalSidedProvider.CLIENTWORLD.get(ctx.get().getDirection().getReceptionSide())
+                .map(level -> level.getEntity(this.entityId))
+                .ifPresent(entity -> entity.readCapabilities(this.capabilityData)))
+            .thenRun(this.capabilityData::release);
+        return true;
+    }
+}
```
ChampionAsh5357: New line at EOF.

src/main/java/net/minecraftforge/network/message/EquipmentSlotCapabilitiesMessage.java
```diff
@@ -0,0 +1,39 @@
+package net.minecraftforge.network.message;
```
ChampionAsh5357, line 20: Missing license header.

src/main/java/net/minecraftforge/network/message/EquipmentSlotCapabilitiesMessage.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.function.Supplier;
+
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.world.entity.EquipmentSlot;
+import net.minecraft.world.entity.LivingEntity;
+import net.minecraftforge.common.util.LogicalSidedProvider;
+import net.minecraftforge.network.NetworkEvent;
+
+public record EquipmentSlotCapabilitiesMessage(int entityId, EquipmentSlot slot, FriendlyByteBuf capabilityData) {
```
ChampionAsh5357: New line braces.

src/main/java/net/minecraftforge/network/message/EquipmentSlotCapabilitiesMessage.java
```diff
@@ -0,0 +1,39 @@
...
+        ctx.get()
+            .enqueueWork(() -> LogicalSidedProvider.CLIENTWORLD.get(ctx.get().getDirection().getReceptionSide())
+                .map(level -> level.getEntity(this.entityId))
+                .filter(LivingEntity.class::isInstance)
+                .map(LivingEntity.class::cast)
+                .ifPresent(livingEntity -> livingEntity.getItemBySlot(this.slot).readCapabilities(this.capabilityData)))
+            .thenRun(this.capabilityData::release);
+        return true;
+    }
+}
```
ChampionAsh5357: New line at EOF.

src/main/java/net/minecraftforge/network/message/SlotCapabilitiesMessage.java
```diff
@@ -0,0 +1,41 @@
+package net.minecraftforge.network.message;
```
ChampionAsh5357, line 20: Missing license header.

src/main/java/net/minecraftforge/network/message/SlotCapabilitiesMessage.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.function.Supplier;
+
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.world.entity.player.Player;
+import net.minecraftforge.common.util.LogicalSidedProvider;
+import net.minecraftforge.network.NetworkEvent;
+
+public record SlotCapabilitiesMessage(int entityId, int slotIndex, FriendlyByteBuf capabilityData) {
```
ChampionAsh5357: New line braces.

src/main/java/net/minecraftforge/network/message/SlotCapabilitiesMessage.java
```diff
@@ -0,0 +1,41 @@
...
+                .filter(Player.class::isInstance)
+                .map(Player.class::cast)
+                .map(player -> player.containerMenu)
+                .filter(menu -> this.slotIndex < menu.slots.size())
+                .map(container -> container.getSlot(this.slotIndex))
+                .ifPresent(slot -> slot.getItem().readCapabilities(this.capabilityData)))
+            .thenRun(this.capabilityData::release);
+        return true;
+    }
+}
```
ChampionAsh5357: New line at EOF.

src/test/java/net/minecraftforge/debug/NetworkCapabilityTest.java
```diff
@@ -0,0 +1,110 @@
+package net.minecraftforge.debug;
```
ChampionAsh5357, line 20: Missing license header.

src/test/java/net/minecraftforge/debug/NetworkCapabilityTest.java
```diff
@@ -0,0 +1,110 @@
...
+import net.minecraftforge.common.util.LazyOptional;
+import net.minecraftforge.event.AttachCapabilitiesEvent;
+import net.minecraftforge.event.entity.living.LivingEntityUseItemEvent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+
+@Mod(NetworkCapabilityTest.ID)
+@Mod.EventBusSubscriber(bus = Mod.EventBusSubscriber.Bus.FORGE, modid = NetworkCapabilityTest.ID)
+public class NetworkCapabilityTest {
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/NetworkCapabilityTest.java
```diff
@@ -0,0 +1,110 @@
...
+        event.getEntity().getCapability(TEST_CAP).ifPresent(cap ->
+        {
+            if (!event.getEntity().level.isClientSide())
+            {
+                cap.update();
+                logger.info("Set new number: " + cap.number);
+            }
+        });
+    }
+}
```
ChampionAsh5357: New line at EOF.

ChampionAsh5357 (commented):
So just general questions and comments while doing a quick skim through:

1. Why is the block entity capabilities its own packet? Why not just enforce or make default saving the data to the update packet? Seems unnecessary to split the packet data here since the two places the packet is sent is where the block entity logic is patched.
2. Is the handling code for `ForgeNetwork#send*Capabilities` necessary, or can it simply be abstracted using `SimpleChannel#send` and supplying the `PacketDistributor` with the necessary trackers?

Random Thoughts:

- There probably should be a distinction between initial synchronization and consistent synchronization. For example, I may have a capability that holds static data that I may only want to send on the initial sync. I can imagine this being simple for entities and block entities, not sure about itemstacks.
- There should be a test mod for block entities, entities, and itemstacks. This is a PR that could potentially cause a lot of issues, so there should be test mods to verify no bugs or edge cases arise.

josephtarbit:
> 1. Why is the block entity capabilities its own packet? Why not just enforce or make default saving the data to the update packet? Seems unnecessary to split the packet data here since the two places the packet is sent is where the block entity logic is patched.

Because block entities can return a `null` update packet which would mean no caps would be able to synch data. In addition, it would mean serializing the entire BE every time a cap was synched.

> 2. Is the handling code for `ForgeNetwork#send*Capabilities` necessary, or can it simply be abstracted using `SimpleChannel#send` and supplying the `PacketDistributor` with the necessary trackers?

What do you mean by this? Create a custom `PacketDistributor`?

> * There probably should be a distinction between initial synchronization and consistent synchronization. For example, I may have a capability that holds static data that I may only want to send on the initial sync. I can imagine this being simple for entities and block entities, not sure about itemstacks.

If a cap returns `false` from `requiresSync` it will only ever be synched upon initial synchronization. If it needs synching subsequent times, you return `true`.

> * There should be a test mod for block entities, entities, and itemstacks. This is a PR that could potentially cause a lot of issues, so there should be test mods to verify no bugs or edge cases arise.

Yeah that's good point, I'll add those.

(P.S. I appreciate giving this PR some attention)

ChampionAsh5357:
> Because block entities can return a null update packet which would mean no caps would be able to synch data. In addition, it would mean serializing the entire BE every time a cap was synched.

The update packet is for syncing data as required for a block entity. It does not serialize the entire block entity, only the position where its located and the type for validation. The tag only includes data the player would like to send. Additionally, `#onDataPacket` by default reads forge caps, so as long as you create the packet when null or add onto the compound tag when not, it should be fine from my understanding.

> What do you mean by this? Create a custom PacketDistributor?

There should already be a `PacketDistributor` that sends packets when tracking a certain entity or tracking a specific chunk which should make it unnecessary to write the distribution methods yourself.

> If a cap returns false from requiresSync it will only ever be synched upon initial synchronization. If it needs synching subsequent times, you return true.

Ah, I seemed to have missed that while reading though. Thanks for the explanation.

josephtarbit:
> The update packet is for syncing data as required for a block entity. It does not serialize the entire block entity, only the position where its located and the type for validation. The tag only includes data the player would like to send. Additionally, `#onDataPacket` by default reads forge caps, so as long as you create the packet when null or add onto the compound tag when not, it should be fine from my understanding.

Block entities attach additional data to the `ClientboundBlockEntityDataPacket` though (usually via `getUpdateTag`), so if you wanted to synchronize caps you'd always be sending additional data alongside them. 

> There should already be a `PacketDistributor` that sends packets when tracking a certain entity or tracking a specific chunk which should make it unnecessary to write the distribution methods yourself.

Ahh, I see what you mean. The reason is to check whether a specific client is a vanilla client so we don't send caps to them.
The methods also perform boiler plate code for sending caps such as constructing a `FriendlyByteBuf`, writing the caps to the buf and in the case of container slot capabilities - filters out equipment slots.

ChampionAsh5357:
> Block entities attach additional data to the ClientboundBlockEntityDataPacket though (usually via getUpdateTag), so if you wanted to synchronize caps you'd always be sending additional data alongside them.

Sure, but you only sending the packets when `#getUpdatePacket` is referenced anyways. So, the only difference is sending two packets instead of one. The only use case I could see where this might be an issue is if the block entity is already syncing a capability outside of the handler which should never happen.

josephtarbit:
> > Block entities attach additional data to the ClientboundBlockEntityDataPacket though (usually via getUpdateTag), so if you wanted to synchronize caps you'd always be sending additional data alongside them.
> 
> Sure, but you only sending the packets when `#getUpdatePacket` is referenced anyways. So, the only difference is sending two packets instead of one. The only use case I could see where this might be an issue is if the block entity is already syncing a capability outside of the handler which should never happen.

Yeah, I suppose my argument is only justified if you wanted to synchronize a capability after the initial block entity synchronization.

src/main/java/net/minecraftforge/common/capabilities/INetworkCapability.java
josephtarbit: You wouldn't really be able to impose a size limit here unless you knew how much data was in the rest of the packet, assuming the caller of this even wants to put the data in a packet as it's not tied specially to that purpose.

ChampionAsh5357: True enough, it could be limited to the max packet size, but at that point it really wouldn't matter since the erroring would always appear at the same place later down the line anyways. It's ore just for throwing errors near the 'correct' call site.

ChampionAsh5357:
It seems all of your formatting changes have been removed.

src/test/java/net/minecraftforge/debug/NetworkCapabilityTest.java
```diff
@@ -0,0 +1,173 @@
...
+@Mod(NetworkCapabilityTest.ID)
+@Mod.EventBusSubscriber(bus = Mod.EventBusSubscriber.Bus.FORGE, modid = NetworkCapabilityTest.ID)
+public class NetworkCapabilityTest
+{
+    public static final String ID = "network_capability_test";
+
+    private static final Logger logger = LogManager.getLogger();
+
+    private static final Capability<TestCapability> TEST_CAP = CapabilityManager.get(new CapabilityToken<>() {
+    });
```
ChampionAsh5357: This can be one line.

src/test/java/net/minecraftforge/debug/NetworkCapabilityTest.java
```diff
@@ -0,0 +1,173 @@
...
+    {
+        FMLJavaModLoadingContext.get().getModEventBus().addListener(NetworkCapabilityTest::handleRegisterCapabilities);
+    }
+
+    private static void handleRegisterCapabilities(RegisterCapabilitiesEvent event)
+    {
+        event.register(TestCapability.class);
+    }
+
+    private static class TestCapability {
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/NetworkCapabilityTest.java
```diff
@@ -0,0 +1,173 @@
...
+        private boolean dirty;
+
+        public void update()
+        {
+            this.number++;
+            this.dirty = true;
+        }
+    }
+
+    private static class TestCapabilityProvider implements ICapabilityProvider, INetworkCapability {
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/NetworkCapabilityTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+    @SubscribeEvent
+    public static void handleAttachBlockEntityCapabilities(AttachCapabilitiesEvent<BlockEntity> event)
+    {
+        event.addCapability(new ResourceLocation(ID, "test_cap"),
+            new TestCapabilityProvider(() -> event.getObject().getBlockState().getBlock().getName().getString()));
+    }
+
+    @SubscribeEvent
+    public static void handleLivingEntityUseItem(LivingEntityUseItemEvent event)
```
ChampionAsh5357: Bad test event, use one of the sub events. It gets difficult figuring out how the capability is updating when it could changed on start, tick, and stop.

src/test/java/net/minecraftforge/debug/NetworkCapabilityTest.java
ChampionAsh5357: Should be serializable to test initial sync.

ChampionAsh5357 (changes requested):
In creative, it seemed like any nbt data that was loaded into a gui on a stack was immediately discarded and synced with the initial value.

For replication purposes:
1. Took out mundane potion from creative inventory.
2. Increased capability value by using item.
3. Open gui again to observe initial value being synced to client.
4. Verify data isn't from gui by using item to increase capability again.

src/test/java/net/minecraftforge/debug/NetworkCapabilityTest.java
josephtarbit: I assume you mean `INBTSerializable`?

ChampionAsh5357: That, or you could use `ICapabilitySerializable` over `ICapabilityProvider`

josephtarbit:
> In creative, it seemed like any nbt data that was loaded into a gui on a stack was immediately discarded and synced with the initial value.

I believe this is because the capability was not serializable (calls to `ItemStack#copy` will not copy the cap data).
This now works after making it serializable.

ChampionAsh5357:
> I believe this is because the capability was not serializable (calls to ItemStack#copy will not copy the cap data).

That seems like a bug that data would get reset if the capability is not serializable. The syncing of an item should not be determined by whether or not the capability is serializable.

josephtarbit:
> > I believe this is because the capability was not serializable (calls to ItemStack#copy will not copy the cap data).
> 
> That seems like a bug that data would get reset if the capability is not serializable. The syncing of an item should not be determined by whether or not the capability is serializable.

I wouldn't necessarily say it's a bug, just the nature of the way item capabilities work. If you want `ItemStack#copy` to copy your capability data you need a serializable capability, which is unrelated to synchronization.

Waterpicker:
I'm curious about how much longer this PR has in triage hell before its gets merged. Networking for capabilities is something I'm interested in.

Shadows-of-Fire:
Probably a while... it requires that it be merged in 1.19.x first and it's fairly involved, which means someone has to heavily test it.

tmvkrpxl0:
> I wouldn't necessarily say it's a bug, just the nature of the way item capabilities work. If you want `ItemStack#copy` to copy your capability data you need a serializable capability, which is unrelated to synchronization.

I still believe this should be addressed regardless. This can be seen as replacement for capability sync logic that modders implemented themselves with custom packets. and it isn't clear that it needs to be INBTSerializable.
Also, it is better to have capability for item that is not saved to disc but still synced across network.

PR #8225, opened 2021-11-30
[1.18] Integrate the gametest framework with Forge

SizableShrimp:
1.18 port of #8024 (see for more details)

Differences from 1.17 version:
* `GameTestMain` is no longer a copy of `server.Main`, it now sets `forge.gametest` and delegates to `server.Main` (with patching)
* License header updates are no longer included
* `--acceptEula` has been removed to limit unnecessary patching
* A new `ForgeGameTestHooks` class is in use to prevent issues with registry being loaded before bootstrapping when classloading `ForgeHooks` from `Main`


patches/minecraft/net/minecraft/client/Minecraft.java.patch
```diff
@@ -67,10 +67,11 @@
        this.f_91064_ = new DebugRenderer(this);
        RenderSystem.m_69900_(this::m_91113_);
        if (this.f_91042_.f_83915_ == this.f_90990_.m_85441_() && this.f_91042_.f_83916_ == this.f_90990_.m_85442_()) {
-@@ -529,6 +_,10 @@
+@@ -529,6 +_,11 @@
           TinyFileDialogs.tinyfd_messageBox("Minecraft", stringbuilder.toString(), "ok", "error", false);
        }
  
++      net.minecraftforge.gametest.ForgeGameTestHooks.registerGametests(java.util.Set.of());
```
sciwhiz12: [`Set#of()`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Set.html#of()) returns an unmodifiable set, which would mean that mods which try to add to the (documented as mutable) set of enabled namespaces during their `RegisterGameTestsEvent` will cause an error.

src/main/java/net/minecraftforge/gametest/BlockPosValueConverter.java
```diff
@@ -0,0 +1,22 @@
+package net.minecraftforge.gametest;
```
sciwhiz12, line 20: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/gametest/ForgeGameTestHooks.java
```diff
@@ -0,0 +1,23 @@
+package net.minecraftforge.gametest;
```
sciwhiz12, line 20: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/gametest/GameTestMain.java
```diff
@@ -0,0 +1,12 @@
+package net.minecraftforge.gametest;
```
sciwhiz12, line 20: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/server/ServerLifecycleHooks.java
```diff
@@ -97,7 +99,12 @@ public static boolean handleServerAboutToStart(final MinecraftServer server)
 
     public static boolean handleServerStarting(final MinecraftServer server)
     {
-        DistExecutor.runWhenOn(Dist.DEDICATED_SERVER, ()->()->LanguageHook.loadLanguagesOnServer(server));
+        DistExecutor.runWhenOn(Dist.DEDICATED_SERVER, ()->()->{
+            LanguageHook.loadLanguagesOnServer(server);
+            // GameTestServer requires the gametests to be registered earlier, so it is done in main and should not be done twice.
+            if (!(server instanceof GameTestServer))
+                net.minecraftforge.gametest.ForgeGameTestHooks.registerGametests(Set.of());
```
sciwhiz12: [`Set#of()`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Set.html#of()) returns an unmodifiable set, which would mean that mods which try to add to the (documented as mutable) set of enabled namespaces during their `RegisterGameTestsEvent` will cause an error.

src/test/java/net/minecraftforge/debug/misc/GameTestTest.java
```diff
@@ -0,0 +1,116 @@
...
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+
+import java.util.List;
+
+@Mod(GameTestTest.MODID)
+public class GameTestTest
+{
+    public static final String MODID = "gametest_test";
+    public static final boolean ENABLE = true;
```
sciwhiz12: Nitpick: the conventional name for this field is `ENABLED`.

patches/minecraft/net/minecraft/gametest/framework/GameTest.java.patch
```diff
@@ -0,0 +1,15 @@
+--- a/net/minecraft/gametest/framework/GameTest.java
++++ b/net/minecraft/gametest/framework/GameTest.java
+@@ -16,6 +_,12 @@
+ 
+    boolean m_177045_() default true;
+ 
++   /** The namespace of where to grab the structure from, generally a mod id. */
++   String templateNamespace() default "minecraft";
++
++   /** Whether to prefix the template structure with the classname and a period. */
```
sciwhiz12: In the interests of modder convenience (because the Forge test mods are not accessible from their IDEs), please include an example here of this prefixing behavior.

src/test/java/net/minecraftforge/debug/misc/GameTestTest.java
```diff
@@ -0,0 +1,116 @@
...
+        BlockPos woodPos = new BlockPos(1, 1, 1);
+
+        // assertBlockState will convert the relative woodPos into a real world block position and check it with the predicate.
+        // Relative positions are made absolute by adding their value to the block position of the structure tile entity,
+        // which is always the lowest northwest corner inside the structure.
+        // If the predicate fails, the String supplier will be used to construct an exception message, which is thrown
+        helper.assertBlockState(woodPos, b -> b.is(Blocks.OAK_LOG), () -> "Block was not an oak log");
+
+        // If we got to this point, the assert succeeded, so the test has succeeded
+        helper.succeed();
```
sciwhiz12: We should possibly note that gametests operate differently from JUnit tests: where JUnit tests imply that no failure means a successful test, a gametest must explicitly declare its success, otherwise it will fail due to the timeout.

patches/minecraft/net/minecraft/server/Main.java.patch
```diff
@@ -66,9 +76,39 @@
...
+@@ -175,22 +_,33 @@
+ 
+          levelstoragesource$levelstorageaccess.m_78287_(registryaccess$registryholder, worlddata);
+          WorldData worlddata1 = worlddata;
+-         final DedicatedServer dedicatedserver = MinecraftServer.m_129872_((p_129697_) -> {
+-            DedicatedServer dedicatedserver1 = new DedicatedServer(p_129697_, registryaccess$registryholder, levelstoragesource$levelstorageaccess, packrepository, serverresources, worlddata1, dedicatedserversettings, DataFixers.m_14512_(), minecraftsessionservice, gameprofilerepository, gameprofilecache, LoggerChunkProgressListener::new);
++         final MinecraftServer dedicatedserver = MinecraftServer.m_129872_((p_129697_) -> {
++            MinecraftServer dedicatedserver1;
++            if (gametestEnabled) {
++               net.minecraftforge.gametest.ForgeGameTestHooks.registerGametests(java.util.Set.copyOf(optionset.valuesOf(enableNamespaceOpt)));
```
sciwhiz12: [`Set#copyOf(Collection)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Set.html#copyOf(java.util.Collection)) returns an unmodifiable set, which would mean that mods which try to add to the (documented as mutable) set of enabled namespaces during their `RegisterGameTestsEvent` will cause an error.

patches/minecraft/net/minecraft/server/Main.java.patch
```diff
@@ -1,11 +1,21 @@
...
++      OptionSpec<net.minecraft.core.BlockPos> spawnPosOpt;
++      OptionSpec<String> enableNamespaceOpt;
++      boolean gametestEnabled = net.minecraftforge.gametest.ForgeGameTestHooks.isGametestEnabled();
++      if (gametestEnabled) {
++         spawnPosOpt = optionparser.accepts("spawnPos").withRequiredArg().withValuesConvertedBy(new net.minecraftforge.gametest.BlockPosValueConverter()).defaultsTo(new net.minecraft.core.BlockPos(0, 60, 0));
++         enableNamespaceOpt = optionparser.accepts("enableNamespace", "Adds an enabled namespace. Only gametests with templates using an enabled namespace will be loaded. If none are added, all gametests are enabled.").withRequiredArg().withValuesSeparatedBy(",");
++      } else {
++          spawnPosOpt = null;
++          enableNamespaceOpt = null;
++      }
```
sciwhiz12: Why are these in the `else` block, rather than in the initializer for the variable declarations?

sciwhiz12 (changes requested):
Ran the two provided gametests, and they seem to work fine to me. I do have two points:

- `forge.enablegametest` should be mentioned or, better yet, enabled by default in the run configs for the MDK.
- We should probably clarify that the `gametest` run is for the `GameTestServer` and automatic running of gametests, and that the above property when included in the regular client/server run configurations enable the `/test` command for manual running and creation of gametests.

patches/minecraft/net/minecraft/server/Main.java.patch
SizableShrimp: They are used in a lambda

patches/minecraft/net/minecraft/gametest/framework/GameTest.java.patch
SizableShrimp: I have updated this documentation, but it could use review and possibly better wording.

src/main/java/net/minecraftforge/event/RegisterGameTestsEvent.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * Returns an immutable view of all registered gametest methods.
+     *
+     * @return an immutable view of all registered gametest methods
+     */
+    public Set<Method> getGameTestMethods() {
+        return this.gameTestMethodsView;
+    }
```
sciwhiz12: I'm not sure if we should be exposing the registered methods as public API, as I'd personally prefer this to be made an implementation detail. 
You could refactor this event such that the caller of the constructor passes in a mutable set controlled by that caller, which this event add the methods to.
(If you decide against this suggestion, you'll need to resolve the brace being on the same line. :shipit:)

src/main/java/net/minecraftforge/event/RegisterGameTestsEvent.java
```diff
@@ -0,0 +1,102 @@
...
+     * If a game test's template does not have a namespace inside this set, it will not be registered.
+     * An empty set means all namespaces are enabled.
+     * By default, this can only be non-empty when running through {@link GameTestServer}.
+     *
+     * @return the mutable set of enabled namespaces
+     */
+    public Set<String> getEnabledNamespaces()
+    {
+        return enabledNamespaces;
+    }
```
sciwhiz12: After a bit of pondering, I don't think giving mods free access to the mutable set of enabled namespaces is a good thing (or, at least, probably unacceptable to the maintainers). Should we have this be an unmodifiable view of the set instead, and provide `enableNamespace`/`disableNamespace` (or other-named equivalents) methods instead?

src/main/java/net/minecraftforge/gametest/ForgeGameTestHooks.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @SuppressWarnings("deprecation")
+    public static void registerGametests(Set<String> enabledNamespaces)
+    {
+        if (isGametestEnabled())
+        {
+            RegisterGameTestsEvent event = new RegisterGameTestsEvent(enabledNamespaces);
+            ModLoader.get().postEvent(event);
+            for (Method gameTestMethod : event.getGameTestMethods()) {
+                GameTestRegistry.register(gameTestMethod, event.getEnabledNamespaces());
```
sciwhiz12: Should we have this set of enabled namespaces be defensively copied first (using `Set.copyOf`) into a local variable before this `for`-loop?

sciwhiz12 (changes requested):
A few more things to possibly change, and this looks fine to me for assignment.

src/main/java/net/minecraftforge/gametest/ForgeGameTestHooks.java
SizableShrimp: I personally find this unnecessary as forge itself should control all calls to this hook method. Any calls from modders are a mistake and shouldn't be accounted for.

patches/minecraft/net/minecraft/server/Main.java.patch
```diff
@@ -1,11 +1,21 @@
...
-@@ -76,6 +_,8 @@
+@@ -76,6 +_,18 @@
        OptionSpec<String> optionspec12 = optionparser.accepts("serverId").withRequiredArg();
        OptionSpec<Void> optionspec13 = optionparser.accepts("jfrProfile");
        OptionSpec<String> optionspec14 = optionparser.nonOptions();
 +      optionparser.accepts("allowUpdates").withRequiredArg().ofType(Boolean.class).defaultsTo(Boolean.TRUE); // Forge: allow mod updates to proceed
 +      optionparser.accepts("gameDir").withRequiredArg().ofType(File.class).defaultsTo(new File(".")); //Forge: Consume this argument, we use it in the launcher, and the client side.
++      OptionSpec<net.minecraft.core.BlockPos> spawnPosOpt;
++      OptionSpec<String> enableNamespaceOpt;
++      boolean gametestEnabled = net.minecraftforge.gametest.ForgeGameTestHooks.isGametestEnabled();
```
sciwhiz12: While these are effectively final because they're used in a lambda, I'd like them to be marked `final` to clarify that.

src/main/java/net/minecraftforge/gametest/ForgeGameTestHooks.java
```diff
@@ -0,0 +1,51 @@
...
+
+public class ForgeGameTestHooks
+{
+    public static boolean isGametestEnabled()
+    {
+        return SharedConstants.IS_RUNNING_IN_IDE || Boolean.getBoolean("forge.enablegametest");
+    }
+
+    @SuppressWarnings("deprecation")
+    public static void registerGametests(Set<String> enabledNamespaces)
```
sciwhiz12: For informational purposes, the set of enabled namespaces should be logged (either here, or at least in `server.Main`).

sciwhiz12 (changes requested):
Two very minor quibbles popped up in my mind while I was doing some final testing.

src/main/java/net/minecraftforge/gametest/ForgeGameTestHooks.java
SizableShrimp: Need more info on the preferred logging level and logger to use for this

sciwhiz12: Because the message should only be logged when the gametests are enabled (via the system property), the logging level should be `INFO`. For the logger, create a new (`private static final`) one for `ForgeGameTestHooks`.

sciwhiz12 (approved):
Looks fine to me. :shipit:

dpeter99:
The code itself looks good to me as well.
But as mentioned by @TheCurle in discord, more examples would help show how modders could use this.
Showing how a modded block can be tested and how capabilities can be checked would add a lot to this.
Here are a few suggestions for test that might add tho it:
- Make a test that tests if a hopper picks up an item. (To show how to check for inventory contents)
- Another one to check if lava dripping is working (To show long running tests)
- And I guess make a simple Block Entity with a capability that... "Emulates" an energy system by counting up and using up a coal.... (To show a more complex case with a custom cap)

TheCurle:
I'll make these tests as i come around to testing the viability of this API extension.

As of right now, i don't want to expose the Game Test system unless it's at least baseline usable for modded items, including reading capability data and custom blockstate properties.

SizableShrimp:
I have a added a new container-related gametest involving dropping a golden apple onto a hopper. `GameTestHelper` contains some nice util for both spawning an item and waiting a certain amount of ticks before checking a container for the existence of an item. One thing that may be considered is patching `GameTestHelper#assertContainerContains` to instead allow a count of anything one or higher for a given item. Currently, only a count of exactly one is supported, otherwise it throws an assertion exception.

I do not agree with adding a long-running test dependent on random ticks. Tests should be generally fast to run so that they can be run in the future to quickly and easily tell if any forge compatibility/functionality has been broken. I believe anything that takes longer than a few seconds to run like random ticks should not be made into a gametest, and more consistent tests should be substituted instead.

Mysterious-Dev (approved):

Mysterious-Dev:
Any news on that?

robotgryphon:
A couple of notes from some initial testing:

- Usage of MOD event bus rather than FORGE is a bit strange.
- Echoing feedback from Shrimp that adding a new class-level annotation (I'd like `ModGameTests` or his `GameTestHolder` suggestion) would eliminate a ton of duplicated code, even though it isn't exactly how vanilla does it
- `prefixTemplateWithClassname` should probably default to `false` rather than true

Something else I'd like to see is splitting the structure info from `GameTest` and instead putting it in `GameTestTemplate` annotation, as it's a bit more readable to me.

```java
public class CodecTests {
    @GameTest(templateNamespace = MOD_ID, prefixTemplateWithClassname = false, template = "empty_5x")
    public static void testCodec(GameTestHelper test) {
        test.succeed();
    }
 }
 ```
 
 vs, assuming a couple of changes:
 
 ```java
 // maybe @GameTestHolder here?
@GameTestTemplate(namespace = MOD_ID)
public class CodecTests {

    @GameTest
    @GameTestTemplate("empty_5x")
    public static void testCodec(GameTestHelper test) {
        test.succeed();
    }
}
```

TheCurle:
We've been talking about this a lot internally.
@robotgryphon has used this live for testing their mod Compact Crafting and has given us some amazing feedback.

With that, i'm going to pull this. 
This is an amazing tool that Mojang have given us, and now we're giving it to modders.

We'll prepare a guide for using this system, and it'll be posted in all the usual places once it's ready.

PR #8229, opened 2021-11-30
[1.18] Adds hook for entity EXP value, fixes mutable experience getter in Zombie, readds missing LivingExperienceDropEvent 

baileyholl:
This is a port of the previous 1.16 version of this PR: https://github.com/MinecraftForge/MinecraftForge/pull/7852

This PR exposes a way for modders to obtain the 'true' experience value of a mob. Currently, shouldDropExperience and getExperienceReward are protected methods. While these could be set to public, this method runs all of the expected logic modders would typically need to do themselves in order to conform to vanilla EXP rules, including a call to the getExperienceDrop forge event.

Additionally, a patch is made to Zombie#getExperienceReward as the existing method mutates and increases the xpReward each time the method is called.

**Adds getTotalExperienceValue to LivingEntity**
Exposes a way for modders to obtain the exp value of a mob.

**Patches Zombie#getExperienceReward**
Changes the method to safely return the exp value of the entity without increasing it each call.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
@@ -600,9 +600,9 @@
 +    * @return The amount of experience
 +    */
 +   public int getTotalExperienceValue(Player player){
-+      if(!this.m_6149_())
++      if(!m_6149_())
 +         return 0;
-+      int i = this.m_6552_(player);
++      int i = m_6552_(player);
 +      return net.minecraftforge.event.ForgeEventFactory.getExperienceDrop(this, player, i);
     }
  }
--- a/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
@@ -611,7 +611,7 @@
 +   }
 +
 +   @Override
-+   public int getTotalExperienceValue(Player player){
++   public int getTotalExperienceValue(Player player) {
 +      if(!m_6149_() && !m_6124_())
 +         return 0;
 +      int i = m_6552_(player);
--- a/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
@@ -5,7 +5,7 @@
  import net.minecraft.world.scores.PlayerTeam;
  
 -public abstract class LivingEntity extends Entity {
-+public abstract class LivingEntity extends Entity implements net.minecraftforge.common.extensions.IForgeLivingEntity{
++public abstract class LivingEntity extends Entity implements net.minecraftforge.common.extensions.IForgeLivingEntity {
     private static final UUID f_20929_ = UUID.fromString("662A6B8D-DA3E-4C1C-8813-96EA6097278D");
     private static final UUID f_20959_ = UUID.fromString("87f46a96-686f-4796-b035-22e16ee9e038");
     private static final UUID f_147184_ = UUID.fromString("1eaf83ff-7207-4596-b37a-d7a07b3ec4ce");
--- a/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
@@ -588,3 +588,21 @@
           }
        } else {
           return EquipmentSlot.HEAD;
+@@ -3287,5 +_,17 @@
+    }
+ 
+    public static record Fallsounds(SoundEvent f_196626_, SoundEvent f_196627_) {
++   }
++
++   /**
++    * Exposes the experience value of the entity after the getExperienceDrop event.
++    * @param player The attacking player involved in obtaining this experience, typically lastHurtByPlayer.
++    * @return The amount of experience
++    */
++   public int getTotalExperienceValue(Player player){
++      if(!this.m_6149_())
++         return 0;
++      int i = this.m_6552_(player);
++      return net.minecraftforge.event.ForgeEventFactory.getExperienceDrop(this, player, i);
+    }
+ }
--- a/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
@@ -1,6 +1,11 @@
 --- a/net/minecraft/world/entity/LivingEntity.java
 +++ b/net/minecraft/world/entity/LivingEntity.java
-@@ -118,7 +_,9 @@
+@@ -114,11 +_,13 @@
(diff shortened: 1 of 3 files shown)


patches/minecraft/net/minecraft/world/entity/monster/Zombie.java.patch
```diff
@@ -1,5 +1,18 @@
...
+@@ -156,11 +_,7 @@
+    }
+ 
+    protected int m_6552_(Player p_34322_) {
+-      if (this.m_6162_()) {
+-         this.f_21364_ = (int)((float)this.f_21364_ * 2.5F);
+-      }
+-
+-      return super.m_6552_(p_34322_);
++      return this.m_6162_() ? (int) (super.m_6552_(p_34322_) * 2.5F) : super.m_6552_(p_34322_);
```
ChampionAsh5357: This logic is incompatible with vanilla. Looking at `Mob#getExperienceReward`, the experience is first cached than added onto based on armor and weapons. This means that the baby multiplier would then be apply 2.5x the amount of the total experience gained instead of just the zombie experience.

patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
```diff
@@ -588,3 +588,21 @@
...
++    * Exposes the experience value of the entity after the getExperienceDrop event.
++    * @param player The attacking player involved in obtaining this experience, typically lastHurtByPlayer.
++    * @return The amount of experience
++    */
++   public int getTotalExperienceValue(Player player){
++      if(!m_6149_())
++         return 0;
++      int i = m_6552_(player);
++      return net.minecraftforge.event.ForgeEventFactory.getExperienceDrop(this, player, i);
+    }
```
ChampionAsh5357: Should be declared within a forge extension and then overridden imo.

ChampionAsh5357 (changes requested):
So, I think I may be missing something, where is `#getTotalExperienceValue` called? Currently, this would mean that `LivingExperienceDropEvent` would only be executed on this forge added method which isn't patched anywhere. It doesn't seem like `LivingExperienceDropEvent` is called anywhere at the moment anyways, so that should probably also be handled by this PR.

baileyholl:
Re getTotalExperienceValue being called, this is an additional method being added so that it can be called by mod authors, as currently there is no publicly accessible way for mod authors to obtain the EXP value of a mob before its death. This is a way for authors to access that value without modifying the signatures of getExperienceReward and shouldDropExperience, both of which are marked as Protected.

As for the LivingExperienceDropEvent not being used, it appears to have previously been used in 1.16 firing before the EXP orbs are dropped. Mojang has moved the orb spawning logic around, but I can add this event hook back as well to maintain the same behaviors as in 1.16.

ChampionAsh5357:
Yeah, my comment was in response to a forge added method being the only reference point for a forge added event. I don't see any issue in the method itself.

patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
baileyholl: I have now created an IForgeLivingEntity extension class, modeled after IForgeEntity. I have moved this method there, and implemented the class in LivingEntity with this method.

patches/minecraft/net/minecraft/world/entity/monster/Zombie.java.patch
baileyholl: I have changed this to now use a temporary variable to restore the xpReward field after the super method is called. I am not happy with this solution, but to me it is the most clear as to what is happening, while preserving the currently intended vanilla mechanic.

```java
protected int getExperienceReward(Player p_34322_) {
      if (this.isBaby()) {
         int oldReward = this.xpReward;
         this.xpReward = (int)((float)this.xpReward * 2.5F);
         int reward = super.getExperienceReward(p_34322_);
         this.xpReward = oldReward;
         return reward;
      }

      return super.getExperienceReward(p_34322_);
   }
```

If we are unhappy with this temp variable assigment, an equivalent EXP calculation can be made by recognizing that 1x of the xpReward is accounted for in the super call, and we can just add the difference back after the calculation without assignment, and without an unintended increase from the equipment caculations in the super call.

```java
protected int getExperienceReward(Player p_34322_) {
      if (this.isBaby()) {
         return super.getExperienceReward(p_34322_) + (int) this.xpReward * 1.5F;
      }

      return super.getExperienceReward(p_34322_);
   }
```

baileyholl:
I have now restored the forge getExperienceDrop event hook located in LivingEntity#dropExperience as it was in 1.16

patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
```diff
@@ -588,3 +604,17 @@
...
           return EquipmentSlot.HEAD;
+@@ -3287,5 +_,13 @@
+    }
+ 
+    public static record Fallsounds(SoundEvent f_196626_, SoundEvent f_196627_) {
++   }
++
++   @Override
++   public int getTotalExperienceValue(Player player){
++      if(!m_6149_() && !m_6124_())
```
ChampionAsh5357, line 615: Can you not get the total experience value of another player? I'm wondering the additional condition that they must not always drop experience has to do with the total experience value they drop.

baileyholl: You cannot get the total experience value of another player from the method, but the total levels is exposed, but not helpful for what the actual raw exp value is.

ChampionAsh5357: Should be fine then.

ChampionAsh5357 (approved):

LexManos:
This has failed to answer the "Why?" question. This needs more justification. Also the experience can change all the time. Do any call before they die would potentially be incorrect.

baileyholl:
> This has failed to answer the "Why?" question. This needs more justification. Also the experience can change all the time. Do any call before they die would potentially be incorrect.

Currently, it is impossible for modders to obtain the experience value of a mob pre-death within forge + mc. The xpReward field is not used consistently anywhere and is not suitable for access, and MC overrides the method that uses this field quite often in order to return a random range of EXP for animals. These EXP value methods do all of the hard work of bonus calculations for geared monsters as well. 

The use case for this is within my own mod. I have an in-game mechanic for the player to passively acquire EXP based on the entities nearby, and this is done without killing the mobs, as the intention is to provide an alternative to mob farming and killing.

The EXP value before death should be sufficient even with slight variation as this randomness is a well documented minecraft mechanic, and cross-mod support is still provided as these methods still fire the getExperienceDrop event. These mods would have no reason to mutate xpReward directly from this method either, as again, that field is not reliable to use even within Minecraft.

sciwhiz12 (changes requested):
- `src/main/java/net/minecraftforge/common/extensions/IForgeLivingEntity.java`

PR #8231, opened 2021-12-01
Adds tag support for MobEffects

bageldotjpg:
MobEffects feel like an obvious inclusion alongside Enchantments, Potions, and BlockEntityType tags that Forge already adds support for. 
This has several possible use cases, such as defining a list of effects a mob is immune to or affected by, categorizing similar effects for compatibility, etc.


patches/minecraft/net/minecraft/world/effect/MobEffect.java.patch
```diff
@@ -17,12 +17,22 @@
...
 +
 +   // FORGE START
++   private final net.minecraftforge.common.util.ReverseTagWrapper<MobEffect> reverseTags = new net.minecraftforge.common.util.ReverseTagWrapper<>(this, () -> net.minecraft.tags.SerializationTags.m_13199_().m_144452_(Registry.f_122900_));
++
++   public java.util.Set<net.minecraft.resources.ResourceLocation> getTags() {
++      return reverseTags.getTagNames();
++   }
++
++   public boolean is(net.minecraft.tags.Tag<MobEffect> tag) {
++      return tag.m_8110_(this);
```
eerussianguy, line 33: These patches can be smaller if you move the braces up all on one line

gigaherz: No need to one-line things. what we want is to minimize the *complexity* of the patches, not the length per se.

eerussianguy: Ah, I had had a similar PR and someone told me this so that is good to know

SmellyModder (approved):

ChampionAsh5357 (approved):

PR #8233, opened 2021-12-01
1.18: Basic Animation Modifier API

JTK222:
Currently, it is not possible to alter animations of entities in any way.
This API is a very rudimentary implementation of such an API.

The main goal right now is to allow players to alter the way entities hold their Items.
Therefore, this API will only be implemented for Humanoid entities, all other ones will be made in a separate one.
(Many other ones will require a large amount of extra patches, those can be avoided if there is no demand for that)

Additionally, the API itself is designed to hopefully stay future-proof, even if entity models become data driven.

A test mod which demonstrates the usage is in the works.


JTK222:
Test Mod has been added.
Here is an image of it's result (& code bellow)
image
```java
public void registerLayersEvent(EntityRenderersEvent.AddLayers event) {
    LivingEntityRenderer[] renderers = {
         event.getSkin("default"),
            event.getSkin("slim")
     };

    for (LivingEntityRenderer renderer : renderers) {
        Model m = renderer.getModel();

       if (m instanceof IForgeEntityModel fModel){
         fModel.addAnimationModifier((entity, modelPart) -> {
             ModelPart leftArm = modelPart.getChild("left_arm");
                leftArm.xRot = (float)Math.PI;
            });
        }
    }
}
```

Of course the above code could be expanded to make custom item use animations and similar.
I'll only add the limb motion parameters and partial ticks to the lambda method and this PR has it's functionality

src/main/java/net/minecraftforge/client/extensions/IForgeEntityModel.java
```diff
@@ -0,0 +1,31 @@
+package net.minecraftforge.client.extensions;
```
sciwhiz12, line 1: Missing license header.

src/main/java/net/minecraftforge/client/extensions/IForgeEntityModel.java
```diff
@@ -0,0 +1,31 @@
...
+    default void prepareAnim(T p_102618_, float p_102619_, float p_102620_, float p_102621_, float p_102622_, float p_102623_)
+    {
+        setupAnim(p_102618_, p_102619_, p_102620_, p_102621_, p_102622_, p_102623_);
+        ModelPart root = getModelRoot();
+
+        for (IEntityModelAnimationModifier modifier : animationModifiers)
+            modifier.modifyAnimations(p_102618_, root, p_102619_, p_102620_, p_102621_, p_102622_, p_102623_);
+    }
+
+    void setupAnim(T p_102618_, float p_102619_, float p_102620_, float p_102621_, float p_102622_, float p_102623_);
```
sciwhiz12, line 23: Please give these parameters proper names befitting their purpose.

sciwhiz12, line 23: This method being named such that it is implemented by the same-named MC methods in `EntityModel` may be a bit iffy on the reobfuscation. My thinking is that, because of reobfuscation of MC methods, the corresponding method in `HumanoidModel` will be reobfuscated back into a SRG name and therefore will not be implementing the method defined by this interface, causing a crash when the JVM verifies the class.

If you are able, please test this in a userdev environment by making a mod with your PR and running it in production. Tse `publishToMavenLocal` to publish the necessary artifacts needed by userdev and then use the version indicated in the Gradle console output. Use `installerJar` to build the installer JAR so you can test it in production.

src/main/java/net/minecraftforge/client/extensions/IForgeEntityModel.java
```diff
@@ -0,0 +1,31 @@
...
+import net.minecraft.client.model.geom.ModelPart;
+import net.minecraft.world.entity.Entity;
+import net.minecraftforge.client.model.animation.IEntityModelAnimationModifier;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public interface IForgeEntityModel<T extends Entity>
+{
+    Set<IEntityModelAnimationModifier> animationModifiers = new HashSet<>();
```
sciwhiz12, line 12: I believe this won't work as you expect, as all fields in an interface are implicitly `public static final`, so all of these animation modifiers will be shared across all `IForgeEntityModel` instances.

src/main/java/net/minecraftforge/client/model/animation/IEntityModelAnimationModifier.java
```diff
@@ -0,0 +1,11 @@
+package net.minecraftforge.client.model.animation;
```
sciwhiz12, line 1: Missing license header.

src/test/java/net/minecraftforge/debug/client/rendering/AnimationModifierTest.java
```diff
@@ -0,0 +1,47 @@
+package net.minecraftforge.debug.client.rendering;
```
sciwhiz12, line 1: Missing license header.

patches/minecraft/net/minecraft/client/model/HumanoidModel.java.patch
```diff
@@ -0,0 +1,29 @@
+--- a/net/minecraft/client/model/HumanoidModel.java
++++ b/net/minecraft/client/model/HumanoidModel.java
+@@ -19,7 +_,8 @@
+ import net.minecraftforge.api.distmarker.OnlyIn;
+ 
+ @OnlyIn(Dist.CLIENT)
+-public class HumanoidModel<T extends LivingEntity> extends AgeableListModel<T> implements ArmedModel, HeadedModel {
++public class HumanoidModel<T extends LivingEntity> extends AgeableListModel<T> implements ArmedModel, HeadedModel, net.minecraftforge.client.extensions.IForgeEntityModel<T>
++{
```
sciwhiz12, line 9: In Minecraft classes (those which patches are generated for), the code style of the class should be followed rather than Forge's code style: 3 tabs, braces on the same line.

src/test/java/net/minecraftforge/debug/client/rendering/AnimationModifierTest.java
```diff
@@ -0,0 +1,47 @@
...
+                event.getSkin("slim")
+            };
+
+        for (LivingEntityRenderer renderer : renderers)
+        {
+            Model m = renderer.getModel();
+
+            if (m instanceof IForgeEntityModel fModel)
+            {
+                fModel.addAnimationModifier((entity, modelPart) ->
```
sciwhiz12, line 39: It seems this lambda hasn't been updated with the new parameters of the functional interface yet.

src/main/java/net/minecraftforge/client/model/animation/IEntityModelAnimationModifier.java
```diff
@@ -0,0 +1,11 @@
...
+
+import net.minecraft.client.model.geom.ModelPart;
+import net.minecraft.world.entity.Entity;
+
+public interface IEntityModelAnimationModifier<T extends Entity>
```
sciwhiz12, line 6: Please mark this functional interface with `@FunctionalInterface`.

sciwhiz12 (changes requested):
I have a big concern regarding the `IForgeEntityModel#setupAnim` method and obfuscation which needs to be tested at the very least. In addition, you need to add license headers (`licenseFormat`/`updateLicenses` task), and I recommend also adding some documentation on the new classes and methods

MrCrayfish:
I like the direction this PR is going and hope it eventually gets merged into Forge. I do have a concern since I've done something similar in a library of mine. A problem is that directly modifying the position and/or rotation of a model part won't reset it back to it's original definitions. It may seem like in your example that it's fine but start wrapping your modifications in a condition and you'll start to see the problem. Feel free to ask me anything.

JTK222:
Thanks for the offer, but I'll likely first have to redo this PR from 0.
I was way too tired when I wrote it and missed out on some important points that sciwhiz brought up.
So it's nothing more than a proof of concept right now.

And good point about resetting, I entirely forget that not all parts of a model might have their rotation set each frame.
I'll try to work out a solution for that.

JTK222:
Closing this in favor of #8278

PR #8234, opened 2021-12-01
Rework world persistence hooks.

### LINKED ISSUE #8125: Loading an existing world with forge triggers two data reloads instead of one

alcatrazEscapee:
### What this does

- Fixes #8125 (Loading a world triggers two resource reloads with Forge, can have significant impacts on world load times with large data packs, especially world gen)
- Fixes an issue where Forge's registry checking would build a error message and then discard it completely (This has existed since 1.17).

### Explanation of patches and changes:

- Removes the redundant `WorldPersistenceHook` in `ForgeMod`. It was marked with `//todo: 1.16 Dimension` and has been completely unused since then.
- Remove `WorldPersistenceHooks` entirely. It enforced an API that is fundamentally incompatible with what we actually need to do here, the parameters were unused, and unhelpful. If this is a mod-desirable API, then a future PR may re-add something similar but I doubt that will be the case.
- Re-implement the actually useful part of `WorldPersistenceHooks` into `ForgeHooks`. Doing this allowed the methods in question to have their signatures changed to what was actually being used rather than carrying unnecessary baggage around, and in particular, means they don't have to do **and discard** and entire reload of resources.
- There were a number of redundant and complicated patches in `LevelStorageSource`. These have been reworked into one patched method, which was necessary to access `LevelStorageSource.this` and other private things (that should be kept private) without some convoluted transformations.

### Testing

- Verified that when adding and removing registry entries (blocks), they are identified correctly, and the log message is spat out.
- As of 1.18, the core cause of this issue doesn't exist anymore (Vanilla's biome ID mismatching, since they switched to palleted biome containers on chunks).
- Loading a world with Forge only triggers one resource reload (two if there are world gen datapacks involved), which is one less than it was previously.

This is an re-implementation of #8163, which actually takes the chance during 1.18 to completely rip out and rewrite the world persistence hooks, rather than shoehorning in the fix into the (already broken, mind you) API.


src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1241,4 +1249,85 @@ public static void onEntityEnterSection(Entity entity, long packedOldPos, long p
...
+    {
+        CompoundTag tag = rootTag.getCompound("fml");
+        if (tag.contains("LoadingModList"))
+        {
+            ListTag modList = tag.getList("LoadingModList", net.minecraft.nbt.Tag.TAG_COMPOUND);
+            for (int i = 0; i < modList.size(); i++)
+            {
+                CompoundTag mod = modList.getCompound(i);
+                String modId = mod.getString("ModId");
+                if (Objects.equals("minecraft",  modId)) {
```
ChampionAsh5357: New line braces.

patches/minecraft/net/minecraft/world/level/storage/LevelStorageSource.java.patch
```diff
@@ -1,42 +1,30 @@
...
-@@ -178,6 +_,10 @@
-    }
+@@ -299,6 +_,19 @@
+          return LevelStorageSource.this.m_78229_(this.f_78271_.toFile(), LevelStorageSource.m_78247_(p_78281_, p_78282_));
+       }
  
-    static BiFunction<File, DataFixer, PrimaryLevelData> m_78247_(DynamicOps<Tag> p_78248_, DataPackConfig p_78249_) {
-+       return getReader(p_78248_, p_78249_, null);
-+   }
++      public void readAdditionalLevelSaveData() {
```
ChampionAsh5357, line 7: This is never called on the server, so I'm not sure how you plan on reading the data there.

ChampionAsh5357 (changes requested):
So, the title of your PR doesn't make much sense. You aren't reworking WPHs, you are removing them completely and then only implementing the part that Forge needs.

In other news, you seem to be missing a server implementation for reading the data.

alcatrazEscapee:
> So, the title of your PR doesn't make much sense. You aren't reworking WPHs, you are removing them completely and then only implementing the part that Forge needs.

I'd disagree, but "Rework WP Hooks" was shorter than "Remove WP hooks, reimplement the parts that Forge needs, fix a bug and remove the unused WP hook implementations" but I digress.

There are a couple things here:

- This PR exists to fix #8125 
- It's (essentially) infeasible to fix the linked issue without breaking and/or reworking WP hooks in some fashion, due to the awkward contract that they require
- Nothing has indicated to me that WP hooks are a public API that Forge is intentionally providing to mods (It's in the `minecraftforge.internal` package, completely undocumented, has been broken - doing double reloads - since 1.16, etc.)

And, moreover, I think a PR reworking / removing the current iteration of WP hooks (and incidentally, reimplementing the parts Forge needs in a much cleaner, and non-buggy fashion) should be separate from one adding a similar mod visible API. Because that begs the question:

 1. Is there a need for such an API?
 2. Where, if anywhere, does this API's calls need to be made (in the same spot as WPH? Or closer to the new patch locations?)
 3. What parameters or info need to be provided to the users of this API? (This is going to affect where the calls be placed.)

I'd strongly argue that no, there is no need for such and API and the questions are moot. And *even if* there is a need, that it be done in a separate PR, because really, this is a bugfix, not a new feature.

---

Anyway

- I've fixed the formatting that was spotted and looked over it again.
- I added the missing patch for servers - the functionality is exactly the same (delay the creation of `builtin` until after the save can be loaded, although this is much simpler on servers)

ChampionAsh5357 (approved):
Looks good to me

pupnewfster:
@alcatrazEscapee Could you please resolve the merge conflicts for this PR

alcatrazEscapee:
@pupnewfster Done.

PR #8235, opened 2021-12-02
[1.18.x] Remove the `BlockEntity#save(CompoundTag)` patch

SuperMartijn642:
## Current problem

In 1.18, the `BlockEntity#save(CompoundTag)` patch is no longer used when saving `BlockEntity`s. It still being present seems to have caused block entities to not be saved properly in other places, and it can cause confusion when updating from 1.16 or 1.17.

## The fix in this PR

This PR removes the `BlockEntity#save(CompoundTag)` patch and replaces its use by `BlockEntity#saveWithFullMetadata`.
This means
- `IForgeBlockEntity#serializeNBT` and `BlockSnapshot#getBlockEntityTag` save metadata such as location and capabilities again, like in 1.16 and 1.17
- `TileFluidHandler`'s and `FullPotsAccessorDemo$DioriteFlowerPotBlockEntity`'s data gets saved again


TheCurle:
I tried to do this during the initial porting and it caused world locking issues.
This needs a LOT of testing, as it's not as simple as a missed patch this time.

Mojang changed the internal semantics of the save and load methods, so we need to be careful with what we do here.

SuperMartijn642:
I did some testing with the proposed changes.

I tested the following scenarios:
- I created a world in 1.17.1 with a chest with items, a banner with a pattern, and a pig with a saddle. I then saved the world and loaded the world in 1.18.1 with this PR.
- I created a world in 1.12.2 with a chest with items, a banner with a pattern, and a pig with a saddle. I then saved the world and loaded the world in 1.18.1 with this PR.
- I loaded a large vanilla world saved in 1.14.4 in 1.18.1 with this PR.
- I created a world in 1.18.1 with this PR and flew around in the overworld, the nether, and the end. I also visited a stronghold and an end city.

In all of the above scenarios, I didn't experience any issues and nothing seemed out of the ordinary.
Here is a zip file containing the worlds I used for the first 3 scenarios: https://drive.google.com/file/d/1J0grlQrf3G6WOMWOybhmPisMhfoy2J76/view?usp=sharing

What else should be tested? And is there a better way I can quantify this testing?

PR #8236, opened 2021-12-02
Change ConfigValue#get() to throw if called before config loaded 

### LINKED ISSUE #7716: [RFC][1.17] Make ConfigValue#get throw if the config is not loaded yet
Calling `ConfigValue#get` before the config is loaded usually indicates a bug, since anything affected by this call can't actually be configured. For newly added config values one could argue that this should be found immediately during testing, but if the bug is introduced during a port or general refactoring this is not necessarily the case. By making any calls to `ConfigValue#get` throw until the config is loaded finding such bugs would become much easier.

In an ideal world this would be all that is necessary, in practice some way of working with non-loaded configs is required as well. One basic case of this is a tooltip that depends on a config value, e.g. on a configurable durability.  Tooltips are queried when the search tree for the creative menu is built, which happens before any configs are loaded. It does not really matter whether this is implemented as `isLoaded`, `getOr`, an explicit `getOrDefault` or something else, as long as it is possible to handle this case in *some* manner.

sciwhiz12:
This PR implements and closes #7716 by changing the behavior of `ConfigValue#get` to throw if called before the config has been loaded. This prevents silent issues where a mod gets the value of the setting before configs are loaded, which means the default value is always returned.

As said in the linked RFC issue, there may be situations where the getting the config setting before configs are loaded is needed, and it is not preferable or maybe even possible to hardcode the default value. For these situations, the original behavior (return the actual value, or the default if config is not yet loaded) is made available through the new `#getRaw` method.

This PR also introduces javadocs for both `#get` and `#getRaw` to clarify the distinction between the two method's behaviors, and to point current users of `#get` which rely on the default-value-returning behavior to use the other method instead.


patches/minecraft/net/minecraft/client/Minecraft.java.patch
```diff
@@ -29,7 +29,7 @@
        }
  
        Util.f_137440_ = RenderSystem.m_69576_();
-+      if (net.minecraftforge.common.ForgeConfig.CLIENT.forceSystemNanoTime.get()) Util.f_137440_ = System::nanoTime;
++      if (net.minecraftforge.common.ForgeConfig.CLIENT.forceSystemNanoTime.getRaw()) Util.f_137440_ = System::nanoTime;
```
malte0811: If this runs before configs are loaded, doesn't that mean that the option does not do anything? This code only runs once during startup, the usual usecase for `getRaw` should be code that is called both before and after the config is loaded.

Thanks for implementing this btw!

sciwhiz12: Yes; as currently implemented, the `forceSystemNanoTime` option on the Forge client config does not affect anything at all. By this reasoning, the config option should be removed (or possibly moved to a later position), but I'm presuming that change would be out of this PR's scope, and am leaving it in for now for a later PR to remove it. 

If wanted, I can mark the option for removal in a later PR with a `// TODO` comment, or fix/remove the option in this PR, though I'd like to keep the scope of this PR as-is and leave the matter of fixing/removing it for someone else more knowledgable on its effects (or lack thereof) on rendering).

ichttt: This is actually the perfect example for the use case of this PR - the current `get` method may cause cases where the config isn't respected without the modder recognising this. I would at least add a `TODO` comment here, but if you want to fix it directly in this PR I personally wouldn't consider this Out of Scope

sciwhiz12: I've marked it as a TODO for now, since this needs to be reevaluated more closely by people more knowledgable of `glfwGetTime` possibly having performance issues (see the commit message of 06804675be18c612138a713aca1a5801a7e5dbf8 which added this).

ichttt (approved):
Code LGTM. I don't know if it's too late in the 1.18.1 lifecycle to introduce this (even though it isn't binary breaking and we would even allow binary breaks as we don't have a RB yet), but it might cause some mods to crash on startup. I'll leave that up to the core team to decide.

sciwhiz12:
After a bit of (internal) discussion, I decided to change the check to be dev-time only for now, with comments and an exception message notifying developers that this will be upgraded to a full check (both dev-time and production-time) in the future, such as during 1.19.

This is to reduce the breakage caused by this PR, by allowing already published mods to run without error, while erroring out in the dev environment so developers can fix their incorrect and untimely usages of `#get`.

sciwhiz12:
After a bit more internal discussion, `getRaw()` has been removed and `getDefault()` added in its place. Any would-be calls to `getRaw()` should now be the expression `spec.isLoaded() ? configValue.get() : configValue.getDefault()` (where `ForgeConfigSpec spec` and `ConfigValue<?> configValue`).

sciwhiz12:
PR has been updated to 1.19.

TheCurle:
lgtm

PR #8237, opened 2021-12-02
Restore missing patch to LevelChunk

SquidDev:
In 1.17.x `ClientPacketListener` was patched to call `handleUpdateTag` instead of load. This logic was moved to `LevelChunk`, and the patch accidentally dropped in the migration.

I discussed this yesterday with Curle on Discord, but I'm not sure what the conclusion was so opening a PR for better visibility/tracking. I know there was some concern about the changed semantics of save/load - I don't think this is an issue here, as `handleUpdateTag` just calls `load` by default, so we're not changing vanilla behaviour at all.


sciwhiz12 (approved):

SquidDev:
Thank you!

Issue #7993, opened 2021-08-08
[1.17] Minecraft exits without a crash report when mod dependencies are missing


SuperMartijn642:
**Minecraft Version:** 1.17.1

**Forge Version:** 1.17.1-37.0.27

**Logs:**
Latest.log: https://gist.github.com/SuperMartijn642/71f4e0a7c69b58831723deb6ff58de0c

**Steps to Reproduce:**
 1. Download and open the MDK for 1.17.1 - 37.0.27
 2. Add the following dependency to `mods.toml`:
 ```
 [[dependencies.examplemod]]
    modId="somemode"
    mandatory=true
    versionRange="[1,2)"
    ordering="NONE"
    side="BOTH"
```
 3. Build the mod using the gradle `build` task
 4. Start Minecraft in the vanilla launcher with the generated jar in the mods folder

**Description of issue:**
Outside of the dev environment, when a mod dependency is not installed the game exits without showing the mod loading error screen or generating a crash report.
Inside the dev environment the game shows the mod loading error screen as expected.

sciwhiz12:
I've confirmed that this issue happens, and found the root cause for the issue.

## What's All This About

The problem is what happens when `ModSorter`, the class in FML responsible for sorting mods based on dependency information, encounters a situation wherein there are unsatisfied mod requirements, such as missing mandatory dependencies or installed but incompatible optional dependencies.

For this situation, it stores a small list internally in a field called `forgeAndMC`. As you may surmise from the name of the field, this list **_should_** contain two mod files: the one containing the `minecraft` mod (which is the client JAR), and the one containing the `forge` mod (the Forge universal JAR). 

This list is passed on as the "full mods list" in three situations: 
- There are errors during scanning of mod files;
- There are unsatisfied dependencies; or
- A cyclic dependency order is detected.

So theoretically, if any of the above were to happen, this would be fine: Forge and Minecraft are the only 'mods' in the mod list (from that `forgeAndMC` field), mod loading continues, __the access transformer configurations are loaded__, and it should just continue until it reaches the point where FML can display the mod loading errors screen.

## UH OH

[...]

PR #8238, opened 2021-12-02
Introduce system mods to mod loading

### LINKED ISSUE #8049: [RFC] System mod concept
### Background

As discussed by [my comment in #7993](https://github.com/MinecraftForge/MinecraftForge/issues/7993#issuecomment-909366998) and my subsequent PR #8043, there is a need to be able to mark certain mods as 'system mods' -- meaning, that these mods are required for the loading of the game, and are given special treatment. 

In the situation of the above linked issue and PR, one of these special treatments is that these mods are filed away by `ModSorter`, and passed on to the loading system as the only contents of the mod list if `ModSorter` encounters an exceptional situation. By passing these 'game core' mods into the `LoadingModList` along with the list of errors, the game can continue loading but at reduced functionality, only loading upto the point where an error screen may be displayed to the user.

The current problem found in the above linked issue is that, because of the recent separation between FML and Forge in 1.17, FML's `ModSorter` only includes the `minecraft` mod in that 'core game' mods list (the `List<ModFile> forgeAndMC` field). This prevented the configuration of the AT subsystem, which meant that an error later manifested due to accessing a field expected to be transformed to `protected` but wasn't.

### Potential solutions

1. **Hardcode the `forge` mod to be a system mod if present.**

   This is less a solution and more a hack, which fixes the issue above but violates the conceptual separation between FML and Forge. Note that this is implemented by my PR above, but is only meant to be temporary until a proper solution, with a possibly better API path, is implemented.

1. **Add a new `systemMod` property to mod files, which are then validated by the running system later.**

   By adding a new property to `ModFile`, breaking changes can be avoided in `ModSorter` as the information on system mods are available directly from the `ModFile`. After mod sorting, the running system -- such as Forge -- can perform its own validation on what system mods are installed and raise an error if unknown ones are installed.

1. **Retrieve the list of system mods from some object/process beforehand.**

   By having another step which runs before mod sorting prepare the list, such as within `FMLLoader` (possibly provided by launch targets?), the responsibility of specifying what the system mods is passed on to whatever system uses FML, such as Forge.
   For example, a method could be added to `CommonLaunchHandler` which returns a `Set<String>` of mod IDs for the system mods, which could then be passed onto `ModSorter`.

I welcome anyone to suggest alternative solutions to the above ones that I just thought up. I would like to note that because this is FML-land that's being discussed, whatever solution that is PR'd will fall into the hands of cpw to accept or reject, and because of his personal and work responsibilities, he will probably only be able to discuss the solution once PR'd and he becomes available to do so.

sciwhiz12:
This PR implements and closes #8049 by introducing the concept of **system mods** to mod loading.

System mods are mods which are required to exist in the environment during mod loading. These may be specially provided mods (for example, the `minecraft` mod), or mods which are vital to the framework which FML is connected to (for example, Minecraft Forge and the `forge` mod). If any of these configured mod IDs cannot be found during mod loading, an exception is immediately thrown to halt the game.

These system mods are used as the only existing mods in the mod list if mod sorting or dependency verification fails. This allows later steps in the which use resources from these mod files to work correctly (up to when the error screen is shown and the game exits). 

As an example: in #7993, because the `forge` mod was missing from this list, the forge universal JAR was not included when the Access Transformers subsystem was searching for AT configuration files to load. This caused a JVM error to be thrown when the first field (that was assumed to be transformed by AT at that point) tried to be accessed by Forge code.

#### Implementation notes

~~The set of mod IDs of system mods are passed in as FML service provider arguments: the `coreGameMods` option (when referenced in the command line: `--fml.systemMods`), which takes in comma-separated mod IDs for those core game mods.~~

~~These are then passed down through `ModDiscoverer` into `ModSorter`, which does the finding of the core game mods and adding them to the `systemMods` field (previously named `forgeAndMC`). The FML-only project is configured with only the `minecraft` mod as the core game mod, while the Forge project is configured with both `minecraft` and `forge` as core game mods.~~

The set of mod IDs of system mods are configured from the value of the `FML-System-Mods` manifest entry from first `ModFile` whose locator is `MinecraftLocator`.


cpw:
I see.. good idea..

cpw (commented):
Not quite sure about passing it all the way in from launch, but otherwise OK.

sciwhiz12:
I was/am unsure if there is a better way to provide the set of core game mod IDs to the mod validator from FMLoader. 
I thought about a file resource containing the mod IDs per line, but that would mean somehow configuring the path to the JAR so the resource can be loaded, which would possibly be a bit of a mess.
If anyone has a better idea, I'd like to hear it so we can see if it can be implemented cleanly. Otherwise, this PR is as ready as it'll ever be (pending reviews from Triage/Core for any possible defects).

cpw:
Manifest? We could add a manifest entry to the forge jar listing the "core mods"?

cpw:
Also, lets call them something other than "core mods". We have core mods already, and this would just confuse matters. These should be call "system mods" because they're part of ensuring basic system function.

sciwhiz12:
I agree with the rename to "system mods", for less confusion with "coremods".

For the manifest idea, I've found a prospective solution which involves iterating over the list of `ModFile`s in `ModSorter#buildUniqueList`, filtering out those which do not use `MinecraftLocator`, and taking the first `FML-System-Mods` manifest value from these mod files manifests. 

However, this does have one big gotcha: the `minecraft` mod is now hardcoded as an always-existing system mod, because there is no clean way for me to introduce a manifest for the FMLOnly jar in the development environment. We could potentially change that by making the `minecraft` mod be added as the only existing system mod if there are no system mods configured via the manifest, but I leave that decision up to you.

cpw:
Is having minecraft hardcoded as a systemmod a bad thing? Given it's a requirement for basically everything else, it seems like a very small loss of generality... Is FML-System-Mod going to act as a flag? Do we expect many other mods to need this? It requires some very special setup to be able to play in this space, so exposing it generally will come at a very high potential for fuck ups..

sciwhiz12:
On further thought, `minecraft` being a hardcoded system mod is not really bad, since it is always present as added by `MinecraftLocator`, but I did want to note it's hardcodedness at least.

And `FML-System-Mods` is not a flag, but rather a comma-separated list of mod IDs which are designated as system mods. Given that the manifest value is (a) only taken from mod files whose locator is `MinecraftLocator`, which excludes any mod provided either via the regular locators or a custom locator, and (b) limited to the first found manifest entry from those mod files, it is highly unlikely if not impossible for mod developers to actually affect FML by inserting this manifest entry in their own JARs.

cpw:
OK, so your idea is to use the MinecraftLocator to locate the "tag". That seems reasonable, agree.

cpw (approved):
Okay that looks much better

PR #8242, opened 2021-12-02
[1.18]: Support for DFU

marchermans:
### Modifiable DFU

#### Goals

#### Implementation

#### Future:

#### Usage:

##### Example:
```java
{

    @SubscribeEvent
    {
            );
        }
    }
}
```

#### Todo:
- [X] Better testing
- [X] Formatting


src/main/java/net/minecraftforge/common/ForgeStatesProvider.java
```diff
@@ -70,4 +83,6 @@ public ThreadSelector threadSelector() {
             return ModLoadingStage::currentState;
         }
     }
+
+
```
pupnewfster: Adding these random blank lines seems a bit odd.

src/main/java/net/minecraftforge/datafix/ForgeDataFixDelegate.java
```diff
@@ -0,0 +1,102 @@
...
+import com.mojang.datafixers.DataFix;
+import com.mojang.datafixers.TypeRewriteRule;
+import net.minecraftforge.fml.util.ObfuscationReflectionHelper;
+import org.jetbrains.annotations.Nullable;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+/**
+ * This delegate allows for rebuilding of the cached rule, which is needed since it is
```
pupnewfster, line 31: Since it is what?

src/main/java/net/minecraftforge/datafix/ForgeDataFixDelegate.java
```diff
@@ -0,0 +1,102 @@
...
+    private TypeRewriteRule rebuildRuleReflectively()
+    {
+        try
+        {
+            //Just invoke it via reflection.
+            final Object candidate = makeRuleMethod.invoke(this.wrapped);
+            //Check if we have a rule.
+            if (candidate instanceof TypeRewriteRule typeRewriteRule)
+            {
+                //Yep, returning.
```
pupnewfster: Some of these comments feel a bit overkill to include,

src/main/java/net/minecraftforge/datafix/ForgeDataFixerBuilder.java
```diff
@@ -0,0 +1,103 @@
...
+    public ForgeDataFixerBuilder(final int dataVersion)
+    {
+        super(dataVersion);
+        this.dataVersion = dataVersion;
+    }
+
+    @Override
+    public Schema addSchema(
+      final int version, final int subVersion, final BiFunction<Integer, Schema, Schema> factory
+    )
```
pupnewfster: Having this signature on three lines feels a bit odd.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerBuilder.java
```diff
@@ -0,0 +1,103 @@
...
+        return wrapped;
+    }
+
+    @Override
+    public void addSchema(final Schema schema)
+    {
+        final Schema parent = schemas.isEmpty() ? null : schemas.get(getLowestSchemaSameVersion(schemas, schema.getVersionKey() - 1));
+        final ForgeSchema wrapped = new ForgeSchema(schema.getVersionKey(), parent, schema);
+        schemas.put(wrapped.getVersionKey(), wrapped);
+        allSchemas.add(wrapped);
```
pupnewfster: It may not be worth it but part of me wonders if having a private helper method for this and the other addSchema method may be of use as it looks like the only difference in this grouping of four lines is key vs passed in key (could become a parameter), and then just making the BiFunction always return schema.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
```diff
@@ -0,0 +1,194 @@
...
+ */
+class ForgeDataFixerDelegateHandler extends DataFixerUpper
+{
+    private static final Logger LOGGER = LogManager.getLogger();
+
+    private final int                              dataVersion;
+    private final List<ForgeSchema>                allSchemas;
+    private final Int2ObjectSortedMap<ForgeSchema> schemas;
+    private final List<ForgeDataFixDelegate> globalList;
+    private final IntSortedSet               fixerVersions;
```
pupnewfster: Spacing is a little odd here, my guess is either they should all be aligned vertically or just not have the extra spacing.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
```diff
@@ -0,0 +1,194 @@
...
+      final Int2ObjectSortedMap<ForgeSchema> schemas,
+      final List<ForgeSchema> allSchemas,
+      final List<ForgeDataFixDelegate> globalList,
+      final Executor executor,
+      final IntSortedSet fixerVersions
+    )
+    {
+        //Supply dummy data to the super constructor
+        //We never invoke and method on the super and so this data does not matter.
+        super(new Int2ObjectAVLTreeMap<>(), new ArrayList<>(), new IntAVLTreeSet());
```
pupnewfster: Presuming at least some of the usages of these are overridden, can any of these dummy values be replaced with an empty collection, for example I believe all uses of the super schemas map is overridden so you could probably pass `Int2ObjectSortedMaps.emptyMap()` instead.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
```diff
@@ -0,0 +1,194 @@
...
+    {
+        //Reset our rules, normal vanilla rules (ours too) keep a cache of their schemas
+        //We can not use that and need to forcefully recompute them.
+        this.globalList.forEach(ForgeDataFixDelegate::resetRule);
+
+        //All schemas are already reset via the mod event bus as this time.
+        //Rebuild of those is not needed.
+        this.activeFixer = new InnerFixer(new Int2ObjectAVLTreeMap<Schema>(schemas), new ArrayList<>(globalList), new IntAVLTreeSet(fixerVersions));
+
+        //This section was added to make it easier to debug the rule set at runtime, since it is now not pre-computed off thread anymore,
```
pupnewfster: > since it is now not pre-computed off thread anymore,

Might read better as

> since it is no longer pre-computed off thread,

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /**
+     * Gives access to all known schemas at the point of building of this wrapper.
+     *
+     * @return All schemas known to DFU.
+     */
+    public List<ForgeSchema> getAllSchemas()
+    {
+        return allSchemas;
```
pupnewfster, line 170: Should this be an unmodifiable view?

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /**
+     * Represents an DFU that gives access to the protected sub methods so data can be properly extracted.
```
pupnewfster: I believe this should be changed `an DFU` -> `a DFU`

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
```diff
@@ -0,0 +1,194 @@
...
+    {
+        private InnerFixer(final Int2ObjectSortedMap<Schema> schemas, final List<DataFix> globalList, final IntSortedSet fixerVersions)
+        {
+            super(schemas, globalList, fixerVersions);
+        }
+
+        @Override
+        public <T> Dynamic<T> update(
+          final DSL.TypeReference type, final Dynamic<T> input, final int version, final int newVersion
+        )
```
pupnewfster: Having this signature on three lines feels a bit odd.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerEventHandler.java
```diff
@@ -0,0 +1,169 @@
...
+     *
+     * The generator will always return the same multi-thread capable event reference regardless
+     * of the passed in mod name.
+     *
+     * @return The stream of event generators used to fire one event per mod per schema.
+     */
+    public static Stream<IModStateTransition.EventGenerator<?>> generateConfigureEvents()
+    {
+        //Grab the schemas
+        List<ForgeSchema> schemas = Lists.newArrayList(getAllSchemas());
```
pupnewfster, line 84: Is there a reason this isn't just `new ArrayList<>(getallSchemas())`

src/main/java/net/minecraftforge/event/datafix/ConfigureDatafixSchemaEvent.java
```diff
@@ -0,0 +1,179 @@
...
+ * of a DFU Schemas type signature.
+ *
+ * If used to register a new entity or block entity to a particular version,
+ * then this also will need to be registered to all newer versions to persist properly!
+ * As opposed to vanilla which automatically transfers the entities and block entities from older
+ * schemas to newer ones, this event does not do that and the modder is on his own to implement this properly.
+ *
+ * We suggest instead of checking for version equality to check for a greater or equal version.
+ */
+public class ConfigureDatafixSchemaEvent extends Event implements IModBusEvent
```
pupnewfster: Given how DataFixers are named elsewhere I believe this should be called `ConfigureDataFixSchemaEvent ` (basically make the `f` capital `F`)

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,298 @@
...
+    /**
+     * Rebuilds the types, with support for the case that no types exists.
+     * (Which should never happen, but you never know!)
+     *
+     * @return The type map.
+     */
+    protected Map<String, Type<?>> buildTypes()
+    {
+        //Result map.
+        final Map<String, Type<?>> types = Maps.newHashMap();
```
pupnewfster: Is there a reason that this method returns a hashmap, but the variable that is set to this is defaulted to a concurrent map?

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,298 @@
...
+                type = family.apply(recurseId);
+            } else {
+                type = getTemplate(name).apply(family).apply(-1);
+            }
+            types.put(name, type);
+        }
+        return types;
+    }
+
+    public Set<String> types()
```
pupnewfster, line 128: Missing override annotation (on this method and a bunch of others in this class)? It also is probably worth adding a one line comment/javadoc for the methods that are carbon copies of super, in case things in DataFixerUpper change so as to make it easier to track.

pupnewfster (commented):
It might be worth creating an example mod to better show how this might be used, even if the "old" format it fixes is just part of the comments.

src/main/java/net/minecraftforge/datafix/ForgeDataFixDelegate.java
```diff
@@ -0,0 +1,102 @@
+package net.minecraftforge.datafix;
```
sciwhiz12, line 20: Missing license header.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerBuilder.java
```diff
@@ -0,0 +1,103 @@
+package net.minecraftforge.datafix;
```
sciwhiz12, line 20: Missing license header.

src/main/java/net/minecraftforge/datafix/ForgeDataFixDelegate.java
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * This delegate allows for rebuilding of the cached rule, which is needed since it is
+ */
+class ForgeDataFixDelegate extends DataFix
+{
+
+    private static final Method makeRuleMethod = ObfuscationReflectionHelper.findMethod(
+      DataFix.class,
+      "makeRule"
+    );
```
sciwhiz12, line 41: I presume ObfuscationReflectionHelper is used here not because of its name remapping ability, but for convenience of retrieving the `Method` instance for reflection (since `com.mojang.datafixers.DataFix` is not an obfuscated class)?

Also, given that `DataFix#makeRule` seems to be `protected abstract`, and this class extends `DataFix`, doesn't it mean you are allowed to call that method without needing to do reflection (or am I misremembering how access works)?

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
```diff
@@ -0,0 +1,194 @@
+package net.minecraftforge.datafix;
```
sciwhiz12, line 20: Missing license header.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
```diff
@@ -0,0 +1,194 @@
...
+        //This section was added to make it easier to debug the rule set at runtime, since it is now not pre-computed off thread anymore,
+        //But only on the main thread, at the moment it is needed.
+        //This basically disables the REWRITE_CACHE prefill and fills it on-demand.
+        //Although it is mainly used for debugging, it also provides a good way of reducing DFU's memory footprint, when it is not needed.
+        //Since the following code will compute an update rule from every version to the current for every known type in the schema,
+        //relating to this version. By default, however this is disabled to preserve vanilla compatibility, since it does do the pre-compute
+        //to fill up the cache.
+        //To enable it start the JVM with the system property set to true, -Dforge.datafixer.disablePreCompute=true needs to be added as
+        //a JVM Launch argument.
+        if (System.getProperty("forge.datafixer.disablePreCompute", "false").trim().toLowerCase(Locale.ROOT).equals("true")) {
```
sciwhiz12: Could use [`Boolean#getBoolean(String)`](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/lang/Boolean.html#getBoolean(java.lang.String)) instead.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
```diff
@@ -0,0 +1,194 @@
...
+                    final TypeRewriteRule rule = activeFixer.getRule(DataFixUtils.getVersion(versionKey), dataVersion);
+
+                    //Optimize the rule to only apply to the given type.
+                    dataType.rewrite(rule, DataFixerUpper.OPTIMIZATION_RULE);
+                }, executor).exceptionally(e -> {
+                    //Fails.
+                    LOGGER.error("Unable to build datafixers", e);
+
+                    //Just kill it, if this fails we have a problem anyway!
+                    Runtime.getRuntime().exit(1);
```
sciwhiz12, line 138: This seems a bit extreme, to quit the application when datafixers cannot be built. Is there no other more user-friendly way to bubble up the exception?

src/main/java/net/minecraftforge/datafix/ForgeDataFixerEventHandler.java
```diff
@@ -0,0 +1,169 @@
+package net.minecraftforge.datafix;
```
sciwhiz12, line 20: Missing license header.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,298 @@
+package net.minecraftforge.datafix;
```
sciwhiz12, line 20: Missing license header.

src/main/java/net/minecraftforge/event/datafix/ConfigureDatafixSchemaEvent.java
```diff
@@ -0,0 +1,179 @@
+package net.minecraftforge.event.datafix;
```
sciwhiz12: Missing license header.

sciwhiz12 (changes requested):
Gave it a semi-quick look-over. There's missing license headers so `licenseFormat`/`updateLicenses` needs to be run, and I didn't note it in the comments but there's also cases of inconsistent placement of braces (some on newlines, some on the same line).

src/main/java/net/minecraftforge/datafix/ForgeDataFixDelegate.java
marchermans: So I though so to, but it turns out, no.
You can not simply extend the base class and then access a protected method from a different instance, which is potentially not in your inheritance tree.

This is only possible if the targeted object is guaranteed to be in your inheritance tree, which is not the case here. Both Lex and I thought too that it was possible, but the compiler just does not allow it.

marchermans: And yes we use ORH here to prevent the reflection gore and provide better future compatibility with regards to reflection and its limitations in the newer JVM version.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
marchermans: This is literally what vanilla MC does.....

Since I do not want to change that behavior and DFU will in that case later on crash with a weird exception, I prefer the log entry and then just stop the engine from there, besides this is run in an executor which does not really allow us to bubble up the exception neatly, in this case the game just stops with the DFU error log line as the last line with the exception as well, which seems to be the cleanest option in my opinion.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerEventHandler.java
marchermans: Not really I just do it that way.......

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
marchermans: Done in the constructor.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
marchermans: Nope just missed it.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
```diff
@@ -0,0 +1,213 @@
...
+        //This section was added to make it easier to debug the rule set at runtime, since it is no longer pre-computed off thread,
+        //But only on the main thread, at the moment it is needed.
+        //This basically disables the REWRITE_CACHE prefill and fills it on-demand.
+        //Although it is mainly used for debugging, it also provides a good way of reducing DFU's memory footprint, when it is not needed.
+        //Since the following code will compute an update rule from every version to the current for every known type in the schema,
+        //relating to this version. By default, however this is disabled to preserve vanilla compatibility, since it does do the pre-compute
+        //to fill up the cache.
+        //To enable it start the JVM with the system property set to true, -Dforge.datafixer.disablePreCompute=true needs to be added as
+        //a JVM Launch argument.
+        if (Boolean.getBoolean("forge.datafixer.disablePreCompute")) {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,329 @@
...
+        {
+            //Check for its existence or nuke.
+            return Maps.newHashMap();
+        }
+        //Build a type family for our recursive types.
+        final TypeTemplate choice = availableTypeTemplate.get();
+        final TypeFamily family = new RecursiveTypeFamily(name, choice);
+
+        //Now process all other known types, taking care of recursion.
+        for (final String name : TYPE_TEMPLATES.keySet()) {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,329 @@
...
+            final Type<?> type;
+            final int recurseId = RECURSIVE_TYPES.getOrDefault(name, -1);
+            if (recurseId != -1)
+            {
+                type = family.apply(recurseId);
+            } else {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,329 @@
...
+        return choiceType.types().get(choiceName);
+    }
+
+    @Override
+    public TaggedChoice.TaggedChoiceType<?> findChoiceType(final DSL.TypeReference type) {
+        return getType(type).findChoiceType("id", -1).orElseThrow(() -> new IllegalArgumentException("Not a choice type"));
+    }
+
+    @Override
+    public void registerTypes(final Schema schema, final Map<String, Supplier<TypeTemplate>> entityTypes, final Map<String, Supplier<TypeTemplate>> blockEntityTypes) {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,329 @@
...
+        final TaggedChoice.TaggedChoiceType<?> choiceType = findChoiceType(type);
+        if (!choiceType.types().containsKey(choiceName))
+        {
+            throw new IllegalArgumentException("Data fixer not registered for: " + choiceName + " in " + type.typeName());
+        }
+        return choiceType.types().get(choiceName);
+    }
+
+    @Override
+    public TaggedChoice.TaggedChoiceType<?> findChoiceType(final DSL.TypeReference type) {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,329 @@
...
+            wrapped.registerTypes(schema, entityTypes, blockEntityTypes);
+        }
+        else if (parent != null)
+        {
+            parent.registerTypes(schema, entityTypes, blockEntityTypes);
+        }
+    }
+
+    @Override
+    public Map<String, Supplier<TypeTemplate>> registerEntities(final Schema schema) {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,329 @@
...
+        if (parent != null)
+        {
+            return parent.registerEntities(schema);
+        }
+
+        return Maps.newHashMap();
+    }
+
+    @Override
+    public Map<String, Supplier<TypeTemplate>> registerBlockEntities(final Schema schema) {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,329 @@
...
+        if (recursive && !RECURSIVE_TYPES.containsKey(type.typeName()))
+        {
+            RECURSIVE_TYPES.put(type.typeName(), RECURSIVE_TYPES.size());
+        }
+    }
+
+    /**
+     * Resets the schema so that the type map can be properly rebuild.
+     */
+    public void resetSchema() {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,329 @@
...
+     * Rebuilds the schema's type map.
+     * This also injects custom types for entities and block entities.
+     *
+     * @param modEntityTypes The mod entity types to inject.
+     * @param modBlockEntityTypes The block entity types to inject.
+     */
+    public void rebuildSchema(
+      final Map<String, Supplier<TypeTemplate>> modEntityTypes,
+      final Map<String, Supplier<TypeTemplate>> modBlockEntityTypes
+    ) {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,198 @@
...
+    {
+        return schema;
+    }
+
+    /**
+     * The version of the schema that is about to be configured.
+     *
+     * @return The version.
+     */
+    public int getVersion() {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,198 @@
...
+        return getSchema().getVersionKey();
+    }
+
+    /**
+     * Registers a new simple (catch-all-nbt) block entity to the current schema and all its descendent versions.
+     *
+     * @param registryName The registry name of the block entity.
+     */
+    public void registerSimpleBlockEntity(final ResourceLocation registryName) {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,198 @@
...
+        registerBlockEntity(registryName, DSL::remainder);
+    }
+
+    /**
+     * Registers a new simple (catch-all-nbt) block entity with an item inventory array on the nbt field with the given name to the current schema and all its descendent versions.
+     *
+     * @param registryName The registry name of the block entity.
+     * @param nbtKey The name of the nbt field to where the item array is located.
+     */
+    public void registerSimpleItemInventory(final ResourceLocation registryName, final String nbtKey) {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /**
+     * Allows for the custom definition of a new block entity type and registration of it to the current schema and all its descendent versions.
+     * The function is given the registry name of the block entity type template to produce.
+     *
+     * @param registryName The registry name of the block entity type.
+     * @param template The function to produce the block entity type template.
+     */
+    public void registerBlockEntity(final ResourceLocation registryName, final Function<String, TypeTemplate> template) {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /**
+     * Allows for the custom definition of a new block entity type and registration of it to the current schema and all its descendent versions.
+     * The supplier is invoked to produce the block entity type template.
+     *
+     * @param registryName The registry name of the block entity type.
+     * @param template The supplier of the type template.
+     */
+    public void registerBlockEntity(final ResourceLocation registryName, final Supplier<TypeTemplate> template) {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,198 @@
...
+        blockEntityTypes.put(registryName.toString(), template);
+    }
+
+    /**
+     * Removes an already registered block entity from the current schema and all its descendent versions.
+     * This allows for the removal of a custom block entity type when it is registered in a parent version and is no longer needed.
+     *
+     * @param registryName The registry name of the block entity to remove.
+     */
+    public void removeBlockEntity(final ResourceLocation registryName) {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,198 @@
...
+        blockEntityTypes.put(registryName.toString(), null);
+    }
+
+    /**
+     * Registers a new simple (catch-all-nbt) entity to the current schema and all its descendent versions.
+     *
+     * @param registryName The registry name of the entity.
+     */
+    public void registerSimpleEntity(final ResourceLocation registryName) {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /**
+     * Allows for the custom definition of a new entity type and registration of it to the current schema and all its descendent versions.
+     * The function is given the registry name of the entity type template to produce.
+     *
+     * @param registryName The registry name of the entity type.
+     * @param template The function to produce the entity type template.
+     */
+    public void registerEntity(final ResourceLocation registryName, final Function<String, TypeTemplate> template) {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /**
+     * Allows for the custom definition of a new entity type and registration of it to the current schema and all its descendent versions.
+     * The supplier is invoked to produce the entity type template.
+     *
+     * @param registryName The registry name of the entity type.
+     * @param template The supplier of the type template.
+     */
+    public void registerEntity(final ResourceLocation registryName, final Supplier<TypeTemplate> template) {
```
sciwhiz12: Formatting: braces on newlines.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,198 @@
...
+        entityTypes.put(registryName.toString(), template);
+    }
+
+    /**
+     * Removes an already registered entity from the current schema and all its descendent versions.
+     * This allows for the removal of a custom entity type when it is registered in a parent version and is no longer needed.
+     * 
+     * @param registryName The registry name of the entity to remove.
+     */
+    public void removeEntity(final ResourceLocation registryName) {
```
sciwhiz12: Formatting: braces on newlines.

sciwhiz12 (changes requested):
As requested, made review comments for formatting, specifically to put braces on newlines.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerEventHandler.java
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * Event handler utility class for managing the events related to data fixers.
+ * This particular class is used to handle firing of the parallel mod loading events
+ * for the DFU setup.
+ *
+ * These events are fired immediately after mod construction and before the rest of the game
+ * loads, even before registries are fired.
+ */
+public class ForgeDataFixerEventHandler
```
pupnewfster, line 48: Part of me feels like this shouldn't really be exposed as public API and just be in a different package and package private, but it would feel sort of weird to not have this in the datafix package. At the very least it may be worth putting a note in the javadoc that this is for internal use only.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerEventHandler.java
```diff
@@ -0,0 +1,188 @@
...
+        //Make one event per schema, and return it reference based from the event generator,
+        //regardless of which mod the generator is invoked for.
+        return schemas.stream()
+                 .map(ConfigureDataFixSchemaEvent::new)
+                 .map(event -> IModStateTransition.EventGenerator.fromFunction(modContainer -> event));
+    }
+
+    /**
+     * Invoked by the event management system to indicate that we are about the fire an event for the given generator to all mods.
+     * This is exactly invoked once, and once only, per generator (and as such per event, and schema).
```
pupnewfster: > This is exactly invoked once, and once only, per generator...

Would read better as:
> This is invoked exactly once, and only once, per generator...

src/main/java/net/minecraftforge/datafix/ForgeDataFixerEventHandler.java
```diff
@@ -0,0 +1,188 @@
...
+     *
+     * @param executor The executor to execute this task on.
+     * @param eventGenerator The event generator that is about to be executed.
+     * @return A future that indicates success or failure.
+     */
+    public static CompletableFuture<List<Throwable>> preDispatch(
+      final Executor executor, final IModStateTransition.EventGenerator<ConfigureDataFixSchemaEvent> eventGenerator
+    )
+    {
+        return CompletableFuture.runAsync(()-> {
```
pupnewfster: Missing space after the parenthesis before the lambda's arrow.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerEventHandler.java
```diff
@@ -0,0 +1,188 @@
...
+            //Log that we are processing that particular event.
+            LOGGER.debug("Registering DFU Schema updates: {}", event.getSchema().getVersionKey());
+
+            //Check if we are a compatible schema.
+            if (event.getSchema() instanceof ForgeSchema forgeSchema)
+            {
+                //And reset if needed.
+                forgeSchema.resetSchema();
+            }
+        }, executor).thenApply(v-> Collections.emptyList());
```
pupnewfster: Missing space after `v` before the lambda's arrow.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerEventHandler.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    /**
+     * Invoked by the event management system to indicate that it has fired all the events for the given schema.
+     * This allows us to collect the data from the events and trigger the given schema to rebuild its type map.
+     *
+     * @param executor The executor to execute the task on.
+     * @param eventGenerator The event generator for the event (and thus) schema in question.
```
pupnewfster: > (and thus) schema in question.

Would read better as:
> (and thus the schema in question).

src/main/java/net/minecraftforge/datafix/ForgeDataFixerEventHandler.java
```diff
@@ -0,0 +1,188 @@
...
+     * @return The future that indicates success or failure.
+     */
+    public static CompletableFuture<List<Throwable>> postDispatch(
+      final Executor executor, final IModStateTransition.EventGenerator<ConfigureDataFixSchemaEvent> eventGenerator
+    )
+    {
+        return CompletableFuture.runAsync(()-> {
```
pupnewfster: Missing space after the parenthesis before the lambda's arrow.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerEventHandler.java
```diff
@@ -0,0 +1,188 @@
...
+     * then an empty list is returned.
+     *
+     * @return The list of all known schemas to handle.
+     */
+    private static List<ForgeSchema> getAllSchemas()
+    {
+        //Check if we are running in a compatible environment.
+        if (DataFixers.getDataFixer() instanceof ForgeDataFixerDelegateHandler forgeDataFixerDelegateHandler)
+        {
+            //Forge DFU wrapper is active, grabbing its configured schemas.
```
pupnewfster: > grabbing its

Would read better as:
> grab its

src/main/java/net/minecraftforge/datafix/ForgeDataFixerEventHandler.java
```diff
@@ -0,0 +1,188 @@
...
+            return forgeDataFixerDelegateHandler.getAllSchemas();
+        }
+
+        //Not compatible environment, return an empty list.
```
pupnewfster: > Not compatible

Would read better as:
> Incompatible

src/main/java/net/minecraftforge/datafix/ForgeDataFixerEventHandler.java
```diff
@@ -0,0 +1,188 @@
...
+            //Check for a compatible schema.
+            if (event.getSchema() instanceof ForgeSchema forgeSchema)
+            {
+                //Rebuild the schema with the mod data.
+                forgeSchema.rebuildSchema(
+                  event.getEntityTypes(),
+                  event.getBlockEntityTypes()
+                );
+            }
+        }, executor).handle((v, t)->t != null ? Collections.singletonList(t): Collections.emptyList());
```
pupnewfster: Missing space before and after the lambda's arrow and also before the `:` in the ternary.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerEventHandler.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    /**
+     * Invoked by the event management system to indicate that it completely fired all events from the event stream.
+     * This indicates now to us that all events for all schemas have been fired and post processed.
```
pupnewfster: > This indicates now to us that all events for all schemas have been fired and post processed.

Would read better as:
> This indicates that all the events for all the schemas have been fired and post processed.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,340 @@
...
+            //Generate a template for the type.
+            templates.add(DSL.check(entry.getKey(), entry.getValue(), getTemplate(entry.getKey())));
+        }
+
+        //Merge them all together into one root recursive type.
+        final Optional<TypeTemplate> availableTypeTemplate = templates.stream().reduce(DSL::or);
+        if (availableTypeTemplate.isEmpty())
+        {
+            //Check for its existence or nuke.
+            return Maps.newHashMap();
```
pupnewfster: Should this be a concurrent map? Given everywhere else TYPES is made as a concurrent map.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,340 @@
...
+            throw new IllegalArgumentException("Data fixer not registered for: " + choiceName + " in " + type.typeName());
+        }
+        return choiceType.types().get(choiceName);
+    }
+
+    @Override
+    public TaggedChoice.TaggedChoiceType<?> findChoiceType(final DSL.TypeReference type)
+    {
+        return getType(type).findChoiceType("id", -1).orElseThrow(() -> new IllegalArgumentException("Not a choice type"));
+    }
```
pupnewfster: I don't think any of these three methods actually need to be overridden because we are overriding `resolveTemplate` and `getType` already.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,340 @@
...
+
+/**
+ * A partial carbon copy of the {@link Schema} class.
+ * But this allows us to delay the setup of the type hierarchy until we need it.
+ */
+class ForgeSchema extends Schema
+{
+    //Can not be final, needs to be initialized on the fly!
+    //Constructor invocation order and the what not!
+    private Map<String, Supplier<TypeTemplate>> TYPE_TEMPLATES = Maps.newConcurrentMap();
```
pupnewfster, line 49: I believe `TYPE_TEMPLATES` can be a normal hash map. It is in `Schema` and in `registerType` it is set to a hash map. If not then the normal hash map variant in `registerType` should be changed to be concurrent as well.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,340 @@
...
+    private Map<String, Type<?>>                TYPES          = Maps.newConcurrentMap();
+    private Map<String, Integer>                RECURSIVE_TYPES = Maps.newConcurrentMap();
```
pupnewfster: I believe `RECURSIVE_TYPES` can be final as it isn't assigned anywhere else, but other than that is there a reason why you are redefining `RECURSIVE_TYPES` (and as a `Map<String, Integer>` rather than `Object2IntMap<String>` like in super) when `Schema.RECURSIVE_TYPES` is protected so I believe you could just make direct use of that. Or is it because of how forge's injection works this need to be concurrent?

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,340 @@
...
+     * Registers a new type to the Schema with its type template.
+     *
+     * @param recursive Indicates if the type is potentially recursive.
+     * @param type The type in question.
+     * @param template Its template.
+     */
+    @Override
+    public void registerType(final boolean recursive, final DSL.TypeReference type, final Supplier<TypeTemplate> template)
+    {
+        if (TYPE_TEMPLATES == null)
```
pupnewfster, line 250: While there is a comment on the field about why it can't be final, having one here saying how this can be null would probably also make sense.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,340 @@
...
+        if (recursive && !RECURSIVE_TYPES.containsKey(type.typeName()))
+        {
+            RECURSIVE_TYPES.put(type.typeName(), RECURSIVE_TYPES.size());
+        }
+    }
+
+    /**
+     * Resets the schema so that the type map can be properly rebuild.
+     */
+    public void resetSchema()
```
pupnewfster: I feel like this potentially should be package private? (Technically it doesn't matter given the class itself is package private but still.)

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    /**
+     * Rebuilds the schema's type map.
+     * This also injects custom types for entities and block entities.
+     *
+     * @param modEntityTypes The mod entity types to inject.
+     * @param modBlockEntityTypes The block entity types to inject.
+     */
+    public void rebuildSchema(
```
pupnewfster: I feel like this should be package private, especially then if we add more support for other types vanilla doesn't, we can easily change the signature. (Technically it doesn't matter given the class itself is package private but still.)

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,340 @@
...
+      final Map<String, Supplier<TypeTemplate>> modEntityTypes,
+      final Map<String, Supplier<TypeTemplate>> modBlockEntityTypes
+    )
+    {
+        //First grab the parent's additional modded types, if the parent is compatible.
+        if (this.parent != null && this.parent instanceof ForgeSchema parentForgeSchema)
```
pupnewfster: This can just be:
```java
if (this.parent instanceof ForgeSchema parentForgeSchema)
```

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,340 @@
...
+        this.MODDED_ENTITY_TYPES.putAll(modEntityTypes);
+        this.MODDED_BLOCK_ENTITY_TYPES.putAll(modBlockEntityTypes);
+
+        //Find all types which the event marked as removed.
+        final Set<String> removedEntityTypes = this.MODDED_ENTITY_TYPES.keySet().stream().filter(key -> this.MODDED_ENTITY_TYPES.get(key) == null).collect(Collectors.toSet());
+        final Set<String> removedBlockEntityTypes = this.MODDED_BLOCK_ENTITY_TYPES.keySet().stream().filter(key -> this.MODDED_BLOCK_ENTITY_TYPES.get(key) == null).collect(Collectors.toSet());
+
+        //And remove those from the type maps.
+        removedEntityTypes.forEach(this.MODDED_ENTITY_TYPES::remove);
+        removedBlockEntityTypes.forEach(this.MODDED_BLOCK_ENTITY_TYPES::remove);
```
pupnewfster: Couldn't all this just be:
```java
this.MODDED_ENTITY_TYPES.values().removeIf(Objects::isNull);
this.MODDED_BLOCK_ENTITY_TYPES.values().removeIf(Objects::isNull);
```

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,208 @@
...
+import java.util.function.Supplier;
+
+/**
+ * Event fired in parallel on the mod event bus to allow for the modification
+ * of a DFU Schemas type signature.
+ *
+ * If used to register a new entity or block entity to a particular version,
+ * then this also will need to be registered to all newer versions to persist properly!
+ * As opposed to vanilla which automatically transfers the entities and block entities from older
+ * schemas to newer ones, this event does not do that and the modder is on his own to implement this properly.
```
pupnewfster: > is on his own to

Would read nicer as:
> is on their own to

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,208 @@
...
+        registerBlockEntity(registryName, DSL::remainder);
+    }
+
+    /**
+     * Registers a new simple (catch-all-nbt) block entity with an item inventory array on the nbt field with the given name to the current schema and all its descendent versions.
+     *
+     * @param registryName The registry name of the block entity.
+     * @param nbtKey The name of the nbt field to where the item array is located.
+     */
+    public void registerSimpleItemInventory(final ResourceLocation registryName, final String nbtKey)
```
pupnewfster, line 106: A clearer name for this method might be `registerSimpleInventoryBlockEntity`

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,208 @@
...
+     *
+     * @param registryName The registry name of the block entity.
+     */
+    public void registerSimpleBlockEntity(final ResourceLocation registryName)
+    {
+        registerBlockEntity(registryName, DSL::remainder);
+    }
+
+    /**
+     * Registers a new simple (catch-all-nbt) block entity with an item inventory array on the nbt field with the given name to the current schema and all its descendent versions.
```
pupnewfster, line 101: Should this be block entity type instead of just block entity?

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,208 @@
...
+     *
+     * @return The version.
+     */
+    public int getVersion()
+    {
+        return getSchema().getVersionKey();
+    }
+
+    /**
+     * Registers a new simple (catch-all-nbt) block entity to the current schema and all its descendent versions.
```
pupnewfster, line 91: Should this be block entity type instead of just block entity?

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,208 @@
...
+    public void registerBlockEntity(final ResourceLocation registryName, final Supplier<TypeTemplate> template)
+    {
+        blockEntityTypes.put(registryName.toString(), template);
+    }
+
+    /**
+     * Removes an already registered block entity from the current schema and all its descendent versions.
+     * This allows for the removal of a custom block entity type when it is registered in a parent version and is no longer needed.
+     *
+     * @param registryName The registry name of the block entity to remove.
```
pupnewfster, line 139: Should this be block entity type instead of just block entity?

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,208 @@
...
+     * @param registryName The registry name of the block entity type.
+     * @param template The supplier of the type template.
+     */
+    public void registerBlockEntity(final ResourceLocation registryName, final Supplier<TypeTemplate> template)
+    {
+        blockEntityTypes.put(registryName.toString(), template);
+    }
+
+    /**
+     * Removes an already registered block entity from the current schema and all its descendent versions.
```
pupnewfster, line 136: Should this be block entity type instead of just block entity?

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,208 @@
...
+     *
+     * @param registryName The registry name of the block entity to remove.
+     */
+    public void removeBlockEntity(final ResourceLocation registryName)
+    {
+        blockEntityTypes.put(registryName.toString(), null);
+    }
+
+    /**
+     * Registers a new simple (catch-all-nbt) entity to the current schema and all its descendent versions.
```
pupnewfster, line 147: Should this be entity type instead of just entity?

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,208 @@
...
+     *
+     * @param registryName The registry name of the entity.
```
pupnewfster, line 149: Should this be entity type instead of just entity?

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,208 @@
...
+     * @param registryName The registry name of the entity type.
+     * @param template The supplier of the type template.
+     */
+    public void registerEntity(final ResourceLocation registryName, final Supplier<TypeTemplate> template)
+    {
+        entityTypes.put(registryName.toString(), template);
+    }
+
+    /**
+     * Removes an already registered entity from the current schema and all its descendent versions.
```
pupnewfster, line 181: Should this be entity type instead of just entity?

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,208 @@
...
+     * This allows for the removal of a custom entity type when it is registered in a parent version and is no longer needed.
+     * 
+     * @param registryName The registry name of the entity to remove.
```
pupnewfster, line 184: Should this be entity type instead of just entity?

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,208 @@
...
+     */
+    public void removeEntity(final ResourceLocation registryName)
+    {
+        entityTypes.put(registryName.toString(), null);
+    }
+
+    /**
+     * An unmodifiable map of all registered additional modded block entity types.
+     * A key maybe mapped to a null value, if it is supposed to be removed from the schema and its descendant versions, overriding the
```
pupnewfster: > A key maybe mapped to a null value, if it is supposed

Would read better as:
> Keys may be mapped to a null value if they are supposed

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,208 @@
...
+     * @return The already registered additional modded block entity types.
+     */
+    public Map<String, Supplier<TypeTemplate>> getEntityTypes()
+    {
+        return entityTypesView;
+    }
+
+    /**
+     * An unmodifiable map of all registered additional modded entity types.
+     * A key maybe mapped to a null value, if it is supposed to be removed from the schema and its descendant versions, overriding the
```
pupnewfster: > A key maybe mapped to a null value, if it is supposed

Would read better as:
> Keys may be mapped to a null value if they are supposed

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
marchermans: Fixed that in already. It needs to be concurrent due to the concurrent firing of the event.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
marchermans: Made it final. It is not a Object2IntMap since those are harder to make properly concurrent (at least I could not find a nice solution) and the performance overhead is not that much of a problem.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
marchermans: This is akin to how Mojang calls this method (it is nearly a blind copy), so I will leave it like this.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
marchermans: No this does not refer to the type, but to the NBT of the actual block entity.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
marchermans: No this does not refer to the type, but to the NBT of the actual block entity.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
marchermans: No this does not refer to the type, but to the NBT of the actual block entity.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
marchermans: No this does not refer to the type, but to the NBT of the actual entity.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
marchermans: No this does not refer to the type, but to the NBT of the actual entity.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
marchermans: No this does not refer to the type, but to the NBT of the actual entity.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
marchermans: No this does not refer to the type, but to the NBT of the actual block entity.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
marchermans: No this does not refer to the type, but to the NBT of the actual entity.

PR #8243, opened 2021-12-02
Add BonusDamage event. Closes #6564

### LINKED ISSUE #6564: Allow enchantments to modify damage based on the entity hit

maxanier:
1.18 version of #8205

Enchantments can only modify the dealt damage based on the entity's CreatureAttribute, not on the entity object itself. Therefore it's not possible to e.g. change the damage based a capability of an attacked player/entity.

## Details
When a player attacks an entity, the dealt damage is calculated in `Player#attack` as follows:
`Total = (0.8*Att * Scale^2 + 0.2 * Att) * Critical + Ench * Scale`
Where
`Att` AttackDamage attribute
`Scale` multiplier from the attack cooldown
`Critical` multiplier based on critical hit
`Ench` Additional damage caused by enchantment

During `LivingAttackEvent` only `Total` is available. The attack cooldown has been reset already, so there is no way to accurately calculate the damage caused by a potential enchantment during that event. Only workaround I can see is to cache the attack cooldown during `AttackEntityEvent` event. But that will get messy, because two different entities (attacker and victim) are involved and the event (or the firing method) is called from many different places.

## Solution
As suggested by @ChampionAsh5357, this PR introduces a new event. This allows modifying the base damage before the scale is applied.
`Total = (0.8*Att * Scale^2 + 0.2 * Att) * Critical + (Ench + Bonus) * Scale`

For Mobs attacking with enchanted weapons `LivingAttackEvent` is sufficient, because they do not have the attack cooldown functionality.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -1242,9 +1242,9 @@ public class ForgeHooks
         MinecraftForge.EVENT_BUS.post(new EntityEvent.EnteringSection(entity, packedOldPos, packedNewPos));
     }
 
-    public static float getDamageBonus(Player player, ItemStack weapon, Entity target, float enchantmentBonus)
+    public static float getDamageBonus(Player player, Entity target, float enchantmentBonus)
     {
-        DamageBonusEvent event = new DamageBonusEvent(player, weapon, target, enchantmentBonus);
+        DamageBonusEvent event = new DamageBonusEvent(player, target, enchantmentBonus);
         MinecraftForge.EVENT_BUS.post(event);
         return event.getBonus();
     }
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -144,6 +144,7 @@ import net.minecraftforge.event.entity.player.AdvancementEvent;
 import net.minecraftforge.event.entity.player.AnvilRepairEvent;
 import net.minecraftforge.event.entity.player.AttackEntityEvent;
 import net.minecraftforge.event.entity.player.CriticalHitEvent;
+import net.minecraftforge.event.entity.player.DamageBonusEvent;
 import net.minecraftforge.event.entity.player.PlayerEvent;
 import net.minecraftforge.event.entity.player.PlayerInteractEvent;
 import net.minecraftforge.event.world.BiomeLoadingEvent;
@@ -1239,6 +1240,13 @@ public class ForgeHooks
     public static void onEntityEnterSection(Entity entity, long packedOldPos, long packedNewPos)
     {
         MinecraftForge.EVENT_BUS.post(new EntityEvent.EnteringSection(entity, packedOldPos, packedNewPos));
+    }
+
+    public static float getDamageBonus(Player player, ItemStack weapon, Entity target, float enchantmentBonus)
+    {
+        DamageBonusEvent event = new DamageBonusEvent(player, weapon, target, enchantmentBonus);
+        MinecraftForge.EVENT_BUS.post(event);
+        return event.getBonus();
     }
 
 }
--- a/patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
@@ -204,7 +204,7 @@
                 f1 = EnchantmentHelper.m_44833_(this.m_21205_(), MobType.f_21640_);
              }
  
-+            f1 = net.minecraftforge.common.ForgeHooks.getDamageBonus(this, this.m_21205_(), p_36347_, f1);
++            f1 = net.minecraftforge.common.ForgeHooks.getDamageBonus(this, p_36347_, f1);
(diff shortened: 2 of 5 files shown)


src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1241,4 +1242,11 @@ public static void onEntityEnterSection(Entity entity, long packedOldPos, long p
         MinecraftForge.EVENT_BUS.post(new EntityEvent.EnteringSection(entity, packedOldPos, packedNewPos));
     }
 
+    public static float getDamageBonus(Player player, ItemStack weapon, Entity target, float enchantmentBonus)
```
ChampionAsh5357: Does the weapon need to be passed in? It's publicly available on the player.

Shadows-of-Fire:
Please merge this with upstream so the CI build succeeds, otherwise looks good.   Might be viable to introduce similar events for the other things that enchantments provide too (such as protection).  Though that can go in other PR's.

maxanier:
I am not sure why the CI build fails. I have rebased onto 1.18.x and locally it builds and runs fine.
I can't see the CI output as I am not willing to grant this Jenkins instance access to my private repositories.

ChampionAsh5357 (approved):

src/main/java/net/minecraftforge/common/ForgeHooks.java
Elenterius: imo yes,
in the context of Mixins, calling `player#getMainHandItem` doesn't guarantee I get the same ItemStack that was used to get the initial damage bonus (e.g. a mod that allows dual wielding attacks could just pass in the offhand stack into `EnchantmentHelper#getDamageBonus`), so its better to have it as a explicit parameter that way anyone doing Mixins knows they have to supply the correct ItemStack to the damage bonus event as well

Elenterius:
In the context of getting the damage bonus from Enchantments i have to disagree with your claim:

> For Mobs attacking with enchanted weapons `LivingAttackEvent` is sufficient, because they do not have the attack cooldown functionality.

This only works for **vanilla** Entities because you know where to find the ItemStack that was used for the attack:
- vanilla Mobs use the Mainhand slot
- ThrownTrident uses a cached ItemStack copy

Now compare that to modded entities, they could use any arbitrary ItemStack to get the damage bonus for their attack. You have no way of knowing what ItemStack was used. 

This way custom damage bonus Enchantments, that are compatible with all modded entities, aren't possible in a sensible manner.

**Solution:**
- don't restrict the damage bonus event to players and widen it so it accepts any `Entity` instance as the attacker
- pass in the ItemStack into the event (the one that was used for `EnchantmentHelper#getDamageBonus`)
- call the damage bonus event from `Mob#doHurtTarget` and `ThrownTrident#onHitEntity` or deprecate getDamageBonus

What i would prefer to have:
```Java

	public class EnchantmentHelper {
               
                //vanilla, this method should only be called by ItemStack#getTooltipLines
		@Deprecated 
		public static float getDamageBonus(ItemStack stack, MobType creatureAttribute) {...}

                //forge version
		public static float getDamageBonus(ItemStack weapon, Entity attacker, Entity victim, MobType mobType) {
			float initialDamageBonus = getDamageBonus(weapon, mobType);
			return ForgeHooks.getDamageBonus(attacker, weapon, victim, initialDamageBonus);
		}

	}

```

--- 
btw you mean `LivingHurtEvent`, the `LivingAttackEvent` event doesn't allow the modification of the damage

maxanier:
Sounds reasonable. Does require more patches though (~5 patches). Before I spend too much time on this, I would like some feedback from the Forge team if such PR has a chance of being accepted

Shadows-of-Fire:
My opinion is that this could be a method on Enchantment instead of an event - event implies you want to change the damage of other enchantments, (which should probably be handled later, at the damage event).
You could make an extension interface to enchantment `IForgeEnchantment` with the single method `public default float getDamageBonus(int level, @Nullable Player player, ItemStack weapon, @Nullable Entity target)` which calls `getDamageBonus` by default.

I only suggest that the entity parameters be nullable because some vanilla call sites for `getDamageBonus` do not have an entity context to pass through, but this allows you to setup a valid contract regardless of that, and still replace those calls with the new one.

Ideally you would also deprecate `getDamageBonus` since this method would replace it. In light of nullability I'm wondering if the new method should also retain a `MobType` parameter.

PR #8244, opened 2021-12-03
Add bulk (stack size ignoring and slot ignoring) functions to IItemHandler

### LINKED ISSUE #8175: Bulk Transfer Support in IItemHandler
## Problem

I'll use the term "large item quantities" for any item stack larger than the stack's max stack size.
Examples for mods dealing in large item quantities are storage drawers, barrels, RS, AE2, bins, and many similar storage systems.

The current `IItemHandler` contract has limited support for such large item quantities by allowing `getStackInSlot` to return a stack with a stack size that exceeds the item's max stack size (See https://github.com/MinecraftForge/MinecraftForge/blob/1.17.x/src/main/java/net/minecraftforge/items/IItemHandler.java#L38).

### Extracting Large Amounts
`extractItem` specifically states that the returned item stack must not exceed the item's maximum stack size.

This has two consequences:

1) it's not possible to simulate the extraction of large item quantities, since splitting the simulation into chunks of the maximum stack size is pointless (every operation will have the same outcome since the state of the underlying inventory does not change inbetween)

2) to actually extract large item quantities, a mod has to loop and extract chunks of max stack size from the same slot

### Inserting Large Amounts
`insertItem` does not specifically state if the inserted stack can exceed the item's maximum stack size or not, so a mod can presumably decide to deal with it or not.

## Proposal

To provide better compatibility between mods that consume or provide large item quantities, I'd propose the following additional methods on IItemHandler. These methods also allow mods to provide optimized versions of what currently are O(n) (where n = slots) operations.

```java
/**
 * Tries to extract as many items that can be stacked with the given template item as requested from this
 * item handler. Will extract regardless of slot and does not respect {@link ItemStack#getMaxStackSize()}.
 *
 * @param template The template item for the items that should be extracted. Only items matching this templates item,
 *             tags and caps will be extracted. {@link ItemStack#EMPTY} is not a valid template.
 *             The handler must not modify this template.
 * @param maxAmount The maximum amount to extract. Must be 0 or greater.
 * @param simulate Whether the operation should actually be performed or only simulated.
 * @return The stack of items that could be extracted. This stack does not adhere to {@link ItemStack#getMaxStackSize()}.
 */
default ItemStack bulkExtractItem(ItemStack template, int maxAmount, boolean simulate) {
    Preconditions.checkArgument(!template.isEmpty(), "what may not be empty");
    Preconditions.checkArgument(maxAmount >= 0, "maxAmount is negative");

    int extracted = 0;
    for (int i = 0; i < getSlots(); ++i) {
        if (ItemHandlerHelper.canItemStacksStack(getStackInSlot(i), template)) {
            extracted += extractItem(i, maxAmount, simulate).getCount();
            if (extracted >= maxAmount) {
                break;
            }
        }
    }

    if (extracted > 0) {
        var result = template.copy();
        result.setCount(extracted);
        return result;
    } else {
        return ItemStack.EMPTY;
    }
}

/**
 * Tries to insert as many of the given item as possible into this item handler, spreading it across
 * slots as needed.
 *
 * @param what The stack that should be inserted. {@link ItemStack#EMPTY} is not valid.
 *             Ownership of this stack is transferred to the item handler.
 *             This stack does not need to adhere to {@link ItemStack#getMaxStackSize()}.
 * @param simulate Whether the operation should actually be performed or only simulated.
 * @return The stack of items that could not be inserted.
 */
default ItemStack bulkInsertItem(ItemStack what, boolean simulate) {
    Preconditions.checkArgument(!what.isEmpty(), "what may not be empty");

    return ItemHandlerHelper.insertItem(this, what, simulate);
}
```

An alternative to implementing this as default methods would be adding a completely different capability for it, but it seems more suited to IItemHandler in general.

I will provide a PR for this in case this proposal not outright rejected.

Technici4n:
Implements #8175.

This PR extends IItemHandler for two uses cases:
- handlers that can store more than max stack size per slot.
- handlers that can efficiently find a requested resource without iterating through all slots.

### More than max stack size
The javadoc of `insertItem` is clarified to mention that all of these should not limit themselves to max stack size.
The javadoc of `extractItem` is clarified to mention that it should limit itself to the max stack size.
`extractItemIgnoreStackSize` is added for handlers and callers that support amounts larger than the max stack size.

This is mostly a correctness/feature augmentation to allow mods that deal with large item quantities to work better (AE2, RS, drawers, deep storage units, etc).

### Slot ignoring
`bulkInsertItem` and `bulkExtractItem` are added to allow for faster insertion or extraction operations when the caller doesn't need explicit per-slot access (that should be the case most of the time).
They ignore the stack size limit because they are mostly meant for mods that can provide very large amounts of items, for example colossal chests, ender rift, or the aforementioned AE2, RS, etc.

Limitation: currently, there is no way to know if a handler supports these bulk functions, or if the default implementation is used. Should a `boolean supportsBulkTransfer()` function be added too, or is that too much?

This is a performance-oriented addition. But note that the default `bulkInsertItem` replaces the helper in `ItemStackHelper`, and can be quite convenient for modders.

### Other stuff
Fixed `ItemStackHandler` not always copying the passed stack. By the contract of `insertItem`, the passed stack should never be modified (or stored, obviously).

### DIFF
--- a/src/main/java/net/minecraftforge/items/IItemHandler.java
+++ b/src/main/java/net/minecraftforge/items/IItemHandler.java
@@ -110,8 +110,7 @@ public interface IItemHandler
 
     /**
      * Extracts an ItemStack from the given slot.
-     * If there is more than the max stack size in the current slot and more than the max stack size is requested,
-     * then more than the max stack size may be returned.
+     * The returned stack's count may not exceed the max stack size of the item stack.
      * <p>
      * The returned value must be empty if nothing is extracted,
      * otherwise its stack size must be less than or equal to {@code amount} and {@link ItemStack#getMaxStackSize()}.
@@ -127,13 +126,18 @@ public interface IItemHandler
     ItemStack extractItem(int slot, int amount, boolean simulate);
 
     /**
-     * Extracts an ItemStack from the given slot, but limiting the amount extracted to {@link ItemStack#getMaxStackSize()}.
+     * Extracts an ItemStack from the given slot, completely ignoring the max stack size.
+     * If there is more than the max stack size in the current slot and more than the max stack size is requested,
+     * then more than the max stack size may be returned.
+     * <p>
+     * By default, this calls the usual extractItem that limits the amount to the max stack size.
+     * Handlers that can store more items in each slot should override both methods.
+     * </p>
      * @see #extractItem
      */
     @Nonnull
-    default ItemStack limitedExtractItem(int slot, int amount, boolean simulate) {
-        int limitedAmount = Math.min(amount, getStackInSlot(slot).getMaxStackSize());
-        return extractItem(slot, limitedAmount, simulate);
+    default ItemStack extractItemIgnoreStackSize(int slot, int amount, boolean simulate) {
+        return extractItem(slot, amount, simulate);
     }
 
     /**
@@ -159,7 +163,7 @@ public interface IItemHandler
         {
             if (ItemHandlerHelper.canItemStacksStack(getStackInSlot(i), template))
             {
-                extracted += extractItem(i, amount - extracted, simulate).getCount();
+                extracted += extractItemIgnoreStackSize(i, amount - extracted, simulate).getCount();
                 if (extracted >= amount) break;
             }
         }
--- a/src/main/java/net/minecraftforge/items/IItemHandler.java
+++ b/src/main/java/net/minecraftforge/items/IItemHandler.java
@@ -19,6 +19,7 @@
 
 package net.minecraftforge.items;
 
+import com.google.common.base.Preconditions;
 import net.minecraft.world.Container;
 import net.minecraft.world.item.ItemStack;
 import net.minecraftforge.fluids.FluidStack;
@@ -26,7 +27,7 @@ import net.minecraftforge.fluids.capability.IFluidHandler;
 
 import javax.annotation.Nonnull;
 
-
+// TODO: replace all the boolean parameters by FluidAction (renamed to also fit items and energy)
 public interface IItemHandler
 {
     /**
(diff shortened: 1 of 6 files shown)


shartte:
I got hung up on the TODOs but remembered later: It's inconsistent right now because `FluidAction` is used on `IFluidHandler`, while `booleans` are used in `IItemHandler` / `IEnergyStorage`.

src/main/java/net/minecraftforge/items/IItemHandler.java
```diff
@@ -61,12 +62,13 @@
     /**
      * <p>
      * Inserts an ItemStack into the given slot and return the remainder.
-     * The ItemStack <em>should not</em> be modified in this function!
```
pupnewfster, line 64: This sounds like a very annoying contract change, is there a reason for it?

Technici4n: This is actually a documentation update, as the current documentation has been wrong for a long time. This comes from an old PR of mine https://github.com/TheCurle/MinecraftForge/pull/17.

Technici4n: image
As you can see in ItemStackHandler, if `reachedLimit` is false the stack is directly set.

Technici4n: Reverted the javadoc change, and fixed the bug in ItemStackHandler instead.

diesieben07:
I don't think changing the contract of normal `insert` and `extract` is a good idea. Various things might not operate correctly with stack sizes > 64, so those "broken" stacks should be kept to a minimal API, if they are used at all. Otherwise these stacks might "leak out" and end up in places that do not expect them.
So the bulk method default implementations should split/combine the items to their max stack size.

shartte:
@diesieben07 I don't actually see a change on the `extract` contract? 

Regarding `insert`: A lot of the existing storage mods (RS/AE2) will happily pass larger stacks to `insert`, to insert into storage drawers, for example. 

They rely on the fact that essentially all implementations of `IItemHandler#insert` will have to clamp the amount to the remaining slot space, which automatically takes care of adhering to max stack size.

I think that's what most if not all mods do. (Forge's default implementation `ItemStackHandler` does this in `getStackLimit`).

sciwhiz12:
Hello! @Technici4n, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.19.2 (or leave an explanation why this PR's changes are not applicable to 1.19.2). If we do not hear back from you after some time, this PR will be closed for inactivity.

Technici4n:
I'm not gonna be working on this PR due to:
- me changing some logic in AE2 that makes this less critical for me,
- the lack of community and team interest.

However anyone should feel free to pick it up, and also consider the changes proposed in #8990 to improve `IItemHandler` overall.

PR #8245, opened 2021-12-03
Refactor parseAxisRotation

coldu:
Refactor parseAxisRotation method by transform if else structure to switch structure

### DIFF
--- a/src/main/java/net/minecraftforge/common/model/TransformationHelper.java
+++ b/src/main/java/net/minecraftforge/common/model/TransformationHelper.java
@@ -324,19 +324,12 @@ public final class TransformationHelper
             Quaternion ret;
             try
             {
-                if (entry.getKey().equals("x"))
-                {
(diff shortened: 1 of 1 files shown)


TheCurle:
Trivial PRs are not accepted on this repository.

coldu:
Hello, I would like to know how to help improve code quality. If this is not possible, could you add this information to your wiki? (If I didn't see it, sorry)

TheCurle:
"Code Quality" is not a metric we use often.

Every commit to this repository starts a lengthy build process that uses a lot of processing power and storage space to perform.
For low effort, tiny changes like this which has zero effect on a modder or an end-user, it's simply not worth the time, effort, energy or cost to entertain.

We have an omnibus PR coming up which collates smaller changes, and this may be considered for that.

Until then, we're not too worried about "code quality" with regards to a 4 entry if ladder.

PR #8246, opened 2021-12-03
Added new 1.18 biomes to the BiomeDictionary

GirafiStudios:
- Added BiomeDictionary tags for the 1.18 biomes
- Added a new UNDERGROUND BiomeDictionary tag, for a way to differentiate between the new cave biomes & a normal biome, with the BiomeDictionary.


src/main/java/net/minecraftforge/common/BiomeDictionary.java
```diff
@@ -328,6 +329,12 @@ private static void registerVanillaBiomes()
         addTypes(Biomes.SAVANNA_PLATEAU, HOT, SAVANNA, PLAINS, SPARSE, RARE, OVERWORLD, PLATEAU);
         addTypes(Biomes.BADLANDS, MESA, SANDY, DRY, OVERWORLD);
         addTypes(Biomes.WOODED_BADLANDS, MESA, SANDY, DRY, SPARSE, OVERWORLD, PLATEAU);
+        addTypes(Biomes.MEADOW, PLAINS, PLATEAU, OVERWORLD);
```
ChampionAsh5357: Isn't this part of the mountain category?

src/main/java/net/minecraftforge/common/BiomeDictionary.java
```diff
@@ -328,6 +329,12 @@ private static void registerVanillaBiomes()
...
+        addTypes(Biomes.GROVE, COLD, CONIFEROUS, FOREST, SNOWY, MOUNTAIN, OVERWORLD);
+        addTypes(Biomes.SNOWY_SLOPES, COLD, SPARSE, SNOWY, MOUNTAIN, OVERWORLD);
+        addTypes(Biomes.JAGGED_PEAKS, COLD, SPARSE, SNOWY, MOUNTAIN, OVERWORLD);
+        addTypes(Biomes.FROZEN_PEAKS, COLD, SPARSE, SNOWY, HILLS, OVERWORLD);
```
ChampionAsh5357: Isn't this part of the mountain category?

src/main/java/net/minecraftforge/common/BiomeDictionary.java
GirafiStudios: Yes, but it´s very low. Just goes up a little bit and the flattens out. Like a plateau.

src/main/java/net/minecraftforge/common/BiomeDictionary.java
GirafiStudios: It generates lower than the other mountain biomes, wasn´t sure if Hills or Mountain would fit best. Will change it to Mountain :)

src/main/java/net/minecraftforge/common/BiomeDictionary.java
GirafiStudios: Could it perhaps makes sense to have both Plateau and Mountain? As it is indeed part of the vanilla mountain category, but it´s "shaped" like a plateau.

FlashFyreMods:
Regarding the categories of the mountain slope biomes (meadow, grove, snowy slopes) it might be better for there to be a "Mountain Peak" and "Mountain Slope" tag and assign them to the latter, and the mountain peak biomes (jagged, stony and frozen peaks) to the former. And remove the plateau tag from all the mountain biomes.

And also remove the generic mountain tag entirely, and make sure that biomes like the windswept hills, forest, gravelly hills and savanna don't have the mountain tag, only the hills tag.

There are some other weird tags too, I don't think the flower forest, old growth birch forest or ice spikes should have the hills tag.

FlashFyreMods:
Took another look at the overworld biome builder and I'm gonna amend my previous message and make it easier to understand.
- Remove "mountain" tag
- Jagged, stony and frozen peaks should have mountain_peak. No other biomes should have this tag.
- Snowy slopes and grove should have mountain_slope as they only generate as slopes.
- Meadows, savanna plateaus and wooded badlands should have both mountain_slope and plateau as they generate both as slopes and plateaus. No other biomes should have this tag now.
- All windswept biomes should have the hills tag, not any mountain tag.
- The hills tag should be removed from the ice spikes, flower forest and old growth birch forest biomes. So the only biomes as hills are the windswept biomes.

GirafiStudios:
Made adjustments based on what Fla5hFyre suggested.
Peak and Slope is added as new tags, instead of Mountain. Mountain is still kept as an overall tag that includes both Peak and Slope.

FlashFyreMods:
Make sure the savanna plateau and the wooded badlands also have the slope tag.

ChampionAsh5357 (approved):
I can agree with this.

PR #8249, opened 2021-12-03
Fix datagen test for sounds definitions provider 

sciwhiz12:
This PR fixes the datageneration test by modifying the output for the sound definitions provider subtest to match the expected vanilla output, and fixing two mistakes in the error checking code. Specifically, it reorders the entries for `item.trident.return` to match the vanilla sound definitions. The error checking code's inputs are now in the correct order (middle of the `Triple` is the vanilla definition, right of the `Triple` is the generated definition), and also checks if the generated definition is `null`.

This also adds in some datagenerated files from the tests which seem to not have been included previously.


ChampionAsh5357 (approved):
Looks good.

Mysterious-Dev (approved):

PR #8250, opened 2021-12-04
Merge values of defaulted optional tags

### LINKED ISSUE #7570: Add toString method to vanilla Tag class

sciwhiz12:
This PR fixes #7570 by merging the default values of a defaulted optional tag when defined in multiple places (whether within the same mod or across multiple mods). For more details as why the linked issue happens and why this solution fixes it, see my comment at https://github.com/MinecraftForge/MinecraftForge/issues/7570#issuecomment-978902329.


patches/minecraft/net/minecraft/tags/StaticTagHelper.java.patch
```diff
@@ -63,7 +63,7 @@
 +         OptionalNamedTag<T> optionalNamedTag = (OptionalNamedTag<T>) namedTag;
 +         optionalNamedTag.defaulted = true;
 +         return optionalNamedTag.resolveDefaulted();
-+      }));
++      }, (existingTag, newTag) -> SetTag.m_13222_(com.google.common.collect.ImmutableSet.<T>builder().addAll(existingTag.m_6497_()).addAll(newTag.m_6497_()).build())));
```
pupnewfster: Part of me wonders if it might be worthwhile to only create a new tag if both tags are not empty and otherwise just use the one that isn't empty or just return either one if they are both empty, though honestly that might not really matter.

On another note something that definitely should be changed on this line for readability, is there is no need to be using the FQN for ImmutableSet as it is already imported due to vanilla using it in this class.

sciwhiz12: That would be worthwhile if we expect an optional tag to be referenced many times (by multiple mods) and be defaulted, as currently this would make N-1 `SetTag` instances for N (defaulted) optional tag references. But it doesn't seem to be a big enough problem to warrant optimizing that.

And I didn't notice that MC already imports `ImmutableSet`. :eyes: (I've been bitten by silent auto-importing of those in the past while adding to patched files, so I wanted to head off that trouble.)

PR #8251, opened 2021-12-05
[1.18.x] Resource Key Tags for biomes, dimensions, dimension types, and noise settings

Commoble:
This is a continuation of [half of this PR from 1.17.x](https://github.com/MinecraftForge/MinecraftForge/pull/7917). The other half (pertaining to the dynamic-registries-loaded event) will be ported in a separate PR at a later time.

## What's the PR add?
* JSON-definable tags for Resource Keys of dynamic registry objects (biomes and other json-definable worldgen data)
  * These are implemented in a similar manner to Function Tags. They are only loaded on the server thread (via reloadable data); they are not loaded on or synced to the render thread.
  * Because they load before dynamic registry objects, Resource Key Tags are only sets of resource keys of those objects, not the objects themselves. Validation is not performed, so these tags may contain keys for objects that do not exists. The tags are therefore more effective at checking whether an object belongs to the tag than at providing collections of objects.
  * Resource Key Tags can be referred to while dynamic registry objects or dimensions are loading (though wariness should be had that dynamic registry objects and dimensions are *not* reloaded when servers `/reload`), or during the server startup lifecycle (ServerAboutToStartEvent and later).
  * Resource Key Tags *cannot* be used during the BiomeLoadingEvent, as they load after the event (the BiomeDIctionary should still be used for that).
  * Resource Key Tags are not meant to be used for static registry objects (blocks, items, etc). Traditional tags should be used for that as they have more features and validation.
* Builtin forge Resource Key Tags for biomes (loosely based on the BiomeDictionary), dimensions or levelstems (vanilla uses both of these terms for the same thing), dimension types, and noise settings (which contain information about what structures can spawn where).
* Ability to load Resource Key Tags for any of the vanilla dynamic registries
* Ability to specify (via an in-mod API) that Resource Key Tags should be loaded from additional folders
* Datagen for Resource Key Tags

## Public APIs
  * This allows mods to notify the tag loader to load tags from custom directories. Forge automatically marks vanilla's dynamic registries for loading resource key tags for (see ForgeMod for an example of doing this), as well as the dimension-or-levelstem-registry. This is safe to call during parallel modloading and can be called from a mod constructor. Mods that add their own loading directories should namespace them with their modid to avoid collisions. If two mods register two different directories for the same master registry key, an error will be logged (only one directory can be loaded per master registry key as the key is used to determine where to load key tags from).
 
It should be emphasized that ResourceKeyTags' values are `100% optional`. Resource Key Tags load before dynamic registry objects do, so a biome tag *cannot* verify whether any of the biomes their keys represent actually exist. Users of resource key tags should be wary of this when iterating over their contents.

## Other Concerns
* This PR does not deprecate BiomeDictionary, as the BiomeLoadingEvent is still currently the best place for mods to modify biomes in (and biome key tags can't be used in the BiomeLoadingEvent). This will be addressed in a separate PR (the other half of #7917) that will add a datapack-sensitive timing for modifying biomes in (after which Resource Key Tags will be able to be used for specifying which biomes should have which modifications, and the BiomeDictionary will be able to be deprecated).
* While Resource Key Tags don't strictly *need* to be in forge due to not requiring any patches, having biome tags in forge provides a centralized implementation of an often-requested feature, rather than having competing implementations of tag loaders that mods must choose from or support all of.


src/test/java/net/minecraftforge/debug/misc/ResourceKeyTagsTest.java
```diff
@@ -0,0 +1,52 @@
...
+    {
+        IEventBus forgeBus = MinecraftForge.EVENT_BUS;
+        forgeBus.addListener(this::onRegisterCommands);
+    }
+    
+    private void onRegisterCommands(RegisterCommandsEvent event)
+    {
+        event.getDispatcher().register(
+            Commands.literal("resource_key_tags_test_am_i_in_a_swamp")
+            .executes(context ->{
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/misc/ResourceKeyTagsTest.java
```diff
@@ -0,0 +1,52 @@
+package net.minecraftforge.debug.misc;
```
ChampionAsh5357, line 20: Missing license header.

src/main/java/net/minecraftforge/common/data/ResourceKeyTagsProvider.java
```diff
@@ -0,0 +1,292 @@
...
+        /**
+         * Specify that lower-priority datapack's tag jsons with the same ID should be ignored instead of being merged into your tag json.
+         * (the default is replace=false, indicating that the tags should be merged -- invoke this to set replace=true)
+         * @return this
+         */
+        public Builder<T> replace() {
+            return replace(true);
+        }
+
+        public Builder<T> replace(boolean value) {
```
ChampionAsh5357: New line braces.

src/main/java/net/minecraftforge/common/data/ResourceKeyTagsProvider.java
```diff
@@ -0,0 +1,292 @@
...
+            }
+            return this;
+        }
+        
+        /**
+         * Specify that lower-priority datapack's tag jsons with the same ID should be ignored instead of being merged into your tag json.
+         * (the default is replace=false, indicating that the tags should be merged -- invoke this to set replace=true)
+         * @return this
+         */
+        public Builder<T> replace() {
```
ChampionAsh5357: New line braces.

src/main/java/net/minecraftforge/common/data/ResourceKeyTagsProvider.java
```diff
@@ -0,0 +1,292 @@
...
+
+    @Override
+    public void run(HashCache cache) throws IOException
+    {
+        this.tagBuilders.clear();
+        this.addTags();
+        Tag<ResourceKey<T>> tag = SetTag.empty();
+        Function<ResourceLocation, Tag<ResourceKey<T>>> tagFromID = id -> this.tagBuilders.containsKey(id) ? tag : null;
+        Function<ResourceLocation, ResourceKey<T>> resourceKeyFromID = ResourceKey.elementKey(this.registryKey);
+        this.tagBuilders.forEach((id,builder) ->{
```
ChampionAsh5357: New line braces.

src/main/java/net/minecraftforge/common/data/ForgeResourceKeyTagsProvider.java
```diff
@@ -0,0 +1,234 @@
...
+        }.run(cache);
+    }
+
+    @Override
+    public String getName()
+    {
+        return "Forge Resource Key Tags Provider";
+    }
+    
+
```
ChampionAsh5357: Remove extra line.

ChampionAsh5357 (changes requested):
I don't believe I see any issues on this since it's purely in Forge. This can be submitted for review after the formatting has been fixed.

src/test/java/net/minecraftforge/debug/misc/ResourceKeyTagsTest.java
Commoble: Fixed in next push

src/main/java/net/minecraftforge/common/data/ResourceKeyTagsProvider.java
Commoble: Fixed in next push

src/main/java/net/minecraftforge/common/data/ResourceKeyTagsProvider.java
Commoble: Fixed in next push

src/main/java/net/minecraftforge/common/data/ResourceKeyTagsProvider.java
Commoble: Fixed in next push

src/main/java/net/minecraftforge/common/data/ForgeResourceKeyTagsProvider.java
Commoble: Fixed in next push

src/test/java/net/minecraftforge/debug/misc/ResourceKeyTagsTest.java
Commoble: Added in next push

ChampionAsh5357 (approved):
Should be good to go.

Commoble:
Closing this as universal vanilla tags in 1.18.2 are expected to supercede this

PR #8253, opened 2021-12-05
[1.18] Fix ChunkWatchEvents not being fired

XFactHD:
This PR reactivates the call to `ForgeEventFactory.fireChunkWatch()` in `ChunkMap#updateChunkTracking()` which was marked as TODO in the porting process and then presumably missed.
The code surrounding the call to the Forge hook has not changed apart from a type change on the packet parameter which is not used by the hook anyway.

According to the existing `ChunkWatchEventTest` test mod the event works as intended.


TheCurle:
God damnit. I really need to be more careful with my capitalization when writing TODOs.

PR #8254, opened 2021-12-06
Add a RenderArmEvent

pupnewfster:
The purpose of this PR is to add a more targeted version of `RenderHandEvent` for mods that specifically want to be adjusting the rendering of a player's arm, such as to make "long sleeve" armor be able to render on the arm. While this is possible via `RenderHandEvent`, it requires copy pasting large portions of vanilla code for purposes of:
- Performing all the transformations to the `PoseStack` to get it rendering in the correct place
- Rendering maps (as vanilla renders arm(s) when holding maps)

Additionally this more targeted PR will allow mods making changes to hands to natively then support any mods that say make the offhand always render to have it render properly with the adjustments desired.

### DIFF
--- a/src/main/java/net/minecraftforge/client/event/RenderArmEvent.java
+++ b/src/main/java/net/minecraftforge/client/event/RenderArmEvent.java
@@ -29,7 +29,7 @@ import net.minecraftforge.eventbus.api.Event;
 /**
  * This is a more targeted version of {@link RenderHandEvent} event that is fired specifically when
  * a player's arm is being rendered in first person, and should be used instead if the desired
- * outcome is just to replace the rendering of the arm, such as to add make armor render on it or
+ * outcome is just to replace the rendering of the arm, such as to make armor render on it or
  * instead of it.
  *
  * This event is fired on the {@link net.minecraftforge.common.MinecraftForge#EVENT_BUS}
@@ -77,7 +77,7 @@ public class RenderArmEvent extends Event
     }
 
     /**
-     * @return the Client player that is having their arm rendered. In general this will be the same as {@link net.minecraft.client.Minecraft#getInstance()#getPlayer()}.
+     * @return the client player that is having their arm rendered. In general this will be the same as {@link net.minecraft.client.Minecraft#player}.
      */
     public AbstractClientPlayer getPlayer()
     {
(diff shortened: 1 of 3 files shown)


marchermans (approved):

src/main/java/net/minecraftforge/client/event/RenderArmEvent.java
```diff
@@ -0,0 +1,86 @@
...
+import net.minecraft.client.renderer.MultiBufferSource;
+import net.minecraft.world.entity.HumanoidArm;
+import net.minecraftforge.eventbus.api.Cancelable;
+import net.minecraftforge.eventbus.api.Event;
+
+/**
+ * This is a more targeted version of {@link RenderHandEvent} event that is fired specifically when
+ * a player's arm is being rendered in first person, and should be used instead if the desired
+ * outcome is just to replace the rendering of the arm, such as to add make armor render on it or
+ * instead of it.
```
SizableShrimp, line 33: This phrasing is a little weird, "add make" looks incorrect.

src/main/java/net/minecraftforge/client/event/RenderArmEvent.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public int getPackedLight()
+    {
+        return packedLight;
+    }
+
+    /**
+     * @return the Client player that is having their arm rendered. In general this will be the same as {@link net.minecraft.client.Minecraft#getInstance()#getPlayer()}.
+     */
+    public AbstractClientPlayer getPlayer()
```
SizableShrimp, line 82: Formatting here is also slightly off with capitalization of Client. Is this javadoc link valid?

SizableShrimp (changes requested):

src/main/java/net/minecraftforge/client/event/RenderArmEvent.java
pupnewfster: Turns out when I changed my mind how to phrase it I missed removing a word.

src/main/java/net/minecraftforge/client/event/RenderArmEvent.java
pupnewfster: Fixed, and turns out it wasn't I just got baited by how IntelliJ was autocompleting stuff and part of it was referencing the method the javadoc is on itself instead of the proper object.

PR #8256, opened 2021-12-07
(1.18) Fixed KeyMappings only checking if they conflict with themselves.

jaredlll08:


sciwhiz12 (approved):
This seems to be a regression introduced during the 1.16 update and patching cycle: https://github.com/MinecraftForge/MinecraftForge/commit/6d15febdb35b5676b1795b2965d6eb527bb336e6#diff-eecc1c634c74c835e5eafda997032fc0c27131444bd7aec7562137933c765471L47-L49

The second argument was errorneously changed to be the same object that `hasKeyCodeModifierConflict` is called upon, which means that it checked if the keybind conflicts with itself since 1.16.

PR #8259, opened 2021-12-08
Renderable API

gigaherz:
This PR adds a Renderable API to allow easier rendering of OBJ and other custom models from Entity and BlockEntity renderers.

This is a redesign of a discarded section of my initial model system rewrite, back in 1.14.

In order to use it with the OBJ loader, you can use OBJLoader.INSTANCE.loadModel to get the OBJModel,
and then call OBJModel#bakeRenderable() to get the object to render with.

TODO:

- [x] Add a test mod
- [x] Consider if the OBJ renderable can make use of any dynamic context value.
  * ~~One possibility for this, may be to provide a way to get transformation matrices per part, such that the model ban be animated.~~ DONE
  * No more feedback after multiple weeks. Have to assume no one has any further use case for this yet.
- [x] Test that the part transformation allows making animations.


gigaherz:
State as of the latest commit: 

- The test mod draws the sugar_glider.obj model from RenderLevelLast.
- The feature supports supplying a map of textures so it can resolve names of texture aliases.
- The texture locations get translated to actual texture paths, so they work from the render types.

java_2021-12-08_23-15-11

gigaherz:
Animation! (in code)

src/main/java/net/minecraftforge/client/model/renderable/SimpleRenderable.java
```diff
@@ -0,0 +1,183 @@
...
+        private Part currentPart;
+
+        private List<Part> parentList;
+
+        public Builder()
+        {
+            parentList = renderable.parts;
+        }
+
+        public Builder pushPart(String partName)
```
marchermans: This concept feels iffy to me. Always has.

I generally prefer it if the part builder is either his own instance and not the same instance that is prefixing or even an instance that support auto closing so that a try with resources block could be used to wrap the part construction and handle the build code automatically

marchermans (commented):
Generally good system, this is just me being nitpicky over how the builder api for parts works.

In this particular concept, you are not building a stack so push and pop make little to no sense to me. Could you see if a different more clearly defined api makes more sense?

src/main/java/net/minecraftforge/client/model/renderable/SimpleRenderable.java
gigaherz: I made it that way to avoid allocating extra objects, but I could switch to individual builders. I'll play around with it.

marchermans: Personally allocating the additional objects should not matter here.
The renderer should cache the renderable anyway, right?

src/main/java/net/minecraftforge/client/model/StandaloneModelConfiguration.java
```diff
@@ -0,0 +1,119 @@
...
+    {
+        return ItemTransforms.NO_TRANSFORMS;
+    }
+
+    @Override
+    public ModelState getCombinedTransform()
+    {
+        return BlockModelRotation.X0_Y0;
+    }
+}
```
ChampionAsh5357: Line feed at EOF.

src/main/java/net/minecraftforge/client/model/renderable/SimpleRenderable.java
```diff
@@ -0,0 +1,161 @@
...
+
+    public static class Builder
+    {
+        private final SimpleRenderable renderable = new SimpleRenderable();
+
+        private Builder()
+        {
+        }
+
+        public PartBuilder child(String name)
```
ChampionAsh5357: I don't see any issues with this. Though the chainability of the builder is lost if there is no reference back to the builder from the internal part.

src/main/java/net/minecraftforge/client/model/renderable/SimpleRenderable.java
```diff
@@ -0,0 +1,161 @@
...
+    public static class PartBuilder
+    {
+        private final Part part;
+
+        private PartBuilder(Part part)
+        {
+            this.part = part;
+        }
+
+        public PartBuilder child(String name)
```
ChampionAsh5357: Same thing here, no reference back to the part from the internal part.

src/main/java/net/minecraftforge/client/model/renderable/BakedRenderable.java
```diff
@@ -0,0 +1,58 @@
+package net.minecraftforge.client.model.renderable;
```
ChampionAsh5357, line 20: Missing license header.

src/main/java/net/minecraftforge/client/model/renderable/BakedRenderable.java
```diff
@@ -0,0 +1,58 @@
...
+import net.minecraft.client.resources.model.BakedModel;
+import net.minecraft.core.Direction;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.inventory.InventoryMenu;
+import net.minecraftforge.client.model.data.IModelData;
+
+import java.util.Random;
+import java.util.function.Function;
+
+public class BakedRenderable implements IRenderable<IModelData>
```
ChampionAsh5357, line 42: Could probably use some documentation.

ChampionAsh5357 (changes requested):
For the scope of what the system should be, I think its fine, just some formatting and builder additions.

src/main/java/net/minecraftforge/client/model/renderable/SimpleRenderable.java
gigaherz: I was hoping to avoid having an ungodly recursive generic but I guess it's not as bad as it looked in my head.
It does however mean that if you are a few levels deep, the actual type of PartBuilder will be `PartBuilder<PartBuilder<PartBuilder<...<Builder>...>>>` but 🤷‍♂️ you don't have to look at the type.

ChampionAsh5357 (approved):

PR #8260, opened 2021-12-08
[1.18] Add ItemUseEvent

Shadows-of-Fire:
This is a continuation of #6615 and adds the ItemUseEvent to enable custom item usage behaviors.  This behavior is IMPOSSIBLE using RightClickBlock as-is due to some tiny specifics in how item usage is processed.

For example:
Apotheosis has a feature where a special "affix" can be applied to any pickaxe, that will enable the pickaxe to place torches at the cost of durability on the pickaxe. This feature currently uses this event as implemented in Placebo.

With this event, the interaction happens seamlessly - it occurs as if you literally were to right click a torch and everything functions as expected. 

Attempting to use RightClickBlock to achieve the same effect will not yield the correct behavior - you will have to forcibly take precedence of either the item or the block (f the block has a use action).  Even more confusing, you cannot even _determine_
if the block has a use action without running the use code, so you cannot actually get this order of precedence right without potentially performing irreversible changes.


alcatrazEscapee (approved):
I can second that this event does fill a specific niche: trying to recreate the ordering of standard actions (i.e., when you want your "thing that happens with the same priority of an item use", to work correctly and interact with the vanilla sequence, incorporating shifting/sneaking, clicking on blocks like chests or furnaces with block "useOn" interactions, using items like books or potions with item use interactions, it can be incredibly difficult to get it right using just `RightClickBlock`. I would absolutely use this event in it's current form.

marchermans (changes requested):
Missing test mod to check if the event still works as intended in the future.

src/main/java/net/minecraftforge/event/entity/player/PlayerInteractEvent.java
```diff
@@ -322,6 +322,28 @@ public LeftClickEmpty(Player player)
...
 
+    /**
+     * This event is fired when an item would be used via {@link Item#useOn}<br>
+     * It allows for the usage to be changed or cancelled.<br>
+     * This event allows finer-tuned control over the actual item usage that cannot be achieved by using {@link RightClickBlock}
+     */
+    @Cancelable
+    public static class ItemUseEvent extends PlayerInteractEvent
+    {
+        protected final UseOnContext ctx;
```
sciwhiz12: Why is this field `protected` rather than `private`?

sciwhiz12 (commented):
Looks fine to me except for a nitpick, though I echo Orion's recommendation that a test mod be made for regression testing and triage testing purposes.

src/main/java/net/minecraftforge/event/entity/player/PlayerInteractEvent.java
Shadows-of-Fire: I just tend to use protected over private unless it explicitly needs to be private.  I can swap it, since forge uses private over protected by default

Shadows-of-Fire:
Event fields swapped to private and a test mod has been added.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -749,6 +748,14 @@ public static void onEmptyLeftClick(Player player)
     {
         MinecraftForge.EVENT_BUS.post(new PlayerInteractEvent.LeftClickEmpty(player));
     }
+    
+    public static InteractionResult onItemUse(ItemStack stack, UseOnContext ctx)
+    {
+        PlayerInteractEvent.ItemUseEvent event = new PlayerInteractEvent.ItemUseEvent(ctx);
+        MinecraftForge.EVENT_BUS.post(event);
+        if (event.isCanceled()) return event.getCancellationResult();
```
sciwhiz12, line 757: Because this effect is affected by calls to `PlayerInteractEvent#setCancellationResult`, the javadoc for that method should be amended to include this new event in its list of affected events.

patches/minecraft/net/minecraft/world/item/ItemStack.java.patch
```diff
@@ -69,10 +69,12 @@
     }
  
     public boolean m_150922_(Tag<Item> p_150923_) {
-@@ -196,6 +_,15 @@
+@@ -196,6 +_,17 @@
     }
  
     public InteractionResult m_41661_(UseOnContext p_41662_) {
++      InteractionResult itemUseEventRes = net.minecraftforge.common.ForgeHooks.onItemUse(this, p_41662_);
++      if(itemUseEventRes != null) return itemUseEventRes;
```
sciwhiz12: Formatting: space after the `if` and before the opening parenthesis.
```suggestion
+      if (itemUseEventRes != null) return itemUseEventRes;
```

src/test/java/net/minecraftforge/debug/item/ItemUseTest.java
```diff
@@ -0,0 +1,56 @@
...
+    @SubscribeEvent
+    public static void itemUse(ItemUseEvent event)
+    {
+        UseOnContext ctx = event.getContext();
+        if (ctx.getItemInHand().getItem() == Items.WOODEN_PICKAXE)
+        {
+            if (Items.TORCH.useOn(ctx).consumesAction())
+            {
+                ctx.getItemInHand().grow(1); // Grow call is because the torch will shrink the stack. The "clean" way is to make a new context.
+                ctx.getPlayer().getItemInHand(ctx.getHand()).hurtAndBreak((int) 1, ctx.getPlayer(),
```
sciwhiz12: The cast seems to be unneeded:
```suggestion
                ctx.getPlayer().getItemInHand(ctx.getHand()).hurtAndBreak(1, ctx.getPlayer(),
```

src/test/java/net/minecraftforge/debug/item/ItemUseTest.java
```diff
@@ -0,0 +1,56 @@
...
+{
+
+    @SubscribeEvent
+    public static void itemUse(ItemUseEvent event)
+    {
+        UseOnContext ctx = event.getContext();
+        if (ctx.getItemInHand().getItem() == Items.WOODEN_PICKAXE)
+        {
+            if (Items.TORCH.useOn(ctx).consumesAction())
+            {
```
sciwhiz12: These two conditions could be simplified into one `if`-condition, as the logical AND operator (`&&`) is short-circuiting (if the left-hand condition is `false`, the right-hand condition is not evaluated).

sciwhiz12 (changes requested):
A bit of formatting and logic simplificaation, and this event looks :shipit: to me.

src/test/java/net/minecraftforge/debug/item/ItemUseTest.java
```diff
@@ -0,0 +1,56 @@
...
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
+
+@Mod("item_use_event")
+@Mod.EventBusSubscriber
+/**
+ * The ItemUseTest will allow Wooden Pickaxes to place torches as if they were a torch item, taking durability damage in the process.
+ * The use of the pickaxe should be completely seamless - as if you were physically holding a torch item, including all block-interactions.
+ * The pickaxe will not take durability damage in creative.
+ */
```
sciwhiz12, line 33: Additionally, this javadoc comment should be moved above the annotations, as otherwise the compiler doesn't know to attach this to the class and considers it a dangling comment.

Shadows-of-Fire:
Not sure what CI is angry about, builds locally fine

sciwhiz12 (approved):
Looks :shipit: to me, as long as the merge conflict is resolved.

LexManos:
We already have enough events fired when you right click a block. We are not going to add an event between every line of code.
Your functionality can be achieved with minimal effort using the events we already have.

Shadows-of-Fire:
I challenge you to accomplish this flawlessly with the existing events.

PR #8261, opened 2021-12-08
[1.18] Add ShieldBlockEvent

Shadows-of-Fire:
This PR adds the `ShieldBlockEvent`, which allows reactions to the actual blocking of damage with a shield, as well as changing how much of the damage is blocked, and if the block occurs at all.

This is useful for a couple of scenarios - making the shield block less damage in certain cases (instead of all or nothing), making something happen when a block is successful, doing additional durability damage to a shield, preventing durability damage to a shield, etc.

The current use cases I employ are allowing the user to "catch" arrows that are blocked with a shield, as well as doing damage to a melee attacker on a successful block.


patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
```diff
@@ -123,6 +123,30 @@
...
++            p_21017_ -= ev.getBlocked();
+             if (!p_21016_.m_19360_()) {
+                Entity entity = p_21016_.m_7640_();
+                if (entity instanceof LivingEntity) {
+@@ -1009,7 +_,7 @@
+             }
+ 
+             flag = true;
+-         }
++         }}
```
pupnewfster: The extra bracket should just be on its own line so the patch doesn't also need to contain removing the existing bracket.

Shadows-of-Fire: updated

Thelnfamous1:
This is a great event. Do you think something like a ShieldDisableEvent could fit in this PR, to have better control over what happens when a weapon tries to disable a shield? Currently, there doesn't seem to be an easy and clean way to say, make a shield immune to being disabled, or modify the time the shield is set on cooldown.

Shadows-of-Fire:
Based on past experiences I would not want to include two events within this PR.

marchermans (changes requested):
A test mod should be included for new events, to check if the event works after porting.

src/main/java/net/minecraftforge/event/entity/living/ShieldBlockEvent.java
```diff
@@ -0,0 +1,82 @@
+package net.minecraftforge.event.entity.living;
```
sciwhiz12, line 20: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/event/entity/living/ShieldBlockEvent.java
```diff
@@ -0,0 +1,82 @@
...
+ * Note: The shield item stack "should" be available from {@link LivingEntity#getUseItem()}
+ * at least for players.
+ */
+@Cancelable
+public class ShieldBlockEvent extends LivingEvent
+{
+    protected final DamageSource source;
+    protected final float originalBlocked;
+    protected float dmgBlocked;
+    protected boolean shieldTakesDamage = true;
```
sciwhiz12: Why `protected` rather than `private`?

src/main/java/net/minecraftforge/event/entity/living/ShieldBlockEvent.java
```diff
@@ -0,0 +1,82 @@
...
+        return this.originalBlocked;
+    }
+
+    /**
+     * @return The current amount of damage blocked, as a result of this event.
+     */
+    public float getBlocked()
+    {
+        return this.dmgBlocked;
+    }
```
sciwhiz12: Prehaps these could be renamed to `get(Original)BlockedDamage`?

src/main/java/net/minecraftforge/event/entity/living/ShieldBlockEvent.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    public void setBlocked(float blocked)
+    {
+        this.dmgBlocked = Mth.clamp(blocked, 0, this.originalBlocked);
+    }
+
+    /**
+     * Set if the shield will take durability damage or not.
+     */
+    public void setTakesDamage(boolean damage)
```
sciwhiz12: The setter should include the `shield` bit of the field name.

src/main/java/net/minecraftforge/event/entity/living/ShieldBlockEvent.java
```diff
@@ -0,0 +1,82 @@
...
+        super(blocker);
+        this.source = source;
+        this.originalBlocked = blocked;
+        this.dmgBlocked = blocked;
+    }
+
+    /**
+     * @return The damage source.
+     */
+    public DamageSource getSource()
```
sciwhiz12: Perhaps this could be fully named as `getDamageSource`? More a suggestion, up to you if you decide to take it.

sciwhiz12 (changes requested):
I echo Orion's recommendation that a test mod be made for regression testing and triage testing purposes.

Shadows-of-Fire:
Method names have been updated, license headers have been added, and a test mod is now included as well.

src/test/java/net/minecraftforge/debug/item/ShieldBlockTest.java
```diff
@@ -0,0 +1,55 @@
...
+		if (event.getDamageSource().getDirectEntity() instanceof AbstractArrow arrow)
+		{
+			if (event.getEntityLiving() instanceof Player player)
+			{
+				player.getInventory().add(new ItemStack(Items.ARROW));
+				event.setBlockedDamage(event.getOriginalBlockedDamage() / 2);
+				arrow.discard();
+			}
+		}
+	}
```
sciwhiz12: This class is using tabs, while the Forge code style mandates using spaces.

src/test/java/net/minecraftforge/debug/item/ShieldBlockTest.java
```diff
@@ -0,0 +1,55 @@
...
+{
+	static final String MOD_ID = "shield_block_event";
+
+	@SubscribeEvent
+	public static void shieldBlock(ShieldBlockEvent event)
+	{
+		if (event.getDamageSource().getDirectEntity() instanceof AbstractArrow arrow)
+		{
+			if (event.getEntityLiving() instanceof Player player)
+			{
```
sciwhiz12: You can simplify the logic by combining these two into one `if`-condition.

sciwhiz12 (changes requested):
The event looks fine to me, just two changes for the test mod and this looks :shipit: to me.

src/test/java/net/minecraftforge/debug/item/ShieldBlockTest.java
Shadows-of-Fire: Bleh, could've sworn I ran a tab->space conversion on it.  Gusss not

src/test/java/net/minecraftforge/debug/item/ShieldBlockTest.java
Shadows-of-Fire: That makes for a really really long single line, which imo isn't great for readability

sciwhiz12: You could have the `if`-expression split across two lines then, which would still mean we get the benefit of the decrease of indentation level. I'll leave it up to you.

sciwhiz12 (approved):
Looks :shipit: to me, as long as the merge conflict is resolved.

PR #8263, opened 2021-12-09
[1.18] Provide way to add new component types to the worldgen datapack registries

Won-Ton:
This provides a way for mods to add new world-gen component types that are loaded alonside vanilla's (biomes, features etc).

The PR introduces a `RegistryAccessExtension` object which wraps the component's data Codec along with a DeferredRegister. Mods can register default values of their component to this and forge will handle adding the extra registries to the RegistryAccess.RegistryHolders, injecting the registry contents to form the default/"builtin"/"seed" registries, and loading the datapack definitions/overrides, at the appropriate times.

### Why it's useful
[A test mod is provided](https://github.com/Won-Ton/MinecraftForge/blob/a5d17a927f35ee44a6abe7fdacde761ca5b9842f/src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java) to show how this can be used to add a new world-gen component:
- [Example instances](https://github.com/MinecraftForge/MinecraftForge/blob/7ace0044bdb1d64e3e4a3191847f811286956bbf/src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java#L55-L57) are registered in code, which becomes a default/builtin entry much like the vanilla components that get registered to the `BuiltinRegistries`. These will always be present in the resulting RegistryAccess but may be overriden by a datapack.
- [Various datapack definitions are also provided](https://github.com/Won-Ton/MinecraftForge/blob/a5d17a927f35ee44a6abe7fdacde761ca5b9842f/src/test/resources/data/registry_access_extension_test/worldgen/example/data_example.json](https://github.com/MinecraftForge/MinecraftForge/tree/7ace0044bdb1d64e3e4a3191847f811286956bbf/src/test/resources/data/registry_access_extension_test/registry_access_extension_test/worldgen)) to show that dynamic entries can also be added.

### Fin
Hopefully it's ok. There are definitely other ways to approach this but this I think is cleanest in terms of patches. Or maybe the PR will at-least seed the idea for someone else to take a crack at it. Thanks :]


SmellyModder:
This PR looks promising, but I have one concern/suggestion.
The goal of this PR is to provide a way to add new component types to worldgen datapacks, but fails to address the inability to dynamically load various user-created files for custom worldgen component types.
For example, a mod may want to decode an unfixed amount of data files for a custom component type to get used in a custom system.
Vanilla does this using `RegistryResourceAccess#forResourceManager()` to dynamically load various components for worldgen registries when loading datapacks for a world.
In other words, this would allow mods to, in a way, add their own worldgen dynamic registries to the game.

Won-Ton:
Not 100% sure I follow,  so you might need to reword your point (if you wouldn't mind) if this doesn't answer what you're saying:

This system allows json-only definitions of the components. I've provided an example of this in the `data_example.json` file. There's no corresponding 'builtin' registry entry for that (the example registered in code is called "builtin_example"). The resulting registry after datapack loading contains both the code-registered `builtin_example` and data-loaded `data_example` entries. Additionally one could override the  `builtin_example` with a json definition (no example provided in the test mod but it works).

SmellyModder:
> Not 100% sure I follow, so you might need to reword your point (if you wouldn't mind) if this doesn't answer what you're saying:
> 
> This system allows json-only definitions of the components. I've provided an example of this in the `data_example.json` file. There's no corresponding 'builtin' registry entry for that (the example registered in code is called "builtin_example"). The resulting registry after datapack loading contains both the code-registered `builtin_example` and data-loaded `data_example` entries. Additionally one could override the `builtin_example` with a json definition (no example provided in the test mod but it works).

Ok, good, it sounds like this PR does achieve what I was wondering about, thanks 👍 
This system will be very useful, and I'm excited to use it 😛

Commoble:
I would use this feature! The main advantages of custom dynamic registries over traditional ReloadListener data loaders are that they can directly refer to other worldgen json data and validate their existence on load, and that they are slightly easier to set up due to the vanilla infrastructure in place for them (The main disadvantage is that they do not support merging of jsons with the same ID from multiple datapacks ala tags).

Another way this could be implemented would involve patching into RegistryAccess.REGISTRIES's builder and adding the custom RegistryData there + adding the registry to BuiltinRegistries at some earlier time, but I have no substantial grievances with wonton's implementation.

src/main/java/net/minecraftforge/common/ForgeRegistryAccessHooks.java
```diff
@@ -0,0 +1,222 @@
...
+import org.apache.logging.log4j.Marker;
+import org.apache.logging.log4j.MarkerManager;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Contains hooks for RegistryAccess creation, injection and loading.
+ */
+@SuppressWarnings("JavadocReference")
```
ChampionAsh5357: Why?

src/main/java/net/minecraftforge/common/ForgeRegistryAccessHooks.java
```diff
@@ -0,0 +1,222 @@
...
+
+    /*
+     * Helper for copying an Iterable of ResourceKey-to-value mappings to a RegistryAccess.
+     */
+    private static <T> void copy(ResourceKey<? extends Registry<T>> key, Iterable<Map.Entry<ResourceKey<T>, T>> source, RegistryAccess dest)
+    {
+        var registry = dest.ownedRegistryOrThrow(key);
+        for (var entry : source)
+        {
+            registry.register(entry.getKey(), entry.getValue(), Lifecycle.stable());
```
ChampionAsh5357: The lifecycle should be obtained from the elements itself. It should not just be declared as stable.

src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java
```diff
@@ -0,0 +1,114 @@
...
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.ForgeRegistryEntry;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.util.function.Supplier;
+
+@Mod(RegistryAccessExtensionTest.MODID)
+public class RegistryAccessExtensionTest {
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java
```diff
@@ -0,0 +1,114 @@
...
+    public static final String MODID = "registry_access_extension_test";
+    public static final Logger LOGGER = LogManager.getLogger(MODID);
+    public static final ResourceLocation REGISTRY_NAME = new ResourceLocation(MODID, "worldgen/example");
+    public static final RegistryAccessExtension<ExampleComponent> EXAMPLE_COMPONENT = new RegistryAccessExtension<>(REGISTRY_NAME, ExampleComponent.class, ExampleComponent.CODEC);
+
+    public RegistryAccessExtensionTest() {
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java
```diff
@@ -0,0 +1,114 @@
...
+        var modBus = FMLJavaModLoadingContext.get().getModEventBus();
+        modBus.addGenericListener(RegistryAccessExtension.class, this::registerExtension);
+
+        EXAMPLE_COMPONENT.getRegister().register(modBus);
+        EXAMPLE_COMPONENT.getRegister().register("builtin_example", builtinSupplier());
+
+        MinecraftForge.EVENT_BUS.addListener(this::onServerStart);
+    }
+
+    private void onServerStart(ServerAboutToStartEvent event) {
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java
```diff
@@ -0,0 +1,114 @@
...
+        var registries = event.getServer().registryAccess();
+
+        var biomes = registries.registryOrThrow(Registry.BIOME_REGISTRY);
+        var features = registries.registryOrThrow(Registry.PLACED_FEATURE_REGISTRY);
+        var biomeFeatures = registries.registryOrThrow(EXAMPLE_COMPONENT.getRegistryKey());
+
+        LOGGER.info("RegistryAccessExtensionTest - ExampleComponent registry:");
+        for (var entry : biomeFeatures.entrySet()) {
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java
```diff
@@ -0,0 +1,114 @@
...
+            // Check that pair holds valid Biome and PlacedFeature references by attempting to
+            // retrieve their key from the RegistryAccess instance. Will return null if invalid.
+            var biomeName = biomes.getKey(pair.getBiome().get());
+            var featureName = features.getKey(pair.getFeature().get());
+
+            LOGGER.info(" - Name: {}, Biome: {}, Feature: {}", name, biomeName, featureName);
+        }
+    }
+
+    private void registerExtension(RegistryEvent.Register<RegistryAccessExtension<?>> event) {
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java
```diff
@@ -0,0 +1,114 @@
...
+        event.getRegistry().register(EXAMPLE_COMPONENT);
+    }
+
+    private static Supplier<ExampleComponent> builtinSupplier() {
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java
```diff
@@ -0,0 +1,114 @@
...
+        return () -> new ExampleComponent(
+                () -> ForgeRegistries.BIOMES.getValue(Biomes.TAIGA.location()),
+                () -> VegetationPlacements.BAMBOO_VEGETATION
+        );
+    }
+
+    public static class ExampleComponent extends ForgeRegistryEntry<ExampleComponent> {
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java
```diff
@@ -0,0 +1,114 @@
...
+        public static final Codec<ExampleComponent> CODEC = RecordCodecBuilder.create(instance -> instance.group(
+                Biome.CODEC.fieldOf("biome").forGetter(p -> p.biome),
+                PlacedFeature.CODEC.fieldOf("feature").forGetter(p -> p.feature)
+        ).apply(instance, ExampleComponent::new));
+
+        private final Supplier<Biome> biome;
+        private final Supplier<PlacedFeature> feature;
+
+        public ExampleComponent(Supplier<Biome> biome, Supplier<PlacedFeature> feature) {
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java
```diff
@@ -0,0 +1,114 @@
...
+            this.biome = biome;
+            this.feature = feature;
+        }
+
+        public Supplier<Biome> getBiome() {
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java
```diff
@@ -0,0 +1,114 @@
...
+            return biome;
+        }
+
+        public Supplier<PlacedFeature> getFeature() {
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java
```diff
@@ -0,0 +1,114 @@
...
+        return () -> new ExampleComponent(
+                () -> ForgeRegistries.BIOMES.getValue(Biomes.TAIGA.location()),
+                () -> VegetationPlacements.BAMBOO_VEGETATION
+        );
+    }
+
+    public static class ExampleComponent extends ForgeRegistryEntry<ExampleComponent> {
+        public static final Codec<ExampleComponent> CODEC = RecordCodecBuilder.create(instance -> instance.group(
+                Biome.CODEC.fieldOf("biome").forGetter(p -> p.biome),
+                PlacedFeature.CODEC.fieldOf("feature").forGetter(p -> p.feature)
```
ChampionAsh5357: When referring to the flat map, I mean here.

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
```diff
@@ -0,0 +1,107 @@
...
+
+    public ResourceKey<? extends Registry<T>> getRegistryKey()
+    {
+        return registryKey;
+    }
+
+    public DeferredRegister<T> getRegister()
+    {
+        return register;
+    }
```
ChampionAsh5357: The point of datapack types is to not have any internal object registries. We should not be giving users the object to add json objects via code.

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * Creates a new RegistryAccessExtension from the given IForgeRegistry and element Codec.
+     * The registry name of the RegistryAccessExtension is derived from the name of the provided IForgeRegistry.
+     *
+     * @param registry The registry that elements will be registered to.
+     * @param codec    The codec for serializing/deserializing the entry to/from data formats (namely json).
+     */
+    public RegistryAccessExtension(IForgeRegistry<T> registry, Codec<T> codec)
```
ChampionAsh5357: We probably shouldn't be creating extensions from existing types. That could lead to a whole slew of desync issues.

ChampionAsh5357 (changes requested):
There are a few more test cases I would like to see in this mod. First would be a data generated object holding a reference to a data generated object. Another would be constructor resolution for example with `BiomeGenerationSettings`, though the flat map with `ExtraCodecs#nonNullSupplierCheck` should suffice that case. Another one would be a list of elements containing both builtin and data examples.

Another thing of note is how to deal with references that are removed between world saves and then added back in at a later time. I'm pretty sure that there is code within Forge for this, but it should also be another test case to consider.

Along that line of thinking, what about data that is synchronized to the client? How would that be handled? Is there a separate network codec? What is the result of the data being successfully synced?

We want to stress test this PR as much as possible. I see huge benefits for this, but also a lot of areas for disaster if not implemented correctly. This will also probably be even more relevant with the second half of #7917 since it could be a basis for a modification API to json components.

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
Commoble: concur'd

Commoble:
> Along that line of thinking, what about data that is synchronized to the client? How would that be handled? Is there a separate network codec? What is the result of the data being successfully synced?

While vanilla does support syncing of dynamic registries, as far as I can tell this is hardcoded to disabled for the custom registries in this PR

src/main/java/net/minecraftforge/common/ForgeRegistryAccessHooks.java
Won-Ton: Jdoc links to private methods used. Can remove if we don't care about the warnings

src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java
Won-Ton: I was under the impression that the codestyle did not apply to test mods?

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
Won-Ton: Builtin registry entries guarantee some baseline set of components that worldgen can be constructed with when there are no datapacks providing that content present (vanilla will crash if there are no biomes, for example)

src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java
gigaherz: We still want some consistency. The braces are on their own lines on every other piece of forge code...

Won-Ton: no problem 👍

Won-Ton:
Thanks!

> There are a few more test cases I would like to see in this mod. First would be a data generated object holding a reference to a data generated object. Another would be constructor resolution for example with `BiomeGenerationSettings`, though the flat map with `ExtraCodecs#nonNullSupplierCheck` should suffice that case. Another one would be a list of elements containing both builtin and data examples.

Sure, I'll add more of these cases.

> Another thing of note is how to deal with references that are removed between world saves and then added back in at a later time. I'm pretty sure that there is code within Forge for this, but it should also be another test case to consider.

Generally speaking world-gen configurations are transient, existing only for the current session. Changes to components between sessions will produce different configurations that could change how new chunks generate in comparison to existing ones but shouldn't really break things otherwise.

I suppose if a mod chooses to persist a reference to a specific entry in the world somewhere they'd need to verify it exists when trying to retrieve that referenece. There was that issue of vanilla not checking structure references in the chunk savedata and NPE'ing when mod-added structures were removed (that wasn't specifically about datapack components but it's the closest example I can think of). In that case I believe forge patched vanilla to add the check. In this case it would be on the mod to handle it correctly.

> Along that line of thinking, what about data that is synchronized to the client? How would that be handled? Is there a separate network codec? What is the result of the data being successfully synced?

and:
> While vanilla does support syncing of dynamic registries, as far as I can tell this is hardcoded to disabled for the custom registries in this PR

I couldn't come up with a usecase that wasn't already (better?) served by forge's other APIs for clientside stuff. My background is world-gen which is all server-side so maybe the blinkers are on a bit. All ears.

edit: but yes, to support client syncing I think it would need:
- An additional codec field where mods can provide the network codec for the component
- Forge to replace the `RegistryAccess.RegistryHolder.NETWORK_CODEC` (or patch uses of it) to use an extended version.
  - Possibly also to enforce the client has the extra codec(s)

src/main/java/net/minecraftforge/common/ForgeRegistryAccessHooks.java
Won-Ton: Forge registries don't hold lifecycles for entries. Thoughts on stable or experimental for those? Will do for the vanilla ones although they are all stable anyway since they're sourced directly from the BuiltinRegistries

Won-Ton:
@ ChampionAsh5357 I've added some extra examples to the test mod. I think I may be misunderstanding you on the bit quoted below; please could you clarify what you mean and go into a bit more detail about what you're looking to confirm about this PR?:
> Another would be constructor resolution for example with BiomeGenerationSettings, though the flat map with ExtraCodecs#nonNullSupplierCheck should suffice that case

ChampionAsh5357:
> Another would be constructor resolution for example with BiomeGenerationSettings, though the flat map with ExtraCodecs#nonNullSupplierCheck should suffice that case

Specifically that when I try to resolve the supplied entry in a constructor, it should give me the actual object. I should have no reason to keep it as a supplier as soon as the constructor passes it in. The resulted method I gave you basically tries to resolve the object to verify that it exists within the codec itself as part of a `flatXmap`.

src/main/java/net/minecraftforge/common/ForgeRegistryAccessHooks.java
ChampionAsh5357: Just use `@code` instead. I believe also there is a compiler setting for the javadoc to allow private methods to be referenced, but that would be part of the omnibus.

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
ChampionAsh5357: That's just bad design on Mojang. The idea of a json based registry is that it can have no elements if so desired. No biomes would just be a void world, same with no generators. There can be some defaulted case sure, but that would be something non-changing that would be abstracted away within the builder of the component registry, not as an exposed built in element.

ChampionAsh5357 (changes requested):
As for the lifecycle of the registry objects, it should be all experimental by default unless otherwise specified by the modder.

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
Won-Ton: Beyond providing a way to configure world-gen/dimensions etc I couldn't comment on Mojang's ideals behind the design of the datapack system. Absence of default values in the registries means that the behaviour where they would have been used has to be hardcoded instead which seems counter to the configurability aspect.

I guess forcing mods to ship with json definitions in their jar makes those more accessible to end-users which is nice. However, if they can't register in code though they can't data-gen them without having to write their own custom system for it

Won-Ton:
> Specifically that when I try to resolve the supplied entry in a constructor, it should give me the actual object. I should have no reason to keep it as a supplier as soon as the constructor passes it in. The resulted method I gave you basically tries to resolve the object to verify that it exists within the codec itself as part of a `flatXmap`.

Ah I see. That would be a limitation of the datapack system itself rather than anything that this PR touches. I believe the standard of using suppliers across vanilla is for datapack reloads

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
ChampionAsh5357: Apologies, I'm not meaning to comment on Mojang's ideals, I'm just stating my own opinion on how I imagine it to be. 

As for registering in code for data generation, that is completely reasonable. However, that would be directly coupled to the data provider. We shouldn't allow the user to just register outside of that provider based context. But again, these are my opinions.

ChampionAsh5357:
> Ah I see. That would be a limitation of the datapack system itself rather than anything that this PR touches. I believe the standard of using suppliers across vanilla is for datapack reloads

Yeah, I just want to make sure we can replicate all features and bugs within the vanilla datapack system.

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
Won-Ton: Seems a bit arbitrary to me to be honest but appreciate you'll have more insight to the kinds of things that need protecting against with such restrictions

src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java
```diff
@@ -0,0 +1,188 @@
...
+@Mod(RegistryAccessExtensionTest.MODID)
+public class RegistryAccessExtensionTest
+{
+    public static final String MODID = "registry_access_extension_test";
+    public static final Logger LOGGER = LogManager.getLogger(MODID);
+
+    public static final ResourceLocation TEST_A_NAME = new ResourceLocation(MODID, "worldgen/test_a");
+    public static final ResourceLocation TEST_B_NAME = new ResourceLocation(MODID, "worldgen/test_b");
+    public static final RegistryAccessExtension<TestA> TEST_A = new RegistryAccessExtension<>(TEST_A_NAME, TestA.class, TestA.DIRECT_CODEC);
+    public static final RegistryAccessExtension<TestB> TEST_B = new RegistryAccessExtension<>(TEST_B_NAME, TestB.class, TestB.DIRECT_CODEC);
```
ChampionAsh5357: Another thing I realized is that these are statically initialized due to how you set up your system. However, these should be deferred for creating until their associated registry event. This would mean that the registry and register for the dynamic objects would need to be not part of the actual registered object which is how it should be. However, it's still relatively trivial to get the necessary registry without that since you already have the registry key of the registry and can pull it from `RegistryManager#getRegistry` in the `ACTIVE` registry set.

ChampionAsh5357 (changes requested):
Just realized something I completely forgot. This system will not work due to registry conflictions. Specifically, the namespace of the registry is not taken into consideration when creating the folders. As such, all created registries must be in the `minecraft` namespace. This means if two mods `a` and `b` create a dynamic registry `x`, then the path for both will be `data/<object_modid>/x/<object_path>.json`. It needs to factor in the namespace of the mod as well either in the path as a subdirectory or a prefix and validated.

Won-Ton:
> Just realized something I completely forgot. This system will not work due to registry conflictions. Specifically, the namespace of the registry is not taken into consideration when creating the folders. As such, all created registries must be in the `minecraft` namespace. This means if two mods `a` and `b` create a dynamic registry `x`, then the path for both will be `data/<object_modid>/x/<object_path>.json`. It needs to factor in the namespace of the mod as well either in the path as a subdirectory or a prefix and validated.

Yep thats true of the vanilla datapack registries too - the 'path' part needs to be unique to the registry. If forge wants to enforce that uniqueness on extension registries then prefixing the path with the modid would be the simplest way

src/test/java/net/minecraftforge/debug/world/RegistryAccessExtensionTest.java
Won-Ton: This was intentional so that the extension object can take responsibility over setting up the resourcekey/register properly. Happy to remove it but it makes it harder to enforce things like the path uniqueness that we discussed or ensure the backing registry is built with the right configuration - I could add some static helper methods for creating keys/registers that handle these requirements for the modder

hYdos:
Based on what I've seen so far, all custom data would end up in 2 folders with the same name. For example a registry created by doing your
`ResourceKey<Registry<TestA>> REGISTRY = RegistryAccessExtension.createRegistryKey(Mod.MOD_ID, "foo");`
 would be in the folder `/data/modid/modid/foo` instead of `/data/modid/foo`. Is there a reason for that? it just seems to make the data folder a bit messier

Won-Ton:
> Based on what I've seen so far, all custom data would end up in 2 folders with the same name. For example a registry created by doing your `ResourceKey<Registry<TestA>> REGISTRY = RegistryAccessExtension.createRegistryKey(Mod.MOD_ID, "foo");` would be in the folder `/data/modid/modid/foo` instead of `/data/modid/foo`. Is there a reason for that? it just seems to make the data folder a bit messier

See the discussion immediately above; it's to enforce registry paths are unique (also explained in the jdocs). The first "modid" in `/data/modid/modid/foo` is the namespace of the registry entry. The second serves as the namespace of the registry itself. It prevents two different mods that define a registry path of `foo/` from conflicting

hYdos:
But how would the end up conflicting if they have different modids? The folders would never end up being the same

Commoble:
> But how would the end up conflicting if they have different modids? The folders would never end up being the same

The folders would never be the same IF the data loaders themselves are namespaced. If the loaders aren't namespaced, then two mods could accidentally register loaders with the same name.

Suppose loaders aren't namespaced. Two mods (`fancy_cheeses` and `fine_cheeses`) both register a "/cheeses" data loader folder (with different data formats). A third mod `extra_cheeses` wants to have their `extra_cheeses:cheddar` json loaded by fancy_cheeses, so they put it in `data/extra_cheeses/cheeses/cheddar.json`. Fine Cheeses's loader also loads the same json from the cheeses folder and logspams because the json doesn't use Fine Cheese's json format.

Now suppose loaders are namespaced. Fancy Cheeses loads from `/fancy_cheeses/cheeses` and Fine Cheeses loads from `/fine_cheeses/cheeses`. Extra Cheeses wants Fancy Cheeses to load their json, so they add a json at `data/extra_cheeses/fancy_cheeses/cheeses/cheddar.json` (the first folder under the data folder must be the object namespace folder due to how vanilla namespaces objects as it loads them). Fancy Cheeses loads `extra_cheeses:cheddar` from the `fancy_cheeses/cheeses` data folder. Fine Cheeses looks in `fine_cheeses/cheeses` and doesn't find any cheese to whine about.

(The existing system we already have for adding reloadable data loaders doesn't enforce this extra namespacing and *does* have this loader collision problem.)

hYdos:
> > But how would the end up conflicting if they have different modids? The folders would never end up being the same
> 
> The folders would never be the same IF the data loaders themselves are namespaced. If the loaders aren't namespaced, then two mods could accidentally register loaders with the same name.
> 
> Suppose loaders aren't namespaced. Two mods (`fancy_cheeses` and `fine_cheeses`) both register a "/cheeses" data loader folder (with different data formats). A third mod `extra_cheeses` wants to have their `extra_cheeses:cheddar` json loaded by fancy_cheeses, so they put it in `data/extra_cheeses/cheeses/cheddar.json`. Fine Cheeses's loader also loads the same json from the cheeses folder and logspams because the json doesn't use Fine Cheese's json format.
> 
> Now suppose loaders are namespaced. Fancy Cheeses loads from `/fancy_cheeses/cheeses` and Fine Cheeses loads from `/fine_cheeses/cheeses`. Extra Cheeses wants Fancy Cheeses to load their json, so they add a json at `data/extra_cheeses/fancy_cheeses/cheeses/cheddar.json` (the first folder under the data folder must be the object namespace folder due to how vanilla namespaces objects as it loads them). Fancy Cheeses loads `extra_cheeses:cheddar` from the `fancy_cheeses/cheeses` data folder. Fine Cheeses looks in `fine_cheeses/cheeses` and doesn't find any cheese to whine about.
> 
> (The existing system we already have for adding reloadable data loaders doesn't enforce this extra namespacing and _does_ have this loader collision problem.)

With this example, you are saying that the extra cheeses mod is adding its data in `data/extra_cheeses/cheeses/cheddar.json`. Shouldnt it go in `data/fancy_cheeses/cheeses/cheddar.json` if you want `fancy_cheeses` to load the data? You wouldnt have any collision issues if you do it like that right?

ChampionAsh5357:
I believe the simplest solution is to use a data loader folder for storing the registry information. For example, if we have some registry `a:b` with some entry `c:d`, it should be located in `data/c/modded_registries/a/b/d.json`. Regardless of what we end up going with, there will always be a potential naming conflict due to how data loaders work. The best we can do is reserve some subdirectory namespace. This should solve most conflict concerns.

hYdos:
> I believe the simplest solution is to use a data loader folder for storing the registry information. For example, if we have some registry `a:b` with some entry `c:d`, it should be located in `data/c/modded_registries/a/b/d.json`. Regardless of what we end up going with, there will always be a potential naming conflict due to how data loaders work. The best we can do is reserve some subdirectory namespace. This should solve most conflict concerns.

The only issue I think modders would have with this is what if the names are very long? It sortof makes your data folder a mess to navigate.

ChampionAsh5357:
> The only issue I think modders would have with this is what if the names are very long? It sortof makes your data folder a mess to navigate.

There's no perfect system when it comes to this. I favor readability with minimal conflicts compared to length. The only concern I would have there is path lengths though its still unlikely to hit on most OS systems.

Commoble:
> With this example, you are saying that the extra cheeses mod is adding its data in `data/extra_cheeses/cheeses/cheddar.json`. Shouldnt it go in `data/fancy_cheeses/cheeses/cheddar.json` if you want `fancy_cheeses` to load the data? You wouldnt have any collision issues if you do it like that right?

No, it's Extra Cheese's json and they want the loaded object to have the `extra_cheeses` namespace. If they put it in the `fancy_cheeses` namespace, that would imply that Fancy Cheeses's own internal datapack included the json. The object namespace folder is the first folder after data, so it must go in `data/extra_cheeses`. This is how all assets and data works.

Won-Ton:
Ello! Re the latest commit:

I've changed the approach to patching ResourceAccess to something that's probably a bit easier to reason about and results in fewer and smaller hook methods for forge to maintain:

- It makes the `ResourceAccess.REGISTRIES` map (holds the codecs for each component type) mutable so that forge can add the extension codecs at an appropriate time.
- Adds a method for `ForgeMod` to call which forwards the mutable maps of `ResourceAccess.REGISTRIES`  and the backing map (already mutable) of `ResourceAccess.BUILTIN` to a hook which handles adding the codecs/builtin-registries for the extensions.
- Tweaks a preexisting patch that delegates registry lookups to the BuiltinRegistries (just class loading things) so that it still does that, but only for the builtin registries. It isn't necessary for extension registries since we inject those to `ResourceAccess.BUILTIN` after mod registrations. (note: this patch previously just filtered out the DimensionType resource key which isn't a built-in registry so the change maintains behaviour).

Functionally it behaves the same as the previous iteration. It also makes it easier to introduce a newtwork codec for extensions in future if that were ever needed. Thoughts welcome. Happy to revert if the other way was preferable.

patches/minecraft/net/minecraft/core/RegistryAccess.java.patch
```diff
@@ -1,11 +1,36 @@
 --- a/net/minecraft/core/RegistryAccess.java
 +++ b/net/minecraft/core/RegistryAccess.java
+@@ -37,7 +_,7 @@
+ 
+ public abstract class RegistryAccess {
+    static final Logger f_123047_ = LogManager.getLogger();
+-   static final Map<ResourceKey<? extends Registry<?>>, RegistryAccess.RegistryData<?>> f_123048_ = Util.m_137537_(() -> {
++   static final Map<ResourceKey<? extends Registry<?>>, RegistryAccess.RegistryData<?>> f_123048_ = new java.util.HashMap<>(Util.m_137537_(() -> {
```
ChampionAsh5357, line 8: I would usually avoid making immutable maps mutable and settle for some joined forge variant, but since this is inaccessible normally, this is more of an opinion.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1241,4 +1245,38 @@ public static void onEntityEnterSection(Entity entity, long packedOldPos, long p
...
+            // Injects the mod-provided built-in registry elements to RegistryAccess.BUILTIN
+            builtin.put(key, registry);
+
+            LOGGER.info("Injected RegistryAccessExtension: {}", key);
+        }
+    }
+
+    private static <E extends IForgeRegistryEntry<E>> RegistryAccess.RegistryData<E> createExtensionData(RegistryAccessExtension<E> extension)
+    {
+        return new RegistryAccess.RegistryData<>(extension.getRegistryKey(), extension.getDirectCodec(), null);
```
ChampionAsh5357, line 1358: So are network codecs not supported by this as well and should be left as a feature to implement of the modders? If so, that should be specified somewhere.

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
```diff
@@ -0,0 +1,123 @@
...
+ * provided at {@link #createRegistryKey(String, String)} and {@link #createRegistryKey(ResourceLocation)}.
+ * <p>
+ * Default or 'built-in' registry entries can be registered via a forge DeferredRegister or IForgeRegistry
+ * of the same registry ResourceKey. These elements will populate the resulting RegistryAccess before datapacks
+ * are loaded and may be overridden by data-definitions. A static helper method for creating a DeferredRegisters
+ * for a RegistryAccessExtension is provided at {@link #createRegister(Class, ResourceKey)}.
+ *
+ * @param <T> The type of world-gen component.
+ */
+public class RegistryAccessExtension<T extends IForgeRegistryEntry<T>> extends ForgeRegistryEntry<RegistryAccessExtension<?>>
```
ChampionAsh5357: Shouldn't this be final? I don't believe there should be a need to expand behavior here.

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
```diff
@@ -0,0 +1,123 @@
...
+            path = prefix + path;
+        }
+
+        return ResourceKey.createRegistryKey(new ResourceLocation(namespace, path));
+    }
+
+    public static <T> ResourceKey<Registry<T>> createRegistryKey(ResourceLocation location)
+    {
+        return createRegistryKey(location.getNamespace(), location.getPath());
+    }
```
ChampionAsh5357, line 98: This could be better handled as a static constructor for this class or as inputs to the constructor itself. It's more strange that you need to call these methods within the constructor itself.

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
```diff
@@ -0,0 +1,123 @@
...
+        register.makeRegistry(key.location().getPath(), () -> new RegistryBuilder<T>().disableSaving().disableSync());
+        return register;
+    }
+
+    private static String getPathPrefix(String namespace)
+    {
+        return namespace + '/';
+    }
+
+    private static void validateRegistryKey(ResourceKey<? extends Registry<?>> registryKey)
```
ChampionAsh5357, line 113: This is just a comment making sure there are no complaints on the current standard before moving forward since there was a discussion about it above.

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
```diff
@@ -0,0 +1,123 @@
...
+
+        return ResourceKey.createRegistryKey(new ResourceLocation(namespace, path));
+    }
+
+    public static <T> ResourceKey<Registry<T>> createRegistryKey(ResourceLocation location)
+    {
+        return createRegistryKey(location.getNamespace(), location.getPath());
+    }
+
+    public static <T extends IForgeRegistryEntry<T>> DeferredRegister<T> createRegister(Class<T> type, ResourceKey<Registry<T>> key)
```
ChampionAsh5357, line 100: Why couldn't this be an instance method since the data already exists on the extension access? It also seems as though this is responsible for creating both the registry and the entries. I think it would be better to split the deferred register into a creation event for the static registry if wanted and then an instance method to create the deferred register for the specified mod id.

ChampionAsh5357 (changes requested):
Will test when I can. Here are just some general observation comments.

patches/minecraft/net/minecraft/core/RegistryAccess.java.patch
Won-Ton: Yeah I had similar feelings but the access swayed it for me, personally. This map is used when statically initializing the BUILTIN registry holder so I think it'd become complicated to have forge inject in the REGISTRIES lambda, for example, at an apropriate time - it'd be frail to classloading.

src/main/java/net/minecraftforge/common/ForgeHooks.java
Won-Ton: It'd be easy to add (an extra codec field in RegAccExtn and using that here instead of the null). 

The consideration to make is that the client needs the same set of codecs in order to connect to a server that sends extra components (see attached). There's no facility for ensuring that at the mo besides implying it by synchronizing the builtin registry for those components (still doesn't guarantee a codec exists on the client). I feel like if there's a strong argument for sync'ing it could be looked at in separate PR. The usefullness of this system imo lies entirely on the server side

_This is a vanilla client connecting to a server that is sending extra world-gen components over the network:_

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
Won-Ton: Yep good point. Would be possible to override the ResourceKey to get around the restrictions on the path prefix as-is

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
Won-Ton: You wanted this class to _not_ be statically initiallized (previous comments on the test class) however we do need static references to the registry ResourceKeys in constructing Codecs that can be referenced by other dynamic components.

I added these helper methods so that users can create static ResourceKeys with the appropriate namespacing, and then extensions can just be initialized with those keys during registration.

For clarity are you suggesting to go back to having statically init'd RegAccExtns? Or that mods should take responsibilty for creating the ResourceKeys with the appropriate path rather than having helper methods? (or other that I've missed?)

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
Won-Ton: I'm not sure if misunderstanding, but is it not a similar chicken-egg situation as the above point where you don't want RegAccExtns statically initialized, but they'd kindof have to be for component deferred registers to be created off instance methods of them?

Could you reword please if I've missed the point?

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
ChampionAsh5357: A static constructor is not the same as a static initialization. A static constructor is simple a method to create an object via a static method. It can still be used to register an object lazily. However, the point I was making is that the registry key should be made as part of the constructor instead of having as a method that would need to be passed in.

```java
static Test create(String name, String path) {
   // create the registry key
}

// or

Test(String name, String path) {
  this.registryKey = // registry value
}
```

instead of

```java
Test(RegistryKey)

// Somewhere to call
Test(methodToCreateKey(String name, String path))
```

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
ChampionAsh5357: Apologies, I think I forgot the order I commented against the last time. You can resolve this comment.

src/main/java/net/minecraftforge/common/world/RegistryAccessExtension.java
Won-Ton: If we’re using this PR to add a custom registry to the RegistryAccess then we’re holding reference to that registry’s ResourceKey in a static final field somewhere so that we can:
a) retrieve our registry from the RegistryHolder and actually use the component instances it contains.
b) create codecs (also typically static final fields) that can be used to reference our component from other components (the Supplier\<T\> codecs).

If we’re holding reference to the registry’s ResourceKey already we may as well be constructing the RegAccExtn with it.

Having helper methods that ‘correctly’ construct our registry ResourceKey that we can hold on to in a static field  is more useful to us than it being constructed inside a lazily computed object that we can’t

sciwhiz12:
As per previous communication from the author, this is effectively superseded by https://github.com/MinecraftForge/MinecraftForge/pull/8630.

PR #8264, opened 2021-12-09
[1.18] Add MobFinalizeSpawnEvent

seymourimadeit:
This PR adds the MobFinalizeSpawnEvent, which allows things to be added to an entity after it has initially spawned. (ie, not loaded from the disk).


patches/minecraft/net/minecraft/world/entity/Mob.java.patch
```diff
@@ -32,3 +32,13 @@
           if (entity != null) {
              double d0 = entity.m_20280_(this);
              int i = this.m_6095_().m_20674_().m_21611_();
+@@ -988,7 +_,8 @@
+       } else {
+          this.m_21559_(false);
+       }
+-
++      net.minecraftforge.event.entity.EntityFinalizeSpawnEvent e = new net.minecraftforge.event.entity.EntityFinalizeSpawnEvent(this, p_21435_, p_21436_, p_21437_, p_21438_);
++      net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(e);
```
SmellyModder: This can just be one line.
`net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.entity.EntityFinalizeSpawnEvent(this, p_21435_, p_21436_, p_21437_, p_21438_));`

src/main/java/net/minecraftforge/event/entity/EntityFinalizeSpawnEvent.java
```diff
@@ -0,0 +1,56 @@
...
+import net.minecraft.world.entity.SpawnGroupData;
+
+/**
+ * This event is called when an entity finalizes its spawn, usually to give it something (eg, skeletons with bows)<br>
+ * This event should be used if a modder wants to add something to an entity the first time it's spawned.<br>
+ * If a modder wants to add something everytime an entity is spawned they should use {@link EntityJoinWorldEvent}
+ */
+public class EntityFinalizeSpawnEvent extends EntityEvent
+{
+
```
SmellyModder: Remove this blank line.
```suggestion
```

src/main/java/net/minecraftforge/event/entity/EntityFinalizeSpawnEvent.java
```diff
@@ -0,0 +1,56 @@
...
+{
+    
+    private final DifficultyInstance difficultyInstance;
+    private final MobSpawnType mobSpawnType;
+    @Nullable
+    private final SpawnGroupData groupData;
+    @Nullable
+    private final CompoundTag compoundTag;
+
+    public EntityFinalizeSpawnEvent(Mob entity, DifficultyInstance difficultyInstance, MobSpawnType mobSpawnType, SpawnGroupData data, CompoundTag compoundTag)
```
SmellyModder: `@Nullable` should also be included in the nullable parameters.
```suggestion
    public EntityFinalizeSpawnEvent(Mob entity, DifficultyInstance difficultyInstance, MobSpawnType mobSpawnType, @Nullable SpawnGroupData data, @Nullable CompoundTag compoundTag)
```

Shadows-of-Fire:
What is the desired use case of this event?

seymourimadeit:
> What is the desired use case of this event?

The desired use case is for modders to use this event is if they want to modify an entity the first time they spawn. A modder might not want to use the EntityJoinWorldEvent because that event fires every time a mob joins a world, or use LivingSpawnEvent's subevents because those only fire when an entity is naturally spawned.

patches/minecraft/net/minecraft/world/entity/Mob.java.patch
```diff
@@ -32,3 +32,12 @@
           if (entity != null) {
              double d0 = entity.m_20280_(this);
              int i = this.m_6095_().m_20674_().m_21611_();
+@@ -988,7 +_,7 @@
+       } else {
+          this.m_21559_(false);
+       }
+-
++      net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.entity.EntityFinalizeSpawnEvent(this, p_21435_, p_21436_, p_21437_, p_21438_));
```
sciwhiz12: While a nitpick, you can shorten this patch by inserting the line and leaving the blank line unaffected.

src/main/java/net/minecraftforge/event/entity/EntityFinalizeSpawnEvent.java
```diff
@@ -0,0 +1,55 @@
+package net.minecraftforge.event.entity;
```
sciwhiz12: Missing license header, please run `updateLicenses`/`licenseFormat`.

src/main/java/net/minecraftforge/event/entity/EntityFinalizeSpawnEvent.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+    
+    @Nullable
+    public SpawnGroupData getGroupData() 
+    {
+        return groupData;
+    }
+
+    @Nullable
+    public final CompoundTag getCompoundTag()
```
sciwhiz12: Why is this getter `final`?

src/main/java/net/minecraftforge/event/entity/EntityFinalizeSpawnEvent.java
```diff
@@ -0,0 +1,55 @@
...
+public class EntityFinalizeSpawnEvent extends EntityEvent
+{
+    private final DifficultyInstance difficultyInstance;
+    private final MobSpawnType mobSpawnType;
+    @Nullable
+    private final SpawnGroupData groupData;
+    @Nullable
+    private final CompoundTag compoundTag;
+
+    public EntityFinalizeSpawnEvent(Mob entity, DifficultyInstance difficultyInstance, MobSpawnType mobSpawnType, @Nullable SpawnGroupData data, @Nullable CompoundTag compoundTag)
```
sciwhiz12: In the interests of convenience of avoiding a cast/pattern-matching `instanceof`, please override the `getEntity` getter and change the return type to `Mob`.[^1]

Also, is this event intended for _any_ `Entity` while finalizing spawning, or only for `Mob`s? If only for `Mob`s, it may be beneficial to rename this event to `MobFinalizeSpawnEvent`, to clarify that this only fires for mobs. If not, then the type of the first parameter should be expanded to `Entity` to signal that it may not necessarily be a `Mob` stored in this event.

[^1]: This is possible because [return types are _covariant_](https://www.baeldung.com/java-covariant-return-type) in Java 5+ -- that is, an override can narrow down its return type to be a subclass of the overriden method's return type.

src/main/java/net/minecraftforge/event/entity/EntityFinalizeSpawnEvent.java
```diff
@@ -0,0 +1,55 @@
...
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.world.DifficultyInstance;
+import net.minecraft.world.entity.Mob;
+import net.minecraft.world.entity.MobSpawnType;
+import net.minecraft.world.entity.SpawnGroupData;
+
+/**
+ * This event is called when an entity finalizes its spawn, usually to give it something (eg, skeletons with bows)<br>
+ * This event should be used if a modder wants to add something to an entity the first time it's spawned.<br>
+ * If a modder wants to add something everytime an entity is spawned they should use {@link EntityJoinWorldEvent}
```
sciwhiz12: Rather then `everytime an entity is spawned`, it may be more understandable if it were reworded to `everytime an entity is added to the world`, as "spawned" is more likely to be understood in the context of "being spawned by something, like a spawner, spawn egg, etc."

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -84,7 +84,6 @@
 import net.minecraft.world.level.LevelReader;
 import net.minecraft.world.level.Level;
 import net.minecraft.server.level.ServerLevel;
-import net.minecraft.world.level.biome.MobSpawnSettings;
```
sciwhiz12: Please do not modify unrelated files.

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/event/entity/MobFinalizeSpawnEvent.java
```diff
@@ -0,0 +1,83 @@
...
+    @Nullable
+    public CompoundTag getCompoundTag()
+    {
+        return compoundTag;
+    }
+    
+    @Override
+    public Mob getEntity()
+    {
+        return mob;
```
sciwhiz12: You can avoid storing the `Mob` instance in a field by replacing this with `(Mob) super.getEntity()` instead. IIRC, the compiler is smart enough to know that it won't cause an unchecked cast and therefore not emit a warning for such.

src/main/java/net/minecraftforge/event/entity/MobFinalizeSpawnEvent.java
```diff
@@ -0,0 +1,80 @@
...
+    @Nullable
+    public CompoundTag getCompoundTag()
+    {
+        return compoundTag;
+    }
+    
+    @Override
+    public Mob getEntity()
+    {
+        return (Mob)super.getEntity();
```
sciwhiz12: A very small formatting nitpick: space between cast and expression.
```suggestion
        return (Mob) super.getEntity();
```

sciwhiz12 (changes requested):
A small formatting nitpick, and then this looks fine to me for assignment. (Please re-request my review once you've applied the change.) Also, please rename your PR's title to match the event's new name.

sciwhiz12 (approved):

TheCurle:
The idea is great, but the API itself is undocumented and we haven't heard from the author in a while, so I'm going to close this.

If somebody wants to make something similar to allow modifying mobs on first spawn, and document it properly, then feel free.

PR #8267, opened 2021-12-10
Fix wrong stage being declared in ModStateProvider

williewillus:
Not sure what exactly this impacts, I believe DeferredWorkQueue might not work properly here?

But either way, it's pretty obviously incorrect. Every other field has a stage that matches the field name, except this one.


sciwhiz12 (approved):
A simple oversight with a simple fix. :shipit:

PR #8278, opened 2021-12-10
[1.18] Entity Model Animation API

MrCrayfish:
~~A set of events to allow applying animations/custom poses to the player model with the safety of knowing modders can't destroy the definitions of the original player model by restoring it's default pose before every call. I've included a test mod so you can understand the use of each event.~~

I've updated to PR to become an entity model animation API. It allows modders to hook into existing entities and apply custom animations to them safely. It could also be used for entities a modder may own. It is currently a draft as it only supports as low as Living Entities, the goal would be to support all entities. Feel free to post feedback.

TODO: Update this to show examples


MrScautHD:
nice work!

MrScautHD:
but a disable Render item Event was also cool if you can add it 🍡

BluSunrize:
Do we have to make this player-only? Wouldn't this be much more useful on a broader scale if it worked for all Bipeds?

MrCrayfish:
> Do we have to make this player-only? Wouldn't this be much more useful on a broader scale if it worked for all Bipeds?

I've been here for a while trying to write a repsonse to this. I don't think adding an event for biped/humanoid model is a great idea based on the fact there would be an extraordinarily more event calls (remember this event is called every frame, not tick). The events in my PR actually never get called for the local player when in first person. This means that if you are in single player, there would be little to no impact on performance. Now let's say there is an event for biped, every zombie, every skeleton, every model that extends the biped model is now calling the event two times (assuming a pre and post impl). I simply believe it's not worth to write a solution to satisfy all.

ChampionAsh5357:
I have a few complaints and concerns about this system.

First, it adds three new events for rendering the player which only have an effect on the player when in third person or a remote player. This means that any outside rendering of the player model in other environments could not have these attributes applied to them.

Second is that it only affects the player. An animation API should be extensible to any entity. Then we would not have to reimplement the system for every single entity, resulting in more events and more patches.

Third is that the poses defaulted are those which are within the system. What is the need of this? It's very inextensible behavior and the movement of the body could just be done by modifying the value directly. The need for more than one event here is strange as it will always be based upon the last person to edit the ModelPart.

Since all baked layers have access to the root model part, you could essentially just make a simple registry that takes in applied animations given the entity's current condition as seen on the client and grab the necessary model parts from the root. That would be my opinion for a more extensible and efficient system.

MrCrayfish:
> First, it adds three new events for rendering the player which only have an effect on the player when in third person or a remote player. This means that any outside rendering of the player model in other environments could not have these attributes applied to them.

I mean that could be changed. Right now the animations wouldn't be applied to the player in the inventory.

> Second is that it only affects the player. An animation API should be extensible to any entity. Then we would not have to reimplement the system for every single entity, resulting in more events and more patches.

Said events are only useful to entites you do no own. I would be curious to know if modders are even attempting to modify the animation of other entities that don't own besides the player. Basically is it even worth it to write an Animaiton API to cover all? 

> Third is that the poses defaulted are those which are within the system. What is the need of this? It's very inextensible behavior and the movement of the body could just be done by modifying the value directly. The need for more than one event here is strange as it will always be based upon the last person to edit the ModelPart.

For some model parts, if you modifiy the position it will stay in that position forever. The whole point is that Forge should handling restoring the original positon/rotation rather than leaving that up to modders. Please read the documentation of the events if you haven't, I didn't just call them at specific points for no reason. Pre is called after regular biped animations are applied, this means you could apply custom arm animations while still having the walking animaiton applied. Also because Pre is called because player model specific animations, the arm/leg layers will be automatically copied to the arm/legs transforms which avoids writing extra code. How it is now is animations are applied on top of previous animations unless you cancel the event to handle your own. I would assume most modders would wrapped their modifications inside of a condition.

ChampionAsh5357:
> Said events are only useful to entites you do no own. I would be curious to know if modders are even attempting to modify the animation of other entities that don't own besides the player. Basically is it even worth it to write an Animaiton API to cover all?

My opinion is that forge systems should be extensible to its target and then specified in layers on top of it. If an extensible system got written for players, it could easily be applied to other models as well. As for what you currently have, it could be applied to the humanoid models. Whether or not modders will use it is up to the modders. How many use cases it suffice is dependent on whoever wants to use the system. I can think of a few which may benefit from mob animations, though whether or not they'd implement is up to them.

> For some model parts, if you modifiy the position it will stay in that position forever. The whole point is that Forge should handling restoring the original positon/rotation rather than leaving that up to modders. Please read the documentation of the events if you haven't, I didn't just call them at specific points for no reason. Pre is called after regular biped animations are applied, this means you could apply custom arm animations while still having the walking animaiton applied. Also because Pre is called because player model specific animations, the arm/leg layers will be automatically copied to the arm/legs transforms which avoids writing extra code. How it is now is animations are applied on top of previous animations unless you cancel the event to handle your own. I would assume most modders would wrapped their modifications inside of a condition.

I do read the entire documentation, I try to be thorough when I go through PRs. The need for three events is still not useful in my opinion. If models relied on the previously set state of the model, then maybe more events would be necessary. However, they act in full isolation. Pre and blank are only useful when cancelled, but they would achieve the same goal if called after post.   If multiple animations are playing, then that is how it will be. Nothing you can do about that. One person will always win in the end. Stacking animations is just as possible with only post regardless anyways. As for finding the default state, it can simply be inferred from the model itself. A modder will know what model and the data each part will hold. However, if you could just loop through the defaults and store them in a map, that I can probably agree with. But setting it for the modder is up to the modder. 

After an additional look through, I don't think your events have enough context. You are only supplying the player, but you are neglecting the other five animation fields passed into the renderer. Yes you could construct them yourself since models in a renderer context are singletons, but the data is there so you don't have to do that. And again, you could probably just store the root part and the default poses of all parts in some kind of map, make a registry for entity type to model animator which has some condition to execute, if it should attempt to stack, and the necessary values and context lookup to set that data as necessary. That would more or less make it extensible to all entities and serve the purpose you are trying to accomplish above.

MrCrayfish:
I've updated this PR to allow creating animations for a larger range of entities. At the moment, it only supports as low as a living entities, any further would require significant code changes. I would like feedback before continuing.

patches/minecraft/net/minecraft/client/model/EntityModel.java.patch
```diff
@@ -0,0 +1,22 @@
+--- a/net/minecraft/client/model/EntityModel.java
++++ b/net/minecraft/client/model/EntityModel.java
+@@ -8,7 +_,8 @@
+ import net.minecraftforge.api.distmarker.OnlyIn;
+ 
+ @OnlyIn(Dist.CLIENT)
+-public abstract class EntityModel<T extends Entity> extends Model {
++public abstract class EntityModel<T extends Entity> extends Model implements net.minecraftforge.client.animation.IAnimatedModel
++{
```
sciwhiz12: Please put the brace back on the same line.

src/main/java/net/minecraftforge/client/animation/IAnimatedModel.java
```diff
@@ -0,0 +1,44 @@
...
+     * the name "body_0", "body_1" and so on.
+     *
+     * @param partMap the map to put the parts
+     * @param parts an iterable of model parts
+     * @param prefix the name to prefix for each part
+     */
+    static void fillFromIterable(Map<String, ModelPart> partMap, Iterable<ModelPart> parts, String prefix)
+    {
+        int i = 0;
+        for(ModelPart part : parts) partMap.put(prefix + "_" + i, part);
```
sciwhiz12: This seems to be missing an increment for `i`; as-is, all the parts are added with the same `0` index.

src/main/java/net/minecraftforge/client/event/EntityRenderersEvent.java
```diff
@@ -144,4 +150,36 @@ public EntityModelSet getEntityModels()
...
+                ((LivingEntityRenderer<T, ?>) renderer).getAnimator().addAnimation(animation);
+            });
+        }
+
+        /**
+         * Gets a map of the registered animations. Used internally.
+         */
+        public Map<EntityType<?>, List<Consumer<LivingEntityRenderer<?, ?>>>> getAnimations()
+        {
+            return ImmutableMap.copyOf(animations);
```
sciwhiz12: This would still mean the lists present in the immutable map can be modified and affect the original list, which may be undesirable. You may want to consider using a `ListMultimap` internally, and returning a (cached) view of the multimap using `Multimaps#asMap` and `Multimaps#unmodifiableListMultimap` (or `ImmutableListMultimap#copyOf` for a copy and not a view of the underlying multimap).

src/test/java/net/minecraftforge/debug/client/AnimateEntityLivingTest.java
```diff
@@ -0,0 +1,190 @@
...
+                rightLeg.yRot = (float) Math.toRadians(90F);
+                root.get("left_pants").copyFrom(leftLeg);
+                root.get("right_pants").copyFrom(rightLeg);
+            }
+        });
+    }
+
+    private void setupZombieAnimationTest(EntityRenderersEvent.AddAnimations event)
+    {
+        event.addAnimation(EntityType.ZOMBIE, new EntityAnimation<>(EntityAnimation.Mode.PASSIVE, EntityAnimation.Priority.FIRST)
```
sciwhiz12: For completeness' sake, please add a comment explaining what this animation does, as you've done for the others.

src/test/java/net/minecraftforge/debug/client/AnimateEntityLivingTest.java
```diff
@@ -0,0 +1,190 @@
...
+                root.get("right_arm").xRot = (float) Math.toRadians(rotation);
+                root.get("left_arm").xRot = (float) Math.toRadians(rotation + 180F);
+            }
+        });
+    }
+
+    // Creates a test to show active animations cancelling later priority active animations
+    private void setupArmorStandAnimationTest(EntityRenderersEvent.AddAnimations event)
+    {
+        event.addAnimation(EntityType.ARMOR_STAND, new EntityAnimation<>(EntityAnimation.Mode.ACTIVE, EntityAnimation.Priority.FIRST)
```
sciwhiz12: For completeness' sake, please add a comment explaining what these two animations do, as you've done for the others.

src/test/resources/META-INF/mods.toml
```diff
@@ -172,5 +172,6 @@ license="LGPL v2.1"
     modId="part_entity_test"
 [[mods]]
     modId="capabilities_test"
-
+[[mods]]
+    modId="animate_entity_test"
```
sciwhiz12, line 182: Insert a blank line after your inserted `modId` line.

patches/minecraft/net/minecraft/client/model/EntityModel.java.patch
```diff
@@ -0,0 +1,22 @@
...
+    public boolean f_102610_ = true;
+@@ -30,5 +_,9 @@
+       p_102625_.f_102608_ = this.f_102608_;
+       p_102625_.f_102609_ = this.f_102609_;
+       p_102625_.f_102610_ = this.f_102610_;
++   }
++
++   @Override
++   public void gatherAnimatedParts(java.util.Map<String, net.minecraft.client.model.geom.ModelPart> partMap) {
+    }
```
sciwhiz12: Is there a particular reason why this is not a defaulted no-op method in the interface, but instead defined here as a no-op?

src/main/java/net/minecraftforge/client/animation/DefaultPoseHolder.java
```diff
@@ -0,0 +1,83 @@
...
+import net.minecraft.client.model.geom.ModelPart;
+import net.minecraft.client.model.geom.PartPose;
+import org.apache.commons.lang3.tuple.Pair;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class DefaultPoseHolder
+{
+    private final ImmutableList<Pair<ModelPart, PartPose>> defaultPoseList;
```
sciwhiz12: I'd refrain from using the concrete `ImmutableList` type over using the general `List` type for this.

src/main/java/net/minecraftforge/client/animation/EntityAnimation.java
```diff
@@ -0,0 +1,173 @@
...
+     * Determines how animations are executed. The order is based on the mode and priority
+     * of the animation. Below is a visual representation of the ordering.
+     * <ul>
+     * <li>PASSIVE - FIRST</li>
+     * <li>PASSIVE - DEFAULT</li>
+     * <li>PASSIVE - LAST</li>
+     * <li>ACTIVE - FIRST</li>
+     * <li>ACTIVE - DEFAULT (Cancelled if an ACTIVE - FIRST is executed)</li>
+     * <li>ACTIVE - LAST (Cancelled if an ACTIVE - FIRST/DEFAULT is executed)</li>
+     * </ul>
```
sciwhiz12: Just to clarify, this is in increasing order; so an element lower in the list is 'greater' than an element higher in the list?

src/main/java/net/minecraftforge/client/animation/EntityAnimation.java
```diff
@@ -0,0 +1,173 @@
...
+         * an active animation will be executed at this priority if one was executed in an early
+         * priority.
+         */
+        LAST
+    }
+
+    /**
+     * A record containing data used when calculating animations with the addition of partial ticks
+     */
+    protected record Context(float animateTicks, float animateSpeed, float bobAnimateTicks, float headYaw, float headPitch, float partialTicks) {}
```
sciwhiz12: Please put the brace on a newline. Also, I believe the convention in Mojang code is that this is called the 'partial tick' (since it's a fraction of a single tick).

patches/minecraft/net/minecraft/client/model/EntityModel.java.patch
MrCrayfish: Fixed

src/main/java/net/minecraftforge/client/animation/IAnimatedModel.java
MrCrayfish: Fixed

src/test/java/net/minecraftforge/debug/client/AnimateEntityLivingTest.java
MrCrayfish: Fixed. Added more comments.

src/test/java/net/minecraftforge/debug/client/AnimateEntityLivingTest.java
MrCrayfish: Fixed. Included more comments and renamed method to clearly indicate test.

patches/minecraft/net/minecraft/client/model/EntityModel.java.patch
MrCrayfish: Just a developer forgetting I could have used a default method. Fixed.

src/main/java/net/minecraftforge/client/animation/DefaultPoseHolder.java
MrCrayfish: Fixed

src/main/java/net/minecraftforge/client/animation/EntityAnimation.java
MrCrayfish: Yes, you are correct. It's "greater" with the exception of an `ACTIVE` animation, which one with `FIRST` priority (if executed) would cancel later prioritised animations (`DEFAULT` and `LAST`)

src/main/java/net/minecraftforge/client/animation/EntityAnimation.java
MrCrayfish: Fixed and changed

src/main/java/net/minecraftforge/client/event/EntityRenderersEvent.java
MrCrayfish: Good catch. I've patched it to use `ListMultimap` and return a copy using`ImmutableListMultimap#copyOf`

MrCrayfish:
At a point where additional changes to patches would be needed to progress futher. Here is where I am in the API and the concerns I going forward.

- Currently this animation system only supports Living Entities since the majority of them use a common renderer `EntityLivingRenderer`. In order to support all entities, the animation system will have to be included in individual `EntityRenderer` implementations unless a better approach is suggested.
- The animation system only supports creating animations for models with ModelParts. Are there any use cases outside of Model Parts that should be considered? I looked at @gigaherz's Renderable API (https://github.com/MinecraftForge/MinecraftForge/pull/8259) and it seems I could supply transform matrices.
- The system has be manually included into entity renderers. This means that modders will have to write support for `EntityAnimator` in their own entity renderers if they want to use the animation system. I don't think there is any way to make this seamless. I am probably going to write an interface to mark which entity renderers support animations and it'll be able to return the `EntityAnimator` instance.

As for anything that I would still like to add. I think the possibility to provide transformations on the pose stack before rendering the model would be useful. Since events like RenderPlayerEvent don't have the local transforms before the player model is rendered, would be good for animations.

src/main/java/net/minecraftforge/client/animation/EntityAnimation.java
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * The condition to test to determine if the animation can be executed. This is where a test like
+     * holding or using a particular item, riding an entity, etc should be used. There should always
+     * be a condition and never just true.
+     *
+     * @param entity the entity the animation is being applied to
+     * @return the result of the test
+     */
+    public abstract boolean canRun(E entity);
```
SmellyModder: I don't see why this method needs to exist.
The doc for this method says, "There should always be a condition and never just true.", but I don't see how that would always be the case. Many animations have no playing conditions, and this method does unnecessary branching for animations that always play.

Also, modders could want to process multiple conditioned stages to an animation(s) in one `EntityAnimation` instance as a separate instance per conditioned animation stage is verbose, and the `EntityAnimation` instances are processed iteratively. If a modder wanted to do this, this method becomes weird as the structure of an `EntityAnimation` suggests conditionals should only get done in this method.

Furthermore, this method only has the `Entity` instance in its parameters, which is pretty limiting for conditional animations as modders may want to check for values in the `ModelComponent` instance and `Context` instance in the `apply()` method.

src/main/java/net/minecraftforge/client/animation/EntityAnimation.java
```diff
@@ -0,0 +1,174 @@
...
+    {
+        if (this.getMode() == o.getMode())
+        {
+            return this.getPriority().ordinal() - o.getPriority().ordinal();
+        }
+        else if (o.getMode() == Mode.ACTIVE)
+        {
+            return -1;
+        }
+        return 1;
```
SmellyModder: This can be one line.
```suggestion
        return this.getMode() == o.getMode() ? this.getPriority().ordinal() - o.getPriority().ordinal() : o.getMode() == Mode.ACTIVE ? -1 : 1;
```

src/main/java/net/minecraftforge/client/animation/EntityAnimation.java
```diff
@@ -0,0 +1,174 @@
...
+         * an active animation will be executed at this priority if one was executed in an early
+         * priority.
+         */
+        LAST
+    }
+
+    /**
+     * A record containing data used when calculating animations with the addition of partial ticks
+     */
+    protected record Context(float animateTicks, float animateSpeed, float bobAnimateTicks, float headYaw, float headPitch, float partialTick)
```
SmellyModder: I don't see why this record should exist.
It seems like the data in this record can just get passed in `apply()`  as parameters to avoid initialization every frame, and getting the data by doing `context.animateTicks`, `context.animateSpeed`, and so on is pretty verbose.

SmellyModder:
This looks very promising 👍 
Hopefully, more animation mods will become compatible if this gets merged.

src/main/java/net/minecraftforge/client/animation/EntityAnimation.java
SizableShrimp: Why does it need to be? Nested ternaries lead to code that is confusing to read and doesn't seem necessary here.

TheCurle: Don't make unreadable code for the sake of unreadability. This can be multiple lines just fine.

src/main/java/net/minecraftforge/client/animation/EntityAnimation.java
MrCrayfish: > The doc for this method says, "There should always be a condition and never just true.", but I don't see how that would always be the case. Many animations have no playing conditions, and this method does unnecessary branching for animations that always play.

I should reword the documentation, you should be allowed to run the animation all the time if necessary. 

> Also, modders could want to process multiple conditioned stages to an animation(s) in one EntityAnimation instance as a separate instance per conditioned animation stage is verbose, and the EntityAnimation instances are processed iteratively. If a modder wanted to do this, this method becomes weird as the structure of an EntityAnimation suggests conditionals should only get done in this method.

I definitely didn't intend for this come across to modders that you need to use multiple `EntityAnimation` instances for an animation with conditional stages. Maybe rewording the method to `canStart` would clear it up since something could start but it doesn't mean it'll finish, whereas `canRun` might indicate the whole animation will be applied. The method exists for two reasons, the main one being that once an active animation is applied, it will ignore the remaining active animations.
```java
if (animation.canRun(entity))
{
    animation.apply(entity, this.root, context);
    if (animation.getMode() == EntityAnimation.Mode.ACTIVE)
    {
        break;
    }
}
```
The second is from a readability stand point. I believe it should be clear when an animation can start. Even if I removed the `canRun` method, it'd still replicate the same bahaviour by making `apply` return a boolean. It's really just a alternaltive way to write it.

> Furthermore, this method only has the Entity instance in its parameters, which is pretty limiting for conditional animations as modders may want to check for values in the ModelComponent instance and Context instance in the apply() method.

I don't think `ModelComponent` would actually be useful. The only thing you can get from `ModelComponent` is model parts. If testing if a model part exists, as a modder you'd already know since you can look at the `LayerDefinition`. As for testing the rotation/position of a model part, I don't know if any modder needs that super fine control. If they do, it can be applied in the `apply` method.

src/main/java/net/minecraftforge/client/animation/EntityAnimation.java
MrCrayfish: I still have a few more things to include into `Context`. I think in this scenario it's much better to put everything into `Context` than have a method with 20 parameters. It's all optional data that can be used for creating animations.  If being a little verbose is better for readability, it's worth it. I can avoid initializing it every frame and just have persistent instance that gets updated before calling `apply`.

SmellyModder: I did not expect many more parameters to get added to the method, and a long strand of hard-to-read parameters does not sound pretty.
Avoiding initialization every frame by using a routinely updated persistence instance sounds good 👍

src/main/java/net/minecraftforge/client/animation/EntityAnimation.java
SmellyModder: > I should reword the documentation, you should be allowed to run the animation all the time if necessary.

Sounds good 👍 

>  The method exists for two reasons, the main one being that once an active animation is applied, it will ignore the remaining active animations.

Ah, I totally overlooked this, and yeah, it is just an alternative way to write it.

> I don't think ModelComponent would actually be useful. The only thing you can get from ModelComponent is model parts. If testing if a model part exists, as a modder you'd already know since you can look at the LayerDefinition. As for testing the rotation/position of a model part, I don't know if any modder needs that super fine control. If they do, it can be applied in the apply method.

I agree those variables would be useless in most cases, but this system should be as versatile as possible when given easy opportunities.
If a modder needed those variables for deciding if the animation can run, there is no simple way to prevent the iteration over the animations from ending when their animation should not get run.

src/main/java/net/minecraftforge/client/animation/EntityAnimation.java
MrCrayfish: Just to be more clear, I was exaggerating to 20 parameters. I wasn't sure on exactly the amount I would be adding. It would still end up being a readability problem.

MrCrayfish: On second thought, I think `Context` needs to be more versatile. As it is right now, all the values (except partialTick) are speicific to a living entity. Maybe a system similar to `ModelComponent` where you can retrieve data with a key.

src/main/java/net/minecraftforge/client/animation/EntityAnimation.java
MrCrayfish: So I've reworked the system and custom data can now be passed and retrieved from `AnimationedData`. I've included this in the `ModelAnimation#canStart` method.

MrCrayfish:
Anyone who stumbles back here to take a look at the new changes, here is a quick update
- Replaced `Context` with `AnimationData` which allows for custom data to be passed easily to animations. Data is added and retrieved from `AnimationData` using an `AnimationKey`.
```java
// Pushing data from renderer
ModelAnimator animator = ...
AnimationKey<Float> YAW = new AnimationKey<>(0F); // 0F is default value
animator.pushData(YAW, entity.yaw);

/* The data then stored into AnimationData and passed to ModelAnimation#apply */

// Retrieving data
AnimationData data = ....
float entityYaw = data.get(YAW);
```
- Animation system is no longer locked down to regular Entities but now allows Block Entities.
```java
// New events
EntityRenderersEvent.AddAnimations.Entity
EntityRenderersEvent.AddAnimations.BlockEntity
```
- `DefaultPoseHolder` now `ModelPoseHolder` has been generalised since it could be used outside of animations. I thought why not.

Having that said, looking for reviews and feedback. Something in particular I thought would be helpful to modders is being able to dump the `ModelTree` structure and `AnimationData`. I'd probably have to add a string id for `AnimationKey` since there would be no way to know which piece of data is tied to which key. Any suggestions on a graceful way to dump this data, please let me know!

patches/minecraft/net/minecraft/client/model/AgeableListModel.java.patch
```diff
@@ -0,0 +1,14 @@
+--- a/net/minecraft/client/model/AgeableListModel.java
++++ b/net/minecraft/client/model/AgeableListModel.java
+@@ -76,4 +_,11 @@
+    protected abstract Iterable<ModelPart> m_5607_();
+ 
+    protected abstract Iterable<ModelPart> m_5608_();
++
++   @Override
++   public void gatherAnimatedParts(java.util.function.BiConsumer<String, ModelPart> consumer) {
++      // This might be bad since changes in the iterables
```
ChampionAsh5357: While true, the convention for these is the iterables to be an unchanging immutable list, so there should be no issue.

patches/minecraft/net/minecraft/client/model/HumanoidModel.java.patch
```diff
@@ -0,0 +1,30 @@
...
+       this.f_102810_ = p_170679_.m_171324_("body");
+@@ -383,6 +_,11 @@
+    private HumanoidArm m_102856_(T p_102857_) {
+       HumanoidArm humanoidarm = p_102857_.m_5737_();
+       return p_102857_.f_20912_ == InteractionHand.MAIN_HAND ? humanoidarm : humanoidarm.m_20828_();
++   }
++
++   @Override
++   public void gatherAnimatedParts(java.util.function.BiConsumer<String, ModelPart> consumer) {
++      this.root.getChildren().forEach(consumer);
```
ChampionAsh5357, line 27: Why not use the super implementation for this?

patches/minecraft/net/minecraft/client/model/geom/ModelPart.java.patch
```diff
@@ -0,0 +1,13 @@
+--- a/net/minecraft/client/model/geom/ModelPart.java
++++ b/net/minecraft/client/model/geom/ModelPart.java
+@@ -152,6 +_,10 @@
+       return Stream.concat(Stream.of(this), this.f_104213_.values().stream().flatMap(ModelPart::m_171331_));
+    }
+ 
++   public Map<String, ModelPart> getChildren() {
++      return this.f_104213_;
```
ChampionAsh5357: This should be an unmodifiable view.

patches/minecraft/net/minecraft/client/renderer/entity/LivingEntityRenderer.java.patch
```diff
@@ -29,6 +43,18 @@
           f8 = Mth.m_14179_(p_115310_, p_115308_.f_20923_, p_115308_.f_20924_);
           f5 = p_115308_.f_20925_ - p_115308_.f_20924_ * (1.0F - p_115310_);
           if (p_115308_.m_6162_()) {
+@@ -110,8 +_,11 @@
+          }
+       }
+ 
++      this.animator.restoreDefaultPose();
```
ChampionAsh5357, line 50: I don't see the use of setting the default pose still. In this design, it wouldn't matter whether the pose was reset or not. This is because `EntityModel#prepareMobModel` and `EntityModel#setupAnim` are both called directly after making any defaults restored to be a bit pointless. It's more a thing that is accessible to the animator instead of something that would need to be executed imo.

src/main/java/net/minecraftforge/client/animation/ModelPoseHolder.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * A class that stores a pose for an entire model
+ */
+public class ModelPoseHolder
```
ChampionAsh5357, line 35: I mentioned this in an earlier comment, but it would probably be nice to have access to the default pose of a particular model part.

ChampionAsh5357 (commented):
Just a few comments at a quick glance.

For the `ModelAnimator` for storing data, it may be a good idea to specify at what point should data be pushed since imo it could be stored in either `prepareMobModel` or `setupAnim`. It could also be set during the animation, but I'm not sure how safe that would be.

Another thing of note is that the implementation for `EnderDragonRenderer$DragonModel` is also not present at the time.

As for the idea of dumping the `ModelTree` and `AnimationData`, that would probably be a good design point. Since it seems as though Mojang is transitioning the model system to json, it would be a valuable addition, although not necessary at this time. The best you could do for dumping would be to probably have some kind of `#toString` or `JsonObject` dump to coagulate the information effectively, though at that point you may just want to start implementing codecs. `AnimationData` is relatively similar with the map codec being constructed after registration of all animation keys.

As for the `BlockEntityRenderer`s, I assume that their animators would also be implemented as some point, though I am unsure how the animations would be attached without be tightly coupled to the rendering process. Though this could also be just left as a modder api and not implemented in the vanilla structure to get around this.

patches/minecraft/net/minecraft/client/renderer/entity/LivingEntityRenderer.java.patch
ExpensiveKoala: The point of the resetting to default pose is because while rotation isn't destructive (destructive in the sense that it negatively influences subsequent animations), changing the rotation axis, part position, etc... Are destructive and need to be reset before each animation frame. You can see this being done in vanilla in certain models. One that comes to mind is the piglin model.

To elaborate further: imagine an arm with rotation axis 0,0,0. For rotation we apply sin/cos so it doesn't matter the previous value.
However, if we move the pivot to 0,10,0 for a specific animation, the next frame after the animation that moved the pivot finishes will have all of the values calculated based on the 0,10,0 pivot.
Ok, we make it up to the 0,10,0 animation to reset it. That means your animation either needs a "resetting" pose at the end or be applied before any and all other animations just for the resetting part.

ChampionAsh5357: You are describing the model holding state for animations, which is not good since the models are considered singletons in context to their associated renderers. Setting the default pose would only matter if you decided to give the model the ability to hold state which should not be done since you already have an object given to you to retrieve state data from. Additionally, it's called before prepare and setup which would set the rotation data to something other than the default making it a unnecessary call since the data is immediately overwritten before rendering. Resetting the pose of a model is not useful unless the modder desires it that way as part of the animation. Animation is a calculation of knowing the previous location, the new location, and the time delta in between where the locations are stored on the object source itself. So, needing to reset the data when you are just setting information anyways will already happen.

MrCrayfish: The problem is Minecraft is not explicit with it's animations. It may change the x and y of a pivot point but not the z. If a custom animation from this new system changes the z, the z will always hold the last value from the animation even if it's no longer running. As you said Minecraft models should not hold a state and that's why the model is reset to the default pose; in the example, the original z value.

patches/minecraft/net/minecraft/client/model/HumanoidModel.java.patch
MrCrayfish: `headParts()` and `bodyParts()` sometimes does not contain all the model parts. Best example is the player, the cloak and ears are not returned in either method. Not sure how this applies to other models but it just guarantees that I get all models parts.

patches/minecraft/net/minecraft/client/model/AgeableListModel.java.patch
MrCrayfish: Alright, will assume immutable.

MrCrayfish:
> For the ModelAnimator for storing data, it may be a good idea to specify at what point should data be pushed since imo it could be stored in either prepareMobModel or setupAnim. It could also be set during the animation, but I'm not sure how safe that would be.

I'll update the docs. By design, it should be pushed before calling `ModelAnimator#execute`. 

> Another thing of note is that the implementation for EnderDragonRenderer$DragonModel is also not present at the time.

DragonModel is just too conviluded, it mixes animations and rendering together and renders some model parts multiple times, for example the neck. I don't know if it's easily possible to integrate this animation system into the dragon model without completely refactoring the rendering process.

> As for the idea of dumping the ModelTree and AnimationData, that would probably be a good design point. Since it seems as though Mojang is transitioning the model system to json, it would be a valuable addition, although not necessary at this time. The best you could do for dumping would be to probably have some kind of #toString or JsonObject dump to coagulate the information effectively, though at that point you may just want to start implementing codecs. AnimationData is relatively similar with the map codec being constructed after registration of all animation keys.

I haven't used Codecs that much but I've looked into them since you mentioned it. I'd say that Codecs wouldn't really be applicable to `ModelTree` and `AnimationData` since they contain sensitive object instances. Only use would be allowing the data to be encoded into the desired format if that's what you originally meant.

> As for the BlockEntityRenderers, I assume that their animators would also be implemented as some point, though I am unsure how the animations would be attached without be tightly coupled to the rendering process. Though this could also be just left as a modder api and not implemented in the vanilla structure to get around this.

I've been reluctand to implementing the animator on individual renderers in order to keep the patches to a minimum. Currently only three block entity renderers (`BedRenderer`, `BellRenderer`, and `ShulkerBoxRenderer`) would work since it would not require the rendering to be refactored. Not sure it'd even be worth it and best to leave as a modder API as suggested.

@ChampionAsh5357

ChampionAsh5357:
> DragonModel is just too conviluded, it mixes animations and rendering together and renders some model parts multiple times, for example the neck. I don't know if it's easily possible to integrate this animation system into the dragon model without completely refactoring the rendering process.

Understandable. It can be left out for now since doing a full refactor is not really worthwhile and too invasive to properly handle it.

> I haven't used Codecs that much but I've looked into them since you mentioned it. I'd say that Codecs wouldn't really be applicable to ModelTree and AnimationData since they contain sensitive object instances. Only use would be allowing the data to be encoded into the desired format if that's what you originally meant.

The codecs were for just easily transforming the information from object to an easily readable json. Another added benefit would be that in the future, a json system could be integrated using this data. But if just wanted pure dumping, overriding `#toString` is more than enough I would say.

> I've been reluctand to implementing the animator on individual renderers in order to keep the patches to a minimum. Currently only three block entity renderers (BedRenderer, BellRenderer, and ShulkerBoxRenderer) would work since it would not require the rendering to be refactored. Not sure it'd even be worth it and best to leave as a modder API as suggested.

I would agree that its probably best to leave it alone. None of these animations are complex enough to warrant something different to happen to the existing parts.

MrCrayfish:
Do you have any follow up to the review converstations?

ChampionAsh5357:
> Do you have any follow up to the review converstations?

Apologies, looking over the two reviews, the arguments are both valid, so I made an assumption of communication without expressing my understanding. I will probably have time to look over everything again this week, so if I think of anything else, I'll mention it when I do.

oliviathevampire:
Does this support keyframe animations? Since vanilla will be getting it in 1.19

MrCrayfish:
Not the intention of this PR. This PR will just give you the hooks to write a keyframe animation system.

MrCrayfish:
Since 1.19 has introduced an animation system, it's likely I'll rewrite this PR to suit the new system.

PR #8283, opened 2021-12-10
Remove IForgeAbstractMinecart::getCartItem

josephtarbit:
Removes `IForgeAbstractMinecart::getCartItem` as it's longer needed (just override `AbstractMinecart::getPickResult`  instead).


Shadows-of-Fire:
Seems good, this is the only call site of this method and the usage is equivalent.
Plus given the current impl `getCartItem` is never invoked anyway, since `item` is never null.

LexManos:
Keep it but mark it as deprecated for removal in 1.19.

PR #8291, opened 2021-12-12
Fix particles going fullbright when picking up items or experience

### LINKED ISSUE #8221: [1.18.1] Particles go fullbright for a bit when picking up items or experience
**Minecraft Version:** 1.17.1

**Forge Version:** 37.0.126

**Steps to Reproduce:**
 1. Go into Survival
 2. Mine a block
 3. Pick up the drops

**Description of issue:**
The particles from the broken block will briefly become fullbright and black upon picking up an item or experience from a block. This is very similar to previously fixed #6706

**Image of issue:**
image

bageldotjpg:
This PR fixes #8221, essentially recreating fc3fa13bf3de2e6e34d5227a9218dffd9133719e which seems to have been left behind at some point.


TheCurle:
Can you provide a recording of with and without this fix?

bageldotjpg:
Here is what the particles look like without the fix: https://imgur.com/XJQLiOD
Here is what the particles look like with the fix: https://imgur.com/LehLD06

PR #8292, opened 2021-12-12
Rework Cobblestone tags to not change vanilla behavior

bageldotjpg:
With the current Cobblestone tag, vanilla recipe behavior is changed, which I don't think should be the case.
Infested Cobblestone and Mossy Cobblestone are likely purposely not able to be used in recipes currently because of their difference in makeup and to distance them from regular cobblestone, otherwise they'd be in the `stone_crafting_materials` tag.

Also, having them in the regular Cobblestone tag and replacing recipes creates weird situations, like mossy cobblestone + vine crafting mossy cobblestone, and using infested or mossy cobblestone in andesite and diorite recipes and not having them be mossy or infested.

This solution introduces `cobblestone/normal`, `cobblestone/infested`, and `cobblestone/mossy` and puts those tags in the Cobblestone tag, so mods can still use those tags for logic, but only uses the `cobblestone/normal` tag for recipes, to not alter vanilla recipe behavior or create odd situations.


SmellyModder (approved):

Mysterious-Dev (approved):
LGTM !

Shadows-of-Fire:
Seems fine, no point in arbitrarily adjusting vanilla behavior (even if it is fairly minor), better to keep it consistent.

bageldotjpg:
Messed things up locally, will be redoing

bageldotjpg:
Fixed up and added `COBBLESTONE_DEEPSLATE` tag because the same issue occurred with cobbled deepslate being in the cobblestone tag.

LexManos:
May be worth reviewing the cobblestone tag in the future and killing Forge's.
So that we just use vanilla's.

bageldotjpg:
Yeah, honestly the `stone_tool_materials` and `stone_crafting_materials` probably cover the use cases pretty well, definitely worth looking into in the future

PR #8293, opened 2021-12-12
[1.12.x] Remove log4j-slf4j18-impl

dualspiral:
In the recent update to log4j, the author of the patch included `log4j-slf4j18-impl` in the installer configuration. However, Forge doesn't use slf4j, and as the installer doesn't download transitive dependecies, this library is useless on its own. I don't think it should have been included (and don't think slf4j was ever part of Forge for 1.12 in the first place), so I'm PRing its removal.

We noticed this because the inclusion of this library broke plugin logging in SpongeForge - we use slf4j 1.7.25 and this dep is for the not yet released version 2.

See https://github.com/SpongePowered/SpongeForge/issues/3275

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -306,7 +306,6 @@ project(':forge') {
         installer 'net.sf.jopt-simple:jopt-simple:5.0.3'
         installer 'org.apache.logging.log4j:log4j-api:2.15.0' //TODO: Unpin in 1.18.1 or when Mojang bumps the Log4J version
         installer 'org.apache.logging.log4j:log4j-core:2.15.0' //TODO: Unpin in 1.18.1 or when Mojang bumps the Log4J version
-        installer 'org.apache.logging.log4j:log4j-slf4j18-impl:2.15.0' //TODO: Unpin in 1.18.1 or when Mojang bumps the Log4J version
 
         testImplementation 'org.junit.jupiter:junit-jupiter-api:5.0.0'
         testImplementation 'org.junit.vintage:junit-vintage-engine:5.+'


marchermans:
I am going to need to review the vanilla client classpath.
I am pretty sure an older version is included there, which means this is needed to be able to patch the whole completely.

dualspiral:
It's not as far as I can tell, on a pure Forge 14.23.5.2859 1.12.2 server I wrote a quick mod that just gets a logger from the slf4j logger factory:

```java
org.slf4j.LoggerFactory.getILoggerFactory().getLogger("this").info("Test");
```

and I got the following (as in the linked issue)

```
Caused by: java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory
        at com.example.examplemod.ExampleMod.init(ExampleMod.java:31) ~[ExampleMod.class:1.0]
...
```

This infers that slf4j is not on the vanilla classpath.

dualspiral:
I should read, I realise you said client. My mistake - but still, the line I gave should be the way to test I guess...

LexManos:
You know what, I really don't give a shit as this is a old version that people shouldn't be using.
So rather then spend the manhours to track this down, and considering the patched code is in core not slf4j, i'll just pull this.
But there was a reason we bumped the version, as something else was pulling in slf4j 2.11~ like the core/api libraries. So feel free to figure out what that was and tell it to stop.

aidanwd:
> You know what, I really don't give a shit as this is a old version that people shouldn't be using. So rather then spend the manhours to track this down, and considering the patched code is in core not slf4j, i'll just pull this. But there was a reason we bumped the version, as something else was pulling in slf4j 2.11~ like the core/api libraries. So feel free to figure out what that was and tell it to stop.

based

PR #8295, opened 2021-12-12
Allow sound instances to play custom audio streams

SquidDev:
### Motivation
There's a couple of mods out there which play dynamically generated sounds, rather than reading from an ogg file. For instance, [Etched](https://www.curseforge.com/minecraft/mc-mods/etched) allows streaming audio from an arbitrary URL, and [mxTune](https://www.curseforge.com/minecraft/mc-mods/mxtune) allows writing and playing MML files. In my case (https://github.com/cc-tweaked/CC-Tweaked/pull/982), players can generate music data through code and play it through "speaker" blocks.

While Minecraft's sound engine does support this, it's not exposed to `Sound`/`SoundInstance`s, meaning you need to poke the guts of the `SoundEngine`.

### This change
This adds a method to `SoundInstance` which creates an audio stream for a given `Sound`. By default it just uses vanilla logic, but mods can override this to return a custom `AudioStream` implementation.

The test mod adds a new sound which just plays a sine tone at 220Hz. This does still require some boilerplate: you do need to specify a `sounds.json` and list a dummy sound. The sound will never be played (and so can be empty), but _must_ exist, which is a little annoying.

### Alternative approaches
The obvious alternative here would be to add a new event instead (or even repurpose the `PlayStreamingSourceEvent` one; I couldn't find anybody actually using it). The current approach felt more natural, as whenever you want this functionality, you're going to be sub-classing `SoundInstance` anyway. But happy to go for either!


diesieben07:
Would it be warranted to add a more data-driven approach that lets resource-pack authors override this behavior instead of the sound implementation being hardcoded in the code? Similar to texture atlas sprite loaders?

SquidDev:
I think you could but I'm not sure how useful it could be in practice.

The main motivation for this was for streams which are tightly coupled to their `SoundInstance` sub-class and the `ResourceLocation` is hard coded elsewhere in the mod. It's not something which makes sense to configure via a resource pack.

My understanding of #7822 is that it's more intended for procedurally generated images, such as textures derived from other textures. This would be neat for sure, but I'm not aware of any mods which do/want to do this sort of thing for audio. I don't know, I'm aware the current approach is sub-par, but it's also simple, and very weary of adding lots of extra functionality which isn't going to be useful.

diesieben07:
It would allow a resource pack to say "no, I don't want your in-code generated audio, take this ogg file instead!"

SquidDev:
Is this a desirable feature? It feels a bit strange to me, in the same way I wouldn't expect a resource pack to mess with a `BlockEntityRenderer`.

diesieben07:
`BlockModelRenderer`? Do you mean a normal block model? If so resource packs can absolutely "mess" with them.
If you mean `BlockEntityRenderer`, I think it is only a matter of time before Mojang will replace those with a more data-driven format.

SquidDev:
Ahh, yes, should have been `BlockEntityRenderer`. Sorry! Still kinda think of them as TESRs. Apparently I'm too asleep right now to think, so will let this garner more feedback before making any changes.

TeamSpen210:
Perhaps a solution would be to modify the `sounds.json` logic, and fix the empty sound requirement by adding say a `"procedural": true` parameter? If `true`, sound files are prohibited, and `getStream()` is called. If `false`, it is ignored and regular sounds are played.

SquidDev:
I think you'd probably want to a accept an object there too, as that's the only place you can specify the attenuation distance. But otherwise this sounds reasonable.

The interface would be a bit nicer, but the patch would be substantially larger too. I don't know, really bad at telling whether these things are worth it or not!

Shadows-of-Fire:
I'm against trying to insert any extra data-driven stuff in there, as procedurally generated audio has no *practical* application being overridden by a datapack.  Sure, the capacity to sounds "neat" in theory, but it doesn't *make sense* to override a dynamically generated sound (like a streamed video or user-generated audio clip) with a static .ogg file.

src/test/java/net/minecraftforge/debug/client/AudioStreamTest.java
```diff
@@ -0,0 +1,91 @@
+package net.minecraftforge.debug.client;
```
sciwhiz12, line 6: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/test/java/net/minecraftforge/debug/client/AudioStreamTest.java
```diff
@@ -0,0 +1,91 @@
...
+import net.minecraftforge.fml.loading.FMLLoader;
+import org.jetbrains.annotations.NotNull;
+import org.lwjgl.BufferUtils;
+
+import javax.sound.sampled.AudioFormat;
+import java.nio.ByteBuffer;
+import java.util.concurrent.CompletableFuture;
+
+@Mod(AudioStreamTest.MOD_ID)
+public class AudioStreamTest
```
sciwhiz12, line 35: Please provide a javadoc for this class which explains what feature this tests, how to test it, and what the expected results are from said test.

src/test/java/net/minecraftforge/debug/client/AudioStreamTest.java
```diff
@@ -0,0 +1,91 @@
...
+            MinecraftForge.EVENT_BUS.addListener(this::onClientChatEvent);
+        }
+    }
+
+    public void onClientChatEvent(ClientChatEvent event)
+    {
+        if (event.getMessage().equalsIgnoreCase("sine wave"))
+        {
+            event.setCanceled(true);
+            Minecraft.getInstance().getSoundManager().play(new SineSound());
```
sciwhiz12: This test mod fails to load on the dedicated server because of the `SoundInstance` client-only class. Please fix this test mod so it does not cause a mod loading error due to classloading on the dedicated server.

sciwhiz12 (changes requested):
From my testing, it seems that the included test mod does not work: typing `sine wave` into chat causes the expected visual result of the chat message not appearing, yet there is no audible result of the sine wave.

src/test/java/net/minecraftforge/debug/client/AudioStreamTest.java
SquidDev: Have updated the license for this file. Worth noting the year is outdated on all files, so by default running `updateLicenses` will touch 600+ files - probably worth that being run separately and committed to trunk.

patches/minecraft/net/minecraft/client/resources/sounds/SoundInstance.java.patch
```diff
@@ -0,0 +1,14 @@
...
++++ b/net/minecraft/client/resources/sounds/SoundInstance.java
+@@ -50,4 +_,11 @@
+       NONE,
+       LINEAR;
+    }
++
++   /*================================ FORGE START ================================================*/
++
++   default java.util.concurrent.CompletableFuture<net.minecraft.client.sounds.AudioStream> getStream(net.minecraft.client.sounds.SoundBufferLibrary soundBuffers, Sound sound, boolean looping)
++   {
```
sciwhiz12: As this is a Minecraft class (in a patch), it should follow the formatting seen in Minecraft classes (as set by Fernflower): brackets on the same line. (Please affect this change in your IDE instead of using GitHub's web committing feature.)
```suggestion
+   default java.util.concurrent.CompletableFuture<net.minecraft.client.sounds.AudioStream> getStream(net.minecraft.client.sounds.SoundBufferLibrary soundBuffers, Sound sound, boolean looping) {
```

src/test/java/net/minecraftforge/debug/client/AudioStreamTest.java
```diff
@@ -0,0 +1,125 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.debug.client;
```
sciwhiz12: Outdated license header, please run the `updateLicenses`/`licenseFormat` task after updating this PR to the latest commit for `1.18.x`.

sciwhiz12 (changes requested):
Works fine to me; tested locally by first merging the latest commit from 1.18.x (no merge conflicts).

This PR still needs to be updated to latest base branch commit, for the new compact license header.

sciwhiz12 (approved):

PR #8296, opened 2021-12-13
[1.18.x] Ensure ScreenEvent doesn't accept null screens

mezz:
I encountered a crash where a mod uses its key event handler to set `minecraft.screen` to `null`, and doesn't cancel the event.
https://github.com/mezz/JustEnoughItems/issues/2597
This results in a crash in the consumers of the event that assume the `screen` field will not be `null`.

This PR hardens some places where these events are fired, and asserts the event's `screen` field isn't `null` in order to crash earlier, closer to the cause of the issue.


Shadows-of-Fire:
Looks good to me, ensures the chain of events is always fired for the same screen and hopefully prevents people from breaking things a bit more.

PR #8300, opened 2021-12-14
[1.18.x] Allow blocks to hide faces on a neighboring block

XFactHD:
This PR adds a hook that allows a block to hide a neighboring block's face that touches it. This is really useful for blocks like glass slabs which should behave identically to a glass block when placed on another glass block.

The level and position context is intended for blocks which need dynamic data from their `BlockEntity` to decide whether they would hide a neighboring block's face*. I have thought about the feasibility of using the `IModelData` system to allow modders to pass dynamic context to this method but this approach runs into multiple roadblocks:

- `Block#shouldRenderFace()` gets a `BlockGetter` while `ModelDataManager#getModelData()` requires a `Level`
- The `BlockGetter` given to `Block#shouldRenderFace()` is very unlikely to be the current client `Level` because the chunk rendering uses additional data structures that act similar to a `Level` but aren't actually one
- This would not work for blocks that use `BakedModel#getModelData()` to create the data object instead of using the `ModelDataManager`

*My usecase is a block which can be camouflaged as almost any full size block. Registering a dedicated variant for every camo block is not possible because some of those blocks come from a tag since it is completely impossible to determine if a block's model actually fills the whole cube for blocks which are not considered solid.

One thing I am not sure about and would love to get feedback on is the addition of `IForgeBlockGetter#getExistingBlockEntity()` and the reference to it from the javadoc on `IForgeBlock#hidesNeighborFace()`.
The intention behind it is to allow modders to access their BEs while making sure that someone passing an actual `Level` to `Block#shouldRenderFace()` doesn't accidentally cause the BE retrieval to create a BE that doesn't exist yet which would cause a CME when this happens in a render worker thread, for example while the world is still loading. The reason for this behavior is that `Level#getBlockEntity()` calls `LevelChunk#getBlockEntity()` with `LevelChunk.EntityCreationType.IMMEDIATE` to immediately create a `BlockEntity` for the given position if it has not been added yet.
The other option would be to document this even more thoroughly on `IForgeBlock#hidesNeighborFace()` and let the modder handle this differentiation themselves, though this feels very prone to error on the modder's side.

### DIFF
--- a/patches/minecraft/net/minecraft/world/level/block/Block.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/block/Block.java.patch
@@ -23,6 +23,15 @@
     }
  
     public static boolean m_152463_(BlockState p_152464_) {
+@@ -193,6 +_,8 @@
+       BlockState blockstate = p_152446_.m_8055_(p_152449_);
+       if (p_152445_.m_60719_(blockstate, p_152448_)) {
+          return false;
++      } else if (blockstate.hidesNeighborFace(p_152446_, p_152449_, p_152445_, p_152448_.m_122424_())) {
++         return false;
+       } else if (blockstate.m_60815_()) {
+          Block.BlockStatePairKey block$blockstatepairkey = new Block.BlockStatePairKey(p_152445_, blockstate, p_152448_);
+          Object2ByteLinkedOpenHashMap<Block.BlockStatePairKey> object2bytelinkedopenhashmap = f_49789_.get();
 @@ -324,7 +_,7 @@
     }
  
--- a/patches/minecraft/net/minecraft/world/level/block/Block.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/block/Block.java.patch
@@ -27,7 +27,7 @@
        BlockState blockstate = p_152446_.m_8055_(p_152449_);
        if (p_152445_.m_60719_(blockstate, p_152448_)) {
           return false;
-+      } else if (blockstate.hidesNeighborFace(p_152446_, p_152449_, p_152445_, p_152448_.m_122424_())) {
++      } else if (p_152445_.supportsExternalFaceHiding() && blockstate.hidesNeighborFace(p_152446_, p_152449_, p_152445_, p_152448_.m_122424_())) {
 +         return false;
        } else if (blockstate.m_60815_()) {
           Block.BlockStatePairKey block$blockstatepairkey = new Block.BlockStatePairKey(p_152445_, blockstate, p_152448_);
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
@@ -32,6 +32,7 @@ import net.minecraft.core.BlockPos;
 import net.minecraft.world.phys.HitResult;
 import net.minecraft.world.phys.Vec3;
 import net.minecraft.server.level.ServerLevel;
+import net.minecraftforge.client.ForgeHooksClient;
 import net.minecraftforge.common.ForgeHooks;
 import net.minecraftforge.common.IPlantable;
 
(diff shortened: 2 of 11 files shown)


patches/minecraft/net/minecraft/world/level/block/Block.java.patch
```diff
@@ -22,6 +22,15 @@
     }
  
     public static boolean m_152463_(BlockState p_152464_) {
+@@ -191,6 +_,8 @@
+       BlockState blockstate = p_152446_.m_8055_(p_152449_);
+       if (p_152445_.m_60719_(blockstate, p_152448_)) {
+          return false;
++      } else if (blockstate.hidesNeighborFace(p_152446_, p_152449_, p_152445_, p_152448_.m_122424_())) {
++         return false;
```
ChampionAsh5357: In my opinion, this is a level and position sensitive `#skipRendering` implementation. So, it would probably be better to replace that call, rename, and default to the vanilla method.

src/main/java/net/minecraftforge/common/extensions/IForgeBlockGetter.java
```diff
@@ -0,0 +1,27 @@
+package net.minecraftforge.common.extensions;
```
ChampionAsh5357, line 6: Missing license header

src/test/java/net/minecraftforge/debug/block/HideNeighborFaceTest.java
```diff
@@ -0,0 +1,95 @@
+package net.minecraftforge.debug.block;
```
ChampionAsh5357, line 6: Missing license header.

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -772,4 +772,26 @@ else if (state.is(Blocks.OBSERVER))
...
         }
     }
+
+    /**
+     * Whether this block hides the neighbors face pointing towards this blocks face in the given direction.
+     * <p>
+     * This method should only be used for blocks you don't control, for your own blocks override
+     * {@link Block#skipRendering(BlockState, BlockState, Direction)} on the respective block instead
+     * <p>
+     * WARNING: This method is likely to be called from a worker thread! If you want to retrieve a
```
ChampionAsh5357, line 761: The warning should also include information regarding to CMEs for any kind of write operation.

ChampionAsh5357 (changes requested):
Besides the fact that it should be treated as a more sensitive method, it looks useful to me. The addition of `#getExistingBlockEntity` also seems like a valid idea to try and mitigate people having to check a specific case each time.

I am concerned on whether or not `LevelChunk#getBlockEntity` is thread safe at all though. It can remove BEs or promote BEs that are being loaded, making it not thread safe.

ChampionAsh5357 (approved):
Another triage member (sciwhiz) has confirmed the safety of the statement. As such, I will approve this now

sciwhiz12 (changes requested):
Please update the license headers in the following files:

- `src/main/java/net/minecraftforge/common/extensions/IForgeBlockGetter.java`
- `src/test/java/net/minecraftforge/debug/block/HideNeighborFaceTest.java`

sciwhiz12 (approved):

marchermans:
We talked about this internally some more.

By default, we don't really like having APIs in forge which allow the modification of behavior of other mods. This would mean that normally this PR would be out of scope for forge, however we also agree with the fact that camouflage is a concept that we should support up to a certain degree.

The discussion ended up with a compromise:
- We would like to shrink the scope of this functionality way down. As it states right now the scope is way to large, and there is no way for mods to opt out of this functionality at all.
- We realized that for solid blocks this behavior is completely unwanted.
- We want an override on the neighbor block (the one whose sides this PR would hide) to completely opt out and fallback to the current behavior.

This al caused us to come up with the following compromise:
1) Add a method to IForgeBlock that functions as an override. Something like `supportsExternalFaceHiding` this method returns a `boolean` which indicates if a neighbor can hide a face or not. Use the DistExecutor to check for the rendering type that the block uses, if it is solid return `false`, if not return `true`. This should result in the required override behaviour.
2) Add a check for this method, before executing your hiding logic, if it is false default to the current behavior.

Would that work for you?

XFactHD:
That's a fair compromise, I can work with that.

marchermans (approved):

sciwhiz12:
:shipit: Published as **1.18.2-40.0.15**.

PR #8303, opened 2021-12-17
Add removal reason for BlockEntity#setRemoved

### LINKED ISSUE #8302: BlockEntity#setRemoved now also being called when a chunk unloads, causing unexpected behavior when performing cleanup
**Minecraft Version:** `1.18.1`

**Forge Version:** {Forge version. *Version number, not latest/rb*} `39.0.0`

**Description of issue:**
In 1.16.5, I used `BlockEntity#setRemoved` to perform cleanup of data that is decoupled from the block entity.
It was the most reliable method to detect a removal of a block entity.

In 1.18.1, `BlockEntity#setRemoved` is now also called when a chunk unloads. This isn't incorrect, in fact, it's probably even better that it's called consistently in all cases (see `ServerLevel#unload` and `LevelChunk#clearAllBlockEntities`).

However, it poses an unexpected change for mods porting from 1.16.5.

**I don't want my decoupled data being removed just because a chunk unloads where the block entity was in.**

Luckily, I can still detect that `#setRemoved` was called because of an unload, because `#onChunkUnloaded` is called first (see  `LevelChunk#clearAllBlockEntities`). See my workaround in https://github.com/refinedmods/refinedstorage/commit/a522edea083d57c14c3309af88018618ee3ed92c

But it's a bit of a flaky/unstable way of detecting it. 

Is there a way to make it more explicit in `#setRemoved` that it was called because of chunk unload?

For clarity: I'm **NOT** asking to remove calls to `#setRemoved`. It would just be helpful to know *why* it was called, since it's quite a generic method.

sciwhiz12:
This PR closes #8302 by implementing an overload for `BlockEntity#setRemoved` with a parameter for the **removal reason** for the `BlockEntity`, modeled after `Entity.RemovalReason`. The enum `IForgeBlockEntity.RemovalReason` provides the four removal reasons that may be passed into the new overload, with documentation as to what they represent and where they are called.

The original `setRemoved()` is kept for compatibility reasons, and redirects to `setRemoved(RemovalReason)` with the catch-all `REMOVED` reason. The new overload is specified in the extension interface for `BlockEntity` for the javadoc.

Furthermore, the `IForgeBlockEntity#onChunkUnloaded` method's call path was adjusted to account for the new removal reason, ensuring that the method is called only upon the chunk being actually unloaded, rather than both due to chunk unload and due to chunk packet synchronization.


src/main/java/net/minecraftforge/common/extensions/IForgeBlockEntity.java
```diff
@@ -189,4 +196,69 @@ default void requestModelDataUpdate()
...
+     * Marks this block entity as having been removed from the level for the given reason.
+     *
+     * @param reason the reason for the block entity's removal
+     */
+     void setRemoved(RemovalReason reason);
+
+    /**
+     * Reasons for the removal of a {@link BlockEntity} from the level.
+     */
+    enum RemovalReason
```
Technici4n, line 194: I'm not sure all these cases, and the `boolean chunk` parameter, are useful. Wouldn't REMOVED and UNLOADED be enough, possibly UNKNOWN too?

sciwhiz12: The various removal reasons were added in an effort to provide as much information to the modder as reasonably possible. We certainly could cut down the amount of values in this enum (though for two values, it'd be better then to use a `boolean`), but I felt at the time that providing the info would be of some potential use and wasn't too hard to implement at least.

Technici4n: More info means more cases to handle for the modder though. I think that most use cases would be covered by a simple boolean `wasUnloaded`, probably with a default of false.

pupnewfster: I am inclined to agree that at least adding a second boolean to `RemovalReason` for removed (and not coming back) vs just unloaded would make sense and be useful for devs (`REMOVED` and `REPLACED_IN_WORLD`, not sure about `UNKNOWN` and I haven't looked into how `CLEARED_CHUNK` is used)

sciwhiz12: @pupnewfster: That sounds like a reasonable idea to me, so modders can rely on the boolean instead of hardcoding the individual reasons (and possibly needing to add any further new reasons should they come up). This way, we retain the information provided by the individual `RemovalReason`s (which was my original reason for the enum) and provide a simple boolean check for modders to use (as @Technici4n notes).

LexManos:
Quick look over and discussion on discord, I think this needs more work and looking into the usecases.
I think {I don't know because I haven't seen enough actual uses} that this would fall more under the modders using the wrong method. It may be more appropriate to make these 'cleanup' things invoked from Block.onRemove. Which is the thing at the Block level that is responsible for cleaning up Block Entities when the blockstate changes in the chunk.

PR #8310, opened 2021-12-20
[1.18] Add block spreading api

JTK222:
Problem this PR solves:
Grass blocks will only spread grass to dirt blocks.
Blocks extending GrassBlock or SpreadableSnowyDirtBlock will do the same unless changed.
They'll also spread their own block type, ignoring the soil type.

Both of those aspects make it relatively hard to make grass spread between vanilla & modded soil types. 
(Same applies for mycelium)
This API extension adds a proper way to handle this behavior without much work & allows extension.

At the state of creating this PR, there are still 3 ToDos:
- [ ] JavaDoc in those new classes
- [x] Setting a spreader type for grass & mycelium blocks with a minimal patch
- [x] Creating a Test mod

Usage example:
```java
        SpreadBehavrios.addSpreaderBehavior(Blocks.DIRT, SpreaderType.GRASS, ((state, level, pos) ->
            Blocks.GRASS_BLOCK.defaultBlockState()
                .setValue(BlockStateProperties.SNOWY, level.getBlockState(pos.above()).is(Blocks.SNOW))));
```
The above example allows all grassy blocks to properly spread grass to dirt blocks.
All blocks in the example can be replaced with modded blocks too.


patches/minecraft/net/minecraft/world/level/block/SpreadingSnowyDirtBlock.java.patch
```diff
@@ -1,13 +1,25 @@
 --- a/net/minecraft/world/level/block/SpreadingSnowyDirtBlock.java
 +++ b/net/minecraft/world/level/block/SpreadingSnowyDirtBlock.java
-@@ -35,8 +_,10 @@
+@@ -10,7 +_,7 @@
+ import net.minecraft.world.level.block.state.BlockState;
+ import net.minecraft.world.level.lighting.LayerLightEngine;
+ 
+-public abstract class SpreadingSnowyDirtBlock extends SnowyDirtBlock {
++public abstract class SpreadingSnowyDirtBlock extends SnowyDirtBlock implements net.minecraftforge.common.extensions.IForgeSpreadingBlock{
```
sciwhiz12: Space before a bracket:
```suggestion
+public abstract class SpreadingSnowyDirtBlock extends SnowyDirtBlock implements net.minecraftforge.common.extensions.IForgeSpreadingBlock {
```

patches/minecraft/net/minecraft/world/level/block/SpreadingSnowyDirtBlock.java.patch
```diff
@@ -1,13 +1,25 @@
...
+       super(p_56817_);
+    }
+@@ -35,9 +_,12 @@
  
     public void m_7455_(BlockState p_56819_, ServerLevel p_56820_, BlockPos p_56821_, Random p_56822_) {
        if (!m_56823_(p_56819_, p_56820_, p_56821_)) {
+-         p_56820_.m_46597_(p_56821_, Blocks.f_50493_.m_49966_());
 +         if (!p_56820_.isAreaLoaded(p_56821_, 1)) return; // Forge: prevent loading unloaded chunks when checking neighbor's light and spreading
-          p_56820_.m_46597_(p_56821_, Blocks.f_50493_.m_49966_());
++         if(net.minecraftforge.common.block.spreader.SpreadBehaviors.canSpread(p_56819_, net.minecraftforge.common.block.spreader.SpreaderType.REVERT))
```
sciwhiz12: Space before an opening parenthesis for an `if`-expression:
```suggestion
+         if (net.minecraftforge.common.block.spreader.SpreadBehaviors.canSpread(p_56819_, net.minecraftforge.common.block.spreader.SpreaderType.REVERT))
```

patches/minecraft/net/minecraft/world/level/block/SpreadingSnowyDirtBlock.java.patch
```diff
@@ -1,13 +1,25 @@
...
++            p_56820_.m_46597_(p_56821_, net.minecraftforge.common.block.spreader.SpreadBehaviors.getSpreadState(p_56819_, p_56820_, p_56821_, net.minecraftforge.common.block.spreader.SpreaderType.REVERT));
        } else {
-+         if (!p_56820_.isAreaLoaded(p_56821_, 3)) return; // Forge: prevent loading unloaded chunks when checking neighbor's light and spreading
           if (p_56820_.m_46803_(p_56821_.m_7494_()) >= 9) {
++            this.spread(p_56819_, p_56820_, p_56821_, p_56822_, 4, 1); if(true) return;
```
sciwhiz12: For the sake of better readability, please move the `if (true) return;` to the next line, and add an explanatory `// Forge: ` comment just in case. Also, space before opening parenthesis for an `if`-expression.

src/main/java/net/minecraftforge/common/block/spreader/SpreadBehaviors.java
```diff
@@ -0,0 +1,109 @@
...
+
+import javax.annotation.Nullable;
+import java.util.HashMap;
+import java.util.Map;
+
+public class SpreadBehaviors
+{
+
+    private static final Logger LOGGER = LogManager.getLogger();
+    private static final Map<IRegistryDelegate<Block>, Map<SpreaderType, ISpreadingBehavior>> SPREADERS = new HashMap<>();
```
sciwhiz12: I feel like a [Guava `Table<IRegistryDelegate<Block>, SpreaderType, ISpreadingBehavior>`](https://github.com/google/guava/wiki/NewCollectionTypesExplained#table) would be cleaner here than a `Map` of `Map`s.

src/main/java/net/minecraftforge/common/block/spreader/SpreadBehaviors.java
```diff
@@ -0,0 +1,109 @@
...
+    {
+        ISpreadingBehavior behavior = getSpreadingBehavior(state.getBlock(), type);
+        if (behavior == null) return state;
+        return behavior.getSpreadingState(state, level, pos);
+    }
+
+    @Nullable
+    private static ISpreadingBehavior getSpreadingBehavior(Block block, SpreaderType type)
+    {
+        return SPREADERS.containsKey(block) ? SPREADERS.get(block).get(type) : null;
```
sciwhiz12: This should be querying using the `delegate`. Also, this may benefit from doing `SPREADERS.getOrDefault(block.delegate, Collections.emptyMap()).get(type)` to avoid the dual query cost (`containsKey` then `get` if the former is successful).

src/main/java/net/minecraftforge/common/block/spreader/SpreaderType.java
```diff
@@ -0,0 +1,51 @@
...
+{
+    private static final Map<String, SpreaderType> TYPES = new HashMap<>();
+
+    public static final SpreaderType GRASS = getType("grass");
+    public static final SpreaderType MYCELIUM = getType("mycelium");
+    public static final SpreaderType CRIMSON = getType("crimson");
+    public static final SpreaderType WARPED = getType("warped");
+    public static final SpreaderType REVERT = getType("revert");
+
+    public static SpreaderType getType(String name)
```
sciwhiz12: A nitpick, but this could simply be named `get`.

src/main/java/net/minecraftforge/common/extensions/IForgeSpreadingBlock.java
```diff
@@ -0,0 +1,45 @@
...
+import net.minecraft.world.level.block.state.BlockState;
+import net.minecraftforge.common.block.spreader.SpreadBehaviors;
+import net.minecraftforge.common.block.spreader.SpreaderType;
+
+import java.util.Random;
+
+public interface IForgeSpreadingBlock
+{
+
+
```
sciwhiz12: Remove the (double) spaces.

src/main/java/net/minecraftforge/common/extensions/IForgeSpreadingBlock.java
```diff
@@ -0,0 +1,45 @@
...
+    default void spread(BlockState state, ServerLevel level, BlockPos pos, Random random, int tries, int range){
+        if (!level.isAreaLoaded(pos, range + 1)) return;
+        range = (range * 2) + 1;
+        for(int i = 0; i < tries; ++i) {
+            BlockPos blockpos = pos.offset(random.nextInt(range) - 1, random.nextInt(5) - 3, random.nextInt(range) - 1);
+            BlockState targetState = level.getBlockState(blockpos);
+            if(SpreadBehaviors.canSpread(targetState, SpreaderType.GRASS)){
```
sciwhiz12: Spaces before opening parentheses of `if` and `for` expressions, and after the closing parenthesis of an `if` expression, and opening braces on newlines:
```suggestion
        for (int i = 0; i < tries; ++i)
        {
            BlockPos blockpos = pos.offset(random.nextInt(range) - 1, random.nextInt(5) - 3, random.nextInt(range) - 1);
            BlockState targetState = level.getBlockState(blockpos);
            if (SpreadBehaviors.canSpread(targetState, SpreaderType.GRASS))
            {
```

It may be beneficial to use a `MutableBlockPos` here instead (with the note that the position passed to `getSpreadState` later should be immutable).

src/main/java/net/minecraftforge/common/extensions/IForgeSpreadingBlock.java
```diff
@@ -0,0 +1,45 @@
...
+                level.setBlockAndUpdate(blockpos, SpreadBehaviors.getSpreadState(targetState, level, pos, SpreaderType.GRASS));
```
sciwhiz12: Is it intended that these are hardcoded for the `GRASS` spreader type?

src/main/java/net/minecraftforge/common/extensions/IForgeSpreadingBlock.java
```diff
@@ -0,0 +1,45 @@
...
+import net.minecraftforge.common.block.spreader.SpreadBehaviors;
+import net.minecraftforge.common.block.spreader.SpreaderType;
+
+import java.util.Random;
+
+public interface IForgeSpreadingBlock
+{
+
+
+    default void spread(BlockState state, ServerLevel level, BlockPos pos, Random random, int tries, int range){
```
sciwhiz12: Space after the closing parenthesis of a method definition, and opening brace should be on a newline:
```suggestion
    default void spread(BlockState state, ServerLevel level, BlockPos pos, Random random, int tries, int range)
    {
```

sciwhiz12 (changes requested):

Shadows-of-Fire:
@JTK222 Do you plan on continuing this PR?  The JavaDoc comments are still unfinished and they are part of your original to-do list.  You will also need to rebase since there are merge conflicts now.

PR #8313, opened 2021-12-21
Fix #8298 (MobBucketItem) and add test mod

### LINKED ISSUE #8298: Javascript coremod for MobBucketItem commented out
**Minecraft Version:** 1.18.1

**Forge Version:** 39.0.0

**Description of issue:** A small regression. In the 1.17 update the coremod that makes the supplier-ified MobBucketItem constructor useful was commented out. It is linked below. As a result trying to construct a custom bucket using the class is useless.

https://github.com/MinecraftForge/MinecraftForge/blob/456a48b12a61ae62bc86a36fe824245781cfab19/src/main/resources/coremods/field_to_method.js#L57-L65

eerussianguy:
This removes the broken coremod for MobBucketItem by just removing the old fields and redirecting everything to the supplier-ified constructor


src/test/java/net/minecraftforge/debug/item/CustomMobBucketTest.java
```diff
@@ -0,0 +1,34 @@
+package net.minecraftforge.debug.item;
```
sciwhiz12, line 20: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/test/java/net/minecraftforge/debug/item/CustomMobBucketTest.java
```diff
@@ -0,0 +1,34 @@
...
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.RegistryObject;
+
+@Mod(CustomMobBucketTest.MODID)
+public class CustomMobBucketTest
+{
+    public static final String MODID = "custom_mob_bucket_test";
+    public static final DeferredRegister<Item> ITEMS = DeferredRegister.create(ForgeRegistries.ITEMS, MODID);
+
+    public static final boolean ENABLE = true;
```
sciwhiz12: Nitpick: the conventional name for this field is `ENABLED`.

src/test/java/net/minecraftforge/debug/item/CustomMobBucketTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public static final RegistryObject<Item> COW_BUCKET = ITEMS.register("cow_bucket", () -> new MobBucketItem(() -> EntityType.COW, () -> Fluids.WATER, () -> SoundEvents.BUCKET_EMPTY_FISH, (new Item.Properties()).stacksTo(1).tab(CreativeModeTab.TAB_MISC)));
```
sciwhiz12: Just for readability purposes, could you split this into 2+ lines?

sciwhiz12 (changes requested):
Testing with the included test mod's mob bucket item (the cow bucket) and the linked issue's code (the pig bucket) shows both work correctly with this PR applied.

Aside from the review comments, please confirm who created the `cow_bucket.png` texture and what the license for the texture is (and if the copyright author grants Forge a license to use, modify, redistribute, etc. the texture under the LGPLv2.1 license as set out in the [CLA](https://github.com/MinecraftForge/MinecraftForge/blob/1.18.x/docs/CONTRIBUTING.md#contributor-license-agreement)).

eerussianguy:
I made that texture, I grant forge full rights to it as covered in the CLA.

sciwhiz12 (changes requested):
After some discussion of this PR in the `#github-discussions` channel, it's decided that the uses of the `type` and `emptySound` fields within the class should be manually patched to redirect to the methods instead of using the transformer (which should now be deleted), along with some other changes.
- The now-`@Deprecated` constructor for `MobBucketItem` should redirect to the supplier-using constructor instead.
- The original non-supplier fields should be deleted, so if this class were to change in the future (for example, other uses of the fields are added), this class will fail to compile during the update and patch cycle as to draw attention to it (rather than silently compiling but having broken logic).

patches/minecraft/net/minecraft/world/item/MobBucketItem.java.patch
```diff
@@ -1,43 +1,67 @@
...
-+   protected EntityType<?> getFishType() {
-+      return fishTypeSupplier.get();
++   protected EntityType<?> getFishType()
++   {
++      return entityTypeSupplier.get();
 +   }
 +
-+   protected SoundEvent getEmptySound() {
++   protected SoundEvent getEmptySound()
++   {
```
sciwhiz12: Because these are Minecraft classes, the code style to be followed is the same as the rest of the class (i.e. the formatting outputted by Fernflower): brackets on the same line.

patches/minecraft/net/minecraft/world/item/MobBucketItem.java.patch
```diff
@@ -1,43 +1,67 @@
...
+-      if (this.f_151134_ == EntityType.f_20489_) {
++      if (getFishType() == EntityType.f_20489_) {
+          CompoundTag compoundtag = p_151155_.m_41783_();
+          if (compoundtag != null && compoundtag.m_128425_("BucketVariantTag", 3)) {
+             int i = compoundtag.m_128451_("BucketVariantTag");
+@@ -80,6 +_,16 @@
              p_151157_.add(mutablecomponent);
           }
        }
 -
```
sciwhiz12: Since you're in this patch anyway, could you introduce a blank link here so this line removal disappears?

sciwhiz12 (changes requested):
Looks fine to me, just two minor quibbles remaining and then this is good to go!

sciwhiz12 (approved):

PR #8314, opened 2021-12-21
Add Environment Tick Event

eerussianguy:
This is a 1.18 port of #7235. I made a test mod with some functionality since the old test mod was just massive log spam.

I put the hook inside the `canTick` check because I can't think of a use case for intentionally attempting to tick chunks that aren't accessible, and it's a guard modders will have to add themselves each time otherwise.

To restate some applications for this event: modifying snowy weather so that the snow builds up in layers, as well as giving better control for mods that want to give blocks extra random ticks in certain conditions.

### DIFF
--- a/src/main/java/net/minecraftforge/event/TickEvent.java
+++ b/src/main/java/net/minecraftforge/event/TickEvent.java
@@ -22,13 +22,14 @@ package net.minecraftforge.event;
 import net.minecraft.world.entity.player.Player;
 import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.world.level.Level;
+import net.minecraft.world.level.chunk.LevelChunk;
 import net.minecraftforge.eventbus.api.Event;
 import net.minecraftforge.fml.LogicalSide;
 
 public class TickEvent extends Event
 {
     public enum Type {
-        WORLD, PLAYER, CLIENT, SERVER, RENDER;
+        WORLD, PLAYER, CLIENT, SERVER, RENDER, ENVIRONMENT;
     }
 
     public enum Phase {
@@ -66,6 +67,22 @@ public class TickEvent extends Event
             this.world = world;
         }
     }
+
+    public static class EnvironmentTickEvent extends TickEvent
+    {
+        public final Level level;
+        public final LevelChunk chunk;
+        public final int randomTickSpeed;
+
+        public EnvironmentTickEvent(LogicalSide side, Phase phase, Level level, LevelChunk chunk, int randomTickSpeed)
+        {
+            super(Type.ENVIRONMENT, side, phase);
+            this.level = level;
+            this.chunk = chunk;
+            this.randomTickSpeed = randomTickSpeed;
+        }
+    }
+
     public static class PlayerTickEvent extends TickEvent {
         public final Player player;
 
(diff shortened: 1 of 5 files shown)


src/main/java/net/minecraftforge/event/TickEvent.java
```diff
@@ -66,6 +67,25 @@ public WorldTickEvent(LogicalSide side, Phase phase, Level world)
             this.world = world;
         }
     }
+
+    /**
+     * Called before and after a chunk is ticked for things like random ticks and weather. Useful for random-tick and weather based actions.
+     */
+    public static class EnvironmentTickEvent extends TickEvent
+    {
+        public final Level level;
```
sciwhiz12: Is there a reason why this is a `Level` rather than a `ServerLevel`, since this event is fired from `ServerChunkCache`?

src/main/java/net/minecraftforge/event/TickEvent.java
```diff
@@ -66,6 +67,25 @@ public WorldTickEvent(LogicalSide side, Phase phase, Level world)
...
+        public final LevelChunk chunk;
+        public final int randomTickSpeed;
```
sciwhiz12: Please expose these as getters and mark these `private`. I've been meaning to refactor these tick events to encapsulate the fields, but I keep missing my windows to do them.

src/test/java/net/minecraftforge/debug/world/EnvironmentTickEventTest.java
```diff
@@ -0,0 +1,63 @@
...
+import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.event.TickEvent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
+
+@Mod(EnvironmentTickEventTest.MODID)
+public class EnvironmentTickEventTest
+{
+    public static final String MODID = "environment_tick_event_test";
+    public static final boolean ENABLED = true;
```
sciwhiz12: I'm not sure if we want this test to be enabled by default, since it seems to be particularly disruptive while testing if it snows (_it's snowing obsidian, hallelujah_).

src/main/java/net/minecraftforge/event/TickEvent.java
```diff
@@ -66,6 +67,25 @@ public WorldTickEvent(LogicalSide side, Phase phase, Level world)
...
+
+        public EnvironmentTickEvent(LogicalSide side, Phase phase, Level level, LevelChunk chunk, int randomTickSpeed)
+        {
+            super(Type.ENVIRONMENT, side, phase);
```
sciwhiz12: Will it ever be possible for this event to be fired on the logical client? If not, it may be good to signal this by hardcoding `LogicalSide.SERVER` (possibly alongside the change to `ServerLevel` as suggested above).

sciwhiz12 (changes requested):
Confirmed it works with the provided test mod. I do question though if this will have a measurable performance impact because this event will be fired for each loaded chunk (nearby players or forced chunks). Other than that and the review comments, this looks fine to me.

sciwhiz12 (approved):

eerussianguy:
Closing in favor of more performant solutions. People should not try to recreate this PR as is in the future.

PR #8316, opened 2021-12-22
[1.18.x] Event for adding biomes to multi-noise biome sources when dimensions load

Commoble:
## Background
* Minecraft 1.18 refactored the vanilla overworld biome source (the part of the chunk generator that determines what biomes are where) to use the MultiNoiseBiomeSource (like the Nether) instead of the legacy biome source, which has been removed. This means that we need a new hook to add biomes to vanilla biome sources, as otherwise the only way to add a biome to a dimension is for a datapack to override the dimension json.

### More on what Biome Sources are
* Dimensions have chunk generators, which have biome sources. Chunk generators and biome sources are both type-dispatched; mods can register custom chunk generator and biome source types, with associated logic and deserializers. This makes making an API to support *all* biome source types infeasible.
* There is no registry or datapack folder for biome sources; biome sources for datapack dimensions are defined inline in the dimension json. This also means that biome sources are constructed before their associated dimensions are, and cannot determine which dimension they are part of (because the dimension does not yet exist).
* Multi-noise biome sources define what biomes spawn where by declaring a list of biome parameter entries, each consisting of a biome and several noise parameters; each position in the world has a calculable value for each of these parameters, and the closest biome to that position's parameters is what biome generates there.
* Multi-noise biome sources have *two* deserializers:
  * A "preset" codec that refers to one of the vanilla dimensions (overworld, biome, or nether), effectively allowing a custom dimension to say "just give me whatever the vanilla dimension's biomes are". A MultiNoiseBiomeSource is constructed with a copy of the preset's parameter list, as well as a reference to the original preset (for datageneration purposes). We can retrieve the vanilla name from these when they are loaded.
  * A "fully-custom" codec wherein which all biome parameters are defined inline.
  * The primary codec that actually reads the biome source json uses a mapEither codec, first attempting to read a preset biome source, and then falling back to the fully-custom format if that fails.
* The three vanilla dimensions use preset-type multinoise sources.

## Requirements for a Hook for Adding Biomes to Vanilla Biome Sources
1. Any modifications to a biome source instance must be applied exactly once (adding a biome entry mustn't cause the parameter list to have two copies of that entry)
2. Mods must be able to determine which biome source they're modifying so they can decide whether to add a biome or not
3. If possible, modifications must apply before chunk generators load so that custom chunk generators that read data from the biome source are reading finalized data
4. If possible, access to json data must be available, so that mods can make data-driven modifications to biome sources if they desire

This PR meets the above requirements.

### Non-Goals
* It is not a goal of this PR to support adding biomes to non-vanilla biome source types; mods with custom biome source types are capable of providing their own APIs, and the type-dispatched nature of biome sources means that no single API can reasonably support adding biomes to all possible biome source types.
* It is not a goal of this PR to support adding biomes to the vanilla end biome source (which is its own unrelated thing); this can be done in a separate PR, if necessary and feasible.
* It is not a goal of this PR to add a data loader for biome source modifications (though this could conceivably done at a later time if necessary)
* It is not a goal of this PR to address the problem inherent to the multi-noise biome source whereby lots of mods adding biomes to the world causes the biomes to become increasingly smaller and patchy; this is a problem left to mods and datapacks to solve.

### Expanded MultiBiomeSource JSON Format
* This PR adds an optional `"forge:name"` string field to the non-preset multi-noise biome source json format, which will be read as a ResourceLocation or namespaced ID (e.g. `"minecraft:nether"`). If datapacks use this field, it is strongly suggested that the field have the same name as the dimension the biome source belongs to, though it is not required. An example use of this field can be found in the `named_test` dimension json included with this PR's test mod.

### MultiNoiseBiomeSourceLoadingEvent
* This event fires whenever a multi-noise biome source instance is deserialized (similar to the BiomeLoadingEvent, though dimensions load differently than biomes in several favorable ways).
* The event fires on the forge bus, and is cancellable. If cancelled, no modifications from any event listener will be made to the biome source.
* The event provides these methods:
  * getParameters()
    * Returns a mutable list of biome parameters. When the event is fired, this contains a copy of the original parameter list the biome source would have used. When the event concludes, the biome source will be constructed with this list (including all modifications by event listeners) instead of the original list.
  * getName()
    * Returns a nullable ResourceLocation representing either the name of the biome source preset (if the biome source is a preset biome source) or the "forge:name" field described earlier, if present (if the biome source is a non-preset biome source). Mods can use this to decide whether to add a biome to the biome source or not.
  * getRegistries()
    * Provides access to the servers dynamic/worldgen registries; event listeners can retrieve biomes and other worldgen datapack data from this.

An example use of this event is found in the test mod included with this PR.

### MultiNoiseBiomeSource Patch
* Adds a `"forge:name"` optional ResourceLocation field to the non-preset codec
* Adds hooks to the codecs to pass incoming data and registry access (by way of RegistryReadOps) to the forge hooks that fire the loading event
  * The hook for the fully-custom multinoise format is added to the fully-custom multinoise codec
  * The hook for loading the preset format is added to the primary (mapEither) codec instead, as this is where the final MultiNoiseBiomeSource instance would be constructed from the preset json's data.
  * The location and timing of these hooks allow modification of the biome entry data, with registry access, after the biome entries are read from json, but just before the final MultiNoiseBiomeSource is constructed, without causing any additional MultiNoiseBiomeSources to be constructed than otherwise would be.
* Adds a field for the biome source name and adds a constructor to accept this field
  * This field isn't used by the event or relevant hooks, but we preserve it here to support data generation

## Test Mod
The provided test mods gives an example of using RegistryWriteOps to datagenerate dimension jsons (with or without the "forge:name" biome source field) and using the event to add biomes to vanilla, preset json, and non-preset json dimensions. It also includes some logging so that requirement (1) could be verified.

## Other Concerns
* While verifying requirement (1), I noticed that when the Recreate World button was used to recreate a world, some dimensions seemed to be being modified more than they should have; however, when I observed these dimensions' biome sources at runtime, all biome sources had the correct number of biome entries, so I'm chalking that up to the singleplayer create world menu and experimental warning screen's general strangeness.


patches/minecraft/net/minecraft/world/level/biome/MultiNoiseBiomeSource.java.patch
```diff
@@ -0,0 +1,39 @@
...
++++ b/net/minecraft/world/level/biome/MultiNoiseBiomeSource.java
+@@ -33,9 +_,13 @@
+          return p_187078_.group(Climate.ParameterPoint.f_186862_.fieldOf("parameters").forGetter(Pair::getFirst), Biome.f_47431_.fieldOf("biome").forGetter(Pair::getSecond)).apply(p_187078_, Pair::of);
+       }).listOf()).xmap(Climate.ParameterList::new, (Function<Climate.ParameterList<Supplier<Biome>>, List<Pair<Climate.ParameterPoint, Supplier<Biome>>>>) Climate.ParameterList::m_186850_).fieldOf("biomes").forGetter((p_187080_) -> {
+          return p_187080_.f_48435_;
+-      })).apply(p_187070_, MultiNoiseBiomeSource::new);
++      }), ResourceLocation.f_135803_.optionalFieldOf("forge:name").forGetter(source -> {return source.dimensionName; // forge: add optional name field to MultiNoiseBiomeSource codec
++      }), net.minecraftforge.common.RegistryAccessCodec.INSTANCE.forGetter(x -> {return null; // forge: add registry access via registryreadops (doesn't need to be reserialized)
++      })).apply(p_187070_, (parameters, name, registryAccess) ->
++         net.minecraftforge.common.ForgeHooks.enhanceMultiNoiseBiomeSource(parameters, name.orElse(null), registryAccess));
```
alcatrazEscapee: Just a thought, most of this could be done without the very contrived line/bracket formatting and still stay consistent with the same patched line(s) from vanilla. i.e.
```java
}), ResourceLocation.f_135803_.optionalFieldOf("forge:name").forGetter(source -> {return source.dimensionName; // forge: add optional name field to MultiNoiseBiomeSource codec
}), net.minecraftforge.common.RegistryAccessCodec.INSTANCE.forGetter(x -> {return null; // forge: add registry access via registryreadops (doesn't need to be reserialized)
})).apply(p_187070_, (parameters, name, registryAccess) ->
net.minecraftforge.common.ForgeHooks.enhanceMultiNoiseBiomeSource(parameters, name.orElse(null), registryAccess));
```
is just kind of unnecessarily convoluted, as opposed to
```java
}),
    ResourceLocation.f_135803_.optionalFieldOf("forge:name").forGetter(source -> source.dimensionName), // forge: add optional name field to MultiNoiseBiomeSource codec
    net.minecraftforge.common.RegistryAccessCodec.INSTANCE.forGetter(x -> null) // forge: add registry access via registryreadops (doesn't need to be reserialized)
).apply(p_187070_, (parameters, name, registryAccess) -> net.minecraftforge.common.ForgeHooks.enhanceMultiNoiseBiomeSource(parameters, name.orElse(null), registryAccess));
```

alcatrazEscapee (approved):
Looks well done, there's maybe a bit of patch cleanup but otherwise from what I've heard and seen about this it looks good.

patches/minecraft/net/minecraft/world/level/biome/MultiNoiseBiomeSource.java.patch
Commoble: Made these changes in latest push

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1241,4 +1251,37 @@ public static void onEntityEnterSection(Entity entity, long packedOldPos, long p
...
+        ParameterList<Supplier<Biome>> parameters = presetInstance.biomeSource().parameters;
+        List<Pair<ParameterPoint, Supplier<Biome>>> parameterBuilder = new ArrayList<>(parameters.values());
+        ResourceLocation name = presetInstance.preset().name;
+        MultiNoiseBiomeSourceLoadingEvent event = new MultiNoiseBiomeSourceLoadingEvent(parameterBuilder, name, registryAccess);
+        MinecraftForge.EVENT_BUS.post(event);
+        ParameterList<Supplier<Biome>> actualParameters = event.isCanceled() ? parameters : new ParameterList<>(parameterBuilder);
+        return new MultiNoiseBiomeSource(actualParameters, Optional.of(presetInstance), Optional.empty());
+    }
+
+    /** Forge internal use, called via the {@link MultiNoiseBiomeSource.DIRECT_CODEC} used for non-preset sources **/
```
sciwhiz12: While I'm not sure whether these methods need to be marked `@Deprecated` because of their residence in `ForgeHooks`, what I can say is that this javadoc does not follow the othe two in having the `@deprecated` block tag.

src/main/java/net/minecraftforge/event/world/MultiNoiseBiomeSourceLoadingEvent.java
```diff
@@ -0,0 +1,97 @@
...
+ * Event for modifying multi-noise biome sources.
+ * This event is fired whenever a MultiNoiseBiomeSource is deserialized during dimension loading.
+ * This occurs after datapacks and worldgen registries load, but before chunk generators are constructed
+ * (and before the biome source itself is constructed).
+ * Standard tag wrappers are not resolvable at this time.
+ * 
+ * Best practice for use of this event:
+ * Additions during {@link EventPriority#HIGH}
+ * Removals during {@link EventPriority#NORMAL}
+ * Modifications during {@link EventPriority#LOW}
```
sciwhiz12: This should probably be an unordered list (`<ul>` and `<li>`).

src/main/java/net/minecraftforge/event/world/MultiNoiseBiomeSourceLoadingEvent.java
```diff
@@ -0,0 +1,97 @@
...
+ * 
+ * Cancelling this event will prevent the biome source from being modified.
+ * 
+ * This event is fired on the {@link MinecraftForge.EVENT_BUS}.
```
sciwhiz12: The paragraphs should be separated by `<p>` tags.

src/main/java/net/minecraftforge/common/RegistryAccessCodec.java
```diff
@@ -0,0 +1,60 @@
...
+import com.mojang.serialization.MapLike;
+import com.mojang.serialization.RecordBuilder;
+
+import net.minecraft.core.RegistryAccess;
+import net.minecraft.resources.RegistryReadOps;
+
+/**
+ * Unit codec for getting access to dynamic registries during dimension loading.
+ * This is only safe to use while loading dimensions (LevelStems, ChunkGenerators, and BiomeSources);
+ * when access to registries is needed while loading dynamic registry objects such as configured features,
```
sciwhiz12: For better readability, split the sentence into two:
```suggestion
 * This is only safe to use while loading dimensions (LevelStems, ChunkGenerators, and BiomeSources).
 * When access to registries is needed while loading dynamic registry objects such as configured features,
```
Also consider `{@link}`ing the reference to `RegistryAccessCodec`.

src/main/java/net/minecraftforge/common/RegistryAccessCodec.java
```diff
@@ -0,0 +1,60 @@
...
+ * then RegistryLookupCodec should be used instead.
+ */
+public final class RegistryAccessCodec extends MapCodec<RegistryAccess>
+{
+    public static final RegistryAccessCodec INSTANCE = new RegistryAccessCodec();
```
sciwhiz12, line 41: If this is meant to be a singleton instance, then the constructor for the class should be created and set to `private`.

src/main/java/net/minecraftforge/common/RegistryAccessCodec.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Override
+    public <T> DataResult<RegistryAccess> decode(DynamicOps<T> ops, MapLike<T> input)
+    {
+        return ops instanceof RegistryReadOps ? DataResult.success(((RegistryReadOps)ops).registryAccess) : DataResult.error("Not a registry ops");
```
sciwhiz12: Nitpick: space between variable and cast type in cast expression:
```suggestion
        return ops instanceof RegistryReadOps ? DataResult.success(((RegistryReadOps) ops).registryAccess) : DataResult.error("Not a registry ops");
```

src/main/java/net/minecraftforge/event/world/MultiNoiseBiomeSourceLoadingEvent.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Retrieves the name of this biome source.
+     * For vanilla biome sources and preset-based biome sources, this will be the name of the preset (typically minecraft:overworld/nether/end).
+     * Non-preset biome sources may specify the preset ID in the biome source json object with a "name" field.
+     * When present, this should generally match the ID of the dimension this biome source belongs to.
+     * If this is not present, this indicates that a datapack has defined this biome source and has not specified a name.
+     * @return Biome source name, if present.
+     */
+    public @Nullable ResourceLocation getName()
```
sciwhiz12: Current convention is that the nullability annotation for the method return type is specified on the line before the method:
```suggestion
    @Nullable
    public ResourceLocation getName()
```

src/main/java/net/minecraftforge/event/world/MultiNoiseBiomeSourceLoadingEvent.java
```diff
@@ -0,0 +1,97 @@
...
+ * 
+ * Cancelling this event will prevent the biome source from being modified.
+ * 
+ * This event is fired on the {@link MinecraftForge.EVENT_BUS}.
+ */
+@Cancelable
+public class MultiNoiseBiomeSourceLoadingEvent extends Event
+{
+    private final List<Pair<ParameterPoint, Supplier<Biome>>> parameters;
+    private final @Nullable ResourceLocation name;
```
sciwhiz12: Current convention is that the nullability annotation for a field is specified on the line before:
```suggestion
    @Nullable
    private final ResourceLocation name;
```

src/test/java/net/minecraftforge/debug/world/MultiNoiseBiomeSourceLoadingEventTest.java
```diff
@@ -0,0 +1,259 @@
...
+                ), () -> end));
+        }
+    }
+    
+    private void onRegisterCommands(RegisterCommandsEvent event)
+    {
+        event.getDispatcher().register(
+            Commands.literal(MODID)
+                .then(Commands.literal("biomes")
+                    .executes(context ->{
```
sciwhiz12: Nitpick: missing space between lambda arrow and opening brace:
```suggestion
                    .executes(context -> {
```
Alternatively, this could be a method reference to a static method to avoid this level of indentation.

src/test/java/net/minecraftforge/debug/world/MultiNoiseBiomeSourceLoadingEventTest.java
```diff
@@ -0,0 +1,259 @@
...
+                dimensionWriter.accept(NAMELESS_TEST, namelessDimension);
+            }
+
+            @Override
+            public String getName()
+            {
+                return MODID + " worldgen data provider";
+            }
+        };
+        generator.addProvider(provider);
```
sciwhiz12, line 191: This should be gated behind a `GatherDataEvent#includeServer()` condition.

src/test/java/net/minecraftforge/debug/world/MultiNoiseBiomeSourceLoadingEventTest.java
```diff
@@ -0,0 +1,259 @@
...
+        Registry<DimensionType> dimensionTypes = registries.registryOrThrow(Registry.DIMENSION_TYPE_REGISTRY);
+        Registry<NoiseGeneratorSettings> noiseSettings = registries.registryOrThrow(Registry.NOISE_GENERATOR_SETTINGS_REGISTRY);
+        Registry<NoiseParameters> noiseParameters = registries.registryOrThrow(Registry.NOISE_REGISTRY);
+        Registry<Biome> biomes = registries.registryOrThrow(Registry.BIOME_REGISTRY);
+        
+        DimensionType overworldDimensionType = dimensionTypes.get(DimensionType.OVERWORLD_LOCATION);
+        NoiseGeneratorSettings overworldNoiseSettings = noiseSettings.get(NoiseGeneratorSettings.OVERWORLD);
+        Biome desert = biomes.get(Biomes.DESERT);
+        
+        // make two three dimensions for testing
```
sciwhiz12: Seems like a small mistake while writing:
```suggestion
        // make three dimensions for testing
```

src/test/java/net/minecraftforge/debug/world/MultiNoiseBiomeSourceLoadingEventTest.java
```diff
@@ -0,0 +1,259 @@
...
+        parameterPoints.add(Pair.of(new ParameterPoint(
+            Parameter.span(0.2F, 0.55F), // temperature
+            Parameter.span(0.3F, 1.0F), // humidity
+            Parameter.span(0.3F, 1.0F), // continentalness
+            Parameter.span(-0.375F, -0.2225F), // erosion
+            Parameter.point(1.0F), // depth
+            Parameter.span(0.9333F, 1.0F), // weirdness
+            0L // offset
+            ), () -> desert));
+        ParameterList<Supplier<Biome>> biomeParams = new ParameterList<>(parameterPoints);
```
sciwhiz12: For readability, this should be moved to above the comment, since this is used by both named and nameless biome sources.

src/test/java/net/minecraftforge/debug/world/MultiNoiseBiomeSourceLoadingEventTest.java
```diff
@@ -0,0 +1,259 @@
...
+        
+        // expected behaviour is that overworld and overworld preset json dimension have an extra end biome,
+        // nether has an extra jungle biome,
+        // and the named test dimension has an extra nether biome
+        if (Objects.equals(event.getName(), Preset.OVERWORLD.name))
+        {
+            Biome end = event.getRegistries().registryOrThrow(Registry.BIOME_REGISTRY).get(Biomes.SMALL_END_ISLANDS);
+            var parameters = event.getParameters();
+            LOGGER.info("Adding end islands biome to overworld biome source with {} biomes", parameters.size());
+            parameters.add(Pair.of(new ParameterPoint(
```
sciwhiz12: I wonder if it would be worth it to add a `addParameter(ParameterPoint, Supplier<Biome>)` method to the event, for modder convenience.

sciwhiz12 (changes requested):
This looks solid to me. Testing using the provided test mod shows that the features work as expected; though I was unable to verify physically the presence of the small end islands biome in the overworld biome sources, the command shows that the biome parameter is present so I'll take its word.

src/main/java/net/minecraftforge/common/ForgeHooks.java
Commoble: Added in next push

src/main/java/net/minecraftforge/event/world/MultiNoiseBiomeSourceLoadingEvent.java
Commoble: Added in next push

src/main/java/net/minecraftforge/event/world/MultiNoiseBiomeSourceLoadingEvent.java
Commoble: Added in next push

src/main/java/net/minecraftforge/common/RegistryAccessCodec.java
Commoble: Made these changes in next push

src/main/java/net/minecraftforge/common/RegistryAccessCodec.java
Commoble: Added constructor in next push

src/main/java/net/minecraftforge/common/RegistryAccessCodec.java
Commoble: Fixed in next push

src/main/java/net/minecraftforge/event/world/MultiNoiseBiomeSourceLoadingEvent.java
Commoble: Fixed in next push

src/main/java/net/minecraftforge/event/world/MultiNoiseBiomeSourceLoadingEvent.java
Commoble: Fixed in next push

src/test/java/net/minecraftforge/debug/world/MultiNoiseBiomeSourceLoadingEventTest.java
Commoble: Moved into a static method in next push

src/test/java/net/minecraftforge/debug/world/MultiNoiseBiomeSourceLoadingEventTest.java
Commoble: Refactored in next push

src/test/java/net/minecraftforge/debug/world/MultiNoiseBiomeSourceLoadingEventTest.java
Commoble: Fixed in next push

src/test/java/net/minecraftforge/debug/world/MultiNoiseBiomeSourceLoadingEventTest.java
Commoble: Rearranged in next push

src/test/java/net/minecraftforge/debug/world/MultiNoiseBiomeSourceLoadingEventTest.java
Commoble: Fair enough, added in next push

50ap5ud5:
Nice work, this is a real need for this sort of API with the new world generation changes. Excited to see this merged in the future!

alcatrazEscapee (approved):

src/test/java/net/minecraftforge/debug/world/MultiNoiseBiomeSourceLoadingEventTest.java
```diff
@@ -0,0 +1,274 @@
...
+        // one with a preset biome source, one with a fully custom biome source and a dimension name, one with a fully custom source without name
+        
+        // preset biome source dimension
+        BiomeSource presetSource = MultiNoiseBiomeSource.Preset.OVERWORLD.biomeSource(biomes);
+        ChunkGenerator presetGenerator = new NoiseBasedChunkGenerator(noiseParameters, presetSource, 0, () -> overworldNoiseSettings);
+        LevelStem presetDimension = new LevelStem(() -> overworldDimensionType, presetGenerator);
+        
+        // some sample points used for the latter dimensions
+        List<Pair<ParameterPoint, Supplier<Biome>>> parameterPoints = new ArrayList<>();
+        parameterPoints.add(Pair.of(new ParameterPoint(
```
sciwhiz12, line 134: A minor quibble: this (and the `onLoadBiomeSource` listener below) should use the newly-added `addParameter` convenience.

sciwhiz12 (changes requested):
After this quibble I have, this is ready for assignment.

sciwhiz12 (approved):

Goldorion:
Does something need to be fixed before LexManos reviews this PR? There is no activity for 16 days, so I'm wondering what's the cause.

TheCurle:
We're all extremely busy. 
I want to test this PR to our usual internal standards when life permits, but we're all struggling between IRL things right now.
As it stands, it isn't up to Lex right now.

Mysterious-Dev (approved):

sciwhiz12 (changes requested):
- `src/main/java/net/minecraftforge/common/RegistryAccessCodec.java`
- `src/main/java/net/minecraftforge/event/world/MultiNoiseBiomeSourceLoadingEvent.java`
- `src/test/java/net/minecraftforge/debug/world/MultiNoiseBiomeSourceLoadingEventTest.java`

Defeatomizer:
@Commoble

mcenderdragon:
In 1.18.2 Biomes no longer define which Blocks they use but instead the SurfaceSystem does this, which loads SurfaceRules from the noise_settings/overworld.jsons (or nether.json you get the point) I think this PR should adress this aswell, or the biomes will not have custom blocks.

Commoble:
> In 1.18.2 Biomes no longer define which Blocks they use but instead the SurfaceSystem does this, which loads SurfaceRules from the noise_settings/overworld.jsons (or nether.json you get the point) I think this PR should adress this aswell, or the biomes will not have custom blocks.

I agree that something does need to be done there but I don't really have any good solutions for that (surface rules are a horrendous glob of json) and someone who does have a good solution for that would be able to address that in a separate PR. This PR just addresses the ability to get mods' biomes to exist in vanilla dimensions in the first place.

mcenderdragon:
> > In 1.18.2 Biomes no longer define which Blocks they use but instead the SurfaceSystem does this, which loads SurfaceRules from the noise_settings/overworld.jsons (or nether.json you get the point) I think this PR should adress this aswell, or the biomes will not have custom blocks.
> 
> I agree that something does need to be done there but I don't really have any good solutions for that (surface rules are a horrendous glob of json) and someone who does have a good solution for that would be able to address that in a separate PR. This PR just addresses the ability to get mods' biomes to exist in vanilla dimensions in the first place.

yeah agree with that. those needa very solit solution.

KlemenDEV:
Would be nice to have an API for this in the Forge itself

LexManos:
TerraBlender is the common Biome/Terrain API. Seems to work for mos things people want.
Haven't looked a this one in ages, but last attempt to do this that I recall was just making everything mutable and shoving it into an event. Which isn't a great approach both performance and compatibility wise.
I'll keep this in mind when the new toolchain is done and I tart modularizing things.

PR #7992, opened 2021-08-08
[Overhaul][1.17.1][Fluids] - Fluid Overhaul PR

Lanse505:
This PR aims to fix the fundamentally hardcoded nature of Fluids and Fluid Behaviour in 1.16 onwards.

In this PR we've greatly expanded upon the feature set of Custom Fluids to make their behaviour far more configurable as well as expanding upon their control over expected fluid behaviours such as hydrating certain blocks, allowing for more granular control over drowning, extinguishing, etc.

Due to the nature of Fluidlogging in minecraft a lot of what would be considered "block" tied behaviour and not direct "fluid-tied" behaviour needs to be part of the fluid, to allow for fluidloggable blocks that would support certain fluids to actually be able to call and use the behaviour of those fluids. 

With this in mind, in the future ``FluidAttributes`` will have to play a far bigger role in terms of defining certain behaviours of fluids. With this, we've introduced a whole new line of methods and fields to FluidAttributes for people to work with:

The following fields has been added:

Behaviour Relevant Fields:
```  
  - canSwim // Used to dictate whether a player can enter the "Swimming" pose and thus "sprinting" state while inside of the fluid, and thus giving them a speed-boost.
  
  - canDrown // Used to dictate whether an LivingEntity can drown while submerged in the Fluid.
  
  - canExtinguish // Used to dictate whether an entity will be extinguished if touching the fluid and is on fire.
  
  - canHydrate // Used to dictate if this block can provide "hydration" to various blocks. With this ConcretePowderBlock, CoralBlock, FarmBlock, SpongeBlock, and SugarCaneBlock have been patched to support this.
  
  - canBoat // Used to dictate whether a boat can both be placed in the Fluid as well as function within normal behaviour similar to water.
```

Fluid Movement-related Fields:
```  
  - horizontalMotionModifier // Used to dictate the horizontal movement modifier when moving inside of the Fluid.
  
  - enchantmentModifier // Used to dictate an additional horizontal movement modifier. By default this also requires you to override IForgeFluid#handleMotion since the default code is a reimplementation of Water fluid physics which only passes in the "Depth Strider" enchantment for checking.
  
  - effectModifier // Used to dictate a final modifier upon horizontal movement. By default this also requires you to override IForgeFluid#handleMotion since the default code is a reimplementation of Water fluid physics which only passes in the "Dolphin's Grace" effect for checking.
  ```
  
  Misc Fields:
  ```
    - motionScale // Used to scale the motion being applied when a flowing fluid is attempting to push you.
[...]

PR #8318, opened 2021-12-22
Make tryEmptyContainer respect doDrain

josephtarbit:
`FluidUtil::tryEmptyContainer` doesn't respect the `doDrain` flag, it fills the specified `IFluidHandler` regardless. 

It broke in this commit: https://github.com/MinecraftForge/MinecraftForge/commit/b294f4d8944d342611e95616152b128e5a03d6b5#diff-a6aa3c7740f816ab284f190e9121980e804de26f228e10af0ef9c4172fb0ac53

### DIFF
--- a/src/main/java/net/minecraftforge/fluids/FluidUtil.java
+++ b/src/main/java/net/minecraftforge/fluids/FluidUtil.java
@@ -190,7 +190,7 @@ public class FluidUtil
                 .map(containerFluidHandler -> {
 
                     // We are acting on a COPY of the stack, so performing changes is acceptable even if we are simulating.
-                    FluidStack transfer = tryFluidTransfer(fluidDestination, containerFluidHandler, maxAmount, true);
+                    FluidStack transfer = tryFluidTransfer(fluidDestination, containerFluidHandler, maxAmount, doDrain);
                     if (transfer.isEmpty())
                         return FluidActionResult.FAILURE;
 


pupnewfster:
See https://github.com/MinecraftForge/MinecraftForge/pull/6243#issuecomment-545140968 from previous PRs attempting to do this

josephtarbit:
> See [#6243 (comment)](https://github.com/MinecraftForge/MinecraftForge/pull/6243#issuecomment-545140968) from previous PRs attempting to do this

Wouldn't that be a bug with `tryFluidTransfer` though? This method is just delegating to that method and retrieving the result.

gigaherz:
It's a combination of a design flaw in the API (can't get the simulated resulting item) + the way tryFluidTransfer/tryEmptyContainer work. None of  them are bugged per se, but they don't work together as it is now.
If you want to fix this right you need to make tryEmptyContainer **not** use tryFluidTransfer, and instead do a non-simulating extraction in a copy of the stack without doing any insertion in order to compute the resulting item.
This WILL slow down the logic, which is why I have been saying since early 1.15 that the api needs to be changed to not have a stateful getContainer.

josephtarbit:
Perhaps `IFluidHandlerItem` could declare an additional fill/drain method which returns a result with a new `ItemStack`?

josephtarbit:
Or you could introduce a generic result object in `IFluidHandler` which allows you to return additional metadata from the fill/drain methods and then something like `IFluidHandlerItem` would have a result type of `ItemStack` (specified by a type parameter on `IFluidHandler`). It would look something like:

```java
interface IFluidHandler<T> {

    Pair<Integer, T> fill(FluidStack resource, FluidAction action);

    Pair<FluidStack, T> drain(FluidStack resource, FluidAction action);
}
```
(`Pair` could become a new class for each result type e.g. `DrainResult` or `FillResult`)

gigaherz:
Something like that was exactly what I suggested, except it would be only in the IFluidHandlerItem, and it would require removing the extends IFluidHandler.
But I believe there were other proposals. You'd have to sort through the conversations on Discord. 
What I can't remember is if #7992 deals with this or not (currently away from my PC).

TheCurle:
I have no idea how this slipped past us.

TheCurle:
I also have no idea why i put 1.16 in that commit message... 🤦

PR #8322, opened 2021-12-24
[1.18.2] Add an event when a living entity gets a projectile item stack to be fired

Jonathing:
Successor to #7715 (1.16 PR) and #8061 (1.17 PR). The commit from the 1.17 PR was cherry-picked and used as a base.

This PR adds an event that is fired every time a living entity gets a projectile (using the getProjectile() event) to be used by a ProjectileWeaponEvent. This usually includes skeletons firing arrows, players using bows/crossbows, etc. etc.

The purpose of this event is to allow modders to modify which ItemStack is used when a player or entity fires a ranged weapon (usually a bow). This can include things like using a different itemstack from the inventory, a completely separate item not from the inventory, or an entirely new item stack with its own properties. There's a test mod that includes an example of how if a player tries to shoot an arrow, a new item stack of 1 spectral arrow is used instead.


src/test/java/net/minecraftforge/debug/entity/living/LivingGetProjectileEventTest.java
```diff
@@ -0,0 +1,39 @@
...
+import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.event.entity.living.LivingGetProjectileEvent;
+import net.minecraftforge.fml.common.Mod;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+@Mod("living_get_projectile_event_test")
+public class LivingGetProjectileEventTest
+{
+    public static final boolean ENABLE = true;
```
sciwhiz12: Nitpick: conventional name for this field is `ENABLED`.

src/main/java/net/minecraftforge/event/entity/living/LivingGetProjectileEvent.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public LivingGetProjectileEvent(LivingEntity livingEntity, ItemStack projectileWeaponItemStack, ItemStack ammo)
+    {
+        super(livingEntity);
+        this.projectileWeaponItemStack = projectileWeaponItemStack;
+        this.projectileItemStack = ammo;
+    }
+
+    /**
+     * Gets the itemstack of the projectile weapon item that is looking for a projectile.
```
sciwhiz12: Use the `@return` javadoc inline tag:
```suggestion
     * {@return the itemstack of the projectile weapon item that is looking for a projectile}
```

src/main/java/net/minecraftforge/event/entity/living/LivingGetProjectileEvent.java
```diff
@@ -0,0 +1,84 @@
...
+     */
+    public ItemStack getProjectileWeaponItemStack()
+    {
+        return this.projectileWeaponItemStack;
+    }
+
+    /**
+     * Gets the itemstack of the projectile found. Initially this is set to the projectile found by vanilla behaviour.
```
sciwhiz12: Use the `@return` javadoc inline tag:
```suggestion
     * {@return the itemstack of the projectile found} Initially, this is set to the projectile found by vanilla behavior.
```

src/test/java/net/minecraftforge/debug/entity/living/LivingGetProjectileEventTest.java
```diff
@@ -0,0 +1,39 @@
...
+        if (event.getEntityLiving() instanceof Player && event.getProjectileItemStack().getItem().equals(Items.ARROW))
+        {
+            event.setProjectileItemStack(new ItemStack(Items.SPECTRAL_ARROW));
+        }
+    }
+
+    public void onLivingGetProjectile(LivingGetProjectileEvent event)
+    {
+        LOGGER.info("{} about to fire {} with a {}", event.getEntity(), event.getProjectileItemStack(), event.getProjectileWeaponItemStack());
+    }
```
sciwhiz12: You can combine the two listeners into one: log each time the event fires, and if the entity is a player and the ammunition is an arrow, set the ammunition to be used to a spectral arrow.

src/test/java/net/minecraftforge/debug/entity/living/LivingGetProjectileEventTest.java
```diff
@@ -0,0 +1,39 @@
...
+        if (ENABLE)
+        {
+            MinecraftForge.EVENT_BUS.addListener(this::onPlayerGetProjectile);
+            MinecraftForge.EVENT_BUS.addListener(this::onLivingGetProjectile);
+        }
+    }
+
+    public void onPlayerGetProjectile(LivingGetProjectileEvent event)
+    {
+        if (event.getEntityLiving() instanceof Player && event.getProjectileItemStack().getItem().equals(Items.ARROW))
```
sciwhiz12: You can use `==` on `Item` instances instead of `.equals()`.

sciwhiz12 (changes requested):
The event looks solid to me. Just needs a bit of changes in javadocs and the test mod, and it looks :shipit: to me.

Jonathing:
I made a small change to the test mod. Nothing major, I just added a condition where if there is an itemstack of spectral arrow in the offhand, the event will use that itemstack. It's just a small example showing how specific itemstacks can be used with the event.

sciwhiz12 (approved):

src/main/java/net/minecraftforge/event/entity/living/LivingGetProjectileEvent.java
```diff
@@ -0,0 +1,81 @@
...
+        this.projectileWeaponItemStack = projectileWeaponItemStack;
+        this.projectileItemStack = ammo;
+    }
+
+    /**
+     * @return The itemstack of the projectile weapon item that is looking for a projectile.
+     */
+    public ItemStack getProjectileWeaponItemStack()
+    {
+        return this.projectileWeaponItemStack;
```
sciwhiz12, line 42: As a follow-up, should this event guarantee that the itemstack provided by this method contains an item which is a subclass of `ProjectileWeaponItem`?

Jonathing: I thought about that, but then thought of the possibility that a modder might make an item that doesn't extend ProjectileWeaponItem that uses `getProjectile()`. Do you want me to make that change anyway?

Jonathing:
Merged in the 1.18.1 changes to fix a mods.toml conflict with the test mod.

src/main/java/net/minecraftforge/event/entity/living/LivingGetProjectileEvent.java
sciwhiz12: I suggest then that the method's javadocs clarify that the returned itemstack _may_ contain an item which is not an instance of `ProjectileWeaponItem`, as the method name suggests it might, to avoid possible confusion.

sciwhiz12 (changes requested):
- `src/main/java/net/minecraftforge/event/entity/living/LivingGetProjectileEvent.java`
- `src/test/java/net/minecraftforge/debug/entity/living/LivingGetProjectileEventTest.java`

Jonathing:
Will do as soon as I am able. Thanks!

sciwhiz12 (approved):

marchermans (approved):

TheCurle:
A nice solution to fixing the shortcomings of our current events;

ArrowNock event only provides the BowItem stack that was used to nock the bow, and it requires a very round-about method of overriding and instantiating to change the arrow that is nocked, and it can easily create some incompatibilities.
ArrowLoose event only provides a boolean that indicates whether an arrow exists, and cannot be used in this instance.

This projectile search event allows for the arrow to be trivially swapped out with another once it is about to be fired, with a single line event handler.

The patch surface is relatively high, but I believe it to be worth it (and, well, it's easy to see what those lines are meant to do.)

PR #8323, opened 2021-12-24
[1.18.x] move OnDatapackSyncEvent after tags and recipes are synced

Minecraftschurli:


pupnewfster:
I am 99% sure the entire point of the event was to be before tags and recipes are sent see #7912

MelanX:
Alternative: Pre and Post sub-event. Like a few other events.

KnightMiner:
Yeah, pleaSe absolutely do not move it after. My mod relies on that context syncing first. Explain your usecase for moving it after.

Minecraftschurli:
Closing since it was intended this way

PR #8325, opened 2021-12-25
[1.18.x] Added a hook to activate spyglass-style scoping

Da-Technomancer:
This PR is a previous PR (https://github.com/MinecraftForge/MinecraftForge/pull/8171), rebased onto 1.18.

Added a hook to trigger the 'scoping' effect the spyglass uses (zoom in, visual overlay, adjusted movement).

Mods can register a predicate which will be tested on the client side to see whether the scoping effect is being externally triggered. If any registered predicate returns true, scoping will be active.

Also wrote a test mod.


Shadows-of-Fire:
Looks fine to me, would go ahead and assign but the CI is failing and I can't check if it compiles manually right now.

Da-Technomancer:
CI is saying it passes now. I didn't make any changes.

Shadows-of-Fire:
Yeah it just had to be restarted, figured that was the case.

src/main/java/net/minecraftforge/common/SpyglassHooks.java
```diff
@@ -0,0 +1,57 @@
...
+
+package net.minecraftforge.common;
+
+import net.minecraft.world.entity.player.Player;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Predicate;
+
+public class SpyglassHooks
```
diesieben07: This doesn't need to be its own class, you can use `ForgeHooks`.

src/main/java/net/minecraftforge/common/SpyglassHooks.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Registers a predicate which will be checked for whether the player is considered to be 'scoping' with a spyglass.
+     * The predicate will only be tested on the client.
+     * Only one registered predicate needs to be true for scoping to occur.
+     * <strong>The passed predicate will be tested very frequently (typically every frame).</strong>
+     *
+     * @param spyglassCondition A predicate which will be checked on the client side.
+     */
+    public static synchronized void registerSpyglassCondition(Predicate<Player> spyglassCondition)
```
diesieben07: This is not truly thread safe.
Please instead fire an event at startup to collect all conditions and then store them in an immutable and therefor threadsafe list.

diesieben07 (changes requested):
Please update the license headers to 2022 as well and resolve merge conflicts.

src/main/java/net/minecraftforge/common/SpyglassHooks.java
Da-Technomancer: Could you clarify why you consider it thread-unsafe? I must be missing something, but I don't see it.

diesieben07: `SPYGLASS_PREDICATES` is accessed from two methods. One of them is synchronized, one of them is not.
However that is irrelevant here, the more important change request is to add a dedicated collection event.

diesieben07 (changes requested):
I don't think these should be a full blown registry. Why does it need to be?

Da-Technomancer:
> I don't think these should be a full blown registry. Why does it need to be?

It had seemed cleaner to use the tried-and-tested registry model than to create a new system for registration at mod startup. I'm open to suggestions.

diesieben07:
Fire an event. That event has an internal list of predicates and a method `addPredicate`. After the event is fired, bake the list into an immutable one that is then stored and used later. This avoids threading issues and the need for synchronization, it improves performance, etc.

Da-Technomancer:
After several revisions, I'm calling it quits.
I was not able to find a way to fire an event only as needed which was both robust and cleanly implemented; I'm sure a way exists, but someone more familiar with the inner workings of Forge would need to find it. 

As a workaround, for anyone who found this PR because they wanted to add their version of a spyglass feature, here's how I suggest doing it without relying on this PR. It's only a rough approximation of what this PR is meant to achieve.
- Achieve the 'zoom' effect using an FOV modifier. There is an existing forge hook for FOV; see 'ForgeHooksClient::getFieldOfView' and set the FOV to 0.1 when zooming is desired. 
- Achieve the spyglass overlay effect using an event handler for the game rendering, and create your own overlay render.
- Achieve the sound by playing the sound (see the vanilla spyglass item class)

PR #8329, opened 2021-12-27
[CVE-2021-45046/CVE-2021-45105] update Log4J and its subcomponents

daerich:
Completely fix Log4J issue as mitigations are insufficient/ineffective according to the Apache Log4J maintainers.


TheCurle:
Minecraft is not vulnerable to the second or third CVE - we don't use ThreadLocal or Logger#printf, so we're safe.

We'll update it in the future, but this is unneeded right now.

PR #8330, opened 2021-12-27
Fix crash with PermissionsAPI

JTK222:
Fixes a seemingly rare crash that was reported in the Forge Discord server today.
<https://discord.com/channels/313125603924639766/313125603924639766/925064802932502579>

That is done by replacing the security measures, which prevent other mods from reseting the PermissionAPI.

### DIFF
--- a/src/main/java/net/minecraftforge/server/permission/PermissionAPI.java
+++ b/src/main/java/net/minecraftforge/server/permission/PermissionAPI.java
@@ -109,9 +109,10 @@ public final class PermissionAPI
      */
     public static void initializePermissionAPI()
     {
-        if(StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE).getCallerClass() != ServerLifecycleHooks.class)
+        Class callerClass = StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE).getCallerClass();
+        if (callerClass != ServerLifecycleHooks.class)
         {
-            LOGGER.warn("Another mod tried to initialize the PermissionAPI, this is not allowed!");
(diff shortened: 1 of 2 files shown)


src/main/java/net/minecraftforge/server/permission/PermissionAPI.java
```diff
@@ -108,7 +109,13 @@ public static ResourceLocation getActivePermissionHandler()
      */
     public static void initializePermissionAPI()
     {
-        if (PermissionAPI.activeHandler != null) throw new IllegalStateException("Tried to initialize PermissionAPI multiple times!");
+        if(StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE).getCallerClass() != ServerLifecycleHooks.class)
```
sciwhiz12: Formatting: space after `if` and before opening parenthesis.
```suggestion
        if (StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE).getCallerClass() != ServerLifecycleHooks.class)
```
Additionally, it may be good to store the calling class and, if not `ServerLifecycleHooks`, log the name of the class.

sciwhiz12 (changes requested):
As the Permission API initialization only happens once during the server starting up, the (already small) cost of walking up 1 (regular) frame up the stack is negligible. Additionally, this also protects from any mods manually calling `resetPermissionAPI()` then reinitializing the API afterwards.
A possible alternative, if this is not accepted by maintainers, is to move the API reset call to `ServerStoppedEvent`.

Aside from a small formatting quibble and a suggestion, this looks :shipit: to me.

sciwhiz12 (approved):

TheCurle:
heck

PR #8331, opened 2021-12-28
[1.18.X] Add support for user-defined render types in static geometry

amadornes:
Minecraft's render engine supports the creation of custom render types, which allow the user to specify the vertex format and render state (face culling, depth sorting, shaders, etc) for certain parts of their geometry. However, natively these only work on dynamic geometry (entity/blockentity renderers), and not static elements like terrain.

This PR adds the necessary hooks to extend the default set of render types supported for static geometry rendering (solid, cutout_mipped, cutout, translucent, tripwire).

## Motivation

The `RenderType` system is incredibly powerful, but it's currently limited to just dynamic geometry, which requires a considerable amount of CPU time and GPU bandwidth to gather and upload every single frame. This means that any effect involving static geometry that requires a custom render state or shader (which is not all that uncommon) is creating an unnecessary load on both the CPU and GPU.

## Implementation

As suggested by @OrionDevelopment, the code from Forge's existing `TierSortingRegistry` was generified and repurposed to provide a method through which developers can register their own `RenderType` instances and customize their ordering relative to both Minecraft's built in render types as well as other mods'. The tier sorting registry has been migrated to this generified registry as well.

Due to the way Minecraft schedules its rendering, the render types have been split into three phases: solid, translucent and tripwire. Solid render types are rendered before entities, translucent render types are rendered after them, and tripwire render types are rendered after the dynamic buffers have been drawn. This is reflected in the events fired during client initialization to register custom render types: `StaticRenderTypeRegisterEvent.Solid`, `StaticRenderTypeRegisterEvent.Translucent` and `StaticRenderTypeRegisterEvent.Tripwire`. These events are fired on the mod bus as they are used to register mod-owned render types.

Along with these events, a `RenderTypeRegisterEvent` has been added for any mod that may need to register custom render types. This event provides a convenient and consistent place for this to take place, and ensures correct initialization of all required resources has taken place by the time it is fired. This event is also fired on the mod bus as it is used to register mod-owned render types.

The render calls for each phase are delegated from `LevelRenderer` to the `LevelRendererAdapter` class, which may decide (based on a Forge client config option) to run vanilla's default code path, or visit each of the vanilla + custom render types for that phase in the configured order. 

An additional set of interfaces and classes has been created to allow the creation of custom render types which receive a callback before rendering every chunk, allowing developers to compute and upload their own sets of uniforms to their shaders, or otherwise update the render state before the draw call is issued.

Forge's `MultiLayerModel` has been updated to support user-registered static render types, and the public field listing the old supported render types has been deprecated and marked for removal.

Lastly, a test mod is provided (mod id `layer_render_type_test`) which adds a block with a custom render type. This render type's shader makes the block's geometry shake, and its color changes depending on the camera's orientation relative to the faces.

## Demo

Here is a short demo gif of the block in the test mod as the camera moves around:

gif

## Performance and memory implications

The overhead of this system is minimal, with a baseline cost of 3 foreach loops iterating over a pre-computed empty list.

The cost of additional render layers is down to each layer's specific settings. The main factors will be:
 - Memory usage increase of at least `RenderType#bufferSize()` for every chunk (negligible under vanilla standards of <= 1MB)
 - Added CPU cost of capturing and rendering the additional geometry (may be considerably optimized by deviating from the current way Forge handles render types in baked models, but this would be another PR)
 - Minor CPU hit during startup while the render types are registered and sorted

In conclusion, the actual cost if unused is negligible, but as with everything, it must be used in moderation (at least until the aforementioned changes to baked models are discussed and implemented).

## Maintainability

This PR moves the maintenance tasks related to `MultiLayerModel` (keeping an updated list of vanilla-supported static render types) to the `LevelRenderPhaseManager` class. `MultiLayerModel` now relies on this, and as such it does not need to be kept updated anymore.

The mipmap fixes in `LevelRenderer` have been moved to `ForgeRenderTypes.MipmapFixTextureShard` and are now applied by replacing vanilla's default mipmap texture render state shard, which is a more idiomatic approach and ensures that the fix is applied to any render types using the mipmapped blocks texture.

Additional maintenance tasks may involve adding/re-ordering hooks in `LevelRenderer` in the case that vanilla adds new static render types. These changes are unlikely and should not result in any changes to the core API.

## Additional notes

During the testing of this pull request, I found that in a Forge dev environment with test mods enabled, reloading resources triggers an invalid GL state which changes the colors of everything in the game and prevents text from rendering. This has been tested with and without this PR's changes, and the effect is the same, so the problem is expected to be in one of the other tests. This has not been tested in an environment without tests because if it works in a mod dev environment (which it does), it should work in an empty Forge dev environment.

If you want, I can open an issue about this.

---

Special thanks to @OrionDevelopment for all the help, feedback and code provided throughout the development of this PR.  
They provided a significant amount of code, but the git authorship has been lost in refactoring and cleanup.

### DIFF
--- a/patches/minecraft/net/minecraft/client/renderer/LevelRenderer.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/LevelRenderer.java.patch
@@ -43,7 +43,7 @@
        profilerfiller.m_6182_("compilechunks");
        this.m_194370_(p_109604_);
        profilerfiller.m_6182_("terrain");
-+      if (!net.minecraftforge.client.renderer.LevelRendererAdapter.getInstance().onRenderSolids(this, p_109600_, p_109604_, p_109607_, profilerfiller)) {
++      if (!net.minecraftforge.client.renderer.LevelRendererAdapter.getInstance().onRenderSolids(this, p_109600_, p_109604_, p_109607_)) {
        this.m_172993_(RenderType.m_110451_(), p_109600_, d0, d1, d2, p_109607_);
        this.m_172993_(RenderType.m_110457_(), p_109600_, d0, d1, d2, p_109607_);
        this.m_172993_(RenderType.m_110463_(), p_109600_, d0, d1, d2, p_109607_);
@@ -94,11 +94,11 @@
           this.f_109413_.m_83954_(Minecraft.f_91002_);
           this.f_109413_.m_83945_(this.f_109461_.m_91385_());
           profilerfiller.m_6182_("translucent");
-+         if (!net.minecraftforge.client.renderer.LevelRendererAdapter.getInstance().onRenderTranslucent(this, p_109600_, p_109604_, p_109607_, profilerfiller)) {
++         if (!net.minecraftforge.client.renderer.LevelRendererAdapter.getInstance().onRenderTranslucent(this, p_109600_, p_109604_, p_109607_)) {
           this.m_172993_(RenderType.m_110466_(), p_109600_, d0, d1, d2, p_109607_);
 +         }
           profilerfiller.m_6182_("string");
-+         if (!net.minecraftforge.client.renderer.LevelRendererAdapter.getInstance().onRenderTripwire(this, p_109600_, p_109604_, p_109607_, profilerfiller)) {
++         if (!net.minecraftforge.client.renderer.LevelRendererAdapter.getInstance().onRenderTripwire(this, p_109600_, p_109604_, p_109607_)) {
           this.m_172993_(RenderType.m_110503_(), p_109600_, d0, d1, d2, p_109607_);
 +         }
           this.f_109415_.m_83954_(Minecraft.f_91002_);
@@ -114,13 +114,13 @@
              this.f_109413_.m_83954_(Minecraft.f_91002_);
           }
  
-+         if (!net.minecraftforge.client.renderer.LevelRendererAdapter.getInstance().onRenderTranslucent(this, p_109600_, p_109604_, p_109607_, profilerfiller)) {
++         if (!net.minecraftforge.client.renderer.LevelRendererAdapter.getInstance().onRenderTranslucent(this, p_109600_, p_109604_, p_109607_)) {
           this.m_172993_(RenderType.m_110466_(), p_109600_, d0, d1, d2, p_109607_);
 +         }
           multibuffersource$buffersource.m_109912_(RenderType.m_110504_());
           multibuffersource$buffersource.m_109911_();
           profilerfiller.m_6182_("string");
-+         if (!net.minecraftforge.client.renderer.LevelRendererAdapter.getInstance().onRenderTripwire(this, p_109600_, p_109604_, p_109607_, profilerfiller)) {
++         if (!net.minecraftforge.client.renderer.LevelRendererAdapter.getInstance().onRenderTripwire(this, p_109600_, p_109604_, p_109607_)) {
           this.m_172993_(RenderType.m_110503_(), p_109600_, d0, d1, d2, p_109607_);
 +         }
           profilerfiller.m_6182_("particles");
--- a/patches/minecraft/net/minecraft/client/renderer/LevelRenderer.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/LevelRenderer.java.patch
@@ -24,7 +24,7 @@
        float f = this.f_109461_.f_91073_.m_46722_(1.0F) / (Minecraft.m_91405_() ? 1.0F : 2.0F);
        if (!(f <= 0.0F)) {
           Random random = new Random((long)this.f_109477_ * 312987231L);
-@@ -1148,20 +_,23 @@
+@@ -1148,21 +_,24 @@
        RenderSystem.m_69421_(16640, Minecraft.f_91002_);
        float f = p_109605_.m_109152_();
        boolean flag2 = this.f_109461_.f_91073_.m_104583_().m_5781_(Mth.m_14107_(d0), Mth.m_14107_(d1)) || this.f_109461_.f_91065_.m_93090_().m_93715_();
@@ -43,13 +43,14 @@
        profilerfiller.m_6182_("compilechunks");
        this.m_194370_(p_109604_);
        profilerfiller.m_6182_("terrain");
++      if (!net.minecraftforge.client.renderer.LevelRendererAdapter.getInstance().onRenderSolids(this, p_109600_, p_109604_, p_109607_, profilerfiller)) {
        this.m_172993_(RenderType.m_110451_(), p_109600_, d0, d1, d2, p_109607_);
-+      this.f_109461_.m_91304_().m_119428_(TextureAtlas.f_118259_).setBlurMipmap(false, this.f_109461_.f_91066_.f_92027_ > 0); // FORGE: fix flickering leaves when mods mess up the blurMipmap settings
        this.m_172993_(RenderType.m_110457_(), p_109600_, d0, d1, d2, p_109607_);
-+      this.f_109461_.m_91304_().m_119428_(TextureAtlas.f_118259_).restoreLastBlurMipmap();
        this.m_172993_(RenderType.m_110463_(), p_109600_, d0, d1, d2, p_109607_);
++      }
        if (this.f_109465_.m_104583_().m_108885_()) {
           Lighting.m_84925_(p_109600_.m_85850_().m_85861_());
+       } else {
 @@ -1191,7 +_,7 @@
        MultiBufferSource.BufferSource multibuffersource$buffersource = this.f_109464_.m_110104_();
  
@@ -89,7 +90,18 @@
        }
  
        PoseStack posestack = RenderSystem.m_157191_();
-@@ -1344,7 +_,7 @@
+@@ -1337,14 +_,18 @@
+          this.f_109413_.m_83954_(Minecraft.f_91002_);
+          this.f_109413_.m_83945_(this.f_109461_.m_91385_());
+          profilerfiller.m_6182_("translucent");
++         if (!net.minecraftforge.client.renderer.LevelRendererAdapter.getInstance().onRenderTranslucent(this, p_109600_, p_109604_, p_109607_, profilerfiller)) {
+          this.m_172993_(RenderType.m_110466_(), p_109600_, d0, d1, d2, p_109607_);
++         }
+          profilerfiller.m_6182_("string");
++         if (!net.minecraftforge.client.renderer.LevelRendererAdapter.getInstance().onRenderTripwire(this, p_109600_, p_109604_, p_109607_, profilerfiller)) {
+          this.m_172993_(RenderType.m_110503_(), p_109600_, d0, d1, d2, p_109607_);
++         }
+          this.f_109415_.m_83954_(Minecraft.f_91002_);
           this.f_109415_.m_83945_(this.f_109461_.m_91385_());
           RenderStateShard.f_110126_.m_110185_();
           profilerfiller.m_6182_("particles");
@@ -98,15 +110,33 @@
           RenderStateShard.f_110126_.m_110188_();
        } else {
           profilerfiller.m_6182_("translucent");
-@@ -1358,7 +_,7 @@
+@@ -1352,13 +_,17 @@
+             this.f_109413_.m_83954_(Minecraft.f_91002_);
+          }
+ 
++         if (!net.minecraftforge.client.renderer.LevelRendererAdapter.getInstance().onRenderTranslucent(this, p_109600_, p_109604_, p_109607_, profilerfiller)) {
+          this.m_172993_(RenderType.m_110466_(), p_109600_, d0, d1, d2, p_109607_);
++         }
+          multibuffersource$buffersource.m_109912_(RenderType.m_110504_());
+          multibuffersource$buffersource.m_109911_();
           profilerfiller.m_6182_("string");
++         if (!net.minecraftforge.client.renderer.LevelRendererAdapter.getInstance().onRenderTripwire(this, p_109600_, p_109604_, p_109607_, profilerfiller)) {
           this.m_172993_(RenderType.m_110503_(), p_109600_, d0, d1, d2, p_109607_);
++         }
           profilerfiller.m_6182_("particles");
 -         this.f_109461_.f_91061_.m_107336_(p_109600_, multibuffersource$buffersource, p_109606_, p_109604_, p_109601_);
 +         this.f_109461_.f_91061_.render(p_109600_, multibuffersource$buffersource, p_109606_, p_109604_, p_109601_, frustum);
        }
  
        posestack.m_85836_();
+@@ -1511,6 +_,7 @@
+                uniform.m_85633_();
+             }
+ 
++            if (p_172994_ instanceof net.minecraftforge.client.renderer.IComplexRenderType complexRenderType) { complexRenderType.beforeChunkRender(this, chunkrenderdispatcher$renderchunk, p_172995_, p_172996_, p_172997_, p_172998_, p_172999_); }
+             vertexbuffer.m_166887_();
+             flag1 = true;
+          }
 @@ -1790,6 +_,11 @@
  
     public void m_181409_(PoseStack p_181410_, Matrix4f p_181411_, float p_181412_, Runnable p_181413_) {
--- a/patches/minecraft/net/minecraft/client/renderer/RenderType.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/RenderType.java.patch
@@ -34,3 +34,12 @@
     }
  
     public static RenderType m_110502_() {
+@@ -375,7 +_,7 @@
+    }
+ 
+    public static List<RenderType> m_110506_() {
+-      return ImmutableList.of(m_110451_(), m_110457_(), m_110463_(), m_110466_(), m_110503_());
++      return net.minecraftforge.client.renderer.LevelRenderPhaseManager.getInstance().getKnownTypes();
+    }
+ 
+    public int m_110507_() {
--- a/src/main/java/net/minecraftforge/client/model/MultiLayerModel.java
+++ b/src/main/java/net/minecraftforge/client/model/MultiLayerModel.java
@@ -44,6 +44,7 @@ import net.minecraftforge.client.MinecraftForgeClient;
 import net.minecraftforge.client.model.data.IDynamicBakedModel;
 import net.minecraftforge.client.model.data.IModelData;
 import net.minecraftforge.client.model.geometry.IModelGeometry;
+import net.minecraftforge.client.renderer.LevelRenderPhaseManager;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -246,6 +247,10 @@ public final class MultiLayerModel implements IModelGeometry<MultiLayerModel>
 
     public static final class Loader implements IModelLoader<MultiLayerModel>
     {
+        /**
+         * Replaced by {@link LevelRenderPhaseManager#getAllKnownTypes()}
+         */
+        @Deprecated(forRemoval = true)
         public static final ImmutableBiMap<String, RenderType> BLOCK_LAYERS = ImmutableBiMap.<String, RenderType>builder()
                 .put("solid", RenderType.solid())
                 .put("cutout", RenderType.cutout())
@@ -269,13 +274,11 @@ public final class MultiLayerModel implements IModelGeometry<MultiLayerModel>
         {
             ImmutableList.Builder<Pair<RenderType, UnbakedModel>> builder = ImmutableList.builder();
             JsonObject layersObject = GsonHelper.getAsJsonObject(modelContents, "layers");
-            for(Map.Entry<String, RenderType> layer : BLOCK_LAYERS.entrySet()) // block layers
-            {
-                String layerName = layer.getKey(); // mc overrides toString to return the ID for the layer
-                if(layersObject.has(layerName))
-                {
-                    builder.add(Pair.of(layer.getValue(), deserializationContext.deserialize(GsonHelper.getAsJsonObject(layersObject, layerName), BlockModel.class)));
-                }
+            var renderTypes = LevelRenderPhaseManager.getInstance().getAllKnownTypes();
+            for (var key : layersObject.keySet()) {
+                var renderType = renderTypes.get(new ResourceLocation(key));
+                if (renderType != null)
+                    builder.add(Pair.of(renderType, deserializationContext.deserialize(GsonHelper.getAsJsonObject(layersObject, key), BlockModel.class)));
             }
             boolean convertRenderTypes = GsonHelper.getAsBoolean(modelContents, "convert_render_types", true);
             return new MultiLayerModel(builder.build(), convertRenderTypes);
--- a/src/main/java/net/minecraftforge/client/renderer/ComplexRenderType.java
+++ b/src/main/java/net/minecraftforge/client/renderer/ComplexRenderType.java
@@ -7,11 +7,19 @@ import net.minecraft.client.renderer.LevelRenderer;
 import net.minecraft.client.renderer.RenderType;
 import net.minecraft.client.renderer.chunk.ChunkRenderDispatcher;
 
+/**
+ * An implementation of {@link net.minecraft.client.renderer.RenderType.CompositeRenderType} with
+ * the {@link IComplexRenderType} interface.
+ * Create using {@link ComplexRenderTypeBuilder}.
+ *
+ * @see ComplexRenderTypeBuilder
+ */
 public final class ComplexRenderType extends RenderType.CompositeRenderType implements IComplexRenderType
 {
     private final IBeforeChunkRenderCallback beforeChunkRenderCallback;
 
-    public static ComplexRenderTypeBuilder builder() {
+    public static ComplexRenderTypeBuilder builder()
+    {
         return new ComplexRenderTypeBuilder();
     }
 
@@ -23,7 +31,8 @@ public final class ComplexRenderType extends RenderType.CompositeRenderType impl
       final boolean affectsCrumbling,
       final boolean sortOnUpload,
       final CompositeState compositeState,
-      final IBeforeChunkRenderCallback beforeChunkRenderCallback) {
+      final IBeforeChunkRenderCallback beforeChunkRenderCallback
+    ) {
         super(name, vertexFormat, mode, bufferSize, affectsCrumbling, sortOnUpload, compositeState);
         this.beforeChunkRenderCallback = beforeChunkRenderCallback;
     }
@@ -36,8 +45,8 @@ public final class ComplexRenderType extends RenderType.CompositeRenderType impl
       final double cameraX,
       final double cameraY,
       final double cameraZ,
-      final Matrix4f projectionMatrix)
-    {
+      final Matrix4f projectionMatrix
+    ) {
         this.beforeChunkRenderCallback.beforeChunkRender(
           levelRenderer, renderChunk, poseStack, cameraX, cameraY, cameraZ, projectionMatrix
         );
--- a/src/main/java/net/minecraftforge/client/renderer/ComplexRenderType.java
+++ b/src/main/java/net/minecraftforge/client/renderer/ComplexRenderType.java
@@ -1,0 +1,45 @@
+package net.minecraftforge.client.renderer;
+
+import com.mojang.blaze3d.vertex.PoseStack;
+import com.mojang.blaze3d.vertex.VertexFormat;
+import com.mojang.math.Matrix4f;
+import net.minecraft.client.renderer.LevelRenderer;
+import net.minecraft.client.renderer.RenderType;
+import net.minecraft.client.renderer.chunk.ChunkRenderDispatcher;
+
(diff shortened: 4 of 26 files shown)


amadornes:
Currently working with Orion on some changes to try to clean up the patches.

amadornes:
I have updated the PR description with all the changes Orion and I have made throughout the day.

The patches have been reduced to a minimal footprint and several quality of life improvements have been made, including the addition of a new event for registering custom render types and hooks for uploading custom uniforms to a render type's shader on a per-chunk basis.

sciwhiz12:
Noting here: @OrionDevelopment made substantial contributions to this PR as noted by the author in the PR description, and therefore cannot triage, review, and assign this PR on grounds of conflict of interest.

LatvianModder:
My new mod which uses shaders/RenderTypes in static geometry could *really* use this PR 😛 https://www.curseforge.com/minecraft/mc-mods/literal-sky-block

src/main/java/net/minecraftforge/client/model/MultiLayerModel.java
```diff
@@ -246,6 +247,10 @@ public boolean isLayered()
 
     public static final class Loader implements IModelLoader<MultiLayerModel>
     {
+        /**
+         * Replaced by {@link LevelRenderPhaseManager#getAllKnownTypes()}
+         */
+        @Deprecated(forRemoval = true)
```
sciwhiz12, line 253: Please add `since = "1.18.1"` to this deprecation annotation.

src/main/java/net/minecraftforge/client/model/MultiLayerModel.java
```diff
@@ -269,13 +274,11 @@ public MultiLayerModel read(JsonDeserializationContext deserializationContext, J
...
             JsonObject layersObject = GsonHelper.getAsJsonObject(modelContents, "layers");
-            for(Map.Entry<String, RenderType> layer : BLOCK_LAYERS.entrySet()) // block layers
-            {
-                String layerName = layer.getKey(); // mc overrides toString to return the ID for the layer
-                if(layersObject.has(layerName))
-                {
-                    builder.add(Pair.of(layer.getValue(), deserializationContext.deserialize(GsonHelper.getAsJsonObject(layersObject, layerName), BlockModel.class)));
-                }
+            var renderTypes = LevelRenderPhaseManager.getInstance().getAllKnownTypes();
+            for (var key : layersObject.keySet()) {
```
sciwhiz12, line 278: Formatting: braces on newlines:
```suggestion
            for (var key : layersObject.keySet())
            {
```

src/main/java/net/minecraftforge/client/renderer/ComplexRenderType.java
```diff
@@ -0,0 +1,54 @@
+package net.minecraftforge.client.renderer;
```
sciwhiz12, line 1: Missing license header.

src/main/java/net/minecraftforge/client/renderer/ComplexRenderType.java
```diff
@@ -0,0 +1,54 @@
...
+import net.minecraft.client.renderer.LevelRenderer;
+import net.minecraft.client.renderer.RenderType;
+import net.minecraft.client.renderer.chunk.ChunkRenderDispatcher;
+
+/**
+ * An implementation of {@link net.minecraft.client.renderer.RenderType.CompositeRenderType} with
+ * the {@link IComplexRenderType} interface.
+ * Create using {@link ComplexRenderTypeBuilder}.
+ *
+ * @see ComplexRenderTypeBuilder
```
sciwhiz12, line 15: This should likely include a `@see` tag for `ComplexRenderType#builder()`.

src/main/java/net/minecraftforge/client/renderer/ComplexRenderTypeBuilder.java
```diff
@@ -0,0 +1,37 @@
+package net.minecraftforge.client.renderer;
```
sciwhiz12, line 1: Missing license header.

src/main/java/net/minecraftforge/client/renderer/ComplexRenderTypeBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.mojang.blaze3d.vertex.VertexFormat;
+import net.minecraft.client.renderer.RenderType;
+
+public final class ComplexRenderTypeBuilder
```
sciwhiz12, line 6: What is the use of a builder if there is only one method which can be built? The other components of a `ComplexRenderType` should have corresponding builder methods to modify them (with either default values or a placeholder which is checked to have been changed at the end of building). I can see candidates of this being the boolean fields of `affectsCrumbling` and `sortOnUpload`.

Furthermore, is there a particular reason this builder class is separate from the class which it is building? Could this perhaps be moved as a nested static class, as in `ComplexRenderType.Builder`?

src/main/java/net/minecraftforge/client/renderer/ComplexRenderTypeBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+
+    private IBeforeChunkRenderCallback beforeChunkRenderCallback = DEFAULT_BEFORE_CHUNK_CALLBACK;
+
+    ComplexRenderTypeBuilder()
+    {
+    }
+
+    public ComplexRenderTypeBuilder beforeChunkRenderCallback(IBeforeChunkRenderCallback beforeChunkRenderCallback)
+    {
+        this.beforeChunkRenderCallback = beforeChunkRenderCallback;
```
sciwhiz12, line 20: Part of me feels like this needs a non-null check.

src/main/java/net/minecraftforge/client/renderer/IBeforeChunkRenderCallback.java
```diff
@@ -0,0 +1,40 @@
+package net.minecraftforge.client.renderer;
```
sciwhiz12, line 1: Missing license header.

src/main/java/net/minecraftforge/client/renderer/IBeforeChunkRenderCallback.java
```diff
@@ -0,0 +1,40 @@
...
+import net.minecraft.client.renderer.chunk.ChunkRenderDispatcher;
+
+/**
+ * Interface which represents a callback that is invoked by the {@link LevelRenderer} before
+ * a non-empty chunk is rendered. This is useful for uploading additional uniforms to the GPU,
+ * or modifying the current GPU state based on what chunk is being rendered.
+ *
+ * For use with render types, must be passed to a {@link ComplexRenderTypeBuilder}.
+ * This render type must be registered via a {@link net.minecraftforge.client.event.LayerRenderTypeRegisterEvent}
+ * of the appropriate phase.
```
sciwhiz12, line 15: This javadoc sentence feels out of place to me.

src/main/java/net/minecraftforge/client/renderer/IComplexRenderType.java
```diff
@@ -0,0 +1,11 @@
+package net.minecraftforge.client.renderer;
```
sciwhiz12, line 1: Missing license header.

src/main/java/net/minecraftforge/client/renderer/IComplexRenderType.java
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * Interface which implemented by custom render types which are to be
+ * notified when a chunk section is rendered using them.
+ *
+ * @see ComplexRenderType
+ */
+public interface IComplexRenderType extends IBeforeChunkRenderCallback
```
sciwhiz12, line 9: Is there a particular reason why this interface exists? From what I can see, this is only used to `instanceof` on a render type, to invoke the method from `IBeforeChunkRenderCallback`. Why then can't that `instanceof` use that interface directly?

src/main/java/net/minecraftforge/client/renderer/LevelRenderPhaseManager.java
```diff
@@ -0,0 +1,120 @@
+package net.minecraftforge.client.renderer;
```
sciwhiz12, line 1: Missing license header.

src/main/java/net/minecraftforge/client/renderer/LevelRendererAdapter.java
```diff
@@ -0,0 +1,100 @@
+package net.minecraftforge.client.renderer;
```
sciwhiz12, line 1: Missing license header.

src/main/java/net/minecraftforge/client/renderer/LevelRendererAdapter.java
```diff
@@ -0,0 +1,100 @@
...
+     * Invoked by the renderer when it is time to render the blocks which are solid.
+     *
+     * @return {@code true} when the renderer should not render the blocks, {@code false} otherwise.
+     */
+    public boolean onRenderSolids(
+      final LevelRenderer levelRenderer,
+      final PoseStack poseStack,
+      final Camera camera,
+      final Matrix4f projectionMatrix
+    ) {
```
sciwhiz12, line 41: Formatting: braces on newlines:
```suggestion
    )
    {
```

src/main/java/net/minecraftforge/client/renderer/LevelRendererAdapter.java
```diff
@@ -0,0 +1,100 @@
...
+     * Invoked by the renderer when it is time to render the blocks which are translucent.
+     *
+     * @return {@code true} when the renderer should not render the blocks, {@code false} otherwise.
+     */
+    public boolean onRenderTranslucent(
+      final LevelRenderer levelRenderer,
+      final PoseStack poseStack,
+      final Camera camera,
+      final Matrix4f projectionMatrix
+    ) {
```
sciwhiz12, line 60: Formatting: braces on newlines:
```suggestion
    )
    {
```

src/main/java/net/minecraftforge/client/renderer/LevelRendererAdapter.java
```diff
@@ -0,0 +1,100 @@
...
+     * Invoked by the renderer when it is time to render the blocks which are tripwire.
+     *
+     * @return {@code true} when the renderer should not render the blocks, {@code false} otherwise.
+     */
+    public boolean onRenderTripwire(
+      final LevelRenderer levelRenderer,
+      final PoseStack poseStack,
+      final Camera camera,
+      final Matrix4f projectionMatrix
+    ) {
```
sciwhiz12, line 79: Formatting: braces on newlines:
```suggestion
    )
    {
```

src/main/java/net/minecraftforge/client/renderer/LevelRendererAdapter.java
```diff
@@ -0,0 +1,100 @@
...
+        return true;
+    }
+
+    private void doRenderPhases(
+      final Collection<RenderType> phases,
+      final Camera camera,
+      final LevelRenderer levelRenderer,
+      final PoseStack poseStack,
+      final Matrix4f projectionMatrix
+    ) {
```
sciwhiz12, line 94: Formatting: braces on newlines:
```suggestion
    )
    {
```

src/main/java/net/minecraftforge/common/ForgeConfig.java
```diff
@@ -159,6 +161,11 @@
                     .translation("forge.configgui.forceSystemNanoTime")
                     .define("forceSystemNanoTime", false);
 
+            useLevelRendererAdapter = builder
+                    .comment("Set to true to use the Forge LevelRendererAdapter instead of the default one.")
+                    .translation("forge.configgui.useLevelRendererAdapter")
+                    .define("useLevelRendererAdapter", true);
```
sciwhiz12, line 167: What situations would necessitate disabling this config option? Since it's not clear to me why one would wish to disable this.

src/main/java/net/minecraftforge/common/ForgeConfig.java
```diff
@@ -159,6 +161,11 @@
                     .translation("forge.configgui.forceSystemNanoTime")
                     .define("forceSystemNanoTime", false);
 
+            useLevelRendererAdapter = builder
+                    .comment("Set to true to use the Forge LevelRendererAdapter instead of the default one.")
```
sciwhiz12, line 165: The wording suggests that there is a vanilla `LevelRendererAdapter`, and that this config option substitutes that class with Forge's own implementation. It should be reworded something, possibly "Enables the rendering of custom static render types through the LevelRendererAdapter." or something similar.

src/main/java/net/minecraftforge/common/util/SortedRegistry.java
```diff
@@ -0,0 +1,219 @@
...
+import com.google.common.graph.ElementOrder;
+import com.google.common.graph.GraphBuilder;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraftforge.fml.loading.toposort.TopologicalSort;
+
+import javax.annotation.Nullable;
+import java.util.*;
+
+/**
+ * A {@link ResourceLocation} to T registry where elements may define arbitrary ordering rules.
```
sciwhiz12, line 32: I think the javadoc could be reworded somewhat:
```suggestion
 * A registry of elements with {@link ResourceLocation} names with topological sorting.
```

src/main/java/net/minecraftforge/common/util/SortedRegistry.java
```diff
@@ -0,0 +1,219 @@
...
+    private final ImmutableBiMap<ResourceLocation, T> names;
+
+    private SortedRegistry(ImmutableList<T> elements, ImmutableBiMap<ResourceLocation, T> names)
+    {
+        this.elements = elements;
+        this.names = names;
+    }
+
+    /**
+     * Returns the list of elements in ascending order as defined by the dependencies.
```
sciwhiz12, line 47: Could this clarify what is meant by "ascending order"?

src/main/java/net/minecraftforge/common/util/SortedRegistry.java
```diff
@@ -0,0 +1,219 @@
...
+        public synchronized <T2 extends T> T2 add(T2 element, ResourceLocation name, List<Object> predecessors, List<Object> successors)
+        {
+            Objects.requireNonNull(predecessors, "Predecessor list may not be null.");
+            Objects.requireNonNull(successors, "Successor list may not be null.");
+            if (entries.containsKey(name))
+                throw new IllegalArgumentException("An entry has already been registered with this name: " + name);
+            entries.put(name, element);
+            for(var other : predecessors)
+                edges.put(getName(other), name);
+            for(var other : successors)
```
sciwhiz12, line 177: Formatting: space before the opening parentheses in a `for`-expression:
```suggestion
            for (var other : successors)
```

src/main/java/net/minecraftforge/common/util/SortedRegistry.java
```diff
@@ -0,0 +1,219 @@
...
+         * @return The element
+         */
+        public synchronized <T2 extends T> T2 add(T2 element, ResourceLocation name, List<Object> predecessors, List<Object> successors)
+        {
+            Objects.requireNonNull(predecessors, "Predecessor list may not be null.");
+            Objects.requireNonNull(successors, "Successor list may not be null.");
+            if (entries.containsKey(name))
+                throw new IllegalArgumentException("An entry has already been registered with this name: " + name);
+            entries.put(name, element);
+            for(var other : predecessors)
```
sciwhiz12, line 175: Formatting: space before the opening parentheses in a `for`-expression:
```suggestion
            for (var other : predecessors)
```

src/main/java/net/minecraftforge/common/util/SortedRegistry.java
```diff
@@ -0,0 +1,219 @@
...
+        {
+            this.type = type;
+        }
+
+        /**
+         * Adds an element to the sorted registry.
+         * @param element The element
+         * @param name The name of the element
+         * @param predecessors A list of elements that must appear before this element (String, ResourceName, and T are allowed)
+         * @param successors A list of elements that must appear after this element (String, ResourceName, and T are allowed)
```
sciwhiz12, line 164:
```suggestion
         * @param predecessors A list of elements that must appear before this element. These may be instances of a registered element, or a name of a (registered or unregistered) element as a String or ResourceLocation.
         * @param successors A list of elements that must appear after this element. These may be instances of a registered element, or a name of a (registered or unregistered) element as a String or ResourceLocation.
```

src/main/java/net/minecraftforge/common/util/SortedRegistry.java
```diff
@@ -0,0 +1,219 @@
...
+
+        /**
+         * Creates a new {@link SortedRegistry} from the registered elements and dependencies.
+         * @return A new registry
+         */
+        @SuppressWarnings("UnstableApiUsage")
+        public SortedRegistry<T> build()
+        {
+            var graph = GraphBuilder.directed().nodeOrder(ElementOrder.<T>insertion()).build();
+            for(var element : entries.values())
```
sciwhiz12, line 205: Formatting: space before the opening parentheses in a for-expression:
```suggestion
            for (var element : entries.values())
```

src/main/java/net/minecraftforge/common/util/SortedRegistry.java
```diff
@@ -0,0 +1,219 @@
...
+package net.minecraftforge.common.util;
+
+import com.google.common.collect.*;
+import com.google.common.graph.ElementOrder;
+import com.google.common.graph.GraphBuilder;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraftforge.fml.loading.toposort.TopologicalSort;
+
+import javax.annotation.Nullable;
+import java.util.*;
```
sciwhiz12, line 29: Please avoid the use of wildcard imports.

src/test/java/net/minecraftforge/debug/client/LayerRenderTypeTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+package net.minecraftforge.debug.client;
+
+import com.mojang.blaze3d.shaders.Uniform;
+import com.mojang.blaze3d.vertex.DefaultVertexFormat;
+import com.mojang.blaze3d.vertex.PoseStack;
+import com.mojang.blaze3d.vertex.VertexFormat;
+import com.mojang.math.Matrix4f;
+import net.minecraft.Util;
+import net.minecraft.client.renderer.*;
```
sciwhiz12, line 28: Please avoid the use of wildcard imports.

src/test/java/net/minecraftforge/debug/client/LayerRenderTypeTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+        private static void beforeChunkRender(
+            LevelRenderer levelRenderer,
+            ChunkRenderDispatcher.RenderChunk renderChunk,
+            PoseStack poseStack,
+            double cameraX,
+            double cameraY,
+            double cameraZ,
+            Matrix4f projectionMatrix
+        ) {
```
sciwhiz12, line 141: Formatting: braces on newlines:
```suggestion
        )
        {
```

src/test/java/net/minecraftforge/debug/client/LayerRenderTypeTest.java
```diff
@@ -0,0 +1,147 @@
...
+            final Uniform TIME = CUSTOM_SHADER.getUniform("SysTime");
```
sciwhiz12, line 142: A bit nitpicky, but the variable name should be in lowerCamelCase.

src/test/java/net/minecraftforge/debug/client/LayerRenderTypeTest.java
```diff
@@ -0,0 +1,147 @@
...
+public class LayerRenderTypeTest
+{
+    private static final boolean ENABLED = true;
+    static final String MOD_ID = "layer_render_type_test";
+
+    private static final DeferredRegister<Block> BLOCKS = DeferredRegister.create(ForgeRegistries.BLOCKS, MOD_ID);
+    private static final DeferredRegister<Item> ITEMS = DeferredRegister.create(ForgeRegistries.ITEMS, MOD_ID);
+
+    private static final RegistryObject<Block> TEST_BLOCK = BLOCKS.register("test", () -> new Block(BlockBehaviour.Properties.of(Material.STONE)));
+    private static final RegistryObject<Item> TEST_ITEM = ITEMS.register("test", () -> new BlockItem(TEST_BLOCK.get(), new Item.Properties()));
```
sciwhiz12, line 64: Could you please add this block to the Miscellaneous item group?

src/test/java/net/minecraftforge/debug/client/LayerRenderTypeTest.java
```diff
@@ -0,0 +1,147 @@
...
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.RegistryObject;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Objects;
+
+@Mod(LayerRenderTypeTest.MOD_ID)
+public class LayerRenderTypeTest
```
sciwhiz12, line 55: Please add a javadoc comment to the test mod explaining what this test mod adds, what it does, and what it tests for.

src/main/java/net/minecraftforge/client/renderer/LevelRenderPhaseManager.java
```diff
@@ -0,0 +1,120 @@
...
+
+    public BiMap<ResourceLocation, RenderType> getAllKnownTypes()
+    {
+        return allKnownTypesView;
+    }
+
+    public List<RenderType> getKnownTypes()
+    {
+        return knownTypesView;
+    }
```
sciwhiz12, line 97: Please document the difference between these two getters.

src/main/java/net/minecraftforge/client/renderer/ComplexRenderType.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public void beforeChunkRender(
+      final LevelRenderer levelRenderer,
+      final ChunkRenderDispatcher.RenderChunk renderChunk,
+      final PoseStack poseStack,
+      final double cameraX,
+      final double cameraY,
+      final double cameraZ,
+      final Matrix4f projectionMatrix
+    ) {
```
sciwhiz12, line 49: Formatting: braces on newlines:
```suggestion
    )
    {
```

src/main/java/net/minecraftforge/client/renderer/ComplexRenderType.java
```diff
@@ -0,0 +1,54 @@
...
+    ComplexRenderType(
+      final String name,
+      final VertexFormat vertexFormat,
+      final VertexFormat.Mode mode,
+      final int bufferSize,
+      final boolean affectsCrumbling,
+      final boolean sortOnUpload,
+      final CompositeState compositeState,
+      final IBeforeChunkRenderCallback beforeChunkRenderCallback
+    ) {
```
sciwhiz12, line 35: Formatting: braces on newlines:
```suggestion
    )
    {
```

src/main/java/net/minecraftforge/client/renderer/LevelRendererAdapter.java
```diff
@@ -0,0 +1,100 @@
...
+import net.minecraft.world.phys.Vec3;
+import net.minecraftforge.common.ForgeConfig;
+
+import java.util.Collection;
+
+/**
+ * Adapter class used to dispatch draw calls for vanilla and user-provided static
+ * render types during level rendering.
+ */
+public final class LevelRendererAdapter
```
sciwhiz12, line 18: This isn't really an [adapter](https://en.wikipedia.org/wiki/Adapter_pattern) class, but more of a hooks class (in the same vein of `ForgeHooksClient`).

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -75,7 +75,9 @@ public net.minecraft.client.particle.ParticleEngine m_107378_(Lnet/minecraft/cor
 public net.minecraft.client.particle.ParticleEngine m_107381_(Lnet/minecraft/core/particles/ParticleType;Lnet/minecraft/client/particle/ParticleProvider;)V # register
 public net.minecraft.client.particle.ParticleEngine$SpriteParticleRegistration
 public net.minecraft.client.renderer.GameRenderer m_109128_(Lnet/minecraft/resources/ResourceLocation;)V # loadEffect
+public net.minecraft.client.renderer.LevelRenderer m_172993_(Lnet/minecraft/client/renderer/RenderType;Lcom/mojang/blaze3d/vertex/PoseStack;DDDLcom/mojang/math/Matrix4f;)V
```
sciwhiz12, line 78: This needs to have the mapped name of the method (`renderChunkLayer`) in a comment after the declaration.

As a thought, is it possible and viable to avoid this AT (and opening up `renderChunkLayer` to everyone) by passing in a method reference to it to the `LevelRendererAdapter` hooks?

patches/minecraft/net/minecraft/client/renderer/LevelRenderer.java.patch
```diff
@@ -89,7 +90,18 @@
        }
  
        PoseStack posestack = RenderSystem.m_157191_();
-@@ -1344,7 +_,7 @@
+@@ -1337,14 +_,18 @@
+          this.f_109413_.m_83954_(Minecraft.f_91002_);
+          this.f_109413_.m_83945_(this.f_109461_.m_91385_());
+          profilerfiller.m_6182_("translucent");
++         if (!net.minecraftforge.client.renderer.LevelRendererAdapter.getInstance().onRenderTranslucent(this, p_109600_, p_109604_, p_109607_)) {
+          this.m_172993_(RenderType.m_110466_(), p_109600_, d0, d1, d2, p_109607_);
```
sciwhiz12, line 98: Looking at the code in `renderChunkLayer` shows some hardcoded checks for `RenderType.translucent()`. Do these checks substantially affect render types which render alongside the vanilla translucent render type?

We may want to expand the test mod to include tests for these three render type 'groups', rather than only registering a custom solid render type.

sciwhiz12 (changes requested):
While I don't fully comprehend rendering, I've gone ahead and looked at this PR, mostly from formatting and documentation angles. I'll leave the bulk of the actual rendering review to @gigaherz. 👀

Aside from the formatting and whatnot, I do have one notable concern in that the test mod seems to only provide a test for a custom solid render type (as I note in one of my review comments). Ideally, it should provide tests for all three render type 'groups', to help look for possible regressions or oversights now and in the future.
In particular, custom translucent render types need testing because of their special interactions with regards to other render types.

---

Please update the license headers of the files added or modified by this PR to reflect the current year.

src/main/java/net/minecraftforge/client/model/MultiLayerModel.java
LatvianModder: This is little off topic but why doesnt this repo have an .editorconfig file?

TheCurle: Because style is a weird and wacky place - patches and code have different styles, and no matter what we provide some people will choose to override it with their own preference anyway, which makes the whole thing moot.

patches/minecraft/net/minecraft/client/renderer/RenderType.java.patch
```diff
@@ -34,3 +34,12 @@
     }
  
     public static RenderType m_110502_() {
+@@ -375,7 +_,7 @@
+    }
+ 
+    public static List<RenderType> m_110506_() {
+-      return ImmutableList.of(m_110451_(), m_110457_(), m_110463_(), m_110466_(), m_110503_());
++      return net.minecraftforge.client.renderer.LevelRenderPhaseManager.getInstance().getKnownTypes();
```
XFactHD, line 42: In my opinion this is missing an accompanying patch in `RenderBuffers`: The `RenderBuffers#fixedBuffers` map which is used to construct the `RenderBuffers#bufferSource` needs to contain all chunk layers because requesting a buffer from said buffer source for a certain `RenderType` after using a different `RenderType` that isn't in the map of fixed buffers will cause the current data in the buffer to be uploaded to the GPU because "non-fixed" `RenderType`s use a shared buffer.

While this is not an issue for the static chunk rendering, which uses the `ChunkBufferBuilderPack` to get its buffers, I would still expect a `RenderType`, which has a pre-allocated buffer and can be used in static chunk rendering, to be useable in a dynamic context (i.e. `EntityRenderer` or `BlockEntityRenderer`) without causing an upload to the GPU every time I switch the `RenderType` in use to render the next thing in the dynamic renderer (which kills performance when a few instances of the related `Entity` or `BlockEntity` are present).

sciwhiz12:
Hello! @amadornes, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.19.2 (or leave an explanation why this PR's changes are not applicable to 1.19.2). If we do not hear back from you after some time, this PR will be closed for inactivity.

amadornes:
I'll be porting this 1.19, yep. Just haven't had the time for it 😅 

I'll close this and open a new PR targetting 1.19 directly once it's ready. I could keep the same one, but it's going to be redone from scratch due to the all changes introduced in this area during the client cleanup/refactor, so I may as well keep them separate and just reference this PR in the other one.

PR #8334, opened 2021-12-30
[1.16.x] Add world and location data to PlaySoundAtEntityEvent

NoOneCanCode:
Expose `World` and x/y/z coordinates in `PlaySoundAtEntityEvent`. This is necessary because the player entity attached to the event is often null and thus the location and world corresponding to the event cannot always be determined. These changes expose those values as separate fields so that they can be accessed by mods that subscribe to the event. This is useful, for instance, if a mod wants to hook into sound events to provide a sense of hearing to a mob (as I was trying to do before writing this PR).

An example mod is provided in `PlaySoundAtEntityEventTest.java`. Set the `ENABLED` flag to `true` to enable the event subscription. When the mod is enabled, there should be lots of log spam showing the coordinates and dimension name of each sound event. An easy way to confirm the coordinates are correct is to tp to an entity making sound.

### DIFF
--- a/patches/minecraft/net/minecraft/client/entity/player/ClientPlayerEntity.java.patch
+++ b/patches/minecraft/net/minecraft/client/entity/player/ClientPlayerEntity.java.patch
@@ -12,7 +12,7 @@
     }
  
     public void func_184185_a(SoundEvent p_184185_1_, float p_184185_2_, float p_184185_3_) {
-+      net.minecraftforge.event.entity.PlaySoundAtEntityEvent event = net.minecraftforge.event.ForgeEventFactory.onPlaySoundAtEntity(this, p_184185_1_, this.func_184176_by(), p_184185_2_, p_184185_3_);
++      net.minecraftforge.event.entity.PlaySoundAtEntityEvent event = net.minecraftforge.event.ForgeEventFactory.onPlaySoundAtEntity(this, p_184185_1_, this.func_184176_by(), p_184185_2_, p_184185_3_, this.field_70170_p, this.func_226277_ct_(), this.func_226278_cu_(), this.func_226281_cx_());
 +      if (event.isCanceled() || event.getSound() == null) return;
 +      p_184185_1_ = event.getSound();
 +      p_184185_2_ = event.getVolume();
--- a/patches/minecraft/net/minecraft/client/world/ClientWorld.java.patch
+++ b/patches/minecraft/net/minecraft/client/world/ClientWorld.java.patch
@@ -42,7 +42,7 @@
     }
  
     public void func_184148_a(@Nullable PlayerEntity p_184148_1_, double p_184148_2_, double p_184148_4_, double p_184148_6_, SoundEvent p_184148_8_, SoundCategory p_184148_9_, float p_184148_10_, float p_184148_11_) {
-+      net.minecraftforge.event.entity.PlaySoundAtEntityEvent event = net.minecraftforge.event.ForgeEventFactory.onPlaySoundAtEntity(p_184148_1_, p_184148_8_, p_184148_9_, p_184148_10_, p_184148_11_);
++      net.minecraftforge.event.entity.PlaySoundAtEntityEvent event = net.minecraftforge.event.ForgeEventFactory.onPlaySoundAtEntity(p_184148_1_, p_184148_8_, p_184148_9_, p_184148_10_, p_184148_11_, this, p_184148_2_, p_184148_4_, p_184148_6_);
 +      if (event.isCanceled() || event.getSound() == null) return;
 +      p_184148_8_ = event.getSound();
 +      p_184148_9_ = event.getCategory();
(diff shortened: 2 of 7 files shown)


diesieben07:
PRs should first target the current latest version (1.18.1). Once they are merged there they can be backported to LTS.

XFactHD:
This change looks completely pointless to me: The event already gives you the entity that fired the event, which means you already have access to the entity's world and coordinates via `EntityEvent#getEntity().getLevel()` and `EntityEvent#getEntity().getX/Y/Z()`.

diesieben07:
The entity is often null, the event name is very misleading.

XFactHD:
~~In those cases the proposed changes would lead to a `NullPointerException` anyway because the `Entity` passed into the methods that fire this event is accessed without null-checks.~~

EDIT: I take that back, just noticed that that they are using the actual `Entity` (not nullable) to get that data (which makes sense) instead of the `PlayerEntity` (nullable in certain cases). IMO the proper fix here would be to pass in the actual entity instead of the player to the entity parameter of the event. Thoug that could probably be considered a breaking change

NoOneCanCode:
[EDIT] I missed XFactHD's edit the first time I replied. Given I'm new to the project, I don't have much of an opinion as to how this gets resolved. I did read in the docs that submitting breaking changes is not allowed. If anyone from the team chimes in I'm happy to do the leg work of implementing a different solution if they decide on one.
[EDIT 2]
> IMO the proper fix here would be to pass in the actual entity instead of the player to the entity parameter of the event

I agree that this would be a nice feature but more work would be required. `ServerWorld#playSound` takes a player because it then passes it to `PlayerList#broadcast`, which uses the parameter to avoid sending a sound packet to the player that created the sound in the first place. If the forge event should use the originating entity, than numerous classes would need to be patched to use the `playSound(@Nullable PlayerEntity p_217384_1_, Entity p_217384_2_, SoundEvent p_217384_3_, SoundCategory p_217384_4_, float p_217384_5_, float p_217384_6_)` signature instead of the `playSound(@Nullable PlayerEntity p_184148_1_, double p_184148_2_, double p_184148_4_, double p_184148_6_, SoundEvent p_184148_8_, SoundCategory p_184148_9_, float p_184148_10_, float p_184148_11_)` signature. Even then, there would still be some cases when the entity is null because some world sounds do not originate from _any_ entity.

NoOneCanCode:
> The entity is often null, the event name is very misleading.

Might it separately be worth changing some of the nomenclature around the event? I'm new to forge so I'm not familiar with naming convention but the javadoc around the class and the class/field names themselves seem to suggest that the entity attached to the event is the entity that generates the event. In fact, the entity is always a player entity if it is not null. The player entity appears to correspond to the player that creates the sound event, but since most sound events aren't player created it is usually null. Also, because the event extends `EntityEvent` it inherits an entity field with a generic entity. This could be misleading because the entity will always be a player when present.

sciwhiz12 (changes requested):
As said by @diesieben07, PRs (features, enhancements, bugfixes) must be made for the active version (1.18.x) before they are accepted for the LTS version (1.16.x), or it must be proved that the PR's contents are not applicable to the active version.

In addition, please title your PR properly.

NoOneCanCode:
> As said by @diesieben07, PRs (features, enhancements, bugfixes) must be made for the active version (1.18.x) before they are accepted for the LTS version (1.16.x), or it must be proved that the PR's contents are not applicable to the active version.
> 
> In addition, please title your PR properly.

Just fixed the title. I can submit a PR with similar functionality to 1.18 tonight or tomorrow afternoon but maybe there should be some more discussion about what the actual fix should be, and whether or not the event should include the entity that generates it and not just the player it comes from.

XFactHD:
> The player entity appears to correspond to the player that creates the sound event, but since most sound events aren't player created it is usually null.

This is incorrect in most cases.

There are five places where the event hook is called:

- Two in `ClientWorld`
- One in `ClientPlayerEntity`
- Two in `ServerWorld`

#### `ClientWorld` and `ServerWorld`
The call-sites in `ClientWorld` and `ServerWorld` are very similar with one important distinction: The player that is passed into the method is handled differently.
On the client side, the player passed into the method must match `Minecraft#player` and only then the sound will be played.
On the server side, a packet will be sent to all players except the one passed in.
The reason for this is that it allows an interaction (like switching a lever for example) to play a sound without a side check and therefore avoid unnecessary network packets and lag between player interaction and sound playback on the client of the player that performed the interaction.
The player passed to these methods is in most cases not the sound source.

#### `ClientPlayerEntity`
`Entity#playSound()` seems to be used when the `Entity` is actually the source of the sound. This method is overriden in `ClientPlayerEntity` and fires the event seperately.
The implementations of `playSound()` in `PlayerEntity` and `Entity` both call one of the `World#playSound()` methods.

---

The problem with all this is that most sounds, even when created by an entity, don't go through the `World#playSound()` method with an additional entity argument because that overload is only used for sounds that move with the entity.

If this event were to be reworked, it should be changed to not be an `EntityEvent` because there is no guarantee that an `Entity` that fits is available, it should be renamed to something like `PlaySoundAtPositionEvent` and provide the `World` and coordinates like you are proposing here. Obviously this would be a breaking change and could only happen in a phase where those are accepted. This is of course only my personal opinion.
After this deepdive into the sound code, your proposed changes would be a decent bandaid-fix for this event IMO.

In light of that I have to apologize for my initial reaction, I completely misinterpreted this whole thing.

---

Further discussion on this might be easier on the Forge Discord in the #github-discussions channel.

PR #8335, opened 2021-12-31
[1.18] Patch log4j vulnerabilities

GrahamCampbell:


sciwhiz12:
Hello! 👋 
According to the information available at the [Apache Log4j Security Vulnerabilities page][log4j-vulns-page], we have reasoned out that Forge (and Minecraft as well) are not affected by the vulnerabilities present in Log4j 2.15.0 up to 2.17.0:

 - **CVE-2021-45046**, which was fixed in Log4j *2.16.0*, affects non-default pattern layouts which use a context lookup (`$${ctx:loginId}` for example) where the attacker has control over the ThreadContext Map. 

   As Forge's pattern layouts make no use of context lookups, it is not affected by this vulnerability.

 - **CVE-2021-45105**, which was fixed in Log4j *2.17.0*, also affects non-default pattern layouts which use a context lookup (`$${ctx:loginId}` for example) where the attacker has control over the ThreadContext Map. 

   As said above, Forge's pattern layouts make no use of context lookups, it is not affected by this vulnerability.

 - **CVE-2021-44832**, which fixed in Log4j *2.17.1*, affects environments where the attacker has permission to modify the logging configuration: specifically, to create a JDBC Appender with a data source referencing a JNDI URI.

   As nothing in Forge uses the JDBC Appender, it is not affected by this vulnerability. 

    Additionally, this vulnerability requires the attacker hold permission to modify the logging configuration, which on clients and servers means write access to the affected machine (to substitute their own logging configuration and modify the command line arguments for the server to include such maliciously crafted configuration). 
    Therefore, if this vulnerability is somehow exploited by an attacker, it is indicative of an already present security problem on the affected machines (i.e. "They have bigger fish to fry").

Therefore, we are closing this PR as unneeded since Minecraft Forge and its end users are unaffected by these vulnerabilities. If new information come to light which show that these vulnerabilities (or future vulnerabilities if they do appear) can demonstrably be shown to affect running Forge installations, we will swiftly work towards updating Log4j in affected versions as needed.
Thank you for your contribution!

[log4j-vulns-page]: https://logging.apache.org/log4j/2.x/security.html

PR #8336, opened 2021-12-31
[1.17] Patch log4j vulnerabilities

GrahamCampbell:


sciwhiz12:
Closing this PR for the same reasons as https://github.com/MinecraftForge/MinecraftForge/pull/8335#issuecomment-1003371736, with the additional reason that 1.17.x is unsupported (1.18.x is the active version with 1.16.x as LTS). However, I will note that we will patch old versions when required to resolve critical security vulnerabilities (as was done for CVE-2021-44228 and versions 1.18 down to 1.12).

PR #8337, opened 2021-12-31
[1.18.X] Add events for collecting static and dynamic level geometry

amadornes:
As things stand right now, the rendering hooks provided by Forge are quite minimal, which limits the options natively available. This pull request aims to provide the two most useful level-related geometry collection events.

## Implementation

Two events are provided: `GatherGeometryEvent.ChunkSectionStatic` and `GatherGeometryEvent.LevelDynamic`.

The former is fired every time a chunk section (16x16x16) is baked before any block geometry is added. This event provides a `MultiBufferSource`, a section origin position, a `RenderChunkRegion` which provides access to the blocks, and the visibility graph used to determine chunk section occlusion.

The latter is fired every frame after `Entity` rendering, but ahead of `BlockEntity` rendering. This event also provides a `MultiBufferSource`, as well as the active level renderer, pose stack, camera and partial tick time.

The static chunk geometry event requires a multi-buffer source to be provided by Forge, as the backing buffer container is a `ChunkBufferBuilderPack`. This buffer source ensures that the requested render type can be used for static rendering (see #8331) and initializes the layer's buffer. After firing the event, the layers containing geometry are marked as such so the chunk section knows whether to render or not.

Lastly, a test mod is provided (mod id `geometry_gathering_test`) which statically renders an obsidian block with full brightness at (0, 80, 0) in every chunk, and dynamically renders a stone block 5 meters above the player's head.

## Demo

Here are some images of the obsidian and stone blocks previously mentioned:

obsidian blocks

stone block

## Performance implications

The chunk section geometry-gathering event is fired for as many vertical sections as a chunk has (dependent on world height), for every chunk that needs to be re-rendered. Due to being fired before any blocks have been populated, the time iterating the visited render types should be minimal, and the overall performance impact should not be noticeable, even when re-rendering the entire level.

The level geometry-gathering event is fired once per frame for the current level, and has the same performance implications as the current `RenderLevelLastEvent` - that is, negligible.

## Maintainability

These events shouldn't require any ongoing maintenance.

### DIFF
--- a/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
+++ b/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
@@ -19,6 +19,7 @@
 
 package net.minecraftforge.client;
 
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.mojang.blaze3d.systems.RenderSystem;
 import com.mojang.blaze3d.vertex.*;
@@ -33,6 +34,9 @@ import net.minecraft.client.multiplayer.MultiPlayerGameMode;
 import net.minecraft.client.multiplayer.ServerData;
 import net.minecraft.client.player.AbstractClientPlayer;
 import net.minecraft.client.player.LocalPlayer;
+import net.minecraft.client.renderer.chunk.ChunkRenderDispatcher;
+import net.minecraft.client.renderer.chunk.RenderChunkRegion;
+import net.minecraft.client.renderer.chunk.VisGraph;
 import net.minecraft.locale.Language;
 import net.minecraft.network.Connection;
 import net.minecraft.network.chat.FormattedText;
@@ -137,6 +141,7 @@ import java.util.Random;
 import java.util.Set;
 import java.util.Stack;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
@@ -1043,6 +1048,43 @@ public class ForgeHooksClient
                         ClientTooltipComponent::create
                 ))
                 .toList();
+    }
+
+    public static void onGatherStaticChunkGeometry(
+            ChunkBufferBuilderPack chunkBufferBuilderPack,
+            RenderChunkRegion region,
+            BlockPos.MutableBlockPos origin,
+            VisGraph visibilityGraph,
+            ChunkRenderDispatcher.CompiledChunk compiledChunk,
+            Consumer<BufferBuilder> beginLayer
+    ) {
+        MinecraftForge.EVENT_BUS.post(new GatherGeometryEvent.ChunkSectionStatic(
+                t -> {
+                    var buf = chunkBufferBuilderPack.builder(t);
+                    Preconditions.checkNotNull(buf, "The requested render type is not configured for static rendering: " + t);
+                    if (compiledChunk.hasLayer.add(t)) {
+                        beginLayer.accept(buf);
+                    }
+                    return buf;
+                },
+                region,
+                origin,
+                visibilityGraph
+        ));
+        for (RenderType type : compiledChunk.hasLayer)
+        {
+            var buf = chunkBufferBuilderPack.builder(type);
+            if (buf.getSortState().vertices > 0)
+            {
+                compiledChunk.hasBlocks.add(type);
+                compiledChunk.isCompletelyEmpty = false;
+            }
+        }
+    }
+
+    public static void onGatherDynamicLevelGeometry(MultiBufferSource bufferSource, LevelRenderer levelRenderer, PoseStack poseStack, float partialTicks, Camera camera)
+    {
+        MinecraftForge.EVENT_BUS.post(new GatherGeometryEvent.LevelDynamic(bufferSource, levelRenderer, poseStack, partialTicks, camera));
     }
 
 }
--- a/src/main/java/net/minecraftforge/client/event/GatherGeometryEvent.java
+++ b/src/main/java/net/minecraftforge/client/event/GatherGeometryEvent.java
@@ -1,0 +1,108 @@
+package net.minecraftforge.client.event;
+
+import com.mojang.blaze3d.vertex.PoseStack;
+import net.minecraft.client.Camera;
+import net.minecraft.client.renderer.LevelRenderer;
+import net.minecraft.client.renderer.MultiBufferSource;
+import net.minecraft.client.renderer.chunk.RenderChunkRegion;
+import net.minecraft.client.renderer.chunk.VisGraph;
+import net.minecraft.core.BlockPos;
+import net.minecraftforge.eventbus.api.Event;
+
+/**
+ * Fired on the {@link net.minecraftforge.common.MinecraftForge#EVENT_BUS}
+ * on {@link net.minecraftforge.api.distmarker.Dist#CLIENT} when geometry
+ * is being collected from its respective sources.
+ *
+ * @see ChunkSectionStatic
+ * @see LevelDynamic
+ */
+public abstract class GatherGeometryEvent extends Event
+{
+    private final MultiBufferSource bufferSource;
+
+    protected GatherGeometryEvent(MultiBufferSource bufferSource)
+    {
+        this.bufferSource = bufferSource;
+    }
+
+    public MultiBufferSource getBufferSource()
+    {
+        return bufferSource;
+    }
+
+    /**
+     * Fired when the client is gathering static geometry and
+     * occlusion data for a 16x16x16 section of a chunk.
+     */
+    public static class ChunkSectionStatic extends GatherGeometryEvent
+    {
+        private final RenderChunkRegion region;
+        private final BlockPos.MutableBlockPos origin;
+        private final VisGraph visibilityGraph;
+
+        public ChunkSectionStatic(MultiBufferSource bufferSource, RenderChunkRegion region, BlockPos.MutableBlockPos origin, VisGraph visibilityGraph)
+        {
+            super(bufferSource);
+            this.region = region;
+            this.origin = origin;
+            this.visibilityGraph = visibilityGraph;
+        }
+
+        public RenderChunkRegion getRegion()
+        {
+            return region;
+        }
+
+        public BlockPos.MutableBlockPos getOrigin()
+        {
+            return origin;
+        }
+
+        public VisGraph getVisibilityGraph()
+        {
+            return visibilityGraph;
+        }
+    }
+
+    /**
+     * Fired when the client is gathering dynamic level geometry.
+     */
+    public static class LevelDynamic extends GatherGeometryEvent
+    {
+        private final LevelRenderer levelRenderer;
+        private final PoseStack poseStack;
+        private final float partialTicks;
+        private final Camera camera;
+
+        public LevelDynamic(MultiBufferSource bufferSource, LevelRenderer levelRenderer, PoseStack poseStack,
+                            float partialTicks, Camera camera)
+        {
+            super(bufferSource);
+            this.levelRenderer = levelRenderer;
+            this.camera = camera;
+            this.partialTicks = partialTicks;
+            this.poseStack = poseStack;
+        }
+
+        public LevelRenderer getLevelRenderer()
+        {
+            return levelRenderer;
+        }
+
+        public PoseStack getPoseStack()
+        {
+            return poseStack;
+        }
+
+        public float getPartialTicks()
+        {
+            return partialTicks;
+        }
+
+        public Camera getCamera()
+        {
+            return camera;
+        }
+    }
+}
--- a/src/main/java/net/minecraftforge/client/event/GatherGeometryEvent.java
+++ b/src/main/java/net/minecraftforge/client/event/GatherGeometryEvent.java
@@ -9,6 +9,8 @@ import net.minecraft.client.renderer.chunk.VisGraph;
 import net.minecraft.core.BlockPos;
 import net.minecraftforge.eventbus.api.Event;
 
+import javax.annotation.Nullable;
+
 /**
  * Fired on the {@link net.minecraftforge.common.MinecraftForge#EVENT_BUS}
  * on {@link net.minecraftforge.api.distmarker.Dist#CLIENT} when geometry
@@ -37,11 +39,12 @@ public abstract class GatherGeometryEvent extends Event
      */
     public static class ChunkSectionStatic extends GatherGeometryEvent
     {
+        @Nullable
         private final RenderChunkRegion region;
         private final BlockPos.MutableBlockPos origin;
         private final VisGraph visibilityGraph;
 
-        public ChunkSectionStatic(MultiBufferSource bufferSource, RenderChunkRegion region, BlockPos.MutableBlockPos origin, VisGraph visibilityGraph)
+        public ChunkSectionStatic(MultiBufferSource bufferSource, @Nullable RenderChunkRegion region, BlockPos.MutableBlockPos origin, VisGraph visibilityGraph)
         {
             super(bufferSource);
             this.region = region;
@@ -49,6 +52,7 @@ public abstract class GatherGeometryEvent extends Event
             this.visibilityGraph = visibilityGraph;
         }
 
+        @Nullable
         public RenderChunkRegion getRegion()
         {
             return region;
--- a/src/main/java/net/minecraftforge/client/event/GatherGeometryEvent.java
+++ b/src/main/java/net/minecraftforge/client/event/GatherGeometryEvent.java
@@ -63,6 +63,13 @@ public abstract class GatherGeometryEvent extends Event
             return origin;
         }
 
+        /**
+         * The visibility graph for this 16x16x16 section of a chunk.
+         *
+         * This is used to mark areas of a chunk as opaque, and is used by
+         * Minecraft to determine whether other chunks are being occluded
+         * by this one or not, and thus whether they need to be rendered.
+         */
         public VisGraph getVisibilityGraph()
         {
             return visibilityGraph;
--- a/src/main/resources/META-INF/accesstransformer.cfg
+++ b/src/main/resources/META-INF/accesstransformer.cfg
@@ -1,3 +1,4 @@
+public com.mojang.blaze3d.vertex.BufferBuilder$SortState f_166818_ # vertices
 public net.minecraft.advancements.CriteriaTriggers m_10595_(Lnet/minecraft/advancements/CriterionTrigger;)Lnet/minecraft/advancements/CriterionTrigger; # register
 default net.minecraft.client.KeyMapping f_90817_ # isDown
 public net.minecraft.client.Minecraft f_90987_ # textureManager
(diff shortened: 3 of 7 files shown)


src/main/java/net/minecraftforge/client/event/GatherGeometryEvent.java
```diff
@@ -0,0 +1,112 @@
...
+
+        public BlockPos.MutableBlockPos getOrigin()
+        {
+            return origin;
+        }
+
+        public VisGraph getVisibilityGraph()
+        {
+            return visibilityGraph;
+        }
```
marchermans, line 76: You might want to add some documentation on how to use the graph here

KitsuneAlex:
Hot

XFactHD:
I have played a bit with this PR, specifically the dynamic geometry event. In general it seems pretty solid to me.
The only thing I noticed is that while solid and cutout render types work perfectly fine, the whole thing falls apart with certain combinations of translucent types and graphics modes. Do note that I have only tested the translucent types that use `DefaultVertexFormat.BLOCK` because I can render a block with those.
The results were as follows:

- `RenderType.translucent()`:
  - __Fast, Fancy:__ hides other translucent types (Stained Glass, Water, block outline, etc.) behind it
  - __Fabulous:__ doesn't render at all
- `RenderType.translucentMovingBlock()`:
  - __Fast, Fancy:__ hides other translucent types (Stained Glass, Water, block outline, etc.) behind it
  - __Fabulous:__ works fine
- `RenderType.translucentNoCrumbling()`:
  - __Fast, Fancy:__ works fine
  - __Fabulous:__ doesn't render at all
- `RenderType.beaconBeam(TextureAtlas.LOCATION_BLOCKS, true)`:
  - __Fast, Fancy, Fabulous:__ draws behind other translucent types (Stained Glass, Water, block outline, etc.)
- `RenderType.crumbling(TextureAtlas.LOCATION_BLOCKS)`:
  - __Fast, Fancy, Fabulous:__ works fine
- `RenderType.tripwire()`:
  - __Fast, Fancy:__ hides other translucent types (Stained Glass, Water, block outline, etc.) behind it
  - __Fabulous:__ works fine

---

I am not sure what the best solution for this would be.
Leaving it this way would make this new event similarly akward to use as the current `RenderLevelLastEvent`.
Trying to fire the event again at a later point closer to translucent rendering and adding a flag to it to distinguish between solid and translucent didn't get me any better results either.

---

This is the code snippet I added to the test mod to test this:
```java
private static final Supplier<BlockState> TRANSLUCENT_STATE = Suppliers.memoize(() ->
        Blocks.GREEN_STAINED_GLASS.defaultBlockState()
);

private static final List<Supplier<RenderType>> TYPES = List.of(
        Suppliers.memoize(() -> RenderType.translucent()),
        Suppliers.memoize(() -> RenderType.translucentMovingBlock()),
        Suppliers.memoize(() -> RenderType.translucentNoCrumbling()),
        Suppliers.memoize(() -> RenderType.beaconBeam(TextureAtlas.LOCATION_BLOCKS, true)),
        Suppliers.memoize(() -> RenderType.crumbling(TextureAtlas.LOCATION_BLOCKS)),
        Suppliers.memoize(() -> RenderType.tripwire())
);

@SubscribeEvent
public static void onGatherDynamicLevelGeometryTranslucent(final GatherGeometryEvent.LevelDynamic event)
{
    if (!ENABLED) return;

    HitResult mouseOver = Minecraft.getInstance().hitResult;
    if (mouseOver == null || mouseOver.getType() != HitResult.Type.BLOCK) { return; }

    BlockHitResult target = (BlockHitResult) mouseOver;

    BlockPos renderPos = target.getBlockPos().relative(target.getDirection());
    Vec3 camera = Minecraft.getInstance().gameRenderer.getMainCamera().getPosition();
    Vec3 offset = Vec3.atLowerCornerOf(renderPos).subtract(camera);

    PoseStack poseStack = event.getPoseStack();

    poseStack.pushPose();
    poseStack.translate(offset.x, offset.y, offset.z);

    int slot = Minecraft.getInstance().player.getInventory().selected;
    RenderType renderType = TYPES.get(slot % TYPES.size()).get();

    BlockRenderDispatcher blockRenderer = Minecraft.getInstance().getBlockRenderer();
    blockRenderer.getModelRenderer().renderModel(
            poseStack.last(),
            event.getBufferSource().getBuffer(renderType),
            TRANSLUCENT_STATE.get(),
            blockRenderer.getBlockModel(TRANSLUCENT_STATE.get()),
            1f, 1f, 1f,
            0x0E00E0,
            OverlayTexture.NO_OVERLAY,
            EmptyModelData.INSTANCE
    );

    poseStack.popPose();
}
```

JTK222:
I am probably blind, and only checked the code so far. Surprisingly, I have something very, very similar in 1.16.
However, there are a few differences I see.
This does not seem to enable new RenderTypes properly if they weren't used in a chunk yet.
However, by checking the 1.18 code, this is still required. (as done in the ChunkRenderDispatcher on line 623 and 624)
I can also imagine it causing the render issues XFactHD had.
However if this is done, and I am just blind please ignore me.

malte0811:
The event for static geometry is (in most cases) fired off-thread AFAICT, this should probably be noted in the JavaDoc. This also means that any data used for rendering in this event needs to be stored in a thread-safe manner. For my own use-case (IE wires) this is not a major issue since only a very small part of the data structure needs to be thread-safe, but in general some equivalent of `BlockEntity#getModelData` (i.e. a way to gather data on-thread) might be nice to have.  
(Note: `BE#getModelData` has JavaDoc claiming that it may be called off-thread, but [according to the designer of the `IModelData` system](https://github.com/MinecraftForge/MinecraftForge/commit/06a30e9f23289a45c3c58e83d84c3dee01757e2b#r36351093) this is intentionally not the case.)

XFactHD:
The `ChunkRenderDispatcher.RenderChunk.RebuildTask` where the static geometry event is fired extends `ChunkRenderDispatcher.RenderChunk.ChunkCompileTask`, which in turn has a copy of all `IModelData` from the chunk the task is working on. Allowing access to that data would be as simple as passing `this.modelData` to the event and adding a `GatherGeometryEvent.ChunkSectionStatic#getModelData(BlockPos)` method.

malte0811:
IMO just giving access to the block model data is not enough, it's not unreasonable to want to render things into a chunk (section) without any blocks controlled by your mod. E.g. IE wires can easily pass through sections that don't contain any connectors (or other IE blocks for that matter).

JTK222:
Agreed with malte, most use cases I can think of won't be tied into blocks at all.
Also, I am not sure whether supplying the IModelData in the event would be a good Idea.
The IModelData was added to be used for block models, if it's provided here, many people would likely just use this event, because it's easier.

XFactHD:
So you essentially want something similar to the existing `ModelDataManager` but not bound to a block in the world and instead to a chunk or chunk section, correct?

malte0811:
My (not very thought-through) approach would have been to fire the event on-thread and only have it gather a list of `ISectionRenderer`s (or similar). This would be a functional interface with a `render` method that takes the current fields of the event. The `ISectionRenderer`s would then be invoked off-thread, ideally doing most of the work there. The usage would be something like
```java
var data = gatherDataForChunk(event.getOrigin());
event.addRenderer((buffers, [...]) -> render(buffers, data));
```
But something more structured like what you suggested should also work.

gigaherz (approved):
I'm not 100% sure what the purpose of the dynamic level geometry is, but the rest looks sane to me, and is a feature I have wanted to see for a long time.
(Sorry for not reviewing sooner, it's rare for me these days to have both time and brain energy, but not be too busy worrying about other personal projects)

Note: this review does not take into account all the concerns from comments, only the code itself.

XFactHD:
> I'm not 100% sure what the purpose of the dynamic level geometry is

The dynamic level geometry is intended to replace `RenderLevelLastEvent` because that event is cumbersome to use and is also completely broken when used for translucent rendering while Fabulous mode is active.

src/main/java/net/minecraftforge/client/event/GatherGeometryEvent.java
```diff
@@ -0,0 +1,119 @@
+package net.minecraftforge.client.event;
```
sciwhiz12, line 1: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/client/event/GatherGeometryEvent.java
```diff
@@ -0,0 +1,119 @@
...
+        @Nullable
+        public RenderChunkRegion getRegion()
+        {
+            return region;
+        }
+
+        public BlockPos.MutableBlockPos getOrigin()
+        {
+            return origin;
+        }
```
sciwhiz12, line 64: What origin within the chunk section does this block position represent? Is this intended for consumers to modify, given that we explicitly declare the return type to be `MutableBlockPos`? If not, could this be changed into `BlockPos` instead and noted that callers must call `immutable()` on this object if they intend on storing the position outside the bounds of the event?

src/main/java/net/minecraftforge/client/event/GatherGeometryEvent.java
```diff
@@ -0,0 +1,119 @@
...
+
+        public PoseStack getPoseStack()
+        {
+            return poseStack;
+        }
+
+        public float getPartialTicks()
+        {
+            return partialTicks;
+        }
```
sciwhiz12, line 112: I prefer `partialTick` myself, since this is a fraction of a tick -- a partial tick.

src/test/java/net/minecraftforge/debug/client/rendering/GeometryGatheringTest.java
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
```
sciwhiz12, line 18: Outdated license header, please update this PR to the latest commit of the base branch and run the `updateLicenses`/`licenseFormat` task.

src/test/java/net/minecraftforge/debug/client/rendering/GeometryGatheringTest.java
```diff
@@ -0,0 +1,90 @@
...
+import net.minecraft.world.level.block.state.BlockState;
+import net.minecraftforge.api.distmarker.Dist;
+import net.minecraftforge.client.event.GatherGeometryEvent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
+
+import java.util.function.Supplier;
+
+@Mod(GeometryGatheringTest.MOD_ID)
+public class GeometryGatheringTest
```
sciwhiz12, line 37: Please provide a javadoc for this class which explains what feature this tests, how to test it, and what the expected results are from said test.

src/test/java/net/minecraftforge/debug/client/rendering/GeometryGatheringTest.java
```diff
@@ -0,0 +1,90 @@
...
+        public static void onGatherStaticChunkGeometry(final GatherGeometryEvent.ChunkSectionStatic event)
+        {
+            if (!ENABLED) return;
+            if (event.getOrigin().getY() >> 4 != 80 >> 4) return;
+
+            event.getVisibilityGraph().setOpaque(event.getOrigin().above(80 & 0xF));
+
+            var bufferSource = event.getBufferSource();
+            var poseStack = new PoseStack();
+            poseStack.translate(0, 80 & 0xF, 0);
```
sciwhiz12, line 58: Please explain what `80 & 0xF` does as a comment.

src/test/java/net/minecraftforge/debug/client/rendering/GeometryGatheringTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+            var blockRenderer = Minecraft.getInstance().getBlockRenderer();
+            blockRenderer.getModelRenderer().renderModel(
+                    poseStack.last(), bufferSource.getBuffer(RenderType.solid()),
+                    STATIC_STATE.get(), blockRenderer.getBlockModel(STATIC_STATE.get()),
+                    1f, 1f, 1f, 0x0E00E0, OverlayTexture.NO_OVERLAY
```
sciwhiz12, line 64: To avoid the magic constant for the packed light, I suggest storing as a constant field the computation of the packed light through `LightTexture#pack`.

src/test/java/net/minecraftforge/debug/client/rendering/GeometryGatheringTest.java
```diff
@@ -0,0 +1,90 @@
...
+        private static final Supplier<BlockState> STATIC_STATE = Suppliers.memoize(() -> Blocks.OBSIDIAN.defaultBlockState());
+        private static final Supplier<BlockState> DYNAMIC_STATE = Suppliers.memoize(() -> Blocks.STONE.defaultBlockState());
+
+        @SubscribeEvent
+        public static void onGatherStaticChunkGeometry(final GatherGeometryEvent.ChunkSectionStatic event)
+        {
+            if (!ENABLED) return;
+            if (event.getOrigin().getY() >> 4 != 80 >> 4) return;
+
+            event.getVisibilityGraph().setOpaque(event.getOrigin().above(80 & 0xF));
```
sciwhiz12, line 54: Please explain these bitwise operations in comments, for easier comprehension by future readers of this code.

src/test/java/net/minecraftforge/debug/client/rendering/GeometryGatheringTest.java
```diff
@@ -0,0 +1,90 @@
...
+public class GeometryGatheringTest
+{
+    private static final boolean ENABLED = true;
+    static final String MOD_ID = "geometry_gathering_test";
+
+    @Mod.EventBusSubscriber(modid = MOD_ID, value = Dist.CLIENT, bus = Mod.EventBusSubscriber.Bus.FORGE)
+    private static class ClientHandler
+    {
+        private static final Supplier<BlockState> STATIC_STATE = Suppliers.memoize(() -> Blocks.OBSIDIAN.defaultBlockState());
+        private static final Supplier<BlockState> DYNAMIC_STATE = Suppliers.memoize(() -> Blocks.STONE.defaultBlockState());
```
sciwhiz12, line 46: Why are these states memoized? Since `Block#defaultBlockState` is a field access, there shouldn't be any adverse effects if `Blocks.OBSIDIAN.defaultBlockState()` is plugged directly into the event handler code, right?

src/main/java/net/minecraftforge/client/event/GatherGeometryEvent.java
```diff
@@ -0,0 +1,119 @@
...
+        public VisGraph getVisibilityGraph()
+        {
+            return visibilityGraph;
+        }
+    }
+
+    /**
+     * Fired when the client is gathering dynamic level geometry.
+     */
+    public static class LevelDynamic extends GatherGeometryEvent
```
sciwhiz12, line 82: Would it be worth including in the javadocs what the consumers should expect the state of the pose stack to be in (e.g. the positioning and rotation with respect to the player or the current camera position, whether thirdperson or firstperson)?

src/main/java/net/minecraftforge/client/event/GatherGeometryEvent.java
```diff
@@ -0,0 +1,119 @@
...
+            this.region = region;
+            this.origin = origin;
+            this.visibilityGraph = visibilityGraph;
+        }
+
+        @Nullable
+        public RenderChunkRegion getRegion()
+        {
+            return region;
+        }
```
sciwhiz12, line 59: Why is this `@Nullable`? Looking at the call site for this event, it seems that this will never be `null`, as the line directly before this event is fired is a `null` check on the variable passed to this.

src/main/java/net/minecraftforge/client/event/GatherGeometryEvent.java
```diff
@@ -0,0 +1,119 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Fired on the {@link net.minecraftforge.common.MinecraftForge#EVENT_BUS}
+ * on {@link net.minecraftforge.api.distmarker.Dist#CLIENT} when geometry
+ * is being collected from its respective sources.
+ *
+ * @see ChunkSectionStatic
+ * @see LevelDynamic
+ */
```
sciwhiz12, line 21: For the sake of completeness, please include where this event and subclasses are fired in their javadocs.

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -1045,4 +1050,41 @@ public static List<ClientTooltipComponent> gatherTooltipComponents(ItemStack sta
...
+            RenderChunkRegion region,
+            BlockPos.MutableBlockPos origin,
+            VisGraph visibilityGraph,
+            ChunkRenderDispatcher.CompiledChunk compiledChunk,
+            Consumer<BufferBuilder> beginLayer
+    ) {
+        MinecraftForge.EVENT_BUS.post(new GatherGeometryEvent.ChunkSectionStatic(
+                t -> {
+                    var buf = chunkBufferBuilderPack.builder(t);
+                    Preconditions.checkNotNull(buf, "The requested render type is not configured for static rendering: " + t);
```
sciwhiz12, line 1064: As a point of technicality, the more appropriate precondition here is `checkArgument` (`buf != null`), as it would be inappropriate for a method to fire `NullPointerException` when the render type given to it is not null but an invalid/illegal argument.

src/main/java/net/minecraftforge/client/event/GatherGeometryEvent.java
```diff
@@ -0,0 +1,119 @@
...
+    public MultiBufferSource getBufferSource()
+    {
+        return bufferSource;
+    }
+
+    /**
+     * Fired when the client is gathering static geometry and
+     * occlusion data for a 16x16x16 section of a chunk.
+     */
+    public static class ChunkSectionStatic extends GatherGeometryEvent
```
sciwhiz12, line 40: Perhaps it would be good to overrider `getBufferSource` here and specify in its javadocs that only render types configured for static rendering (as defined currently in `RenderType#chunkBufferLayers()` and in the future by #8337) may be used with the buffer source, and render types not configured for such will cause an exception to be thrown.

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -126,6 +127,9 @@ public net.minecraft.client.renderer.entity.ItemRenderer m_115162_(Lcom/mojang/b
 public net.minecraft.client.renderer.entity.ItemRenderer m_115189_(Lnet/minecraft/client/resources/model/BakedModel;Lnet/minecraft/world/item/ItemStack;IILcom/mojang/blaze3d/vertex/PoseStack;Lcom/mojang/blaze3d/vertex/VertexConsumer;)V # renderModelLists
 public net.minecraft.client.renderer.entity.LivingEntityRenderer m_115326_(Lnet/minecraft/client/renderer/entity/layers/RenderLayer;)Z # addLayer
 public net.minecraft.client.renderer.texture.TextureAtlasSprite m_118415_()I # getFrameCount
+public net.minecraft.client.renderer.chunk.ChunkRenderDispatcher$CompiledChunk f_112749_ # hasBlocks
+public net.minecraft.client.renderer.chunk.ChunkRenderDispatcher$CompiledChunk f_112750_ # hasLayer
+public net.minecraft.client.renderer.chunk.ChunkRenderDispatcher$CompiledChunk f_112751_ # isCompletelyEmpty
```
sciwhiz12, line 132: Please run the `checkATs` task to automatically move these lines to their proper sorted position.

sciwhiz12 (changes requested):
Looks fine to me, pending some testing after the requested changes are made.

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
diesieben07: I disagree. From the NullPointerException Javadocs:
> Applications should throw instances of this class to indicate other illegal uses of the null object

Additionally, the JDK built-in null validation method `Objects.requireNonNull`, which "is designed primarily for doing parameter validation in methods and constructors" also throws `NullPointerException`

sciwhiz12: Respectfully, I disagree with your disagreement. If the value being validated by `checkNotNull` is `t`, then it would be appropriate to throw `NullPointerException`, as the validation checks that the caller's provided value is not a `null` object.

However, the value under validation is `buf`, which is the result of `ChunkBufferBuilderPack#builder` with the `t` parameter passed in as an argument. This is why `NullPointerException` is inappropriate here: a caller which provides a non-`null` argument should not be expecting a `NullPointerException` thrown when their argument is somehow invalid (as they'd expect `IllegalArgumentException`). 

The details of how the `t` parameter is determined to be invalid is an implementation detail[^1] -- in this case, checking if the return value of `ChunkBufferBuilderPack#builder` for the given render type is `null`, which indicates that the render type is not configured for static rendering and therefore invalid. `NullPointerException` leaks this implementation detail to the caller.

[^1]: It is an implementation detail because the way by which the validity of the argument can change without affecting the contract of the method, such as invoking a helper method in another class which returns a `boolean` (and the way by which that method checks the validity of the argument is an implementation detail itself).

diesieben07: I accept your disagreement :smile:

amadornes:
I'll be re-evaluating the implementation as per everyone's comments and port it to 1.19 soon™. I'll close this PR and open a new one once it's ready. I could keep the same one, but it's going to be redone from scratch due to the all changes introduced in this area during the client cleanup/refactor, so I may as well keep them separate and just reference this PR in the other one.

PR #8338, opened 2021-12-31
Cleanup FogRenderer patch

eerussianguy:
So this is from a discussion in discord about what is really just some old leftover code that can be removed.

First is the removal of the `isInfinite` bit of the patch. This was added in #4383, fixing [MC-4647](https://bugs.mojang.com/browse/MC-4647) and [MC-10480](https://bugs.mojang.com/browse/MC-10480), As you can see in the patch, there is a guard against any of the values being zero now that there was not before, added around 1.17. So therefore the forge fix is no longer necessary. I also verified that it is fixed ingame.

Second part is an erroneous // TODO comment in the patch. Per the [discord discussion](https://discord.com/channels/313125603924639766/852298000042164244/925421207690575893) it is likely this was a leftover from the porting process and no longer refers to anything.

Last part is adding a couple call sites for `onFogRender`. Currently the event is only fired if FogDensityEvent is not used and if the type of fog is not water fog. I am unable to verify the github history since the event is from 2014, but I believe it makes more sense if the event is fired for all types of fog.


sciwhiz12 (approved):

eerussianguy:
Going to reopen a different PR since this one is quite broken.

PR #8340, opened 2022-01-01
Clear local variable table on RuntimeEnumExtender transformation

gabizou:
[1.18.x PR](https://github.com/MinecraftForge/MinecraftForge/pull/8502) | **1.16.x backport**

This is a "proper" fix for clearing the state of the LVT in the `create` functions due to an exposure of debugging with Mixins. 

## The Crash

Traditionally, I am always writing mixins for SpongeForge, and lately I discovered a crash that was invariably caused by Mixin's debugging tool (`-Dmixin.debug.export=true`) when a Mixin targeted `net.minecraft.entity.EntityClassification`, it'd cause the export to trigger `ClassNode.accept()` to dump the class, and the LVT for the method modified by `RuntimeEnumExtender` doesn't match the expected LVT (the LVT still had the original LVT for the thrown exception).

The various logs and exports that were related to Mixin troubleshooting: https://gist.github.com/gabizou/beef7061791cdf3fdebb46b810362300

## The reproduction

1. Add a Mixin:
```java
@Mixin(EntityClassification)
public abstract class EntityClassificationMixin {
}
```
2. Run Minecraft with `-Dmixin.debug.export=true`
3. Crash occurs as above, and the exported Mixin code appears as above (but this isn't the final class loaded by ModLauncher)

## The Resulting class
The javap output of the loaded class in both with the mixing above and without the mixing: https://gist.github.com/gabizou/9d4926c4d4eaddac128db73ae770759d

Tagging @Mumfrey to help get any further details that I might've missed.

### DIFF
--- a/src/fmllauncher/java/net/minecraftforge/common/asm/RuntimeEnumExtender.java
+++ b/src/fmllauncher/java/net/minecraftforge/common/asm/RuntimeEnumExtender.java
@@ -1,6 +1,6 @@
 /*
  * Minecraft Forge
- * Copyright (c) 2016-2021.
+ * Copyright (c) 2016-2022.
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -150,6 +150,19 @@ public class RuntimeEnumExtender implements ILaunchPluginService {
 
             mtd.access |= Opcodes.ACC_SYNCHRONIZED;
             mtd.instructions.clear();
+            mtd.localVariables.clear();
+            if (mtd.tryCatchBlocks != null)
+            {
+                mtd.tryCatchBlocks.clear();
+            }
+            if (mtd.visibleLocalVariableAnnotations != null)
+            {
+                mtd.visibleLocalVariableAnnotations.clear();
+            }
+            if (mtd.invisibleLocalVariableAnnotations != null)
+            {
+                mtd.invisibleLocalVariableAnnotations.clear();
+            }
             InstructionAdapter ins = new InstructionAdapter(mtd);
 
             int vars = 0;


Mumfrey:
> Tagging @Mumfrey to help get any further details that I might've missed.

Essentially this happens because the `RuntimeEnumExtender` removes the labels referenced in the LVT (when it clears the method instructions) which normally doesn't end up being a problem because the Label's internal `bytecodeOffset` doesn't get resolved until the method is written out. Since the label doesn't get visited (as it's been removed) this leaves the offsets at `zero` so the LVT is useless but not invalid.

This is a weird side-effect of calling `accept` on the ClassNode, since it causes those internal bytecode offsets to be *resolved* as the labels in the method body are visited. One would expect that calling `accept` just causes the ClassNode to visit the supplied visitor with its current state, but the fact that it causes resolution of the Label offsets actually alters internal state.

It's probably reasonable to argue that calling `accept` isn't expected, but by the same token you wouldn't necessarily expect it (from the contract) to break anything either. I certainly didn't.

Essentially you could argue that ASM should not respect the values of orphaned `Label`s, but I think it's reasonable for ASM to expect that orphaned Labels are a bad practice or undefined behaviour.

The fix proposed by gabi is probably the most sensible, though preserving the original `LabelNode`s and reappending them to the method would also work if there was a desire to preserve the functionality of the LVT.

LexManos:
That is weird indeed, you would assume that ASM just traverses the AST and not try to resolve anything.
But meh, it doens't hurt anything to nuke the LVT as well. Would there be any other attributes we need to nuke?

Mumfrey:
> That is weird indeed, you would assume that ASM just traverses the AST and not try to resolve anything.

It's a reasonable assumption, and it's kind of what ASM does, the main problem is that orphaned `Label`s are already a loose cannon so I don't think we can blame ASM here. Normally the `Label`'s offset will be updated when the code is visited, the issue is caused by the `LabelNode` being removed without the corresponding `Label` being removed.

> But meh, it doens't hurt anything to nuke the LVT as well. Would there be any other attributes we need to nuke?

Strictly speaking `tryCatchBlocks` should be nuked as well but for the methods in question I don't think it matters, though that's probably what the original author thought when they ignored the LVT.

If there's a desire for the code to be archetypal though - in that some other person might copy it thinking it's the right way to do things - then I'd say clear that collection too just for completeness.

gabizou:
> If there's a desire for the code to be archetypal though - in that some other person might copy it thinking it's the right way to do things - then I'd say clear that collection too just for completeness.

Done.

I ended up reading more of the accessible list nodes and cleared out the local variable annotations lists as well as wrapping them with null checks (because honestly I don't know whether null checks here are really required, but I don't know enough of the characteristics of asm)

gabizou:
Just gonna bump again since it's almost a month past now and running into this again trying to verify the byte code output of some things in production :/

TheCurle:
Calling @Mumfrey for the final decision on this.

gabizou:
> Calling @Mumfrey for the final decision on this.

Dunno if it’s really necessary to keep waiting. The problem keeps showing itself when trying to debug mod mixin incompatibilities on Forge….

src/fmllauncher/java/net/minecraftforge/common/asm/RuntimeEnumExtender.java
```diff
@@ -150,6 +150,16 @@ public int processClassWithFlags(final Phase phase, final ClassNode classNode, f
...
+            mtd.localVariables.clear();
+            if (mtd.tryCatchBlocks != null) {
+                mtd.tryCatchBlocks.clear();
+            }
+            if (mtd.visibleLocalVariableAnnotations != null) {
+                mtd.visibleLocalVariableAnnotations.clear();
+            }
+            if (mtd.invisibleLocalVariableAnnotations != null) {
+                mtd.invisibleLocalVariableAnnotations.clear();
+            }
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
            if (mtd.tryCatchBlocks != null)
            {
                mtd.tryCatchBlocks.clear();
            }
            if (mtd.visibleLocalVariableAnnotations != null)
            {
                mtd.visibleLocalVariableAnnotations.clear();
            }
            if (mtd.invisibleLocalVariableAnnotations != null)
            {
                mtd.invisibleLocalVariableAnnotations.clear();
            }
```

sciwhiz12 (changes requested):
Confirmed that the issue exists on 1.16.5-36.2.29 and 1.18.2-40.0.5, and that the issue is fixed by this PR's changes.

---

Our version policy requires that changes (new features or bugfixes) must first be integrated into the Latest version's branch before it will be merged into the LTS version's branch (unless it can be shown that the change is only relevant to the LTS branch). 

In line with this, please update this PR or create a new PR targeting the Latest version (as of writing, `1.18.x`).

---

Please update the license header in the following file(s) by updating the year from `2021` to `2022`:
- `src/fmllauncher/java/net/minecraftforge/common/asm/RuntimeEnumExtender.java`
 
When you do update or create the PR to `1.18.x`, please update the license header in the file(s) to match the new style license headers found in the main Forge code of the `1.18.x` branch. (The PR targeting the `1.16.x` branch should still have the license header year updated as said above.)

gabizou:
Did the following:
- Opened a new [PR for the 1.18.x mainline branch](https://github.com/MinecraftForge/MinecraftForge/pull/8502)
- Updated license year for this PR
- Fixed formatting for brackets

sciwhiz12 (changes requested):
The changes look fine (as I've tested them locally earlier), but this needs to be rebased against the latest changes for 1.16.x (or the latest changes merged) so the repository state has the configs needed by the CI to build it.

sciwhiz12 (approved):
Looks good to me. Per our version policy, this LTS PR is waiting on #8502, its active version counterpart, to be merged before this is assigned for merging.

sciwhiz12 (approved):
Latest version PR is merged.

PR #8341, opened 2022-01-01
[1.18.x] Add datapack-sensitive event for modifying biomes

Commoble:
This is a continuation of #7917 for 1.18, with the resource key tags having been split out into a separate PR in #8251.

## Background Information
* The BiomeLoadingEvent cannot be used to read loaded datapack data to decide how to modify biomes or what to add to them, because vanilla biomes are constructed before datapacks are fully loaded.

This makes modifying biomes in a data-driven manner difficult, as the only part of the forge API where the behavior of the BiomeLoadingEvent can be modified outside of java is in common configs, which aren't great for this purpose and can't really be used for mods to add compat to each other to the extent that datapacks can.

To resolve this, we add a new event that provides an API for modifying biomes after worldgen datapacks have been loaded.

## RegistryAccessLoadedEvent
This event fires after worldgen registries (biomes, configuredfeatures, etc) load from json, but before levelstems/chunkgenerators/biomesources load from json, providing access to the worldgen registries (containing biomes and similar objects loaded from worldgen datapacks). It also provides an API for modifying biomes at this point.

This event also allows mods to retrieve configuredfeatures and similar objects from the RegistryAccess and add them to biomes, meaning that unlike the BiomeLoadingEvent, this event allows mods to define their custom worldgen features entirely in json without needing to create or register those features in java first.

A test mod is provided with this PR; this test mod acts as an example of how to use this new event to add feature jsons to existing biomes, without needing to create them in java or register them to the BuiltinRegistries (it does create them in java for datagen purposes but this is optional). It also performs some modifications via the BiomeLoadingEvent to ensure that that event still works as expected when used alongside the RegistryAccessLoadedEvent.

## Other Things Added
* In order to implement the biome modification API, some extra biome builders were made that create copies of existing biomes to use as a base. These can also be used by mods for datagen purposes.

## Explanation of Patches
In order to fire the RegistryAccessLoadedEvent before dimensions load, we must fire it immediately after RegistryReadOps imports datapacks into a RegistryAccess instance. This works perfectly fine in dedicated servers, but during the singeplayer create world screen, datapacks can be can be imported multiple times into the same RegistryAccess instance (this occurs in several places, particularly when using the Recreate World feature). This causes problems with modifications via the event happening to non-json biomes multiple times.

* Firstly, we add a RegistryReadOps field to RegistryAccess instances, allowing us to cache a RegistryReadOps instance in the registries instance.
* Secondly, we have the RegistryReadOps.createAndLoad(...ResourceManager...) static method redirect to ForgeHooks. This lets us determine whether a fresh RegistryAccess instance is being created, or datapacks are being imported into an existing set of registries.
* Thirdly, in this forge hook, we check if datapacks have already been imported into the given RegistryAccess. If so, we return a previously cached RegistryReadOps instance.
  * Since RegistryAccess is a wrapper around a DynamicOps for parsing either json or nbt, and we must provide a readops wrapped around the correct type when returning early, we also patch a map into RegistryReadOps to be able to cache and return a wrapper of the correct dynamicops (vanilla already does this but just for jsonops, we need this map to be able to cache NBT ops as well).
* Finally, if we allowed a new RegistryReadOps to be created and we are importing datapacks into the dynamic registries, we fire the RegistryAccessLoadedEvent. Since we've previously ensured that datapacks aren't imported into this registries instance if they already have been, the event is not fired more than once for the same set of registry imports. (changing or reorganizing datapacks in the create world gui doesn't cause issues here because vanilla creates a new registries instance with freshly recreated builtin objects before it imports these datapacks here)

## Deprecation of Existing APIs
* BiomeLoadingEvent has been deprecated and marked for future removal by this PR as the dynamic registries loaded event serves the same purpose, but is also sensitive to datapacks. If there are any use cases for the BiomeLoadingEvent that this new event does not address, they should be brought to attention so that the RegistryAccessLoadingEvent can be refactored to support those cases, or the BiomeLoadingEvent un-deprecated should that be infeasible.

## Testing
We must ensure that the RegistryAccessLoadedEvent can read data from jsons, alter json-defined biomes, and be used alongside the BiomeLoadingEvent (the RegistryAccessLoadedEvent fires after the BiomeLoadingEvent and will take precedence for non-additive changes). We must also ensure that changes via the RegistryAccessLoadedEvent do not apply more than once per biome instance.

Three biomes are modified via the RegistryAccessLoadedEvent:
* Windswept Hills, which are also modified in BiomeLoadingEvent to have a black sky and add a black wool feature
* Windswept Forest, which are also overridden via a datapack json to have a red sky
* Snowy Plains, which are not modified except during the RegistryAccessLoadedEvent

The RegistryAccessLoadedEvent adds a red wool feature to these biomes and changes the sky color to green. Expected observations of these biomes:
* Windswept Hills will have a green sky and one red and black wool block per chunk
* Windswept Forest will have a green sky and one red wool block per chunk
* Snowy Plains will have a green sky and one red wool block per chunk

These biomes behave as expected when the test mod is used (including when rejoining or recreating worlds).

## Other Concerns
* This PR doesn't cause any binary-breaking changes
* This PR definalizes many Biome fields. However, it is recommended not to modify biomes outside of the event's provided API for doing so. The event fires before chunk generators and biome sources load, so if mods are only modifying biomes via the offical APIs, mods with custom chunk generators and biome sources (which are loaded during dimension load) can assume that the biomes will not be changed post-dimension-load.
* Previous iterations of this PR also included an API for adding structures to noise settings. There are complexities with noise generators that require additional research and design, and so that API has been left out of this iteration in the interest of providing the biome API in a timely manner.


Kasim115:
Ad my friend I can help you in minecraft 1.18. mod apk

50ap5ud5:
Yes! Thank you Commoble, this will be very useful for any mod that touches world generation.  Hope it gets merged soon!

TelepathicGrunt (approved):

50ap5ud5:
Hi @Commoble, are you still interested in updating this PR? It appears to have some merge conflicts at the moment. This would be really great to have in anticipation for 1.18.2!

Commoble:
> Hi @Commoble, are you still interested in updating this PR? It appears to have some merge conflicts at the moment. This would be really great to have in anticipation for 1.18.2!

Will do that after 1.18.2 as there may be changes that affect this PR

Commoble:
1.18.2 has an implementation change to data loading that substantially affects this PR. Previously, reloadable data loaded first, then dynamic registries, then dimensions. In 1.18.2, dynamic registries and dimensions load, and *then* reloadable data.

This means that a listener to this event would no longer be able to use reloadable data to determine what features or mob spawners to add to which biomes (at the current timing, or in BiomeLoadingEvent's timing, or at any timing before dimensions load), which sort of defeats the purpose of this event. However, dynamic registries are still readable. The ability to add custom dynamic registries is therefore particularly important for the continued usefulness of this event; there are several open PRs such as https://github.com/MinecraftForge/MinecraftForge/pull/8522 to add such a feature.

Commoble:
Closing this as I'm rewriting a completely different impl

PR #8348, opened 2022-01-03
Complete Overhaul of PlantType system

OneLemonyBoi:
Lots of redundant code is removed, now all you need to do to set where a plant can be placed is:

1. Inherit IPlantable if not already inherited
2. Create a PlantType or use a preexisting one
3. Override getPlantType

Hypothetically not breaking, if there is something breaking, lmk


src/main/java/net/minecraftforge/common/IPlantable.java
```diff
@@ -29,18 +29,7 @@
...
-        if (this instanceof FlowerBlock) return PlantType.PLAINS;
-        if (this == Blocks.DEAD_BUSH)      return PlantType.DESERT;
-        if (this == Blocks.LILY_PAD)       return PlantType.WATER;
-        if (this == Blocks.RED_MUSHROOM)   return PlantType.CAVE;
-        if (this == Blocks.BROWN_MUSHROOM) return PlantType.CAVE;
-        if (this == Blocks.NETHER_WART)    return PlantType.NETHER;
-        if (this == Blocks.TALL_GRASS)      return PlantType.PLAINS;
-        return net.minecraftforge.common.PlantType.PLAINS;
-    }
+    PlantType getPlantType(BlockGetter world, BlockPos pos);
```
pupnewfster, line 32: Is there a reason instead of having a default here like there was previously, just with updated values? That would greatly reduce the maintenance overhead of the system.

patches/minecraft/net/minecraft/world/level/block/AttachedStemBlock.java.patch
```diff
@@ -0,0 +1,14 @@
+--- a/net/minecraft/world/level/block/AttachedStemBlock.java
++++ b/net/minecraft/world/level/block/AttachedStemBlock.java
+@@ -39,8 +_,9 @@
+       return !p_48850_.m_60713_(this.f_48831_) && p_48849_ == p_48848_.m_61143_(f_48830_) ? this.f_48831_.m_7161_().m_49966_().m_61124_(StemBlock.f_57013_, Integer.valueOf(7)) : super.m_7417_(p_48848_, p_48849_, p_48850_, p_48851_, p_48852_, p_48853_);
+    }
+ 
+-   protected boolean m_6266_(BlockState p_48863_, BlockGetter p_48864_, BlockPos p_48865_) {
+-      return p_48863_.m_60713_(Blocks.f_50093_);
```
pupnewfster, line 8: Is here a reason for this being removed in so many patches? To represent the method shouldn't be used you could just add a `@Deprecated //Forge: Use x instead` at the base declaration, as that would reduce the size and complexity of a lot of these patches.

sciwhiz12:
Hello! @OneLemonyBoi, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.19.2 (or leave an explanation why this PR's changes are not applicable to 1.19.2). If we do not hear back from you after some time, this PR will be closed for inactivity.

sciwhiz12:
Hello! As per my previous comment, we will be closing this PR for inactivity. If you or another contributor wishes to continue the work of this PR, please create a new PR with a link to this PR for reference.

PR #8349, opened 2022-01-03
Simplify usage of IItemRenderProperties::getArmorModel

KnightMiner:
The current version of `IItemRenderProperties::getArmorModel` has two flaws:
* The return type of the method is `A extends HumanoidModel`, and it is basically impossible to create an instance of `A` leading to [class cast exceptions](https://pastebin.com/aWrTMnwk) with some extensions of the armor layer.
* The method runs before `_default` is set up, and if you plan to render the default model in addition to extra components, this means you have to manually copy over all the setup data to the base model.

The second can be worked around by copying all `ModelPart` from the base model into your Humanoid model extension so they get the proper setup data, but that means you lose potential overrides of the rendering logic (e.g. the model adds in extra model parts).

----

My solution is threefold:
* Move the hook to run after `_default` is set up. This means if you plan to render it directly, minimal extra work is required.
* Copy the relevant properties from `_default` to the returned model as part of a ForgeClientHook to prevent the unsafe casts when calling the copy functions.
* Added `IItemRenderProperties::getBaseArmorModel` which relaxes the hook type to `Model`. This means for the case of wrapping the default model with extra configuration (my usecase), you can skip the redundant initialization and instantiation of unneeded model parts. Fully custom armor mesh can continue to use `IItemRenderProperties::getArmorModel` with the automatic setup described above.

As a note, the solution has one unchecked cast, from `HumanoidModel<?>` to `HumanoidModel<T>` to call `HumanoidModel::copyPropertiesTo`. I have deemed this cast to be safe as the only usages of the generic `T` in `HumanoidModel` are as function parameters, which cannot be called in `copyPropertiesTo` as that method receives no `T` parameter.

----

To test this PR, I have included a custom model layer test mod which adds three items under the test mod `custom_armor_model_test`:
* `red_leggings`: Worn in the pants slot, uses a basic `Model` to wrap the diamond leggings model and apply a red tint
* `enderman_chestplate`: Worn in the chest slot, uses the enderman mesh as the chestplate model to test custom 3D models. The arms will properly move and sway with the player arms and no legs/head will be visible
* `enderman_boots`: Worn in the boot slot, testing the same model as the chestplate. The legs will properly move with the player legs (may need to fly to see them properly). No arms, head, or chest will be visible (unless wearing the chestplate).

Screenshot of test mod (while flying forwards):
image


KnightMiner:
I made some changes based on a [discussion on Discord](https://discord.com/channels/313125603924639766/852298000042164244/930946537910046761) to lessen the burden of devs updating to the new hook:

* The PR now provides two armor model hooks:
    * `getArmorModel` which returns a `HumanoidModel` and will automatically set it up based on properties from the default, similar to the old hook. Small difference in how, as it directly copies over the visibility rather than calling `setVisible`. It does still require an unsafe cast from `HumanoidModel<?>` to `HumanoidModel<T>`, but in inspection of the method that generic should have no impact.
    * `getBaseArmorModel` which returns a `Model`. This hook is much more ideal for the case of rendering the vanilla model with additional added elements, or replacing the vanilla model with an entirely different component (e.g. elytra wings). The default implementation simply calls `getArmorModel` then performs the automatic setup discussed in that hook.
* The PR test mod was updated
    * I realized the original test mod could be accomplished without a custom model with the new hook simply by calling `setVisible` in the `getArmorModel` hook, so I removed that test
    * A new test is `red_leggings` which tints the diamond armor model red using a `Model` wrapper
    * A second new test is `enderman_chestplate` and `enderman_boots` which renders parts of the enderman model using the properties from the default model.

pupnewfster (who approved the PR before the changes) [did not seem to object to the changes](https://discord.com/channels/313125603924639766/852298000042164244/930954437923577896), though if preferred we can mark this PR for review again rather than assigned.

----

As a part of this PR, it might be worth reducing the patches further on `HumanoidArmorLayer`. The method `getArmorModelHook` for instance just redirects to the forge hook, so could easily be inlined. `getArmorResource` also feels a bit like it should be moved to a forge hook instead of being included as a larger patch. I have not made either of these changes as I am unsure if those two hooks were intended to be overridden.

PR #8351, opened 2022-01-03
[1.18] Make it easier to register custom skull blocks models

KnightMiner:
In 1.16, it was possible to add a model to a skull block using an AT on the map in `SkullBlockRenderer`. 1.18 replaced that map with a method returning an `ImmutableMap`, which is called in 3 places in vanilla code to inject the map into a final field (one of those three places is called as a model layer for every living entity, meaning several dozen instances of the map exist). If a skull is missing from the map, it causes a `NullPointerException` in the vanilla layer when looking a the item form or placing it on an entity's head. The block entity renderer would also cause a null pointer exception if you use the vanilla one, though that can be worked around more easily using a custom block entity/block entity renderer.

This PR fixes the problem by adding `EntityRenderersEvent.CreateSkullModels`, an event which is fired during `SkullBlockRenderer.createSkullRenderers` to allow adding your head models into the builder. In addition, this PR adds an AT making `SkullBlockRenderer.SKIN_BY_TYPE` public to allow adding the skin for the skull.

Included in the PR is a test mod that adds a blaze head block/item, similar to the custom sign test. When testing, make sure to wear the skull on your head, place it on an armor stand, and place it as a wall block and a ground block.


Shadows-of-Fire:
By the test mod this works as advertised 
Let me go poke the formatting and make sure that's in order as well.

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -811,6 +814,7 @@ public static TextureAtlasSprite loadTextureAtlasSprite(
 
     private static final Map<ModelLayerLocation, Supplier<LayerDefinition>> layerDefinitions = new HashMap<>();
 
+    /** Call during {@link EntityRenderersEvent.RegisterLayerDefinitions} via the event method */
```
diesieben07: Please deprecate this method instead and move it to the event so that it is impossible to call it outside the event.

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -820,6 +824,19 @@ public static void loadLayerDefinitions(ImmutableMap.Builder<ModelLayerLocation,
         layerDefinitions.forEach((k, v) -> builder.put(k, v.get()));
     }
 
+    private static final Map<SkullBlock.Type, Function<EntityModelSet,SkullModelBase>> skullModelConstructors = new HashMap<>();
+
+    /** Call during {@link EntityRenderersEvent.RegisterLayerDefinitions} via the event method */
+    public static void registerSkullModel(SkullBlock.Type type, Function<EntityModelSet,SkullModelBase> constructor)
+    {
+        skullModelConstructors.put(type, constructor);
+    }
```
diesieben07: Please move this method to the event, so that it is impossible use it outside of the event.
If necessary, make a new event.

diesieben07 (changes requested):
Please also update the license headers in the necessary files to 2022.

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
KnightMiner: In what way do you suggest moving it into the event? Should the static map be a field in the event? Not sure of another good way to grant mutable access to a private variable from another class

LexManos: Yes, the idea is that we limit exactly what can mutate the map and when.
Passing the maps as constructor arguments to the event is the simplest way. May need some rejiggering to where the event is constructed tho.
Main reason is we want to be explicitly clear when these things should be registered, by making it impossible {by non hacky means} to register it at another time. Instead of relying on modders reading the comment.

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
KnightMiner: I ended up just removing all updates to ForgeHooksClient as the hook is entirely self contained in the event now, so I reverted the comment there. If preferred I can patch to deprecate it, but with the PR no longer based on that hook it felt less related.

src/main/java/net/minecraftforge/client/event/EntityRenderersEvent.java
```diff
@@ -144,4 +149,36 @@ public EntityModelSet getEntityModels()
             return entityModels;
         }
     }
+
+    /**
+     * Provides a safe place to register models for the skull block
+     */
+    public static class CreateSkullModels extends EntityRenderersEvent
+    {
+        private final Builder<Type,SkullModelBase> builder;
```
sciwhiz12: Could we please have the import reduced to `ImmutableMap` instead, so this is qualified as `ImmutableMap.Builder`, for readability's sake?

src/main/java/net/minecraftforge/client/event/EntityRenderersEvent.java
```diff
@@ -144,4 +149,36 @@ public EntityModelSet getEntityModels()
...
+
+        public CreateSkullModels(Builder<Type,SkullModelBase> builder, EntityModelSet entityModelSet)
+        {
+            this.builder = builder;
+            this.entityModelSet = entityModelSet;
+        }
+
+        public EntityModelSet getEntityModelSet()
+        {
+            return  entityModelSet;
```
sciwhiz12: Formatting: duplicate space.
```suggestion
            return entityModelSet;
```

src/main/java/net/minecraftforge/client/event/EntityRenderersEvent.java
```diff
@@ -144,4 +149,36 @@ public EntityModelSet getEntityModels()
...
+        }
+
+        /**
+         * Registers the constructor for a skull block with the given {@link SkullBlock.Type}.
+         * These will be inserted into the maps used by the item, entity, and block model renderers at the appropiate time.
+         *
+         * @param type   Unique skull type. Cannot be used to replace vanilla skull models
```
sciwhiz12: Might it be worth clarifying in the javadocs that if one does try to register for already registered skull types (such as the vanilla skulls), then this will fail later when the `ImmutableMap` is built?

src/test/java/net/minecraftforge/debug/block/CustomHeadTest.java
```diff
@@ -0,0 +1,152 @@
...
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.common.Mod.EventBusSubscriber.Bus;
+import net.minecraftforge.fml.event.lifecycle.FMLClientSetupEvent;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.RegistryObject;
+
+@Mod(CustomHeadTest.MODID)
+public class CustomHeadTest
```
sciwhiz12, line 52: Please provide a javadoc for this class which explains what feature this tests, how to test it, and what the expected results are from said test.

sciwhiz12 (changes requested):
Currently, running this PR and scrolling down to the end of the decorations tab triggers a crash. Here is the [crash report](https://github.com/MinecraftForge/MinecraftForge/files/8140640/crash-2022-02-25_19.07.25-client.txt) (uploaded as a file rather than a Gist for posterity), as neither the debug.log nor the latest.log contain the stacktrace of the crash.

KnightMiner:
Seems when I last updated it the patch never got regenerated after I realized the FORGE event bus was insufficient for this hook, corrected it to the proper bus and made the stylistic changes you suggested.

src/test/java/net/minecraftforge/debug/block/CustomHeadTest.java
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        @Override
+        public BlockEntityType<?> getType()
+        {
+            return CUSTOM_SKULL.get();
+        }
+    }
+
+    private enum SkullType implements SkullBlock.Type { BLAZE }
```
sciwhiz12: While I appreciate the attempt to keep this one-value enum to be as small as possible, we should follow the formatting here as usual and put the braces on newlines.

sciwhiz12 (changes requested):
Works from my testing using the provided test mod, just one formatting change and this should be good to go.

KnightMiner:
Made the change

sciwhiz12 (approved):

KnightMiner:
Took a little to get my forge workspace to rebuild, but I tested this after merging on 1.18.2 and everything is working fine, so this is ready for testing on 1.18.2

PR #8352, opened 2022-01-04
Fix clients being unable to deserialize tags for custom registries

pupnewfster:


TheCurle:
For context: #7898 introduced a regression whereby non-optional tags using a custom type (via a registry) could not be deserialized.

Todo for the future: implement a more thorough custom tag test system.

StraleXY:
So will the mod be available for 1.18.1?

PR #8357, opened 2022-01-05
Provide access to the state in BucketPickup#getPickupSound

pupnewfster:
In Mekanism I have a few blocks that support both water logging and lava logging, but without access to the state there is no way to determine which pickup sound should be used as I can't determine if the block is water logged or lava logged. I debated providing access to the level and position but decided against it as fluid states aren't position sensitive, *and* it would have required a larger patch in BucketItem to move getting the pickup sound to above actually picking the fluid up.

### DIFF
--- a/patches/minecraft/net/minecraft/world/item/BucketItem.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/BucketItem.java.patch
@@ -29,6 +29,15 @@
        if (blockhitresult.m_6662_() == HitResult.Type.MISS) {
           return InteractionResultHolder.m_19098_(itemstack);
        } else if (blockhitresult.m_6662_() != HitResult.Type.BLOCK) {
+@@ -56,7 +_,7 @@
+                   ItemStack itemstack1 = bucketpickup.m_142598_(p_40703_, blockpos, blockstate1);
+                   if (!itemstack1.m_41619_()) {
+                      p_40704_.m_36246_(Stats.f_12982_.m_12902_(this));
+-                     bucketpickup.m_142298_().ifPresent((p_150709_) -> {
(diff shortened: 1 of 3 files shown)


cech12:
Is this change compatible to the fill/empty sound of the FluidUtil? As far as I know, the sound is fluid specific and not block state specific. Maybe a better solution would be to change the line in the BucketItem to get the sound out of the fluid inside after picking the fluid up like the FluidUtils does:
https://github.com/MinecraftForge/MinecraftForge/blob/335cb5edc4467a213a01a315e75aad5aa3569f34/src/main/java/net/minecraftforge/fluids/FluidUtil.java#L155
That should be more compatible to modded tanks/buckets that are using the FluidUtils to pickup fluids.

cech12:
Oh, I forgot the Powder Snow Bucket. It has no fluid inside and a gets its sound from the BucketPickup interface. Then a blockstate sensitive method is a good idea. 
But something feels odd to me getting the sound out of the FluidAttributes (in FluidUtils) and ouf of the BucketPickup interface (in BucketItem).

pupnewfster:
FluidUtil is filling an IFluidHandler and likely has no access to the interface. I do agree that it is weird (and in fact in Mekanism I do [this](https://github.com/mekanism/Mekanism/blob/eff44e61a7ccdacf9968942e2ce37347e00938ed/src/main/java/mekanism/common/util/WorldUtils.java#L406-L414) as a fallback in case the optional is empty due to people not bothering to implement it when they implement BucketPickup). But honestly forge's FluidUtils is a relatively big mess in general in terms of implementation, so using what it does or doesn't do as a reference to how vanilla code should be expanded to allow methods to be properly implemented is probably counterproductive.

PR #8358, opened 2022-01-06
Allow items to hide parts of their tooltips by default

diesieben07:
Adds `IForgeItem#getHiddenTooltipParts`, which acts in addition to the `HideFlags` tag set on individual stacks. This allows items to hide parts of their tooltip by default.

See: https://github.com/MinecraftForge/MinecraftForge/pull/7879#issuecomment-1006321680


src/test/java/net/minecraftforge/debug/item/HiddenTooltipPartsTest.java
```diff
@@ -0,0 +1,79 @@
...
+            return ImmutableMultimap.<Attribute, AttributeModifier>builder()
+                    .put(Attributes.ARMOR, MODIFIER)
+                    .build();
+        }
+
+        @NotNull
+        @Override
+        public Set<ItemStack.TooltipPart> getHiddenTooltipParts(@NotNull ItemStack stack)
+        {
+            return Set.of(ItemStack.TooltipPart.MODIFIERS);
```
pupnewfster: Part of me wonders if this should be stored in a variable so that a new set isn't created each time `getTooltipLines` is called (at least in the example/test mod's case as it doesn't actually depend on the data in the stack), similar to how vanilla for attributes on an item cache by default.

diesieben07: Yes, I didn't do that because this is just a test mod so it doesn't really matter. But I can change it.

sciwhiz12: Because the test mod is essentially an example of this feature which may be referenced by modders, I suggest this use an `EnumSet` instead.

diesieben07: `EnumSet` is mutable and especially for 1 and 2 sized sets the immutable sets are more compact.

patches/minecraft/net/minecraft/world/item/ItemStack.java.patch
```diff
@@ -205,6 +205,16 @@
...
     }
  
+@@ -736,7 +_,8 @@
+    }
+ 
+    private int m_41618_() {
+-      return this.m_41782_() && this.f_41590_.m_128425_("HideFlags", 99) ? this.f_41590_.m_128451_("HideFlags") : 0;
++      int stackFlags = this.m_41782_() && this.f_41590_.m_128425_("HideFlags", 99) ? this.f_41590_.m_128451_("HideFlags") : 0;
++      return stackFlags | net.minecraftforge.common.ForgeHooks.getDefaultHideFlags(this);
+    }
```
LexManos, line 214: Personally, I think this should be JUST replacing the 0 value. Because if you set the HideFlags in the NBT then it should be the end-all. They should not be 'in addition to the default'.

Also, I get what you're going for with the thing being a set, however, performance wise I think it'd be better to have the value return a int. So it doesn't need to recalculate the values all the time. And it's more likely that a modder will screw up creating a new set every call then it is they will screw up `return CAN_PLACE.getMask() | DYE.getMask()`

PR #8361, opened 2022-01-06
Fix and improve Ingredient invalidation

### LINKED ISSUE #8359: Race condition / deadlock loading world in `Ingredient` constructor.
Forge version: 39.0.5
Minecraft version: 1.18.1
This was initially reproduced with JEI 9.1.1.48 in a mod dev environment. 
The affected bits of Forge code in this issue are identical to the latest at time of writing (39.0.10)

---

[Forge patches `Ingredient`](https://github.com/MinecraftForge/MinecraftForge/blob/1.18.x/patches/minecraft/net/minecraft/world/item/crafting/Ingredient.java.patch) to add all instances to a `Collections.newSetFromMap(new WeakHashMap<>())`. This is not a thread safe data structure and in particular during, i.e. recipe reloading, ingredients can be created from multiple threads. This has the *potential* to cause a deadlock.

I have reproduced this irregularly (three times now) in a mod development environment, and observed the following conditions:

- During world loading or creation (both have caused this), everything seems to halt, just after JEI finishes logging:

```
[20:54:03] [Render thread/INFO]: Building recipe registry...
[20:54:03] [Render thread/INFO]: Building recipe registry took 13.32 ms
```

- Inspecting via the debugger shows that most threads are parked except for one:
- The thread that is not parked is suck in a stacktrace of `Ingredient.<init>`, somewhere within a `WeakHashMap` method call.

I have written up a [mock example](https://gist.github.com/alcatrazEscapee/531cee8ec0cb5736dfc23399eda31cba), which confirms that this data structure can indeed deadlock under some unpredictable threaded creation of instances. I believe this is what is occurring here as well. I do not have a minimal test case to reproduce it Minecraft.

---

Suggestion: wrap the cached map in a `Collections.synchronizedMap`. I have done this in my mod dev environment via Mixin, and can (anecdotally) say that I have not experienced the issue since doing so.

diesieben07:
As it turns out Ingredient.invalidateAll has not been called at all since 1.13. It was supposed to be fixed by #7953, but that did not fix it.
This PR actually re-adds the call to Ingredient.invalidateAll.

Additionally the invalidation tracking is greatly improved. Instead of keeping a set of all existing Ingredient instances in memory (which has its own problems, see #8359), a global counter is used to enable the ingredient itself to check whether invalidation has to happen. To do this, `checkInvalidation` must be called by the ingredient before it accesses any invalidation-sensitive data. It will call `invalidate` if necessary. Then `markValid` must be called once the invalidation-sensitive data has been computed.

A test mod is included that checks that ingredients that survive across world loads are invalidated.

Fixes #8359


alcatrazEscapee (approved):

src/test/java/net/minecraftforge/debug/world/item/IngredientInvalidationTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+    public IngredientInvalidationTest()
+    {
+        if (!ENABLED)
+            return;
+
+        MinecraftForge.EVENT_BUS.register(ForgeEventListener.class);
+    }
+
+    private static class ForgeEventListener
```
sciwhiz12: Why does this use a full class for a simple event listener, rather than adding a method reference as an event listener?

src/test/java/net/minecraftforge/debug/world/item/IngredientInvalidationTest.java
```diff
@@ -0,0 +1,95 @@
...
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.RegistryObject;
+
+import java.util.stream.Stream;
+
+@Mod(IngredientInvalidationTest.MOD_ID)
+@Mod.EventBusSubscriber(bus = Mod.EventBusSubscriber.Bus.MOD, modid = IngredientInvalidationTest.MOD_ID)
```
sciwhiz12: Why does this class need an `@Mod.EventBusSubscriber` annotation? The event listener is in an inner class (which I also question in another comment), and there are no other `static` methods annotated with `@SubscribeEvent` in this class.

src/test/java/net/minecraftforge/debug/world/item/IngredientInvalidationTest.java
```diff
@@ -0,0 +1,95 @@
...
+public class IngredientInvalidationTest
+{
+    public static final String MOD_ID = "ingredient_invalidation";
+
+    private static final boolean ENABLED = true;
+
+    private static final DeferredRegister<Item> ITEMS = DeferredRegister.create(ForgeRegistries.ITEMS, MOD_ID);
+    private static final RegistryObject<Item> TEST_ITEM = ITEMS.register("test_item", () -> new Item(new Item.Properties()));
```
sciwhiz12: These two fields seem to be unused: the item is unused by anything, and the deferred register is not registered to the mod-specific event bus which renders it and the item useless.

src/test/java/net/minecraftforge/debug/world/item/IngredientInvalidationTest.java
```diff
@@ -0,0 +1,95 @@
...
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.RegistryObject;
+
+import java.util.stream.Stream;
+
+@Mod(IngredientInvalidationTest.MOD_ID)
+@Mod.EventBusSubscriber(bus = Mod.EventBusSubscriber.Bus.MOD, modid = IngredientInvalidationTest.MOD_ID)
+public class IngredientInvalidationTest
```
sciwhiz12, line 39: For the sake of documentation, please add a javadoc to this class explaining what this test is checking for and how to confirm the behavior it is checking works as expected.

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -167,6 +167,7 @@ public ForgeMod()
 
         MinecraftForge.EVENT_BUS.addListener(VillagerTradingManager::loadTrades);
         MinecraftForge.EVENT_BUS.register(MinecraftForge.INTERNAL_HANDLER);
+        MinecraftForge.EVENT_BUS.addListener(this::mappingChanged);
```
sciwhiz12, line 170: Should the license header of this file also be updated with the new year, given that you have modified it?

sciwhiz12 (changes requested):
The changes seem solid to me; most of my concerns are with the included test mod.

diesieben07:
I have addressed the concerns, the test mod issues were a result of me doing a lazy copypasta job.

sciwhiz12 (approved):

TheCurle:
This is causing a very hard to debug hard-crash during worldloading, so I'm going to accelerate this. Thanks for the fix, diesieben.

PR #8363, opened 2022-01-08
Allow adding to Weathering and Waxing maps

bageldotjpg:
Currently, it is impossible to add to WeatheringCopper's `NEXT_BY_BLOCK` map, which makes it quite hard to add new copper blocks that can oxidize. Making a HashBiMap instead makes it a lot easier as modders can just add to the map directly by getting the value in the supplier. While it is possible to make stuff oxidize without adding to this map, it would require a lot more work to get all the quirks functional especially stuff such as lightning reversing, when it can be made easier just by being able to add to this map.

The same kind of map is used in HoneycombItem's `WAXABLES` to determine what copper blocks can be waxed and scraped off, making this a HashBiMap also makes it simpler to modify. 

A test mod is included, registering new copper blocks, that can be observed to be oxidizing over time (though, you may want to turn up the tick speed). You can also confirm that lightning reverses the custom copper's oxidation, and axes can scrape off the oxidation. There is also a test that shows the HoneycombItem's `WAXABLES` map is updated, using dirt and sand. Axe scraping works here as well.


diesieben07:
This can be done without any patches and instead using an AT and then setting the field. Forge should do that and then provide methods to add to them in the MInecraftForge class instead of this huge patch.

bageldotjpg:
Oops, you are right about `WAXABLES` but I am pretty sure `NEXT_BY_BLOCK` can not be AT'd.

diesieben07:
Why not?

bageldotjpg:
Not sure, just did some testing and `WAXABLES` updates and runs just fine when I AT it but `NEXT_BY_BLOCK` does not. It no longer has the red line under it in IntelliJ but running says I cannot assign a value to it because its final, unlike `WAXABLES`.

thelegitdolt:
wow

bageldotjpg:
This has been updated as last discussed in github-discussions, not sure if any other changing needs being done

src/test/java/net/minecraftforge/debug/block/CustomCopperTest.java
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
```
sciwhiz12, line 18: Outdated license header, please update your PR to the latest commit of the base branch and run the `updateLicenses`/`licenseFormat` task.

src/test/java/net/minecraftforge/debug/block/CustomCopperTest.java
```diff
@@ -0,0 +1,73 @@
...
+import net.minecraftforge.eventbus.api.IEventBus;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.event.lifecycle.FMLCommonSetupEvent;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.RegistryObject;
+
+@Mod(CustomCopperTest.MODID)
+public class CustomCopperTest
```
sciwhiz12, line 39: Please provide a javadoc for this class which explains what feature this tests, how to test it, and what the expected results are from said test.

src/test/java/net/minecraftforge/debug/block/CustomCopperTest.java
```diff
@@ -0,0 +1,73 @@
...
+    private static final DeferredRegister<Item> ITEMS = DeferredRegister.create(ForgeRegistries.ITEMS, MODID);
+    public static final RegistryObject<Item> TEST_COPPER_BLOCK_ITEM = ITEMS.register("test_copper_block", () -> new BlockItem(TEST_COPPER_BLOCK.get(), (new Item.Properties())));
+    public static final RegistryObject<Item> TEST_OXIDIZED_COPPER_BLOCK_ITEM = ITEMS.register("test_oxidized_copper_block", () -> new BlockItem(TEST_OXIDIZED_COPPER_BLOCK.get(), (new Item.Properties())));
+
+    public CustomCopperTest()
+    {
+        if (ENABLED)
+        {
+            final IEventBus eventBus = FMLJavaModLoadingContext.get().getModEventBus();
+            MinecraftForge.EVENT_BUS.register(this);
```
sciwhiz12, line 57: As there are no listeners for events fired on the main Forge event bus, this line is unneeded.

sciwhiz12 (changes requested):
The current implementation looks fine to me.

However, this raises the design question of whether giving modders almost full control over the contents of the map is intended, and whether the use of the `BiMap<Block, Block>` is suitable for Forge which prefers data structures store the delegate to a registry object (like a block).

As a possible alternative, we could store two `BiMap<IRegistryDelegate<Block>, IRegistryDelegate<Block>>`s in `ForgeHooks`, and patch `HoneycombItem` and `WeatheringCopper` to add static, synchronized methods which register new blocks to the map and to modify the existing query methods to query the new `BiMap`s in `ForgeHooks` (the ordering would be dependent on whether we wish to allow mods to override regular vanilla behavior).

ChampionAsh5357:
As a quick question, is there a reason why this injects into the map instead of adding an extension method in `IForgeBlock`? This seems like a similar usecase of how Forge handled flammability by the `FireBlock` or burn time for furnaces.

Random832:
I think waxing is better done by adding a ToolAction.

For weathering, it may make more sense to have custom blocks override getNext, and add a defaulted getFirst method that can likewise be called from lightning.

sciwhiz12:
Hello! @bageldotjpg, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.19.2 (or leave an explanation why this PR's changes are not applicable to 1.19.2). If we do not hear back from you after some time, this PR will be closed for inactivity.

sciwhiz12:
Hello! As per my previous comment, we will be closing this PR for inactivity. If you or another contributor wishes to continue the work of this PR, please create a new PR with a link to this PR for reference.

PR #8364, opened 2022-01-08
Create parent directories when creating empty config files

diesieben07:
If a config file name is for a subfolder, Forge will fail to create the config file (because the parent folder does not exist). This leads to [crashes like this](https://forums.minecraftforge.net/topic/107412-forge-server-crashing/?tab=comments#comment-480429).


TheCurle:
This is a straight port of a 1.16 feature that has been confusing modders for a while. 
Confirmed from me, let's wait for the butler.

TheCurle (approved):

diesieben07:
FYI: 1.16 has the same problem.

TheCurle:
Go for it.

PR #8371, opened 2022-01-09
[1.18] Add shears_dig tool action to glow lichen

cech12:
Problem: 
The ToolAction shears_dig is not added to the loot table of glow lichen and modded tools cannot be used to collect them.

Solution:
The MatchTool condition of Glow Lichen was not supported by the ForgeLootTableProvider because it is listed as a condition of a loot entry and is not nested or a pool condition. A small change to the ForgeLootTableProvider and an execution of the data generation were enough to solve the problem.


sciwhiz12 (approved):
Looks :shipit: to me. 

Will note for other reviewers that the other loot table JSON changes are caused by #8184, which fixed serialization of loot functions (by removing an errorneous check, in that it checked the loot _conditions_ list when serializing loot _functions_). Confirmed by running datageneration locally on commit 6683e9e660696e8fbef4d3860846dd267d53d704 and comparing between the files in the PR vs the files as modified in my local environment.

PR #8375, opened 2022-01-09
Missing item tags

XxRexRaptorxX:
adding missing item tags related to copper


pupnewfster:
First off that is not at all how datagen'd resource work you don't just add files in the output it is meant to be datagenerated. Secondly #7891

PR #8376, opened 2022-01-09
[1.18.x] Implement full support for IPv6

PaintNinja:
This PR fixes and enhances the small existing IPv6 patches in Forge, including:
- IPv6 support for Open to LAN and local server discovery
- Dual-stack resolution support for hostnames
- Compressed representation of Open to LAN IPv6 addresses with config option (enabled by default)

Successor of https://github.com/ForgeForce/MinecraftForge/pull/13

### DIFF
--- a/patches/minecraft/net/minecraft/client/server/LanServerDetection.java.patch
+++ b/patches/minecraft/net/minecraft/client/server/LanServerDetection.java.patch
@@ -5,21 +5,26 @@
           this.setUncaughtExceptionHandler(new DefaultUncaughtExceptionHandler(LanServerDetection.f_120082_));
           this.f_120088_ = new MulticastSocket(4445);
 -         this.f_120087_ = InetAddress.getByName("224.0.2.60");
-+         if (net.minecraftforge.network.NetworkConstants.USE_IPV6) {
-+            this.f_120087_ = InetAddress.getByName("FF7E:230::60"); // todo: contact IANA and ask if this is the right equiv and compliant with the IPv6 Multicast Address Space Registry
++         if (net.minecraftforge.network.DualStackUtils.isIPv6(net.minecraftforge.network.DualStackUtils.getLocalAddress())) {
++            this.f_120087_ = InetAddress.getByName("FF75:230::60"); // todo: determine the correct/ideal multicast IPv6 address
 +         } else {
 +            this.f_120087_ = InetAddress.getByName("224.0.2.60");
 +         }
           this.f_120088_.setSoTimeout(5000);
           this.f_120088_.joinGroup(this.f_120087_);
        }
-@@ -88,7 +_,11 @@
+@@ -88,7 +_,16 @@
           String s = LanServerPinger.m_120111_(p_120097_);
           String s1 = LanServerPinger.m_120116_(p_120097_);
           if (s1 != null) {
 -            s1 = p_120098_.getHostAddress() + ":" + s1;
-+            if (net.minecraftforge.network.NetworkConstants.USE_IPV6) {
-+               s1 = "[" + p_120098_.getHostAddress() + "]:" + s1;
++            if (net.minecraftforge.network.DualStackUtils.isIPv6(p_120098_)) {
++               // compress to short form
++               final String compressedIP = p_120098_.getHostAddress();/*
++                       .replaceAll("((?::0\\b){2,}):?(?!\\S*\\b\\1:0\\b)(\\S*)", "::$2")
++                       .replaceFirst("^0::","::");*/
++
++               s1 = "[" + compressedIP + "]:" + s1;
 +            } else {
 +               s1 = p_120098_.getHostAddress() + ":" + s1;
 +            }
--- a/patches/minecraft/net/minecraft/client/server/LanServerDetection.java.patch
+++ b/patches/minecraft/net/minecraft/client/server/LanServerDetection.java.patch
@@ -1,0 +1,28 @@
+--- a/net/minecraft/client/server/LanServerDetection.java
++++ b/net/minecraft/client/server/LanServerDetection.java
+@@ -33,7 +_,11 @@
+          this.setDaemon(true);
+          this.setUncaughtExceptionHandler(new DefaultUncaughtExceptionHandler(LanServerDetection.f_120082_));
+          this.f_120088_ = new MulticastSocket(4445);
+-         this.f_120087_ = InetAddress.getByName("224.0.2.60");
++         if (net.minecraftforge.network.NetworkConstants.USE_IPV6) {
++            this.f_120087_ = InetAddress.getByName("FF7E:230::60"); // todo: contact IANA and ask if this is the right equiv and compliant with the IPv6 Multicast Address Space Registry
++         } else {
++            this.f_120087_ = InetAddress.getByName("224.0.2.60");
++         }
+          this.f_120088_.setSoTimeout(5000);
+          this.f_120088_.joinGroup(this.f_120087_);
+       }
+@@ -88,7 +_,11 @@
+          String s = LanServerPinger.m_120111_(p_120097_);
+          String s1 = LanServerPinger.m_120116_(p_120097_);
+          if (s1 != null) {
+-            s1 = p_120098_.getHostAddress() + ":" + s1;
++            if (net.minecraftforge.network.NetworkConstants.USE_IPV6) {
++               s1 = "[" + p_120098_.getHostAddress() + "]:" + s1;
++            } else {
++               s1 = p_120098_.getHostAddress() + ":" + s1;
++            }
+             boolean flag = false;
+ 
+             for(LanServer lanserver : this.f_120092_) {
--- a/patches/minecraft/net/minecraft/client/server/LanServerDetection.java.patch
+++ b/patches/minecraft/net/minecraft/client/server/LanServerDetection.java.patch
@@ -1,15 +1,11 @@
 --- a/net/minecraft/client/server/LanServerDetection.java
 +++ b/net/minecraft/client/server/LanServerDetection.java
-@@ -33,7 +_,11 @@
+@@ -33,7 +_,7 @@
           this.setDaemon(true);
           this.setUncaughtExceptionHandler(new DefaultUncaughtExceptionHandler(LanServerDetection.f_120082_));
           this.f_120088_ = new MulticastSocket(4445);
 -         this.f_120087_ = InetAddress.getByName("224.0.2.60");
-+         if (net.minecraftforge.network.DualStackUtils.isIPv6(net.minecraftforge.network.DualStackUtils.getLocalAddress())) {
-+            this.f_120087_ = InetAddress.getByName("FF75:230::60");
-+         } else {
-+            this.f_120087_ = InetAddress.getByName("224.0.2.60");
-+         }
++         this.f_120087_ = InetAddress.getByName(LanServerPinger.f_174974_);
           this.f_120088_.setSoTimeout(5000);
           this.f_120088_.joinGroup(this.f_120087_);
        }
@@ -17,7 +13,6 @@
           String s = LanServerPinger.m_120111_(p_120097_);
           String s1 = LanServerPinger.m_120116_(p_120097_);
           if (s1 != null) {
--            s1 = p_120098_.getHostAddress() + ":" + s1;
 +            if (net.minecraftforge.network.DualStackUtils.isIPv6(p_120098_)) {
 +               final String ip;
 +
@@ -29,7 +24,7 @@
 +
 +               s1 = "[" + ip + "]:" + s1;
 +            } else {
-+               s1 = p_120098_.getHostAddress() + ":" + s1;
+             s1 = p_120098_.getHostAddress() + ":" + s1;
 +            }
              boolean flag = false;
  
--- a/patches/minecraft/net/minecraft/client/server/LanServerDetection.java.patch
+++ b/patches/minecraft/net/minecraft/client/server/LanServerDetection.java.patch
@@ -13,7 +13,7 @@
           String s = LanServerPinger.m_120111_(p_120097_);
           String s1 = LanServerPinger.m_120116_(p_120097_);
           if (s1 != null) {
-+            if (net.minecraftforge.network.DualStackUtils.isIPv6(p_120098_)) {
++            if (net.minecraftforge.network.DualStackUtils.checkIPv6(p_120098_)) {
 +               final String ip;
 +
 +               // compress to short form if enabled in config
--- a/patches/minecraft/net/minecraft/client/server/LanServerDetection.java.patch
+++ b/patches/minecraft/net/minecraft/client/server/LanServerDetection.java.patch
@@ -6,25 +6,28 @@
           this.f_120088_ = new MulticastSocket(4445);
 -         this.f_120087_ = InetAddress.getByName("224.0.2.60");
 +         if (net.minecraftforge.network.DualStackUtils.isIPv6(net.minecraftforge.network.DualStackUtils.getLocalAddress())) {
-+            this.f_120087_ = InetAddress.getByName("FF75:230::60"); // todo: determine the correct/ideal multicast IPv6 address
++            this.f_120087_ = InetAddress.getByName("FF75:230::60");
 +         } else {
 +            this.f_120087_ = InetAddress.getByName("224.0.2.60");
 +         }
           this.f_120088_.setSoTimeout(5000);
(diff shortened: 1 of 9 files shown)


PaintNinja:
I've already thoroughly tested this but to help speed up approval by a Triager, here's some instructions on how to test this PR yourself if needed. How thoroughly you test it is up to you - at a certain point it's more about combatting paranoia by covering every edge-case. Note that you do *not* need a native IPv6 Internet connection to fully test this, although having it does make it easier.

# How to test
First, clone the PR and set up forgedev environment as usual

## Open to LAN features

### Forgedev
1. Create a copy of the forge client run configuration and change the username arg from `Dev` to `Dev###`
    - We'll call the original run config "client 1" and the copy "client 2" from now on for simplicity's sake
2. Run both client 1 and client 2
3. Make a new flat world on client 1 and Open to LAN or `/publish`
4. Click the Multiplayer screen on client 2

If it works, client 1 should be discovered by client 2 and show up at the bottom of the list with a compressed IPv6 address and client 2 should be able to join client 1's world.

Repeat the process after toggling the new `compressLanIPv6Addresses` config option in `projects\forge\run\config\forge-client.toml` to test the LAN IPv6 address compression feature.

Repeat the process after disabling IPv4 on your network adapter to confirm that legacy connections still work. Don't forget to re-enable IPv6 support on your network adapter afterwards!

### Userdev
1. Make an MDK and `publishToMavenLocal`
2. Add `mavenLocal()` to repositories in the MDK's build.gradle
3. Now follow the forgedev instructions

### Production clients
1. Build Forge
2. Run installer and install client
3. Start client twice
4. Now follow forgedev instructions 3 and 4

### Interaction between forgedev <-> userdev <-> production clients
For completeness, you may want to test each combination of client types such as forgedev client opening to lan and production client discovering and joining the world.

## Dual-stack resolution feature
If you do not have a native IPv6 Internet connection, you can use a tunnel broker or setup local DNS entries. These steps assume you have already tested the Open To LAN features.

### Forgedev
1. Start the forge server run configuration
2. Determine your computer's public IPv4 and v6 addresses with `ipconfig` or use https://ipv4.icanhazip.com and https://ipv6.icanhazip.com respectively. Press Ctrl + F5 a couple of times if the wrong IP address version is returned by the site.
3. Setup an `A` DNS record that points to your computer's public IPv4 address and an `AAAA` record for your IPv6 one
4. Verify that you can connect over both IP versions with your OS' ping command (e.g. `ping -6 -p 25565 ipv6patch-test.example.com.invalid` for IPv6 and `ping -4 -p 25565 ipv6patch-test.example.com.invalid for IPv4`)
4.1 If you cannot connect over IPv4, temporarily enable DMZ for your test system or setup port forwarding or port triggering
5. Start the client 1 run configuration
6. Click the Multiplayer screen on client 2 and add the server hostname created in step 3

If it works, client 1 should be able to ping and connect to the server. The server console should show a connection over IPv6 if your operating system returned the AAAA record with higher priority (due to OS preference, lower ping or no connection over IPv4)

7. Disable IPv6 connectivity for the server but keep the AAAA record

If it works, client 1 should still be able to ping and connect to the server and the server console should show a connection over IPv4.

8. Disable DMZ or port forwarding/triggering over IPv4 and re-enable IPv6 connectivity for the server, but keep the A record

If it works, client 1 should still be able to ping and connect to the server. The server console should show a connection over IPv6. 

Without this PR, the client will fail to connect as it will always resolve to the IPv4 A record in all circumstances if one is available - regardless of connectivity or OS logic. With this patch, your operating system should return the AAAA record if it knows there's no connectivity over IPv4 but is able to connect over IPv6.

### Userdev
Same as Forgedev, but you can skip steps 2 and 3 as they only need to be performed once

### Production server and client
1. Build Forge
2. Run installer and install the server and client.
3. Follow forgedev instructions 4-8. Production client in this case would be "client 1"

### Interaction between forgedev <-> userdev <-> production servers and clients
For completeness, you may want to test each combination of client and server types such as forgedev client joining a production server, production client joining userdev server and so on.

### Interaction between different Internet connections
If you want to be as thorough as I was, you can run these tests...
- ...on a known IPv4-only Internet connection
- ...on a known IPv6-only Internet connection
- ...on on a known dual-stack (IPv4 and IPv6) Internet connection
- ...and different combinations of each, such as running the server on a known IPv6-only Internet connection and trying to connect to it as a client on a known dual-stack one.

## Interaction between different operating systems
While of course different OSes have different implementations of IPv6 and their own quirks, support has existed for a long time in major operating systems (Windows XP supported IPv6, Windows Vista enabled it by default. Linux supported it without experimental status around version 2.6.12) so it is safe to reasonably assume that interaction between the different major operating systems is sufficiently mature and stable.

It should be noted that with this PR, clients may still use IPv4 if their OS determines that IPv6 connectivity is poor or unavailable. The 

## Finishing up
1. Remove the DNS entries to your computer's public IP addresses created in forgedev instructions step 3 if desired
2. Re-enable IPv6 on your network adapters if you disabled it during testing

patches/minecraft/net/minecraft/client/server/LanServerDetection.java.patch
```diff
@@ -0,0 +1,36 @@
+--- a/net/minecraft/client/server/LanServerDetection.java
++++ b/net/minecraft/client/server/LanServerDetection.java
+@@ -33,7 +_,11 @@
+          this.setDaemon(true);
+          this.setUncaughtExceptionHandler(new DefaultUncaughtExceptionHandler(LanServerDetection.f_120082_));
+          this.f_120088_ = new MulticastSocket(4445);
+-         this.f_120087_ = InetAddress.getByName("224.0.2.60");
++         if (net.minecraftforge.network.DualStackUtils.isIPv6(net.minecraftforge.network.DualStackUtils.getLocalAddress())) {
++            this.f_120087_ = InetAddress.getByName("FF75:230::60");
```
sciwhiz12: This should probably reference the patched constant field in `LanServerPinger`, as I presume the address string was inlined by the compiler from there anyway.

patches/minecraft/net/minecraft/client/server/LanServerDetection.java.patch
```diff
@@ -0,0 +1,36 @@
...
++               // compress to short form if enabled in config
++               if (net.minecraftforge.common.ForgeConfig.CLIENT.compressLanIPv6Addresses.get())
++                  ip = com.google.common.net.InetAddresses.toAddrString(p_120098_);
++               else
++                  ip = p_120098_.getHostAddress();
++
++               s1 = "[" + ip + "]:" + s1;
++            } else {
++               s1 = p_120098_.getHostAddress() + ":" + s1;
++            }
```
sciwhiz12: You can avoid the removal of the line by not indenting it.

patches/minecraft/net/minecraft/client/server/LanServerDetection.java.patch
```diff
@@ -0,0 +1,36 @@
...
+@@ -33,7 +_,11 @@
+          this.setDaemon(true);
+          this.setUncaughtExceptionHandler(new DefaultUncaughtExceptionHandler(LanServerDetection.f_120082_));
+          this.f_120088_ = new MulticastSocket(4445);
+-         this.f_120087_ = InetAddress.getByName("224.0.2.60");
++         if (net.minecraftforge.network.DualStackUtils.isIPv6(net.minecraftforge.network.DualStackUtils.getLocalAddress())) {
++            this.f_120087_ = InetAddress.getByName("FF75:230::60");
++         } else {
++            this.f_120087_ = InetAddress.getByName("224.0.2.60");
++         }
```
sciwhiz12: You can avoid the removal of the line by not indenting it.

patches/minecraft/net/minecraft/client/server/LanServerPinger.java.patch
```diff
@@ -0,0 +1,21 @@
+--- a/net/minecraft/client/server/LanServerPinger.java
++++ b/net/minecraft/client/server/LanServerPinger.java
+@@ -16,7 +_,8 @@
+ public class LanServerPinger extends Thread {
+    private static final AtomicInteger f_120101_ = new AtomicInteger(0);
+    private static final Logger f_120102_ = LogManager.getLogger();
+-   public static final String f_174974_ = "224.0.2.60";
++   public static final String f_174974_ =
++           net.minecraftforge.network.DualStackUtils.isIPv6(net.minecraftforge.network.DualStackUtils.getLocalAddress()) ? "FF75:230::60" : "224.0.2.60";
```
sciwhiz12: Could you please explain here (and if possible, in a comment in all similar spaces) how this IPv6 address was chosen? Also, we'll want to keep this on one line instead, and perhaps extract the condition to a special helper method in `DualStackUtils`.

src/main/java/net/minecraftforge/network/DualStackUtils.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * Checks if an address is an IPv6 one or an IPv4 one, then lets netty know accordingly and returns the result
+     *
+     * @author Paint_Ninja
+     * @param inetAddress The address you want to check
+     * @return true if IPv6, false if IPv4
+     */
+    public static boolean isIPv6(final InetAddress inetAddress) {
```
sciwhiz12: The method's name (and its overload) should be modified to better reflect that this method has side-effects, such as perhaps `checkIPv6`. Also, formatting: braces on newlines.
```suggestion
    public static boolean isIPv6(final InetAddress inetAddress)
    {
```

src/main/java/net/minecraftforge/network/DualStackUtils.java
```diff
@@ -0,0 +1,90 @@
...
+import java.net.InetSocketAddress;
+import java.net.UnknownHostException;
+import java.util.Optional;
+
+public class DualStackUtils {
+    /**
+     * Resolve the address and see if Java and the OS return an IPv6 or IPv4 one, then let netty know
+     * accordingly (it doesn't understand the java.net.preferIPv6Addresses=system property)
+     *
+     * @author Paint_Ninja
```
sciwhiz12: I don't think we accept the use of the `@author` tag in javadocs. @ChampionAsh5357 and @TheCurle, your thoughts?

src/main/java/net/minecraftforge/network/DualStackUtils.java
```diff
@@ -0,0 +1,90 @@
...
+import net.minecraft.util.HttpUtil;
+
+import javax.annotation.Nullable;
+import java.net.Inet6Address;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.UnknownHostException;
+import java.util.Optional;
+
+public class DualStackUtils {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
public class DualStackUtils
{
```

src/main/java/net/minecraftforge/network/DualStackUtils.java
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Resolve the address and see if Java and the OS return an IPv6 or IPv4 one, then let netty know
+     * accordingly (it doesn't understand the java.net.preferIPv6Addresses=system property)
+     *
+     * @author Paint_Ninja
+     * @param hostAddress The address you want to check
+     * @return true if IPv6, false if IPv4
+     */
+    public static boolean isIPv6(final String hostAddress) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    public static boolean isIPv6(final String hostAddress)
    {
```

src/main/java/net/minecraftforge/network/DualStackUtils.java
```diff
@@ -0,0 +1,90 @@
...
+    
+    /**
+     * Get the device's local IP address, taking into account scenarios where the client's network adapter
+     * supports IPv6 and has it enabled but the router's LAN does not.
+     *
+     * @author Paint_Ninja
+     * @return the client's local IP address or null if unable
+     */
+    @Nullable
+    public static InetAddress getLocalAddress() {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    public static InetAddress getLocalAddress()
    {
```

src/main/java/net/minecraftforge/network/DualStackUtils.java
```diff
@@ -0,0 +1,90 @@
...
+        final InetAddress localAddr = new InetSocketAddress(HttpUtil.getAvailablePort()).getAddress();
+        if (localAddr.isAnyLocalAddress()) return localAddr;
+
+        try {
+            return InetAddress.getByName("localhost");
+        } catch (final UnknownHostException e) {
+            return null;
+        }
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
        try
        {
            return InetAddress.getByName("localhost");
        } catch (final UnknownHostException e)
        {
            return null;
        }
```

src/main/java/net/minecraftforge/network/DualStackUtils.java
```diff
@@ -0,0 +1,90 @@
...
+import java.net.Inet6Address;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.UnknownHostException;
+import java.util.Optional;
+
+public class DualStackUtils {
+    /**
+     * Resolve the address and see if Java and the OS return an IPv6 or IPv4 one, then let netty know
+     * accordingly (it doesn't understand the java.net.preferIPv6Addresses=system property)
```
sciwhiz12: Some stylistic javadoc changes:
```suggestion
     * Resolve the address and see if Java and the OS return an IPv6 or IPv4 one, then let Netty know
     * accordingly (as it doesn't understand the {@code java.net.preferIPv6Addresses=system} property).
```

src/main/java/net/minecraftforge/network/DualStackUtils.java
```diff
@@ -0,0 +1,90 @@
...
+                ServerNameResolver.DEFAULT
+                        .resolveAddress(ServerAddress.parseString(hostAddress))
+                        .map(ResolvedServerAddress::asInetSocketAddress);
+
+        if (hostAddr.isPresent()) return isIPv6(hostAddr.get().getAddress());
+        else return false;
+    }
+
+    /**
+     * Checks if an address is an IPv6 one or an IPv4 one, then lets netty know accordingly and returns the result
```
sciwhiz12: Some stylistic javadoc changes:
```suggestion
     * Checks if an address is an IPv6 one or an IPv4 one, lets Netty know accordingly and returns the result.
```

src/main/java/net/minecraftforge/network/DualStackUtils.java
```diff
@@ -0,0 +1,90 @@
...
+            return false;
+        }
+    }
+    
+    /**
+     * Get the device's local IP address, taking into account scenarios where the client's network adapter
+     * supports IPv6 and has it enabled but the router's LAN does not.
+     *
+     * @author Paint_Ninja
+     * @return the client's local IP address or null if unable
```
sciwhiz12: Some stylistic javadoc changes:
```suggestion
     * @return the client's local IP address or {@code null} if unable to determine it
```

sciwhiz12 (changes requested):
My testing shows that this works, though I haven't followed the testing guidelines you gave fully (partly because of the DNS part). Anyone, please also try out this PR yourselves and comment here on your results so we can avoid any potential breaks that might not be caught by our testing.

---

Please update the license headers in the following files by changing the year from `2021` to `2022`:
- `src/main/java/net/minecraftforge/common/ForgeConfig.java`

src/main/java/net/minecraftforge/network/DualStackUtils.java
ChampionAsh5357: I concur. It's a bit pointless if we have the git logs to track contributions.

src/main/java/net/minecraftforge/network/DualStackUtils.java
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
```
TheCurle, line 4: Old style license, please update to SPDX header

PR #8380, opened 2022-01-10
Add patches to enable MobEffects with IDs > 255

diesieben07:
This PR adds patches to enable Minecraft to handle more than 255 mob effects in the registry.

The problematic places in the code are:
- `ClientboundUpdateMobEffectPacket` and `ClientboundRemoveMobEffectPacket` use `writeByte` for the MobEffect ID and are thus limiting the ID range to 0-255. This PR adds a network filter to use a custom network channel instead which does not limit the ID range.
- Saving to NBT in `MobEffectInstance`, `MushroomCow`  and `SuspiciousStewItem`. The respective places are patched to also save (and load) the registry name of the `MobEffect` to NBT with a fallback to using the ID saved by vanilla.
- 
A test mod is included which registers 255 MobEffects that can be used to test the effectiveness of this filter. Without the filter, effects with an ID above 255 do not get sent to the client correctly and cannot persist properly across reloads.


src/test/java/net/minecraftforge/debug/ManyMobEffectsTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+    static
+    {
+        for (int i = 0; i < 255; i++)
+        {
+            final var index = i;
+            MOB_EFFECTS.register("effect_" + i, () -> new MobEffect(MobEffectCategory.NEUTRAL, 0)
+            {
+                @Override
+                public void applyEffectTick(LivingEntity p_19467_, int p_19468_)
```
ChampionAsh5357: Parameters should be given names instead of left as srg.

src/test/java/net/minecraftforge/debug/ManyMobEffectsTest.java
```diff
@@ -0,0 +1,79 @@
...
+                {
+                    if (p_19467_.level.isClientSide)
+                    {
+                        LOGGER.info("Effect Tick for {} on the client", index);
+                    }
+                }
+
+                @Override
+                public boolean isDurationEffectTick(int p_19455_, int p_19456_)
```
ChampionAsh5357: Parameters should be given names instead of left as srg.

src/test/java/net/minecraftforge/debug/ManyMobEffectsTest.java
```diff
@@ -0,0 +1,79 @@
...
+    private static final Logger LOGGER = LogManager.getLogger();
+
+    private static final DeferredRegister<MobEffect> MOB_EFFECTS = DeferredRegister.create(ForgeRegistries.MOB_EFFECTS, MODID);
+
+    static
+    {
+        for (int i = 0; i < 255; i++)
+        {
+            final var index = i;
+            MOB_EFFECTS.register("effect_" + i, () -> new MobEffect(MobEffectCategory.NEUTRAL, 0)
```
ChampionAsh5357: I would choose a different color mainly just to show particle effects on other mobs for an easy way that the effect was applied successfully.

ChampionAsh5357 (changes requested):
Although this removes the limit for packet sending, it will still need to be patched in all other entities and usecases as otherwise the data stored will still be defaulted to the original 255.

However, that's just something for the next PR, this looks fine besides from a few text changes.

diesieben07:
I've updated the PR to add patches for the places that store the byte-ID in NBT as well.

ChampionAsh5357 (approved):

diesieben07:
Updated to 1.18.2, removed custom packet because vanilla packet now transmits VarInt.

sciwhiz12 (changes requested):
Please update the license headers in the following file(s):
- `src/test/java/net/minecraftforge/debug/ManyMobEffectsTest.java`

ChampionAsh5357 (approved):

sciwhiz12 (approved):

PR #8381, opened 2022-01-10
Replace test source set's `mods.toml` with custom mod locator

sciwhiz12:
One point of friction that contributors to Forge have is the test sourceset's `mods.toml`. When a PR adds a test mod, a declaration must be inserted in this file to declare the test mod to Forge, so it loads correctly. However, when a PR is merged, the added mod declaration creates a merge conflict in all other PRs which add a test mod.

This friction introduces affects both contributors and maintainers: contributors have to constantly fix the simple merge conflict in the test source set's `mods.toml` after each merged PR with a test mod, and maintainers have to either fix the merge conflict themselves locally or ask contributors to fix the merge conflict (which delays the merging of the PR).

This PR aims to solve that by completely replacing the test source set's `mods.toml` with a custom `IModLocator`, which scans the test source set for `@Mod`-annotated classes and automatically provides the required information (in what is essentially an in-memory TOML file) to FML. By using the presence of the `@Mod` annotation to declare a test mod, there is no longer any need to insert a change in a file which will conflict with other PRs, and there is now one less step when adding a test mod: you only need to declare a test mod with `@Mod`, and the locator will find it automatically.

### Implementation details

The `TestModLocator` is the main class added by this PR, which implements a mod locator and draws inspiration from `MinecraftLocator`'s code for generating the TOML config on-the-fly. 

The sources where the locator finds the test mods from is configured by a combination of the existing `MOD_CLASSES` environment variable and a new system property `forge.test.id`. This new system property is a 'mod ID' referenced in `MOD_CLASSES`. The value of this property should match the name of the mod config in the run config:

```gradle
property 'forge.test.id', 'modName'
mods {
    modName { sources sourceSets.test }
}
```

The tests sources are excluded from the paths in `MinecraftLocator`, to avoid a modloading warning when it fails to locate a `mods.toml` for the test sources (as that is generated on-the-fly by `TestModLocator`).

Currently, it searches for the `@Mod` annotation provided by the FML Java language provider and uses the mod ID value in that annotation. It is possible to construct a custom annotation for the test sourceset to allow specifying more data for the test mod, for inclusion in its mod declaration in the generated TOML file.

The mod locator is activated if both the above mentioned environment variable and system property is present; otherwise, it will do nothing and be silent.


Technici4n:
Another way to reduce conflicts: instead of appending test mods, keep them sorted. :P

sciwhiz12:
That would still leave the possibility of two PRs conflicting if their tests mods so happen to be in the same sorted position (however unlikely that may be). This solution outright removes that possibility by **eliminating** the `mods.toml` file altogether. :shipit:

Will also note that @TheCurle independently privately implemented this on her private working tree for the tests source set overhaul, which is why her review is requested here. I made this PR in the hopes of it getting in earlier than the test source set overhaul, to hopefully lessen the burden on contributors now while we go around assigning PRs before a merging pass happens.

LexManos:
I will need to talk to Curle about this more. Specifically I want this done when we do the main test mod refactor and codeify the standard structure of a test mod/packages.
The other issue I want to be VERY clear about on this one. Is this MUST be ForgeDev time only, It can not exist in Userdev or Runtime.

sciwhiz12:
The core team has decided that this PR will be indefinitely delayed until there is a proper standardization of the test mods.

TheCurle:
Standardization of the test mods has been achieved, and a new testing framework has been implemented in the 1.19 pre-releases and will be included with the release.

With that, we no longer have a need for this, so it will be closed.

PR #8383, opened 2022-01-10
Fix RegistryObject not working if created after registry events

### LINKED ISSUE #8339: RegistryObject cannot retrieve registry objects if created after registries load
**Minecraft Version:** 1.18.1

**Forge Version:** 39.0.9

**Logs:** https://gist.github.com/Commoble/a09deca41db8a30b3445ef9b21d78e68
(I was in the middle of some forge dev and needed to whip up a quick example of RegistryObjects crashing, so I added this to the start of ForgeMod's common setup event)

 ```java
     public void preInit(FMLCommonSetupEvent evt) // occurs after registration
    {
        RegistryObject<Block> dirtObj = RegistryObject.of(new ResourceLocation("minecraft:dirt"), ForgeRegistries.BLOCKS);
        Block dirt = dirtObj.get(); // NPEs
    }
````

**Steps to Reproduce:**
 1. Create a RegistryObject after standard registries have concluded using RegistryObject#of
 2. Call RegistryObject#get
 3. NPE

**Description of issue:**
RegistryObjects, rather than using conventional lazy evaluation or memoization, are created by registering a callback to the ObjectHolderRegistry, telling forge to populate the registry object when registration is performed. This works perfectly well when the RegistryObject is created before registration occurs, but causes NPEs when the RegistryObject is created post-registration.

### PR DESCRIPTION
**Pull request by diesieben07**
Fixes #8339 by checking at construction time if the registry entry already exists in the registry.

### CODE REVIEW

File: src/main/java/net/minecraftforge/registries/RegistryObject.java [hunk 1]
```diff
@@ -102,6 +103,12 @@ public void accept(Predicate<ResourceLocation> pred)
                     RegistryObject.this.value = registry.containsKey(RegistryObject.this.name) ? (T)registry.getValue(RegistryObject.this.name) : null;
             }
         });
+        IForgeRegistry<V> registry = RegistryManager.ACTIVE.getRegistry(baseType);
+        // allow registry to be null, this might be for a custom registry that does not exist yet
```
pupnewfster, line 107: Is this actually needed? I feel like it potentially should fail given custom registries are added/registered before any registry events have fired. That means if we are *after* registry events the registry shouldn't be null unless it is missing/invalid? (See a few lines above this where the registry is null checked.)

Or is this related to it trying to instantly set it and if it is null that means it was created before which currently doesn't have issues as it doesn't run any code at that time?

diesieben07: Yes it is needed. Consider the following setup for a custom registry:
```java
private static final DeferredRegister<Custom> CUSTOMS = DeferredRegister.create(Custom.class, MODID);
private static final RegistryObject<Custom> CUSTOM = CUSTOMS.register("test", () -> new Custom());
```
The 2nd line (indirectly) calls the `RegistryObject` constructor from a static initializer, while the custom registry is nowhere to be found yet. It is yet to be created by the `DeferredRegister` once `RegistryEvent.NewRegistry` fires.

PR #8387, opened 2022-01-11
Fix LivingDropsEvent not having all drops for foxes

### LINKED ISSUE #8019: spawnDrops of FoxEntity calls the forge hooks getLootingLevel and onLivingDrops too late
**Minecraft Version:** 1.16.5

**Forge Version:** 36.2.2 

(and probably still present in 1.17.1 versions, haven't tested yet)

**Steps to Reproduce:**
1: subscribe an event for LivingDropsEvent with a filter for only foxes
```
@Mod.EventBusSubscriber
public class SomeClassThatHandlesEvents {

    @SubscribeEvent
    public static void testingLivingDropsEventForFoxes(LivingDropsEvent event) {
        if (event.getEntity() instanceof FoxEntity) LogManager.getLogger().info("Killed fox dropped: " + event.getDrops());
    }
}
```
2: give an item to a fox/spawn a fox with an item
3: kill the fox
4: the console should print: `[INFO]: Killed fox dropped: []`, expected output was something like `[INFO]: Killed fox dropped: EntityItem['Dimension', l='ServerLevel[|World Name]', x=0.00, y=0.00, z=0.00]`

**Description of issue:**
the issue comes from the way `spawnDrops` in `FoxEntity` was overwritten:
```
protected void spawnDrops(DamageSource damageSourceIn) {
      ItemStack itemstack = this.getItemStackFromSlot(EquipmentSlotType.MAINHAND);
      if (!itemstack.isEmpty()) {
         this.entityDropItem(itemstack);
         this.setItemStackToSlot(EquipmentSlotType.MAINHAND, ItemStack.EMPTY);
      }

      super.spawnDrops(damageSourceIn);
   }
```
because the drop of the main hand (mouth of the fox) is handled before the `super.spawnDrops`, when the event arrives in the moder's hands, the drops are already empty and can't be processed.

I'm sorry to say this, but many foxes where hurt in the making of this issue, don't sue me, they are pixels that refused to behave with my items.

diesieben07:
Fixes #8019


patches/minecraft/net/minecraft/world/entity/animal/Fox.java.patch
```diff
@@ -1,5 +1,22 @@
 --- a/net/minecraft/world/entity/animal/Fox.java
 +++ b/net/minecraft/world/entity/animal/Fox.java
+@@ -646,13 +_,15 @@
+    }
+ 
+    protected void m_6668_(DamageSource p_28536_) {
++      super.m_6668_(p_28536_);
++   }
++   protected void m_5907_() {
```
sciwhiz12: This added method should have a comment indicating this is patched in by Forge (and possibly with a link to the relevant issue).

sciwhiz12 (changes requested):
Looks fine to me, though I wonder why Mojang left this here. Just needs a small comment indicating the change for the future.

patches/minecraft/net/minecraft/world/entity/animal/Fox.java.patch
```diff
@@ -1,5 +1,22 @@
 --- a/net/minecraft/world/entity/animal/Fox.java
 +++ b/net/minecraft/world/entity/animal/Fox.java
+@@ -646,13 +_,15 @@
+    }
+ 
+    protected void m_6668_(DamageSource p_28536_) {
++      super.m_6668_(p_28536_);
++   }
++   protected void m_5907_() { // Forge: move extra drops to dropEquipment to allow them to be captured by LivingDropsEvent
+       ItemStack itemstack = this.m_6844_(EquipmentSlot.MAINHAND);
```
sciwhiz12: To follow the other overrides of `dropEquipment`, this should call the superclass method as well.

sciwhiz12 (changes requested):
Just noticed this small detail while doing a final review.

sciwhiz12 (approved):

PR #8115, opened 2021-09-22
Re-add hook for LivingExperienceDropEvent, fixes #7900

AterAnimAvis:
Re-adds the hook that got missed during the 1.17 port.

Closes #7900

Testing:
----- 
- Enable Test Mod
- Join World
- Spawn a `pig` and kill it
- Verify the logged message
- Verify via `/xp query points` that the received xp was double the reported original.

PR #8388, opened 2022-01-11
Re-added the patch for LivingExperienceDropEvent and enable it for dragons

### LINKED ISSUE #7900: [1.17.1] LivingExperienceDropEvent is not fired (no event injected)
**Minecraft Version:** 1.17.1

**Forge Version:** 37.0.1

**Logs:** N/A

**Steps to Reproduce:**
 1. Utilize the LivingExperienceDropEvent
 2. It doesn't get fired

**Description of issue:**

LivingExperienceDropEvent is not fired at all. Analyzing Forge's code, in 1.16 you can trace back from the event to ForgeEventFactory#getExperienceDrop, and find usages from there to bring you to LivingEntity#dropExperience wherein you can find the method being called. In 1.17, this method call is missing entirely.

diesieben07:
Reimplemented the patch for `LivingExperienceDropEvent` in `LivingEntity`. This is an updated version of #8115.

Additionally add a new patch for `EnderDragon` to also fire the event when experience is spawned at the end of the dragon fight.
To do this, a field is added that keeps `lastHurtByPlayer` for longer than the normal 100 ticks, so that the last attacking player can still be given with the event. A note is added, that the attacking player may be `null` in the event, in case they have left the game.

Fixes #7900


patches/minecraft/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java.patch
```diff
@@ -1,5 +1,16 @@
 --- a/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
 +++ b/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
+@@ -150,7 +_,10 @@
+       return adouble;
+    }
+ 
++   @javax.annotation.Nullable private Player unlimitedLastHurtByPlayer = null;
+    public void m_8107_() {
++      if (this.f_20888_ != null) this.unlimitedLastHurtByPlayer = f_20888_;
++      if (this.unlimitedLastHurtByPlayer != null && this.unlimitedLastHurtByPlayer.m_146910_()) this.unlimitedLastHurtByPlayer = null;
```
ichttt, line 11: Could you maybe add a comment here explaining why we need to capture the player here? I know you already did this in the PR desc, but for future reference a code comment wouldn't hurt.

ichttt (commented):
One small nitpick, otherwise looking good

ichttt (approved):

PR #8389, opened 2022-01-13
Introduce an attribute for step height

williewillus:
Very similar to the reach distance attribute. Basically this is a primitive field modders like to tweak a lot, and it's tricky to make sure you apply/remove things at the right time lest you get into a bad state. Perfect use for attributes, something Mojang chronically forgets exists.

Adds the new attribute, and new method that queries the combined old field + attribute value. Writes to the old field are kept while all reads of the old field are redirected to the new method.

Test: https://gfycat.com/compassionatehonoredlhasaapso
I show what happens with three helmets, with +1, +2, and +5 step height. Note that the default player step height is 0.6, so with the +1 helmet I can step up 1.6 blocks and clear the grass + slab.

### DIFF
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeEntity.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeEntity.java
@@ -8,6 +8,7 @@ package net.minecraftforge.common.extensions;
 import java.util.Collection;
 import javax.annotation.Nullable;
 
+import net.minecraft.world.entity.LivingEntity;
 import net.minecraft.world.entity.boss.enderdragon.EnderDragon;
 import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.block.state.BlockState;
@@ -19,6 +20,7 @@ import net.minecraft.world.item.ItemStack;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.core.BlockPos;
 import net.minecraft.world.phys.HitResult;
+import net.minecraftforge.common.ForgeMod;
 import net.minecraftforge.common.ForgeSpawnEggItem;
 import net.minecraftforge.common.capabilities.ICapabilitySerializable;
 import net.minecraftforge.entity.PartEntity;
@@ -188,5 +190,23 @@ public interface IForgeEntity extends ICapabilitySerializable<CompoundTag>
     default PartEntity<?>[] getParts()
     {
         return null;
+    }
+
+    /**
+     * {@return Return the height in blocks the Entity can step up without needing to jump}
+     * This is the sum of vanilla's {@link Entity#maxUpStep} field and the current value
+     * of the {@link net.minecraftforge.common.ForgeMod#STEP_HEIGHT_ADDITION} attribute,
+     * (if this Entity is a {@link LivingEntity}), clamped at 0.
+     */
+    default float getStepHeight()
+    {
+        var vanillaStep = self().maxUpStep;
+        if (self() instanceof LivingEntity living)
+        {
+            return (float) Math.max(0, vanillaStep + living.getAttributeValue(ForgeMod.STEP_HEIGHT_ADDITION.get()));
+        } else
+        {
+            return vanillaStep;
+        }
     }
 }
--- a/patches/minecraft/net/minecraft/world/entity/Entity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/Entity.java.patch
@@ -17,6 +17,14 @@
     private final EntityType<?> f_19847_;
     private int f_19848_ = f_19843_.incrementAndGet();
     public boolean f_19850_;
+@@ -175,6 +_,7 @@
+    public double f_19790_;
+    public double f_19791_;
+    public double f_19792_;
++   @Deprecated // Forge - see IForgeEntity#getStepHeight
+    public float f_19793_;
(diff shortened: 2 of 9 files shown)


Random832:
If multiple mods use this with the design intent of an item that grants "step assist" [+0.4 or perhaps +0.5], will these interact in a way that lets you use more than one and cause the player to be able to step up multiple blocks? Is there a way for modders to determine whether their modifier applies if another one is already in effect [maybe some sort of extension to attribute modifiers, and possibly determine what order the modifiers are applied in since any such logic will be sensitive to this]

williewillus:
> will these interact in a way that lets you use more than one and cause the player to be able to step up multiple blocks

yes, that's the point of using the attribute system over the old way, where mods just clobber each other writing to the same field

> Is there a way for modders to determine whether their modifier applies if another one is already in effect [maybe some sort of extension to attribute modifiers, and possibly determine what order the modifiers are applied in since any such logic will be sensitive to this]

Outside of what methods vanilla makes available, no, but as you noted this is more a general feature request for the attribute system than a specific thing this diff needs to address

diesieben07:
A possible suggestion:
- Introduce `IForgeLivingEntity#onAttributeModified` and call it from `AttributeMap#onAttributeModified` (patch `AttributeMap` to store the `LivingEntity` optionally)
- From `onAttributeModified` set `maxUpStep` according to the attribute value. Then all the patches could be saved.

Not sure if this is really viable, but it sounds like a better solution in my head.

williewillus:
I'm not sure what you mean -- all the patches in this PR are for vanilla code to read the attribute instead of the old field. I don't see a way to avoid them since redirecting reads of that field is the purpose of the PR. Writing to the old field by vanilla code is still fine (but discouraged).

diesieben07:
If you keep the field in sync with the attribute then you don't need to patch the reads.

williewillus:
Then you need to patch all the raw writes vanilla does to the field, of which there's many as well.

src/main/java/net/minecraftforge/common/extensions/IForgeEntity.java
```diff
@@ -200,4 +202,19 @@ default boolean isMultipartEntity()
...
+     *  (if this Entity is a {@link LivingEntity}).
+     */
+    default float getStepHeight() {
+        var vanillaStep = self().maxUpStep;
+        if (self() instanceof LivingEntity living) {
+            return (float) (vanillaStep + living.getAttributeValue(ForgeMod.STEP_HEIGHT.get()));
+        } else {
+            return vanillaStep;
+        }
+    }
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    default float getStepHeight()
    {
        var vanillaStep = self().maxUpStep;
        if (self() instanceof LivingEntity living)
        {
            return (float) (vanillaStep + living.getAttributeValue(ForgeMod.STEP_HEIGHT.get()));
        } else
        {
            return vanillaStep;
        }
    }
```

src/main/java/net/minecraftforge/common/extensions/IForgeEntity.java
```diff
@@ -200,4 +202,19 @@ default boolean isMultipartEntity()
     {
         return null;
     }
+
+    /**
+     * Return the height in blocks the Entity can step up without needing to jump
+     * @return The sum of vanilla's {@code maxUpStep} field and the current value
+     *  of the {@link net.minecraftforge.common.ForgeMod#STEP_HEIGHT} attribute,
+     *  (if this Entity is a {@link LivingEntity}).
```
sciwhiz12: Suggestion for improving the javadocs:
```suggestion
     * {@return the height in blocks that the Entity can step up without needing to jump} This is the sum of 
     * the entity's {@link Entity#maxUpStep} and the value of the of the 
     * {@link net.minecraftforge.common.ForgeMod#STEP_HEIGHT} attribute (if this is a {@link LivingEntity}).
```

sciwhiz12 (changes requested):
Locally updating the PR and testing shows that it works as expected. 

However, I do have a concern with how the attribute is expressed right now: because the attribute is merely added to the default step height (`maxUpStep`) of the entity, the base value of the attribute is 0 rather than the intuitive value of whatever the entity's `maxUpStep` is.

We should find a solution so the attribute's base value is correctly synchronized with the entity's `maxUpStep` (such that, for example, checking the player's base attribute value correctly reports `0.5`). An alternative is to rename the attribute to something like `step_height_addition` to correctly reflect that it merely adds to the existing step height rather than representing what it actually is.

williewillus:
Sure, I can rename it to have `addition` in the name. I think trying to shoehorn vanilla's defaults in to the attribute base value isn't worth the trouble.

Alpvax:
I personally would prefer to have it as the entire step height, rather than an addition. Otherwise there is no way to easily double the height, you have to check what the base value is and add it into any calculations you perform.
I haven't looked into the source though, so maybe that is not reasonable to maintain.

TheCurle (approved):
Seems like this is a relatively easy to maintain addition that will save on a lot of unnecessary adjustments to the same field, 

Concerns about requiring calculations to be performed on the addition value rather than absolute are reciprocated, but in the absence of collaboration, i think we can pull this now.

Thanks for the contribution, williewillus.

PR #8391, opened 2022-01-13
feat(api): add PlayerEvent.SignChangeEvent and relevant test

danorris709:
Adds the SignChangeEvent to the PlayerEvent class and implements it in the minecraft ServerGamePacketListenerImpl class.

**Proof of test succeeding:**
Test Success
Test Success 2

**Reasoning:**
Adding this event allows for mods such as [chest shop](https://www.spigotmc.org/resources/chestshop.51856/) and grief prevention mods.


Rasgnarok:
Following, cheers for the PR @danorris709.

patches/minecraft/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
```diff
@@ -61,6 +61,21 @@
...
+ 
+    public void m_5527_(ServerboundSignUpdatePacket p_9921_) {
+-      List<String> list = Stream.of(p_9921_.m_134663_()).map(ChatFormatting::m_126649_).collect(Collectors.toList());
++      List<String> list = net.minecraftforge.event.ForgeEventFactory.fireSignChange(this.f_9743_, p_9921_.m_134660_(), p_9921_.m_134663_());
++
++      for (int i = 0; i < list.size(); i++) {
++         String strippedLine = ChatFormatting.m_126649_(list.get(i));
++         list.set(i, strippedLine);
++      }
++
```
diesieben07: This code should be in the forge hook, not in the patch.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -19,74 +19,50 @@
 
 package net.minecraftforge.event;
 
-import java.io.File;
```
diesieben07: Please do not reorganize imports.

diesieben07 (changes requested):
- Please fix the formatting: braces should be on new lines.
- Please fix the license headers. It should have be updated to the current year for any file that you change.

src/test/resources/META-INF/mods.toml
```diff
@@ -176,7 +176,8 @@ license="LGPL v2.1"
     modId="capabilities_test"
 [[mods]]
     modId="custom_particle_type_test"
+[[mods]]
+    modId="sign_changed_event_test"
 [[mods]]
     modId="ingredient_invalidation"
-
```
sciwhiz12, line 191: Please keep this blank line between the last mod ID and the comment line.

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -19,27 +19,27 @@
...
-import net.minecraft.world.entity.player.Player;
-import net.minecraft.core.BlockPos;
-import net.minecraft.network.chat.Component;
 import net.minecraftforge.event.entity.living.LivingEvent;
+import net.minecraftforge.eventbus.api.Cancelable;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import java.io.File;
+import java.util.List;
```
sciwhiz12: Please avoid reorganizing imports.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -20,73 +20,51 @@
...
-import net.minecraft.world.level.LevelAccessor;
-import net.minecraft.world.level.LevelReader;
-import net.minecraft.world.level.Level;
-import net.minecraft.server.level.ServerLevel;
-import net.minecraft.world.level.biome.MobSpawnSettings;
-import net.minecraft.world.level.storage.ServerLevelData;
-import net.minecraft.world.level.storage.PlayerDataStorage;
 import net.minecraftforge.api.distmarker.Dist;
 import net.minecraftforge.api.distmarker.OnlyIn;
 import net.minecraftforge.client.MinecraftForgeClient;
```
sciwhiz12: Please avoid reorganizing imports or using wildcard imports.

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -553,4 +553,46 @@ public void setNewGameMode(GameType newGameMode)
...
+        }
+
+        public BlockPos getPos()
+        {
+            return this.pos;
+        }
+
+        public List<String> getLines()
+        {
+            return this.lines;
```
sciwhiz12: This should be an unmodifiable view of the list since this currently allows event listeners to modify the list directly, bypassing the setter method in the event.

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -553,4 +553,46 @@ public void setNewGameMode(GameType newGameMode)
             this.newGameMode = newGameMode;
         }
     }
+
+    /**
+     *
+     * Fired when a player clicks the "done" button in the sign edit UI.
+     * Cancelling will set the edited sign to be blank (i.e. lines will be an empty list)
```
sciwhiz12: This gives the impression that this event is fired on the client side, with the reference to the sign editing screen. Please mention in the javadocs which [logical side](https://mcforge.readthedocs.io/en/1.18.x/concepts/sides/) this event is fired on.

Furthermore, looking at the code suggests that this event is fired whenever any player sends a valid sign change packet to the server, even if the change is rejected later in `SeverGamePacketListenerImpl#updateSignText` due to the sign not being editable to that user or the block not even being a sign. 
It bears worth mentioning that aspect in this event, and possibly even adding a helper method to this event for checking if the sign block entity is actually editable by the player in this event (`isEditableByPlayer`). My main concern for this is a mod which naively logs all sign changes: this could possibly allow malicious actors to spam any such logs.

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -553,4 +553,46 @@ public void setNewGameMode(GameType newGameMode)
...
+    @Cancelable
+    public static class SignChangeEvent extends PlayerEvent
+    {
+
+        private final BlockPos pos;
+        private List<String> lines;
+
+        public SignChangeEvent(Player player, BlockPos pos, List<String> lines)
+        {
+            super(player);
```
sciwhiz12, line 581: Since this event is fired on the (logical) server, it may be good to override `getPlayer()` and have it return a `ServerPlayer` (thanks to covariant return types in Java).

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -553,4 +553,46 @@ public void setNewGameMode(GameType newGameMode)
...
+            this.pos = pos;
+            this.lines = lines;
+        }
+
+        public BlockPos getPos()
+        {
+            return this.pos;
+        }
+
+        public List<String> getLines()
```
sciwhiz12, line 623: This should likely be documented as the return list having only four entries (for the four lines on a sign).

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
danorris709: The wildcard import already existed and is used across multiple files in the forge project (https://i.gyazo.com/26808eefd63ff06786a7ae21d28adc42.png)

TheCurle: The fact that it's used elsewhere (which is a mistake) does not change the fact that we require PRs to not *add* them.

TheCurle: The fact that it's used elsewhere (which is a mistake) does not change the fact that we require PRs to not *add* them.

danorris709: Well given it's unused in an uncountably infinite number of places across the project already it feels like an arbitrary and futile thing to require. 

In addition to this, if you read what I wrote, my point was that I did NOT add this and as such it's not a worthwhile comment to make.

sciwhiz12: Regardless of the use (or lack thereof) of wildcard imports in other places within the project, the official code style used by Forge enforced for PRs mandates the lack of use of the wildcard import. The existing wildcard import _is_ a mistake (one which we will correct at the next opportunity), and should not be taken as a greenlight to add further wildcard imports.

Furthermore, the more pressing change here is the reorganization of the imports (usually caused by performing the IDE's optimize/organize imports functionality), causing the mess of import changes when reviewing files.

src/test/resources/META-INF/mods.toml
danorris709: I believe this was solved in the last merge from Master, just commenting to double check

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
danorris709: I believe this was solved with my last commit, just commenting to double check

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
danorris709: Yep I completely agree and I hadn't noticed that I had avoided the later checks. Although now looking at it this could make it more useful as it provides a wider range of use cases. I've added the isEditableByPlayer and commented about the logical side (and updated the description to include it firing after receiving the packet)

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
danorris709: Should be fixed with latest commit, just commenting to ensure

PR #8393, opened 2022-01-13
feat(api): add PlayerEvent.PrePlayerLoggedInEvent

danorris709:
Adds the PrePlayerLoggedInEvent to the PlayerEvent class and implements it in the minecraft ServerLoginPacketListenerImpl class.

**Proof of test succeeding:**
Test Success
Test Success 2
Test Success 3

**Reasoning:**
Adding this event allows for better protection mods when it becomes to banning/blacklisting specific players and IPs. The normal PlayerLoggedInEvent is fired too late to prevent some more technical players from causing issues.

In addition to the above, it's also useful for loading data and ensuring the data is loaded before the player has been loaded into the world.


Rasgnarok:
Following, a good improvement to server protection in general.

patches/minecraft/net/minecraft/server/network/ServerLoginPacketListenerImpl.java.patch
```diff
@@ -15,6 +15,21 @@
...
+                this.f_10019_ = ServerLoginPacketListenerImpl.State.DELAY_ACCEPT;
+                this.f_10024_ = serverplayer1;
+             } else {
++               Component reason = net.minecraftforge.event.ForgeEventFactory.firePrePlayerLoggedIn(serverplayer1, this.f_10013_);
++               if (reason != null) {
++                  this.f_10013_.m_129514_(new ClientboundDisconnectPacket(reason), (p_9828_) -> this.f_10013_.m_129507_(reason));
++                  this.f_10013_.m_129540_();
++                  this.f_10018_.m_18709_(this.f_10013_::m_129541_);
++                  return;
++               }
```
diesieben07: Why is this code _here_ specifically? I am not sure when the above case (`DELAY_ACCEPT`) happens, but when it does, this code is never executed and thus the checks are bypassed.
Why is this hook not in `PlayerList.canPlayerLogin`?

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -805,6 +807,18 @@ public static void firePlayerChangedDimensionEvent(Player player, ResourceKey<Le
         MinecraftForge.EVENT_BUS.post(new PlayerEvent.PlayerChangedDimensionEvent(player, fromDim, toDim));
     }
 
+    public static Component firePrePlayerLoggedIn(Player player, Connection connection)
+    {
+        PlayerEvent.PrePlayerLoggedInEvent event = new PlayerEvent.PrePlayerLoggedInEvent(player, connection);
+        MinecraftForge.EVENT_BUS.post(event);
+
+        if (event.getResult() == Result.DENY) {
```
diesieben07: Formatting

diesieben07 (changes requested):
Adding to my above comments, please review the previous pull request on this topic (#6889) and account for the concerns and comments made there.

patches/minecraft/net/minecraft/server/network/ServerLoginPacketListenerImpl.java.patch
danorris709: So I put it here because as I understand it the DELAY_ACCEPT only happens when the user with said UUID is already logged in and it's delaying to see if they're just an artifact of lag (and if they're not the connection gets denied for "You're already logged in"). 

I didn't add it to PlayerList.canPlayerLogin as that's where the whitelist is checked. I suppose it could go in there. However, at that point you don't have access to the player object so I figured maybe it'd be more useful for mods utilizing this API to have such access.

danorris709:
> Adding to my above comments, please review the previous pull request on this topic (#6889) and account for the concerns and comments made there.

Went through that thread and found majority of the discussion was about overriding the whitelist which is not really a goal of this PR (So I hope that doesn't apply here)? And the rest was about the name of the class. I've refactored it to be hopefully more appropriate

patches/minecraft/net/minecraft/server/network/ServerLoginPacketListenerImpl.java.patch
diesieben07: Please put all this code into the event hook the instead of into the patch. Keep patches small if possible.
Is there a reason you are not calling `ServerLoginPacketListenerImpl#disconnect`, like vanilla does for disconnections at this stage?

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -462,6 +463,39 @@ public ItemStack getSmelting()
         }
     }
 
+    /**
+     *
+     * Event fired before the player is accepted to the world
+     *
+     */
+    @HasResult
+    public static class PrePlayerLoginEvent extends PlayerEvent {
```
diesieben07: Please document what the result means for this event. Why `@HasResult` and not `@Cancellable`? What does `Result.DEFAULT` do vs. `Result.ALLOW`?
Also: Formatting, braces on new lines.

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -462,6 +463,39 @@ public ItemStack getSmelting()
...
+        private Component denyMessage = null;
+        private Connection connection;
```
diesieben07: Field should be final.

src/test/java/net/minecraftforge/debug/entity/player/PlayerLoginEventTest.java
```diff
@@ -0,0 +1,47 @@
...
+public class PlayerLoginEventTest
+{
+    private static final Logger LOGGER = LogManager.getLogger(PlayerLoginEventTest.class);
+    private static final boolean ENABLED = false;
+
+    @SubscribeEvent
+    public static void onPlayerChangeGameModeEvent(PlayerEvent.PrePlayerLoginEvent event)
+    {
+        LOGGER.info("{} logged in with connection {}.", event.getPlayer().getName(), event.getConnection());
+        if (ENABLED) {
```
diesieben07: Formatting: Braces on new lines.
And to properly disable the mod, you should not even register to the event bus (i.e. you should not use `@EventBusSubscriber`).

src/test/resources/META-INF/mods.toml
```diff
@@ -176,7 +176,8 @@ license="LGPL v2.1"
     modId="capabilities_test"
 [[mods]]
     modId="custom_particle_type_test"
+[[mods]]
+    modId="player_login_event_test"
 [[mods]]
     modId="ingredient_invalidation"
-
```
diesieben07: unnecessary whitespace change

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -30,9 +30,9 @@
 import com.mojang.blaze3d.vertex.PoseStack;
```
diesieben07, line 30: Missing license header update

patches/minecraft/net/minecraft/server/network/ServerLoginPacketListenerImpl.java.patch
danorris709: Moving the logic to the hook. And when using the implementation from `ServerLoginPacketListenerImpl#disconnect` I get [this](https://i.gyazo.com/78681e94197d139d2c63dfdd9c2e3508.png) disconnect error. However, when using the one above it disconnects with the correct message. I attempted to try find out why and I came to the conclusion that the logic was too late for this packet to be sent (?). Not 100% certain on that explanation but it's something about that specific packet. And I also did attempt to change the text in the kick message but regardless of the length of the disconnect message it always resulted in the same buffer error.

src/test/java/net/minecraftforge/debug/entity/player/PlayerLoginEventTest.java
danorris709: The ENABLED boolean is not used for disabling the test. It's used for testing if the kick works or not.

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
diesieben07: The documentation here has not been fixed. Neither has the formatting.

src/test/java/net/minecraftforge/debug/entity/player/PlayerLoginEventTest.java
diesieben07: Then it is still wrong. It _should_ disable the test.

patches/minecraft/net/minecraft/server/network/ServerLoginPacketListenerImpl.java.patch
```diff
@@ -15,6 +15,17 @@
...
        } else if (this.f_10019_ == ServerLoginPacketListenerImpl.State.DELAY_ACCEPT) {
           ServerPlayer serverplayer = this.f_10018_.m_6846_().m_11259_(this.f_10021_.getId());
+@@ -116,6 +_,10 @@
+                this.f_10019_ = ServerLoginPacketListenerImpl.State.DELAY_ACCEPT;
+                this.f_10024_ = serverplayer1;
+             } else {
++               if (!net.minecraftforge.event.ForgeEventFactory.firePrePlayerLoggedIn(serverplayer1, this.f_10013_)) {
++                  return;
++               }
++
```
diesieben07: This could be a 1 line patch.

diesieben07 (changes requested):
> Went through that thread and found majority of the discussion was about overriding the whitelist which is not really a goal of this PR (So I hope that doesn't apply here)? And the rest was about the name of the class. I've refactored it to be hopefully more appropriate

[This comment](https://github.com/MinecraftForge/MinecraftForge/pull/6889#issuecomment-666719173) specifically has not been addressed.

A use case has also not been provided. What "better protection mods" are necessary beyond the already present banning of IPs and usernames in vanilla?

danorris709:
> > Went through that thread and found majority of the discussion was about overriding the whitelist which is not really a goal of this PR (So I hope that doesn't apply here)? And the rest was about the name of the class. I've refactored it to be hopefully more appropriate
> 
> [This comment](https://github.com/MinecraftForge/MinecraftForge/pull/6889#issuecomment-666719173) specifically has not been addressed.
> 
> A use case has also not been provided. What "better protection mods" are necessary beyond the already present banning of IPs and usernames in vanilla?

That comment was what I specifically said wasn't an issue. As a result of this event being later in the login process there is no controversy about overriding the whitelist. As the whitelist has already been processed before this event even occurs. 

As for the use case, there are two that I have stated. One being to pre-load player data before they hit the world. This is important as to ensure that mods can always have the player data loaded in time for the player to be in the game. There are plenty of examples of similar events in both Sponge and Spigot (look at AsyncPrePlayerLoginEvent if you're wondering). Even in these cases, these platforms allow for overriding the whitelist. Which is not what this event does regardless.

The second usecase would be better IP banning. IP banning can be quite trivial and there are some mods that have a wider check than the simple Minecraft one (I know a friend has one in particular). The specific instance they told me about was where they had players logging in using bots and VPNs and by the time their mod was able to identify and kick them (using existing events) they had already loaded into the world and begun griefing.

patches/minecraft/net/minecraft/server/network/ServerLoginPacketListenerImpl.java.patch
danorris709: How could this be a one-line patch without directly contradicting the code styling document for this project?

src/test/java/net/minecraftforge/debug/entity/player/PlayerLoginEventTest.java
danorris709: It should not disable the test. That would require a second boolean, or as you said earlier, removing it from the mods.toml.

diesieben07:
The comment is not about overriding the whitelist.

src/test/java/net/minecraftforge/debug/entity/player/PlayerLoginEventTest.java
diesieben07: Why not? That is how all other test mods work.

patches/minecraft/net/minecraft/server/network/ServerLoginPacketListenerImpl.java.patch
diesieben07: Code style for patches is separate from code style in Forge code.

danorris709:
> The comment is not about overriding the whitelist.

Okay, my apologies but it's a very trivial answer. Forge has a priority system for events, thus the entire premise of the question is futile. 

Your response to such an answer previously was about "What if X developer tries to do Y solution" well that's not really something I should have to worry about when making an event. One could equally argue, what if someone attempts to register a block listener at the very end but then finds someone else has already handled cancelling the event.

In addition to the priority system there is also cancelling the event.

patches/minecraft/net/minecraft/server/network/ServerLoginPacketListenerImpl.java.patch
danorris709: In future can you press the unresolve button so I can tell when you reply to a comment on a resolved thread?

Also, there are multiple (recent) patches where similar if statements are not a one line diff.

sciwhiz12: Unfortunately, marking review threads as (un)resolved requires one to either be the PR author, or have write access to the repository (which we do not have as Triage Team members).

src/test/java/net/minecraftforge/debug/entity/player/PlayerLoginEventTest.java
danorris709: It's not how all other test mods work, and as I stated originally that specific boolean is for testing the kick handling (i.e. not always wanting that enabled). 

In addition to this, your initial comment was "And to properly disable the mod, you should not even register to the event bus" but in every other instance of the ENABLED boolean being used it is done exactly how I've implemented it (i.e. it is registered to the bus)

patches/minecraft/net/minecraft/server/network/ServerLoginPacketListenerImpl.java.patch
danorris709: Well that is unfortunate, but it also makes it very difficult to see what's going on

diesieben07:
> Forge has a priority system for events, thus the entire premise of the question is futile.

It does, however it is very limited and does not at all solve this problem.

> One could equally argue, what if someone attempts to register a block listener at the very end but then finds someone else has already handled cancelling the event.

Block registration is not cancellable.

> In addition to the priority system there is also cancelling the event.

What do you mean by "in addition"? These two things combined are specifically what is the problem.

patches/minecraft/net/minecraft/server/network/ServerLoginPacketListenerImpl.java.patch
diesieben07: This is why you should not just resolve conversations without being sure they are actually resolved.

danorris709:
> > Forge has a priority system for events, thus the entire premise of the question is futile.
> 
> It does, however it is very limited and does not at all solve this problem.
> 
> > One could equally argue, what if someone attempts to register a block listener at the very end but then finds someone else has already handled cancelling the event.
> 
> Block registration is not cancellable.
> 
> > In addition to the priority system there is also cancelling the event.
> 
> What do you mean by "in addition"? These two things combined are specifically what is the problem.

Well if the issue sis that the Forge priority system is limited then it is an issue with that system and not my event. As such the question goes far outside the scope of this merge. 

You really misread what I said. I said they register a block LISTENER to listen to the BlockPlaceEvent at the end using the forge priority and then as a result of the issue you bring up it doesn't work how they intended. Nonetheless, my point was this argument could be made for ANY event ANYWHERE. As such, and as I stated above, this issue isn't for this merge request but should be dealt with by improving the priority system itself.

By In addition I meant that if the event was being listened to properly they could determine if something with a different priority had already cancelled it (think isCanceled method)

diesieben07:
> Well if the issue sis that the Forge priority system is limited then it is an issue with that system and not my event. As such the question goes far outside the scope of this merge.

> You really misread what I said. I said they register a block LISTENER to listen to the BlockPlaceEvent at the end using the forge priority and then as a result of the issue you bring up it doesn't work how they intended. Nonetheless, my point was this argument could be made for ANY event ANYWHERE. As such, and as I stated above, this issue isn't for this merge request but should be dealt with by improving the priority system itself

While that is correct, it is more relevant here, because whether or not players can connect to a server is a security critical system.

> By In addition I meant that if the event was being listened to properly they could determine if something with a different priority had already cancelled it (think isCanceled method)

Correct. And then (potentially nefariously) call setCancelled(false)

danorris709:
> > Well if the issue sis that the Forge priority system is limited then it is an issue with that system and not my event. As such the question goes far outside the scope of this merge.
> 
> > You really misread what I said. I said they register a block LISTENER to listen to the BlockPlaceEvent at the end using the forge priority and then as a result of the issue you bring up it doesn't work how they intended. Nonetheless, my point was this argument could be made for ANY event ANYWHERE. As such, and as I stated above, this issue isn't for this merge request but should be dealt with by improving the priority system itself
> 
> While that is correct, it is more relevant here, because whether or not players can connect to a server is a security critical system.
> 
> > By In addition I meant that if the event was being listened to properly they could determine if something with a different priority had already cancelled it (think isCanceled method)
> 
> Correct. And then (potentially nefariously) call setCancelled(false)

But then how does one resolve it in this if the issue lies elsewhere? It would seem rather futile to add a whole new priority system when one already exists? 

I'm happy to open a merge in which a better priority system is implemented? But I'm uncertain the solution lies within this event. Perhaps a warning is logged to console if someone attempts to (un)cancel this when it was already cancelled?

src/test/java/net/minecraftforge/debug/entity/player/PlayerLoginEventTest.java
danorris709: Added global ENABLED variable and refactored old variable to KICK_ENABLED

patches/minecraft/net/minecraft/server/network/ServerLoginPacketListenerImpl.java.patch
danorris709: Updated to be inline if

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -462,6 +463,50 @@ public ItemStack getSmelting()
         }
     }
 
+    /**
+     *
+     * Event fired before the player is accepted to the world
+     *
+     */
+    @Cancelable
+    public static class PrePlayerLoginEvent extends PlayerEvent {
```
diesieben07: Wrong formatting here still.
Also please document what cancelling does.

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -462,6 +463,50 @@ public ItemStack getSmelting()
...
+        {
+            return this.connection;
+        }
+
+        public Component getDenyMessage()
+        {
+            return this.denyMessage;
+        }
+
+        public void setDenyMessage(Component denyMessage)
```
diesieben07, line 525: This should have documentation.

src/test/java/net/minecraftforge/debug/entity/player/PlayerLoginEventTest.java
diesieben07: Why are the two variables necessary?

diesieben07 (changes requested):

src/test/java/net/minecraftforge/debug/entity/player/PlayerLoginEventTest.java
danorris709: Well I figured it'd probably be best to test the two cases (using the variable). So that is checking the event fires regardless of cancellation, and kick status, and then also checking the disconnect works properly. 

Can be removed if you think it's unnecessary

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
danorris709: Added javadoc comments at the top of the class and on the setCancelled override (and fixed formatting)

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
danorris709: Added documentation, and also noticed missing default value. So I added a default value and then explained in the documentation that it only exists to prevent future NPEs

diesieben07: If a NPE is possible for null values, then setDenyMessage should throw when called with null

src/test/java/net/minecraftforge/debug/entity/player/PlayerLoginEventTest.java
diesieben07: It's fine I guess. The method just has the wrong name and there is still some formatting problems here.

diesieben07 (changes requested):

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -805,6 +805,21 @@ public static void firePlayerChangedDimensionEvent(Player player, ResourceKey<Le
...
 
+    public static boolean firePrePlayerLoggedIn(Player player, Connection connection)
+    {
+        PlayerEvent.PrePlayerLoginEvent event = new PlayerEvent.PrePlayerLoginEvent(player, connection);
+        MinecraftForge.EVENT_BUS.post(event);
+
+        if (event.isCanceled())
+        {
+            connection.send(new ClientboundDisconnectPacket(event.getDenyMessage()));
+            connection.disconnect(event.getDenyMessage());
```
diesieben07, line 816: What's the reason for not using `ServerLoginPacketListenerImpl#disconnect`?

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -462,6 +464,69 @@ public ItemStack getSmelting()
...
+         * This will throw a {@link RuntimeException} if you attempt to uncancel a previously cancelled event
+         *
+         * @param cancel If the event should be cancelled (default false)
+         */
+        @Override
+        public void setCanceled(boolean cancel)
+        {
+            if (this.isCanceled() && !cancel)
+            {
+                throw new RuntimeException("You cannot un-cancel the PrePlayerLogin event");
```
diesieben07: I think IllegalArgumentException would be more suitable

diesieben07 (changes requested):
Github stop eating my comments

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
danorris709: I initially attempted to use that method and got a buffer overflow exception on the client end. Couldn't figure out why it wasn't working (https://i.gyazo.com/78681e94197d139d2c63dfdd9c2e3508.png). I believe it's something to do with when the event is fired (too late to use this method?)

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
danorris709: Updated to use IllegalArgumentException

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
danorris709: Added an `Objects#requireNonNull` check before assigning variable (I believe this should throw an error if null)

src/test/java/net/minecraftforge/debug/entity/player/PlayerLoginEventTest.java
danorris709: Updated name and formatting

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
diesieben07: The method literally does the same thing, it just has a try-catch around it. So that sounds improbable.

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
```diff
@@ -462,6 +465,70 @@ public ItemStack getSmelting()
...
+        {
+            super(player);
+
+            this.connection = connection;
+        }
+
+        /**
+         *
+         * Cancels the player logging into the server
+         * This will throw a {@link RuntimeException} if you attempt to uncancel a previously cancelled event
```
diesieben07: Still mentions `RuntimeException`

diesieben07 (changes requested):

src/main/java/net/minecraftforge/event/entity/player/PlayerEvent.java
danorris709: Updated

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
danorris709: Actually, not quite. The method in ServerLoginPacketListenerImpl uses a "ClientboundLoginDisconnectPacket" whereas I'm using a "ClientboundDisconnectPacket"

diesieben07 (approved):

diesieben07 (changes requested):
More justification for this feature has been requested due to the high potential for abuse.
Specifically please elaborate on your mention of "more technical players".

danorris709:
> More justification for this feature has been requested due to the high potential for abuse. Specifically please elaborate on your mention of "more technical players".

Okay so my personal main use would be for loading player data before the player hits the world. This is important particularly in the case of database storage as otherwise it leads to a potential of server-side mods not having player's data loaded before they start using commands.

As for "more technical players" I was more speaking in terms of more advanced hacks (and other types of abuse) where by the time the current events hit the player has loaded into the world. This means that before the mod can ban them they have access to the world data, and a like, and as such can use this data to send malicious packets (etc) that would damage the server at hand.

I hope this provides some insight. Personally, I am most interested in the first use-case. However, when I began making this PR the second was brought to my attention.

diesieben07:
Personally I think the first use-case is invalid. Database access is something asynchronous anyways, so immediate access cannot be guaranteed at all and must be dealt with by the consumer, usually via mechanisms like `CompletableFuture`.

danorris709:
> Personally I think the first use-case is invalid. Database access is something asynchronous anyways, so immediate access cannot be guaranteed at all and must be dealt with by the consumer, usually via mechanisms like `CompletableFuture`.

Yeah the database access is asynchronous but that doesn't change the fact that it still takes time to load. The sooner that async database call is made the sooner the data can be fully loaded. Whereas, if it's called with the current event that data doesn't get loaded for up to several seconds into the player being loaded into the world.

PR #8394, opened 2022-01-15
Fix FoliagePlacerType and TreeDecoratorType registry

dipo33:
Currently `FoliagePlacerType` and `TreeDecoratorType ` forge registries are not working but they exist. The actual registries weren't wrapped by forge. Probably was missed during an update. This PR patches this and wraps the registries to make it work.

### DIFF
--- a/patches/minecraft/net/minecraft/core/Registry.java.patch
+++ b/patches/minecraft/net/minecraft/core/Registry.java.patch
@@ -149,7 +149,7 @@
        return BlockStateProviderType.f_68752_;
     });
 -   public static final Registry<FoliagePlacerType<?>> f_122858_ = m_122999_(f_122848_, () -> {
-+   public static final Registry<FoliagePlacerType<?>> f_122858_ = forge(f_122848_, () -> {
++   @Deprecated public static final Registry<FoliagePlacerType<?>> f_122858_ = forge(f_122848_, () -> {
        return FoliagePlacerType.f_68591_;
     });
     public static final Registry<TrunkPlacerType<?>> f_122859_ = m_122999_(f_122849_, () -> {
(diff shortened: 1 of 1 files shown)


patches/minecraft/net/minecraft/core/Registry.java.patch
```diff
@@ -140,15 +140,19 @@
...
     public static final ResourceKey<Registry<Codec<? extends SurfaceRules.RuleSource>>> f_194572_ = m_122978_("worldgen/material_rule");
     public static final ResourceKey<Registry<StructureProcessorType<?>>> f_122854_ = m_122978_("worldgen/structure_processor");
     public static final ResourceKey<Registry<StructurePoolElementType<?>>> f_122855_ = m_122978_("worldgen/structure_pool_element");
 -   public static final Registry<BlockStateProviderType<?>> f_122856_ = m_122999_(f_122846_, () -> {
 +   @Deprecated public static final Registry<BlockStateProviderType<?>> f_122856_ = forge(f_122846_, () -> {
        return BlockStateProviderType.f_68752_;
     });
-    public static final Registry<FoliagePlacerType<?>> f_122858_ = m_122999_(f_122848_, () -> {
+-   public static final Registry<FoliagePlacerType<?>> f_122858_ = m_122999_(f_122848_, () -> {
++   public static final Registry<FoliagePlacerType<?>> f_122858_ = forge(f_122848_, () -> {
```
sciwhiz12: Please add a `@Deprecated` annotation to this modified field, as is done with the other modified registry fields.
```suggestion
+   @Deprecated public static final Registry<FoliagePlacerType<?>> f_122858_ = forge(f_122848_, () -> {
```

sciwhiz12 (changes requested):
Confirmed to be a regression caused during the 1.18 patch and update cycle.

In addition to the `FoliagePlacerType` registry, it seems the `TreeDecoratorType` also seems to have a Forge registry without a corresponding `forge()`-modified registry field. Please also fix this field in your PR.

patches/minecraft/net/minecraft/core/Registry.java.patch
dipo33: Done

sciwhiz12 (approved):

PR #8398, opened 2022-01-16
[1.18.x] Fix Mob Spawner logic with CustomSpawnRules

SilverDavidMC:
When using a spawner entry with custom spawn rules added in 1.18, those rules are not used properly when attempting to spawn the mob, resulting in the spawner failing even when it should succeed. This PR aims to fix that, matching vanilla behavior.

This could be a breaking change, since `ForgeEventFactory.canEntitySpawnSpawner` needed an additional argument, however this could be made nullable with a method using the original arguments if that's a concern.

Supplied are two commands for mob spawner blocks, one that spawns Creepers with normal light restrictions, and another that only spawns Creepers in bright skylight. For testing, I placed both spawners exposed on the surface during the day. The normal spawner shouldn't work under these conditions, while the one with custom rules should. Doing this underground, in the dark, switches it such that the normal spawner works and the custom rule spawner stops.

Normal Spawner:

`/setblock ~ ~ ~ minecraft:spawner{Delay:20s,MaxNearbyEntities:6s,MaxSpawnDelay:100s,MinSpawnDelay:90s,RequiredPlayerRange:16s,SpawnCount:1s,SpawnPotentials:[{data:{entity:{id:"minecraft:creeper"}},weight:1}],SpawnRange:4s}`

Bright Skylight only (Non-functional in current Forge):

`/setblock ~ ~ ~ minecraft:spawner{Delay:20s,MaxNearbyEntities:6s,MaxSpawnDelay:100s,MinSpawnDelay:90s,RequiredPlayerRange:16s,SpawnCount:1s,SpawnPotentials:[{data:{custom_spawn_rules:{sky_light_limit:[10,15]},entity:{id:"minecraft:creeper"}},weight:1}],SpawnRange:4s}`

### DIFF
--- a/src/main/java/net/minecraftforge/event/ForgeEventFactory.java
+++ b/src/main/java/net/minecraftforge/event/ForgeEventFactory.java
@@ -216,7 +216,7 @@ public class ForgeEventFactory
     {
         Result result = canEntitySpawn(entity, world, x, y, z, spawner, MobSpawnType.SPAWNER);
         if (result == Result.DEFAULT)
-            return !(spawnRules.isEmpty() && !entity.checkSpawnRules(world, MobSpawnType.SPAWNER) || !entity.checkSpawnObstruction(world)); // vanilla logic (inverted)
+            return (!spawnRules.isEmpty() || entity.checkSpawnRules(world, MobSpawnType.SPAWNER)) && entity.checkSpawnObstruction(world); // vanilla logic (inverted)
         else
             return result == Result.ALLOW;
     }
--- a/src/main/java/net/minecraftforge/event/ForgeEventFactory.java
+++ b/src/main/java/net/minecraftforge/event/ForgeEventFactory.java
@@ -205,7 +205,7 @@ public class ForgeEventFactory
         return event.getResult();
     }
 
-    public static boolean canEntitySpawnSpawner(SpawnData spawnData, Mob entity, Level world, float x, float y, float z, BaseSpawner spawner)
+    public static boolean canEntitySpawnSpawner(SpawnData spawnData, Mob entity, LevelAccessor world, float x, float y, float z, BaseSpawner spawner)
     {
         Result result = canEntitySpawn(entity, world, x, y, z, spawner, MobSpawnType.SPAWNER);
         if (result == Result.DEFAULT)
@@ -214,7 +214,7 @@ public class ForgeEventFactory
             return result == Result.ALLOW;
     }
 
-    public static boolean doSpecialSpawn(Mob entity, Level world, float x, float y, float z, BaseSpawner spawner, MobSpawnType spawnReason)
+    public static boolean doSpecialSpawn(Mob entity, LevelAccessor world, float x, float y, float z, BaseSpawner spawner, MobSpawnType spawnReason)
     {
         return MinecraftForge.EVENT_BUS.post(new LivingSpawnEvent.SpecialSpawn(entity, world, x, y, z, spawner, spawnReason));
     }
--- a/src/main/java/net/minecraftforge/event/ForgeEventFactory.java
+++ b/src/main/java/net/minecraftforge/event/ForgeEventFactory.java
@@ -83,6 +83,7 @@ import net.minecraft.world.level.GameRules;
 import net.minecraft.world.level.LevelAccessor;
 import net.minecraft.world.level.LevelReader;
 import net.minecraft.world.level.Level;
+import net.minecraft.world.level.SpawnData;
 import net.minecraft.server.level.ServerLevel;
 import net.minecraft.world.level.biome.MobSpawnSettings;
 import net.minecraft.world.level.storage.ServerLevelData;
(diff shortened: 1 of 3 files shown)


sciwhiz12 (approved):
Looks :shipit: to me, and my tests confirm it works. The tiny breaking change should be fine if this gets in before the RB; if it doesn't, then we'll need to give it the `@Nullable` treatment.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -204,11 +205,11 @@ public static Result canEntitySpawn(Mob entity, LevelAccessor world, double x, d
         return event.getResult();
     }
 
-    public static boolean canEntitySpawnSpawner(Mob entity, Level world, float x, float y, float z, BaseSpawner spawner)
+    public static boolean canEntitySpawnSpawner(SpawnData spawnData, Mob entity, Level world, float x, float y, float z, BaseSpawner spawner)
```
Shadows-of-Fire: I suggest this `Level` parameter be changed to type `LevelAccessor`, since all subsequent calls are of type `LevelAccessor`.  This probably should've been done a while ago, but since we're doing breaking changes we may as well do it now.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -204,16 +205,16 @@ public static Result canEntitySpawn(Mob entity, LevelAccessor world, double x, d
         return event.getResult();
     }
 
-    public static boolean canEntitySpawnSpawner(Mob entity, Level world, float x, float y, float z, BaseSpawner spawner)
+    public static boolean canEntitySpawnSpawner(SpawnData spawnData, Mob entity, LevelAccessor world, float x, float y, float z, BaseSpawner spawner)
     {
```
LexManos, line 216: Even tho we are not at a RB yet {waiting for Mojang to drop 1.18.2} we're past the point if binary breaks.
So these changes need to be bounced through a deprecated method to keep bin compat.
Also, when adding new arguments to a method, they go at the END to make life a ton less confusing for people calling the method...

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -204,16 +205,16 @@ public static Result canEntitySpawn(Mob entity, LevelAccessor world, double x, d
...
         Result result = canEntitySpawn(entity, world, x, y, z, spawner, MobSpawnType.SPAWNER);
         if (result == Result.DEFAULT)
-            return entity.checkSpawnRules(world, MobSpawnType.SPAWNER) && entity.checkSpawnObstruction(world); // vanilla logic (inverted)
+            return !(spawnData.getCustomSpawnRules().isEmpty() && !entity.checkSpawnRules(world, MobSpawnType.SPAWNER) || !entity.checkSpawnObstruction(world)); // vanilla logic (inverted)
         else
             return result == Result.ALLOW;
     }
 
-    public static boolean doSpecialSpawn(Mob entity, Level world, float x, float y, float z, BaseSpawner spawner, MobSpawnType spawnReason)
+    public static boolean doSpecialSpawn(Mob entity, LevelAccessor world, float x, float y, float z, BaseSpawner spawner, MobSpawnType spawnReason)
```
Shadows-of-Fire, line 231: This signature change (and the SpecialSpawn constructor change) will also need to be implemented as a redirect-and-deprecate fashion.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -204,16 +205,30 @@ public static Result canEntitySpawn(Mob entity, LevelAccessor world, double x, d
...
+    {
+        return canEntitySpawnSpawner(entity, world, x, y, z, spawner, Optional.empty());
+    }
+
+    public static boolean canEntitySpawnSpawner(Mob entity, LevelAccessor world, float x, float y, float z, BaseSpawner spawner, Optional<SpawnData.CustomSpawnRules> spawnRules)
     {
         Result result = canEntitySpawn(entity, world, x, y, z, spawner, MobSpawnType.SPAWNER);
         if (result == Result.DEFAULT)
-            return entity.checkSpawnRules(world, MobSpawnType.SPAWNER) && entity.checkSpawnObstruction(world); // vanilla logic (inverted)
+            return !(spawnRules.isEmpty() && !entity.checkSpawnRules(world, MobSpawnType.SPAWNER) || !entity.checkSpawnObstruction(world)); // vanilla logic (inverted)
```
Shadows-of-Fire: This is the same as `!spawnRules.isEmpty() || entity.checkSpawnRules(world, MobSpawnType.SPAWNER) && entity.checkSpawnObstruction(world);` just a lot less readable

SilverDavidMC: In this case, since the main goal of this line is to mimic vanilla code, I feel that matching vanilla code is more important than readability for the sake of future maintenance. 

May be worth noting that the `||` part would need to be wrapped in parentheses to keep the order of operations intact.

TheCurle: > I feel that matching vanilla code is more important than readability for the sake of future maintenance. 

We don't. 

Why would you ever prioritise keeping the dodgy decompiler output over something much simpler and easier for us to maintain in the future?

That seems like an intentional FU to the people who have to fix this stuff when it breaks.

PR #8402, opened 2022-01-17
[1.18.x] Added new Disconnection screen to display better channel/registry mismatch information

RedstoneDubstep:
This PR adds a new Disconnection screen which appears instead of the Vanilla disconnection screen in case the disconnect happened within Forge's Handshake. If any disconnect happens because of a channel versions mismatch, or because of a channel missing on either side, or because of a mismatching registry, this screen will provide the user with most of the details sufficient to fix the incompatibility without having to look through (maybe incomplete) log data. 
The data appearing in the screen includes the name of the mod adding the mismatched channel/registry (which is given in most circumstances), the mod/channel id of said channels/registries, and the client and server versions of the channel (if applicable).
A test mod has been added for the purpose of intentionally causing the handshake to fail due to a channel mismatch and thus displaying the new Disconnection screen.

Currently, this PR makes use of a new Handshake Packet (`S2CModMismatchData`) to notify the client that the handshake failed, as well as to ensure data like the server mod names and channel versions are sent to the screen for proper display. ~~However, I think this makes the PR incompatible with older versions of Forge and thus a breaking change. Feedback is appreciated to how this can be resolved or if it's even worth keeping the custom packet.~~ Edit: This PR also fixes this exact incompatibility, which previously was caused by outdated clients receiving the new packets not completing the handshake, and thus it should™ no longer be a breaking change.

For the updated version of this mod's screen, please scroll further down, the screen below is the initial version of it:
screen


SirEndii:
This is a neat addition

marchermans (changes requested):
This is a very good start. But since this is user facing and actually provides him information in an error case, without translations this is not really useful.

Could you replace the hard coded messages with translation keys?

Claycorp:
Your general mod user isn't going to have any clue what a "Client/Server Chanel Version" is nor how to figure out how to fix that, even with the basic info given. 

Thus I would consider reformatting the screen and making the information more user clear.

- The "columns" should start below the spaces they matter in. Making them shorter will also save space, just refining them to `Client` & `Server` should suffice, if not, adding `Version` to the end of each would work too.
- Also either alternating color coded or fill the empty space between as having a list with many options that spaced out will be hard to read. Especially in small UI scales.

For the missing server/client list,
- Outright listing of missing mods it would likely be better and just list them out like so:
"`modname/ID` is missing from the `side`"

For the mismatching version list, 
- `The following mods have a mismatching client and server channel version, please consider updating either mod version:` is not something people will get. This error isn't for programers so it's best to not use "programer speak" when dealing with it. I would instead go with something more generic like `The following mod versions do not match, consider downloading the server matching version of these mods:`
- It might also benefit from color coding the version that is lower in yellow as a way to help differentiate them. Though that could also be confusing to users.

Overall, this is currently very developer oriented and that's going to result in poor reception from the people actually relying on it.

RedstoneDubstep:
Claycorp: Thank you for your feedback! I already tried to design the gui as user-friendly as possible, though I can agree that it is quite improveable. 
I agree with most of your points, however I have some questions:
- What do you mean with your first adressed point? Should I make the columns only get into use in the `Mismatched Mod Channel` section of the list? And/Or should I make the version columns smaller (judging from your suggested column header changes)?
- Since the screen will always ever list mods missing on the server OR mods missing on the client (due to the mechanics of the handshake), I think it's a bit redundant to say "`modname` is missing from the `alwaysthesamedist`" for every single missing mod, so I would prefer the old way of specifying the dist once at the top and then listing the missing. If you have a different opinion about that, feel free to convince me of it.

Claycorp:
- Yes, separate the screen in half. Keep all your non-version related things at the top and then have the columns below. This would also keep the "These are absolutely critical to have" parts separated from the "Well this _might_ be important to take care of" parts.
Example: 
image

- The only reasoning behind what I suggested is because screen space is prime real estate. You have more than enough room to add to the end of the mod names and the likelihood of a name being long enough with text is very, very slim. So instead of wasting a line on something that just defines the space below, why not have room for another entry or two?

diesieben07:
I want to point out that this is about channel versions (i.e. protocol versions), _not_ mod versions. They do not need to align and often don't. Labelling the channel version as "mod version" will be even more confusing, because those versions might not exist as an actual version of the mod and if they do, not correspond to the network version.

Claycorp:
> I want to point out that this is about channel versions (i.e. protocol versions), _not_ mod versions. They do not need to align and often don't. Labelling the channel version as "mod version" will be even more confusing, because those versions might not exist as an actual version of the mod and if they do, not correspond to the network version.

Then it would likely be best to entirely cull that information from screen. If there's no way for someone to know what to update to and it often doesn't need to match. There's little point in surfacing it for users.

diesieben07:
The protocol version should not be surfaced, correct. The mod version should be shown - if the network version has a mismatch.

RedstoneDubstep:
I mainly decided to display the channel version instead of the mod version because the server channel version is already available on the client (through the data from `S2CModList`), while the mod version of the server mod isn't. If I wanted to display the mod version instead (which, yes, would improve readability) I would need to add another packet or enhance the `mods` field of `S2CModList` to also send the mod versions. However, that would make this PR even more of a breaking chance, which is not desirabe. 
It's true that you can't really read useful information out of the channel versions, but the user can at least see on what side the version is more outdated (which could be highlighted even more with a different color, as Claycorp already suggested)

diesieben07:
It might be possible to send the mod version in a backwards compatible way, in my opinion it is worth investigating.

RedstoneDubstep:
I've changed the layout in the screen a bit according to the feedback I got (thanks to everyone who helped), the updated look of the screen can be seen below. I'll update the texts to be translatable soon, in the meantime further design feedback is always appreciated.

Edit: making the strings used in the screen translatable wasn't as big of a deal as I initially assumed, so that's now done as well. So (I think) everything that has been adressed here so far has been fixed/changed, other than the issue of this PR being a breaking change, which I still haven't found a solution for.

https://user-images.githubusercontent.com/42962686/150394696-d175795d-0d72-46c2-972a-6397fd6005ea.mp4

Claycorp:
Here's a few points for the new layout. Looks way better than the last one, that's for sure.

- `"Your client is missing the following mods, you need to download them to play on this server:"` I would personally go with `"Your client is missing the following mods, install these mods to join this server:"`
- The long length wrap is too long. It's too easy to mistake the version as part of the name. If you can wrap over another 2-3 characters it would likely work fine. You may also want to test with a solid string of W or other wide characters too as you have a fair few thin characters in there so it might bleed into the version worse yet.

- `"The following mod versions do not match, you need to download a matching version of these mods to play on this server:"` This doesn't sound right and doesn't reflect what needs to be done. `"The following mod versions do not match, install the same version this server has to join:"`
- The last section may also happen due to configs not matching I believe thus it might be wise to add something in there mentioning configs. Perhaps something like `"The following mods have mismatched registry data, this can be caused by the common configs of the mod not matching, your client mod version does not match the servers or your client does not have the mod installed."`


- Is there a reason for not displaying the full list if the window is already using a scroll bar in the last section?
- Perhaps it would be a good idea to get the `displayURL` from the `mods.toml` file (if there is one set) and make all the mod names a link to open the URL (if they aren't already). This would give the a fast way to get files for incorrect mods. If this is done, you would need to add info somewhere saying you can click the names to go to the site they list.

diesieben07:
> The last section may also happen due to configs not matching I believe thus it might be wise to add something in there mentioning configs. Perhaps something like "The following mods have mismatched registry data, this can be caused by the common configs of the mod not matching, your client mod version does not match the servers or your client does not have the mod installed."

If this is caused by configs not matching, the mod in question has a bug. Forge specifically syncs server configs already. Common configs should not need to be identical between server and client. If they need to be, the mod is badly programmed.

src/main/java/net/minecraftforge/network/ConnectionData.java
```diff
@@ -45,9 +52,9 @@
      *    People using hacked clients WILL hack the mod lists to make them look correct.
      *    Do not use this as an anti-cheat feature!
      *  </b>
-     * @return An immutable list of MODIDs
+     * @return An immutable list of MODIDs and their respective mod versions.
      */
-    public ImmutableList<String> getModList()
+    public ImmutableMap<String, String> getModList()
```
diesieben07: This is a breaking change.

src/main/java/net/minecraftforge/network/HandshakeMessages.java
```diff
@@ -68,7 +69,7 @@ public int getAsInt() {
 
         public S2CModList()
         {
-            this.mods = ModList.get().getMods().stream().map(IModInfo::getModId).collect(Collectors.toList());
+            this.mods = ModList.get().getMods().stream().map(mod -> mod.getModId() + "#" + mod.getVersion().toString()).collect(Collectors.toList());
```
diesieben07: This is not backwards compatible. Old clients do not understand this format.

diesieben07 (changes requested):

src/main/java/net/minecraftforge/network/HandshakeMessages.java
RedstoneDubstep: Sending the version within the mod id string was the only reasonable way I could think of for sending the mod version without changing the packet (if you can think of a better one, I'm open for suggestions). Old clients won't crash when receiving this packet, they will just store the `modid#version` string, which needs to be handled specially by mods who use `getModList()`. That, in my opinion, is not too bad of a trade-off for sending the mod version without a crash.

diesieben07: You could send another packet as prefix to this one.

RedstoneDubstep: in case you mean an entirely new packet for sending the versions to the client: I don't know if I'm a fan of adding yet another packet for this feature to work, especially if it's not needed (since the current approach works without it in a semi-backwards-compatible way, as described above). Of course, if you think that old clients not understanding the new server's format is a problem to that extent that the addition of another packet (or at that point, an edit in `S2CModList` because it'd be as backwards incompatible as a new packet) is warranted, I'll try to do that - though then this PR is probably not gonna get merged until 1.19.

SirEndii: Sure, backwards compatibility is important.
But clients and servers should always share the same forge version. I know, there are some who don't do that.
But updating to a newer version of forge isn't magic, just updating it in the launcher of choice or downloading a newer version from the forge site.
Would be sad to wait for 1.19 for this wonderful new feature.

diesieben07: Backwards compatibility is an important part of Forge and I doubt this PR will be merged in 1.18 without being backwards compatible.

A new packet is not backwards incompatible, it will simply be ignored by older clients. Your current change is not even "semi backwards compatible", because it changes the data in NetworkHooks.getConnectionData

RedstoneDubstep: Wait, so a packet sent from the server that doesn't get recorgnized by the client (for example, if the client is outdated) does not cause the client to crash/not be able to complete the handshake? I thought that adding a new packet would inevitably make this PR a breaking change for that reason. If that's not the case, sorry for the misunderstanding, I will definitely use a packet for sending the data from the server to the client

diesieben07: See `ClientHandshakePacketListenerImpl#handleCustomQuery` and `NetworkHooks.onCustomPayload`. If I read the code correctly, an unrecognized packet should not do anything harmful.

Claycorp:
> If this is caused by configs not matching, the mod in question has a bug. Forge specifically syncs server configs already. Common configs should not need to be identical between server and client. If they need to be, the mod is badly programmed.

And according to many (especially Forge devs), like 70% of the mods made for MC are made poorly. Just because something was made poorly isn't an excuse to not notify the user of a possible issue that could commonly cause the problem. Especially when dealing with stuff provided by Forge itself. It would be understandable to not notify if it's due to some other API or Library mod, but that's not the case.

There will always be people using things "wrong", it's better for the user to know what's going on vs punishing them for something someone else did because of whatever reason that other person decided. If you don't want people doing "wrong" things, you can't allow them to do "wrong" things, otherwise you should warn people about those common "wrong" things. That's like the basics of every error handling system under the sun.

XFactHD:
> If you don't want people doing "wrong" things, you can't allow them to do "wrong" things, otherwise you should warn people about those common "wrong" things. That's like the basics of every error handling system under the sun.

The thing is that this is already the case for configs: they only load after the registries are filled and locked. There should literally be no way for a config setting to produce a mismatched registry.
If someone goes out of their way to break through this restriction by manually loading the config through the Forge config system or with their own homebrew config system (which people have done), then it is IMO not the responsibility of Forge to safe-guard against or warn about that.

RedstoneDubstep:
I've now, as previously suggested, used a second packet (called `S2CModData`, prefixing `S2CModList`) to send server mod data to the client at the beginning of the handshake. I've also fixed that clients receiving unknown handshake packets could not connect to the server, and because of that, this PR is no longer a breaking change. 
My results of the backwards compatibility tests:

Old client -> New server (comparing to before this PR):
- Mismatch on the client: 1 clientside warning log message about unknown packet (`S2CModData`), then default behavior (old disconnection screen with correct message)
- Mismatch on the server:  2 clientside warning log messages about unkown packets (`S2CModData` and `S2CModMismatchData`), then default behavior (old disconnection screen with uninformative message)
- No mismatch: 1 clientside warning log message about unknown packet (`S2CModData`), then default behavior (client joins the server)

New client -> Old server (comparing to before this PR):
- Mismatch on the client: Since some data is missing because `S2CModData` never arrived, the information on the new disconnection screen will be incomplete (only affects the server mod version in case of a mismatch though), but otherwise behaves like New client -> New server
- Mismatch on the server: The client will never be notified of a mismatch on the server, so default behavior applies (old disconnection screen with uninformative message)
- No mismatch: The new packets aren't actually needed on the client if no mismatch is present, so default behavior applies (client joins the server)

src/main/java/net/minecraftforge/network/ConnectionData.java
```diff
@@ -49,7 +58,22 @@
      */
     public ImmutableList<String> getModList()
     {
-        return modList;
+        return modData.keySet().asList();
+    }
+
+    /**
+     * Returns the list of mods and respective mod names and versions present in the remote.
```
diesieben07: Still mentioning list here

src/main/java/net/minecraftforge/network/ConnectionData.java
```diff
@@ -66,4 +90,76 @@
     {
         return channels;
     }
+
+    /**
+     * A class for holding the mod mismatch data of a failed handshake.
+     * Contains a list of mismatched channels, the channels present on the side the handshake failed on, the mods with mismatching registries (if available) and the information of whether the mismatching data's origin is the server.
+     * @param mismatchedChannelData The ids and respective mod versions of the mismatched or missing channels
+     * @param presentChannelData The ids and respective mod names and versions of the present channels corresponding to the mismatched channels, stored like this: [channelId -> [modName, modVersion]]
+     * @param missingRegistryModData The ids and names of mods that have mismatched registry data
```
diesieben07: Missing `@param` for `mismatchedDataFromServer`

src/main/java/net/minecraftforge/network/ConnectionData.java
```diff
@@ -66,4 +90,76 @@
...
+     */
+    public record ModMismatchData(Map<ResourceLocation, String> mismatchedChannelData, Map<ResourceLocation, Pair<String, String>> presentChannelData, Map<String, String> missingRegistryModData, boolean mismatchedDataFromServer)
```
diesieben07: In this class you should use `Map.of` for empty maps instead of new `HashMap`s every time.

src/main/java/net/minecraftforge/network/ConnectionData.java
```diff
@@ -66,4 +90,76 @@
...
+        public static Map<ResourceLocation, String> toRemoteChannelData(Map<ResourceLocation, String> mismatchedChannels, ConnectionData connectionData, boolean mismatchedDataFromServer)
+        {
+            Map<String, String> remoteModVersions;
+
+            if (mismatchedDataFromServer) { //enhance with data from the server
+                remoteModVersions = connectionData != null ? connectionData.getModData().entrySet().stream().collect(Collectors.toMap(Entry::getKey, e -> e.getValue().getRight())) : new HashMap<>();
+            }
+            else { //enhance with local data
+                remoteModVersions = ModList.get().getMods().stream().map(info -> Pair.of(info.getModId(), info.getVersion().toString())).collect(Collectors.toMap(Pair::getLeft, Pair::getRight));
+            }
```
diesieben07: Formatting (braces on new lines)

src/main/java/net/minecraftforge/network/ConnectionData.java
```diff
@@ -66,4 +90,76 @@
...
+
+        public static Map<ResourceLocation, Pair<String, String>> getLocalChannelData(Map<ResourceLocation, String> mismatchedChannelsFilter, ConnectionData connectionData, boolean mismatchedDataFromServer)
+        {
+            Map<ResourceLocation, String> localChannelData;
+
+            if (mismatchedDataFromServer) //mismatch happened on the client, use client data
+                localChannelData = NetworkRegistry.buildChannelVersions();
+            else { //mismatch happened on the server, use server data
+                localChannelData = connectionData != null ? connectionData.getChannels() : new HashMap<>();
+            }
```
diesieben07: Formatting (missing braces / braces should be on new lines)

src/main/java/net/minecraftforge/network/ConnectionData.java
```diff
@@ -66,4 +90,76 @@
...
+        }
+
+        public static Pair<ResourceLocation, Pair<String, String>> getLocalModDataFromChannel(ResourceLocation channel, ConnectionData connectionData, boolean mismatchedDataFromServer)
+        {
+            if (mismatchedDataFromServer) {
+                return ModList.get().getModContainerById(channel.getNamespace()).map(modContainer -> Pair.of(channel, Pair.of(modContainer.getModInfo().getDisplayName(), modContainer.getModInfo().getVersion().toString()))).orElse(Pair.of(channel, Pair.of(channel.getNamespace(), "")));
+            } else {
+                Map<String, Pair<String, String>> modData = connectionData != null ? connectionData.getModData() : new HashMap<>();
+                return Pair.of(channel, modData.getOrDefault(channel.getNamespace(), Pair.of(channel.getNamespace(), "")));
+            }
```
diesieben07: Formatting (braces on new lines)

src/main/java/net/minecraftforge/network/HandshakeHandler.java
```diff
@@ -211,18 +222,31 @@ void handleServerModListOnClient(HandshakeMessages.S2CModList serverModList, Sup
...
+            NetworkConstants.handshakeChannel.reply(new HandshakeMessages.S2CModMismatchData(mismatchedChannels), c.get());
             c.get().getNetworkManager().disconnect(new TextComponent("Connection closed - mismatched mod channel list"));
             return;
         }
         LOGGER.debug(FMLHSMARKER, "Accepted client connection mod list");
     }
 
+    void handleModMismatchData(HandshakeMessages.S2CModMismatchData modMismatchData, Supplier<NetworkEvent.Context> c)
+    {
+        if (!modMismatchData.getMismatchedChannelData().isEmpty()) {
```
diesieben07: Formatting (braces on new lines)

src/main/java/net/minecraftforge/network/HandshakeMessages.java
```diff
@@ -68,7 +69,7 @@ public int getAsInt() {
 
         public S2CModList()
         {
-            this.mods = ModList.get().getMods().stream().map(IModInfo::getModId).collect(Collectors.toList());
+            this.mods = ModList.get().getMods().stream().map(mod -> mod.getModId()).collect(Collectors.toList());
```
diesieben07: Why?

src/main/java/net/minecraftforge/network/HandshakeMessages.java
```diff
@@ -128,6 +129,43 @@ public void encode(FriendlyByteBuf output)
...
+    public static class S2CModData extends LoginIndexedMessage implements INoResponse
+    {
+        private final Map<String, Pair<String, String>> mods;
+
+        public S2CModData()
+        {
+            this.mods = ModList.get().getMods().stream().map(info -> Pair.of(info.getModId(), Pair.of(info.getDisplayName(), info.getVersion().toString()))).collect(Collectors.toMap(Pair::getLeft, Pair::getRight));
+        }
+
+        private S2CModData(Map<String, Pair<String, String>> mods) {
```
diesieben07: Formatting (braces on new lines)

src/main/java/net/minecraftforge/network/NetworkEvent.java
```diff
@@ -121,6 +121,10 @@ public void add(FriendlyByteBuf buffer, ResourceLocation channelName, String con
             collected.add(new NetworkRegistry.LoginPayload(buffer, channelName, context));
         }
 
+        public void add(FriendlyByteBuf buffer, ResourceLocation channelName, String context, boolean needsResponse) {
```
diesieben07: Formatting (braces on new lines)

src/main/java/net/minecraftforge/network/NetworkRegistry.java
```diff
@@ -330,10 +330,20 @@ public static void lock() {
...
+        /**
+         * If the connection should await a response to this packet to continue
+         */
+        private final boolean needsResponse;
+
         public LoginPayload(final FriendlyByteBuf buffer, final ResourceLocation channelName, final String messageContext) {
+            this(buffer, channelName, messageContext, true);
+        }
+
+        public LoginPayload(final FriendlyByteBuf buffer, final ResourceLocation channelName, final String messageContext, final boolean needsResponse) {
```
diesieben07: Formatting (braces on new lines)

src/main/java/net/minecraftforge/network/NetworkRegistry.java
```diff
@@ -347,6 +357,10 @@ public ResourceLocation getChannelName() {
         public String getMessageContext() {
             return messageContext;
         }
+
+        public boolean needsResponse() {
```
diesieben07: Formatting (braces on new lines)

diesieben07 (changes requested):
Just a couple generic formatting reviews for now, I will go through this in detail on the weekend

src/main/java/net/minecraftforge/network/HandshakeMessages.java
RedstoneDubstep: That's a remnant of the old way of sending the mod version via a `#` as a separator, will fix

src/main/java/net/minecraftforge/network/NetworkRegistry.java
RedstoneDubstep: It kind of feels off changing this one method to have braces on newlines if most, if not all of the other braces of this inner class are not on a new line either, but whatever

src/main/java/net/minecraftforge/network/HandshakeMessages.java
```diff
@@ -128,6 +129,44 @@ public void encode(FriendlyByteBuf output)
...
+     * Prefixes S2CModList by sending additional data about the mods installed on the server to the client
+     * The mod data is stored as follows: [modId -> [modName, modVersion]]
+     */
+    public static class S2CModData extends LoginIndexedMessage implements INoResponse
+    {
+        private final Map<String, Pair<String, String>> mods;
+
+        public S2CModData()
+        {
+            this.mods = ModList.get().getMods().stream().map(info -> Pair.of(info.getModId(), Pair.of(info.getDisplayName(), info.getVersion().toString()))).collect(Collectors.toMap(Pair::getLeft, Pair::getRight));
```
diesieben07:
```suggestion
            this.mods = ModList.get().getMods().stream().collect(Collectors.toMap(IModInfo::getModId, info -> Pair.of(info.getDisplayName(), info.getVersion().toString())));
```

src/main/java/net/minecraftforge/network/HandshakeMessages.java
```diff
@@ -275,4 +314,36 @@ public String getFileName() {
             return fileData;
         }
     }
+
+    public static class S2CModMismatchData extends LoginIndexedMessage {
```
diesieben07: Please add documentation when this is sent.

src/main/java/net/minecraftforge/network/HandshakeMessages.java
```diff
@@ -275,4 +314,36 @@ public String getFileName() {
...
+        public Map<ResourceLocation, String> getMismatchedChannelData()
+        {
+            return mismatchedChannelData;
+        }
+    }
+
+    /**
+     * Marker interface for all server -> client packets that the server should not await a response from the client for
+     */
+    public interface INoResponse {
```
diesieben07: Why is this a marker interface instead of a setting on the channel? If it stays as a marker interface (which I do not like, it should instead be a setting in `SimpleChannel.MessageBuilder`), it definitely does not belong in this file.
Also: Formatting, braces on new lines.

src/main/java/net/minecraftforge/network/simple/IndexedMessageCodec.java
```diff
@@ -150,6 +150,7 @@ MSG newInstance() {
     void consume(FriendlyByteBuf payload, int payloadIndex, Supplier<NetworkEvent.Context> context) {
         if (payload == null) {
             LOGGER.error(SIMPLENET, "Received empty payload on channel {}", Optional.ofNullable(networkInstance).map(NetworkInstance::getChannelName).map(Objects::toString).orElse("MISSING CHANNEL"));
+            context.get().setPacketHandled(true); //don't disconnect because of empty packets
```
diesieben07: This should only be done for packets that we actually know of and that do not expect a response.

src/main/java/net/minecraftforge/network/simple/SimpleChannel.java
```diff
@@ -70,7 +71,7 @@ private void networkLoginGather(final NetworkEvent.GatherLoginPayloadsEvent gath
             packetGenerator.apply(gatherEvent.isLocal()).forEach(p->{
                 FriendlyByteBuf pb = new FriendlyByteBuf(Unpooled.buffer());
                 this.indexedCodec.build(p.getRight(), pb);
-                gatherEvent.add(pb, this.instance.getChannelName(), p.getLeft());
+                gatherEvent.add(pb, this.instance.getChannelName(), p.getLeft(), !(p.getRight() instanceof INoResponse));
```
diesieben07: See above, this should not be using a marker interface.

src/main/java/net/minecraftforge/network/ConnectionData.java
```diff
@@ -66,4 +89,86 @@
...
+            Map<ResourceLocation, String> mismatchedRegistryModData = mismatchedRegistryMods.stream().map(id -> ModList.get().getModContainerById(id.getNamespace()).map(modContainer -> Pair.of(id, modContainer.getModInfo().getVersion().toString())).orElse(Pair.of(id, NetworkRegistry.ABSENT))).collect(Collectors.toMap(Pair::getLeft, Pair::getRight));
+            Map<ResourceLocation, Pair<String, String>> presentModData = getServersideModData(mismatchedRegistryModData.keySet(), connectionData).entrySet().stream().collect(Collectors.toMap(e -> new ResourceLocation(e.getKey().getNamespace(), ""), Entry::getValue));
+
+            return new ModMismatchData(mismatchedRegistryModData, presentModData, false); //false because mods marked with NetworkRegistry.ABSENT should be displayed as missing from the client, not the server
+        }
+
+        public static Map<ResourceLocation, String> enhanceWithModData(Map<ResourceLocation, String> mismatchedChannels, ConnectionData connectionData, boolean mismatchedDataFromServer)
```
diesieben07: Why is this method public?

src/main/java/net/minecraftforge/network/ConnectionData.java
```diff
@@ -66,4 +89,86 @@
...
+                remoteModVersions = connectionData != null ? connectionData.getModData().entrySet().stream().collect(Collectors.toMap(Entry::getKey, e -> e.getValue().getRight())) : Map.of();
+            }
+            else //enhance with data from the client
+            {
+                remoteModVersions = ModList.get().getMods().stream().map(info -> Pair.of(info.getModId(), info.getVersion().toString())).collect(Collectors.toMap(Pair::getLeft, Pair::getRight));
+            }
+            return mismatchedChannels.keySet().stream().map(channel -> Pair.of(channel, mismatchedChannels.get(channel).equals(NetworkRegistry.ABSENT) ? NetworkRegistry.ABSENT : remoteModVersions.getOrDefault(channel.getNamespace(), NetworkRegistry.ABSENT))).collect(Collectors.toMap(Pair::getLeft, Pair::getRight));
+        }
+
+        public static Map<ResourceLocation, Pair<String, String>> getChannelDataFromSide(Set<ResourceLocation> mismatchedChannelsFilter, ConnectionData connectionData, boolean fromClientSide)
```
diesieben07: Why is this method public?

src/main/java/net/minecraftforge/network/ConnectionData.java
```diff
@@ -66,4 +89,86 @@
...
+    {
+        public static ModMismatchData channel(Map<ResourceLocation, String> mismatchedChannels, ConnectionData connectionData, boolean mismatchedDataFromServer)
+        {
+            Map<ResourceLocation, String> mismatchedChannelData = enhanceWithModData(mismatchedChannels, connectionData, mismatchedDataFromServer);
+            Map<ResourceLocation, Pair<String, String>> presentChannelData = getChannelDataFromSide(mismatchedChannels.keySet(), connectionData, mismatchedDataFromServer);
+
+            return new ModMismatchData(mismatchedChannelData, presentChannelData, mismatchedDataFromServer);
+        }
+
+        public static ModMismatchData registry(Multimap<ResourceLocation, ResourceLocation> mismatchedRegistryData, ConnectionData connectionData)
```
diesieben07, line 107: This method should have documentation.

src/main/java/net/minecraftforge/network/ConnectionData.java
```diff
@@ -66,4 +89,86 @@
...
+    /**
+     * A class for holding the mod mismatch data of a failed handshake.
+     * Contains a list of mismatched channels, the channels present on the side the handshake failed on, the mods with mismatching registries (if available) and the information of whether the mismatching data's origin is the server.
+     * @param mismatchedModData The ids and respective mod versions of the mismatched or missing mods/channels
+     * @param presentModData The ids and respective mod names and versions corresponding to the mismatched mods/channels, stored like this: [id -> [modName, modVersion]]
+     * @param mismatchedDataFromServer Whether the mismatched data originates from the server. Note that if the mismatched data is from the server, the actual mismatch check happened on the client and vice versa
+     */
+    public record ModMismatchData(Map<ResourceLocation, String> mismatchedModData, Map<ResourceLocation, Pair<String, String>> presentModData, boolean mismatchedDataFromServer)
+    {
+        public static ModMismatchData channel(Map<ResourceLocation, String> mismatchedChannels, ConnectionData connectionData, boolean mismatchedDataFromServer)
```
diesieben07, line 94: This method should have documentation.

src/main/java/net/minecraftforge/network/ConnectionData.java
```diff
@@ -66,4 +89,86 @@
...
+                channelData = NetworkRegistry.buildChannelVersions();
+            }
+            else //mismatch happened on the server, use server data
+            {
+                channelData = connectionData != null ? connectionData.getChannels() : Map.of();
+            }
+            return channelData.keySet().stream().filter(mismatchedChannelsFilter::contains).map(id -> getModDataFromChannel(id, connectionData, fromClientSide)).collect(Collectors.toMap(Pair::getLeft, Pair::getRight));
+        }
+
+        public static Pair<ResourceLocation, Pair<String, String>> getModDataFromChannel(ResourceLocation channel, ConnectionData connectionData, boolean fromClientSide)
```
diesieben07: Why is this method public?

src/main/java/net/minecraftforge/network/ConnectionData.java
```diff
@@ -66,4 +89,86 @@
...
+                return ModList.get().getModContainerById(channel.getNamespace()).map(modContainer -> Pair.of(channel, Pair.of(modContainer.getModInfo().getDisplayName(), modContainer.getModInfo().getVersion().toString()))).orElse(Pair.of(channel, Pair.of(channel.getNamespace(), "")));
+            }
+            else
+            {
+                Map<String, Pair<String, String>> modData = connectionData != null ? connectionData.getModData() : Map.of();
+                return Pair.of(channel, modData.getOrDefault(channel.getNamespace(), Pair.of(channel.getNamespace(), "")));
+            }
+        }
+
+        public static Map<ResourceLocation, Pair<String, String>> getServersideModData(Set<ResourceLocation> mismatchedModsFilter, ConnectionData connectionData)
```
diesieben07: Why is this method public?
Also `Serverside` => `ServerSide`

src/main/java/net/minecraftforge/network/ConnectionData.java
```diff
@@ -66,4 +89,86 @@
...
+        {
+            Map<String, Pair<String, String>> localChannelData = connectionData != null ? connectionData.getModData() : Map.of();
+            Set<String> modIdFilter = mismatchedModsFilter.stream().map(ResourceLocation::getNamespace).collect(Collectors.toSet());
+            return localChannelData.entrySet().stream().filter(e -> modIdFilter.contains(e.getKey())).collect(Collectors.toMap(e -> new ResourceLocation(e.getKey(), ""), Entry::getValue));
+        }
+
+        public boolean containsMismatches()
```
diesieben07: This method should be up with the other public methods and should have documentation

diesieben07 (changes requested):
Generally looks good, but some style and functionality changes are required

src/main/java/net/minecraftforge/network/simple/IndexedMessageCodec.java
RedstoneDubstep: Because the payload is null at this point, it's difficult to look up what packet on the client caused this empty-payload-packet to be sent to the server. I thought logging an error about an empty payload packet being received is sufficient and that the server should just move on with the handshake if the client isn't able to decode a packet (which currently only happens to `S2CModData` anyways, a packet that the server doesn't care about if the client was able to decode it or not). If a packet that is not declared as not needing a response (and thus, for that packet, it is crucial for a response of this packet to be sent) can't be read by the client, the empty payload packet sent because of it will be ignored/not disconnect at this point in the code, but the handshake will fail anyways because the needed response never arrives (see HandshakeHandler#tickServer).

diesieben07: Can't you look it up based on the login index?

RedstoneDubstep: Just thought about it, I could maybe test if the given `payloadIndex` is contained in `sentMessages` (if it is, the packet requires a response, and if not, it doesn't because of the filter in `HandshakeHandler#tickServer`). Should I try to do this (and then not mark the packet as handled if an actual response is required for the packet with that login index)?

diesieben07: If it's possible without too much hackery, sure. If not, it should be OK because of the other thing you mentioned (I will have to test)

RedstoneDubstep: I now added a check as described in my comment above at the desired place, hopefully it isn't considered too hacky (I at least don't think so, but if you have another opinion feel free to tell me)

RedstoneDubstep:
This PR has just been merged with the 1.18.x branch to be up to date with the RB. The last few commits significantly improved documentation and the bundled test mod (which now actually covers all cases in which a ModMismatchDisconnectedScreen would appear), so a re-review by a triager would be greatly appreciated to figure out if this PR goes into the right direction or if something should be added/changed or cut entirely.

PR #8405, opened 2022-01-19
[1.18.x] Added a Hunger-System API

Parker8283:
_This is the 1.18 version of #7266. The below is mostly copy-pasta'd from that PR's description since there were no substantial changes in the port._

---

**This is an event-driven API that allows for control of the hunger system in the game. Things like food, health regen, hunger regen, exhaustion, starvation, etc. This was actually a PR I had made 6 years ago (#1758), but broke and never fixed. Until now!**

The goal of this system is to allow modders to change values/behaviors that were previously hard-coded in the game. Some examples include:

* Changing the nutrition and saturation values of foods eaten
* Allow foods to have different nutrition and saturation values per-player
* Change the amount of exhaustion added for each exhausting action the player does
* Control how and when health and hunger regeneration occurs
* Control how and when starvation occurs
* and more!

There is a lot that is allowed by this system. A good overview of everything would be to check the test mods that are included with this PR.

Other more technical things that this PR does include:

* `FoodData` now takes a `Player` upon construction that is used throughout the class when querying values. This allows `FoodData` to operate more player-dependently than before, allowing for operations to occur based on what player is trying to perform them.
* The `foodTimer` in `FoodData` has been separated into the `foodTimer` and `starveTimer`. This allows starvation to happen independently from the exhaustion/regeneration timer.
* A `FoodValues` class is provided to allow for easy use of nutrition and saturation values for food objects.
* An `IEdible` interface is provided that can be applied to non-standard food objects so that they can be included in the system. For a vanilla example, Cake now implements this interface. It is still recommended and preferred to include a `FoodProperties` object when constructing your item over implementing this interface.
  * The original author of this system, @squeek502, notes that he was never completely satisfied with this approach of including non-standard food items and wonders if there is a better solution. If you have an idea, please let me know!
* Fire events while in peaceful mode to allow modders to control how health/hunger regen happens in peaceful

---

This system would be used by mods such as [Hunger Overhaul](https://www.curseforge.com/minecraft/mc-mods/hunger-overhaul) and [The Spice Of Life](https://www.curseforge.com/minecraft/mc-mods/the-spice-of-life) to fundamentally change how the hunger system works in the game.

This system was originally written by @squeek502 for [AppleCore](https://www.curseforge.com/minecraft/mc-mods/applecore), and was first attempted to be upstreamed by me way back in the 1.7.10/1.8 days (as noted above). He has offered his code into the [Public Domain](https://github.com/squeek502/AppleCore/blob/1.12.2/LICENSE), and as such I am re-attempting to put this system into Forge! Since this code was originally written for a different Minecraft version, there is a good chance that there are better ways to do some of what we are trying to do. If that is the case, please let me know and I will gladly learn the new systems and how best to be doing what we are attempting. (I haven't really been actively modding Minecraft since the 1.8 days, haha)

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
@@ -8,6 +8,15 @@
     public static final String f_150081_ = "OfflinePlayer:";
     public static final int f_150082_ = 16;
     public static final int f_150083_ = 20;
+@@ -136,7 +_,7 @@
+    protected PlayerEnderChestContainer f_36094_ = new PlayerEnderChestContainer();
+    public final InventoryMenu f_36095_;
+    public AbstractContainerMenu f_36096_;
+-   protected FoodData f_36097_ = new FoodData();
++   protected FoodData f_36097_ = new FoodData(this);
+    protected int f_36098_;
+    public float f_36099_;
+    public float f_36100_;
 @@ -162,6 +_,9 @@
     private final ItemCooldowns f_36087_ = this.m_7478_();
     @Nullable
@@ -63,6 +72,22 @@
        if (this.m_20175_(Pose.SWIMMING)) {
           Pose pose;
           if (this.m_21255_()) {
+@@ -469,11 +_,13 @@
+ 
+       if (this.f_19853_.m_46791_() == Difficulty.PEACEFUL && this.f_19853_.m_46469_().m_46207_(GameRules.f_46139_)) {
+          if (this.m_21223_() < this.m_21233_() && this.f_19797_ % 20 == 0) {
+-            this.m_5634_(1.0F);
++            net.minecraftforge.event.hunger.HealthRegenEvent.PeacefulRegen peacefulHealthRegenEvent = net.minecraftforge.event.ForgeEventFactory.onPeacefulHealthRegen(this);
++            if (!peacefulHealthRegenEvent.isCanceled()) this.m_5634_(peacefulHealthRegenEvent.getDeltaHealth());
+          }
+ 
+          if (this.f_36097_.m_38721_() && this.f_19797_ % 10 == 0) {
+-            this.f_36097_.m_38705_(this.f_36097_.m_38702_() + 1);
++            net.minecraftforge.event.hunger.HungerRegenEvent.PeacefulRegen peacefulHungerRegenEvent = net.minecraftforge.event.ForgeEventFactory.onPeacefulHungerRegen(this);
++            if (!peacefulHungerRegenEvent.isCanceled()) this.f_36097_.addFood(peacefulHungerRegenEvent.getDeltaHunger());
+          }
+       }
+ 
 @@ -569,6 +_,7 @@
     }
  
@@ -141,7 +166,7 @@
              });
              if (this.f_20935_.m_41619_()) {
                 if (interactionhand == InteractionHand.MAIN_HAND) {
-@@ -877,10 +_,13 @@
+@@ -877,20 +_,23 @@
  
     protected void m_6475_(DamageSource p_36312_, float p_36313_) {
        if (!this.m_6673_(p_36312_)) {
@@ -155,9 +180,11 @@
           float f = p_36313_ - f2;
           if (f > 0.0F && f < 3.4028235E37F) {
              this.m_36222_(Stats.f_12933_, Math.round(f * 10.0F));
-@@ -889,8 +_,8 @@
+          }
+ 
           if (f2 != 0.0F) {
-             this.m_36399_(p_36312_.m_19377_());
+-            this.m_36399_(p_36312_.m_19377_());
++            this.m_36399_(net.minecraftforge.event.ForgeEventFactory.onExhaustionAdded(this, net.minecraftforge.event.hunger.ExhaustionEvent.ExhaustingActions.DAMAGE_TAKEN, p_36312_.m_19377_()));
              float f1 = this.m_21223_();
 -            this.m_21153_(this.m_21223_() - f2);
              this.m_21231_().m_19289_(p_36312_, f1, f2);
@@ -259,6 +286,15 @@
                          this.m_21008_(InteractionHand.MAIN_HAND, ItemStack.f_41583_);
                       }
                    }
+@@ -1176,7 +_,7 @@
+                      }
+                   }
+ 
+-                  this.m_36399_(0.1F);
++                  this.m_36399_(net.minecraftforge.event.ForgeEventFactory.onExhaustionAdded(this, net.minecraftforge.event.hunger.ExhaustionEvent.ExhaustingActions.ATTACK_ENTITY, 0.1F));
+                } else {
+                   this.f_19853_.m_6263_((Player)null, this.m_20185_(), this.m_20186_(), this.m_20189_(), SoundEvents.f_12315_, this.m_5720_(), 1.0F, 1.0F);
+                   if (flag4) {
 @@ -1200,7 +_,7 @@
        }
  
@@ -285,6 +321,58 @@
        } else {
           boolean flag = block.m_5568_();
           boolean flag1 = p_36131_.m_8055_(p_36132_.m_7494_()).m_60734_().m_5568_();
+@@ -1337,9 +_,9 @@
+       super.m_6135_();
+       this.m_36220_(Stats.f_12926_);
+       if (this.m_20142_()) {
+-         this.m_36399_(0.2F);
++         this.m_36399_(net.minecraftforge.event.ForgeEventFactory.onExhaustionAdded(this, net.minecraftforge.event.hunger.ExhaustionEvent.ExhaustingActions.SPRINTING_JUMP, 0.2F));
+       } else {
+-         this.m_36399_(0.05F);
++         this.m_36399_(net.minecraftforge.event.ForgeEventFactory.onExhaustionAdded(this, net.minecraftforge.event.hunger.ExhaustionEvent.ExhaustingActions.NORMAL_JUMP, 0.05F));
+       }
+ 
+    }
+@@ -1397,19 +_,19 @@
+             int i = Math.round((float)Math.sqrt(p_36379_ * p_36379_ + p_36380_ * p_36380_ + p_36381_ * p_36381_) * 100.0F);
+             if (i > 0) {
+                this.m_36222_(Stats.f_12924_, i);
+-               this.m_36399_(0.01F * (float)i * 0.01F);
++               this.m_36399_(net.minecraftforge.event.ForgeEventFactory.onExhaustionAdded(this, net.minecraftforge.event.hunger.ExhaustionEvent.ExhaustingActions.MOVEMENT_SWIM, 0.01F * (float)i * 0.01F));
+             }
+          } else if (this.m_204029_(FluidTags.f_13131_)) {
+             int j = Math.round((float)Math.sqrt(p_36379_ * p_36379_ + p_36380_ * p_36380_ + p_36381_ * p_36381_) * 100.0F);
+             if (j > 0) {
+                this.m_36222_(Stats.f_13001_, j);
+-               this.m_36399_(0.01F * (float)j * 0.01F);
++               this.m_36399_(net.minecraftforge.event.ForgeEventFactory.onExhaustionAdded(this, net.minecraftforge.event.hunger.ExhaustionEvent.ExhaustingActions.MOVEMENT_WALK_UNDERWATER, 0.01F * (float)j * 0.01F));
+             }
+          } else if (this.m_20069_()) {
+             int k = Math.round((float)Math.sqrt(p_36379_ * p_36379_ + p_36381_ * p_36381_) * 100.0F);
+             if (k > 0) {
+                this.m_36222_(Stats.f_12997_, k);
+-               this.m_36399_(0.01F * (float)k * 0.01F);
++               this.m_36399_(net.minecraftforge.event.ForgeEventFactory.onExhaustionAdded(this, net.minecraftforge.event.hunger.ExhaustionEvent.ExhaustingActions.MOVEMENT_WALK_ONWATER, 0.01F * (float)k * 0.01F));
+             }
+          } else if (this.m_6147_()) {
+             if (p_36380_ > 0.0D) {
+@@ -1420,13 +_,13 @@
+             if (l > 0) {
+                if (this.m_20142_()) {
+                   this.m_36222_(Stats.f_12996_, l);
+-                  this.m_36399_(0.1F * (float)l * 0.01F);
++                  this.m_36399_(net.minecraftforge.event.ForgeEventFactory.onExhaustionAdded(this, net.minecraftforge.event.hunger.ExhaustionEvent.ExhaustingActions.MOVEMENT_SPRINT, 0.1F * (float)l * 0.01F));
+                } else if (this.m_6047_()) {
+                   this.m_36222_(Stats.f_12995_, l);
+-                  this.m_36399_(0.0F * (float)l * 0.01F);
++                  this.m_36399_(net.minecraftforge.event.ForgeEventFactory.onExhaustionAdded(this, net.minecraftforge.event.hunger.ExhaustionEvent.ExhaustingActions.MOVEMENT_CROUCH, 0.0F * (float)l * 0.01F));
+                } else {
+                   this.m_36222_(Stats.f_12994_, l);
+-                  this.m_36399_(0.0F * (float)l * 0.01F);
++                  this.m_36399_(net.minecraftforge.event.ForgeEventFactory.onExhaustionAdded(this, net.minecraftforge.event.hunger.ExhaustionEvent.ExhaustingActions.MOVEMENT_WALK_ONLAND, 0.0F * (float)l * 0.01F));
+                }
+             }
+          } else if (this.m_21255_()) {
 @@ -1465,6 +_,7 @@
  
     public boolean m_142535_(float p_150093_, float p_150094_, DamageSource p_150095_) {
--- a/patches/minecraft/net/minecraft/world/food/FoodData.java.patch
+++ b/patches/minecraft/net/minecraft/world/food/FoodData.java.patch
@@ -1,0 +1,176 @@
+--- a/net/minecraft/world/food/FoodData.java
++++ b/net/minecraft/world/food/FoodData.java
+@@ -14,64 +_,95 @@
+    private float f_38698_;
+    private int f_38699_;
+    private int f_38700_ = 20;
++   // FORGE START
++   private Player player;
++   private int starveTimer;
++   // FORGE END
+ 
++   @Deprecated // Forge: Provided for backwards compatibility. Use the Player-specific constructor instead.
+    public FoodData() {
++      this(null);
++   }
++
++   public FoodData(Player player) {
+       this.f_38697_ = 5.0F;
++      this.player = player;
+    }
+ 
++   // Forge: Don't call this unless the player is not consuming food. Use eat(Item, ItemStack) instead.
+    public void m_38707_(int p_38708_, float p_38709_) {
+-      this.f_38696_ = Math.min(p_38708_ + this.f_38696_, 20);
++      if (net.minecraftforge.event.ForgeEventFactory.onFoodDataAddition(this.player, new net.minecraftforge.common.hunger.FoodValues(p_38708_, p_38709_))) return;
++      this.f_38696_ = Math.min(p_38708_ + this.f_38696_, net.minecraftforge.event.ForgeEventFactory.getMaxHunger(this.player));
+       this.f_38697_ = Math.min(this.f_38697_ + (float)p_38708_ * p_38709_ * 2.0F, (float)this.f_38696_);
+    }
+ 
+    public void m_38712_(Item p_38713_, ItemStack p_38714_) {
+-      if (p_38713_.m_41472_()) {
+-         FoodProperties foodproperties = p_38713_.m_41473_();
+-         this.m_38707_(foodproperties.m_38744_(), foodproperties.m_38745_());
++      net.minecraftforge.common.hunger.FoodValues modifiedFoodValues = net.minecraftforge.common.hunger.FoodValues.get(p_38714_, this.player);
++      if (modifiedFoodValues != null) {
++         int prevFoodLevel = this.f_38696_;
++         float prevSaturationLevel = this.f_38697_;
++         this.m_38707_(modifiedFoodValues.nutrition(), modifiedFoodValues.saturationModifier());
++         net.minecraftforge.event.ForgeEventFactory.onFoodEaten(modifiedFoodValues, this.f_38696_ - prevFoodLevel, this.f_38697_ - prevSaturationLevel, p_38714_, this.player);
+       }
+ 
+    }
+ 
+    public void m_38710_(Player p_38711_) {
+-      Difficulty difficulty = p_38711_.f_19853_.m_46791_();
++      if (this.player == null) this.player = p_38711_; // Forge: Safety-check, in-case the class was constructed without the player-specific constructor
+       this.f_38700_ = this.f_38696_;
+-      if (this.f_38698_ > 4.0F) {
+-         this.f_38698_ -= 4.0F;
+-         if (this.f_38697_ > 0.0F) {
+-            this.f_38697_ = Math.max(this.f_38697_ - 1.0F, 0.0F);
+-         } else if (difficulty != Difficulty.PEACEFUL) {
+-            this.f_38696_ = Math.max(this.f_38696_ - 1, 0);
++      net.minecraftforge.eventbus.api.Event.Result allowExhaustionResult = net.minecraftforge.event.ForgeEventFactory.fireAllowExhaustionEvent(p_38711_);
++      float maxExhaustion = net.minecraftforge.event.ForgeEventFactory.getMaxExhaustion(p_38711_);
++      if (allowExhaustionResult == net.minecraftforge.eventbus.api.Event.Result.ALLOW || (allowExhaustionResult == net.minecraftforge.eventbus.api.Event.Result.DEFAULT && this.f_38698_ > maxExhaustion)) {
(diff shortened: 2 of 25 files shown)


Minecraftschurli (changes requested):
A few things:
- How does this interact with vanillas `FoodProperties` set in the item properties?
- add doc comments on the setters and getters in the events

Parker8283:
* `FoodProperties` is still the primary and preferred way to create food items, as noted in the header of `IEdible`. This gets incorporated in the `FoodValues` class, when someone calls one of the `get` methods in that class to get the `FoodValues` of a given `ItemStack`. As such, modders should generally _not_ need to change anything to take full advantage of this system. However, someone directly accessing the values from the `FoodProperties` for their own uses might bypass parts of this system. Note that this does not break anything and should still result in the expected behavior from that mod's perspective. It just wouldn't fire the new `GetFoodValues` event to allow other mods to modify those food values dynamically. There might be room for a patch to `FoodProperties` to notate this somehow to modders, and add a new function that does return the "more proper" values. However since that function is already in `FoodValues`, coupled with my (unsubstantiated) notion that most modders probably don't directly call the getters in `FoodProperties`, and the want to keep patches small, I have forgone this. Let me know if its something that is wanted. (I had a comment of similar nature with more detail in the 1.16 PR #7266, feel free to read that as well)
* Done. Wasn't sure if these getters/setters were too trivial to have comments, but I guess they couldn't hurt.

patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
```diff
@@ -63,6 +72,22 @@
...
+          if (this.m_21223_() < this.m_21233_() && this.f_19797_ % 20 == 0) {
+-            this.m_5634_(1.0F);
++            net.minecraftforge.event.hunger.HealthRegenEvent.PeacefulRegen peacefulHealthRegenEvent = net.minecraftforge.event.ForgeEventFactory.onPeacefulHealthRegen(this);
++            if (!peacefulHealthRegenEvent.isCanceled()) this.m_5634_(peacefulHealthRegenEvent.getDeltaHealth());
+          }
+ 
+          if (this.f_36097_.m_38721_() && this.f_19797_ % 10 == 0) {
+-            this.f_36097_.m_38705_(this.f_36097_.m_38702_() + 1);
++            net.minecraftforge.event.hunger.HungerRegenEvent.PeacefulRegen peacefulHungerRegenEvent = net.minecraftforge.event.ForgeEventFactory.onPeacefulHungerRegen(this);
++            if (!peacefulHungerRegenEvent.isCanceled()) this.f_36097_.m_38705_(this.f_36097_.m_38702_() + peacefulHungerRegenEvent.getDeltaHunger());
```
Shadows-of-Fire: This sort of call `FoodData#setFoodLevel(FoodData#getFoodLevel + change)` seems common enough to warrant a new method on `FoodData` that performs this operation.  Likely for both Food and Saturation.
Possibly `addFood` and `addSaturation`

patches/minecraft/net/minecraft/world/food/FoodData.java.patch
```diff
@@ -0,0 +1,145 @@
...
+    public FoodData() {
++      this(null);
++   }
++
++   public FoodData(Player player) {
+       this.f_38697_ = 5.0F;
++      this.player = player;
+    }
+ 
++   @Deprecated // Forge: Don't call this unless the player is not consuming food. Use eat(Item, ItemStack) instead.
```
Shadows-of-Fire: I don't believe this method deserves deprecation, it has valid use-cases, since there are plenty of non-item sources which apply food (even one in vanilla, being the saturation potion).  The comment being there is fine though.

patches/minecraft/net/minecraft/world/food/FoodData.java.patch
```diff
@@ -0,0 +1,145 @@
...
+-         this.f_38698_ -= 4.0F;
+-         if (this.f_38697_ > 0.0F) {
+-            this.f_38697_ = Math.max(this.f_38697_ - 1.0F, 0.0F);
+-         } else if (difficulty != Difficulty.PEACEFUL) {
+-            this.f_38696_ = Math.max(this.f_38696_ - 1, 0);
++      net.minecraftforge.eventbus.api.Event.Result allowExhaustionResult = net.minecraftforge.event.ForgeEventFactory.fireAllowExhaustionEvent(p_38711_);
++      float maxExhaustion = net.minecraftforge.event.ForgeEventFactory.getMaxExhaustion(p_38711_);
++      if (allowExhaustionResult == net.minecraftforge.eventbus.api.Event.Result.ALLOW || (allowExhaustionResult == net.minecraftforge.eventbus.api.Event.Result.DEFAULT && this.f_38698_ > maxExhaustion)) {
++         net.minecraftforge.event.hunger.ExhaustionEvent.Exhausted exhaustedEvent = net.minecraftforge.event.ForgeEventFactory.onExhausted(p_38711_, maxExhaustion, this.f_38698_);
++         this.f_38698_ += exhaustedEvent.getDeltaExhaustion();
```
Shadows-of-Fire: I think it makes sense for this value to not be updated if the event is cancelled as well, so it should probably be moved into the if-block.

patches/minecraft/net/minecraft/world/food/FoodData.java.patch
```diff
@@ -0,0 +1,145 @@
+--- a/net/minecraft/world/food/FoodData.java
++++ b/net/minecraft/world/food/FoodData.java
+@@ -14,64 +_,95 @@
+    private float f_38698_;
+    private int f_38699_;
+    private int f_38700_ = 20;
++   // FORGE START
++   private Player player;
++   private int starveTimer;
```
Shadows-of-Fire, line 9: The starve timer needs to be de/serialized in `readAdditionalSaveData` / `addAdditionalSaveData`

src/main/java/net/minecraftforge/client/gui/ForgeIngameGui.java
```diff
@@ -570,7 +571,24 @@ public void renderFood(int width, int height, PoseStack mStack)
...
+        int level;
+        // Only set level to 0 if the real hunger value is 0
+        if (realHunger == 0)
+        {
+            level = 0;
+        }
+        else
+        {
+            // Return a scaled value so that the HUD can still use the same logic as if the max was 20.
+            float scale = 20.0F / ForgeEventFactory.getMaxHunger(minecraft.player);
```
Shadows-of-Fire: I can't say I agree with scaling the hunger bar to fit the current maximum (because healing still works at a flat 20, and not at the current maximum).

I think higher maximums should just require the mod causing higher maximums to draw additional food shanks, unless we want forge to also draw them.

The player has some expectation the a single food shank represents a fixed value, not a percentage of their current food max.

src/main/java/net/minecraftforge/common/hunger/FoodValues.java
```diff
@@ -0,0 +1,165 @@
...
+ *
+ * To get food values for any given food, use any of the static {@link #get} methods.
+ *
+ * <pre>
+ * {@code
+ * FoodValues appleFoodValues = FoodValues.get(new ItemStack(Items.APPLE));
+ * }
+ * </pre>
+ */
+public class FoodValues
```
Shadows-of-Fire: This class could be a record `public record FoodValues(int nutrition, float saturationModifier)`
That would have the added benefit of removing the need to override `hashCode` and `equals`

src/main/java/net/minecraftforge/common/hunger/FoodValues.java
```diff
@@ -0,0 +1,165 @@
...
+            else if (stack.getItem() instanceof IEdible)
+            {
+                return ((IEdible)stack.getItem()).getFoodValues(stack);
+            }
+            else if (stack.getItem() instanceof BlockItem blockItem)
+            {
+                if (blockItem.getBlock() instanceof IEdible)
+                {
+                    return ((IEdible)blockItem.getBlock()).getFoodValues(stack);
+                }
```
Shadows-of-Fire: These instanceofs can probably make use of the new syntax to look a little cleaner
ex:
```
if (stack.getItem() instanceof IEdible edible)
    {
        return edible.getFoodValues(stack);
    }
```

src/main/java/net/minecraftforge/common/hunger/FoodValues.java
```diff
@@ -0,0 +1,165 @@
...
+    {
+        return get(stack, null);
+    }
+
+    /**
+     * Get player-specific food values.
+     *
+     * @return The food values, or null if none were found.
+     */
+    public static FoodValues get(@Nonnull ItemStack stack, Player player)
```
Shadows-of-Fire: Player arg should be marked `@Nullable`

src/main/java/net/minecraftforge/event/hunger/ExhaustionEvent.java
```diff
@@ -0,0 +1,303 @@
...
+        ATTACK_ENTITY,
+        DAMAGE_TAKEN,
+        EFFECT_HUNGER,
+        MOVEMENT_SWIM,
+        MOVEMENT_WALK_UNDERWATER,
+        MOVEMENT_WALK_ONWATER,
+        MOVEMENT_SPRINT,
+        MOVEMENT_CROUCH,
+        MOVEMENT_WALK_ONLAND
+    }
```
Shadows-of-Fire: These may want to use a string-backed type system instead of an enum, similar to `ToolAction`.  That would allow for modders to define additional types used in their systems.

src/test/java/net/minecraftforge/debug/hunger/ExhaustionEventTest.java
```diff
@@ -0,0 +1,97 @@
...
+ * - Maximum exhaustion level a player can have increased from 4.0 -> 30.0
+ * - Maximum exhaustion addition at one time increased from 40.0 -> 90.0
+ * - Non-sprinting jumps should apply a random amount of exhaustion each time
+ * - Sprinting jumps should apply 90 exhaustion each time
+ * - All other exhausting actions should be 1.5x their normal rate
+ * - Hunger loss should occur in peaceful difficulty
+ */
+@Mod("exhaustion_event_test")
+@Mod.EventBusSubscriber
+public class ExhaustionEventTest
```
Shadows-of-Fire, line 28: This test does work but I think it might be best if it also disabled automatic hunger regen in peaceful, since attempting to test if hunger goes down while it constantly regenerates is a bit weird.

Shadows-of-Fire (changes requested):
Looking very good so far, Just a couple things I think should be addressed.
All the test mods work as described.  

Finally, the PR needs to be rebased since it currently has merge conflicts.

src/main/java/net/minecraftforge/client/gui/ForgeIngameGui.java
Parker8283: I can buy that. I guess it comes down to if Forge wants to be opinionated on this. I could maybe see an issue where leaving it to the mod devs who change the max hunger could cause a potential incompatibility if multiple mods draw multiple lines of drumsticks in one instance. Not sure what the precedent is for that, so I'll leave that as an open question for someone else to answer.

patches/minecraft/net/minecraft/world/food/FoodData.java.patch
Parker8283: Moving this line into the if block would make the behavior the same as if the mod listened to `ExhaustionEvent.AllowExhaustion` and set that to `DENY`. The purpose of the `Exhausted` event is more to be a notification that max exhaustion was reached, while also allowing control over how hunger and saturation values are changed because of the exhaustion occurred. Because of this, unless there's some data `Exhausted` has that `AllowExhaustion` doesn't and should, I think it's better to leave it as is.

Parker8283:
Thanks for the review and feedback @Shadows-of-Fire! I have updated to account for most of your comments. There are two outstanding issues that I left responses to your comments on.

patches/minecraft/net/minecraft/world/food/FoodData.java.patch
Shadows-of-Fire: It is reasonable that cancelling exhaustion have the same impact as returning `DENY` from `AllowExhaustion`, as denying it being allowed, and cancelling it from happening, should technically be the same thing.

src/main/java/net/minecraftforge/client/gui/ForgeIngameGui.java
Shadows-of-Fire: Even if multiple mods were to be drawing additional drumstick lines, technically they would just be drawing them over eachother with the same data (thus not really conflicting beyond wasting time).  It would be simple to sort out at the modpack level with configuration files.

src/main/java/net/minecraftforge/common/hunger/FoodValues.java
```diff
@@ -0,0 +1,128 @@
...
+            return false;
+        if (getClass() != obj.getClass())
+            return false;
+        FoodValues other = (FoodValues) obj;
+        if (nutrition != other.nutrition)
+            return false;
+        if (Float.floatToIntBits(saturationModifier) != Float.floatToIntBits(other.saturationModifier))
+            return false;
+        return true;
+    }
```
Shadows-of-Fire: These are automatically implemented for record classes based on the fields.  Technically fine to keep, but unnecessary.

Parker8283: Huh...I was pretty sure I removed these...but I guess not! Thanks for double-checking.

patches/minecraft/net/minecraft/world/food/FoodData.java.patch
Parker8283: While that could be reasonable, I'm not sure how I feel about that from a design perspective. Having that basically yields us the below:
```java
if (eventA) {
  if (eventB) {
    doWork();
  }
}
```
To me when I see that, the first thing I think is that there's a chance that one of these events isn't needed, since firing two events one after another to basically ask the same question doesn't seem right. While I'm not against asking the question of whether or not we actually need both events (I technically didn't write this system so I'd have to do some research as to why both were created in the first place), I think if we do need both events, then not moving the line is probably the best course of action. I can't really think of a use case where a modder would say "I do want to have `AllowExhaustion` to go through, but I want to cancel any further action in `Exhausted`". I agree that is does look a little odd having one action from the event go through before checking if it was canceled, so maybe there's a better way to do what is attempting to be done here. I'm not sure. But unless I'm missing something, I don't think the answer is moving the line into the if-check.

If that is what really is wanted in order to be merged, I guess I don't mind doing it. Just wanted to throw in my perspective on this.

Shadows-of-Fire: See - Initially I was going to write that - why can't we just ditch `AllowExhaustion`
However, the case for `AllowExhaustion` is typically going to be that of a mod which wants to trigger exhaustion for some reason or another.  The use case of the other event would likely be a mod not interacting with `AllowExhaustion` at all, but wants to optionally do something when an exhaustion occurs.  
Semantically, it makes sense that cancelling `Exhausted` would cause the same impact as denying `AllowExhaustion`, but these are different time points, and would likely be different mods interacting in some fashion
Say mod A listens to `AllowExhaustion` and sets it to allow every 2 ticks while using some item, and mod B only listens to `Exhausted` to prevent exhaustion while under some potion effect.  You would expect that mod B only needs to cancel `Exhausted` to trigger no changes, but they would have to cancel it and also set getDeltaExhaustion to zero.

Parker8283: I kinda follow you for the most part, save for your example:
> mod B only listens to `Exhausted` to prevent exhaustion while under some potion effect

That's basically the exact contract of the `AllowExhaustion` event. Mod B wants to disallow "full exhaustion" processing to happen while under some potion effect, so they would listen to that and `DENY` it. They don't need to listen to `Exhausted` to prevent it, they need to listen to `AllowExhaustion` and then _not_ allow it. To me, that's just misusing the system as it currently exists.

I would agree that if mod B had to cancel and set `deltaExhaustion` to 0 as the only option to get that desired behavior, that would be bad/counter-intuitive. But for this specific example, they don't have to do that. The current contracts of the events is that `AllowExhaustion` is asking "should I go through the 'player is fully exhausted' processing or not?", whereas `Exhausted` is asking "hey, the player is fully exhausted, here's how much exhaustion is going to be removed (or added). How should we handle the hunger/saturation that should be removed (or added)?". The most correct option if a modder wanted to not do any exhaustion processing, which as your suggesting can be achieved by cancelling `Exhausted` and setting `deltaExhaustion` to 0, would actually be done by `DENY`ing `AllowExhaustion` and not even allowing `Exhausted` to be posted (saving a tiny bit of processing with it). The use case of someone not wanting to listen to `AllowExhaustion` but instead cancel `Exhausted` (and setting `deltaExhaustion` to 0, either explicitly or automatically as you are proposing) is the part that still isn't clicking with me.

Perhaps a different example will click better with me, because part of your reasoning I agree with, but I'm not sure I'm fully sold yet.

Shadows-of-Fire: I think that resonates the actual purpose correctly, so it will be fine as-is.

However, I think we should clamp this subtraction here so the value cannot go below zero.

Shadows-of-Fire:
Last batch of changes looks good to me.

marchermans:
I have read through this code a bit, and (as an author of a Entity mod with villager like entities) how does this behave for `LivingEntities`, I think it might be a worthwhile investment to see if this PR can be extended to reach such behavior.

Shadows-of-Fire:
I don't think any part of this system can be expanded to `LivingEntity`. Most of it is concerned with the use of `FoodData` or is a patch of a method with a `Player` param, so it cannot be generified.  `LivingEntity` does not have a `FoodData`.

For food interactions at the `LivingEntity` level, I believe #8477 handles that, but the scope of this PR is player-specific.

marchermans:
> I don't think any part of this system can be expanded to `LivingEntity`. Most of it is concerned with the use of `FoodData` or is a patch of a method with a `Player` param, so it cannot be generified. `LivingEntity` does not have a `FoodData`.
> 
> For food interactions at the `LivingEntity` level, I believe #8477 handles that, but the scope of this PR is player-specific.

Well you are adding a `Player` to food data, and then to those events. I am saying: Make that an `LivingEntity` so that mods which use this system for their villagers etc. can use this as well then.
That also accounts for of the events etc.

I am not asking you to implement it on living entity, I am asking for you to not make it impossible to use it for a `LivingEntity`.

Parker8283:
Hey Orion, that's a good point. I'm not against looking into that. There is one potential problem that I see currently that might hinder that however. Even though prior to this PR `FoodData` did not track which `Player` was using it, the vanilla `FoodData#tick` method _does_ take a `Player` as its parameter. So even if I change the `Player` field to a `LivingEntity` field and some mod attached a `FoodData` instance to a `LivingEntity`, they would still not be able to call `tick` properly. This would mean most events could not be fired without the modder redoing most of the work themselves. Or I guess the alternative would be for Forge to modify the method signature or some other workaround, but I'm not sure what the best way forward for that would be. If you've got ideas, let me know and I'll happily do them. And even with that, there might be some room for some events to relax the `Player` restriction. I can look at that as well.

Shadows-of-Fire:
If the calls within `tick` are not really player-specific, you can add a method `tick(LivingEntity)` and make `tick(Player)` just call that one.

Parker8283:
Oh yeah good point...well I can do that if need be. I'll have to go through everything and figure out which methods/events/etc can be abstracted out to be `LivingEntity` generic, if they can be. A lot of this code just uses/hardcodes the default player values on initial event firing since that's essentially what `FoodData` did, so I'm not sure what the best course of action is. In reality, `FoodData` appears to have been written to be very Player-centric overall (makes sense for vanilla obviously), so there may need to be a bit more work than just switching a field type to allow `FoodData` to be used on non-Players, like a way to set the default values for things like `maxExhaustionLevel`, `maxHunger`, etc. for a given entity upon construction or whatever. I'm starting to wonder if that starts to cause scope-creep on this PR and if that wouldn't be better suited for a subsequent PR, or if people don't care and such changes should just be rolled into this. For whatever it's worth, I'm willing to provide that subsequent PR if that's the path agreed upon. When I get time to go over this in full again, I can consider all that.

Shadows-of-Fire:
My opinion on it is that up-typing to `LivingEntity` could be done later by a different PR in 1.19 when breaking changes are permitted.

marchermans:
@Parker8283 Please hit me up on discord.
After talking internally we consider the idea of this PR valid, but the current implementation is not something I think that fits within the way forge tries to achieve the same kind of niche systems.

Parker8283:
For those not in the Discord, [Orion's requests and my response](https://i.imgur.com/E4Kut4H.png). TLDR is I will look into everything and how best to reconfigure this PR to match the design desires of the Forge team when I get the chance.

patches/minecraft/net/minecraft/world/food/FoodData.java.patch
```diff
@@ -0,0 +1,176 @@
+--- a/net/minecraft/world/food/FoodData.java
++++ b/net/minecraft/world/food/FoodData.java
+@@ -14,64 +_,95 @@
+    private float f_38698_;
+    private int f_38699_;
+    private int f_38700_ = 20;
++   // FORGE START
++   private Player player;
```
Shadows-of-Fire, line 8: expand to `LivingEntity`

patches/minecraft/net/minecraft/world/food/FoodData.java.patch
```diff
@@ -0,0 +1,176 @@
...
++      if (modifiedFoodValues != null) {
++         int prevFoodLevel = this.f_38696_;
++         float prevSaturationLevel = this.f_38697_;
++         this.m_38707_(modifiedFoodValues.nutrition(), modifiedFoodValues.saturationModifier());
++         net.minecraftforge.event.ForgeEventFactory.onFoodEaten(modifiedFoodValues, this.f_38696_ - prevFoodLevel, this.f_38697_ - prevSaturationLevel, p_38714_, this.player);
+       }
+ 
+    }
+ 
+    public void m_38710_(Player p_38711_) {
```
Shadows-of-Fire, line 44: add a new method `public void tick(LivingEntity entity) {` and redirect this one to that.

Proper patch syntax for that is 

```
public void m_38710_(Player p_38711_) {
   tick((LivingEntity) p_38711_);
}

public void tick(LivingEntity entity) {
   <patched method body of m_38710_>

File: patches/minecraft/net/minecraft/world/food/FoodData.java.patch [hunk 14]
```diff
@@ -0,0 +1,176 @@
...
+-         if (this.f_38699_ >= 80) {
+-            if (p_38711_.m_21223_() > 10.0F || difficulty == Difficulty.HARD || p_38711_.m_21223_() > 1.0F && difficulty == Difficulty.NORMAL) {
+-               p_38711_.m_6469_(DamageSource.f_19313_, 1.0F);
+-            }
+-
++      net.minecraftforge.eventbus.api.Event.Result allowSaturatedRegenResult = net.minecraftforge.event.ForgeEventFactory.fireAllowSaturatedRegenEvent(p_38711_);
++      boolean shouldDoSaturatedRegen = allowSaturatedRegenResult == net.minecraftforge.eventbus.api.Event.Result.ALLOW || (allowSaturatedRegenResult == net.minecraftforge.eventbus.api.Event.Result.DEFAULT && flag && this.f_38697_ > 0.0F && p_38711_.m_36325_() && this.f_38696_ >= 20);
++      net.minecraftforge.eventbus.api.Event.Result allowRegenResult = shouldDoSaturatedRegen ? net.minecraftforge.eventbus.api.Event.Result.DENY : net.minecraftforge.event.ForgeEventFactory.fireAllowRegenEvent(p_38711_);
++      boolean shouldDoRegen = allowRegenResult == net.minecraftforge.eventbus.api.Event.Result.ALLOW || (allowRegenResult == net.minecraftforge.eventbus.api.Event.Result.DEFAULT && flag && this.f_38696_ >= 18 && p_38711_.m_36325_());
```
Shadows-of-Fire, line 91: The method at the end here `m_36325_` -> `Player#isHurt` is player-specific, but it calls LivingEntity-level methods.  Could probably be substituted with a util method on FoodData `boolean isHurt(LivingEntity entity)` which does the same as the player method.

Shadows-of-Fire (changes requested):

Shadows-of-Fire:
The following event hook calls should be converted to attributes on `LivingEntity` 
* getMaxHunger
* getExhaustionCap
  
  
The following tick-based event hook calls *could* be converted to attributes on `LivingEntity` 
* getSaturatedRegenTickPeriod
* getHealthRegenTickPeriod
* getStarveTickPeriod

These events are a bit weirder, however.  They may not even be necessary unless someone expresses a need for changing a tick period that cannot be accomplished by simply changing the values on the actual events.  My suggestion is to scrap them, because one could always change the values to match the "standard" tick period.

Shadows-of-Fire:
Then finally, all events except `PeacefulRegen` (both types) can be expanded to be `LivingEntity`

marchermans:
> Then finally, all events except `PeacefulRegen` (both types) can be expanded to be `LivingEntity`

PeacefulRegen should also be expanded. Why can it not be expanded?

mont-40:
@Parker8283 @Shadows-of-Fire @Minecraftschurli Is this PR dead? Would be great to see this added as AppleCore was discontinued.

Parker8283:
Not dead, I just have a day job :)

James103:
@Parker8283 Do you have any plans to port this forward to 1.19.2? 1.19 has been released since this PR was made, making it the new full-support version and relegating 1.18 to LTS.

Furthermore, how would this interact with https://github.com/MinecraftForge/MinecraftForge/pull/9134?

Parker8283:
Yes I plan to get this up to 1.19 once I have the time, and then this will be the LTS version after that. As for that PR, there is likely some overlap between the two PRs on first glance, but I haven't looked super in-depth on it.

LexManos:
At the end of the day I think something like this is vastly too complicated and prone to breaking during updates for it to be in Forge itself. It has existed as external api for quite a while, and with the plans I have for the future building systems like this that are loaded alongside forge should become simpler.

PR #8406, opened 2022-01-19
[1.18.x] [event] add EnderManAngerEvent

Minecraftschurli:
Adds a new event that is fired when an enderman checks if it should get angry at a player to make it possible to prevent endermen getting angry at that player by cancelling the event.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/monster/EnderMan.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/monster/EnderMan.java.patch
@@ -21,7 +21,7 @@
     boolean m_32534_(Player p_32535_) {
        ItemStack itemstack = p_32535_.m_150109_().f_35975_.get(3);
 -      if (itemstack.m_150930_(Blocks.f_50143_.m_5456_())) {
-+      if (itemstack.isEnderMask(p_32535_, this)) {
++      if (net.minecraftforge.common.ForgeHooks.shouldSuppressEnderManAnger(this, p_32535_, itemstack)) {
           return false;
        } else {
           Vec3 vec3 = p_32535_.m_20252_(1.0F).m_82541_();
--- a/patches/minecraft/net/minecraft/world/entity/monster/EnderMan.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/monster/EnderMan.java.patch
@@ -16,15 +16,12 @@
     }
  
     protected void m_8097_() {
-@@ -194,7 +_,10 @@
+@@ -194,7 +_,7 @@
  
     boolean m_32534_(Player p_32535_) {
        ItemStack itemstack = p_32535_.m_150109_().f_35975_.get(3);
 -      if (itemstack.m_150930_(Blocks.f_50143_.m_5456_())) {
-+      net.minecraftforge.eventbus.api.Event.Result result = net.minecraftforge.common.ForgeHooks.onEnderManAnger(this, p_32535_);
-+      if (result != net.minecraftforge.eventbus.api.Event.Result.DEFAULT) {
-+         return result == net.minecraftforge.eventbus.api.Event.Result.ALLOW;
-+      } else if (itemstack.isEnderMask(p_32535_, this)) {
++      if (net.minecraftforge.common.ForgeHooks.shouldSuppressEnderManAnger(this, p_32535_, itemstack)) {
           return false;
        } else {
           Vec3 vec3 = p_32535_.m_20252_(1.0F).m_82541_();
--- a/patches/minecraft/net/minecraft/world/entity/monster/EnderMan.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/monster/EnderMan.java.patch
@@ -16,12 +16,15 @@
     }
  
     protected void m_8097_() {
-@@ -194,7 +_,7 @@
+@@ -194,7 +_,10 @@
  
     boolean m_32534_(Player p_32535_) {
        ItemStack itemstack = p_32535_.m_150109_().f_35975_.get(3);
 -      if (itemstack.m_150930_(Blocks.f_50143_.m_5456_())) {
-+      if (net.minecraftforge.common.ForgeHooks.shouldSuppressEnderManAnger(this, p_32535_, itemstack)) {
++      net.minecraftforge.eventbus.api.Event.Result result = net.minecraftforge.common.ForgeHooks.onEnderManAnger(this, p_32535_);
++      if (result != net.minecraftforge.eventbus.api.Event.Result.DEFAULT) {
++         return result == net.minecraftforge.eventbus.api.Event.Result.ALLOW;
++      } else if (itemstack.isEnderMask(p_32535_, this)) {
           return false;
        } else {
           Vec3 vec3 = p_32535_.m_20252_(1.0F).m_82541_();
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -1457,7 +1457,7 @@ public class ForgeHooks
 
     public static boolean shouldSuppressEnderManAnger(EnderMan enderMan, Player player, ItemStack mask)
     {
-        return MinecraftForge.EVENT_BUS.post(new EnderManAngerEvent(enderMan, player)) || mask.isEnderMask(player, enderMan);
+        return mask.isEnderMask(player, enderMan) || MinecraftForge.EVENT_BUS.post(new EnderManAngerEvent(enderMan, player));
     }
 
 }
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -1294,7 +1294,7 @@ public class ForgeHooks
         boolean generateBonusChest = wgs.generateBonusChest();
         Registry<LevelStem> originalRegistry = wgs.dimensions();
         Optional<String> legacyCustomOptions = wgs.legacyCustomOptions;
-
+        
         // make a copy of the dimension registry; for dimensions that specify that they should use the server seed instead
         // of the hardcoded json seed, recreate them with the correct seed
         MappedRegistry<LevelStem> seededRegistry = new MappedRegistry<>(Registry.LEVEL_STEM_REGISTRY, Lifecycle.experimental(), (Function<LevelStem, Holder.Reference<LevelStem>>)null);
@@ -1311,10 +1311,10 @@ public class ForgeHooks
                 seededRegistry.register(key, dimension, originalRegistry.lifecycle(dimension));
             }
         }
-
+        
         return new WorldGenSettings(seed, generateFeatures, generateBonusChest, seededRegistry, legacyCustomOptions);
     }
-
+    
     /** Called in the LevelStem codec builder to add extra fields to dimension jsons **/
     public static App<Mu<LevelStem>, LevelStem> expandLevelStemCodec(RecordCodecBuilder.Instance<LevelStem> builder, Supplier<App<Mu<LevelStem>, LevelStem>> vanillaFieldsSupplier)
     {
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -155,7 +155,6 @@ import net.minecraftforge.event.entity.player.PlayerInteractEvent;
 import net.minecraftforge.event.world.BiomeLoadingEvent;
 import net.minecraftforge.event.world.BlockEvent;
 import net.minecraftforge.event.world.NoteBlockEvent;
-import net.minecraftforge.eventbus.api.Event;
 import net.minecraftforge.eventbus.api.Event.Result;
 import net.minecraftforge.fluids.FluidAttributes;
 import net.minecraftforge.fml.ModContainer;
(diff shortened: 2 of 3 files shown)


src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1332,4 +1333,9 @@ public static void readAdditionalLevelSaveData(CompoundTag rootTag)
         }
     }
 
+    public static boolean shouldSuppressEnderManAnger(EnderMan enderMan, Player player, ItemStack mask)
+    {
+        return MinecraftForge.EVENT_BUS.post(new EnderManAngerEvent(enderMan, player)) || mask.isEnderMask(player, enderMan);
```
marchermans: Flip this the other way around.
It is more performant that way.

marchermans (changes requested):

src/main/java/net/minecraftforge/event/entity/living/EnderManAngerEvent.java
```diff
@@ -0,0 +1,26 @@
...
+
+import net.minecraft.world.entity.monster.EnderMan;
+import net.minecraft.world.entity.player.Player;
+import net.minecraftforge.eventbus.api.Cancelable;
+
+/**
+ * Event that is fired when an enderman checks if the player is looking at him.
+ * If cancelled the enderman does not get angry at the player.
+ */
+@Cancelable
```
diesieben07, line 21: This could benefit from `@HasResult`:
- `DENY`: do not let the enderman get angry (regardless of what `isEnderMask` says)
- `ALLOW`: allow the enderman to get angry (regardless of what `isEnderMask` says)
- `DEFAULT`: refer to vanilla behavior (i.e. call `ItemStack#isEnderMask`)

src/main/java/net/minecraftforge/event/entity/living/EnderManAngerEvent.java
```diff
@@ -0,0 +1,26 @@
+package net.minecraftforge.event.entity.living;
```
diesieben07, line 6: Missing license header.
Please also check the formatting in this file (braces on new lines)

src/main/java/net/minecraftforge/event/entity/living/EnderManAngerEvent.java
```diff
@@ -0,0 +1,26 @@
...
+public class EnderManAngerEvent extends LivingEvent {
+    private final Player player;
+
+    public EnderManAngerEvent(EnderMan enderman, Player player) {
+        super(enderman);
+        this.player = player;
+    }
+
+    /**
+     * The player that is being checked.
```
diesieben07, line 33: Should use `@return` here

diesieben07 (changes requested):
Why does this need to be an event? What use case is there that is not covered by overriding `isEnderMask`?

Minecraftschurli:
> Why does this need to be an event? What use case is there that is not covered by overriding isEnderMask?

isEnderMask only applies to items worn on the head. An event allows things like potion effects or other things to affect it too.

src/main/java/net/minecraftforge/event/entity/living/EnderManAngerEvent.java
```diff
@@ -0,0 +1,48 @@
...
+
+package net.minecraftforge.event.entity.living;
+
+import net.minecraft.world.entity.monster.EnderMan;
+import net.minecraft.world.entity.player.Player;
+import net.minecraftforge.eventbus.api.Event;
+
+/**
+ * Event that is fired when an enderman checks if the player is looking at him.
+ * If cancelled the enderman does not get angry at the player.
```
sciwhiz12: Please update the javadoc to follow and explain the new behavior of the event.

sciwhiz12 (changes requested):

patches/minecraft/net/minecraft/world/entity/monster/EnderMan.java.patch
```diff
@@ -16,12 +16,15 @@
...
+@@ -194,7 +_,10 @@
  
     boolean m_32534_(Player p_32535_) {
        ItemStack itemstack = p_32535_.m_150109_().f_35975_.get(3);
 -      if (itemstack.m_150930_(Blocks.f_50143_.m_5456_())) {
-+      if (net.minecraftforge.common.ForgeHooks.shouldSuppressEnderManAnger(this, p_32535_, itemstack)) {
++      net.minecraftforge.eventbus.api.Event.Result result = net.minecraftforge.common.ForgeHooks.onEnderManAnger(this, p_32535_);
++      if (result != net.minecraftforge.eventbus.api.Event.Result.DEFAULT) {
++         return result == net.minecraftforge.eventbus.api.Event.Result.ALLOW;
++      } else if (itemstack.isEnderMask(p_32535_, this)) {
```
diesieben07: Why did you pull out this logic into the patch instead of just changing `ForgeHooks.shouldSuppressEnderManAnger`?

diesieben07 (changes requested):

patches/minecraft/net/minecraft/world/entity/monster/EnderMan.java.patch
Minecraftschurli: because just a boolean can not be used to determine what to do because ALLOW/DENY means it skips the vanilla distance and line of sight checks and forges isEnderMask check

diesieben07: `EnderMan#isLookingAtMe` returns a boolean though:
```
if (ForgeHooks.shouldCancelEndermanAnger(this, stack, player)) {
    return false;
} else { 
 ....
 }
```

Minecraftschurli: your code would make ALLOW and DEFAULT do the same since ALLOW also skips the other vanilla calculations

Shadows-of-Fire:
Just looking at this I think using `Result` is overkill, do you have a use case where you would ever want to override the behavior of the head item?  This event could just be a simple `@Cancellable` event otherwise, only run if the vanilla check would return true.

Minecraftschurli:
It wasn't my idea to use result. I used Cancellable but it was requested I change it to a result.

Shadows-of-Fire:
I am willing to move forward with the current iteration but I believe the original iteration (cancellable) was more realistic as well as more performant (since the event will be fired less).  I also believe the first iteration is more likely to be accepted due to performance concerns.

Unless you have a compelling reason for the use of `Result` (beyond the suggestion that was made), please revert this back to the original iteration and notify me of that change. If you think this iteration is better than the former, and the utilization of `Result` is significant, please provide said reason.

Once one of those two things happens, this PR can move forward.

Minecraftschurli:
@Shadows-of-Fire I reverted it to `@Cancellable` since the `Result` wasn't even my idea and it made the PR unnecessarily bulky

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1432,4 +1433,9 @@ public static MobEffect loadMobEffect(CompoundTag nbt, String key, @Nullable Mob
...
```
Shadows-of-Fire: Echoing Orion's earlier comment - flip the order here, because if `isEnderMask` returns true, the event does not need to be fired.
`return mask.isEnderMask(player, enderMan) || MinecraftForge.EVENT_BUS.post(new EnderManAngerEvent(enderMan, player));`

src/main/java/net/minecraftforge/event/entity/living/EnderManAngerEvent.java
```diff
@@ -0,0 +1,33 @@
...
+
+package net.minecraftforge.event.entity.living;
+
+import net.minecraft.world.entity.monster.EnderMan;
+import net.minecraft.world.entity.player.Player;
+import net.minecraftforge.eventbus.api.Cancelable;
+
+/**
+ * Event that is fired when an enderman checks if the player is looking at him.
+ * If cancelled the enderman does not get angry at the player.
```
Shadows-of-Fire: Proposing a more formal javadoc
```
/**
 * This event is fired when an Enderman detects that a player is looking at them.
 * It will not be fired if the detection is prevented by {@link IForgeItem#isEnderMask}
 * 
 * This event is {@link Cancelable}.
 * If this event is canceled, the Enderman will not target the player.
 * 
 * This event does not have a {@link Result}.
 */

**Review by Shadows-of-Fire — CHANGES_REQUESTED** [review]

File: src/main/java/net/minecraftforge/common/ForgeHooks.java [hunk 9]
```diff
@@ -1404,7 +1405,7 @@ public static void readAdditionalLevelSaveData(CompoundTag rootTag)
             LOGGER.error(WORLDPERSISTENCE, buf.toString());
         }
     }
-
+
```
Shadows-of-Fire: I'm on the fence about the other whitespace changes but this one is against style guides, and needs to be reverted.

Minecraftschurli:
@Shadows-of-Fire Done

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1293,7 +1294,7 @@ public static WorldGenSettings loadDimensionsWithServerSeed(WorldGenSettings wgs
         boolean generateBonusChest = wgs.generateBonusChest();
         Registry<LevelStem> originalRegistry = wgs.dimensions();
         Optional<String> legacyCustomOptions = wgs.legacyCustomOptions;
-        
+
```
TheCurle: What's with the whitespace?

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1310,10 +1311,10 @@ public static WorldGenSettings loadDimensionsWithServerSeed(WorldGenSettings wgs
                 seededRegistry.register(key, dimension, originalRegistry.lifecycle(dimension));
             }
         }
-        
+
```
TheCurle: Whitespace.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1310,10 +1311,10 @@ public static WorldGenSettings loadDimensionsWithServerSeed(WorldGenSettings wgs
...
         return new WorldGenSettings(seed, generateFeatures, generateBonusChest, seededRegistry, legacyCustomOptions);
     }
-    
+
```
TheCurle: I don't understand why you're changing so much whitespace.

src/main/java/net/minecraftforge/event/entity/living/EnderManAngerEvent.java
```diff
@@ -0,0 +1,38 @@
...
+import net.minecraftforge.eventbus.api.Cancelable;
+
+/**
+ * This event is fired when an Enderman detects that a player is looking at them.
+ * It will not be fired if the detection is prevented by {@link IForgeItem#isEnderMask}
+ * <p>
+ * This event is {@link Cancelable}.
+ * If this event is canceled, the Enderman will not target the player.
+ * <p>
+ * This event does not have a {@link Result}.
```
TheCurle, line 19: Probably worth mentioning which bus this fires on (as a quick and easy reference so that one doesn't have to follow through the event hierarchy looking for IModBusEvent) as well as which side(s) it fires on.

TheCurle (changes requested):

src/main/java/net/minecraftforge/common/ForgeHooks.java
Minecraftschurli: I have no idea why that keeps happening

src/main/java/net/minecraftforge/common/ForgeHooks.java
Minecraftschurli: I don't know either

Minecraftschurli:
I can't fix the whitespace since the problem reappears when I try to commit (even from the cli)

(edit) fixed through github web interface

Minecraftschurli:
@TheCurle I fixed the whitespace

TheCurle (approved):

src/main/java/net/minecraftforge/event/entity/living/EnderManAngerEvent.java
```diff
@@ -0,0 +1,38 @@
...
+ * This event is fired on the forge bus before an Enderman detects that a player is looking at them.
+ * It will not be fired if the detection is already prevented by {@link IForgeItem#isEnderMask}
+ * <p>
+ * This event is {@link Cancelable}.
+ * If this event is canceled, the Enderman will not target the player.
+ * <p>
+ * This event does not have a {@link Result}.
+ */
+@Cancelable
+public class EnderManAngerEvent extends LivingEvent {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
public class EnderManAngerEvent extends LivingEvent
{
```

src/main/java/net/minecraftforge/event/entity/living/EnderManAngerEvent.java
```diff
@@ -0,0 +1,38 @@
...
+    private final Player player;
+
+    public EnderManAngerEvent(EnderMan enderman, Player player)
+    {
+        super(enderman);
```
sciwhiz12, line 28: As the event's entity is an `EnderMan`, it will be helpful if this event overrides `getEntity()` and changes the return type to `EnderMan`[^1] (returning the result from the super method using a cast). 

[^1]: This is possible because [return types are covariant](https://www.baeldung.com/java-covariant-return-type) in Java -- an override of a method may have a return type which is a subclass of the overriden method's return type.

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/event/entity/living/EnderManAngerEvent.java
Minecraftschurli: `getEntity` or `getEntityLiving` or both?

Shadows-of-Fire: I don't think we have a standard, as far as existing source goes it looks like such overrides are not common at all - `AnimalTameEvent`, and `PlayerEvent`, which are both subclasses of `LivingEvent` provide a new method to access their strict types, instead of overriding getEntityLiving.

Arguably `getEntityLiving` should instead be an override on `getEntity` to return a `LivingEntity`
Thus, I would say you should override `getEntity`, which supports potential phasing-out of `getEntityLiving` in the future.

Minecraftschurli: Done

sciwhiz12 (approved):

PR #8410, opened 2022-01-20
Provide NPE protection against out of order init of the TYPES and REF

marchermans:
This provides a hotfix against the crash that occurs because of #8242 somehow.
It just checks against a null field when it is called from the super and sets the map early.

This just works now, no clue why it worked before....


MrGullible:
New error running 39.0.54: https://www.toptal.com/developers/hastebin/ovuxuniqux.yaml @OrionDevelopment

PR #8412, opened 2022-01-21
DFU - The second.

marchermans:
### Second version of: #8242 and #8410 
This version of the PR finally fixes the NPEs that where seen yesterday during the initial attempt at this.
I finally tracked them down to the concurrent calculation of the rules while forge rebuilds the types in the schema. So @cpw was right.....again.
This version now also includes a test mod 👍, not sure why I did not include it in the first one.

### Modifiable DFU

#### Goals

#### Implementation

#### Future:

#### Usage:

##### Example:
```java
{

    @SubscribeEvent
    {
            );
        }
    }
}
```

#### Todo:
- [X] Better testing
- [X] Formatting


Raycoms:
Any news on that?

Raycoms:
Tested it locally with our Schematic World with a lot of mod specific tile entities. Worked great.
Test-code: https://github.com/ldtteam/minecolonies/blob/forge-dfu/src/main/java/com/minecolonies/coremod/MineColonies.java#L120

Raycoms:
A small question on this, why is this not the default (By default all mod tileentities recover correctly)

marchermans:
So the reason why that does not work is related to the usage reason for DFU.
Assume you have two versions A and B.

In A your BlockEntity is called `mod:some_tile_entity` and in B `mod:some_moved_tile_entity` (note this is basically what has happened between 1.17 and 1.18 (The key path from the root of the NBT to list of TileEntities in the chunk/world has changed). If you are in 1.18 how would you know form just the registry names what the value was in 1.17?

This example can even be expanded upon: Howabout you change the layout of how you store items in your NBT?

Yes it would be easy to say: This is stupid busy work and should not be a problem for modders. But it is, Mojang puts great care into making saves compatible between versions, and I personally think modders should to.

PR #8419, opened 2022-01-23
[1.18.x] Custom usage animations for items

rehwinkel:
This PR adds the ability to add custom usage animations to items. By default, `Item#getUseAnimation` returns a hardcoded Enum. Instead, I've made the enum extensible. Subsequently registering a custom `IForgeArmPose` for a custom `UseAnim` in the `ClientRegistry` allows you to define what happens when this UseAnim is active. I decided against using the EventBus for this because animation updates happen very frequently and because there will (almost) always be a 1-to-1 relation of UseAnims and subscribers for that UseAnim.


patches/minecraft/net/minecraft/client/model/HumanoidModel.java.patch
```diff
@@ -0,0 +1,195 @@
+--- a/net/minecraft/client/model/HumanoidModel.java
++++ b/net/minecraft/client/model/HumanoidModel.java
+@@ -32,8 +_,8 @@
+    public final ModelPart f_102812_;
+    public final ModelPart f_102813_;
+    public final ModelPart f_102814_;
+-   public HumanoidModel.ArmPose f_102815_ = HumanoidModel.ArmPose.EMPTY;
+-   public HumanoidModel.ArmPose f_102816_ = HumanoidModel.ArmPose.EMPTY;
++   public net.minecraftforge.common.extensions.IForgeArmPose f_102815_ = HumanoidModel.ArmPose.EMPTY;
++   public net.minecraftforge.common.extensions.IForgeArmPose f_102816_ = HumanoidModel.ArmPose.EMPTY;
```
diesieben07: This is a breaking change. Why not make `ArmPose` an extensible enum, too?

patches/minecraft/net/minecraft/client/model/HumanoidModel.java.patch
```diff
@@ -0,0 +1,195 @@
...
+-         boolean flag4 = flag2 ? this.f_102815_.m_102897_() : this.f_102816_.m_102897_();
++         boolean flag4 = flag2 ? this.f_102815_.isPoseTwoHanded() : this.f_102816_.isPoseTwoHanded();
+          if (flag2 != flag4) {
+             this.m_102878_(p_102866_);
+             this.m_102875_(p_102866_);
+@@ -227,73 +_,85 @@
+    }
+ 
+    private void m_102875_(T p_102876_) {
+-      switch(this.f_102816_) {
```
diesieben07: This patch could be a lot smaller.
If `ArmPos` is extensible like suggested above, just add a `default` block that calls `applyTransform`.

src/main/java/net/minecraftforge/client/ClientRegistry.java
```diff
@@ -19,19 +19,21 @@
...
+import net.minecraft.client.Minecraft;
 import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.item.UseAnim;
+import net.minecraftforge.common.extensions.IForgeArmPose;
 import org.apache.commons.lang3.ArrayUtils;
 
-import net.minecraft.client.Minecraft;
-import net.minecraft.client.KeyMapping;
-
```
diesieben07: Don't reorder the imports

src/main/java/net/minecraftforge/client/ClientRegistry.java
```diff
@@ -58,4 +60,15 @@ public static ResourceLocation getEntityShader(Class<? extends Entity> entityCla
...
     }
+
+    public static void registerUseAnimation(UseAnim animation, IForgeArmPose pose)
+    {
+        useAnimationMap.put(animation, pose);
+    }
+
+    public static IForgeArmPose getUseAnimation(UseAnim animation)
+    {
+        return useAnimationMap.get(animation);
```
diesieben07: This should call a method on the enum instead of just a plain map lookup.

diesieben07 (changes requested):

src/main/java/net/minecraftforge/client/ClientRegistry.java
rehwinkel: IntelliJ loves to mangle imports :/

rehwinkel:
@diesieben07 I kept the HashMap lookup because putting any code in the UseAnim enum that returns an object that only exists on the client is a bad idea.

XFactHD:
The `create(String name, boolean twoHanded)` method (https://github.com/MinecraftForge/MinecraftForge/pull/8419/files#diff-749c2dfa55244df387a9ad0634c39aa2c8b29d4e6c1c3a7827e29bbf592d10f8R42-R44) is unnecessary IMO because modders should never register a custom `ArmPose` without an accompanying `IForgeArmPose`.
The `forgeArmPose` parameter of the `create(String name, boolean twoHanded, IForgeArmPose forgeArmPose)` method (https://github.com/MinecraftForge/MinecraftForge/pull/8419/files#diff-749c2dfa55244df387a9ad0634c39aa2c8b29d4e6c1c3a7827e29bbf592d10f8R46-R48) should also be explicitly annotated with `@NotNull`.

I would also argue that `applyTransform(HumanoidModel<T> model, T entity)` should throw an error if it is called on a vanilla `ArmPose` by checking the `forgeArmPose` with `Preconditions.checkNotNull(forgeArmPose, "applyTransform() must not be called on a vanilla ArmPose!")`.

rehwinkel:
@XFactHD given this information from the IExtensibleEnum doc:

>  * To be implemented on vanilla enums that should be enhanced with ASM to be
>  * extensible. If this is implemented on a class, the class must define a static
>  * method called "create" which takes a String (enum name), and the rest of the
>  * parameters matching a constructor.

I assumed I *must* implement a `create` method per constructor, as stated in the text. Making the IForgeArmPose nonnull is a good call though!

XFactHD:
I see can how this may be confusing, but to clear it up: you only need to match *a* constructor, matching all of them is not necessary.

src/main/java/net/minecraftforge/client/ClientRegistry.java
```diff
@@ -58,4 +61,14 @@ public static ResourceLocation getEntityShader(Class<? extends Entity> entityCla
     {
         return entityShaderMap.get(entityClass);
     }
+
+    public static void registerArmPose(UseAnim animation, HumanoidModel.ArmPose pose)
```
diesieben07: There should be documentation for this method.

src/main/java/net/minecraftforge/client/ClientRegistry.java
```diff
@@ -58,4 +61,14 @@ public static ResourceLocation getEntityShader(Class<? extends Entity> entityCla
...
+    {
+        useAnimationMap.put(animation, pose);
+    }
+
+    public static HumanoidModel.ArmPose getArmPose(UseAnim animation)
```
diesieben07: Should have a `@Nullable` annotation.

src/main/java/net/minecraftforge/client/IForgeArmPose.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * An ArmPose that can be defined by the user.
+ * Register one by creating a custom {@link net.minecraft.client.model.HumanoidModel.ArmPose}
+ * and use {@link net.minecraftforge.client.ClientRegistry#registerArmPose(UseAnim, HumanoidModel.ArmPose)} to bind it to a UseAnim.
+ */
+public interface IForgeArmPose
+{
+
+    /**
+     * This method should be used to apply all wanted transformations to the player if when the UseAnim is active.
```
diesieben07: typo ("if when")

src/main/java/net/minecraftforge/client/ClientRegistry.java
```diff
@@ -32,6 +34,7 @@
 public class ClientRegistry
 {
     private static Map<Class<? extends Entity>, ResourceLocation> entityShaderMap = new ConcurrentHashMap<>();
+    private static Map<UseAnim, HumanoidModel.ArmPose> useAnimationMap = new ConcurrentHashMap<>();
```
diesieben07: Should be final

patches/minecraft/net/minecraft/client/renderer/entity/player/PlayerRenderer.java.patch
```diff
@@ -10,6 +10,19 @@
...
+@@ -123,6 +_,12 @@
+             if (useanim == UseAnim.SPYGLASS) {
+                return HumanoidModel.ArmPose.SPYGLASS;
+             }
++
++            HumanoidModel.ArmPose forgeArmPose = net.minecraftforge.client.ClientRegistry.getArmPose(useanim);
++            if (forgeArmPose == null) {
++               throw new RuntimeException("No ArmPose registered for UseAnim '" + useanim + "'");
++            }
++            return forgeArmPose;
```
diesieben07: This should just be a generic hook that allows setting the arm pose. I.e. call `IItemRenderProperties#getArmPose` here regardless of whether the item is in use or not. Then you do not need the global map from UseAnim to ArmPos and it also allows greater flexibility.

diesieben07 (changes requested):
You also have an empty patch file commited (MouseHandler)

src/main/java/net/minecraftforge/client/ClientRegistry.java
rehwinkel: Should I make the entityShaderMap final aswell?

diesieben07: No, but this comment is moot anyways if you apply the changes in my last comment

patches/minecraft/net/minecraft/client/renderer/entity/player/PlayerRenderer.java.patch
```diff
@@ -10,6 +10,18 @@
     }
  
     public Vec3 m_7860_(AbstractClientPlayer p_117785_, float p_117786_) {
+@@ -123,6 +_,11 @@
+             if (useanim == UseAnim.SPYGLASS) {
+                return HumanoidModel.ArmPose.SPYGLASS;
+             }
++
++            HumanoidModel.ArmPose forgeArmPose = net.minecraftforge.client.RenderProperties.get(itemstack).getArmPose(p_117795_, p_117796_, itemstack);
```
diesieben07: This patch needs to be _outside_ the if statement, so that it is not limited to items in use.

src/test/java/net/minecraftforge/debug/entity/player/ItemUseAnimationTest.java
```diff
@@ -0,0 +1,93 @@
...
+            return SWING;
+        }
+
+        @Override
+        public void initializeClient(Consumer<IItemRenderProperties> consumer)
+        {
+            consumer.accept(new IItemRenderProperties()
+            {
+
+                private final HumanoidModel.ArmPose SWING_POSE = HumanoidModel.ArmPose.create("SWING", false, this::applyTransform);
```
diesieben07: This is in the wrong place now, it should be a top level static final, like `SWING` is.

patches/minecraft/net/minecraft/client/renderer/entity/player/PlayerRenderer.java.patch
```diff
@@ -10,6 +10,18 @@
...
++            if (forgeArmPose != null) {
++               return forgeArmPose;
++            }
```
diesieben07: this could be on one line.

diesieben07 (changes requested):
just very few small nitpicks then I am happy :)

src/test/java/net/minecraftforge/debug/entity/player/ItemUseAnimationTest.java
rehwinkel: I didn't wanna put it in the Item class cause ArmPose is client-only.

rehwinkel: Yup, as I had anticipated I get this error when putting it a higher level up:

`Attempted to load class net/minecraft/client/model/HumanoidModel for invalid dist DEDICATED_SERVER`

src/main/java/net/minecraftforge/client/IForgeArmPose.java
```diff
@@ -35,9 +35,10 @@
      * This method should be used to apply all wanted transformations to the player when the ArmPose is active.
      * You can use {@link LivingEntity#getTicksUsingItem()} and {@link LivingEntity#getUseItemRemainingTicks()} for moving animations.
      *
-     * @param model  The humanoid model
-     * @param entity The humanoid entity
+     * @param model   The humanoid model
+     * @param entity  The humanoid entity
+     * @param isRight True if the animation is applied to the right arm, false if to the left arm.
```
TeamSpen210, line 42: Instead of a boolean, it might be better to use `net.minecraft.world.entity.HumanoidArm.LEFT` / `RIGHT`. That is meaningful by itself, without needing to check the parameter name / docs.

src/main/java/net/minecraftforge/client/IForgeArmPose.java
```diff
@@ -0,0 +1,44 @@
...
+package net.minecraftforge.client;
+
+import net.minecraft.client.model.HumanoidModel;
+import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.item.UseAnim;
+
+/**
+ * An ArmPose that can be defined by the user.
+ * Register one by creating a custom {@link net.minecraft.client.model.HumanoidModel.ArmPose}
+ * and use {@link net.minecraftforge.client.ClientRegistry#registerArmPose(UseAnim, HumanoidModel.ArmPose)} to bind it to a UseAnim.
```
diesieben07: This doc is now outdated.

diesieben07 (commented):
A minor JavaDoc change you missed, other than that this looks solid now :+1:

diesieben07 (approved):

patches/minecraft/net/minecraft/world/item/UseAnim.java.patch
```diff
@@ -0,0 +1,20 @@
...
+    DRINK,
+@@ -9,4 +_,9 @@
+    SPEAR,
+    CROSSBOW,
+    SPYGLASS;
++
++   public static UseAnim create(String name)
++   {
++       throw new IllegalStateException("Enum not extended");
++   }
```
sciwhiz12: In Minecraft classes (patches), code should follow the formatting as defined in the rest of the class (by Fernflower): braces on newlines, and three spaces indentation.
```suggestion
+   public static UseAnim create(String name) {
+      throw new IllegalStateException("Enum not extended");
+   }
```

sciwhiz12 (changes requested):
Please update this PR to the latest commit of the `1.18.x` branch and update the license headers in the following files:
- `src/main/java/net/minecraftforge/client/IForgeArmPose.java`
- `src/main/java/net/minecraftforge/client/IItemRenderProperties.java`
- `src/test/java/net/minecraftforge/debug/entity/player/ItemUseAnimationTest.java`

tmvkrpxl0:
I have updated this PR to latest 1.18 commit in my branch here:
https://github.com/tmvkrpxl0/MinecraftForge/tree/updated-pr

sciwhiz12:
@tmvkrpxl0, you'll either have to coordinate with @rehwinkel to update this PR's branch (via force-pushing their branch to match yours), or you can open a new PR using your branch.

I will note that because 1.18.x is now LTS, we'll also need justification on why the feature of this PR is not applicable to 1.19.x, or else the PR needs to be ported and merged to 1.19.x.

tmvkrpxl0:
It should also be targeted for 1.19.x, I just wanted to do it on 1.18.x first.
I'll port this to 1.19.x soon

tmvkrpxl0:
I have ported this PR to 1.19.x in #8932

rehwinkel:
@tmvkrpxl0 i merged your changes finally

sciwhiz12 (changes requested):
Please update this PR to match the changes from #8932, or provide explanation on why these two PRs should differ in implementation.

tmvkrpxl0:
Note that #8932 had a critical bug and changes HumanoidModel.java.patch from #8987 should be there too

PaintNinja:
Closing as stale. If you'd still like this, please make a new PR

PR #8420, opened 2022-01-23
[1.19.x] Add support for items to override NBT enchantments

KnightMiner:
Many enchantments in vanilla are hardcoded to checks using `EnchantmentHelper::getItemEnchantmentLevel`, for instance the curse of binding or soul speed. Some mods add enchantment alternatives, such as Tinkers' Construct's modifiers, augments on various handheld tools in tech mods, or tools that start with a specific enchantment. Adding enchantments directly to the tool can easily lead to an XP dupe using the grindstone, prevents re-enchanting the tool, and in general conflicts with map-maker tools. It is sometimes possible to replicate the effect of enchantments, though it often partially successful or very likely to conflict with other mods (such as the global loot modifier example mod reimplementing silk touch by regenerating the entire drops list).

----

This PR solves the problem by adding two new methods to `IForgeItem` and `IForgeItemStack`:
* `getEnchantmentLevel`: Allows an item to override the level of an enchantment from NBT, for hooks using `EnchantmentHelper::getItemEnchantmentLevel`
* `getAllEnchantments`: Allows an item to override the method returning a full enchantment list, for hooks using `EnchantmentHelper.deserializeEnchantments(stack.getEnchantmentTags())`.

For some implementation notes:
* Enchantments from the hooks are ignored in vanilla enchanting blocks such as the grindstone, anvil, and enchanting table. It is assumed that if you are using these hooks, your intention is to implement a system parallel to enchantments.
* Enchantments from the hooks are not shown in the tooltip. If the modder wishes to show them in the tooltip, they can override `Item::appendHoverText` (which notably runs right before vanilla adds enchantment names)
* Returning enchantments will not prevent enchanting the tool in a vanilla table or anvil or make the tool have the enchantment glint, as between forge and vanilla there are enough methods to control that behavior.
* `getAllEnchantments` does not return book enchantments, it is meant for gameplay logic, not enchanting.

To minimize patches:
* `EnchantmentHelper::getItemEnchantmentLevel` was patched to call the new method, and a new method `EnchantmentHelper::getTagEnchantmentLevel` for cases that specifically want to read from NBT (two cases in vanilla enchanting logic and the default for the hook). As a result, the forge hook and vanilla method are interchangeable. 
* `EnchantmentHelper::getEnchantments` was every in vanilla that used that method was part of enchanting logic (anvil, enchantment table, etc.). Mods should not have been calling that method anyways as it returns book enchantments, which should not be used for gameplay.

----

Included with this PR is a test mod that adds "soul boots". These boots use the hook to include both soul speed and the curse of binding, two enchantments I have noted are incredibly hard to impossible to replicate without this hook.


sciwhiz12 (changes requested):
Please update the license headers of the following modified files by changing the year of `2021` to the year `2022`:
 - `src/main/java/net/minecraftforge/common/extensions/IForgeItem.java`
 - `src/main/java/net/minecraftforge/common/extensions/IForgeItemStack.java`
 - `src/test/java/net/minecraftforge/debug/gameplay/loot/GlobalLootModifiersTest.java`
 - `src/test/java/net/minecraftforge/debug/misc/CustomTagTypesTest.java`

src/main/java/net/minecraftforge/common/extensions/IForgeItemStack.java
```diff
@@ -178,6 +181,25 @@ default int getItemEnchantability()
...
+     */
+    default int getEnchantmentLevel(Enchantment enchantment)
+    {
+        return self().getItem().getEnchantmentLevel(self(), enchantment);
+    }
+
+    /**
+     * Gets a map of all enchantments present on the stack.
+     * Use in place of {@link EnchantmentHelper#getEnchantments(ItemStack)} for gameplay enchantments, as this enforces the enchantment is not from an enchanted book.
+     */
```
pupnewfster: This could use having the javadocs updated to match the newer javadocs on the methods added in IForgeItem

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -533,6 +535,33 @@ default boolean canApplyAtEnchantingTable(ItemStack stack, Enchantment enchantme
         return enchantment.category.canEnchant(stack.getItem());
     }
 
+    /**
+     * Gets the level of the enchantment currently present on the stack. By default, returns the enchantment level present in NBT.
+     * Most enchantment implementations rely upon this method.
+     * For consistency, results of this method should be the same as getting the enchantment from {@link #getAllEnchantments(ItemStack)}
```
pupnewfster, line 541: I wonder if it is worth mentioning something along the lines of the difference between these two methods and the ones left in EnchantmentHelper so people can have a better idea of what cases they may care about NBT and ignore non NBT ones.

patches/minecraft/net/minecraft/world/item/enchantment/EnchantmentHelper.java.patch
```diff
@@ -1,5 +1,31 @@
...
+@@ -59,7 +_,12 @@
+       return Registry.f_122825_.m_7981_(p_182433_);
+    }
+ 
++   @Deprecated // forge: use getTagEnchantmentLevel or ItemStack::getEnchantmentLevel
+    public static int m_44843_(Enchantment p_44844_, ItemStack p_44845_) {
++      return p_44845_.getEnchantmentLevel(p_44844_);
++   }
++
++   public static int getTagEnchantmentLevel(Enchantment p_44844_, ItemStack p_44845_) {
```
pupnewfster, line 12: Just pointing this out so it isn't missed but I am unsure if this counts as a logical breaking change or not as there are times when a mod may be calling this but only want the tag enchantment levels (though it would be a lot more maintenance heavy to patch it a different way). Mostly just want more opinions on this given cases where even in vanilla (the grindstone menu) it needed to be changed to be explicitly NBT based.

src/test/java/net/minecraftforge/debug/item/EnchantmentLevelTest.java
```diff
@@ -0,0 +1,79 @@
...
+                }
+                if (enchantment == Enchantments.BINDING_CURSE)
+                {
+                    return 1;
+                }
+                return super.getEnchantmentLevel(stack, enchantment);
+            }
+
+            @Override
+            public Map<Enchantment,Integer> getAllEnchantments(ItemStack stack)
```
pupnewfster: Missing space after comma

src/test/java/net/minecraftforge/debug/item/EnchantmentLevelTest.java
```diff
@@ -0,0 +1,79 @@
...
+            {
+                Map<Enchantment,Integer> map = super.getAllEnchantments(stack);
```
pupnewfster: Missing space after comma

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
KnightMiner: Not on this doc, but maybe on `IForgeItemStack`. I treated the docs on `IForgeItem` as info for the modder overriding the method, while the docs on `IForgeItemStack` were for modders wanting enchantments.

Edit: added

patches/minecraft/net/minecraft/world/item/enchantment/EnchantmentHelper.java.patch
KnightMiner: I agree its probably a breaking change, but I suspect it will affect minimal mods. Mods that implement custom enchantments shouldn't have any trouble, while mods that add custom enchanting methods or machines to remove enchantments will need to update to the tag based methods. I suspect even among custom enchantment methods, most of them are probably using the map method making it a very small list of mods affected, so my opinion is an announcement would be sufficient.

If that is deemed as too big of a breaking change, sounds like 1.18.2 is coming soon, so could always include it there.

KnightMiner:
I have converted this PR to a draft as its my understanding the change to `EnchantmentHelper::getItemEnchantmentLevel` makes this a breaking change, as there is no sensible behavior that will catch both the people using that method to get enchantments for enchanting sources and the people wanting enchantments for gameplay mechanics.

It will be updated again once 1.19 comes around, or if anyone has a good suggestion for how to make this not a breaking change.

KnightMiner:
Closing as I opened a new PR for 1.19, ended up being a bit easier than rebasing.

PR #8421, opened 2022-01-23
[1.18.x] Allow new Multi-noise Climate Parameters to be added

SmellyModder:
# Background
Minecraft 1.18 introduced the `MultiNoiseBiomeSource`, a biome source in which biomes get distributed by climate parameter points. Every climate parameter point is made up of 7 climate parameters: `temperature`, `humidity`, `continentalness`, `erosion`, `depth`, `weirdness`, and `offset`.
A `ParameterList` contains parameter points paired with biomes and is used to search for the most fitting biome for a sampled `TargetPoint`. A `TargetPoint` contains sampled values for every climate parameter, excluding `offset`.

There is a video by Henrik Kniberg that goes into detail about how the multi-noise biome source works and what each of the climate parameters does: https://www.youtube.com/watch?v=TycBrFKEteU

All of these parameters uniquely customize the distribution of biomes, but there is a flaw with the system. When people add a new biome to an existing configured `MultiNoiseBiomeSource`, there tend to be awkward distribution issues because specific areas have gotten sliced for specific biome sections.
For example, a mod that adds a new ocean biome to the overworld has tight constraints because the only climate parameter that differentiates the overworld's oceans is `temperature`. One solution to this problem is to add a new climate parameter for a slice of the world that will use that mod's ocean distribution, but that is easier said than done.

## Implementation
`Climate` and `NoiseSampler` have gotten patched to allow for extensible, and highly customizable, climate parameters.
A new system has gotten added to Forge's `BiomeManager` class that handles the extensible climate parameters.
New climate parameters are added by calling the `BiomeManager#addClimateParameter(...)` method during mod loading.

However, this is only one step needed for compatible modification of multi-noise biome sources as a simple way to add new entries to configured multi-noise biome sources currently doesn't exist.
Fortunately, there is a PR open right now that adds this missing step: <https://github.com/MinecraftForge/MinecraftForge/pull/8316>

## Test Mod
The `climate_parameter_extensions_test` test mod adds two new climate parameters: `height` and `moddedness`.

`height` is a parameter value between 0 and 1 inclusive that's a percentage of the dimension's height.
`moddedness` is a parameter value that uses a new noise that comes in a large form when large biomes are enabled.

A new dimension file is located at `/data/climate_parameter_extensions_test/dimension/extended.json` and makes use of these two added climate parameters.

### Performance Impacts
Multi-noise sampling is frequently processed, and a large amount of added climate parameters may slow world generation down for all multi-noise biome sources. During my testing with 7 added climate parameters (double the amount of vanilla), I did not notice any slowing of worldgen performance on my laptop and PC, but other machines should get tested.
### Risks of affecting existing systems
There may be mods with custom systems that assume only 7 climate parameters are possible, and this PR could mess up distributions or even break them.
### Distribution issues can still occur if used incorrectly
If new biome entries get added to configured multi-noise biome sources without accounting for the possibility that other mods may want to do the same, distribution issues can still occur. Essentially, modders need to be aware that other mods may add their own slices and should modify the entires of configured multi-noise biome sources with that in mind.

### DIFF
--- a/patches/minecraft/net/minecraft/world/level/biome/Climate.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/biome/Climate.java.patch
@@ -1,0 +1,88 @@
+--- a/net/minecraft/world/level/biome/Climate.java
++++ b/net/minecraft/world/level/biome/Climate.java
+@@ -145,7 +_,7 @@
+       }
+    }
+ 
+-   public static record ParameterPoint(Climate.Parameter f_186863_, Climate.Parameter f_186864_, Climate.Parameter f_186865_, Climate.Parameter f_186866_, Climate.Parameter f_186867_, Climate.Parameter f_186868_, long f_186869_) {
++   public static record ParameterPoint(Climate.Parameter temperature, Climate.Parameter humidity, Climate.Parameter continentalness, Climate.Parameter erosion, Climate.Parameter depth, Climate.Parameter weirdness, long offset, net.minecraftforge.common.BiomeManager.ParameterPointExtensions extensions) {
+       public static final Codec<Climate.ParameterPoint> f_186862_ = RecordCodecBuilder.create((p_186885_) -> {
+          return p_186885_.group(Climate.Parameter.f_186812_.fieldOf("temperature").forGetter((p_186905_) -> {
+             return p_186905_.f_186863_;
+@@ -161,15 +_,18 @@
+             return p_186888_.f_186868_;
+          }), Codec.floatRange(0.0F, 1.0F).fieldOf("offset").xmap(Climate::m_186779_, Climate::m_186796_).forGetter((p_186881_) -> {
+             return p_186881_.f_186869_;
+-         })).apply(p_186885_, Climate.ParameterPoint::new);
++         }), net.minecraftforge.common.BiomeManager.ParameterPointExtensions.CODEC.optionalFieldOf("extensions").xmap(o -> o.orElse(net.minecraftforge.common.BiomeManager.ParameterPointExtensions.getDefault()), a -> java.util.Objects.equals(a, net.minecraftforge.common.BiomeManager.ParameterPointExtensions.getDefault()) ? java.util.Optional.empty() : java.util.Optional.of(a)).forGetter(ParameterPoint::extensions)).apply(p_186885_, Climate.ParameterPoint::new);
+       });
+ 
++      public ParameterPoint(Climate.Parameter f_186863_, Climate.Parameter f_186864_, Climate.Parameter f_186865_, Climate.Parameter f_186866_, Climate.Parameter f_186867_, Climate.Parameter f_186868_, long f_186869_) {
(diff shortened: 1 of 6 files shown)


alcatrazEscapee:
So, I do quite like the concept of this API. It's - as far as I can tell - provides similar capabilities to the [existing TerraBlender mod](https://github.com/Glitchfiend/TerraBlender) for modifying the multinoise biome source via adding new noises/climate parameters. It also appears to enable the [suggested method for composing overworld biomes](https://twitter.com/kingbdogz/status/1466672508250247170) by kingbdogz on twitter.

However looking at this PR there appear to be two major concerns that stand out:

1. The continued use of `BiomeManager`. To my understanding, everything in `BiomeManager` currently is *completely* unused, deprecated, and moreover, fundamentally incompatible at this point with multi noise anyway (The proposed resolution, #multi. And adding *new functionality* to that class I think is a huge mistake given it's current situation.
    - `BiomeManager` itself, should be either removed, or deprecated for removal as soon as possible.
    - Everything here that is added to `BiomeManager` should be added to it's own class, i.e. `MultiNoiseBiomeSourceManager` or similar.

However, both of those points hint at another major problem I see with this:

2. Breaking changes. To my knowledge:
    - Removal of `BiomeManager` would be a breaking change (this can just be deprecated instead).
    - **The modifications to the signature of [`ParamaterPoint`](https://github.com/MinecraftForge/MinecraftForge/pull/8421/files#diff-d1a3a1e4792eb8f2f32655e8cfbb884755076e6321bef64b080ae85540b3fc1cR7-R8) and [`TargetPoint`](https://github.com/MinecraftForge/MinecraftForge/pull/8421/files#diff-d1a3a1e4792eb8f2f32655e8cfbb884755076e6321bef64b080ae85540b3fc1cR77-R78) are breaking changes**. (@sciwhiz12 correct me if I'm wrong (which I appear to have been, for I didn't notice the added bouncer ctor), ~~this should give this the breaking change label.~~)

Now, Forge is still in beta (although for how long?) and those breaking changes seem kind of crucial to this PR, which puts this in a bit of an awkward situation. It's made worse by this PR's dependence on other world generation API PRs (most importantly, the multi noise biome source loaded event, but also resource key tags to a lesser extent).

---

It does beg the question of if "just use terrablender" is an valid option, if it will end up being the defacto solution instead of this.

SmellyModder:
I agree that the existing code in `BiomeManager` is essentially useless these days. Deprecating `BiomeManager` and making a new class specifically for this API sounds good. I only put it in `BiomeManager` because that was about the most fitting existing place.

I am well aware of this API's dependence on other PRs, and the option to "just use terrablender" may be acceptable for some developers. This PR is made to be a compatible solution for adding new climate parameters to multi-noise built into Forge, so developers don't have to depend on other mods or come up with their own solutions.
In the end, if many developers don't mind depending on Terrablender or don't mind making their own solutions, then this API may really not end up being useful.

SmellyModder:
Closing this now because I've discovered a clean, configurable, easy to re-create, and compatible solution for modded biome distributions: <https://github.com/team-abnormals/blueprint/commit/7a65df5aab627d31dff13af322d4b1f88b58952c>

PR #8424, opened 2022-01-25
[1.18.x] Changing hardcoded Items.FISHING_ROD to allow for custom fishing rods

rehwinkel:
1.18 port of #8168.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ToolActions.java
+++ b/src/main/java/net/minecraftforge/common/ToolActions.java
@@ -114,6 +114,9 @@ public class ToolActions
      */
     public static final ToolAction SHIELD_BLOCK = ToolAction.get("shield_block");
 
+    /**
+     * This action corresponds to right-clicking the fishing rod.
+     */
     public static final ToolAction FISHING_ROD_CAST = ToolAction.get("fishing_rod_cast");
 
     // Default actions supported by each tool type
--- a/src/main/java/net/minecraftforge/common/ToolActions.java
+++ b/src/main/java/net/minecraftforge/common/ToolActions.java
@@ -114,6 +114,8 @@ public class ToolActions
      */
     public static final ToolAction SHIELD_BLOCK = ToolAction.get("shield_block");
 
+    public static final ToolAction FISHING_ROD_CAST = ToolAction.get("fishing_rod_cast");
+
     // Default actions supported by each tool type
     public static final Set<ToolAction> DEFAULT_AXE_ACTIONS = of(AXE_DIG, AXE_STRIP, AXE_SCRAPE, AXE_WAX_OFF);
     public static final Set<ToolAction> DEFAULT_HOE_ACTIONS = of(HOE_DIG /* TODO: , HOE_TILL */);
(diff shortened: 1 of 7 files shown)


pupnewfster:
Shouldn't these be using forge's [instanceof transformer](https://github.com/MinecraftForge/MinecraftForge/blob/1.18.x/build.gradle#L1084)?

KnightMiner:
Alternatively, is this a case that would benefit from tool actions?

Yes I know, repeat of mine and pup's comments from #8168.

eerussianguy:
There are only two relevant patches, one of which is in the renderer hook. I don't think we need to overthink this.

KnightMiner:
Using tool actions instead of instanceof is not about overthinking it, its  about flexibility. instanceof means you prevent anyone from benefitting from this patch unless they extend the vanilla class. Is there a case where someone would want this and not be extending the vanilla class?

See also the shears patches, they use the tool action instead of instanceof shears.

----

The transformer suggested is also not overthinking it. Forge does not want to maintain any patches that are trivial transformers. It is a 3 line "config" file addition, does not need to be a patch.

rehwinkel:
@KnightMiner I implemented it using ToolAction!

src/test/java/net/minecraftforge/debug/item/CustomFishingRodTest.java
```diff
@@ -0,0 +1,101 @@
...
+        public CustomFishingRodItem(Item.Properties props)
+        {
+            super(props);
+        }
+
+        @Override
+        public boolean canPerformAction(ItemStack stack, ToolAction toolAction)
+        {
+            return toolAction.equals(ToolActions.FISHING_ROD_CAST);
+        }
```
pupnewfster: This shouldn't be needed as  this class is extending FishingRodItem

src/main/java/net/minecraftforge/common/ToolActions.java
```diff
@@ -114,6 +114,8 @@
      */
     public static final ToolAction SHIELD_BLOCK = ToolAction.get("shield_block");
 
+    public static final ToolAction FISHING_ROD_CAST = ToolAction.get("fishing_rod_cast");
```
pupnewfster, line 120: This should probably have a one sentence javadoc like the other tool actions do.

GirafiStudios:
@rehwinkel looks like this PR unfortunately have some conflicts now. Could you resolve those? :)

mcenderdragon:
um any news on this ?

sciwhiz12:
Hello! @rehwinkel, please leave a comment whether you wish to continue this PR. As a similar PR was merged to 1.19.2 recently -- #8681 -- this should conform to the changes in that PR, or an explanation be given on why it should differ in implementation. If we do not hear back from you after some time, this PR will be closed for inactivity.

rehwinkel:
This PR can be closed, I'm not going to be working on this any time soon.

sciwhiz12:
Closing for now. Thank you for your contributions!

PR #8425, opened 2022-01-25
[1.18.x] New Function for when an Item stops being used

rehwinkel:
This PR basically just passes `LivingEntity#stopUsingItem` onto the last Item being held. The new Item method is also called when switching slots while using an item, which the `releaseUsing` and `finishUsingItem` functions don't do.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
@@ -489,11 +489,14 @@
           if (this.f_20935_.m_41781_()) {
              this.m_21329_();
           }
-@@ -2951,6 +_,7 @@
+@@ -2951,6 +_,10 @@
           this.m_21155_(1, false);
        }
  
-+      this.f_20935_.stopUsingItem(this);
++      if (!this.f_20935_.m_41619_())
++      {
++         this.f_20935_.onStopUsingItem(this);
++      }
        this.f_20935_ = ItemStack.f_41583_;
        this.f_20936_ = 0;
     }
--- a/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
@@ -489,6 +489,14 @@
           if (this.f_20935_.m_41781_()) {
              this.m_21329_();
           }
+@@ -2951,6 +_,7 @@
+          this.m_21155_(1, false);
+       }
+ 
++      this.f_20935_.stopUsingItem(this);
+       this.f_20935_ = ItemStack.f_41583_;
+       this.f_20936_ = 0;
+    }
 @@ -3084,8 +_,8 @@
        }
  
--- a/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
@@ -489,14 +489,12 @@
           if (this.f_20935_.m_41781_()) {
              this.m_21329_();
           }
-@@ -2951,6 +_,10 @@
+@@ -2951,6 +_,8 @@
           this.m_21155_(1, false);
        }
  
-+      if (!this.f_20935_.m_41619_())
-+      {
-+         this.f_20935_.onStopUsingItem(this);
-+      }
++      if (!this.f_20935_.m_41619_()) this.f_20935_.onStopUsingItem(this);
++
        this.f_20935_ = ItemStack.f_41583_;
        this.f_20936_ = 0;
     }
--- a/patches/minecraft/net/minecraft/world/item/ItemStack.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/ItemStack.java.patch
@@ -217,7 +217,7 @@
        return multimap;
     }
  
-@@ -954,6 +_,23 @@
+@@ -954,6 +_,28 @@
  
     public boolean m_41614_() {
        return this.m_41720_().m_41472_();
@@ -238,6 +238,11 @@
 +         this.gatherCapabilities(() -> f_41589_.initCapabilities(this, this.capNBT));
 +         if (this.capNBT != null) deserializeCaps(this.capNBT);
 +      }
++   }
++
++   public void stopUsingItem(net.minecraft.world.entity.LivingEntity entity)
++   {
++      this.m_41720_().stopUsingItem(entity, this);
     }
  
     public SoundEvent m_41615_() {
--- a/patches/minecraft/net/minecraft/world/item/ItemStack.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/ItemStack.java.patch
@@ -217,7 +217,7 @@
        return multimap;
     }
  
-@@ -954,6 +_,28 @@
+@@ -954,6 +_,23 @@
  
     public boolean m_41614_() {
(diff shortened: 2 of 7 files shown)


thedarkcolour:
Would it be better to have that method in `IForgeItem`?

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -109,6 +109,15 @@ default InteractionResult onItemUseFirst(ItemStack stack, UseOnContext context)
         return InteractionResult.PASS;
     }
 
+    /**
```
diesieben07, line 112: Please add some documentation about when this is called

patches/minecraft/net/minecraft/world/item/ItemStack.java.patch
```diff
@@ -238,6 +238,11 @@
 +         this.gatherCapabilities(() -> f_41589_.initCapabilities(this, this.capNBT));
 +         if (this.capNBT != null) deserializeCaps(this.capNBT);
 +      }
++   }
++
++   public void stopUsingItem(net.minecraft.world.entity.LivingEntity entity)
++   {
++      this.m_41720_().stopUsingItem(entity, this);
```
diesieben07: This should be in `IForgeItemStack`

src/test/java/net/minecraftforge/debug/item/StopUsingItemHookTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private static class ThingItem extends Item
+    {
+
+        public ThingItem(Item.Properties props)
+        {
+            super(props);
+        }
+
+        public int getUseDuration(ItemStack stack)
```
diesieben07, line 65: Missing `@Override`

src/test/java/net/minecraftforge/debug/item/StopUsingItemHookTest.java
```diff
@@ -0,0 +1,64 @@
...
+            return ItemUtils.startUsingInstantly(level, player, hand);
+        }
+
+        @Override
+        public void stopUsingItem(LivingEntity e, ItemStack itemStack)
+        {
+            super.stopUsingItem(e, itemStack);
+            if (!e.getLevel().isClientSide())
+            {
+                System.out.println("Item Over!");
```
diesieben07: Should use a Logger for this.

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -109,6 +109,15 @@ default InteractionResult onItemUseFirst(ItemStack stack, UseOnContext context)
...
+     * @param entity    The entity using the item
+     * @param itemStack The last {@link ItemStack} held before it stopped being used.
+     */
+    default void stopUsingItem(net.minecraft.world.entity.LivingEntity entity, ItemStack itemStack)
```
diesieben07: No need for fully qualified names here.
The parameter name for `ItemStack` parameters is usually just `stack`

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -109,6 +109,15 @@ default InteractionResult onItemUseFirst(ItemStack stack, UseOnContext context)
         return InteractionResult.PASS;
```
diesieben07, line 109: Please update the license header in this file

src/test/java/net/minecraftforge/debug/item/StopUsingItemHookTest.java
```diff
@@ -0,0 +1,64 @@
+package net.minecraftforge.debug.item;
```
diesieben07, line 20: Missing license header

diesieben07 (changes requested):

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -109,6 +109,17 @@ default InteractionResult onItemUseFirst(ItemStack stack, UseOnContext context)
...
     }
 
+    /**
+     * When an item stops being used, by switching slots, releasing the mouse button or whatever, this method is called.
```
diesieben07:
```suggestion
     * This method is called when this item stops being used by switching slots, releasing the "use item" button or any other means.
```

src/main/java/net/minecraftforge/common/extensions/IForgeItemStack.java
```diff
@@ -124,6 +124,17 @@ default boolean canPerformAction(ToolAction toolAction)
         return self().getItem().canPerformAction(self(), toolAction);
     }
 
+    /**
+     * This method is called in {@link LivingEntity#stopUsingItem()}.
+     * When an item stops being used, by switching slots, releasing the mouse button or whatever, this method is called.
```
diesieben07:
```suggestion
     * This method is called when this item stops being used by switching slots, releasing the "use item" button or any other means.
```

diesieben07 (changes requested):
In testing I have found that `stopUsingItem` is called even if the item is currently not in use and you just switched slots.

rehwinkel:
I've added a check in the stopUsingItem function to avoid calling the hook on air items. When switching slots, it is now only called when the "used" item is not empty. In just a sec I'll change the condition in the patch to be in-line.

patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
```diff
@@ -489,6 +489,17 @@
...
              this.m_21329_();
           }
+@@ -2951,6 +_,10 @@
+          this.m_21155_(1, false);
+       }
+ 
++      if (!this.f_20935_.m_41619_())
++      {
++         this.f_20935_.onStopUsingItem(this);
++      }
```
diesieben07: You can keep this patch smaller by either putting this on one line or moving the if statement into the IForgeItemStack method.
Although I now realize that my point about this being called every time is moot because you call it on the currently used item, which will be EMPTY at that point and do nothing. So sorry, but you can revert adding this if statement

diesieben07 (changes requested):

patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
rehwinkel: I think saving two function calls every time a slot is switched is still worth it probably.

diesieben07: Okay, fair enough

diesieben07 (approved):

sciwhiz12 (changes requested):
- `src/main/java/net/minecraftforge/common/extensions/IForgeItem.java`
- `src/main/java/net/minecraftforge/common/extensions/IForgeItemStack.java`
- `src/test/java/net/minecraftforge/debug/item/StopUsingItemHookTest.java`

sciwhiz12:
Hello! @rehwinkel, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.19.2 (or leave an explanation why this PR's changes are not applicable to 1.19.2). If we do not hear back from you after some time, this PR will be closed for inactivity.

rehwinkel:
Im currently not planning on continuing work on this PR. You can close it if you like.

sciwhiz12:
Closing for now. Thank you for your contributions!

PR #8437, opened 2022-01-29
Fix long keybinds making shorter ones unreadable

sekwah41:
In mods such as JEI at the default scaling value their translation strings can be too long making the text unreadable for other mods.
image

image

Fix
image

Just adding a max limit to how far they can offset so only mods with long names are affected.

I was considering making it change the text to ... but that would cause issues where you are right at the border and want another symbol, so decided this was probably better.

### DIFF
--- a/patches/minecraft/net/minecraft/client/gui/screens/controls/KeyBindsList.java.patch
+++ b/patches/minecraft/net/minecraft/client/gui/screens/controls/KeyBindsList.java.patch
@@ -1,6 +1,38 @@
 --- a/net/minecraft/client/gui/screens/controls/KeyBindsList.java
 +++ b/net/minecraft/client/gui/screens/controls/KeyBindsList.java
-@@ -51,7 +_,7 @@
+@@ -40,18 +_,37 @@
+          }
+ 
+          Component component = Component.m_237115_(keymapping.m_90860_());
+-         int i = p_193862_.f_91062_.m_92852_(component);
++
++         // FORGE: max width for the keybind descriptions to make all readable, clip any strings that are too long
++         String plainText = component.getString();
++         String clippedMessage = "";
++
++         int i = 0;
++
++         for (int j = 0; j <= plainText.length(); j++) {
++            String checkString = plainText.substring(0, j);
++            i = p_193862_.f_91062_.m_92895_(plainText.substring(0, j));
++            if(i <= 185) {
++               clippedMessage = checkString;
++            } else {
++               clippedMessage = clippedMessage.substring(0, Math.max(clippedMessage.length() - 3, 0));
++               clippedMessage += "...";
++               i = p_193862_.f_91062_.m_92895_(clippedMessage);
++               break;
++            }
++         }
++
+          if (i > this.f_193859_) {
+             this.f_193859_ = i;
+          }
+ 
+-         this.m_7085_(new KeyBindsList.KeyEntry(keymapping, component));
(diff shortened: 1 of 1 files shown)


patches/minecraft/net/minecraft/client/gui/screens/controls/KeyBindsList.java.patch
```diff
@@ -1,6 +1,16 @@
 --- a/net/minecraft/client/gui/screens/controls/KeyBindsList.java
 +++ b/net/minecraft/client/gui/screens/controls/KeyBindsList.java
-@@ -53,7 +_,7 @@
+@@ -47,13 +_,17 @@
+             this.f_193859_ = i;
+          }
+ 
++         if(this.f_193859_ > 175) {
++            this.f_193859_ = 175;
++         }
```
sciwhiz12: Please include a Forge comment marker (`// FORGE: ...`) explaining why this change is in place, for future maintainability.

Also, this is better placed after the `for`-loop, to check and set the field's value once after all keybinds have been processed rather than checking each time.

sciwhiz12 (changes requested):
This works fine, though I wonder how complicated it would be to implement other solutions. 

For example, splitting the keybind name across multiple lines (whether in `KeyBindsList` automatically or by the mod manually in its language file or something similar), or adding a tooltip when hovering over a keybind's name to show the full name if it is clipped (the text width exceeding the 175px limit).

Shadows-of-Fire:
Good catch, we don't want to have longer names breaking the whole gui.  Though we probably want to make that full name visible somehow...

A solution adding a hovered tooltip or a scrollable name when hovered should both be viable, though the former is much easier to do.

patches/minecraft/net/minecraft/client/gui/screens/controls/KeyBindsList.java.patch
sekwah41: Fair point I'll add that when I get a moment in the next few days.

sekwah41:
@Shadows-of-Fire I can take a look at multi lining or a hover UI when I'm back as I'm on a break. multi lining would probably make most sense otherwise you have hover then multi line.
Though hover may just be easier to implement.

sciwhiz12:
Hello! @sekwah41, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.19.2 (or leave an explanation why this PR's changes are not applicable to 1.19.2). If we do not hear back from you after some time, this PR will be closed for inactivity.

sekwah41:
Ah sorry I've been quite busy. I'll add it to the list to get to sometime this or next week hopefully. It shouldn't take much changing.

sekwah41:
Ill make a new pr with these changes for 1.19 but backport it to this too.

sekwah41:
image

sekwah41:
Switched over to forcefully shorten the string rather than shot it behind the button. Can revert it if that is preferred to just spill behind the button if its too long.

I was planning to possibly make it so it shows a tooltip when you hover. But sadly this section of code doesn't extend screen or doesn't render directly so its a bit more complex

sekwah41:
Ill push the changes and other parts to fix it now

sekwah41:
ignore the fact my branch says 1.18, I've force pushed 1.19 to it on my end so its actually the right one

sekwah41:
Not sure if this is still an issue or not though I am going to leave this or someone else to handle as I've updated it a few times.

I know none of the mods I have atm at least cause this issue.

PR #8441, opened 2022-01-31
Fix PlayerChangeGameModeEvent and improve test mod

### LINKED ISSUE #8439: PlayerChangeGameModeEvent doesn't use the game mode set in the event
**Minecraft Version:** 1.16.3 -> 1.18.1

**Forge Version:** 34.1.20 -> 39.0.64 (since the event was added)

**Logs:** No logs

**Steps to Reproduce:**
 1. Setup an event handler for `PlayerChangeGameModeEvent`
 2. Try to set a new game mode with `setNewGameMode` and nothing happens

**Description of issue:**
The `PlayerChangeGameModeEvent` is not hooked up to use the value changed with `setNewGameMode`.
To work around this I just cancel the event and set the game mode on the player.

The doc of the `setNewGameMode` method:
> Sets the game mode the player will be changed to if this event is not canceled.

I think it should either be hooked up to use that value if not canceled like the doc say or just remove it as it's just as easy to set the game mode on the player manually. With the method being like it is right now it's confusing why it doesn't work until you start debugging why it doesn't work.

The event was added in #7355

diesieben07:
Fixes #8439

### DIFF
--- a/patches/minecraft/net/minecraft/server/level/ServerPlayer.java.patch
+++ b/patches/minecraft/net/minecraft/server/level/ServerPlayer.java.patch
@@ -134,11 +134,12 @@
     }
  
     protected void m_142540_(MobEffectInstance p_143393_, @Nullable Entity p_143394_) {
-@@ -1181,6 +_,7 @@
+@@ -1181,6 +_,8 @@
     }
  
     public boolean m_143403_(GameType p_143404_) {
-+      if (!net.minecraftforge.common.ForgeHooks.onChangeGameMode(this, this.f_8941_.m_9290_(), p_143404_)) return false;
++      p_143404_ = net.minecraftforge.common.ForgeHooks.onChangeGameType(this, this.f_8941_.m_9290_(), p_143404_);
++      if (p_143404_ == null) return false;
        if (!this.f_8941_.m_143473_(p_143404_)) {
           return false;
        } else {
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -1,6 +1,6 @@
 /*
  * Minecraft Forge
- * Copyright (c) 2016-2021.
+ * Copyright (c) 2016-2022.
  *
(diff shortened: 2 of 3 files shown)


patches/minecraft/net/minecraft/server/level/ServerPlayer.java.patch
```diff
@@ -134,11 +134,12 @@
     }
  
     protected void m_142540_(MobEffectInstance p_143393_, @Nullable Entity p_143394_) {
-@@ -1181,6 +_,7 @@
+@@ -1181,6 +_,8 @@
     }
  
     public boolean m_143403_(GameType p_143404_) {
-+      if (!net.minecraftforge.common.ForgeHooks.onChangeGameMode(this, this.f_8941_.m_9290_(), p_143404_)) return false;
++      p_143404_ = net.minecraftforge.common.ForgeHooks.onChangeGameType(this, this.f_8941_.m_9290_(), p_143404_);
```
TheCurle, line 141: This should be onChangeGameMode (the boolean), not onChangeGameType (the Game Type)

TheCurle (changes requested):
This is a behavior change as implemented, probably a typo.

patches/minecraft/net/minecraft/server/level/ServerPlayer.java.patch
SizableShrimp: But then you can't set the parameter?

diesieben07: No it should not...? onChangeGameMode is the old, deprecated one.
A boolean is not enough as a return type, because the event can change the new gametype.

TheCurle: Oh I see, I misunderstood the next line. Ignore my comment, then.

TheCurle (approved):

TheCurle:
Okay yes, i misunderstood how these functions relate to each other.
With that in mind, this now looks good to me.

PR #8442, opened 2022-02-01
Add tags for various EntityTypes

MichaelTRoss0:
Added multiple EntityType tags to allow for common interaction between certain mobs in different mods.

This PR is currently a work in progress.


KnightMiner:
I think you misunderstand how the tags in forge's `Tags`  class work. Nothing stops a modder from adding their own tags otherwise, and the purpose of them is partly for datagen.

In addition, you cannot place a file under `generated` without adding the appropriate data generator. The tags you added will be deleted next time datagen is run.

----

So, two fixes suggested:
* This PR is not adding new functionality, you are simply proposing some standard tags. Update the description to say that to avoid confusion
* Use datagenerators to create the tags.

MichaelTRoss0:
Thank you for the pointers. I've made all the changes you've suggested.

src/main/java/net/minecraftforge/common/Tags.java
```diff
@@ -416,6 +420,19 @@ private static void init(){}
         }
     }
 
+    public static class EntityTypes
+    {
+        private static void init() {}
+
+        public static final Tags.IOptionalNamedTag<EntityType<?>> BOSSES = tag("bosses");
+        public static final Tags.IOptionalNamedTag<EntityType<?>> CONSTRUCTS = tag("constructs");
```
ChampionAsh5357: What is constructs? You have elder guardian and guardian along with iron and snow golems in here. I'm unsure what this tag would mean. If you're talking about entities that can be created, then the former two wouldn't be valid. Additionally, ender dragons or villagers could be created as well. If you're talking about entities that can spawn in structures, then there is a massive list missing, snow golems not one of them which spawn in a structure.

MichaelTRoss0: Constructs are artificial creatures that were made of nonliving materials and operate fully independently. Guardians fit because they seem to be made of prismarine and don't suffocate when removed from water.  I'll remove the tag if you guys don't think it's too helpful, or I'll add clarification on what this tag means.

SizableShrimp: This seems rather niche and something I think would best fit as a tag for your specific mod (if you even have a use). What would be the use case of such a tag existing in forge?

MichaelTRoss0: There is no real reason. I just saw that Forge didn't have any preexisting entity tags generated in it, so I decided to add one or two. I'll go ahead and remove it and just leave the bosses tag.

KnightMiner:
Based on

> There is no real reason. I just saw that Forge didn't have any preexisting entity tags generated in it, so I decided to add one or two. I'll go ahead and remove it and just leave the bosses tag.

Its really worth clarifying the goal of this PR. It sounds like you don't directly need the tags you are proposing, so a PR adding a single tag does not seem super useful. Afterall, mods can agree on whatever convention they want without forge adding the tags.

So basically, if your goal is to add tags so forge has tags, we should add more than one standard tag (be worth discussing on discord). If you goal is to add a boss tag because we want a way to identify bosses, then this PR could simply be framed as "added a boss entity tag" and you should suggest some reasons why the tag is useful

MichaelTRoss0:
Alright, I’m at my job right now, so I won’t be able to work in this until
I come back. I’ll check out the Discord once I get home, and discuss it
with you guys then. This is my first time making a pull request to
anything, but I’m eager to figure things out.

Thelnfamous1:
A forge defined tag for bosses is quite good for general use. There’s no good generalized way to check if an entity is a boss currently.

src/main/java/net/minecraftforge/common/data/ForgeEntityTypeTagsProvider.java
```diff
@@ -0,0 +1,45 @@
...
+package net.minecraftforge.common.data;
+
+import net.minecraft.data.DataGenerator;
+import net.minecraft.data.tags.EntityTypeTagsProvider;
+import net.minecraft.world.entity.EntityType;
+import net.minecraftforge.common.Tags;
+
+public class ForgeEntityTypeTagsProvider extends EntityTypeTagsProvider {
+
+    public ForgeEntityTypeTagsProvider(DataGenerator generator, String modId, ExistingFileHelper existingFileHelper)
```
pupnewfster: The call initializing and adding this provider appears to be missing.

MichaelTRoss0: That's in common/Tags.java, isn't it?

pupnewfster: No, if you actually try running the datagen you will see the file you manually created in the generated folder get deleted... Providers forge needs to run should be added in forge's [GatherDataEvent listener](https://github.com/MinecraftForge/MinecraftForge/blob/1.18.x/src/main/java/net/minecraftforge/common/ForgeMod.java#L209).

MichaelTRoss0: I'll add that in my next commit and push.

Alpvax:
I think you have gone into far too much detail with these. about half of them are either empty or redirect to a single other tag.
For example, there is no reason to have a felids/small tag when there are no medium or large felids. The very few mods which would like this distinction can manage it themselves. Alternatively, you could group all entities by their size (i.e. small: < 1block, medium: 1-2 blocks, large: > 2 blocks) but again, I think that is unnecessary.

I would much prefer to have simpler tags, not tags based on the latin names: cats (not animals/cats, although if others would prefer that then fine), birds, etc.
Then have tags ground/flying/water. Entities can be in multiple tags, and they should be as simple to organise as possible, otherwise modders just won't bother.

Tags like "creepers" is unnecessary, there is only 1 creeper in vanilla. If another mod adds more creepers, they can add the tag (and modders should be encouraged to create tags to query, so that modpack authors / other mods can easily add compatibility.
But in my opinion, most of these tags should not be in forge.

In addition, the "enemies/animals" tag (and other similar tags) should not exist, you are already adding entities to "enemies", and "animals". Expecting people to make sure they also add it to compound tags is crazy. Just find the entities with both of the requested tags.

Thutmose:
Latin names or english names are fine IMO, so long as there is a standard defined here (ie to prevent a mix of some mods using latin, others using english)

PaintNinja:
Closing as stale, please open a new PR if you'd like to continue this.

PR #8445, opened 2022-02-03
Add Forge default Fluid Physic tag

MrScautHD:
This is a PR that add the Tag ```"forge:fluid_physics"``` if you add your fluid to it, It has the Water Physic (so can it not anymore brake recpies or a other things that you need for fluid Tag checks)

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/Entity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/Entity.java.patch
@@ -1,6 +1,10 @@
 --- a/net/minecraft/world/entity/Entity.java
 +++ b/net/minecraft/world/entity/Entity.java
-@@ -114,7 +_,7 @@
+@@ -111,10 +_,11 @@
+ import net.minecraft.world.phys.shapes.VoxelShape;
+ import net.minecraft.world.scores.PlayerTeam;
+ import net.minecraft.world.scores.Team;
++import net.minecraftforge.common.Tags;
  import org.apache.logging.log4j.LogManager;
  import org.apache.logging.log4j.Logger;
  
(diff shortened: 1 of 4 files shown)


XFactHD:
You may want to explain what this is supposed to achieve because at the moment this PR only adds a patch that most likely shouldn't be a patch (the tag should be in the `Tags.Fluids` class - nothing is stopping you from creating that class if it doesn't exist) and would also crash with "Tag used before it was bound" when the tag is used because there is no json file associated with this tag.

MrScautHD:
yea its not done

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -1,6 +1,10 @@
 --- a/net/minecraft/world/entity/Entity.java
 +++ b/net/minecraft/world/entity/Entity.java
-@@ -114,7 +_,7 @@
+@@ -111,10 +_,11 @@
+ import net.minecraft.world.phys.shapes.VoxelShape;
+ import net.minecraft.world.scores.PlayerTeam;
+ import net.minecraft.world.scores.Team;
++import net.minecraftforge.common.Tags;
```
TheCurle, line 7: Don't add imports to vanilla classes.

src/generated/resources/data/forge/tags/fluids/default_fluid_physic.json
```diff
@@ -0,0 +1,5 @@
+{
```
TheCurle, line 1: Why is this file here without a data generator? It'll be deleted the next time you run datagen.

src/main/java/net/minecraftforge/common/Tags.java
```diff
@@ -421,6 +421,7 @@ private static void init(){}
         private static void init() {}
 
         public static final IOptionalNamedTag<Fluid> MILK = tag("milk");
+        public static final IOptionalNamedTag<Fluid> DEFAULT_FLUID_PHYSIC = tag("default_fluid_physic");
```
TheCurle, line 424: Choose a better name. Default means nothing, and physic isn't a word.

TheCurle (changes requested):
I have no idea what you're trying to accomplish here, but this isn't the way to do it.

src/generated/resources/data/forge/tags/fluids/default_fluid_physic.json
MrScautHD: how can i gen it?

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -125,6 +129,33 @@
...
++      } else if (this.m_19943_(FluidTags.f_13131_, 0.014D) || this.m_19943_(Tags.Fluids.DEFAULT_FLUID_PHYSIC, 0.014D)) {
+          if (!this.f_19798_ && !this.f_19803_) {
+             this.m_5841_();
+          }
+@@ -1073,7 +_,7 @@
+    }
+ 
+    private void m_20323_() {
+-      this.f_19800_ = this.m_19941_(FluidTags.f_13131_);
++      this.f_19800_ = this.m_19941_(FluidTags.f_13131_) || this.m_19941_(Tags.Fluids.DEFAULT_FLUID_PHYSIC);
```
ChampionAsh5357, line 155: Just as a heads up, you need about 10 more patches to get the minimum necessary to enable proper fluid physics for the player alone. Tag determining physics here is also not that great since it tends to make custom fluid logic difficult to achieve. If you would like to help contribute for this, you can work on making #7992 without any breaking changes since Simon nor I have the time to work on it consistently.

From just looking at the patches, I'm pretty sure this will just keep you in the swimming animation but won't start it in the fluid, do some falling logic without any proper swimming, and desync the eyes in water fluid thus breaking vanilla compatibility.

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
MrScautHD: oh, is fixed by the next pull

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
MrScautHD: ik because the patches i has worked already on it but if you has already a better way then is it better to close my one

PR #8454, opened 2022-02-07
Add 11

oscie57:
Windows 11 is released and kind of popular now. This file stated it went up to Windows 10, when it infact supports Windows 11. To act on that, I simply added an `11` so people reading also understand that.


SizableShrimp:
Since this is simply a comment change, it would be better to target this towards an omnibus PR or lumping it in with something that actually changes the functionality of Forge in some way. Otherwise, it just becomes wasted processing and disk space as a result of every commit being built. For further discussion on this topic, please consult the [Forge discord](https://discord.gg/forge).

PR #8463, opened 2022-02-09
Add missing Locale parameter to String.format calls

diesieben07:
For example on `ar_SA` `1` is formatted as `١`, which is not desired when creating a path or ResourceLocation.

This becomes apparent when wearing armor with a locale of `ar_SA`, which will immediately crash the game if the armor model is on screen.

### DIFF
--- a/src/test/java/net/minecraftforge/debug/world/BiomeLoadingEventTest.java
+++ b/src/test/java/net/minecraftforge/debug/world/BiomeLoadingEventTest.java
@@ -40,6 +40,6 @@ public class BiomeLoadingEventTest {
     }
     public void onBiomeLoading(BiomeLoadingEvent event){
         ResourceLocation biome = event.getName();
-        LOGGER.info(String.format(Locale.ENGLISH, "Biome loaded: %s", biome.toString()));
+        LOGGER.info("Biome loaded: {}", biome);
     }
 }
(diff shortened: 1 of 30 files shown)


src/test/java/net/minecraftforge/debug/world/BiomeLoadingEventTest.java
```diff
@@ -38,6 +40,6 @@ public BiomeLoadingEventTest(){
     }
     public void onBiomeLoading(BiomeLoadingEvent event){
         ResourceLocation biome = event.getName();
-        LOGGER.info(String.format("Biome loaded: %s", biome.toString()));
+        LOGGER.info(String.format(Locale.ENGLISH, "Biome loaded: %s", biome.toString()));
```
sciwhiz12: This use of `String.format` should be removed in favor of using logger parameter substitution:
```suggestion
        LOGGER.info("Biome loaded: %s", biome.toString());
```

sciwhiz12 (changes requested):
Looks fine to me, and confirmed via test that it fixes the crash when wearing armor while on the `ar_SA` locale.

There are three instances in the FML subprojects which `String.format` is used but are not touched by this PR. Should these be changed as well?
- https://github.com/MinecraftForge/MinecraftForge/blob/be5446b91c17a3dbed880028741c430610bd59f1/fmlcore/src/main/java/net/minecraftforge/fml/ModList.java#L88
- https://github.com/MinecraftForge/MinecraftForge/blob/be5446b91c17a3dbed880028741c430610bd59f1/fmlcore/src/main/java/net/minecraftforge/fml/config/ModConfig.java#L57
- https://github.com/MinecraftForge/MinecraftForge/blob/be5446b91c17a3dbed880028741c430610bd59f1/fmlloader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/InvalidModFileException.java#L30

sciwhiz12 (approved):

PR #8468, opened 2022-02-18
[1.16.x] Prevent setting window icon on OSX

mezz:
When forcing LWJGL to version 3.3.0 in order to support M1 Macs, setting a window icon crashes the game.
This has been fixed in Minecraft 1.18 but still happens in 1.16, so this is sort of a weird backport fix!


TheCurle:
A lot of things crash the game on M1 with LWJGL 3.3, to my knowledge. Is this the only thing that breaks? I recall something about minimised windows breaking on some systems.

mezz:
You also need `java-objc-bridge:1.1 ` in order to support window maximizing, but everything else seems to work fine. I’ve been using it in dev on 1.18 for a while now.

PR #8469, opened 2022-02-18
Add ability to modify data gen output location

Yamahari:
add ability to modify data gen output location
AT is there for making overriding RecipeProvider#run possible with less copy&paste

### DIFF
--- a/patches/minecraft/net/minecraft/data/advancements/AdvancementProvider.java.patch
+++ b/patches/minecraft/net/minecraft/data/advancements/AdvancementProvider.java.patch
@@ -19,7 +19,7 @@
     public void m_6865_(HashCache p_123969_) {
        Path path = this.f_123962_.m_123916_();
        Set<ResourceLocation> set = Sets.newHashSet();
-@@ -45,6 +_,16 @@
+@@ -45,13 +_,23 @@
           }
        };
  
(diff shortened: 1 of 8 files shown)


ChampionAsh5357:
There are three things that come to mind for this PR that I think you need to address.

1. Why can't you just modify the output location of the generated entries within the FG args in the run block?
2. Why can you just make the registry path of the entry have a slash in it which ultimately does this?
3. If the above two answers are valid, why haven't you done this for **ALL** data providers currently in vanilla and Forge?

There are plenty of ways to implement this without this PR, so I'm unsure what's the reasoning here.

Yamahari:
1. I would have 30 different runs at the current state, only more to come x)
2. because that still doesn't allow me to have a folder "above" data/asset (custom pack)
3. for tags there's already a getPath function, what other providers are that I missed? I would have everything I need with this

ChampionAsh5357:
1. I'm pretty sure you can create a task to run other tasks in gradle.
3. If you're using this for a custom pack, probably all the ones listed on the [docs](https://mcforge.readthedocs.io/en/1.18.x/datagen/intro/).

Yamahari:
I think starting new instance of MC 30 times would take me a long time, 1 run takes like a minute or two to finish.

Yamahari:
I will take a look at the other providers though.

Yamahari:
added the other providers

Yamahari:
something similar can be archieved by creating new data generators at runtime like
```    
private static GatherDataEvent.DataGeneratorConfig getConfig(final GatherDataEvent event) {
   return ObfuscationReflectionHelper.getPrivateValue(GatherDataEvent.class, event, "config");
}

private static DataGenerator makeGenerator(final GatherDataEvent.DataGeneratorConfig config, final String root) {
    return config.makeGenerator(outputPath -> outputPath.resolve(root), true);
}
```
works for my case, but may not be suitable for others (?)

PR #8470, opened 2022-02-18
Provide access to the haveTime supplier in WorldTickEvent and ServerTickEvent

josephtarbit:
Allows modders access to the haveTime supplier in server tick events.


src/main/java/net/minecraftforge/event/TickEvent.java
```diff
@@ -45,9 +48,18 @@ public TickEvent(Type type, LogicalSide side, Phase phase)
...
+
+        private final BooleanSupplier haveTime;
+
+        public ServerTickEvent(Phase phase, BooleanSupplier haveTime)
         {
             super(Type.SERVER, LogicalSide.SERVER, phase);
+            this.haveTime = haveTime;
+        }
+
+        public boolean haveTime()
```
sciwhiz12, line 62: I think this and its sister method in `WorldTickEvent` will benefit from having a small explanatory javadoc about what this is and possibly what it is useful for.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -856,14 +857,14 @@ public static void onPlayerPostTick(Player player)
...
-        MinecraftForge.EVENT_BUS.post(new TickEvent.WorldTickEvent(LogicalSide.SERVER, TickEvent.Phase.START, world));
+        MinecraftForge.EVENT_BUS.post(new TickEvent.WorldTickEvent(LogicalSide.SERVER, TickEvent.Phase.START, world, haveTime));
     }
 
-    public static void onPostWorldTick(Level world)
+    public static void onPostWorldTick(Level world, BooleanSupplier haveTime)
     {
-        MinecraftForge.EVENT_BUS.post(new TickEvent.WorldTickEvent(LogicalSide.SERVER, TickEvent.Phase.END, world));
+        MinecraftForge.EVENT_BUS.post(new TickEvent.WorldTickEvent(LogicalSide.SERVER, TickEvent.Phase.END, world, haveTime));
     }
```
sciwhiz12: As a note for other reviewers, this hardcoded passing in of `LogicalSide.SERVER` to the world tick event seems to have been done all the way back in 95f0447881997533e5d352d96704e222bd5c36be when the tick events were first refactored to use the event system. So while this _seems_ to imply that the tick event may be fired on the logical client, it has never been fired on the logical client.

sciwhiz12 (changes requested):
For the record, could you please state the use case for which you need access to the value given by this supplier?

---

Please update the license headers of the following modified files by changing the year of 2021 to the year 2022:

 - `src/main/java/net/minecraftforge/event/ForgeEventFactory.java`
 - `src/main/java/net/minecraftforge/event/TickEvent.java`

josephtarbit:
> For the record, could you please state the use case for which you need access to the value given by this supplier?

It's useful for checking if there's enough time to perform non-essential/additional tasks during the tick loop, specifically IO related ones.  I'm using it for an implementation of `net.minecraft.world.level.chunk.storage.SectionStorage` which requires it to be passed into it's tick method as it uses it to check if there's enough time to save dirty chunks to the disk.

src/main/java/net/minecraftforge/event/TickEvent.java
```diff
@@ -45,9 +48,24 @@ public TickEvent(Type type, LogicalSide side, Phase phase)
...
             super(Type.SERVER, LogicalSide.SERVER, phase);
+            this.haveTime = haveTime;
+        }
+
+        /**
+         * Returns whether the server has enough time to perform any
+         * additional tasks (usually IO related) during the current tick.
+         * 
+         * @return <code>true</code> if there is time, <code>false</code> otherwise
+         */
```
sciwhiz12: We could simplify it a bit with the use of the inline `{@return}` tag:
```suggestion
        /**
         * {@return {@code true} whether the server has enough time to perform any additional
         * tasks (usually IO related) during the current tick, otherwise {@code false}}
         */
```

src/main/java/net/minecraftforge/event/TickEvent.java
```diff
@@ -60,10 +78,20 @@ public ClientTickEvent(Phase phase)
...
+        public WorldTickEvent(LogicalSide side, Phase phase, Level world, BooleanSupplier haveTime)
         {
             super(Type.WORLD, side, phase);
             this.world = world;
+            this.haveTime = haveTime;
+        }
+
+        /**
+         * See {@link ServerTickEvent#haveTime()}.
+         */
```
sciwhiz12: This should copy the javadoc from that method and use a `@see` tag to link the two.

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/event/TickEvent.java
josephtarbit: It would seem Eclipse doesn't know what that is...

src/main/java/net/minecraftforge/event/TickEvent.java
```diff
@@ -60,10 +78,25 @@ public ClientTickEvent(Phase phase)
...
+            this.haveTime = haveTime;
+        }
+
+        /**
+         * Returns whether the server has enough time to perform any
+         * additional tasks (usually IO related) during the current tick.
+         * 
+         * @return <code>true</code> if there is time, <code>false</code> otherwise
+         * 
+         * @see {@link ServerTickEvent#haveTime()}.
```
sciwhiz12: No need for that `{@link}` tag usage:
```suggestion
         * @see ServerTickEvent#haveTime()
```

src/main/java/net/minecraftforge/event/TickEvent.java
sciwhiz12: Hmm, then Eclipse may not yet have updated itself with the new tag which was introduced in Java 16 ([JDK-8075778](https://bugs.openjdk.java.net/browse/JDK-8075778)). Remove the outermost braces to make it a block tag for now.

sciwhiz12 (approved):

sciwhiz12 (changes requested):
Please add appropriate overloads for the methods and constructor and deprecate the existing overloads, to avoid the breaking changes in binary compatibility.

sciwhiz12 (changes requested):
Please add `since = "1.18.1"` to the `@Deprecated` annotations, as well as accompanying javadocs for these deprecated methods with a `@deprecated` tag, akin to the following:
```java
/**
 * @deprecated To be removed in 1.19. Use {@link #newMethodHere(Level, BooleanSupplier)} instead.
 */
```

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -856,14 +857,36 @@ public static void onPlayerPostTick(Player player)
         MinecraftForge.EVENT_BUS.post(new TickEvent.PlayerTickEvent(TickEvent.Phase.END, player));
     }
 
+    /**
+     * TODO: Remove in 1.19
+     * 
+     * @deprecated Use {@link #onPreWorldTick(Level, BooleanSupplier)}
+     */
+    @Deprecated(forRemoval = true)
```
sciwhiz12: This annotation is missing `since = "1.18.1"`.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -856,14 +857,36 @@ public static void onPlayerPostTick(Player player)
...
+    {
+        MinecraftForge.EVENT_BUS.post(new TickEvent.WorldTickEvent(LogicalSide.SERVER, TickEvent.Phase.START, world, haveTime));
+    }
+
+    /**
+     * TODO: Remove in 1.19
+     * 
+     * @deprecated Use {@link #onPostWorldTick(Level, BooleanSupplier)}
+     */
+    @Deprecated(forRemoval = true)
```
sciwhiz12: This annotation is missing `since = "1.18.1"`.

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -876,13 +899,35 @@ public static void onPostClientTick()
...
+    {
+        MinecraftForge.EVENT_BUS.post(new TickEvent.ServerTickEvent(TickEvent.Phase.START, haveTime));
+    }
+
+    /**
+     * TODO: Remove in 1.19
+     * 
+     * @deprecated Use {@link #onPostServerTick(BooleanSupplier)}
+     */
+    @Deprecated(forRemoval = true)
```
sciwhiz12: This annotation is missing `since = "1.18.1"`.

sciwhiz12 (changes requested):

sciwhiz12 (approved):

sciwhiz12 (changes requested):
- `src/main/java/net/minecraftforge/event/ForgeEventFactory.java`
- `src/main/java/net/minecraftforge/event/TickEvent.java`

sciwhiz12 (approved):

PR #8471, opened 2022-02-18
Add missing shear and elytra game events

pupnewfster:
While testing another PR I am writing I noticed that forge's redirecting of shearing entities to `IForgeShearable#onSheared` had not been updated to also fire vanilla's `SHEAR` GameEvent. This PR just adds those missing calls in.


marchermans (approved):
Code looks good and is working, not sure if these are all of them though.

PR #8472, opened 2022-02-18
Add VanillaGameEvent to allow for globally listening to vanilla's GameEvents

pupnewfster:
This PR adds an event to allow mods to globally listen to vanilla's GameEvents. My use case for this PR is specific to having an item that depending on what upgrades are installed can act as shears in exchange for energy. Thanks to the changes in #7997 I can more easily support this without having to recreate all of vanilla's interactions with the one exception of applying the energy cost to actions like carving a pumpkin.

The first solution I thought of was to have some sort of way to tell a stack when it was used for a specific action (specifically in my case shears), but as the majority of action types don't have a clear "was used as" spot that wouldn't work well. The alternate solution I thought of which led to this PR was to use vanilla's `GameEvent` system and listeners to listen for each player and if they were the one who caused a shear event to do a bit of validation and then use the energy that way. The issue with this is that vanilla's `GameEventListener` system is strongly tied to position of the listener (specifically `y` position) which wouldn't work well for how much the player moves.

This PR aims to solve that issue and provide modders with easy access to being able to listen to vanilla's game events by exposing all fired game events as a forge event.

One slight modification to this PR that I can make if people think it would be of use is to:
- Patch `LevelAccessor#gameEvent(GameEvent, Entity)` to call `LevelAccessor#gameEvent(Entity, GameEvent, Entity)`
- Override `LevelAccessor#gameEvent(Entity, GameEvent, Entity)` in `ServerLevel` to directly call `Level#postGameEventInRadius(Entity, GameEvent, BlockPos, int)` (instead of calling that from `ServerLevel#gameEvent(Entity, GameEvent, BlockPos)` via a couple redirects)
- Add a different event, a subclass of the `VanillaGameEvent` event that provides access to the last entity param (which would be the "target entity")

I initially have held off on this second modification as it is *slightly* more maintenance and I am unsure how much use it would be (and can easily be done later without being a breaking change).

One final thing to note about the test mod this PR is adding is that without #8471 it doesn't actually fire like the comment says it does for the case of a player shearing a sheep.

### DIFF
--- a/patches/minecraft/net/minecraft/server/level/ServerLevel.java.patch
+++ b/patches/minecraft/net/minecraft/server/level/ServerLevel.java.patch
@@ -198,14 +198,14 @@
        this.f_8548_.m_6846_().m_11241_(p_8689_, p_8690_.m_20185_(), p_8690_.m_20186_(), p_8690_.m_20189_(), p_8693_ > 1.0F ? (double)(16.0F * p_8693_) : 16.0D, this.m_46472_(), new ClientboundSoundEntityPacket(p_8691_, p_8692_, p_8690_, p_8693_, p_8694_));
     }
  
-@@ -838,6 +_,7 @@
- 
-    public void m_142346_(@Nullable Entity p_143268_, GameEvent p_143269_, BlockPos p_143270_) {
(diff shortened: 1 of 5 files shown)


marchermans (approved):
I don't think it needs further extension.

PR #8473, opened 2022-02-18
Provide damage source context to Item#onDestroyed(ItemEntity)

pupnewfster:
Relatively straightforward PR that provides access to the damage source as part of the context for `Item#onDestroyed`. My use case is that I am planning on making my items able to drop their contents if they are destroyed (similar to shulker boxes), but only if they are destroyed when they don't have any security mode set OR if the player that caused the destruction has "ownership" of the item via our security system. To achieve this I need to be able to access the player (or entity) via `DamageSource#getEntity`


marchermans (approved):
Yep simple, and works.
Looks good.

PR #8475, opened 2022-02-19
[1.18.x] Add Attribute to prevent EnderMan angering

Minecraftschurli:
This PR is an alternative to #8406 using an attribute instead of an event as requested by @diesieben07 on discord.

It adds an attribute called `ender_masking` (name open for discussion) which prevents endermen from angering if the value is greater than 0 (which is default).

This PR also adds a test mod for it adding an effect that modifies the attribute to add 1.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/monster/EnderMan.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/monster/EnderMan.java.patch
@@ -21,7 +21,7 @@
     boolean m_32534_(Player p_32535_) {
        ItemStack itemstack = p_32535_.m_150109_().f_35975_.get(3);
 -      if (itemstack.m_150930_(Blocks.f_50143_.m_5456_())) {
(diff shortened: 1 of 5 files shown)


Shadows-of-Fire:
I don't think an attribute is the proper solution here.  Attributes are not really meant to represent boolean values, and should be understandable by the player since they can show up as modifiers on items.  If a player saw something that said 
`+0.01 Ender Masking` they would likely be confused.

Minecraftschurli:
It was requested that I move this from an event to an attribute (It should decrease the agro distance but the vanilla calculation is weird)

TheCurle:
As mentioned by Shadows, attributes are scalar values - having a value of "> 0" trigger some behavior is unintuitive at best.

I'm not sure what the best route here is. There's some potential in having a scalar value, but it shouldn't represent a boolean "does or does not anger an enderman". 

I'm going to close this for the mean time and we can figure out the best route on your other PR.

PR #8476, opened 2022-02-22
Avoid exclusive synchronisation in block render type lookups

Gegy:
This removes the exclusive synchronisation introduced by Forge for looking up render types, which limits concurrency in chunk building. This is instead replaced by a check during writing to the render type map that ensures writes can only occur during client mod loading, avoiding any case of a race condition with no overhead for lookups.

This should increase the efficiency of chunk building, especially when many worker threads are allocated.

Writes to the render type map are still synchronised to avoid concurrent writes due to parallel mod loading.

### DIFF
--- a/patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
@@ -42,7 +42,7 @@
           if (!Minecraft.m_91085_()) {
              return Sheets.m_110792_();
           } else {
-@@ -340,9 +_,86 @@
+@@ -340,9 +_,75 @@
        }
     }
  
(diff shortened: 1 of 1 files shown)


patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
```diff
@@ -54,50 +54,78 @@
...
++      return canRenderInLayer(type, fluidRenderChecks, fluid.m_76152_().delegate);
++   }
++
++   private static <T> boolean canRenderInLayer(RenderType type, Map<T, java.util.function.Predicate<RenderType>> map, T key) {
++      java.util.concurrent.locks.StampedLock lock = RENDER_CHECK_LOCK;
++
++      // attempt optimistic read to the render type predicate without blocking
++      long stamp = lock.tryOptimisticRead();
++      if (stamp != 0) {
++         java.util.function.Predicate<RenderType> predicate = map.get(key);
```
Technici4n: Does this not risk crashing if the map is being modified concurrently?

Technici4n (changes requested):
Due to the issue mentioned below, I would recommend using a read-write lock instead and check if that is fast enough before going for a fancier optimization (e.g. copy-on-write).

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
Gegy: A get query should not throw an exception with concurrent access (at least, with `Object2ObjectOpenHashMap`)- but it might return invalid data, though this is protected by the stamp validation.
From previous testing, though it was a while ago, acquiring a read lock instead of performing an optimistic read was significantly more expensive, mainly due to having to acquire and mutate the lock state in such hot code. Using a copy-on-write map might be a much more elegant solution, though! Would this be preferred? 🙂

Technici4n: > A get query should not throw an exception with concurrent access

Afaik, all bets are off when it comes to concurrent access with fastutil maps. :P

The fancy solution would be to have a `volatile boolean modified` that is set after each modification, and have read synchronize then copy the map if the flag is set. Possibly a bit overkill. :P

Gegy: It's definitely not ideal to rely on fastutil internals to not throw any exceptions. 😄 
Might it make sense to re-assign the map right away after a render type is set, just to avoid any overhead in the lookup logic, even if it means setting into the map is a bit more expensive?

XFactHD:
I think this could be done in a far simpler way.

There should be no need to check for concurrent modification of the maps as `setRenderLayer()` should only be called in the startup process. Anything else would be a bug in the mod calling `setRenderLayer()` after startup.

This rule plus the assumption from the `HashMap` javadoc that `HashMap#get()` is safe to call concurrently when you are guaranteed to have no write accesses would allow simplifying this PR to two simple changes:
1. Remove the `synchronized` blocks from the `canRenderInLayer()` methods without switching the type of map that is used
2. If you want to be safe, add a check in `setRenderLayer()` to throw an exception when it is called after startup (`ClientModLoader.isLoading()` could be used for this check).

Gegy:
@XFactHD This would definitely be the ideal solution here. 🙂 I went instead with these solutions to make sure to not create any breaking changes, but is such a breaking change acceptable? Or would this be something to delay to a breaking release?

In the case that such a breaking change *is* unacceptable, it may still be worth introducing a more complex patch for now, as the synchronisation can have a noticeable impact to chunk building performance.

XFactHD:
IMO it's not a breaking change because a mod doing that has a serious bug but I am not the one to decide that.
The Forge team needs to decide whether this would classify as a breaking change.

TheCurle:
As long as a mod compiled before this change can still run after this change - so-called binary compatibility, then it is not a binary breaking change, and we can accept it with solid enough reasoning - of which exists already.

Basically, make sure this won't make a mod crash for being compiled before these changes and you're good.

PR #7702, opened 2021-03-27
Added ItemStack sensitive version of getFoodProperties

queekusme:
Fixes #7701 

This PR adds an ItemStack sensitive version of `getFoodProperties`.
This allows custom Items to override the `Food` class data so that the output can be modified before the data is used.

- Existing calls have been updated to use the ItemStack sensitive version.
- The new method calls the existing method by default so that existing behaviour can be retained.
- The existing method has been left unchanged except for a Deprecation annotation with a comment indicating to use the ItemStack sensitive version of the function.

PR #8477, opened 2022-02-23
Implement ItemStack and LivingEntity sensitive method to get the FoodProperties for an Item

canitzp:
This pr is a new version for #7702 and #7543.
Both pr got as reply that an stack transformer is added (#7638), but to this day this hasn't happened.

With this approach the stack transformer wouldn't work, since the new method is not only sensitive to ItemStack, but also to LivingEntity. This can be useful to change the food properties for eating entities, currently wolves and cats.

All calls to Item#getFoodProperties() are changed to use the newer method, which also includes changing FoodData#eat(Item, ItemStack). This was done by adding a new eat method, which also includes the LivingEntity as third parameter, but it can be null. This Nullable was added so that Item#getUseDuration(ItemStack) doesn't need an LivingEntity sensitive method too.

It is important to note here that this pr doesn't change the behavior of Item#isEditable(), since this would involve much more methods to be changed and LexManos explicitly didn't want a item to be both, food and not food (https://github.com/MinecraftForge/MinecraftForge/pull/7543#issuecomment-767257386)

**Edit:** There would be a advantage in changing Item#isEditable() too, because Item#getFoodProperties() is annotated with Nullabe, but the vanilla codes never actually checks for null on the returned value, instead it relies on isEditable() on being false. For Vanilla this works, but when someone overwrites isEditable() to be true, but returns null in getFoodProperties() (which should be possible due to the annotation), the game would crash with an NPE at some point.


patches/minecraft/net/minecraft/world/food/FoodData.java.patch
```diff
@@ -0,0 +1,20 @@
...
+    }
+ 
++   @Deprecated
+    public void m_38712_(Item p_38713_, ItemStack p_38714_) {
+-      if (p_38713_.m_41472_()) {
+-         FoodProperties foodproperties = p_38713_.m_41473_();
++      this.eat(p_38713_, p_38714_, null);
++   }
++
++   public void eat(Item item, ItemStack stack, @javax.annotation.Nullable net.minecraft.world.entity.LivingEntity entity)
```
pupnewfster: I feel like this overloaded eat method should have the item and stack use the srg names and then just be added to forge's exc file so that they can be properly mapped. That way if more things that use them get added later in the method at some point they don't need to be replaced via patch to reference these names.

patches/minecraft/net/minecraft/world/food/FoodData.java.patch
```diff
@@ -0,0 +1,20 @@
+--- a/net/minecraft/world/food/FoodData.java
++++ b/net/minecraft/world/food/FoodData.java
+@@ -24,9 +_,15 @@
+       this.f_38697_ = Math.min(this.f_38697_ + (float)p_38708_ * p_38709_ * 2.0F, (float)this.f_38696_);
+    }
+ 
++   @Deprecated
```
pupnewfster, line 8: This should have a comment/javadoc saying to use the more sensitive version in addition to the deprecated annotation.

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -764,4 +765,17 @@ default int getDefaultTooltipHideFlags(@Nonnull ItemStack stack)
...
+     * Get the food properties for this item.
+     * Use this instead of the {@link Item#getFoodProperties()} method, for ItemStack sensitivity.
+     *
+     * @param stack The ItemStack the entity wants to eat.
+     * @param entity The entity which wants to eat the food. Be aware that this can be null!
+     * @return The current FoodProperties for the item.
+     */
+    default FoodProperties getFoodProperties(ItemStack stack, @Nullable LivingEntity entity)
+    {
+        return self().getFoodProperties();
```
pupnewfster, line 800: The base food properties method should be deprecated with a comment/javadoc saying to use this more sensitive version.

patches/minecraft/net/minecraft/world/food/FoodData.java.patch
```diff
@@ -0,0 +1,20 @@
...
+    public void m_38712_(Item p_38713_, ItemStack p_38714_) {
+-      if (p_38713_.m_41472_()) {
+-         FoodProperties foodproperties = p_38713_.m_41473_();
++      this.eat(p_38713_, p_38714_, null);
++   }
++
++   public void eat(Item item, ItemStack stack, @javax.annotation.Nullable net.minecraft.world.entity.LivingEntity entity)
++   {
```
pupnewfster: Bracket should be on previous line as this is part of a patch.

src/main/resources/forge.exc
```diff
@@ -51,4 +51,5 @@ net/minecraft/world/level/chunk/PalettedContainer.setBits(IZ)V=|p_63122_,forceBi
...
-net/minecraft/world/level/storage/loot/LootPool.<init>([Lnet/minecraft/world/level/storage/loot/entries/LootPoolEntryContainer;[Lnet/minecraft/world/level/storage/loot/predicates/LootItemCondition;[Lnet/minecraft/world/level/storage/loot/functions/LootItemFunction;Lnet/minecraft/world/level/storage/loot/providers/number/NumberProvider;Lnet/minecraft/world/level/storage/loot/providers/number/NumberProvider;Ljava/lang/String;)V=|p_165128_,p_165129_,p_165130_,p_165131_,p_165132_,name
\ No newline at end of file
+net/minecraft/world/level/storage/loot/LootPool.<init>([Lnet/minecraft/world/level/storage/loot/entries/LootPoolEntryContainer;[Lnet/minecraft/world/level/storage/loot/predicates/LootItemCondition;[Lnet/minecraft/world/level/storage/loot/functions/LootItemFunction;Lnet/minecraft/world/level/storage/loot/providers/number/NumberProvider;Lnet/minecraft/world/level/storage/loot/providers/number/NumberProvider;Ljava/lang/String;)V=|p_165128_,p_165129_,p_165130_,p_165131_,p_165132_,name
+net/minecraft/world/food/FoodData.eat(Lnet/minecraft/world/item/Item;Lnet/minecraft/world/item/ItemStack;Lnet/minecraft/world/entity/LivingEntity;)V=|p_38713_,p_38714_,entity
```
pupnewfster: We try to keep this file in more or less alphabetical order (and which means `world/food` should be before `world/level` (and before `world/item`

canitzp: Oh ok, I'll change it accordingly

src/main/resources/forge.exc
```diff
@@ -43,6 +43,7 @@ net/minecraft/world/item/ItemStack.onItemUse(Lnet/minecraft/world/item/context/U
 net/minecraft/world/item/ItemStack.onItemUseFirst(Lnet/minecraft/world/item/context/UseOnContext;)Lnet/minecraft/world/InteractionResult;=|p_41662_
 net/minecraft/world/item/MapItem.getCustomMapData(Lnet/minecraft/world/item/ItemStack;Lnet/minecraft/world/level/Level;)Lnet/minecraft/world/level/saveddata/maps/MapItemSavedData;=|p_42910_,p_42911_
 net/minecraft/world/item/alchemy/PotionBrewing$Mix.<init>(Lnet/minecraftforge/registries/ForgeRegistryEntry;Lnet/minecraft/world/item/crafting/Ingredient;Lnet/minecraftforge/registries/ForgeRegistryEntry;)V=|p_43536_,p_43537_,p_43538_
+net/minecraft/world/food/FoodData.eat(Lnet/minecraft/world/item/Item;Lnet/minecraft/world/item/ItemStack;Lnet/minecraft/world/entity/LivingEntity;)V=|p_38713_,p_38714_,entity
```
pupnewfster: As I said `f` also comes before `i`

canitzp: oh crap, I should probably learn the alphabet xd

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -764,4 +765,18 @@ default int getDefaultTooltipHideFlags(@Nonnull ItemStack stack)
...
 
+    /**
+     * Get the food properties for this item.
+     * Use this instead of the {@link Item#getFoodProperties()} method, for ItemStack sensitivity.
+     *
+     * @param stack The ItemStack the entity wants to eat.
+     * @param entity The entity which wants to eat the food. Be aware that this can be null!
+     * @return The current FoodProperties for the item.
+     */
+    @Nullable
```
pupnewfster: Are there any restrictions on when this should be nullable? After looking at the PR in more depth it seems there are a few places patched that if a modder returns null from here but marks their item as edible a null pointer will be thrown. It is probably worth specifying in the javadocs for this that it shouldn't return null if the item returns true for is edible. (At least assuming that is actually true)

canitzp: Yeah I wrote about exactly this in the description. With the Nullable annotation it is exactly like Vanilla, since this NPE already is there.

pupnewfster: That is why I believed what I was saying is correct that there are restrictions, so then that definitely should be mentioned here (in the javadocs) to make sure modders know they shouldn't just return null unless their item is marked as never being edible.

canitzp: For me the current question is, should I just remove the Nullable annotation? There is no need for it and the method isn't really Nullable anyways

pupnewfster: I feel like yes as the same as the method it wraps around, it just should have a clarification in the javadocs that it only should be null if isEdible is false.

Shadows-of-Fire:
Is this superseded in any way by #8405?

canitzp:
Well there are certainly similarities. Their system is more complex and offers more features, but even with their events, they only control player hunger. My pr is more focused on the eating for every living entity.

Shadows-of-Fire:
Please fix the merge conflicts on this PR so it can be reassigned.

canitzp:
Sorry for the delay. I haven't noticed the conflicts, but they are resolved now.

pupnewfster (changes requested):
After a bit of internal discussion it was pointed out that as this is a method adding item stack sensitivity, a bouncer should also be added to `IForgeItemStack` (like a bunch of the other methods in it) that make calling `getFoodProperties` easier as instead of calling `stack.getItem().getFoodProperties(stack, entity)` it will just be `stack.getFoodProperties(entity)`

canitzp:
@pupnewfster You are right, the bouncer is a great addition to use the method more conveniently, therefor it is now added and I also make use of it in one occasion.

pupnewfster:
Is there any reason to not make use of it in the other locations? As even if you already have the item reference it would make it a bit more readable as then you just switch that item reference to a stack reference vs adding the stack as a parameter.

canitzp:
Well I didn't want to change the base code more than needed and in this one location, it was very straightforward. If you want, than I also can implement these changes

LexManos:
Yes it is preferred to move things to calling the bouncer to make the code in the patches simpler. It will be the same number of changes, just changing item.getFoodProperties(stack, entity) to stack.getFoodProperties(entity)

canitzp:
So the requested changes were done. All callings of getFoodProperties now use the stack bouncer method

patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
```diff
@@ -549,6 +558,18 @@
     }
  
     public ItemStack m_5584_(Level p_21067_, ItemStack p_21068_) {
+@@ -3158,9 +_,8 @@
+    }
+ 
+    private void m_21063_(ItemStack p_21064_, Level p_21065_, LivingEntity p_21066_) {
+-      Item item = p_21064_.m_41720_();
+-      if (item.m_41472_()) {
```
pupnewfster: Don't change these two lines as that just makes it a larger patch, only change the call to `getFoodProperties`

patches/minecraft/net/minecraft/world/entity/animal/Wolf.java.patch
```diff
@@ -9,3 +18,13 @@
                 this.m_21828_(p_30412_);
                 this.f_21344_.m_26573_();
                 this.m_6710_((LivingEntity)null);
+@@ -400,8 +_,7 @@
+    }
+ 
+    public boolean m_6898_(ItemStack p_30440_) {
+-      Item item = p_30440_.m_41720_();
+-      return item.m_41472_() && item.m_41473_().m_38746_();
```
pupnewfster: Don't change the first line or the `item#isEdible` call as that just makes it a larger patch, only change the call to `getFoodProperties`

canitzp:
@pupnewfster done

PR #8478, opened 2022-02-24
Add Attack Range attribute and Update Reach Distance

Shadows-of-Fire:
The way that Reach Distance has interacted with entities has always been strange, to say the least, and there have been a couple previous PR's that tried to address this - https://github.com/MinecraftForge/MinecraftForge/pull/7808 https://github.com/MinecraftForge/MinecraftForge/pull/7807 https://github.com/MinecraftForge/MinecraftForge/pull/4736

This PR follows the spirit of those ones, by adding a new Attribute - `Attack Range` - which represents the range (in blocks) at which a player may attack an entity.  This attribute carries a default of 3 (same as vanilla) and has various places where it is increased by 3 for creative players.  All range checks relating to attack range now redirect to this attribute.

However, this PR also makes some consistency updates to `Reach Distance`.  The default value has been changed from `5.0` to `4.5` which is the proper default for survival players, and is increased by the respective constant values where necessary. This means effects that double reach distance will cause it to properly rise from 4.5 to 9.0, instead of 4.5 to 9.5.  Also, reach distance now impacts entity interactions, meaning an increase in reach distance will allow you to right-click entities from further away.

This PR does not add a test mod - this is by intention - nothing in this PR requires a mod to be tested.  You can generate an item with an attribute modifier on it and utilize that for testing.

Shears with Attack Range +2
`/give @p shears{AttributeModifiers:[{AttributeName:"forge:attack_range",Amount:2,Slot:mainhand,Name:"test",UUID:[I;-122124,24674,3193,-49348]}]} 1`
Shears with Reach Distance +2
`/give @p shears{AttributeModifiers:[{AttributeName:"forge:reach_distance",Amount:2,Slot:mainhand,Name:"test",UUID:[I;-122124,24674,3193,-49348]}]} 1`
The effects of these can be compared to default shears to see that the attributes work as intended.

TLDR: 
`Reach Distance` default value updated to 4.5 (proper survival value) from 5.0 (creative value).  Now increases distance you can right-click on entities from as well.
`Attack Range` added.  Responsible for how far you may attack an entity from.  Default value is 3 blocks.


patches/minecraft/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
```diff
@@ -53,7 +53,16 @@
...
-@@ -1222,6 +_,7 @@
+@@ -1216,12 +_,14 @@
+             return;
+          }
+ 
+-         double d0 = 36.0D;
+-         if (this.f_9743_.m_20280_(entity) < 36.0D) {
++         double maxRange = this.f_9743_.m_21133_(net.minecraftforge.common.ForgeMod.ATTACK_RANGE.get()) + 3; //Forge: Change constant value to max possible range. Equal to current value + 3 (because creative gives +3)
++         maxRange = Math.max(maxRange, this.f_9743_.m_21133_(net.minecraftforge.common.ForgeMod.REACH_DISTANCE.get()) + 0.5); //Used for both attacks and interacts, needs to be the highest of the two for sanity.
++         if (this.f_9743_.m_20280_(entity) < maxRange * maxRange) {
```
Elenterius: This solution won't work with entities that have a very large bounding box and/or players with reduced attack range.
The method distanceSqTo uses the **position** of both entities to calculate the distance which is problematic when the bounding box of the victim keeps the player out of reach of being able to make a attack that is "visually" in range.

For testing purposes you can reduce the attack range and summon a big slime (e.g. `/summon minecraft:slime ~ ~ ~ {Size: 15}`)
(note: don't forget to reduce the block reach as well -> see line 63)

Elenterius: distanceIssue

Elenterius:
Btw the way applying **multiply by zero attribute modifiers** on attack range and block reach doesn't properly work (in creative) since your adding constants on top without checking if the reach/range is not zero. Not sure if it is desirable/intended behavior that you are still able to interact with entities (in creative) even though the reach/range should be zero.

patches/minecraft/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
```diff
@@ -53,7 +53,16 @@
...
  
-@@ -1222,6 +_,7 @@
+@@ -1216,12 +_,14 @@
+             return;
+          }
+ 
+-         double d0 = 36.0D;
+-         if (this.f_9743_.m_20280_(entity) < 36.0D) {
++         double maxRange = this.f_9743_.m_21133_(net.minecraftforge.common.ForgeMod.ATTACK_RANGE.get()) + 3; //Forge: Change constant value to max possible range. Equal to current value + 3 (because creative gives +3)
++         maxRange = Math.max(maxRange, this.f_9743_.m_21133_(net.minecraftforge.common.ForgeMod.REACH_DISTANCE.get()) + 0.5); //Used for both attacks and interacts, needs to be the highest of the two for sanity.
```
Elenterius: Shouldn't you use ServerboundInteractPacket.Action to determine if it is an attack action and only then use the attack range as the max range? Because atm this tightly couples both attack range and block reach together, meaning to properly validate the attack range on the server side you have to change both. (else the client could potentially send attack packets from further away)

Also shouldn't you test first if the player is in creative before applying the +3?

Shadows-of-Fire:
The question of what to do with creative players is rather difficult, because the reach distance and attack range defaults are different for players in creative.  The question is should players in creative have their attributes scaled or increased by a constant...

The original implementation of reach distance was using a constant different (So setting to zero was unsupported behavior previously).  To maintain those values without adding constants, it would mean creative players always receive doubled attack range as compared to survival.

patches/minecraft/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
Shadows-of-Fire: Note that this is what vanilla does in the first place - so this issue (with arbitrarily large entities) exists there as well.  I'm not sure if it is documented on Mojang's end.

The only possible fix would be to create a vector from the player's eye position to somewhere on the entity (maybe the center) and ask the AABB for an intersection, then use the distance between that intersection point and the player's eye position.

I'm not sure what other problems that might lead to, however...

Jackiecrazy: The simplified version's what I've done for a while now, and it seems fine.
https://github.com/Jackiecrazy/WarDance/blob/51a0da44570073dfb683e6e925d139c74c8dd0a1/src/main/java/jackiecrazy/wardance/utils/GeneralUtils.java#L112-L123
I know with some certainty that Mojang does do some sort of special handling, because attempts to make the system more sane have inadvertently broken it. Vanilla accepts all attacks within 6 blocks or so, and the client is responsible for sending the target.

patches/minecraft/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
Shadows-of-Fire: The edit to this check was to mirror vanilla (it checks that the distance is within the absolute maximum possible).
I suppose further sanity might be desirable.

patches/minecraft/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
Elenterius: i could also imagine that vanilla accepts all attacks within 6 blocks in order to mitigate issues related to network lag. But who knows for sure, could just be legacy code.

Elenterius: > I'm not sure if it is documented on Mojang's end.

I couldn't find anything specific in the minecraft bug tracker related to attacking large bounding boxes.

Elenterius: > The only possible fix would be to create a vector from the player's eye position to somewhere on the entity (maybe the center) and ask the AABB for an intersection, then use the distance between that intersection point and the player's eye position.

I did something like this. Haven't heard any complaints yet.
https://github.com/Elenterius/Combat-Commons/blob/1.18.x/src/main/java/com/github/elenterius/combat_commons/utils/RayTraceUtil.java#L57-L75

patches/minecraft/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
Elenterius: Depends on how paranoid of clients we want to be and how network delay (in multiplayer) impacts this. I could imagine it will be harder to hit fast moving players/mobs with stricter range checks.

patches/minecraft/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
Shadows-of-Fire: Resolving as outdated - see https://github.com/MinecraftForge/MinecraftForge/pull/8478/files#diff-fad76e2aed5c02fd9756e9dda64030fb777b549c5fcbdfc56466af63a07efee6R67

patches/minecraft/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
Shadows-of-Fire: Checks are now done at a lower level - thus validation should be a bit better.

Shadows-of-Fire:
That last commit altered the underlying structure pretty heavily, but a lot of the concerns voiced should be addressed.  Zero-Modifiers should now work, as well.

src/main/java/net/minecraftforge/common/extensions/IForgePlayer.java
```diff
@@ -0,0 +1,88 @@
...
+     * @param entity The entity being range-checked.
+     * @param padding Extra validation distance.
+     * @return If the player can interact with the passed entity.
+     */
+    default boolean canInteractWith(Entity entity, double padding)
+    {
+        return isCloseEnough(entity, getReachDistance() + padding);
+    }
+
+    default boolean isCloseEnough(Entity entity, double dist)
```
sciwhiz12, line 74: Is this meant to be exposed as API? If not, this should be marked as `private`.

src/main/java/net/minecraftforge/common/extensions/IForgePlayer.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.util.Optional;
+
+import net.minecraft.core.BlockPos;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.phys.Vec3;
+import net.minecraftforge.common.ForgeMod;
+
+public interface IForgePlayer {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
public interface IForgePlayer
{
```

sciwhiz12 (changes requested):
Preliminary testing shows the PR currently has an error, where the player cannot place blocks and can only break blocks when very close to the block, and the block selection outline does not render (as if in adventure mode).

I will take a deeper look at the changes once this error (and any similar errors that may be occurring alongside it) is resolved.

src/main/java/net/minecraftforge/common/extensions/IForgePlayer.java
Shadows-of-Fire: Yeah, I meant to add a javadoc, it has use as a utility for other interaction checks, think I just forgot.

patches/minecraft/net/minecraft/server/level/ServerPlayerGameMode.java.patch
```diff
@@ -1,19 +1,17 @@
 --- a/net/minecraft/server/level/ServerPlayerGameMode.java
 +++ b/net/minecraft/server/level/ServerPlayerGameMode.java
-@@ -125,7 +_,15 @@
+@@ -125,7 +_,13 @@
        double d1 = this.f_9245_.m_20186_() - ((double)p_9282_.m_123342_() + 0.5D) + 1.5D;
```
sciwhiz12, line 4: Looks like there's a `1.5D` difference which needs to be accounted for here.

sciwhiz12: Ah, disregard this comment, I believe this is the eye height of the player, which is accounted for by the use of the eye position in `IForgePlayer#canInteractWith(BlockPos, double)` rather than the raw position of the player.

sciwhiz12 (commented):
The code looks fine to me, though admittedly I don't have much experience with entities. 

This will need more rigorous testing to ensure that this doesn't have any adverse or unwanted effects with the default reach distance and attack range. Particularly, these tests need to be done between a Forge server and vanilla clients, and between a vanilla server and Forge clients.

Shadows-of-Fire:
So, I have performed a series of exhaustive tests regarding how this all works and come up with the following results that will need to indicate changes on this PR:

The following PR bug was uncovered (this is a bug that would have been introduced by the PR, not an existing bug):
*  The Attack Range check was using player center to entity center hit checks `Entity#distanceToSqr(Entity)` instead of a proper eye-corner hit check, which is now in place https://github.com/MinecraftForge/MinecraftForge/pull/8478/commits/469be5a3b46b392aaef62d5bfcbf32169daf10f4#diff-fad76e2aed5c02fd9756e9dda64030fb777b549c5fcbdfc56466af63a07efee6L78-L80

The following (undocumented) *vanilla* bug was uncovered:
*  There is a check in `GameRenderer` that is `(this.minecraft.hitResult != null)`, but should really be `(this.minecraft.hitResult != null && this.minecraft.hitResult.getType() != HitResult.Type.MISS)`
    *  This check is used to clamp how far a valid entity trace result may be if a block trace has already been made, as to enforce that if an entity ray trace would be selected, the entity must be closer than the hit block.
    *  In Vanilla, this missing check clamps the max entity distance to `5.0`, regardless of the desired value.  In Forge, without this check, it effectively clamps the entity trace distance (Attack Range AND Reach Distance) to be at most the Reach Distance.
    *  The impact of this patch is low-reaching, the bug it fixes only has one actual issue, which is that the Creative Player Entity Trace Distance (which is both attack range and reach distance) is set to 6.0 and then clamped back down to 5.0, which seems unintentional.  There are no other side-effects.

The following side-effect of this PR was uncovered:
* In vanilla, the Entity Trace Distance (which represents both Attack Range and Entity Reach Distance for vanilla) is 3.0, and the [Block] Reach Distance is 4.5.  In this PR, the Attack Range is made to be 3.0 and the Reach Distance (both block and entity) is made to be 4.5,
    * In a Forge (S) to Forge (C) environment, this works as expected.
    * In a Forge (S) to Vanilla (C) environment, the clients retain their original reach distances of `Entity [Attack and Reach]: 3.0` and `Block [Reach]: 4.5`
    * In a Vanilla (S) to Forge (C) environment, the clients now have an increased entity reach distance `Entity [Reach]: 4.5` but still retain their `Entity [Attack]: 3.0` and `Block [Reach]: 4.5`
    * I believe the benefit of this change outweights the small discrepancy, as vanilla permits these interactions anyway.  Vanilla's serverside check for allowing an interaction is `distanceSq < 36.0` or a hard-cap of `6.0` blocks, and `4.5 < 6.0`.

Shadows-of-Fire:
I believe, from a player perspective, having separate `Reach Distance` and `Attack Range` is the most intuitive method of going about it.
From a code perspective, vanilla would rather we had `Entity Reach` and `Block Reach`, but I feel like these stats are less intuitive with regards to what they actually mean.  If the change above (the 3.0 entity interaction range -> 4.5 on forge) is deemed too much of a break from vanilla, I think that would be the best route to pursue, having two stats, one for blocks and one for entities.

The other alternative is three stats (Attack Range, Entity Reach, and Block Reach), but that seems even more cumbersome than having the two be separate, and I would prefer two stats over this approach in all cases.

Shadows-of-Fire:
The change from an entity interaction range of 3.0 to 4.5 can be considered a fix to these issues: [MC-172289](https://bugs.mojang.com/browse/MC-172289) [MC-92484](https://bugs.mojang.com/browse/MC-92484)
The added [!= MISS check in GameRenderer](https://github.com/MinecraftForge/MinecraftForge/pull/8478/files#diff-3fb0d83ef26ef7092b984b34839c29f114505fa768daca844b05f100671eb90cR41) has been marked as not a user-facing bug in Vanilla ([MC-250858](https://bugs.mojang.com/browse/MC-250858)), but it is still required to have the attack range exceed the reach distance, and is thus required for this PR to provide the expected functionality.

Shadows-of-Fire:
Merge conflicts have been resolved.

Tslat:
Any progress on this?
I'm still getting numerous crash reports from users who are finding their mods colliding because of this being unfixed and mods not knowing to override their require in mixins

PR #8481, opened 2022-02-24
Re-add missing Shulker patch for EntityTeleportEvent

### LINKED ISSUE #8474: `EntityTeleportEvent.EnderEntity` is never dispatched for Shulkers
**Minecraft Version:** 1.18.1

**Forge Version:** 39.0.64

**Logs:** _unapplicable_

**Steps to Reproduce:**
 1. Read the javadocs of `EntityTeleportEvent.EnderEntity` and think that it will be dispatched for shulkers;
 2. Subscribe to the event and try to use it with this assumption;
 3. Be horribly disappointed.

**Description of issue:**
Doesn't seem like compatibility with aforementioned event is implemented at all for Shulkers.

sciwhiz12:
This PR fixes #8474 by re-adding a missing patch to `Shulker` which is responsible for firing `EntityTeleportEvent.EnderEntity`. This patch seems to have been removed during the 1.17 update and patch cycle, as I have noted in my comment in the linked issue.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/monster/Shulker.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/monster/Shulker.java.patch
@@ -1,0 +1,15 @@
+--- a/net/minecraft/world/entity/monster/Shulker.java
++++ b/net/minecraft/world/entity/monster/Shulker.java
+@@ -356,6 +_,12 @@
+             if (blockpos1.m_123342_() > this.f_19853_.m_141937_() && this.f_19853_.m_46859_(blockpos1) && this.f_19853_.m_6857_().m_61937_(blockpos1) && this.f_19853_.m_45756_(this, (new AABB(blockpos1)).m_82406_(1.0E-6D))) {
+                Direction direction = this.m_149810_(blockpos1);
+                if (direction != null) {
++                  net.minecraftforge.event.entity.EntityTeleportEvent.EnderEntity.EnderEntity event = net.minecraftforge.event.ForgeEventFactory.onEnderTeleport(this, blockpos1.m_123341_(), blockpos1.m_123342_(), blockpos1.m_123343_());
++                  if (event.isCanceled()) direction = null;
++                  blockpos1 = new BlockPos(event.getTargetX(), event.getTargetY(), event.getTargetZ());
++               }
++
++               if (direction != null) {
+                   this.m_19877_();
+                   this.m_149788_(direction);
+                   this.m_5496_(SoundEvents.f_12418_, 1.0F, 1.0F);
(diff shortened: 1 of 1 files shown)


ChampionAsh5357 (commented):
Just as a question, why do we not simply return false when the shulker's teleportation is canceled?  It seems a bit strange since the event could fire up to 5 times for a single shulker trying to teleport to a valid space.

KnightMiner:
Its possible the cancelation is based on the teleport target and not the teleport source. Suppose we cancel teleports to negative coordinates and a shulker is at 1,1. Just because we canceled the teleport to -4, -2 does not mean the shulker is not allowed to teleport to 3, 5

If you logic is based on source, it means you just cancel that shulker a few extra times. Think of it as the shulker tried to teleport 5 times and failed 5 times.

ChampionAsh5357:
> Just because we canceled the teleport to -4, -2 does not mean the shulker is not allowed to teleport to 3, 5.

Yeah, my immediate thought for a counterargument is that the modder could simply force the target to the direction of movement. The only reason I could think of not translating it using a teleport target would be if you just want to make it less likely for an entity to teleport or use some level data, which both could be handled with a single event call.

KnightMiner:
That requires the modder to know what is a valid teleport target for the entity, which is both a redundant effort to determine and may not be accurate with all "ender teleporting" entities (I will note this event is not shulker specific, so unless you create a tag list of mobs that may teleport multiple times so you know to redirect instead of canceling, that solution sounds absolutely terrible). Plus, based on the lore behind the teleport canceling, it may be more appropiate to cancel the teleport rather than redirecting it, I see it as a bad idea to disallow that route simply to fire a few less events.

Its a lot simplier to either redirect the shulker if you are writing teleport redirection or to cancel the teleport if you are writing teleport canceling, forcing modders who are writing teleport canceling to instead redirect the shulker is way worse than having a single event fire more than once. I think unless you can come up with a way to fire just one event that is not a burden to the modders, multiple events is better.

ChampionAsh5357 (approved):
Fair enough, mainly my curiosity talking for not making multiple calls.

patches/minecraft/net/minecraft/world/entity/monster/Shulker.java.patch
```diff
@@ -0,0 +1,15 @@
+--- a/net/minecraft/world/entity/monster/Shulker.java
++++ b/net/minecraft/world/entity/monster/Shulker.java
+@@ -356,6 +_,12 @@
+             if (blockpos1.m_123342_() > this.f_19853_.m_141937_() && this.f_19853_.m_46859_(blockpos1) && this.f_19853_.m_6857_().m_61937_(blockpos1) && this.f_19853_.m_45756_(this, (new AABB(blockpos1)).m_82406_(1.0E-6D))) {
+                Direction direction = this.m_149810_(blockpos1);
+                if (direction != null) {
++                  net.minecraftforge.event.entity.EntityTeleportEvent.EnderEntity.EnderEntity event = net.minecraftforge.event.ForgeEventFactory.onEnderTeleport(this, blockpos1.m_123341_(), blockpos1.m_123342_(), blockpos1.m_123343_());
```
SizableShrimp: `net.minecraftforge.event.entity.EntityTeleportEvent.EnderEntity.EnderEntity` does not look correct

sciwhiz12: Whoops, my copy-paste of the class name must've been faulty. Fixed.

PR #8482, opened 2022-02-25
Fix item durability dar to respect custom max damage

### LINKED ISSUE #8294: [1.18.1] [Bug] Item duration bar renders as black after item takes damage if IForgeItem.getMaxDamage is overriden.
**Minecraft Version:** 1.18.1
**Forge Version:** 39.0.0

**Steps to Reproduce:**
- override `IForgeItem.getMaxDamage(ItemStack stack)` in an item class.
- have this item take damage (via `ItemStack.hurt` or `ItemStack.hurtAndBreak`)

**Description of Issue:**
- Item duration bar renders as black after item takes damage.

**Image of Issue:**
image

sciwhiz12:
This PR fixes #8294 by changing `Item#getBarWidth` and `Item#getBarColor` to use the maximum damage returned by `IForgeItem#getMaxDamage(ItemStack)` instead of using the `maxDamage` field.

As an implementation note, the patch to `Item#getBarColor` uses a variable to store the returned value of `getMaxDamage(ItemStack)` to avoid calling the method twice in the same code block.


patches/minecraft/net/minecraft/world/item/Item.java.patch
```diff
@@ -42,6 +42,21 @@
...
+ 
+    public int m_142158_(ItemStack p_150900_) {
+-      return Math.round(13.0F - (float)p_150900_.m_41773_() * 13.0F / (float)this.f_41371_);
++      return Math.round(13.0F - (float)p_150900_.m_41773_() * 13.0F / (float)this.getMaxDamage(p_150900_));
+    }
+ 
+    public int m_142159_(ItemStack p_150901_) {
+-      float f = Math.max(0.0F, ((float)this.f_41371_ - (float)p_150901_.m_41773_()) / (float)this.f_41371_);
++      int stackMaxDamage = this.getMaxDamage(p_150901_);
++      float f = Math.max(0.0F, ((float)stackMaxDamage - (float)p_150901_.m_41773_()) / (float)stackMaxDamage);
```
ChampionAsh5357: Wouldn't it make sense to just convert the int to a float once and supply that value to the local variable?

sciwhiz12: I did think about that, but decided to go ahead with the current way unless told otherwise. Shall I also remove the redundant `(float)` casts as well?

ChampionAsh5357: I don't see why not. The argument against is that the logic doesn't encompass any changes, but it shouldn't necessarily matter as the patch can always be replaced between versions if it happens to get updated.

ChampionAsh5357 (approved):

cech12 (approved):
Looks good to me :)

TheCurle:
Oops.
That's my fault.

Nice catch, great fix.

PR #8486, opened 2022-02-27
Use UTF-8 charset for Java compilation

PaintNinja:
Gradle only treats buildscripts as UTF-8 by default. Java compilation tasks use the OS' default character set, which can vary and can sometimes cause issues with certain characters such as box drawings in Javadocs. This patch sets Java compilation tasks in the MDK and Forge to use UTF-8 instead.


sciwhiz12 (approved):

TheCurle:
Do you have a log or something from where not having this causes problems?
Just for posterity.

PaintNinja:
> Do you have a log or something from where not having this causes problems? Just for posterity.

When using box drawings from ASCII, building on Windows gives me this:
image

image

TheCurle:
Perfect. Thanks.

PR #8488, opened 2022-02-28
Replace '-' with '.' in mod JPMS module ids to avoid crash

kb-1000:
Rationale:
 * `-` is valid in modids (or was before 1.17, but is still treated as valid and is still a valid resource location namespace), while `.` is not
 * `.` is valid in JPMS module names, while `-` is not

To restore the old behavior, one could simply replace `-` with `.` while constructing the module name (which is done in this PR)

### DIFF
--- a/fmlloader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
+++ b/fmlloader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
@@ -191,7 +191,7 @@ public class ModFileInfo implements IModFileInfo, IConfigurable
 
     @Override
     public String moduleName() {
-        return getMods().get(0).getModId();
+        return getMods().get(0).getModId().replace('-', '.');
     }
 
     @Override


sciwhiz12:
The current opinion of the Core team is that, instead of modifying the module name to replace `-` to `.`, the regex for mod IDs should be modified to ban the use of `-`. 

The rationale is that currently, mods on 1.18 (and 1.17) whose mod IDs have `-` already do not work (so no currently existing and working mods for 1.18 are broken). So, instead of inserting a workaround so module names formed from mod IDS are fixed to work, mod IDs will instead be required to conform to module name restrictions (that is, no use of `-`).

KnightMiner:
To be clear, this crash only affects forge mods as its converting the mod ID into a module ID? Its not possible for a resource pack/datapack to define a domain that triggers this crash?

If so, using sci's fix also gives us a nice feature for datapacks, you can ensure a certain resource ID will never be used by a mod.

LexManos:
That is correct, Resource Packs are not injected into the Java Module System and as such do not need to have valid Module names.

PR #8490, opened 2022-03-02
Re-add access transformer for RecipeProvider#has with tag argument

SuperMartijn642:
The `RecipeProvider#has(TagKey<Item>)` in 1.18.2 appears to be the equivalent of `RecipeProvider#has(Tag<Item>)` in 1.18.1. The method provides an easy way to obtain a criterion for a player having items from a certain tag in their inventory and can be used as the trigger to unlock recipes.
The 1.18.1 method had an entry in Forge's access transformer to change its visibility from private to protected. 
https://github.com/MinecraftForge/MinecraftForge/blob/4f4849b60c1a5f956016b8ce2ad2c128a98dcc1c/src/main/resources/META-INF/accesstransformer.cfg#L188
This entry seems to have been removed in the update to 1.18.2, likely due to the changes to tags.

This PR adds an access transformer entry to change the visibility of `RecipeProvider#has(TagKey<Item>)` from private to protected, so mods can use the method in their own classes extending `RecipeProvider`.


LexManos:
Ya, a few ATs may of been missed/mistranslated with reguards to all the tag stuff. Tags will be the major bain of the 1.18.2 update. Keep reporting things as you find them. I'll do a pass of PRs/Issues tomarrow morning when i wake up.

sciwhiz12 (changes requested):
Looks fine to me. Checking the [1.18.2 update commit](https://github.com/MinecraftForge/MinecraftForge/commit/425eff39ff46affdceb0bcbe0738c24b72f707e3#diff-3c5e61e6b3303cd31e5882573871953441b3b1d9a4cc7b018b1ff5eda25ce27d), it seems to me that this is the only access transformer line of note which got removed.

Please run the `checkATs` task so the access transformer files' entries are checked, re-ordered, and the newline removed.

SuperMartijn642:
Yup, the `RecipeProvider#has(TagKey<Item>)` is the only one that stood out to me and the only one I came across when updating from 1.18.1 to 1.18.2.

I did run the checkATs task now, so the ordering and new line at the end should be correct now.

sciwhiz12 (changes requested):
After some small internal discussion, it's been decided that the other helper methods in the `RecipeProvider` class also be AT'd to `protected` in this PR. Broadly, this means the `get*Name` helpers and the various currently-`private` helpers below the `buildCraftingRecipes` method.

Please do not forget to run `checkATs` afterwards for checking, re-ordering, and un-EOF-newlining.

SuperMartijn642:
I have added AT entries for all the helper functions in `RecipeProvider` now.

I also added AT entries to expose the `COAL_SMELTABLES`, `IRON_SMELTABLES`, etc. and `shapeBuilders` fields in `RecipeProvider`.
The `*_SMELTABLES` fields are used by vanilla to generate smelting recipes and I think they could be useful to someone making an alternative smelting system and wanting to add the same recipes to it as vanilla.
The `shapeBuilders` field is only used by `RecipeProvider#generateRecipes` which applies every shape builder to a given `BlockFamily`. The field itself may be useful to someone wanting to apply only a subset of all shape builders instead of applying all of them.

marchermans (approved):

sciwhiz12 (approved):

TheCurle:
Nobody has fielded any other suggestions yet, so given the long waiting period (almost a month) i'm going to pull this now.
If anyone wants anything else added, open a discussion on our Discord server.

PR #8492, opened 2022-03-02
Rework fog events

eerussianguy:
This is a continuation of #8338

The 1.18.1 and 1.18.2 updates changed some stuff related to how the setupFog method works, fixing some oddities, and notably adding a fog shape feature. The 8+ year old FogDensity event no longer serves a real purpose, given that the render event has had the same capabilities for a few updates now (since RenderSystem calls can be made anywhere).

This PR deprecates FogDensityEvent, the old RenderFogEvent constructor, and adds helper methods for the necessary render system calls, for which it also provides getters for vanilla's values. It also removes an old redundant patch (see previous pr description). There should not be any breaking changes.

### DIFF
--- a/patches/minecraft/net/minecraft/client/renderer/FogRenderer.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/FogRenderer.java.patch
@@ -1,13 +1,6 @@
 --- a/net/minecraft/client/renderer/FogRenderer.java
 +++ b/net/minecraft/client/renderer/FogRenderer.java
-@@ -173,24 +_,42 @@
- 
-       if (f_109010_ != 0.0F && f_109011_ != 0.0F && f_109012_ != 0.0F) {
-          float f8 = Math.min(1.0F / f_109010_, Math.min(1.0F / f_109011_, 1.0F / f_109012_));
-+         // Forge: fix MC-4647 and MC-10480
-+         if (Float.isInfinite(f8)) f8 = Math.nextAfter(f8, 0.0);
-          f_109010_ = f_109010_ * (1.0F - f6) + f_109010_ * f8 * f6;
-          f_109011_ = f_109011_ * (1.0F - f6) + f_109011_ * f8 * f6;
+@@ -178,19 +_,36 @@
           f_109012_ = f_109012_ * (1.0F - f6) + f_109012_ * f8 * f6;
        }
  
@@ -36,6 +29,7 @@
        FogShape fogshape = FogShape.SPHERE;
        float f;
        float f1;
++      // TODO: remove this hook in 1.19
 +      float hook = net.minecraftforge.client.ForgeHooksClient.getFogDensity(p_109026_, p_109025_, partialTicks, 0.1F);
 +      if (hook >= 0) {
 +         f = -8.0f;
@@ -48,7 +42,7 @@
        RenderSystem.m_157445_(f);
        RenderSystem.m_157443_(f1);
        RenderSystem.m_202160_(fogshape);
-+      net.minecraftforge.client.ForgeHooksClient.onFogRender(p_109026_, p_109025_, partialTicks, f1);
++      net.minecraftforge.client.ForgeHooksClient.onFogRender(p_109026_, p_109025_, partialTicks, f, f1, fogshape);
     }
  
     public static void m_109036_() {
--- a/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
+++ b/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
@@ -6,6 +6,8 @@
 package net.minecraftforge.client;
 
 import com.google.common.collect.ImmutableMap;
+
+import com.mojang.blaze3d.shaders.FogShape;
 import com.mojang.blaze3d.systems.RenderSystem;
 import com.mojang.blaze3d.vertex.*;
 import com.mojang.datafixers.util.Either;
@@ -389,9 +391,9 @@ public class ForgeHooksClient
         return -1;
     }
 
-    public static void onFogRender(FogMode type, Camera camera, float partialTick, float distance)
+    public static void onFogRender(FogMode type, Camera camera, float partialTick, float nearDistance, float farDistance, FogShape shape)
     {
-        MinecraftForge.EVENT_BUS.post(new EntityViewRenderEvent.RenderFogEvent(type, camera, partialTick, distance));
+        MinecraftForge.EVENT_BUS.post(new EntityViewRenderEvent.RenderFogEvent(type, camera, partialTick, nearDistance, farDistance, shape));
     }
 
     public static EntityViewRenderEvent.CameraSetup onCameraSetup(GameRenderer renderer, Camera camera, float partial)
--- a/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
+++ b/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
@@ -391,9 +391,24 @@ public class ForgeHooksClient
         return -1;
     }
 
+    /**
+     * @deprecated to be removed in 1.19, use other onFogRender hook with more params
+     */
+    @Deprecated(forRemoval = true, since = "1.18.2")
+    public static void onFogRender(FogMode type, Camera camera, float partialTick, float distance)
+    {
+        MinecraftForge.EVENT_BUS.post(new EntityViewRenderEvent.RenderFogEvent(type, camera, partialTick, distance));
+    }
+
     public static void onFogRender(FogMode type, Camera camera, float partialTick, float nearDistance, float farDistance, FogShape shape)
     {
-        MinecraftForge.EVENT_BUS.post(new EntityViewRenderEvent.RenderFogEvent(type, camera, partialTick, nearDistance, farDistance, shape));
+        EntityViewRenderEvent.RenderFogEvent event = new EntityViewRenderEvent.RenderFogEvent(type, camera, partialTick, nearDistance, farDistance, shape);
+        if (MinecraftForge.EVENT_BUS.post(event))
+        {
+            RenderSystem.setShaderFogStart(event.getNearPlaneDistance());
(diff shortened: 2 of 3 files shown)


src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -389,9 +391,9 @@ public static float getFogDensity(FogMode type, Camera camera, float partialTick
         return -1;
     }
 
-    public static void onFogRender(FogMode type, Camera camera, float partialTick, float distance)
+    public static void onFogRender(FogMode type, Camera camera, float partialTick, float nearDistance, float farDistance, FogShape shape)
```
ChampionAsh5357, line 403: This should also be deprecated instead of replaced as its still public API.

src/main/java/net/minecraftforge/client/event/EntityViewRenderEvent.java
```diff
@@ -83,23 +90,67 @@ public void setDensity(float density)
...
     /**
-     * Event that allows any feature to customize the rendering of fog.
+     * This event allows for customization of parameters related to fog rendering. The plane distances are based on the player's render distance.
+     *
+     * A FogMode of FOG_SKY is the skybox's fog.
+     * A FogMode of FOG_TERRAIN is what the fog induced by render distance uses. This works best for reducing the camera's visibility.
      */
-    @HasResult
-    public static class RenderFogEvent extends FogEvent
+    public static class RenderFogEvent extends EntityViewRenderEvent
```
ChampionAsh5357: Breaking change

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -389,9 +391,9 @@ public static float getFogDensity(FogMode type, Camera camera, float partialTick
...
     {
-        MinecraftForge.EVENT_BUS.post(new EntityViewRenderEvent.RenderFogEvent(type, camera, partialTick, distance));
+        MinecraftForge.EVENT_BUS.post(new EntityViewRenderEvent.RenderFogEvent(type, camera, partialTick, nearDistance, farDistance, shape));
     }
```
sciwhiz12: This still is counted as a breaking change to my understanding, and therefore must be treated appropriately (terminally deprecated, overload with new parameters).

src/main/java/net/minecraftforge/client/event/EntityViewRenderEvent.java
```diff
@@ -43,6 +46,7 @@ public double getPartialTicks()
         return partialTicks;
     }
 
+    //TODO: remove in 1.19
```
sciwhiz12: This should be a `@Deprecated(forRemoval = true, since = "1.18.2")` annotation and a javadoc with a `@deprecated` tag explaining that it will be removed in 1.19 and why it will be removed.

src/main/java/net/minecraftforge/client/event/EntityViewRenderEvent.java
```diff
@@ -57,10 +61,13 @@ protected FogEvent(FogMode mode, Camera camera, double partialTick)
     }
 
     /**
+     * MODDERS: use RenderFogEvent
+     *
      * Event that allows any feature to customize the fog density the player sees.
      * NOTE: In order to make this event have an effect, you must cancel the event
      */
     @Cancelable
+    @Deprecated //TODO: remove in 1.19
```
sciwhiz12: This should be a `@Deprecated(forRemoval = true, since = "1.18.2")` annotation and a javadoc with a `@deprecated` tag explaining that it will be removed in 1.19, why it is terminally deprecated, and linking to or showing alternatives to the event. (And to clarify, the `MODDERS:` part can be removed in favor of the `@deprecated` javadoc tag.)

src/main/java/net/minecraftforge/client/event/EntityViewRenderEvent.java
```diff
@@ -83,23 +90,67 @@ public void setDensity(float density)
...
         private final float farPlaneDistance;
+        private final float nearPlaneDistance;
+        private final FogShape fogShape;
 
+        @Deprecated //TODO: remove in 1.19
         public RenderFogEvent(FogMode type, Camera camera, float partialTicks, float distance)
         {
-            super(type, camera, partialTicks);
-            this.farPlaneDistance = distance;
+            this(type, camera, partialTicks, 0f, distance, FogShape.SPHERE);
```
ChampionAsh5357: Near plane is treated as `-8f` as default in the hook so that probably should be reflected here.

src/main/java/net/minecraftforge/client/event/EntityViewRenderEvent.java
sciwhiz12: These are _technically_ breaking changes:
 - The [javadoc for `Event#setResult`](https://github.com/MinecraftForge/EventBus/blob/6d3407ccf8a8f74bc66c759cbf40f3909c918c8f/src/main/java/net/minecraftforge/eventbus/api/Event.java#L120-L127) states that attempting to set the result for an event which does not have one (i.e. no `@HasResult`) causes an `IllegalArgumentException`. Regardless of the actual visible implementation, that is the specification in the javadocs and we must abide by it. 

   Since there is already no use for the event result, you can keep the annotation and mark it with a `// TODO: ` to remove the annotation in 1.19.

 - The change in superclass now means that any possibly mod which expect `RenderFogEvent` to extend `FogEvent` will fail. While this doesn't sound like an issue, practically speaking, it is still technically a breaking change. For now, it should extend `FogEvent` and marked with a `// TODO: ` to change its superclass to `EntityViewRenderEvent` in 1.19.

src/main/java/net/minecraftforge/client/event/EntityViewRenderEvent.java
```diff
@@ -83,23 +90,67 @@ public void setDensity(float density)
...
-    @HasResult
-    public static class RenderFogEvent extends FogEvent
+    public static class RenderFogEvent extends EntityViewRenderEvent
     {
+        private final FogMode type;
         private final float farPlaneDistance;
+        private final float nearPlaneDistance;
+        private final FogShape fogShape;
 
+        @Deprecated //TODO: remove in 1.19
```
sciwhiz12: This should be a `@Deprecated(forRemoval = true, since = "1.18.2")` annotation and a javadoc with a `@deprecated` tag saying that it will be removed in 1.19.

src/main/java/net/minecraftforge/client/event/EntityViewRenderEvent.java
```diff
@@ -83,23 +90,67 @@ public void setDensity(float density)
     }
 
     /**
-     * Event that allows any feature to customize the rendering of fog.
+     * This event allows for customization of parameters related to fog rendering. The plane distances are based on the player's render distance.
```
ChampionAsh5357, line 92: This sentence seems to be immediately contradicted by the descriptions of  `FogMode`.

ChampionAsh5357 (changes requested):

src/main/java/net/minecraftforge/client/event/EntityViewRenderEvent.java
```diff
@@ -83,23 +90,67 @@ public void setDensity(float density)
...
+
+        public void setNearPlaneDistance(float distance)
+        {
+            RenderSystem.setShaderFogStart(distance);
+        }
+
+        public void setFogShape(FogShape shape)
+        {
+            RenderSystem.setShaderFogShape(shape);
+        }
```
sciwhiz12: Currently, this sets the near and far planes and shape immediately, without changing the values returned by the getters. This may pose possible mod compatibility problems if multiple mods wish to simply "add on" to the near/far plane distances, as currently they would always overwrite any previously set distance.

Perhaps it may be better to store the originals, and have mutable fields in the event with getters and setters, which are then provided to `RenderSystem` in the hook?

src/main/java/net/minecraftforge/client/event/EntityViewRenderEvent.java
eerussianguy: I don't think this is a breaking change, I copied over the getMode method so any uses of the event will operate exactly the same

patches/minecraft/net/minecraft/client/renderer/FogRenderer.java.patch
```diff
@@ -36,6 +29,7 @@
        FogShape fogshape = FogShape.SPHERE;
        float f;
        float f1;
++      // TODO: remove this hook in 1.19
 +      float hook = net.minecraftforge.client.ForgeHooksClient.getFogDensity(p_109026_, p_109025_, partialTicks, 0.1F);
```
sciwhiz12, line 33: The hook method should be terminally deprecated (`@Deprecated(forRemoval = true, since = "1.18.2")`, `/** @deprecated To be removed in 1.19, because ... */`).

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/client/event/EntityViewRenderEvent.java
```diff
@@ -83,23 +90,74 @@ public void setDensity(float density)
...
+
+        public void setNearPlaneDistance(float distance)
+        {
+            nearPlaneDistance = distance;
+        }
+
+        public void setFogShape(FogShape shape)
+        {
+            fogShape = shape;
+        }
```
ChampionAsh5357, line 159: You could make incrementing setters as well for these to make it such that adding onto the existing value is not simply `set(get + newVal)`.

ChampionAsh5357 (approved):

marchermans (changes requested):
<Ignore my stupidity>

eerussianguy:
I chose to do multipliers rather than adding, because the initial parameter passed into the fog event is based on the render distance, which is different based on the settings

ChampionAsh5357 (approved):

sciwhiz12 (approved):

marchermans (approved):

PR #8494, opened 2022-03-02
Make has(TagKey) protected

SirEndii:
In forge 1.18.2, the method `net.minecraft.data.recipes.RecipeProvider#has(net/minecraft/tags/TagKey)` is private not as `has(ItemLike)`.
This PR adds an entry to the forge AT config to make this method public.

This is the first time I contribute to Forge, so I'm sorry if I forgot something.


SirEndii:
I saw that already someone made a PR for this... oops

PR #8495, opened 2022-03-03
[1.18.2] Fix tags for custom forge registries

SizableShrimp:
This PR addresses a few tag-related regressions due to the 1.18.2 update:
* Allow custom forge registries to use tags again. A custom registry must call `RegistryBuilder#hasTags` to enable tags, otherwise __they are off by default__. 
* Allow querying a Holder based on ResourceKey, ResourceLocation, or value for tag-enabled registries
* Fix ForgeRegistryTagsProvider to use the right tag directory method
* Cleanup nullability annotations in `IForgeRegistry` and `ForgeRegistry` and move to jetbrains annotations
* Add custom forge registries to the root registry (global registries)
* Expose new `ITagManager` system for querying tag-related info from a forge registry
* Big change: **tagFolder() is gone. Modded registries must place their tags in `data/<namespace>/tags/<registryNamespace>/<registryPath>`** - For example with a registry name `examplemod:test_registry`, the tags folder may look like`data/mydatapackid/tags/examplemod/test_registry/`

Checking whether a custom registry entry is part of a tag can be done with something like so:
```java
    public boolean is(TagKey<CoolThing> key) {
        return coolThingForgeRegistry.tags().getTag(key).contains(this);
    }
```

Optional tags have been reintroduced with `ITagManager#createOptionalTagKey(ResourceLocation, Set<Supplier<V>>)`.


Cyborgmas:
You can probably delete the ForgeTagHandler class completely? Also if optional tags are no longer needed, it means the "default set" behaviour would be manual by modders, haven't tested but would imagine it requires checking if a tag is empty on the client and providing their own default behaviour. Not a big deal but thought it was worth mentioning.

The one problem I see is that you changed the tag folder logic for previously custom registries (i.e. enchantments) and mod added custom registries. For vanilla registries it's a bit forced as they don't have the "s" added, and I would say it's a bit too much to consider both folders to be valid. However, for custom registries, compatibility for the tag folder could be maintained and deprecated to be removed in 1.19 perhaps?

src/main/java/net/minecraftforge/registries/IForgeRegistry.java
```diff
@@ -34,17 +39,24 @@
...
+    @NotNull Set<Entry<ResourceKey<V>, V>> getEntries();
 
-    @Nonnull Codec<V> getCodec();
+    @NotNull Codec<V> getCodec();
+
+    @NotNull Optional<Holder<V>> getHolder(ResourceKey<V> key);
+    @NotNull Holder<V> getHolderOrThrow(ResourceKey<V> key);
+    @NotNull HolderSet.Named<V> getOrCreateTag(TagKey<V> name);
+    @NotNull Optional<HolderSet.Named<V>> getTag(TagKey<V> name);
+    @NotNull Iterable<Holder<V>> getTagOrEmpty(TagKey<V> name);
```
pupnewfster: I believe it may also make sense to expose the following methods that also exist on `MappedRegistry`:
- `boolean isKnownTagName(TagKey)`
- `Stream<Pair<TagKey<T>, HolderSet.Named<T>>> getTags()`
- `Stream<TagKey<T>> getTagNames()`

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -167,12 +168,13 @@ public ResourceLocation getRegistryName()
         return superType;
     }
 
-    @Nullable
-    public String getTagFolder()
+    @Override
+    public boolean supportsTags()
     {
-        return tagFolder;
+        return this.builder.getHasWrapper();
```
pupnewfster: I feel like this potentially should just be stored in a local variable the same as everything else that references the builder.

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -234,26 +236,87 @@ public ResourceLocation getKey(V value)
...
+    @SuppressWarnings("unchecked")
+    @NotNull
+    NamespacedHolderHelper<V> getHolderHelper()
+    {
+        if (!this.builder.getHasWrapper())
+            throw new IllegalStateException("Cannot query holder helper for non-wrapped registry");
+
+        return this.defaultKey != null
+                ? this.getSlaveMap(NamespacedDefaultedWrapper.Factory.ID, NamespacedDefaultedWrapper.class).getHolderHelper()
+                : this.getSlaveMap(NamespacedWrapper.Factory.ID, NamespacedWrapper.class).getHolderHelper();
```
pupnewfster: This looks like it could be "simplified" (reduce the long slave map calls) to:
```java
Registry<V> wrapper = getWrapper();
if (wrapper == null)
    throw new IllegalStateException("Cannot query holder helper for non-wrapped registry");
return wrapper.getHolderHelper();
```

And then the suppress warnings for unchecked could probably be removed as well.

pupnewfster (commented):
I also agree with Cyborg that `ForgeTagHandler` can probably go as everything is commented out in it except for one method that is just returning an always empty set now.

I am not sure how worth it is keeping tag folders for custom registries until 1.19 as with 1.18.2 the tag changes are so breaking they will have to rewrite their things anyway.

pupnewfster:
One final thought is for:
> Expose API methods for holders and holder sets based on related Registry methods - only works for wrapped forge registries (all tag-enabled registries will have a wrapped Registry variant)

It likely makes sense to mention with a very short javadocs on the given methods that they will fail if the registry doesn't have a backing wrapped vanilla registry.

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
SizableShrimp: This is not possible because the two slave maps have no common super type with getHolderHelper() declared

SizableShrimp:
> You can probably delete the ForgeTagHandler class completely? Also if optional tags are no longer needed, it means the "default set" behaviour would be manual by modders, haven't tested but would imagine it requires checking if a tag is empty on the client and providing their own default behaviour. Not a big deal but thought it was worth mentioning.

I am open to suggestions on being able to default a tag in the event it does not exist. Maybe this is possible with a simple change to NamespacedHolderHelper#bindTags.

> However, for custom registries, compatibility for the tag folder could be maintained and deprecated to be removed in 1.19 perhaps?

This is just unnecessary patching. Tags were already a breaking change for mods with TagKey so keeping this in is not worthwhile.

src/main/java/net/minecraftforge/registries/ForgeRegistries.java
```diff
@@ -129,7 +129,7 @@
 
         //Forge
         public static final ResourceKey<Registry<DataSerializerEntry>> DATA_SERIALIZERS = key("data_serializers");
-        public static final ResourceKey<Registry<GlobalLootModifierSerializer<?>>> LOOT_MODIFIER_SERIALIZERS = key("forge:loot_modifier_serializers");
+        public static final ResourceKey<Registry<DataSerializerEntry>> LOOT_MODIFIER_SERIALIZERS = key("forge:loot_modifier_serializers");
```
pupnewfster: Is this an intended change? It is hard to tell without being able to browse the code.

src/main/java/net/minecraftforge/registries/NamespacedHolderHelper.java
```diff
@@ -120,6 +133,11 @@ boolean isKnownTagName(TagKey<T> name)
         return named;
     }
 
+    void addOptionalTag(TagKey<T> name, Set<Supplier<T>> defaults)
+    {
+        this.optionalTags.computeIfAbsent(name, k -> new HashSet<>()).addAll(defaults);
```
pupnewfster: I don't think this and the injecting of works quite right in terms of duplicates. As it is a set of suppliers it won't remove duplicates if there are two different instances supplying the same value. I think it may need a distinct call or something before the toList.

src/main/java/net/minecraftforge/registries/ForgeRegistries.java
SizableShrimp: No, this was likely a dumb IntelliJ refactor thing I missed. Resolved.

src/test/java/net/minecraftforge/debug/DuplicateOptionalTagTest.java
```diff
@@ -41,46 +42,47 @@
...
+   {
+       // if (!TAG_A.isDefaulted())
+       // {
+       //     LOGGER.warn("First instance of optional tag is not defaulted!");
+       // }
+       //
+       // if (!TAG_B.isDefaulted())
+       // {
+       //     LOGGER.warn("Second instance of optional tag is not defaulted!");
+       // }
```
sciwhiz12: If there are no plans at the moment to reintroduce the method to check if an optional tag is 'defaulted', this commented-out block should be removed.

src/main/java/net/minecraftforge/registries/NamespacedHolderHelper.java
```diff
@@ -49,6 +52,7 @@
     private ObjectList<Holder.Reference<T>> holdersById = new ObjectArrayList<>(256);
     private Map<ResourceLocation, Holder.Reference<T>> holdersByName = new HashMap<>();
     private Map<T, Holder.Reference<T>> holders = new IdentityHashMap<>();
+    private Map<TagKey<T>, Set<Supplier<T>>> optionalTags = new IdentityHashMap<>();
```
sciwhiz12: The field can be declared `final`. Additionally, it may be nicer here to use a `Multimap<TagKey<T>, Supplier<T>>` instead. To retain the use of `IdentityHashMap`, use a multimap created through `Multimaps.newSetMultimap(new IdentityHashMap<>(), HashSet::new)`.

src/test/java/net/minecraftforge/debug/DuplicateOptionalTagTest.java
```diff
@@ -41,46 +42,47 @@
...
+           LOGGER.error("Values of both optional tag instances are not the same: first instance: {}, second instance: {}", tagAValues, tagBValues);
+           return;
+       }
+
+       final Set<Block> expected = Sets.union(TAG_A_DEFAULTS, TAG_B_DEFAULTS).stream()
+               .map(Supplier::get)
+               .collect(Collectors.toUnmodifiableSet());
+       if (!tagAValues.equals(expected))
+       {
+           // TODO Turn this into an actual exception? This can easily get lost in logs if not looking carefully
```
sciwhiz12: Perhaps, we should include a `Throwable` parameter at the end of the logging call to print out a visible stacktrace, instead of throwing an exception which would halt loading?

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -234,26 +240,164 @@ public ResourceLocation getKey(V value)
...
+    @NotNull
+    @Override
+    public TagKey<V> createOptionalTagKey(ResourceLocation location, @NotNull Set<Supplier<V>> defaults)
+    {
+        TagKey<V> tagKey = createTagKey(location);
+
+        getHolderHelper().addOptionalTag(tagKey, defaults);
+
+        return tagKey;
+    }
```
sciwhiz12: This probably should be advertised when this PR is merged as the replacement for the old `createOptional` with defaults.

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -234,26 +240,164 @@ public ResourceLocation getKey(V value)
...
+
+    @NotNull
+    @Override
+    public Optional<Holder<V>> getHolder(V value)
+    {
+        return getHolderHelper().getHolder(value);
+    }
+
+    @Override
+    public @NotNull Holder<V> getHolderOrThrow(V value)
```
sciwhiz12: The location of this `@NotNull` annotation is different from the others.

src/main/java/net/minecraftforge/network/filters/VanillaConnectionNetworkFilter.java
```diff
@@ -42,8 +45,10 @@
 @ChannelHandler.Sharable
 public class VanillaConnectionNetworkFilter extends VanillaPacketFilter
 {
-
     private static final Logger LOGGER = LogManager.getLogger();
+    private static final Set<ResourceLocation> FORGE_ADDED_REGISTRIES = Stream.of(ForgeRegistries.Keys.DATA_SERIALIZERS, ForgeRegistries.Keys.LOOT_MODIFIER_SERIALIZERS, ForgeRegistries.Keys.WORLD_TYPES)
+            .map(ResourceKey::location)
+            .collect(Collectors.toUnmodifiableSet());
```
sciwhiz12: This set's existence should likely be noted in `ForgeRegistries`, for ease of maintenance should a new Forge-exclusive registry be added.

src/main/java/net/minecraftforge/network/filters/VanillaConnectionNetworkFilter.java
```diff
@@ -42,8 +45,10 @@
 @ChannelHandler.Sharable
 public class VanillaConnectionNetworkFilter extends VanillaPacketFilter
 {
-
     private static final Logger LOGGER = LogManager.getLogger();
```
sciwhiz12: We should likely also change the use of `LogManager` with `LogUtils`.

src/main/java/net/minecraftforge/registries/RegistryManager.java
```diff
@@ -133,9 +141,25 @@ public String getName()
             this.synced.add(name);
         for (ResourceLocation legacyName : builder.getLegacyNames())
             addLegacyName(legacyName, name);
+        if (registerToRoot && this == RegistryManager.ACTIVE && builder.getHasWrapper() && !Registry.REGISTRY.containsKey(reg.getRegistryName()) && Registry.REGISTRY instanceof WritableRegistry<?> rootRegistry)
+            registerRegistry(rootRegistry, reg);
         return getRegistry(name);
     }
 
+    @SuppressWarnings("unchecked")
+    private <T> void registerRegistry(WritableRegistry<T> rootRegistry, ForgeRegistry<?> forgeReg) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    private <T> void registerRegistry(WritableRegistry<T> rootRegistry, ForgeRegistry<?> forgeReg)
    {
```

src/main/java/net/minecraftforge/registries/RegistryManager.java
```diff
@@ -133,9 +141,25 @@ public String getName()
...
+        Registry<?> wrapper = forgeReg.getWrapper();
+        if (wrapper != null)
+            rootRegistry.register((ResourceKey<T>) forgeReg.getRegistryKey(), (T) wrapper, Lifecycle.experimental());
+    }
+
+    public static CompletableFuture<List<Throwable>> preNewRegistryEvent(final Executor executor, final IModStateTransition.EventGenerator<? extends RegistryEvent.NewRegistry> eventGenerator) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    public static CompletableFuture<List<Throwable>> preNewRegistryEvent(final Executor executor, final IModStateTransition.EventGenerator<? extends RegistryEvent.NewRegistry> eventGenerator)
    {
```

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/registries/NamespacedHolderHelper.java
SizableShrimp: Using a Multimap provides minimal cleanup, but I will take the suggestion.

patches/minecraft/net/minecraft/data/tags/BiomeTagsProvider.java.patch
```diff
@@ -0,0 +1,15 @@
+--- a/net/minecraft/data/tags/BiomeTagsProvider.java
++++ b/net/minecraft/data/tags/BiomeTagsProvider.java
+@@ -7,8 +_,12 @@
+ import net.minecraft.world.level.biome.Biomes;
+ 
+ public class BiomeTagsProvider extends TagsProvider<Biome> {
++   @Deprecated
```
sciwhiz12, line 8: This should have a javadoc with `@deprecated` directing users to the other constructor.

patches/minecraft/net/minecraft/data/tags/ConfiguredStructureTagsProvider.java.patch
```diff
@@ -0,0 +1,15 @@
+--- a/net/minecraft/data/tags/ConfiguredStructureTagsProvider.java
++++ b/net/minecraft/data/tags/ConfiguredStructureTagsProvider.java
+@@ -7,8 +_,12 @@
+ import net.minecraft.world.level.levelgen.structure.BuiltinStructures;
+ 
+ public class ConfiguredStructureTagsProvider extends TagsProvider<ConfiguredStructureFeature<?, ?>> {
++   @Deprecated
```
sciwhiz12, line 8: This should have a javadoc with `@deprecated` directing users to the other constructor.

patches/minecraft/net/minecraft/data/tags/BiomeTagsProvider.java.patch
```diff
@@ -0,0 +1,15 @@
...
+    public BiomeTagsProvider(DataGenerator p_211094_) {
+       super(p_211094_, BuiltinRegistries.f_123865_);
++   }
++   public BiomeTagsProvider(DataGenerator p_211094_, String modId, @javax.annotation.Nullable net.minecraftforge.common.data.ExistingFileHelper existingFileHelper) {
```
sciwhiz12: Should we be using the `javax.annotation` annotation here, or the JetBrains one? Also, the `forge.exc` file needs to be updated to add this.

patches/minecraft/net/minecraft/data/tags/ConfiguredStructureTagsProvider.java.patch
```diff
@@ -0,0 +1,15 @@
...
+    public ConfiguredStructureTagsProvider(DataGenerator p_211098_) {
+       super(p_211098_, BuiltinRegistries.f_123862_);
++   }
++   public ConfiguredStructureTagsProvider(DataGenerator p_211098_, String modId, @javax.annotation.Nullable net.minecraftforge.common.data.ExistingFileHelper existingFileHelper) {
```
sciwhiz12: Should we be using the `javax.annotation` annotation here, or the JetBrains one? Also, the `forge.exc` file needs to be updated to add this.

patches/minecraft/net/minecraft/data/tags/GameEventTagsProvider.java.patch
```diff
@@ -8,8 +8,8 @@
     public GameEventTagsProvider(DataGenerator p_176826_) {
        super(p_176826_, Registry.f_175412_);
 +   }
-+   public GameEventTagsProvider(DataGenerator p_126517_, String modId, @javax.annotation.Nullable net.minecraftforge.common.data.ExistingFileHelper existingFileHelper) {
-+      super(p_126517_, Registry.f_175412_, modId, existingFileHelper);
++   public GameEventTagsProvider(DataGenerator p_176826_, String modId, @javax.annotation.Nullable net.minecraftforge.common.data.ExistingFileHelper existingFileHelper) {
++      super(p_176826_, Registry.f_175412_, modId, existingFileHelper);
```
sciwhiz12: The `forge.exc` file needs to be updated with this change in SRG parameter name.

src/main/java/net/minecraftforge/registries/IHolderHelperHolder.java
```diff
@@ -0,0 +1,6 @@
+package net.minecraftforge.registries;
```
sciwhiz12, line 6: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -234,26 +240,165 @@ public ResourceLocation getKey(V value)
...
+                ? this.getSlaveMap(NamespacedDefaultedWrapper.Factory.ID, NamespacedDefaultedWrapper.class)
+                : this.getSlaveMap(NamespacedWrapper.Factory.ID, NamespacedWrapper.class);
+    }
+
+    @SuppressWarnings("unchecked")
+    @NotNull
+    NamespacedHolderHelper<V> getHolderHelper()
+    {
+        Registry<V> wrapper = getWrapper();
+        if (!(wrapper instanceof IHolderHelperHolder))
```
sciwhiz12, line 262: As this cannot be a pattern-matching `instanceof` as my local testing reveals, this should be noted as a comment so future maintainers know at a glance why this does not use pattern-matching `instanceof`.

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/registries/tags/IReverseTag.java
```diff
@@ -0,0 +1,22 @@
...
+
+package net.minecraftforge.registries.tags;
+
+import net.minecraft.tags.TagKey;
+
+import java.util.stream.Stream;
+
+public interface IReverseTag<V>
+{
+    Stream<TagKey<V>> getTagKeyStream();
```
pupnewfster: As there is no other getter and it is obvious this is a stream due to the return type I feel like a better name for this may just be `getTagKeys`

src/main/java/net/minecraftforge/registries/tags/ITagManager.java
```diff
@@ -0,0 +1,45 @@
...
+ * This should be preferred to any {@link Holder}-related methods.
+ */
+public interface ITagManager<V extends IForgeRegistryEntry<V>> extends Iterable<ITag<V>>
+{
+    /**
+     * Will create an empty tag if it does not exist.
+     */
+    @NotNull ITag<V> getTag(@NotNull TagKey<V> name);
+
+    @NotNull Optional<IReverseTag<V>> getReverseTag(@NotNull V value);
```
pupnewfster, line 42: I feel like there should be some more javadocs on this interface, especially for example on this method giving a brief explanation of what a "reverse tag" even is (or maybe better yet have `IReverseTag` and `ITag` have javadocs both on the class level and method level.

src/main/java/net/minecraftforge/registries/tags/ITagManager.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @NotNull Stream<TagKey<V>> getTagNames();
+
+    @NotNull TagKey<V> createTagKey(@NotNull ResourceLocation location);
+
+    /**
+     * Creates a tag key that will use the set of defaults if no tag JSON is found.
+     * Useful on the client side when a server may not provide a specific tag.
+     */
+    @NotNull TagKey<V> createOptionalTagKey(@NotNull ResourceLocation location, @NotNull Set<Supplier<V>> defaults);
```
pupnewfster: Would there potentially be a slight benefit to expanding the set's generic to `Set<? extends Supplier<V>>`? That way if someone is passing a set of say registry objects they don't have to worry about explicitly declaring the set as being of suppliers?

src/main/java/net/minecraftforge/registries/ForgeRegistryTagManager.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    void bind(Map<TagKey<V>, HolderSet.Named<V>> holderTags)
+    {
+        IdentityHashMap<TagKey<V>, ITag<V>> newTags = new IdentityHashMap<>(this.tags);
+        Set<TagKey<V>> bound = new HashSet<>();
+
+        holderTags.forEach((key, holderSet) -> {
+            bound.add(key);
+            ((ForgeRegistryTag<V>) newTags.computeIfAbsent(key, ForgeRegistryTag::new)).bind(holderSet);
```
pupnewfster: Rather than having this cast could you just have the value type of `newTags` be `ForgeRegistryTag<V>`? and then just have `tags` be `Map<TagKey<V>, ? extends ITag<V>>` so that you can set it to `newTags` without any casting trickery.

src/main/java/net/minecraftforge/registries/ForgeRegistryTagManager.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @NotNull
+    @Override
+    public TagKey<V> createOptionalTagKey(@NotNull ResourceLocation location, @NotNull Set<Supplier<V>> defaults)
+    {
+        TagKey<V> tagKey = createTagKey(location);
+
+        this.owner.getHolderHelper().ifPresent(h -> h.addOptionalTag(tagKey, defaults));
+
+        return tagKey;
```
pupnewfster, line 127: Part of me wonders if this method should have some sort of "failure" return for if it was unable to add an optional tag (such as if there is no holder helper), rather than just returning a new tag key. (This might not make sense to do, just figured I would ask as it did cross my mind)

src/main/java/net/minecraftforge/server/command/TagsCommand.java
```diff
@@ -0,0 +1,266 @@
...
+            new TranslatableComponent("Unknown element '%s' in registry '%s'", tag, registry));
+
+    public static ArgumentBuilder<CommandSourceStack, ?> register()
+    {
+        /*
+         * /forge tags <registry> list [page]
+         * /forge tags <registry> get <tag> [page]
+         * /forge tags <registry> query <element> [page]
+         */
+        return Commands.literal("tags")
```
pupnewfster, line 82: I don't see a major reason not to allow this, but do we want this command to be accessible by everyone by default (like it is now)? Or do we want to potentially require the sender to be an op

src/main/java/net/minecraftforge/server/command/TagsCommand.java
```diff
@@ -0,0 +1,266 @@
...
+    {
+        final ResourceKey<? extends Registry<?>> registryKey = getResourceKey(ctx, "registry", ROOT_REGISTRY_KEY)
+                .orElseThrow(); // Expect to be always retrieve a resource key for the root registry (registry key)
+        final Registry<?> registry = ctx.getSource().getServer().registryAccess().registry(registryKey)
+                .orElseThrow(() -> UNKNOWN_REGISTRY.create(registryKey.location()));
+
+        final long tagCount = registry.getTags().count();
+
+        ctx.getSource().sendSuccess(createMessage(
+                new TranslatableComponent("%s", new TextComponent(registryKey.location().toString()).withStyle(ChatFormatting.GOLD)),
```
pupnewfster: What is with the random TranslatableTextComponent of just `%s`? Why doesn't this just pass the created `TextComponent` directly and then change `createMessage` to take a `MutableComponent` instead of a `TranslatableComponent`

src/main/java/net/minecraftforge/server/command/TagsCommand.java
```diff
@@ -0,0 +1,266 @@
...
+            final String name,
+            final ResourceKey<Registry<T>> registryKey)
+    {
+        // Don't inline to avoid an unchecked cast warning due to raw types
+        final ResourceKey<?> key = ctx.getArgument(name, ResourceKey.class);
+        return key.cast(registryKey);
+    }
+
+    @SuppressWarnings("unchecked")
+    private static <I, O> O cast(final I input)
```
pupnewfster: Is the `I` being a generic actually needed? Or can this just be `private static <O> cast(final Object input)`

src/main/java/net/minecraftforge/registries/ForgeRegistryTagManager.java
SizableShrimp: Failure return or failure exception?

src/main/java/net/minecraftforge/server/command/TagsCommand.java
SizableShrimp: @sciwhiz12

src/main/java/net/minecraftforge/registries/ForgeRegistryTagManager.java
pupnewfster: Not sure of what the best would be. Maybe an optional or null? Might get a bit messy or even maybe an exception though that might not make sense. This is in part why I was mentioning it as I wonder if it makes sense to do something here, but I don't know *what*.

src/main/java/net/minecraftforge/registries/ForgeRegistryTagManager.java
SizableShrimp: This does not seem to work.

src/main/java/net/minecraftforge/server/command/TagsCommand.java
sciwhiz12: This was a remnant of when I refactored the common code between `listTags` and `listTagElements` into the separate helper method `createMessage`. As can be seen in the other methods, `TranslatableComponent` is used because of `%s / %s` for listing tag elements and `%s : %s` for listing tags which hold a certain element.

Additionally, the use of `TranslatableComponent` allows flexibility in changing the format of the command's output at either the programmer's discretion (by changing the string literal) or the final translator's discretion (through the translation keys), which would not otherwise be attainable if passing the text component directly.

patches/minecraft/net/minecraft/data/tags/TagsProvider.java.patch
```diff
@@ -9,21 +9,25 @@
...
++   @Deprecated
     protected TagsProvider(DataGenerator p_126546_, Registry<T> p_126547_) {
 +      this(p_126546_, p_126547_, "vanilla", null);
 +   }
-+   protected TagsProvider(DataGenerator p_126546_, Registry<T> p_126547_, String modId, @javax.annotation.Nullable net.minecraftforge.common.data.ExistingFileHelper existingFileHelper) {
++   protected TagsProvider(DataGenerator p_126546_, Registry<T> p_126547_, String modId, @org.jetbrains.annotations.Nullable net.minecraftforge.common.data.ExistingFileHelper existingFileHelper) {
 +      this(p_126546_, p_126547_, modId, existingFileHelper, null);
 +   }
-+   protected TagsProvider(DataGenerator p_126546_, Registry<T> p_126547_, String modId, @javax.annotation.Nullable net.minecraftforge.common.data.ExistingFileHelper existingFileHelper, @javax.annotation.Nullable String folder) {
++   protected TagsProvider(DataGenerator p_126546_, Registry<T> p_126547_, String modId, @org.jetbrains.annotations.Nullable net.minecraftforge.common.data.ExistingFileHelper existingFileHelper, @org.jetbrains.annotations.Nullable String folder) {
```
pupnewfster: Is this constructor with the folder being passed in just being kept for backward compat? As it seems to be unneeded now that `TagManager.getTagDir` exists and supports custom registries properly as of this PR? If so it should either be removed (as tags got broken so much and it isn't worth maintaining until 1.19 as people will have to adjust anyway) or it should be deprecated and the one without the folder passed in be preferred.

In fact I think what probably should happen is:
- remove this constructor (just have the one without the folder passed in)
- remove the class level `folder` variable as it isn't needed anymore
- Don't substring the `tags/` out of the returned folder from `getTagDir` and just pass it directly to the `ResourceType` constructor without prefixing it with `tags/` there in case any tag directories ever end up in a different subfolder

src/main/java/net/minecraftforge/registries/ForgeRegistries.java
```diff
@@ -127,7 +127,8 @@
         // Vanilla Dynamic
         public static final ResourceKey<Registry<Biome>> BIOMES = key("worldgen/biome");
 
-        //Forge
+        // Forge
+        // New forge-added registries should be populated in VanillaConnectionNetworkFilter#FORGE_ADDED_REGISTRIES
         public static final ResourceKey<Registry<DataSerializerEntry>> DATA_SERIALIZERS = key("data_serializers");
```
pupnewfster, line 136: Is there a reason this isn't in forge's domain? If not it may make sense to add a TODO for 1.19 to move it to forge's domain and remove having to add them to `FORGE_ADDED_REGISTRIES` and instead just check if the namespace is forge's? (Or would that set still make sense to have for purposes if forge ever decides to add any registries itself that supports tags)

src/main/java/net/minecraftforge/registries/RegistryManager.java
```diff
@@ -133,9 +144,28 @@ public String getName()
...
+        Registry<?> wrapper = forgeReg.getWrapper();
+        if (wrapper != null)
+            rootRegistry.register((ResourceKey<T>) forgeReg.getRegistryKey(), (T) wrapper, Lifecycle.experimental());
+    }
+
+    public static CompletableFuture<List<Throwable>> preNewRegistryEvent(final Executor executor,
+            final IModStateTransition.EventGenerator<? extends RegistryEvent.NewRegistry> eventGenerator)
+    {
+        return CompletableFuture.runAsync(() -> {
+            registerToRoot = true;
```
pupnewfster: This could use a brief comment explaining why it isn't just "always" true as it is not that obvious from looking at the code that it is because we don't want to register any of the registries that get staticly initialized in `GameData` to root.

src/main/java/net/minecraftforge/registries/tags/ITag.java
```diff
@@ -0,0 +1,30 @@
...
+
+import net.minecraft.tags.TagKey;
+
+import java.util.stream.Stream;
+
+/**
+ * A tag is a collection of elements with an identifying {@link TagKey tag key}.
+ * For Forge, these are bound on world load.
+ * Tags will always be empty until they are bound.
+ * A tag instance provided for a given {@link TagKey} from a given {@link ITagManager} will always return the same instance on future invocations.
```
pupnewfster, line 18: Is this statement of same instance also true for if a reload occurs? Either way it is probably worth clarifying it.

src/main/java/net/minecraftforge/registries/ForgeRegistryTagManager.java
pupnewfster: Ahh you are right, I missed `newTags` taking `tags` as a parameter of defaults initially when I suggested it

src/main/java/net/minecraftforge/server/command/TagsCommand.java
pupnewfster: A couple counter points to your mention of flexibility:
1. The programmer can always wrap it in a translatable component if they want to change it further in the future, that isn't a great reason to add an extra wrapper around it in the meantime.
2. This is a more valid reason, except that the key isn't exposed as a lang entry? So there is no way for the final translator to adjust it; at which point having it doesn't gain anything.

src/main/java/net/minecraftforge/registries/ForgeRegistries.java
SizableShrimp: I cannot comment on why this isn't in forge's domain. However, I would rather hardcode and provide a nice reference directly above the relevant code rather than rely on the namespace.

src/main/java/net/minecraftforge/registries/RegistryManager.java
SizableShrimp: This is a good point. The real reason is that we don't want to try and register vanilla-wrapped registries twice. As Registry is patched with all the forge vanilla-wrapped registries which queries GameData, and then returns it back to vanilla to be registered to the root registry. So the RegistryManager will try to register all the GameData registries to the root registry first, and then the vanilla Registry patches will attempt it as well and crash. We could maybe instead simplify this by always registering to root and changing the patches in Registry to return the instance but *not* register it to the root registry. Thoughts?

src/main/java/net/minecraftforge/registries/tags/ITag.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+public interface ITag<V> extends Iterable<V>
```
pupnewfster, line 22: I feel like it would be useful to expose `size` for how many elements are in the tag and maybe even `get(int)` for getting a specific element of the tag if you want to do an indexed loop.

src/main/java/net/minecraftforge/registries/ForgeRegistryTagManager.java
```diff
@@ -0,0 +1,133 @@
...
+    @Override
+    public TagKey<V> createTagKey(@NotNull ResourceLocation location)
+    {
+        Objects.requireNonNull(location);
+        return TagKey.create(this.owner.getRegistryKey(), location);
+    }
+
+    @NotNull
+    @Override
+    public TagKey<V> createOptionalTagKey(@NotNull ResourceLocation location, @NotNull Set<? extends Supplier<V>> defaults)
```
pupnewfster, line 121: One thing that appears to be missing is a way to create an optional tag key *before* the registry may exist. The use of this is for any mods that may want to statically define their tag keys (almost all do this) AND have some they want to declare that are optional for custom registries (where the registry may not exist yet).

For non optional tags this is possible by just manually calling `TagKey#create` using a resource key that will represent the registry once it is created, but for optional tags I am not sure of the best way to provide a helper method to addons to allow them to statically create optional tags for the case of the registry not having been created yet.

src/main/java/net/minecraftforge/registries/tags/ITag.java
SizableShrimp: Same forge registry instance means same tag manager and same tags. Reload doesn't recreate these registries afaik.

src/main/java/net/minecraftforge/registries/ForgeRegistryTagManager.java
SizableShrimp: I could probably provide some sort of utility on DeferredRegister that stores its own set of this data, then invokes it on the ITagManager when it is available.

src/main/java/net/minecraftforge/registries/tags/ITag.java
SizableShrimp: No guarantee is made that the tag or reverse tag are ordered. This is why Stream is used; no inherent order is guaranteed and no collection view is necessary. Providing a size without an indexed lookup seems pretty useless to me.

pupnewfster: Size still has a few benefits, such as to check if it is empty. I suppose just exposing a helper for if it is empty or not could work, but I figure size gives more flexibility to the modder for example if they want to compare the contents of two tags and want to star by validating they both have the same number of elements.

src/main/java/net/minecraftforge/registries/ForgeRegistryTagManager.java
SizableShrimp: Implemented in latest push.

src/main/java/net/minecraftforge/registries/tags/ITag.java
SizableShrimp: Alright.

patches/minecraft/net/minecraft/data/tags/TagsProvider.java.patch
```diff
@@ -1,32 +1,28 @@
...
     protected final Map<ResourceLocation, Tag.Builder> f_126543_ = Maps.newLinkedHashMap();
 +   protected final String modId;
-+   protected final String folder;
 +   protected final net.minecraftforge.common.data.ExistingFileHelper existingFileHelper;
 +   private final net.minecraftforge.common.data.ExistingFileHelper.IResourceType resourceType;
  
-+   /**@deprecated Forge: Use the ModID version**/ @Deprecated
++   /**
++    * @see #TagsProvider(DataGenerator, Registry, String, net.minecraftforge.common.data.ExistingFileHelper)
++    * @see #TagsProvider(DataGenerator, Registry, String, net.minecraftforge.common.data.ExistingFileHelper, String)
```
pupnewfster: This second "see" should be removed as the constructor for it was removed.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1117,8 +1117,14 @@ public static int getSerializerId(EntityDataSerializer<?> serializer, CrudeIncre
         int id = vanilla.getId(serializer);
         if (id < 0)
         {
-            DataSerializerEntry entry = serializerEntries.get(serializer);
-            if (entry != null) id = ((ForgeRegistry<DataSerializerEntry>)ForgeRegistries.DATA_SERIALIZERS).getID(entry);
+            if (serializerEntries == null)
+                serializerEntries = GameData.getSerializerMap();
```
pupnewfster, line 1123: I feel like a brief comment saying why this might be null is worthwhile just as it may not be immediately obvious.

patches/minecraft/net/minecraft/data/tags/TagsProvider.java.patch
```diff
@@ -1,32 +1,27 @@
...
     protected final Map<ResourceLocation, Tag.Builder> f_126543_ = Maps.newLinkedHashMap();
 +   protected final String modId;
-+   protected final String folder;
 +   protected final net.minecraftforge.common.data.ExistingFileHelper existingFileHelper;
 +   private final net.minecraftforge.common.data.ExistingFileHelper.IResourceType resourceType;
  
-+   /**@deprecated Forge: Use the ModID version**/ @Deprecated
++   /**
++    * @see #TagsProvider(DataGenerator, Registry, String, net.minecraftforge.common.data.ExistingFileHelper)
++    * @deprecated Forge: Use one of the mod id variants
```
pupnewfster: Didn't notice this wording when mentioning to remove the second see, but given there is only one modid variant this should probably be rephrased.

src/main/java/net/minecraftforge/registries/RegistryManager.java
```diff
@@ -136,6 +155,47 @@ public String getName()
         return getRegistry(name);
     }
 
+    @SuppressWarnings("unchecked")
+    private static <V extends IForgeRegistryEntry<V>> void registerToRootRegistry(ForgeRegistry<V> forgeReg)
+    {
+        WritableRegistry<Registry<V>> registry = (WritableRegistry<Registry<V>>) Registry.REGISTRY;
+        Registry<V> wrapper = forgeReg.getWrapper();
+        if (wrapper != null)
+            registry.register(forgeReg.getRegistryKey(), wrapper, Lifecycle.experimental());
```
pupnewfster, line 163: Do we want forge registries as experimental like this or stable like vanilla?

src/main/java/net/minecraftforge/registries/RegistryObject.java
```diff
@@ -97,6 +133,36 @@ public void accept(Predicate<ResourceLocation> pred)
...
+            private IForgeRegistry<V> registry;
+
+            @Override
+            public void accept(Predicate<ResourceLocation> pred)
+            {
+                if (registry == null)
+                {
+                    this.registry = RegistryManager.ACTIVE.getRegistry(registryName);
+                    if (registry == null)
+                        throw new IllegalStateException("Unable to find registry with key " + registryName + " for mod \"" + modid + "\". Check the 'caused by' to see futher stack.", callerStack);
```
pupnewfster: `futher` -> `further`

src/main/java/net/minecraftforge/registries/tags/IReverseTag.java
```diff
@@ -0,0 +1,27 @@
...
+import net.minecraft.core.Holder;
+import net.minecraft.tags.TagKey;
+
+import java.util.stream.Stream;
+
+/**
+ * A reverse tag is an object aware of what tags it is contained in.
+ * {@link Holder}s implement this interface.
+ */
+public interface IReverseTag<V>
```
pupnewfster, line 19: It is probably worth clarifying in this javadoc whether this instance persists or not. I am guessing it doesn't but given it is mentioned in `ITag` some modders may make incorrect assumptions.

src/main/java/net/minecraftforge/registries/tags/ITag.java
```diff
@@ -0,0 +1,34 @@
...
```
pupnewfster, line 22: I feel like we may want to expose the equivalent of the old `getRandomElement`. Tags used to expose this and vanilla's `HolderSet` still has this.

src/main/java/net/minecraftforge/registries/tags/ITagManager.java
```diff
@@ -0,0 +1,85 @@
...
+    @NotNull TagKey<V> createTagKey(@NotNull ResourceLocation location);
+
+    /**
+     * Creates a tag key that will use the set of defaults if the tag is not loaded from any datapacks.
+     * Useful on the client side when a server may not provide a specific tag.
+     *
+     * @see #createTagKey(ResourceLocation)
+     * @see #addOptionalTagDefaults(TagKey, Set)
+     */
+    @NotNull TagKey<V> createOptionalTagKey(@NotNull ResourceLocation location, @NotNull Set<? extends Supplier<V>> defaults);
```
pupnewfster, line 80: It might be worth also linking a see or mentioning the way via `DeferredRegister` (especially for use with custom registries.

src/main/java/net/minecraftforge/registries/NamespacedHolderHelper.java
```diff
@@ -184,8 +207,22 @@ void bindTags(Map<TagKey<T>, List<Holder<T>>> newTags)
 
         Map<TagKey<T>, HolderSet.Named<T>> tmpTags = new IdentityHashMap<>(this.tags);
         newTags.forEach((k, v) -> tmpTags.computeIfAbsent(k, this::createTag).bind(v));
+        this.optionalTags.asMap().forEach((k, v) -> {
+            if (!tmpTags.containsKey(k))
```
pupnewfster: I feel like this might be slightly cleaner if it were to use `tmpTags.computeIfAbsent`

src/main/java/net/minecraftforge/registries/ForgeRegistryTagManager.java
```diff
@@ -0,0 +1,142 @@
...
+
+    ForgeRegistryTagManager(ForgeRegistry<V> owner)
+    {
+        this.owner = owner;
+    }
+
+    void bind(Map<TagKey<V>, HolderSet.Named<V>> holderTags)
+    {
+        IdentityHashMap<TagKey<V>, ITag<V>> newTags = new IdentityHashMap<>(this.tags);
+        Set<TagKey<V>> bound = new HashSet<>();
```
pupnewfster: Is there a reason each element is added to a set tracking the bound tags instead of just using `holderTags#keySet` for comparing against the previously bound tags?

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -51,31 +61,62 @@
...
+    /**
+     * DeferredRegister factory for custom registries that are made during the {@link RegistryEvent.NewRegistry} event
+     * or with {@link #makeRegistry(Class, Supplier)}.
+     *
+     * @param registryName The name of the registry, should include namespace. May come from another DeferredRegister through {@link #getRegistryName()}.
+     * @param modid The namespace for all objects registered to this DeferredRegister
+     */
+    public static <B extends IForgeRegistryEntry<B>> DeferredRegister<B> create(ResourceLocation registryName, String modid)
+    {
+        return new DeferredRegister<B>(registryName, null, modid);
```
pupnewfster: Explicitly declaring the generic as `B` here seems to be unneeded.

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -51,31 +61,62 @@
...
+    /**
+     * DeferredRegister factory for custom registries that are made during the {@link RegistryEvent.NewRegistry} event
+     * or with {@link #makeRegistry(Class, Supplier)}.
+     *
+     * @param key The key of the registry to reference
+     * @param modid The namespace for all objects registered to this DeferredRegister
+     */
+    public static <B extends IForgeRegistryEntry<B>> DeferredRegister<B> create(ResourceKey<? extends Registry<B>> key, String modid)
+    {
+        return new DeferredRegister<B>(key.location(), null, modid);
```
pupnewfster: Explicitly declaring the generic as `B` here seems to be unneeded.

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -118,15 +162,72 @@ else if (this.superType != null)
...
+        return makeRegistry(this.registryName, base, sup);
+    }
+
+    /**
+     * Creates a tag key based on the current modid and provided path as the location and the registry name linked to this DeferredRegister.
+     *
+     * @see #createOptionalTagKey(String, Set)
+     */
+    @NotNull
+    public TagKey<T> createTagKey(@NotNull String path)
```
pupnewfster, line 199: I think instead of a string this might make more sense to be a `ResourceLocation`? While I can see why it may be useful to have this be based on the deferred register's modid, I feel like especially for custom registries (which this and `createOptionalTagKey` are more likely to be used for as you can't easily static initialize optional tags for registries that aren't static init otherwise), it would be overly annoying to have to create a deferred register in another mod's domain *purely* for purposes of creating tags in their domain.

For example if an addon wanted to add an optional tag that has defaults for one of Mekanism's registries (say gases), instead of just being able to specify they may want the tag in our domain or in some other domain for another addon they are interacting with. They would have to create a second deferred register using our namespace in order to be able to create tags in it, rather than just being able to specify in the tag key creation they want it to be in our domain.

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -162,17 +263,59 @@ public void handleEvent(RegistryEvent.Register<?> event) {
...
     {
-        if (this.type == null && this.registryFactory == null)
+        IForgeRegistry<T> storedType = this.getForgeRegistry();
+        if (storedType == null && this.registryFactory == null)
         {
             //If there is no type yet and we don't have a registry factory, attempt to capture the registry
             //Note: This will only ever get run on the first registry event, as after that time,
             // the type will no longer be null. This is needed here rather than during the NewRegistry event
             // to ensure that mods can properly use deferred registers for custom registries added by other mods
             captureRegistry();
```
pupnewfster, line 366: This needs to be adjusted as `captureRegistry` captures the registry to `this.type`, but as references lower in this method use `storedType` without first updating storedType to point to the updated `this.type` after capturing is done, then `storedType` is still going to be null.

src/main/java/net/minecraftforge/registries/tags/IReverseTag.java
SizableShrimp: Persists relative to what?

pupnewfster: Where it is gotten from or through reloads? I don't know, that is why I presume the answer is no

src/main/java/net/minecraftforge/registries/RegistryManager.java
SizableShrimp: Most of the vanilla registries and all of the forge registries inside of `Registry` use `Lifecycle.experimental()` when registering to the root registry. That's what I based this off of. Only biome sources and chunk generators are registered as stable lifecycles to the root registry. (I also don't really get the purpose of lifecycles.)

src/main/java/net/minecraftforge/registries/tags/IReverseTag.java
SizableShrimp: I think it's easiest to specify that reverse tags make no guarantee to their persistence. Mods should be looking up a reverse tag every time they need one rather than storing them.

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -118,15 +162,110 @@ else if (this.superType != null)
...
+     * To use the current modid as the namespace, use {@link #createTagKey(String)}.
+     *
+     * @see #createTagKey(String)
+     * @see #createOptionalTagKey(ResourceLocation, Set)
+     */
+    @NotNull
+    public TagKey<T> createTagKey(@NotNull ResourceLocation location)
+    {
+        if (this.registryName == null)
+            throw new IllegalStateException("The registry name was not set, cannot create a tag key");
```
sciwhiz12, line 218: The conditions of when a `IllegalStateException` is thrown should be documented.

pupnewfster (approved):

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -118,15 +162,120 @@ else if (this.superType != null)
...
+     * Only used for custom registries to fill the forge registry held in this DeferredRegister.
+     *
+     * Calls {@link RegistryBuilder#setName} and {@link RegistryBuilder#setType} automatically.
+     *
+     * @param base The base type to use in the created {@link IForgeRegistry}
+     * @param sup Supplier of a RegistryBuilder that initializes a {@link IForgeRegistry} during the {@link NewRegistryEvent} event
+     * @return A supplier of the {@link IForgeRegistry} created by the builder.
+     * Will always return null until after the {@link NewRegistryEvent} event fires.
+     */
+    public Supplier<IForgeRegistry<T>> makeRegistry(final Class<T> base, final Supplier<RegistryBuilder<T>> sup) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    public Supplier<IForgeRegistry<T>> makeRegistry(final Class<T> base, final Supplier<RegistryBuilder<T>> sup)
    {
```

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -162,17 +311,60 @@ public void handleEvent(RegistryEvent.Register<?> event) {
...
+    /**
+     * @return The registry name stored in this deferred register. Useful for creating new deferred registers based on an existing one.
+     */
+    @Nullable
+    public ResourceLocation getRegistryName()
+    {
+        return this.registryName;
+    }
+
+    private Supplier<IForgeRegistry<T>> makeRegistry(final ResourceLocation registryName, final Class<T> superType, final Supplier<RegistryBuilder<T>> sup) {
```
sciwhiz12, line 325: Formatting: braces on newlines.
```suggestion
    private Supplier<IForgeRegistry<T>> makeRegistry(final ResourceLocation registryName, final Class<T> superType, final Supplier<RegistryBuilder<T>> sup)
    {
```

src/main/java/net/minecraftforge/registries/NewRegistryEvent.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Register new registries when you receive this event through {@link RegistryBuilder} and {@link #create(RegistryBuilder)}.
+ */
+public class NewRegistryEvent extends Event implements IModBusEvent
+{
+    private final List<RegistryData<?>> registries = new ArrayList<>();
+
+    public NewRegistryEvent() {}
```
sciwhiz12, line 28: Formatting: braces on newlines.
```suggestion
    public NewRegistryEvent()
    {
    }
```

src/main/java/net/minecraftforge/registries/NewRegistryEvent.java
```diff
@@ -0,0 +1,98 @@
...
+                    data.onFill.accept(registry);
+            }
+        }
+    }
+
+    private record RegistryData<V extends IForgeRegistryEntry<V>>(
+            RegistryBuilder<V> builder,
+            RegistryHolder<V> registryHolder,
+            Consumer<IForgeRegistry<V>> onFill
+    ) {}
```
sciwhiz12, line 85: Formatting: braces on newlines.
```suggestion
    )
    {
    }
```

src/main/java/net/minecraftforge/registries/RegistryBuilder.java
```diff
@@ -227,7 +216,24 @@
...
+    {
+        // Tag system heavily relies on Registry<?> objects, so we need a wrapper for this registry to take advantage
+        this.hasWrapper();
+        return this;
+    }
+
+    /**
+     * Modders: Use {@link NewRegistryEvent#create(RegistryBuilder)} instead
+     */
+    IForgeRegistry<T> create()
```
sciwhiz12, line 236: Should this be `@Deprecated`, to discourage developers from using this method?

src/test/java/net/minecraftforge/debug/DuplicateOptionalTagTest.java
```diff
@@ -41,46 +40,35 @@
     private static final Set<Supplier<Block>> TAG_A_DEFAULTS = Set.of(Blocks.BEDROCK.delegate);
     private static final Set<Supplier<Block>> TAG_B_DEFAULTS = Set.of(Blocks.WHITE_WOOL.delegate);
 
-    // TODO-PATCHING: fix or remove this accordingly with optional tags feature (specifically, default values for opt. tags)
-//    private static final Tags.IOptionalNamedTag<Block> TAG_A = ForgeTagHandler.createOptionalTag(ForgeRegistries.BLOCKS, TAG_NAME,
-//            TAG_A_DEFAULTS);
-//    private static final Tags.IOptionalNamedTag<Block> TAG_B = ForgeTagHandler.createOptionalTag(ForgeRegistries.BLOCKS, TAG_NAME,
-//            TAG_B_DEFAULTS);
+   private static final TagKey<Block> TAG_A = ForgeRegistries.BLOCKS.tags().createOptionalTagKey(TAG_NAME, TAG_A_DEFAULTS);
+   private static final TagKey<Block> TAG_B = ForgeRegistries.BLOCKS.tags().createOptionalTagKey(TAG_NAME, TAG_B_DEFAULTS);
```
sciwhiz12: Formatting: indentation must be four spaces in Forge classes.
```suggestion
    private static final TagKey<Block> TAG_A = ForgeRegistries.BLOCKS.tags().createOptionalTagKey(TAG_NAME, TAG_A_DEFAULTS);
    private static final TagKey<Block> TAG_B = ForgeRegistries.BLOCKS.tags().createOptionalTagKey(TAG_NAME, TAG_B_DEFAULTS);
```

src/test/java/net/minecraftforge/debug/DuplicateOptionalTagTest.java
```diff
@@ -41,46 +40,35 @@
...
 
     public DuplicateOptionalTagTest()
     {
-//        MinecraftForge.EVENT_BUS.addListener(this::onServerStarted);
+       MinecraftForge.EVENT_BUS.addListener(this::onServerStarted);
```
sciwhiz12: Formatting: indentation must be four spaces in Forge classes.
```suggestion
        MinecraftForge.EVENT_BUS.addListener(this::onServerStarted);
```

src/test/java/net/minecraftforge/debug/DuplicateOptionalTagTest.java
```diff
@@ -41,46 +40,35 @@
...
+       if (!tagAValues.equals(expected))
+       {
+           IllegalStateException e = new IllegalStateException("Optional tag values do not match!");
+           LOGGER.error("Values of the optional tag do not match the expected union of their defaults: expected {}, got {}", expected, tagAValues, e);
+           return;
+       }
+
+       LOGGER.info("Optional tag instances match each other and the expected union of their defaults");
+   }
 }
```
sciwhiz12: Formatting: indentation must be four spaces in Forge classes.
```suggestion
    private void onServerStarted(ServerStartedEvent event)
    {
        Set<Block> tagAValues = ForgeRegistries.BLOCKS.tags().getTag(TAG_A).stream().collect(Collectors.toUnmodifiableSet());
        Set<Block> tagBValues = ForgeRegistries.BLOCKS.tags().getTag(TAG_B).stream().collect(Collectors.toUnmodifiableSet());

        if (!tagAValues.equals(tagBValues))
        {
            LOGGER.error("Values of both optional tag instances are not the same: first instance: {}, second instance: {}", tagAValues, tagBValues);
            return;
        }

        final Set<Block> expected = Sets.union(TAG_A_DEFAULTS, TAG_B_DEFAULTS).stream()
                .map(Supplier::get)
                .collect(Collectors.toUnmodifiableSet());
        if (!tagAValues.equals(expected))
        {
            IllegalStateException e = new IllegalStateException("Optional tag values do not match!");
            LOGGER.error("Values of the optional tag do not match the expected union of their defaults: expected {}, got {}", expected, tagAValues, e);
            return;
        }

        LOGGER.info("Optional tag instances match each other and the expected union of their defaults");
    }
}
```

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/registries/RegistryBuilder.java
SizableShrimp: It's already package private. The documentation is a nicety.

src/main/java/net/minecraftforge/registries/NewRegistryEvent.java
SizableShrimp: Lex said {} was fine so that's what I'm going with.

src/main/java/net/minecraftforge/registries/NewRegistryEvent.java
SizableShrimp: Lex said {} was fine so that's what I'm going with.

PR #8499, opened 2022-03-05
Fix swim speed attribute

Frostygames0:
This PR fixes the swim speed attribute, which in 1.18.2 multiplies the wrong variable,
and because of this it does not work.


TheCurle:
God damn local variable shifts.

Nice catch, thanks for the fix.

PR #8500, opened 2022-03-06
Add missing module exports arg to server arguments list

sciwhiz12:
This PR adds the missing JVM argument line `--add-exports jdk.naming.dns/com.sun.jndi.dns=java.naming` to the server arguments list (`server_files/args.txt`), matching the JVM arguments list for the client profile as defined in the buildscript.

To reduce future similar errors, a comment was also added to the aforementioned buildscript section to remind editors that changes to those lines should be mirrored to the server arguments list.

This PR was prompted by @Sm0keySa1m0n's discovery and subsequent report of this issue's existence on the dedicated server on the official Forge Project discord, and requested by @SizableShrimp.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -436,6 +436,7 @@ def sharedFmlonlyForge = { Project prj ->
             run.jvmArgs '-p', prj.configurations.moduleonly.files.collect { it.path }.join(File.pathSeparator)
         }
         run.jvmArgs '--add-modules', 'ALL-MODULE-PATH'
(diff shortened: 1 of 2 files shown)


SizableShrimp (approved):

LexManos:
I thought the real solution was to advise server hosts to update their java version to one that doesnt have this issue? As it was an issue in openjdk. Either way, as long as this doesn't crash when redundant, then this is fine.

PR #8501, opened 2022-03-06
DFU - The Third

marchermans:
### Third version of: #8242, #8410 and #8412
This version of the PR finally fixes the NPEs that where seen previously during the initial attempt at this.
I finally tracked them down to the concurrent calculation of the rules while forge rebuilds the types in the schema. So @cpw was right.....again.
This version now also includes a test mod 👍, not sure why I did not include it in the first one.

### Modifiable DFU

#### Goals

#### Implementation

#### Future:

#### Usage:

##### Example:
```java
{

    @SubscribeEvent
    {
            );
        }
    }
}
```

### Custom Fixers
Besides the modifiable DFU system this third version of the DFU PR now also adds support for registering custom Fixers to schemas.

This way modders can update their own BlockEntities etc between Minecraft Versions.

#### Goals
- Provide a way for modders to register their own fixes to schemas Mojang already creates.

#### Implementation
A second DFU mod state transition was added which fires registration events (one per schema and `ModContainer` this time) on the mod bus.
This event can then be used to selectively add one or more fixers to the schemas fixer list and bakes them into the vanilla fixes which are applied for that version.

#### Example:
```java
{

    @SubscribeEvent
    private void onDataFixerRegistration(RegisterFixesEvent registerFixesEvent) {
        LOGGER.info("Registering DFU Fixer " + registerFixesEvent.getVersion());

        registerFixesEvent.addFixer(schema -> new DataFix(schema, false) {
            @Override
            protected TypeRewriteRule makeRule() {
                LOGGER.warn("Creating test rewrite rule for schema version: " + schema.getVersionKey());
                return TypeRewriteRule.nop();
            }
        });
    }
}
```

### Todo:
- [X] Better testing
- [X] Formatting


src/main/java/net/minecraftforge/common/ForgeSerialModStateTransition.java
```diff
@@ -0,0 +1,47 @@
+package net.minecraftforge.common;
```
sciwhiz12, line 6: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/common/ForgeSerialModStateTransition.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
+
+public record ForgeSerialModStateTransition<T extends Event & IModBusEvent>(
+  Supplier<Stream<EventGenerator<?>>> eventStream,
+  BiFunction<Executor, ? extends EventGenerator<T>, CompletableFuture<List<Throwable>>> preDispatchHook,
+  BiFunction<Executor, ? extends EventGenerator<T>, CompletableFuture<List<Throwable>>> postDispatchHook,
+  BiFunction<Executor, CompletableFuture<List<Throwable>>, CompletableFuture<List<Throwable>>> finalActivityGenerator
+  ) implements IModStateTransition {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
  ) implements IModStateTransition
  {
```

src/main/java/net/minecraftforge/common/ForgeSerialModStateTransition.java
```diff
@@ -0,0 +1,47 @@
...
+    public static <T extends Event & IModBusEvent> ForgeSerialModStateTransition<T> of(Supplier<Stream<EventGenerator<?>>> eventStream) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    public static <T extends Event & IModBusEvent> ForgeSerialModStateTransition<T> of(Supplier<Stream<EventGenerator<?>>> eventStream)
    {
```

src/main/java/net/minecraftforge/common/ForgeSerialModStateTransition.java
```diff
@@ -0,0 +1,47 @@
...
+        return new ForgeSerialModStateTransition<T>(eventStream,
+          (t, f) -> CompletableFuture.completedFuture(Collections.emptyList()),
+          (t, f) -> CompletableFuture.completedFuture(Collections.emptyList()),
+          (e, prev) -> prev.thenApplyAsync(
+            Function.identity(), e));
+    }
+
+    @Override
+    public Supplier<Stream<EventGenerator<?>>> eventFunctionStream() {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    public Supplier<Stream<EventGenerator<?>>> eventFunctionStream()
    {
```

src/main/java/net/minecraftforge/common/ForgeSerialModStateTransition.java
```diff
@@ -0,0 +1,47 @@
...
+        return eventStream;
+    }
+
+    @Override
+    public ThreadSelector threadSelector() {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    public ThreadSelector threadSelector()
    {
```

src/main/java/net/minecraftforge/common/ForgeSerialModStateTransition.java
```diff
@@ -0,0 +1,47 @@
...
+        return ThreadSelector.SYNC;
+    }
+
+    // These never progress the ModLoadingStage..
+    @Override
+    public BiFunction<ModLoadingStage, Throwable, ModLoadingStage> nextModLoadingStage() {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    public BiFunction<ModLoadingStage, Throwable, ModLoadingStage> nextModLoadingStage()
    {
```

src/main/java/net/minecraftforge/common/ForgeSerialModStateTransition.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
+
+public record ForgeSerialModStateTransition<T extends Event & IModBusEvent>(
```
sciwhiz12, line 24: What is the reason for this class being refactored/moved to outside of `ForgeStatesProvider` and renamed? Also, this should probably be a package-private class.

src/main/java/net/minecraftforge/datafix/ForgeDataFixDelegate.java
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
```
sciwhiz12: Outdated license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/datafix/ForgeDataFixDelegate.java
```diff
@@ -0,0 +1,123 @@
...
+ * cached in {@link DataFix} and the recompute of schema will need to invalidate this cached
+ * type rewrite rule.
+ */
+class ForgeDataFixDelegate extends DataFix
+{
+
+    private static final Method makeRuleMethod = ObfuscationReflectionHelper.findMethod(
+      DataFix.class,
+      "makeRule"
+    );
```
sciwhiz12: I personally think these can be on one line.

src/main/java/net/minecraftforge/datafix/ForgeDataFixDelegate.java
```diff
@@ -0,0 +1,123 @@
...
+    public TypeRewriteRule getRule()
+    {
+        //Check if OUR cached rule is available.
+        if (this.wrappedRule == null)
+        {
+            //Not available get a new one.
+            this.wrappedRule = rebuildRuleReflectively();
+        }
+
+        //Cached version available, return it.
```
sciwhiz12: While I admire the dedication to documentation, at the very least these two comments are unneeded. (One could also argue the first one is also unneeded, but I'll defer it to you.)

src/main/java/net/minecraftforge/datafix/ForgeDataFixerBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.datafix;
```
sciwhiz12: Outdated license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/datafix/ForgeDataFixDelegate.java
```diff
@@ -0,0 +1,123 @@
...
+            {
+                return typeRewriteRule;
+            }
+
+            //WTF Who put this here? Who is returning a none TypeRewriteRule from a method that should return it.!
+            throw new IllegalStateException("Failed to get a rewrite rule. The returned object was not a TypeRewriteRule!"); //Should never be reached.
+        }
+        catch (IllegalAccessException | InvocationTargetException e)
+        {
+            throw new IllegalStateException("Could not get or invoke makeRule() Method on data fixer!", e);
```
sciwhiz12, line 79: `RuntimeException` is probably the more 'correct' exception here:
> Signals that a method has been invoked at an illegal or inappropriate time. In other words, the Java environment or Java application is not in an appropriate state for the requested operation.
> ~ _[`java.lang.IllegalStateException`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/IllegalStateException.html), Java SE 17 & JDK 17_

src/main/java/net/minecraftforge/datafix/ForgeDataFixerBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+ * Some methods are carbon copies of the DFU counterpart, but are copied so that we can control the data.
+ */
+public class ForgeDataFixerBuilder extends DataFixerBuilder
+{
+    private static final Logger LOGGER = LogManager.getLogger();
+
+    private final int                              dataVersion;
+    private final List<ForgeSchema>                allSchemas    = new ArrayList<>();
+    private final Int2ObjectSortedMap<ForgeSchema> schemas       = new Int2ObjectAVLTreeMap<>();
+    private final List<ForgeDataFixDelegate>       globalList    = Lists.newArrayList();
```
sciwhiz12: Use `new ArrayList<>()` instead.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * The purpose of this custom data fixer builder is to gain full control over the datafixer that is in play at runtime.
+ * <p>
+ * This means that schemas become adaptable and that the ruleset can be dynamically rebaked to adapt to the changes mod authors make to the schemas.
+ * <p>
+ * Some methods are carbon copies of the DFU counterpart, but are copied so that we can control the data.
+ */
+public class ForgeDataFixerBuilder extends DataFixerBuilder
+{
+    private static final Logger LOGGER = LogManager.getLogger();
```
sciwhiz12: Use `LogUtils` instead.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerCustomFixersModStateTransition.java
```diff
@@ -0,0 +1,220 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.datafix;
```
sciwhiz12: Outdated license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerCustomFixersModStateTransition.java
```diff
@@ -0,0 +1,220 @@
...
+public class ForgeDataFixerCustomFixersModStateTransition
+{
+    private static final Logger LOGGER = LogUtils.getLogger();
+
+    /**
+     * Builds a new mod state transition which can be used to fire the appropriate fixer registration events on the mod bus.
+     *
+     * @return The mod state transition used to add DFU fixers.
+     */
+    public static IModStateTransition get() {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    public static IModStateTransition get()
    {
```

src/main/java/net/minecraftforge/datafix/ForgeDataFixerCustomFixersModStateTransition.java
```diff
@@ -0,0 +1,220 @@
...
+     *
+     * @param executor The executor to execute this task on.
+     * @param eventGenerator The event generator that is about to be executed.
+     * @return A future that indicates success or failure.
+     */
+    private static CompletableFuture<List<Throwable>> preDispatchFixerEvents(
+      final Executor executor, final IModStateTransition.EventGenerator<RegisterFixesEvent> eventGenerator
+    )
+    {
+        if (!(eventGenerator instanceof InternalEventGenerator internalEventGenerator)) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
        if (!(eventGenerator instanceof InternalEventGenerator internalEventGenerator))
        {
```

src/main/java/net/minecraftforge/datafix/ForgeDataFixerCustomFixersModStateTransition.java
```diff
@@ -0,0 +1,220 @@
...
+     * This allows us to collect the data from the events and trigger the given schema to rebuild its type map.
+     *
+     * @param executor The executor to execute the task on.
+     * @param eventGenerator The event generator for the event (and thus the schema in question).
+     * @return The future that indicates success or failure.
+     */
+    private static CompletableFuture<List<Throwable>> postDispatchFixerEvents(
+      final Executor executor, final IModStateTransition.EventGenerator<RegisterFixesEvent> eventGenerator
+    ) {
+        if (!(eventGenerator instanceof InternalEventGenerator internalEventGenerator)) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    )
    {
        if (!(eventGenerator instanceof InternalEventGenerator internalEventGenerator))
        {
```

src/main/java/net/minecraftforge/datafix/ForgeDataFixerCustomFixersModStateTransition.java
```diff
@@ -0,0 +1,220 @@
...
+     * As such we can now safely reconfigure the DFU wrapper so that the new schema configurations are picked
+     * up by the system.
+     *
+     * @param executor The executor to execute the task on.
+     * @param listCompletableFuture The list of previous failures.
+     * @return The future which indicates success or failure.
+     */
+    private static CompletableFuture<List<Throwable>> rebuildRulesAfterSchemaEvents(
+      final Executor executor, final CompletableFuture<List<Throwable>> listCompletableFuture
+    ) {
```
sciwhiz12:
```suggestion
    )
    {
```

src/main/java/net/minecraftforge/datafix/ForgeDataFixerCustomFixersModStateTransition.java
```diff
@@ -0,0 +1,220 @@
...
+            }
+        }, executor);
+    }
+
+    private static class InternalEventGenerator implements IModStateTransition.EventGenerator<RegisterFixesEvent> {
+
+        private static final Logger LOGGER = LogUtils.getLogger();
+
+        private final ForgeSchema                           schema;
+        private final Map<ModContainer, RegisterFixesEvent> eventPerContainerMap = Maps.newConcurrentMap();
```
sciwhiz12: What is the reason for the use of this instead of `new ConcurrentHashMap<>()`?

src/main/java/net/minecraftforge/datafix/ForgeDataFixerCustomFixersModStateTransition.java
```diff
@@ -0,0 +1,220 @@
...
+                    //Do not do this on the mod executor!
+                    //This will crash the game.
+                    //Rebuild the DFU wrapper on the bootstrap executor
+                    forgeDataFixerDelegateHandler.rebuildFixer(Util.bootstrapExecutor(), true);
+                }
+            }
+        }, executor);
+    }
+
+    private static class InternalEventGenerator implements IModStateTransition.EventGenerator<RegisterFixesEvent> {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    private static class InternalEventGenerator implements IModStateTransition.EventGenerator<RegisterFixesEvent>
    {
```

src/main/java/net/minecraftforge/datafix/ForgeDataFixerCustomFixersModStateTransition.java
```diff
@@ -0,0 +1,220 @@
...
+
+        private static final Logger LOGGER = LogUtils.getLogger();
+
+        private final ForgeSchema                           schema;
+        private final Map<ModContainer, RegisterFixesEvent> eventPerContainerMap = Maps.newConcurrentMap();
+
+        private InternalEventGenerator(final ForgeSchema schema) {this.schema = schema;}
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
        private InternalEventGenerator(final ForgeSchema schema)
        {
        	this.schema = schema
      	}
```

src/main/java/net/minecraftforge/datafix/ForgeDataFixerCustomFixersModStateTransition.java
```diff
@@ -0,0 +1,220 @@
...
+
+        @Override
+        public RegisterFixesEvent apply(final ModContainer modContainer) {
+            return eventPerContainerMap.computeIfAbsent(modContainer, this::createEvent);
+        }
+
+        private RegisterFixesEvent createEvent(final ModContainer modContainer) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
        public RegisterFixesEvent apply(final ModContainer modContainer)
        {
            return eventPerContainerMap.computeIfAbsent(modContainer, this::createEvent);
        }

        private RegisterFixesEvent createEvent(final ModContainer modContainer)
        {
```

src/main/java/net/minecraftforge/datafix/ForgeDataFixerCustomFixersModStateTransition.java
```diff
@@ -0,0 +1,220 @@
...
+        private static final Logger LOGGER = LogUtils.getLogger();
+
+        private final ForgeSchema                           schema;
+        private final Map<ModContainer, RegisterFixesEvent> eventPerContainerMap = Maps.newConcurrentMap();
+
+        private InternalEventGenerator(final ForgeSchema schema) {this.schema = schema;}
+
+        @Override
+        public RegisterFixesEvent apply(final ModContainer modContainer) {
+            return eventPerContainerMap.computeIfAbsent(modContainer, this::createEvent);
```
sciwhiz12, line 202: Is this event fired multiple times for a mod container?

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
```diff
@@ -0,0 +1,244 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.datafix;
```
sciwhiz12: Outdated license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
```diff
@@ -0,0 +1,244 @@
...
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+
+/**
+ * A delegate handler which can rebuild a data fixer on the fly.
+ */
+class ForgeDataFixerDelegateHandler extends DataFixerUpper
+{
+    private static final Logger LOGGER = LogManager.getLogger();
```
sciwhiz12: Use `LogUtils` instead.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
```diff
@@ -0,0 +1,244 @@
...
+                }
+
+                final ForgeDataFixDelegate delegate = new ForgeDataFixDelegate(fix);
+
+                globalList.add(delegate);
+                fixerVersions.add(fix.getVersionKey());
+            });
+          });
+
+        //Reset our rules, normal vanilla rules (ours too) keep a cache of their schemas
```
sciwhiz12: Just as clarification, does this comment say that all rules -- both vanilla and ours -- keep their schemas? Because if so, we can reword it to be more simple:
```suggestion
        //Reset our rules, rules keep a cache of their schemas
```

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
```diff
@@ -0,0 +1,244 @@
...
+        //This section was added to make it easier to debug the rule set at runtime, since it is no longer pre-computed off thread,
+        //But only on the main thread, at the moment it is needed.
+        //This basically disables the REWRITE_CACHE prefill and fills it on-demand.
+        //Although it is mainly used for debugging, it also provides a good way of reducing DFU's memory footprint, when it is not needed.
+        //Since the following code will compute an update rule from every version to the current for every known type in the schema,
+        //relating to this version. By default, however this is disabled to preserve vanilla compatibility, since it does do the pre-compute
+        //to fill up the cache.
+        //To enable it start the JVM with the system property set to true, -Dforge.datafixer.disablePreCompute=true needs to be added as
+        //a JVM Launch argument.
+        if (Boolean.getBoolean("forge.datafixer.disablePreCompute"))
```
sciwhiz12: THis can be moved to a `static final` field.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
```diff
@@ -0,0 +1,244 @@
...
+        {
+            LOGGER.warn("The pre-compute of the data fixer rules is disabled, this is not recommended for normal use, but can be useful for debugging or in low-memory situations.");
+            return;
+        }
+
+        //We only rebuild once we are told to do so.
+        if (!triggerOffThreadCompute) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
        if (!triggerOffThreadCompute)
        {
```

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
```diff
@@ -0,0 +1,244 @@
...
+                    final TypeRewriteRule rule = activeFixer.getRule(DataFixUtils.getVersion(versionKey), dataVersion);
+
+                    //Optimize the rule to only apply to the given type.
+                    dataType.rewrite(rule, DataFixerUpper.OPTIMIZATION_RULE);
+                }, executor).exceptionally(e -> {
+                    //Fails.
+                    LOGGER.error("Unable to build datafixers", e);
+
+                    //Just kill it, if this fails we have a problem anyway!
+                    Runtime.getRuntime().exit(1);
```
sciwhiz12: I believe we've discussed this already that this is what vanilla does in `DataFixerBuilder`. 

While I'm not so confident on the idea of immediately terminating the JVM for this (because of the possibility of mods causing problems and therefore causing an unfriendly error to users), I will at least require this be noted as a comment to alleviate the concerns of future reviewers and maintainers. Additionally, this can use `System.exit` instead.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerSchemaConfigurationModStateTransition.java
```diff
@@ -0,0 +1,212 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.datafix;
```
sciwhiz12: Outdated license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerSchemaConfigurationModStateTransition.java
```diff
@@ -0,0 +1,212 @@
...
+ * for the DFU schema setup.
+ *
+ * These events are fired immediately after mod construction and before the rest of the game
+ * loads, even before registries are fired.
+ *
+ * THIS CLASS IS FOR INTERNAL USE ONLY!
+ */
+public class ForgeDataFixerSchemaConfigurationModStateTransition
+{
+    private static final Logger LOGGER = LogManager.getLogger();
```
sciwhiz12: Use `LogUtils` instead.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerSchemaConfigurationModStateTransition.java
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     * Builds a new mod state transition which can be used to fire the appropriate schema configuration events on the mod bus.
+     *
+     * @return The mod state transition used to configure DFU schemas.
+     */
+    public static IModStateTransition get() {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    public static IModStateTransition get()
    {
```

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,384 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.datafix;
```
sciwhiz12: Outdated license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,384 @@
...
+class ForgeSchema extends Schema
+{
+    //Can not be final, needs to be initialized on the fly!
+    //Constructor invocation order and the what not!
+    private Map<String, Type<?>>                TYPES;
+    private Map<String, Supplier<TypeTemplate>> TYPE_TEMPLATES;
+    private Map<String, Integer>                RECURSIVE_TYPES;
+
+    private final Map<String, Supplier<TypeTemplate>> MODDED_ENTITY_TYPES = Maps.newConcurrentMap();
+    private final Map<String, Supplier<TypeTemplate>> MODDED_BLOCK_ENTITY_TYPES = Maps.newConcurrentMap();
```
sciwhiz12: What is the reason for using this over `new ConcurrentHashMap<>()`? (Ditto for the other uses of that method in this class.)

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,384 @@
...
+
+        if (TYPE_TEMPLATES == null)
+        {
+            TYPE_TEMPLATES = Maps.newConcurrentMap();
+        }
+
+        this.wrapped = wrapped;
+
+        final int subVersion = DataFixUtils.getSubVersion(versionKey);
+        name = "V" + DataFixUtils.getVersion(versionKey) + (subVersion == 0 ? "" : "." + subVersion);
```
sciwhiz12: This assignment to an instance field should be prefixed with `this`.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,384 @@
...
+    @Override
+    public Type<?> getType(final DSL.TypeReference type)
+    {
+        final String name = type.typeName();
+        final Type<?> type1 = TYPES.computeIfAbsent(name, key -> {
+            throw new IllegalArgumentException("Unknown type: " + name);
+        });
+        if (type1 instanceof RecursivePoint.RecursivePointType<?>)
+        {
+            return type1.findCheckedType(-1).orElseThrow(() -> new IllegalStateException("Could not find choice type in the recursive type"));
```
sciwhiz12, line 168: Please explain the magic constant `-1`.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,384 @@
...
+        if (wrapped != null)
+        {
+            return wrapped.registerBlockEntities(schema);
+        }
+        if (parent != null)
+        {
+            return parent.registerBlockEntities(schema);
+        }
+
+        return Maps.newHashMap();
```
sciwhiz12: Use `new HashMap<>()` instead.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,384 @@
...
+        if (wrapped != null)
+        {
+            return wrapped.registerEntities(schema);
+        }
+        if (parent != null)
+        {
+            return parent.registerEntities(schema);
+        }
+
+        return Maps.newHashMap();
```
sciwhiz12: Use `new HashMap<>()` instead.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
sciwhiz12: Formatting: instance fields should be in `lowerCamelCase` (`SCREAMING_SNAKE_CASE` is reserved for class/`static` fields).

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,384 @@
...
+        registerTypes(this, entityTypes, blockEntityTypes);
+
+        //Build the type map again.
+        TYPES = buildTypes();
+    }
+
+    public void resetCustomFixes()
+    {
+
+    }
```
sciwhiz12: Why is this empty?

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,214 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.event.datafix;
```
sciwhiz12: Outdated license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,214 @@
...
+ *
+ * We suggest instead of checking for version equality to check for a greater or equal version, and to unregister (or override with a newer version)
+ * when the type changes again in a newer version.
+ *
+ * If a DFU touches a vanilla type (like a list of block entities in a chunk), then all the possible variations of that type need to be known,
+ * in other words, when mojang changes the type signature of a vanilla type, all the mods that touch that type need also to register their types in that schema to persist.
+ * An example of this is when Mojang moved the block entity list in the chunk during the upgrade from 1.17 to 1.18, any block entity type which is
+ * not registered to the 1.18 schema will be lost, since the rule considers a not registered block entity type, as a type that got removed from the game.
+ */
+public class ConfigureDataFixSchemaEvent extends Event implements IModBusEvent
```
sciwhiz12: For the sake of readability, please separate the paragraphs using `<p>`.

src/main/java/net/minecraftforge/event/datafix/RegisterFixesEvent.java
```diff
@@ -0,0 +1,113 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.event.datafix;
```
sciwhiz12: Oudated license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/event/datafix/RegisterFixesEvent.java
```diff
@@ -0,0 +1,113 @@
...
+public class RegisterFixesEvent extends Event implements IModBusEvent
+{
+    private final Schema schema;
+    private final ModContainer                      container;
+    private final List<Function<Schema, DataFix>> fixerFactories;
+
+    public RegisterFixesEvent(
+      final Schema schema,
+      final ModContainer container
+    ) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    )
    {
```

src/main/java/net/minecraftforge/event/datafix/RegisterFixesEvent.java
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * A chainable registration method which can be used to add fixers to the current schema.
+     *
+     * @param fixerFactory The fixer factory to add.
+     * @param others The other fixer factories to add. (Optional)
+     * @return The event itself, with the fixer factories added.
+     */
+    @SafeVarargs
+    public final RegisterFixesEvent addFixer(final Function<Schema, DataFix> fixerFactory, final Function<Schema, DataFix>... others) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    public final RegisterFixesEvent addFixer(final Function<Schema, DataFix> fixerFactory, final Function<Schema, DataFix>... others)
    {
```

src/test/java/net/minecraftforge/debug/datafix/ForgeDataFixTest.java
```diff
@@ -0,0 +1,72 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.debug.datafix;
```
sciwhiz12: Outdated license header, please run the `updateLicenses`/`licenseFormat` task.

src/test/java/net/minecraftforge/debug/datafix/ForgeDataFixTest.java
```diff
@@ -0,0 +1,72 @@
...
+import com.mojang.logging.LogUtils;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraftforge.event.datafix.ConfigureDataFixSchemaEvent;
+import net.minecraftforge.event.datafix.RegisterFixesEvent;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import org.slf4j.Logger;
+
+@Mod("forge_datafix_test")
+public class ForgeDataFixTest
```
sciwhiz12, line 26: Please provide a javadoc for this class which explains what feature this tests, how to test it, and what the expected results are from said test.

src/test/java/net/minecraftforge/debug/datafix/ForgeDataFixTest.java
```diff
@@ -0,0 +1,72 @@
...
+{
+    public static final  boolean ENABLE = true;
```
sciwhiz12: The conventional name for the field which controls whether a test mod is enabled is `ENABLED`.

src/test/java/net/minecraftforge/debug/datafix/ForgeDataFixTest.java
```diff
@@ -0,0 +1,72 @@
...
+    private static final Logger  LOGGER = LogUtils.getLogger();
+
+    public ForgeDataFixTest()
+    {
+        if (ENABLE) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
        if (ENABLE)
        {
```

src/test/java/net/minecraftforge/debug/datafix/ForgeDataFixTest.java
```diff
@@ -0,0 +1,72 @@
...
+            registerEventHandler();
+        }
+    }
+
+    private void registerEventHandler() {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    private void registerEventHandler()
    {
```

src/test/java/net/minecraftforge/debug/datafix/ForgeDataFixTest.java
```diff
@@ -0,0 +1,72 @@
...
+        FMLJavaModLoadingContext.get().getModEventBus().addListener(
+          this::onDataFixSchemaTypeConfiguration
+        );
+        FMLJavaModLoadingContext.get().getModEventBus().addListener(
+          this::onDataFixerRegistration
+        );
+    }
+
+    private void onDataFixSchemaTypeConfiguration(ConfigureDataFixSchemaEvent configureDataFixSchemaEvent) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    private void onDataFixSchemaTypeConfiguration(ConfigureDataFixSchemaEvent configureDataFixSchemaEvent)
    {
```

src/test/java/net/minecraftforge/debug/datafix/ForgeDataFixTest.java
```diff
@@ -0,0 +1,72 @@
...
+        LOGGER.info("Initializing DFU schema: " + configureDataFixSchemaEvent.getVersion());
+
+        //Register something to all the schemas.
+        configureDataFixSchemaEvent.registerSimpleBlockEntity(new ResourceLocation("forge_datafix_test:simple_block_entity"));
+    }
+
+    private void onDataFixerRegistration(RegisterFixesEvent registerFixesEvent) {
```
sciwhiz12:
```suggestion
    private void onDataFixerRegistration(RegisterFixesEvent registerFixesEvent) 
    {
```

src/test/java/net/minecraftforge/debug/datafix/ForgeDataFixTest.java
```diff
@@ -0,0 +1,72 @@
...
+        LOGGER.info("Registering DFU Fixer " + registerFixesEvent.getVersion());
+
+        registerFixesEvent.addFixer(schema -> new DataFix(schema, false) {
+            @Override
+            protected TypeRewriteRule makeRule() {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
        registerFixesEvent.addFixer(schema -> new DataFix(schema, false)
        {
            @Override
            protected TypeRewriteRule makeRule()
            {
```

sciwhiz12 (changes requested):
I haven't tested this PR yet, so this will be mostly for formatting. On that note, _please_ standardize the formatting across the files (braces on newlines, field names should have natural formatting, space between oneline comment header and comment body).

src/main/java/net/minecraftforge/common/ForgeSerialModStateTransition.java
marchermans: It can't be package private because it is used by the DFU Adaptation system.

It was moved because this reduces the total average exposure. Yes it exposes 1 additional record here, but it removes access to 8 internal methods, which I consider a much better API surface.

src/main/java/net/minecraftforge/datafix/ForgeDataFixDelegate.java
marchermans: That is actually not the case.

These are the checked exceptions that invoke can directly throw, catching any others would result in a behavior where I would need to process those other two events separately.

I then throw an illegal state exception since that is literally the case. If I can not get the rewrite rule build, that means that I can not proceed with the current application state and that I reached an invalid processing state within the DFU specification. As such, as you stated in your comment on the use of an ISE, that type of exception is completely appropriate.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerCustomFixersModStateTransition.java
marchermans: No but I need to be able to pull the fired result back per mod container later.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
marchermans: What is not to understand.
All rules keep a cache of their schema.....
Not just ours.

marchermans: Fixed.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
marchermans: Since this is literally what vanilla does I added the comment, but did not change it to System.exit.

marchermans: image

System.exit is exactly the same anyway....

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
marchermans: Just how I learned to do it: <https://ahdak.github.io/blog/effective-java-part-1/> Factory methods are better for testability and future expansion.

marchermans: This follows the pattern set by Mojang in the parent class, so I won't change it here.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerDelegateHandler.java
sciwhiz12: While `Runtime#exit` redirects to `System.exit` internally, it is noted by the javadocs of Runtime that

> The [`System.exit`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/System.html#exit(int)) method is the conventional and convenient means of invoking this method.
> ~ _[`Runtime#exit(int)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Runtime.html#exit(int)), Java SE 17 & JDK 17_

And while there is credence to the argument of following what vanilla does exactly, there is also the argument that this is akin to the use of, for example, `Lists.newArrayList()` versus `new ArrayList<>()`. However, as this is not too big a deal, I'll leave this point to your choice.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
sciwhiz12: I feel like that this deliberate discontinuity in this class of the project's formatting should be commented upon, then.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
sciwhiz12: This has been neither addressed nor resolved as far as I can see.

ChampionAsh5357:
As an additional comment for whenever block entities or entities are registered, usually an `AddNewChoices` data fix is registered for said type as well. Since we are allowing users to add block entities and entities whenever, an `AddNewChoices` data fix should also be added for schema versions which are being used with these values. Alternatively, the schemas for adding these entities should be those that already have the data fixer attached to them.

marchermans:
> As an additional comment for whenever block entities or entities are registered, usually an `AddNewChoices` data fix is registered for said type as well. Since we are allowing users to add block entities and entities whenever, an `AddNewChoices` data fix should also be added for schema versions which are being used with these values. Alternatively, the schemas for adding these entities should be those that already have the data fixer attached to them.

So I thought about this for a while and maybe I need to ask fry again, but from testing it indicates that it is not needed to add the relevant fix.
Just adding the type (marked as a copy everything) will get the job done to help with retaining the (block) entities during the upgrade.

An `AddNewChoices` datafix would require the modders to also specify the type more exactly and would tie the events even further together. Which is not exactly what I wanted to achieve, but yes to make it work perfectly fine you would need to provide a proper type schema and a choice datafix.

@ChampionAsh5357 how would you personally proceed here?

ChampionAsh5357:
> Just adding the type (marked as a copy everything) will get the job done to help with retaining the (block) entities during the upgrade.

I believe that would be the case. From what I understand, `AddNewChoices` specifically deals with when data is serialized involving what types can current be chosen between (a good example would be the `TaggedChoiceType` instance in `MissingDimensionFix` as that provides different generation types to choose from).

> An `AddNewChoices` datafix would require the modders to also specify the type more exactly and would tie the events even further together. Which is not exactly what I wanted to achieve, but yes to make it work perfectly fine you would need to provide a proper type schema and a choice datafix.

Modders would be specifying the data version the block entity/entity would be registered on anyways so it wouldn't be difficult to add one fixer when those methods are used.

> how would you personally proceed here?

I think it should definitely be looked into more to see where `AddNewChoices` is relevant if it would be needed at all in terms of a querying result. So, it does not need to be a part of this PR and be its own separate matter. It's more or less just adapting a specific fixer to be handled by Forge rather than by the modder.

If this PR wanted to add it, it would just add a new `AddNewChoices` if the method on the event is called. I doubt there would be any issue if the schema happened to be added twice (once by vanilla and once by forge) since the data would already be transformed.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,200 @@
...
+
+    /**
+     * Removes an already registered block entity from the current schema and all its descendent versions.
+     * This allows for the removal of a custom block entity type when it is registered in a parent version and is no longer needed.
+     *
+     * @param registryName The registry name of the block entity to remove.
+     */
+    public void removeBlockEntity(final ResourceLocation registryName)
+    {
+        blockEntityTypes.put(registryName.toString(), null);
```
ChampionAsh5357: Concurrent maps throw NPE on adding null values.

src/main/java/net/minecraftforge/event/datafix/ConfigureDataFixSchemaEvent.java
```diff
@@ -0,0 +1,200 @@
...
+
+    /**
+     * Removes an already registered entity from the current schema and all its descendent versions.
+     * This allows for the removal of a custom entity type when it is registered in a parent version and is no longer needed.
+     * 
+     * @param registryName The registry name of the entity to remove.
+     */
+    public void removeEntity(final ResourceLocation registryName)
+    {
+        entityTypes.put(registryName.toString(), null);
```
ChampionAsh5357: Concurrent maps throw NPE on adding null values.

ChampionAsh5357 (changes requested):
Found a few bugs while trying to validate the use of `AddNewChoices`.

So, coming back to this after a bit of testing and playing around. I think the only real issue is if modded non-contextual data changes in the same Minecraft version.

For example, if a capability provider is going to be renamed from `examplemod:cap1` to `examplemod:cap2`, then the data fixer couldn't be run. Additionally, since the data is attached to a vanilla object, creating a custom data fixer would also not fix the issue.

In these cases, I would say that the modder has the responsibility to keep data consistent in the same minecraft version and should modify modded non-contextual data in different Minecraft versions. If they need to change contextual data, they should create their own data fixer to handle those changes.

The other scenario that could play out is not related now, but will screw with things if we ever let people build snapshots *cough* Curle *cough*. Since we are only able to add to existing schemas and not create new ones, there must be at least one snapshot version that changes data in-between the two Minecraft versions. If a user tries to upgrade worlds between snapshots or between Minecraft versions that change no data, then the data fixer will also do nothing if data needs to be changed.

However, that's a low priority since the likelihood is Mojang will change how something is stored at least somewhere once per version.

marchermans:
> So, coming back to this after a bit of testing and playing around. I think the only real issue is if modded non-contextual data changes in the same Minecraft version.
> 
> For example, if a capability provider is going to be renamed from `examplemod:cap1` to `examplemod:cap2`, then the data fixer couldn't be run. Additionally, since the data is attached to a vanilla object, creating a custom data fixer would also not fix the issue.

I talked with Fry about this, and there is not really a way to fix this. Due to the nature of DFU combined with the fact that the version object they use is an int, we don't really have the option to support an arbitrary schema order with arbitrary versioning.

I initially investigated a second DFU instance that could be used to do this, but since both DFU instances need the types to be known then to not cause them to break, this would not work since the second DFU instance would fail because the fixes in the first would not produce the correct type output.

That said I don't think it is a big deal, as you indeed already stated: Modders can you MC boundary schemas to achieve the desired functionality if they want.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,368 @@
...
+
+        //Now add our own modded types, and potentially override the data.
+        this.MODDED_ENTITY_TYPES.putAll(modEntityTypes);
+        this.MODDED_BLOCK_ENTITY_TYPES.putAll(modBlockEntityTypes);
+
+        //Remove those types from the map, which the event marked for removal.
+        this.MODDED_ENTITY_TYPES.values().removeIf(Objects::isNull);
+        this.MODDED_BLOCK_ENTITY_TYPES.values().removeIf(Objects::isNull);
+        this.MODDED_ENTITY_TYPES.keySet().removeIf(modEntityTypesToRemove::contains);
+        this.MODDED_BLOCK_ENTITY_TYPES.keySet().removeIf(modBlockEntityTypesToRemove::contains);
```
pupnewfster: Is there a reason this isn't basically just:
```java
this.MODDED_ENTITY_TYPES.entrySet().removeIf(entry -> entry.getValue() == null || modEntityTypesToRemove.contains(entry.getKey()));
this.MODDED_BLOCK_ENTITY_TYPES.entrySet().removeIf(entry -> entry.getValue() == null || modBlockEntityTypesToRemove.contains(entry.getKey()));
```

That way it would only have to iterate the entries once instead of twice.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,368 @@
...
+        return TYPES.keySet();
+    }
+
+    @Override
+    public Type<?> getTypeRaw(final DSL.TypeReference type)
+    {
+        final String name = type.typeName();
+        return TYPES.computeIfAbsent(name, key -> {
+            throw new IllegalArgumentException("Unknown type: " + name);
+        });
```
pupnewfster, line 154: No need to make this a capturing lambda
```java
return TYPES.computeIfAbsent( type.typeName(), name -> {
    throw new IllegalArgumentException("Unknown type: " + name);
});
```

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    @Override
+    public Type<?> getType(final DSL.TypeReference type)
+    {
+        final String name = type.typeName();
+        final Type<?> type1 = TYPES.computeIfAbsent(name, key -> {
+            throw new IllegalArgumentException("Unknown type: " + name);
+        });
```
pupnewfster, line 163: I think this would be cleaner as:
```java
final Type<?> type1 = getTypeRaw(type);
```
unless there is some technical reason not to such as if it is expected that `getTypeRaw` may be overridden?

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
```diff
@@ -0,0 +1,368 @@
...
+            return type1.findCheckedType(-1).orElseThrow(() -> new IllegalStateException("Could not find choice type in the recursive type"));
+        }
+        return type1;
+    }
+
+    @Override
+    public TypeTemplate resolveTemplate(final String name)
+    {
+        //This is potentially invoked from the super constructor, so our field init code has not run yet.
+        //Set it to a none null value, so the super type can do its initial construction.
```
pupnewfster, line 177: If this is actually invoked by the super constructor doesn't that mean this method will always throw an `IllegalArgumentException` then as the current logic for being invoked early via the super constructor is:
- initialize to empty map
- get from empty map if not present (which is always the case here) throw IAE

src/main/java/net/minecraftforge/event/datafix/RegisterFixesEvent.java
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * A chainable registration method which can be used to add fixers to the current schema.
+     *
+     * @param fixerFactory The fixer factory to add.
+     * @param others The other fixer factories to add. (Optional)
+     * @return The event itself, with the fixer factories added.
+     */
+    @SafeVarargs
+    public final RegisterFixesEvent addFixer(final Function<Schema, DataFix> fixerFactory, final Function<Schema, DataFix>... others)
```
pupnewfster, line 94: What is the benefit of this being chainable given you can just pass other factories via varargs? It seems having both is a little redundant.

src/main/java/net/minecraftforge/datafix/ForgeDataFixerCustomFixersModStateTransition.java
```diff
@@ -0,0 +1,215 @@
...
+     */
+    private static Stream<IModStateTransition.EventGenerator<?>> generateFixersEvents()
+    {
+        //Grab the schemas
+        List<ForgeSchema> schemas = Lists.newArrayList(getAllSchemas());
+        //Sort them based on their version key.
+        schemas.sort(Comparator.comparingInt(Schema::getVersionKey));
+
+        //Create one generator per schema.
+        return schemas.stream().map(InternalEventGenerator::new);
```
pupnewfster: Given this method's use of streams I feel like you might as well keep it entirely using streams and just do
```java
return getAllSchemas().stream()
              .sorted(Comparator.comparingInt(Schema::getVersionKey))
              .map(InternalEventGenerator::new);
```

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
marchermans: This is a straight copy from Mojang, I thought of fixing this but I think in this case it is better to leave it like this.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
marchermans: This is a straight copy from Mojang, I thought of fixing this but I think in this case it is better to leave it like this.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
marchermans: It is just a safe-guard for people being stupid.
It is called from `buildTypes()` which initializes the map properly before hand.

src/main/java/net/minecraftforge/event/datafix/RegisterFixesEvent.java
marchermans: This is just for the different coding styles.
Personally I am a "chainer" and would always use the chaining way of adding multiple instances, but there are people which prefer the varargs. Each their own I guess.

src/main/java/net/minecraftforge/datafix/ForgeSchema.java
pupnewfster: But given it will always error anyway, wouldn't it make more sense to throw an `IllegalStateException` there rather than initialize it.

ChampionAsh5357 (approved):
Should be fine, waiting on merge conflicts to assign.

ChampionAsh5357 (approved):

JayemCeekay:
This PR works great! I'm going to be able to safely update a server using Conquest-Reforged and Create from 1.16.5 to 1.18.2 using it! It let me preserve all conquest paintings and all the create contraptions with no issue!

TheCurle:
Going to be able to, or are able to?

In my testing there are some edge cases with the implementation that need to be ironed out. If it actually works in practice for you, I'd like to get in touch on Discord to figure out how.

marchermans:
> Going to be able to, or are able to?
> 
> In my testing there are some edge cases with the implementation that need to be ironed out. If it actually works in practice for you, I'd like to get in touch on Discord to figure out how. 

What are the edge cases then curle?  Cause I am not aware of them else I would have addressed them by now.

LexManos:
I know this is assigned to me, however due to its nature and the fact that it's broken things quite a bit before. I am going to need more user feedback on this with reproduceable tests.
It doesn't *appear* to break anything in our current tests but neither did the last one.
The other issue is the entire DFU process is very deep magic from Mojang and very much not meant for us to extend. 
So i'm super weary of touching it.

marchermans:
@Raycoms Could you also leave a comment on your experience with this?

JayemCeekay:
I can confirm that this worked perfectly for me. I was able to update the ArdaCraft server from 1.16.5 to 1.18.2 while preserving all Conquest Reforged paintings/entities and Create entities without any issues by registering them in a custom mod utilizing the features provided by this pull request. 10/10 would recommend!

PR #8502, opened 2022-03-06
Clear local variable table on RuntimeEnumExtender transformation

gabizou:
**1.18.x PR** | [1.16.x PR](https://github.com/MinecraftForge/MinecraftForge/pull/8340)

This is a "proper" fix for clearing the state of the LVT in the `create` functions due to an exposure of debugging with Mixins. 

## The Crash

Traditionally, I am always writing mixins for SpongeForge, and lately I discovered a crash that was invariably caused by Mixin's debugging tool (`-Dmixin.debug.export=true`) when a Mixin targeted `net.minecraft.entity.EntityClassification`, it'd cause the export to trigger `ClassNode.accept()` to dump the class, and the LVT for the method modified by `RuntimeEnumExtender` doesn't match the expected LVT (the LVT still had the original LVT for the thrown exception).

The various logs and exports that were related to Mixin troubleshooting: https://gist.github.com/gabizou/beef7061791cdf3fdebb46b810362300

## The reproduction

1. Add a Mixin:
```java
@Mixin(EntityClassification)
public abstract class EntityClassificationMixin {
}
```
2. Run Minecraft with `-Dmixin.debug.export=true`
3. Crash occurs as above, and the exported Mixin code appears as above (but this isn't the final class loaded by ModLauncher)

## The Resulting class
The javap output of the loaded class in both with the mixing above and without the mixing: https://gist.github.com/gabizou/9d4926c4d4eaddac128db73ae770759d

Tagging @Mumfrey to help get any further details that I might've missed.

### DIFF
--- a/fmlloader/src/main/java/net/minecraftforge/fml/common/asm/RuntimeEnumExtender.java
+++ b/fmlloader/src/main/java/net/minecraftforge/fml/common/asm/RuntimeEnumExtender.java
@@ -1,20 +1,6 @@
 /*
- * Minecraft Forge
- * Copyright (c) 2016-2021.
- *
- * This library is free software; you can redistribute it and/or
- * modify it under the terms of the GNU Lesser General Public
- * License as published by the Free Software Foundation version 2.1
- * of the License.
- *
- * This library is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


sciwhiz12 (approved):
Tested the same changes locally earlier for #8340, and confirmed to work.

PR #8503, opened 2022-03-07
[1.18.x] Rework the tool action system to use `ToolModificationResult`

SizableShrimp:
This PR reworks the tool action system, specifically its relation to `IForgeBlock` and `IForgeBlockState`, to allow for fine-grained tool modification results with failure and pass states. This has the direct benefit of allowing the `HOE_TILL` tool action to be reinstated without the hackery that was necessary in the 1.17 PR #8025. (This PR is a successor to #8025.)

Additions:
* `IForgeBlock#getToolModificationResult(BlockState, UseOnContext, ToolAction)`
* `IForgeBlockState#getToolModificationResult(UseOnContext, ToolAction)`
* `BlockEvent.BlockToolInteractEvent#<init>(UseOnContext, BlockState, ToolAction)`
* `ForgeHooks#getHoeTillablePair`
* `ToolModificationResult`

Currently, `IForgeBlockState#getToolModifiedState` has two meanings for null.
One represents a cancellation of `BlockEvent.BlockToolInteractEvent` and therefore, the event.
The other represents that the action is simply not supported by the Block/ItemStack and therefore returns null.
This is not enough information for HoeItem's till action because of the abstractions that Mojang implemented in 1.17.
Now, a `TILLABLES` map exists holding `Pair<Predicate<UseOnContext>, Consumer<UseOnContext>>`.
If the action was canceled, we need to return early rather than look in this TILLABLES map.
If the tool-modified blockstate is null, we need to default back to the TILLABLES map.
And finally, if the tool-modified blockstate is not null, we need to provide our own custom Pair that will run the action with the returned blockstate.
This tristate cannot be represented by nullability (binary state).

Therefore, I propose a ToolModificationResult that holds a tool-modified `BlockState` and PASS or FAIL `Type`.
If the modification result has a failure state, the tool-modified state will always be null.
If the modification result has a pass state, then the tool-modified state can be null or non-null:
* null - Fallback on default behavior (vanilla or modded depending on the calling context)
* non-null - Use the returned block state to perform the action

These three states are used accordingly in `ForgeHooks#getHoeTillablePair` with comments and a patch to `HoeItem` exposing the `TILLABLES` map to be able to integrate hoe tilling with the existing tool action system.

Any tool action related methods that do not take the proper `UseOnContext` have been deprecated and annotated with appropriate javadocs pointing modders to the correct methods.
Because of the way overriding works, the default implementation in `IForgeBlock#getToolModifiedState` must stay there until 1.19, at which point it should be moved to getToolModificationResult and the deprecated method removed.
This has been marked as such with a `// TODO`.
Additionally, the new `HOE_TILL` tool action will not be passed to `IForgeBlock#getToolModifiedState` by default. This is so that the default behavior can be a null PASS state with the new ToolModificationResult, which is only possible from `IForgeBlock#getToolModificationResult`.

Suggestions and testing are both welcome. I have done some basic testing which shows that this new system works with new and old tool actions, but more comprehensive testing is necessary.


src/main/java/net/minecraftforge/common/ToolModificationResult.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * If this tool modification result has a {@link #failed() fail state}, this return value is always null.
+     * <p>
+     * If this tool modification result has a {@link #passed() pass state}, a null return value signifies a fallback on the default behavior.
+     *
+     * @return the nullable tool-modified state
+     */
+    @Nullable
+    public BlockState toolModifiedState()
```
marchermans: This is a personal opinion:
Nullable issues are harder to trace and to keep track of.

API wise it would be much better if this method returned an `Optional<BlockState>`

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -660,9 +661,35 @@ default boolean shouldDisplayFluidOverlay(BlockState state, BlockAndTintGetter l
...
+     * @param state The current state
+     * @param context The use on context that the action was performed in
+     * @param toolAction The action being performed by the tool
+     * @return The result holder containing the tool modified state attached to a pass or failure state
+     */
+    @NotNull
+    default ToolModificationResult getToolModificationResult(BlockState state, UseOnContext context, ToolAction toolAction)
+    {
+        if (toolAction == ToolActions.HOE_TILL)
+            return ToolModificationResult.pass(null);
```
marchermans: Make this a utility method with a proper name instead of passing in `null` and depending on uncodified behaviour.

src/main/java/net/minecraftforge/common/extensions/IForgeBlockState.java
```diff
@@ -603,21 +608,47 @@ default boolean shouldDisplayFluidOverlay(BlockAndTintGetter level, BlockPos pos
...
      * @param pos The block position in level
      * @param player The player clicking the block
      * @param stack The stack being used by the player
      * @param toolAction The tool type to be considered when performing the action
      * @return The resulting state after the action has been performed
+     * @deprecated Use {@link #getToolModificationResult(net.minecraft.world.item.context.UseOnContext, ToolAction)} instead
      */
+    @SuppressWarnings("removal")
     @Nullable
+    @Deprecated(forRemoval = true, since = "1.18.2")
```
marchermans, line 648: Was the `since` field not for the version it is supposed to be deleted in?

src/main/java/net/minecraftforge/event/world/BlockEvent.java
```diff
@@ -415,32 +419,57 @@ public PortalShape getPortalSize()
...
         private final ToolAction toolAction;
+        @Nullable
+        private final UseOnContext context;
         private BlockState state;
+        private ToolModificationResult toolModificationResult = ToolModificationResult.pass(null);
 
+        /**
+         * @deprecated Use {@link #BlockToolInteractEvent(UseOnContext, BlockState, ToolAction)} instead
+         */
+        @Deprecated(forRemoval = true, since = "1.18.2")
```
marchermans, line 442: I am not sure why we should deprecate this?
Is there a thrifty reason for this?

marchermans (commented):
Some general comments on the state, the idea is good, but the API and deprecations need some polishing IMHO.

src/main/java/net/minecraftforge/common/extensions/IForgeBlockState.java
SizableShrimp: We had discussion on this previously and we ended up deciding on switching to using `since` properly, where you put the version it was deprecated in (since XXX). And the next major version removes it.

src/main/java/net/minecraftforge/event/world/BlockEvent.java
SizableShrimp: Because being able to post this event from a place without `UseOnContext` is wrong. All places this can be called from should have the UseOnContext. And that way, we can get rid of all the fields/helper methods in 1.19 and just return the one non-null UseOnContext.

src/main/java/net/minecraftforge/common/extensions/IForgeBlockState.java
marchermans: Okiedokie, I made a note :D

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
SizableShrimp: You are mentioning the .pass(null) specifically?

src/main/java/net/minecraftforge/common/ToolModificationResult.java
SizableShrimp: I will see what I can do.

src/main/java/net/minecraftforge/event/world/BlockEvent.java
marchermans: Okey.

patches/minecraft/net/minecraft/world/item/AxeItem.java.patch
```diff
@@ -9,9 +9,9 @@
...
 -         return p_150694_.m_152465_(blockstate);
 -      });
-+      Optional<BlockState> optional = Optional.ofNullable(blockstate.getToolModifiedState(level, blockpos, player, p_40529_.m_43722_(), net.minecraftforge.common.ToolActions.AXE_STRIP));
-+      Optional<BlockState> optional1 = Optional.ofNullable(blockstate.getToolModifiedState(level, blockpos, player, p_40529_.m_43722_(), net.minecraftforge.common.ToolActions.AXE_SCRAPE));
-+      Optional<BlockState> optional2 = Optional.ofNullable(blockstate.getToolModifiedState(level, blockpos, player, p_40529_.m_43722_(), net.minecraftforge.common.ToolActions.AXE_WAX_OFF));
++      Optional<BlockState> optional = Optional.ofNullable(blockstate.getToolModificationResult(p_40529_, net.minecraftforge.common.ToolActions.AXE_STRIP).toolModifiedState());
++      Optional<BlockState> optional1 = Optional.ofNullable(blockstate.getToolModificationResult(p_40529_, net.minecraftforge.common.ToolActions.AXE_SCRAPE).toolModifiedState());
++      Optional<BlockState> optional2 = Optional.ofNullable(blockstate.getToolModificationResult(p_40529_, net.minecraftforge.common.ToolActions.AXE_WAX_OFF).toolModifiedState());
```
marchermans: You forgot to regen patches, since this is does not compie.... `toolModifiedState` returns an optional, and `Optional.ofNullable(...toolModifiedState())` would as such be an `Optional<Optional<BlockState>>` not an `Optional<BlockState>`

src/main/java/net/minecraftforge/common/ToolModificationResult.java
```diff
@@ -0,0 +1,130 @@
...
+        return FAIL_RESULT;
+    }
+
+    /**
+     * Signifies that the default behavior should be invoked.
+     *
+     * @return An immutable tool modification result that signifies a fallback on default behavior
+     */
+    @NotNull
+    public static ToolModificationResult passEmpty()
```
marchermans: `passWithDefaultBehaviour` better describes the usecase of the method, but this is nitpicking at this point.

marchermans (changes requested):
Minor things then this looks good.

patches/minecraft/net/minecraft/world/item/AxeItem.java.patch
SizableShrimp: This didn't happen, *ever*.

src/main/java/net/minecraftforge/common/ToolModificationResult.java
SizableShrimp: I don't like "With", maybe `passFallbackBehavior` or simply `passFallback`?

patches/minecraft/net/minecraft/world/item/AxeItem.java.patch
marchermans: Gotya 😄

src/main/java/net/minecraftforge/common/ToolModificationResult.java
marchermans: The problem is that you are indicating two different things.
1) The fact that you want the result to indicate a `pass` as primary behavior
and
2) The fact that the default behavior should be executed

`passFallback` or `passFallbackBehaviour` don't indicate that the results indicates _both_ of these behaviors.

But I would already agree with `passToFallbackBehaviour` or something in that direction.

SizableShrimp: Got it.

marchermans (approved):

SizableShrimp:
Internal discussion has led to the decision to deny this PR, with the reasoning being that it is too complex and overal not enough of a reason to warrant the changes w/ just Hoes needing it. Going forward, the values in `HoeItem#TILLABLES` will be hardcoded in Forge's default implementation of `IForgeBlock#getToolModifiedState`. Different PR incoming soon.

PR #8508, opened 2022-03-11
Implement IPlantable in BambooBlock

SirEndii:
This PR implements IPlantable into BambooBlock. So It can be used in forge methods like `IForgeBlock#canSustainPlant`
Bamboo is the only plant which missed the IPlantable implementation.


Shadows-of-Fire:
PR looks fine, implementation is the same as in most other `IPlantable` cases.  Plant type will be `PLAINS` for anyone who wants to know that.

PR #8511, opened 2022-03-13
Fix NullPointerException in ShapelessRecipeBuilder when item category is null.

marios-andr:
This is a rather simple fix for the `ShapelessRecipeBuilder`. When the item category of an item is null, and it's passed in as a result, the `save` method attempts to call `getItemCategory().getRecipeFolderName()` but since `getItemCategory()` is null, a `NullPointerException` is thrown. This patch checks whether the category of the item is null and if true sets the string to "empty", otherwise uses the original method. This allows to generate shapeless recipes for items that do not belong to any category.
The string "empty" can be replaced if anyone finds a more suitable name. 
Thank you.

### DIFF
--- a/patches/minecraft/net/minecraft/data/recipes/ShapelessRecipeBuilder.java.patch
+++ b/patches/minecraft/net/minecraft/data/recipes/ShapelessRecipeBuilder.java.patch
@@ -1,0 +1,12 @@
+--- a/net/minecraft/data/recipes/ShapelessRecipeBuilder.java
++++ b/net/minecraft/data/recipes/ShapelessRecipeBuilder.java
+@@ -85,7 +_,8 @@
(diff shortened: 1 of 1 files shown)


ChampionAsh5357:
Can you provide use case and justification for this change?

I have two arguments against this that also should be addressed:

1. This only modifies the `ShapelessRecipeBuilder` when it affects all recipe builders since they all use the recipe folder name from the creative tab.
2. A recipe is meant to be obtained within survival mode. The creative inventory is a superset of all items that can be obtained in survival mode. As such, every item that can be crafted should be able to be easily accessed from the creative inventory. If not, this would be a missing feature as determined by the modder. Even if the item doesn't belong to a single creative tab, it can go in `misc` since it's miscellaneous or `search` since every item in the creative inventory can be searched.

It does make sense to improve the error message with a precondition so it's easy to understand. However, as I see it, there is no need to make the creative tab optional.

PR #8512, opened 2022-03-13
[1.18.x] Fix FMLOnly in forgedev and userdev

SizableShrimp:
This is a simple PR that fixes a regression introduced by [Forge version 39.0.59](https://github.com/MinecraftForge/MinecraftForge/commit/476e20ae96014a13731068c96dfb5d095134d0f8). FMLOnly now loads properly in forgedev and userdev. Production was unaffected by this regression. Forgedev, userdev, and production were all tested and seem to work with this fix. Providing a `mods.toml` and declaring FMLOnly as a mod allows fmlloader to detect and load it properly. The `mods.toml` info is basically a copy of the forge variant. Because fmlonly does not add hooks for the mods button, this will basically never be seen during normal play.


TheCurle:
Better late than never.

sciwhiz12:
:shipit: Published as **1.18.2-40.0.14**.

PR #8514, opened 2022-03-15
[1.18.2] Add hook for making custom boots that can walk on powdered snow

KnightMiner:
Pretty straightforward patch, this replaces the hardcoded check for leather boots with a new `IForgeItem` hook that allow modded boots to act as "snow boots". Since the method already had access to a living entity, I decided to include that as additional context to the method, allowing entity sensitive snow boots.

My usecase for this is I make a mod that adds armor variants that are intended to replace the vanilla armor progression, but at this time I have no way to replicate leather boots powdered snow behavior. So I really only need item NBT sensitive ultimately.

Included in this PR is a test mod that adds "snow boots", which allow the player to walk on the powdered snow if they have less than 5 hearts remaining.


TheCurle:
LGTM. A simple addition yet useful for many winter-style mods.

PR #8516, opened 2022-03-16
Add onServerChatEvent null check

josephtarbit:
`onServerChatEvent` will return `null` if cancelled but no null check is performed before sending the chat message leading to null pointer exceptions.


Shadows-of-Fire:
Yep that component does indeed go down into `MinecraftServer#sendMessage` which will try to call `Component#getString` and error.  Good catch.

PR #8517, opened 2022-03-16
[1.18.x] Add 3 new ingredient types, and make existing ingredients compatible with datagen

KnightMiner:
This PR adds 3 useful types of ingredients, and improves forge's two existing ingredients by adding datagen constructors. It also adds tests for each ingredient as part of the datagen test mod.

## Changes

Changes to existing ingredients:

* `CompoundIngredient` gained a static constructor that merges ingredients, returning either a vanilla ingredient or a compound as needed
* `NBTIngredient` gained a static constructor

New ingredients:

* `PartialNBTIngredient`: for partial NBT matches.
    * See #8004 for a previous PR adding this functionality. Compared to that PR, this one fixes the issue where tools would always match on damage value and supports a set of items for input.
    * Unlike `NBTIngredient`, this allows you to match on a subset of tags, such as matching just an items name and ignoring its durability.
* `IntersectionIngredient`: matches if all child ingredients match, effectively a set intersection
    * Designed for queries such as "non-flammable planks" (anything in non-flammable wood and in planks) or "iron ore in stone" (anything in `ores/iron` and in `ores_in_ground/stone`, see #7891)
* `DifferenceIngredient` matches if the base child matches and the subtracted child does not.
    * Designed for queries such as "flammable planks" (planks and not non-flammable wood) or "non-dense copper ore" (`ores/copper` and not `ore_rates/dense`)
    * I have personally used this ingredient in my mods to make fallback recipes for wood types. That is, I make a recipe for each vanilla wood type, then make a fallback for "planks without vanilla planks"

In addition, this PR adds in the class `AbstractIngredient`, which abstracts several methods that must always be overridden in subclasses and deprecates all the static constructors of `Ingredient`, making them throw. The second change is to prevent a common mistake where modders call `of` through a subclass thinking it constructs the custom ingredient type, notably in cases like `NBTIngredient.of(Stream<ItemStack>)` which looks like it could be valid.

## Testing

As mentioned above, this PR adds these ingredients to the datagenerator test. I specifically added 7 recipes:

* `data_gen_test:exact_nbt_ingredient`: This test ensures `forge:nbt` works. It allows you to craft a golden pickaxe from any iron pickaxe with exactly 3 damage. Test you can craft the pickaxe, and test that renaming the pickaxe in an anvil (making the NBT a non-exact match) does not let you craft the pickaxe
* `data_gen_test:contains_nbt_ingredient_single_item`: This test is the same as the first test, but using `forge:partial_nbt`. Using a stone pickaxe with exactly 3 damage, you can craft an iron pickaxe. You can also rename the pickaxe (making the NBT a non-exact match) and it still works for the recipe
* `data_gen_test:contains_nbt_ingredient_item_set`: `forge:partial_nbt` supports having a set of valid items, so this is a simple test to ensure set matching works. Any wood, stone, or iron pickaxe renamed to `Diamond Pickaxe` in an anvil will craft into a diamond pickaxe. You can damage the pickaxe any amount of times and the recipe will still work
* `data_gen_test:intersection_ingredient`: This recipe tests `forge:intersection`. It only matches if you use non-flammable planks, which are crimson or warped in vanilla (though using the tag means modded planks will also work for the recipe)
* `data_gen_test:difference_ingredient`: This recipe tests `forge:difference`. It only matches if you use flammable fences, which is any of the 6 non-nether woods. Note I intentionally did the opposite of the intersection test to show where each of these is useful.
* `data_gen_test:compound_ingredient_only_vanilla`: Serializer test for the compound ingredient. It should merge all vanilla ingredients into one, and result in a vanilla ingredient leading to a single non-nested array
* `data_gen_test:compound_ingredient_custom_types`: Serializer test for the compound ingredient. It should merge the vanilla ingredients into one, but due to the presence of the custom NBT ingredient, it is forced to have an outer array making the vanilla ingredients nested inside.

## Other discussion points

This PR could potentially be split, one for the datagen helpers and one or more for the custom ingredient types. I personally feel all three new ingredients fit in forge as intersection is basically the same as the AND conditions forge adds elsewhere, and difference is as close to a NOT as I feel ingredients should go.

I am not certain we need the set matching for partial NBT, but it was pretty trivial to add so I figured I might as well add it. I could simplify the serializer using just a single item match if preferred.

Rather than try to merge `PartialNBTIngredient` and `NBTIngredient`, I think keeping them separate is simpler. Original reason against merging was to prevent a network break, but it also let me make the fields more specialized to what partial NBT needed. I would suggest renaming `NBTIngredient` to `forge:strict_nbt` in 1.19; I can include a comment in the PR but want opinions on whether it should be merged first.


src/main/java/net/minecraftforge/common/crafting/DifferenceIngredient.java
```diff
@@ -0,0 +1,141 @@
...
+            return false;
+        return base.test(stack) && !subtracted.test(stack);
+    }
+
+    @Override
+    public ItemStack[] getItems()
+    {
+        if (this.filteredMatchingStacks == null)
+            this.filteredMatchingStacks = Arrays.stream(base.getItems())
+						                                    .filter(stack -> !subtracted.test(stack))
```
ChampionAsh5357: Weirdly out of line.

src/main/java/net/minecraftforge/common/crafting/DifferenceIngredient.java
```diff
@@ -0,0 +1,141 @@
...
+    {
+        JsonObject json = new JsonObject();
+        json.addProperty("type", CraftingHelper.getID(Serializer.INSTANCE).toString());
+        json.add("base", base.toJson());
+        json.add("subtracted", subtracted.toJson());
+        return json;
+    }
+
+    @Override
+    public IIngredientSerializer<DifferenceIngredient> getSerializer() {
```
ChampionAsh5357: New line braces.

src/main/java/net/minecraftforge/common/crafting/DifferenceIngredient.java
```diff
@@ -0,0 +1,141 @@
...
+        return Serializer.INSTANCE;
+    }
+
+    public static class Serializer implements IIngredientSerializer<DifferenceIngredient> {
```
ChampionAsh5357: New line braces.

src/main/java/net/minecraftforge/common/crafting/IntersectionIngredient.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    @Override
+    public ItemStack[] getItems()
+    {
+        if (this.intersectedMatchingStacks == null)
+        {
+            this.intersectedMatchingStacks = Arrays
+                    .stream(children.get(0).getItems())
+                    .filter(stack -> {
```
ChampionAsh5357: New line braces.

src/main/java/net/minecraftforge/common/crafting/IntersectionIngredient.java
```diff
@@ -0,0 +1,169 @@
...
+    {
+        return Serializer.INSTANCE;
+    }
+
+    public static class Serializer implements IIngredientSerializer<IntersectionIngredient>
+    {
+        public static final IIngredientSerializer<IntersectionIngredient> INSTANCE = new Serializer();
+
+        @Override
+        public IntersectionIngredient parse(JsonObject json) {
```
ChampionAsh5357: New line braces.

src/main/java/net/minecraftforge/common/crafting/PartialNBTIngredient.java
```diff
@@ -0,0 +1,157 @@
...
+
+/** Ingredient that matches the given items, performing a partial NBT match. Use {@link NBTIngredient} if you want exact match on NBT */
+public class PartialNBTIngredient extends AbstractIngredient
+{
+    private final Set<Item> items;
+    private final CompoundTag nbt;
+    private final NbtPredicate predicate;
+    protected PartialNBTIngredient(Set<Item> items, CompoundTag nbt)
+    {
+        super(items.stream().map(item -> {
```
ChampionAsh5357: New line braces.

src/main/java/net/minecraftforge/common/crafting/PartialNBTIngredient.java
```diff
@@ -0,0 +1,157 @@
...
+            Set<Item> items = Stream.generate(() -> buffer.readRegistryIdUnsafe(ForgeRegistries.ITEMS)).limit(buffer.readVarInt()).collect(Collectors.toSet());
+            CompoundTag nbt = buffer.readNbt();
+            return new PartialNBTIngredient(items, Objects.requireNonNull(nbt));
+        }
+
+        @Override
+        public void write(FriendlyByteBuf buffer, PartialNBTIngredient ingredient)
+        {
+            buffer.writeVarInt(ingredient.items.size());
+            for (Item item : ingredient.items) {
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/DataGeneratorTest.java
```diff
@@ -222,6 +234,75 @@ protected void buildCraftingRecipes(Consumer<FinishedRecipe> consumer)
                     )
                     .generateAdvancement()
                     .build(consumer, new ResourceLocation("data_gen_test", "conditional2"));
+
+            // ingredient tests
+            // strict NBT match - should match an unnamed iron pickaxe that lost 3 durability
+            Ingredient nbtIngredient = NBTIngredient.of(Util.make(() -> {
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/DataGeneratorTest.java
```diff
@@ -222,6 +234,75 @@ protected void buildCraftingRecipes(Consumer<FinishedRecipe> consumer)
...
+            }));
+            ShapelessRecipeBuilder.shapeless(Items.GOLDEN_PICKAXE)
+                                  .requires(nbtIngredient)
+                                  .unlockedBy("has_pickaxe", has(Items.IRON_PICKAXE))
+                                  .save(consumer, new ResourceLocation("data_gen_test", "exact_nbt_ingredient"));
+
+            // ingredient tests
+            // contains NBT match - should match a stone pickaxe that lost 3 durability, regardless of setting its name
+            ShapelessRecipeBuilder.shapeless(Items.IRON_PICKAXE)
+                                  .requires(PartialNBTIngredient.of(Items.STONE_PICKAXE, Util.make(() -> {
```
ChampionAsh5357: New line braces.

src/test/java/net/minecraftforge/debug/DataGeneratorTest.java
```diff
@@ -222,6 +234,75 @@ protected void buildCraftingRecipes(Consumer<FinishedRecipe> consumer)
...
+                                      CompoundTag nbt = new CompoundTag();
+                                      nbt.putInt(ItemStack.TAG_DAMAGE, 3);
+                                      return nbt;
+                                  })))
+                                  .unlockedBy("has_pickaxe", has(Items.STONE_PICKAXE))
+                                  .save(consumer, new ResourceLocation("data_gen_test", "contains_nbt_ingredient_single_item"));
+
+            // contains NBT match - should match a wood, stone, or iron pickaxe that was named "Diamond Pickaxe", regardless of damage
+            ShapelessRecipeBuilder.shapeless(Items.DIAMOND_PICKAXE)
+                                  .requires(PartialNBTIngredient.of(Util.make(() -> {
```
ChampionAsh5357: New line braces.

ChampionAsh5357 (changes requested):
Formatting issues I noticed before testing the PR.

src/main/java/net/minecraftforge/common/crafting/DifferenceIngredient.java
KnightMiner: Seems I need to argue with my formatter more, this is due to github's "tabs are 8 characters wide"

KnightMiner:
Formatting fixed, I was not entirely sure where to indent things in lambdas with the curly bracket on the newline, so I just indented an extra level when making the change

src/main/java/net/minecraftforge/common/crafting/PartialNBTIngredient.java
```diff
@@ -0,0 +1,157 @@
...
+        }
+        this.items = Collections.unmodifiableSet(items);
+        this.nbt = nbt;
+        this.predicate = new NbtPredicate(nbt);
+    }
+
+    /** Creates a new ingredient matching any item from the list, containing the given NBT */
+    public static PartialNBTIngredient of(CompoundTag nbt, ItemLike... items)
+    {
+        return new PartialNBTIngredient(Arrays.stream(items).map(ItemLike::asItem).collect(Collectors.toSet()), nbt);
```
ChampionAsh5357, line 59: As a question, what is the benefit instead of using a `CompoundIngredient` of `PartialNBTIngredient`s to hold the different items and their preconditions?

src/main/java/net/minecraftforge/common/crafting/PartialNBTIngredient.java
```diff
@@ -0,0 +1,157 @@
...
+
+            return new PartialNBTIngredient(items, nbt);
+        }
+
+        @Override
+        public PartialNBTIngredient parse(FriendlyByteBuf buffer)
+        {
+            Set<Item> items = Stream.generate(() -> buffer.readRegistryIdUnsafe(ForgeRegistries.ITEMS)).limit(buffer.readVarInt()).collect(Collectors.toSet());
+            CompoundTag nbt = buffer.readNbt();
+            return new PartialNBTIngredient(items, Objects.requireNonNull(nbt));
```
ChampionAsh5357, line 145: I think the nullability check would be more useful in the constructor for modders that are constructing this ingredient on the fly.

ChampionAsh5357 (changes requested):
All tests work, just a question and a change.

src/main/java/net/minecraftforge/common/crafting/PartialNBTIngredient.java
KnightMiner: Mostly less redundancy. So less network traffic to sync (sync the tag once instead of once per item) and cleaner recipe JSON. As I said in the main post, it was trivial to do but not strictly needed

src/main/java/net/minecraftforge/common/crafting/PartialNBTIngredient.java
KnightMiner: This check is when reading from network. We never write a null tag to the network, but the way it's implemented means we can technically read a null one (the return is marked nullable, in this case only possible from corrupt network data I assume). The constructor being nullable is not a bad idea, I guess it would serve the same need but the stacktrace would be a bit longer then for what is a network error. Plus, then we ask why I don't have require not null on all parameters

ChampionAsh5357: My thought process was the way this would come about would be from someone who passed in a potentially nullable tag in `#of` since the nbt tag must be present during decoding which would throw an error on parsing instead. So, unless the user was declaring an ingredient in code, I don't know how this would become null without  as you said some corruption in which case I would think it should just error out altogether.

KnightMiner: The goal here was to error immediately, instead of erroring later. Which is typically when I use that method.

The constructor is non null by default and this method returns nullable so I wrap it. If a modder is ignoring nullability and passing potentially null that is a bug in their code

ChampionAsh5357 (approved):
You can resolve all conversations.

TheCurle:
Perfect! 

I've had a few requests for this in the past, and this will help a lot with the common confusion regarding the various `.of` methods.

Great work with the comprehensive tests, too.

PR #8522, opened 2022-03-19
[1.18.2] Allow Modders to create custom datapack registries

Silverminer007:
This PR allows Modders to easily create a custom registry that loads elements not only from code, but from data-packs too:
- Allow Modders to set a registry to be a data pack in their Registry Builder (this enables tags too) (these registires will get injected to BuiltinRegistries#REGISTRY)
- Allow registering elements trough the DeferredRegister
- Added DataPackRegistriesHooks to create a data-pack DeferedRegister to register elements in code
- Added AddBuiltinRegistryEvent to add the direct codec/and network codec to RegistryAcess#REGISTRIES
- All elements of the registry 'mod_a:registry_a' are read from 'data/modid/mod_a/registry_a/'
- All tags of the registry 'mod_a:registry_a' are read from 'data/modid/tags/mod_a/registry_a'
- Elements registered in code can be overridden from data-packs

**Technical Approach** 
- Patch Registry Access to use a immutable view of the mutable REGISTRIES Map
- On Load Complete, a event is fired to add the direct codec to that event
- Add a utility method to create Holder objects for these registries in DataPackRegistriesHooks
- Expand NewRegistryEvent#fill to inject data-pack registries to BuiltinRegistries#REGISTRY (Builtin Root Registry)
- The ResourceKey of a data-pack registry is verified to be prefixed with the modid, to prevent mod conflicts

**Why is this useful**
- Data-pack registries allows better customisability of mod registries for users
- Data-pack based registries can improve inter mod compatibility

**Fin**
This PR is based on #8263. It aims to make it easier to add data-pack registries and bring a port for 1.18.2.  I think there are other ways to implement this feature, but I think this brings the smallest patch size and the easiest way for modders to a new data-pack registry. If you want me to add any additional test case, just let me know. Thanks :)


ChampionAsh5357:
Looking over this, it seems to be the same as the aforementioned PR besides the coupling to the vanilla registry system compared to the forge based registry.

Currently, since the registry system for Forge would need to be redesigned to support all the changes Mojang has made, this is not really worthwhile. It seems to use operations that were added in tags fix which was mainly patchwork. In addition, I don't believe custom datapack registries should be coupled to vanilla's system since a much better implementation could be made.

My opinion as such would be to redesign this system once a more concrete standpoint has been made on the future on how Forge will handle the registry changes. Until then, if you want this system, it would be just as easy to make this into a mod.

Silverminer007:
I don't think this can be done in a mod without core modding, because RegistryAccess is an interface that doesn't even allow mixins to be used. 
I'm willing to make a new attempt to implement this, but I'd like to know exactly what points to change. I have considered making RegistryAccess#REGISTRIES non-modifiable, but I think that is very difficult without major changes. I tried creating an event that fires as soon as the lambda is run, but that fires before the mods start loading, so mods weren't able to subscribe to the event. Any implementation ideas are welcome.

ChampionAsh5357:
> I don't think this can be done in a mod without core modding, because RegistryAccess is an interface that doesn't even allow mixins to be used.

It can. Just purely using datapacks with reload listeners works just as well since they are both loaded around the same time. The reason I mentioned it is because I have done it.

> I'm willing to make a new attempt to implement this, but I'd like to know exactly what points to change.

As I've mentioned, it makes more sense to wait for this PR until Forge finalizes how they plan to change registries to deal with Mojang changes since otherwise this would just be building upon a patchwork fix which is not ideal for a new major subsystem.

marchermans (approved):

hYdos (approved):

hYdos:
> As I've mentioned, it makes more sense to wait for this PR until Forge finalizes how they plan to change registries to deal with Mojang changes since otherwise this would just be building upon a patchwork fix which is not ideal for a new major subsystem.

The only issue with this is that some modders may want to just work on their mods and having data driven registries may be core features of mods.

Commoble:
> > I don't think this can be done in a mod without core modding, because RegistryAccess is an interface that doesn't even allow mixins to be used.
> 
> It can. Just purely using datapacks with reload listeners works just as well since they are both loaded around the same time. The reason I mentioned it is because I have done it.

A big problem here is that dynamic registries and dimensions now (as of 1.18.2) load before reloadable data (as opposed to loading after reloadable data in earlier versions). This matters for other WIP features, such as https://github.com/MinecraftForge/MinecraftForge/pull/8341, which intends to provide a means of making data-driven modifications to biomes prior to dimensions loading. If reloadable data hasn't loaded yet, then the only data available at that time is via the dynamic registries, so we'd need some sort of way to make custom dynamic registries to make that viable.

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
```diff
@@ -0,0 +1,101 @@
...
+
+public class DataPackRegistriesHooks
+{
+   private static Map<ResourceKey<? extends Registry<?>>, RegistryAccess.RegistryData<?>> REGISTRY_ACCESS_REGISTRIES_COPY;
+
+   /** Modders, DO NOT USE. Internal use only */
+   public static Map<ResourceKey<? extends Registry<?>>, RegistryAccess.RegistryData<?>> addBuiltinRegistriesEvent(ImmutableMap.Builder<ResourceKey<? extends Registry<?>>, RegistryAccess.RegistryData<?>> builder)
+   {
+      REGISTRY_ACCESS_REGISTRIES_COPY = new HashMap<>(builder.build());
+      return REGISTRY_ACCESS_REGISTRIES_COPY;
```
SizableShrimp: This method should return an immutable view instead of the mutable map.

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+public class DataPackRegistriesHooks
+{
+   private static Map<ResourceKey<? extends Registry<?>>, RegistryAccess.RegistryData<?>> REGISTRY_ACCESS_REGISTRIES_COPY;
+
+   /** Modders, DO NOT USE. Internal use only */
+   public static Map<ResourceKey<? extends Registry<?>>, RegistryAccess.RegistryData<?>> addBuiltinRegistriesEvent(ImmutableMap.Builder<ResourceKey<? extends Registry<?>>, RegistryAccess.RegistryData<?>> builder)
```
SizableShrimp, line 29: This method and others like it with `Modders, DO NOT USE. Internal use only` as documentation should be expanded to have `@Deprecated`.

src/main/java/net/minecraftforge/event/world/AddBuiltinRegistryEvent.java
```diff
@@ -0,0 +1,33 @@
...
+import net.minecraft.core.RegistryAccess;
+import net.minecraft.resources.ResourceKey;
+import net.minecraftforge.eventbus.api.Event;
+import net.minecraftforge.fml.event.IModBusEvent;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.function.BiConsumer;
+
+public class AddBuiltinRegistryEvent extends Event implements IModBusEvent
```
SizableShrimp, line 32: This whole event class needs some serious documentation.

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
```diff
@@ -0,0 +1,101 @@
...
+
+   /** Modders, DO NOT USE. Internal use only */
+   public static void fireAddBuiltinRegistriesEvent()
+   {
+      AddBuiltinRegistryEvent registryEvent = new AddBuiltinRegistryEvent((resourceKey, registryData) ->
+      {
+         validateRegistryKey(resourceKey);
+         if (!BuiltinRegistries.REGISTRY.containsKey(resourceKey.location()))
+         {
+            throw new IllegalArgumentException("DataPack registry must be marked as DataPack Registry in RegistryBuilder");
```
SizableShrimp: Weird formatting, this should probably directly reference the method on RegistryBuilder in the format `RegistryBuilder#isDataPackRegistry` and DataPack capitalization is weird.

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
```diff
@@ -0,0 +1,101 @@
...
+         throw new IllegalArgumentException(String.format("Registry path must be prefixed with '%s'", prefix));
+      }
+   }
+
+   public static <T extends IForgeRegistryEntry<T>> DeferredRegister<T> createDataPackRegistry(ResourceKey<? extends Registry<T>> registryKey, String modid, Class<T> baseClass)
+   {
+      DeferredRegister<T> deferredRegister = DeferredRegister.create(registryKey, modid);
+      deferredRegister.makeRegistry(baseClass, () -> new RegistryBuilder<T>().disableSync().disableSaving().isDataPackRegistry());
+      return deferredRegister;
+   }
```
SizableShrimp: All of this seems like unnecessary util.

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
```diff
@@ -0,0 +1,101 @@
...
+
+      return Optional.empty();
+   }
+
+   private static boolean registryExists(ResourceLocation registryName)
+   {
+      return RegistryManager.ACTIVE.getRegistry(registryName) != null
+            || Registry.REGISTRY.containsKey(registryName)
+            || BuiltinRegistries.REGISTRY.containsKey(registryName);
+   }
```
SizableShrimp: Unnecessary util, superseded by #8548.

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -364,6 +366,7 @@ public void register(IEventBus bus)
             bus.addListener(this::createRegistry);
         }
     }
+
```
SizableShrimp: This needs to be deleted/reverted.

src/main/java/net/minecraftforge/registries/NewRegistryEvent.java
```diff
@@ -62,8 +63,15 @@ void fill()
...
-                RegistryManager.registerToRootRegistry((ForgeRegistry<?>) reg);
+            if(builder.getIsDataPackRegistry())
+            {
+                if (builder.getHasWrapper() && !BuiltinRegistries.REGISTRY.containsKey(reg.getRegistryName()))
+                    RegistryManager.registerToBuiltinRegistry((ForgeRegistry<?>) reg);
+            } else
+            {
+                if (builder.getHasWrapper() && !Registry.REGISTRY.containsKey(reg.getRegistryName()))
+                    RegistryManager.registerToRootRegistry((ForgeRegistry<?>) reg);
+            }
```
SizableShrimp: This else and if should be merged.

src/main/java/net/minecraftforge/registries/NewRegistryEvent.java
```diff
@@ -62,8 +63,15 @@ void fill()
                 .collect(Collectors.toMap(RegistryData::builder, d -> d.builder().create(), (a, b) -> a, IdentityHashMap::new));
 
         builtRegistries.forEach((builder, reg) -> {
-            if (builder.getHasWrapper() && !Registry.REGISTRY.containsKey(reg.getRegistryName()))
-                RegistryManager.registerToRootRegistry((ForgeRegistry<?>) reg);
+            if(builder.getIsDataPackRegistry())
```
SizableShrimp: Missing space after `if`.

src/main/java/net/minecraftforge/registries/RegistryBuilder.java
```diff
@@ -230,6 +231,13 @@ public RegistryBuilder<T> hasTags()
         return this;
     }
 
+    public RegistryBuilder<T> isDataPackRegistry()
+    {
+        this.hasWrapper();// A Wrapper is required for data pack registries
```
SizableShrimp: Wrapper shouldn't be capitalized, and there should be a space after the semicolon and before the comment start.

src/main/java/net/minecraftforge/registries/RegistryBuilder.java
```diff
@@ -387,4 +395,9 @@ boolean getHasWrapper()
     {
         return this.hasWrapper;
     }
+
+    boolean getIsDataPackRegistry()
```
SizableShrimp: Kinda weirder getter name.

src/main/java/net/minecraftforge/registries/RegistryManager.java
```diff
@@ -163,6 +165,15 @@ static <V extends IForgeRegistryEntry<V>> void registerToRootRegistry(ForgeRegis
...
     }
 
+    @SuppressWarnings("unchecked")
+    static <V extends IForgeRegistryEntry<V>> void registerToBuiltinRegistry(ForgeRegistry<V> forgeReg)
+    {
+        WritableRegistry<Registry<V>> registry = (WritableRegistry<Registry<V>>) BuiltinRegistries.REGISTRY;
+        Registry<V> wrapper = forgeReg.getWrapper();
+        if (wrapper != null)
+            registry.register(forgeReg.getRegistryKey(), wrapper, Lifecycle.experimental());
+    }
```
SizableShrimp: Instead of code duplication, probably better to extract the last 3 lines of both functions, and possibly the unchecked cast too, to a private static helper function.

src/test/java/net/minecraftforge/debug/world/DataPackRegistriesTest.java
```diff
@@ -0,0 +1,210 @@
...
+import net.minecraftforge.registries.ForgeRegistryEntry;
+import net.minecraftforge.registries.RegistryObject;
+import org.jetbrains.annotations.Nullable;
+import org.slf4j.Logger;
+
+import java.util.List;
+import java.util.Optional;
+
+@Mod(DataPackRegistriesTest.MODID)
+public class DataPackRegistriesTest
```
SizableShrimp, line 35: Please add an ENABLED boolean flag. If it is particularly log spammy or modifies vanilla behavior, please default it to disabled (false).

SizableShrimp (changes requested):
This looks like a good start. But it still needs some work, and also need to run `updateLicenses`.

src/main/java/net/minecraftforge/event/world/AddBuiltinRegistryEvent.java
Silverminer007: I hope, I haven't missed anything important in the new documentation. Is there anything that I should add?

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
Silverminer007: I'm not sure what you mean with "immutable view". I need to pass the mutable map, because it changes afterwards. If I would return the immutable map, all changes through the event would be ignored. Could you specify what you mean?

SizableShrimp: A collection "view" is a wrapper instance around an original collection that is updated when the original collection is updated. It is a view because it does not have the same properties / allowed functions. In this case, the immutable view would not allow mutating the map, only reading. Please see `Collections#unmodifiableMap`.

src/main/java/net/minecraftforge/event/world/AddBuiltinRegistryEvent.java
```diff
@@ -0,0 +1,60 @@
...
+ * <p>
+ *    This event is used to register the Codec to a data-pack-registry. All Registries passed here should be marked as data-pack-registries
+ *    with RegistryBuilder#dataPackRegistry.
+ * </p>
+ * <p>
+ *    You should avoid the usage of {link AddBuiltinRegistryEvent#addRegistry(ResourceKey, Codec, Codec)},
+ *    because clients missing the required network codec can cause strange issues
+ * </p>
+ */
+public class AddBuiltinRegistryEvent extends Event implements IModBusEvent
```
SizableShrimp, line 32: Use `{@link}` more here, instead of just dropping capitalized names.
Missing a period on the last sentence.
Last sentence is also not specific. What strange issues? When will a client being missing the network codec?

src/main/java/net/minecraftforge/event/world/AddBuiltinRegistryEvent.java
```diff
@@ -0,0 +1,60 @@
...
+   public AddBuiltinRegistryEvent(BiConsumer<ResourceKey<? extends Registry<?>>, RegistryAccess.RegistryData<?>> dataPackRegistryConsumer)
+   {
+      this.dataPackRegistryConsumer = dataPackRegistryConsumer;
+   }
+
+   /**
+    * @param registryKey The passed ResourceKey should belong to a Registry marked as data-pack-registry with RegistryBuilder#dataPackRegistry
+    * @param codec this codec is used to serialize elements from data-packs to the registry
+    */
+   public <T> AddBuiltinRegistryEvent addRegistry(@NotNull ResourceKey<? extends Registry<T>> registryKey, @NotNull Codec<T> codec)
```
SizableShrimp, line 45: Be consistent with your capitalization on docs please. Prefer lowercase to start a parameter doc line if you can. A link to what you can call on RegistryBuilder would also be more appropriate.

src/main/java/net/minecraftforge/event/world/AddBuiltinRegistryEvent.java
```diff
@@ -0,0 +1,60 @@
...
+   {
+      return this.addRegistry(registryKey, codec, null);
+   }
+
+   /**
+    * @param registryKey The passed ResourceKey should belong to a Registry marked as data-pack-registry with RegistryBuilder#dataPackRegistry
+    * @param codec this codec is used to serialize elements from data-packs to the registry
+    * @param networkCodec the network codec is used to serialize all registry elements for client-server sync. If you want to disable sync, pass {@code null}
+    */
+   public <T> AddBuiltinRegistryEvent addRegistry(@NotNull ResourceKey<? extends Registry<T>> registryKey, @NotNull Codec<T> codec, @Nullable Codec<T> networkCodec)
```
SizableShrimp, line 55: Same deal as the other addRegistry method.

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
```diff
@@ -0,0 +1,61 @@
...
+
+   private static void validateRegistryKey(ResourceKey<? extends Registry<?>> registryKey)
+   {
+      var location = registryKey.location();
+      var prefix = location.getNamespace() + '/';
+
+      if (!location.getPath().startsWith(prefix))
+      {
+         throw new IllegalArgumentException(String.format("Registry path must be prefixed with '%s'", prefix));
```
SizableShrimp, line 58: This should be using `Locale.ENGLISH`. But I don't really get the point of this validation method at all. Can you explain?

src/main/java/net/minecraftforge/registries/NewRegistryEvent.java
```diff
@@ -62,8 +63,14 @@ void fill()
                 .collect(Collectors.toMap(RegistryData::builder, d -> d.builder().create(), (a, b) -> a, IdentityHashMap::new));
 
         builtRegistries.forEach((builder, reg) -> {
-            if (builder.getHasWrapper() && !Registry.REGISTRY.containsKey(reg.getRegistryName()))
+            if (builder.isDataPackRegistry())
+            {
+                if (builder.getHasWrapper() && !BuiltinRegistries.REGISTRY.containsKey(reg.getRegistryName()))
+                    RegistryManager.registerToBuiltinRegistry((ForgeRegistry<?>) reg);
+            } else if (builder.getHasWrapper() && !Registry.REGISTRY.containsKey(reg.getRegistryName()))
```
SizableShrimp, line 70: Small thing but this would follow the formatting by having the `else if` start on a new line separate from the closing brace.

src/main/java/net/minecraftforge/registries/RegistryBuilder.java
```diff
@@ -230,6 +231,13 @@ public RegistryBuilder<T> hasTags()
         return this;
     }
 
+    public RegistryBuilder<T> dataPackRegistry()
```
SizableShrimp, line 234: Would it be good to explain what enabling a data pack registry does here?

SizableShrimp (changes requested):

sciwhiz12:
As per previous communication from the author, this is effectively superseded by #8630, which is a fork of this PR.

PR #8524, opened 2022-03-21
[1.18.x] [event] Add event for lightning strikes

Minecraftschurli:
This PR adds a LightningStrikeEvent that fires on each of the 3 ticks of a LightningBolt which when cancelled prevents vanilla behavior and removes the lightning bolt.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/LightningBolt.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/LightningBolt.java.patch
@@ -8,7 +8,7 @@
  
     public LightningBolt(EntityType<? extends LightningBolt> p_20865_, Level p_20866_) {
        super(p_20865_, p_20866_);
-@@ -74,6 +_,14 @@
+@@ -74,8 +_,17 @@
  
     }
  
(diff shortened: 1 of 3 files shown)


patches/minecraft/net/minecraft/world/entity/LightningBolt.java.patch
```diff
@@ -22,7 +22,10 @@
 +
     public void m_8119_() {
        super.m_8119_();
++      if (net.minecraftforge.common.ForgeHooks.onLightningStrike(this, this.m_147162_())) {this.m_142687_(RemovalReason.DISCARDED);return;}
```
Shadows-of-Fire, line 25: Spaces
`{this.m_142687_(RemovalReason.DISCARDED);return;}` -> `{ this.m_142687_(RemovalReason.DISCARDED); return; }`

sciwhiz12: Also, why does this call `remove(RemovalReason.DISCARDED)` instead of `discard()`?

Shadows-of-Fire:
So aside from being one tick later (since it fires on the first call to `tick()` rather than immediately) - what does this accomplish that cannot be accomplished with `EntityJoinWorldEvent`

patches/minecraft/net/minecraft/world/entity/LightningBolt.java.patch
Minecraftschurli: because I didn't know it exists

Minecraftschurli:
I made the event because it was suggested by gigahertz but since after looking at it again everything can be done with `EntityJoinWorldEvent` I'll gonna close this PR

PR #8527, opened 2022-03-23
[1.18.x] Allow using DeferredRegisters for vanilla registries

SizableShrimp:
This PR expands the functionality of `DeferredRegister`s and `RegistryObject`s to be able to use them for vanilla registries which have no associated forge registry.
Unlike #8495, this PR contains no binary or source compatibility breaks and should therefore be only a benefit to modders.

To use a `DeferredRegister` for a vanilla registry, use the `create` factory that takes a `ResourceKey`. Example:
```java
public static final DeferredRegister<RecipeType<?>> RECIPE_TYPES = DeferredRegister.create(Registry.RECIPE_TYPE_REGISTRY, MODID);
```
Then, register objects like you would with any normal `DeferredRegister`. Example:
```java
    public static final RegistryObject<RecipeType<?>> TEST_RECIPE_TYPE = RECIPE_TYPES.register("test", () -> new RecipeType<>() {});
```
This example has also been added to `DeferredRegistryTest`.

This should cleanup the admittedly ugly previous solution of "just sticking it wherever" and using `Lazy.of()` or static non-final fields that start out as null, then registering from a random `RegistryEvent.Register<?>`.
More testing would be appreciated, e.g. worldgen people.


src/main/java/net/minecraftforge/registries/VanillaRegisterEvent.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+package net.minecraftforge.registries;
+
+import net.minecraft.core.Registry;
+import net.minecraftforge.eventbus.api.Event;
+import net.minecraftforge.fml.event.IModBusEvent;
+import org.jetbrains.annotations.NotNull;
+
+class VanillaRegisterEvent extends Event implements IModBusEvent
```
KnightMiner, line 20: Any reason this is package protected instead of public? An event like this would be useful outside deferred registers for registering things into vanilla registries using the classic event style.

Also, could this be made a generic event? That way you could subscribe to `VanillaRegisterEvent<RecipeType<?>` for instance.

LexManos: The entire concept is that we can no longer garentee the type. The plan is to remove types all together in 1.19.
And the reason it is private is because We do not wish modders to use it. Instead, using DR is the desired API. I even told Shrimp that this shouldn't be a event at all, instead it should be a manual callback on our end. 

In 1.19 the Register event will become non-generic, and instead be keyed on Registry Name.

This event is purely a stop gap until we can make that breaking change.

sciwhiz12:
This PR is currently blocked as both myself and the PR author are experiencing a very strange issue where the `@ObjectHolder`-annotated field in [`MultiLayerModelTest`](https://github.com/SizableShrimp/MinecraftForge/blob/e8ec6b892760f28467b891058bf99b70cbae2fef/src/test/java/net/minecraftforge/debug/client/model/MultiLayerModelTest.java#L30-L31) seems to fail almost at random (with a bias for not working). Somehow, the injection for that field fails seemingly at random, causing a `NullPointerException` when its corresponding `BlockItem` is created and registered afterwards.

Until this issue is debugged and fixed, this PR cannot move forward to assignment. We appreciate it if others try this PR locally to see if they experience this issue as well.

src/main/java/net/minecraftforge/registries/RegistryObject.java
```diff
@@ -111,15 +115,25 @@ private RegistryObject() {
...
                 {
                     this.registry = RegistryManager.ACTIVE.getRegistry(baseType);
                     if (registry == null)
-                        throw new IllegalStateException("Unable to find registry for type " + baseType.getName() + " for mod \"" + modid + "\". Check the 'caused by' to see further stack.", callerStack);
+                    {
+                        invalidRegistry = true;
+                        LOGGER.error(
+                                "Unable to find registry for type " + baseType.getName() + " for mod \"" + modid + "\". Check the 'caused by' to see further stack.",
+                                callerStack
+                        );
```
sciwhiz12: Use logger parameter substitution, and move this all to one line:
```suggestion
                        LOGGER.error("Unable to find registry for type {} for mod \"{}\". Check the 'caused by' to see further stack.", baseType.getName(), modid, callerStack);
```

src/main/java/net/minecraftforge/registries/RegistryObject.java
```diff
@@ -140,16 +154,24 @@ private RegistryObject(final ResourceLocation name, final ResourceLocation regis
...
+                if (invalidRegistry)
+                    return;
                 if (!registryExists)
                 {
                     if (!registryExists(registryName))
-                        throw new IllegalStateException(
+                    {
+                        invalidRegistry = true;
+                        LOGGER.error(
                                 "Unable to find registry with key " + registryName + " for mod \"" + modid + "\". Check the 'caused by' to see further stack.",
```
sciwhiz12: Move to logger parameter substitution, and move to one line (see previous comment).

src/main/java/net/minecraftforge/registries/VanillaRegisterEvent.java
sciwhiz12, line 20: While this event is a stopgap measure as said by @LexManos, I feel that it should still be documented stating the same, along with referencing and linking to `DeferredRegister`. It may also be worth considering (terminally) deprecating this, to definitely mark this for deletion come 1.19.

src/main/resources/coremods/registry_object_binary_compat.js
```diff
@@ -0,0 +1,54 @@
+var Opcodes = Java.type('org.objectweb.asm.Opcodes')
+var MethodNode = Java.type('org.objectweb.asm.tree.MethodNode')
+var VarInsnNode = Java.type('org.objectweb.asm.tree.VarInsnNode')
+var MethodInsnNode = Java.type('org.objectweb.asm.tree.MethodInsnNode')
+var TypeInsnNode = Java.type('org.objectweb.asm.tree.TypeInsnNode')
+var InsnNode = Java.type('org.objectweb.asm.tree.InsnNode')
+
+// TODO 1.19: Delete this whole coremod. We won't need to maintain binary compatibility then.
+function initializeCoreMod() {
```
sciwhiz12: Even though this will be gone in 1.19, please still add comments explaining what this coremod does, to avoid having to read through the coremod's logic. (Some may say that is unnecessary, but I say that if we're adding something here that not many understand, then it's a good place to add comments.)

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -90,34 +90,32 @@
...
+    public static <B> DeferredRegister<B> create(ResourceLocation registryName, String modid)
     {
-        return new DeferredRegister<>(registryName, null, modid);
+        return new DeferredRegister<>(ResourceKey.createRegistryKey(registryName), null, modid);
     }
 
-    private final ResourceLocation registryName;
-    private final Class<T> superType;
+    private final ResourceKey<? extends Registry<T>> registryKey;
+    private final Class<? extends IForgeRegistryEntry<?>> superType;
```
sciwhiz12: These should be marked as `@Nullable`. Additionally, add comments which denote which of the two are nullable at which time.

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -90,34 +90,32 @@
...
     private final String modid;
     private final Map<RegistryObject<T>, Supplier<? extends T>> entries = new LinkedHashMap<>();
     private final Set<RegistryObject<T>> entriesView = Collections.unmodifiableSet(entries.keySet());
 
-    private IForgeRegistry<T> type;
-    private Supplier<IForgeRegistry<T>> typeSupplier;
-    private Supplier<RegistryBuilder<T>> registryFactory;
+    private Supplier<RegistryBuilder<?>> registryFactory;
     private SetMultimap<TagKey<T>, Supplier<T>> optionalTags;
```
sciwhiz12: These should be marked as `@Nullable`.

patches/minecraft/net/minecraft/core/RegistryAccess.java.patch
```diff
@@ -1,11 +1,10 @@
 --- a/net/minecraft/core/RegistryAccess.java
 +++ b/net/minecraft/core/RegistryAccess.java
-@@ -185,6 +_,8 @@
-    private static <E> void m_211081_(RegistryResourceAccess.InMemoryStorage p_211082_, RegistryAccess.RegistryData<E> p_211083_) {
+@@ -186,6 +_,7 @@
        ResourceKey<? extends Registry<E>> resourcekey = p_211083_.f_123101_();
        Registry<E> registry = BuiltinRegistries.f_206379_.m_175515_(resourcekey);
-+      if (!resourcekey.equals(Registry.f_122818_))
-+         registry = ((Registry<Registry<E>>)BuiltinRegistries.f_123858_).m_6246_((ResourceKey<Registry<E>>)resourcekey);
```
sciwhiz12, line 8: Why was this patch deleted? (Asking so it's documented here for future reference.)

sciwhiz12:
The above noted issue was resolved. To my understanding, the issue was caused by three reasons:

  1. Because the `PlacedFeature` registry did not exist in the root (regular) registry at `Registry.REGISTRY` (and neither exists as a Forge-added registry), the handler registered by `RegistryObject` to `ObjectHolderRegistry` threw an `IllegalStateException`.

  2. The above noted exception was swallowed because of the internals of FML.[^1] However, the exception still would interrupt `ObjectHolderRegistry`'s calling of its internal collection. If the `RegistryObject` was called (and threw) before the `ObjectHolder`s, then this issue happens.

  3. `ObjectHolderRegistry` uses a `HashSet` for its internal collection of callbacks. The `HashSet`'s ordering is unspecified and may change between runs of the JVM, causing the random behavior observed earlier.

The issue was mitigated by changing the exception into a simple `Logger` statement. While this increases the chance of `RegistryObject`s being created for invalid registries, it is an acceptable compromise given that mod developer should be testing their mods and catching these errors themselves, as well as the daunting task of cleaning up and refactoring FML (which is out-of-scope for this PR).

[^1]: FML uses `CompletableFuture<List<Throwable>>`s for the various tasks and hooks during the state transition process, where the contained list value seemed to be originally meant as the list of exceptions thrown during the task or hook. 
However, investigations conclude that later refactors of the state transition code eliminated that use of the list but, for unknown reasons, was kept in as the type parameter to `CompletableFuture`. 
Furthermore, the registry event dispatch hooks kept using the list as it was originally intended, handling any exceptions thrown during the hook into the list. Because the list is never queried in the code, this had the effect of discarding the exception.

sciwhiz12 (changes requested):

patches/minecraft/net/minecraft/core/RegistryAccess.java.patch
SizableShrimp: Per my commit description:
> Deleted hunk that is no longer needed - BuiltinRegistries.ACCESS already wraps around BuiltinRegistries.REGISTRY

sciwhiz12 (approved):

sciwhiz12 (approved):

src/test/java/net/minecraftforge/debug/DeferredRegistryTest.java
```diff
@@ -57,7 +71,19 @@ public DeferredRegistryTest() {
...
         ITEMS.register(modBus);
         CUSTOMS.register(modBus);
+        RECIPE_TYPES.register(modBus);
+        PLACED_FEATURES.register(modBus);
         modBus.addListener(this::gatherData);
+
+        MinecraftForge.EVENT_BUS.addListener(this::serverStarted);
+    }
+
+    public void serverStarted(ServerStartedEvent event) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    public void serverStarted(ServerStartedEvent event)
    {
```

sciwhiz12 (approved):

src/main/java/net/minecraftforge/registries/DeferredRegister.java
```diff
@@ -68,38 +77,89 @@
...
     {
-        return new DeferredRegister<>(null, base, modid);
+        return new DeferredRegister<>(null, base, modid, false);
     }
 
     /**
-     * DeferredRegister factory for custom registries that are made during the {@link NewRegistryEvent} event
-     * or with {@link #makeRegistry(Class, Supplier)}.
+     * DeferredRegister factory for custom forge registries, {@link Registry vanilla registries},
+     * or {@link BuiltinRegistries built-in registries}.
```
sciwhiz12: For clarity reasons, the first sentence of the javadocs for the following four methods should distinguish each method from each other (as this is used as the "summary" of the method in javadoc).

src/main/java/net/minecraftforge/registries/ForgeRegistries.java
```diff
@@ -147,12 +145,13 @@ private static void init() {}
     }
 
     /**
-     * This function is just to make sure static initializers in other classes have run and setup their registries before we query them.
+     * This function is just to make sure static inializers in other classes have run and setup their registries before we query them.
```
sciwhiz12: Could you keep the typo correction?

src/main/java/net/minecraftforge/registries/ObjectHolderRegistry.java
```diff
@@ -171,7 +171,23 @@ public static void applyObjectHolders()
...
+            {
+                objectHolder.accept(filter);
+            }
+            catch (Exception e)
+            {
+                aggregate.addSuppressed(e);
+            }
+        });
+
+        if (aggregate.getSuppressed().length > 0) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
        if (aggregate.getSuppressed().length > 0)
        {
```

src/main/java/net/minecraftforge/registries/RegistryObject.java
```diff
@@ -51,27 +56,75 @@
...
      */
-    public static <T, U extends T> RegistryObject<U> of(final ResourceLocation name, final ResourceKey<? extends Registry<T>> key, String modid) {
-        return new RegistryObject<>(name, key.location(), modid);
+    public static <T, U extends T> RegistryObject<U> of(final ResourceLocation name, final ResourceKey<? extends Registry<T>> registryKey, String modid) {
+        return new RegistryObject<>(name, registryKey.location(), modid, false);
     }
 
     /**
-     * Factory for a {@link RegistryObject} that stores the value of an object in a registry once it is ready.
+     * Factory for a {@link RegistryObject} that stores the value of an object from a registry once it is ready.
```
sciwhiz12: The first sentence should differentiate this factory from the other factories, rather than being uniform.

src/main/java/net/minecraftforge/registries/RegistryObject.java
```diff
@@ -51,27 +56,75 @@
...
-     * @return A {@link RegistryObject} that stores the value of an object in a registry once it is ready
+     * @param name the name of the object to look up in a registry
+     * @param registryKey the key of the registry. Supports lookups on {@link BuiltinRegistries}, {@link Registry}, and {@link RegistryManager#ACTIVE}.
+     * @param modid the mod id calling context
+     * @return a {@link RegistryObject} that stores the value of an object from a registry once it is ready
+     * @see #of(ResourceLocation, ResourceKey, String)
+     * @see #of(ResourceLocation, IForgeRegistry)
+     * @see #of(ResourceLocation, ResourceLocation, String)
+     */
+    public static <T, U extends T> RegistryObject<U> optional(final ResourceLocation name, final ResourceKey<? extends Registry<T>> registryKey, String modid) {
```
sciwhiz12: First, `ofOptional` would probably be more conforming to the current conventional names of the factories (akin to `Optional.of` and `Optional.ofNullable`).
Formatting: braces on newlines.
```suggestion
    public static <T, U extends T> RegistryObject<U> ofOptional(final ResourceLocation name, final ResourceKey<? extends Registry<T>> registryKey, String modid)
    {
```

src/main/java/net/minecraftforge/registries/RegistryObject.java
```diff
@@ -51,27 +56,75 @@
...
+     *
+     * @param name the name of the object to look up in a registry
+     * @param registryName the name of the registry. Supports lookups on {@link BuiltinRegistries}, {@link Registry}, and {@link RegistryManager#ACTIVE}.
+     * @param modid the mod id calling context
+     * @return a {@link RegistryObject} that stores the value of an object from a registry once it is ready
+     * @see #of(ResourceLocation, ResourceLocation, String)
+     * @see #of(ResourceLocation, IForgeRegistry)
+     * @see #of(ResourceLocation, ResourceKey, String)
+     */
+    public static <T, U extends T> RegistryObject<U> optional(final ResourceLocation name, final ResourceLocation registryName, String modid) {
```
sciwhiz12: Same comment for `ofOptional`. Formatting: braces on newlines.
```suggestion
    public static <T, U extends T> RegistryObject<U> optional(final ResourceLocation name, final ResourceLocation registryName, String modid)
    {
```

sciwhiz12 (changes requested):

PR #8528, opened 2022-03-23
[1.18.x] Move to Jetbrains nullability annotations

SizableShrimp:
This PR moves every reference to javax annotations in the Forge codebase (excluding Minecraft source obviously) to Jetbrains nullability annotations. This is just to standardize everything as the current state of the codebase uses a mix of both. All classes affected by this move had an import optimizer run over them which reordered some imports / got rid of wildcard imports.


marchermans (approved):
It took me a hot second to realize that "in forge code" also means in forge code in patches....
But yeah this looks good, the import rework though... Maybe should have been in a separate PR, but what ever.

sciwhiz12 (changes requested):
With the agreement of @OrionDevelopment, this PR will temporarily be put on hold for the following concerns:
 1. The change of nullability annotations from `javax.annotation` to `org.jetbrains.annotations` means that all open (1.18) PRs need to be reviewed and checked for uses of the old annotations and changed accordingly.

    The PR reviews will have to done manually, as some PRs may not use these annotations or use them in a way which does not conflict. Also, the CI will not detect these conflicts unless and until the PR is merged against the latest commit of its base branch.

 2. The reordering of imports in this PR will cause merge conflicts against all open (1.18) PRs which add or modify importss in existing classes. This means that all PRs have to reviewed and changes requested for merge conflicts brought about due to the import reordering.

    This review will be a tad more complicated due to the possibility of merge conflicts caused by the test sourceset's `mods.toml` rather than the import reordering. 

    GitHub only shows the list of files with merge conflicts to those with write access to the PR (the PR author and repository collaborators), so the Triage Team would have to manually confirm the merge conflicts are due to the import reordering (whether by reasoning or by checking the conflicts locally).

 3. The reordering of imports in this PR means the style configuration of your import optimizer (which I presume to be your IDE) is applied to all classes. 

    While this does make import ordering consistent across all classes, alongside the point made above, we would like to codify this import ordering as a Forge style guideline (and produce autoformatter config files for the same), to avoid constant reordering of these imports by PRs from contributors in the future (due to their own IDE's configured import ordering).

Dropping the imports changes from this PR would delay points 2 and 3 until such time that we are confident in performing these changes without affecting all open PRs, such as perhaps during the 1.19 update and patch cycle. 

However, point 1 stands regardless, which alone means the Triage Team must be made aware of when this PR is merged, to plan a review of all open PRs for correcting uses of the old nullability annotations.

PR #8529, opened 2022-03-23
Add support for OnWorldRegistriesLoadedEvent

marchermans:
### New Subsystems:
- Task system
- Async Events (more on those later, the name is misleading)
- OnWorldRegistriesLoadedEvent (First example of an async event, currently gives access to IO and world Registry in one common place)

#### Task system
This system is designed to work together with executors, and event loops to create a way of processing a set of tasks in one particular place while being able to collect the tasks in another.
The system is designed to feel similar to a `ResourceReloadListener` with support for both an async phase as well as a sync phase. But has an API which is easier to understand.

The system comes with a set of configurators which can be used to configure a single task that is supposed to be executed.

Additionally the system supports barrier based splitting of the phases (if a barrier is not used, then the async and sync parts are executed immediately after each other on the relevant executor, regardless of if there are async parts of other tasks still in the queue), this splitting guarantees that first all async parts are executed on its executor, and once those are all completed the sync parts are executed on their own executor.

#### Async Events
The name is misleading here, but I could not find a better one. An async event is a normal event, fired on a normal synchronous event bus, which then has a post-processing phase which is split in an async and a sync part. This functions then similar to a `ResourceReloadListener` but as a post processing phase to the events firing. (Note if the event is cancelled, then no post-processing is supposed to be executed).

#### OnWorldRegistriesLoadedEvent 
This is the first example of the async event system at play. It is fired when the world registries have been loaded (so after there datapack information is overriden etc) but before the control is handed back to world loading. The event gives a modder access to a common place where both the world specific registries exists and are properly loaded (so Tags can be retrieved) while also the required `ResourceManager` for datapack access is available. Additionally the async post processing allows the IO to be performed asynchronously speeding up the loading of the data if a lot of mods use this endpoint.

### The test mod
The included testmod verifies that the event is properly fired and that the async post processing happens in the correct order.

### Patches:
Both `Minecraft.java` and the server `Main.java` are patched to fired the event.

### DIFF
--- a/patches/minecraft/net/minecraft/client/Minecraft.java.patch
+++ b/patches/minecraft/net/minecraft/client/Minecraft.java.patch
@@ -313,8 +313,11 @@
           return p_205187_::m_46934_;
        }, (p_210718_) -> {
           return (p_210712_, p_210713_) -> {
-@@ -1831,10 +_,10 @@
+@@ -1829,12 +_,13 @@
+                return WorldGenSettings.f_64600_.parse(dynamicops1, p_210687_);
+             });
              WorldGenSettings worldgensettings = dataresult.getOrThrow(false, Util.m_137489_("Error reading worldgen settings after loading data packs: ", f_90982_::error));
++            net.minecraftforge.event.async.ForgeAsyncEventFactories.doFireWorldRegistriesLoadedEvent(registryaccess$writable, dynamicops1, p_210712_, this, this::m_18701_);
              return Pair.of(new PrimaryLevelData(p_205187_, worldgensettings, dataresult.lifecycle()), registryaccess$writable.m_203557_());
           };
 -      }, false, Minecraft.ExperimentalDialogType.CREATE);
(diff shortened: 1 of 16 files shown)


TheCurle:
What specific advantages does this have over #8341?

marchermans:
The idea is different, that PR is focused on dealing with Biomes and their friends.

However this PR is in particular related to handling IO related to custom objects which either reference Tags or other objects from the registry, basically it is meant to replace the StaticTags access you had previously while deserializing custom objects that use tags.

In particular I previously deserialized custom recipe representations for Aequivaleo that way: ResourceLocation of item tag -> Use StaticTags to get the tag reference out.

That does not work anymore since the tags are not static anymore but need to be retrieved from the registry properly, and only completely after the tags are loaded. 
Additionally to be able to properly load the data from the correct datapacks access to the `ResourceManager` is needed, hence the "async" addition so that files can be read properly in parallel.

PR #8530, opened 2022-03-23
[1.18.x] Fix Banner Pattern Extension

### LINKED ISSUE #8518: Banner Patterns Do Not Work for a Few Reasons
**Minecraft Version:** 1.16.5, 1.18.2

**Forge Version:** 36.2.31+

**Steps to Reproduce:**
Issue 1
 1. Add a new `BannerPattern` using `BannerPattern#create`.
 2. Compare the size of `BannerPattern#COUNT` to the length of `BannerPattern#values`.
 3. Notice the inconsistent difference between the two values.
 
 Issue 2 (After fixing Issue 1)
 1. Add a new `BannerPattern` using `BannerPattern#create` that is not a banner item.
 2. Load the game and open a loom.
 3. Notice that the `GLOBE` banner pattern now appears on the banner pattern selection menu.

Issue 3 (After fixing Issue 2)
 1. Add a new `BannerPattern` using `BannerPattern#create` that is not a banner item.
 2. Add a texture at `assets/<modid>/entity/banner/<banner_pattern_name>`.
 3. Load the game and open a loom.
 4. Notice that the banner pattern texture is missing.

**Description of issue:**
1. `BannerPattern` caches the result of the values to generate a static value representing the number of patterns, which are part of an item, and the available patterns to choose from in the loom. Since all of these values are static init, as soon as the class is initialized, the value is calculated and stable. As such, there can only be 41 patterns within the game.
2. `BannerPattern` displays the patterns in the loom using an index lookup based upon the registered order of the enum constants. As such, if the above case wasn't an issue, the pattern that would appear in the loom would be the `GLOBE` even though its a pattern item. As such, only banner patterns that are items can be properly added if issue 1 was fixed.
3. The location of the banner pattern texture will always exist within the `minecraft` namespace. This is because `BannerPattern#location` uses the inputted filename as the end string and does not check for a namespace. This additionally affects how the translation key appears within `BannerItem` since it hardcodes the namespace to `minecraft`.

ChampionAsh5357:
Closes #8518

Banner patterns to cache the values and counts of its available patterns and non pattern item instances as soon as the enum is first loaded. This PR changes that behavior to cache the value on first access which happens after enums are made extensible.

Additionally, this fixes some bugs within the loom where if pattern and non pattern items are not ordered, it would display the wrong pattern and store it incorrectly on the item.

The test mod creates three banner patterns with the second pattern also being an item. Open a loom and try applying the item or the patterns and notice that they appear on the menu and are also able to be properly put on a banner.


src/test/java/net/minecraftforge/debug/block/BannerPatternTest.java
```diff
@@ -0,0 +1,30 @@
+package net.minecraftforge.debug.block;
```
sciwhiz12, line 6: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

sciwhiz12 (changes requested):
Tested locally and it works fine. 

The fields `COUNT`, `PATTERN_ITEM_COUNT`, and `AVAILABLE_PATTERNS` should be deprecated with a comment redirecting to the relevant `ForgeHook` method (and a small blurb about subtracting one, in the case of `AVAILABLE_PATTERNS`).

sciwhiz12 (approved):

PR #8533, opened 2022-03-24
[1.18.x] Fix lost validation of registry data on singleplayer world load

SizableShrimp:
During the 1.18.2 porting process, the piece of code which calls `LevelStorageSource#readAdditionalLevelSaveData` during a singleplayer world load in `Minecraft` was commented out with a relevant `// TODO` referencing the changed world loading process. The TODO has been removed and the patch uncommented. The patch line modifying the previously existing `registryHolder` parameter has been removed and at this time, I am not sure if it needs to be reintroduced nor where that would be. However, the important first line has been put back in place. This means `GameData#injectSnapshot` is now run for singleplayer worlds again like in previous versions, and `RegistryEvent.MissingMappings` is now correctly fired. The EXC file has been updated to include the relevant `Minecraft#doLoadLevel` patched method which takes a `creating` boolean. I also took this time to fix up any invalid excs and update SRG parameters where it was necessary.


marchermans (approved):

patches/minecraft/net/minecraft/server/Main.java.patch
```diff
@@ -101,8 +101,7 @@
 +               // FORGE: Halting as GameTestServer will cause issues as it always calls System#exit on both crash and normal exit, so skip it
 +               if (!(dedicatedserver instanceof net.minecraft.gametest.framework.GameTestServer))
                 dedicatedserver.m_7570_(true);
-+               //TODO-PATCHING: Figure out what needs to be done here after the logging update.
-+               //LogManager.shutdown(); // we're manually managing the logging shutdown on the server. Make sure we do it here at the end.
++               org.apache.logging.log4j.LogManager.shutdown(); // we're manually managing the logging shutdown on the server. Make sure we do it here at the end.
```
TheCurle, line 103: Mojang use SLF4J now. We should see if there's a way to properly decouple this from hardcoded Log4j paths.

SizableShrimp: I believe this decoupling should be left to another PR and this just restore previous behavior that still works *at the moment*. There are still many usages of LogManager across the Forge codebase and the use of log4j markers instead of slf4j meaning not everything can easily be moved to LogUtils.

sciwhiz12: I agree with @SizableShrimp here. SLF4J is only a facade, and needs a backing provider to properly function. Currently, Log4j is that backing provider. Until such time that Mojang replaces Log4j with another logging framework (like logback) and Forge also moves away from Log4j to SLF4J, we should keep (or in this case, return) this shutdown call to Log4j's `LogManager`.

(Moving Forge away from Log4j to SLF4J should be done at another time, not in this PR.)

SizableShrimp:
After analyzing the changes made to registry loading in 1.18.2 and the goal of the previous code delaying the initialization of `RegistryAccess#builtin()` in 1.18.1, I have come to the conclusion that this code is no longer necessary.

With some simple breakpointing, one can determine that `GameData#injectSnapshot` always fires before `RegistryAccess#builtinCopy` is fired for a world load. (The specific invocation of `builtinCopy` can be found in `WorldStem$WorldDataSupplier#loadFromWorld`.)

Therefore, I think deleting all the patches and TODOs relating to deferring this registry copy can be safely deleted. As always, the feature can be reintroduced later if it is deemed necessary then.

src/main/resources/forge.exc
```diff
@@ -38,8 +38,8 @@ net/minecraft/world/entity/npc/AbstractVillager.changeDimension(Lnet/minecraft/s
 net/minecraft/world/entity/player/Player.getDigSpeed(Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/core/BlockPos;)F=|p_36282_,pos
 net/minecraft/world/entity/projectile/ThrownEnderpearl.changeDimension(Lnet/minecraft/server/level/ServerLevel;Lnet/minecraftforge/common/util/ITeleporter;)Lnet/minecraft/world/entity/Entity;=|p_37506_,teleporter
 net/minecraft/world/item/BoneMealItem.applyBonemeal(Lnet/minecraft/world/item/ItemStack;Lnet/minecraft/world/level/Level;Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/entity/player/Player;)Z=|p_40628_,p_40629_,p_40630_,player
-net/minecraft/world/item/DyeableHorseArmorItem.<init>(ILnet/minecraft/resources/ResourceLocation;Lnet/minecraft/world/item/Item$Properties;)V=|p_41110_,p_41111_,p_41112_
-net/minecraft/world/item/HorseArmorItem.<init>(ILnet/minecraft/resources/ResourceLocation;Lnet/minecraft/world/item/Item$Properties;)V=|p_41364_,p_41111_,p_41366_
+net/minecraft/world/item/DyeableHorseArmorItem.<init>(ILnet/minecraft/resources/ResourceLocation;Lnet/minecraft/world/item/Item$Properties;)V=|p_41110_,p_41365_,p_41112_
```
sciwhiz12: Why is this using the SRG identifier `p_41365_` from the superclass method, rather than the SRG identifier `p_41111_` from the sibling constructor (following the pattern from the superclass)?

sciwhiz12 (approved):

PR #8534, opened 2022-03-24
[1.18.x] Allow Brewing Stand to Properly Handle Stacked Items in Bottle Slots

### LINKED ISSUE #8521: Brewing stand doesn't allow placing stacked items in the bottle slots without unstacking them manually first
**Minecraft Version:** 1.18.2

**Forge Version:** 40.0.19

**Logs:** no relevant log

**Steps to Reproduce:**
 1. Create a recipe with a stackable item [such as honey bottles] as the potion input
 2. Attempt to place a stack of more than one of this item in the brewing stand's bottle input slots [left click or right click while holding the items, or shift click from inventory]
 3. Expected: places a single item from the stack in the slot (or one in each slot for shift click), shrinking the source stack by one [or up to three for the shift click case]
 4. Actual: does not apparently accept the item at all. The recipe itself works, and if the stack is separated into single-item stacks before attempting to insert they work normally [side note: I got inconsistent results trying to insert through the item handler capability, though hoppers work fine.]

**Description of issue:**
The code responsible for this issue is in BrewingStandMenu, but I don't know the best way to approach fixing it while keeping patch size in mind.

This issue does not arise in vanilla because in vanilla the brewing stand only accepts potion bottles in these slots.

ChampionAsh5357:
Closes #8521.

Removes the count checks from the brewing menu and registry to allow stacked items to be properly handled. This has no adverse effects since the menu handles all splitting logic normally without issue.

### DIFF
--- a/patches/minecraft/net/minecraft/world/inventory/BrewingStandMenu.java.patch
+++ b/patches/minecraft/net/minecraft/world/inventory/BrewingStandMenu.java.patch
@@ -1,5 +1,14 @@
 --- a/net/minecraft/world/inventory/BrewingStandMenu.java
 +++ b/net/minecraft/world/inventory/BrewingStandMenu.java
+@@ -75,7 +_,7 @@
+                if (!this.m_38903_(itemstack1, 3, 4, false)) {
+                   return ItemStack.f_41583_;
+                }
+-            } else if (BrewingStandMenu.PotionSlot.m_39133_(itemstack) && itemstack.m_41613_() == 1) {
++            } else if (BrewingStandMenu.PotionSlot.m_39133_(itemstack)) {
+                if (!this.m_38903_(itemstack1, 0, 3, false)) {
+                   return ItemStack.f_41583_;
+                }
 @@ -146,7 +_,7 @@
        }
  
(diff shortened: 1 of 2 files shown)


marchermans (approved):

patches/minecraft/net/minecraft/world/inventory/BrewingStandMenu.java.patch
```diff
@@ -1,5 +1,14 @@
 --- a/net/minecraft/world/inventory/BrewingStandMenu.java
 +++ b/net/minecraft/world/inventory/BrewingStandMenu.java
+@@ -75,7 +_,7 @@
+                if (!this.m_38903_(itemstack1, 3, 4, false)) {
+                   return ItemStack.f_41583_;
+                }
+-            } else if (BrewingStandMenu.PotionSlot.m_39133_(itemstack) && itemstack.m_41613_() == 1) {
```
marchermans, line 7: Does this have other concequences?

ChampionAsh5357: As far as I can tell, no. The `m_38903_` calls a safe insert method which handles proper splitting based on the slot's maximum size. So, there are no standing consequences in relation to the single item slot output. The only major consequence would be if someone tried to modify the brewing stand to take in more than one input; however, that would be on the responsibility of the modder to fix that issue as they tried to coremod that existence.

sciwhiz12 (approved):

PR #8536, opened 2022-03-24
[1.18.2] Make GameData methods caller sensitive and add better frozen registry error messages

Matyrobbrt:
This PR makes a few public methods in `GameData` caller sensitive, in order to prevent modders calling them. The frozen registry crash is also improved by logging a message which provides modders basic information on how to register objects properly, and tell players the cause of the crash, with the mod to blame.


patches/minecraft/net/minecraft/core/MappedRegistry.java.patch
```diff
@@ -1,17 +1,35 @@
 --- a/net/minecraft/core/MappedRegistry.java
 +++ b/net/minecraft/core/MappedRegistry.java
-@@ -279,6 +_,14 @@
-       return this.f_205842_.containsKey(p_175392_);
+@@ -75,7 +_,13 @@
+ 
+    private void m_205921_(ResourceKey<T> p_205922_) {
+       if (this.f_205845_) {
+-         throw new IllegalStateException("Registry is already frozen (trying to add key " + p_205922_ + ")");
```
SizableShrimp, line 7: This should still be an exception. Warning changes the codepath.

src/main/java/net/minecraftforge/internal/ForgeExceptionFactories.java
```diff
@@ -0,0 +1,11 @@
+/*
+ * Minecraft Forge - Forge Development LLC
+ * SPDX-License-Identifier: LGPL-2.1-only
+ */
+
+package net.minecraftforge.internal;
+
+public class ForgeExceptionFactories {
+
+	public static final IllegalCallerException INTERNAL_METHOD = new IllegalCallerException("This is an internal forge method and cannot be used by mods");
```
SizableShrimp: This whole class for one field seems a bit weird...

SizableShrimp (changes requested):

src/main/java/net/minecraftforge/internal/ForgeExceptionFactories.java
KnightMiner: Also, won't static init here really mess with the displayed stacktrace?

src/main/java/net/minecraftforge/server/ServerLifecycleHooks.java
```diff
@@ -113,6 +113,10 @@ public static void expectServerStopped()
 
     public static void handleServerStopped(final MinecraftServer server)
     {
+        if (StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE).getCallerClass() != MinecraftServer.class)
+        {
+            throw new IllegalCallerException("This is an internal forge method and cannot be used by mods");
```
PaintNinja: Although the ServerStoppedEvent is just below here in the code, it might still be a good idea to mention it in the exception in case they don’t look at this method’s contents.

Shadows-of-Fire:
Merge conflicts need to be resolved for this PR to move forward.

Matyrobbrt:
Conflicts should be fixed now.

shartte:
Ignoring how StackWalker is a crappy modder-hostile hack, Forge could also just prevent the initial freeze and only freeze *once* after all register events have fired successfully.

We currently do our unfreeze/freeze dance because we HAVE to have initialized some of our game object classes (mind you, not registered, just ctor'd) before going into the registry event phase due to allowing addons access to them at otherwise undefined points during startup. And again, this is not about registering them, but as a side-effect of any Item/Block ctor creating holders in their constructor, impossible BEFORE the first register event.

KnightMiner:
@shartte would be worth discussing on discord why you need to allow access pre-register event to your items (easier  to write what we conclude than bog down this PR discussion with the full discussion). I cannot think of a usecase where addons need items to exist before registration events, I write a mod with a ton of addons. To me, sounds like you intentionally wrote APIs that require item access instead of using suppliers as you should so registry event order does not matter.

shartte:
> @shartte would be worth discussing on discord why you need to allow access pre-register event to your items (easier to write what we conclude than bog down this PR discussion with the full discussion). I cannot think of a usecase where addons need items to exist before registration events, I write a mod with a ton of addons. To me, sounds like you intentionally wrote APIs that require item access instead of using suppliers as you should so registry event order does not matter.

I don't *have* to, obviously. But we've put in the effort to make our API thread-safe after Forge made Mod constructors run in parallel (I still don't get why exactly). And I honestly don't have the time to rewrite the mod yet again, not because of technical necessity but rather because "the way to do things" gets narrowed down yet again. 

Why is it necessary to keep Vanilla's freeze in Bootstrap to unfreeze and refreeze again later? It seems much more sensible to just patch out the first freeze in Bootstrap, don't even patch-in the unfreeze method in the fist place and simply freeze once directly after the register events finish, since items/blocks constructed after that can't be registered anyway.

TheCurle:
The end goal is to be able to completely ~~eliminate~~ wipe clean the registries and re-run the register events to create a ~~fresh snapshot~~ new registry set, for example with mods added or removed at runtime.
Allowing people to register any time except the registry events is actively against our goals.

There's a reason we made the system this way, though it's far in the future. That doesn't mean we should stop moving towards that goal, though.

shartte:
> The end goal is to be able to completely ~eliminate~ wipe clean the registries and re-run the register events to create a ~fresh snapshot~ new registry set, for example with mods added or removed at runtime. Allowing people to register any time except the registry events is actively against our goals.
> 
> There's a reason we made the system this way, though it's far in the future. That doesn't mean we should stop moving towards that goal, though.

Pursuing this will leave yet another batch of mods behind on old version. Not all maintainers can keep investing insane amounts of time rewriting again and again and again to follow your vision. They already have to do it often enough to follow Mojang's. You should be conscious of that.

TheCurle:
I feel like this is venturing into "Mojang already updates the game a lot, why are you updating Forge so much too?" territory.
Minecraft is an unstable platform. Extensive and widespread changes happen extremely often. Modders are already aware of that.

1.16.2 and 1.18.2 are prime examples of this, breaking worldgen and registries, tags etc in "minor updates". 
Forge's relatively slow progress towards an extremely highly requested technical feature is nothing compared to that.

If your opinion is that we should stay exactly the same and wallow in our own technical debt brought about by the exact changes that Mojang make that you mentioned, then it is an opinion we note and summarily ignore.

Technici4n:
Shartte's point is that we had a working mod in 1.18.1, and we don't want to spend days or even weeks refactoring some stuff and breaking the existing and working API, when that time could be spent implementing features and fixing bugs.

This has been a constant with each of the previous forge updates, where we first ported to the new MC version on fabric where we only had to deal with MC-induced breakage usually. And then we'd have to spend a few more hours fixing random forge-introduced changes that could have been avoided.

This PR per se will just break our mod, yet it doesn't solve anything since any mod wishing to unfreeze registries for object creation can and will find a way. It will just force us to spend even more time fixing the forge branch instead of adding features.

As for adding mods at runtime, you can always dream, but I don't see us writing an API that supports clearing various static registries and re-populating them. And forcing everyone to break their mod for a vague and long term goal that's not nearly in sight, and imo unattainable, doesn't seem like a great idea.

My 0.02$.

SizableShrimp:
Forge is not Minecraft. Minecraft is Minecraft. Maintaining stuff across platforms requires modifications per platform. That's just how it works. Intentionally undoing Mojang's changes to freeze registries is a step in the wrong direction, and it won't be done to make the lives of modder's easier. Modding is hard. Undoing this change does no good, it causes more issues than just fixing your mod(s) to work with the new stuff. If you don't have enough time/energy to keep up with Forge/Minecraft changes and maintain, then that's just how it is...

KnightMiner:
Also worth noting, Forge has not been pushing towards using registry events just for 1.18.2. They have been pushing for using registry events since 1.12 (probably earlier, but that is long enough ago to make my point). You likely needed to rewrite your mod to make it compatible with 1.16, and you made the choice then to intentionally ignore Forge telling you to create items during the events. Thus, the extra work caused by Mojang freezing registries could have been entirely avoided if you just listened to the recommendations from Forge back when they were optional (this registry freezing issue caused me no trouble for instance as I have been using registry events since 1.12).

LexManos:
TLDR: Modding is hard, things change, You have to put in effort to updating.
This PR was made to stop lazy people from doing the exact opposite of what we've been telling them to do for literal years.
Closing and locking this because it's become nothing but complaining and lack of any effort to understand why we make changes.

I called it from the beginning that any effort to try and get people to do things correctly would result in this. Which is why this PR will never be accepted. Because if it was, it'd just cause more drama, and we've have the loud minority going else ware to cause us headaches.

PR #8538, opened 2022-03-26
[1.19.x] Rework forge registry system and move away from IForgeRegistryEntry

### LINKED ISSUE #7969: RFC - Remove ForgeRegistryEntry and ForgeRegistries being Type bound

### LINKED ISSUE #8513: MissingMappings Event Bus
**Minecraft Version:** 1.18.2

**Forge Version:** 40.0.15

**Description of issue:**
The MissingMappings event is fired on the MinecraftForge event bus rather than the Mod event bus, even though it inherits from RegistryEvent, which is fired on the mod bus. I feel as though it should either not inherit from a class that implements IModEvent or that it should fire on the mod bus.

SizableShrimp:
This draft PR serves as the basis for the rework of forge's registry system to be more in line with vanilla registries by supporting registries of interfaces and no longer requiring a supertype of IForgeRegistryEntry. The existence of this PR is largely for code review and visibility, as it is targeted towards 1.19 with the goal of being ported and merged between the launch of Minecraft 1.19 and the release of forge 1.19. Most notable changes should be explained in commit descriptions.

Fixes #8513.
Fixes #7969.


patches/minecraft/net/minecraft/world/level/block/state/BlockBehaviour.java.patch
```diff
@@ -1,21 +1,19 @@
 --- a/net/minecraft/world/level/block/state/BlockBehaviour.java
 +++ b/net/minecraft/world/level/block/state/BlockBehaviour.java
-@@ -66,7 +_,8 @@
+@@ -66,6 +_,7 @@
  import net.minecraft.world.phys.shapes.Shapes;
  import net.minecraft.world.phys.shapes.VoxelShape;
  
--public abstract class BlockBehaviour {
 +//TODO, Delegates are weird here now, because Block extends this.
```
sciwhiz12: Looks like the TODO can be removed now.

patches/minecraft/net/minecraft/world/level/block/state/BlockBehaviour.java.patch
```diff
@@ -1,21 +1,19 @@
...
     protected static final Direction[] f_60441_ = new Direction[]{Direction.WEST, Direction.EAST, Direction.NORTH, Direction.SOUTH, Direction.DOWN, Direction.UP};
     protected final Material f_60442_;
-    protected final boolean f_60443_;
 @@ -93,6 +_,8 @@
        this.f_60449_ = p_60452_.f_60893_;
        this.f_60438_ = p_60452_.f_60903_;
        this.f_60439_ = p_60452_;
 +      final ResourceLocation lootTableCache = p_60452_.f_60894_;
-+      this.lootTableSupplier = lootTableCache != null ? () -> lootTableCache : p_60452_.lootTableSupplier != null ? p_60452_.lootTableSupplier : () -> new ResourceLocation(this.getRegistryName().m_135827_(), "blocks/" + this.getRegistryName().m_135815_());
++      this.lootTableSupplier = lootTableCache != null ? () -> lootTableCache : p_60452_.lootTableSupplier != null ? p_60452_.lootTableSupplier : () -> new ResourceLocation(net.minecraftforge.registries.ForgeRegistries.BLOCKS.getKey((Block)this).m_135827_(), "blocks/" + net.minecraftforge.registries.ForgeRegistries.BLOCKS.getKey((Block)this).m_135815_());
```
sciwhiz12: Use a local variable for the registry name, to avoid doubling the query.

patches/minecraft/net/minecraft/world/level/block/FlowerPotBlock.java.patch
```diff
@@ -42,7 +42,7 @@
        ItemStack itemstack = p_53543_.m_21120_(p_53544_);
        Item item = itemstack.m_41720_();
 -      BlockState blockstate = (item instanceof BlockItem ? f_53524_.getOrDefault(((BlockItem)item).m_40614_(), Blocks.f_50016_) : Blocks.f_50016_).m_49966_();
-+      BlockState blockstate = (item instanceof BlockItem ? getEmptyPot().fullPots.getOrDefault(((BlockItem)item).m_40614_().getRegistryName(), Blocks.f_50016_.delegate).get() : Blocks.f_50016_).m_49966_();
++      BlockState blockstate = (item instanceof BlockItem ? getEmptyPot().fullPots.getOrDefault(net.minecraftforge.registries.ForgeRegistries.BLOCKS.getKey(((BlockItem)item).m_40614_()), net.minecraftforge.registries.ForgeRegistries.BLOCKS.getDelegateOrThrow(Blocks.f_50016_)).get() : Blocks.f_50016_).m_49966_();
```
sciwhiz12, line 45: Use a local variable for the registry name, to avoid doubling the query.

src/main/java/net/minecraftforge/registries/IdMappingEvent.java
```diff
@@ -0,0 +1,72 @@
+package net.minecraftforge.registries;
```
sciwhiz12, line 6: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/registries/MissingMappingsEvent.java
```diff
@@ -0,0 +1,171 @@
+package net.minecraftforge.registries;
```
sciwhiz12, line 6: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/registries/IdMappingEvent.java
```diff
@@ -0,0 +1,72 @@
...
+ * Fired on the {@link net.minecraftforge.common.MinecraftForge#EVENT_BUS}
+ */
+public class IdMappingEvent extends Event
+{
+    public static class ModRemapping
+    {
+        public final ResourceLocation registry;
+        public final ResourceLocation key;
+        public final int oldId;
+        public final int newId;
```
sciwhiz12, line 38: Avoid exposing fields; prefer using getters.

src/main/java/net/minecraftforge/registries/IdMappingEvent.java
```diff
@@ -0,0 +1,72 @@
...
+            this.key = key;
+            this.oldId = oldId;
+            this.newId = newId;
+        }
+    }
+
+    private final Map<ResourceLocation, ImmutableList<ModRemapping>> remaps;
+    private final ImmutableSet<ResourceLocation> keys;
+
+    public final boolean isFrozen;
```
sciwhiz12: Avoid exposing fields; prefer using getters.

src/main/java/net/minecraftforge/registries/MissingMappingsEvent.java
```diff
@@ -0,0 +1,171 @@
...
+         */
+        REMAP
+    }
+
+    public static class Mapping<T> implements Comparable<Mapping<T>>
+    {
+        public final IForgeRegistry<T> registry;
+        private final IForgeRegistry<T> pool;
+        public final ResourceLocation key;
+        public final int id;
```
sciwhiz12: Avoid exposing fields; prefer using getters.

src/main/java/net/minecraftforge/registries/MissingMappingsEvent.java
```diff
@@ -0,0 +1,171 @@
...
+        // internal
+        public Action getAction()
+        {
+            return this.action;
+        }
+
+        public T getTarget()
+        {
+            return target;
+        }
```
sciwhiz12: Could these methods be made package private?

src/main/java/net/minecraftforge/registries/ObjectHolder.java
```diff
@@ -15,15 +17,23 @@
...
+    /**
+     * The name of the registry to load registry entries from.
+     * This string is parsed as a {@link ResourceLocation} and can contain a namespace.
+     *
+     * @return The registry name
+     */
+    String registryName();
+
     /**
      * If used on a class, this represents a modid only.
```
sciwhiz12: This javadoc needs to be updated, as the annotation can only be applied to fields now.

src/main/java/net/minecraftforge/registries/RegisterEvent.java
```diff
@@ -0,0 +1,89 @@
...
+ * Fired on the {@link IModBusEvent mod bus}.
+ *
+ * @see #register(ResourceKey, ResourceLocation, Object)
+ */
+public class RegisterEvent extends Event implements IModBusEvent
+{
+    @NotNull
+    private final ResourceKey<? extends Registry<?>> registryKey;
+    @Nullable
+    final ForgeRegistry<?> forgeRegistry;
```
sciwhiz12, line 33: Why is this package-private, when this already has a getter and its sibling fields are `private`?

src/main/java/net/minecraftforge/registries/RegisterEvent.java
```diff
@@ -0,0 +1,89 @@
+package net.minecraftforge.registries;
```
sciwhiz12, line 6: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/registries/RegisterEvent.java
```diff
@@ -0,0 +1,89 @@
...
+    public <T> Registry<T> getVanillaRegistry()
+    {
+        return (Registry<T>) vanillaRegistry;
+    }
+
+    @Override
+    public String toString()
+    {
+        return "RegisterEvent";
+    }
```
sciwhiz12, line 113: Where is this used? 🤔 If this is kept, it may be best to add the registry name to this: `RegisterEvent{<registry name>}`.

src/test/java/net/minecraftforge/debug/block/FlowerPotTest.java
```diff
@@ -30,17 +30,20 @@
...
+            BLOCK_ID + "_oak", () -> new FlowerPotBlock(EMPTY_FLOWER_POT, ForgeRegistries.BLOCKS.getDelegateOrThrow(Blocks.OAK_SAPLING), Block.Properties.copy(Blocks.FLOWER_POT)));
     
     static {
         ITEMS.register(BLOCK_ID, () -> new BlockItem(EMPTY_FLOWER_POT.get(), new Item.Properties()));
     }
     
     @SubscribeEvent
-    public static void onItemRegister(RegistryEvent.Register<Item> event) {
-        EMPTY_FLOWER_POT.get().addPlant(Blocks.OAK_SAPLING.getRegistryName(), OAK_FLOWER_POT);
+    public static void onItemRegister(RegisterEvent event) {
```
sciwhiz12: Formatting: braces on newlines. (Might as well fix it since you're already touching it.)
```suggestion
    public static void onItemRegister(RegisterEvent event)
    {
```

src/main/java/net/minecraftforge/registries/RegisterEvent.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Registers the value with the given name to the stored registry if the provided registry key matches this event's registry key.
+     *
+     * @param registryKey The key of the registry to register the value to
+     * @param name The name of the object to register as its key
+     * @param value The object value
+     * @param <T> The type of the registry
+     */
```
sciwhiz12: The test mods show that the convention for registering using this event is to first check if the registry key in this event matches the key they expect, and calling this event if so. To my understanding, this is to avoid the creation of the to-be-registered object.

This convention should be documented in the javadoc of this class. Additionally or alternatively, a similar method could be added which accepts a `Supplier<T>` (though, this will still incur the cost of the `ResourceLocation` creation).

src/test/java/net/minecraftforge/debug/recipe/recipebook/RecipeBookExtensionTest.java
```diff
@@ -49,24 +47,18 @@
...
         RECIPE_SERIALIZER.register(FMLJavaModLoadingContext.get().getModEventBus());
         MENU_TYPE.register(FMLJavaModLoadingContext.get().getModEventBus());
         MinecraftForge.EVENT_BUS.addListener(this::onRightClick);
     }
 
-    private void commonSetup(FMLCommonSetupEvent event)
-    {
-        event.enqueueWork(() ->
-        {
-            RecipeType.register(getId("test_recipe").toString());
```
sciwhiz12, line 66: Should `RecipeType.register` and similar methods be marked deprecated, to warn modders who are currently calling it to move to `DeferredRegister`s instead (as seen in this test mod)?

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -226,48 +226,50 @@ public void missingSoundMapping(RegistryEvent.MissingMappings<SoundEvent> event)
     }
 
     // done in an event instead of deferred to only enable if a mod requests it
-    public void registerFluids(RegistryEvent.Register<Fluid> event)
+    public void registerFluids(RegisterEvent event)
```
sciwhiz12, line 241: Should these three `register*` methods be kept separate, or consolidated into a single method?

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -226,48 +226,50 @@ public void missingSoundMapping(RegistryEvent.MissingMappings<SoundEvent> event)
...
     {
-        // Ignore the event itself: this is done only not to statically initialize our custom LootConditionType
-        Registry.register(Registry.LOOT_CONDITION_TYPE, new ResourceLocation("forge:loot_table_id"), LootTableIdCondition.LOOT_TABLE_ID);
-        Registry.register(Registry.LOOT_CONDITION_TYPE, new ResourceLocation("forge:can_tool_perform_action"), CanToolPerformAction.LOOT_CONDITION_TYPE);
+        if (event.getRegistryKey().equals(ForgeRegistries.Keys.LOOT_MODIFIER_SERIALIZERS))
+        {
+            // Ignore the event itself: this is done only not to statically initialize our custom LootConditionType
+            Registry.register(Registry.LOOT_CONDITION_TYPE, new ResourceLocation("forge:loot_table_id"), LootTableIdCondition.LOOT_TABLE_ID);
+            Registry.register(Registry.LOOT_CONDITION_TYPE, new ResourceLocation("forge:can_tool_perform_action"), CanToolPerformAction.LOOT_CONDITION_TYPE);
+        }
```
sciwhiz12: Looks like this existed as a case of registering objects to vanilla registries in the `RegistryEvent.Register` of the closest type that is related to what's being registered. This should be modified to listen for the `LOOT_CONDITION_TYPE` event and register using the event's `register` method.

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -226,48 +226,50 @@ public void missingSoundMapping(RegistryEvent.MissingMappings<SoundEvent> event)
...
+            CraftingHelper.register(OrCondition.Serializer.INSTANCE);
+            CraftingHelper.register(TrueCondition.Serializer.INSTANCE);
+            CraftingHelper.register(TagEmptyCondition.Serializer.INSTANCE);
+
+            CraftingHelper.register(new ResourceLocation("forge", "compound"), CompoundIngredient.Serializer.INSTANCE);
+            CraftingHelper.register(new ResourceLocation("forge", "nbt"), NBTIngredient.Serializer.INSTANCE);
+            CraftingHelper.register(new ResourceLocation("forge", "partial_nbt"), PartialNBTIngredient.Serializer.INSTANCE);
+            CraftingHelper.register(new ResourceLocation("forge", "difference"), DifferenceIngredient.Serializer.INSTANCE);
+            CraftingHelper.register(new ResourceLocation("forge", "intersection"), IntersectionIngredient.Serializer.INSTANCE);
+            CraftingHelper.register(new ResourceLocation("minecraft", "item"), VanillaIngredientSerializer.INSTANCE);
```
sciwhiz12, line 272: Perhaps `IConditionSerializer` and `IIngredientSerializer` could be made into proper (unsynced, unsaved) Forge registries, instead of being their own separate special registries in `CraftingHelper`.

src/main/java/net/minecraftforge/registries/ObjectHolderRegistry.java
```diff
@@ -80,24 +91,46 @@ public static void findObjectHolders()
...
 
-        // double pass - get all the class level annotations first, then the field level annotations
-        annotations.stream().filter(a -> OBJECT_HOLDER.equals(a.annotationType())).filter(a -> a.targetType() == ElementType.TYPE)
-        .forEach(data -> scanTarget(classModIds, classCache, data.clazz(), null, (String)data.annotationData().get("value"), true, data.clazz().getClassName().startsWith("net.minecraft.")));
+        // Validate all the vanilla class-level object holders then scan those first
+        VANILLA_OBJECT_HOLDERS.forEach(data -> {
+            try
+            {
+                Class<?> holderClass = Class.forName(data.holderClass(), true, ObjectHolderRegistry.class.getClassLoader());
+                Class<?> registryClass = Class.forName(data.registryType(), true, ObjectHolderRegistry.class.getClassLoader());
```
sciwhiz12, line 108: Why do we need to specify the class loader? Is the classloader of this class different from the thread context classloader? Add a comment explaining why.

src/main/java/net/minecraftforge/registries/ObjectHolderRef.java
```diff
@@ -69,55 +54,32 @@ public ObjectHolderRef(Field field, ResourceLocation injectedObject)
...
             catch (ResourceLocationException e)
             {
                 throw new IllegalArgumentException("Invalid @ObjectHolder annotation on \"" + field.toString() + "\"", e);
             }
         }
 
-        if (this.injectedObject == null || !isValid())
-        {
+        if (injectedObjectName == null)
             throw new IllegalStateException(String.format(Locale.ENGLISH, "The ObjectHolder annotation cannot apply to a field that does not map to a registry. Ensure the registry was created during the RegistryEvent.NewRegistry event. (found : %s at %s.%s)", field.getType().getName(), field.getDeclaringClass().getName(), field.getName()));
```
sciwhiz12: The exception message needs to be updated to refer to `NewRegistryEvent`.

fmlloader/src/main/java/net/minecraftforge/fml/common/asm/ObjectHolderDefinalize.java
```diff
@@ -109,4 +120,5 @@ public int processClassWithFlags(final Phase phase, final ClassNode classNode, f
         return changes.get() ? ComputeFlags.SIMPLE_REWRITE : ComputeFlags.NO_REWRITE;
     }
 
+    public record VanillaObjectHolderData(String holderClass, String registryName, String registryType) {}
```
sciwhiz12: It may be worth just duplicating this in `ObjectHolderRegistry` and marking this class as `private`.

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -166,25 +168,19 @@ public ResourceLocation getRegistryName()
...
-        for (V value : values)
-            register(value);
-    }
+    // @SuppressWarnings("unchecked")
+    // @Override
+    // public void registerAll(V... values)
+    // {
+    //     for (V value : values)
+    //         register(value);
+    // }
```
sciwhiz12: Why commented out?

sciwhiz12 (changes requested):
As a general question of style, the various `RegistryEvent` listeners in the test mods (and `ForgeMod`) could be modified to use short-circuiting `if`s at the start, to reduce the level of indentation in the methods.

It may be worth considering going through all the `ForgeRegistries` fields and aligning their names to their keys and types.

patches/minecraft/net/minecraft/world/level/block/FlowerPotBlock.java.patch
SizableShrimp: There's no double query here.

sciwhiz12: Ah, my mistake. Carry on. :shipit:

src/main/java/net/minecraftforge/common/ForgeMod.java
sciwhiz12: Nevermind this comment, I've been told that `IConditionSerializer` and `IIngredientSerializer` are to remain as their own special, lighter registries in `CraftingHelper`.

src/main/java/net/minecraftforge/registries/IdMappingEvent.java
SizableShrimp: It's basically just a record with a private constructor. I think this is fine personally.

src/main/java/net/minecraftforge/registries/RegisterEvent.java
SizableShrimp: ForgeRegistry stored in field vs IForgeRegistry in getter. Just so GameData doesn't have to cast it a bunch.

src/main/java/net/minecraftforge/registries/RegisterEvent.java
SizableShrimp: I have no idea; just preserving behavior.

src/main/java/net/minecraftforge/common/ForgeMod.java
SizableShrimp: I think keeping them separate is fine.

src/main/java/net/minecraftforge/registries/ObjectHolderRegistry.java
SizableShrimp: This is just following the convention from line 144.

src/main/java/net/minecraftforge/registries/ForgeRegistry.java
SizableShrimp: Because you need the registry name for each value now. It would require two arrays, and I didn't decide whether to try and include that or not. The comment should probably be deleted at this point.

src/test/java/net/minecraftforge/debug/recipe/recipebook/RecipeBookExtensionTest.java
```diff
@@ -49,24 +47,18 @@
     public static final RegistryObject<MenuType<RecipeBookTestMenu>> RECIPE_BOOK_TEST_MENU_TYPE =
             MENU_TYPE.register("test_recipe_menu", () -> IForgeMenuType.create(RecipeBookTestMenu::new));
 
+    public static final DeferredRegister<RecipeType<?>> RECIPE_TYPE = DeferredRegister.create(ForgeRegistries.RECIPE_TYPES, MOD_ID);
```
XFactHD, line 51: This DR is missing the event bus registration

src/main/java/net/minecraftforge/registries/RegisterEvent.java
SizableShrimp: After changing this method to take a supplier and providing a RegisterHelper, I don't think this is important to specify anymore.

src/test/java/net/minecraftforge/debug/recipe/recipebook/RecipeBookExtensionTest.java
SizableShrimp: What similar methods?

TheCurle:
This has been implemented into the 1.19 pre-releases and will be included with the release. 
The PR will remain open so that people can see what code changes will be made, and prepare accordingly.

SizableShrimp:
This is now public in Forge 1.19, starting with 41.0.1. Use that.

PR #8543, opened 2022-03-26
[1.18.x] Simplfy default behavior of isSimple

KnightMiner:
isSimple should only return true if the ingredient is not NBT sensitive. Simply matching against a pickaxe or another damagable tool is not NBT sensitive so it should still be considered simple

Also prevents fetching tag values too early, as tags are not ready during the ingredient constructor.

### DIFF
--- a/patches/minecraft/net/minecraft/world/item/crafting/Ingredient.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/crafting/Ingredient.java.patch
@@ -1,6 +1,6 @@
 --- a/net/minecraft/world/item/crafting/Ingredient.java
 +++ b/net/minecraft/world/item/crafting/Ingredient.java
-@@ -29,17 +_,26 @@
+@@ -29,12 +_,19 @@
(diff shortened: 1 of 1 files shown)


ChampionAsh5357:
I agree with this change as `Ingredient#isSimple` was added to fix an issue with checking any data on the stack besides the item in an ingredient would cause an issue. Though, because of that, I think this should more accurately address `StackedContents` altogether in correlation to the recipe book since `#isSimple` checks to properly handle the ghost recipes is not present.

Shadows-of-Fire:
This looks good as well, not sure why the current `isSimple` implementation is the way that it is (probably just legacy and hasn't been changed), but the code in `Ingredient#test` boils down to `ItemStack#is` calls, and thus all vanilla ingredients are implicitly simple based on being item equality checks.

As an aside... I wonder if performance could be improved by using a `Set` for storing dissolved ingredient items and `Set#contains` in `Ingredient#test`

PR #8546, opened 2022-03-31
Add ICondition.IContext, fix TagEmptyCondition

Technici4n:
Allows recipe conditions to also receive a Context, currently it only contains the tags, for use by the `TagEmptyCondition`. The motivation is that this fixes the tag empty condition in 1.18.2.

The context might be expanded in the future to include more information.


pupnewfster:
Didn't really look at this at all, but one thing that I noticed is missing is adding the overload constructors and methods to the `forge.exc` file

patches/minecraft/net/minecraft/server/ReloadableServerResources.java.patch
```diff
@@ -1,5 +1,29 @@
...
+-   private final ServerAdvancementManager f_206853_ = new ServerAdvancementManager(this.f_206850_);
++   private final ServerAdvancementManager f_206853_;
+    private final ServerFunctionLibrary f_206854_;
+ 
+    public ReloadableServerResources(RegistryAccess.Frozen p_206857_, Commands.CommandSelection p_206858_, int p_206859_) {
+       this.f_206849_ = new TagManager(p_206857_);
+       this.f_206847_ = new Commands(p_206858_);
+       this.f_206854_ = new ServerFunctionLibrary(p_206859_, this.f_206847_.m_82094_());
++      // Forge: pass tagManager too
++      this.f_206848_ = new RecipeManager(this.f_206849_);
```
KnightMiner: Rather than passing in a tag manager, I would suggest passing in an IContext. It means we are constructing fewer instances, and if we want to add more to the context later we don't have to deprecate the old forge constructor. Have the default constructor set the field to an empty context.

src/main/java/net/minecraftforge/common/crafting/CraftingHelper.java
```diff
@@ -235,4 +253,24 @@ public static ICondition getCondition(JsonObject json)
             throw new JsonSyntaxException("Unknown condition type: " + condition.getID().toString());
         return serializer.getJson(condition);
     }
+
+    @Nullable
+    public static ICondition.IContext buildContext(@Nullable TagManager tagManager) {
+        if (tagManager == null) return null;
+
+        Map<ResourceKey<?>, Map<ResourceLocation, Tag<Holder<?>>>> tagsMap = new IdentityHashMap<>();
```
KnightMiner: Is making a copy of the tag map needed? Or could the context just return the tag manager itself? Have not looked into what is public on tag manager too much so copy might be the best move, just want to hear your reason.

If the copy is needed, I would suggest lazy loading this so it works with my above suggestion.

----

See also my below suggestion about making the parameter `TagKey<T>` and the return `Tag<Holder<T>>` as that might allow us to avoid copying.

src/main/java/net/minecraftforge/common/crafting/conditions/ICondition.java
```diff
@@ -5,11 +5,33 @@
...
+import net.minecraft.tags.Tag;
+
+import javax.annotation.Nullable;
+import java.util.Map;
 
 public interface ICondition
 {
     ResourceLocation getID();
 
+    default boolean test(@Nullable IContext context)
```
KnightMiner: What about if instead of making this nullable, you create `IContext.EMPTY` as a default to pass in, which always returns null for tags? Saves having to do a bunch of extra null checks in conditions.

src/main/java/net/minecraftforge/common/crafting/CraftingHelper.java
```diff
@@ -235,4 +253,24 @@ public static ICondition getCondition(JsonObject json)
...
+        for (var result : tagManager.getResult())
+        {
+            tagsMap.put(result.key(), Collections.unmodifiableMap((Map) result.tags()));
+        }
+
+        return new ICondition.IContext() {
+            @Override
+            @Nullable
+            public <T> Map<ResourceLocation, Tag<Holder<T>>> getTags(ResourceKey<? extends Registry<T>> key)
```
KnightMiner: For an alternative to consider, could we make this `Tag<Holder<T>> getTags(TagKey<T>)` instead? Its arguably slightly less efficient if you need multiple tags, but makes the interface a lot cleaner.

src/main/java/net/minecraftforge/common/crafting/conditions/ICondition.java
Technici4n: Good idea, implemented!

src/main/java/net/minecraftforge/common/crafting/CraftingHelper.java
Technici4n: Yes absolutely, implemented!

src/main/java/net/minecraftforge/common/crafting/CraftingHelper.java
Technici4n: > Is making a copy of the tag map needed?

It's not a deep copy, it only turns the List<Pair<registry key, tag map>> into a Map<registry key, tag map> to avoid having to iterate over all the elements.
See also:
image
image

The idea is to prevent iterating over all load results every time a tag is queried.

Technici4n: It's indeed lazy loaded now, but the list is still turned into a map.

patches/minecraft/net/minecraft/server/ReloadableServerResources.java.patch
Technici4n: Implemented.

Technici4n:
> Didn't really look at this at all, but one thing that I noticed is missing is adding the overload constructors and methods to the `forge.exc` file

Done, hopefully they are correct.

src/main/java/net/minecraftforge/common/crafting/conditions/ConditionContext.java
```diff
@@ -0,0 +1,43 @@
...
+			var tags = tagManager.getResult();
+			if (tags.isEmpty()) throw new IllegalStateException("Tags have not been loaded yet.");
+
+			loadedTags = new IdentityHashMap<>();
+			for (var loadResult : tags)
+			{
+				loadedTags.put(loadResult.key(), loadResult);
+			}
+		}
+		return (Tag) loadedTags.get(key.registry()).tags().get(key.location());
```
KnightMiner: This probably needs a null check on `loadedTags.get()` in case someone passes in a tag key for a registy that does not support tags. I know actually know what would happen if a tag supporting registry got no tags, whether that would be null or not.

Technici4n: Should now be handled properly with `getAllTags` returning an empty map by default.

PR #8547, opened 2022-03-31
implement a server starter jar

diesieben07:
Adds a `server_starter.jar` generated by the installer. Allows starting the server via `java -jar start_server.jar`.

~~So far only tested in a Linux console, marked as Draft until further testing (Windows, possibly server hosting consoles) has been done.~~

### DIFF
--- a/launcherscript/src/main/java/net/minecraftforge/launcherscript/LauncherScript.java
+++ b/launcherscript/src/main/java/net/minecraftforge/launcherscript/LauncherScript.java
@@ -1,0 +1,48 @@
+package net.minecraftforge.launcherscript;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.util.*;
+
+public class LauncherScript
+{
+
+    public static void main(String[] args) throws Exception
+    {
+
+        List<String> command = new ArrayList<>();
+        Collections.addAll(command, findJavaBinary(), "@user_jvm_args.txt");
+
+        command.add(String.format(Locale.ROOT, "@libraries/%s/%s_args.txt", loadMavenPath(), isWindows() ? "win" : "unix"));
+        Collections.addAll(command, args);
+        new ProcessBuilder(command)
+                .directory(null)
+                .inheritIO()
+                .start().waitFor();
+    }
+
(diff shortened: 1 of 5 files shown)


diesieben07:
I have now successfully tested it in Windows, too.
I have been unable to get Multicraft running, however I have no reason to believe it would struggle with this.

Seniorendi:
Neat!

> I have been unable to get Multicraft running

I have a running multicraft instance, I'll see if I can test it today

IzzelAliz:
Maybe you can pass jvm arguments to subprocess by `ManagementFactory.getRuntimeMXBean().getInputArguments()`.

diesieben07:
> Maybe you can pass jvm arguments to subprocess by ManagementFactory.getRuntimeMXBean().getInputArguments().

Good point, I will look into this. However the problem here is that you now have two JVMs running with the same arguments.

diesieben07:
Okay, no I won't be doing that. The server_starter only exists as a jumping off point, JVM arguments should go into `user_jvm_args.txt`. The server_starter should only be started with a minimum amount of RAM.

launcherscript/src/main/java/net/minecraftforge/launcherscript/LauncherScript.java
```diff
@@ -0,0 +1,57 @@
...
+
+        List<String> command = new ArrayList<>();
+        Collections.addAll(command, findJavaBinary(), "@user_jvm_args.txt");
+
+        command.add(String.format(Locale.ROOT, "@libraries/%s/%s_args.txt", loadMavenPath(), isWindows() ? "win" : "unix"));
+        Collections.addAll(command, args);
+        new ProcessBuilder(command)
+                .directory(null)
+                .inheritIO()
+                .start().waitFor();
```
Matyrobbrt, line 32: Is there a reason as to why the launcher process isn't shut down?

diesieben07: Because that would make whatever server software is running (usually the panel of a hoster or something like Multicraft) think that the server stopped.

IzzelAliz:
What about adding a shutdown hook that terminates the subprocess?

diesieben07:
Closing this, as this is the wrong approach and has way too many problems.

PR #8548, opened 2022-04-02
[1.18.x] Expose `getHolder()` method on RegistryObject

SizableShrimp:
This PR exposes a `getHolder()` method on RegistryObject which returns an `Optional<Holder<T>>` to provide better compatibility with vanilla code that requires the use of a Holder, especially worldgen (see `FeatureUtils#simpleRandomPatchConfiguration`, `BiomeGenerationSettings#addFeature`, `WeightedPlacedFeature#<init>`, `RandomBooleanFeatureConfiguration#<init>` for some examples).

### DIFF
--- a/src/main/java/net/minecraftforge/registries/RegistryObject.java
+++ b/src/main/java/net/minecraftforge/registries/RegistryObject.java
@@ -578,6 +578,19 @@ public final class RegistryObject<T> implements Supplier<T>
         }
     }
 
+    /**
+     * Returns an optional {@link Holder} instance pointing to this RegistryObject's name and value.
+     * <p>
+     * This should <b>only</b> be used in cases where vanilla code requires passing in a Holder.
+     * Mod-written code should rely on RegistryObjects or Suppliers instead.
+     * <p>
+     * The returned optional will be empty if the registry does not exist
+     * or if the deprecated factories {@link #of(ResourceLocation, Class, String)}
+     * or {@link #of(ResourceLocation, Supplier)} are used.
+     * Otherwise, the optional Holder will be present even if {@link #isPresent()} returns false.
+     *
+     * @return an optional {@link Holder} instance pointing to this RegistryObject's name and value
+     */
     @SuppressWarnings("unchecked")
     @NotNull
     public Optional<Holder<T>> getHolder()
--- a/src/main/java/net/minecraftforge/registries/RegistryObject.java
+++ b/src/main/java/net/minecraftforge/registries/RegistryObject.java
@@ -396,6 +396,18 @@ public final class RegistryObject<T> implements Supplier<T>
         return this.name;
     }
 
+    /**
+     * Returns the resource key that points to the registry and name of this registry object.
+     * Nullable only when the deprecated functions {@link #of(ResourceLocation, Class, String)} or {@link #of(ResourceLocation, Supplier)} are used.
+     *
+     * @return the resource key that points to the registry and name of this registry object
+     */
+    @Nullable
+    public ResourceKey<T> getKey()
+    {
+        return this.key;
+    }
+
     public Stream<T> stream() {
         return isPresent() ? Stream.of(get()) : Stream.of();
     }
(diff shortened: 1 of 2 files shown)


src/main/java/net/minecraftforge/registries/RegistryObject.java
```diff
@@ -297,32 +303,68 @@ public void accept(Predicate<ResourceLocation> pred)
...
     {
-        this.value = registry.containsKey(this.name) ? registry.get(this.name) : null;
+        if (this.name == null)
+            return;
+        if (registry.containsKey(this.name))
+        {
+            this.value = registry.get(this.name);
+            if (this.key == null)
+                this.key = (ResourceKey<T>) ResourceKey.create(registry.key(), this.name);
+            this.holder = ((Registry<T>) registry).getHolder(ResourceKey.create((ResourceKey<? extends Registry<T>>) registry.key(), this.name)).orElse(null);
```
pupnewfster: Is there a reason this isn't just:
```java
this.holder = ((Registry<T>) registry).getHolder(this.key).orElse(null)
```
as if the key was null, then it is initialized to the same value the line above, and otherwise we already have it?

SizableShrimp: No reason probably just missed this when I was modifying stuff

maxanier:
Thank you for this addition.
I assume `RegistryObject#getHolder` should always be non-empty when the registry exists, right?
Because when trying to use `#getHolder` to create dependent objects and register them via DeferredRegister, I do not see an alternative to using `Optional#get`.

Example code: 
```
public static final DeferredRegister<PlacedFeature> PLACED_FEATURES = DeferredRegister.create(Registry.PLACED_FEATURE_REGISTRY, REFERENCE.MODID);
public static final DeferredRegister<ConfiguredFeature<?,?>> CONFIGURED_FEATURES = DeferredRegister.create(Registry.CONFIGURED_FEATURE_REGISTRY, REFERENCE.MODID);

 public static final RegistryObject<ConfiguredFeature<?,?>> vampire_tree = CONFIGURED_FEATURES.register("vampire_tree", () -> new ConfiguredFeature<>(Feature.TREE, new TreeConfiguration.TreeConfigurationBuilder(BlockStateProvider.simple(Blocks.SPRUCE_LOG.defaultBlockState()), new StraightTrunkPlacer(4, 2, 0), BlockStateProvider.simple(ModBlocks.vampire_spruce_leaves.defaultBlockState()), new BlobFoliagePlacer(ConstantInt.of(2), ConstantInt.of(0), 3), new TwoLayersFeatureSize(1, 0, 1)).ignoreVines().build()));
public static final RegistryObject<PlacedFeature> vampire_tree_placed = PLACED_FEATURES.register("vampire_tree_placed", () -> new PlacedFeature(vampire_tree.getHolder().get(), List.of(PlacementUtils.filteredByBlockSurvival((ModBlocks.vampire_spruce_sapling)))));
```
When creating the `PlacedFeature` I need the holder instance. I cannot properly handle empty optionals, because within the supplier passed to `DeferredRegister#register` I cannot return null/empty.
I also cannot do `vampire_tree.getHolder().map(holder -> PLACED_FEATURES.register("",()->new PlacedFeature(holder,...)).orElse(???)` because I cannot get an empty RegistryObject to assign to my static field.

Maybe it would be possible to have some sort of `RegistryObject<I> DeferredRegister#registerOptional(id, Optional<T>, Function<T,I>`. That only calls the supplier function when the passed optional is present.

I'm not sure, maybe this doesn't make sense or I am missing something obvious. It's getting late.

SizableShrimp:
As long as the registry exists and you don't use deprecated RegistryObject factories (already done by DeferredRegister for you), the holder optional should be filled. For vanilla builtin registries, this means the moment you make a RegistryObject, it should return a non-empty holder optional. Thinking about it, it probably makes more sense to make getHolder() return a Nullable Holder? Thoughts? (cc. @pupnewfster)

pupnewfster:
Potentially, in my testing I just used `orElseThrow` on the optional as if it isn't present for some reason something is drastically wrong.

src/main/java/net/minecraftforge/registries/RegistryObject.java
```diff
@@ -523,6 +578,24 @@ public T orElseGet(Supplier<? extends T> other) {
         }
     }
 
+    @SuppressWarnings("unchecked")
+    @NotNull
+    public Optional<Holder<T>> getHolder()
```
LexManos, line 596: No javadocs to explain what this is, and why it should/shouldn't be used?

PR #8549, opened 2022-04-02
Fix return value of Recipe#isIncomplete being inaccurate for empty tags

pupnewfster:
Forge adds barriers to the output items for empty tags to fix issues with how vanilla does recipe matching. This adjusts vanilla's code that checks for if a recipe is incomplete and should be added to the recipe book to properly ignore this added item.

Eventually hopefully mojang will make `Recipe#isIncomplete` also be used as a precheck sort of thing before matching recipes, but until then as it would be a lot of spots for forge to patch and potentially might cause the method to be used in a way mojang doesn't intend, this is a simpler and less intrusive fix.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -58,6 +58,7 @@ import net.minecraft.world.entity.ai.attributes.AttributeSupplier;
 import net.minecraft.world.entity.ai.attributes.DefaultAttributes;
 import net.minecraft.world.entity.EquipmentSlot;
 import net.minecraft.world.item.*;
+import net.minecraft.world.item.crafting.Ingredient;
 import net.minecraft.world.item.crafting.RecipeType;
 import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.gameevent.GameEvent;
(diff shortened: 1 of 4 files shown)


src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1093,6 +1094,19 @@ public static int canEntitySpawn(Mob entity, LevelAccessor world, double x, doub
...
 
+    public static boolean hasNoElements(Ingredient ingredient)
+    {
+        ItemStack[] items = ingredient.getItems();
+        if (items.length == 0) return true;
+        if (items.length == 1 && !ForgeConfig.SERVER.treatEmptyTagsAsAir.get())
+        {
+            //If we potentially added a barrier due to the ingredient being an empty tag, try and check if it is the stack we added
+            ItemStack item = items[0];
+            return item.getItem() == Items.BARRIER && item.getHoverName() instanceof TextComponent hoverName && hoverName.getText().startsWith("Empty Tag: ");
```
KnightMiner, line 1105: Would it be worth just adding a custom NBT tag to the forge item for empty tags? Something like `forge_empty_tag: "forge:ingots/iron"`, and here just checking for that custom to exist. This feels super hacky as its comparing displayable text for gameplay logic. The tag would also be useful for things like recipe viewers to easily detect it without their own display name check hacks.

LexManos: The reason we used display text is because if for some reason someone ever got this item in a GUI. Such as NEI/RecipeBook it would tell them what tag it wants. Without having to lookup the NBT tags.

KnightMiner: Yeah, was not suggesting the display text be removed. Suggesting the extra tag in addition to the display text. Feels wrong to have game behavior based on display test (outside test mods that is)

LexManos: It feels wrong yes, but its no more wrong then this ugly ugly hack. Which I hope one day never has to happen because Mojang will of fixed their issues.

Issue #8498, opened 2022-03-05
`Ingredient.fromJson` in a custom recipe class results in an Empty Tag.


HaydenBelanger:
**Minecraft Version:** 1.18.2

**Forge Version:** 40.0.3

**Logs:** N/A

**Description of issue:**

Certain tags used within certain recipe Serializers result in an Empty Tag once in-world.

**Steps to Reproduce:**
 1. Use `Ingredient.fromJson` within any non-vanilla recipe Serializer to try to create an Ingredient from JSON.
 2. Enter game and use JEI/recipe book to view that the game shows an Empty Tag. Recipe is non-processable by expected functions.
 3. Arbitrarily, tag may fill itself at some point during operation until world is unloaded again, possibly related to /reload.

*JEI used to demonstrate*
![View with JEI](https://user-images.githubusercontent.com/7004666/156865539-ac841770-8582-4551-b82e-ee79c72e31fb.PNG)

[Link to image serializer](https://github.com/HaydenBelanger/assemblylinemachines/blob/530fdbeeea6c9d13fe57fdebb3ef65e6fb9bb4dc/src/main/java/me/haydenb/assemblylinemachines/crafting/BathCrafting.java#L213-L290)

[Link to image recipe JSON](https://github.com/HaydenBelanger/assemblylinemachines/blob/530fdbeeea6c9d13fe57fdebb3ef65e6fb9bb4dc/src/main/resources/data/assemblylinemachines/recipes/enchanted_book/looting.json)

SizableShrimp:
Can you provide the specific ingredient JSON for clarity?

HaydenBelanger:
Edited to include.

pupnewfster:
As of 1.18.2 I don't believe vanilla's ingredient's in recipes (at least for tag ingredients) return valid results from getItems until the reload is complete. So your serializer needs to delay the consumption of this until tags are resolved. (The reason it works after a reload is that you are making use of the previous reload's data, which might even be out of date so should not be used)

A similar issue I believe exists with forge's code and how it initializes Ingredient.isSimple in the constructor, even though getItems no longer will provide an accurate result for that for tags during deserialization.

Edit: My best guess as to why they do this is so that they don't need to keep two sets of tags around during reloading and then ensure they are referencing the correct one, rather than just referencing the "key" and then not accessing it until the reload is complete and the tags have been updated.

Eelt:
[...]

PR #8550, opened 2022-04-02
Fix issues with custom forge ingredients causing sub ingredients to be prematurely and invalidly cached

pupnewfster:
This PR fixes an issue I ran into when trying to use the `DifferenceIngredient` in a `ShapelessRecipe` as part of the deserialization logic for `ShapelessRecipe`s checks if the ingredient is empty. Vanilla's default empty implementation seems to be: `if no values and hasn't been initialized then it is empty`, so I went ahead and added empty checks in a semi similar way based on what my intuition tells me is right (this should definitely be validated to see if it makes sense though):
- Compound Ingredient: If all child ingredients are empty the compound is empty
- Intersection Ingredient: If any child ingredient is empty the intersection is empty. This doesn't necessarily return it being empty for non intersecting sets, but as vanilla in many places uses `getItems().length == 0` to check if the ingredient has no elements, I believe this is a reasonable implementation.
- Difference Ingredient: If the base ingredient is empty the difference is empty. I ignore the `subtracted` ingredient as if it is empty then the ingredient will just be equal to the `base` ingredient.

To prevent similar issues I also made a [commit](https://github.com/pupnewfster/MinecraftForge/commit/2f838577b928190e90018f518f49c040ea566681) on another branch that in theory would fix any issues like this where modders may accidentally initialize an ingredients' item cache during resource reloading (when tags are invalid) by invalidating all ingredients when tags update on the server (as client side there is no issue as the server doesn't tell the client the ingredient is a tag). I didn't include that commit in this PR for now as I am not quite sure if this extra safety handling is something that forge wants to do as it will add a tiny bit of overhead each time `getItems` or `test` is ran on ingredients to check if the ingredient's cache is valid. Additionally ensuring it is valid like that and updating it then would likely end up masking bugs in mods that then would never be discovered. Though if this is something that is decided is worthwhile to have I can add that commit to this PR.

### DIFF
--- a/patches/minecraft/net/minecraft/world/item/crafting/ShapelessRecipe.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/crafting/ShapelessRecipe.java.patch
@@ -56,3 +56,12 @@
           } else {
              ItemStack itemstack = ShapedRecipe.m_151274_(GsonHelper.m_13930_(p_44291_, "result"));
              return new ShapelessRecipe(p_44290_, s, itemstack, nonnulllist);
+@@ -88,7 +_,7 @@
+ 
+          for(int i = 0; i < p_44276_.size(); ++i) {
+             Ingredient ingredient = Ingredient.m_43917_(p_44276_.get(i));
+-            if (!ingredient.m_43947_()) {
++            if (net.minecraftforge.common.ForgeConfig.SERVER.skipEmptyShapelessCheck.get() || !ingredient.m_43947_()) {
+                nonnulllist.add(ingredient);
+             }
+          }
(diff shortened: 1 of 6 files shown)


Technici4n:
Note that due to 1.18.2 changes, querying an ingredient in a recipe deserializer will cause it to be empty forever. See also #8498.

pupnewfster:
That is what my second point of this in regards to the commit I made but didn't include yet in this PR (for the reasons I mentioned above) and making it then invalidate all ingredients before they actually get put into use outside of reload listeners.

Shadows-of-Fire:
As noted the proper emptiness check for Intersection ingredient requires resolution of the ingredient items, but that is not an option at this point, and I don't think there is a sensible method to flag "when" it will resolve.

However, there is a call to `Ingredient#isEmpty` within `StackedContents$RecipePicker#tryPick`.  While it is unlikely that someone constructs an empty intersection ingredient intentionally, it may act "weird" if it marked as non-empty yet is actually empty through disjoint sets.
Similar concern for subtracted ingredient, where a subtraction(a - a) could yield a "empty but not" scenario again.  This is (still) very unlikely to happen in practice.

We could simply mark those cases as undefined behavior, and let it be, unless we can figure out some kind of mechanism to allow them to perform their proper checks at runtime while deferring to the simpler checks at load time.

PR #8557, opened 2022-04-09
[1.18.x] Add use context and simulate flag to getToolModifiedState

SizableShrimp:
This is a different approach to the problem faced in #8503.
Hoes have been fixed to call `getToolModifiedState` when tilling with an appropriate patch.
A new overload of `getToolModifiedState` has been introduced which takes a `UseOnContext` and `simulate` boolean flag.
The old overload has been deprecated for removal in 1.19.
To preserve mods that do world-modifying actions from `BlockToolInteractEvent`, a new event called `BlockToolModificationEvent` has been made.
When `getToolModifiedState` is called with simulate as true, then `BlockToolModificationEvent` will be fired.
Otherwise, `BlockToolInteractEvent` will be fired.
`BlockToolInteractEvent` is a subclass of `BlockToolModificationEvent`.
This means that listeners of `BlockToolModificationEvent` will receive both simulate = true and simulate = false events, while older listeners of the now-deprecated `BlockToolInteractEvent` will only receive the old behavior of simulate = false.
Appropriate deprecation annotations and `// TODO`s have been added to `BlockToolInteractEvent` and `BlockToolModificationEvent` for 1.19 when the former can be removed.

### DIFF
--- a/patches/minecraft/net/minecraft/world/item/HoeItem.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/HoeItem.java.patch
@@ -9,7 +9,7 @@
        Level level = p_41341_.m_43725_();
        BlockPos blockpos = p_41341_.m_8083_();
 -      Pair<Predicate<UseOnContext>, Consumer<UseOnContext>> pair = f_41332_.get(level.m_8055_(blockpos).m_60734_());
-+      BlockState toolModifiedState = p_41341_.m_43725_().m_8055_(p_41341_.m_8083_()).getToolModifiedState(p_41341_, net.minecraftforge.common.ToolActions.HOE_TILL, false);
++      BlockState toolModifiedState = level.m_8055_(blockpos).getToolModifiedState(p_41341_, net.minecraftforge.common.ToolActions.HOE_TILL, false);
 +      Pair<Predicate<UseOnContext>, Consumer<UseOnContext>> pair = toolModifiedState == null ? null : Pair.of(ctx -> true, m_150858_(toolModifiedState));
        if (pair == null) {
           return InteractionResult.PASS;
--- a/patches/minecraft/net/minecraft/world/item/HoeItem.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/HoeItem.java.patch
@@ -1,6 +1,6 @@
 --- a/net/minecraft/world/item/HoeItem.java
 +++ b/net/minecraft/world/item/HoeItem.java
-@@ -30,6 +_,8 @@
+@@ -30,9 +_,11 @@
     }
  
     public InteractionResult m_6225_(UseOnContext p_41341_) {
@@ -8,7 +8,11 @@
 +      if (hook != 0) return hook > 0 ? InteractionResult.SUCCESS : InteractionResult.FAIL;
        Level level = p_41341_.m_43725_();
        BlockPos blockpos = p_41341_.m_8083_();
-       Pair<Predicate<UseOnContext>, Consumer<UseOnContext>> pair = f_41332_.get(level.m_8055_(blockpos).m_60734_());
+-      Pair<Predicate<UseOnContext>, Consumer<UseOnContext>> pair = f_41332_.get(level.m_8055_(blockpos).m_60734_());
++      Pair<Predicate<UseOnContext>, Consumer<UseOnContext>> pair = Pair.of(ctx -> ctx.m_43725_().m_8055_(ctx.m_8083_()).getToolModifiedState(ctx, net.minecraftforge.common.ToolActions.HOE_TILL, true) != null, ctx -> ctx.m_43725_().m_8055_(ctx.m_8083_()).getToolModifiedState(ctx, net.minecraftforge.common.ToolActions.HOE_TILL, false));
+       if (pair == null) {
+          return InteractionResult.PASS;
+       } else {
 @@ -72,5 +_,10 @@
  
     public static boolean m_150856_(UseOnContext p_150857_) {
--- a/patches/minecraft/net/minecraft/world/item/HoeItem.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/HoeItem.java.patch
@@ -1,6 +1,6 @@
 --- a/net/minecraft/world/item/HoeItem.java
 +++ b/net/minecraft/world/item/HoeItem.java
-@@ -30,9 +_,11 @@
+@@ -30,9 +_,12 @@
     }
  
     public InteractionResult m_6225_(UseOnContext p_41341_) {
@@ -9,7 +9,8 @@
        Level level = p_41341_.m_43725_();
        BlockPos blockpos = p_41341_.m_8083_();
 -      Pair<Predicate<UseOnContext>, Consumer<UseOnContext>> pair = f_41332_.get(level.m_8055_(blockpos).m_60734_());
-+      Pair<Predicate<UseOnContext>, Consumer<UseOnContext>> pair = Pair.of(ctx -> ctx.m_43725_().m_8055_(ctx.m_8083_()).getToolModifiedState(ctx, net.minecraftforge.common.ToolActions.HOE_TILL, true) != null, ctx -> ctx.m_43725_().m_8055_(ctx.m_8083_()).getToolModifiedState(ctx, net.minecraftforge.common.ToolActions.HOE_TILL, false));
++      BlockState toolModifiedState = p_41341_.m_43725_().m_8055_(p_41341_.m_8083_()).getToolModifiedState(p_41341_, net.minecraftforge.common.ToolActions.HOE_TILL, false);
++      Pair<Predicate<UseOnContext>, Consumer<UseOnContext>> pair = toolModifiedState == null ? null : Pair.of(ctx -> true, m_150858_(toolModifiedState));
        if (pair == null) {
           return InteractionResult.PASS;
        } else {
--- a/patches/minecraft/net/minecraft/world/item/HoeItem.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/HoeItem.java.patch
@@ -1,5 +1,18 @@
 --- a/net/minecraft/world/item/HoeItem.java
 +++ b/net/minecraft/world/item/HoeItem.java
+@@ -21,6 +_,12 @@
+ import net.minecraft.world.level.block.state.BlockState;
+ 
+ public class HoeItem extends DiggerItem {
++   /**
++    * @deprecated Forge: This map is patched out of vanilla code.
++    * Listen to {@link net.minecraftforge.event.world.BlockEvent.BlockToolModificationEvent}
++    * or override {@link net.minecraftforge.common.extensions.IForgeBlock#getToolModifiedState(BlockState, UseOnContext, net.minecraftforge.common.ToolAction, boolean)}.
++    */
++   @Deprecated
+    protected static final Map<Block, Pair<Predicate<UseOnContext>, Consumer<UseOnContext>>> f_41332_ = Maps.newHashMap(ImmutableMap.of(Blocks.f_50440_, Pair.of(HoeItem::m_150856_, m_150858_(Blocks.f_50093_.m_49966_())), Blocks.f_152481_, Pair.of(HoeItem::m_150856_, m_150858_(Blocks.f_50093_.m_49966_())), Blocks.f_50493_, Pair.of(HoeItem::m_150856_, m_150858_(Blocks.f_50093_.m_49966_())), Blocks.f_50546_, Pair.of(HoeItem::m_150856_, m_150858_(Blocks.f_50493_.m_49966_())), Blocks.f_152549_, Pair.of((p_150861_) -> {
+       return true;
+    }, m_150849_(Blocks.f_50493_.m_49966_(), Items.f_151017_))));
 @@ -30,9 +_,12 @@
     }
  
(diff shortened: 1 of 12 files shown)


KnightMiner:
The name `BlockToolInteractEventNew` is really gross, why not give it a name closer to the method on IForgeItemStack, like `BlockToolModificationEvent`? Means `BlockToolInteractionEvent` can just directly be removed in 1.19 instead of renaming the new event to the old name (causing a silent break for anyone who did not deal with the deprecation).

SizableShrimp:
Sounds reasonable enough to me

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -673,18 +715,20 @@ default boolean shouldDisplayFluidOverlay(BlockState state, BlockAndTintGetter l
      * @param stack The stack being used by the player
      * @param toolAction The action being performed by the tool
      * @return The resulting state after the action has been performed
+     * @deprecated Override and use {@link #getToolModifiedState(BlockState, UseOnContext, ToolAction, boolean)} instead
```
pupnewfster, line 716: This feels *slightly* weird to me just because of if any mods haven't updated yet to call the new `getToolModifiedState`, then any mods that have moved their implementation to it won't work with those other mods yet. I don't think there is really a good fix though other than mods that don't need the extra info just implementing the old method if they really want to stay compatible with any mods that may be calling the old one.

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -662,9 +663,50 @@ default boolean shouldDisplayFluidOverlay(BlockState state, BlockAndTintGetter l
...
+     * @param simulate If {@code true}, no actions that modify the world in any way should be performed. If {@code false}, the world may be modified.
+     * @return The resulting state after the action has been performed
+     */
+    @Nullable
+    default BlockState getToolModifiedState(BlockState state, UseOnContext context, ToolAction toolAction, boolean simulate)
+    {
+        BlockState toolModifiedState = getToolModifiedState(state, context.getLevel(), context.getClickedPos(),
+                context.getPlayer(), context.getItemInHand(), toolAction);
+
+        if (toolModifiedState == null && ToolActions.HOE_TILL == toolAction)
```
pupnewfster: Is it intentional that unlike the old method this one doesn't check if the stack can perform the hoe till action?

SizableShrimp: No

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
SizableShrimp: Technically a mod could override both for the most compatibility, just knowing that you don't get the full context from the deprecated method, and having to make some assumptions / adjust code accordingly.

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -662,9 +663,50 @@ default boolean shouldDisplayFluidOverlay(BlockState state, BlockAndTintGetter l
...
+    {
+        BlockState toolModifiedState = getToolModifiedState(state, context.getLevel(), context.getClickedPos(),
+                context.getPlayer(), context.getItemInHand(), toolAction);
+
+        if (toolModifiedState == null && ToolActions.HOE_TILL == toolAction && context.getItemInHand().canPerformAction(ToolActions.HOE_TILL))
+        {
+            // Logic copied from HoeItem#TILLABLES; needs to be kept in sync during updating
+            Block block = state.getBlock();
+            if (context.getLevel().getBlockState(context.getClickedPos().above()).isAir() &&
+                    (block == Blocks.GRASS_BLOCK || block == Blocks.DIRT_PATH || block == Blocks.DIRT || block == Blocks.COARSE_DIRT))
```
pupnewfster: I feel like checking block type before if the above state is air would be more efficient as that would remove having to lookup the above block in world if the block being tilled is rooted dirt.

src/test/java/net/minecraftforge/debug/block/ToolInteractTest.java
```diff
@@ -9,21 +9,21 @@
...
     {
         MinecraftForge.EVENT_BUS.addListener(this::onToolInteraction);
     }
-	
-    private void onToolInteraction(final BlockToolInteractEvent event)
+
+    private void onToolInteraction(final BlockToolModificationEvent event)
     {
         //Test 1: No Changes, just test if event is called. State and Final State should be the same
         LOGGER.info("BlockState {} is modified to {} at position {} by {} with {}", event.getState(), event.getFinalState(), event.getPos(), event.getPlayer(), event.getHeldItemStack());
```
pupnewfster: Might be worth printing if it is simulated or not just for test purposes?

src/main/java/net/minecraftforge/event/world/BlockEvent.java
```diff
@@ -415,65 +417,153 @@ public PortalShape getPortalSize()
...
+        }
+
+        /**
+         * Returns {@code true} if this event should not perform any actions that modify the world.
+         * If {@code false}, then world-modifying actions can be performed.
+         *
+         * @return {@code true} if this event should not perform any actions that modify the world.
+         * If {@code false}, then world-modifying actions can be performed.
+         */
+        public boolean isSimulate()
```
pupnewfster: Part of me feels like `isSimulated` might be a better method name, other than I understand the `is` part is for `isBoolean`

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
pupnewfster: Fair, I wonder if it is worth us pointing out this case... probably not.

SizableShrimp: We could, maybe in a FUN announcement. But I don't think it's something that should be covered by this PR specifically unless you think that would be a benefit. But it seems pretty niche

pupnewfster: Yeah, I don't think it is worth covering in this PR as trying to explain it would be somewhat convoluted and probably confuse more people.

src/test/java/net/minecraftforge/debug/block/ToolInteractTest.java
SizableShrimp: Sure. It'll never be fired as true in forge test mods currently, but maybe that can be expanded later.

patches/minecraft/net/minecraft/world/item/HoeItem.java.patch
```diff
@@ -1,14 +1,19 @@
...
     }
  
     public InteractionResult m_6225_(UseOnContext p_41341_) {
 +      int hook = net.minecraftforge.event.ForgeEventFactory.onHoeUse(p_41341_);
 +      if (hook != 0) return hook > 0 ? InteractionResult.SUCCESS : InteractionResult.FAIL;
        Level level = p_41341_.m_43725_();
        BlockPos blockpos = p_41341_.m_8083_();
-       Pair<Predicate<UseOnContext>, Consumer<UseOnContext>> pair = f_41332_.get(level.m_8055_(blockpos).m_60734_());
+-      Pair<Predicate<UseOnContext>, Consumer<UseOnContext>> pair = f_41332_.get(level.m_8055_(blockpos).m_60734_());
++      BlockState toolModifiedState = p_41341_.m_43725_().m_8055_(p_41341_.m_8083_()).getToolModifiedState(p_41341_, net.minecraftforge.common.ToolActions.HOE_TILL, false);
```
pupnewfster: Given vanilla already puts level and blockpos in local variables this would be much easier to read as `level.getBlockState(blockpos)...`

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -662,9 +663,49 @@ default boolean shouldDisplayFluidOverlay(BlockState state, BlockAndTintGetter l
...
+
+        if (toolModifiedState == null && ToolActions.HOE_TILL == toolAction && context.getItemInHand().canPerformAction(ToolActions.HOE_TILL))
+        {
+            // Logic copied from HoeItem#TILLABLES; needs to be kept in sync during updating
+            Block block = state.getBlock();
+            if (block == Blocks.ROOTED_DIRT)
+            {
+                if (!simulate)
+                {
+                    Block.popResourceFromFace(context.getLevel(), context.getClickedPos(), context.getClickedFace(), new ItemStack(Items.HANGING_ROOTS));
```
pupnewfster, line 691: I am not sure if it actually matters but for hoes this part is only ran server side.

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -662,9 +663,49 @@ default boolean shouldDisplayFluidOverlay(BlockState state, BlockAndTintGetter l
...
+            }
+        }
+
+        return toolModifiedState;
+    }
+
+    /**
+     * Returns the state that this block should transform into when right-clicked by a tool.
+     * For example: Used to determine if {@link ToolActions#AXE_STRIP an axe can strip},
+     * {@link ToolActions#SHOVEL_FLATTEN a shovel can path}, or {@link ToolActions#HOE_TILL a hoe can till}.
```
pupnewfster: I feel like this shouldn't mention HOE_TILL as an example given the fact it doesn't handle vanilla hoe tilling by default in this older now deprecated method.

src/main/java/net/minecraftforge/common/extensions/IForgeBlockState.java
```diff
@@ -603,18 +605,40 @@ default boolean shouldDisplayFluidOverlay(BlockAndTintGetter level, BlockPos pos
...
+    default BlockState getToolModifiedState(UseOnContext context, ToolAction toolAction, boolean simulate)
+    {
+        BlockState eventState = net.minecraftforge.event.ForgeEventFactory.onToolUse(self(), context, toolAction, simulate);
+        return eventState != self() ? eventState : self().getBlock().getToolModifiedState(self(), context, toolAction, simulate);
+    }
+
+    /**
+     * Returns the state that this block should transform into when right-clicked by a tool.
+     * For example: Used to determine if {@link ToolActions#AXE_STRIP an axe can strip},
+     * {@link ToolActions#SHOVEL_FLATTEN a shovel can path}, or {@link ToolActions#HOE_TILL a hoe can till}.
```
pupnewfster: I feel like this shouldn't mention HOE_TILL as an example given the fact it doesn't handle vanilla hoe tilling by default in this older now deprecated method.

src/main/java/net/minecraftforge/event/world/BlockEvent.java
```diff
@@ -415,65 +417,153 @@ public PortalShape getPortalSize()
...
+            this.context = null;
+            this.state = originalState;
+            this.toolAction = toolAction;
+            this.simulate = false;
+        }
+
+        /**
+         * @return the player using the tool
+         */
         public Player getPlayer()
```
pupnewfster, line 509: I feel like this should be marked as nullable and then just marked as nonnull in the override? Mostly given the fact that `Player` in `UseOnContext` is nullable

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
SizableShrimp: This is inaccurate as it passes through the deprecated method first before trying to do the default hoe till action.

src/main/java/net/minecraftforge/common/extensions/IForgeBlockState.java
SizableShrimp: This is inaccurate as it passes through the deprecated method first before trying to do the default hoe till action.

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
SizableShrimp: Looks like it doesn't matter based on the code in Block.popResourceFromFace, but it does some wasted calculations first, so I will fix this.

src/main/java/net/minecraftforge/common/extensions/IForgeBlockState.java
SizableShrimp: On further thought, the blockstate deprecated method won't be called for hoe till, but the block deprecated method always will be. Now, do we need to care? The only time this becomes an issue is if people try to coremod into this deprecated IForgeBlockState method and expect to receive HOE_TILL, which would require direct JS ASM since Mixins can't target interfaces. So I don't think we need to worry about this javadoc being technically slightly inaccurate for this unsupported usecase.

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
pupnewfster: Yes, except for the fact that if you call this with HOE_TILL on say dirt, you don't get farmland as a result you get null. You only get the proper result for the vanilla blocks when using the newer method. Hence why I think listing it as an one of the examples doesn't make sense.

SizableShrimp: Ok, that makes sense.

pupnewfster (approved):

PR #8561, opened 2022-04-10
Remove bad patch for AbstractFurnaceBlockEntity

sciwhiz12:
This bad patch was accidentally introduced in the 1.18 update and patch cycle, and prevented the block entity from being marked as changed when an item finished smelting. (Credits to @SizableShrimp for pointing this out.)

https://github.com/MinecraftForge/MinecraftForge/commit/40a174310d9eaf31e089f18b66a26d45c96d5587#diff-ff8896e0334590ab16251db3baa8ef3446a1ab83d4bdc880bc648efbc7287211
image


TheCurle:
Maximum oops!

PR #8562, opened 2022-04-10
[1.18] Let BaseRailBlocks decide if the calculated RailShape for placement or connection to neighbors is valid

XFactHD:
This PR adds a method in `IForgeBaseRailBlock` that allows modded rail blocks to decide whether the new `RailShape` calculated for block placement in `RailState#place()` or for neighbor connection in `RailState#connectTo()` is valid for this block.
This allows mods to add rail blocks where the `BlockState` property that is returned from `BaseRailBlock#getShapeProperty()` has limitations that are not covered by the existing checks (`IForgeBaseRailBlock#isFlexibleRail()` for allowing corners and `IForgeBaseRailBlock#canMakeSlopes()` for allowing slopes).
This has no influence on the behaviour of existing (vanilla) rails that do not use this new method.

My usecase is a rail block that can only have a ramp shape (essentially an underfilled rail ramp), which currently causes an exception and server-side log spam when another rail block is placed at the high end of the ramp on the y level of the lower end and tries to modify the shape of the ramp to be flat.

### DIFF
--- a/patches/minecraft/net/minecraft/world/level/block/RailState.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/block/RailState.java.patch
@@ -37,6 +37,17 @@
           if (BaseRailBlock.m_49364_(this.f_55414_, blockpos3.m_7494_())) {
              railshape = RailShape.ASCENDING_EAST;
           }
+@@ -203,6 +_,10 @@
+          railshape = RailShape.NORTH_SOUTH;
+       }
+ 
++      if (!this.f_55416_.isValidRailShape(railshape)) { // Forge: allow rail block to decide if the new shape is valid
++         this.f_55419_.remove(p_55442_.f_55415_);
++         return;
++      }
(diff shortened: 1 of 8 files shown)


TheCurle:
Rail hooks (:aniblobsweat:)

This seems like a great addition, but we should be careful for ancient API that intended to do this in the past.
We used to have an extensive rail API, but it slowly ran down and got purged part-by-part.

Some of it remains - I've been working on it over time, since some vanilla changes have made updates rather difficult with those hooks. 

We should make sure that none of those hooks do the same as this, and remove the old ones if they do. (they're likely not working now anyway)

TheCurle:
We've confirmed through searching the repository that there aren't any legacy patches that affect this, so i'm going to call it good to go.

Note to self and to triage team: We need to think through a decent set of maintainable, extensible rail patches. The last system was hell, and it's slowly becoming more highly requested.

PR #8564, opened 2022-04-11
1.18.x Port BiomeDictionary over to biome tags

Flemmli97:
This PR ports the current `BiomeDictionary` over to the new tag system.

The tags are under the minecraft namespace which if unfit i can change and use the vanilla style name so e.g. if the `BiomeDictionary#Type` was HOT the tags path would be "is_hot".

Note: This does not replace the current `BiomeDictionary` fully since tags aren't loaded at `BiomeLoadingEvent`


ChampionAsh5357:
The tags should be under the `forge` namespace as Forge provides them, not vanilla. Additionally, I don't believe prefixing with `is_` is necessary since it's specifying what the tag is already for.

Is this only a tag port, or does the logic provided by the `BiomeDictionary` also interface with the newly created tags?

eerussianguy:
> The tags should be under the `forge` namespace as Forge provides them, not vanilla. Additionally, I don't believe prefixing with `is_` is necessary since it's specifying what the tag is already for.
> 
> Is this only a tag port, or does the logic provided by the `BiomeDictionary` also interface with the newly created tags?

It's worth noting the `is_` prefix is vanilla's style. Of course we are not obliged to keep vanilla's style but stating it just in case.

Flemmli97:
changed to forge namespace. also yes `is_` is cause vanilla.
Unless im missing something there is no real logic used by forge itself for the `BiomeDictionary` except for checking if a biome has a `BiomeDictionary.Tag` which tags already provide that

Minecraftschurli:
> Unless im missing something there is no real logic used by forge itself for the `BiomeDictionary` except for checking if a biome has a `BiomeDictionary.Tag` which tags already provide that

It does add logic with `getBiomes`, `getTypes`, `areSimilar`, `hasType` and `hasAnyType` which you would need to change the implementation for. Also I think at the moment `BiomeDictionaries` are sometimes used before tags are loaded which would cause problems.

src/main/java/net/minecraftforge/common/data/ForgeBiomeTagsProvider.java
```diff
@@ -0,0 +1,117 @@
...
+        tag(Biomes.BAMBOO_JUNGLE, List.of(Tags.Biomes.IS_HOT, Tags.Biomes.IS_WET, Tags.Biomes.IS_RARE, Tags.Biomes.IS_OVERWORLD));
+        tag(Biomes.LUSH_CAVES, List.of(Tags.Biomes.IS_UNDERGROUND, Tags.Biomes.IS_LUSH, Tags.Biomes.IS_WET, Tags.Biomes.IS_OVERWORLD));
+        tag(Biomes.DRIPSTONE_CAVES, List.of(Tags.Biomes.IS_UNDERGROUND, Tags.Biomes.IS_SPARSE, Tags.Biomes.IS_OVERWORLD));
+        tag(Biomes.SOUL_SAND_VALLEY, List.of(Tags.Biomes.IS_HOT, Tags.Biomes.IS_DRY));
+        tag(Biomes.CRIMSON_FOREST, List.of(Tags.Biomes.IS_HOT, Tags.Biomes.IS_DRY));
+        tag(Biomes.WARPED_FOREST, List.of(Tags.Biomes.IS_HOT, Tags.Biomes.IS_DRY));
+        tag(Biomes.BASALT_DELTAS, List.of(Tags.Biomes.IS_HOT, Tags.Biomes.IS_DRY));
+    }
+
+    private void tag(ResourceKey<Biome> biome, List<TagKey<Biome>> tags)
```
pupnewfster: I feel like calls to this might be more readable if this just takes the tags via vararg like so:
```java
private void tag(ResourceKey<Biome> biome, TagKey<Biome>... tags)
```
(And if needed I forget if it will warn about possible pollution at which point you could just annotate the method as `@SafeVarargs`)

src/generated/resources/data/forge/tags/worldgen/biome/is_plains.json
```diff
@@ -0,0 +1,10 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:plains",
+    "minecraft:savanna",
+    "minecraft:savanna_plateau",
```
TelepathicGrunt: I would say remove this from plains tag as plains a low lying flat land. Seems odd to have a plateau in a plains tag

src/generated/resources/data/forge/tags/worldgen/biome/is_plateau.json
```diff
@@ -0,0 +1,8 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:savanna_plateau",
+    "minecraft:wooded_badlands",
+    "minecraft:meadow"
```
TelepathicGrunt, line 6: As for meadow here, should this be removed from plateau since it is in the plains tag already? it's tougher as meadows are high flat land which signals a plateau but visually, they are very similar to a plains.

src/generated/resources/data/forge/tags/worldgen/biome/is_slope.json
```diff
@@ -0,0 +1,10 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:savanna_plateau",
+    "minecraft:wooded_badlands",
```
TelepathicGrunt, line 5: do plateaus and wooded badlands belong in slopes? I see slopes as the sides of the giant mountainous areas that end with peak biomes in center. Savannas plateau and badlands are more their own things separate from that mountainous area. Here's a pic of what I mean. You can see how these mountains have a pretty clear area of what is a slope for them around the peaks and lower down. Basically anywhere that biomes can spawn alongside Snowy Slopes biome
image

KnightMiner:
Is `hasAnyType` even needed for tags? My understanding of it was to check "are they actually using the biome dictionary for this biome", which was a good way to identify biomes added via datapacks. Datapacks can add biome tags though, so there should be no reason a biome has no tags

src/generated/resources/data/forge/tags/worldgen/biome/is_slope.json
FlashFyreMods: Savanna plateaus and wooded badlands were added to slopes with the forge biomedict tags because they do have the possibility to generate as slopes. If you can find an area like that screenshot in warmer biomes you should run into plateaus in place of the cold slope biomes. The OverworldBiomeBuilder class shows this.

TelepathicGrunt: Leaving my final 2 cents over the slopes tag here. The slopes tags is for biomes that are dedicated to spawning around the main body of giant peaking mountains like snowy slopes, and groves do. these biomes spawn nowhere else except for those areas. Savanna Plateaus spawn frequently throughout the savanna area and often nowhere near any giant peaking mountain. Thus they are plateaus and not slopes. Wooded Badlands do seem to be spawning as mountains themselves or higher plateaus. I'll concede that wooded badlands can stay in the slopes tag but they are a trickier biome since they do the peak, slope, and plateau areas all together.

image
Untitled

Still, one has to think of the usability of the tag as well. Worldgen structures json can only be given one structure to spawn in and other usages of these tags may also be highly tailored to the slopes of the main giant mountains everyone thinks about with regards to groves and snowy slopes and meadows. Having savanna and/or badlands biomes would cause mods using the slopes to have their stuff be out of place and not designed to fit in a badland's area. It's up to the PR maintainer to decide what to do but if anything, adding a cold_slopes, cold_peaks, warm_slopes, warm_peaks may be beneficial in the long run to help people target the right temperature mountain. Or maybe simplified to just cold_mountain and warm_mountain and contains the slopes and peaks for them. 

Although, savanna plateau should probably stay only in the plateau and savanna tag. Since forests and jungle biomes are crowded around stony peaks a ton, would that means forest and jungles have to be added to slopes to match savanna plateaus if they are kept in slopes? Also, meadows may be better to remove from slopes and kept only in the plains and plateau tags.

src/generated/resources/data/forge/tags/worldgen/biome/is_plains.json
TelepathicGrunt: also, maybe savanna should also be removed alongside savanna plateau. We already have a savanna tag for targeting savannas. This way people can target plains-like biomes better

src/generated/resources/data/forge/tags/worldgen/biome/is_dry.json
```diff
@@ -0,0 +1,16 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:desert",
+    "minecraft:nether_wastes",
+    "minecraft:the_end",
```
TelepathicGrunt, line 6: is there a reason only one end biome is dry but not any of the others? maybe we should add all end biomes to this tag.

src/generated/resources/data/forge/tags/worldgen/biome/is_cold.json
```diff
@@ -0,0 +1,21 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:taiga",
+    "minecraft:the_end",
```
TelepathicGrunt, line 5: same deal here as the hot tag. Should all the other end biomes be added to this tag or split off into a cold_otherwordly biome tag? Make it easier for mods and datapacks using the mojang json structures system to add to cold overworld biomes.

src/generated/resources/data/forge/tags/worldgen/biome/is_hot.json
```diff
@@ -0,0 +1,20 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:desert",
+    "minecraft:nether_wastes",
```
TelepathicGrunt, line 5: A friend and I were talking and they pointed out that it is odd that nether biomes are in a hot tag. Specifically as this makes it harder for mods with json structures to use as they only take a single biome tag. So if they want a structure to spawn in hot biomes but only in overworld, they cannot use this tag. Ideally, tags could do an intersection system and only import entries that exists in all tags in a given list (like overworld and hot tag together). 

Since union doesn't exist, maybe the nether biomes should be split off of here to a hot_otherwordly biome tag?

src/generated/resources/data/forge/tags/worldgen/biome/is_slope.json
FlashFyreMods: Firstly, I should state that meadows and savanna plateaus have identical elevation spawning, so if you want savanna plateaus removed from the slope tag you should also remove the meadows, leaving the slope tag dedicated for the two slope only biomes, the snowy slopes and the grove.

However, I think both should stay in the slope tag as well as the plateau tag as they can spawn in both areas, to put it simply. If they shouldn't be in the slope tag because they don't spawn on slopes 100% of the time then you should remove them from the plateau tag as well surely, as they don't spawn as plateaus 100% of the time either. Even pure slope biomes, the grove and the snowy slope, don't always generate next to mountains (if the erosion noise didn't peak high enough). I definitely think it's better to put them in tags for areas that they _can_ spawn rather than put them in tags for areas that are the only place they can spawn.

Some tag changes of some sort are probably needed in the future tho, since the functionality possible with biome dictionary tags doesn't seem to be easily translatable into the vanilla system.

src/generated/resources/data/forge/tags/worldgen/biome/is_hot.json
KnightMiner: Could just do nested tags like forge does with ingots. `forge:hot/overworld`, `forge:hot/nether`, etc.

Flemmli97:
> > Unless im missing something there is no real logic used by forge itself for the `BiomeDictionary` except for checking if a biome has a `BiomeDictionary.Tag` which tags already provide that
> 
> It does add logic with `getBiomes`, `getTypes`, `areSimilar`, `hasType` and `hasAnyType` which you would need to change the implementation for. Also I think at the moment `BiomeDictionaries` are sometimes used before tags are loaded which would cause problems.

The tagsystem do provide the logic already for all the current `BiomeDictionary` stuff. One thing though that do indeed need some further looking into is the `BiomeLoadingEvent` cause tags are not populated yet at the time its called

Flemmli97:
also regarding [TelepathicGrunt](https://github.com/TelepathicGrunt) i simply copied over the definitions from the current `BiomeDictionary` so if something doesn't seem to fit (anymore) i can change it

FlashFyreMods:
> also regarding [TelepathicGrunt](https://github.com/TelepathicGrunt) i simply copied over the definitions from the current `BiomeDictionary` so if something doesn't seem to fit (anymore) i can change it

Yeah I'd definitely change obvious things like the end and nether biome temperature issues, Anything else should be fine as is, as the tags were updated for 1.18 biomes a little while ago.

TelepathicGrunt:
Don’t forget savannas in plains tag. Savanna’s don’t fit as they are a separate wood type and very visually different from the existing entries in plains tag. Even Mojang considers them a completely separate thing by making a savanna category instead of reusing the plains category for savannas and even have separate plains and savanna villages for them. If a modder wants both plains and savannas, they would make a tag that includes both of those tags together

FlashFyreMods:
> Don’t forget savannas in plains tag. Savanna’s don’t fit as they are a separate wood type and very visually different from the existing entries in plains tag. Even Mojang considers them a completely separate thing by making a savanna category instead of reusing the plains category for savannas and even have separate plains and savanna villages for them. If a modder wants both plains and savannas, they would make a tag that includes both of those tags together

Yep. I'm guessing the plains and savanna were grouped because they are both "plains-like" biomes, or at least that's the only logic I can think of. Since the forge biome dictionary always seemed to be more about "kinds" of biomes because of tags being associated with biomes rather than the other way round, I think?

Flemmli97:
i have split some tags into overworld/nether/end variants e.g. for `is_hot` we have now `is_hot/overworld`,  `is_hot/nether`,  `is_hot/end` tags. and removed some tags for some biomes like e.g. the `is_plains` tag from savannas.

also asking: should the `is_` prefix be kept or removed? cause keeping it is more inline with how vanilla does it

Tslat:
I think one of the biggest disconnects we have here (and in numerous other modding-related discussions), is forgetting that datapacks and mods are aimed at _users_, not modders. Tags especially are intended for datapack devs, not modders.

To a user/datapack dev, it makes the most sense to have a system of consistency, in which this case it would be to keep the `is_` prefix, since vanilla requires that for their tags.

Differentiating in the forge tags because it makes less semantic sense is stupid, because that's prioritising modder grammatical comfort over the user experience and ease of use of the people who are actually going to use the system.

TheCurle:
I'm firmly of the opinion that Forge should follow the lead of vanilla.
It's what we've done so far, it's what we're going to continue doing for long to come, and it's what we're going to do here.

There's no valid reason for us to differentiate from the naming system that vanilla uses and create inconsistency for our own sake.

sciwhiz12:
Regarding the `is_` prefix, it makes sense to me that we should keep the prefix on these tags, to copy the convention laid down by vanilla MC itself (as @eerussianguy noted). No reason to complicate matters by adding our own conflicting conventions.

Flemmli97:
fixed dark forest tag missing from dense/overworld, else dont think anything other is missing/wrong but maybe someone else might go over the tags too in case i overlooked something

GirafiStudios:
I my opinion it would make the most sense to wait with this until 1.19. (Not the biggest fan of it in general, but I can understand the desire to go in this direction, as vanilla is starting to do the same)
While this is technically not a breaking change, as the BiomeDictionary is not being removed, it would end up causing incompatibility between mods.
For example if Mod A is only spawning mobs within biomes with certain BiomeDictionary tags, but Mod B is only adding their biomes to the json tags, Mod A´s mobs would not end up spawning in Mob B´s biomes.
It´s difficult enough to get people to support one system. Having people have support for two different systems for the same thing, seems like an impossible task to me.

GirafiStudios:
Looking at the comments in this PR - is the aim more to provide the tags for datapack/modpack creators?
If so the title/description is quite misleading.

TelepathicGrunt:
@GirafiStudios And modders too. We use biome tags as well. Especially for structure generation

GirafiStudios:
> @GirafiStudios And modders too. We use biome tags as well. Especially for structure generation

In ways that you can´t with the BiomeDictionary, or?

TelepathicGrunt:
@GirafiStudios Structures are now json. Mojang is pushing everyone to doing stuff in json and structure are redone to add themselves to biome given to them by biome tags in json.
https://github.com/TelepathicGrunt/StructureTutorialMod/blob/c5dd7932909222fe92cb9027f8261cdbde3620bd/src/main/resources/data/structure_tutorial/worldgen/configured_structure_feature/json_only_house.json#L15

Furthermore, biome tags are better for user configurability imo as they are able to be overridden right off the bat whereas biome dictionary requires the modder to make a config to expose that configurability directly. So mods that are able to switch to biome tags would be less likely to be hardcoded in ways that cannot be changed by config or datapack as there's one less config entry to forget to add.

Imo, better to push for people to switch to biome tags now so the transition to 1.19 is smoother for modders and less conflicts or misunderstanding about what goes where. The people who add features or mobs to biomes cannot use biome tags as the current BiomeLoadingEvent cannot use biome tags (although a replacement for that event is in the works that would allow people to use biome tags) and the json for features/mob spawns is within the biome json itself instead of using biome tags like structures do.

Tslat:
With that being the case, are we looking for a way to allow usage of these in BiomeLoadingEvent?
It's all well and good switching to tags, but if they're entirely unusable in the place where people actually need them in modding, it's not so great

GirafiStudios:
> @GirafiStudios Structures are now json. Mojang is pushing everyone to doing stuff in json and structure are redone to add themselves to biome given to them by biome tags in json. https://github.com/TelepathicGrunt/StructureTutorialMod/blob/c5dd7932909222fe92cb9027f8261cdbde3620bd/src/main/resources/data/structure_tutorial/worldgen/configured_structure_feature/json_only_house.json#L15
> 
> Furthermore, biome tags are better for user configurability imo as they are able to be overridden right off the bat whereas biome dictionary requires the modder to make a config to expose that configurability directly. So mods that are able to switch to biome tags would be less likely to be hardcoded in ways that cannot be changed by config or datapack as there's one less config entry to forget to add.
> 
> Imo, better to push for people to switch to biome tags now so the transition to 1.19 is smoother for modders and less conflicts or misunderstanding about what goes where. The people who add features or mobs to biomes cannot use biome tags as the current BiomeLoadingEvent cannot use biome tags (although a replacement for that event is in the works that would allow people to use biome tags) and the json for features/mob spawns is within the biome json itself instead of using biome tags like structures do.

Ahh okay. Haven´t looked at structures yet for 1.18.2, didn´t realize that was the case. Definitely needed for that then!

As Tslat also just mentioned, I´m not a big fan of the limitations it currently would bring. Not being able to use tags in BiomeLoadingEvent would break a lot of mods, including 2-3 of my own.

As the BiomeLoadingEvent have not been fixed to allow for json tags as of yet & removing the BiomeDictionary would a huge breaking change, I don´t see how this could go through for 1.18.2 - with Forges own policy of not having breaking changes after a recommended build have been released.

Having two systems that does almost the same thing for 1.18 would also be a huge mess, as if some mod rely entirely of either system, it would break compatibility between mods. For example if Mod A adds mob spawns to biomes with a certain BiomeDictionary tag, but Mod B is only adding its biomes to the json tags, Mod A´s mobs would not spawns Mod B´s biomes.
It´s also difficult enough to get people to use just one tag system. Having people add them to both would be borderline impossible.

TelepathicGrunt:
@Tslat " they're entirely unusable in the place where people actually need them in modding". Only one place. BiomeLoadingEvent. And that's mainly just features and mob spawning. Everywhere else can use biome tags just fine really. The event cannot ever be tag sensitive because it fires in the construction of the biome itself while the dynamic registry is still being made before any tags is made. This is the replacement for BiomeLoadingEvent which I was told a while ago when BiomeLoadingEvent was first made, that BiomeLoadingEvent was supposed to be temporary anyway until a better solution comes along: https://github.com/MinecraftForge/MinecraftForge/pull/8341

@GirafiStudios There's already two systems in place. Mojang introduced biome tags and biomes mods are using it. See Biomes o Plenty and Oh The Biomes You'll Go as they already use the mojang biome tags. So blocking this PR doesn't save anyone from the incompatibility issue. It just makes mods still be difficult to be compat with datapacks out of the box if the mods keep relying on biome dictionary outside BiomeLoadingEvent and refusing to add their biomes to biome tags. I will continue to push biomes mods to use biome tags so my structures can spawn in their biomes by default even without this pr. 

What this PR accomplishes is it gives more options of biome tags that people know are standardized and will be more willing to use for their biome mods. Which in turn will make mods and datapacks start to use those tags for detecting if players are in desert tagged biomes or other stuff that are fitting to be tag based. The biome mods just has to add their biomes to biome dictionary only for the BiomeLoadingEvent mods but all other use cases can switch to tags just fine. And the big biome mods already use biome dictionary and biome tags together so the belief of getting biome mods to adapt tags is impossible was already proven false.

There's going to be transition period whether we like it or not. The question is, do we set up the tags now so people can make the switch quickly and easily or just wait while people start making their own forge biome tags and create a bigger mess of incompat that has to be undone. Mojang already started the process. Might as well complete it.

src/main/java/net/minecraftforge/common/data/ForgeBiomeTagsProvider.java
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
```
TheCurle, line 4: Please rerun the license formatting task - we use a different format now.

TheCurle:
Given the concerns stated above, it would be a good idea to throw @Deprecated onto the BiomeDictionary with an explanation as to why it shouldn't be used by modders in the future.

It would also be a good idea to mention the existence of these biome tags in documentation - be it in the BiomeDictionary javadoc, the Documentation repo or the FCW.

Tslat:
Telepathic, my issue is exactly what you've just done.

People are dismissing the fact that BiomeLoadingEvent isn't compatible with this as a 'well it's only this _one_ thing', which is bs because Biome modifications are a huge chunk of modded features & decorations. It's a critical component of modding, and it not working with tags is a massive issue.
I'm fully aware that it's not possible to support it at current, but I think there needs to be a bigger emphasis put on installing its replacement ASAP as without it this tag system has a huge hole

I'm fully aware that's not within the scope of this particular PR, but I do think it needs to have a higher priority with this becoming a thing

src/generated/resources/data/forge/tags/worldgen/biome/is_cold.json
```diff
@@ -0,0 +1,25 @@
...
```
sciwhiz12, line 5: `forge:is_cold/end` and `forge:is_cold/overworld` were added, but it did not replace the same entries already here.

src/generated/resources/data/forge/tags/worldgen/biome/is_dense.json
```diff
@@ -0,0 +1,9 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:jungle",
+    "minecraft:dark_forest",
+    "minecraft:old_growth_birch_forest",
+    "minecraft:old_growth_spruce_taiga"
```
sciwhiz12, line 7: `forge:is_dense/overworld` was added, but it did not replace the same entries already here.

src/generated/resources/data/forge/tags/worldgen/biome/is_dry.json
```diff
@@ -0,0 +1,20 @@
...
```
sciwhiz12, line 6: `forge:is_dry/overworld`, `forge:is_dry/nether`, and `forge:is_dry/end` were added, but it did not replace the same entries already here.

src/generated/resources/data/forge/tags/worldgen/biome/is_hot.json
sciwhiz12, line 5: `forge:is_hot/nether` and `forge:is_hot/overworld` were added, but it did not replace the same entries already here.

src/generated/resources/data/forge/tags/worldgen/biome/is_sparse.json
```diff
@@ -0,0 +1,16 @@
...
+    "minecraft:savanna",
+    "minecraft:savanna_plateau",
+    "minecraft:wooded_badlands",
+    "minecraft:snowy_slopes",
+    "minecraft:jagged_peaks",
+    "minecraft:frozen_peaks",
+    "minecraft:windswept_gravelly_hills",
+    "minecraft:windswept_savanna",
+    "minecraft:eroded_badlands",
+    "minecraft:dripstone_caves"
```
sciwhiz12, line 14: `forge:is_sparse/overworld` wasadded, but it did not replace the same entries already here.

src/generated/resources/data/forge/tags/worldgen/biome/is_wet.json
```diff
@@ -0,0 +1,11 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:swamp",
```
sciwhiz12, line 4: `forge:is_wet/overworld` wasadded, but it did not replace the same entries already here.

src/main/java/net/minecraftforge/common/BiomeDictionary.java
```diff
@@ -25,6 +26,11 @@
 
 import net.minecraft.world.level.biome.Biome;
 
+/**
+ * Use {@link Tags.Biomes} when possible.
+ * Note that tags are not usable yet during {@link BiomeLoadingEvent} so BiomeDictionary are still necessary there
+ */
+@Deprecated
```
sciwhiz12, line 33: `@Deprecated` needs a `since` tag set to `1.18.2`, and the javadoc info relating to its deprecation should be added to the `@deprecated` block javadoc tag.

src/generated/resources/data/forge/tags/worldgen/biome/is_beach.json
```diff
@@ -0,0 +1,8 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:beach",
+    "minecraft:stony_shore",
+    "minecraft:snowy_beach"
```
sciwhiz12, line 6: Vanilla Minecraft has the `is_beach` tag, which contains the same entries except the `stony_shore`.

src/generated/resources/data/forge/tags/worldgen/biome/is_end.json
```diff
@@ -0,0 +1,10 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:the_end",
+    "minecraft:small_end_islands",
+    "minecraft:end_midlands",
+    "minecraft:end_highlands",
+    "minecraft:end_barrens"
```
sciwhiz12, line 8: Vanilla Minecraft has the `is_end` tag, which contains the same entries.

src/main/java/net/minecraftforge/common/data/ForgeBiomeTagsProvider.java
```diff
@@ -0,0 +1,107 @@
...
+        tag(Biomes.BAMBOO_JUNGLE, Tags.Biomes.IS_HOT, Tags.Biomes.IS_HOT_OVERWORLD, Tags.Biomes.IS_WET, Tags.Biomes.IS_WET_OVERWORLD, Tags.Biomes.IS_RARE, Tags.Biomes.IS_OVERWORLD);
+        tag(Biomes.LUSH_CAVES, Tags.Biomes.IS_UNDERGROUND, Tags.Biomes.IS_LUSH, Tags.Biomes.IS_WET, Tags.Biomes.IS_WET_OVERWORLD, Tags.Biomes.IS_OVERWORLD);
+        tag(Biomes.DRIPSTONE_CAVES, Tags.Biomes.IS_UNDERGROUND, Tags.Biomes.IS_SPARSE, Tags.Biomes.IS_SPARSE_OVERWORLD, Tags.Biomes.IS_OVERWORLD);
+        tag(Biomes.SOUL_SAND_VALLEY, Tags.Biomes.IS_HOT, Tags.Biomes.IS_HOT_NETHER, Tags.Biomes.IS_DRY, Tags.Biomes.IS_DRY_NETHER);
+        tag(Biomes.CRIMSON_FOREST, Tags.Biomes.IS_HOT, Tags.Biomes.IS_HOT_NETHER, Tags.Biomes.IS_DRY, Tags.Biomes.IS_DRY_NETHER);
+        tag(Biomes.WARPED_FOREST, Tags.Biomes.IS_HOT, Tags.Biomes.IS_HOT_NETHER, Tags.Biomes.IS_DRY, Tags.Biomes.IS_DRY_NETHER);
+        tag(Biomes.BASALT_DELTAS, Tags.Biomes.IS_HOT, Tags.Biomes.IS_HOT_NETHER, Tags.Biomes.IS_DRY, Tags.Biomes.IS_DRY_NETHER);
+
+        tag(Tags.Biomes.IS_WATER).addTag(BiomeTags.IS_OCEAN).addTag(BiomeTags.IS_RIVER);
+        tag(BiomeTags.IS_MOUNTAIN).addTag(Tags.Biomes.IS_PEAK).addTag(Tags.Biomes.IS_SLOPE);
```
sciwhiz12, line 91: This should have a comment noting that this is modifying a Minecraft tag, to avoid that detail being glossed over.

src/main/java/net/minecraftforge/common/Tags.java
```diff
@@ -411,4 +416,71 @@ private static TagKey<Fluid> tag(String name)
...
+        public static final TagKey<Biome> IS_DENSE_END = tag("is_dense/end");
+
+        public static final TagKey<Biome> IS_WET = tag("is_wet");
+        public static final TagKey<Biome> IS_WET_OVERWORLD = tag("is_wet/overworld");
+        public static final TagKey<Biome> IS_WET_NETHER = tag("is_wet/nether");
+        public static final TagKey<Biome> IS_WET_END = tag("is_wet/end");
+        public static final TagKey<Biome> IS_DRY = tag("is_dry");
+        public static final TagKey<Biome> IS_DRY_OVERWORLD = tag("is_dry/overworld");
+        public static final TagKey<Biome> IS_DRY_NETHER = tag("is_dry/nether");
+        public static final TagKey<Biome> IS_DRY_END = tag("is_dry/end");
```
sciwhiz12, line 450: All entries should be in the dimension-specific tags, and those tags should be children of the main tag. Currently, most biomes are put both in the main tag and in dimension-specific tags, which duplicates data unnecessarily, as well as the empty dimension-specific tags not being set as children of the main tag.

src/main/java/net/minecraftforge/common/data/ForgeBiomeTagsProvider.java
```diff
@@ -0,0 +1,107 @@
...
+    private void tag(ResourceKey<Biome> biome, TagKey<Biome>... tags)
+    {
+        for(TagKey<Biome> key : tags)
+        {
+            tag(key).add(biome);
+        }
+    }
+
+    @Override
+    public String getName() {
```
sciwhiz12, line 104: Formatting: braces on newlines.
```suggestion
    public String getName()
    {
```

src/main/java/net/minecraftforge/common/data/ForgeBiomeTagsProvider.java
```diff
@@ -0,0 +1,107 @@
...
+        tag(Biomes.BASALT_DELTAS, Tags.Biomes.IS_HOT, Tags.Biomes.IS_HOT_NETHER, Tags.Biomes.IS_DRY, Tags.Biomes.IS_DRY_NETHER);
+
+        tag(Tags.Biomes.IS_WATER).addTag(BiomeTags.IS_OCEAN).addTag(BiomeTags.IS_RIVER);
+        tag(BiomeTags.IS_MOUNTAIN).addTag(Tags.Biomes.IS_PEAK).addTag(Tags.Biomes.IS_SLOPE);
+    }
+
+    @SafeVarargs
+    private void tag(ResourceKey<Biome> biome, TagKey<Biome>... tags)
+    {
+        for(TagKey<Biome> key : tags)
```
sciwhiz12, line 97: Formatting: space between `for` keyword and opening parenthesis.
```suggestion
        for (TagKey<Biome> key : tags)
```

src/main/java/net/minecraftforge/common/data/ForgeBiomeTagsProvider.java
```diff
@@ -0,0 +1,107 @@
...
+import net.minecraft.resources.ResourceKey;
+import net.minecraft.tags.BiomeTags;
+import net.minecraft.tags.TagKey;
+import net.minecraft.world.level.biome.Biome;
+import net.minecraft.world.level.biome.Biomes;
+import net.minecraftforge.common.Tags;
+
+public final class ForgeBiomeTagsProvider extends BiomeTagsProvider
+{
+
```
sciwhiz12, line 19: Formatting: no blank line between class declaration and first class member.
```suggestion
```

src/main/java/net/minecraftforge/common/BiomeDictionary.java
```diff
@@ -25,6 +26,11 @@
 
 import net.minecraft.world.level.biome.Biome;
 
+/**
+ * Use {@link Tags.Biomes} when possible.
+ * Note that tags are not usable yet during {@link BiomeLoadingEvent} so BiomeDictionary are still necessary there
```
sciwhiz12, line 31: Grammatical error: `are` to `is`.
```suggestion
 * Note that tags are not usable yet during {@link BiomeLoadingEvent} so BiomeDictionary is still necessary there
```

sciwhiz12 (changes requested):
There's some issues to be resolved with this PR before we can push for assignment (hopefully before 1.19 drops so we can get this in for 1.19's final checking).

- Regarding the dimension-specific tags, such as `forge:is_hot/end` for `forge:is_hot` (the main tag):
	- These tags need to be added as children of the main tag, which leads to the next point;
	- These tags should contain all (vanilla) biome entries instead of being in the main tag, so the main tag _only_ contains references to its child dimension-specific tags.

- ~~There are certain inconsistencies between the biome entries in the new tags, the `BiomeDictionary`, and the biome tags added by vanilla itself.
  These three sources should match, for consistent behavior whether a mod uses the tags or `BiomeDictionary`.~~
	- ~~The biomes `minecraft:small_end_islands`, `minecraft:end_midlands`, `minecraft:end_highlands`, and `minecraft:end_barrens` are listed in the new tags, but are not marked as `DRY` in the `BiomeDictionary`.~~
	- ~~The vanilla tag `minecraft:is_forest` misses some biomes that are marked with the equivalent `BiomeDictionary#FOREST`: `minecraft:taiga`, `minecraft:sparse_jungle`, `minecraft:snowy_taiga`, `minecraft:old_growth_pine_taiga`, `minecraft:windswept_forest`, `minecraft:crimson_forest`, and `minecraft:warped_forest`.~~
	- ~~The new tag `forge:is_plains` contains `minecraft:snowy_plains` which is not marked with `BiomeDictionary#PLAINS`, while the tag misses `minecraft:savanna` and `minecraft:savanna_plateau` which are marked with `BiomeDictionary#PLAINS`.~~
	- ~~The new tag `forge:is_beach` and `BiomeDictionary#BEACH` contains `minecraft:stony_shore`, but that is not contained in the vanilla tag `minecraft:is_beach`.~~

- What is the difference between `forge:is_coniferous` (the equivalent to `BiomeDictionary#CONIFEROUS`) and vanilla's `minecraft:is_taiga`? Do they refer to the same set of biomes (with regards to the features of said biome) with only different names?
	- If they are the same, then there are inconsistencies: `minecraft:grove` is present in the new tag and `BiomeDictionary` while not in the vanilla tag, while `old_growth_spruce_taiga` is present in the vanilla tag while neither in the new tag nor `BiomeDictionary`.

- Should we note that the corresponding vanilla tag for `BiomeDictionary#MESA` is `minecraft:is_badlands` in its javadoc?

TelepathicGrunt:
@sciwhiz12 please read the conversation above for why savanna is not in the plains tag. The new tags do not need to match the old messy and flawed biome dictionary entries. This is a great time to clean it up and fix the entries so that they actually make sense instead of just matching old entries for the sake of matching old entries. Remember, people can’t filter out tag entries easily when using the tags for structures and stuff. So if the biome tags can be as clean and accurate as possible, it will greatly help everyone

sciwhiz12:
If we are to clean up the entries in the new tags, should we then match up the `BiomeDictionary` to whatever's in the new tags, so they are both equal in contents for the benefit of those still using `BiomeDictionary`? I'm thinking that there may possibly be issues if the `BiomeDictionary` and the new biome tags, which are intended to replace the `BiomeDictionary` in the future if I understand it right, do not match in their contents.

TelepathicGrunt:
The biome dictionary has different usage abilities. One can get dry overworld biome by combining the biome dictionary's dry tag and overworld tag in an intersection way. However, we cannot do this with biome tags as tags only allow unions. So adding stuff like end biomes into the dry biome tag means that tag is no longer usable by anyone because now if they use that tag, their structure spawns in both overworld and end dimensions. 

Imo, the biome tags should not match biome dictionary at all. If the biome dictionary should change, that's fine but may be better as a separate pr to determine what to match and what to not due to how one can do more tailored filters with biome dictionary than they can with tags.

KnightMiner:
You can do tag intersections in code just as easily as you can do biome dictionary intersections in code, so I am not convinced we need biome dict just to do intersections. Personally I think its better to do away with the biome dict entirely, datapack biomes have no way to register with that causing incompatibilities. Plus, forcing modders to maintain both will likely lead to inconsistencies when a modder forgets to add their biomes to one.

TheCurle:
A great step forward, and combined with some incoming changes, this will provide a great replacement for an entire suite of hooks.

Thanks, Flemmli97!

PR #8565, opened 2022-04-12
Deprecate alwaysSetupTerrainOffThread option

williewillus:
It is wholly superseded by vanilla's "Video Settings > Chunk Builder" option. The config field is kept in case anyone is reading it.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeConfig.java
+++ b/src/main/java/net/minecraftforge/common/ForgeConfig.java
@@ -110,6 +110,7 @@ public class ForgeConfig {
      * Client specific configuration - only loaded clientside from forge-client.toml
      */
     public static class Client {
+        @Deprecated(forRemoval = true) // Superseded by vanilla option
         public final BooleanValue alwaysSetupTerrainOffThread;
 
         public final BooleanValue experimentalForgeLightPipelineEnabled;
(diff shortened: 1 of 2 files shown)


src/main/java/net/minecraftforge/common/ForgeConfig.java
```diff
@@ -110,6 +110,7 @@ public static class Common {
      * Client specific configuration - only loaded clientside from forge-client.toml
      */
     public static class Client {
+        @Deprecated(forRemoval = true) // Superseded by vanilla option
```
sciwhiz12, line 113: Please specify the version through `since`:
```suggestion
        @Deprecated(forRemoval = true, since = "1.18.2") // Superseded by vanilla option
```

sciwhiz12 (changes requested):
Should we modify the config comment to reflect its obsolescence by the vanilla config setting and removal in a future version? 
What about keeping the config setting working for now (modifying the patch as needed) so it still works as users/modpack developers expect it should, and removing it completely in a future version?

williewillus:
> What about keeping the config setting working for now (modifying the patch as needed) so it still works as users/modpack developers expect it should

if we wanted that, we'd just leave the current patch in place

TheCurle:
I concur. This option suddenly breaking with no explanation (from an end-user's perspective) is an unacceptable change.

This change has been put on the changelist for 1.19, and i'm going to close this PR.

PR #8566, opened 2022-04-12
Make it possible to add new perspectives / item transform types

gigaherz:
## Rationale

Mods often add new ways to display items, such as my Tool Belt mod, which displays items on either side of the player, attached to the belt.

Those mods often have to resort to using vanilla perspectives, which may not work equally for all items.

Example: In my Tool Belt mod, I could use either the FIXED (item frame) perspective, or the third person hand perspective. If I used FIXED, some items that would display upright on an item frame, should display hanging upside down in the belt. But when using the third person hand perspectives and rotating them so they look hanging, the items sometimes show hanging upside down when they should look upright. There is no perspective that will satisfy all items in all cases.

## Design

To achieve this, I have made the ItemTransforms.TransformType extensible, and I have added the serialize name of the perspective as a field to the enum.

All a modder needs to do is call the create method, and the game will do the rest.
```java
public static final ItemTransforms.TransformType MY_TYPE = ItemTransforms.TransformType.create("MODNAME_TYPENAME", new ResourceLocation("modid", "typename"));

// or with a fallback type
public static final ItemTransforms.TransformType MY_TYPE_WITH_FALLBACK = ItemTransforms.TransformType.create("MODNAME_TYPENAME2", new ResourceLocation("modid", "typename2", ItemTransforms.TransformType.FIXED));
```

Then transforms and perspectives can be used via `getTransform` and `handlePerspective` respectively, using the extended type entry.


patches/minecraft/net/minecraft/client/renderer/block/model/BlockModel.java.patch
```diff
@@ -78,6 +78,27 @@
...
+       ItemTransform itemtransform6 = this.m_111444_(ItemTransforms.TransformType.GROUND);
+       ItemTransform itemtransform7 = this.m_111444_(ItemTransforms.TransformType.FIXED);
+-      return new ItemTransforms(itemtransform, itemtransform1, itemtransform2, itemtransform3, itemtransform4, itemtransform5, itemtransform6, itemtransform7);
++
++      var builder = com.google.common.collect.ImmutableMap.<ItemTransforms.TransformType, ItemTransform>builder();
++      for(ItemTransforms.TransformType type : ItemTransforms.TransformType.values()) {
++         if (type.isModded()) {
++            var transform = this.m_111444_(type);
++            if (transform != ItemTransform.f_111754_)
++            {
```
pupnewfster: Patches should have brackets on the same line

patches/minecraft/net/minecraft/client/renderer/block/model/ItemTransforms.java.patch
```diff
@@ -1,26 +1,122 @@
...
++      this.moddedTransforms = p_111807_.moddedTransforms;
     }
  
 +   @Deprecated
     public ItemTransforms(ItemTransform p_111798_, ItemTransform p_111799_, ItemTransform p_111800_, ItemTransform p_111801_, ItemTransform p_111802_, ItemTransform p_111803_, ItemTransform p_111804_, ItemTransform p_111805_) {
++      this(p_111798_, p_111799_, p_111800_, p_111801_, p_111802_, p_111803_, p_111804_, p_111805_, com.google.common.collect.ImmutableMap.of());
++   }
++
++   public ItemTransforms(ItemTransform p_111798_, ItemTransform p_111799_, ItemTransform p_111800_, ItemTransform p_111801_, ItemTransform p_111802_, ItemTransform p_111803_, ItemTransform p_111804_, ItemTransform p_111805_,
++           com.google.common.collect.ImmutableMap<TransformType, ItemTransform> moddedTransforms) {
```
pupnewfster, line 24: Missing exc entry

patches/minecraft/net/minecraft/client/renderer/block/model/ItemTransforms.java.patch
```diff
@@ -1,26 +1,122 @@
...
+          ItemTransform itemtransform6 = this.m_111815_(p_111822_, jsonobject, "ground");
+          ItemTransform itemtransform7 = this.m_111815_(p_111822_, jsonobject, "fixed");
+-         return new ItemTransforms(itemtransform1, itemtransform, itemtransform3, itemtransform2, itemtransform4, itemtransform5, itemtransform6, itemtransform7);
++
++         var builder = com.google.common.collect.ImmutableMap.<TransformType, ItemTransform>builder();
++         for(TransformType type : TransformType.values()) {
++            if (type.isModded()) {
++               var transform = this.m_111815_(p_111822_, jsonobject, type.getSerializeName());
++               if (transform != ItemTransform.f_111754_)
++               {
```
pupnewfster: Bracket on same line in patches

patches/minecraft/net/minecraft/client/renderer/block/model/ItemTransforms.java.patch
```diff
@@ -1,26 +1,122 @@
...
++         isModded = false;
++      }
++
++      private TransformType(net.minecraft.resources.ResourceLocation serializeName){
++         this.serializeName = java.util.Objects.requireNonNull(serializeName, "Modded TransformTypes must have a non-null serializeName").toString();
++         isModded = true;
++      }
++
++      public boolean isModded()
++      {
```
pupnewfster: Brackets on same line in patches

src/main/java/net/minecraftforge/client/model/ModelLoaderRegistry.java
```diff
@@ -214,18 +214,10 @@ public static Optional<ModelState> deserializeTransform(JsonDeserializationConte
...
-
-            deserializeTRSR(context, transforms, transform, "firstperson", ItemTransforms.TransformType.FIRST_PERSON_RIGHT_HAND);
-            deserializeTRSR(context, transforms, transform, "firstperson_righthand", ItemTransforms.TransformType.FIRST_PERSON_RIGHT_HAND);
-            deserializeTRSR(context, transforms, transform, "firstperson_lefthand", ItemTransforms.TransformType.FIRST_PERSON_LEFT_HAND);
-
-            deserializeTRSR(context, transforms, transform, "head", ItemTransforms.TransformType.HEAD);
-            deserializeTRSR(context, transforms, transform, "gui", ItemTransforms.TransformType.GUI);
-            deserializeTRSR(context, transforms, transform, "ground", ItemTransforms.TransformType.GROUND);
-            deserializeTRSR(context, transforms, transform, "fixed", ItemTransforms.TransformType.FIXED);
+            for(var type : ItemTransforms.TransformType.values())
```
pupnewfster: Space before parenthesis

KnightMiner:
Would it be possible to have new transforms default to an existing one? That is, specify 
```java
ItemTransforms.TransformType.create("MODNAME_TYPENAME", new ResourceLocation("modid", "typename"), ItemTransforms.TransformType.FIXED);
```
to say if missing, default to `FIXED`? Most modded items will not be adding to my transform type by default, so it would be nice if we could choose some fallback that looks pretty good for the mods that don't add specific compat.

Code wise, I think you just need a nullable getter on the transform type enum for the fallback, and when fetching the transforms from the map, if missing in the map fetch the default.

patches/minecraft/net/minecraft/client/renderer/block/model/ItemTransforms.java.patch
```diff
@@ -1,26 +1,122 @@
...
+-      THIRD_PERSON_LEFT_HAND,
+-      THIRD_PERSON_RIGHT_HAND,
+-      FIRST_PERSON_LEFT_HAND,
+-      FIRST_PERSON_RIGHT_HAND,
+-      HEAD,
+-      GUI,
+-      GROUND,
+-      FIXED;
++   public static enum TransformType implements net.minecraftforge.common.IExtensibleEnum {
++      NONE("none"),
```
KnightMiner, line 85: These look to be pretty uniformly the lowercase of the enum identifier, so you could reduce patch size a lot by doing a zero arg constructor
```java
private TransformType() {
  this.serializedName = this.name().toLowerCase(Locale.ROOT);
  isModded = false;
}
```

gigaherz: Sadly no, as I just learned while writing the test mod:
```
      THIRD_PERSON_LEFT_HAND("thirdperson_lefthand"),
      THIRD_PERSON_RIGHT_HAND("thirdperson_righthand"),
      FIRST_PERSON_LEFT_HAND("firstperson_lefthand"),
      FIRST_PERSON_RIGHT_HAND("firstperson_righthand"),
```
the names don't quite match up.

KnightMiner: I assume having both a zero arg (for exceptions) and a 1 arg parameter won't save much in patch size.

I suppose you could could use a switch for the exceptions, but not convinced that would be cleaner either.

patches/minecraft/net/minecraft/client/renderer/block/model/ItemTransforms.java.patch
```diff
@@ -54,7 +52,12 @@
 +         for (TransformType type : TransformType.values()) {
 +            if (type.isModded()) {
 +               var transform = this.m_111815_(p_111822_, jsonobject, type.getSerializeName());
-+               if (transform != ItemTransform.f_111754_) {
```
KnightMiner: Is there any concern about order here for fallbacks? Might need to document that a fallback must be an earlier element in the enum, or require fallbacks to be vanilla.

gigaherz: Either the key exists in the json, or it doesn't. The ordering is irrelevant... unless I missed something.

patches/minecraft/net/minecraft/client/renderer/block/model/ItemTransforms.java.patch
gigaherz: It doesn't feel like we gain anything by doing it.

patches/minecraft/net/minecraft/client/renderer/block/model/ItemTransforms.java.patch
KnightMiner: I could also have missed something. Do you process fallbacks at the same time as you serialize, or do fallbacks in a second pass? In other words, is it possible when I see the key does not exist (and need a fallback) that the fallback has not yet parsed?

gigaherz: The code doesn't use existing parsed data, it just re-parses the same data for the new key. The fallback is only used to decide which key to read from.

gigaherz:
I can't think of anything else that may be necessary, so I'm considering this not a draft anymore.

src/test/java/net/minecraftforge/debug/client/model/CustomTransformTypeTest.java
```diff
@@ -0,0 +1,285 @@
+package net.minecraftforge.debug.client.model;
```
TheCurle, line 6: License?

src/test/java/net/minecraftforge/debug/client/model/CustomTransformTypeTest.java
```diff
@@ -0,0 +1,285 @@
...
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.forge.event.lifecycle.GatherDataEvent;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.RegistryObject;
+import org.jetbrains.annotations.Nullable;
+
+@Mod(CustomTransformTypeTest.MODID)
```
TheCurle, line 61: Please explain what this is testing; what it does, and what should be looked for to consider the test a "pass".

gigaherz: Done. And the license too.

TheCurle:
Looks good. 
Someone will have to remind one of us to pull this tomorrow when the GradleUtils issue is fixed (and the new version is detected by Gradle).

PR #8567, opened 2022-04-12
Add Gzip support to version checker (partial backport of #8148 from 1.17)

PaintNinja:
This PR is a partial backport of #8148 to 1.16, adding gzip compression support for version checking.


PaintNinja:
Related: https://github.com/MinecraftForge/Web/pull/20 (makes a promotions_slim.json.gz at pagegen rather than on each request)

LexManos:
Meh, doesn't hurt.

PR #8568, opened 2022-04-12
Update IExtensionPoint examples to reflect API changes

TheMrMilchmann:
Pretty self-explanatory.


TheCurle:
We usually don't do merges like this for javadoc additions, due to the cost of the build and no user-facing changes. 

However, this is an extremely common (and very correct) complaint we receive a lot, so i'm going to make the executive decision to waive that concern.

Thanks for the contribution, TheMrMilchmann.

PR #8570, opened 2022-04-14
[1.15.x] Add Gzip support to version checker (backport of #8567)

PaintNinja:
Thought I might as well as the more users that have this, the more bandwidth saved `¯\_(ツ)_/¯`


TheCurle:
This is blocked on a lot of other changes to the 1.15.x build, due to changes in our build pipeline that requires FG5. 

Since 1.15 is not LTS, it's hard to justify such extensive changes for such little gain, so i'm going to close this PR.

PR #8571, opened 2022-04-15
Add event for hooking into StructuresBecomeConfiguredFix

### LINKED ISSUE #8505: [1.18.2-40.0.12] Modded structures deletes and regens chunks when updating 1.18.1 or older worlds into 1.18.2
So if you put on a structure mod in 1.18.1 like Repurposed Structures and make a world and save it while looking at a modded structure, updating the world to 1.18.2 without the structure mod on (or even still have it on) will make those chunks with the structure be regenerated entirely. Like everything placed in those chunks will be gone as if the structure or your changes never existed.

The old behavior before 1.18.1 was the chunk keeps the chunk's blocks the same and not regen it. It would logspam about missing structures tho. In 1.181, it now regens the chunks which is destructive while also logspamming about missing structures. This should be something forge should fix (the chunk regen and logspam) as otherwise, people updating worlds are going to have a lot of pain and lost builds when they think they are just updating to a newer minor mc version.

Mod: https://www.curseforge.com/minecraft/mc-mods/repurposed-structures

Example 1.18.1 Repurposed Structures world to test with. Go to /tp @s -448 ~ 7248 for a mansion in 1.18.1. Save. And then load the world in 1.18.2 
https://www.dropbox.com/s/1qa1vd2kagmhuve/oldworld.zip?dl=0

Pics:
image
image
image

The desired result is that when updating 1.18.1 or older worlds to 1.18.2, all blocks remain as is in the chunks and no chunk is regenerated regardless if the 1.18.2 mod is on or not.

sciwhiz12:
This PR fixes #8505 by implementing options 1 and 3 as described in https://github.com/MinecraftForge/MinecraftForge/issues/8505#issuecomment-1062676787: old structure IDs which are namespaced are passed along as the new structure IDs, and a new `RegisterStructureConversionsEvent` is added for registering structure conversions.

The new event has been documented thoroughly for the benefit of consumers. I've not implemented a test mod, mostly because testing this event will require manual work on behalf of the tester (creating the world in an older version with a strucuture datapack/mod, loading the same world on this version with the same structure datapack/mod).

I've tested this PR using the same setup I used to confirm the linked issue, and it does work as expected. I request people to test this PR out locally, especially in using the new event for converting old structure IDs to new ones based on biomes.

<details>
<summary>Screenshot of the PR in action -- using the same pre-1.18.2 world from the linked issue, loaded on a production installation of my PR</summary>

image

</details>

### DIFF
--- a/src/main/java/net/minecraftforge/event/RegisterStructureConversionsEvent.java
+++ b/src/main/java/net/minecraftforge/event/RegisterStructureConversionsEvent.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright (c) Forge Development LLC and contributors
+ * SPDX-License-Identifier: LGPL-2.1-only
+ */
+
+package net.minecraftforge.event;
+
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.util.datafix.fixes.StructuresBecomeConfiguredFix;
+import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.eventbus.api.Cancelable;
+import net.minecraftforge.eventbus.api.Event;
+
+import java.util.Locale;
+import java.util.Map;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Fired for registering structure conversions for pre-1.18.2 worlds. This is used by {@link StructuresBecomeConfiguredFix}
+ * for converting old structure IDs in pre-1.18.2 worlds to their new equivalents, which can be differentiated per biome.
+ *
+ * <p>By default, structures whose old ID has a namespace which is not equal to {@value ResourceLocation#DEFAULT_NAMESPACE}
+ * will be assumed to belong to a modded structure and will be used as the new ID. Mods may choose to register structure
+ * conversions for their structures, if they wish to override this default behavior.</p>
+ *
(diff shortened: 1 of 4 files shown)


src/main/java/net/minecraftforge/event/RegisterStructureConversionsEvent.java
```diff
@@ -0,0 +1,105 @@
...
+     *     <li>A <em>biome-mapped</em> conversion, created using {@link
+     *     StructuresBecomeConfiguredFix.Conversion#biomeMapped(Map, String)}, contains a fallback structure ID, and a
+     *     biome-specific conversion map. Each entry in the map is composed of a list of biome IDs and the new structure
+     *     ID.
+     *
+     *     <p>If a structure is in a biome which exists in the map, the
+     *     structure ID in the corresponding entry is used as the new structure ID. If there is no such biome found,
+     *     the new structure ID will be the fallback structure ID.</p>
+     *
+     *     <p>For example, the following registers a biome-mapped conversion for {@code "xampleStructure} with the
```
alcatrazEscapee:
```suggestion
     *     <p>For example, the following registers a biome-mapped conversion for {@code "exampleStructure} with the
```
Typo

sciwhiz12: screm.

marchermans (approved):

PR #8575, opened 2022-04-19
[1.18.x] Fix ContextNbtProvider serializing wrong name

SizableShrimp:
In vanilla, ContextNbtProvider serializes `LootContext.EntityTarget` using `Enum#name`. This leads to serializing with `"THIS"`, `"KILLER"`, `"DIRECT_KILLER"`, or `"KILLER_PLAYER"`. However, deserializing through `ContextNbtProvider#createFromContext` uses `LootContext.EntityTarget#getByName`. This does not use the enum name, but rather a private field named `name` which is in lowercase. This leads to issues when trying to datagen a `ContextNbtProvider` made for an entity target, as the serialization will be invalid since the datagen produced invalid results. This PR fixes the issue by changing `getId()` to use a new patched `getName()` method on EntityTarget, since the `name` field is private. Testing from @Tslat would be appreciated as he was the user to experience this issue.

### DIFF
--- a/patches/minecraft/net/minecraft/world/level/storage/loot/LootContext.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/storage/loot/LootContext.java.patch
@@ -66,12 +66,13 @@
              }
           }
        }
-@@ -230,6 +_,10 @@
+@@ -230,6 +_,11 @@
  
        public LootContextParam<? extends Entity> m_79003_() {
           return this.f_78995_;
 +      }
 +
++      // Forge: This method is patched in to expose the same name used in getByName so that ContextNbtProvider#forEntity serializes it properly
 +      public String getName() {
 +         return this.f_78994_;
        }
--- a/patches/minecraft/net/minecraft/world/level/storage/loot/LootContext.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/storage/loot/LootContext.java.patch
@@ -66,3 +66,14 @@
              }
           }
        }
+@@ -230,6 +_,10 @@
+ 
+       public LootContextParam<? extends Entity> m_79003_() {
+          return this.f_78995_;
++      }
++
++      public String getName() {
++         return this.f_78994_;
+       }
+ 
+       public static LootContext.EntityTarget m_79006_(String p_79007_) {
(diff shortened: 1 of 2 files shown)


marchermans (approved):

marchermans:
Awaiting testing by @tslat before assigning.

Tslat:
Should probably patch both of the Serializers for ContextNbtProvider
the class has both an InlineSerializer and a Serializer, which both have the same flaw
image

Tslat:
I can get to it in time
At face value though it looks perfect, there's really not a lot to this one

patches/minecraft/net/minecraft/world/level/storage/loot/LootContext.java.patch
```diff
@@ -66,3 +66,14 @@
              }
           }
        }
+@@ -230,6 +_,10 @@
+ 
+       public LootContextParam<? extends Entity> m_79003_() {
+          return this.f_78995_;
++      }
++
++      public String getName() {
```
sciwhiz12, line 76: I'd like for this patched-in method to be clearly marked as such with a comment (javadoc or regular).

sciwhiz12 (approved):
Tested locally with a small kludge in a test mod, and it works fine.

It should be noted that this doesn't occur in vanilla's (datagenerated) loot tables because only `ContextNbtProvider.BLOCK_ENTITY` is used (for block loot tables).

~~I do wonder if this should be filed as a bug on Mojira? I'm not sure on the precedent on reporting code bugs which aren't exactly user-facing issues.~~  Though, in the case that Mojang ever adds datagenerates something using `ContextNbtProvider#forContextEntity` (such as through `CopyNbtFunction#copyData(LootContext.EntityTarget)`, they will likely encounter this same issue.

**Edit:** After a bit of internal discussion, we won't be needing a bug filed on Mojira for this.

Tslat:
Thanks Sci

patches/minecraft/net/minecraft/world/level/storage/loot/LootContext.java.patch
TheCurle: I concur. This can be very misleading during a port update.

sciwhiz12 (approved):

PR #8576, opened 2022-04-21
Multipart entity id desync fix

Meldexun:
Fixes entity id desync of multipart entities by also sending the ids of part entities.

Continuation of #8190


patches/minecraft/net/minecraft/network/protocol/game/ClientboundAddMobPacket.java.patch
```diff
@@ -0,0 +1,44 @@
...
+       this.f_131539_ = p_178566_.readShort();
++      this.partIds = p_178566_.m_130100_();
+    }
+ 
+    public void m_5779_(FriendlyByteBuf p_131554_) {
+@@ -70,6 +_,7 @@
+       p_131554_.writeShort(this.f_131537_);
+       p_131554_.writeShort(this.f_131538_);
+       p_131554_.writeShort(this.f_131539_);
++      p_131554_.m_130089_(this.partIds);
```
ChampionAsh5357: This is not suitable for FS -> VC. This will most definitely throw an `IndexOutOfBoundsException` when loading a multipart entity. You will need to implement a network filter to handle connections to a vanilla client with this method. 

Otherwise, a different solution would need to be used such as deferring the creation of the entity id until when the entity is added to a world.

Meldexun: I think you mean VanillaServer and ForgeClient (With a ForgeServer and a VanillaClient the client will just ignore the extra data that is being sent).
I would add some `if (Minecraft.getInstance().getCurrentServer().forgeData.type.equals("VANILLA"))` statements to check if the server is a vanilla server . I would create a helper method for that in ForgeHooksClient to handle the sidedness of `Minecraft`. Or did you mean something else with a network filter? Or is there already a helper method to check if the server is a vanilla server?

diesieben07: You also have to handle forge servers that do not have this patch yet.
And please verify that additional data is actually ignored - I am not certain on that.

diesieben07: I checked the code and as I suspected: extra data is _not_ ignored (from `PacketDecoder`):
> `throw new IOException("Packet " + p_130535_.channel().attr(Connection.ATTRIBUTE_PROTOCOL).get().getId() + "/" + j + " (" + packet.getClass().getSimpleName() + ") was larger than I expected, found " + friendlybytebuf.readableBytes() + " bytes extra whilst reading packet " + j);`

Meldexun: Hm ok. That means I have to check if the connected client is a forge client and what forge version the client is. But I would have to do this inside the `ClientboundAddMobPacket.write` method where I don't know to which client the packet will be sent to. So do you have an idea how I could handle the server side of this problem.

Meldexun: I found the `ForgeConnectionNetworkFilter` class. So this is what you meant with network filter. Will work on a fix for this problem now.

Meldexun: Changed the PR to not make changes to the Vanilla network protocol.

Meldexun:
Changed the PR according to the discussion on discord. Instead of changing vanilla packets to also send the part entity ids it now makes sure that the ids of the part entities are the direct successors of the parent mob id in the constructor of the enderdragon mob. Maybe not the nicest solution for mod authors but they can do the same in their multipart entity class to fix the problem.

~~Also I added a fix for the `World.getEntities` methods to not ignore part entities when the parent mob is in a chunk that doesn't intersect with the AABB.~~
EDIT: A new PR will be created for this fix

LexManos:
> Also I added a fix for the `World.getEntities` methods to not ignore part entities when the parent mob is in a chunk that doesn't intersect with the AABB.

No

Meldexun:
Why no? If forge doesn't patch the issue modders will have like three options in my opinion:
1. Implement their own fix for the bug for their own multipart entities
2. Make/Use a small utility mod which fixes the bug for all multipart entities
3. Wait for vanilla to fix it...
These options all don't sound really good and the problem affects vanilla and modded multipart entities. Also forge is already patching these methods but that patch is not fixing the issue. So I don't really understand why you prefer a patch that does not fix the issue over a slightly different patch that does fix the issue.

marchermans:
> Why no? If forge doesn't patch the issue modders will have like three options in my opinion:
> 1. Implement their own fix for the bug for their own multipart entities
> 2. Make/Use a small utility mod which fixes the bug for all multipart entities
> 3. Wait for vanilla to fix it...
> These options all don't sound really good and the problem affects vanilla and modded multipart entities. Also forge is already patching these methods but that patch is not fixing the issue. So I don't really understand why you prefer a patch that does not fix the issue over a slightly different patch that does fix the issue.

That is a secondary change.
If you also want that fixed you will need to make a second pr.

patches/minecraft/net/minecraft/world/level/Level.java.patch
```diff
@@ -250,6 +266,15 @@
...
++
     public final boolean m_46659_() {
+       return this.f_46424_;
+    }
+@@ -900,4 +_,6 @@
+    public long m_183596_() {
+       return (long)(this.f_186455_++);
+    }
++
++   public abstract java.util.Collection<net.minecraftforge.entity.PartEntity<?>> getPartEntities();
```
diesieben07: Does this count as a backwards-incompatible change? Do we care about custom Level subclasses in mods?

marchermans: They are actually quite common.....
Especially when rendering structures etc. in the world.

diesieben07: Okay, then it should be a default method in `IForgeLevel`

LexManos: This entire 'fix' is not needed/not justified. There is a reason why we tell people not to add random unrelated things in PRs. No idea where this came from, our conversation on discord was related to the ID issue.

Meldexun: Reverted the last commit. A new PR will be created for the `getEntities` method fix.

TheCurle:
I'm not entirely sure what this fixes, and what the effects of this bug are.
However, i've had testimony from some other modders that this is relatively annoying to work around.

That being said, modders may need to be careful when updating to this version because of said workarounds. 
Currently modders are not *expecting* the IDs to be contiguous, whereas they now will be.

I'll make a FUN announcement for this.

Thanks for the contribution, Meldexun.

LexManos:
This was waiting on a cleanup  of the patch, but we can do that later.
And the thing is, the game expects the ids to be contiguous, thee problem is that they are not.

PR #8578, opened 2022-04-21
Grindstone event

ferriarnus:
An event implementation for the grindstone. This is heavily based on the anvil event. I've made this implementation based on the talks in the discord chat, so any comments/remarks are certainly welcome if I've missed something.

A quick question: similar to the anvil events, these events do not handle changes in the input items. Would another event (or a "post" type event) be good to achieve this? Currently the vanilla behavior is to set the slots to empty, deleting whatever was in them, even if it was a full stack.

### DIFF
--- a/patches/minecraft/net/minecraft/world/inventory/GrindstoneMenu.java.patch
+++ b/patches/minecraft/net/minecraft/world/inventory/GrindstoneMenu.java.patch
@@ -1,5 +1,45 @@
 --- a/net/minecraft/world/inventory/GrindstoneMenu.java
 +++ b/net/minecraft/world/inventory/GrindstoneMenu.java
+@@ -35,6 +_,7 @@
+       }
+    };
+    private final ContainerLevelAccess f_39561_;
++   private int xp = -1;
+ 
+    public GrindstoneMenu(int p_39563_, Inventory p_39564_) {
+       this(p_39563_, p_39564_, ContainerLevelAccess.f_39287_);
+@@ -45,12 +_,12 @@
+       this.f_39561_ = p_39568_;
+       this.m_38897_(new Slot(this.f_39560_, 0, 49, 19) {
+          public boolean m_5857_(ItemStack p_39607_) {
+-            return p_39607_.m_41763_() || p_39607_.m_150930_(Items.f_42690_) || p_39607_.m_41793_();
++            return true;
+          }
+       });
+       this.m_38897_(new Slot(this.f_39560_, 1, 49, 40) {
+          public boolean m_5857_(ItemStack p_39616_) {
+-            return p_39616_.m_41763_() || p_39616_.m_150930_(Items.f_42690_) || p_39616_.m_41793_();
++            return true;
+          }
+       });
+       this.m_38897_(new Slot(this.f_39559_, 2, 129, 34) {
+@@ -71,6 +_,7 @@
+          }
+ 
+          private int m_39631_(Level p_39632_) {
++            if (xp > -1) return xp;
+             int l = 0;
+             l += this.m_39636_(GrindstoneMenu.this.f_39560_.m_8020_(0));
+             l += this.m_39636_(GrindstoneMenu.this.f_39560_.m_8020_(1));
+@@ -123,6 +_,8 @@
+       ItemStack itemstack1 = this.f_39560_.m_8020_(1);
+       boolean flag = !itemstack.m_41619_() || !itemstack1.m_41619_();
+       boolean flag1 = !itemstack.m_41619_() && !itemstack1.m_41619_();
++      this.xp = -1;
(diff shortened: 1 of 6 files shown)


src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -674,6 +677,24 @@ public static float onAnvilRepair(Player player, @Nonnull ItemStack output, @Non
...
+            xp = e.getXp();
+            return false;
+        }
+        if (e.getOutput().isEmpty()) 
+        {
+            return true;
+        }
+
+        outputSlot.setItem(0, e.getOutput());
+        xp = e.getXp();
```
XFactHD: All assignments to the `xp` parameter are dropped when this method returns, they will never reach the `GrindstoneMenu#xp` variable you added.

patches/minecraft/net/minecraft/world/inventory/GrindstoneMenu.java.patch
```diff
@@ -1,5 +1,45 @@
...
++   private int xp = -1;
+ 
+    public GrindstoneMenu(int p_39563_, Inventory p_39564_) {
+       this(p_39563_, p_39564_, ContainerLevelAccess.f_39287_);
+@@ -45,12 +_,12 @@
+       this.f_39561_ = p_39568_;
+       this.m_38897_(new Slot(this.f_39560_, 0, 49, 19) {
+          public boolean m_5857_(ItemStack p_39607_) {
+-            return p_39607_.m_41763_() || p_39607_.m_150930_(Items.f_42690_) || p_39607_.m_41793_();
++            return true;
```
KnightMiner, line 16: This may cause a desync between a forge server and a vanilla client or vice versa, have you tested that case? Should be able to launch both through the dev environment, use offline mode for the server.

src/main/java/net/minecraftforge/event/GrindstoneUpdateEvent.java
```diff
@@ -0,0 +1,95 @@
...
+ * If the event is not canceled, but the output is not empty, it will set the output and not run vanilla behavior. <br>
+ * if the output is empty, and the event is not canceled, vanilla behavior will execute. <br>
+ * if the amount of experience is larger than or equal 0, the vanilla behavior for calculating experience will not run. <br>
+ */
+@Cancelable
+public class GrindstoneUpdateEvent extends Event
+{
+	
+    private final ItemStack top;
+    private final ItemStack bottem;
```
KnightMiner: typo, should be `bottom`

src/main/java/net/minecraftforge/common/ForgeHooks.java
ferriarnus: Oh that's stupid of me. The simpelst way to fix this is make use of an int[] or some "mutable int" instead. What would be the best way?

XFactHD: I would say `int[]` or the `MutableInt` from apache-commons.

diesieben07: Can you not return the int instead and use two negative numbers to represent true and false?

diesieben07: Can you not return the int instead and use two negative numbers to represent true and false?

ferriarnus: Of course that also works! I'm really not paying attention.

patches/minecraft/net/minecraft/world/inventory/GrindstoneMenu.java.patch
ferriarnus: I've checked forge server vanilla client, the items would go in the grindstone no problem. I've tried to join a vanilla world from lan using forge, but that did not work (I couldn't join). I'll need to setup a vanilla server to test this?

SizableShrimp: If you remove the --uuid 0 arg from the forge run config then you should be able to join LAN stuff

ferriarnus: Thanks for the help. I don't see any obvious issues. When forge is the server any item can go in the slot no issue. If vanilla is the server the item won't go in (if you press really fast it will go in for a fraction of a second, but it's hard to notice).

ghost:
I just wanted to throw my hat into the ring and strongly encourage the forge team to add these events into the game, mods should be able to control what happens to their content in vanilla containers.

src/test/java/net/minecraftforge/debug/GrindstoneEventTest.java
```diff
@@ -0,0 +1,59 @@
+package net.minecraftforge.debug;
```
TheCurle, line 1: License, please.

Matyrobbrt:
This has been effectively superseded by #9090.

PR #8583, opened 2022-04-24
[1.18.x] Fix ForgeHooks.canEntityDestroy causing lag if a pos is not loaded

SilverDavidMC:
When `ForgeHooks.canEntityDestroy` is called, it does not do a check for if the position is actually loaded. This can cause a significant amount of server lag for entities that call this regularly, such as zombified piglins and zombies searching for turtle eggs at the default position (0, 0, 0), which has no guarantee of being loaded.

Screenshot comparing TPS in The End at (0, 0, 10,000) with a zombie searching for turtle eggs.

Note: The lag not happen in vanilla since they don't run `level.getBlockState` during the gamerule check in `RemoveBlockGoal`, but Forge does since `ForgeHooks.canEntityDestroy` has additional checks.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -1161,6 +1161,8 @@ public class ForgeHooks
 
     public static boolean canEntityDestroy(Level level, BlockPos pos, LivingEntity entity)
     {
+        if (!level.isLoaded(pos))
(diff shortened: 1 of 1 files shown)


TheCurle:
I remember encountering this issue myself, but i was never sure of an effective fix.

Turns out i've been overthinking it this whole time.

Fix confirmed, thanks for the contribution SilverDavidMC.

PR #8584, opened 2022-04-24
Allow translation keys to be set for config folder/level comments, and provide access to those comments and keys

Phylogeny:
Under the old config annotation system, back when Forge had an in-built config GUI, folder/level/category (whatever you want to call them) names would render as localized buttons, and their comments would render as localized tooltips for those buttons. Under the new config builder system, comments for folders can be set, but are only stored internally in `ForgeConfigSpec#levelComments` with no getter, and folder translation keys are disallowed (`BuilderContext#ensureEmpty` will throw an exception upon calling `Builder#push` if the key is non-null).

This PR allows config GUI mods (like _Configured_) to replicate previous functionality by modifying `ForgeConfigSpec` to treat translation keys like it does comments. Parity between `ValueSpec` and `ForgeConfigSpec` regarding comments and translation keys is established, in that `ValueSpec` has `getComment`/`getTranslationKey`, returning the value (or null if nonexistent), and `ForgeConfigSpec` now has `getLevelComment(path)`/`getLevelTranslationKey(path)`, returning the same for a given folder/level path. I've modified _Configured_ to use these changes to render localized comments as tooltips, and it works fine.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
+++ b/src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
@@ -55,17 +55,27 @@ import com.google.common.collect.ObjectArrays;
  */
 public class ForgeConfigSpec extends UnmodifiableConfigWrapper<UnmodifiableConfig> implements IConfigSpec<ForgeConfigSpec>//TODO: Remove extends and pipe everything through getSpec/getValues?
 {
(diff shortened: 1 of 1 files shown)


TheCurle:
A simple addition, bringing back a highly requested feature.

Thanks for the contribution, Phylogeny.

PR #8588, opened 2022-04-26
Fix entity parts being ignored when collecting entities in an AABB

Meldexun:
Currently entity parts are only taken into consideration when the parent entity is in a chunk that intersects with the AABB. This PR fixes that issue.

This issue is part of [MC-158205](https://bugs.mojang.com/browse/MC-158205).

Another solution to this problem would be to reimplement `MAX_ENTITY_RADIUS` and replace the hardcoded `2.0D` in the `forEachAccessibleNonEmptySection` method in the `EntitySectionStorage` class. But this has two problems:
1. Performance: The ender dragon is the largest entity with a width of 16 which means passing small AABB to `forEachAccessibleNonEmptySection` will result in at least 3x3x3=27 sections being checked.
2. This only works when the bounding box of the parent entity contains all part entities at all time. This is definitely not the case for all modded multipart entities.

### DIFF
--- a/patches/minecraft/net/minecraft/client/multiplayer/ClientLevel.java.patch
+++ b/patches/minecraft/net/minecraft/client/multiplayer/ClientLevel.java.patch
@@ -1,5 +1,13 @@
 --- a/net/minecraft/client/multiplayer/ClientLevel.java
 +++ b/net/minecraft/client/multiplayer/ClientLevel.java
+@@ -117,6 +_,7 @@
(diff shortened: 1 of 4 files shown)


TheCurle:
A nice, simple bug fix. 
Maintainance seems a little expensive, but i've seen first hand the issues this can cause, so i'm willing to make that trade-off.

Thanks for the contribution, Meldexun.

PR #8590, opened 2022-04-27
Allow Exception passed through CommandEvent to propagate

LXGaming:
I want to throw a `CommandSyntaxException` through the `CommandEvent` so that the client is notified of the error, currently Forge is ignoring it as it's a checked exception (only unchecked exceptions are being re-thrown). Because the surrounding try-catch is handling `Exception` I've added a re-throw for it as it can/should be handled instead of being ignored.

Example Event Handler:
```java
@SubscribeEvent
public void onCommand(CommandEvent event) {
    event.setException(CommandSyntaxException.BUILT_IN_EXCEPTIONS.dispatcherUnknownCommand().createWithContext(event.getParseResults().getReader()));
    event.setCanceled(true);
}
```

### DIFF
--- a/patches/minecraft/net/minecraft/commands/Commands.java.patch
+++ b/patches/minecraft/net/minecraft/commands/Commands.java.patch
@@ -17,7 +17,7 @@
  
        this.f_82090_.findAmbiguities((p_82108_, p_82109_, p_82110_, p_82111_) -> {
           f_82089_.warn("Ambiguity between arguments {} and {} with inputs: {}", this.f_82090_.getPath(p_82109_), this.f_82090_.getPath(p_82110_), p_82111_);
-@@ -220,7 +_,15 @@
+@@ -220,7 +_,17 @@
  
        try {
           try {
@@ -25,7 +25,9 @@
 +            ParseResults<CommandSourceStack> parse = this.f_82090_.parse(stringreader, p_82118_);
 +            net.minecraftforge.event.CommandEvent event = new net.minecraftforge.event.CommandEvent(parse);
 +            if (net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(event)) {
-+               if (event.getException() != null) {
++               if (event.getException() instanceof Exception) {
++                  throw (Exception) event.getException();
++               } else if (event.getException() != null) {
 +                  com.google.common.base.Throwables.throwIfUnchecked(event.getException());
 +               }
 +               return 1;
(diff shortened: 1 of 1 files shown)


patches/minecraft/net/minecraft/commands/Commands.java.patch
```diff
@@ -17,15 +17,17 @@
...
  
        try {
           try {
 -            return this.f_82090_.execute(stringreader, p_82118_);
 +            ParseResults<CommandSourceStack> parse = this.f_82090_.parse(stringreader, p_82118_);
 +            net.minecraftforge.event.CommandEvent event = new net.minecraftforge.event.CommandEvent(parse);
 +            if (net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(event)) {
-+               if (event.getException() != null) {
++               if (event.getException() instanceof Exception) {
++                  throw (Exception) event.getException();
```
sciwhiz12: This can use a pattern variable in the `instanceof` instead:
```suggestion
+               if (event.getException() instanceof Exception exception) {
+                  throw exception;
```

sciwhiz12 (approved):
This looks fine to me, as this technically fixes an issue where a non-unchecked exception in the cancelled event would be ignored entirely, rather than the expected behavior of the exception being propagated to the exception handler below.

patches/minecraft/net/minecraft/commands/Commands.java.patch
```diff
@@ -17,15 +17,17 @@
...
           try {
 -            return this.f_82090_.execute(stringreader, p_82118_);
 +            ParseResults<CommandSourceStack> parse = this.f_82090_.parse(stringreader, p_82118_);
 +            net.minecraftforge.event.CommandEvent event = new net.minecraftforge.event.CommandEvent(parse);
 +            if (net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(event)) {
-+               if (event.getException() != null) {
++               if (event.getException() instanceof Exception exception) {
++                  throw exception;
++               } else if (event.getException() != null) {
 +                  com.google.common.base.Throwables.throwIfUnchecked(event.getException());
```
diesieben07, line 31: The instanceof check is not necessary. From the documentation of `throwIfUnchecked`, the expected usage would be:
```java
if (event.getException() != null) {
    Throwables.throwIfUnchecked(event.getException());
    throw new RuntimeException(event.getException());
}
```

diesieben07 (changes requested):

sciwhiz12 (approved):

LexManos:
I'm gunna pull this, however, we should mark it as a TODO for 1.19 to depreciate the getException part of the event. As it's  a  [old event](https://github.com/MinecraftForge/MinecraftForge/commit/d9423a4f25d1e431220979af00105048eb9e7039).
If we implement it as `if (LambdaExceptionUtils.rethrowSupplier(() -> MinecraftForge.EVENT_BUS.post(event)).get()) return 1;` then us manually handleing any exceptions is not needed.

PR #8596, opened 2022-05-01
Expose ICondition$IContext to modded reload listeners.

Shadows-of-Fire:
Currently the ConditionContext is only available for recipes.  This change makes it available to all reload listeners, so modded reload listeners can perform actions that require this information (such as loading conditions with support for `forge:tag_empty`).

### DIFF
--- a/patches/minecraft/net/minecraft/server/ReloadableServerResources.java.patch
+++ b/patches/minecraft/net/minecraft/server/ReloadableServerResources.java.patch
@@ -44,21 +44,20 @@
     }
  
     private static <T> void m_206870_(RegistryAccess p_206871_, TagManager.LoadResult<T> p_206872_) {
-@@ -96,5 +_,17 @@
+@@ -96,5 +_,16 @@
           return p_206874_.getValue().m_6497_();
        }));
        p_206871_.m_175515_(resourcekey).m_203652_(map);
 +   }
 +
-+   private final net.minecraftforge.common.crafting.conditions.ConditionContext context;
++   private final net.minecraftforge.common.crafting.conditions.ICondition.IContext context;
 +
 +   /**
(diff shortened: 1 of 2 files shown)


patches/minecraft/net/minecraft/server/ReloadableServerResources.java.patch
```diff
@@ -44,3 +44,21 @@
...
++
++   private final net.minecraftforge.common.crafting.conditions.ConditionContext context;
++
++   /**
++    * Exposes the currently staged TagManager wrapped in an IContext so that mods can use<br>
++    * tags in their custom reload listeners.<br>
++    * This is only useful for reload listeners who need the currently staged tag values.<br>
++    * @return A ConditionContext with access to the current TagManager.
++    */
++   public net.minecraftforge.common.crafting.conditions.ConditionContext getTagContext() {
```
KnightMiner: The whole reason the interface was named `IContext` and not `ITagContext` was so we could add things besides tags to it in the future. Call this `getConditionContext`

KnightMiner:
For the record, docs could still mention tags are included in the context (as long as they get updated when something besides tags is added). My goal with the rename suggestion as so we don't have to either have a poor name or break modder code once something else is added to context

Technici4n (approved):

Shadows-of-Fire:
I think explicitly noting what the context supports in a javadoc (aside from the class-level javadoc for IContext) is likely to not be updated and cause confusion.  The interface is self-documenting enough by looking at the available methods.

I might consider adding a util method to `AddReloadListenerEvent` which directly returns the `IContext` (as opposed to making modders call `#getServerResources()#getConditionContext()`

TheCurle:
Nice addition. Will help a lot with boilerplate code that's required in Reload Listeners.

Ready to be pulled once TeamCity gives it green.

Thanks for the contribution, Shadows.

PR #8597, opened 2022-05-01
Fix shears passing on a client when interacting with a shearable entity

KnightMiner:
Forge replaces the default shear logic with logic that supports `IShearableEntity`, how ever the current code causes it to pass on the client. This leads to both hands interacting with the entity, as the main hand passing falls back to the client. To see this bug with just forge, put shears in your main hand, and a renamed name tag in your offhand. Shearing a sheep will also immediately rename it.

Fix is pretty straight forward, just return success on the client if the entity is shearable (so move the check after the instance of check)


Shadows-of-Fire:
Looks good to me, since the code path always returns success if the instanceof passes, the early success return is fine.

PR #8598, opened 2022-05-01
Fix EntityInteraction event running twice on client side

KnightMiner:
`MultiplayerGameMode#interact(Player,Entity,InteractionHand)` fired the `EntityInteract` event, but right after doing so calls `Player#interactOn(Entity,InteractionHand)` which then fires `EntityInteract` event a second time on client. Since most listeners cancel the event to set a new result, this bug went unnoticed, but its inefficient for all mods to run their logic to check they cannot interact twice.

Easiest way to test this is just adding a breakpoint to the forge event consturctor, see it gets constructed on the render thread twice, and after the fix on the render thread just once.


TheCurle:
Bug confirmed. 
This is a relatively small performance concern, but this is somehow easier to maintain than the alternative, so let's go with it.

PR #8599, opened 2022-05-03
Implement PlayerNegotiationEvent

LXGaming:
I'm currently attempting to add Forge support to LuckPerms (https://github.com/LuckPerms/LuckPerms/pull/3262) however there are some shortcomings in Forge that @lucko would rather me address directly in Forge instead of attempting to work around them.

The earliest event currently available is the `PlayerLoggedInEvent`, however, this is too late for LuckPerms as block/claim protection mods (GriefDefender / GriefPrevention) or other mods that wish to block player actions upon logging in, won't be able to perform permission checks until LuckPerms has finished loading the user data. To reduce the likelihood of login timeouts on slow connections and unnecessary stalls on the Server Thread (affecting TPS), LuckPerms must load the user data as early as possible, due to the various supported storage providers taking an undetermined amount of time to load.

This PR adds the `PlayerNegotiationEvent` which is posted on the first tick of [`HandshakeHandler.tickServer`](https://github.com/LXGaming/MinecraftForge/blob/efccafe3a9c74efae2d9fc56df1b2c8699a74c02/src/main/java/net/minecraftforge/network/HandshakeHandler.java#L295). This event facilitates the ability for mods to enqueue work that must be completed before the login can proceed.

This functionality already exists in the following platforms and providing it here in Forge would allow plugins traditionally targeting these platforms, such as LuckPerms, to exist on Forge:
- Bukkit [AsyncPlayerPreLoginEvent](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/browse/src/main/java/org/bukkit/event/player/AsyncPlayerPreLoginEvent.java)
- Fabric [LoginSynchronizer](https://github.com/FabricMC/fabric/blob/1d2d03ecd5a4508d14d4bf0647324afef1b8d2bd/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/api/networking/v1/ServerLoginNetworking.java#L159) and [Impl](https://github.com/FabricMC/fabric/blob/1d2d03ecd5a4508d14d4bf0647324afef1b8d2bd/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/server/ServerLoginNetworkAddon.java#L70-L111)
- Sponge [ServerSideConnectionEvent.Auth](https://github.com/SpongePowered/SpongeAPI/blob/44f1788068a046618636856858cab8d30dc4d257/src/main/java/org/spongepowered/api/event/network/ServerSideConnectionEvent.java#L86), [Hook](https://github.com/SpongePowered/Sponge/blob/api-8/src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerLoginPacketListenerImpl_1Mixin.java) and [Impl](https://github.com/SpongePowered/Sponge/blob/47d863402e6145daca63d6f426c313d3a45bfbcf/src/mixins/java/org/spongepowered/common/mixin/core/server/network/ServerLoginPacketListenerImplMixin.java#L200)
- Velocity [PreLoginEvent](https://github.com/PaperMC/Velocity/blob/aa38d3e561862d92c1a62f296cfdb6cdf983871a/api/src/main/java/com/velocitypowered/api/event/connection/PreLoginEvent.java#L32)

### DIFF
--- a/src/main/java/net/minecraftforge/event/entity/player/PlayerNegotiationEvent.java
+++ b/src/main/java/net/minecraftforge/event/entity/player/PlayerNegotiationEvent.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright (c) Forge Development LLC and contributors
+ * SPDX-License-Identifier: LGPL-2.1-only
+ */
+
+package net.minecraftforge.event.entity.player;
+
+import com.mojang.authlib.GameProfile;
+import net.minecraft.network.Connection;
+import net.minecraftforge.eventbus.api.Event;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
+
+/**
+ * This event is fired when a connection has started the Forge Handshake.<br>
+ * Forge will wait for all enqueued work to be completed before proceeding further with the login process.
+ */
+public class PlayerNegotiationEvent extends Event
+{
+
+    private final Connection connection;
+    private final GameProfile profile;
+    private final List<Future<Void>> futures;
+
+    public PlayerNegotiationEvent(Connection connection, GameProfile profile, List<Future<Void>> futures)
+    {
+        this.connection = connection;
+        this.profile = profile;
+        this.futures = futures;
+    }
+
+    /**
+     * Enqueue work to be completed asynchronously before the login proceeds.
+     */
+    public void enqueueWork(Runnable runnable)
+    {
+        enqueueWork(CompletableFuture.runAsync(runnable));
+    }
+
+    /**
+     * Enqueue work to be completed asynchronously before the login proceeds.
+     */
+    public void enqueueWork(Future<Void> future)
+    {
+        futures.add(future);
+    }
+
+    public Connection getConnection()
+    {
+        return connection;
+    }
+
+    public GameProfile getProfile()
+    {
+        return profile;
+    }
+}
--- a/src/main/java/net/minecraftforge/event/entity/player/PlayerNegotiationEvent.java
+++ b/src/main/java/net/minecraftforge/event/entity/player/PlayerNegotiationEvent.java
@@ -7,6 +7,7 @@ package net.minecraftforge.event.entity.player;
 
 import com.mojang.authlib.GameProfile;
 import net.minecraft.network.Connection;
+import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.eventbus.api.Event;
 
 import java.util.List;
@@ -14,8 +15,12 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Future;
 
 /**
- * This event is fired when a connection has started the Forge Handshake.<br>
+ * This event is fired on the server when a connection has started the Forge handshake,
  * Forge will wait for all enqueued work to be completed before proceeding further with the login process.
+ * <br>
+ * This event can be used to delay the player login until any necessary work such as preloading user data has completed.
+ * <br>
+ * This event is fired on the {@link MinecraftForge#EVENT_BUS}.
  */
 public class PlayerNegotiationEvent extends Event
 {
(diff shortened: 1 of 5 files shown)


src/main/java/net/minecraftforge/network/HandshakeHandler.java
```diff
@@ -99,6 +105,8 @@ static boolean tickLogin(Connection networkManager)
     private Map<ResourceLocation, ForgeRegistry.Snapshot> registrySnapshots;
     private Set<ResourceLocation> registriesToReceive;
     private Map<ResourceLocation, String> registryHashes;
+    private boolean negotiationStarted;
```
TheCurle: It seems like this field is never initialized.

src/main/java/net/minecraftforge/event/entity/player/PlayerNegotiationEvent.java
```diff
@@ -0,0 +1,59 @@
...
+import net.minecraftforge.eventbus.api.Event;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
+
+/**
+ * This event is fired when a connection has started the Forge Handshake.<br>
+ * Forge will wait for all enqueued work to be completed before proceeding further with the login process.
+ */
```
TheCurle, line 24: Some more documentation here would be great.
What side does this event fire on? Which bus?

Document the use-case here too, please.

TheCurle (commented):
Seems like a simple, easy addition.
I have some documentation concerns, please take a look:

src/test/java/net/minecraftforge/debug/entity/player/PlayerNegotiationEventTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import net.minecraftforge.event.entity.player.PlayerNegotiationEvent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+@Mod("player_negotiation_event_test")
+@Mod.EventBusSubscriber()
+public class PlayerNegotiationEventTest
```
sciwhiz12, line 20: Please provide a javadoc for this class which explains what feature this tests, how to test it, and what the expected results are from said test.

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/network/HandshakeHandler.java
LXGaming: The field is implicitly false, however I can explicitly set it if you'd like?

TheCurle: That would be much better for reading at a glance, yeah.

src/main/java/net/minecraftforge/event/entity/player/PlayerNegotiationEvent.java
LXGaming: I've added the requested documentation

src/test/java/net/minecraftforge/debug/entity/player/PlayerNegotiationEventTest.java
LXGaming: I've added the requested documentation

TheCurle:
Perfect! I'm going to take another pass over this tomorrow and perform some of my own testing as per our internal guidelines.

Other than that, the code looks good.

joaoespanhol:
> Perfect! I'm going to take another pass over this tomorrow and perform some of my own testing as per our internal guidelines.
> 
> Other than that, the code looks good.

how have feedback?

TheCurle:
Whoops! This slipped past me when i was reviewing contributions the last time.

It's all good on my end. Can't think of anything to change, or any improvements to make.

joaoespanhol:
> Whoops! This slipped past me when i was reviewing contributions the last time.
> 
> It's all good on my end. Can't think of anything to change, or any improvements to make.

why no using same code for 1.16.5 for using .

TheCurle:
That's simply not how code works.

joaoespanhol:
> That's simply not how code works.
but the idea would be the same, you can base the code made for 1.18.2 of this function for 1.16.5

TheCurle:
Locking discussion on this issue since it's not productive any more.

PR #8600, opened 2022-05-03
Shears played no breaking sound after mob interaction

cech12:
This is a small bugfix which affects the issue that no breaking sound was played when breaking shears by shearing an entity (sheep, mooshroom, snow golem). 

You can test it when you give you shears with only one durability

    /give @s minecraft:shears{Damage:237}

and rightclicking with them on a sheep. (in survival mode)

The shears are breaking without a breaking sound. 

In the `interactLivingEntity` of the `ShearsItem` the second parameter of the `hurtAndBreak` call was wrong. I changed it from the sheared entity to the acting player. After this, a breaking sound is played. :)


TheCurle:
Oops, looks like this one is my fault.

Thanks for the contribution, cech12.

PR #8601, opened 2022-05-03
[1.18.x] Catch and aggregate exceptions during NewRegistryEvent

SizableShrimp:
Custom forge registries which throw exceptions during creation will cause `NewRegistryEvent#fill` to prematurely exit and fail to create any registries that come after the failed one.
This issue is compounded by the fact that the fill method is handled from a CompletableFuture, where FML's ModLoadingState system unfortunately silences exceptions.

A similar approach is taken to #8527 with the fix to ObjectHolderRegistry.
This fix both properly logs exceptions and will continue to fill registries even if there is an errored one, to prevent a cascade of dozens of registries not being created due to one erroring registry creation.

### DIFF
--- a/src/main/java/net/minecraftforge/registries/NewRegistryEvent.java
+++ b/src/main/java/net/minecraftforge/registries/NewRegistryEvent.java
@@ -5,10 +5,12 @@
 
 package net.minecraftforge.registries;
 
+import com.mojang.logging.LogUtils;
 import net.minecraft.core.Registry;
 import net.minecraftforge.eventbus.api.Event;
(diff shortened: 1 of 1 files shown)


marchermans (approved):

marchermans:
This works and indeed follows the original precedent set to fix this.
@SizableShrimp What is the actuall swallower? FML or ML?

SizableShrimp:
It's FML and its CompletableFuture state provider stuff

marchermans:
Ahhh right the state providers....

SizableShrimp:
Yes, also know as *hell*

PR #8603, opened 2022-05-05
[1.18.x] Add RenderLevelStageEvent to expand on and replace RenderLevelLastEvent

SilverDavidMC:
The goal of this PR is to introduce more times to render things into the level as a direct replacement for `RenderLevelLastEvent`, fixing issues with Fabulous graphics and allowing more specific render timings.

# Problems with RenderLevelLastEvent

The primary issue with this event is that it happens too late to work with Fabulous graphics, resulting in many renders simply not working at all. Simply moving it to a time that works with Fabulous graphics would technically solve the issue, but due to there being many times that someone may want to render something, having a variety would be far more more useful.

# Implementation

The main issues above can be resolved with this PR by allowing modders to render things into the level at more specific times that better fit their use case. Some of the Stages (listed below) are rendered during times that allow them to work with Fabulous graphics, resolving that issue and giving fine-tuned control over when the rendering happens. This also enables modders to render things with specific parts of the game, such as an overlay onto the skybox.

## RenderLevelStageEvent:

A replacement for `RenderLevelLastEvent` that functions nearly identical. This event fires for every `RenderLevelLastEvent.Stage`, so verifying that you're rendering during the appropriate `Stage` is necessary.

## RenderLevelStageEvent.Stage:

A time when rendering should occur. Custom stages can be created with `RenderLevelStageEvent.RegisterStageEvent` for scenarios where they may be needed, such as if a PR like #8331 goes through. As it stands, the most likely use case for a custom `Stage` would be to render during an existing `Stage` after some amount of setup.

Stages:
- `AFTER_SKY`: After attempting to render the skybox. Use this to render custom objects into the sky.
- `AFTER_PARTICLES`: This is a sort of "catch all" and would likely be the most useful since it happens after blocks, particles, and entities. Works with Fabulous graphics. Rendering translucent blocks during this is intended to be done with `ForgeRenderTypes.TRANSLUCENT_ON_PARTICLES_TARGET`, also added in this PR.
- `AFTER_WEATHER`: Mainly useful for custom weather effects. Works with Fabulous graphics.
- Stages during `LevelRenderer.renderChunkLayer` that correspond to their matching `RenderType` of the same name. Each of these would be used to render custom geometry into the world.
    - `AFTER_SOLID_BLOCKS`
    - `AFTER_CUTOUT_MIPPED_BLOCKS_BLOCKS`
    - `AFTER_CUTOUT_BLOCKS`
    - `AFTER_TRANSLUCENT_BLOCKS`: Happens after block entities render as well. Called during a fabulous graphics target, but may not work with it due to how Minecraft handles transparency sorting. `AFTER_TRIPWIRE_BLOCKS` and `AFTER_PARTICLES` are more reliable for transparency.
    - `AFTER_TRIPWIRE_BLOCKS`

## ForgeHooksClient.partialTicks:

- Gives access to the current partial ticks used by the `LevelRenderer` at the start of `LevelRenderer.renderLevel`.
- Mainly for rendering during Stages where partial ticks are normally not accessible.

# Maintaining

The bulk of maintenance would be in the `LevelRenderer` patch since this is a somewhat patch-heavy PR, making sure the various `RenderLevelStageEvent` calls happen when they should and pass the required data for rendering.

`ForgeRenderTypes.TRANSLUCENT_ON_PARTICLES_TARGET` is based directly off of `RenderType.TRANSLUCENT` and may need updated if that changes.

# Test Mod

This PR reuses the existing `RenderableTest`, modifying it to be able to use `RenderLevelStageEvent`. The changes involve rendering stained glass instead of a diamond block for transparency testing, and rendering at (0, 120, 0). Various dogs are being rendered to test the various Stages, with offsets on the X axis to prevent overlap. The boolean `USE_LEVEL_RENDERER_HOOKS` can be set to false to use the existing `RenderLevelLastEvent` if need be.


TheCurle:
This is a wonderful addition that addresses a very common pitfall and setback that a lot of modders encounter during rendering.

However, I would like to set some expectations for the community; this is a very large, very thorough and very far-reaching addition that will have to be maintained for a long long time to come, so we're going to take our time, pick through this, and request lots and lots of feedback, to verify that this is what the people want, and that people are interested in using this.

On our side, there are currently two triage team members that are capable of reviewing this to its' fullest - me and @gigaherz, the latter is currently having struggles to find time for management like this, so this is going to take a while to fully review and get up to scratch.


With all of that out of the way, I'd like to say; this is wonderful. I'm glad all of this is finally being sorted out properly, and your work seems thorough and well thought through.

With that last point in mind, I'd like to ask what the reasoning behind the hooks requiring a callback is - this means that we need to deal with extra synchronisation due to the parallel nature of ClientSetupEvent, etc. It seems much easier to simply fire a RenderStageEvent with the type as an enum value that can be trivially compared, for eg. a switch statement.

The only other places that callbacks like this are used, are in the event bus itself, and various registerX methods in the vanilla code base, such as entity renderers.

It seems weird to me to go against the established and expected pattern, and instead register render callbacks like this.

SilverDavidMC:
The main reason I have it using a callback system like this was to reduce overhead and only attempt to render something during a specific Phase. Since realistically the "overhead" would be minimal, there's no reason it couldn't be moved to an event bus based system and I'd be glad to make that change if need be. 

Alternatively a proper event could be fired to register these. I forgot about client setup being parallel.

src/main/java/net/minecraftforge/client/LevelRendererHooks.java
```diff
@@ -0,0 +1,118 @@
...
+     * Renders registered hooks for the {@link Phase} passed.
+     */
+    public static void render(Phase phase, LevelRenderer levelRenderer, PoseStack poseStack, Matrix4f projectionMatrix, int ticks, double camX, double camY, double camZ)
+    {
+        List<Consumer<RenderContext>> hooks = HOOKS.get(phase);
+        if (hooks != null)
+        {
+            Minecraft.getInstance().getProfiler().popPush(phase.toString());
+            for (Consumer<RenderContext> hook : hooks)
+                hook.accept(new RenderContext(levelRenderer, poseStack, projectionMatrix, ticks, partialTicks, camX, camY, camZ));
```
PepperCode1: It would be better to create the context once before the loop and pass that object in instead of creating it for every hook. This is of course assuming that `RenderContext` is an immutable structure, which seems like it is.

src/main/java/net/minecraftforge/client/LevelRendererHooks.java
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Renders registered hooks to render during their appropriate {@link Phase}.
+ * Renderer hooks may be registered at any time.
+ * 
+ * @see #register
+ */
+public class LevelRendererHooks
+{
+    private static final HashMap<Phase, List<Consumer<RenderContext>>> HOOKS = new HashMap<>();
```
josephtarbit: `HashMap` -> `Map` on the field types, maybe use a multimap for `HOOKS`.

src/main/java/net/minecraftforge/client/LevelRendererHooks.java
```diff
@@ -0,0 +1,123 @@
...
+        HOOKS.put(phase, hook);
+    }
+
+    /**
+     * Renders registered hooks for the {@link Phase} passed.
+     */
+    public static void render(Phase phase, LevelRenderer levelRenderer, PoseStack poseStack, Matrix4f projectionMatrix, int ticks, double camX, double camY, double camZ)
+    {
+        Collection<Consumer<RenderContext>> hooks = HOOKS.get(phase);
+        if (hooks != null)
```
pupnewfster: I don't believe this ever can be null as the `Multimap` javadocs says
> Note that when containsKey(key) is false, this returns an empty collection, not null.

src/main/java/net/minecraftforge/client/LevelRendererHooks.java
```diff
@@ -0,0 +1,123 @@
...
+    private static final Map<RenderType, Phase> RENDER_TYPE_PHASES = new HashMap<>();
+    public static float partialTicks = 0.0F;
+
+    /**
+     * @see RegisterLevelRendererHooksEvent
+     */
+    public static void register(Phase phase, Consumer<RenderContext> hook)
+    {
+        HOOKS.put(phase, hook);
+    }
```
pupnewfster: Instead of having this be public API and potentially being called at the wrong time if you add a helper method that the javadocs says is internal (or actually if it is only ever called once during startup and not say every time a new world is joined that then errors if called a second time) that fires the `RegisterLevelRendererHooksEvent` and make the `RegisterLevelRendererHooksEvent` take a `Multimap<Phase, Consumer<RenderContext>>` as a parameter. Then the only way people will have to actually register to the map is by doing so at the proper time (in the event) as that is the only place that the `register(Phase, Consumer<RenderContext>)` will be exposed

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -223,6 +223,7 @@ public static boolean onDrawHighlight(LevelRenderer context, Camera camera, HitR
         }
     }
 
+    @Deprecated
```
XFactHD: This should also be marked for removal.

src/main/java/net/minecraftforge/client/event/RenderLevelLastEvent.java
```diff
@@ -10,6 +10,10 @@
 import net.minecraft.client.renderer.LevelRenderer;
 import com.mojang.math.Matrix4f;
 
+/**
+ * @see net.minecraftforge.client.LevelRendererHooks
+ */
+@Deprecated
```
XFactHD: This should also be marked for removal.

src/main/java/net/minecraftforge/client/LevelRendererHooks.java
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Called internally on startup.
+     * 
+     * @see RegisterLevelRendererHooksEvent
+     */
+    public static void registerHooks()
+    {
+        if (hooksMap == null)
+        {
+            hooksMap = ArrayListMultimap.create();
```
XFactHD: Why is the map field not initialized in static init? `registerHooks()` is only called exactly once because the `Minecraft` constructor only runs exactly once.
I don't see a good point in trying to guard against mods being stupid. If you really want to guard against it, then I think a boolean flag is more appropriate and an invalid call to this method should also throw IMO.

src/main/java/net/minecraftforge/client/LevelRendererHooks.java
```diff
@@ -0,0 +1,131 @@
...
+        private Phase(String name)
+        {
+            this.name = name;
+        }
+
+        /**
+         * @param name The name of your Phase. Used for profiling.
+         * @param chunkLayer If your Phase should render during LevelRenderer.renderChunkLayer, supply a RenderType. If not, pass null.
+         */
+        public static Phase create(ResourceLocation name, @Nullable RenderType chunkLayer)
```
XFactHD: I can see how it would be interesting to create a `Phase` for a custom `RenderType` when something like https://github.com/MinecraftForge/MinecraftForge/pull/8331 is merged, but is there any reason for a mod to add a `Phase` that doesn't correspond to a `RenderType`? When and how would such a `Phase` be dispatched?

XFactHD (commented):
Overall I think this is a great addition, especially because it handles Fabulous graphics flawlessly AFAICT from my own tests.

I do think that using events for this instead of registering hooks would be more straightforward though, like Curle said. If a specific event per special phase and a "block layer" event for the phases based on a `RenderType` were used, then non-`RenderType` phases wouldn't need any phase checks, making it very similar to the current "hook registry" approach.

Another thing I am not sure about is what a few of the phases are intended for. It might be a good idea to add some examples for why certain phases should exist.

- `AFTER_SKY`: Could be interesting for adding things like northern lights to the sky without completely overriding sky rendering
- `AFTER_<RenderType>`: Makes sense for rendering block-like things (i.e. block placement previews)
- `AFTER_ENTITIES`, `AFTER_PARTICLES`, `AFTER_CLOUDS`: I can't think of anything for these three, maybe someone else has an idea
- `AFTER_WEATHER`: Could be interesting for adding additional weather effects (i.e. hail) without completely overriding weather rendering and/or weather particle rendering
- `LAST`: I don't really see a point in this, considering the existing `RenderLevelLastEvent` is completely broken in Fabulous mode for anything with translucency

src/main/java/net/minecraftforge/client/LevelRendererHooks.java
SilverDavidMC: A custom `Phase` would be an incredibly niche scenario in general, but ones not fired during a `RenderType` would likely be a "sub-Phase" of sorts, being called by a mod rendering during an existing `Phase` after doing some setup.

SilverDavidMC:
Given some valid points I'm going to change this to fully use a Forge event instead of this registry approach for simplicity. I did a test and the performance is nearly identical. 

`AFTER_PARTICLES` and `AFTER_CLOUDS`, they mainly exist to fire during their respective Fabulous graphics target. `AFTER_ENTITIES` would give you a time to render custom entity-like things into the world, though it may be more fitting to simply use `AFTER_PARTICLES` for that purpose so I will probably end up removing it.
Seeing as `AFTER_WEATHER` would achieve the same goal as `AFTER_CLOUDS` that will also be removed.
And given that `LAST` is completely broken in Fabulous just like the event it's intended to replace, keeping it does seem a bit pointless so it'll get the boot as well.

XFactHD (approved):
There are two dangling empty files (`LevelRendererHooks` and `RegisterLevelRendererHooksEvent`).

I personally preferred calling the different parts of the rendering process "phases" but that's purely subjective.

Apart from that, looks good to me.

SilverDavidMC:
I opted to call them Stages instead since events naturally have a `getPhase()` method for `EventPriority` and I figured this would avoid confusion.

XFactHD:
> I opted to call them Stages instead since events naturally have a `getPhase()` method for `EventPriority` and I figured this would avoid confusion.

Yeah, that makes sense, I completely forgot about that.

src/main/java/net/minecraftforge/client/event/RenderLevelStageEvent.java
```diff
@@ -0,0 +1,145 @@
+package net.minecraftforge.client.event;
```
sciwhiz12, line 6: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -223,11 +223,33 @@ public static boolean onDrawHighlight(LevelRenderer context, Camera camera, HitR
...
     public static void dispatchRenderLast(LevelRenderer context, PoseStack poseStack, float partialTick, Matrix4f projectionMatrix, long finishTimeNano)
     {
         MinecraftForge.EVENT_BUS.post(new RenderLevelLastEvent(context, poseStack, partialTick, projectionMatrix, finishTimeNano));
     }
 
+    /**
+     * The current partial ticks when LevelRenderer.renderLevel is called.
+     * Use this for places where partial ticks are unavailable, but do not modify it.
+     */
+    public static float partialTicks = 0.0F;
```
sciwhiz12: I'd prefer this be named `partialTick` to be consistent with Mojang's naming of `net.minecraft.client.Timer#partialTick` (which is the root of the partial tick values during rendering) and the naming of the field in the event.

src/main/java/net/minecraftforge/client/event/RenderLevelLastEvent.java
```diff
@@ -10,6 +10,10 @@
 import net.minecraft.client.renderer.LevelRenderer;
 import com.mojang.math.Matrix4f;
 
+/**
+ * @see net.minecraftforge.client.event.RenderLevelStageEvent
+ */
```
sciwhiz12: Ideally, there should be a `@deprecated` tag explaining what the closest alternative is (using `{@link}`) and why. For example, `Use {@link RenderLevelStageEvent} for more flexibility for when to render additional content, allowing better compatibility with translucency in certain graphics modes. There is no explicit replacement render stage for this event; a decision must be taken on the appropriate stage for rendering.`

The corresponding method in `ForgeHooksClient` does not need a javadoc of the same; the deprecation annotation will suffice there.

src/main/java/net/minecraftforge/client/event/RenderLevelStageEvent.java
```diff
@@ -0,0 +1,145 @@
...
+ */
+public class RenderLevelStageEvent extends Event
+{
+    public final Stage stage;
+    public final LevelRenderer levelRenderer;
+    public final PoseStack poseStack;
+    public final Matrix4f projectionMatrix;
+    public final int ticks;
+    public final float partialTick;
+    public final double camX, camY, camZ;
```
sciwhiz12: Why `public` fields and not getters?

src/main/java/net/minecraftforge/client/event/RenderLevelStageEvent.java
```diff
@@ -0,0 +1,145 @@
...
+ * Fires at various times during LevelRenderer.renderLevel. 
+ * Check {@link #stage} to render during the appropriate time for your use case.
+ */
+public class RenderLevelStageEvent extends Event
+{
+    public final Stage stage;
+    public final LevelRenderer levelRenderer;
+    public final PoseStack poseStack;
+    public final Matrix4f projectionMatrix;
+    public final int ticks;
```
sciwhiz12: What does `ticks` represent? The name is unclear on this, so it should be clarified in the name or in a javadoc comment.

src/main/java/net/minecraftforge/client/event/RenderLevelStageEvent.java
```diff
@@ -0,0 +1,145 @@
...
+        this.camX = camX;
+        this.camY = camY;
+        this.camZ = camZ;
+    }
+
+    /**
+     * Use to create a custom {@link RenderLevelStageEvent.Stage}s.
+     * Fired after the LevelRenderer has been created.
+     */
+    public static class RegisterStageEvent extends Event implements IModBusEvent
```
sciwhiz12, line 108: I'm a bit unsure on putting a mod-bus event inside a non-mod-bus event, because of the potential confusion by modders. I won't say this event should be moved, but at least it should be noted in the javadoc.

src/main/java/net/minecraftforge/client/event/RenderLevelStageEvent.java
```diff
@@ -0,0 +1,145 @@
...
+
+        private Stage(String name)
+        {
+            this.name = name;
+        }
+
+        private static Stage register(ResourceLocation name, @Nullable RenderType renderType)
+        {
+            Stage stage = new Stage(name.toString());
+            if (renderType != null && RENDER_TYPE_STAGES.put(renderType, stage) != null)
```
sciwhiz12: Use [`Map#putIfAbsent`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Map.html#putIfAbsent(K,V)) instead of [`Map#put`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Map.html#put(K,V)), as `putIfAbsent` only modifies the map if they key had no value previously, while `put` modifies the map regardless of whether or not the key had a value previously.

src/main/java/net/minecraftforge/client/event/RenderLevelStageEvent.java
```diff
@@ -0,0 +1,145 @@
...
+    public static class RegisterStageEvent extends Event implements IModBusEvent
+    {
+        /**
+         * @param name The name of your Stage.
+         * @param renderType
+         *     If not null, called automatically by LevelRenderer.renderChunkLayer if the RenderType passed into it matches this one.
+         *     If null, needs to be called manually by whoever implements it.
+         * 
+         */
+        public Stage register(ResourceLocation name, @Nullable RenderType renderType)
```
sciwhiz12: The javadocs should note the `IllegalArgumentException` thrown by `Stage#register` and the conditions thereof.

src/test/java/net/minecraftforge/debug/client/rendering/RenderableTest.java
```diff
@@ -30,19 +36,23 @@
...
 import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
 import net.minecraftforge.fml.loading.FMLEnvironment;
 
 import java.util.Map;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
 @Mod(RenderableTest.MODID)
 public class RenderableTest
```
sciwhiz12, line 61: Please provide a javadoc for this class which explains what feature this tests, how to test it, and what the expected results are from said test. 

As you are only modifying an existing test mod, you need only include what is relevant to this PR, with an note about there being other behaviors tested by this mod which aren't documented yet.

src/main/java/net/minecraftforge/client/event/RenderLevelStageEvent.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        @Override
+        public String toString()
+        {
+            return this.name;
+        }
+
+        /**
+         * @return The Stage bound to the RenderType passed or null if no value is present.
```
sciwhiz12: We can use the [`{@return}` inline javadoc tag](https://docs.oracle.com/en/java/javase/17/docs/specs/javadoc/doc-comment-spec.html#return) here:
```suggestion
         * {@return the Stage bound to the RenderType, or null if no value is present}
```

src/main/java/net/minecraftforge/client/event/RenderLevelStageEvent.java
```diff
@@ -0,0 +1,145 @@
...
+        {
+            Stage stage = new Stage(name.toString());
+            if (renderType != null && RENDER_TYPE_STAGES.put(renderType, stage) != null)
+                throw new IllegalArgumentException("Attempted to replace an existing RenderLevelStageEvent.Stage for a RenderType: Stage = " + stage + ", RenderType = " + renderType);
+            return stage;
+        }
+
+        private static Stage register(String name, @Nullable RenderType renderType)
+        {
+            return register(new ResourceLocation("forge", name), renderType);
```
sciwhiz12: Should these be under `forge` or under `minecraft`?

sciwhiz12 (changes requested):

sciwhiz12 (approved):
This looks fine to me in its current form. I've heard that there is demand for this PR with valid use cases, chiefly the rendering of ghost previews that won't disappear in *Fabulous!* graphics mode or cause issues with translucent objects in Fancy graphics mode. 

The PR works according to my tests, considering the limitations of translucent rendering, so I'll leave the final determinations to the rendering folks at the core team. ~~may they have mercy on this PR~~

amadornes:
Really happy to see this finally happening! However, I do have a couple notes:

What is the _actual_ purpose of `ForgeHooksClient.partialTick` in the current implementation? In your PR description you say it's useful for querying the partial tick time during level render, but we already have `MinecraftForgeClient.partialTick` for that purpose, which is being set earlier in the render chain. Moreover, its practical use case within the PR seems to just be passing the partial tick time during `RenderLevelStageEvent` construction, but all the call sites for the event dispatch have a partial tick time available. Wouldn't it make more sense to just pass it as an argument during the dispatch call?

I'd like to also request a couple of additions to the event signature, namely a `Camera` instance in place of the current camera coordinates, and the `Frustum` instance which can be found in the render level method. They're both readily available at all call sites, and are incredibly useful when drawing custom geometry.

Other than that, I think it looks good! 👍 

Seems like it'll be compatible with my static render types PR, and doesn't overlap with the geometry collection events PR (this one allows you to fire your own render calls, whereas mine just lets you add geometry to the `MultiBufferSource`, both dynamically and during chunk baking).

sciwhiz12 (changes requested):
Requesting changes per @amadornes' comment above.

SilverDavidMC:
The main purpose of storing the partial ticks was that partial ticks are not passed into `LevelRenderer.renderChunkLayer`, but seeing as there's already a place to get partial ticks from `MinecraftForgeClient`, I'll remove that bit and replace it with calling that.

marchermans:
Please update this PR to 1.19.

robotgryphon:
This appears to have been closed, will there be updates now that the 1.19 version is merged?

AterAnimAvis:
Seems like a Github bug, reopening for now, if this was actually meant to be closed then apologies

PR #8607, opened 2022-05-05
Re-introduce the attribute for entity step height in a non-breaking way

pupnewfster:
This PR reimplements #8389 in a non-breaking way so that any mods that have actually updated to making use of the feature can actually work instead of being broken by the then breaking change of having reverted the PR in #8604.
Changes compared to the original PR:
- Validate the entity has the step height attribute in `IFogeEntity#getStepHeight` rather than just assuming it has it
- Slightly reworded the javadocs for `IFogeEntity#getStepHeight` to mention that it only adds it if the entity actually has the attribute
- As it is only one extra line as part of the same hunk: cache the retrieved step height in `Entity#collide` so as to not have to look up the attribute four times
- Modified the spawn egg item test mod to spawn the entity explicitly without the step height attribute for purposes of validating things don't break when an entity doesn't expose the attribute. (This isn't part of its own test mod as it would basically be a copy of the entire thing for one small change)

Tests performed:
- Validated that the modified test mod crashes when the implementation of `IFogeEntity#getStepHeight` from #8389 is used
- Validated it doesn't crash with the modified version (this PR)
- Tested behavior when a forge server is updated to have the attribute but the connecting forge client isn't:
  - The result is a message being logged on the client side like this: `Entity Pig['entity.forge_spawnegg_test.test_entity'/457, l='ClientLevel', x=153.50, y=64.00, z=-172.50] does not have attribute forge:step_height_addition`
  - When a player's step height is modified the player doesn't respect the changes (as movement is initially calculated client side), but just requiring mods to bump their network version to ensure both sides of their mod is using a min forge version seems quite reasonable for if they want to make use of it

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/Entity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/Entity.java.patch
@@ -17,6 +17,14 @@
     private final EntityType<?> f_19847_;
     private int f_19848_ = f_19843_.incrementAndGet();
     public boolean f_19850_;
+@@ -175,6 +_,7 @@
+    public double f_19790_;
+    public double f_19791_;
+    public double f_19792_;
++   @Deprecated // Forge - see IForgeEntity#getStepHeight
+    public float f_19793_;
+    public boolean f_19794_;
+    protected final Random f_19796_ = new Random();
 @@ -229,6 +_,7 @@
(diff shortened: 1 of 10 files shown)


TheCurle:
Perfect. Thanks for a more robust implementation.

PR #8609, opened 2022-05-07
Add event for intercepting narration chat messages

sciwhiz12:
This PR adds the `NarrationChatEvent` to allow mods to intercept narration chat messages and replace them as needed for a better accessibility experience. This provides a test mod to check and showcase the functionality of the PR.

## Use cases

My usecase for this PR is my mod, Concord, which adds chat messages based on messages from a connected Discord channel. I wish to replicate the functionality that Minecraft hardcodes for regular chat messages for my own chat messages: changing to a more humanized message by replacing the `chat.type.text` key with the `chat.type.text.narrate` key (which contains a more humanized message "%s says %s" for chat messages). 

Other mods may use this in a similar manner, to replace their chat messages with narration-friendly alternatives for the benefit of users who use the narrator functionality.

## Alternatives

It may be possible for a mod to extend `NarratorChatListener`, override its methods to create one's own hook into the chat message functionality, and replace the instance held in `NarratorChatListener#INSTANCE` as well as `Gui#chatListeners`. However, this is disfavored and error-prone for two main reasons:

- A subclass would need to deal with the additional `Narrator` instance constructed by their superclass' constructor, leading to multiple `Narrator` instances. 
	- This could be mitigated with the use of `Unsafe#allocateInstance` to allocate an instance of the subclass without invoking its constructor (and therefore the superclass constructor), but that seems icky to me.
- A subclass would need to deal with the original singleton instance, due to management of the contained `Narrator` within (calls to `#destroy` and such) and due to the possibility of other objects other than `Gui#chatListeners` storing the narrator instance.

### DIFF
--- a/patches/minecraft/net/minecraft/client/gui/chat/NarratorChatListener.java.patch
+++ b/patches/minecraft/net/minecraft/client/gui/chat/NarratorChatListener.java.patch
@@ -1,0 +1,10 @@
+--- a/net/minecraft/client/gui/chat/NarratorChatListener.java
++++ b/net/minecraft/client/gui/chat/NarratorChatListener.java
+@@ -30,6 +_,7 @@
+             this.m_168787_(p_93323_.getString());
+          } else {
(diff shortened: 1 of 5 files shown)


sciwhiz12:
Marking this as a draft for now until I can investigate the changes in 1.19 to adapt this PR to.

PR #8612, opened 2022-05-08
Add pack-type-specific format versions in pack metadata

### LINKED ISSUE #8563: Single pack format version declaration causing pack incompatibility warnings
**Minecraft Version:** 1.18.2
**Forge Version:** 1.18.2-40.0.44
**Logs:** _not applicable_

## Steps to reproduce

 1. Unzip a fresh copy of the MDK for 1.18.2-40.0.44, and set it up as usual (open in IDE, generate runs).
 2. Insert the following line into `src/main/resources/META-INF/mods.toml`, at the beginning section before the first `[[mods]]` declaration: 
    ```toml
    showAsResourcePack=true
    ```
    This tells FML to show your mod as a separate resource pack, instead of being combined into the global "Mod resources" pack created by FML.
 3. Confirm that the pack format version in `src/main/resources/pack.mcmeta` is set to `9`.
 4. Run the game.
 5. Click on the `Options` button then the `Resource Packs...` button to open the resource pack selection menu.
    - _Observe that the resource pack for `examplemod` is under the `Available` column, is highlighted in red, and shows on hover text declaring the resource pack as incompatible and built for a newer version of Minecraft._
 6. Return to the main menu, click on the `Singleplayer` button (which should show the world creation screen) and then click the `Data Packs` button to open the data pack selection menu.
    - _Observe that the data pack for `examplemod` is under the `Selected` column and is not highlighted in red, unlike the other two data packs present._
 7. Close the game.
 8. In the `src/main/resources/pack.mcmeta` file, change the pack format version from `9` to `8`.
 9. Repeat steps 4 and 5 (running the game and opening the resource pack selection menu in the options screen).
    - _Observe that the resource pack for `examplemod` is still under the `Available` column, but no longer highlighted in read nor showing text on hover declaring its incompatibility._
 10. Repeat step 6 (opening the data pack selection menu in the world creation screen).
    - _Observe that the data pack for `examplemod` is still under the `Selected` column, but is now highlighted in red and show text on hover declaring it as incompatible and built for an older version of Minecraft, like the other two data packs present._

### Background
In the 1.17 snapshot 20w45a, the pack format version was split into data and resource versions. This meant that data packs and resource packs now have concurrent and separate pack formats. Presumably, this allows the Minecraft developers to communicate breaking changes of the formats of either data packs or resource packs without affecting the other. Until 1.18.2, both pack format versions were the same due to coincidence.[^1]

In 1.18.2 Pre-Release 1, this consistency between the two pack format versions was broken: the data pack format version was bumped to `9`, while the resource pack version remained at `8`.

### Problem
This creates a small problem in the world of mods. Mods generally contain both a data pack and a resource pack within them, which share the same `pack.mcmeta` file and therefore the same pack format version declaration. However, because of this disparity between the two pack format versions, **declaring the pack format version to be compatible with the current resource pack format means it will not be compatible with the current data pack format, and vice-versa**.

This is demonstrated above using the data pack and resource pack selection menus (through the options screen and the world creation screen, respectively). When the version matches `9` (the current data pack format version), the resource pack shows as incompatible. When the version matches `8` instead (the current resource pack format version), the data pack shows as incompatible.

As can also be observed in the above steps, the Forge and Minecraft data packs are also marked as incompatible because of this issue. Comparing with a vanilla 1.18.2 installation, the built-in Minecraft resource pack and data pack are neither highlighted in red or marked as incompatible.

### Potential solutions
I'm not fully sure what the best solution here will be. Here's a few options, as food for thought:

 1. **Do nothing.** 🙈
 2. **Separate the locations of the resource and data packs within mod JARs to separate folders.** For example, `resourcepack/` within the JAR contains the mod's resource pack (with its meta file at `resourcepack/pack.mcmeta`), and `datapack/` contains the mod's data pack (with its meta file at `datapack/pack.mcmeta`).
 3. **Modify the pack loading code to search under the `assets`/`data` folder for a `pack.mcmeta` file first.** This change may be limited to only the mod packs (those in the JARs, or folders for the dev env), or applied to all packs (but that may be too a broad scope).
 4. **Modify the pack metadata loading code to add new keys for both pack format versions.** This means the same `pack.mcmeta` may serve both the data pack and resource pack of the mod, with differing pack formats between then. The original `pack_format` key may still be required, and used as a fallback if the corresponding pack type's pack format key is not present.

[^1]: Between 1.17 snapshots 21w37a and 21w39a, there was a temporary disparity between the pack format versions. The former snapshot bumped the data pack format version, and the latter snapshot bumping the resource pack format version, bringing them back to the same number. For our purposes, we will disregard this, but it is neat to note.

sciwhiz12:
This PR fixes #8563 by implementing pack-type-specific format versions in the pack metadata file, `pack.mcmeta`, as suggested by @gigaherz in the linked issue.

For each of Minecraft's `PackType`[^1], a new key is deserialized if present in the pack metadata file which is the pack format version for that pack type. This key is derived from the name of the Mojang `PackType` present in the Minecraft `PackType`, which is lowercased then prepended with `"forge:"` and appended with `"_pack_format"`.

 - For Minecraft's `PackType#CLIENT_RESOURCES` and Mojang's `PackType#RESOURCE`, the key is `forge:resource_pack_format`.
 - For Minecraft's `PackType#SERVER_DATA` and Mojang's `PackType#DATA`, the key is `forge:data_pack_format`.

The original `pack_format` key is still required, and acts as the fallback if the more specific pack-typed format version key is absent.

As an example of use and as a test case for reviewers, Forge's pack metadata files for the main and test source sets are modified to use the new pack-typed format version keys.

---

**NBs:** The change in the last line of `src/main/resources/forge.exc` is the removal of the newline, which is done automatically by the `checkExcs` task.

[^1]: Minecraft's `PackType` is `net.minecraft.server.packs.PackType`, and should not be confused with Mojang's `PackType` at `com.mojang.bridge.game.PackType`.


patches/minecraft/net/minecraft/server/packs/metadata/pack/PackMetadataSection.java.patch
```diff
@@ -0,0 +1,32 @@
+--- a/net/minecraft/server/packs/metadata/pack/PackMetadataSection.java
++++ b/net/minecraft/server/packs/metadata/pack/PackMetadataSection.java
+@@ -6,17 +_,29 @@
+    public static final PackMetadataSectionSerializer f_10366_ = new PackMetadataSectionSerializer();
+    private final Component f_10367_;
+    private final int f_10368_;
++   private final java.util.Map<net.minecraft.server.packs.PackType, Integer> packTypeVersions;
+ 
+    public PackMetadataSection(Component p_10371_, int p_10372_) {
```
marchermans, line 9: Given that this is the original constructor and should not be used anymore, should this not be marked as deprecated?

marchermans (approved):
This looks good, and works. 
One question remaining on the PackType constructor.

patches/minecraft/net/minecraft/server/packs/metadata/pack/PackMetadataSection.java.patch
sciwhiz12: I thought about deprecating it, but decided against it as some may use it to construct pack metadata with the correct pack format for the pack type. Having them specify the empty map manually is unnecessary when this constructor already exists which does the same.

marchermans: Okiedokie.
Good enough reason for me.
I will assign.

PR #8616, opened 2022-05-08
Fix command redirects by replacing Commands#fillUsableCommands

### LINKED ISSUE #7551: Command redirect Issues
**Minecraft Version:**1.16.4

**Forge Version:** 31.0.7

**Steps to Reproduce:**
```java
LiteralCommandNode<CommandSource> command = dispatcher.register(LiteralArgumentBuilder.<CommandSource>literal("test")
            .requires(x -> x.hasPermissionLevel(4))
            .then(Commands.literal("one")
                .executes(ctx -> {
                    ctx.getSource().sendFeedback(new StringTextComponent("Working"), true);
                    return 1;
                })));
        dispatcher.register(Commands.literal("te")
            .redirect(command));
```

**Description of issue:**
With the setup above the command `/test one` is as expected inaccessible to people without permission level 4.
The redirect below however is NOT restricted. So `/te one` is working even without permissions. <- this does not synchronize the client making it very difficult to recognize (if it was intended at all).

I'm unsure if this is a bug in Brigadier or caused by forge's changes to commands. 

EDIT: Also might be related. Redirects to intermediate commandnodes don't sync either.

sciwhiz12:
This PR fixes #7551 by replacing `Commands#fillUsableCommands` with `CommandHelper#mergeCommandNode`, which is `ClientCommandHandler#mergeCommandNode` but moved to a non-client-only class. See my comment on the linked issue for a bit more context on this change: https://github.com/MinecraftForge/MinecraftForge/issues/7551#issuecomment-1030610010.

### DIFF
--- a/patches/minecraft/net/minecraft/commands/Commands.java.patch
+++ b/patches/minecraft/net/minecraft/commands/Commands.java.patch
@@ -36,3 +36,13 @@
           } catch (CommandRuntimeException commandruntimeexception) {
              p_82118_.m_81352_(commandruntimeexception.m_79226_());
              return 0;
+@@ -276,7 +_,8 @@
+       Map<CommandNode<CommandSourceStack>, CommandNode<SharedSuggestionProvider>> map = Maps.newHashMap();
+       RootCommandNode<SharedSuggestionProvider> rootcommandnode = new RootCommandNode<>();
+       map.put(this.f_82090_.getRoot(), rootcommandnode);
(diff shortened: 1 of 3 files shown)


marchermans:
Since this is a vanilla bug, a Mojira issue number will need to be added as a comment for this to be eligible for merging.

TheCurle:
This is not a user facing vanilla bug, so submitting to Mojira is not necessary. I can confirm this is indeed a vanilla bug, though.

PR #8617, opened 2022-05-08
[1.18] Add callback after a BlockState was changed

XFactHD:
This PR adds a callback after a blocks `BlockState` is changed in the world. The method is very similar to `BlockBehaviour#onPlace()` (which is called on the new `BlockState`) with two significant differences:

1. The new method is also called on the client side, while `BlockBehaviour#onPlace()` is only called on the server side
2. The new method is called after neighboring blocks are also updated, which allows taking changes to neighboring blocks into account

`BlockBehaviour#onRemove()` is not entirely comparable to the new method because `BlockBehaviour#onRemove()` is called on the old `BlockState`. Apart from that, both of the above restrictions also apply to that method.

The added method intentionally takes a `LevelReader` instead of a `Level` because it is intended to be used for read-only accesses, mainly because it doesn't include any guards against infinite recursion.

My usecase for this is to update a client-only occlusion cache when blockstates change. This works fine for most blocks when a neighbor changes (handled through `Block#updateShape()`), but with that approach some blocks need to have the re-calculation of that cache deferred to the end of the current tick because the new `BlockState` is placed into the world after `Block#updateShape()` returns (this is the case for any blocks that use said method for it's intended purpose, like walls and fences) and for some blocks this approach doesn't work at all because there is no way to catch a block update on the client side (this is the case for "self-modifying" blocks like doors and trapdoors).


src/test/java/net/minecraftforge/debug/block/BlockStateChangeTest.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+package net.minecraftforge.debug.block;
+
+import com.mojang.logging.LogUtils;
+import net.minecraft.core.BlockPos;
+import net.minecraft.world.InteractionHand;
+import net.minecraft.world.InteractionResult;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.*;
```
sciwhiz12: Please avoid the use of wildcard imports. (There's another one down the list.)

src/test/java/net/minecraftforge/debug/block/BlockStateChangeTest.java
```diff
@@ -0,0 +1,86 @@
...
+import net.minecraft.world.level.material.Material;
+import net.minecraft.world.phys.BlockHitResult;
+import net.minecraftforge.eventbus.api.IEventBus;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.registries.*;
+import org.slf4j.Logger;
+
+@Mod(BlockStateChangeTest.MOD_ID)
+public class BlockStateChangeTest
```
sciwhiz12, line 45: Please provide a javadoc for this class which explains what feature this tests, how to test it, and what the expected results are from said test.

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -832,4 +832,20 @@ default boolean supportsExternalFaceHiding(BlockState state)
...
+     * Useful for calculating additional data based on the new state and the neighbor's reactions to the state change.
+     *
+     * @param level The level the state was modified in
+     * @param pos The blocks position in the level
+     * @param oldState The previous state of the block at the given position, may be a different block than this one
+     * @param newState The new state of the block at the given position
+     */
+    default void onBlockStateChange(LevelReader level, BlockPos pos, BlockState oldState, BlockState newState)
+    {
+
```
sciwhiz12: The blank space can be omitted.
```suggestion
```

sciwhiz12 (changes requested):
This looks solid and the test works fine (after some mishaps with pistons and the `PistonEventTest`); I just have a few formatting notes.

sciwhiz12 (approved):

LexManos:
Discussed on discord, this hook is not needed as the existing Block.updateIndirectNeighbourShapes should suffice.

sciwhiz12:
Reopening upon request of PR author, see #8686.

sciwhiz12 (approved):
1.19 PR has been merged.

PR #8624, opened 2022-05-14
add an option for OBJ Loder to load .obj file with same part name

zomb-676:
This PR adds the ability for the OBJ Loder to load .obj file with same part name  

# Why need this?

Some modeling software such as BlockBench , which allows homonym cube names .  When its model is exported as an .obj model file . The name subsequents to  .obj format command/keyworld `o` will duplicate

Due to the current OBJ Loader implementation , the collection of the parts is a `Map<String, ModelGroup>` , so the later will override the former. As a reuslt , only the last part will remain

# Implementation

add an option bool for OBJModel.Setting called `allowHomonymObjectName`  which is false by default in order to keep forward compatibility

when the option is true ,  we will check weather the same part name occours before , if it is , we re-name it by adding a positive growth after the origin name until no repeat occour
The similar way as another modeling software called Blender , it can correctly load .obj file with homonym object name

re-name only works on the `o` .obj command/keyword now

# Maybe a problem

```java
    public Optional<? extends IModelGeometryPart> getPart(String name)
    {
        return Optional.ofNullable(parts.get(name));
    }
```

after re-name , the getter method may fail to get the correct named part 
but if your model has homonym part name , you can't expect to get it by this way
you should name the needed part yourself or you could call this method  
 `public Collection<? extends IModelGeometryPart> getParts()`
and filter according to its keys by methods such as startWith to group them

### DIFF
--- a/src/main/java/net/minecraftforge/client/model/obj/OBJLoader.java
+++ b/src/main/java/net/minecraftforge/client/model/obj/OBJLoader.java
@@ -48,10 +48,11 @@ public class OBJLoader implements IModelLoader<OBJModel>
         boolean diffuseLighting = GsonHelper.getAsBoolean(modelContents, "diffuseLighting", false);
         boolean flipV = GsonHelper.getAsBoolean(modelContents, "flip-v", false);
         boolean ambientToFullbright = GsonHelper.getAsBoolean(modelContents, "ambientToFullbright", true);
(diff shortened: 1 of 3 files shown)


sciwhiz12:
Hello! @zomb-676, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.19.2 (or leave an explanation why this PR's changes are not applicable to 1.19.2). If we do not hear back from you after some time, this PR will be closed for inactivity.

PR #8625, opened 2022-05-15
[1.18.x] Fix 2 crossbow animations not applying to custom variants

Autovw:
This PR aims to fix 2 visual bugs with custom crossbows introduced by the 1.17.x update:

* Custom crossbow disappears of the screen while recharging (first person)
* Custom charged crossbow is not displayed with the ``hold`` armpose (third person)

Currently, the most common approach to fix this by modders is with the use of coremods.
I've also included a test mod with a data generator as example implementation in this PR.

Here's a picture of the armposes in third person: 
Image


Rainwangpupil (approved):

Thelnfamous1:
Things like this used to be fixed by the field-to-instanceof transformer, probably needs a rework for the new ItemStack#is method.

TheCurle:
Honestly, i think i'd prefer if you skip the test, and just add the instanceof thing. 

Once we fix the transformer it'll be easier to remove all of this.

Autovw:
I've removed the test mod 👍

PR #8629, opened 2022-05-21
[1.18.x] Enable "Experimental Forge Light Pipeline" by default

XFactHD:
This PR enables the "Experimental Forge Light Pipeline" by default, as requested by @sciwhiz12.
The experimental Forge light pipeline is a vital feature for a lot of mods with more intricate block models that are not lit properly with the vanilla light pipeline.

To quote @TheCurle on Forgecord:
> It's been in incubation for a few years, but since it's now fixing more problems than it's causing, it's now default on for 1.19. [...] And will probably be made default on for 1.18 in a few days.

### DIFF
--- a/src/main/java/net/minecraftforge/client/model/pipeline/ForgeBlockModelRenderer.java
+++ b/src/main/java/net/minecraftforge/client/model/pipeline/ForgeBlockModelRenderer.java
@@ -39,7 +39,7 @@ public class ForgeBlockModelRenderer extends ModelBlockRenderer
     @Override
     public boolean tesselateWithoutAO(BlockAndTintGetter level, BakedModel model, BlockState state, BlockPos pos, PoseStack poseStack, VertexConsumer buffer, boolean checkSides, Random rand, long seed, int packedOverlay, IModelData modelData)
     {
-        if(ForgeConfig.CLIENT.experimentalForgeLightPipelineEnabled.get())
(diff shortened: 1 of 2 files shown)


sciwhiz12 (approved):
As quoted, @TheCurle already signalled her approval on this.

pupnewfster:
This isn't the correct fix. The reason that there even is an experimental pipeline config is so that when it becomes default instead of not being able to update old instances to the new default (due to the configs already being generated) we switch from using the `experimentalForgeLightPipelineEnabled` to the `forgeLightPipelineEnabled` which upon closer inspection seems to have been removed in the port to 1.17 and likely should be added back with the experimental config option being deprecated for removal in 1.19

XFactHD:
If that is the preferred way to fix this, then I'd be happy to reintroduce that setting (blame seems to indicate it was lost in the 1.13 update)

pupnewfster:
That is because the lighting pipeline didn't survive the 1.12 -> higher port. But when it got added back both options got added. 

And it definitely is [still there](https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/src/main/java/net/minecraftforge/common/ForgeConfig.java#L171-L178) in 1.16.

XFactHD:
Re-introduced the non-experimental setting, moved all usages to this setting and deprecated the experimental setting as requested by @pupnewfster.

pupnewfster (approved):

PR #8630, opened 2022-05-21
1.18.x Custom datapack registries, round three-ish

Commoble:
This is a fork of #8522, Silverminer's custom datapack registries PR, with some improvements to documentation and error presentation and streamlining of userdev APIs.

## Datapack Registries, what are they and why might modders want them
Datapack registries (sometimes called dynamic registries or worldgen registries) are a system added by Mojang several major versions ago, neither static registries nor reloadable data, but a third system separate from the other two. Datapack registries are created by registering a loading codec and (optionally) a network codec (if the network codec exists, we consider the registry to be a Synced Datapack Registry). Examples of vanilla datapack registries include biomes, dimensiontypes, and placed features.

Features of datapack registries (relative to reloadable data)
* Datapack registries have a simplified standard implementation for loading and optionally syncing (which cannot be changed, as opposed to reloadable data which provides lower-level access to data)
* Datapack registries have half-builtin support for datagen, using their loading codecs and RegistryOps
* Datapack registries cannot be reloaded with /reload, they can only be reloaded by rebooting the server
* Datapack registries automatically support tag jsons
* Unsynced datapack registries' elements can directly reference any other datapack registries' elements

Lifecycle of datapack registries:
* The datapack registry and its codec(s) are registered (custom datapacks can be registered via the NewRegistryEvent or DeferredRegisters)
* "Builtin" objects can optionally be registered in java (via DeferredRegister/RegistryObjects). This is required to datagen objects using RegistryOps.
* When the server starts, using the registries' loading codecs, builtin objects are deep-copied and pasted into the server's personal RegistryAccess, and then json data from datapacks is loaded and pasted into the server's RegistryAccess (possibly overriding builtin objects).
* When a client logs in, any synced datapack registries and their contents are sent to the client using the network codecs, and stored in the client's personal RegistryAccess (which is stored in the client's connection). Clients do not have to have any of the server's registrables pre-registered (and any builtin registrables are ignored on the client), but the synced registries themselves are required to have been registered on the client.

## Userdev APIs
When using a RegistryBuilder to register a registry (either via DeferredRegister or NewRegistryEvent), the userdev may invoke `RegistryBuilder#dataPackRegistry(Codec<T> codec, @Nullable Codec<T> networkCodec)`, which marks the registry as a datapack registry and registers codecs for it. An overload that only takes the loading codec is also provided (using `null` for the network codec). If the network codec is non-null, the registry is considered to be a synced registry. Unsynced datapack registries registered on the server do not need to be registered on clients, but synced registries do. RegistryBuilder#disableSync is ignored as datapack registries use the presence of the network codec (or lack thereof) to determine whether to sync the registry.

Registries that are named `modid:folder` cause data to be loaded from `data/<datapack_namespace>/modid/folder/`. The folder path can include `/` for subdirectorying, e.g. `worldgen/biome`.

## Network API Changes
* Bump netversion from 2 to 3.
* The S2CModList packet now may or may not contain a list of synced non-vanilla datapack registry IDs at the end of the packet bytestream.
* Backwards-compatibility is kept for clients on netversion 2, see below. This backwards-compatibility may be dropped in 1.19 (sending an empty list instead of nothing).

## End-User Concerns
If no changes are made to login netcode, then if a client were to attempt to connect to a server when the server has syncable datapack registries that don't exist on the client, the login packet fails to decode and the client is presented with [an unreadable error message](https://cdn.discordapp.com/attachments/852298000042164244/975493141966753842/unknown.png) (due to the decoder dumping the entire contents of all synced datapack registries into the login error screen).

To avoid this, we validate the syncability of custom datapack registries in these ways:
* [The server disallows forge clients on netversion 2 and vanilla clients from joining the server if the server has any non-vanilla syncable datapack registries](https://media.discordapp.net/attachments/852298000042164244/978817467075919893/unknown.png)
* Forge clients on netversion 3 check incoming registries in the mod list packet and terminate their connection if any of the incoming datapack registries don't exist on the client

Forge clients on netversion 2 will experience an "[incompatible server](https://media.discordapp.net/attachments/852298000042164244/978818144665763940/unknown.png)" warning in the server list menu due to the netcode version mismatch, however they can still join the server if the server has no syncable custom datapack registries..

## Summary of Patches
* RegistryAccess has a hook patched into its codec registry so forge can add custom datapack registries' codecs later
* RegistryResourceAccess has a patch to include the registry namespace in the directory (like tags) if namespace isn't "minecraft" and the registry was created via the RegistryBuilder API


src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
```diff
@@ -0,0 +1,71 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import com.google.common.collect.ImmutableMap;
+
+import net.minecraft.core.Registry;
+import net.minecraft.core.RegistryAccess;
+import net.minecraft.resources.ResourceKey;
+import net.minecraft.resources.ResourceLocation;
+
+public class DataPackRegistriesHooks
```
sciwhiz12: Is there anything in here which may be used by modders?

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
```diff
@@ -0,0 +1,71 @@
...
+    private static Map<ResourceKey<? extends Registry<?>>, RegistryAccess.RegistryData<?>> REGISTRY_ACCESS_REGISTRIES_COPY;
+    private static Set<ResourceKey<? extends Registry<?>>> SYNCED_CUSTOM_REGISTRIES = new HashSet<>();
+    private static Set<ResourceKey<? extends Registry<?>>> SYNCED_CUSTOM_REGISTRIES_VIEW = Collections.unmodifiableSet(SYNCED_CUSTOM_REGISTRIES); 
+
+    /** Modders, DO NOT USE. Internal use only */
+    @Deprecated
+    public static Map<ResourceKey<? extends Registry<?>>, RegistryAccess.RegistryData<?>> grabBuiltinRegistries(ImmutableMap.Builder<ResourceKey<? extends Registry<?>>, RegistryAccess.RegistryData<?>> builder)
+    {
+        REGISTRY_ACCESS_REGISTRIES_COPY = new HashMap<>(builder.build());
+        SYNCED_CUSTOM_REGISTRIES = new HashSet<>();
```
sciwhiz12: Why create anew instead of clearing the existing set?

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
```diff
@@ -0,0 +1,71 @@
...
+        ResourceKey<? extends Registry<T>> registryKey = data.key();
+        REGISTRY_ACCESS_REGISTRIES_COPY.put(registryKey, data);
+        if (data.networkCodec() != null)
+        {
+            SYNCED_CUSTOM_REGISTRIES.add(registryKey);
+        }
+    }
+    
+    /**
+     * @return unmodifiable view of the set of synced non-vanilla datapack registry IDs.
```
sciwhiz12: Probably want to use an inline return tag here.
```suggestion
     * {@return unmodifiable view of the set of synced non-vanilla datapack registry IDs}
```

src/main/java/net/minecraftforge/network/NetworkConstants.java
```diff
@@ -20,7 +20,7 @@
 public class NetworkConstants
 {
     public static final String FMLNETMARKER = "FML";
-    public static final int FMLNETVERSION = 2;
+    public static final int FMLNETVERSION = 3;
```
sciwhiz12, line 27: It may be good to add a comment here explaining what necessitated the bump to version 3, for ease of future readers.

src/main/java/net/minecraftforge/network/HandshakeMessages.java
```diff
@@ -83,7 +90,18 @@ public static S2CModList decode(FriendlyByteBuf input)
...
+            // Datapack Registries may or may not be sent in 1.18.2 due to netcode changes.
+            // TODO 1.19: Remove optionalness of datapack registry list in the mod list packet.
+            List<ResourceKey<? extends Registry<?>>> dataPackRegistries = new ArrayList<>();
+            if (input.isReadable())
+            {
+                len = input.readVarInt();
+                for (int x = 0; x < len; x++)
+                {
+                    dataPackRegistries.add(ResourceKey.createRegistryKey(input.readResourceLocation()));
+                }
```
sciwhiz12: Could you instead use `FriendlyByteBuf#readCollection` here (which is the counterpart to `FriendlyByteBuf#writeCollection`)?

src/main/java/net/minecraftforge/network/HandshakeMessages.java
```diff
@@ -99,6 +117,16 @@ public void encode(FriendlyByteBuf output)
...
             registries.forEach(output::writeResourceLocation);
+            
+            // The list of synced datapack registry names is not sent in 1.18.2 if the list is empty.
+            // TODO 1.19: should send an empty list if the list is empty.
+            Set<ResourceKey<? extends Registry<?>>> dataPackRegistries = DataPackRegistriesHooks.getSyncedCustomRegistries();
+            int size = dataPackRegistries.size();
+            if (size > 0)
+            {
+                output.writeVarInt(size);
+                dataPackRegistries.forEach(key -> output.writeResourceLocation(key.location()));
```
sciwhiz12: Could you instead use `FriendlyByteBuf#writeCollection` here (which is the counterpart to `FriendlyByteBuf#readCollection`)?

src/main/java/net/minecraftforge/registries/NewRegistryEvent.java
```diff
@@ -62,8 +64,19 @@ void fill()
                 .collect(Collectors.toMap(RegistryData::builder, d -> d.builder().create(), (a, b) -> a, IdentityHashMap::new));
 
         builtRegistries.forEach((builder, reg) -> {
-            if (builder.getHasWrapper() && !Registry.REGISTRY.containsKey(reg.getRegistryName()))
+            RegistryAccess.RegistryData<?> dataPackRegistryData = builder.getDataPackRegistryData();
+            if (dataPackRegistryData != null) // if this is a datapack registry
+            {
+                if (builder.getHasWrapper() && !BuiltinRegistries.REGISTRY.containsKey(reg.getRegistryName()))
```
sciwhiz12: Is it possible for a datapack registry to reach this part and have `getHasWrapper()` to be `false`? If not, should we remove the check since we 'know' that it is always `true`, move it to something like a `Validate` call (to indicate it should never be `false` but we check it anyway), or keep it as-is as a potential safeguard if _somehow_ something manages to make a datapack registry without a wrapper?

src/main/java/net/minecraftforge/registries/RegistryBuilder.java
```diff
@@ -230,6 +244,57 @@ public RegistryBuilder<T> hasTags()
...
+     * <p>Register this registry as an unsynced datapack registry, which will cause data to be loaded from
+     * a datapack folder based on the registry's name. The mod that registers this registry does not need to exist
+     * on the client to connect to servers with the mod/registry.</p>
+     * <p>The registry's name must be of the form {@code modid:modid/folder}.
+     * Data JSONs will be loaded from {@code data/<datapack_namespace>/modid/folder/}, where modid is the mod that registered this registry.</p>
+     * 
+     * @param codec the codec to be used for loading data from datapacks on servers
+     * @return this builder
+     * 
+     * @see {@link #dataPackRegistry(Codec, Codec)} for registering a synced datapack registry
```
sciwhiz12: According to the documentation for the `@see` tag[^1], when the first character is neither a `"` (for standard text entry) nor `<` (for an external HTML link), then `javadoc` expects the text immediately following `@see` to be a class member reference. In line with this, the use of `{@link}` should be removed.

Furthermore, javadocs take the text after the reference in a `@see` tag as the label for the link[^2], akin to how `<a href="bleh.html">boop</a>` shows "boop" for the link. In line with this, the label should also be removed.

```suggestion
     * @see #dataPackRegistry(Codec, Codec)
```

If you wish to attach prose explanation when linking to the method, you'll need to add it as sa paragraph, though I would think the `@see` tag would be visible enough.

[^1]: [Java 17](https://docs.oracle.com/en/java/javase/17/docs/specs/javadoc/doc-comment-spec.html#see) or [Java 7 (with examples)](https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#see) documentation on the `@see` tag
[^2]: `@see` custom labels are used to change the label from the label auto-generated from the reference.[^1]

src/main/java/net/minecraftforge/registries/RegistryBuilder.java
```diff
@@ -230,6 +244,57 @@ public RegistryBuilder<T> hasTags()
...
+     * 
+     * @param codec the codec to be used for loading data from datapacks on servers
+     * @param networkCodec the codec to be used for syncing loaded data to clients.<br>
+     * If networkCodec is null, data will not be synced, and clients without the mod that registered this registry can
+     * connect to servers with the mod.<br>
+     * If networkCodec is not null, then data will be synced (accessible via {@link ClientPacketListener#registryAccess()}),
+     * and the mod must be present on a client to connect to servers with the mod.
+     * @return this builder
+     * 
+     * @see {@link #dataPackRegistry(Codec)} for registering an unsynced datapack registry
```
sciwhiz12: Same notes as my earlier comment on the other use of `@see`.
```suggestion
     * @see #dataPackRegistry(Codec)
```

src/main/java/net/minecraftforge/registries/RegistryBuilder.java
```diff
@@ -230,6 +244,57 @@ public RegistryBuilder<T> hasTags()
...
+    {
+        this.hasWrapper(); // A wrapper is required for data pack registries.
+        this.disableSync(); // Datapack registries are synced using a different system than static registries.
+        // Supplier averts having to set the registry name before calling this.
+        this.dataPackRegistryData = Suppliers.memoize(() -> {
+            // Validate registry key.
+            if (this.registryName == null)
+                throw new IllegalStateException("Registry builder cannot build a datapack registry: registry name not set");
+            if (!DataPackRegistriesHooks.isNamespacedRegistryFolder(this.registryName))
+                throw new IllegalStateException(String.format(Locale.ENGLISH, "Invalid datapack registry name %s: Registry path must be prefixed with '%s'", registryName, registryName.getNamespace() + '/'));
```
sciwhiz12: Should we document these being thrown in the javadoc (noting also that these are thrown when the registry is actually built, not thrown at this method's invocation)?

src/main/java/net/minecraftforge/registries/RegistryManager.java
```diff
@@ -161,10 +162,21 @@ <V extends IForgeRegistryEntry<V>> ForgeRegistry<V> createRegistry(ResourceLocat
         return getRegistry(name);
     }
 
-    @SuppressWarnings("unchecked")
     static <V extends IForgeRegistryEntry<V>> void registerToRootRegistry(ForgeRegistry<V> forgeReg)
     {
-        WritableRegistry<Registry<V>> registry = (WritableRegistry<Registry<V>>) Registry.REGISTRY;
+        injectForgeRegistry(forgeReg, Registry.REGISTRY);
+    }
+
```
sciwhiz12, line 169: Formatting: there should be at most one blank line between lines with content.
```suggestion
```

src/main/java/net/minecraftforge/server/ServerLifecycleHooks.java
```diff
@@ -147,9 +151,22 @@ public static boolean handleServerLogin(final ClientIntentionPacket packet, fina
...
-                rejectConnection(manager, connectionType, "This modded server is not impl compatible with your modded client. Please verify your Forge version closely matches the server. Got net version "+ versionNumber + " this server is net version "+ NetworkConstants.FMLNETVERSION);
-                return false;
+            if (connectionType == ConnectionType.MODDED)
+            {
+                // Allow clients with incorrect netcode version to connect to netversion 3 servers,
+                // if and only if client is netversion 2 and server has no syncable non-vanilla datapack registries.
+                // TODO 1.19: Remove netcode backwards-compatability in 1.19, as there will be no clients on netversion 2 in 1.19. 
+                Set<ResourceKey<? extends Registry<?>>> customDatapackRegistries = DataPackRegistriesHooks.getSyncedCustomRegistries();
+                if (versionNumber == 2) {
+                    if (!customDatapackRegistries.isEmpty()) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
                if (versionNumber == 2)
                {
                    if (!customDatapackRegistries.isEmpty())
                    {
```

src/main/java/net/minecraftforge/server/ServerLifecycleHooks.java
```diff
@@ -147,9 +151,22 @@ public static boolean handleServerLogin(final ClientIntentionPacket packet, fina
...
+                        rejectConnection(manager, connectionType, "Missing required datapack registries: " + String.join(", ", customDatapackRegistries.stream().map(key -> key.location().toString()).toList()));
+                        return false;
+                    }
+                }
+                else if (versionNumber != NetworkConstants.FMLNETVERSION)  {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
                else if (versionNumber != NetworkConstants.FMLNETVERSION)
                {
```

src/test/java/net/minecraftforge/debug/world/DataPackRegistriesTest.java
```diff
@@ -0,0 +1,309 @@
...
+import net.minecraftforge.fml.common.Mod.EventBusSubscriber;
+import net.minecraftforge.fml.common.Mod.EventBusSubscriber.Bus;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistryEntry;
+import net.minecraftforge.registries.RegistryBuilder;
+import net.minecraftforge.registries.RegistryObject;
+
+@Mod(DataPackRegistriesTest.MODID)
+public class DataPackRegistriesTest
```
sciwhiz12: Please provide a javadoc for this class which explains what feature this tests, how to test it, and what the expected results are from said test.

src/test/java/net/minecraftforge/debug/world/DataPackRegistriesTest.java
```diff
@@ -0,0 +1,309 @@
...
+        {
+            return testA;
+        }
+
+        public HolderSet<TestA> getTestAList()
+        {
+            return testAList;
+        }
+    }
+    public static class TestC extends ForgeRegistryEntry<TestC>
```
sciwhiz12: Given the consistency in spacing between other members in this file, there should also be a blank line between these two.
```suggestion
    }

    public static class TestC extends ForgeRegistryEntry<TestC>
```

sciwhiz12 (changes requested):
I've not tested the PR yet, but the logic seems good to me and the conversations regarding it in the Discord server implies that this works. (Still needs testing, though)

src/main/java/net/minecraftforge/network/HandshakeMessages.java
Commoble: That's a neat function! Will change in next push

src/main/java/net/minecraftforge/network/HandshakeMessages.java
Commoble: Changed in next push

src/main/java/net/minecraftforge/network/NetworkConstants.java
Commoble: Added in next push

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
Commoble: Changed in next push

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
Commoble: I keep forgetting maps and sets can be cleared, changed in next push

src/main/java/net/minecraftforge/registries/NewRegistryEvent.java
Commoble: Yeah, the registry builder automatically creates wrappers for all datapack registries, so we can remove the check

src/main/java/net/minecraftforge/registries/RegistryBuilder.java
Commoble: Changed in next push

src/main/java/net/minecraftforge/registries/RegistryBuilder.java
Commoble: Changed in next push

src/main/java/net/minecraftforge/registries/RegistryBuilder.java
Commoble: Changed in next push

src/main/java/net/minecraftforge/registries/RegistryBuilder.java
Commoble: Changed in next push

src/main/java/net/minecraftforge/registries/RegistryBuilder.java
Commoble: We can add a throws doc in getDataPackRegistryData, since that actually will throw on invocation if registry name hasn't been set yet (my understanding is that all registries need to be named eventually anyway so saying "make sure to call setRegistryName eventually" here is probably redundant).

src/main/java/net/minecraftforge/registries/RegistryManager.java
Commoble: Fixed in next push

src/main/java/net/minecraftforge/server/ServerLifecycleHooks.java
Commoble: Changed in next push

src/main/java/net/minecraftforge/server/ServerLifecycleHooks.java
Commoble: Changed in next push

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
Commoble: Only the view getter is safe to use, and I'm not if there's any use case to use it in userdev -- I can mark the whole class as forge-internal, in any case

src/test/java/net/minecraftforge/debug/world/DataPackRegistriesTest.java
Commoble: Rewrote test class for next push

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
```diff
@@ -49,22 +49,12 @@ public static <T> void addRegistryCodec(@NotNull RegistryAccess.RegistryData<T>
     }
     
     /**
-     * @return unmodifiable view of the set of synced non-vanilla datapack registry IDs.
+     * {@return unmodifiable view of the set of synced non-vanilla datapack registry IDs.
      * Clients must have each of a server's synced datapack registries to be able to connect to that server;
-     * vanilla clients therefore cannot connect if this list is non-empty on the server.
+     * vanilla clients therefore cannot connect if this list is non-empty on the server.}
```
sciwhiz12: The scope of the inline `{@return}` tag should only extend to the first sentence, as that will be taken as the first sentence of the javadoc and as the contents of the `Returns:` line.
```java
    /**
     * {@return unmodifiable view of the set of synced non-vanilla datapack registry IDs}
     * Clients must have each of a server's synced datapack registries to be able to connect to that server;
     * vanilla clients therefore cannot connect if this list is non-empty on the server.
     */
```

src/test/java/net/minecraftforge/debug/misc/DataPackRegistriesTest.java
```diff
@@ -0,0 +1,228 @@
...
+                () -> new RegistryBuilder<Syncable>().disableSaving().dataPackRegistry(Syncable.DIRECT_CODEC, Syncable.DIRECT_CODEC));
+            
+            // Datapack registry elements can be datagenerated, but they must be registered as builtin objects first.
+            this.datagenTestObject = unsyncables.register("datagen_test", () -> new Unsyncable("Datagen Success"));
+            
+            unsyncables.register(modBus);
+            syncables.register(modBus);
+            
+            modBus.addListener(this::onGatherData);
+            forgeBus.addListener(this::onServerStarting);
```
sciwhiz12: Formatting: there should be no extraneous trailing whitespace.
```suggestion
            forgeBus.addListener(this::onServerStarting);
```

src/test/java/net/minecraftforge/debug/misc/DataPackRegistriesTest.java
```diff
@@ -0,0 +1,228 @@
...
+    public static final String MODID = "data_pack_registries_test";
+    private static final Logger LOGGER = LogUtils.getLogger();
+    private static final ResourceLocation TEST_RL = new ResourceLocation(MODID, "test");
+    
+    private final RegistryObject<Unsyncable> datagenTestObject;
+
+    public DataPackRegistriesTest()
+    {
+        if (ENABLED)
+        {
```
sciwhiz12: It might be better to invert this into a `if (!ENABLED) return;`, to reduce the indentation.

src/test/java/net/minecraftforge/debug/misc/DataPackRegistriesTest.java
```diff
@@ -0,0 +1,228 @@
...
+            // but only one mod can register the internal registry with makeRegistry.
+            final DeferredRegister<Unsyncable> unsyncables = DeferredRegister.create(Unsyncable.REGISTRY_KEY, MODID);
+            final DeferredRegister<Syncable> syncables = DeferredRegister.create(Syncable.REGISTRY_KEY, MODID);
+            
+            // RegistryBuilder#dataPackRegistry marks the registry as a datapack registry rather than a static registry.
+            unsyncables.makeRegistry(Unsyncable.class,
+                () -> new RegistryBuilder<Unsyncable>().disableSaving().dataPackRegistry(Unsyncable.DIRECT_CODEC));
+            // The overload of #dataPackRegistry that takes a second codec marks the datapack registry as syncable.
+            syncables.makeRegistry(Syncable.class,
+                () -> new RegistryBuilder<Syncable>().disableSaving().dataPackRegistry(Syncable.DIRECT_CODEC, Syncable.DIRECT_CODEC));
```
sciwhiz12: What are the cases that a data pack registry registers a network codec that is different from the main codec?

src/test/java/net/minecraftforge/debug/misc/DataPackRegistriesTest.java
```diff
@@ -0,0 +1,228 @@
...
+            {
+                Unsyncable.DIRECT_CODEC.encodeStart(ops, element)
+                    .resultOrPartial(msg -> LOGGER.error("Failed to encode {}: {}", path, msg)) // log error on encode failure
+                    .ifPresent(json -> // output to file on encode success
+                    {
+                        try
+                        {
+                            DataProvider.save(gson, cache, json, path);
+                        }
+                        catch(IOException e) // we're inside this ifpresent so the throws can't deal with this
```
sciwhiz12: Formatting: space after `catch` keyword.
```suggestion
                        catch (IOException e) // we're inside this ifpresent so the throws can't deal with this
```
Also, perhaps you mean `as` ("as the throws can't deal with this") and not `so`? Unless I'm interpreting the comment's intention incorrectly, the reason the `catch` exists is because we cannot propagate the `IOException` upwards normally because of the use of a lambda, so `as` would be more semantically correct here.

sciwhiz12 (changes requested):
I tried to test this PR using the provided test mod, but I got an error while trying to load a singleplayer world in `DataPackRegistriesTest.ClientEvents#onClientTagsUpdated` because the `player` is `null` (and the attempt to read the `connection` field fails). I've fixed it in my local copy by getting the `RegistryAccess` from `TagsUpdatedEvent#getTagRegistry` instead.

---

> * Datapack registries cannot be reloaded with /reload, they can only be reloaded by rebooting the server
> 
> _[...]_
>
> Lifecycle of datapack registries:
>
> _[...]_
> 
> * When the server starts, using the registries' loading codecs, builtin objects are deep-copied and pasted into the server's personal RegistryAccess, and then json data from datapacks is loaded and pasted into the server's RegistryAccess (possibly overriding builtin objects).

To clarify for me, this means that only the JSON data from datapacks which affect a datapack registry that are present during server startup will affect the contents of the datapack registry, even if said JSON data is removed later on and a `/reload` is issued?

src/test/java/net/minecraftforge/debug/misc/DataPackRegistriesTest.java
SizableShrimp: See vanilla usages. Stuff that is only ever needed/used/processed by the server is just wasted networking data if sent through the main codec over networking.

Commoble: There also exist codecs that don't work at all when used for networking because they're hardcoded to be used by RegistryOps and the netcode doesn't use RegistryOps.

Commoble:
> To clarify for me, this means that only the JSON data from datapacks which affect a datapack registry that are present during server startup will affect the contents of the datapack registry, even if said JSON data is removed later on and a /reload is issued?

Yes, after the first load of datapack registries during server startup, the datapack registry jsons are never read again until the server reboots.

> I tried to test this PR using the provided test mod, but I got an error while trying to load a singleplayer world in DataPackRegistriesTest.ClientEvents#onClientTagsUpdated because the player is null (and the attempt to read the connection field fails). I've fixed it in my local copy by getting the RegistryAccess from TagsUpdatedEvent#getTagRegistry instead.

I really want to make sure the player's connection's registryaccess gets populated, as that's what clientside mods should be querying during server runtime -- I'll try moving the test to a different event again.

src/test/java/net/minecraftforge/debug/misc/DataPackRegistriesTest.java
Commoble: You're interpreting it correctly, but I don't think that would be a correct use of "as" either. I'll use "because" for minimum ambiguity -- "The throws can't deal with this exception, because we're inside this ifPresent."

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
Commoble: Fixed in next iteration

src/test/java/net/minecraftforge/debug/misc/DataPackRegistriesTest.java
Commoble: Fixed in next iteration

src/test/java/net/minecraftforge/debug/misc/DataPackRegistriesTest.java
Commoble: Fixed in next iteration

sciwhiz12 (approved):
The test works now without changes. :shipit: 

---

As a side-note, during testing of this PR, I noticed that the `/forge tags` command doesn't suggest dynamic/datapack registries, but works fine when manually entered. 

With help from @Commoble, the cause and solution was found. The cause was because the registry suggestor used `Registry.REGISTRY`, which only contains the static registries and excludes the dynamic registries. As a fix, one can replace `TagsCommand#suggestRegistries` with the following snippet:

```java
private static CompletableFuture<Suggestions> suggestRegistries(final CommandContext<CommandSourceStack> ctx,
        final SuggestionsBuilder builder)
{
    ctx.getSource().registryAccess().registries()
            .map(RegistryAccess.RegistryEntry::key)
            .map(ResourceKey::location)
            .map(ResourceLocation::toString)
            .forEach(builder::suggest);
    return builder.buildFuture();
}
```

This will be fixed soon, either through a PR from me to 1.18 or in the 1.19 branch.

PR #8636, opened 2022-05-28
Improve TagsUpdatedEvent

diesieben07:
- Add UpdateCause information. This tells the user what caused the event.
- Add hint for when static (shared in SP) data should be refreshed. This is now used in Forge to update the burn time map, which previously would refresh twice in SP.

### DIFF
--- a/src/main/java/net/minecraftforge/event/TagsUpdatedEvent.java
+++ b/src/main/java/net/minecraftforge/event/TagsUpdatedEvent.java
@@ -6,15 +6,33 @@
 package net.minecraftforge.event;
 
 import net.minecraft.core.RegistryAccess;
+import net.minecraft.network.protocol.game.ClientGamePacketListener;
 import net.minecraftforge.eventbus.api.Event;
+import net.minecraftforge.server.ServerLifecycleHooks;
 
+/**
+ * Fired when tags are updated on either server or client. This event can be used to refresh data that depends on tags.
+ */
 public class TagsUpdatedEvent extends Event
 {
     private final RegistryAccess registries;
+    private final UpdateCause updateCause;
+    private final boolean integratedServer;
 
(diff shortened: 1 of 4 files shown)


src/main/java/net/minecraftforge/event/TagsUpdatedEvent.java
```diff
@@ -24,4 +42,39 @@ public RegistryAccess getTagManager()
...
+         * on the client thread.
+         */
+        SERVER_DATA_LOAD,
+        /**
+         * The tag update is caused by the client receiving the tag data from the server.
+         */
+        CLIENT_PACKET_RECEIVED
+    }
+
+
```
sciwhiz12: Formatting: empty lines should be removed.

src/main/java/net/minecraftforge/event/TagsUpdatedEvent.java
```diff
@@ -6,15 +6,33 @@
...
+         var currentServer = ServerLifecycleHooks.getCurrentServer();
+         this.integratedServer = currentServer != null && !currentServer.isDedicatedServer();
+     }
+
+     public TagsUpdatedEvent(RegistryAccess registries, boolean fromClientPacket, boolean isIntegratedServerConnection)
+     {
+         this.registries = registries;
+         this.updateCause = fromClientPacket ? UpdateCause.CLIENT_PACKET_RECEIVED : UpdateCause.SERVER_DATA_LOAD;
+         this.integratedServer = isIntegratedServerConnection;
      }
```
sciwhiz12: Since you're already modifying the constructor, could you please correct the indentation from 5 spaces to 4 spaces?

sciwhiz12 (changes requested):

sciwhiz12 (approved):

Technici4n (approved):

PR #8640, opened 2022-05-29
Fix attack cooldown bar showing improperly

### LINKED ISSUE #8639: Reduced ATTACK_RANGE - Cooldown Bar below Crosshair shows when out of range
**Minecraft Version:** 1.18.2

**Forge Version:** 1.18.2-40.1.21

**Logs:** N/A

**Steps to Reproduce:**
 1. Create a new class that either extends an item that implements `Attributes.ATTACK_SPEED` and `Attributes.ATTACK_DAMAGE` or implements these itself. Override `IForgeItem#getAttributeModifiers` and make sure `ForgeMod.ATTACK_RANGE` is implemented into the `Multimap<Attribute, AttributeModifier>` which the method returns. The `ATTACK_RANGE` attribute take negative value to reduce the attack range.
 2. Register the item and launch the game.
 3. Enter a world. Go into Survival Mode. Place down an entity (armor stands are very convenient for testing this). Give yourself the item you just added and mouse over the entity you placed, ideally at the range where the cooldown bar below the crosshair is about to disappear.
 4. Attack the entity. No damage is dealt (correct behavior!), however the cooldown bar still shows, indicating that an attack should be possible.

**Description of issue:**
When using the new `ATTACK_RANGE` attribute forge added recently to reduce a weapon's attack reach, the cooldown bar just under the crosshair still shows, even when out of range and no damage can actually be dealt. This bar typically indicates when something is hittable, as it does if the `ATTACK_RANGE` is increased. If the player in creative mode, the cooldown bar shows correctly given the reduction in range depending on how much it is reduced. If it is reduced by more than 3.0, the cooldown bar will still show when out of range. 

Interestingly, implementing `ForgeMod.REACH_DISTANCE` with a negative value that is proportional to the negative value passed into `ATTACK_RANGE` seems to be able to make the cooldown bar under the crosshair go away when out of range. Even so, I think it would be more logical for `ATTACK_RANGE` to control this behavior, as the cooldown bar doesn't even show when someone wants to, say, open a chest or shear a sheep, which is what `REACH_DISTANCE` ought to control as I understand it.

I believe this happens in `GameRenderer#pick` when `double d0 = (double)this.minecraft.gameMode.getPickRange();` and `this.minecraft.crosshairPickEntity = entity1;`, as in `Gui#renderCrosshair` crosshairPickEntity is checked when rendering the crosshair and cooldown bar.

Here is the code I used to produce this issue: https://gist.github.com/HIPPOKLEIDES/6d1099f0953030a661e8e44099dbd4b8

Shadows-of-Fire:
Closes #8639

Fixes the attack cooldown bar showing even when the attack range is reduced and the entity cannot be attacked.
Also has minor javadoc updates so modders find the correct methods for interfacing with these attributes.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeMod.java
+++ b/src/main/java/net/minecraftforge/common/ForgeMod.java
@@ -95,20 +95,23 @@ public class ForgeMod
     public static final RegistryObject<Attribute> ENTITY_GRAVITY = ATTRIBUTES.register("entity_gravity", () -> new RangedAttribute("forge.entity_gravity", 0.08D, -8.0D, 8.0D).setSyncable(true));
 
     /**
-     * Reach Distance represents the distance at which a player may interact with the world.  The default is 4.5 blocks.  Players in creative mode have an additional 0.5 blocks of reach distance.<br>
-     * See {@link IForgePlayer#getReachDistance} and {@link IForgePlayer#canInteractWith}
+     * Reach Distance represents the distance at which a player may interact with the world.  The default is 4.5 blocks.  Players in creative mode have an additional 0.5 blocks of reach distance.
+     * @see IForgePlayer#getReachDistance()
+     * @see IForgePlayer#canInteractWith(BlockPos, double)
+     * @see IForgePlayer#canInteractWith(Entity, double)
      */
     public static final RegistryObject<Attribute> REACH_DISTANCE = ATTRIBUTES.register("reach_distance", () -> new RangedAttribute("generic.reachDistance", 4.5D, 0.0D, 1024.0D).setSyncable(true));
 
     /**
-     * Attack Range represents the distance at which a player may attack an entity.  The default is 3 blocks.  Players in creative mode have an additional 3 blocks of attack reach.<br>
-     * See {@link IForgePlayer#getAttackRange} and {@link IForgePlayer#canHit}
+     * Attack Range represents the distance at which a player may attack an entity.  The default is 3 blocks.  Players in creative mode have an additional 3 blocks of attack reach.
+     * @see IForgePlayer#getAttackRange()
+     * @see IForgePlayer#canHit(Entity, double)
      */
     public static final RegistryObject<Attribute> ATTACK_RANGE = ATTRIBUTES.register("attack_range", () -> new RangedAttribute("generic.attack_range", 3.0D, 0.0D, 1024.0D).setSyncable(true));
 
     /**
-     * Step Height Addition modifies the amount of blocks an entity may walk up without jumping.<br>
-     * See {@link IForgeEntity#getStepHeight}
+     * Step Height Addition modifies the amount of blocks an entity may walk up without jumping.
+     * @see IForgeEntity#getStepHeight()
      */
     public static final RegistryObject<Attribute> STEP_HEIGHT_ADDITION = ATTRIBUTES.register("step_height_addition", () -> new RangedAttribute("forge.stepHeight", 0.0D, -512.0D, 512.0D).setSyncable(true));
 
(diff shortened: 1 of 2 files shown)


src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -93,14 +95,21 @@
     public static final RegistryObject<Attribute> ENTITY_GRAVITY = ATTRIBUTES.register("entity_gravity", () -> new RangedAttribute("forge.entity_gravity", 0.08D, -8.0D, 8.0D).setSyncable(true));
 
     /**
-     * Reach Distance represents the distance at which a player may interact with the world.  The default is 4.5 blocks.  Players in creative mode have an additional 0.5 blocks of reach distance.
+     * Reach Distance represents the distance at which a player may interact with the world.  The default is 4.5 blocks.  Players in creative mode have an additional 0.5 blocks of reach distance.<br>
+     * See {@link IForgePlayer#getReachDistance} and {@link IForgePlayer#canInteractWith}
```
sciwhiz12: Use the `@see` tag instead of `{@link}` for linking to other members outside of inline in a sentence. (You may also need to qualify the method references with the parameter types.) Also, remove the trailing `<br>`.
```suggestion
     * @see IForgePlayer#getReachDistance
     * @see IForgePlayer#canInteractWith
```

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -93,14 +95,21 @@
...
      */
     public static final RegistryObject<Attribute> REACH_DISTANCE = ATTRIBUTES.register("reach_distance", () -> new RangedAttribute("generic.reachDistance", 4.5D, 0.0D, 1024.0D).setSyncable(true));
 
     /**
-     * Attack Range represents the distance at which a player may attack an entity.  The default is 3 blocks.  Players in creative mode have an additional 3 blocks of attack reach.
+     * Attack Range represents the distance at which a player may attack an entity.  The default is 3 blocks.  Players in creative mode have an additional 3 blocks of attack reach.<br>
+     * See {@link IForgePlayer#getAttackRange} and {@link IForgePlayer#canHit}
```
sciwhiz12: Use the `@see` tag instead of `{@link}` for linking to other members outside of inline in a sentence. (You may also need to qualify the method references with the parameter types.) Also, remove the trailing `<br>`.
```suggestion
     * @see IForgePlayer#getAttackRange
     * @see IForgePlayer#canHit
```

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -93,14 +95,21 @@
...
      */
     public static final RegistryObject<Attribute> ATTACK_RANGE = ATTRIBUTES.register("attack_range", () -> new RangedAttribute("generic.attack_range", 3.0D, 0.0D, 1024.0D).setSyncable(true));
+
+    /**
+     * Step Height Addition modifies the amount of blocks an entity may walk up without jumping.<br>
+     * See {@link IForgeEntity#getStepHeight}
```
sciwhiz12: Use the `@see` tag instead of `{@link}` for linking to other members outside of inline in a sentence. (You may also need to qualify the method references with the parameter types.) Also, remove the trailing `<br>`.
```suggestion
     * @see IForgeEntity#getStepHeight
```

sciwhiz12 (changes requested):
Tested and confirmed the fix works (after a mishap of not realizing I was in creative mode). I just have some changes needed for the javadoc changes.

Shadows-of-Fire:
Done, didn't realize the see tag was a thing.

sciwhiz12 (approved):

PR #8644, opened 2022-06-01
[1.18.x] Add checkJarCompatibility task

SizableShrimp:
Utilizes the newest version of ForgeGradle, `5.1.39`, to implement a `checkJarCompatibility` test.
By default, this will use the value of `LAST_RB` as the base Forge version/jar. If there is no LAST_RB, then it will default to null and attempting to run the `checkJarCompatibility` task will skip.
If there are errors, they will be logged to the gradle console and the task will fail due to a non-zero exit value.
If there are warnings, they will still be logged to the gradle console but not count towards the error count / exit value.
This can be used both by PR authors to check their feature/patches do not break compatibility, and also can be added as an additional check during PRs (cc @OrionDevelopment).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'net.minecraftforge.gradle:ForgeGradle:[5.1.26,)'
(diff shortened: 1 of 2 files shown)


TheCurle:
Taking a dive here, I trust you but if this blows up it'll be on your head.

PR #8648, opened 2022-06-03
[1.18] Allow custom rarities to directly modify the style

### LINKED ISSUE #8647: [RFC][1.18.x] Rarity does not support colors outside of ChatFormatting
It would be a really nice Quality of Life feature to change it so Rarity can use Hex or Int colors instead of just being limited to ChatFormatting

diesieben07:
Fixes #8647.
A test mod is included that tests the new version and sets the style to italic in addition to a color.

### DIFF
--- a/patches/minecraft/net/minecraft/world/item/Rarity.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/Rarity.java.patch
@@ -1,6 +1,6 @@
 --- a/net/minecraft/world/item/Rarity.java
 +++ b/net/minecraft/world/item/Rarity.java
-@@ -2,15 +_,35 @@
+@@ -2,15 +_,36 @@
  
  import net.minecraft.ChatFormatting;
  
@@ -11,7 +11,8 @@
     RARE(ChatFormatting.AQUA),
     EPIC(ChatFormatting.LIGHT_PURPLE);
  
-+   @Deprecated // Forge: Use getStyleModifier
++   /** @deprecated Forge: Use IForgeRarity#applyStyle */
++   @Deprecated
     public final ChatFormatting f_43022_;
 +   private final java.util.function.UnaryOperator<net.minecraft.network.chat.Style> styleModifier;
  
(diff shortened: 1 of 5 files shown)


patches/minecraft/net/minecraft/world/item/Rarity.java.patch
```diff
@@ -9,13 +9,31 @@
...
+    public final ChatFormatting f_43022_;
++   private final java.util.function.UnaryOperator<net.minecraft.network.chat.Style> styleModifier;
  
     private Rarity(ChatFormatting p_43028_) {
        this.f_43022_ = p_43028_;
++      this.styleModifier = style -> style.m_131140_(p_43028_);
++   }
++
++   Rarity(java.util.function.UnaryOperator<net.minecraft.network.chat.Style> styleModifier) {
++      this.f_43022_ = ChatFormatting.BLACK;
```
Elenterius, line 25: i would change this to ChatFormatting.RESET this way anyone who uses it to get the color via ChatFormatting#getColor() will get null

diesieben07: Which they might not be prepared for at all.

Elenterius: Well the return type of `ChatFormatting#getColor()` is a nullable Integer object, so they always had to be prepared for it to be null 😆 

Anyways i don't really care, `ChatFormatting.BLACK` is acceptable

patches/minecraft/net/minecraft/world/item/Rarity.java.patch
```diff
@@ -9,13 +9,31 @@
     COMMON(ChatFormatting.WHITE),
     UNCOMMON(ChatFormatting.YELLOW),
     RARE(ChatFormatting.AQUA),
-@@ -12,5 +_,9 @@
+    EPIC(ChatFormatting.LIGHT_PURPLE);
+ 
++   @Deprecated // Forge: Use getStyleModifier
```
SizableShrimp: This is functionally different from a `@deprecated` javadoc, which would be preferred.

SizableShrimp (approved):

PR #8650, opened 2022-06-04
[1.18] Fix reload listener registration in ModelLoaderRegistry

### LINKED ISSUE #8649: ModelRegistryLoader could cause ConcurrentModificationException
**Minecraft Version:** 1.18.2

**Forge Version:** 40.1.14

**Logs:** 
Log contains 2 part:
1. the `ConcurrentModificationException` stack trace that provides no information
2. I added a mixin in `ReloadableResourceManager` to reveal who cause this exception
https://gist.github.com/lcy0x1/abcccaafbf1e5ab261874d5a9bdf6de7

**Steps to Reproduce:**
https://github.com/Minecraft-LightLand/ModPackHelper

**Description of issue:**
`ModelRegistryEvent` is fired at the same time as `ReloadableResourceManager.createReload`, causing the error

More details in discord discussion with diesieben07:
https://discord.com/channels/313125603924639766/915304642668290119/979821254133575801

diesieben07:
Currently `ModelLoaderRegistry.registerLoader` also registers the `IModelLoader` as a resource reload listener. `registerLoader` is supposed to be called during `ModelRegistryEvent`, which is fired _during the `prepare` stage of first resource reload_. This happens on a background worker thread and potentially while the loop over the list of reload listeners in `SimpleReloadInstance` is still going on. As such this is entirely the wrong place to register reload listeners and doing so can result in `ConcurrentModificationException`.

This PR fixes this by delaying the registration of loaders until _after_ the first reload has finished. Additionally a note is added to the javadoc to deprecated this automatic behavior and encourage mods to use `RegisterClientReloadListenersEvent` instead.

The registration of built-in loaders is also updated to actually use the required events.

Fixes #8649

### DIFF
--- a/src/main/java/net/minecraftforge/client/model/ModelLoaderRegistry.java
+++ b/src/main/java/net/minecraftforge/client/model/ModelLoaderRegistry.java
@@ -66,12 +66,12 @@ public class ModelLoaderRegistry
     public static void init()
     {
         var builtInLoaders = Map.of(
-            new ResourceLocation("minecraft","elements"), VanillaProxy.Loader.INSTANCE,
-            new ResourceLocation("forge","obj"), OBJLoader.INSTANCE,
-            new ResourceLocation("forge","bucket"), DynamicBucketModel.Loader.INSTANCE,
-            new ResourceLocation("forge","composite"), CompositeModel.Loader.INSTANCE,
-            new ResourceLocation("forge","multi-layer"), MultiLayerModel.Loader.INSTANCE,
-            new ResourceLocation("forge","item-layers"), ItemLayerModel.Loader.INSTANCE,
+            new ResourceLocation("minecraft", "elements"), VanillaProxy.Loader.INSTANCE,
+            new ResourceLocation("forge", "obj"), OBJLoader.INSTANCE,
+            new ResourceLocation("forge", "bucket"), DynamicBucketModel.Loader.INSTANCE,
+            new ResourceLocation("forge", "composite"), CompositeModel.Loader.INSTANCE,
+            new ResourceLocation("forge", "multi-layer"), MultiLayerModel.Loader.INSTANCE,
+            new ResourceLocation("forge", "item-layers"), ItemLayerModel.Loader.INSTANCE,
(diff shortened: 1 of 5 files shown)


KnightMiner:
If the automatic registration of the loader as a resource loader is removed in the future, I'd advise removing the interface/on reload method entirely from the loader interface. Would both make it more obvious it no longer works when that functionality is removed and remove a requ8red merhod thst thr majority of loaders never use.

For a more immediate change, might also be worth overriding the on reload method in a model loader to deprecate it, and probably also no-op it by default to make the transition simplier when removed.

diesieben07:
Unfortunately Java does not have a way to deprecate the fact that something is inherited from. If `onResourceManagerReload` is overridden and deprecated, it'll be deprecated even for someone who "legitimately" inherits from `IResourceManagerReloadListener` and registers the listener properly.

src/main/java/net/minecraftforge/client/model/ModelLoaderRegistry.java
```diff
@@ -63,17 +65,23 @@
...
-        registerLoader(new ResourceLocation("forge","multi-layer"), MultiLayerModel.Loader.INSTANCE);
-        registerLoader(new ResourceLocation("forge","item-layers"), ItemLayerModel.Loader.INSTANCE);
-        registerLoader(new ResourceLocation("forge", "separate-perspective"), SeparatePerspectiveModel.Loader.INSTANCE);
+        var builtInLoaders = Map.of(
+            new ResourceLocation("minecraft","elements"), VanillaProxy.Loader.INSTANCE,
+            new ResourceLocation("forge","obj"), OBJLoader.INSTANCE,
+            new ResourceLocation("forge","bucket"), DynamicBucketModel.Loader.INSTANCE,
+            new ResourceLocation("forge","composite"), CompositeModel.Loader.INSTANCE,
+            new ResourceLocation("forge","multi-layer"), MultiLayerModel.Loader.INSTANCE,
+            new ResourceLocation("forge","item-layers"), ItemLayerModel.Loader.INSTANCE,
```
pupnewfster: Missing space after the comma before the RL path (yes I know it was missing before as well but might as well fix it)

PR #8652, opened 2022-06-04
Add create functions (with resource location) for biome, configured structure & game event tags

Mysterious-Dev:
This PR add three patch with one function in each file (`create` function with resource location inside).

The goal is too add these functions like for blocks, items & fluids ^^.


diesieben07:
Why add patches for something in that already has a proper API in Forge (`ITagManager`, `DeferredRegister#create[Optional]TagKey`)?

Mysterious-Dev:
Because I don't know the existence of this function before. It's my mistake, I close this PR :/

PR #8655, opened 2022-06-04
1.16.x added a refresh  button

amongfr:
added a refresh button for mods to reload


TheCurle:
What you've done here is tried to make 1.18 the same as 1.15.

Let alone the feature you want to implement, reloading, is not possible any more.

Please learn how to use Git before creating useless spam PRs.

PR #8660, opened 2022-06-05
[1.19.x] Data-driven biome modifiers

Commoble:
This adds a serializer registry and json loader for biome modifier jsons, leveraging custom datapack registries.

This PR is currently branched from 1.18.x, but requires some coremods to Biome that aren't compatible with existing mods using access transformers to access the private fields in Biome, making this a breaking change that can't be made until 1.19.

## Design Requirements
* Biome modifiers must be able to modify biomes' generation features, mob spawns, climate, and client effects.
* Biome modifiers must be able to use biome tags to determine which biomes to modify.
* Biome modifiers must be able to refer to and use placedfeature jsons loaded via the vanilla placedfeature loader.
* Custom biome sources and chunk generators must be able to respect biomes' modified state.

## APIs
* Mods can create and register biome modifier serializers to the associated forge registry. This can be done by creating a DeferredRegister using ForgeRegistries.Keys.BIOME_MODIFIER_SERIALIZERS. The provided test mod has an example biome modifier and serializer.
* Biome modifier jsons are loaded into a custom datapack registry. Biome modifier jsons have the following format:
```
{
  "type": "modid:serializer", // indicates the registered serializer type
  // additional fields defined by the serializer can be read
}
```
* Biome Modifiers must implement a `modify` function, which provides the name and ID of the biome-being modified, a modification phase, and a builder for modified biome data.
  * The modification phases are intended to resemble the phases previously suggested by BiomeLoadingEvent, with removals happening after additions and modifications happening after removals.
* Forge provides a "forge:none" noop modifier type, so that datapacks can always selectively disable biome modifiers via datapack override.
* Biomes' original and modified state are readable via a field patched into Biome. Modified state is set after dimensions load but before serverconfigs are read.
* BiomeSource now has an additional constructor that allows its set of possible biomes to be lazily evaluated, delaying this evaluation until after biome modifiers apply.

## 1.18 -> 1.19 Growing Concerns
* BiomeLoadingEvent and BiomeDictionary can be removed in the future, being superceded by biome modifiers and biome tags, respectively.
* The changes to forge registries in 1.19 mean that BiomeModifierSerializer and BiomeModifier no longer need to extend ForgeRegistryEntry. BiomeModifierSerializer as a class can be discarded in favor of registering BiomeModifier Codecs directly; BiomeModifier can be made an interface.

## Patches and Coremods
* Biome requires several of these:
  * A `modifiableBiomeInfo` field is patched into Biome. This is initialized in Biome's constructor, to preserve a mod-readable original state of the biome. This also holds a modified biome state, which is set later during server init.
  * Biome's getters for `generationSettings` and `mobSettings` are patched to redirect to modifiableBiomeInfo.
  * As Biome uses many direct field accesses for `climateSettings` and `generationSettings`, a coremod is applied to redirect Biome's field reads to getter methods. Biome's loading codec is patched to bypass the coremod, to ensure read/write consistency (if the Biome is dumped to json during server runtime, it should serialize in the same state the Biome was loaded in, not the modified state).
* BiomeSource is patched to allow lazy evaluation of its possible biomes. This involves:
  * Patching `possibleBiomes` to be a Supplier
  * Patching in a new constructor that accepts a Supplier of possible biomes
  * Redirecting existing constructors to this new constructor
  * Patching `possibleBiomes()` to use the supplier


patches/minecraft/net/minecraft/world/level/biome/Biome.java.patch
```diff
@@ -20,3 +26,51 @@
...
++   public ClimateSettings getModifiedClimateSettings()
++   {
++       return this.modifiableBiomeInfo.get().climateSettings();
++   }
++   /**
++    * {@return The biome's client effects, with modifications if called after modifiers are applied in server init.
++    */
++   public BiomeSpecialEffects getModifiedSpecialEffects()
++   {
++       return this.modifiableBiomeInfo.get().effects();
```
SizableShrimp: Two things to note here:
1. Patches can use as many lines as they want so long as they are in one logical hunk. This is still the exact same amount of context lines with whitespace. Please prefer whitespace here.
2. Inline brackets should be used in patch code, instead of brackets on newlines.

Additionally, I think `// forge start` and `// forge end` markers would be good for these methods & field. cc @sciwhiz12 please advise formatting.

patches/minecraft/net/minecraft/world/level/biome/Biome.java.patch
```diff
@@ -20,3 +26,51 @@
...
++   public final net.minecraftforge.common.world.ModifiableBiomeInfo modifiableBiomeInfo;
++   /**
++    * {@return The biome's climate settings, with modifications if called after modifiers are applied in server init.
++    */
++   public ClimateSettings getModifiedClimateSettings()
++   {
++       return this.modifiableBiomeInfo.get().climateSettings();
++   }
++   /**
++    * {@return The biome's client effects, with modifications if called after modifiers are applied in server init.
```
SizableShrimp: Missing closing bracket in docs.

SizableShrimp: Missing closing bracket in docs.

patches/minecraft/net/minecraft/world/level/biome/Biome.java.patch
```diff
@@ -20,3 +26,51 @@
...
+    public int m_47539_() {
+@@ -491,5 +_,21 @@
+       public static Biome.TemperatureModifier m_47756_(String p_47757_) {
+          return f_47739_.get(p_47757_);
+       }
++   }
++   /** FORGE: Cache of biome data modified by biome modifiers, set by server after datapacks and serverconfigs load. Climate and effects field reads are coremodded to redirect to this.**/
++   public final net.minecraftforge.common.world.ModifiableBiomeInfo modifiableBiomeInfo;
++   /**
++    * {@return The biome's climate settings, with modifications if called after modifiers are applied in server init.
```
SizableShrimp: Missing closing bracket in docs.

SizableShrimp: Missing closing bracket in docs.

patches/minecraft/net/minecraft/world/level/biome/Biome.java.patch
```diff
@@ -20,3 +26,51 @@
...
+    }
+ 
+    public int m_47539_() {
+@@ -491,5 +_,21 @@
+       public static Biome.TemperatureModifier m_47756_(String p_47757_) {
+          return f_47739_.get(p_47757_);
+       }
++   }
++   /** FORGE: Cache of biome data modified by biome modifiers, set by server after datapacks and serverconfigs load. Climate and effects field reads are coremodded to redirect to this.**/
++   public final net.minecraftforge.common.world.ModifiableBiomeInfo modifiableBiomeInfo;
```
SizableShrimp: Instead of this weird docs I think it would make more sense to have this private and a public getter.

SizableShrimp: Instead of this weird docs I think it would make more sense to have this private and a public getter.

patches/minecraft/net/minecraft/world/level/biome/BiomeSource.java.patch
```diff
@@ -0,0 +1,39 @@
...
+    }
+ 
+    protected BiomeSource(List<Holder<Biome>> p_47894_) {
+-      this.f_47891_ = new ObjectLinkedOpenHashSet<>(p_47894_);
+-      this.f_186695_ = Suppliers.memoize(() -> {
+-         return this.m_186727_(p_47894_, true);
+-      });
++       this(() -> p_47894_); // FORGE: Redirect vanilla constructor to forge's lazy constructor.
++   }
++   protected BiomeSource(Supplier<List<Holder<Biome>>> biomes)
```
SizableShrimp, line 23: Newline here please.

SizableShrimp, line 23: Newline here please.

src/main/java/net/minecraftforge/common/world/BiomeModifier.java
```diff
@@ -0,0 +1,93 @@
...
+{
+    "type": "yourmod:yourserializer", // Indicates a registered serializer/codec.
+    // Additional fields can be specified here according to the serializer codec.
+}
+</pre>
+ * <p>The loading conditions are parsed first (if any), the rest of the json will not be loaded if the conditions fail.
+ * If the loading conditions pass, the serializer codec is read secondly, which is used to parse the rest of the BiomeModifier json.
+ * Datapacks can also disable a biome modifier by overriding the json with an empty {} object.</p>
+ * <p>TODO 1.19: Remove `extends ForgeRegistryEntry` and make this an interface.</p>
+ */
```
SizableShrimp: This formatting doesn't look right

SizableShrimp: This formatting doesn't look right

src/main/java/net/minecraftforge/common/world/BiomeModifier.java
```diff
@@ -0,0 +1,93 @@
...
+     * Codec for referring to biome modifiers by id, list of id, or tags. Can only be used with RegistryOps.
+     */
+    public static final Codec<HolderSet<BiomeModifier>> LIST_CODEC = RegistryCodecs.homogeneousList(ForgeRegistries.Keys.BIOME_MODIFIERS, DIRECT_CODEC);
+    
+    /**
+     * @param biome the named biome being modified (with original data readable).
+     * @param phase Biome modification phase.
+     * @param builder mutable biome info builder. Apply changes to this.
+     */
+    public abstract void modify(Holder<Biome> biome, Phase phase, BiomeInfo.Builder builder);
```
SizableShrimp: Can you please elaborate with a description of what this does / can do?

SizableShrimp: Can you please elaborate with a description of what this does / can do?

src/main/java/net/minecraftforge/common/world/BiomeModifier.java
```diff
@@ -0,0 +1,93 @@
...
+     * @return Registered BiomeModifierSerializer.
+     */
+    public abstract BiomeModifierSerializer<?> type();
+    
+    public static enum Phase
+    {
+        /**
+         * Catch-all for anything that needs to run before standard phases.
+         */
+        BEFORE_ADD,
```
SizableShrimp: BEFORE_EVERYTHING maybe?

SizableShrimp: BEFORE_EVERYTHING maybe?

src/main/java/net/minecraftforge/common/world/BiomeModifier.java
```diff
@@ -0,0 +1,93 @@
...
+         */
+        REMOVE,
+        /**
+         * Alteration of values such as climate or colors.
+         */
+        MODIFY,
+        /**
+         * Catch-all for anything that needs to run after standard phases.
+         */
+        AFTER_MODIFY
```
SizableShrimp: AFTER_EVERYTHING maybe?

SizableShrimp: AFTER_EVERYTHING maybe?

src/main/java/net/minecraftforge/common/world/BiomeSpecialEffectsBuilder.java
```diff
@@ -0,0 +1,110 @@
...
+import net.minecraft.world.level.biome.AmbientAdditionsSettings;
+import net.minecraft.world.level.biome.AmbientMoodSettings;
+import net.minecraft.world.level.biome.AmbientParticleSettings;
+import net.minecraft.world.level.biome.BiomeSpecialEffects;
+
+/**
+ * Extension of the vanilla builder but also provides read access and a copy-from-existing-data helper.
+ * Also, the base builder crashes if certain values aren't specified on build, so this enforces the setting of those.
+ **/
+public class BiomeSpecialEffectsBuilder extends BiomeSpecialEffects.Builder
```
SizableShrimp: Ending line should be `*/` not `**/`.

SizableShrimp: Ending line should be `*/` not `**/`.

src/main/java/net/minecraftforge/common/world/ModifiableBiomeInfo.java
```diff
@@ -0,0 +1,163 @@
...
+public class ModifiableBiomeInfo
+{
+    private final BiomeInfo originalBiomeInfo;
+    @Nullable
+    private BiomeInfo modifiedBiomeInfo = null;
+    
+    /**
+     * @param originalBiomeInfo BiomeInfo representing the original state of a biome when the biome was constructed.
+     */
+    public ModifiableBiomeInfo(final BiomeInfo originalBiomeInfo)
```
SizableShrimp: Probably `@NotNull` would be good here and on the field and on getOriginalBiomeInfo.

SizableShrimp: Probably `@NotNull` would be good here and on the field and on getOriginalBiomeInfo.

src/main/java/net/minecraftforge/common/world/ModifiableBiomeInfo.java
```diff
@@ -0,0 +1,163 @@
...
+ * Copyright (c) Forge Development LLC and contributors
+ * SPDX-License-Identifier: LGPL-2.1-only
+ */
+
+package net.minecraftforge.common.world;
+
+import java.util.List;
+import java.util.Locale;
+
+import javax.annotation.Nullable;
```
SizableShrimp: Wrong annotation. Use jetbrains.

SizableShrimp: Wrong annotation. Use jetbrains.

src/main/java/net/minecraftforge/common/world/NoBiomeModifier.java
```diff
@@ -0,0 +1,29 @@
...
+    {
+        // noop
+    }
+
+    @Override
+    public BiomeModifierSerializer<?> type()
+    {
+        return ForgeMod.NO_BIOME_MODIFIER.get();
+    }
+
```
SizableShrimp: Do we normally have trailing spaces at the end of classes?

SizableShrimp: Do we normally have trailing spaces at the end of classes?

src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -174,6 +175,11 @@ static RegistryBuilder<ForgeWorldPreset> getWorldTypesRegistryBuilder()
         return makeRegistry(WORLD_TYPES, ForgeWorldPreset.class).disableSaving().disableSync();
     }
 
+    static RegistryBuilder<BiomeModifierSerializer<?>> getBiomeModifierSerializersRegistryBuilder()
+    {
+        return makeRegistry(BIOME_MODIFIER_SERIALIZERS, c(BiomeModifierSerializer.class)).disableSaving().disableSync();
+    }
```
SizableShrimp: This makes no sense to be inconsistent about. The other new registry builder is inlined; this one should be too.

SizableShrimp: This makes no sense to be inconsistent about. The other new registry builder is inlined; this one should be too.

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -350,6 +350,18 @@ public net.minecraft.world.level.biome.Biome$ClimateSettings f_47683_ # downfall
...
+protected net.minecraft.world.level.biome.BiomeSpecialEffects$Builder f_48007_ # waterFogColor
+protected net.minecraft.world.level.biome.BiomeSpecialEffects$Builder f_48008_ # skyColor
+protected net.minecraft.world.level.biome.BiomeSpecialEffects$Builder f_48009_ # foliageColorOverride
+protected net.minecraft.world.level.biome.BiomeSpecialEffects$Builder f_48010_ # grassColorOverride
+protected net.minecraft.world.level.biome.BiomeSpecialEffects$Builder f_48011_ # grassColorModifier
+protected net.minecraft.world.level.biome.BiomeSpecialEffects$Builder f_48012_ # ambientParticle
+protected net.minecraft.world.level.biome.BiomeSpecialEffects$Builder f_48013_ # ambientLoopSoundEvent
+protected net.minecraft.world.level.biome.BiomeSpecialEffects$Builder f_48014_ # ambientMoodSettings
+protected net.minecraft.world.level.biome.BiomeSpecialEffects$Builder f_48015_ # ambientAdditionsSettings
+protected net.minecraft.world.level.biome.BiomeSpecialEffects$Builder f_48016_ # backgroundMusic
```
SizableShrimp, line 365: Is this every single field in the file? Prefer a group AT then.

SizableShrimp, line 365: Is this every single field in the file? Prefer a group AT then.

SizableShrimp (changes requested):

patches/minecraft/net/minecraft/world/level/biome/Biome.java.patch
Commoble: Fixed in next push

Commoble: Fixed in next push

patches/minecraft/net/minecraft/world/level/biome/Biome.java.patch
Commoble: Fixed in next push

Commoble: Fixed in next push

patches/minecraft/net/minecraft/world/level/biome/Biome.java.patch
Commoble: Changed in next push

Tslat:
`BiomeLoadingEvent and BiomeDictionary are superceded by biome modifiers and biome tags and can be removed in the future`.
While functionally true, I really don't think this is at all likely. Mojang has conventionally moved towards tags over time, and seem more inclined in doing so and keeping it that way.
I do not at all foresee mojang moving away from this, unless it's to replace it with another data-driven system, which we should be able to patch to use just fine
I don't see this as any kind of concern worth spending a moment of thought on

`The changes to forge registries in 1.19 mean that BiomeModifierSerializer and BiomeModifier no longer need to extend ForgeRegistryEntry. BiomeModifierSerializer as a class can be discarded in favor of registering BiomeModifier Codecs directly; BiomeModifier can be made an interface.`

I'd agree with this - I wouldn't even bother submitting the PR with the current implementation without first cleaning it up for 1.19
1.18 hasn't been fully moved to data-driven biomes yet, so it can still survive on the current system, and not having to distinguish between 1.18-style biomemodifiers and 1.19-style biomemodifiers when the topic comes up in help channels will be very nice

The rest of this looks pretty well thought out and covers all the bases I can think of to ask, so it gets a thumbs up from me

The only thing I want to question is this:
image
I'm not sure whether we're intentionally avoiding using the getter for the modifiableBiomeSettings, but if not, we should use that instead of referring directly to the field for all our getters
This allows custom biomes in-code to override the call if they want without needing to replace the final field value

TelepathicGrunt (approved):

Commoble:
Closed, merged into 1.19 internally

PR #8663, opened 2022-06-06
Revert "[1.18.x] Enable "Experimental Forge Light Pipeline" by default"

XFactHD:
As it turns out there are a few mods that can't handle the Forge light pipeline being enabled. As such it was decided to revert the change for now.


TheCurle:
Heck.
It pains me to revert a commit that would fix so much, but it turns out that several mods are hardcoded to only work with the old method, and springing a breaking change on them inadvertently is a really bad thing to do for many reasons.

Alas, this will be implemented in 1.19, so there's no excuses.

williewillus:
Can you elaborate more on how mods are "harcoded" to only work with the vanilla pipeline? My mental model is that I should code to the vanilla API and the Forge improvements are automagically done under the hood for me.

I've gotten a few scattered reports of Botania crashing on newer Forge versions with this enabled so would like to understand why. If this is something that will be changed and needs modder action, it needs to be clearly communicated, unless you want a horde of modders confused why their rendering crashes when porting to 1.19.

pupnewfster:
I don't believe this will end up requiring modder action due to #8665 (beyond potentially requiring some call paths to override the `getVertexFormat` method in 1.19

PR #8664, opened 2022-06-06
Fix: use copy-on-write render check map to maintain ability to register render checks outside of client setup

Gegy:
This PR reverts to an earlier approach from #8476 which is less likely to cause any breakage for mods. Instead of erroring on registration from outside of client setup, a copy-on-write map is used. This maintains the intent of the previous PR by not requiring synchronisation, but also resolves a crash in the case where a mod registers render type checks outside of client setup. For example, some mods such as CTM are registering render type checks from resource reload.

### DIFF
--- a/patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
@@ -42,7 +42,7 @@
           if (!Minecraft.m_91085_()) {
              return Sheets.m_110792_();
           } else {
-@@ -340,9 +_,70 @@
+@@ -340,9 +_,65 @@
        }
     }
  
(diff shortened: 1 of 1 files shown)


Technici4n:
I wonder how many blocks use this method in the average modpack - it might be too much for an eager copy-on-write approach. An alternative approach might be the following (sort of a lazy copy-on-write, pseudocode):

```java
private static final Object lock = new Object();
private static volatile Map<> readOnlyMap = null;
private static final Map<> editorMap = ...;

public static void register(key, value) {
    synchronized(lock) {
        editorMap.put(key, value);
        readOnlyMap = null;
    }
}

private static Map<> getReadOnlyMap() {
    var ret = readOnlyMap;
    if (ret == null) {
        synchronized (lock) {
            readOnlyMap = ret = copy(editorMap);
        }
    }
    return ret;
}
```

It should have the same reading performance cost as the copy-on-write approach, namely a single volatile read, but make subsequent registrations much faster. The additional branching is probably not a major concern.

Gegy:
I like that a lot better! It's a bit extra code, but I've given this a try 🙂

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
```diff
@@ -42,74 +42,91 @@
...
-+           Map<T, RenderType> vanillaMap
-+   ) {
-+      return Util.m_137469_(new it.unimi.dsi.fastutil.objects.Object2ObjectOpenHashMap<>(vanillaMap.size(), 0.5F), map -> {
-+         map.defaultReturnValue(SOLID_PREDICATE);
-+         for (Map.Entry<T, RenderType> entry : vanillaMap.entrySet()) {
-+            map.put(entry.getKey().delegate, createMatchingLayerPredicate(entry.getValue()));
-+         }
-+      });
-+   }
++   private static final RenderChecks<Block> BLOCK_RENDER_CHECKS = new RenderChecks<>(f_109275_);
```
XFactHD, line 58: Please revert the name of these two maps to the previous camel cased name, otherwise this change will still break CTM because they are accessing this map via reflection

zml2008: are private forge-added fields considered stable api now? what a world we live in

Technici4n: This doesn't matter since the type of the field changed anyway. And CTM should not be accessing private fields anyway, it's on them if it breaks.

Technici4n (approved):

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
Gegy: What is CTM depending on from these internals? Would it make sense in this PR to expand the API to handle CTM's use-case?

XFactHD: They are retrieving the predicates from the map and, under certain conditions, adding them to their own map and overwriting the predicates stored in `ItemBlockRenderTypes` with their own.
https://github.com/Chisel-Team/ConnectedTexturesMod/blob/1.18/dev/src/main/java/team/chisel/ctm/client/util/CTMPackReloadListener.java#L64-L79
https://github.com/Chisel-Team/ConnectedTexturesMod/blob/1.18/dev/src/main/java/team/chisel/ctm/client/util/CTMPackReloadListener.java#L130-L136

Technici4n: So it would seem that adding getters for the predicates should be enough?

Gegy:
Closing due to being replaced by #8671 🙂

PR #8666, opened 2022-06-06
[1.18.x] Fix #8530 (Banner Pattern) NPE

ChampionAsh5357:
This fixes #8530 by providing a default for the uninitialized static variables if no patterns are extensibly registered.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -1485,9 +1485,13 @@ public class ForgeHooks
         }
     }
   
-    private static BannerPattern[] nonPatternItems= Arrays.stream(BannerPattern.values()).filter(p -> !p.hasPatternItem).toArray(BannerPattern[]::new);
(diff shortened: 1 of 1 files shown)


src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1485,8 +1485,8 @@ public static MobEffect loadMobEffect(CompoundTag nbt, String key, @Nullable Mob
         }
     }
   
-    private static BannerPattern[] nonPatternItems;
-    private static int totalPatternRows;
+    private static BannerPattern[] nonPatternItems= Arrays.stream(BannerPattern.values()).filter(p -> !p.hasPatternItem).toArray(BannerPattern[]::new);
+    private static int totalPatternRows = (nonPatternItems.length + 2) / 4;
```
sciwhiz12: Perhaps the `refreshBannerPatternData` method could be called from a static initializer immediately below this method, to avoid duplicating code between the field initializers and the methods?

sciwhiz12 (changes requested):

sciwhiz12 (approved):

PR #8667, opened 2022-06-06
[1.18] Fix EntityEvent.Size not being fired

Meldexun:
`EntityEvent.Size` is only fired in the entity constructor and in `refreshDimensions`. But the event should be fired whenever `getDimensions(Pose)` or `getEyeHeight(Pose, EntitySize)` is called.

For example creating an event listener that just scales the player size by 0.5 will result in the player only being 0.9 blocks high. But when going into a one block gap the player still starts crawling.

Firing the event inside the two methods is problematic because already in vanilla the methods are overridden by several subclasses.
Firing the event whenever the methods are called is also not ideal because the methods get called at several locations in vanilla.
That's why the vanilla methods are now marked as deprecated and forge versions of these methods were added which just call the vanilla method and fire the event.

### Problems and Future
The `refreshDimensions` method wasn't changed and still uses the vanilla `getDimensions` and `getEyeHeight` methods and fires the event manually. The reason for not changing the method is that the event here also gets the old dimensions and old eye height of the entity.
Also `getEyeHeight(Pose)` is called the event gets fired twice because `getEyeHeightForge` and `getDimensionsForge` is called.

In the future I would split the event into `Dimensions` and `EyeHeight` and also change the `oldSize` and `oldEyeHeight` fields of the event to just hold the size and eyeheight when the event was created. This would make things a lot cleaner but splitting the event would be a breaking change and changing the old fields is a logical change (or however you would call this then).

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/Entity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/Entity.java.patch
@@ -372,7 +372,7 @@
  
     }
  
-@@ -3061,6 +_,89 @@
+@@ -3061,6 +_,76 @@
     public boolean m_142265_(Level p_146843_, BlockPos p_146844_) {
        return true;
     }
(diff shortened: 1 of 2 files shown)


patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -388,6 +444,19 @@
...
++      net.minecraftforge.event.entity.EntityEvent.Size e = net.minecraftforge.event.ForgeEventFactory.getEntitySizeForge(this, pose, size, eyeHeight);
++      return e.getNewSize();
++   }
++
++   protected float getEyeHeightForge(Pose pose, EntityDimensions size) {
++      float eyeHeight = this.m_6380_(pose, size);
++      net.minecraftforge.event.entity.EntityEvent.Size e = net.minecraftforge.event.ForgeEventFactory.getEntitySizeForge(this, pose, size, eyeHeight);
++      return e.getNewEyeHeight();
++   }
++
```
SizableShrimp: We have IForgeEntity for this stuff

Meldexun: Moved the methods into IForgeEntity

sciwhiz12:
Hello! @Meldexun, please leave a comment whether you wish to continue this PR and update it to the latest active version of 1.19.2 (or leave an explanation why this PR's changes are not applicable to 1.19.2). If we do not hear back from you after some time, this PR will be closed for inactivity.

Meldexun:
I created a new PR for 1.19

Jonathing:
1.18 is no longer supported. This has been addressed and fixed in updates to supported versions.

PR #8671, opened 2022-06-07
[1.18.x] Allow safely registering RenderType predicates at any time

XFactHD:
This PR effectively supersedes #8664 by providing a solution that keeps the performance benefit of the initial change in #8476 while avoiding breaking mods that need access to the predicate map, such as CTM.
This is done by using a lazy "copy-on-write" implementation where the map is only actually copied on the first read access after modification.
The PR also adds getters to allow mods like CTM to access an unmodifiable view of the maps instead of accessing internals via reflection.

Quote of the description of #8664 for completeness:
> This PR reverts to an earlier approach from #8476 which is less likely to cause any breakage for mods. Instead of erroring on registration from outside of client setup, a copy-on-write map is used. This maintains the intent of the previous PR by not requiring synchronisation, but also resolves a crash in the case where a mod registers render type checks outside of client setup. For example, some mods such as CTM are registering render type checks from resource reload.

### DIFF
--- a/patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
@@ -42,7 +42,7 @@
           if (!Minecraft.m_91085_()) {
              return Sheets.m_110792_();
           } else {
-@@ -340,9 +_,70 @@
+@@ -340,9 +_,106 @@
        }
     }
  
@@ -55,8 +55,12 @@
 +   // FORGE START
 +
 +   private static final java.util.function.Predicate<RenderType> SOLID_PREDICATE = type -> type == RenderType.m_110451_();
++   private static final Object LOCK = new Object();
++   // Access to the two following editable maps is guarded by synchronization and they are lazily copied to the "readonly" maps on first read after modification
 +   private static final Map<net.minecraftforge.registries.IRegistryDelegate<Block>, java.util.function.Predicate<RenderType>> blockRenderChecks = createRenderCheckMap(f_109275_);
 +   private static final Map<net.minecraftforge.registries.IRegistryDelegate<Fluid>, java.util.function.Predicate<RenderType>> fluidRenderChecks = createRenderCheckMap(f_109276_);
++   @org.jetbrains.annotations.Nullable private static volatile Map<net.minecraftforge.registries.IRegistryDelegate<Block>, java.util.function.Predicate<RenderType>> blockRenderChecksReadOnly = null;
++   @org.jetbrains.annotations.Nullable private static volatile Map<net.minecraftforge.registries.IRegistryDelegate<Fluid>, java.util.function.Predicate<RenderType>> fluidRenderChecksReadOnly = null;
 +
 +   private static <T extends net.minecraftforge.registries.ForgeRegistryEntry<T>> Map<net.minecraftforge.registries.IRegistryDelegate<T>, java.util.function.Predicate<RenderType>> createRenderCheckMap(
 +           Map<T, RenderType> vanillaMap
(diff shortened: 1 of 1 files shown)


Technici4n (approved):

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
```diff
@@ -74,37 +78,69 @@
...
 +   public static boolean canRenderInLayer(FluidState fluid, RenderType type) {
-+      return fluidRenderChecks.get(fluid.m_76152_().delegate).test(type);
++      return getFluidLayerPredicates().get(fluid.m_76152_().delegate).test(type);
 +   }
 +
 +   public static void setRenderLayer(Block block, RenderType type) {
 +      setRenderLayer(block, createMatchingLayerPredicate(type));
 +   }
 +
 +   public static synchronized void setRenderLayer(Block block, java.util.function.Predicate<RenderType> predicate) {
```
Gegy: Very tiny thing- maybe makes sense to remove the `synchronized` modifier now?

Gegy (approved):

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
XFactHD: Yeah, that can be safely removed, I'll do that real quick

sciwhiz12 (approved):

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
```diff
@@ -74,37 +78,69 @@
...
++
++   public static Map<net.minecraftforge.registries.IRegistryDelegate<Block>, java.util.function.Predicate<RenderType>> getBlockLayerPredicatesView() {
++      return java.util.Collections.unmodifiableMap(getBlockLayerPredicates());
++   }
++
++   private static Map<net.minecraftforge.registries.IRegistryDelegate<Block>, java.util.function.Predicate<RenderType>> getBlockLayerPredicates() {
++      Map<net.minecraftforge.registries.IRegistryDelegate<Block>, java.util.function.Predicate<RenderType>> map = blockRenderChecksReadOnly;
++      if (map == null) {
++         synchronized (LOCK) {
++            blockRenderChecksReadOnly = map = copy(blockRenderChecks);
```
tterrag1098: Should null check again inside the synch block: https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
```diff
@@ -74,37 +78,69 @@
...
++
++   public static Map<net.minecraftforge.registries.IRegistryDelegate<Fluid>, java.util.function.Predicate<RenderType>> getFluidLayerPredicatesView() {
++      return java.util.Collections.unmodifiableMap(getFluidLayerPredicates());
++   }
++
++   private static Map<net.minecraftforge.registries.IRegistryDelegate<Fluid>, java.util.function.Predicate<RenderType>> getFluidLayerPredicates() {
++      Map<net.minecraftforge.registries.IRegistryDelegate<Fluid>, java.util.function.Predicate<RenderType>> map = fluidRenderChecksReadOnly;
++      if (map == null) {
++         synchronized (LOCK) {
++            fluidRenderChecksReadOnly = map = copy(fluidRenderChecks);
```
tterrag1098: Ditto

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
```diff
@@ -74,37 +78,69 @@
...
++      if (map == null) {
++         synchronized (LOCK) {
++            blockRenderChecksReadOnly = map = copy(blockRenderChecks);
++         }
++      }
++      return map;
++   }
++
++
++   public static Map<net.minecraftforge.registries.IRegistryDelegate<Fluid>, java.util.function.Predicate<RenderType>> getFluidLayerPredicatesView() {
```
tterrag1098, line 128: Why are these view methods necessary? The map being wrapped is already a copy, creating an unmodifiable instance seems like unnecessary overhead.

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
```diff
@@ -55,8 +55,12 @@
 +   // FORGE START
 +
 +   private static final java.util.function.Predicate<RenderType> SOLID_PREDICATE = type -> type == RenderType.m_110451_();
++   private static final Object LOCK = new Object();
```
tterrag1098: Why does there need to be an explicit lock object? Isn't the contention between the two maps completely independent of one another? The lock object could simply be the map itself, right? It would also reduce lock contention during assignments.

tterrag1098 (changes requested):

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
XFactHD: Fixed

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
XFactHD: Fixed

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
XFactHD: You are correct, there is no difference between locking on an explicit lock object and locking on the relevant final map. Changed the locking accordingly

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
XFactHD: These methods only provide a view to prevent callers from modifying the map that may be concurrently accessed by the rendering process.

marchermans (approved):

tterrag1098 (approved):

tterrag1098:
This looks completely fine to me, but I will leave it to another team member to merge, as I am too close to the issue and a bit out of the loop in general.

PR #8675, opened 2022-06-07
[1.19.x] Update MDK for 1.19 changes

SizableShrimp:


iChun:
Update `pack.mcmeta` ?

SizableShrimp:
That's already been done before this PR

iChun:
Does `pack_format` no longer need updating if `forge:data_pack_format` is set?

AterAnimAvis:
Mojang split the packformat and dataformat versioning
Resourcepacks are current at 9
Datapacks are current at 10
See <https://github.com/MinecraftForge/MinecraftForge/pull/8612> and the related issue for more details

SizableShrimp:
Main thing to note:
> The original pack_format key is still required, and acts as the fallback if the more specific pack-typed format version key is absent.

Would need clarification from @sciwhiz12 but I would also assume that this also applies for pack format lookups from modded code or whatever that do not specify. (Not sure if this is even possible)

sciwhiz12:
The value for `pack_format` is used in two situations:
 1. When calling `PackMetadataSection#getPackFormat(PackType)` and there is no specific pack format for that pack type, as described in the PR description; and
 2. When calling the original `PackMetadataSection#getPackFormat()` method, which is deprecated and all uses of this method in vanilla code are redirected to use the `PackType`-sensitive method instead.

mdk/build.gradle
```diff
@@ -81,19 +80,10 @@ minecraft {
...
-            // "REGISTRIES": For firing of registry events.
-            // "REGISTRYDUMP": For getting the contents of all registries.
             property 'forge.logging.markers', 'REGISTRIES'
 
-            // Recommended logging level for the console
-            // You can set various levels here.
-            // Please read: https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels
             property 'forge.logging.console.level', 'debug'
 
-            // Comma-separated list of namespaces to load gametests from. Empty = all namespaces.
```
LexManos, line 96: Why were all the comments removed?
Yes they are verbose, but they are there to direct people where to look.

PaintNinja: Because they're duplicates of the same comments found in the client run config further up

LexManos: ah, works for me, once the pack format is resolved this looks pullable.

SizableShrimp: The pack.mcmeta is already resolved. <https://github.com/MinecraftForge/MinecraftForge/blob/ade2b8cc80254f050b38ff3d32655f08b259a21a/mdk/src/main/resources/pack.mcmeta>

PaintNinja:
Just a minor thing - the documentation link in the MDK's readme.txt should probably be updated to `https://` and the Discord link from https://discord.gg/UvedJ9m to https://discord.gg/forge

SizableShrimp:
The old discord link is still functional and more reliable if the vanity url was ever lost. I'll update the readme

sciwhiz12 (approved):

PR #8679, opened 2022-06-08
Override initialize

ferriarnus:
Override initialize in SlotItemHandler, so it uses the itemhandler instead of container

### DIFF
--- a/src/main/java/net/minecraftforge/items/SlotItemHandler.java
+++ b/src/main/java/net/minecraftforge/items/SlotItemHandler.java
@@ -49,7 +49,8 @@ public class SlotItemHandler extends Slot
     }
     
     @Override
-    public void initialize(ItemStack stack) {
+    public void initialize(ItemStack stack)
+    {
         ((IItemHandlerModifiable) this.getItemHandler()).setStackInSlot(index, stack);
         this.setChanged();
     }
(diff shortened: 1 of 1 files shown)


src/main/java/net/minecraftforge/items/SlotItemHandler.java
```diff
@@ -47,6 +47,13 @@ public void set(@NotNull ItemStack stack)
         ((IItemHandlerModifiable) this.getItemHandler()).setStackInSlot(index, stack);
         this.setChanged();
     }
+    
+    @Override
+    public void initialize(ItemStack stack)
+    {
+        ((IItemHandlerModifiable) this.getItemHandler()).setStackInSlot(index, stack);
+        this.setChanged();
```
pupnewfster, line 56: Would a potentially cleaner solution be to make initialize just call set by default? that way people don't have to override both if their IItemHandler doesn't implement IItemHandlerModifiable? Alternatively the comment about that should be on initialize as well.

ferriarnus: Sounds good. I was hesitant about it, since I was not too sure about the difference with "set", so there might have been a reason for it being different. But since the default vanilla method is the same as "set", it would be fair to do the same here.

diesieben07: The difference is that set should trigger side effects such as equipment sounds (that's what vanilla uses it for), where as initialize should not triggers these. I don't think they should just delegate.

pupnewfster: Yeah that is what I was unsure about whether or not it makes sense. So then in that case add a comment like there is on set regarding IItemHandlerModifiable

ferriarnus: Like this?

PR #8681, opened 2022-06-08
Make fishing rods use tool actions for relevant logic

eerussianguy:
This is a 1.19 port of #8424 which is a port of #8168. There is not much more to be said about this PR that has not been said in the previous conversations. To summarize, this removes hardcoded references to `Items.FISHING_ROD` in a couple places that make life harder for modders.


ChampionAsh5357:
Looks fine to me, though this should have a test mod with a custom fishing rod to verify behavior.

ChampionAsh5357 (approved):

PR #8685, opened 2022-06-08
[1.19] Allow safely registering RenderType predicates at any time

XFactHD:
This PR is the 1.19 port of #8671.

This PR effectively supersedes https://github.com/MinecraftForge/MinecraftForge/pull/8664 by providing a solution that keeps the performance benefit of the initial change in https://github.com/MinecraftForge/MinecraftForge/pull/8476 while still allowing mods to register render type predicates at any point in time.
This is done by using a lazy "copy-on-write" implementation where the map is only actually copied on the first read access after modification.
The PR also adds getters to allow mods like CTM to access an unmodifiable view of the maps instead of accessing internals via reflection.

Quote of the description of https://github.com/MinecraftForge/MinecraftForge/pull/8664 for completeness:

> This PR reverts to an earlier approach from https://github.com/MinecraftForge/MinecraftForge/pull/8476 which is less likely to cause any breakage for mods. Instead of erroring on registration from outside of client setup, a copy-on-write map is used. This maintains the intent of the previous PR by not requiring synchronisation, but also resolves a crash in the case where a mod registers render type checks outside of client setup. For example, some mods such as CTM are registering render type checks from resource reload.

### DIFF
--- a/patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
@@ -1,9 +1,11 @@
 --- a/net/minecraft/client/renderer/ItemBlockRenderTypes.java
 +++ b/net/minecraft/client/renderer/ItemBlockRenderTypes.java
-@@ -19,6 +_,7 @@
+@@ -18,7 +_,8 @@
+ import net.minecraftforge.api.distmarker.OnlyIn;
  
  @OnlyIn(Dist.CLIENT)
- public class ItemBlockRenderTypes {
+-public class ItemBlockRenderTypes {
++public class ItemBlockRenderTypes extends net.minecraftforge.client.extensions.ForgeItemBlockRenderTypes {
 +   @Deprecated
     private static final Map<Block, RenderType> f_109275_ = Util.m_137469_(Maps.newHashMap(), (p_109296_) -> {
        RenderType rendertype = RenderType.m_110503_();
@@ -42,7 +44,7 @@
           if (!Minecraft.m_91085_()) {
              return Sheets.m_110792_();
           } else {
-@@ -350,9 +_,103 @@
+@@ -350,10 +_,68 @@
        }
     }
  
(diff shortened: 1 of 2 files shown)


src/main/java/net/minecraftforge/client/extensions/ForgeItemBlockRenderTypes.java
```diff
@@ -0,0 +1,103 @@
+package net.minecraftforge.client.extensions;
```
sciwhiz12, line 6: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/client/extensions/ForgeItemBlockRenderTypes.java
```diff
@@ -0,0 +1,103 @@
...
+        var newMap = new Object2ObjectOpenHashMap<>(map);
+        newMap.defaultReturnValue(SOLID_PREDICATE);
+        return newMap;
+    }
+
+    private static <T> Map<Holder.Reference<T>, Predicate<RenderType>> createMap()
+    {
+        return Util.make(new Object2ObjectOpenHashMap<>(Object2ObjectOpenHashMap.DEFAULT_INITIAL_SIZE, .75F), map ->
+                map.defaultReturnValue(SOLID_PREDICATE)
+        );
```
sciwhiz12: This is more a matter of style, but should we use `Util.make` here? To my knowledge, that method is designed for use in field initializers which need to modify the value further and don't wish to use a `static` class initializer block due to ugliness, such as adding entries to a modifiable map (or configuring a map, in this case). 

In my opinion, we can (and should) do this without `Util.make`, as we already do in `#copy` above.

src/main/java/net/minecraftforge/client/extensions/ForgeItemBlockRenderTypes.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    protected static void initFromVanillaMaps(Map<Block, RenderType> typeByBlock, Map<Fluid, RenderType> typeByFluid)
+    {
+        for(Map.Entry<Block, RenderType> entry : typeByBlock.entrySet())
```
sciwhiz12: Formatting: space after `for` keyword.
```suggestion
        for (Map.Entry<Block, RenderType> entry : typeByBlock.entrySet())
```

src/main/java/net/minecraftforge/client/extensions/ForgeItemBlockRenderTypes.java
```diff
@@ -0,0 +1,103 @@
...
+        {
+            RenderType type = entry.getValue();
+            blockRenderChecks.put(ForgeRegistries.BLOCKS.getDelegateOrThrow(entry.getKey()), t -> t == type);
+        }
+
+        for(Map.Entry<Fluid, RenderType> entry : typeByFluid.entrySet())
```
sciwhiz12: Formatting: space after `for` keyword.
```suggestion
        for (Map.Entry<Fluid, RenderType> entry : typeByFluid.entrySet())
```

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
```diff
@@ -42,72 +44,62 @@
...
++   public static Map<net.minecraft.core.Holder.Reference<Block>, java.util.function.Predicate<RenderType>> getBlockLayerPredicatesView() {
++      return java.util.Collections.unmodifiableMap(getBlockLayerPredicates());
 +   }
 +
-+   private static java.util.function.Predicate<RenderType> createMatchingLayerPredicate(RenderType type) {
-+      java.util.Objects.requireNonNull(type);
-+      return type::equals;
-    }
++   public static Map<net.minecraft.core.Holder.Reference<Fluid>, java.util.function.Predicate<RenderType>> getFluidLayerPredicatesView() {
++      return java.util.Collections.unmodifiableMap(getFluidLayerPredicates());
```
sciwhiz12: Should we cache these unmodifiable views of these maps, or it is too much effort to cache them (given that their backing maps can and will change) with too little gain? 

Also, should we note via javadocs that these returned views should not be cached, as their backing map will become outdated as soon as something registers a new render layer predicate?

sciwhiz12 (changes requested):

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
XFactHD: Caching them is pointless due to their volatile nature. I'll add the suggested javadoc.

src/main/java/net/minecraftforge/client/extensions/ForgeItemBlockRenderTypes.java
XFactHD: I'll get rid of it, it's a leftover from copying this back and forth between a method and a field initializer.

sciwhiz12 (approved):

marchermans:
This pr is superseded by amas refactor and can be closed.

It's backport is still needed.

LexManos:
Pulling this into 1.19, as Ama's PR will be for 1.19.1. So it'll get reverted shortly, but worth it for 1.19.

tterrag1098:
I wish this was held off until it was synced with its parent PR #8671 ... but if it's truly going to be replaced I suppose it's not a huge problem.

LexManos:
We've been waiting on you (and giga) to review this. It was pulled so that 1.19.0 was functional, while we know that it's gunna be replaced wholesale for 1.19.1.

PR #8687, opened 2022-06-08
[1.19.x] Update Door datagen

ferriarnus:
Updated door data generators to 1.19 (since someone mentioned them in forgecord). 

I also checked the wall. The check does fail, but this seems to be because the checker can not handle that the parameter order is different. I did not touch that part, since it seems to not be my place to change.

P.S. : I also have "fixed" the mods.toml for dev, should I push it?


Minecraftschurli (approved):

AnonymousHacker1279 (approved):

src/main/java/net/minecraftforge/client/model/generators/BlockStateProvider.java
```diff
@@ -534,27 +534,30 @@ public void doorBlock(DoorBlock block, String name, ResourceLocation bottom, Res
...
             if (open) {
                 yRot += 90;
             }
-            if (rh && open) {
+            if (right && open) {
                 yRot += 180;
             }
             yRot %= 360;
-            return ConfiguredModel.builder().modelFile(state.getValue(DoorBlock.HALF) == DoubleBlockHalf.LOWER ? (right ? bottomRight : bottomLeft) : (right ? topRight : topLeft))
+            return ConfiguredModel.builder().modelFile(state.getValue(DoorBlock.HALF) == DoubleBlockHalf.LOWER ? (right ? ( open ? bottomRightOpen : bottomRight ) : ( open ? bottomLeftOpen : bottomLeft )) : (right ? ( open ? topRightOpen : topRight) : ( open ? topLeftOpen : topLeft)))
```
SizableShrimp, line 560: This line is way too complex. It should instead be nested ifs that store to a variable or something.

src/main/java/net/minecraftforge/client/model/generators/BlockStateProvider.java
```diff
@@ -534,27 +534,30 @@ public void doorBlock(DoorBlock block, String name, ResourceLocation bottom, Res
...
+        ModelFile bottomRight = models().doorBottomRight(baseName + "_bottom_right", bottom, top);
+        ModelFile bottomRightOpen = models().doorBottomRightOpen(baseName + "_bottom_right_open", bottom, top);
+        ModelFile topLeft = models().doorTopLeft(baseName + "_top_left", bottom, top);
+        ModelFile topLeftOpen = models().doorTopLeftOpen(baseName + "_top_left_open", bottom, top);
+        ModelFile topRight = models().doorTopRight(baseName + "_top_right", bottom, top);
+        ModelFile topRightOpen = models().doorTopRightOpen(baseName + "_top_right_open", bottom, top);
+        doorBlock(block, bottomLeft, bottomLeftOpen, bottomRight, bottomRightOpen, topLeft, topLeftOpen, topRight, topRightOpen);
+    }
+    
+    public void doorBlock(DoorBlock block, ModelFile bottomLeft, ModelFile bottomLeftOpen , ModelFile bottomRight, ModelFile bottomRightOpen, ModelFile topLeft, ModelFile topLeftOpen, ModelFile topRight, ModelFile topRightOpen) {
```
SizableShrimp, line 548: Extra space here before a comma that shouldn't be

src/test/java/net/minecraftforge/debug/DataGeneratorTest.java
```diff
@@ -798,7 +798,7 @@ protected void registerStatesAndModels()
             stairsBlock((StairBlock) Blocks.ACACIA_STAIRS, "acacia", mcLoc("block/acacia_planks"));
             slabBlock((SlabBlock) Blocks.ACACIA_SLAB, ForgeRegistries.BLOCKS.getKey(Blocks.ACACIA_PLANKS), mcLoc("block/acacia_planks"));
 
-            // TODO 1.19: fix fenceBlock, wallBlock, doorBlock, and co -SS
+            // TODO 1.19: fix fenceBlock, wall and co -SS
```
SizableShrimp, line 801: it's still wallBlock so shouldn't be changed from the method name in the comment.
Additionally, would it be possible for you to look into these other cases and fix them at the same time? Would be a lot nicer all in one PR

SizableShrimp (changes requested):

ferriarnus:
Unfortunately, due to both messing up a rebase and my time running out, I won't be able to finish this for now. The door part works, however, the wall verification fails during the datagen test, while it is correct. I did not see a clean way to solve this, which is why it isn't done yet. I will close the PR for now, so someone else can pick it up instead of waiting for me. If it is not yet done by the time I'm back, I will pick it up again.

PR #8695, opened 2022-06-09
[1.19.x] Fluid Overhaul Again

### LINKED ISSUE #8608: [1.19.x] The Fluid API Overhaul
# Introduction

[Previous PR](https://github.com/MinecraftForge/MinecraftForge/pull/7992) | [Current Fork](https://github.com/ChampionAsh5357/MinecraftForge/tree/fluid-redesign)

One of the few things that has been massively unfinished is the implementation of a feature complete fluid API within Forge. This is due to the massively hardcoded nature of fluids and their behavior since the Flattening back in 1.13. This issue aims to discuss the current conditions of fluids within Forge along with what changes have been taken to remedy their changes. Additionally, this issue will look at fluid implementations within the Forge ecosystem and fix any issues or concerns that may not have been addressed in previous versions, along with provide new extensions and hooks as needed to support an extensible use case.

If there are any suggestions or changes that should be discussed, please leave it below or ping me in the discord. I would like to incorporate as many usecases as possible to make Forge's fluid API as feature complete as possible.

## Current Progress

#7992 which defined the previous changes made to overhaul fluids is currently under a full redesign due to certain issues that have arisen as a result of choices made by Forge's original PR. For a current view of the progress done on fluids, you can view [this branch](https://github.com/ChampionAsh5357/MinecraftForge/tree/fluid-redesign) on my fork. I will be updating it as needed such that when 1.19 comes around, the current progress can be easily transitioned.

For redesign purposes, all previous fluid code within Forge has been marked as deprecated. This does not mean it will be removed, just that it hasn't been looked over yet during the redesign process. The only types mentioned for removal are those annotated with `forRemoval` set to true.

## Limitations

There are certain things that will not be up for discussion as they were decided beforehand or do not have a fully defined agreement. Any new limitations after this issue opens will also appear here.

- The Fluid Size of a bucket will remain at 1000. This has been fully decided by Forge when #8166 was closed.

# The Fluid API

This will be broken down into four sections: Common, Physics, Rendering, and Capability. These sections may be left empty and completed over time depending on how much work has currently been done in those sections. Additionally, sections that contain checkboxes will determine whether the task has been completed and pushed to the branch on the above fork.

## Common

- [x] Introduce a new method of linking fluids together between their source block and flowing block.

This method was necessary as a result of wanting to decouple fluid logic from tags. As such, fluids were given a new required registry object called `FluidType`. This defines the behavior and traits of a fluid whether flowing or the source. This will be used to represent the fluid as a whole and not as a source block or flowing block. This deprecates the behavior of `FluidAttributes` for removal as it was never a unique static object to accurately represent the fluid. Most of the properties for `FluidAttributes` will be moving onto the `FluidType` as a result. `FluidType` can also be additionally extended to provide for complex logic on overrides which are not typically available through the builder.

As `FluidType` is a registry object, it must be registered to be able to use within your fluid. It can be set on a fluid by overriding `Fluid#getFluidType`.

- [x] Give fluids a description id.

Previously known as `translationKey`, the description id, when specified, represents the translation key of the `FluidType`. If none is specified during building, it will default to `fluid_type.modid.registry_name` as defined by vanilla syntax.

- [x] Sounds

Fluids tend to have different sounds depending on certain actions whether filling/emptying a bucket or stepping into and swimming. Sounds in fluids are represented using a SoundAction which defines what sound should be played during the given action for a sound. Current SoundActions are BUCKET_FILL and BUCKET_EMPTY to replicate previous behavior.

- [x] Hydration

Fluids are able to hydrate blocks around them such as concrete, farmland and sponges. Hydration follow the following call hierarchy; at any point the property can be overridden to perform a different behavior from the global behavior defined by the method beneath.

- IForgeBlockState#canBeHydrated
- IForgeBlock#canBeHydrated
- IForgeFluidState#canHydrate
- IForgeFluid#canHydrate
- FluidType#canHydrate

These methods take in a `BlockGetter`, the state of the block being hydrated, the position of where the `BlockState` is/will be located, the state of the fluid hydrating the block, and the position of where the fluid is. Currently, hydration is supported for farmland, sugar cane, coral, sponge, and concrete powder.

- [x] Fluid Interactions

Fluid interactions are defined through the FluidInteractionRegistry. Each interaction is registered by the source which is checking the surrounding fluids, the predicate on when to execute a given interaction, and what to do on interaction. Vanilla uses lava as the source for checking the surrounding fluids for water or blue ice with soul soil underneath.

- [x] Properties

Fluids have common properties that can be applied in different circumstances. They may have different accessors depending on the use case.

- lightLevel: The light level emitted by the fluid
    - Empty: 0
    - Water: 0
    - Lava: 15
    - Default: 0
- density: Density of the fluid, arbitrary number where negative density is lighter than air. Approximately realistic to actual fluids.
    - Empty: 0
    - Water: 1000
    - Lava: 3000
    - Default: 1000
- temperature: Temperature of the fluid, arbitrary number where higher temperature indicates it is hotter. Approximately realistic to actual fluids.
    - Empty: 0
    - Water: 300
    - Lava: 1300
    - Default: 300
- viscosity: Resistance of movement through the fluid, arbitrary number where higher numbers means the entity moves slower through the fluid and vice versa. Negative values are not permitted. Approximately realistic to actual fluids.
    - Empty: 0
    - Water: 1000
    - Lava: 6000
    - Default: 1000

## Physics

- [x] Replace vanilla fluid physics using tags with a new system.

To accommodate the move from tag physics, the `FluidType` system will be used to denote the commonalities between multiple fluids instead. This system deprecates vanilla's system of handling fluid physics and as such, no longer handles any logic in relation to physics. A `FluidType` can define the following properties. Additionally, overrides which call these methods can be found in the associated entity extensions (`IForgeEntity`, `IForgeLivingEntity`, `IForgeBoat`).

- Motion scale: A scalar which determines an entities acceleration in the fluid
    - Empty: No scaling (1)
    - Water: Scaled by 0.014
    - Lava: Scaled by 0.007 in Ultrawarm dimensions, otherwise 0.0023
    - Default: Scaled by 0.014
- Pushing Entity: Whether an entity can be pushed by the fluid
    - Empty: False
    - Water: True
    - Lava: True
    - Default: True
- Fluid motion: How an entity can move through the specified fluid
    - Empty: As defined by vanilla
    - Water: As defined by vanilla
    - Lava: As defined by vanilla
    - Default: Uses water movement when false, otherwise custom movement when true
- Swimming: If the entity is able to swim in the fluid
    - Empty: False
    - Water: True
    - Lava: False
    - Default: True
- Drowning: If the entity is able to drown in the fluid
    - Empty: False
    - Water: True
    - Lava: False
    - Default: True
- Fall Distance Modifier: When the entity comes into contact with the fluid, how much should the fall distance be decreased every tick
    - Empty: No scaling (1)
    - Water: Scaled by 0
    - Lava: Scaled by 0.5
    - Default: Scaled by 0.5
- Extinguishing: If the entity can be extinguished by the fluid (this typically refers to fire in water, but it can be left up to the modder's discretion)
    - Empty: False
    - Water: True
    - Lava: False
    - Default: False
- Multiply Source Blocks: If the fluid can multiply its source blocks
    - Empty: False
    - Water: True
    - Lava: False
    - Default: False

Additionally, the entity defines some more properties which can be modified based on the fluid:

- Swimming: If the fluid permits swimming, then this will be checked to see whether the entity can swim in the fluid
    - Default: If the entity is sensitive to water, then as long as the fluid type is not water, otherwise true
- Rising/Sinking: Determines the movement of the entity when attempting to rise or sink in the fluid deliberately
    - Default: Copied from LivingEntity jumpInLiquid and sinkInLiquid respectively
- Drowning: If the fluid permits drowning, then this will be checked to see whether the entity can drown in the fluid, only is checked for standard air similar to the player uses
    - Default: If the type is water, then if the entity cannot breather underwater, else if it is a WaterAnimal but not a dolphin not in water then return true, otherwise if there is a fluid.

- [x] Handle Item Movements

Items are a special case of entity as their fluid logic is typically held internally. As such, the item logic can be specified by overriding `FluidType#setItemMovement`.

- Empty: As defined by vanilla
- Water: As defined by vanilla
- Lava: As defined by vanilla
- Default: Water movement

- [x] Boating

Boats are also a special case of entity since they are typically only reserved for water. This allows the mod to specify whether a fluid can be boated on via `FluidType#supportsBoating`.

- Empty: False
- Water: True
- Lava: False
- Default: False

- [x] Pathing

Entities may pathfind around fluids as blocks depending on their node evaluators. Fluids should be able to specify how an entity should pathfind in a given situation.

Pathing is implemented for the WalkNodeEvaluator for the current block and surrounding adjacent blocks. BlockPathTypes is also made an extensible enum for this purpose. Additionally, to support the difference between fluidloggable blocks, the logic can be applied at different times as necessary, though this is only used by water in vanilla.

- Empty: Null, Null
- Water: WATER, WATER_BORDER
- Lava: LAVA, Null
- Default: WATER, WATER_BORDER

## Rendering

- [x] IFluidRenderProperties

Implements the base type for handling rendering of a given a `FluidType` in Forge.

- [x] View Overlay

Fluids allow for an overlay to applied over the screen when submerged in a fluid. This can be done by specifying a texture for rendering in `IFluidTypeRenderProperties#getRenderOverlayTexture`. Additionally, custom rendering logic can be implemented via `#renderOverlay` instead.

- [x] Fog Rendering

Fluids allow for fog to be rendered as a way of giving the effect that view is limited under fluids. Custom fluids can either modify the color or rendering method used via `IFluidTypeRenderProperties#modifyFogColor` and `#modifyFogRender`. Otherwise, the default fog rendering will occur.

- [x] Fluid Models

Fluids are typically rendered through `LiquidBlockRenderer` which makes their rendering method hardcoded. This should be replaced with a custom model loader which defines the basic properties a fluid can take on. Additionally, the model loader should also be able to accept a dynamic renderer similar to block entities to allow for more diverse fluid renderings. The model loader will hold the information for the still, flowing, and overlay texture.

Didn't have time to finish before 1.19, so the still, flowing, and overlay textures will just be moved to `IFluidTypeRenderProperties` instead.

- [x] Color

The fluid's color will be determined by `IFluidTypeRenderProperties#getColorTint` which can be modifiable depending on the state of the fluid and the world around it.

## Capability

No time to look over before 1.19, so they will be left alone.

ChampionAsh5357:
Closes #8608.

Implements custom fluid physics, all explanations of what is added and modified is on the issue. Note, this removes physics handling from tags altogether and places it on a common registry object instead.


src/main/java/net/minecraftforge/fluids/FluidType.java
```diff
@@ -0,0 +1,1244 @@
...
+    /**
+     * Returns a sound to play when a certain action is performed by the
+     * entity in the fluid. If no sound is present, then the sound will be
+     * {@code null}.
+     *
+     * @param entity the entity in the fluid
+     * @param action the action being performed
+     * @return the sound to play when performing the action
+     */
+    public SoundEvent getSound(Entity entity, SoundAction action)
```
alcatrazEscapee, line 391: Missing `@Nullable` ?

src/main/java/net/minecraftforge/fluids/FluidType.java
```diff
@@ -0,0 +1,1244 @@
...
+    {
+        return this.adjacentPathType;
+    }
+
+    /**
+     * Returns whether the fluid can create a source.
+     *
+     * @return {@code true} if the fluid can create a source, {@code false} otherwise
+     */
+    public boolean canConvertToSource()
```
alcatrazEscapee: I'm not convinced by the presence of these no-argument accessors. This looks like a pattern seen in vanilla, where a no-argument method (say, `Block.getFriction`), is basically patched out in favor of a Forge added one, with additional context (say, `IForgeBlock.getFriction`). However, this entirely relies on the fact that `Block.getFriction` is essentially deprecated, with instructions to never call this method (because blocks may be expected to override the `IForgeBlock` method, and will return inconsistent results).

In a `FluidType`, if a mod overrides `canConvertToSource(FluidState, LevelReader, BlockPos)`, and they (for example) only allow you to convert to sources in a certain dimension. The return value of `canConvertToSource()` is going to be just plain wrong. And yet there's no indication that this is a method that *should not be called*

It appears to me that `FluidType` is apparently designed to allow mods to override these methods and use the additional context they provide, except they *also* need to be able to function without any additional context, making said additional context a bit meaningless.

---

**Suggestion**:

Unless there's a vanilla call path that is documented (i.e. in `FluidType.getLightLevel`), then there should not be no-arg accessors for these properties. The only ones should be through the accessors with additional arguments, to actually support mods that override the latter.

ChampionAsh5357: That's a fair enough point. Though, the one you are suggesting with `#canConvertToSource` needs to be dealt with in `FlowingFluid` unless we are completing deprecating the context, which I'm not sure how much I should do in vanilla. For right now, I'll assume that contexts which have been replaced will have their accessors removed as well.

alcatrazEscapee: I don't see a way in which the methods can be effectively overridden from an API perspective, if there are still consistent enough call paths - either used by vanilla or from mods - that call no-arg methods. That just seems to be asking for inconsistent behavior depending on where the method is invoked from. So I'd say either:

- Remove the additional context, because it *will* be called through places that don't have it - so whatever is doing there needs to be able to handle said context. or,
- Fully patch in the additional context, and remove all (or "essentially" all) vanilla call paths for methods without context. (The path Forge typically takes with cases i.e. `IForgeBlock.getFriction()`). In this case also throwing `@Deprecated /** @deprecated call FluidType#canConvertToSource(...additional context...)` on the vanilla method `FlowingFluid#convertToSource()` would also be useful to indicate they shouldn't be used.

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -82,12 +101,12 @@
  
 -            this.f_19853_.m_46473_().m_7238_();
 -         }
-+            if (this.f_146810_ && (this.f_146808_ || this.m_20071_())) {
++            if (this.f_146810_ && (this.f_146808_ || this.m_20071_() || this.isInFluidType((fluidType, aDouble) -> this.canFluidExtinguish(fluidType)))) {
```
pupnewfster: Instead of using `aDouble` as the predicate might as well call it `height` even if it is unused here

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -82,12 +101,12 @@
...
 +               this.m_146873_();
 +            }
 +         }
 +
-+         if (this.m_6060_() && (this.f_146808_ || this.m_20071_())) {
++         if (this.m_6060_() && (this.f_146808_ || this.m_20071_() || this.isInFluidType((fluidType, aDouble) -> this.canFluidExtinguish(fluidType)))) {
```
pupnewfster: Instead of using `aDouble` as the predicate might as well call it `height` even if it is unused here

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -141,7 +160,62 @@
...
++         this.m_20282_(this.m_20142_() && (this.m_20069_() || this.isInFluidType((fluidType, height) -> this.canSwimInFluidType(fluidType))) && !this.m_20159_());
+       } else {
+-         this.m_20282_(this.m_20142_() && this.m_5842_() && !this.m_20159_() && this.f_19853_.m_6425_(this.f_19826_).m_205070_(FluidTags.f_13131_));
++         this.m_20282_(this.m_20142_() && (this.m_5842_() || this.canStartSwimming()) && !this.m_20159_());
+       }
+ 
+    }
+ 
+    protected boolean m_20073_() {
+-      this.f_19799_.clear();
```
pupnewfster: Is there any actual benefit to patching out the call to `fluidHeight#clear` rather than leaving it in and adding the `forgeFluidTypeHeight#clear` call as well? Seems like it just increases patch complexity slightly. And given the fact that `Entity.fluidHeight` is protected someone might end up adding to it directly if they are doing something weird.

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -141,7 +160,62 @@
...
++      this.forgeFluidTypeHeight.clear();
+       this.m_20074_();
+-      double d0 = this.f_19853_.m_6042_().f_63857_() ? 0.007D : 0.0023333333333333335D;
+-      boolean flag = this.m_204031_(FluidTags.f_13132_, d0);
+-      return this.m_20069_() || flag;
++      this.f_19789_ *= this.forgeFluidTypeHeight.object2DoubleEntrySet().stream().filter(e -> !e.getKey().isAir() && !e.getKey().isVanilla()).map(e -> this.getFluidFallDistanceModifier(e.getKey())).min((f1, f2) -> Float.compare(f1, f2)).orElse(1F);
++      if (this.isInFluidType((fluidType, height) -> !fluidType.isAir() && !fluidType.isVanilla() && this.canFluidExtinguish(fluidType))) this.m_20095_();
```
pupnewfster: `(f1, f2) -> Float.compare(f1, f2)` can just become `Float::compare`

Also this method is a bit hard to follow in terms of why fall distance and fire extinguishing is being patched in until I noticed that it is just when it isn't vanilla that it runs for and that vanilla does a similar thing in `updateInWaterStateAndDoWaterCurrentPushing` with one important difference: It only does it if the entity is not currently riding a boat (and given there are patches to make boats support custom fluids) I think it is important to make sure that also happens for these?

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -141,7 +160,62 @@
...
++      return this.isInFluidType();
+    }
+ 
+    void m_20074_() {
+@@ -1101,8 +_,8 @@
+    }
+ 
+    private void m_20323_() {
+-      this.f_19800_ = this.m_204029_(FluidTags.f_13131_);
+-      this.f_19801_.clear();
```
pupnewfster: Same as before, I don't see much of a benefit to be patching out clearing the old set rather than just setting the type on eyes and letting the vanilla set be cleared.

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -287,6 +382,14 @@
           }
        } else {
           return null;
+@@ -2408,6 +_,7 @@
+       return this.f_19821_;
+    }
+ 
++   @Deprecated // FORGE: Use FluidType sensitive version
```
pupnewfster: It is somewhat inconsistent in this class whether you are using `Forge` or `FORGE` for comments by deprecations

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -153,6 +227,18 @@
        }
  
     }
+ 
++   @Deprecated // FORGE: Use isEyeInFluidType instead
```
pupnewfster: It is somewhat inconsistent in this class whether you are using `Forge` or `FORGE` for comments by deprecations.

Also I feel like rather than having patches replacing calls for water and lava to this all over the place it would be less maintenance heavy to instead patch it to reroute calls to vanilla types like so:
```java
if (p_204030_ == FluidTags.WATER) {
   return isEyeInFluidType(net.minecraftforge.common.ForgeMod.WATER_TYPE.get())
} else if (p_204030_ == FluidTags.LAVA) {
   return isEyeInFluidType(net.minecraftforge.common.ForgeMod.LAVA_TYPE.get())
}
```

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -310,6 +413,108 @@
           });
        }
  
+@@ -2772,9 +_,17 @@
+       this.f_19859_ = this.m_146908_();
+    }
+ 
++   @Deprecated // FORGE: Use no parameter version instead, only for vanilla Tags
```
pupnewfster: It is somewhat inconsistent in this class whether you are using `Forge` or `FORGE` for comments by deprecations

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -310,6 +413,108 @@
...
+    public boolean m_204031_(TagKey<Fluid> p_204032_, double p_204033_) {
++      this.updateFluidHeightAndDoFluidPushing();
++      if(p_204032_ == FluidTags.f_13131_) return this.isInFluidType(net.minecraftforge.common.ForgeMod.WATER_TYPE.get());
++      else if (p_204032_ == FluidTags.f_13132_) return this.isInFluidType(net.minecraftforge.common.ForgeMod.LAVA_TYPE.get());
++      else return false;
++   }
++
++   public void updateFluidHeightAndDoFluidPushing() {
```
pupnewfster, line 429: I am not sure if the exc file supports methods with fewer parameters, but assuming it does I feel like this method should be there so that it gets the same srg name as the method that calls it and redirects to it.

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -310,6 +413,108 @@
...
+-         return false;
++         return;
+       } else {
+          AABB aabb = this.m_20191_().m_82406_(0.001D);
+          int i = Mth.m_14107_(aabb.f_82288_);
+@@ -2789,25 +_,28 @@
+          Vec3 vec3 = Vec3.f_82478_;
+          int k1 = 0;
+          BlockPos.MutableBlockPos blockpos$mutableblockpos = new BlockPos.MutableBlockPos();
++         it.unimi.dsi.fastutil.objects.Object2ObjectMap<net.minecraftforge.fluids.FluidType, org.apache.commons.lang3.tuple.MutableTriple<Double, Vec3, Integer>> interimCalcs = new it.unimi.dsi.fastutil.objects.Object2ObjectArrayMap<>(net.minecraftforge.fluids.FluidType.SIZE.get());
```
pupnewfster: Should this be `- 1` for the size? Given I presume the EMPTY type will never be added into this map?

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -310,6 +413,108 @@
...
++         this.setFluidTypeHeight(fluidType, interim.getLeft());
++         });
+       }
+    }
+ 
+@@ -2848,6 +_,7 @@
+       return !this.f_19853_.m_151572_(i, k, j, l);
+    }
+ 
++   @Deprecated // FORGE: Use getFluidTypeHeight instead
```
pupnewfster: It is somewhat inconsistent in this class whether you are using `Forge` or `FORGE` for comments by deprecations

Also I feel like rather than having patches replacing calls for water and lava to this all over the place it would be less maintenance heavy to instead patch it to reroute calls to vanilla types like so:
```java
if (p_204037_ == FluidTags.WATER) {
   return getFluidTypeHeight(net.minecraftforge.common.ForgeMod.WATER_TYPE.get())
} else if (p_204037_ == FluidTags.LAVA) {
   return getFluidTypeHeight(net.minecraftforge.common.ForgeMod.LAVA_TYPE.get())
}
```

patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
```diff
@@ -29,6 +35,25 @@
...
+       return this.m_6336_() == MobType.f_21641_;
+    }
+@@ -337,8 +_,9 @@
+             }
+          }
+ 
+-         if (this.m_204029_(FluidTags.f_13131_) && !this.f_19853_.m_8055_(new BlockPos(this.m_20185_(), this.m_20188_(), this.m_20189_())).m_60713_(Blocks.f_50628_)) {
+-            boolean flag1 = !this.m_6040_() && !MobEffectUtil.m_19588_(this) && (!flag || !((Player)this).m_150110_().f_35934_);
++         boolean canDrown = this.canDrownInFluidType(this.getEyeInFluidType());
++         if (!this.getEyeInFluidType().isAir() && canDrown && !this.f_19853_.m_8055_(new BlockPos(this.m_20185_(), this.m_20188_(), this.m_20189_())).m_60713_(Blocks.f_50628_)) {
```
pupnewfster: This is a minor functionality change as for water `canDrown` will be false for undead, but vanilla doesn't check that until checking about updating the air supply and applying drowning damage. Which means that an undead mob that is a passanger of a vehicle that cannot be ridden in water won't be kicked off with this change, whereas it would be kicked off the vehicle in vanilla.

Though that also then checks `canBeRiddenInWater` which potentially needs a more generic `canBeRiddenIn(FluidType)` method?

patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
```diff
@@ -350,25 +375,27 @@
...
        }
  
        this.f_19812_ = true;
 +      net.minecraftforge.common.ForgeHooks.onLivingJump(this);
     }
  
++   @Deprecated // FORGE: use sinkInFluid instead
     protected void m_21208_() {
 -      this.m_20256_(this.m_20184_().m_82520_(0.0D, (double)-0.04F, 0.0D));
 +      this.m_20256_(this.m_20184_().m_82520_(0.0D, (double)-0.04F * this.m_21051_(net.minecraftforge.common.ForgeMod.SWIM_SPEED.get()).m_22135_(), 0.0D));
```
pupnewfster: Either this method should not be patched at all due to being deprecated or it should be patched to call `sinkInFluid(WATER)`

src/main/java/net/minecraftforge/fluids/FluidType.java
```diff
@@ -0,0 +1,1128 @@
...
+     * Returns a sound to play when a certain action is performed at a
+     * position. If no sound is present, then the sound will be {@code null}.
+     *
+     * @param getter the getter which can get the fluid
+     * @param pos the position of the fluid
+     * @param action the action being performed
+     * @return the sound to play when performing the action
+     */
+    @Nullable
+    public SoundEvent getSound(BlockGetter getter, BlockPos pos, SoundAction action)
```
pupnewfster: Is it potentially worth adding a `@Nullable Player player` parameter? Given both places that call this have one so then it would provide more context to the fluid type if it wanted to do player specific sound differences.

patches/minecraft/net/minecraft/world/level/block/ConcretePowderBlock.java.patch
```diff
@@ -0,0 +1,44 @@
+--- a/net/minecraft/world/level/block/ConcretePowderBlock.java
++++ b/net/minecraft/world/level/block/ConcretePowderBlock.java
+@@ -20,7 +_,7 @@
+    }
+ 
+    public void m_48792_(Level p_52068_, BlockPos p_52069_, BlockState p_52070_, BlockState p_52071_, FallingBlockEntity p_52072_) {
+-      if (m_52080_(p_52068_, p_52069_, p_52071_)) {
++      if (m_52080_(p_52068_, p_52069_, p_52070_)) {
```
pupnewfster: This feels like a bug changing which parameter is checked. And if it isn't this definitely needs a comment so that during ports and stuff it doesn't get accidentally broken.

patches/minecraft/net/minecraft/world/level/block/ConcretePowderBlock.java.patch
```diff
@@ -0,0 +1,44 @@
...
+ 
+-   private static boolean m_52064_(BlockGetter p_52065_, BlockPos p_52066_) {
++   private static boolean touchesLiquid(BlockGetter p_52065_, BlockPos p_52066_, BlockState state) {
+       boolean flag = false;
+       BlockPos.MutableBlockPos blockpos$mutableblockpos = p_52066_.m_122032_();
+ 
+       for(Direction direction : Direction.values()) {
+          BlockState blockstate = p_52065_.m_8055_(blockpos$mutableblockpos);
+-         if (direction != Direction.DOWN || m_52088_(blockstate)) {
++         if (direction != Direction.DOWN || state.canBeHydrated(p_52065_, p_52066_, p_52065_.m_6425_(blockpos$mutableblockpos), blockpos$mutableblockpos)) {
```
pupnewfster: This would be more readable as `p_52065_.getFluidState(blockpos$mutableblockpos)` -> `blockstate.getFluidState()`

patches/minecraft/net/minecraft/world/level/block/ConcretePowderBlock.java.patch
```diff
@@ -0,0 +1,44 @@
...
+             blockpos$mutableblockpos.m_122159_(p_52066_, direction);
+             blockstate = p_52065_.m_8055_(blockpos$mutableblockpos);
+-            if (m_52088_(blockstate) && !blockstate.m_60783_(p_52065_, p_52066_, direction.m_122424_())) {
++            if (state.canBeHydrated(p_52065_, p_52066_, p_52065_.m_6425_(blockpos$mutableblockpos), blockpos$mutableblockpos) && !blockstate.m_60783_(p_52065_, p_52066_, direction.m_122424_())) {
```
pupnewfster: This would be more readable as `p_52065_.getFluidState(blockpos$mutableblockpos)` -> `blockstate.getFluidState()`.

patches/minecraft/net/minecraft/world/level/pathfinder/WalkNodeEvaluator.java.patch
```diff
@@ -9,12 +9,51 @@
        }
  
        double d0 = this.m_142213_(new BlockPos(p_77641_.f_77271_, p_77641_.f_77272_, p_77641_.f_77273_));
+@@ -471,6 +_,11 @@
+                if (l != 0 || j1 != 0) {
+                   p_77609_.m_122178_(i + l, j + i1, k + j1);
+                   BlockState blockstate = p_77608_.m_8055_(p_77609_);
++                  BlockPathTypes blockPathType = blockstate.getAdjacentBlockPathType(p_77608_, p_77609_, null, p_77610_);
++                  if (blockPathType != null) return blockPathType;
++                  FluidState fluidState = p_77608_.m_6425_(p_77609_);
```
pupnewfster: This can just be `blockstate.getFluidState()` to avoid having to query the world again.

patches/minecraft/net/minecraft/world/level/pathfinder/WalkNodeEvaluator.java.patch
```diff
@@ -9,12 +9,51 @@
...
++                  if (fluidPathType != null) return fluidPathType;
+                   if (blockstate.m_60713_(Blocks.f_50128_)) {
+                      return BlockPathTypes.DANGER_CACTUS;
+                   }
+@@ -483,7 +_,7 @@
+                      return BlockPathTypes.DANGER_FIRE;
+                   }
+ 
+-                  if (p_77608_.m_6425_(p_77609_).m_205070_(FluidTags.f_13131_)) {
++                  if (fluidState.m_205070_(FluidTags.f_13131_)) {
```
pupnewfster: While this is a minor performance improvement compared to getting the fluid state twice, I am unsure if the minor benefit is worth increasing the patch size by changing out the reference here.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -803,30 +803,25 @@ public static LootTable loadLootTable(Gson gson, ResourceLocation name, JsonElem
...
+    {
+        ResourceLocation name = ForgeRegistries.FLUIDS.getKey(fluid);
+        if (name.equals(new ResourceLocation("empty")))
+            return ForgeMod.EMPTY_TYPE.get();
+        if (name.equals(new ResourceLocation("water")) || name.equals(new ResourceLocation("flowing_water")))
+            return ForgeMod.WATER_TYPE.get();
+        if (name.equals(new ResourceLocation("lava")) || name.equals(new ResourceLocation("flowing_lava")))
+            return ForgeMod.LAVA_TYPE.get();
+        if (name.equals(ForgeMod.MILK.getId()) || name.equals(ForgeMod.FLOWING_MILK.getId()))
+            return ForgeMod.MILK_TYPE.get();
```
pupnewfster: Is there a reason this is comparing based on name instead of instance, especially for empty, water, and lava?

patches/minecraft/net/minecraft/client/renderer/ScreenEffectRenderer.java.patch
```diff
@@ -15,10 +15,12 @@
...
  
        if (!p_110719_.f_91074_.m_5833_()) {
-          if (p_110719_.f_91074_.m_204029_(FluidTags.f_13131_)) {
+-         if (p_110719_.f_91074_.m_204029_(FluidTags.f_13131_)) {
++         if (player.isEyeInFluidType(net.minecraftforge.common.ForgeMod.WATER_TYPE.get())) {
 +            if (!net.minecraftforge.client.ForgeHooksClient.renderWaterOverlay(player, p_110720_))
              m_110725_(p_110719_, p_110720_);
-          }
+-         }
++         } else if (!player.getEyeInFluidType().isAir()) net.minecraftforge.client.RenderProperties.get(player.getEyeInFluidType()).renderOverlay(p_110719_, p_110720_);
```
pupnewfster: Put the `else if` on the next line so that there isn't a patch for removing the `}` in order to add it

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -130,9 +143,160 @@
...
+                    .descriptionId("block.minecraft.air")
+                    .motionScale(1D)
+                    .canPushEntity(false)
+                    .canSwim(false)
+                    .fallDistanceModifier(1F)
+                    .pathType(null)
+                    .adjacentPathType(null)
+                    .density(0)
+                    .temperature(0)
+                    .lightLevel(0)
```
pupnewfster: There is no reason for this to manually be specifying `lightLevel(0)` when that is the default

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -130,9 +143,160 @@
...
+                {
+                    consumer.accept(new IFluidTypeRenderProperties()
+                    {
+                        @Override
+                        public int getColorTint()
+                        {
+                            return 0;
+                        }
+                    });
+                }
```
pupnewfster: Is there any reason to not just let this default to the white default of `getColorTint` and then the fluid properties defaulting to the DUMMY that otherwise matches the fluid properties for empty?

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -130,9 +143,160 @@
...
+            });
+    public static final RegistryObject<FluidType> WATER_TYPE = VANILLA_FLUID_TYPES.register("water", () ->
+            new FluidType(FluidType.Properties.create()
+                    .descriptionId("block.minecraft.water")
```
pupnewfster, line 179: Would it maybe be of use to have a helper that takes a `Block` and then calls `Block#getDescriptionId` on it to grab the description from there?

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -266,14 +431,44 @@ public void missingSoundMapping(MissingMappingsEvent event)
...
-        if (enableMilkFluid && event.getRegistryKey().equals(ForgeRegistries.Keys.FLUIDS))
+        if (enableMilkFluid)
         {
-            // set up attributes
-            FluidAttributes.Builder attributesBuilder = FluidAttributes.builder(new ResourceLocation("forge", "block/milk_still"), new ResourceLocation("forge", "block/milk_flowing")).density(1024).viscosity(1024);
-            ForgeFlowingFluid.Properties properties = new ForgeFlowingFluid.Properties(MILK, FLOWING_MILK, attributesBuilder).bucket(() -> Items.MILK_BUCKET);
-            // register fluids
-            event.register(ForgeRegistries.Keys.FLUIDS, MILK.getId(), () -> new ForgeFlowingFluid.Source(properties));
-            event.register(ForgeRegistries.Keys.FLUIDS, FLOWING_MILK.getId(), () -> new ForgeFlowingFluid.Flowing(properties));
+            if (event.getRegistryKey().equals(ForgeRegistries.Keys.FLUID_TYPES))
```
pupnewfster: I feel like this check is unneeded as it is performed automatically by the event.register helper

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -480,16 +481,43 @@ default BlockState getStateAtViewpoint(BlockState state, BlockGetter level, Bloc
...
-     * @return the PathNodeType
+     * @param state the state of the block
+     * @param level the level which contains this block
+     * @param pos the position of the block
+     * @param mob the mob currently pathfinding, may be {@code null}
+     * @return the path type of this block
      */
     @Nullable
-    default BlockPathTypes getAiPathNodeType(BlockState state, BlockGetter level, BlockPos pos, @Nullable Mob entity)
+    default BlockPathTypes getBlockPathType(BlockState state, BlockGetter level, BlockPos pos, @org.jetbrains.annotations.Nullable Mob mob)
```
pupnewfster: The nullable annotation is already imported, no need for the FQN here

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -480,16 +481,43 @@ default BlockState getStateAtViewpoint(BlockState state, BlockGetter level, Bloc
...
+     *
+     * @param state the state of the block
+     * @param level the level which contains this block
+     * @param pos the position of the block
+     * @param mob the mob currently pathfinding, may be {@code null}
+     * @param originalType the path type of the source the entity is on
+     * @return the path type of this block
+     */
+    @Nullable
+    default BlockPathTypes getAdjacentBlockPathType(BlockState state, BlockGetter level, BlockPos pos, @org.jetbrains.annotations.Nullable Mob mob, BlockPathTypes originalType)
```
pupnewfster: The nullable annotation is already imported, no need for the FQN here

src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -816,4 +844,21 @@ default boolean supportsExternalFaceHiding(BlockState state)
         }
         return true;
     }
+
+    /**
+     * Returns whether the block can be hydrated by a fluid.
+     *
+     * <p>Hydration is an arbitrary word which depends on the block.
```
pupnewfster, line 851: Probably worth giving some examples of what it might mean

src/main/java/net/minecraftforge/common/extensions/IForgeBlockState.java
```diff
@@ -541,25 +541,36 @@ default boolean isBurning(BlockGetter level, BlockPos pos)
...
      *
-     * @return the PathNodeType
+     * @param level the level which contains this block
+     * @param pos the position of the block
+     * @param mob the mob currently pathfinding, may be {@code null}
+     * @return the path type of this block
      */
     @Nullable
-    default BlockPathTypes getBlockPathType(BlockGetter level, BlockPos pos)
+    default BlockPathTypes getBlockPathType(BlockGetter level, BlockPos pos, @org.jetbrains.annotations.Nullable Mob mob)
```
pupnewfster: The nullable annotation is already imported, no need for the FQN here

src/main/java/net/minecraftforge/common/extensions/IForgeBlockState.java
```diff
@@ -541,25 +541,36 @@ default boolean isBurning(BlockGetter level, BlockPos pos)
...
-     * @return the PathNodeType
+     * @param level the level which contains this block
+     * @param pos the position of the block
+     * @param mob the mob currently pathfinding, may be {@code null}
+     * @param originalType the path type of the source the entity is on
+     * @return the path type of this block
      */
     @Nullable
-    default BlockPathTypes getBlockPathType(BlockGetter level, BlockPos pos, @Nullable Mob mob)
+    default BlockPathTypes getAdjacentBlockPathType(BlockGetter level, BlockPos pos, @org.jetbrains.annotations.Nullable Mob mob, BlockPathTypes originalType)
```
pupnewfster: The nullable annotation is already imported, no need for the FQN here

src/main/java/net/minecraftforge/common/extensions/IForgeBlockState.java
```diff
@@ -676,4 +687,20 @@ default boolean supportsExternalFaceHiding()
     {
         return self().getBlock().supportsExternalFaceHiding(self());
     }
+
+    /**
+     * Returns whether the block can be hydrated by a fluid.
+     *
+     * <p>Hydration is an arbitrary word which depends on the block.
```
pupnewfster, line 694: Probably worth giving some examples of what it might mean

src/main/java/net/minecraftforge/common/extensions/IForgeLivingEntity.java
```diff
@@ -0,0 +1,69 @@
...
+    default LivingEntity self()
+    {
+        return (LivingEntity) this;
+    }
+
+    @Override
+    default boolean canSwimInFluidType(FluidType type)
+    {
+        if (self().isSensitiveToWater() && type == ForgeMod.WATER_TYPE.get()) return false;
+        else return IForgeEntity.super.canSwimInFluidType(type);
```
pupnewfster: I think this might read better as:
```java
if (type == ForgeMod.WATER_TYPE.get())
    return !self().isSensitiveToWater();
return IForgeEntity.super.canSwimInFluidType(type);
```

Given I don't think there is any functional difference for water.

src/main/java/net/minecraftforge/common/extensions/IForgeLivingEntity.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * Performs what to do when an entity attempts to go up or "jump" in a fluid.
+     *
+     * @param type the type of the fluid
+     */
+    default void jumpInFluid(FluidType type)
+    {
+        self().setDeltaMovement(self().getDeltaMovement().add(0.0D, (double)0.04F * self().getAttribute(ForgeMod.SWIM_SPEED.get()).getValue(), 0.0D));
```
pupnewfster: Is there a reason (other than the patches didn't do so before) not to have this just be `getAttributeValue`?

src/main/java/net/minecraftforge/common/extensions/IForgeLivingEntity.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * Performs what to do when an entity attempts to go down or "sink" in a fluid.
+     *
+     * @param type the type of the fluid
+     */
+    default void sinkInFluid(FluidType type)
+    {
+        self().setDeltaMovement(self().getDeltaMovement().add(0.0D, (double)-0.04F * self().getAttribute(ForgeMod.SWIM_SPEED.get()).getValue(), 0.0D));
```
pupnewfster: Is there a reason (other than the patches didn't do so before) not to have this just be `getAttributeValue`?

src/main/java/net/minecraftforge/common/extensions/IForgeLivingEntity.java
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Returns whether the entity can drown in the fluid.
+     *
+     * @param type the type of the fluid
+     * @return {@code true} if the entity can drown in the fluid, {@code false} otherwise
+     */
+    default boolean canDrownInFluidType(FluidType type)
+    {
+        if (type == ForgeMod.WATER_TYPE.get()) return !self().canBreatheUnderwater();
+        else return type.canDrownIn(self());
```
pupnewfster: No need for the `else return` this can just be `return` as it is cleaner to read. Also it may be worth considering just making this a ternary expression.

src/main/java/net/minecraftforge/fluids/FluidInteractionRegistry.java
```diff
@@ -0,0 +1,179 @@
...
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
+
+    static
+    {
```
pupnewfster, line 75: Is water + lava -> stone just missing here or is there some reason it doesn't fall under this interaction system

src/main/java/net/minecraftforge/fluids/FluidType.java
```diff
@@ -0,0 +1,1128 @@
...
+     * @param stack the stack holding the fluid being placed
+     * @return {@code true} if this fluid should be vaporized on placement, {@code false} otherwise
+     *
+     * @see BucketItem#emptyContents(Player, Level, BlockPos, BlockHitResult)
+     */
+    public boolean isVaporizedOnPlacement(Level level, BlockPos pos, FluidStack stack)
+    {
+        BlockState state = this.getBlockForFluidState(level, pos, this.getStateForPlacement(level, pos, stack));
+        if (state == null) return false;
+        else return level.dimensionType().ultraWarm() && state.getMaterial() == Material.WATER;
```
pupnewfster: I feel like checking the dimension for the default before looking up the state from the world might have slightly better performance.

Lanse505:
My plan is to start addressing these required changes this week.
Hopefully they'll be either discussed or handled by wednesday evening is the plan.

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
Lanse505: I defer to @ChampionAsh5357 on this one

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
Lanse505: I defer to @ChampionAsh5357 on this one

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
Lanse505: Will fix, standardizing your comments to ``// FORGE: ...``

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
Lanse505: Will fix, standardizing your comments to ``// FORGE: ...``

src/main/java/net/minecraftforge/common/ForgeMod.java
ChampionAsh5357: Not really, a block isn't always available.

src/main/java/net/minecraftforge/fluids/FluidInteractionRegistry.java
ChampionAsh5357: Because the fluid interaction for stone was defined in `LavaFluid` instead of in the `LiquidBlock` so I missed it. Need to look over still.

patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
ChampionAsh5357: Need to look into this more.

KingLemming:
From what I can tell, this is pretty damn solid. I say merge it and let's get any undiscovered bugs hammered out in prod. MVP achieved.

TheCurle:
You're the boss, boss.

PR #8699, opened 2022-06-09
Fix Reach Distance / Attack Range being clamped at 6.0

### LINKED ISSUE #8689: Forge 41.0.4 for 1.19 issue with Forge.REACH_DISTANCE.get() attribute modifier
**Minecraft Version:** 1.19

**Forge Version:** 41.0.4

**Description of issue:**

The Forge.REACH_DISTANCE.get() attribute modifier does not appear to be working correctly. The block inside the normal tool reach distance break as expected. Block outside of the normal tool reach distance break and reappear immediately. The tool does show the +X Reach Distance tool tip, and the same code works fine in Forge 40.1.0. The logs do not show any warnings or errors. It seems that the server side is not applying the reach modifier as expected.

Shadows-of-Fire:
In 1.19 mojang added `ServerGamePacketListenerImpl.MAX_INTERACTION_DISTANCE` and added checks against it that need to be removed to support extended reach/range.  This PR removes the checks and deprecates the field.

Fixes #8689

### DIFF
--- a/patches/minecraft/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
+++ b/patches/minecraft/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
@@ -1,5 +1,13 @@
 --- a/net/minecraft/server/network/ServerGamePacketListenerImpl.java
 +++ b/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+@@ -171,6 +_,7 @@
+ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, ServerGamePacketListener {
+    static final Logger f_9744_ = LogUtils.getLogger();
+    private static final int f_143608_ = 15000;
++   @Deprecated // Forge: See Reach Distance / Attack Range attributes.
+    public static final double f_215198_ = Mth.m_144952_(6.0D);
+    private static final int f_215199_ = -1;
+    public final Connection f_9742_;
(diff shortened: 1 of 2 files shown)


patches/minecraft/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch
```diff
@@ -1,5 +1,13 @@
 --- a/net/minecraft/server/network/ServerGamePacketListenerImpl.java
 +++ b/net/minecraft/server/network/ServerGamePacketListenerImpl.java
+@@ -171,6 +_,7 @@
+ public class ServerGamePacketListenerImpl implements ServerPlayerConnection, ServerGamePacketListener {
+    static final Logger f_9744_ = LogUtils.getLogger();
+    private static final int f_143608_ = 15000;
++   @Deprecated // Forge: See Reach Distance / Attack Range attributes.
```
sciwhiz12: This should be a javadoc comment with (fully-qualified) references to the attributes in `ForgeMod`.

sciwhiz12 (changes requested):
Tested that it fixes the issue. Just need a small change to the added deprecation.

---

Please rename this PR to have a title fitting its contents, and please add `fixes #8689` to the PR description so GitHub links this PR to the issue.

Shadows-of-Fire:
PR and comment updated.

sciwhiz12 (approved):

PR #8700, opened 2022-06-09
Replace support bot with support action

coehlrich:
The public instance of the support bot was shutdown and replaced with a github action

Issue mentioning the shutdown: https://github.com/dessant/support-requests/issues/8

Here's an example issue using the github action: https://github.com/coehlrich/MinecraftForge/issues/1


.github/workflows/support.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+    - uses: dessant/support-requests@v2
+      with:
+        support-label: Forum
+        issue-comment: |
+            :wave: We use the issue tracker exclusively for final bug reports and feature requests.
+            However, this issue appears to be better suited for the
+            [Forge Support Forums](http://www.minecraftforge.net/forum/forum/18-support-bug-reports/).
+            Please create a new topic on the support forum with this issue, and the conversation can
+            continue there.
```
sciwhiz12: Could we change the support message to the following instead:

```
:wave: We use the issue tracker exclusively for final bug reports and feature requests. 
However, this issue appears to be better suited for the 
[Forge Support Forums](https://forums.minecraftforge.net/) or 
[Forge Discord](https://discord.gg/forge). Please create a new topic on the support forum
with this issue or ask in the `#tech-support` channel in the Discord server, and the 
conversation can continue there.
```

Additionally, this should use the `>` (folded style[^1]) instead of `|` (literal style[^2]), to not preserve any of the linebreaks[^3] (so it's a single paragraph in the comment) as was done in the original `support.yml`.

[^1]: [YAML Ain’t Markup Language (YAML™) Version 1.2, 8.1.3. Folded Style](https://yaml.org/spec/1.2/spec.html#style/block/folded)
[^2]: [YAML Ain’t Markup Language (YAML™) Version 1.2, 8.1.2. Literal Style](https://yaml.org/spec/1.2/spec.html#style/block/literal)
[^3]: [StackOverflow, _How do I break a string in YAML over multiple lines?_, answer by Steve Bennett](https://stackoverflow.com/a/21699210/14416954)

sciwhiz12 (changes requested):
Aside from my requested changes, this looks fine to me. The maintainer pulling this will need to (re)enable GitHub Actions on the repository for this to have any effect, though.

.github/workflows/support.yml
coehlrich: Changed the message, made it folded style, and the updated message is on the test issue in the pr description

sciwhiz12 (approved):

PR #8703, opened 2022-06-09
Handle modded packets on the network thread

### LINKED ISSUE #8642: SimpleChannel messages are handled on the main thread
**Minecraft Version:** 1.18.2
**Forge Version:** 40.1.0

**Description of issue:**

SimpleChannel (and the rest of Forge's networking stack) reads, writes and handles messages on the main thread (render/server thread, depending on side) rather than on the dedicated Netty threads.

**Steps to Reproduce:**
 1. Create and register a simple channel and message, and add a breakpoint inside its encoder, decoder and handler. Alternatively, add breakpoints inside Forge's built-in `HandshakeMessages`.
 2. Send your registered message.
 3. Observe the breakpoints are hit on the render or server thread, and not the network thread as expected.

**Further comments:**
It's not clear whether this is intended or not. `NetworkEvent.Context.enqueueWork` exists, which implies the initial packet handler is expected to be called off-thread. However, [the patch to handle packets server-side](https://github.com/MinecraftForge/MinecraftForge/blob/727c7780b30c4e5ded10fd6689b9b968266ac224/patches/minecraft/net/minecraft/server/network/ServerGamePacketListenerImpl.java.patch#L83-L86) explicitly runs the handler on the main thread (and has done so since the 1.13 networking rewrite).

If this is a bug, fixing this may have to wait until 1.19, as I suspect there's quite a lot of mods which rely on the current behaviour.

SquidDev:
Fixes #8642. This ensures custom modded packets are both read and handled on the network thread by default, rather than the main server/client thread.

Tested on both single player and a server, with Netty's Advanced leak detection enabled. Custom packets work as expected (tested with [a mod of mine](https://github.com/cc-tweaked/CC-Tweaked)), and it does not appear to introduce any `ByteBuf` leaks.

I did also look at changing the custom packets to write their contents on the network thread (they are currently written when the packet is created instead), but this would require a more invasive change, and I wasn't comfortable doing that without further discussion.

### DIFF
--- a/src/main/java/net/minecraftforge/network/simple/SimpleChannel.java
+++ b/src/main/java/net/minecraftforge/network/simple/SimpleChannel.java
@@ -8,25 +8,16 @@ package net.minecraftforge.network.simple;
 import io.netty.buffer.Unpooled;
 import net.minecraft.client.Minecraft;
 import net.minecraft.network.Connection;
-import net.minecraft.network.protocol.Packet;
 import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.network.protocol.Packet;
 import net.minecraftforge.network.NetworkDirection;
 import net.minecraftforge.network.NetworkEvent;
 import net.minecraftforge.network.NetworkInstance;
 import net.minecraftforge.network.PacketDistributor;
 import org.apache.commons.lang3.tuple.Pair;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.function.BiConsumer;
-import java.util.function.Consumer;
-import java.util.function.Function;
-import java.util.function.IntSupplier;
-import java.util.function.Supplier;
+import java.util.*;
+import java.util.function.*;
 
 @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
 public class SimpleChannel
@@ -252,7 +243,7 @@ public class SimpleChannel
          * @return The message builder, for chaining.
          * @deprecated Use {@link #consumerMainThread(BiConsumer)} or {@link #consumerNetworkThread(BiConsumer)}.
          */
-        @Deprecated(forRemoval = true)
+        @Deprecated(forRemoval = true, since = "1.19")
         public MessageBuilder<MSG> consumer(BiConsumer<MSG, Supplier<NetworkEvent.Context>> consumer) {
             consumerMainThread(consumer);
             return this;
@@ -320,6 +311,7 @@ public class SimpleChannel
          * @return The message builder, for chaining.
          * @deprecated Use {@link #consumerMainThread(BiConsumer)} or {@link #consumerNetworkThread(BiConsumer)}.
          */
+        @Deprecated(forRemoval = true, since = "1.19")
         public MessageBuilder<MSG> consumer(ToBooleanBiFunction<MSG, Supplier<NetworkEvent.Context>> handler) {
             consumerMainThread(handler::applyAsBool);
             return this;
--- a/src/main/java/net/minecraftforge/network/simple/SimpleChannel.java
+++ b/src/main/java/net/minecraftforge/network/simple/SimpleChannel.java
@@ -183,11 +183,11 @@ public class SimpleChannel
 
         /**
          * Set the message encoder, which writes this message to a {@link FriendlyByteBuf}.
-         *
-         * The encoder is called <em>immediately</em> {@link #send(PacketDistributor.PacketTarget, Object) when the
+         * <p>
+         * The encoder is called <em>immediately</em> {@linkplain #send(PacketDistributor.PacketTarget, Object) when the
          * packet is sent}. This means encoding typically occurs on the main server/client thread rather than on the
          * network thread.
-         *
+         * <p>
          * However, this behaviour should not be relied on, and the encoder should try to be thread-safe and not
          * interact with the current game state.
          *
@@ -201,10 +201,10 @@ public class SimpleChannel
 
         /**
          * Set the message decoder, which reads the message from a {@link FriendlyByteBuf}.
-         *
+         * <p>
          * The decoder is called when the message is received on the network thread. The decoder should not attempt to
-         * access or mutate any game state, deferring that until the {@link #consumer(ToBooleanBiFunction) the message
-         * is handled}.
+         * access or mutate any game state, deferring that until the {@linkplain #consumer(ToBooleanBiFunction) the
+         * message is handled}.
          *
          * @param decoder The message decoder.
          * @return The message builder, for chaining.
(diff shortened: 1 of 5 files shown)


sciwhiz12 (changes requested):
While I know @cpw has commented on the issue that packets being handled on the network thread is the original design, I'd like for an example situation to be provided in which handling the packet on the network thread is more preferrable than handling it on the render/server thread. Other than that, this PR looks fine to me.

---

Please add `fixes #8642` to your PR description so this PR is automatically linked to the issue.

SquidDev:
What originally caused me to spot this was that packet decoding is also handled on the main thread. I compress some of the bulkier packets when sending[^1] and the corresponding decompression was showing up in profiles during some stress tests. If that can be shunted off onto the network thread, it does result in moderate FPS improvements.

[^1]: The server-side compression will still be done on the main thread as mentioned in the OP, but that's much harder to fix.

sciwhiz12:
Hmmm, so this would introduce an asymmetry for which thread packets are encoded and decoded -- packets would be encoded on the main thread while decoded on the network thread. Maybe we should document this behavior somewhere (perhaps as javadocs on `SimpleChannel.MessageBuilder#encoder` and `SimpleChannel.MessageBuilder#decoder`?) so developers which somehow rely on this behavior are aware of this.

src/main/java/net/minecraftforge/network/simple/SimpleChannel.java
```diff
@@ -181,11 +181,34 @@ private static <MSG> MessageBuilder<MSG> forType(final SimpleChannel channel, fi
...
 
+        /**
+         * Set the message encoder, which writes this message to a {@link FriendlyByteBuf}.
+         *
+         * The encoder is called <em>immediately</em> {@link #send(PacketDistributor.PacketTarget, Object) when the
+         * packet is sent}. This means encoding typically occurs on the main server/client thread rather than on the
+         * network thread.
+         *
+         * However, this behaviour should not be relied on, and the encoder should try to be thread-safe and not
+         * interact with the current game state.
```
sciwhiz12: Paragraphs other than the first should either be separated by or surrounded with `<p>` tags. (I personally prefer the latter.) Additionally, `{@linkplain}` should be used when the label would normally not be in monospace.

```suggestion
         * <p>The encoder is called <em>immediately</em> {@linkplain #send(PacketDistributor.PacketTarget, Object) when the
         * packet is sent}. This means encoding typically occurs on the main server/client thread rather than on the
         * network thread.</p>
         *
         * <p>However, this behaviour should not be relied on, and the encoder should try to be thread-safe and not
         * interact with the current game state.</p>
```

src/main/java/net/minecraftforge/network/simple/SimpleChannel.java
```diff
@@ -181,11 +181,34 @@ private static <MSG> MessageBuilder<MSG> forType(final SimpleChannel channel, fi
...
             this.encoder = encoder;
             return this;
         }
 
+        /**
+         * Set the message decoder, which reads the message from a {@link FriendlyByteBuf}.
+         *
+         * The decoder is called when the message is received on the network thread. The decoder should not attempt to
+         * access or mutate any game state, deferring that until the {@link #consumer(ToBooleanBiFunction) the message
+         * is handled}.
```
sciwhiz12: Paragraphs other than the first should either be separated by or surrounded with `<p>` tags. (I personally prefer the latter.) Additionally, `{@linkplain}` should be used when the label would normally not be in monospace.

```suggestion
         * <p>The decoder is called when the message is received on the network thread. The decoder should not attempt to
         * access or mutate any game state, deferring that until the {@linkplain #consumer(ToBooleanBiFunction) the message
         * is handled}.</p>
```

src/main/java/net/minecraftforge/network/simple/SimpleChannel.java
```diff
@@ -223,6 +246,17 @@ public MessageBuilder<MSG> noResponse()
             return this;
         }
 
+        /**
+         * Set the message consumer, which is called once a message has been decoded. This accepts the decoded message
+         * object and the message's context.
+         *
+         * The consumer is called on the network thread, and so should not interact with most game state by default.
+         * {@link NetworkEvent.Context#enqueueWork(Runnable)} can be used to handle the message on the main server or
+         * client thread.
```
sciwhiz12: Paragraphs other than the first should either be separated by or surrounded with <p> tags. (I personally prefer the latter.)

```suggestion
         * <p>The consumer is called on the network thread, and so should not interact with most game state by default.
         * {@link NetworkEvent.Context#enqueueWork(Runnable)} can be used to handle the message on the main server or
         * client thread.</p>
```

sciwhiz12 (changes requested):
Looks fine, just want to fixup the formatting of the javadocs. Also, run the `:forge:javadoc` task to see if the `javadoc` tool complains about the new javadocs.

src/main/java/net/minecraftforge/network/simple/SimpleChannel.java
SquidDev: > Additionally, `{@linkplain}` should be used when the label would normally not be in monospace.

Oh, I wasn't that existed. Oh, that's very nice.

SquidDev:
`./gradlew :forgeJavadoc` seems fine. It gives a single error, but appears unrelated to this change.

```
> Task :forge:javadoc
/home/squid/Documents/software/MinecraftForge/src/main/java/net/minecraftforge/common/world/ModifiableBiomeInfo.java:114: error: @param name not found
             * @param biome A biome that has not had its modified biome data generated yet
                      ^
1 error
```

sciwhiz12 (approved):

src/main/java/net/minecraftforge/network/simple/SimpleChannel.java
```diff
@@ -223,11 +246,44 @@ public MessageBuilder<MSG> noResponse()
             return this;
         }
 
-        public MessageBuilder<MSG> consumer(BiConsumer<MSG, Supplier<NetworkEvent.Context>> consumer) {
```
cpw, line 226: There's no need to binary break here. Mark consumer as @Deprecated and redirect to consumerMainThread in it. Otherwise this looks fine. I need to look at encode/decode offthread too.

cpw (changes requested):

cpw (approved):

SquidDev:
Thanks! ❤️

KingLemming:
So this appears to have completely busted my networking. I think there's a bug introduced here beyond the scope of just fixing a perceived issue.

SquidDev:
@KingLemming Can you clarify what you mean by "busted your networking"? Is there a bug report/reproduction case I can look at?

gigaherz:
We have troubleshooted on one of the internal channels. Was simply a misplaced call to read() from the packet. It didn't break before this PR because of enqueueWork running immediately.

KingLemming:
Can safely ignore, as ghz said.

PR #8705, opened 2022-06-09
Move patch invalidateCaps in Entity

### LINKED ISSUE #8690: [1.19] Entity capabilities invalidated at the wrong time
**Minecraft Version:** 1.19

**Forge Version:** 1.19-41.0.4

**Steps to Reproduce:**
 1. Attach a capability to the player
 2. Jump
 3. Capabilities on the player has now been invalidated

**Description of issue:**

Entity capabilities are invalidated during `processFlappingMovement()` (mojang mappings), instead of on entity removal.
I assume that a patch from `remove(Entity.RemovalReason)` in mc 1.18.2 mistakenly ended up in `processFlappingMovement()` when ported to 1.19.

ferriarnus:
Move misaligned patch in Entity from processFlappingMovement to setRemoved. 
fixes #8690

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/Entity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/Entity.java.patch
@@ -95,14 +95,6 @@
        }
     }
  
-@@ -710,6 +_,7 @@
-             this.m_146850_(GameEvent.f_157815_);
-          }
-       }
-+      this.invalidateCaps();
- 
-    }
- 
 @@ -728,10 +_,10 @@
        int j = Mth.m_14107_(this.f_19825_.f_82480_ - (double)p_216987_);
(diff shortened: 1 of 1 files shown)


sciwhiz12 (changes requested):
Looking at the 1.18.x patch, the `invalidateCaps` method was patched into `Entity#remove`, so this fix should follow that (unless there is a reason for moving it into `Entity#setRemoved` rather then `#remove`).

ferriarnus:
Apologies, it seems I made a mistake. The patch in 1.18 is indeed on "remove", and as it appears nothing changed with it's calls, should stay there.

sciwhiz12 (approved):
Please add `fixes #8690` to your PR description, so GitHub automatically links the issue to this PR.

PR #8706, opened 2022-06-09
Network Capabilities 1.19.x

josephtarbit:
1.19 port of #8224

The aim of this PR is to provide a simple system for capabilities to synchronize data along side vanilla's synchronization of it's own data.

I have also changed a few `ItemStack::matches` checks to use `ItemStack::equals` in order to compare share tags instead of full tags. This is unrelated to the implementation of `INetworkCapability` however I thought I might as well fix it.

### `ItemStack` Capabilities

*Note: it's also now impossible to send your own custom sync packets for arbitrary container menus as there is no way to access a player from a `ContainerSynchronizer` instance.*

### `INetworkCapability`

```java
void writeCapabilities(FriendlyByteBuf out, boolean writeAll);

void readCapabilities(FriendlyByteBuf in);

boolean requiresSync();
```


MagnusHJensen:
My guy has been consistent with this for over 2 years! 
Keep it up 🥂

tmvkrpxl0:
During testing, I discovered that throwing an item into nether portal prints that its number is 0
And going through the portal to pick it up also prints that the number is zero.
However after picking it up, it prints correct number, is this concern?

To be specific, these are what it printed:
*Uses Apple*
[12:26:48] [Server thread/INFO] [ne.mi.de.NetworkCapabilityTest/]: [Dev] Sending new number: 30
[12:26:48] [Render thread/INFO] [ne.mi.de.NetworkCapabilityTest/]: [Dev] Received new number: 30
[12:26:48] [Server thread/INFO] [ne.mi.de.NetworkCapabilityTest/]: [[Apple]] Sending new number: 1
[12:26:48] [Render thread/INFO] [ne.mi.de.NetworkCapabilityTest/]: [[Apple]] Received new number: 1
*Throws apple into nether portal*
[12:26:50] [Server thread/INFO] [ne.mi.de.NetworkCapabilityTest/]: [Apple] Sending new number: 0
[12:26:50] [Render thread/INFO] [ne.mi.de.NetworkCapabilityTest/]: [Apple] Received new number: 0
*Enters nether portal*
[12:26:53] [Server thread/INFO] [ne.mi.de.NetworkCapabilityTest/]: [Apple] Sending new number: 0
[12:26:53] [Server thread/INFO] [ne.mi.de.NetworkCapabilityTest/]: [Dev] Sending new number: 30
[12:26:53] [Render thread/INFO] [ne.mi.de.NetworkCapabilityTest/]: [Apple] Received new number: 0
[12:26:53] [Render thread/INFO] [ne.mi.de.NetworkCapabilityTest/]: [Dev] Received new number: 30
*Picks up apple and uses it*
[12:26:56] [Server thread/INFO] [ne.mi.de.NetworkCapabilityTest/]: [Dev] Sending new number: 31
[12:26:56] [Render thread/INFO] [ne.mi.de.NetworkCapabilityTest/]: [Dev] Received new number: 31
[12:26:56] [Server thread/INFO] [ne.mi.de.NetworkCapabilityTest/]: [[Apple]] Sending new number: 2
[12:26:56] [Render thread/INFO] [ne.mi.de.NetworkCapabilityTest/]: [[Apple]] Received new number: 2

As you can see, Dev's number stays at 30 when passing through the portal, but apple's number temporarily becomes 0

tmvkrpxl0:
oh, never mind, it sends 0 as throwing apple has spawned new item entity and it is giving 0 to it

PR #8710, opened 2022-06-10
[1.19 - Draft/RFC] Rewrite how dynamic (attached) capabilities function

Shadows-of-Fire:
This PR makes major changes to CapabilityProvider, CapabilityDispatcher, and AttachCapabilitiesEvent (as well as a slew of things inbetween) to change how dynamically attached capabilities work.

This is currently a _very large_ work-in-progress and only exists as a PR as this stage to generate constructive feedback about the direction of this system.  None of this has been tested, javadocs are wrong, and I'm sure a thousand things are broken.  That will improve over time.

Big Things:

* Capabilities now have numeric ID's.  These IDs are assigned after registration and are only assigned to actual "registered" caps (not "dead" cap instances that can be created when a mod is not actually providing the required interface).
* AttachCapabilitiesEvent now has specific subclasses for each type of capability, and caps are attached on a per-cap instance, instead of attaching a capability provider.  At the point of attachment, the attached cap can be marked as serializable, or comparable (for itemstack merging).
* CapabilityDispatcher is now only populated when needed, instead of at object construction.  This means AttachCapabilitiesEvent will not fire until something decides it needs to access caps on a specific object.
* Capability comparisons are now limited to capabilities marked as "comparable" instead of those marked as "serializable".  Non-comparable capabilities will be ignored during the process. (I think this doesn't actually work in the current impl, but that's the idea!)
* Attached capabilities are now array-backed, meaning CapabilityDispatcher#getCapability boils down to arr[capId][dirId] instead of calling a list of providers.

There's also more things in there that I am probably forgetting to write about - I'll try to update this as I make changes and receive feedback.  There's also a lot more in here than is typical for a scope of a single PR - if some stuff needs to get dropped, that can happen as well.  The important component is the improvement on attached capabilities.

Things that have been discussed and are not happening:
* Removal of Direction from `ICapabilityProvider#getCapability`.  This will not be occurring, however, there is now `ICapabilityProvider#isDirectionSensitive` where a provider can indicate that it does not support the concept of directions.


src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
```diff
@@ -0,0 +1,116 @@
...
+		if (!this.obj.isDirectionSensitive())
+			sides = new Direction[] { null };
+		AttachedCapability<C> attached = new AttachedCapability<>(id, cap, inst, sides);
+		for (Direction side : sides) {
+			int dKey = side == null ? 6 : side.ordinal();
+			if (caps[cap.getId()][dKey] == null) {
+				caps[cap.getId()][dKey] = attached;
+				this.size++;
+			} else {
+				throw new RuntimeException("Duplicated capabilities!"); // TODO: Change to a better message.
```
alcatrazEscapee: I don't think this should throw an exception here, for one main reason: when this event fires it's not easy to know if such a capability already exists. For example, mod A wants to attach a fluid handler capability to all entities, and mod B that provides an entity with a fluid handler capability already. This needs to support defaulting, or overriding, or first-to-get-there-wins, but *not* runtime crash.

Calling `getCapabilities` and doing the exact logic of this method is very cumbersome and unintuitive:
```java
AttachedCapability<?>[][] caps = event.getCapabilities();
boolean canAttach = true;
for (Direction side : sides) {
    int dKey = side == null ? 6 : side.ordinal();
    if (caps[cap.getId()][dKey] != null) {
        canAttach = false;
        break;
    }
}
if (canAttach) {
    event.addCapability(id, cap, inst, sides)

```

Or worse but less code:
```java
try { event.addCapability(...); }
catch (RuntimeException ignored) {}
```
both of which are annoying.

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	/**
+	 * A unmodifiable view of the capabilities that will be attached to this object.
+	 */
+	public AttachedCapability<?>[][] getCapabilities() {
+		return this.caps;
+	}
+
+	public static final class Items extends AttachCapabilitiesEvent<ItemStack> {
```
alcatrazEscapee: This just seems odd that it's called `Items` and not `ItemStacks` because those are in fact two different things and moving capabilities from item stacks to items would be a huge and unhelpful change.

src/main/java/net/minecraftforge/common/capabilities/AttachedCapability.java
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	@Override
+	public int hashCode() {
+		return this.hash;
+	}
+
+	@Override
+	public boolean equals(Object obj) {
+		return obj instanceof AttachedCapability && obj.hashCode() == this.hashCode();
```
alcatrazEscapee: This doesn't look correct at all but moreover is there a reason these need to implement `hashCode` and `equals` at all instead of relying on identity equality? I don't see any instance in which these are being compared.

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
Shadows-of-Fire: I was struggling to figure out the best method of what to do here - because this method has to return the AttachedCapability instance unless we want the consumer to make that instance on their own.

If we did it that way, we could have this method return a Direction[] which has the sides it was successfully attached to.

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
Shadows-of-Fire: Yeah I'll change that for consistency.

src/main/java/net/minecraftforge/common/capabilities/AttachedCapability.java
Shadows-of-Fire: This is something I forgot to remove - previously I was tracking uniqueCaps in CapabilityDispatcher in a HashSet, but now they're in a map keyed by their ID.

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
alcatrazEscapee: 1. Decide which wins: newly attached capabilities or preexisting ones (I'd lean towards preexisting ones), however having newly attached capabilities would make the implementation a tad cleaner.
2. Either silently overwrite, or silently ignore, as per (1).
3. Return the `AttachedCapability` instance if *at least one* side was attached, otherwise return `null`. If preexisting capabilities win, first filter the `sides` as to only comprise sides that were actually attached.

This seems to me like it solves all the stated problems:

- The caller receives their attached capability instance
- The attached capability's `sides` / `getSides()` can be used to see what sides it *actually* attached to.
- A return value of `null` indicates that the capability didn't attach to anything.

Shadows-of-Fire:
Additionally I would like to ask opinions on a rename of LazyOptional to CapabilityInstance for clarity and conciseness.

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
Shadows-of-Fire: The function is now updated to prefer existing caps over overwriting old ones.  Ideally nobody ever has conflicts, but we can't guarantee that in practice.

josephtarbit:
Yeah I definitely think `LazyOptional` should be renamed, although perhaps to `CapabilityReference` because it's not the instance itself but instead refers to the instance.

KingLemming:
Just curious, what are the benefits of this? Some of the naming I can certainly understand, but offhand I don't have a scenario in mind for numerical IDs being useful.

Shadows-of-Fire:
The numerical id's thing was lex's idea - allows for array access from getCapability which is significantly faster than iterating over a list of cap providers

Technici4n:
So, it won't be possible to have dynamically attached capabilities anymore? What if an entity exposes a capability conditionally, e.g. a Player could expose an `IFluidHandler` only if it has a fluid backpack. That doesn't seem possible anymore.

Or in AE2's case, I want to expose the `IItemHandler` capability only a tile exposes `GenericInternalInventory` (internal AE2 cap), but that can be dependent on the part that is attached on a given side which can change at any time, and should be dynamically re-queried every time.

> The numerical id's thing was lex's idea - allows for array access from getCapability which is significantly faster than iterating over a list of cap providers

Has there been any profile suggesting that this is an issue in practice? And I thought the whole point of LazyOptional was to avoid having to redo the iteration for every access, so why do we care now? It certainly isn't _that_ slow.

Another problem with this PR is that it all it does for the performance of ItemStack capabilities is moving from a generic event to a specialized event. And while that's a good start, I don't think it's going to solve the performance issue. (All you're saving is the generic filter for the generic event.)

Shadows-of-Fire:
> So, it won't be possible to have dynamically attached capabilities anymore? What if an entity exposes a capability conditionally, e.g. a Player could expose an IFluidHandler only if it has a fluid backpack. That doesn't seem possible anymore.  

Not as designed - though I'm not sure why that would be necessary, the cap could always be present and just not be valid when the backpack is not attached, which is legal behavior.  If it is strictly necessary there could be an option of a "complex" cap through attaching a provider, but this is significantly more complex.

> Or in AE2's case, I want to expose the IItemHandler capability only a tile exposes GenericInternalInventory (internal AE2 cap), but that can be dependent on the part that is attached on a given side which can change at any time, and should be dynamically re-queried every time.

See above

> Has there been any profile suggesting that this is an issue in practice? And I thought the whole point of LazyOptional was to avoid having to redo the iteration for every access, so why do we care now? It certainly isn't that slow.

I wasn't given anything, but it was mentioned that it is a desired goal of any such redesign (though not with much context)...  I do recall profiles where getCapability was a pain point from back in the 1.12 era, but I imagine the lazy system likely negates most of that issue since the problem was repeated calls without caching (since LazyOptional wasn't introduced yet).

> Another problem with this PR is that it all it does for the performance of ItemStack capabilities is moving from a generic event to a specialized event. And while that's a good start, I don't think it's going to solve the performance issue. (All you're saving is the generic filter for the generic event.)

This PR does quite a bit for stacks - cap attachment is deferred to initial access of capabilities, rather than on construction.  Additionally, caps will be able to provide their own copy() and compare() functions which should allow developers to alleviate the main contention points w.r.t. stack caps.  If I can get that to work right, anyway.

Technici4n:
> Not as designed - though I'm not sure why that would be necessary, the cap could always be present and just not be valid when the backpack is not attached, which is legal behavior.

If the cap is provided and multiple mods wish to attach the capability when some condition is met, then only the first mod will win, regardless of whether the cap is available or not. For example, imagine two mods that add the exact same fluid backpack. If both try to return a dummy fluid handler then it's not going to work cause one of them will win (even if it's a dummy).

> See above (AE2 case)

So we should just attach an IItemHandler to every single tile in case it might ever happen to expose a GenericInternalInventory? This is not going to work very well if another mod does the same...

> This PR does quite a bit for stacks - cap attachment is deferred to initial access of capabilities, rather than on construction.

This is already done in 1.16+, and has been shown to not be sufficient.

> Additionally, caps will be able to provide their own copy() and compare() functions which should allow developers to alleviate the main contention points w.r.t. stack caps.

Yes, the copy() is what's needed here. I don't think the compare() is particularly useful though, comparison performance is not an issue in the typical modded scenarios I've encountered with AE2, and I suspect many will fail to implement it correctly. I guess you could say the same about copy().

covers1624:
> Attached capabilities are now array-backed

I'm concerned about the potential additional memory usage this could result in. Do we have any numbers on how many capabilities exit in a 'regular' modpack (Kitchen sink for worst case)? With `n` being the number of registered 'non-dead' capabilities, that array will be `n * 8` bytes of ram usage on each game object that supports capabilities, even if that capability will never exist on that object ever. That's not even factoring in the arrays for sided/global likely `n * 8 * 7` to be more accurate, I haven't stared at the implementation much, it's likely these arrays aren't filled until there's stuff there.

I think ids are a good idea overall, but using a faster backing map such as `Int2ObjectArrayMap` (or similar), would be a better fit, but may not provide much more _lookup_ performance than previous versions, however, would provide better memory overhead when no or fewer than the maximum capabilities exist.

'Dynamically attached' capabilities come in several forms.
- A Tile/Entity/whatever decides at some point that _it_ wants to expose a new Capability. Provided that the `getCapability` function remains overridable, I don't see this as a problem. From my knowledge, this is how most mods currently expose caps on Tiles, or at least the ones I've seen. I don't foresee any reasons as to why this should be made an implementation detail, and only removes use cases that are currently possible (AE2's parts being a good example. Same with CBMultipart, these need to be exposed at runtime as parts are added to the space.)
- A mod decides it wants to attach a capability to another mods/vanilla/whatever Tile/Entity/whatever at some point through that object's lifecycle. AFAIK these cases were not supported anyway and should remain unsupported. If a mod wishes to do this, they should tell the object upfront that they want to attach a capability to _that specific object_ not all of that type of object. However, I don't think this is in debate here.

Technici4n:
> A mod decides it wants to attach a capability to another mods/vanilla/whatever Tile/Entity/whatever at some point through that object's lifecycle. AFAIK these cases were not supported anyway and should remain unsupported.

Why would there be such an artificial limitation? It is currently possible since attached caps use ICapabilityProvider, I don't think this was ever said to be unsupported, and in any case AE2 requires this feature. Not for its multiparts, but to be able to expose its generic AEKey-containing inventories to any capability. For example, this is how Applied Mekanistics can turn any AE2 inventory into a gas storage that Mekanism can use: https://github.com/AppliedEnergistics/Applied-Mekanistics/blob/0c579cbc54999ab1a0b0461e3fc8afb69d150055/src/main/java/me/ramidzkh/mekae2/AppliedMekanistics.java#L102.

covers1624:
> It is currently possible since attached caps use ICapabilityProvider, I don't think this was ever said to be unsupported

Right, that's true, I guess I was thinking of attaching new `ICapabilityProvider` instances at runtime.

> AE2 requires this feature. Not for its multiparts, but to be able to expose its generic AEKey-containing inventories to any capability.

I would say that AE's case is pretty special. Ideally, most mods should be filtering what Tiles/Entities/whatever they are putting caps onto, and if they aren't, then they are likely doing something like AE. AE would be a valid use case for this existing feature that is removed by this RFC, AE really should implement this differently through the host tile instead of blindly attaching things to every single tile, but, this would not account for other circumstances where AE's capability is exposed by other mods, these additional 'mapped' capabilities would be left behind.

I think AE's use case should be taken into consideration for this RFC and proper dynamic capabilities kept, perhaps this warrants some form of API for mapping capabilities around, although off hand I'm not seeing many more use cases.

josephtarbit:
I use dynamic capabilities too on the player as my cap instance can be recreated and sometimes not available as it utilises a network connection (which can be re-connected or disconnected etc.). It's not a typical use case but it is one never the less. The alternative would be having a capability which is always attached which wraps an inner instance which isn't *as* nice but still works I suppose.

Shadows-of-Fire:
An alternate design for this system is available at https://github.com/MinecraftForge/MinecraftForge/pull/8744

Shadows-of-Fire:
>>This PR does quite a bit for stacks - cap attachment is deferred to initial access of capabilities, rather than on construction.

>This is already done in 1.16+, and has been shown to not be sufficient.

The insufficiency of this may lie in the fact that the "lazy" implementation was poor, and many actions that did not need to resolve the dispatcher would resolve the dispatcher anyway.

>A Tile/Entity/whatever decides at some point that it wants to expose a new Capability. Provided that the getCapability function remains overridable, I don't see this as a problem. From my knowledge, this is how most mods currently expose caps on Tiles, or at least the ones I've seen. I don't foresee any reasons as to why this should be made an implementation detail, and only removes use cases that are currently possible (AE2's parts being a good example. Same with CBMultipart, these need to be exposed at runtime as parts are added to the space.)

This would not count as a dynamically "attached" cap, simply a dynamically exposed cap from the underlying game object.  This will not be changing nor is it part of the scope of this PR.

>Right, that's true, I guess I was thinking of attaching new ICapabilityProvider instances at runtime.

I wonder if a system where `AttachCapabilitiesEvent` is dropped entirely and providers are added at runtime as needed would be superior to the existing variants.

Technici4n:
> The insufficiency of this may lie in the fact that the "lazy" implementation was poor, and many actions that did not need to resolve the dispatcher would resolve the dispatcher anyway.

I believe that only `getCapability` and `areCapsCompatible` resolve the dispatcher with the current lazy impl, and it's not really possible to avoid one of these.

> I wonder if a system where AttachCapabilitiesEvent is dropped entirely and providers are added at runtime as needed would be superior to the existing variants.

_Maybe_, however keep in mind that this will make porting to this PR (or to the other one) even harder.

Shadows-of-Fire:
Closing in favor of #8744

PR #8712, opened 2022-06-10
[1.19] Re-add PotentialSpawns event

TheMuso:
This commit also includes a test mod, disabled by default, that uses the
re-added event to allow mobs of category type monster to spawn only on hard
difficulty.

### DIFF
--- a/patches/minecraft/net/minecraft/world/level/NaturalSpawner.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/NaturalSpawner.java.patch
@@ -4,7 +4,7 @@
  import net.minecraft.world.phys.Vec3;
  import org.slf4j.Logger;
  
-+// TODO: ForgeEventFactory.getPotentialSpawns, ForgeHooks.canEntitySpawn
++// TODO: ForgeHooks.canEntitySpawn
  public final class NaturalSpawner {
     private static final Logger f_46977_ = LogUtils.getLogger();
     private static final int f_151589_ = 24;
@@ -35,12 +35,13 @@
                                return;
                             }
  
-@@ -266,7 +_,7 @@
+@@ -266,7 +_,8 @@
     }
  
     private static WeightedRandomList<MobSpawnSettings.SpawnerData> m_220443_(ServerLevel p_220444_, StructureManager p_220445_, ChunkGenerator p_220446_, MobCategory p_220447_, BlockPos p_220448_, @Nullable Holder<Biome> p_220449_) {
 -      return m_220455_(p_220448_, p_220444_, p_220447_, p_220445_) ? NetherFortressStructure.f_228517_ : p_220446_.m_223133_(p_220449_ != null ? p_220449_ : p_220444_.m_204166_(p_220448_), p_220445_, p_220447_, p_220448_);
-+      return m_220455_(p_220448_, p_220444_, p_220447_, p_220445_) ? p_220445_.m_220521_().m_175515_(Registry.f_235725_).m_123013_(BuiltinStructures.f_209859_).m_226612_().get(MobCategory.MONSTER).f_210044_() : p_220446_.m_223133_(p_220449_ != null ? p_220449_ : p_220444_.m_204166_(p_220448_), p_220445_, p_220447_, p_220448_);
++      // Forge: Add in potential spawns, and replace hardcoded nether fortress mob list
++      return net.minecraftforge.event.ForgeEventFactory.getPotentialSpawns(p_220444_, p_220447_, p_220448_, m_220455_(p_220448_, p_220444_, p_220447_, p_220445_) ? p_220445_.m_220521_().m_175515_(Registry.f_235725_).m_123013_(BuiltinStructures.f_209859_).m_226612_().get(MobCategory.MONSTER).f_210044_() : p_220446_.m_223133_(p_220449_ != null ? p_220449_ : p_220444_.m_204166_(p_220448_), p_220445_, p_220447_, p_220448_));
     }
  
     public static boolean m_220455_(BlockPos p_220456_, ServerLevel p_220457_, MobCategory p_220458_, StructureManager p_220459_) {
--- a/src/main/java/net/minecraftforge/event/ForgeEventFactory.java
+++ b/src/main/java/net/minecraftforge/event/ForgeEventFactory.java
@@ -18,9 +18,12 @@ import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.ReloadableServerResources;
 import net.minecraft.server.players.PlayerList;
 import net.minecraft.util.RandomSource;
+import net.minecraft.util.random.WeightedRandomList;
 import net.minecraft.world.Container;
+import net.minecraft.world.entity.MobCategory;
 import net.minecraft.world.entity.projectile.Projectile;
 import net.minecraft.world.entity.projectile.ThrownEnderpearl;
+import net.minecraft.world.level.biome.MobSpawnSettings;
 import net.minecraft.world.level.chunk.LevelChunk;
 import net.minecraft.world.level.portal.PortalShape;
 import net.minecraft.world.level.block.state.BlockState;
@@ -800,5 +803,13 @@ public class ForgeEventFactory
     public static void onPostServerTick(BooleanSupplier haveTime, MinecraftServer server)
     {
         MinecraftForge.EVENT_BUS.post(new TickEvent.ServerTickEvent(TickEvent.Phase.END, haveTime, server));
+    }
+
+    public static WeightedRandomList<MobSpawnSettings.SpawnerData> getPotentialSpawns(LevelAccessor level, MobCategory category, BlockPos pos, WeightedRandomList<MobSpawnSettings.SpawnerData> oldList)
+    {
+        LevelEvent.PotentialSpawns event = new LevelEvent.PotentialSpawns(level, category, pos, oldList);
+        if (MinecraftForge.EVENT_BUS.post(event))
+            return WeightedRandomList.create();
+        return WeightedRandomList.create(event.getSpawnerDataList());
     }
 }
--- a/src/main/java/net/minecraftforge/event/level/LevelEvent.java
+++ b/src/main/java/net/minecraftforge/event/level/LevelEvent.java
@@ -5,13 +5,21 @@
 
 package net.minecraftforge.event.level;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.gui.screens.Screen;
 import net.minecraft.client.multiplayer.ClientLevel;
+import net.minecraft.core.BlockPos;
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.level.ServerLevel;
 import net.minecraft.util.ProgressListener;
+import net.minecraft.util.random.WeightedRandomList;
+import net.minecraft.world.entity.MobCategory;
 import net.minecraft.world.level.LevelAccessor;
+import net.minecraft.world.level.biome.MobSpawnSettings;
 import net.minecraft.world.level.storage.ServerLevelData;
 import net.minecraftforge.common.ForgeInternalHandler;
 import net.minecraftforge.common.MinecraftForge;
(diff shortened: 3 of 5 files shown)


src/test/java/net/minecraftforge/debug/world/PotentialSpawnsEventTest.java
```diff
@@ -0,0 +1,41 @@
...
+    {
+        if (enabled)
+        {
+            MinecraftForge.EVENT_BUS.addListener(PotentialSpawnsEventTest::onlySpawnHostileMobs);
+        }
+    }
+
+    public static void onlySpawnHostileMobs (WorldEvent.PotentialSpawns event)
+    {
+        Level level = (Level) event.getWorld();
```
sciwhiz12: Why is this being casted to `Level`, given that the only use of the value is to call `LevelAccessor#getCurrentDifficultyAt`?

src/test/java/net/minecraftforge/debug/world/PotentialSpawnsEventTest.java
```diff
@@ -0,0 +1,41 @@
...
+import net.minecraft.world.entity.MobCategory;
+import net.minecraft.world.level.Level;
+import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.event.world.WorldEvent;
+import net.minecraftforge.fml.common.Mod;
+
+@Mod("potentialspawnsevent_test")
+public class PotentialSpawnsEventTest
+{
+    public static final boolean enabled = false;
```
sciwhiz12: Formatting: `static final` constants should use `SCREAMING_CASE`.
```suggestion
    public static final boolean ENABLED = false;
```

src/test/java/net/minecraftforge/debug/world/PotentialSpawnsEventTest.java
```diff
@@ -0,0 +1,41 @@
...
+import net.minecraft.core.BlockPos;
+import net.minecraft.world.Difficulty;
+import net.minecraft.world.entity.MobCategory;
+import net.minecraft.world.level.Level;
+import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.event.world.WorldEvent;
+import net.minecraftforge.fml.common.Mod;
+
+@Mod("potentialspawnsevent_test")
+public class PotentialSpawnsEventTest
```
sciwhiz12, line 22: Please provide a javadoc for this class which explains what feature this tests, how to test it, and what the expected results are from said test.

src/test/java/net/minecraftforge/debug/world/PotentialSpawnsEventTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public PotentialSpawnsEventTest()
+    {
+        if (enabled)
+        {
+            MinecraftForge.EVENT_BUS.addListener(PotentialSpawnsEventTest::onlySpawnHostileMobs);
+        }
+    }
+
+    public static void onlySpawnHostileMobs (WorldEvent.PotentialSpawns event)
```
sciwhiz12: Formatting: no space between method name and opening parenthesis.
```suggestion
    public static void onlySpawnHostileMobs(WorldEvent.PotentialSpawns event)
```

src/main/java/net/minecraftforge/event/world/WorldEvent.java
```diff
@@ -118,4 +124,44 @@ public ServerLevelData getSettings()
...
         }
     }
+
+    /**
+     * Called by NaturalSpawner to gather a list of all possible entities that can spawn at the specified location.
+     * If an entry is added to the list, it needs to be a globally unique instance.
+     * The event is called in {@link NaturalSpawner#getRandomSpawnMobAt()}.
+     * Canceling the event will result in an empty list, meaning no entity will be spawned.
+     */
+    @net.minecraftforge.eventbus.api.Cancelable
```
sciwhiz12: Formatting: an import should be used, instead of a fully-qualified name.

src/main/java/net/minecraftforge/event/world/WorldEvent.java
```diff
@@ -118,4 +124,44 @@ public ServerLevelData getSettings()
...
+        private final BlockPos pos;
+        private final List<MobSpawnSettings.SpawnerData> list;
+
+        public PotentialSpawns(LevelAccessor world, MobCategory type, BlockPos pos, WeightedRandomList<MobSpawnSettings.SpawnerData> oldList)
+        {
+            super(world);
+            this.pos = pos;
+            this.type = type;
+            if (!oldList.isEmpty())
+              this.list = new ArrayList<>(oldList.unwrap());
```
sciwhiz12: Formatting: 4 spaces for indentation.
```suggestion
                this.list = new ArrayList<>(oldList.unwrap());
```

src/main/java/net/minecraftforge/event/world/WorldEvent.java
```diff
@@ -118,4 +124,44 @@ public ServerLevelData getSettings()
...
+            else
+                this.list = new ArrayList<>();
+        }
+
+        public MobCategory getType()
```
sciwhiz12: The name of the getter and field should reflect the class name of `MobCategory`.

src/main/java/net/minecraftforge/event/world/WorldEvent.java
```diff
@@ -118,4 +124,44 @@ public ServerLevelData getSettings()
...
+        {
+            return type;
+        }
+
+        public BlockPos getPos()
+        {
+            return pos;
+        }
+
+        public List<MobSpawnSettings.SpawnerData> getList()
```
sciwhiz12: `getList` is too generic a name for this; perhaps `getSpawnerData()`?

src/main/java/net/minecraftforge/event/ForgeEventFactory.java
```diff
@@ -823,4 +826,12 @@ public static void onPostServerTick(BooleanSupplier haveTime)
     {
         MinecraftForge.EVENT_BUS.post(new TickEvent.ServerTickEvent(TickEvent.Phase.END, haveTime));
     }
+
+    public static WeightedRandomList<MobSpawnSettings.SpawnerData> getPotentialSpawns(LevelAccessor world, MobCategory type, BlockPos pos, WeightedRandomList<MobSpawnSettings.SpawnerData> oldList)
+    {
+        WorldEvent.PotentialSpawns event = new WorldEvent.PotentialSpawns(world, type, pos, oldList);
+        if (MinecraftForge.EVENT_BUS.post(event))
+            return WeightedRandomList.create ();
```
sciwhiz12: Formatting: no space between method name and opening parenthesis.
```suggestion
            return WeightedRandomList.create();
```

sciwhiz12 (changes requested):

src/test/java/net/minecraftforge/debug/world/PotentialSpawnsEventTest.java
TheMuso: Good catch, a holdover from when I was playing with different ideas for the test. Fixed locally.

src/main/java/net/minecraftforge/event/world/WorldEvent.java
```diff
@@ -118,4 +124,44 @@ public ServerLevelData getSettings()
             return settings;
         }
     }
+
+    /**
+     * Called by NaturalSpawner to gather a list of all possible entities that can spawn at the specified location.
+     * If an entry is added to the list, it needs to be a globally unique instance.
+     * The event is called in {@link NaturalSpawner#getRandomSpawnMobAt()}.
```
pupnewfster: This should link to the FQN so that the IDE can properly link them when clicking to go to target even if it shows an error due to it being private. `{@link net.minecraft.world.level.NaturalSpawner#getRandomSpawnMobAt(ServerLevel, StructureManager, ChunkGenerator, MobCategory, RandomSource, BlockPos)}`

src/main/java/net/minecraftforge/event/world/WorldEvent.java
```diff
@@ -118,4 +124,44 @@ public ServerLevelData getSettings()
...
+        {
+            return mobcategory;
+        }
+
+        public BlockPos getPos()
+        {
+            return pos;
+        }
+
+        public List<MobSpawnSettings.SpawnerData> getSpawnerDataList()
```
pupnewfster: I know the 1.16 version didn't have it but I feel like it may be better API wise to instead of exposing a mutable list directly like this, expose:
- an unmodifiable view of the list
- a method for adding spawner data to the list
- a method for removing spawner data from the list

It also would be beneficial to add brief javadocs to the various methods in this event

pupnewfster (commented):
Left a couple comments, the other thing that I am a bit unsure about (though I am aware it wasn't there in the 1.16 version) is that I think this might have issues with adding new spawns currently due to the following chain of events
- `NaturalSpawner#spawnCategoryForPosition` calls `NaturalSpawner#getRandomSpawnMobAt` and passes it to `NaturalSpawner#isValidSpawnPostitionForType`
-  `NaturalSpawner#isValidSpawnPostitionForType` calls `NaturalSpawner#canSpawnMobAt` which checks `NaturalSpawner#mobsAt`
- But before picking the random mob spawner we wrap `mobsAt` via the `getPotentialSpawns` method
This means that if we added any then when we are validating it can spawn it won't be in the list of valid mobs for the location. Which presuming I am correct then I think a better target for the patch would be in `mobsAt` rather than `getRandomSpawnMobAt`

src/main/java/net/minecraftforge/event/level/LevelEvent.java
```diff
@@ -117,4 +124,82 @@ public ServerLevelData getSettings()
...
+        }
+
+        /**
+         * Returns the list ofmobs that can potentially be spawned.
+         *
+         * @return the list ofmobs that can potentially be spawned.
+         */
+        public WeightedRandomList<MobSpawnSettings.SpawnerData> getSpawnerDataList()
+        {
+            return WeightedRandomList.create(list);
```
pupnewfster: This feels needlessly complex. just add another field: `private final List<MobSpawnSettings.SpawnerData> view` and initialize it to `Collections.unmodifiableList(list)` and then just make this method return a list instead of a weighted random list. Lists are easier to interact with anyway

TheMuso: Ok let me know if the latest push is what you were thinking, or whether I have totally missed the point. :)

patches/minecraft/net/minecraft/world/level/NaturalSpawner.java.patch
```diff
@@ -40,7 +40,7 @@
...
 -      return m_220455_(p_220448_, p_220444_, p_220447_, p_220445_) ? NetherFortressStructure.f_228517_ : p_220446_.m_223133_(p_220449_ != null ? p_220449_ : p_220444_.m_204166_(p_220448_), p_220445_, p_220447_, p_220448_);
-+      return m_220455_(p_220448_, p_220444_, p_220447_, p_220445_) ? p_220445_.m_220521_().m_175515_(Registry.f_235725_).m_123013_(BuiltinStructures.f_209859_).m_226612_().get(MobCategory.MONSTER).f_210044_() : p_220446_.m_223133_(p_220449_ != null ? p_220449_ : p_220444_.m_204166_(p_220448_), p_220445_, p_220447_, p_220448_);
++      return net.minecraftforge.event.ForgeEventFactory.getPotentialSpawns(p_220444_, p_220447_, p_220448_, m_220455_(p_220448_, p_220444_, p_220447_, p_220445_) ? p_220445_.m_220521_().m_175515_(Registry.f_235725_).m_123013_(BuiltinStructures.f_209859_).m_226612_().get(MobCategory.MONSTER).f_210044_() : p_220446_.m_223133_(p_220449_ != null ? p_220449_ : p_220444_.m_204166_(p_220448_), p_220445_, p_220447_, p_220448_));
```
pupnewfster: Would like input from other triager's/other people who contribute to ports but I feel like it may be worth adding a single line comment above this basically saying: `//Forge: Add in potential spawns, and replace hardcoded nether fortress mob list` (just so that it is easier to not accidentally lose one of the pieces of the patch between ports).

src/main/java/net/minecraftforge/event/level/LevelEvent.java
```diff
@@ -117,4 +125,84 @@ public ServerLevelData getSettings()
...
+         */
+        public BlockPos getPos()
+        {
+            return pos;
+        }
+
+        /**
+         * Returns the list ofmobs that can potentially be spawned.
+         *
+         * @return the list ofmobs that can potentially be spawned.
```
pupnewfster: Both of these lines are missing a space in `of mobs`. But realistically instead of having two identical lines for this javadocs and other, it seems more reasonable to just have the one for `@return`

src/main/java/net/minecraftforge/event/level/LevelEvent.java
pupnewfster: Yeah that is what I was thinking in regards to this method.

src/main/java/net/minecraftforge/event/level/LevelEvent.java
TheMuso: Whoops, good catch re the typos.

patches/minecraft/net/minecraft/world/level/NaturalSpawner.java.patch
TheMuso: That makes sense to me, will leave open for now for others to comment.

pupnewfster: After some internal discussions, the general consensus is yes it would be worth having a comment like that

src/main/java/net/minecraftforge/event/level/LevelEvent.java
```diff
@@ -117,4 +125,78 @@ public ServerLevelData getSettings()
...
+        {
+            super(level);
+            this.pos = pos;
+            this.mobcategory = category;
+            if (!oldList.isEmpty())
+                this.list = new ArrayList<>(oldList.unwrap());
+            else
+                this.list = new ArrayList<>();
+
+            this.view = Collections.unmodifiableList(list);        }
```
sciwhiz12: Formatting: incorrect positioning of brace.
```suggestion
            this.view = Collections.unmodifiableList(list);
        }
```

src/main/java/net/minecraftforge/event/level/LevelEvent.java
```diff
@@ -117,4 +125,78 @@ public ServerLevelData getSettings()
...
+
+        /**
+         * @return the category of the mobs in the spawn list.
```
sciwhiz12: Use the inline `{@return}` tag instead.
```suggestion
         * {@return the category of the mobs in the spawn list}
```

src/main/java/net/minecraftforge/event/level/LevelEvent.java
```diff
@@ -117,4 +125,78 @@ public ServerLevelData getSettings()
...
+         */
+        public MobCategory getMobCategory()
+        {
+            return mobcategory;
+        }
+
+        /**
+         * @return the block position where the chosen mob will be spawned.
```
sciwhiz12: Use the inline `{@return}` tag instead.
```suggestion
         * {@return the block position where the chosen mob will be spawned}
```

src/main/java/net/minecraftforge/event/level/LevelEvent.java
```diff
@@ -117,4 +125,78 @@ public ServerLevelData getSettings()
...
+         */
+        public BlockPos getPos()
+        {
+            return pos;
+        }
+
+        /**
+         * @return the list of mobs that can potentially be spawned.
```
sciwhiz12: Use the inline `{@return}` tag instead.
```suggestion
         * {@return the list of mobs that can potentially be spawned}
```

src/main/java/net/minecraftforge/event/level/LevelEvent.java
```diff
@@ -117,4 +125,78 @@ public ServerLevelData getSettings()
...
         }
     }
+
+    /**
+     * Called by NaturalSpawner to gather a list of all possible entities that can spawn at the specified location.
+     * If an entry is added to the list, it needs to be a globally unique instance.
+     * The event is called in {@link net.minecraft.world.level.NaturalSpawner#getRandomSpawnMobAt(ServerLevel,
+     * StructureManager, ChunkGenerator, MobCategory, RandomSource, BlockPos)}.
+     * Canceling the event will result in an empty list, meaning no entity will be spawned.
+     */
```
sciwhiz12: Could you please reformat this to follow the general structure of the javadocs for client events (the `net.minecraftforge.client.event` package)?

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/event/level/LevelEvent.java
TheMuso: Whoops, good catch, thanks.

src/main/java/net/minecraftforge/event/level/LevelEvent.java
TheMuso: Is what I have in my latest push what you had in mind?

sciwhiz12 (approved):

PR #8716, opened 2022-06-10
[1.19.x] Add CommandBuildContext to Register Command Events

ChampionAsh5357:
This expands `RegisterCommandsEvent` and `RegisterClientCommandsEvent` to accept a `CommandBuildContext` for holder lookups.

### DIFF
--- a/src/main/java/net/minecraftforge/client/event/RegisterClientCommandsEvent.java
+++ b/src/main/java/net/minecraftforge/client/event/RegisterClientCommandsEvent.java
@@ -6,6 +6,7 @@
 package net.minecraftforge.client.event;
 
 import com.mojang.brigadier.CommandDispatcher;
+import net.minecraft.commands.CommandBuildContext;
 import net.minecraft.commands.CommandSourceStack;
 import net.minecraft.commands.arguments.ObjectiveArgument;
 import net.minecraft.commands.arguments.ResourceLocationArgument;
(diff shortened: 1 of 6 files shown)


src/main/java/net/minecraftforge/client/event/RegisterClientCommandsEvent.java
```diff
@@ -53,4 +56,12 @@ public CommandDispatcher<CommandSourceStack> getDispatcher()
     {
         return dispatcher;
     }
+
+    /**
+     * @return the context to build the commands for
```
sciwhiz12: This should use an inline `{@return}` tag:
```suggestion
     * {@return the context to build the commands for}
```

src/main/java/net/minecraftforge/client/event/RegisterClientCommandsEvent.java
```diff
@@ -53,4 +56,12 @@ public CommandDispatcher<CommandSourceStack> getDispatcher()
...
+     */
+    public CommandBuildContext getContext()
```
sciwhiz12: Could it be `getBuildContext()` instead?

src/main/java/net/minecraftforge/event/RegisterCommandsEvent.java
```diff
@@ -40,4 +43,9 @@ public Commands.CommandSelection getEnvironment()
     {
         return environment;
     }
+
+    public CommandBuildContext getContext()
```
sciwhiz12: Could it be `getBuildContext()` instead? Also, could copy over the javadocs from the client commands event into here as well.

sciwhiz12 (changes requested):

pupnewfster (approved):

PR #8719, opened 2022-06-11
[1.19] Add support for items to add enchantments without setting them in NBT

KnightMiner:
Update for #8420 for the 1.19 branch.

----

Many enchantments in vanilla are hardcoded to checks using `EnchantmentHelper::getItemEnchantmentLevel`, for instance the curse of binding or soul speed. Some mods add enchantment alternatives, such as Tinkers' Construct's modifiers, augments on various handheld tools in tech mods, or tools that start with a specific enchantment. Adding enchantments directly to the tool can easily lead to an XP dupe using the grindstone, prevents re-enchanting the tool, and in general conflicts with map-maker tools. It is sometimes possible to replicate the effect of enchantments, though it often partially successful or very likely to conflict with other mods (such as the global loot modifier example mod reimplementing silk touch by regenerating the entire drops list).

----

This PR solves the problem by adding two new methods to `IForgeItem` and `IForgeItemStack`:
* `getEnchantmentLevel`: Allows an item to override the level of an enchantment from NBT, for hooks using `EnchantmentHelper::getItemEnchantmentLevel`
* `getAllEnchantments`: Allows an item to override the method returning a full enchantment list, for hooks using `EnchantmentHelper.deserializeEnchantments(stack.getEnchantmentTags())`.

For some implementation notes:
* Enchantments from the hooks are ignored in vanilla enchanting blocks such as the grindstone, anvil, and enchanting table. It is assumed that if you are using these hooks, your intention is to implement a system parallel to enchantments.
* Enchantments from the hooks are not shown in the tooltip. If the modder wishes to show them in the tooltip, they can override `Item::appendHoverText` (which notably runs right before vanilla adds enchantment names)
* Returning enchantments will not prevent enchanting the tool in a vanilla table or anvil or make the tool have the enchantment glint, as between forge and vanilla there are enough methods to control that behavior.
* `getAllEnchantments` does not return book enchantments, it is meant for gameplay logic, not enchanting.
* This is technically a breaking change for callers of `EnchantmentHelper::getItemEnchantmentLevel`. The method was deprecated so modders can decide whether they need the NBT enchant or the gameplay one.

To minimize patches:
* `EnchantmentHelper::getItemEnchantmentLevel` was patched to call the new method, and a new method `EnchantmentHelper::getTagEnchantmentLevel` for cases that specifically want to read from NBT (two cases in vanilla enchanting logic and the default for the hook). As a result, the forge hook and vanilla method are interchangeable. 
* `EnchantmentHelper::getEnchantments` in vanilla is designed for encahnting logic (anvil, enchantment table, etc.). Mods should not have been calling that method anyways as it returns book enchantments, which should not be used for gameplay.

----

~~Included with this PR is~~ [This commit includes](https://github.com/MinecraftForge/MinecraftForge/pull/8719/commits/6adeb1dd31401a96c85ced1702292b35872e3769#diff-083efbebadb812e4d8e5ce319f01ab59a7651dd3c0fb9a2ed6f2a99f6a88842c) a test mod that adds "soul boots". These boots use the hook to include both soul speed and the curse of binding, two enchantments I have noted are incredibly hard to impossible to replicate without this hook.


Minecraftschurli:
> Enchantments from the hooks are ignored in vanilla enchanting blocks such as the grindstone, anvil, and enchanting table. It is assumed that if you are using these hooks, your intention is to implement a system parallel to enchantments.

What about adding another callback method to determine this?

KnightMiner:
Why add another callback? They already exist. There are hooks on item stacks to make them enchantable or not without my PR. That comment simply says I did not change the default behavior to use the new hooks,

For the anvil and grindstone, just use the existing event/the event from the other PR, much better than me adding a likely insufficient hook here

PR #8721, opened 2022-06-11
1.19 - Remove GlobalLootModifierSerializer and move to Codecs

Cyborgmas:
With the registry changes, the serializer class is no longer needed, and it can be converted to a registry of `Codec<? extends IGlobalLootModifier>`.
I ran the test mod, and confirmed the data gen was the same and that the loot modifiers were triggering.
The `dungeon_loot` json was changed but only because the codec has a default value.


src/main/java/net/minecraftforge/common/data/GlobalLootModifierProvider.java
```diff
@@ -91,11 +87,17 @@ public void run(CachedOutput cache) throws IOException
...
-    public <T extends IGlobalLootModifier> void add(String modifier, GlobalLootModifierSerializer<T> serializer, T instance)
+    public <T extends IGlobalLootModifier> void add(String modifier, Codec<T> serializer, T instance)
     {
-        this.toSerialize.put(modifier, new Tuple<>(serializer, serializer.write(instance)));
+        JsonElement element = serializer.encodeStart(JsonOps.INSTANCE, instance).getOrThrow(false, s -> {});
+        if (!(element instanceof JsonObject obj))
+            throw new RuntimeException("Expecting loot modifier to be a Json Object!");
+        String serializerName = ForgeRegistries.LOOT_MODIFIER_SERIALIZERS.get().getKey(serializer).toString();
+        obj.addProperty("type", serializerName);
+        this.toSerialize.put(modifier, obj);
```
SizableShrimp: Aren't there dispatch codecs for this kind of stuff?

Cyborgmas: indeed, mb

src/main/java/net/minecraftforge/common/data/GlobalLootModifierProvider.java
```diff
@@ -87,17 +88,12 @@ public void run(CachedOutput cache) throws IOException
...
-    public <T extends IGlobalLootModifier> void add(String modifier, Codec<T> serializer, T instance)
+    public <T extends IGlobalLootModifier> void add(String modifier, T instance)
     {
-        JsonElement element = serializer.encodeStart(JsonOps.INSTANCE, instance).getOrThrow(false, s -> {});
-        if (!(element instanceof JsonObject obj))
-            throw new RuntimeException("Expecting loot modifier to be a Json Object!");
-        String serializerName = ForgeRegistries.LOOT_MODIFIER_SERIALIZERS.get().getKey(serializer).toString();
-        obj.addProperty("type", serializerName);
-        this.toSerialize.put(modifier, obj);
+        JsonElement json = IGlobalLootModifier.DIRECT_CODEC.encodeStart(JsonOps.INSTANCE, instance).getOrThrow(false, s -> {});
```
SizableShrimp, line 92: Vanilla uses change the consumer to `LOGGER::error`. I think u should also do that here

Cyborgmas: Confirmed `getOrThrow` throws an exception already and it gets logged.

Alpvax:
The test mod modifiers break the usual rules of using DeferredRegistry: they register codecs created statically.
There should either be a comment stating why it is acceptable to break that rule, or they should not break the rule.
Otherwise it will just cause confusion ("Sometimes you can use static initialisers, but when?")

Cyborgmas:
In this case it was fine (afaik) due to them being in a separate (inner) class, and thus not actually being statically init.
But wrapped them in suppliers just to be safe, and for a better example.

Shadows-of-Fire:
Please fix the merge conflicts so this PR can be moved forward.

Shadows-of-Fire:
Everything else looks fine, test mods work.  Data generates and parses as expected, and the loot modifiers work once in-game.

PR #8722, opened 2022-06-11
Add dataprovider for datapack registries, and JsonCodecProvider

Commoble:
This adds a dataprovider for datapack registries that serializes objects to the correct directory for them. Modders can use this via `JsonCodecProvider#forDatapackRegistry`.

```java
    /**
     * {@return DatapackRegistryProvider that encodes using the registered loading codec for the provided registry key}
     * Ensures the correct directory and codec are used.
     * Only vanilla datapack registries enumerated in {@link RegistryAccess} and custom forge datapack registries can
     * be generated this way.
     * 
     * @param <T> Registry element type, e.g. Biome
     * @param dataGenerator DataGenerator provided by {@link GatherDataEvent}.
     * @param modid namespace of the mod adding this DataProvider, for logging purposes.
     * @param registryOps RegistryOps to encode values to json with.
     * @param registryKey ResourceKey identifying the registry and its directory.
     * @param entries Map of entries to encode and their ResourceKeys.
     */
    public static <T> JsonCodecProvider<T> forDatapackRegistry(DataGenerator dataGenerator, String modid, RegistryOps<JsonElement> registryOps, ResourceKey<Registry<T>> registryKey, Map<ResourceLocation, T> entries)
```

This is tested by adjusting the BiomeModifierTest mod to use this dataprovider.

JsonCodecProvider itself is a more general-purpose dataprovider that can datagen any object type for which a codec exists (allowing the codec to perform any validation if necessary).

### DIFF
--- a/src/main/java/net/minecraftforge/common/data/JsonCodecProvider.java
+++ b/src/main/java/net/minecraftforge/common/data/JsonCodecProvider.java
@@ -22,6 +22,7 @@ import net.minecraft.resources.RegistryOps;
 import net.minecraft.resources.ResourceKey;
 import net.minecraft.resources.ResourceLocation;
 import net.minecraft.server.packs.PackType;
+import net.minecraftforge.common.data.ExistingFileHelper.ResourceType;
 import net.minecraftforge.forge.event.lifecycle.GatherDataEvent;
 
 /**
@@ -35,13 +36,23 @@ import net.minecraftforge.forge.event.lifecycle.GatherDataEvent;
  * @param dataGenerator DataGenerator provided by {@link GatherDataEvent}.
  * @param dynamicOps DynamicOps to encode values to jsons with using the provided Codec, e.g. {@link JsonOps.INSTANCE}.
  * @param packType PackType specifying whether to generate entries in assets or data.
- * @param registryFolder String representing the registry folder, e.g. "dimension" or "cheesemod/cheese".
+ * @param directory String representing the directory to generate jsons in, e.g. "dimension" or "cheesemod/cheese".
  * @param codec Codec to encode values to jsons with using the provided DynamicOps.
  * @param entries Map of named entries to serialize to jsons. Paths for values are derived from the ResourceLocation's entryid:entrypath as specified above.
  */
-public record JsonCodecProvider<T>(DataGenerator dataGenerator, String modid, DynamicOps<JsonElement> dynamicOps, PackType packType, String registryFolder, Codec<T> codec, Map<ResourceLocation, T> entries) implements DataProvider
+public record JsonCodecProvider<T>(DataGenerator dataGenerator, ExistingFileHelper existingFileHelper, String modid, DynamicOps<JsonElement> dynamicOps, PackType packType, String directory, Codec<T> codec, Map<ResourceLocation, T> entries) implements DataProvider
 {
     private static final Logger LOGGER = LogManager.getLogger();
+    
+    public JsonCodecProvider
+    {
+        // Track generated data so other dataproviders can validate if needed.
+        final ResourceType resourceType = new ResourceType(packType, ".json", directory);
+        for (ResourceLocation id : entries.keySet())
+        {
+            existingFileHelper.trackGenerated(id, resourceType);
+        }
+    }
     
     /**
      * {@return DatapackRegistryProvider that encodes using the registered loading codec for the provided registry key}
@@ -56,7 +67,7 @@ public record JsonCodecProvider<T>(DataGenerator dataGenerator, String modid, Dy
      * @param registryKey ResourceKey identifying the registry and its directory.
      * @param entries Map of entries to encode and their ResourceLocations. Paths for values are derived from the ResourceLocation's entryid:entrypath.
      */
-    public static <T> JsonCodecProvider<T> forDatapackRegistry(DataGenerator dataGenerator, String modid, RegistryOps<JsonElement> registryOps, ResourceKey<Registry<T>> registryKey, Map<ResourceLocation, T> entries)
+    public static <T> JsonCodecProvider<T> forDatapackRegistry(DataGenerator dataGenerator, ExistingFileHelper existingFileHelper, String modid, RegistryOps<JsonElement> registryOps, ResourceKey<Registry<T>> registryKey, Map<ResourceLocation, T> entries)
     {
         final ResourceLocation registryId = registryKey.location();
         // Minecraft datapack registry folders are in data/json-namespace/registry-name/
@@ -65,7 +76,7 @@ public record JsonCodecProvider<T>(DataGenerator dataGenerator, String modid, Dy
             ? registryId.getPath()
             : registryId.getNamespace() + "/" + registryId.getPath();
         final Codec<T> codec = (Codec<T>)RegistryAccess.REGISTRIES.get(registryKey).codec();
-        return new JsonCodecProvider<>(dataGenerator, modid, registryOps, PackType.SERVER_DATA, registryFolder, codec, entries);
+        return new JsonCodecProvider<>(dataGenerator, existingFileHelper, modid, registryOps, PackType.SERVER_DATA, registryFolder, codec, entries);
     }
 
     @Override
@@ -77,16 +88,16 @@ public record JsonCodecProvider<T>(DataGenerator dataGenerator, String modid, Dy
         {
             final ResourceLocation id = entry.getKey();
             final T value = entry.getValue();
-            final Path path = outputFolder.resolve(String.join("/", dataFolder, id.getNamespace(), this.registryFolder(), id.getPath()+".json"));
+            final Path path = outputFolder.resolve(String.join("/", dataFolder, id.getNamespace(), this.directory(), id.getPath()+".json"));
             this.codec().encodeStart(this.dynamicOps(), value)
-                .resultOrPartial(msg -> LOGGER.error("Failed to encode {}: {}", path, msg)) // log error on encode failure
-                .ifPresent(LamdbaExceptionUtils.rethrowConsumer(json -> DataProvider.saveStable(cache, json, path)));// output to file on encode success
+                .resultOrPartial(msg -> LOGGER.error("Failed to encode {}: {}", path, msg)) // Log error on encode failure.
+                .ifPresent(LamdbaExceptionUtils.rethrowConsumer(json -> DataProvider.saveStable(cache, json, path)));// Output to file on encode success.
         }
     }
 
     @Override
     public String getName()
     {
-        return String.format("%s generator for %s", this.registryFolder(), this.modid());
+        return String.format("%s generator for %s", this.directory(), this.modid());
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/net/minecraftforge/common/data/JsonCodecProvider.java
```diff
@@ -0,0 +1,92 @@
...
+ * 
+ * @param <T> the type of thing being generated.
+ * @param dataGenerator DataGenerator provided by {@link GatherDataEvent}.
+ * @param dynamicOps DynamicOps to encode values to jsons with using the provided Codec, e.g. {@link JsonOps.INSTANCE}.
+ * @param packType PackType specifying whether to generate entries in assets or data.
+ * @param registryFolder String representing the registry folder, e.g. "dimension" or "cheesemod/cheese".
+ * @param codec Codec to encode values to jsons with using the provided DynamicOps.
+ * @param entries Map of named entries to serialize to jsons. Paths for values are derived from the ResourceLocation's entryid:entrypath as specified above.
+ */
+public record JsonCodecProvider<T>(DataGenerator dataGenerator, String modid, DynamicOps<JsonElement> dynamicOps, PackType packType, String registryFolder, Codec<T> codec, Map<ResourceLocation, T> entries) implements DataProvider
```
Minecraftschurli: Maybe rename registry folder to something else for more generic (non dynamic registry) use cases

src/test/java/net/minecraftforge/debug/world/BiomeModifierTest.java
```diff
@@ -89,86 +88,49 @@ public BiomeModifierTest()
...
-        // prepare to datagenerate our biome modifier
-        final ResourceLocation biomeModifiersRegistryID = ForgeRegistries.Keys.BIOME_MODIFIERS.location();
-        final String biomeModifierPathString = String.join("/", directory, MODID, biomeModifiersRegistryID.getNamespace(), biomeModifiersRegistryID.getPath(), MODIFY_BADLANDS + ".json");
-        final Path biomeModifierPath = outputFolder.resolve(biomeModifierPathString);
+        final DataGenerator generator = event.getGenerator();
+        final ExistingFileHelper existingFileHelper = event.getExistingFileHelper();
+        
+        // Any reference holders our objects have must come from this same RegistryAccess instance,
+        // or encoding our objects will fail.
+        final RegistryAccess registries = RegistryAccess.builtinCopy();
```
SizableShrimp, line 102: It is my understanding that making a builtin copy is slow. If every mod is doing this for datagen, that seems a bit unnecessarily slow. Perhaps Forge should store a memoized builtinCopy() that just isn't frozen, for datagen purposes.

On the other hand, only one mod should be datagenning at a time realistically. However, mods may be doing this from GatherDataEvent (exactly like this test mod), which would still fire the event listener even if disabled generation AFAIK. So it would be slow for mods datagenning with deps on the runtime classpath.

src/main/java/net/minecraftforge/common/data/JsonCodecProvider.java
```diff
@@ -0,0 +1,103 @@
...
+ * 
+ * @param <T> the type of thing being generated.
+ * @param dataGenerator DataGenerator provided by {@link GatherDataEvent}.
+ * @param dynamicOps DynamicOps to encode values to jsons with using the provided Codec, e.g. {@link JsonOps.INSTANCE}.
+ * @param packType PackType specifying whether to generate entries in assets or data.
+ * @param directory String representing the directory to generate jsons in, e.g. "dimension" or "cheesemod/cheese".
+ * @param codec Codec to encode values to jsons with using the provided DynamicOps.
+ * @param entries Map of named entries to serialize to jsons. Paths for values are derived from the ResourceLocation's entryid:entrypath as specified above.
+ */
+public record JsonCodecProvider<T>(DataGenerator dataGenerator, ExistingFileHelper existingFileHelper, String modid, DynamicOps<JsonElement> dynamicOps, PackType packType, String directory, Codec<T> codec, Map<ResourceLocation, T> entries) implements DataProvider
```
SizableShrimp, line 43: Making this a record means no one can extend it, which is potentially not what we want. Providing the frame while also allowing extensibility sounds best to me.

src/main/java/net/minecraftforge/common/data/JsonCodecProvider.java
```diff
@@ -0,0 +1,103 @@
...
+        final Path outputFolder = this.dataGenerator().getOutputFolder();
+        final String dataFolder = this.packType().getDirectory();
+        for (final var entry : this.entries().entrySet())
+        {
+            final ResourceLocation id = entry.getKey();
+            final T value = entry.getValue();
+            final Path path = outputFolder.resolve(String.join("/", dataFolder, id.getNamespace(), this.directory(), id.getPath()+".json"));
+            this.codec().encodeStart(this.dynamicOps(), value)
+                .resultOrPartial(msg -> LOGGER.error("Failed to encode {}: {}", path, msg)) // Log error on encode failure.
+                .ifPresent(LamdbaExceptionUtils.rethrowConsumer(json -> DataProvider.saveStable(cache, json, path)));// Output to file on encode success.
```
SizableShrimp, line 94: Extracting this to a method that can be overriden sounds best as part of making this provider extensible.

src/main/java/net/minecraftforge/common/data/JsonCodecProvider.java
```diff
@@ -0,0 +1,103 @@
+package net.minecraftforge.common.data;
```
sciwhiz12, line 1: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/common/data/JsonCodecProvider.java
```diff
@@ -0,0 +1,103 @@
...
+ * @param dataGenerator DataGenerator provided by {@link GatherDataEvent}.
+ * @param dynamicOps DynamicOps to encode values to jsons with using the provided Codec, e.g. {@link JsonOps.INSTANCE}.
+ * @param packType PackType specifying whether to generate entries in assets or data.
+ * @param directory String representing the directory to generate jsons in, e.g. "dimension" or "cheesemod/cheese".
+ * @param codec Codec to encode values to jsons with using the provided DynamicOps.
+ * @param entries Map of named entries to serialize to jsons. Paths for values are derived from the ResourceLocation's entryid:entrypath as specified above.
+ */
+public record JsonCodecProvider<T>(DataGenerator dataGenerator, ExistingFileHelper existingFileHelper, String modid, DynamicOps<JsonElement> dynamicOps, PackType packType, String directory, Codec<T> codec, Map<ResourceLocation, T> entries) implements DataProvider
+{
+    private static final Logger LOGGER = LogManager.getLogger();
```
sciwhiz12, line 45: Use `LogUtils` and SLF4J's `Logger` instead of Log4J.

src/main/java/net/minecraftforge/common/data/JsonCodecProvider.java
```diff
@@ -0,0 +1,103 @@
...
+     */
+    public static <T> JsonCodecProvider<T> forDatapackRegistry(DataGenerator dataGenerator, ExistingFileHelper existingFileHelper, String modid, RegistryOps<JsonElement> registryOps, ResourceKey<Registry<T>> registryKey, Map<ResourceLocation, T> entries)
+    {
+        final ResourceLocation registryId = registryKey.location();
+        // Minecraft datapack registry folders are in data/json-namespace/registry-name/
+        // Non-vanilla registry folders are data/json-namespace/registry-namespace/registry-name/
+        final String registryFolder = registryId.getNamespace().equals("minecraft")
+            ? registryId.getPath()
+            : registryId.getNamespace() + "/" + registryId.getPath();
+        final Codec<T> codec = (Codec<T>)RegistryAccess.REGISTRIES.get(registryKey).codec();
```
sciwhiz12, line 78: Formatting: space after cast.
```suggestion
        final Codec<T> codec = (Codec<T>) RegistryAccess.REGISTRIES.get(registryKey).codec();
```

src/main/java/net/minecraftforge/common/data/JsonCodecProvider.java
```diff
@@ -0,0 +1,103 @@
...
+    @Override
+    public void run(final CachedOutput cache) throws IOException
+    {
+        final Path outputFolder = this.dataGenerator().getOutputFolder();
+        final String dataFolder = this.packType().getDirectory();
+        for (final var entry : this.entries().entrySet())
+        {
+            final ResourceLocation id = entry.getKey();
+            final T value = entry.getValue();
+            final Path path = outputFolder.resolve(String.join("/", dataFolder, id.getNamespace(), this.directory(), id.getPath()+".json"));
```
sciwhiz12, line 91: Formatting: spaces around `+` operator.
```suggestion
            final Path path = outputFolder.resolve(String.join("/", dataFolder, id.getNamespace(), this.directory(), id.getPath() + ".json"));
```

src/main/java/net/minecraftforge/common/data/JsonCodecProvider.java
```diff
@@ -0,0 +1,103 @@
...
+            this.codec().encodeStart(this.dynamicOps(), value)
+                .resultOrPartial(msg -> LOGGER.error("Failed to encode {}: {}", path, msg)) // Log error on encode failure.
```
sciwhiz12, line 93: Should we consider throwing an exception instead of logging the error?

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/common/data/JsonCodecProvider.java
LexManos: yes

src/test/java/net/minecraftforge/debug/world/BiomeModifierTest.java
marchermans: Since this is datagen where only a single mod or at least a handfull of mods is running, this is not a problem here.

src/main/java/net/minecraftforge/common/data/JsonCodecProvider.java
marchermans: I would argue against your position here @SizableShrimp.

This API is stable and contains everything needed to handle the Codecs writing to disk.
If you want a custom handler here as a modder, I don't think this API is something for you anyway.

Leaving it a record is fine in my opinion.

src/main/java/net/minecraftforge/common/data/JsonCodecProvider.java
marchermans: Why? what use case are you seeing which I am missing here? Just curious....

src/main/java/net/minecraftforge/common/data/JsonCodecProvider.java
SizableShrimp: too late, it's already a definalized class and merged in forge through a separate PR

Commoble:
Superceded by and cherry-picked into https://github.com/MinecraftForge/MinecraftForge/pull/8717

PR #8723, opened 2022-06-12
A potential approach to registering order-dependent objects

lukebemish:
This is a draft of a possible approach to solving an issue I've had come up in my own mods, that requires a very hacky solution at present.

An example of the issue: Blocks need certain things (blockstate properties) to be defined prior to the constructor call, and other pieces (block properties) are extremely difficult to fully change after the constructor call. If you are registering a block that depends on the properties of another block, from a (possibly unknown at compile-time) different mod you have no control over, then it is necessary to ensure that your block registers after the other block in question. At present, this is extremely difficult to do. I have put together a first attempt at making this a bit easier - this PR simply provides a single ObjectRegisteredEvent, which is fired after an object is registered in a RegisterEvent and includes a reference to the ResourceLocation, the object in question, and the RegisterEvent it was registered in. There are definitely some aspects I'd like to clean up (for isntance, the object registered is currently a raw `Object`), but I figured I'd place it here as a draft PR so that I don't put too much time into this if it isn't going anywhere.

### DIFF
--- a/src/main/java/net/minecraftforge/registries/ObjectRegisteredEvent.java
+++ b/src/main/java/net/minecraftforge/registries/ObjectRegisteredEvent.java
@@ -1,0 +1,30 @@
+package net.minecraftforge.registries;
+
+import net.minecraft.resources.ResourceLocation;
+import net.minecraftforge.eventbus.api.Event;
+import net.minecraftforge.fml.ModContainer;
+import net.minecraftforge.fml.event.IModBusEvent;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.function.Consumer;
+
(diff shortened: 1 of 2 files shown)


DaemonUmbra:
Have you considered asking the dev of the mod you are extending or adding onto to provide a utility to let you do this?
i.e. extracting their block properties to a method that you can call?

lukebemish:
The issue is that there is no specific mod. All I have is a block id. (For context, my mod here is Excavated Variants; it adds variants for config-defined ore/stone combinations)

lukebemish:
(the disadvantage of this approach is the number of events it will generate; I have an alternative approach that would avoid this issue that I will try to implement instead in the future)

PR #8724, opened 2022-06-12
[1.19] Fix data-gen output path of custom data-pack registries

Silverminer007:
This aims to fix output path of custom data-pack registries for datagen.

In #8630, there was a discussion about prefixing the resource key with the namespace or just prefix it on usage. It was decided to go for option two, but one usage was missed. Here is a fix to align registry dump report output path with the path it reads from.

### Patches:
- Patch WorldgenRegistryDumpReport to prefix the output path with the namespace if it's not "minecraft"

### DIFF
--- a/patches/minecraft/net/minecraft/resources/RegistryResourceAccess.java.patch
+++ b/patches/minecraft/net/minecraft/resources/RegistryResourceAccess.java.patch
@@ -1,13 +1,5 @@
 --- a/net/minecraft/resources/RegistryResourceAccess.java
 +++ b/net/minecraft/resources/RegistryResourceAccess.java
-@@ -21,6 +_,7 @@
- import net.minecraft.core.Registry;
- import net.minecraft.core.RegistryAccess;
- import net.minecraft.server.packs.resources.ResourceManager;
-+import net.minecraftforge.registries.DataPackRegistriesHooks;
- import org.slf4j.Logger;
- 
- public interface RegistryResourceAccess {
 @@ -47,7 +_,7 @@
  
                       DataResult dataresult;
@@ -39,7 +31,7 @@
  
           private static String m_214239_(ResourceLocation p_214240_) {
 -            return p_214240_.m_135815_();
-+            return DataPackRegistriesHooks.prefixNamespace(p_214240_);// FORGE: add non-vanilla registry namespace to loader directory, same format as tag directory (see net.minecraft.tags.TagManager#getTagDir(ResourceKey))
++            return net.minecraftforge.registries.DataPackRegistriesHooks.prefixNamespace(p_214240_); // FORGE: add non-vanilla registry namespace to loader directory, same format as tag directory (see net.minecraft.tags.TagManager#getTagDir(ResourceKey))
           }
  
           private static <E> ResourceLocation m_214268_(ResourceKey<E> p_214269_) {
--- a/patches/minecraft/net/minecraft/resources/RegistryResourceAccess.java.patch
+++ b/patches/minecraft/net/minecraft/resources/RegistryResourceAccess.java.patch
@@ -1,5 +1,13 @@
 --- a/net/minecraft/resources/RegistryResourceAccess.java
 +++ b/net/minecraft/resources/RegistryResourceAccess.java
+@@ -21,6 +_,7 @@
+ import net.minecraft.core.Registry;
+ import net.minecraft.core.RegistryAccess;
+ import net.minecraft.server.packs.resources.ResourceManager;
++import net.minecraftforge.registries.DataPackRegistriesHooks;
+ import org.slf4j.Logger;
+ 
+ public interface RegistryResourceAccess {
 @@ -47,7 +_,7 @@
  
                       DataResult dataresult;
@@ -18,7 +26,7 @@
                       } catch (Throwable throwable1) {
                          if (reader != null) {
                             try {
-@@ -107,12 +_,15 @@
+@@ -107,13 +_,14 @@
              });
           }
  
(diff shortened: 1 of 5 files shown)


Commoble:
If we're using this logic in multiple places, we may want to consolidate it into a static method somewhere (ForgeRegistries or DatapackRegistriesHooks, maybe?)

Silverminer007:
I would favor DatapackRegistriesHooks, because it doesn't really belong to ForgeRegistries IMO. Can you think of any further places where the wrong path might be used?

patches/minecraft/net/minecraft/resources/RegistryResourceAccess.java.patch
```diff
@@ -1,5 +1,13 @@
 --- a/net/minecraft/resources/RegistryResourceAccess.java
 +++ b/net/minecraft/resources/RegistryResourceAccess.java
+@@ -21,6 +_,7 @@
+ import net.minecraft.core.Registry;
+ import net.minecraft.core.RegistryAccess;
+ import net.minecraft.server.packs.resources.ResourceManager;
++import net.minecraftforge.registries.DataPackRegistriesHooks;
```
SizableShrimp: Nope this is not valid

patches/minecraft/net/minecraft/resources/RegistryResourceAccess.java.patch
```diff
@@ -30,8 +38,8 @@
           }
  
           private static String m_214239_(ResourceLocation p_214240_) {
-+            if (net.minecraftforge.registries.RegistryManager.FROZEN.getRegistry(p_214240_) != null && !(p_214240_.m_135827_().equals("minecraft")))
-+               return p_214240_.m_135827_() + "/" + p_214240_.m_135815_(); // FORGE: add non-vanilla registry namespace to loader directory, same format as tag directory (see net.minecraft.tags.TagManager#getTagDir(ResourceKey))
-             return p_214240_.m_135815_();
+-            return p_214240_.m_135815_();
++            return DataPackRegistriesHooks.prefixNamespace(p_214240_);// FORGE: add non-vanilla registry namespace to loader directory, same format as tag directory (see net.minecraft.tags.TagManager#getTagDir(ResourceKey))
```
SizableShrimp: Try to have a space between the line termination and comments please. Same for the other patch file.

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
```diff
@@ -57,4 +58,13 @@ public static Set<ResourceKey<? extends Registry<?>>> getSyncedCustomRegistries(
...
         return SYNCED_CUSTOM_REGISTRIES_VIEW;
     }
+
+    /**
+     *  FORGE: add non-vanilla registry namespace to loader directory, same format as tag directory (see net.minecraft.tags.TagManager#getTagDir(ResourceKey))
+     * @param registryKey original registry key
+     * @return path of the registry key. Prefixed with the modid if it wasn't minecraft
+     */
+    public static String prefixNamespace(ResourceLocation registryKey) {
+        return registryKey.getNamespace().equals("minecraft") ? registryKey.getPath() : registryKey.getNamespace() +  "/"  + registryKey.getPath();
```
SizableShrimp: This is not the same behavior as the old patch. Please fix.

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
```diff
@@ -57,4 +58,13 @@ public static Set<ResourceKey<? extends Registry<?>>> getSyncedCustomRegistries(
     {
         return SYNCED_CUSTOM_REGISTRIES_VIEW;
     }
+
+    /**
+     *  FORGE: add non-vanilla registry namespace to loader directory, same format as tag directory (see net.minecraft.tags.TagManager#getTagDir(ResourceKey))
```
SizableShrimp: This is a crude copy of the comment from the patch. We are in a forge class, this does not need a forge prefix. It should also not have two spaces. And there should be a {@link} used instead of just parentheses. Also a blank line between the description and the tags `@param`, etc.

SizableShrimp (changes requested):

src/main/java/net/minecraftforge/registries/DataPackRegistriesHooks.java
SizableShrimp: I see why this was changed from the discussion in #github-discussions. For anyone else here: Using RegistryManager#FROZEN was only necessary on 1.18.2 to filter out datapack registries that were not made through Forge's system to preserve their datapack format and behavior. It was decided that everyone on 1.19+ should be using the forge system, so this is reasonable to remove.

SizableShrimp (approved):

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1424,4 +1424,28 @@ public static Map<PackType, Integer> readTypedPackFormats(JsonObject json)
...
+     * <p>
+     *    The fix, which is applied here prefixes the path of the registry with the namespace,
+     *    so {@code fancy_cheese}'s registry stores its elements in {@code data/<namespace>/fancy_cheese/cheeses}
+     *    and {@code awesome_cheese}'s registry stores its elements in {@code data/namespace/awesome_cheese/cheeses}
+     * </p>
+     *
+     * @param registryKey key of the registry
+     * @return path of the registry key. Prefixed with the namespace if it is not "minecraft"
+     */
+    public static String prefixNamespace(ResourceLocation registryKey) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    public static String prefixNamespace(ResourceLocation registryKey)
    {
```

sciwhiz12 (changes requested):
Looks fine aside from a single formatting change.

Commoble:
#8717's [JsonCodecProvider](https://github.com/MinecraftForge/MinecraftForge/blob/1.19.x/src/main/java/net/minecraftforge/common/data/JsonCodecProvider.java) can use the registry folder function too as well

sciwhiz12 (approved):

PR #8725, opened 2022-06-13
Update EventBus, ML and BSL

cpw:
Update EventBus to 6.0, ModLauncher to 10.0.1 and BootstrapLauncher to 1.1


cpw:
Tested forgedev and runtime, both work. Looks like I didn't break eventbus by updating and modularizing it.

PR #8726, opened 2022-06-13
Move is_mountain to forge's tag instead of vanilla's

TelepathicGrunt:
Simply put, I was wondering why the heck my stone mountain themed Mineshaft was spawning in badlands biomes. Turns out, Forge is putting its peak and slope tag entries into minecraft:is_mountain tag as a result of this PR. https://github.com/MinecraftForge/MinecraftForge/pull/8564

It was just a simple mistake overlooked by accident. This PR is to fix that and move the tag to under forge. Though, I will point out, this is a great example of why the slopes tag should not have badlands or savannas in it as even vanilla does not consider them as part of a mountain in their is_mountain tag. But that's a discussion for another day on if forge's is_mountain tag is even usable by anyone doing mountain themed stuff. For now, let's just get the is_mountain tag back under forge instead of modifying vanilla's tag.

Edit with more info: Looking at what vanilla uses with it's is_mountain tag, Forge is now spawning pillager outposts in wooded badlands and windswept savanna when it could not before with vanilla. And regular mineshafts can now spawn in wooded badlands when it should only be mesa mineshafts spawning. So vanilla behavior is being changed right now by forge
https://minecraft.fandom.com/wiki/Tag


TheCurle:
You're the biome boss, boss

PR #8727, opened 2022-06-13
[1.19.x] Fix experimental confirmation screen

### LINKED ISSUE #8713: [1.19] java.lang.IllegalStateException: Lock is no longer valid Error
**Minecraft Version:** 1.19

**Forge Version:** 41.0.16

**Logs:** https://gist.github.com/KedisPL/25c7a91b7e7be7b2430f2a4cd5e834fa

**Steps to Reproduce:**
1. Create a new world.
2. Once the world has been created, leave it.
3. Enter the world again.
4. When you try to enter, a message about experimental settings appears. Click proceed.
5. The game crashes.

**Description of issue:**
Mod with built-in worldgen datapack crashes when you re-enter the world after creating it and saving it.
After restarting the game and entering the world, the game no longer crashes.

SizableShrimp:
A porting error (by me) led to the wrong method being used in `ForgeHooksClient#createWorldConfirmationScreen` when the user confirmed that they wanted to proceed on loading the experimental world.
This is due to the fact that the directory lock is closed directly after `createWorldConfirmationScreen` is invoked.
However, the same directory lock instance is reused without being reopened once the user confirms.
This is incorrect and leads to a crash 100% of the time. Instead, the proper fix is to use `WorldOpenFlows#doLoadLevel` (which has similar method contexts to what was used in 1.18.2 inside `ForgeHooksClient#createWorldConfirmationScreen`).
We must now patch this method to be able to set `PrimaryLevelData#withConfirmedWarning` to true, as there is no way to get the world stem inside `createWorldCreationScreen`.
Simple fix, oopsie on my part.

Also unrelated is an inclusion which fixes checkJarCompatibility, which I also broke during the 1.19 update.
So that people can run it locally, and (potentially) during CI.
Didn't make a separate PR for this cuz it's minor and a waste of a build imo.

Fixes #8713.
Can be reproduced with the datapack provided on the original issue [here](https://github.com/MinecraftForge/MinecraftForge/issues/8713#issuecomment-1152347183).
Without the PR -> game crashes after clicking proceed on an existing world w/ experimental settings.
With the PR -> world proceeds to load after clicking proceed on an existing world w/ experimental settings.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1192,7 +1192,7 @@ project(':forge') {
 
     def baseForgeVersion = System.getProperty("forge.compatibility_check_version")
     if (baseForgeVersion == null)
-        baseForgeVersion = "${MC_VERSION}-${GIT_INFO.tag}.0"
+        baseForgeVersion = "1.19-41.0.1" // "${MC_VERSION}-${GIT_INFO.tag}.0"
     if (baseForgeVersion == "null")
         baseForgeVersion = null // Allow disabling from system property if not wanted
     def baseForgeUniversal = null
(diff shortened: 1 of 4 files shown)


sciwhiz12 (approved):
Tested and confirmed works.

Issue #8732, opened 2022-06-14
Dead bushes can be placed on glazed terracotta, unlike vanilla.


ChiefArug:
**Minecraft Version:** 1.18.2
**Forge Version:** 40.1.51

**Logs:** N/A

**Steps to Reproduce:**
 1. Launch a 1.18.2 Forge instance
 2. Place down any glazed terracotta block, then try place a dead bush on top.
 3. Launch vanilla and repeat step 2. Observer the behaviour is that you cannot place the dead bush

**Description of issue:**
I believe that this line in forges patches to the Block class is the cause: https://github.com/MinecraftForge/MinecraftForge/blob/be0facc2aa36a0bbc71e13496034007ce2cdce01/patches/minecraft/net/minecraft/world/level/block/Block.java.patch#L150
This line is also present in 1.19 and 1.17, making me think the issue persists in those versions too (not that 1.17 is going to get an update).

Bug in action:
It can be placed on both terracotta and glazed terracotta, unlike vanilla which does not allow glazed terracotta
![image](https://user-images.githubusercontent.com/73862885/173540418-41fbe57a-305d-4cbe-9ff2-78fe1127b925.png)

ferriarnus:
This line has been part of forge for a long time from what I can see. Is the glazed terracotta needed or can it simply be removed? I don't think **any** plant can be placed on glazed terracotta. If so I'll push a quick fix.

sciwhiz12:
I believe there is no plant that can be placed on glazed terracotta, so it can simply be removed.

SizableShrimp:
Removing that one bug does not solve all the other inconsistencies that PlantTypes bring to the table and their general invisibleness to modders just trying to replicate vanilla behavior. (If else instanceof chains in interfaces are bad mmkay). I'd still advocate for the removal of the entire system until a better one can be decided upon or do it all at once with the people who still need some form of the system.

sciwhiz12:
If we are to remove the entire `PlantType` system, it should be done now during 1.19's breaking changes window, so that a new system can potentially be introduced at a later point without needing to weave about with backwards compat with `PlantType`, if it happens after the breaking changes window.

ferriarnus:
From what I see, most logic can simply be implemented by the modder, since it is a simple Block tag that is used (except BEACH).

ferriarnus:
[...]

PR #8735, opened 2022-06-14
[1.20] Replace planttype with tags

ferriarnus:
In #8732 it was mentioned that the planttype system for forge plants could use an update. I've given it a go, and it seems like this will work.

Changes:
`canSustainPlant` in `Block` got updated. This function is used to check if a plant can grow on it. It is used in the vanilla `survive` method, so that custom blocks can overwrite it and allow plants to grow on it. Since most plants are `Bushblocks`, the patch in block could be compressed to check their method of `canPlaceOn`.

Tags for each plant "type" (like crop, plains) are/will be added. These serve to help modders with their `canSustainPlant`, bundeling similar plants. It is important to note that this means that additional checks (like water close for sugarcane) are not included, and should be implemented by the modder in  `canSustainPlant`, based on the requirements of the tag. 

Villager farming is handled by `isVillagerPlantable` in `IPlantable`.

For modded plants, `Bushblocks` and it's children seem to be the correct way to go, but it is not strictly needed (but can be nice for modded compat?). 

Open to suggestions.


ferriarnus:
Restarted design. Hopefully this one does not violate the forge <-> vanilla compat.

marchermans:
So this is not a good idea.
We simply can't use tags for logic which runs on the server like this.

This is because the vanilla client still needs to be able to connect to a server with server side only mods, these mods can modify these tags which causes a desync then, but because the client has no way of knowing this.

Please redesign this system to be sure to not use tags.

ferriarnus:
I don't think this is an issue, as they are only used on the server, not the client. This would not result in a desync.

marchermans:
> I don't think this is an issue, as they are only used on the server, not the client. This would not result in a desync.

That is exactly the point, the tag is ONLY used on the server side. A vanilla client has no idea what this tag is and it might decide not to place the plant, while a server might. Since this is a vanilla client you might not even get the packet on the server when the player tries to place it. Again tags are not allowed in selective logic, please refactor.

ferriarnus:
> A vanilla client has no idea what this tag is and it might decide not to place the plant, while a server might

I thought that, since the tag only influences harvest logic of villagers, this would be fine. Since the client has no connection what so ever with the ai of mobs. The actual placement of crops has no connection with the tag. But if this is wanted I could add a single "`isCrop`/`isVillagerHarvestable`" method to the `IPlantable` interface, to replace the tag with.

P.S. I'm in the discord if you want to tell me directly.

ferriarnus:
Now using the vanilla crop tag, so the tag desync is gone. All other tags except crops are optional, and should be used for mod compat only. 

Example, modded sand (soil):
- Override `canSustainPlant` to check for `plants/desert` tag.
- Override `canSustainPlant` to check for `plants/beach` tag and water present.

Example, modded sugarcane (plant):
- Implement `IPlantable`
- Add it to the `plants/beach` tag, so other mods can use it

As you can see, the tags themselves don't do anything, it is based on the fact that an other mod will have their soil act in a way befitting the tag. Tags just help other mods recognize each others plants and soil.

eerussianguy:
This may be a good time to revisit what goes into the tags. For me, I see 'beach' plants, OK does that mean gravel counts? probably not, even though there are gravel beaches. Does gravel count for caves? Maybe? It might make sense to see it check block type based stuff, like `forge:plantable/gravel` and `forge:plantable/sand`, since the specific *biome* something might generate in has not really any bearing on what it *can be planted on*, which is what we're worried about here. The planttype thing is a very legacy system, so since we are delaying this to 1.20, we should see about making it more robust. My theory is that this system has gotten so informally deprecated that people are not really using it anymore, so giving people something to use and use well would be awesome.

Edit -- just remembered, some plants only need a solid black (see weeping vines), this needs some sort of indicator as well

KnightMiner:
The tag desync issue is not based on vanilla or forge tags. The issue is the client side and the serverside checking different block lists.
* If the behavior is entirely defined in code, it would take a mod to change the behavior, and such mod could require being installed on both client and server.
* If the behavior is defined in datapacks, a datapack can change the behavior, and datapacks by definition have no client side component. As a result of a forge server working with a vnailla client (or vice versa) there is a potential desync in crop blocks. This will happen regardless of whether the tag is declared in the mojang tag key classes.

This tag issue is not a problem if you can guarantee that the method is only used serverside, though the forge team might have other reasons to dislike using tags for gameplay logic

ferriarnus:
It is only server side, as it is used in villager ai.

ferriarnus:
> This may be a good time to revisit what goes into the tags. For me, I see 'beach' plants, OK does that mean gravel counts? probably not, even though there are gravel beaches. Does gravel count for caves? Maybe? It might make sense to see it check block type based stuff, like `forge:plantable/gravel` and `forge:plantable/sand`, since the specific _biome_ something might generate in has not really any bearing on what it _can be planted on_, which is what we're worried about here. The planttype thing is a very legacy system, so since we are delaying this to 1.20, we should see about making it more robust. My theory is that this system has gotten so informally deprecated that people are not really using it anymore, so giving people something to use and use well would be awesome.

The names for the tags are just based on the previous system, you can name them whatever you want. The tags themselves don't mean anything, and are only to "group" blocks together. I can also not include them and let mods figure it out for themselves, but these names felt clear. `sandbeach` can be used if you think it's not good enough.

MelanX:
I think you need to run the `runData` task. You’re also missing to add something to the `#plants/crops` tag. Not sure if that’s intended.

eerussianguy:
Well the whole point of this being in forge is for modders to have common stuff to use. If you are asking me to make my own conventions as a response to me making a suggestion what’s even the point?

ferriarnus:
> Well the whole point of this being in forge is for modders to have common stuff to use. If you are asking me to make my own conventions as a response to me making a suggestion what’s even the point?

When reading this again I might have misunderstood you and acted to aggressive. My point was that the tags are a general guideline for mods. We can indeed look if new ones are needed, but since none of the vanilla logic is based on these tags, I didn't think about changing the existing ones. If a mod has a very special conditions, they can simply add their own tag, and doesn't need to be in forge. Though looking now, you were talking about more general tags, so see this as my late apologie.

ferriarnus:
I have rebased and updated the PR. I have dealt with the syncing issue by delegating the logic to a new method for villager behavior. Are there other concerns with the current idea? I am thinking if it could be useful to add some harvest and growing logic to `IPlantable` (or a new `IGrowable`). What would be within the scope of forge to provide for modders?

PR #8737, opened 2022-06-15
[1.19.x] Affect ItemEntity Motion in Custom Fluids

ChampionAsh5357:
Readds an accidental patch removal when reducing patch sizes allowing custom fluids to affect item entity motion.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/item/ItemEntity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/item/ItemEntity.java.patch
@@ -32,7 +32,7 @@
           Vec3 vec3 = this.m_20184_();
           float f = this.m_20192_() - 0.11111111F;
 +         net.minecraftforge.fluids.FluidType fluidType = this.getMaxHeightFluidType();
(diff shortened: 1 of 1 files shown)


patches/minecraft/net/minecraft/world/entity/item/ItemEntity.java.patch
```diff
@@ -32,7 +32,7 @@
           Vec3 vec3 = this.m_20184_();
           float f = this.m_20192_() - 0.11111111F;
 +         net.minecraftforge.fluids.FluidType fluidType = this.getMaxHeightFluidType();
-+         if (!fluidType.isAir() && !fluidType.isVanilla() && this.getFluidTypeHeight(fluidType) > (double)f) fluidType.setItemMovement(this);
++         if (!fluidType.isAir() && !fluidType.isVanilla() && this.getFluidTypeHeight(fluidType) > (double)f) fluidType.setItemMovement(this); else
```
sciwhiz12: Should we consider expanding this into two additional lines, so the detail of the `else` can't be glossed over?

marchermans: Yes we should.

marchermans (changes requested):

marchermans (approved):

PR #8740, opened 2022-06-15
[1.19.x] Fix fish swim not in custom liquid

MrScautHD:
This PR fix, fish can not swim in Custom Fluid


TheCurle:
That image is way to dark to see anything.
Please explain what this PR actually does, and the problem it intends to solve, rather than a vague almost perfectly black image that describes nothing.

MrScautHD:
> That image is way to dark to see anything. Please explain what this PR actually does, and the problem it intends to solve, rather than a vague almost perfectly black image that describes nothing.

updated

TheCurle:
That doesn't help anything. You've posted a picture of some fish.

DaemonUmbra:
Is the image intended to show the issue or proof that the fix works?

MrScautHD:
i updated the text too

ah yea the pic is before the fix

PR #8741, opened 2022-06-15
[1.19.x] Fix Jumping in Fluid

ChampionAsh5357:
Adds back in a missed patch during reduction that allows proper jumping in fluid instead of assuming that the fluid is water.


TheCurle:
I'm gonna hold off on this one so you can combine other missed patches that we discover.

Someone remind me to merge this in a day or two :P

ChampionAsh5357:
Merged into #8754 as requested.

PR #8742, opened 2022-06-15
[1.19.x] Implement full support for IPv6

PaintNinja:
Port of #8376 to 1.19.


PaintNinja:
Finished testing and all seems good - wasn't as thorough this time as there didn't seem to be any changes going from 1.18->1.19 so everything should work the same as the 1.18 version of this PR.

Tested forgedev client<->local server, userdev client<->client open to lan, prod client<->local server and prod client<->client open to lan, over IPv4/v6 dual-stack LAN.

PR #8744, opened 2022-06-16
[1.20] [RFC] The Capability API

Shadows-of-Fire:
# Capability API Changes

## Class Renames
* `Capability` -> `CapabilityType`.  This name is more representative of what this class actually is, because it is just a "key" that dictates what cap you want to query.
* `LazyOptional` -> `Capability`.  This name is more concise in that it represents an actual cap, and isn't some arbitrary object, meaning code that interacts with or caches these will be more readable.  Alternatively, this could be called `CapabilityInstance` or `CapabilityReference` (or a shorthand, like `CapReference`)
## Class Removals
* `CapabilityTokenSubclass` - No longer needed since `CapabilityType` is now statically named, instead of magically named through ASM.
* `CapabilityTyoken` - Also no longer needed for the same reason as above.
* `ICapabilityProviderImpl` - Methods from this class have been rolled into the appropriate place (`IForgeItemStack` for comparison checks and `ICapabilityProvider` for invalidate/revive).
* `ICapabilitySerializable` - Not needed, anything that is implementing this interface can just implement the two interfaces directly
* `CapabilityEnergy` - Rolled into `CapabilityTypes`
* `CapabilityFluidHandler` - Same as above
* `CapabilityItemHandler` - Same as above
## Class Additions
* `IAttachedCapabilityProvider` - The new base interface for all attached providers.
* `CapabilityTypes` - Holder for all of forge's added cap types.

### Capability Types
Current System:
```java
    public static final Capability<IItemHandler> ITEM_HANDLER_CAPABILITY = CapabilityManager.get(new CapabilityToken<>(){});

    public static void register(RegisterCapabilitiesEvent event)
    {
        event.register(IItemHandler.class);
    }
```
New System
```java
    public static final CapabilityType<IItemHandler> ITEMS = CapabilityManager.get(new ResourceLocation("forge", "items"));

    public static void register(RegisterCapabilitiesEvent event)
    {
        event.register(ITEMS);
    }
```
These changes retain the ability to have multiple mods make fields set to the same capability type, just through a name rather than the class type.  It has the added benefit of allowing multiple capability types to use the same interface, without having to declare a marker subinterface.
### Attaching Capabilities
The current system simply has users implement `ICapabilityProvider` and attach it to objects with a name.  If that provider implements `INBTSerializable`, it will be saved to disk.
```java
    public static void attach(AttachCapabilitiesEvent<ItemStack> event)
    {
        event.addCapability(new ResourceLocation("test", "test_cap"), providerInstance);
    }
```
This has the following drawbacks 
* No collision handling - Multiple providers can attach to an object and attempt to provide the same cap, but only one of them will be successful.  This will cause silent errors since it will appear that the cap is still exposed, but it may not be the type a mod is expecting.
* Expensive Comparison - Providers do not have a way to compare themselves to another version of themselves when itemstacks are merging.  Currently, all providers are serialized, and the serialized versions are compared.
* Awkward Cloning - Providers cannot copy themselves, and are instead serialized, re-attached, and deserialized.

The new system introduces an interface specific to attached capabilities, `IAttachedCapabilityProvider`, and two subinterfaces with added functionality, `IComparableCapabilityProvider` and `ICopyableCapabilityProvider`.  The subinterfaces are only required as necessary by the parent object.
You only need implement these interfaces if you want advanced functionality, otherwise you can utilize the `ProviderBuilder` class obtained from `AttachCapabilitiesEvent` to generate a provider for any object, as long as you can provide the required functionality.
Finally, AttachCapabilitiesEvent is no longer generic, and has specific subclasses, the function `AttachCapabilitiesEvent#addCapability` takes in the capability type instead of a name (the provider is responsible for providing its name), and the event only permits a single provider to be attached to any given type to prevent silent collisions.

Examples:
```java
    public static void attach(AttachCapabilitiesEvent.ItemStacks event)
    {
        event.addCapability(CapabilityTypes.ITEMS, attachedProvider);
    }
```
```java
    event.builder(() -> new ItemStackHandler(), CapabilityTypes.ITEMS)
        .withSerializer(ItemStackHandler::serializeNBT, ItemStackHandler::deserializeNBT)
        .withComparator(MyClass::areHandlersEqual)
        .withCopier(MyClass::cloneHandler)
        .attach(new ResourceLocation("test", "test_cap"));
```
#### Attached Cap Comparisons
ItemStack requires that all capabilities be compared when checking if those itemstacks may merge or be considered equal.  To support this, the interface `IComparableCapabilityProvider` is required for all providers attached to stacks.  It has the following method:
```java
boolean isEquivalentTo(@Nullable IComparableCapabilityProvider<O> other);
```
This allows the provider to implement a high-speed comparison function instead of serializing both and comparing the resulting nbt data.  It is guaranteed that the argument `other` is another instance of the same provider that you attached, which means that `this.getClass() == other.getClass()` as long as `other` is not null.
#### Attached Cap Cloning
ItemStack and Players can be copied.  Stacks are copied very frequently, and players are copied whenever they respawn.  Previously, stack caps would be copied through serialization, re-attachment, and deserialization, which was inefficient.  Player caps required manual copying though `PlayerEvent.Clone`.  Now, both cases are handled through the subinterface `ICopyableCapabilityProvider`, which has the following method:
```java
@Nullable ICopyableCapabilityProvider<O> copy(O copiedParent);
```
When an object is copied, all attached providers will be given the chance to clone themselves.  The `AttachCapabilitiesEvent` will NOT be fired for copied objects.  Returning null to this method means your capability will not be copied to the new target.
#### Capability Dispatcher
The time complexity of `CapabilityDispatcher#getCapability` has been improved.  Prior to this PR, for any capability, the dispatcher would have to query each individual attached provider for any call to getCap, meaning it was O(n) with the number of providers.  Now it does a map lookup and only queries the provider registered to that capability type, making it O(1).
### Invalidation and Revival
Invalidation `invalidateCaps()` and Revival `reviveCaps()` are now properly implemented across all game objects which support it (which is everything except `ItemStack`).  Prior to this PR, invalidation was sporadic and unreliable.  Attached capabilities were never invalidated and could not be revived.  Levels and Chunks were never invalidated (they are now invalidated upon unloading).  Additionally, a proper contract has been established for what invalidation/revival means, so that modders can properly implement these functions to allow for an ecosystem where caching `Capability` instances is reliable.  This is described in the javadocs of `ICapabilityProvider`.
### End Notes
It is possible I may have missed, overlooked, or under-described something in this write-up.  For complete information about functions or how things work, refer to the diffs and the javadocs within them.


Shadows-of-Fire:
docs have been updated for all classes in `net.minecraftforge.common.capabilities`

josephtarbit:
Big fan of this, looks really good.

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
```diff
@@ -0,0 +1,140 @@
...
+ * The AttachCapabilitiesEvent is used to attach {@link IAttachedCapabilityProvider}s to game objects.<br>
+ * The firing of this event is determined by the provider object, and is not consistent across usages.<br>
+ * For most cases, this is fired whenever the {@link CapabilityDispatcher} must be resolved.<br>
+ * This means that typically, this event is not fired until capabilities are attempted to be accessed for a particular object.
+ */
+public abstract class AttachCapabilitiesEvent<T extends ICapabilityProvider> extends Event
+{
+
+    protected final T obj;
+    protected final Map<CapabilityType<?>, IAttachedCapabilityProvider<?, T>> providers = new HashMap<>();
```
Technici4n: Should use IdentityHashMap for better performance.

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
```diff
@@ -0,0 +1,140 @@
...
+     * Additionally, the ID of all providers must be unique across the all {@link CapabilityType}s.
+     * 
+     * @param provider The provider being attached to this game object.
+     * @return true, if this provider was successfully attached, otherwise false.
+     */
+     public boolean addCapability(IAttachedCapabilityProvider<?, T> provider)
+     {
+        CapabilityType<?> type = provider.getType();
+        // Review Note : Potentially throw on duplicate name?  Would indicate major issues.
+        if(this.providers.containsKey(type) || this.byName.containsKey(provider.getId())) return false;
```
Technici4n: Imo duplicate CapabilityTypes are fine to just ignore with `return false`, however duplicate names should be a programming error.

src/main/java/net/minecraftforge/common/capabilities/Capability.java
```diff
@@ -5,71 +5,345 @@
...
     /**
-     * @return The unique name of this capability, typically this is
-     * the fully qualified class name for the target interface.
+     * Some state of this Capability is based on the value of this object.<br>
+     * If this field is null, it means the supplier has not been resolved.<br>
+     * If this {@link Mutable} does not contain a value, the supplier was invalid.<br>
+     * If this {@link Mutable} contains a value, the supplier has been resolved.<br>
      */
-    public String getName() { return name; }
+    private Mutable<T> resolved;
```
Technici4n: This should be made `volatile` to ensure that double-checked locking is safe. (Yes it's an issue with the current system already, might as well fix it now).

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
Technici4n: I would prefer a multimap (map to list is fine too), it is nicer for dynamic caps, and still yields high performance improvements.

src/main/java/net/minecraftforge/common/capabilities/Capability.java
```diff
@@ -5,71 +5,345 @@
...
  *
- * The CapabilityManager is in charge of creating this class.
+ * @param <T> The type of the optional value.
+ * 
+ * @see {@link ICapabilityProvider#getCapability(CapabilityType, Direction)} for retrieving capabilities.
  */
-public class Capability<T>
+@ParametersAreNonnullByDefault
+@MethodsReturnNonnullByDefault
+public final class Capability<T>
```
Technici4n, line 51: Not sure about the renames, they introduce additional migration difficulty with little benefit. Maybe just `LazyOptional` to `CapabilityInstance`. Not very important though, I don't really care.

Technici4n, line 51: I still find the whole lazy approach useless, but I doubt this is going to change, so ignore this comment.

src/main/java/net/minecraftforge/common/capabilities/CapabilityDispatcher.java
```diff
@@ -5,139 +5,138 @@
...
-        if (parent != null) // Parents go first!
-        {
-            lstCaps.add(parent);
-            if (parent instanceof INBTSerializable)
-            {
-                lstWriters.add((INBTSerializable<Tag>)parent);
-                lstNames.add("Parent");
-            }
-        }
+        this.caps = other.caps.values().stream().map(old -> old.copy(newOwner)).collect(Collectors.toMap(IAttachedCapabilityProvider::getType, t -> t));
```
Technici4n: Please avoid any kind of stream, this code path is super hot, so make it as fast and explicit as possible.

src/main/java/net/minecraftforge/common/capabilities/CapabilityDispatcher.java
```diff
@@ -5,139 +5,138 @@
...
+     * 
+     * @param other The other dispatcher being checked against.
+     * @return If this dispatcher is equivalent to the other dispatcher.
+     * 
+     * @see {@link IAttachedCapabilityProvider#isEquivalentTo(IAttachedCapabilityProvider)}
+     */
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    public boolean isEquivalentTo(@Nullable CapabilityDispatcher<T> other)
+    {
+        Set<ResourceLocation> keys = new HashSet<>();
```
Technici4n: This is also a super hot code path, it should be done without any allocation.

src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
```diff
@@ -5,206 +5,139 @@
...
-    @VisibleForTesting
-    static boolean SUPPORTS_LAZY_CAPABILITIES = true;
 
-    private final @NotNull Class<B> baseClass;
-    private @Nullable CapabilityDispatcher capabilities;
-    private boolean valid = true;
+    @Nullable
+    private CapabilityDispatcher<T> attachedCaps = null;
+    protected boolean capsValid = true;
+    protected CompoundTag lazyCapNbt = new CompoundTag();
```
Technici4n: I think this can be made `null` since you're already checking for `!= null` below. It's also one less object allocation.

src/main/java/net/minecraftforge/common/capabilities/IAttachedCapabilityProvider.java
```diff
@@ -0,0 +1,107 @@
...
+     * If you want to attach to copied itemstacks, you MUST override this method.
+     * 
+     * @param copiedParent The newly copied parent object.
+     * @return A deep copy of this {@link IAttachedCapabilityProvider}, or null, if it should not be copied.<br>
+     *         Note that copies will not fire {@link AttachCapabilitiesEvent}, so you must use this method.
+     */
+    default @Nullable IAttachedCapabilityProvider<C, O> copy(O copiedParent)
+    {
+        return null;
+    }
```
Technici4n: Would it make sense to move this to a subclass such as `IItemStackCapabilityProvider`, and have the ItemStack event gather instances of `IItemStackCapabilityProvider` directly. Then you could make `copy` abstract to force implementors to override if it necessary. This is imo nicer than a silent failure to copy caps if the modders forgets it.

src/main/java/net/minecraftforge/common/capabilities/IAttachedCapabilityProvider.java
```diff
@@ -0,0 +1,107 @@
...
+     * <p>
+     * The default implementation of this method always returns true, meaning this provider
+     * will be ignored during comparison.
+     * <p>
+     * If your capability must be equivalent for itemstacks to merge, you MUST override this method.
+     * 
+     * @param other Another instance of this provider from another object, or null, if it was not present.
+     * @return true if these providers are equivalent, otherwise false.
+     */
+    default boolean isEquivalentTo(@Nullable IAttachedCapabilityProvider<C, O> other)
```
Technici4n: I think it would be reasonable for the default implementation to compare the results of serializing the caps, i.e.:
```java
if (other == null) return false;
return Object.equals(serializeCaps(new CompoundTag()), other.serializeCaps(new CompoundTag()))`
```

Then modders can override this if they have a more performant implementation that they wish to use.

src/main/java/net/minecraftforge/common/capabilities/ICapabilityProvider.java
```diff
@@ -6,30 +6,61 @@
...
-     * This should never be OVERRIDDEN, modders should only ever implement the sided version.
+    /**
+     * Checks if this capability provider is direction-sensitive.<br>
+     * The return of this method should be static, i.e. a provider should always or never be direction-sensitive.<br>
+     * Of the default providers, {@link ItemStack} is not direction sensitive.
+     * @return If directions have importance in {@link #getCapability(Capability, Direction)}
      */
-    @NotNull default <T> LazyOptional<T> getCapability(@NotNull final Capability<T> cap) {
-        return getCapability(cap, null);
+    default boolean isDirectionSensitive()
```
Technici4n: I don't think this is useful, you know that if you're attaching to `ItemStack`s the direction is useless in most cases. And there might be some use cases where a specific cap wants direction-awareness inside of an item. (e.g. something like an AE2 storage cell _could_ in theory be direction-aware if the ME Drive directly forwards the insertion request with the side context).

src/main/java/net/minecraftforge/common/capabilities/IComparableCapabilityProvider.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package net.minecraftforge.common.capabilities;
+
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Defines a Capability Provider that can be checked for equivalence with another provider.
+ */
+public interface IComparableCapabilityProvider<B extends IComparableCapabilityProvider<B>> extends ICapabilityProvider
```
Technici4n: These methods could maybe be made protected methods in CapabilityProvider that ItemStack overrides to make public, to reduce class and interface spamming.

Technici4n (commented):
I like the general direction this is going, much better than the other PR imo. Left a few comments.

src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
Shadows-of-Fire: Yeah, this needs a bit of cleanup.  I was also planning to drop the capNbt field on itemstack and direct it to this one

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
Shadows-of-Fire: Yeah I figured that would be the case, I think I'll update that accordingly.  Duplicate names should indicate a hard error.

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
Shadows-of-Fire: Without a reasonable use-case I'm hesitant to allow multiple providers control over the same capability type, since you can induce cases where a provider can attach but silently fail due to another earlier provider providing that capability type.

src/main/java/net/minecraftforge/common/capabilities/Capability.java
Shadows-of-Fire: It's a bit of port-work but it shouldn't be so bad, and it is much nicer for readability.  Still not certain if Capability (old LazyOptional) should be Capability or CapabilityInstance / CapabilityReference, kinda waiting on more opinions for that.

Shadows-of-Fire: Yeah that's not going anywhere, even if it was it would be out of scope for this PR.

josephtarbit: `Capability` is probably nicest although might be confusing when porting in which case `CapabilityReference` would be the next best imo.

src/main/java/net/minecraftforge/common/capabilities/IAttachedCapabilityProvider.java
Shadows-of-Fire: Maybe, it makes the API on the method a bit of a mess, though not impossible to resolve - since it would require changes to `AttachCapabilitiesEvent.ItemStack` that essentially makes it error if you don't attach the subclass, which would technically break the type-contract of addCapability.
Not saying it's not doable, but it might be ugly.

src/main/java/net/minecraftforge/common/capabilities/IAttachedCapabilityProvider.java
josephtarbit: I think it might be better to force modders to define their own implementation because I think most would overlook overriding it purely for performance.

Shadows-of-Fire: I was thinking about providing that default, but then it is likely people utilize that heavily rather than write a proper compare function, which is why I omitted it.

src/main/java/net/minecraftforge/common/capabilities/ICapabilityProvider.java
Shadows-of-Fire: Technically this is overridable by TE's and Entities, which are free to notify the universe that they do not have special capability returns on a side-wise basis.  It might end up just being annoying, however.

src/main/java/net/minecraftforge/common/capabilities/Capability.java
Shadows-of-Fire: I would say it's worth it to shift some burden on porting to have a better classname in the long run.

Shadows-of-Fire:
As an aside I'd like to ask @Sm0keySa1m0n if you know how much this would impact https://github.com/MinecraftForge/MinecraftForge/pull/8706
Would it make it (easier/harder/impossible) to achieve given the new design?

src/main/java/net/minecraftforge/common/capabilities/Capability.java
Shadows-of-Fire: @gigaherz is this a known/documented issue with LazyOpt?

src/main/java/net/minecraftforge/common/capabilities/IComparableCapabilityProvider.java
Shadows-of-Fire: They can just go onto IForgeItemStack, tbh

Shadows-of-Fire:
Further changes:

* `ItemStack#capNBT` removed in favor of direct usage of `CapabilityProvider#lazyCapNbt`
* `IComparableCapabilityProvider` removed and rolled into `IForgeItemStack`
* `AttachCapabilitiesEvent.ItemStacks` updated to only accept `IItemStackCapabilityProvider`
* `isEquivalentTo` and `copy` removed from `IAttachedCapabilityProvider` and added to subclass `IItemStackCapabilityProvider`
* `AttachCapabilitiesEvent` will now throw on duplicate names.
* Performance-critical methods on `CapabilityDispatcher` have been improved.

marchermans:
> 

I would suggest to rename `IItemStackCapabilityProvider` to `IComparableCapabilityProvider` there are likely other situations where this tech could and should be used.

Shadows-of-Fire:
I don't like IComparable for that simply because it also adds copy() on top of compare().

I'm not sure if we'll ever see a viable usecase for it other than stacks, but if it's really desired we can generify it.  Gonna make the code in dispatcher a nightmare though.

josephtarbit:
> As an aside I'd like to ask @Sm0keySa1m0n if you know how much this would impact https://github.com/MinecraftForge/MinecraftForge/pull/8706
> 
> Would it make it (easier/harder/impossible) to achieve given the new design?

Shouldn't cause any issues, it'll only require a few minor refactors.

josephtarbit:
> I don't like IComparable for that simply because it also adds copy() on top of compare().
> 
> 
> 
> I'm not sure if we'll ever see a viable usecase for it other than stacks, but if it's really desired we can generify it.  Gonna make the code in dispatcher a nightmare though.

Could it be used for player cloning or is that too far fetched?

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
Technici4n: This is useful in case multiple dynamic providers are attached.

src/main/java/net/minecraftforge/common/capabilities/IAttachedCapabilityProvider.java
Technici4n: Imo making life easier for modders should be the priority of any API.

src/main/java/net/minecraftforge/common/capabilities/ICapabilityProvider.java
Technici4n: That sounds like it will cause issues at some point, e.g. maybe I want to attach a direction-aware capability to all subclasses of `BaseContainerBlockEntity`, however some mod might just randomly decide to override that method and that would break my attachment. Imo the system should try to be as broad as possible instead of attempting to limit use cases.

src/main/java/net/minecraftforge/common/capabilities/IAttachedCapabilityProvider.java
josephtarbit: I wouldn’t say that implementing a compare function isn’t easy. I also think performance is almost equally important in the case of this API.

Technici4n: `areCapsCompatible` is not the performance issue at the moment, so it doesn't really make sense to ask modders to prematurely optimize it for now, imo.

src/main/java/net/minecraftforge/common/capabilities/Capability.java
Technici4n: It's a known issue. See also <https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html>... volatile fixes the issue since java 5, as mentioned at the very end of that article.

src/main/java/net/minecraftforge/common/capabilities/CapabilityDispatcher.java
```diff
@@ -5,139 +5,177 @@
...
-            ICapabilityProvider prov = entry.getValue();
-            lstCaps.add(prov);
-            if (prov instanceof INBTSerializable)
+            for(ResourceLocation id : this.byName.keySet())
             {
-                lstWriters.add((INBTSerializable<Tag>)prov);
-                lstNames.add(entry.getKey().toString());
+                var ourProv = (IItemStackCapabilityProvider) this.byName.get(id);
+                var theirProv = other == null ? null : (IItemStackCapabilityProvider) other.byName.get(id);        
+                if(ourProv == null && !theirProv.isEquivalentTo(null)) return false;
```
Technici4n: How can `ourProv` be null if it's retrieved from `this.byName`?

src/main/java/net/minecraftforge/common/capabilities/CapabilityDispatcher.java
```diff
@@ -5,139 +5,177 @@
...
+        for(Map.Entry<ResourceLocation, IAttachedCapabilityProvider<?, T>> entry : other.byName.entrySet())
         {
-            lstCaps.add(parent);
-            if (parent instanceof INBTSerializable)
+            IAttachedCapabilityProvider<?, T> copy = ((IItemStackCapabilityProvider) entry.getValue()).copy(newOwner);
+            // Ideally we would ensure the type and key don't change here, but it's an expensive check that we don't "need" to do.
+            caps.put(copy.getType(), copy);
+            byName.put(copy.getId(), copy);
+        }
+        this.caps = Collections.unmodifiableMap(caps);
```
Technici4n, line 68: I don't think this is necessary... is there any place where the maps are mutated or exposed?

Shadows-of-Fire: This is for consistency, so that the maps in the fields are always unmodifiable.

src/main/java/net/minecraftforge/common/capabilities/CapabilityDispatcher.java
Shadows-of-Fire: Ah, yeah, that's a remnant from the old method, will drop the dead if-statement

Shadows-of-Fire:
| Could it be used for player cloning or is that too far fetched?

Possibly, but it would require separating copy() from compare() to allow a subinterface for players that can just be copied.
Tricky part here is managing the revivals / invalidations.

I think stacks don't even properly manage revivals/invalidates right now.

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
Shadows-of-Fire: Well that's my point, I don't think we should allow multiple providers for a single type.

src/main/java/net/minecraftforge/common/capabilities/ICapabilityProvider.java
Shadows-of-Fire: Yeah, probably just going to drop it based on lack of utility.

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
Technici4n: Well once again my GenericInternalInventory to IItemHandler adapter would like to disagree. :P

Shadows-of-Fire:
Latest changeset

* `IItemStackCapabilityProvider` split into `IComparableCapabilityProvider` and `ICopyableCapabilityProvider`.
* `AttachCapabilitiesEvent.Players` added.  Attaching to `Player` now requires subclassing `ICopyableCapabilityProvider`.
* Caps that are attached to Players will now automatically be copied through `ICopyableCapabilityProvider#copy`.  Cloned players will NOT fire AttachCapabilitiesEvent, similar to cloned itemstacks.
* `ICapabilityProvider#isDirectionSensitive()` has been removed, as it didn't seem to provide any useful value.
* Default fluid item providers have been updated to implement `IItemStackCapablityProvider`.

Notes: `CapabilitiesTest` needs fixing and is currently broken, and invalidation procedures currently need to be checked to make sure they work properly.

Shadows-of-Fire:
`ItemStack` will now invalidate / revive as necessary as the empty cache flag changes.  As of now, stacks, entities, and block entities will now properly manage invalidations / revivals.  Chunks and Levels have no such handling, and I'm not entirely sure where to put it.

Shadows-of-Fire:
Upon further discussion, it has been deemed not useful for `ItemStack` to invalidate/revive, since there is no meaning to "when" a stack becomes invalid.  Instead, stacks will just stop returning caps when empty.

Shadows-of-Fire:
Forgot to genPatches before that last commit, will fix tomorrow

Shadows-of-Fire:
Fixed up and ran the capabilities test, was able to verify that comparing, copying, stacking, and all functions of dispatcher/provider/attach work as intended.
With that, I think this is ready for full review.

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
```diff
@@ -0,0 +1,203 @@
...
+    /**
+     * Adds a {@link IAttachedCapabilityProvider} to this object.<br>
+     * Only one provider may exist for any given {@link CapabilityType}, on a first-come-first-serve basis.<br>
+     * Additionally, the ID of all providers must be unique across the all {@link CapabilityType}s.
+     * 
+     * @param provider The provider being attached to this game object.
+     * @return true, if this provider was successfully attached, otherwise false.
+     * @throws UnsupportedOperationException If a provider with the same ID already exists.
+     */
+    public boolean addCapability(IAttachedCapabilityProvider<?, T> provider)
```
alcatrazEscapee: So, looking at this, I have a use case where I have an item stack that exposes four separate capabilities through the same object: a fluid handler, an item handler, and a heat handler (a capability unique to my mod). Currently, this looks like this:

```java
static class Instance implements ICapabilityProvider, ICapabilitySerializable, IItemHandler, IFluidHandler, IFluidHandlerItem, IHeatHandler
```
This single instance gets attached to the item stack, and returns all the above capabilities when queried.

Under this new system, (correct me if I'm wrong), this would **not be the case**. I would be required to create a separate `IAttachedCapabilityProvider` for each capability, which then just acts as a proxy to look into the internal capability? Meaning I now am effectively creating my custom capability object `Instance` above, and four other other custom proxy objects which then get attached.

This seems... inconvenient, to say the least. Is there not a way this can support attaching single objects which can provide multiple capabilities without this indirection?

src/main/java/net/minecraftforge/common/capabilities/IAttachedCapabilityProvider.java
```diff
@@ -0,0 +1,137 @@
...
+         * <code>other != null</code>.
+         * 
+         * @param other Another instance of this provider from another object, or null, if it was not present.
+         * @return true if these providers are equivalent, otherwise false.
+         */
+        boolean isEquivalentTo(@Nullable IComparableCapabilityProvider<C, O> other);
+    }
+
+    /**
+     * Specialized subclass of {@link IAttachedCapabilityProvider} for use with {@link ItemStack}.
```
alcatrazEscapee: This comment (and most of the ones on this interface) seem to be specific to `ItemStack`, yet in `AttachCapabilitiesEvent.Players`, it specifies that these capabilities must be copyable. I would think this should be reflected in the documentation here.

src/main/java/net/minecraftforge/common/capabilities/CapabilityManager.java
```diff
@@ -5,48 +5,60 @@
...
+/**
+ * The Capability Manager is responsible for making and 
+ * providing {@link CapabilityType} instances.
+ */
 public enum CapabilityManager
 {
     INSTANCE;
-    static final Logger LOGGER = LogManager.getLogger();
 
+	private static final Logger LOGGER = LogManager.getLogger();
```
alcatrazEscapee: Tabs

src/main/java/net/minecraftforge/common/capabilities/CapabilityManager.java
```diff
@@ -5,48 +5,60 @@
...
 import static net.minecraftforge.fml.Logging.CAPABILITIES;
 
+/**
+ * The Capability Manager is responsible for making and 
+ * providing {@link CapabilityType} instances.
+ */
 public enum CapabilityManager
 {
     INSTANCE;
-    static final Logger LOGGER = LogManager.getLogger();
```
alcatrazEscapee: Make a decision - are you going to have a `enum` with single `INSTANCE`, and require everything to access the instance? Or are you just making everything static in which case it doesn't make any sense for this to be an `enum` in the first place.

src/main/java/net/minecraftforge/common/capabilities/CapabilityManager.java
```diff
@@ -5,48 +5,60 @@
...
     {
-        Capability<T> cap;
+        CapabilityType<T> cap;
 
-        synchronized (providers)
+        synchronized (PROVIDERS)
         {
-            realName = realName.intern();
-            cap = (Capability<T>)providers.computeIfAbsent(realName, Capability::new);
+            cap = (CapabilityType<T>)PROVIDERS.computeIfAbsent(id, CapabilityType::new);
```
alcatrazEscapee, line 49: This does no checks that the IDs are unique? Meaning this just happily allows you to registry replace capabilities?

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
Shadows-of-Fire: No, unfortunately.  Allowing a single object to provide multiple types is basically regressing back to the current system where attached providers have getCap(type, side), which means you have to iterate all of them whenever any cap is asked for.

alcatrazEscapee: The attached capabilities are still stored by type on the attached object though, are they not? Why can a provider not say "I provide capabilities A, B, C", which is then baked into the by-type map, except now the same instance is used for several capabilities. It doesn't need to be a runtime queryable check of which capabilities are provided.

Technici4n: (Note: you probably shouldn't expose both `IFluidHandler` and `IFluidHandlerItem` - only the latter should be used?)

src/main/java/net/minecraftforge/common/capabilities/CapabilityManager.java
Shadows-of-Fire: It does not need to - capabilities are not unique, nor can they be replaced.  it is design this way such that a mod can have a capability api where one mod does
cap = manager.get("mymod:cap")
and register it, and another mod can also have
otherModCap = manager.get("mymod:cap")

They can then query the `otherModCap` to see if the cap has been registered, instead of requiring field access to wherever the other mod is storing `cap`.

src/main/java/net/minecraftforge/common/capabilities/CapabilityManager.java
```diff
@@ -5,48 +5,60 @@
...
+ * The Capability Manager is responsible for making and 
+ * providing {@link CapabilityType} instances.
+ */
 public enum CapabilityManager
 {
     INSTANCE;
-    static final Logger LOGGER = LogManager.getLogger();
 
+	private static final Logger LOGGER = LogManager.getLogger();
+    private static final IdentityHashMap<ResourceLocation, CapabilityType<?>> PROVIDERS = new IdentityHashMap<>();
```
alcatrazEscapee: `IdentityHashMap` seems to be a mistake here.

src/main/java/net/minecraftforge/common/capabilities/CapabilityManager.java
alcatrazEscapee: My bad, I was thinking too much about the `IdentityHashMap` and forgot to leave a comment on that explicitly.

src/main/java/net/minecraftforge/common/capabilities/CapabilityManager.java
Shadows-of-Fire: Hm, yeah, not sure how that got there.  Think I got it mixed up with some other map that needs to be identity.

src/main/java/net/minecraftforge/common/capabilities/CapabilityManager.java
Shadows-of-Fire: Enum was legacy from the prior system, though there's not much point to have it.  I'll just drop it.

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
Shadows-of-Fire: I see the following problems with multityping attached providers
* The generic bound <C> has to be removed.
* getType() must return an array instead of a single object.
* The previous issue means that addCapability in AttachCapabilitiesEvent needs a higher level of complexity to attempt to add to multiple maps, and would need more expensive error reporting (rather than true/false success).
* All exposed capabilities are serialized under the same ID, and are not disjoint anymore.

There is a burden of complexity to allow a single attached object to provide multiple caps, and there is a desire to reduce complexity here.

Considering your design there - why do you have one object which is simultaneously an item handler, fluid handler, and heat handler?  Assuming these handlers have no interoperability, I see no reason they should be merged into a single class like this.

alcatrazEscapee: > Assuming these handlers have no interoperability

*Narrator: These handlers had interoperability.*

I'd still say this is inconvenient as compared to the current system, objectives of this PR notwithstanding. I'd either have to attach proxy capabilities, or be constantly querying capability A from capability B to know the actual state which ends up to be the same thing. If the point is really to "reduce complexity" and just shove that onto the modder, then sure, why not.

Shadows-of-Fire: You can still construct internal shared objects and wrap that object into the attached object, though I suppose the problem therein lies in cloning.  I'll see how complex it actually makes it to share objects like that.  I don't think the generics will like me very much.

Shadows-of-Fire:
The latest changeset re-adds the `CapabilityType` param to `IAttachedCapabilityProvider#getCapability` and allows the same provider object to be attached to multiple capability types.  The type passed into `getCapability` will only be one of the types that the provider has been registered to, meaning that providers which only attach to a single type may ignore the parameter.

`AttachCapabilitiesEvent#addCapability` has had a `CapabilityType` parameter added to support this behavior, and `IAttachedCapabilityProvider#getType` has been removed.

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
Shadows-of-Fire: fcee166

src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
```diff
@@ -5,206 +5,158 @@
...
-    {
-        return (B)this;
-    }
-
-    protected final @Nullable CapabilityDispatcher getCapabilities()
+    /**
+     * <b>This method will initialize capabilities when called.</b>
+     * @return The {@link CapabilityDispatcher}, or null, if nothing attached.
+     */
+    protected final @Nullable CapabilityDispatcher<T> getDispatcher()
```
kashike: I'd recommend naming this `getCapDispatcher`, just for consistency with the rest of the methods.
```suggestion
    protected final @Nullable CapabilityDispatcher<T> getCapDispatcher()
```

Shadows-of-Fire: Yep, good point.  I wasn't using that name because I was using it for a different function for the same signature, but that method was removed, freeing the name up.

Shadows-of-Fire:
The latest commit adds a much simpler modder-facing API for attaching capabilities to objects through a builder system.  Usage of this system would be like such
```java
 event.builder(() -> new ItemStackHandler(), CapabilityTypes.ITEMS)
     .withSerializer(INBTSerializable::serializeNBT, INBTSerializable::deserializeNBT)
     .withComparator(MyClass::areHandlersEqual)
     .withCopier(MyClass::cloneHandler)
     .attach(new ResourceLocation("my_mod_id", "my_items"));
```
The option to implement `IAttachedCapabilityProvider` is still available for additional complexity, but this system should make it much easier to attach "simple" objects.

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
```diff
@@ -10,18 +10,32 @@
...
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
+import java.util.function.BiPredicate;
+import java.util.function.Function;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import com.google.common.base.Preconditions;
+import com.google.common.base.Supplier;
```
sciwhiz12: This should be using the JDK Supplier, not the Guava Supplier.

Shadows-of-Fire: Unfortunately `Suppliers.memoize` will only accept the guava supplier, so to keep using that we have to use the guava variant.

SizableShrimp: Use Lazy.of then. But it should have a variant for jdk supplier and or you can always use supplier::get

Shadows-of-Fire: Should this be Lazy.of or Lazy.concurrentOf?  Capability (LazyOptional) has support for concurrent-safe resolution, but imo it's a bit inconsistent and not well defined anywhere if caps should be threadsafe (and its likely most caps are not threadsafe)

src/main/java/net/minecraftforge/common/capabilities/AttachCapabilitiesEvent.java
```diff
@@ -0,0 +1,473 @@
...
+        /**
+         * Validates, builds, and attaches the provider to all specified capability types.<br>
+         * This method has weak error resolution, meaning any failed attachments will be ignored.<br>
+         * If you need strong error handling, use {@link #build(ResourceLocation)} and 
+         * {@link AttachCapabilitiesEvent#addCapability(CapabilityType, IAttachedCapabilityProvider)} directly.
+         * @param id The ID of this provider.
+         */
+        public void attach(ResourceLocation id)
+        {
+        	ICompleteCapabilityProvider<T> prov = build(id);
```
alcatrazEscapee: Tabs

Shadows-of-Fire: I found-replaced all the tabs and yet there's still some there.  Bleh.

marchermans:
@Shadows-of-Fire could you write a much simpler write up of this PR somewhere, explicitly as to what the state of the API is currently.

This changes has gone what I feel like at least 20 different versions with different approaches to the problem, and slowly I am loosing track of what was now implemented and what was discarded for what ever reason.

Shadows-of-Fire:
Will do.  A lot has been buried in this and the previous PR, as well as in the discord channel.  I'll try my best to summarize.

Shadows-of-Fire:
@OrionDevelopment The main post of this PR has been updated to reflect the current state in detail.

Shadows-of-Fire:
Marking this PR as RFC again as the target for the cap changes has been set to 1.20.  Hopefully this can serve as a discussion board for what people want from this system and what changes need to be made to get there.

src/main/java/net/minecraftforge/common/capabilities/Capability.java
```diff
@@ -5,71 +5,341 @@
...
+        }
+        return resolved.getValue();
+    }
+
+    private T getValueUnsafe()
+    {
+        T ret = getValue();
+        if (ret == null)
+        {
+            throw new IllegalStateException("LazyOptional is empty or otherwise returned null from getValue() unexpectedly");
```
Alpvax, line 136: LazyOptional -> Capability in the exception message

Alpvax (changes requested):
Missed LazyOptional -> Capability rename.
LazyOptionalTest should probably be renamed (maybe to LazyCapabilityTest rather than just CapabilityTest for clarification).

tmvkrpxl0:
Will this actually land on 1.20 once it comes out?

marchermans:
> Will this actually land on 1.20 once it comes out?

It is currently not planned.

PR #8745, opened 2022-06-16
[1.19.x] Make command argument types a forge registry

SizableShrimp:
This PR makes command argument types (classname `ArgumentTypeInfo`) a non-saving, syncing forge registry.
This PR also changes gametest argument type infos to always register on a forge client and server.
Before this PR, trying to join a forge/vanilla server with gametest turned off while the client has gametests enabled results in this incomprehensible error: 
It turns out this is due to the fact that `ClientboundCommandsPacket` serializes and deserializes data `ArgumentTypeInfo`s using their registry int id.
This causes issues when the gametest arguments are registered on the client, as it shifts anything registered after it by 2.
All modded entries are registered after the vanilla bootstrap, and this includes forge defaults `forge:enum` and `forge:mod`.
By shifting the int id, all argument type infos registered after `minecraft:test_argument` and `minecraft:test_class` deserialized on the client are the wrong ones, which do not properly read the bytes in the packet buffer.
The proper fix is to:
1. Make command argument types into a syncing forge registry which matches up the int ids.
2. Always register gametest arguments to allow connecting with or without gametest commands enabled.

Because the registry snapshot is sent and injected on the client before `ClientboundCommandsPacket` is ever sent, this works out nicely.
Additionally, we must move the argument type info `minecraft:uuid` to be registered before the gametest arguments, to provide interop with vanilla <-> forge connections as those will not properly sync int ids and cause confusing suggestions for commands which use the uuid argument type (specifically in the vanilla command `/attribute <entity> <attribute> modifier add <uuid>`). Example without moving when `minecraft:uuid` is registered: 

Some testing has been done.
I have found vanilla client <-> forge server with my PR to function correctly (including joining, playing, and writing commands).
I have found forge client <-> vanilla server with my PR to function correctly, besides the issue presented above with `/attribute`. This is a very minor issue, it is purely visual on the client with the suggestions, as hitting enter will have the UUID processed as normal on the server.
This reordering of `minecraft:uuid`, `minecraft:test_argument`, and `minecraft:test_class` does not present a concern of unconsumed bytes in the packet buffer for `ClientboundCommandsPacket`, because all 3 send 0 bytes as they are all instances of `SingletonArgumentInfo`. So it is just a visual bug that nothing can be done about.

Forge client on/after my PR (regardless of gametest enabled or not) trying to connect to a forge server before my PR (with gametest disabled, aka. production) __will always present the original incomprehensible disconnection error__, as we now always register the gametest arguments (see point 2) which will always shift ids on the client.
Forge client before my PR (with gametest disabled) connecting to a forge server after my PR (regardless of gametest enabled or not) presents a different incomprehensible error but in the same vein as the `ClientboundCommandsPacket` internal exception disconnection screen due to the client-vs-server shift in ids for modded entries (see screenshot [here](https://i.imgur.com/yvMqfwY.png)).
Since this is still a breaking change period, I find this to be valid and a necessary change.
__This will require people to either upgrade/downgrade their servers or upgrade/downgrade their clients, as playing across this barrier will not be possible.__


SizableShrimp:
Merged internally into 1.19.1 dev work. This makes a lot more sense to change then alerady with a netbreak so it's in 42.0.0 from the start.

PR #8747, opened 2022-06-16
Fire ChunkWatchEvents after sending packets

SquidDev:
Sorry for submitting this breaking change so close to the deadline!

A common use-case for the `ChunkWatchEvent.Watch` event is to sync chunk-relevant data to the client. However, since Mojang's various rewrites of chunk loading, it's much more awkward to achieve this.

Currently, `ChunkWatchEvent.Watch` is fired when the player starts watching a chunk but before the chunk is actually loaded or sent to the client. This means mods either schedule a task next tick to sync their additional data or (incorrectly) forcibly load the chunk while handling the event (which defeats the whole point of async chunk loading!).

This change moves where the `ChunkWatchEvent`s events are fired from:
 - `ChunkWatchEvent.Watch` is now fired just after the chunk data is sent to the client.
 - `ChunkWatchEvent.UnWatch` is now fired after `ClientboundForgetLevelChunkPacket` is sent to the client.

For what it's worth, this matches the original semantics of this event - though I realise chunk loading has changed a lot since then so those behaviours may not be entirely desirable.

---

I'm also happy to keep the current events as is and add a new one which fires after chunk data is sent, if that's preferred?

### DIFF
--- a/patches/minecraft/net/minecraft/server/level/ChunkMap.java.patch
+++ b/patches/minecraft/net/minecraft/server/level/ChunkMap.java.patch
@@ -31,14 +31,14 @@
              this.m_63502_(chunkpos, compoundtag);
              this.m_140229_(chunkpos, chunkstatus.m_62494_());
              return true;
-@@ -848,6 +_,7 @@
+@@ -862,6 +_,7 @@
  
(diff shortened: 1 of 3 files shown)


marchermans (approved):
From having used these events myself: The current situation is awkward at best, this is a good change and makes using them way more reliable.

Tiny improvement:
Could you add a comment on the event that this is a server side only event, which can be used to sync additional data?

marchermans (approved):

PR #8749, opened 2022-06-16
Sanity check the version specified in the mod file

### LINKED ISSUE #8748: Faulty mod versions can crash both minecraft and java resulting in no crash report or latest.log files
**Minecraft Version:** 1.19

**Forge Version:** 41.0.35

**Logs:** No logs, that's the problem

**Steps to Reproduce:**
 1. Place a mod that has version set as "unspecified" in the mods folder. (MANIFEST.MF has Implementation-Version: unspecified)
 2. Start MC
 3. Get both half a crash report then Java crash (Exit code 1) in the middle of the crash report

**Description of issue:**
When you Place a mod that has version set as "unspecified" in the mods folder. (MANIFEST.MF has Implementation-Version: unspecified) minecraft will crash but in the middle of the crash report the java itself would crash with exit code 1 resulting in no crash report and no latest.log (sometimes it prints like the first two lines).
I know that a faulty mod is causing this but it should not crash bot minecraft and java at the same time, and should create a proper crash report.
Here is the crash report I managed to pull from launcher console:
https://paste.ee/p/W8glB

cpw:
Make sure it's compatible with JPMS standards for version strings.

Closes #8748


Su5eD:
I've tested this, and it appears that while the added check helps identify the invalid file, the exception from SJH still isn't logged.
As a workaround, we could set invalid file versions to a placeholder (e.g. `0.0BROKEN`) to avoid a crash, then throw an exception during modloading to display the `LoadingErrorScreen`.

cpw:
The SJH exception can never show up because it's deep inside the ModuleClassloading system. We have very little control and access there. The point was to _not_ get to that point. Does this fix not do that?

marchermans:
Note: For now this is on hold I think. @cpw you are still researching why the exception is not properly caught and propagated to the error screen correct?

cpw:
This is ready for merge..

marchermans (approved):

PR #8750, opened 2022-06-16
[1.19.x] Fix Water Animals not swimming in Custom Fluids

MrScautHD:
Fixes WaterAnimal swim not in Custom Fluid.

I Patched `Entity#isInWaterOrBubble` this is used just for Entities that lifes in Water and they need it to swim or for swim Animations.


TheCurle:
You were told in no uncertain terms to stop these small low effort PRs with no proof of the fix working or even proof that this is a problem. 

Since you seem incapable of doing this voluntarily, I'm going to make that happen for you.

PR #8753, opened 2022-06-16
Remove IForgeShearable, convert entities to use ToolActions.SHEARS_HARVEST

Random832:
Right now, Shears hooks are a little bit awkward because block actions use the ToolAction system [some of the blocks that dig faster and/or have special drops with shears are tagged by class with IForgeShearable, but this is never actually checked or used, at least not by Forge itself] but entity actions use IForgeShearable, and Dispensers have always used the vanilla Shearable interface.

This PR removes IForgeShearable and patches the Sheep, Mooshroom, and Snow Golem entities to use ToolActions instead, which leads to much lighter overall patch size in these entities and in ShearsItem. The secondary purpose of IForgeShearable, getting drops as a list rather than putting them in the world, can still be accomplished with the standard captureDrops hook.

Dispensers still use the Shearable interface, this is unaffected by this change and does mean that entities can't easily react to shears enchanted with Fortune being used in a Dispenser, but this is the status quo. If necessary I can look at what would be needed to add a hook for this (possibly reintroducing IForgeShearable in another form, as an extension to Shearable for entities only)

This is a breaking change, since it removes IForgeShearable.


patches/minecraft/net/minecraft/world/item/ShearsItem.java.patch
```diff
@@ -1,34 +1,12 @@
...
 +
 +   @Override
 +   public boolean canPerformAction(ItemStack stack, net.minecraftforge.common.ToolAction toolAction) {
 +      return net.minecraftforge.common.ToolActions.DEFAULT_SHEARS_ACTIONS.contains(toolAction);
 +   }
-+
-    public InteractionResult m_6225_(UseOnContext p_186371_) {
-       Level level = p_186371_.m_43725_();
-       BlockPos blockpos = p_186371_.m_8083_();
+ }
```
marchermans, line 12: This patch really threw me for a loop....

marchermans (approved):
Principally this is good.

Could you create another Draft PR which looks into figuring out if we can implement the same mechanic for Dispensers as you suggested?

LexManos:
Rejected as it defeats the entire point of IShearable.

SizableShrimp:
Removing this system in favor of tool actions would make it more cumbersome for automation purposes.

PR #8754, opened 2022-06-17
[1.19.x] Support Mob-Specific Behavior in Custom Fluids

ChampionAsh5357:
Fixes the interactions of mobs which implement specific methods of swimming and drowning in relation to custom fluids. Some behaviors of entities are also fixed to their entanglement with either swimming or drowning.

This is for the following mobs:
* axolotl
* cod
* dolphin
* drowned
* guardian
* elder guardian
* pufferfish
* salmon
* squid
* glow squid
* tropical fish
* frog
* bee
* magma cube
* slime
* tadpole
* turtle

### DIFF
--- a/patches/minecraft/net/minecraft/client/model/AxolotlModel.java.patch
+++ b/patches/minecraft/net/minecraft/client/model/AxolotlModel.java.patch
@@ -1,0 +1,11 @@
+--- a/net/minecraft/client/model/AxolotlModel.java
++++ b/net/minecraft/client/model/AxolotlModel.java
+@@ -81,7 +_,7 @@
+          this.m_170388_(p_170395_);
+       } else {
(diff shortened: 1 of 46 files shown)


marchermans (approved):
Principally this is okey, it fixes the underlying issue.
However this links some of the dolphin abilities to the fluids swim ability, not sure if more or less specific endpoints should be designed into the API.

Regardless. Approved.

LexManos:
Delaying this for later, the patches could be done A LOT simpler/cleaner by reworking some of the logic to a central place.
We can talk more about it on discord, but A lot of the patches are doing the exact same thing, modifying the isInWater call on the entity to add a fluid check. And technically, you're logically wrong as the value for that i the cached value not the exact world value. And threads and other weird stuff in custom implementations can be a hazard. Like I said we can talk more.

ChampionAsh5357:
Gonna close this for now as I plan to take a second look later. Since we transitioned over to teamcity, want to make use of that and get some smaller PRs going to address the issues reported over the last while.

PR #8756, opened 2022-06-17
Remove ForgeWorldPreset and related code

sciwhiz12:
This PR removes all remaining traces of `ForgeWorldPreset` that I could find. The functionality of `ForgeWorldPreset` was removed during the 1.19 update and patch cycle, leaving `ForgeWorldPreset` to be effectively dead code, mistakenly left in.

To delineate the changes:
 - `ForgeWorldPreset` was removed entirely.
 - `GameData`, `ForgeRegistries`, and `ForgeDeferredRegistriesSetup` were modified to remove the existing registry for `ForgeWorldPresets`, formerly accessible at `ForgeRegistries#WORLD_TYPES`.
 - The (unused) hook for the default world preset in `ForgeHooks` was removed.
 - The common Forge configuration (`ForgeConfig.Common`) was emptied out, its sole previous member being the `defaultWorldType` config setting.
   I elected to keep the config file instead of removing the whole section. If requested, I can remove the entire `Common` class entirely.
 - The `ForgeWorldTypeTest` test mod -- whose code was previously commented out by me -- was removed.


TheCurle:
am a derp. 
Thanks!

PR #8757, opened 2022-06-17
Fix SP-Devtime world loading crash due to missing server configs

marchermans:
When #8236 was merged, dev-time config access started check for the loading state of configuration entries.

If a single player world was opened straight from the main menu the configuration entry for handling empty ingredients was not loaded yet, since this is a server config yet no server exists at this point in time, causing a hard crash.

This PR removes the config option entirely and enables the behaviour by default.
Discussed internally.

Validated that the world now loads. Additionally checked that the advancement cleanup (Which has a similar check) now works, which it does, since that is run at a point where the server is loaded and configs have been synced.

### DIFF
--- a/patches/minecraft/net/minecraft/world/item/crafting/ShapelessRecipe.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/crafting/ShapelessRecipe.java.patch
@@ -59,7 +59,7 @@
           for(int i = 0; i < p_44276_.size(); ++i) {
              Ingredient ingredient = Ingredient.m_43917_(p_44276_.get(i));
 -            if (!ingredient.m_43947_()) {
-+            if (net.minecraftforge.common.ForgeConfig.SERVER.skipEmptyShapelessCheck.get() || !ingredient.m_43947_()) {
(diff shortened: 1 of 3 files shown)


patches/minecraft/net/minecraft/world/item/crafting/ShapelessRecipe.java.patch
```diff
@@ -59,7 +59,7 @@
           for(int i = 0; i < p_44276_.size(); ++i) {
              Ingredient ingredient = Ingredient.m_43917_(p_44276_.get(i));
 -            if (!ingredient.m_43947_()) {
-+            if (net.minecraftforge.common.ForgeConfig.SERVER.skipEmptyShapelessCheck.get() || !ingredient.m_43947_()) {
++            if (true || !ingredient.m_43947_()) { //We skip checking if an ingredient is empty during shapeless recipe deserialization to prevent complex ingredients from caching tags too early. Can not be done using a config value due to sync issues.
```
sciwhiz12: Comment should have `FORGE:`, and (as a matter of preference) it should use the imperative tone:
```suggestion
+            if (true || !ingredient.m_43947_()) { // FORGE: Skip checking if an ingredient is empty during shapeless recipe deserialization to prevent complex ingredients from caching tags too early. Can not be done using a config value due to sync issues.
```

sciwhiz12 (changes requested):

sciwhiz12 (approved):

PR #8764, opened 2022-06-18
Fix JEI mdk

Edivad99:
Based on the changes made to the JEI, I have updated the build.gradle file of the mdk. [JEI 10 guide](https://github.com/mezz/JustEnoughItems/wiki/Getting-Started-%5BJEI-10-or-higher-for-Forge-or-Fabric%5D)


SizableShrimp:
It would be better to get rid of these specific examples entirely to remove preferential treatment and instead use generic package names and artifact ids.

PR #8765, opened 2022-06-18
[1.19.x] The ServerTickEvent will now contain a reference to the MinecraftServer.

Budschie:
This PR adds the ability for modders to retrieve an instance of a `MinecraftServer` directly with a getter in the `ServerTickEvent` instead of having to call `ServerLifecycleHooks.getCurrentServer()`.

This greatly improves readability in my opinion, instead of having to write
```java
@SubscribeEvent
public static void onServerTick(ServerTickEvent event)
{
        System.out.println("Current tick: " + ServerLifecycleHooks.getCurrentServer().getTickCount());
}
```
you can just write
```java
@SubscribeEvent
public static void onServerTick(ServerTickEvent event)
{
        System.out.println("Current tick: " + event.getServer().getTickCount());
}
```

This also has the added advantage of presenting the modder with a more obvious solution; Knowing of the existence of `ServerLifecycleHooks.getCurrentServer()` is not trivial, one expects to find a reference to a `MinecraftServer` in the `ServerTickEvent`.

### DIFF
--- a/src/main/java/net/minecraftforge/event/TickEvent.java
+++ b/src/main/java/net/minecraftforge/event/TickEvent.java
@@ -59,9 +59,9 @@ public class TickEvent extends Event
          * @return Returns the {@link MinecraftServer} that is currently ticking.
          */
         public MinecraftServer getServer()
-		{
-			return server;
-		}
+        {
+            return server;
+        }
     }
 
     public static class ClientTickEvent extends TickEvent {
--- a/src/main/java/net/minecraftforge/event/TickEvent.java
+++ b/src/main/java/net/minecraftforge/event/TickEvent.java
@@ -56,7 +56,7 @@ public class TickEvent extends Event
         }
         
         /**
-         * @return Returns the {@link MinecraftServer} that is currently ticking.
+         * {@return the {@link MinecraftServer} instance}
          */
         public MinecraftServer getServer()
         {
--- a/src/main/java/net/minecraftforge/event/TickEvent.java
+++ b/src/main/java/net/minecraftforge/event/TickEvent.java
@@ -6,6 +6,7 @@
 package net.minecraftforge.event;
 
 import net.minecraft.world.entity.player.Player;
+import net.minecraft.server.MinecraftServer;
 
 import java.util.function.BooleanSupplier;
 
@@ -35,11 +36,13 @@ public class TickEvent extends Event
 
     public static class ServerTickEvent extends TickEvent {
         private final BooleanSupplier haveTime;
+        private MinecraftServer server;
 
-        public ServerTickEvent(Phase phase, BooleanSupplier haveTime)
+        public ServerTickEvent(Phase phase, BooleanSupplier haveTime, MinecraftServer server)
         {
             super(Type.SERVER, LogicalSide.SERVER, phase);
             this.haveTime = haveTime;
+            this.server = server;
         }
 
         /**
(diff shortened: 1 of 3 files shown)


src/main/java/net/minecraftforge/event/TickEvent.java
```diff
@@ -35,11 +36,13 @@ public TickEvent(Type type, LogicalSide side, Phase phase)
 
     public static class ServerTickEvent extends TickEvent {
         private final BooleanSupplier haveTime;
+        private MinecraftServer server;
```
sciwhiz12: This field should be `final`, given that nothing else modifies it outside of the constructor.
```suggestion
        private final MinecraftServer server;
```

src/main/java/net/minecraftforge/event/TickEvent.java
```diff
@@ -51,6 +54,14 @@ public boolean haveTime()
...
             return this.haveTime.getAsBoolean();
         }
+        
+        /**
+         * @return Returns the {@link MinecraftServer} that is currently ticking.
+         */
+        public MinecraftServer getServer()
+		{
+			return server;
+		}
```
sciwhiz12: Formatting: indentation is 4 spaces, not tabs.
```suggestion
        {
            return server;
        }
```

src/main/java/net/minecraftforge/event/TickEvent.java
```diff
@@ -51,6 +54,14 @@ public boolean haveTime()
         {
             return this.haveTime.getAsBoolean();
         }
+        
+        /**
+         * @return Returns the {@link MinecraftServer} that is currently ticking.
+         */
```
sciwhiz12:
```suggestion
        /**
         * {@return the server instance}
         */
```

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/event/TickEvent.java
Budschie: Manually merged: 610a74eff95f16fc9a7bc3090622fe02c3115074 (I replaced the `server` bit with `{@link MinecraftServer}`.

sciwhiz12: Technically, `{@link MinecraftServer}` would be redundant as the return type already specifies that information.

Budschie: I agree that this is a bit redundant, though I think that one should be able to easily understand a method just by taking a glance at the Javadoc so that you don't have to look at the method signature as well. The Javadoc of the JDK does seem to contain this type of redundancy too a few times (e.g. `Integer#describeConstable`).

At the end though, it is up to you to decide whether I should keep the redundant link or not, I'll happily change the Javadoc if you wish so.

sciwhiz12: My preference is to leave out the redundant link, since the method signature is always visible alongside the javadoc (whether in web, IntelliJ IDEA, or Eclipse) and the developer is likely already aware of the type of "the server instance" in this event.

Budschie: Alright, I've updated the Javadoc. It should now line up with the change you've suggested above. e585195ce6a30423d4a6d03fe996b1a6717a95b2

sciwhiz12 (approved):

PR #8766, opened 2022-06-18
Method parity between vanilla registries and Forge registry wrappers

### LINKED ISSUE #8758: [1.19] NamespacedDefaultedWrapper missing method overrides
There are methods in the vanilla DefaultedRegistry that Forge's NamespacedDefaultedWrapper does not override to supply correct values. 

The two methods not implemented from the parent classes are `getOrCreateHolderOrThrow` and `registryKeySet`.

I would be happy to make a PR, if appropriate

dhyces:
Fixes #8758. This changes the implementation of `getOrCreateHolder` to more closely match the vanilla implementation (it can return an errored DataResult in cases where the registry value is needed or registry is frozen, rather than throwing) and moves the original implementation to `getOrCreateHolderOrThrow`. We also add the method `getResourceKeys` to IForgeRegistry. This also extends the relevant methods in NamespacedWrapper and NamespacedDefaultedWrapper.

As a side-note, I had a case before where I mistakenly did not call `ResourceKey#location` and when the method returned the errored DataResult, the error did not happen to have much helpful information regarding what the key and registry was. Perhaps when the holderLookup field is nonnull it should still say what key was attempted in both "getOrCreateHolder..." methods?

### DIFF
--- a/src/main/java/net/minecraftforge/registries/IForgeRegistry.java
+++ b/src/main/java/net/minecraftforge/registries/IForgeRegistry.java
@@ -46,6 +46,7 @@ public interface IForgeRegistry<V> extends Iterable<V>
     @NotNull Optional<ResourceKey<V>> getResourceKey(V value);
 
     @NotNull Set<ResourceLocation>         getKeys();
+    @NotNull Set<ResourceKey<V>>           getResourceKeys();
     @NotNull Collection<V>                 getValues();
     @NotNull Set<Entry<ResourceKey<V>, V>> getEntries();
 
--- a/src/main/java/net/minecraftforge/registries/IForgeRegistry.java
+++ b/src/main/java/net/minecraftforge/registries/IForgeRegistry.java
@@ -46,7 +46,6 @@ public interface IForgeRegistry<V> extends Iterable<V>
     @NotNull Optional<ResourceKey<V>> getResourceKey(V value);
 
     @NotNull Set<ResourceLocation>         getKeys();
-    @NotNull Set<ResourceKey<V>>           getResourceKeys();
     @NotNull Collection<V>                 getValues();
     @NotNull Set<Entry<ResourceKey<V>, V>> getEntries();
 
--- a/src/main/java/net/minecraftforge/registries/NamespacedHolderHelper.java
+++ b/src/main/java/net/minecraftforge/registries/NamespacedHolderHelper.java
@@ -95,7 +95,7 @@ class NamespacedHolderHelper<T>
         if (ref == null)
         {
             if (this.holderLookup != null)
-                return DataResult.error("This registry can't create new holders without value");
+                return DataResult.error("This registry can't create new holders without value. (requested key" + key + ")");
 
             if (this.frozen)
                 return DataResult.error("Registry is already frozen (trying to add key " + key + ")");
@@ -111,7 +111,7 @@ class NamespacedHolderHelper<T>
         return this.holdersByName.computeIfAbsent(key.location(), k ->
         {
             if (this.holderLookup != null)
-                throw new IllegalStateException("This registry can't create new holders without value");
+                throw new IllegalStateException("This registry can't create new holders without value. (requested key" + key + ")");
             if (this.frozen)
                 throw new IllegalStateException("Registry is already frozen (trying to add key " + k + ")");
             return Holder.Reference.createStandAlone(this.self, key);
--- a/src/main/java/net/minecraftforge/registries/NamespacedHolderHelper.java
+++ b/src/main/java/net/minecraftforge/registries/NamespacedHolderHelper.java
@@ -22,6 +22,7 @@ import java.util.stream.Stream;
 
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
+import com.mojang.serialization.DataResult;
 import net.minecraft.util.RandomSource;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -88,7 +89,24 @@ class NamespacedHolderHelper<T>
         return Optional.ofNullable(this.holders.get(value));
     }
 
-    Holder<T> getOrCreateHolder(ResourceKey<T> key)
+    DataResult<Holder<T>> getOrCreateHolder(ResourceKey<T> key)
+    {
+        Holder.Reference<T> ref = this.holdersByName.get(key);
+        if (ref == null)
+        {
+            if (this.holderLookup != null)
+                return DataResult.error("This registry can't create new holders without value");
+
+            if (this.frozen)
+                return DataResult.error("Registry is already frozen (trying to add key " + key + ")");
+
+            ref = Holder.Reference.createStandAlone(this.self, key);
+            this.holdersByName.put(key.location(), ref);
+        }
+        return DataResult.success(ref);
+    }
+
+    Holder<T> getOrCreateHolderOrThrow(ResourceKey<T> key)
     {
         return this.holdersByName.computeIfAbsent(key.location(), k ->
         {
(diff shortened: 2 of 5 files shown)


src/main/java/net/minecraftforge/registries/IForgeRegistry.java
```diff
@@ -46,6 +46,7 @@
     @NotNull Optional<ResourceKey<V>> getResourceKey(V value);
 
     @NotNull Set<ResourceLocation>         getKeys();
+    @NotNull Set<ResourceKey<V>>           getResourceKeys();
```
SizableShrimp: Exposing this is not necessarily for functionality. Please delete this from the interface, and make the relevant method in ForgeRegistry package-private.

src/main/java/net/minecraftforge/registries/NamespacedHolderHelper.java
```diff
@@ -88,7 +89,24 @@ Optional<Holder<T>> getHolder(T value)
...
     }
 
-    Holder<T> getOrCreateHolder(ResourceKey<T> key)
+    DataResult<Holder<T>> getOrCreateHolder(ResourceKey<T> key)
+    {
+        Holder.Reference<T> ref = this.holdersByName.get(key.location());
+        if (ref == null)
+        {
+            if (this.holderLookup != null)
+                return DataResult.error("This registry can't create new holders without value");
```
SizableShrimp: Can you add `" (requested key" + key + ")"` to the end of this error message like vanilla?
This additional context would also probably be good in the same error message in `getOrCreateHolderOrThrow`

SizableShrimp (changes requested):

src/main/java/net/minecraftforge/registries/NamespacedHolderHelper.java
```diff
@@ -88,12 +89,29 @@ Optional<Holder<T>> getHolder(T value)
...
     }
 
-    Holder<T> getOrCreateHolder(ResourceKey<T> key)
+    DataResult<Holder<T>> getOrCreateHolder(ResourceKey<T> key)
+    {
+        Holder.Reference<T> ref = this.holdersByName.get(key.location());
+        if (ref == null)
+        {
+            if (this.holderLookup != null)
+                return DataResult.error("This registry can't create new holders without value. (requested key" + key + ")");
```
sciwhiz12: Small detail:
```suggestion
                return DataResult.error("This registry can't create new holders without value (requested key: " + key + ")");
```

src/main/java/net/minecraftforge/registries/NamespacedHolderHelper.java
```diff
@@ -88,12 +89,29 @@ Optional<Holder<T>> getHolder(T value)
...
+        return DataResult.success(ref);
+    }
+
+    Holder<T> getOrCreateHolderOrThrow(ResourceKey<T> key)
     {
         return this.holdersByName.computeIfAbsent(key.location(), k ->
         {
             if (this.holderLookup != null)
-                throw new IllegalStateException("This registry can't create new holders without value");
+                throw new IllegalStateException("This registry can't create new holders without value. (requested key" + key + ")");
```
sciwhiz12: Small detail:
```suggestion
                throw new IllegalStateException("This registry can't create new holders without value. (requested key: " + key + ")");
```

sciwhiz12 (changes requested):

sciwhiz12 (changes requested):
Also, please add `fixes #8758` to your PR description so this PR is linked to that issue automatically by GitHub.

src/main/java/net/minecraftforge/registries/NamespacedHolderHelper.java
```diff
@@ -88,12 +89,29 @@ Optional<Holder<T>> getHolder(T value)
...
+    Holder<T> getOrCreateHolderOrThrow(ResourceKey<T> key)
     {
         return this.holdersByName.computeIfAbsent(key.location(), k ->
         {
             if (this.holderLookup != null)
-                throw new IllegalStateException("This registry can't create new holders without value");
+                throw new IllegalStateException("This registry can't create new holders without value (requested key: " + key + ")");
             if (this.frozen)
                 throw new IllegalStateException("Registry is already frozen (trying to add key " + k + ")");
             return Holder.Reference.createStandAlone(this.self, key);
```
sciwhiz12: Could we reimplement this method to instead call `getOrCreateHolder` and unwrap the given `DataResult`, turning an error message into an `IllegalStateException` (perhaps using `LambdaExceptionUtils#rethrowConsumer` to bypass the pesky pain of checked exceptions)?

sciwhiz12 (changes requested):
The new method looks fine to me, but there's deduplication of code that can be done here. (Also, please rerequest my review when you've applied your changes. @SizableShrimp can't be rerequested for review because of GitHub limitations.)

src/main/java/net/minecraftforge/registries/NamespacedHolderHelper.java
sciwhiz12: I've been advised that the correct lambda exception utility class to suggest is <code>cpw.mods.modlauncher.api.Lam<i>db</i>aExceptionUtils</code> and **not** `cpw.mods.util.LambdaExceptionUtils` -- yes, the one with the misspell of `Lamdba`. Apparently, the one under `cpw.mods.util` is from securejarhandler, which does not export that package, causing an illegal access error during runtime.

dhyces: Right, so, there is a method in DataResult called "getOrThrow" that throws a RuntimeException. Should we use that and have a RuntimeException or have a more specific IllegalStateException?

SizableShrimp: I'd prefer to keep IllegalStateException

src/main/java/net/minecraftforge/registries/NamespacedHolderHelper.java
```diff
@@ -88,15 +90,31 @@ Optional<Holder<T>> getHolder(T value)
...
+    }
+
+    Holder<T> getOrCreateHolderOrThrow(ResourceKey<T> key)
+    {
+        return LamdbaExceptionUtils.uncheck(() ->
+        {
+            DataResult<Holder<T>> dataResult = getOrCreateHolder(key);
+            if (dataResult.error().isPresent())
+                throw new IllegalStateException(dataResult.error().get().message());
+            return dataResult.result().get();
```
sciwhiz12: You may have misunderstood what I meant by using `LamdbaExceptionUtils`. In this form,  `LamdbaExceptionUtils#uncheck` effectively does nothing as `IllegalStateException` is already an unchecked exception. What I meant to suggest was to call `DataResult#getOrThrow` like so:
```java
return getCreateHolder(key).getOrThrow(false, LamdbaExceptionUtils.rethrowConsumer((String message) -> {
    throw new IllegalStateException(message);
}));
```

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/registries/NamespacedHolderHelper.java
SizableShrimp: Is it even necessary in this form?

sciwhiz12: Actually, that's a fair point which I hadn't considered 👀 -- we don't even need `LamdbaExceptionUtils` because `IllegalStateException` is an unchecked exception.

dhyces: Yes, unfortunately, misunderstood its use. Figured we only wanted the illegal state, but I suppose it doesn't impact anything if we throw both. I'll push the revision now.

sciwhiz12 (approved):

PR #8775, opened 2022-06-20
Fix ticking chunk tickets from forge's chunk manager not causing chunks to fully tick

pupnewfster:
Either during the 1.17 or 1.18 port (I didn't check which), the extended forced ticking system that forge added in #7525 to allow tickets to force a chunk to tick broke. due to vanilla checking two maps, and the methods that had been copies of `addRegionTicket` and `removeRegionTick` never got updated to also add to `DistanceManager.tickingTicketsTracker`.  This PR fixes that by adjusting how the patches are implemented to add an overload argument determining whether the ticket should `forceTicks` so that if the body of the method changes in other ways then they don't have to be manually copied into duplicate methods.


patches/minecraft/net/minecraft/server/level/ServerChunkCache.java.patch
```diff
@@ -24,11 +24,11 @@
...
 +
 +   public <T> void registerTickingTicket(TicketType<T> type, ChunkPos pos, int distance, T value) {
-+      this.f_8327_.registerTicking(type, pos, distance, value);
++      this.f_8327_.addRegionTicket(type, pos, distance, value, true);
 +   }
 +
 +   public <T> void releaseTickingTicket(TicketType<T> type, ChunkPos pos, int distance, T value) {
-+      this.f_8327_.releaseTicking(type, pos, distance, value);
++      this.f_8327_.removeRegionTicket(type, pos, distance, value, true);
     }
```
sciwhiz12: Since we're already doing breaking changes, might as well rename these methods to `addForcedTickingRegionTicket` and `removeForcedTickingRegionTicket` to bring them in line with the `addRegionTicket` and `removeRegionTicket` methods immediately preceeding these and the renames you did in `DistanceManager`, as well as communicating that these are region tickets which force ticking.

We could also rename these to exactly match `addRegionTicket` and `removeRegionTicket`, adding the `forceTicks` parameter at the end and using obfuscated parameters (with corresponding EXC entries).

sciwhiz12 (changes requested):

sciwhiz12 (approved):

covers1624:
I can confirm this is also an issue in 1.18.2, backporting this PR fixes the issues as expected. 

Have not ported to 1.19 just yet, but will test this pr and update my comment.

PR #8780, opened 2022-06-21
[1.19] Fix certain user-configured options being overwritten

### LINKED ISSUE #8770: [1.19] Option reload in ClientModLoader breaks GUI scale setting
**Minecraft Version:** 1.19

**Forge Version:** 1.19-41.0.38

**Steps to Reproduce:**
 1. Start the game
 2. Maximize the game
 3. Set the `GUI Scale` setting to 3 or higher (but not Auto)
 4. Restart the game
 5. Maximize the game
 6. Observe that the `GUI scale` setting has been reset to 2

**Description of issue:**
Minecraft initially loads the options in the `Options` constructor and after that constructor returns, `Minecraft#running` is set to `true`. That flag changes the behaviour of most option validators.
Forge later reloads the options in `ClientModLoader.finishModLoading()`, which runs from a resource reload listener (and therefore also on every resource reload). Due to the aforementioned flag being true at that point in time, the GUI scale value read from the options file is clamped to the maximum value allowed by the current window size [^1], which is 2 for the default window size, overwriting any non-"Auto" value higher than 2 that was previously set by the user.
Commenting out [this line](https://github.com/MinecraftForge/MinecraftForge/blob/1.19.x/src/main/java/net/minecraftforge/client/loading/ClientModLoader.java#L130) prevents the issue.

[^1]: This is the GUI scale "validator": `!minecraft.isRunning() ? 2147483646 : minecraft.getWindow().calculateScale(0, minecraft.isEnforceUnicode());`

**Possible fix**
Forge needs to reload the options after mods are loaded to allow key binds added by mods to be properly saved and loaded because the initial option loading happens before mods can register their key binds.
Therefore it seems like the most sensible solution would be to restrict the option reloading to just the key binds.
It may also make sense to limit the option reload to only the actual end of mod loading instead of doing it on every resource reload.

XFactHD:
This PR fixes user-configured options being overwritten with incorrectly "corrected" values when Forge reloads options after mod loading is done and on every resource reload. Fixes #8770.

I am opening this as a draft for now because I have not found a way to fully test the impact of this patch and I am also not particularly happy with this solution, so critical feedback would be highly appreciated.

### DIFF
--- a/patches/minecraft/net/minecraft/client/Options.java.patch
+++ b/patches/minecraft/net/minecraft/client/Options.java.patch
@@ -8,8 +8,22 @@
        this.f_92060_ = p_92138_;
        this.f_92110_ = new File(p_92139_, "options.txt");
        boolean flag = p_92138_.m_91103_();
-@@ -887,10 +_,14 @@
+@@ -807,6 +_,10 @@
+    }
+ 
+    private void m_168427_(Options.FieldAccess p_168428_) {
++      processOptions(p_168428_, false);
++   }
++   private void processOptions(Options.FieldAccess p_168428_, boolean onlyKeyBinds) {
++      if (!onlyKeyBinds) { //FORGE: prevent overwriting user's settings when reloading options after game startup
+       p_168428_.m_213982_("autoJump", this.f_92036_);
+       p_168428_.m_213982_("autoSuggestions", this.f_92037_);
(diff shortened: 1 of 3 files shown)


XFactHD:
In a discussion with Lex, a better solution was found and implemented and as such this is ready for review now.

PR #8781, opened 2022-06-21
[1.19.x] Add getter for correct BiomeSpecialEffectsBuilder to BiomeInfo$Builder

Commoble:
One of the intentions of the BiomeModifier api was to provide read access to the current and modified state of a biome to a BiomeModifier, so that the BiomeModifier can read the state to determine whether to make further modifications.

To facilitate this, forge extensions of vanilla's builders for climate and client special effects were used, to provide read access to the builders' fields (the vanilla builders have setters but no getters), in a similar manner to the generation and mob spawn builder extensions originally created for the BiomeLoadingEvent.

However, due to a clerical error, the getter for the special effects builder provides the vanilla builder type, not the forge extension. The actual implementation is still the forge extension, but mods have to cast to the forge builder to use its getters. This PR adds a getter to the API that returns the correct builder type.

### DIFF
--- a/src/main/java/net/minecraftforge/common/world/ModifiableBiomeInfo.java
+++ b/src/main/java/net/minecraftforge/common/world/ModifiableBiomeInfo.java
@@ -157,10 +157,10 @@ public class ModifiableBiomeInfo
             {
                 return effects;
             }
-            
+
             public BiomeSpecialEffectsBuilder getSpecialEffects()
             {
-            	return effects;
+                return effects;
             }
 
             public BiomeGenerationSettingsBuilder getGenerationSettings()
--- a/src/main/java/net/minecraftforge/common/world/ModifiableBiomeInfo.java
+++ b/src/main/java/net/minecraftforge/common/world/ModifiableBiomeInfo.java
@@ -152,7 +152,7 @@ public class ModifiableBiomeInfo
              * 
              * @deprecated Use {@link #getSpecialEffects()} as it provides read access. TODO remove this by 1.20
              */
-            @Deprecated(forRemoval=true)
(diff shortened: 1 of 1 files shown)


src/main/java/net/minecraftforge/common/world/ModifiableBiomeInfo.java
```diff
@@ -147,10 +147,21 @@ public ClimateSettingsBuilder getClimateSettings()
                 return climateSettings;
             }
 
+            /**
+             * {@return Builder for client special effects}
+             * 
+             * @deprecated Use {@link #getSpecialEffects()} as it provides read access. TODO remove this by 1.20
+             */
+            @Deprecated(forRemoval=true)
```
pupnewfster: This should have a since 1.19 in the annotation

src/main/java/net/minecraftforge/common/world/ModifiableBiomeInfo.java
```diff
@@ -147,10 +147,21 @@ public ClimateSettingsBuilder getClimateSettings()
...
+             */
+            @Deprecated(forRemoval=true, since="1.19")
             public BiomeSpecialEffects.Builder getEffects()
             {
                 return effects;
             }
+            
+            public BiomeSpecialEffectsBuilder getSpecialEffects()
+            {
+            	return effects;
```
sciwhiz12: Formatting: 4 spaces for indentation, not tabs
```suggestion
                return effects;
```

sciwhiz12 (changes requested):

sciwhiz12 (approved):

PR #8783, opened 2022-06-23
[DRAFT] ItemStack-sensitive method for shears in dispensers

Random832:
Dispensers currently have no way to send the level of fortune/looting to the entity being sheared. This could have been done in principle with the IForgeShearable interface but wasn't ever implemented to do so, and #8753 is a proposed change to remove that system in favor of tool actions.

Ultimately I'm not particularly happy with where this leaves block shearing - the fact that dispensers have never supported *all* vanilla block shearing actions [you can't e.g. carve a pumpkin, or trim vines, with a dispenser] makes it difficult to consider implementing this in a general way [two well-known mods that add custom beehives register their own replacement dispenser actions, and even prior to my other PR beehives never implemented IForgeShearable].

### DIFF
--- a/patches/minecraft/net/minecraft/core/dispenser/ShearsDispenseItemBehavior.java.patch
+++ b/patches/minecraft/net/minecraft/core/dispenser/ShearsDispenseItemBehavior.java.patch
@@ -1,0 +1,25 @@
+--- a/net/minecraft/core/dispenser/ShearsDispenseItemBehavior.java
++++ b/net/minecraft/core/dispenser/ShearsDispenseItemBehavior.java
+@@ -26,7 +_,7 @@
+       Level level = p_123580_.m_7727_();
+       if (!level.m_5776_()) {
+          BlockPos blockpos = p_123580_.m_7961_().m_121945_(p_123580_.m_6414_().m_61143_(DispenserBlock.f_52659_));
+-         this.m_123573_(m_123576_((ServerLevel)level, blockpos) || m_123582_((ServerLevel)level, blockpos));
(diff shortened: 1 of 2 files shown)


patches/minecraft/net/minecraft/core/dispenser/ShearsDispenseItemBehavior.java.patch
```diff
@@ -0,0 +1,25 @@
...
+-         this.m_123573_(m_123576_((ServerLevel)level, blockpos) || m_123582_((ServerLevel)level, blockpos));
++         this.m_123573_(m_123576_((ServerLevel)level, blockpos) || tryShearLivingEntity((ServerLevel)level, blockpos, p_123581_));
+          if (this.m_123570_() && p_123581_.m_220157_(1, level.m_213780_(), (ServerPlayer)null)) {
+             p_123581_.m_41764_(0);
+          }
+@@ -53,11 +_,11 @@
+       return false;
+    }
+ 
+-   private static boolean m_123582_(ServerLevel p_123583_, BlockPos p_123584_) {
```
KnightMiner, line 16: I believe there are problems with removing private methods, as someone could have ATed that method. Probably better to keep the old private one around and have it pass in `ItemStack.EMPTY` or `new ItemStack(Items.SHEARS)`

Also, since this method contains mapped parameters, you need to add the new method to the forge.exc file.

Shadows-of-Fire: Yeah - don't patch out methods unless explicitly necessary.  What you can do is add the new method right below the old one, and change the body of the old one to invoke the new one with `ItemStack.EMPTY` as the stack param.  This keeps the method in place and prevents the patch from looking really weird as well.

PR #8788, opened 2022-06-24
[1.18.x] Add LivingEntity damage source modification on hurt

rbasamoyai:
This PR adds a damage source modification event to the beginning of `LivingEntity#hurt` and `PlayerEntity#hurt`. The event in question is `LivingDamageSourceEvent`.

Use case: Modifying damage sources in certain ways, e.g. "wrapper" damage source that appends text to end of wrapped damage source's death messages. Actually, probably most likely just for easily modifying death messages.


rbasamoyai:
Ah my bad, forgot patches.

KnightMiner:
If the goal is modifying death messages, wouldn't a death message event be much cleaner than a third event fired every time an entity is damaged?

Also, pretty sure new PRs need to target 1.19 first, then can be backported to 1.18 after merge.

rbasamoyai:
Ah, didn't know about the new version merge thing
As for death messages; possibly; if there is some other potential use for `DamageSource` switching then perhaps this could be redone for 1.19. If not I will just put out a PR for a death message event only in 1.19.x

rbasamoyai:
yeah, closing this for now

PR #8790, opened 2022-06-24
Add tag_exists condition.

KingLemming:
It's a LOT cleaner than having to chain a forge:not with tag:empty and is the more useful thing to query, typically.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeMod.java
+++ b/src/main/java/net/minecraftforge/common/ForgeMod.java
@@ -541,6 +541,7 @@ public class ForgeMod
             CraftingHelper.register(OrCondition.Serializer.INSTANCE);
             CraftingHelper.register(TrueCondition.Serializer.INSTANCE);
             CraftingHelper.register(TagEmptyCondition.Serializer.INSTANCE);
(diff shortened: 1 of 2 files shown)


Technici4n (commented):
Is `not(tagEmpty(tagKey))` an issue? Ideally this would always be done via datagen anyway, and a helper function would be easy to add there.

If the PR is accepted in spite of that, I think it would be nice to make it work with any `TagKey`, not just `TagKey<Item>`. I would suggest not serializing the registry if it's the item registry, and otherwise serializing the registry key.

KnightMiner:
I have an example making TagEmptyCondition work with arbitrary registries here if you want code to adapt for this: <https://github.com/SlimeKnights/Mantle/blob/1.18.2/src/main/java/slimeknights/mantle/recipe/helper/TagEmptyCondition.java>

Shadows-of-Fire:
I would agree that this should be generified beyond strictly `TagKey<Item>`.  It might be acceptable to update tag_empty to be generified as well within the scope of this PR.

marchermans:
@KingLemming What is the state of this?

KingLemming:
The state is that it works and is good to go, but since people apparently want some genericized version, then they can write it instead. I'll stop wasting my time.

PR #8792, opened 2022-06-24
Add MC-105317 fix into patch to rotate entities in structures properly

TelepathicGrunt:
Mojang fixed the bug where entities spawned by structures do not rotate properly. Here's their bug report:
https://bugs.mojang.com/browse/MC-105317

Forge's patch into Structure Template is still using the old bugged way of rotating entities. 
The bugged old way: mirror + (rotEntity - rotStruct)
The Mojang fixed way: rotStruct + (mirror - rotEntity)

To test, put on this datapack and do `/locate structure minecraft:village_plains` and teleport to many villages. The datapack replaced the town center of these non-zombified villages to have a sad warden and sad creeper facing the same direction as the granite arrow.
[test_structure_rotation.zip](https://github.com/MinecraftForge/MinecraftForge/files/8976188/test_structure_rotation.zip)
image

Facing east with unpatched forge looking at a south rotated piece with warden and creeper facing north.
image

Patched forge with the same conditions as above:
image

This will need to be backported to 1.18.2 as mojang fixed the bug there but forge hasn't.


sciwhiz12 (approved):
Tested to work with provided datapack.

PR #8786, opened 2022-06-23
[1.19.X] Client code cleanup, updates, and other refactors

amadornes:
As has been discussed both publicly and internally, several parts of the client codebase have been in need of maintenance in one way or another for some time now. The team has a lot of responsibilities and it's just not possible for them to dedicate all their time and energy for close to 15 hours a day over the span of 3 weeks to cleaning things up and getting them up to spec.

But as it turns out, I do lol

### This PR is the result of a larger effort with several goals:
 - Audit and document the purpose of all current client code. This provides better insight into what exactly we're dealing with
 - Clean up the codebase, removing everything that is not in use and updating older systems to match current standards and user demands. This removes tech debt and improves both maintainability and user-facing accessibility. Some examples:
     - Adapt all code to current naming conventions
     - Reorganize and rename classes and members to improve organization and legibility
     - Move away from static registration in `FMLClientSetupEvent` and over to dedicated registration events
     - Remove old vertex pipelines and substitute them with new ones based on vanilla's `VertexConsumer` interface
     - Decouple flat and shaded lighting pipelines since they are completely unrelated, and make them into vertex producers instead of part of the vertex pipeline
     - Improve functionality of vanilla's model display transforms, removing the need for Forge's largely overlapping replacement
 - Introduce several breaking or semi-breaking changes, like:
     - Make `IModelData` into a final class backed by an immutable map, solving thread safety concerns
     - Move model data manager ownership to `ClientLevel`, solving issues with multiple active client levels
     - Make render types a first-class citizen in `BakedModel`. This also involves transferring over control of render types from `ItemBlockRenderTypes`, and may result in a slight speedup in chunk geometry baking
 - Provide clear and concise documentation for the most relevant elements of the client API as well as an update guide (WIP)

The above listed changes are not an exhaustive list and more can be found in the comments of each commit.

### Non-goals of this PR:
 - Clean up hooks in patches. This is very much needed, as a large number of events and other hooks are dispatched directly from Minecraft classes instead of through `ForgeHooksClient`, but is outside of the scope of this PR as these would be non-breaking changes for developers and could therefore be safely introduced mid-version
 - Optimize existing code for speed and/or cleanliness. While minor improvements have been made in multiple areas, this is generally just a side effect of the updating/upgrading effort, and not a direct focus of this PR

### Code organization:
[...]

PR #8794, opened 2022-06-24
Added ItemDecorator Map

justliliandev:
This is a retry of #8070 and #8079
So beforehand I'll reply to some of the comments of those PRs, if applicable:

DieSieben07 on #8079
> Who has the responsibility to add the decorator capability? How are mods supposed to add decorators to "foreign" items? If they add the capability using AttachCapabilitiesEvent this can cause issues if the capability is already added by the item itself or other mods.

To solve this, there is a helper method in ForgeHooksClient to add an IItemDecorator to the ItemStack, during AttachCapabilitiesEvent, so no one has to check if the cap is present and add a provider themselves if it isn't.

ChampionAsh on #8079 
> I would suggest implementing a positioning system similar to https://github.com/MinecraftForge/MinecraftForge/pull/7770.

I don't think a positioning system is necessary, as the blitOffset (height) is given to the renderer and everyone can render at a height they want. A positioning System would require the rendering of the durability bar and the count to be made an IItemDecorator as well and as soon as every ItemStack needs this capability, a field in the ItemStack will always be faster.
If this is intended I can change that system to not use a capability, but some kind of positioned List

**Performance:**
Test done on a full chest gui (63 Slots in total) and all slots filled with the same item
With 2 IItemDecorations:
gui needed on average 4.2 ms or ca. 0.066ms per item
Without IItemDecorations:
gui needed on average 2.7ms or ca. 0.042ms per item


Shadows-of-Fire:
This is not a good way to go about doing this.  Render code should not be present or relevant to the server (which capabilities are),  Capabilities also add other complexity when present that is additional overhead that is strictly not necessary in a render context (the cost of getCap is significant to be running every frame).

justliliandev:
diesieben07 wrote in #8048 
> As an alternative to the event, we could add a capability that mods can add to either their own or other ItemStacks which would then be called to do this rendering. That way no event dispatching needs to happen but the system is still pluggable even for items from other mods or vanilla.

Which is why this thing was designed as a capability.
The only real thing exposed to the server is the registration of the capability, but as adding IItemDecorations is mostly done using a helper in ForgeHooksClient and rendering obviously happens clientside, mod authors should be able to know, that this code should only be used on the logical Client.

Getting the capability, once initialized, is just an array lookup through the ICapabilityProvider in CapabilityDispatcher until one returns the cap you want.
As shown by the performance test I did, I don't think it's as bad as you describe it to be (especially comparing this to RenderLivingEvent.Pre and Post). A List in the ItemStack class will be faster, but that is imo much more exposed on the serverside and has a higher maintenance cost.

Also during normal gameplay only 9 ItemStacks will be usually visible, so the impact in normal gameplay is more likely around/below 0.25ms on each frame

marchermans (changes requested):
This is a flatout no.
Capabilities are not a great idea to handle rendering, especially on itemstacks where they are already a problem.

Redesign this to be an event please.

justliliandev:
What kind of an event?
one fired for each ItemStack being rendered at the same patch location this PR does this atm similar to #7693 or #8048?
one fired for each gui layer (see #6927)?
or one for each frame rendered?
The first one might has performance concerns as previously commented to the two PRs, but I can create that and provide the data, if wanted.
The latter two have the added difficulty to gather all render positions [discord conversation](https://discord.com/channels/313125603924639766/852298000042164244/991317881981763604). This would require to get the position of dragged stacks, creativetab icons, merchantoffers etc. to be collected (and a way for mods to add their "slots"/rendered Items to that).
Another way to handle this, would be to not add those decorations to an ItemStack using a Capability, but to add them using the IClientItemExtension added by #8786. That however would be a breaking change as the implementations now need a field to store the Decorations (which would be fine, since no RB is out yet).

Just for my own knowledge: Why is rendering with capabilities on itemstacks such a big problem?

marchermans:
Because capabilities are not designed for it and they are not supposed to be Client only. But i can ask internally if we might want to add support for it.

Shadows-of-Fire:
PR's for events that fire on stack rendering have historically been automatically denied, so it may only be accepted as an extension to the client item interface

justliliandev:
The ItemDecorations have been moved to the IClientItemExtension. Notable changes:
Each Item has by default it's own IClientItemExtension created by `IClientItemExtension#DEFAULT_FACTORY` so that each one has it's own List Instance.
This is a breaking change, since IClientItemExtension got a new non default method

Shadows-of-Fire:
Just make the method default and return `Collections.emptyList()` to make it non-breaking.  Then you also don't need all these default impl classes.

Shadows-of-Fire:
Nevermind, I see the intention is to be able to add to other items, which would require an existing mutable list to be accessible.

Shadows-of-Fire:
You could, however, get around the binary break by adding two methods - one `getItemDecorators` that returns an immutable view, and one `addItemDecorator` that allows for adding a decorator.

Then `getItemDecorators` could default to `Collections.emptyList` and `addItemDecorator` could default to doing nothing until implementers fix their implementations.

justliliandev:
That is true, but would mean that mod authors that already implement this interface might not be aware of this change. I've also considered a throw in dev scenario, but that would only throw if someone tries to add a a decorator in dev which might not happen. Logging the default impl with warning might be an option, but also has the problem of only being fixed if someone reports it to the mod author, which is hard to do as there is no context to which item a decorator was tried to be added.

diesieben07:
The List should probably not live in `IClientItemExtensions`, but rather another client-only data structure for the items, so that mods using `IClientItemExtensions` do not have to care about it.

XFactHD:
diesieben sniped me a little here:
I don't think putting the "burden" for such a feature on everyone implementing `IClientItemExtension` is a good idea. Every implementation would either be exactly the same or completely broken because the modder thinks "I don't need this feature -> return unmodifiable empty list".

justliliandev:
They could extend `ClientItemExtensionImpl` and override the methods they want similar to how I did it with the `CustomArmorModelTest`. A seperate client only object stored in `Item` would increase the patch size and maintenance cost of such a feature as well as duplicate a bit of the code already used by `IClientItemExtension` and potentially hinders discoverability of this feature.
If another client-only data structure is wanted, I would be open for name suggestions.

XFactHD:
That would still make it a breaking change because you would need to enforce the use of `ClientItemExtensionImpl` by changing the generic type in the consumer in `Item#initializeClient()` and also technically still put the burden on the modder.

justliliandev:
It's a breaking change, but there was not Recommended Build yet and with this interface being kinda new I doubt many people will be effected by this break. The generic change in initializeClient is not necessary, and anyone can create their own implementation of `IClientItemExtension` and provide the list themselves.

XFactHD:
The `IClientItemExtension` interface is not new, it was just renamed from `IItemRenderProperties`

justliliandev:
I've missed that completly. A new data structure sounds more plausible now.
To not making it a breaking change it could also be defaulted with a `Map<IClientItemExtension, List<IItemDecorator>>` with a logged warning if this method is not overriden and with the faster direct field way for the `ClientItemExtensionImpl`

XFactHD:
I still think a completely separate data structure is cleaner than trying to tack this onto `IClientItemExtension`

justliliandev:
yeah, would probably be better.
Should the datastructure just be called ItemDecorations or ItemDecoratorHandler or something like that?
or maybe something like ClientItemData?

justliliandev:
ItemDecorators have been moved to their own data structure

patches/minecraft/net/minecraft/world/item/Item.java.patch
```diff
@@ -167,6 +168,9 @@
 +   public Object getRenderPropertiesInternal() {
 +      return renderProperties;
 +   }
++   public Object getItemDecoratorHandler() {
```
diesieben07: Now that we have the JetBrains annotations, this should probably be annotated with `@ApiStatus.Internal`.

marchermans (changes requested):
So this is already a good start, but it can be simplified a lot further down.
Instead of attaching it to an item (so making a field for the handler on the item itself) make a central manager which holds a map from Item delegate to decorator handler. Then to collect the decorators fire an event and freeze the map afterwards (prevents issues with multi-threading and gives everybody a nice moment to register them all at once).

That said i think the functionaly of this idea is good, but I would like performance numbers, so please collect some statistics for the following scenarios:

JEI current situation: how long does it take on average to render a single frame in the current situation (open the creative menu + jei for measurements)
JEI new situation, no decorators: again average frametime please, again open creative menu + jei for measurements
JEI new situation, 1 decorator: same as above
JEI new situation, 50 decorators, same as above

Good luck!
Note this won't be merged without actual indicators as to how it will effect the performance, both in common mondane situations as well as in the extremes

src/main/java/net/minecraftforge/client/IItemDecorator.java
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+package net.minecraftforge.client;
+
+import net.minecraft.client.gui.Font;
+import net.minecraft.world.item.ItemStack;
+
+public interface IItemDecorator {
+
+    void render(Font font, ItemStack stack, int xOffset, int yOffset, float blitOffset);
```
diesieben07: Please add Javadocs to this interface and its method.
When is this called? What is the expected render state? Are there considerations regarding modifying render state in this method (e.g. enabling/disabling of textures, etc.).

src/main/java/net/minecraftforge/client/ItemDecoratorHandler.java
```diff
@@ -0,0 +1,50 @@
...
+
+import net.minecraft.client.gui.Font;
+import net.minecraft.world.item.Item;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.registries.ForgeRegistries;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ItemDecoratorHandler
```
diesieben07: This should be a final class.

src/main/java/net/minecraftforge/client/ItemDecoratorHandler.java
```diff
@@ -0,0 +1,50 @@
...
+        {
+            return e;
+        }
+        throw new IllegalStateException("No ItemDecoratorHandler was added to: " + ForgeRegistries.ITEMS.getKey(item).toString());
+    }
+
+    /**
+     *
+     * Add a decorator during {@link net.minecraftforge.fml.event.lifecycle.FMLClientSetupEvent} with {@link net.minecraftforge.fml.event.lifecycle.FMLClientSetupEvent#enqueueWork(Runnable)}.
+     */
```
diesieben07: Forge is moving towards dedicated registration events. This should be an dedicated event and the list of decorators should be immutable. Refer to for example the new `RegisterTextureAtlasSpriteLoadersEvent`.

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -196,7 +196,7 @@
                     ).fieldOf("spawners").forGetter(AddSpawnsBiomeModifier::spawners)
             ).apply(builder, AddSpawnsBiomeModifier::new))
         );
-    
+
```
diesieben07: Unnecessary white space changes.

src/test/java/net/minecraftforge/debug/client/rendering/CustomItemDecorationsTest.java
```diff
@@ -0,0 +1,80 @@
...
+@Mod(CustomItemDecorationsTest.MOD_ID)
+public class CustomItemDecorationsTest
+{
+    public static final String MOD_ID = "custom_item_decorations_test";
+    private static final boolean ENABLED = true;
+
+    public CustomItemDecorationsTest() { }
+
+    @Mod.EventBusSubscriber(modid = CustomItemDecorationsTest.MOD_ID, bus = Mod.EventBusSubscriber.Bus.MOD, value = Dist.CLIENT)
+    public static class ClientEvents {
```
diesieben07: Formatting (braces on new lines)

src/test/java/net/minecraftforge/debug/client/rendering/CustomItemDecorationsTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+        private static final IItemDecorator decorator = new StackSizeDurabilityBar();
```
diesieben07: Why is this a static final field?

diesieben07 (changes requested):

patches/minecraft/net/minecraft/world/item/Item.java.patch
marchermans: If he applies my suggestion it will be nuked and moved anyway.
i would like this in an external manager not in the item.

src/main/java/net/minecraftforge/client/ItemDecoratorHandler.java
marchermans: Yes hence why I want this in a dedicated manager not in the item as well.
Freeze the map after the event fired.

diesieben07:
> Instead of attaching it to an item (so making a field for the handler on the item itself) make a central manager which holds a map from Item delegate to decorator handler.

This contradicts the rest of your message focusing on performance. The "fastest" would most likely be a field holding a plain array on the `Item` class. Otherwise you need a map lookup and an iterator.

marchermans:
> > Instead of attaching it to an item (so making a field for the handler on the item itself) make a central manager which holds a map from Item delegate to decorator handler.
> 
> This contradicts the rest of your message focusing on performance. The "fastest" would most likely be a field holding a plain array on the `Item` class. Otherwise you need a map lookup and an iterator.

I am not interested in the "fastest" solution.
I am interested in the numbers so I can get an understanding of the performance implications.
A central manager is easier to maintain, easier to initialize and freeze, so it is the superior option even if we leave some performance on the table due to the map lookup.

justliliandev:
# Performance tests:
## How I acquired the data
- All measures where taken over 10000 frames
- The `IItemDecoration` used for this is the `StackSizeDurabilityBar` from the TestMod or an implementation that does nothing for the "without rendering happening" column
- Map implementation used is the `Object2ObjectFastOpenHashMap` from fastutil
- Jei measurements where taken with default maxsize of 9x16 items
- Jei measurements where taken over `GuiEventHandler#onDrawBackgroundPost` from this https://github.com/mezz/JustEnoughItems/pull/2886 PR by pupnewster
- Creative screen measurements where taken with all 5 CreativeRows filled and the hotbar filled (so 54 itemstacks being catched by the measurement)
- Creative screen measurements where taken over the `AbstractContainerScreen#render` call without JEI
- Time measurements are obtained using `System#nanoTime` and averaged before rounding to ms with three decimal places
- All the data was obtained with a i5-7500 and Geforce GTX 1060 with a few (but the same programs running)

## Time Measurements
### JEI
| decoration system | no decorations | 1 decoration | 50 decorations | 50 decorations without rendering happening |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| no DecorationSystem | 3.007  | ~  | ~  | ~ |
| field in item | 3.157 | 5.688 | 102.672 | 3.331 |
| map | 3.208 | 5.701 | 103.608 | 3.382 |

### Creative Menu
| decoration system | no decorations | 1 decoration | 50 decorations | 50 decorations without rendering happening |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| no DecorationSystem | 1.819  | ~  | ~  | ~ |
| field in item | 1.840 | 2.791 | 47.148 | 1.860 |
| map | 1.843 | 2.845 | 46.866 | 1.886 |

## My Conclusion
Performance is not a big problem, if not too many decorators are used. Most of the code execution time is coming from rendering and not the overhead of this feature, hence modders should be careful what they do with this feature, but providing this feature is not too big of a problem.
Using a Map is surprisingly fast, even though the direct field is usually faster. Patchwise I would say that the added field in Item is not too big of a problem, as a similar patch in the exact same place is already present and the performance benefits might when mods add a lot of items.

# Further actions
I'm going to fix the things mentioned by diesieben and add javadocs where necessary.

With the data provided, I await a decision regarding Map or field in Item

src/main/java/net/minecraftforge/client/ItemDecoratorHandler.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@ApiStatus.Internal
+public final class ItemDecoratorHandler
+{
+    private final List<IItemDecorator> itemDecorators;
+
+    private static Map<Item, ItemDecoratorHandler> DECORATOR_LOOKUP = new Object2ObjectOpenHashMap<>();
```
diesieben07: The default should be `ImmutableMap.of`

src/main/java/net/minecraftforge/client/ItemDecoratorHandler.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public static void init()
+    {
+        var decorators = new HashMap<Item, List<IItemDecorator>>();
+        var event = new RegisterItemDecorationsEvent(decorators);
+        ModLoader.get().postEventWithWrapInModOrder(event, (mc, e) -> ModLoadingContext.get().setActiveContainer(mc), (mc, e) -> ModLoadingContext.get().setActiveContainer(null));
+        var decoratorLookUpMap = new HashMap<Item, ItemDecoratorHandler>();
+        decorators.forEach((item, itemDecorators) -> decoratorLookUpMap.put(item, new ItemDecoratorHandler(itemDecorators)));
+        DECORATOR_LOOKUP = ImmutableMap.copyOf(decoratorLookUpMap);
```
diesieben07: You can use `ImmutableMap.Builder` here

src/main/java/net/minecraftforge/client/ItemDecoratorHandler.java
```diff
@@ -0,0 +1,72 @@
...
+    {
+        resetRenderState();
+        for (IItemDecorator itemDecorator : itemDecorators)
+        {
+            if (itemDecorator.render(font, stack, xOffset, yOffset, blitOffset))
+                resetRenderState();
+        }
+    }
+
+    private void resetRenderState() {
```
diesieben07: Formatting (braces on new lines)

src/test/java/net/minecraftforge/debug/client/rendering/CustomItemDecorationsTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public CustomItemDecorationsTest() { }
+
+    @Mod.EventBusSubscriber(modid = CustomItemDecorationsTest.MOD_ID, bus = Mod.EventBusSubscriber.Bus.MOD, value = Dist.CLIENT)
+    public static class ClientEvents
+    {
+        @SubscribeEvent
+        public static void onRegisterItemDecorations(final RegisterItemDecorationsEvent event)
+        {
+            if (ENABLED) {
```
diesieben07: Formatting (braces on new lines)

src/main/java/net/minecraftforge/client/ItemDecoratorHandler.java
```diff
@@ -0,0 +1,72 @@
...
+public final class ItemDecoratorHandler
+{
+    private final List<IItemDecorator> itemDecorators;
+
+    private static Map<Item, ItemDecoratorHandler> DECORATOR_LOOKUP = new Object2ObjectOpenHashMap<>();
+
+    private static final ItemDecoratorHandler EMPTY = new ItemDecoratorHandler();
+    private ItemDecoratorHandler()
+    {
+        this.itemDecorators = List.of();
```
diesieben07: Use `ImmutableList.of()` here, this way it's an `ImmutableList` in both cases and the JVM does not need to deal with two different iterator implementations in `render`.

diesieben07 (changes requested):

src/main/java/net/minecraftforge/client/IItemDecorator.java
```diff
@@ -0,0 +1,28 @@
...
+import net.minecraft.client.renderer.entity.ItemRenderer;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.level.ItemLike;
+import net.minecraftforge.client.event.RegisterItemDecorationsEvent;
+
+/**
+ * An ItemDecorator that is used to render something on specific items, when the DurabilityBar and StackCount is rendered.
+ * Add it to an item using {@linkplain RegisterItemDecorationsEvent#register(ItemLike, IItemDecorator)}.
+ */
+public interface IItemDecorator {
```
diesieben07: Formatting (braces on new lines).
Sorry, I missed this the first time.

diesieben07 (changes requested):

diesieben07 (approved):

PR #8798, opened 2022-06-26
Create PieceBeardifierModifier to re-enable piecewise beardifier definitions

Drullkus:
As of Minecraft 1.18.2, `StructurePiece`s had a method `getNoiseEffect()` that returned a `NoiseEffect`. This method was referenced by the `Beardifier` class inside `compute()` and enabled Modded Structure Piece implementations to control their bounding box's influence on the noise-generated terrain.

With the introduction of Minecraft 1.19's `Beardifier.Rigid`, it introduced a regression where `StructurePiece` implementations were no longer able to be referenced for the type of contribution to the terrain. This functionality has been relegated to the `Structure` implementation.

This pull request restores this functionality to the `StructurePiece` if they opt in by implementing `PieceBeardifierModifier`, and uses its position in the code to also cede more control to each individual `PieceBeardifierModifier` implementer over the noise beardification.


marchermans (approved):
Sad that the documentation for the new interface is a bit lack luster, but it is fine as is. It describes the use case.

PR #8801, opened 2022-06-27
Bump Log4j to address 2 exploits

JT122406:
https://mvnrepository.com/artifact/org.apache.logging.log4j/log4j-core


marchermans:
After internal review of the vulnerabilities mentioned, we come to the conclusion that these do not effect Forge nor will they effect Minecraft in general. 

We tend to follow mojangs library releases closely and as long as they don't update, we see no reason to differentiate from their official versions.

PR #8803, opened 2022-06-28
Add API for tab list header/footer

zml2008:
image

With the existing state of affairs, any single mod could update both tab list header and footer at the same time, just by sending the existing vanilla packet. However, if one mod wants to set a header, and another mod wants to set a footer, they would need some shared way of tracking that state.

This PR adds such a way of tracking state, with wrapper API on ServerPlayer to hold the last sent header and footer.

I don't do anything to step anyone from constructing a tab list packet directly -- is that worth doing, by deprecating the constructor or something?

## testing steps

I did up a quick command to test this API:

```java
    @SubscribeEvent
    public void registerCommands(final RegisterCommandsEvent event) {
      event.getDispatcher().register(literal("tablist")
        .then(literal("header").then(argument("value", ComponentArgument.textComponent()).executes(ctx -> {
          ctx.getSource().getPlayerOrException().setTabListHeader(ComponentArgument.getComponent(ctx, "value"));
          return Command.SINGLE_SUCCESS;
        })))
        .then(literal("footer").then(argument("value", ComponentArgument.textComponent()).executes(ctx -> {
          ctx.getSource().getPlayerOrException().setTabListFooter(ComponentArgument.getComponent(ctx, "value"));
          return Command.SINGLE_SUCCESS;
        })))
      );
    }
```

This was tested with dev client, connected to dev dedicated server (since the tab list doesn't show up in SP)

Then, I confirmed that:

1. header was maintained while setting footer
2. footer was maintained while setting header
3. After executing `/kill`, that executing either command wouldn't clear the other's value on the tab list

### DIFF
--- a/patches/minecraft/net/minecraft/server/level/ServerPlayer.java.patch
+++ b/patches/minecraft/net/minecraft/server/level/ServerPlayer.java.patch
@@ -115,7 +115,7 @@
        this.f_36096_ = this.f_36095_;
     }
  
-@@ -1133,6 +_,13 @@
+@@ -1133,6 +_,15 @@
        this.m_36362_(p_9016_.m_36331_());
        this.m_36364_(p_9016_.m_36332_());
        this.m_219749_(p_9016_.m_219759_());
@@ -126,6 +126,8 @@
 +      if (old.m_128441_(PERSISTED_NBT_TAG))
 +          getPersistentData().m_128365_(PERSISTED_NBT_TAG, old.m_128423_(PERSISTED_NBT_TAG));
 +      net.minecraftforge.event.ForgeEventFactory.onPlayerClone(this, p_9016_, !p_9017_);
++      this.tabListHeader = p_9016_.tabListHeader;
++      this.tabListFooter = p_9016_.tabListFooter;
     }
  
     protected void m_142540_(MobEffectInstance p_143393_, @Nullable Entity p_143394_) {
@@ -210,7 +212,7 @@
           this.f_19853_.m_7967_(itementity);
           ItemStack itemstack = itementity.m_32055_();
           if (p_9087_) {
-@@ -1485,6 +_,29 @@
+@@ -1485,6 +_,76 @@
        }
     }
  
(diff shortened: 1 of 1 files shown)


marchermans (approved):
Reading into this, is it not better to make this an event instead of a single field where last comes, last wins rules the game?
This way the event could potentially make this automatically multi-line with an explicit clear option that people can use?

marchermans (changes requested):
Sorry previous comment was supposed to be a request changes.
Not an approve.

zml2008:
I'm not clear what exactly an event would be triggered by? There's nothing in the game that changes the header and footer here -- it's only there for use by mods. The fields seem like the least confusing variant of this API for me.

TheCurle:
Yeah, events are reactionary, and there's nothing to react to here.
This implementation looks perfectly fine to me.

marchermans:
> Yeah, events are reactionary, and there's nothing to react to here. This implementation looks perfectly fine to me.

The problem that I see with this, and which is generally a problem which we really try to avoid, is that this is a last comes last serves kind of a deal, where only a single mod can show something and all others be damned?

I would rather see an event here which combined the inputs of all mods together.

sciwhiz12:
How would such an event be fired? Whenever a mod attempts to modify the tab list header/footer, fire an event and let all other mods (including the original mod itself, potentially) take a stab at the value?

TheCurle:
COMBINING all the messages is a terrible terrible idea.
It *should* be all or nothing. Otherwise you have long, mismatched, disparate messages being chained one after the other in a visual soup of distracting nothingness.

marchermans:
Then this should be configurable by the end user.

I am aware that an event is far from ideal. But also a system where the last mod wins. We have several examples how we deal with these kinds of scenarios, the permission handler is an example.

So I would like to see this configurable by the end user / server owner at least to prevent weirdness from occuring.

LexManos:
This isn't bad, I just need some justification for it that isn't skeezy microtransactions, which is literally the only thing i've ever seen this feature used for.

zml2008:
These fields can be used for server branding (used especially effectively with custom fonts in Vanilla resource packs), and short status information, such as:

- current time
- active world information
- extra information about player count (current/max, # of users truncated from the player list)
- information about current server events

A relatively tasteful example from a server that seems popular (unfortunately overlayed on some spam in the boss bar position):

image

PR #8805, opened 2022-06-28
[1.19.X] Allow LazyOptional listeners to be weakly registered

amadornes:
In the current LazyOptional implementation, if BlockEntity A registers a listener on BlockEntity B's LazyOptional, even if BlockEntity A is removed from the world, it won't be evicted from memory, as it's still being referenced by the listener callback held by BlockEntity B. Moreover, even if BlockEntity B's LazyOptional is invalidated, the listeners are still retained in memory and won't be considered for GC until the LazyOptional itself has no references.

This PR addresses both of those issues by adding a new method to `LazyOptional` which allows listeners to be registered via a `WeakReference`, as well as clearing the list of listeners after they have been notified of the LazyOptional's invalidation.

An example usage would look like: `myLazyOptional.addListener(myObject, MyObjectType::callback)` as opposed to the current `myLazyOptional.addListener(myObject::callback)`.

### DIFF
--- a/src/main/java/net/minecraftforge/common/util/LazyOptional.java
+++ b/src/main/java/net/minecraftforge/common/util/LazyOptional.java
@@ -5,6 +5,7 @@
 
 package net.minecraftforge.common.util;
 
+import java.lang.ref.WeakReference;
 import java.util.HashSet;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


covers1624:
This is still quite fragile as if anyone pulls reference to `myObject` into the lambda they supply (probably `this` in most cases), then the weak reference is serving no purpose, as the original method.
```
.addListener(this, (e, l) -> this.clearSomeRandomCache(l))
```

This should be clarified in the docs that you MUST use the instance variable passed back into the lambda instead of grabbing your own. The current docs example with the use of a MethodRef does not cover the semantics of this exact common issue.

```
.addListener(this, (e, l) -> e.clearSomeRandomCache(l))
```

Shadows-of-Fire:
as discussed on discord, the old method may want to be either deprecated or removed, as otherwise it is unlikely that mod code is updated to use this method

marchermans (changes requested):
This needs better documentation.
I agree with @covers1624 That the implementation is weak, pun intended. It is easy to screw up and only provides safe guards.

Is there no other way to clean up the listener lists here? Like clearing them entirely out when getting invalidated?

PR #8814, opened 2022-07-02
Fixed edge case with wrongly thrown getOrCreateDirectory exception.

MarkusBordihn:
There is an edge case, where the first check could be passed and a exception happen later within the createDirectory call under Windows 11.

In my case several mods share the same config directory and so it could be that `java.nio.file.FileAlreadyExistsException` is thrown wrongly for an existing directory.
This additional check makes sure that it's only throws a exception if the error is not related to an existing directory, which was created in the mean time, between the former check.

Example exception with the current logic:
https://gist.github.com/MarkusBordihn/d6ae16139733c7c088c2e8aea02e2f5e

### DIFF
--- a/fmlloader/src/main/java/net/minecraftforge/fml/loading/FileUtils.java
+++ b/fmlloader/src/main/java/net/minecraftforge/fml/loading/FileUtils.java
@@ -29,7 +29,10 @@ public class FileUtils
             try {
                 Files.createDirectory(dirPath);
             } catch (IOException e) {
-                if (e instanceof FileAlreadyExistsException) {
+                if (e instanceof FileAlreadyExistsException && Files.isDirectory(dirPath)) {
+                    LOGGER.warn(CORE, "Failed to create {} directory, found existing directory : {}", dirLabel, dirPath);
+                    return dirPath;
+                } else if (e instanceof FileAlreadyExistsException) {
                     LOGGER.error(CORE, "Failed to create {} directory - there is a file in the way", dirLabel);
                 } else {
                     LOGGER.error(CORE, "Problem with creating {} directory (Permissions?)", dirLabel, e);


fmlloader/src/main/java/net/minecraftforge/fml/loading/FileUtils.java
```diff
@@ -29,7 +29,10 @@ public static Path getOrCreateDirectory(Path dirPath, String dirLabel) {
             try {
                 Files.createDirectory(dirPath);
             } catch (IOException e) {
-                if (e instanceof FileAlreadyExistsException) {
+                if (e instanceof FileAlreadyExistsException && Files.isDirectory(dirPath)) {
+                    LOGGER.warn(CORE, "Failed to create {} directory, found existing directory : {}", dirLabel, dirPath);
```
Shadows-of-Fire, line 33: Do we even need a warning message here?  The exception is erroneous (sorta), so it should probably be `debug` or lower.

MarkusBordihn: Warning would more helpful in the case of an user error report to see that it could be related to this edge case.
But happy to change this to debug if needed.

diesieben07: This method (`getOrCreateDirectory`) is a re-implementation of `Files.createDirectories`. Why is it even necessary?

MarkusBordihn: Not sure if I understand your question. The current method will produce a client crash in 1.19 in specific but reproducible cases, like:

image

In general it's best practice to re-use existing helper and utils provided by required classes like the `FileUtils.js` instead of duplicate their logic or function.

In my case the try and catch inside `getOrCreateDirectory` make sure that the client is not crashing for the most common issues, but it's currently not covering the edge case described in the summary.
I assume that the same applies for the mod loader itself, why this method was created, but I'm not a forge developer so can't comment on this.

If needed I could build my own logic and using `Files.createDirectories` directly, but not really seeing the benefit of duplicating the logic instead of using the public available method from `FileUtils.js` and fixing the broken logic.

diesieben07: What I mean is that getOrCreateDirectory shouldn't exist in the first place, Files.createDirectories should be used in its place.

MarkusBordihn: But why ? According a quick code search the method `getOrCreateDirectory` is used three times in this code-base, see: https://github.com/MinecraftForge/MinecraftForge/search?q=getOrCreateDirectory

If it would only be used once or so, I would totally agree. But repeating the same logic 2-3 times would be counter productive and is blowing up the code base.

Anyways, it's not my decision if the function should be deprecated and removed.

gigaherz: You seem to be misunderstanding what diesieben07 is trying to say.

What does `getOrCreateDirectory` do? It makes sure a directory exists, creating it and any of its parents that don't exist.

What does `Files.createDirectories` do? It makes sure a directory exists, creating it and any of its parents that don't exist.

So, what use is there for Forge to have a method, that already exists in java, and does the same? Wouldn't it be best to replace uses of that method with the java.nio.Files method, and remove it?

MarkusBordihn: Using `Files.createDirectories` directly would not fix the described edge case and you still need to use a try / catch to fix it, like `getOrCreateDirectory` with this PR would do.

If this is no concern for this code-base, then I agree getOrCreateDirectory could be deprecated and removed.

From the point of view of a mod developer, I would prefer that Forge is providing a utils method with additional checks so that less experience developers not running into issues and could avoid to build such checks them self. 
This was my motivation to submit this PR, instead of just fixing it inside of my own mods.

cpw: getOrCreateDirectory is required because Files.createDirectory throws an exception in a bunch of cases. What I don't understand is why you need to check if it's a directory when it exists, and is a directory. You don't explain why you're logging this message at all..

cpw: It's possible there's a problem where the directory exists but isn't writeable for some reason, I suppose. but that's MY speculation, NOT your explanation.

cpw:
I'm going to close this. This is an explicit error condition in the scenarios it can happen - specifically, the directory exists but isn't writeable for some reason. That is a VALID crash situation.. I don't know why you're experiencing a race here - I suspect you have some dodgy parallel code trying to somehow reuse a directory in an unexpected manner.

fmlloader/src/main/java/net/minecraftforge/fml/loading/FileUtils.java
MarkusBordihn: It's fine closing this PR, but let me explain the edge case more in detail. It's not an access or write issue, as you could see on the details in the first comment.

- In general `Files.createDirector` needs some time to create the directory and there is no locking or blocking for other processes for the same directory.
- If several mods has the same base config directory in their config file, the checks like ` if (!Files.isDirectory(dirPath))`  are passed for all of them.
- But after this check and when all of them try to create the directory over `Files.createDirector`, an exception is thrown, because the directory is already created by some of the other `Files.createDirector` calls before.
- So this additional check makes sure that even if several `Files.createDirector` calls are made for the same directory that no exception is thrown.

PR #8815, opened 2022-07-02
[1.19] Add RegisterPresetEditorEvent to allow adding PresetEditors 

MelanX:
This PR re-adds the ability to add custom screens to customize world presets. It works like in 1.16 to 1.18 with the ForgeWorldType.
This allows modders to add their own customize screens for their own world presets.


TheCurle:
Mojang have a system that works for this. I can't find specifics, but there's a map that contains screen factories that mods can assign to.

This is, as far as I understand it, completely unnecessary. That's why I removed it in the 1.19 update.

TheCurle:
PresetEditor.EDITORS it seems, is usable for this directly. Right above where you patch in an extra class and more state to manage.

Is there a reason, despite maintaining the old class, that this was added back?

MelanX:
I re-added this because `PresetEditor.EDITORS` is an immutable map and I thought it would be better to have an entry point which makes it easier to add own editors. I'm not quite sure how multiple modders would add their own editor without just replacing the vanilla map by using AccessTransformer. This would also override the maps created by other modders which results in only one editor by the modder which replaces the map at last.
If I'm wrong with this, I'm sorry and you can close this PR!

sciwhiz12:
Given that the type of the field is `Map` instead of an `ImmutableMap`, we could patch the field to be a modifiable map instead of maintaining a separate map in `WorldPresetScreens`. That would mean that mods would have to access the field in a thread-safe manner (`ParallelDispatchEvent#enqueueWork` in the mod lifecycle events). 

I believe we have done this (patching an immutable map to a modifiable one) already in the past, but an example evades my memory at the moment.

TheCurle:
Yeah. The map itself is stored and exposed as a Map, not an ImmutableMap. We should just patch where it's created and make it modifiable.

This goes against our usual thoughts about making more things mutable but in this case it's very clearly the sensible thing to do.

MelanX:
Alright, will change it when back on pc.

MelanX:
Based on a small talk with SizableShrimp on Discord, I added an event which modders can use to add their own screens properly.
Start message: https://discord.com/channels/313125603924639766/852298000042164244/993097186143383622

SizableShrimp (changes requested):
This would work a lot better if the original Map.of() was kept and a mutable map was stored internally in a forge class, with an immutable view stored and returned for this field. Similar to DatapackRegistriesHooks. And then fire the event later on during setup of Minecraft or something. This is just too much patching and not good formatting.

MelanX:
Thank you for the feedback. I'm too insecure how to do this and I feel like I can't do that with my current knowledge. I close this for now. If anyone else wants to add this, it would be cool. Otherwise, I may try it in the future again.

PR #8350, opened 2022-01-03
Load custom ITransformationServices from the classpath in dev

noeppi-noeppi:
Continuation of #8118 for minecraft 1.18.

This fixes #8095 by adding an `ITransformerDiscoveryServices` to locate transformers from the classpath when launching a development launch target. It works similar to `ClasspathLocator` that locates mod files from the classpath.

Updates modlauncher to `9.0.21` to make use of the changes made in https://github.com/McModLauncher/modlauncher/pull/84

PR #8818, opened 2022-07-03
Load custom ITransformationServices from the classpath in dev

### LINKED ISSUE #8095: Custom ITransformationServices don't work in development environment
**Minecraft Version:** 1.17.1

**Forge Version:** 37.0.53

**Description of issue:**

If I understand this correctly, there is [ModDirTransformerDiscoverer](https://github.com/MinecraftForge/MinecraftForge/blob/1.17.x/fmlloader/src/main/java/net/minecraftforge/fml/loading/ModDirTransformerDiscoverer.java) to find jars with custom transformer services in the mods folder and load them into the service layer, so their transform services can be loaded. However, in development, this does not work as there is no mods folder. So the transform service is just not found.

noeppi-noeppi:
1.19 version of #8350.

This fixes #8095 by adding an `ITransformerDiscoveryService` to locate transformers from the classpath when launching a development launch target. It works similar to `ClasspathLocator` that locates mod files from the classpath.


sciwhiz12 (commented):
Why was `ClasspathLocator#findJarPathFor` not removed here, while it was removed in your 1.18 PR?

PR #8819, opened 2022-07-04
[1.19] Fixed broken BufferBuilder.putBulkData(ByteBuffer) added by Forge

CreativeMD:
Hey there,

Forge adds `putBulkData(ByteBuffer buffer)` to `BufferBuilder`. This method is not used by Forge itself, so that is probably why the issue was never noticed. I recently started using it and ran into a few weird issues.

The problem is the position of the `byteBuffer` in `BufferBuilder`, which for some reason I do not know always needs to be `0`. `putBulkData` changes the position of the buffer to add the new data. Unfortunately the position is never reset to zero, so things go wrong after this method is called. I could easily fix the issue by resetting the position of the buffer after the method call myself. To save headaches for other users I created this small pull request, which should eliminate the issue.


patches/minecraft/com/mojang/blaze3d/vertex/BufferBuilder.java.patch
```diff
@@ -10,6 +10,7 @@
 +      m_85722_(buffer.limit() + this.f_85658_.m_86020_());
 +      this.f_85648_.position(this.f_85654_ * this.f_85658_.m_86020_());
```
covers1624: This can be changed to just use `nextElementByte` instead of this math.

CreativeMD: Sure, changed it

patches/minecraft/com/mojang/blaze3d/vertex/BufferBuilder.java.patch
```diff
@@ -1,14 +1,14 @@
...
+@@ -486,4 +_,17 @@
           this.f_166822_ = p_166829_;
        }
     }
 +
 +   // Forge start
 +   public void putBulkData(ByteBuffer buffer) {
 +      m_85722_(buffer.limit() + this.f_85658_.m_86020_());
-+      this.f_85648_.position(this.f_85654_ * this.f_85658_.m_86020_());
++      this.f_85648_.position(this.f_85652_);
```
OrionDevelopment, line 11: This should be the next element byte, so `f_85654_`

OrionDevelopment (changes requested):

patches/minecraft/com/mojang/blaze3d/vertex/BufferBuilder.java.patch
CreativeMD, line 11: `f_85652_` is `nextElementByte`. `f_85654_` is `vertices`.

marchermans (approved):

PR #8820, opened 2022-07-04
[1.19.x] Added RenderLevelStageEvent to replace RenderLevelLastEvent

SilverDavidMC:
This is a direct port of [this PR](https://github.com/MinecraftForge/MinecraftForge/pull/8603) to 1.19.

In short, `RenderLevelLastEvent` has its issues, namely with _Fabulous_ graphics, and this aims to resolve those issues by allowing for rendering during specific Stages in the new event. The original PR for 1.18 goes into more detail.

This does not remove `RenderLevelLastEvent`, but it does mark it for removal in a future update.


marchermans (approved):

cpw (dismissed):
This says it fires on the ModBus (which is definitely wrong) but it actually fires on the main FORGE event bus. Please fix..

cpw:
Nevermind. I was confused because some of the classes are EVENT_BUS but the register stuff is modbus. this is correct. Closing my review request.

PR #8825, opened 2022-07-05
[1.19] Add FogMode to ViewportEvent.RenderFog

### LINKED ISSUE #8762: [1.19] RenderFogEvent.getMode() returns a FogType when it should return a FogMode
**Minecraft Version:** 1.19
**Forge Version:** 41.0.38

**Description of issue:**
In 1.18, RenderFogEvent holds the current FogMode being used in setting up the fog, which can be retrieved through getMode(). In 1.19, it holds a FogType instead, which is returned by getMode(). Having the type is nice, but since there's no other means of retrieving the FogMode, it's difficult to determine whether the event is being fired for the sky fog or the terrain fog.

Jonathing:
Fixes #8762.

Most likely due to a mojang mapping rename, the `FogType` enum which contained an identifier to differ from terrain fog and sky fog was moved to `FogRenderer.FogMode`, which was removed from `ViewportEvent.RenderFog`. While the new `FogType` enum is no doubt helpful, it also removed the ability to identify if the fog that was being modified was either terrain fog or sky fog, and this PR aims to solve this.

Essentially, `getMode()` was renamed to `getType()`, and I added a new `getMode()` method which contains the instance of `FogMode`. I have also edited the constructor to accept a `FogMode` and have edited all instances of the construction of the event class (in which case it was just in `ForgeHooksClient`).

As always and with my last PR, feedback and edit requests are welcome.

### DIFF
--- a/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
+++ b/src/main/java/net/minecraftforge/client/ForgeHooksClient.java
@@ -438,7 +438,7 @@ public class ForgeHooksClient
         if (camera.getPosition().y < (double)((float)camera.getBlockPosition().getY() + state.getHeight(camera.getEntity().level, camera.getBlockPosition())))
             IClientFluidTypeExtensions.of(state).modifyFogRender(camera, mode, renderDistance, partialTick, nearDistance, farDistance, shape);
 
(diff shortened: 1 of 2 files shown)


marchermans (approved):

Jonathing:
PR updated to account for the new client rendering refactor that was recently merged into Forge 1.19.

PR #8827, opened 2022-07-06
Default IItemHandler capability for shulker box itemstacks

lukebemish:
Forge provides an `IItemHandler` capability for most built-in block entities and entities, and provides fluid capabilities for buckets, but does not provide a built-in item capabillity for shulker box items. This PR includes a simple implementation of such a capability, as discussed a bit on discord; concerns about performance were taken into account mostly only when attaching the capability, as it does not seem likely that many shulker box items will be interacted with in this way in a single tick. Otherwise, the `IItemHandler` is very similar to the wrappers given to vanilla container inventories.

### DIFF
--- a/src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
+++ b/src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
@@ -9,13 +9,12 @@ import net.minecraft.core.Direction;
 import net.minecraft.core.NonNullList;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.resources.ResourceLocation;
-import net.minecraft.tags.BlockTags;
 import net.minecraft.world.ContainerHelper;
 import net.minecraft.world.item.BlockItem;
 import net.minecraft.world.item.Item;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.Items;
-import net.minecraft.world.level.block.ShulkerBoxBlock;
+import net.minecraft.world.level.block.entity.BlockEntityType;
 import net.minecraftforge.common.capabilities.Capability;
 import net.minecraftforge.common.capabilities.ICapabilityProvider;
 import net.minecraftforge.common.util.LazyOptional;
@@ -198,7 +197,7 @@ public class ShulkerItemStackInvWrapper implements IItemHandler, ICapabilityProv
     public NonNullList<ItemStack> getItemList()
     {
         NonNullList<ItemStack> itemStacks = NonNullList.withSize(getSlots(), ItemStack.EMPTY);
-        CompoundTag rootTag = this.stack.getTag();
+        CompoundTag rootTag = BlockItem.getBlockEntityData(this.stack);
         if (rootTag != null && rootTag.contains("Items", 9)) {
             ContainerHelper.loadAllItems(rootTag, itemStacks);
         }
@@ -207,7 +206,9 @@ public class ShulkerItemStackInvWrapper implements IItemHandler, ICapabilityProv
 
     public void setItemList(NonNullList<ItemStack> itemStacks)
     {
-        ContainerHelper.saveAllItems(this.stack.getOrCreateTag(), itemStacks);
+        CompoundTag existing = BlockItem.getBlockEntityData(this.stack);
+        BlockItem.setBlockEntityData(this.stack, BlockEntityType.SHULKER_BOX,
+                ContainerHelper.saveAllItems(existing == null ? new CompoundTag() : existing, itemStacks));
     }
 
     @Override
--- a/src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
+++ b/src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
@@ -33,13 +33,40 @@ import java.util.Set;
 @ApiStatus.Internal
 public class ShulkerItemStackInvWrapper implements IItemHandlerModifiable, ICapabilityProvider
 {
+    @ApiStatus.Internal
+    @Nullable
+    public static ICapabilityProvider createDefaultProvider(ItemStack itemStack)
+    {
+        var item = itemStack.getItem();
+        if (item == Items.SHULKER_BOX ||
+            item == Items.BLACK_SHULKER_BOX ||
+            item == Items.BLUE_SHULKER_BOX ||
+            item == Items.BROWN_SHULKER_BOX ||
+            item == Items.CYAN_SHULKER_BOX ||
+            item == Items.GRAY_SHULKER_BOX ||
+            item == Items.GREEN_SHULKER_BOX ||
+            item == Items.LIGHT_BLUE_SHULKER_BOX ||
+            item == Items.LIGHT_GRAY_SHULKER_BOX ||
+            item == Items.LIME_SHULKER_BOX ||
+            item == Items.MAGENTA_SHULKER_BOX ||
+            item == Items.ORANGE_SHULKER_BOX ||
+            item == Items.PINK_SHULKER_BOX ||
+            item == Items.PURPLE_SHULKER_BOX ||
+            item == Items.RED_SHULKER_BOX ||
+            item == Items.WHITE_SHULKER_BOX ||
+            item == Items.YELLOW_SHULKER_BOX
+        )
+            return new ShulkerItemStackInvWrapper(itemStack);
+        return null;
+    }
+
     private final ItemStack stack;
     private final LazyOptional<IItemHandler> holder = LazyOptional.of(() -> this);
 
     private CompoundTag cachedTag;
     private NonNullList<ItemStack> itemStacksCache;
 
-    public ShulkerItemStackInvWrapper(ItemStack stack)
+    private ShulkerItemStackInvWrapper(ItemStack stack)
     {
         this.stack = stack;
     }
@@ -147,7 +174,7 @@ public class ShulkerItemStackInvWrapper implements IItemHandlerModifiable, ICapa
         }
     }
 
-    protected void validateSlotIndex(int slot)
+    private void validateSlotIndex(int slot)
     {
         if (slot < 0 || slot >= getSlots())
             throw new RuntimeException("Slot " + slot + " not in valid range - [0," + getSlots() + ")");
@@ -175,7 +202,7 @@ public class ShulkerItemStackInvWrapper implements IItemHandlerModifiable, ICapa
         setItemList(itemStacks);
     }
 
-    protected NonNullList<ItemStack> getItemList()
+    private NonNullList<ItemStack> getItemList()
     {
         CompoundTag rootTag = BlockItem.getBlockEntityData(this.stack);
         if (cachedTag == null || !cachedTag.equals(rootTag))
@@ -183,7 +210,7 @@ public class ShulkerItemStackInvWrapper implements IItemHandlerModifiable, ICapa
         return itemStacksCache;
     }
 
-    protected NonNullList<ItemStack> refreshItemList(CompoundTag rootTag)
+    private NonNullList<ItemStack> refreshItemList(CompoundTag rootTag)
     {
         NonNullList<ItemStack> itemStacks = NonNullList.withSize(getSlots(), ItemStack.EMPTY);
         if (rootTag != null && rootTag.contains("Items", CompoundTag.TAG_LIST))
@@ -194,7 +221,7 @@ public class ShulkerItemStackInvWrapper implements IItemHandlerModifiable, ICapa
         return itemStacks;
     }
 
-    protected void setItemList(NonNullList<ItemStack> itemStacks)
+    private void setItemList(NonNullList<ItemStack> itemStacks)
     {
         CompoundTag existing = BlockItem.getBlockEntityData(this.stack);
         CompoundTag rootTag = ContainerHelper.saveAllItems(existing == null ? new CompoundTag() : existing, itemStacks);
@@ -207,39 +234,5 @@ public class ShulkerItemStackInvWrapper implements IItemHandlerModifiable, ICapa
     public <T> LazyOptional<T> getCapability(@NotNull Capability<T> cap, @Nullable Direction side)
     {
         return CapabilityItemHandler.ITEM_HANDLER_CAPABILITY.orEmpty(cap, this.holder);
-    }
-
-    @ApiStatus.Internal
-    public static class AttachmentHandler
-    {
-        private static final Set<ResourceLocation> SHULKER_ITEMS = Set.of(
-                new ResourceLocation("shulker_box"),
-                new ResourceLocation("black_shulker_box"),
-                new ResourceLocation("blue_shulker_box"),
-                new ResourceLocation("brown_shulker_box"),
-                new ResourceLocation("cyan_shulker_box"),
-                new ResourceLocation("gray_shulker_box"),
-                new ResourceLocation("green_shulker_box"),
-                new ResourceLocation("light_blue_shulker_box"),
-                new ResourceLocation("light_gray_shulker_box"),
-                new ResourceLocation("lime_shulker_box"),
-                new ResourceLocation("magenta_shulker_box"),
-                new ResourceLocation("orange_shulker_box"),
-                new ResourceLocation("pink_shulker_box"),
-                new ResourceLocation("purple_shulker_box"),
-                new ResourceLocation("red_shulker_box"),
-                new ResourceLocation("white_shulker_box"),
-                new ResourceLocation("yellow_shulker_box"));
-
-        @ApiStatus.Internal
-        @Nullable
-        public static ICapabilityProvider listenCapabilitiesAttachment(ItemStack itemStack)
-        {
-            if (SHULKER_ITEMS.contains(ForgeRegistries.ITEMS.getKey(itemStack.getItem())))
-            {
-                return new ShulkerItemStackInvWrapper(itemStack);
-            }
-            return null;
-        }
     }
 }
--- a/src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
+++ b/src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
@@ -55,11 +55,8 @@ public class ShulkerItemStackInvWrapper implements IItemHandlerModifiable, ICapa
     @NotNull
     public ItemStack getStackInSlot(int slot)
     {
-        if (slot < getSlots())
-        {
-            return getItemList().get(slot);
-        }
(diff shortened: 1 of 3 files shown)


src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
```diff
@@ -0,0 +1,247 @@
...
+import net.minecraftforge.items.CapabilityItemHandler;
+import net.minecraftforge.items.IItemHandler;
+import net.minecraftforge.items.ItemHandlerHelper;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Set;
+
+@Mod.EventBusSubscriber(modid = "forge")
+public class ShulkerItemStackInvWrapper implements IItemHandler, ICapabilityProvider
```
diesieben07: Implementing `ICapabilityProvider` here makes this public API while it should be an implementation detail.
Additionally, `IItemHandlerModifiable` should be implemented as well.

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
```diff
@@ -0,0 +1,247 @@
...
+
+    @Override
+    @NotNull
+    public ItemStack getStackInSlot(int slot)
+    {
+        if (slot < getSlots())
+        {
+            return getItemList().get(slot);
+        }
+        return ItemStack.EMPTY;
```
diesieben07: This should throw an exception instead.

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    @Override
+    @NotNull
+    public ItemStack insertItem(int slot, @NotNull ItemStack stack, boolean simulate)
```
diesieben07, line 90: The logic here seems significantly more complex than the equivalent in `ItemStackHandler`. Why is that?

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
```diff
@@ -0,0 +1,247 @@
...
+        ItemStack stackInSlot = itemStacks.get(slot);
+
+        if (stackInSlot.isEmpty())
+            return ItemStack.EMPTY;
+
+        if (simulate)
+        {
+            if (stackInSlot.getCount() < amount)
+            {
+                return stackInSlot.copy();
```
diesieben07: This stack was just deserialized from NBT, there is no need to copy it.

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
```diff
@@ -0,0 +1,247 @@
...
+        }
+        return itemStacks;
+    }
+
+    public void setItemList(NonNullList<ItemStack> itemStacks)
+    {
+        CompoundTag existing = BlockItem.getBlockEntityData(this.stack);
+        BlockItem.setBlockEntityData(this.stack, BlockEntityType.SHULKER_BOX,
+                ContainerHelper.saveAllItems(existing == null ? new CompoundTag() : existing, itemStacks));
+    }
```
diesieben07: This should really use a cache and only deserialize when the stack's NBT tag has not changed.

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
```diff
@@ -0,0 +1,247 @@
...
+            Items.MAGENTA_SHULKER_BOX,
+            Items.ORANGE_SHULKER_BOX,
+            Items.PINK_SHULKER_BOX,
+            Items.PURPLE_SHULKER_BOX,
+            Items.RED_SHULKER_BOX,
+            Items.WHITE_SHULKER_BOX,
+            Items.YELLOW_SHULKER_BOX);
+
+    @SubscribeEvent
+    public static void listenCapabilitiesAttachment(AttachCapabilitiesEvent<ItemStack> event)
```
diesieben07: This should not be public API.

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    @Override
+    @NotNull
+    public <T> LazyOptional<T> getCapability(@NotNull Capability<T> cap, @Nullable Direction side)
+    {
+        return CapabilityItemHandler.ITEM_HANDLER_CAPABILITY.orEmpty(cap, this.holder);
+    }
+
+    private static final Set<Item> SHULKER_ITEMS = Set.of(Items.SHULKER_BOX,
```
diesieben07: This sounds like a candidate for a tag.

diesieben07 (changes requested):

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
diesieben07: Additionally, `IItemHandlerModifiable` should be implemented as well.

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
lukebemish: This was a discussion had on discord. While I agree that this could be a tag, we also discussed whether this should be an internal implementation; if it is an internal implementation, then it wouldn't make sense to make this a tag. (after all, other mods' shulker boxes might store items in nbt differently). I could see this going either way; just let me know what everyone thinks is best.

lukebemish: (copied from discord) My personal opinion is that it should not, and the list of targets should be immutable, but that I could separate out the IItenHandler implementation from the ICapabilityProvider so that modders could implement the prior for their own items. (honestly, I think this is a bit ill-advised. My implementation is meant to work for vanilla items. Modded items should use something more efficient with a serializable cap or whatever)

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
lukebemish: This logic was copied from InvWrapper; it may be a sign that InvWrapper needs a refactor. I'll look at using the logic from ItemStackHandler instead.

lukebemish: (Fixed by [4076409](https://github.com/MinecraftForge/MinecraftForge/pull/8827/commits/40764098c55531042d61406596c758b5a659664e))

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
lukebemish: (Fixed by [4076409](https://github.com/MinecraftForge/MinecraftForge/pull/8827/commits/40764098c55531042d61406596c758b5a659664e))

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
lukebemish: (Fixed by [8eb35a4](https://github.com/MinecraftForge/MinecraftForge/pull/8827/commits/8eb35a4984587249bfa5e8811c797466ba90b580), I think, though my caching setup could use a once-over)

lukebemish:
I went through and addressed most of your feedback; the one thing I am not sure how you'd like to have set up is the public API vs implementation details part. When I had written this, I'd intended it *all* to be basically implementation details; this wrapper is targeted at vanilla shulker boxes, and any modded items wouldn't be expected to use it. I can definitely separate out the ICapabilityProvider part and the listener somewhere else, so that that becomes an implementation detail and the capability something modders could use.

lukebemish:
I modeled the combined capability provider/item handler off of how Forge adds stuff to buckets at present, but I'm not sure what the best arrangement would be for that

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
alcatrazEscapee: Why should it be different than any of the other item stack capability wrappers that Forge providers? Considering that every item stack capability wrapper already follows this pattern, including [`FluidBucketWrapper`](https://github.com/MinecraftForge/MinecraftForge/blob/c2225ab1da58cf8eb3ac3aab1aab73f2d7a696fa/src/main/java/net/minecraftforge/fluids/capability/wrappers/FluidBucketWrapper.java#L32), [`FluidHandlerItemStack`](https://github.com/MinecraftForge/MinecraftForge/blob/c2225ab1da58cf8eb3ac3aab1aab73f2d7a696fa/src/main/java/net/minecraftforge/fluids/capability/templates/FluidHandlerItemStack.java#L30), and [`FluidHandlerItemStackSimple`](https://github.com/MinecraftForge/MinecraftForge/blob/c2225ab1da58cf8eb3ac3aab1aab73f2d7a696fa/src/main/java/net/minecraftforge/fluids/capability/templates/FluidHandlerItemStackSimple.java#L27).

TheCurle:
Looks good to me for now.
It's currently the weekend, so we'll have a closer look tomorrow, and hopefully pull.

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
```diff
@@ -0,0 +1,239 @@
...
+    private final ItemStack stack;
+    private final LazyOptional<IItemHandler> holder = LazyOptional.of(() -> this);
+
+    private CompoundTag cachedTag;
+    private NonNullList<ItemStack> itemStacksCache;
+
+    public ShulkerItemStackInvWrapper(ItemStack stack)
+    {
+        this.stack = stack;
+        this.itemStacksCache = refreshItemList(BlockItem.getBlockEntityData(this.stack));
```
diesieben07: Why load the data immediately here? It is not needed yet.

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
```diff
@@ -0,0 +1,239 @@
...
+import net.minecraftforge.items.CapabilityItemHandler;
+import net.minecraftforge.items.IItemHandler;
+import net.minecraftforge.items.IItemHandlerModifiable;
+import net.minecraftforge.items.ItemHandlerHelper;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Set;
+
+public class ShulkerItemStackInvWrapper implements IItemHandlerModifiable, ICapabilityProvider
```
diesieben07, line 34: This class should not be public.

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
```diff
@@ -0,0 +1,239 @@
...
+        CompoundTag rootTag = BlockItem.getBlockEntityData(this.stack);
+        if ((cachedTag != null && !cachedTag.equals(rootTag)) || rootTag == null)
+            itemStacksCache = refreshItemList(rootTag);
+        return itemStacksCache;
+    }
+
+    protected NonNullList<ItemStack> refreshItemList(CompoundTag rootTag)
+    {
+        NonNullList<ItemStack> itemStacks = NonNullList.withSize(getSlots(), ItemStack.EMPTY);
+        if (rootTag != null && rootTag.contains("Items", 9)) {
```
diesieben07: - Formatting, braces on new lines
- Use `Tag.TAG_LIST` instead of the magic number

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
```diff
@@ -0,0 +1,239 @@
...
+        if (!isItemValid(slot, stack)) throw new RuntimeException("Invalid stack " + stack + " for slot " + slot + ")");
+        NonNullList<ItemStack> itemStacks = getItemList();
+        itemStacks.set(slot, stack);
+        setItemList(itemStacks);
+    }
+
+    protected NonNullList<ItemStack> getItemList()
+    {
+        CompoundTag rootTag = BlockItem.getBlockEntityData(this.stack);
+        if ((cachedTag != null && !cachedTag.equals(rootTag)) || rootTag == null)
```
diesieben07: `Objects.equals`

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
```diff
@@ -0,0 +1,239 @@
...
+        cachedTag = rootTag;
+        return itemStacks;
+    }
+
+    protected void setItemList(NonNullList<ItemStack> itemStacks)
+    {
+        CompoundTag existing = BlockItem.getBlockEntityData(this.stack);
+        CompoundTag rootTag = ContainerHelper.saveAllItems(existing == null ? new CompoundTag() : existing, itemStacks);
+        BlockItem.setBlockEntityData(this.stack, BlockEntityType.SHULKER_BOX,
+                rootTag);
```
diesieben07: Why the line break?

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
```diff
@@ -0,0 +1,239 @@
...
+                Items.MAGENTA_SHULKER_BOX,
+                Items.ORANGE_SHULKER_BOX,
+                Items.PINK_SHULKER_BOX,
+                Items.PURPLE_SHULKER_BOX,
+                Items.RED_SHULKER_BOX,
+                Items.WHITE_SHULKER_BOX,
+                Items.YELLOW_SHULKER_BOX);
+
+        @SubscribeEvent
+        public static void listenCapabilitiesAttachment(AttachCapabilitiesEvent<ItemStack> event) {
```
diesieben07: This should use `initCapabilities` in the Item class, not the event.

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
```diff
@@ -0,0 +1,239 @@
...
+                Items.LIGHT_BLUE_SHULKER_BOX,
+                Items.LIGHT_GRAY_SHULKER_BOX,
+                Items.LIME_SHULKER_BOX,
+                Items.MAGENTA_SHULKER_BOX,
+                Items.ORANGE_SHULKER_BOX,
+                Items.PINK_SHULKER_BOX,
+                Items.PURPLE_SHULKER_BOX,
+                Items.RED_SHULKER_BOX,
+                Items.WHITE_SHULKER_BOX,
+                Items.YELLOW_SHULKER_BOX);
```
diesieben07: The formatting here is weird, the items should all be on their own line

diesieben07 (changes requested):

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
lukebemish: As discussed multiple times, shulker boxes use normal BlockItems.

lukebemish: The only way to make this use initCapabilities would be to make an otherwise-unused ShulkerBoxBlockItem class and patch it in for shulkers, which seemed like overkill to me. If this is the preferred solution, please let me know

LexManos: No, it's not. You *could* change  `public static void listenCapabilitiesAttachment(AttachCapabilitiesEvent<ItemStack> event) `  to `@ApiInternal @Nullable public static ICapabilityPrvider getVanillaShulkerInventory(Item) ` and call it from `IForgeItem.initCapabilities`'s default implementation.

Also, don't use a static set. Or make the set based on the delegate.

lukebemish:
Except the whole `initCapabilities` thing, this should be good to go; if you want me to use `initCapabilities`, that would require creating an otherwise-unused `ShulkerBoxBlockItem` class and patching shulker boxes to use it. I could, but I'd like some confirmation first.

src/main/java/net/minecraftforge/items/wrapper/ShulkerItemStackInvWrapper.java
lukebemish: As discussed on discord already, we *don't* want to use delegates here, because if someone registry-replaces the shulker boxes, we don't want to attach out capability.

LexManos: And I disagree, because if you implement it in initCaps, then they can opt out via NOT calling  super.

lukebemish: That doesn't solve the issue: imagine we attach this in `initCapabilities` checking equality with a delegate. Someone else replaces shulker boxes with their own normal `BlockItem`. We've now attached a capability to that BlockItem, even though they never asked us to.

LexManos: That is their problem. Opt out vs Opt In. If they break vanilla API by replacing the item, and then breaking how they store data, that is on them.

lukebemish: Alright, I should rephrase: if we do this with delegates, it becomes a breaking change. If you're alright with that, I'll go ahead and switch it to delegates then.

LexManos: By that definition this is a breaking change by adding it at all.
So no, you're wrong. 
And i've already told you what to do. So do it. I've already told this to my triage team about 5 times.

lukebemish:
Should be switched to `initCapabilities` now; I would note that this could be considered a breaking change now, though fairly minor. If a mod had previously registry-replaced shulker boxes with a version that stored less/more items, they would now need to use a custom BlockItem class that overrides the default behavior.

LexManos (commented):
Change costly thing to after cheap thing

PR #8822, opened 2022-07-05
Update SJH to 2.1 and JIJ to .2.26 (subsumes #8817)

cpw:
Closes #8817

PR #8829, opened 2022-07-07
Cache resource listing calls in resource packs

### LINKED ISSUE #8813: PathResourcePack#getResources() throws exception when other mods have improperly-named resources
**Minecraft Version:** 1.19

**Forge Version:** 41.0.63

**Logs:** [example stacktrace](https://gist.github.com/soir20/215052334db7d3aa83243843d0ef0328)

**Steps to Reproduce:**
 1. Include a mod with resources that are improperly named ("read me.txt").
 2. Use `PathResourcePack#getResources(PackType, String, String, Predicate)` or a method that calls it to retrieve resource names from another mod.
 3. Observe that a `ResourceLocationException` is thrown.

**Description of issue:**
I'm the author of a mod that uses `PackResources#getResources(PackType, String, String, Predicate)` to retrieve files with a particular extension.

The issue is that other mods include resources that are not valid `ResourceLocation`s, such as "read me.txt" or "Capitalized.psd." In such cases, Forge's `PathResourcePack#getResources(PackType, String, String, Predicate)` throws a `ResourceLocationException` even if the invalid resources would be filtered out by the provided predicate.

I'm reporting this here because Minecraft's `FilePackResources` and `FolderPackResources` use `ResourceLocation#tryBuild(String, String)` with a null check, filtering out these improperly-named resources and preventing an exception. It would be ideal if Forge's `PathResourcePack` could use `tryBuild()` as well to prevent the exception.

I have submitted issues to the mods with these resources that have been reported to me. However, someone will have to submit an issue for every offending mod that pops up, or every mod author that wants to use this method will have to catch the exception. This seems like a minor change on Forge's end that improves compatibility and is more consistent with existing Minecraft implementations.

I'd be fine with submitting a PR for this if the Forge team wants to go forward with it.

marchermans:
### Core feature: Caching of resource listings
Currently, most of the time during world loads is spent collecting resources from resource packs like registry data, recipes, etc. This is done through calls like: `getResources`, 'listResources' and `listResourceStacks` in the end this all funnels through to a call on 'getResources' in the PackResources interface and its implementations.

Of note here is that the two major implementations that are currently in use when running Forge are `VanillaPackResources` and `PathPackResources`. (For vanilla and mods respectively).

This PR will expand their capabilities and introduce caching on their instances.
The caches will list the contained resources in the pack exactly once, and then hold on them for repeated access (which happens a lot when resources are reloaded) instead of walking the entire file tree every time a list of resources is requested.

### Stats:
Logs of a run without the cache enabled:
<https://gist.github.com/OrionDevelopment/f6d7c3fe452fe4e6e7f6aca47bce4acd>
Logs of a run with the cache enabled:
<https://gist.github.com/OrionDevelopment/2ae8f3894fc2f1ba20df58a3568c7df9>
If NPS profiling runs are needed I have those as well, but Github does not allow uploading those.

As can be seen from the logs (as well as the NPS if you request them) the speed up is roughly 300% (3x compared to the default).

Closes #8813
### Still todo:
- [x] Implement cache clearing on F3+T (Automatically handled since it creates new PackResource instances and reinitializes them)
- [x] Clean up the patches
- [x] Clean up the new cache manager.
- [x] Make the caching behavior configurable.


patches/minecraft/net/minecraft/server/packs/VanillaPackResources.java.patch
```diff
@@ -1,5 +1,63 @@
...
+       if (!p_10358_.contains("/") && !p_10358_.contains("\\")) {
+          if (f_10312_ != null) {
+@@ -146,7 +_,9 @@
+ 
+       try {
+          Path path = f_182296_.get(p_215346_);
+-         if (path != null) {
++         if (shouldUseCache() && path != null && this.cacheManager.hasCached(p_215346_, p_215347_)) {
++            set.addAll(this.cacheManager.getResources(p_215346_, p_215347_, path.getFileSystem().getPath(p_215348_), p_215349_));
++         } else if (path != null) {
```
pupnewfster: If this line is just `} else` and the `it (path != null) {` is on the next line then that slightly simplifies the patch by not having it remove a line.

patches/minecraft/net/minecraft/server/packs/VanillaPackResources.java.patch
```diff
@@ -1,5 +1,63 @@
...
+          ImmutableMap.Builder<PackType, Path> builder = ImmutableMap.builder();
+@@ -92,6 +_,19 @@
+       this.f_10314_ = ImmutableSet.copyOf(p_143762_);
+    }
+ 
++   @Override
++   public void initForNamespace(final String nameSpace) {
++      if (!shouldUseCache()) return;
++
++      this.cacheManager.index(nameSpace);
```
pupnewfster: I feel like this would read better as `if (shouldUseCache())` rather than having a return statement when there is only one line as the other case

src/main/java/net/minecraftforge/resource/PathResourcePack.java
```diff
@@ -17,10 +19,7 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Locale;
-import java.util.Set;
+import java.util.*;
```
pupnewfster: No wildcard imports

src/main/java/net/minecraftforge/resource/PathResourcePack.java
```diff
@@ -50,13 +72,25 @@ public PathResourcePack(String packName, final Path source)
         this.packName = packName;
     }
 
+    @Override
+    public void initForNamespace(final String namespace) {
+        if (!shouldUseCache()) return;
+
+        this.cacheManager.index(namespace);
```
pupnewfster: Same thing as in the patch version I feel like this reads better as a non negated if

src/main/java/net/minecraftforge/resource/PathResourcePack.java
```diff
@@ -50,13 +72,25 @@ public PathResourcePack(String packName, final Path source)
...
+
     /**
      * Returns the source path containing the resource pack.
      * This is used for error display.
      *
      * @return the root path of the resources.
-     */
-    public Path getSource() {
+                */
+        public Path getSource() {
```
pupnewfster: Weird spacing changes

src/main/java/net/minecraftforge/resource/PathResourcePack.java
```diff
@@ -36,6 +35,29 @@ public class PathResourcePack extends AbstractPackResources
...
 
+    private final ResourceCacheManager cacheManager = new ResourceCacheManager(() -> true, PathResourcePack::shouldIndexOffThread, (packType, namespace) -> resolve(packType.getDirectory(), namespace).toAbsolutePath());
+
+
+    private static boolean shouldIndexOffThread() {
+        return !getConfigValue(ForgeConfig.COMMON.indexModPackCachesOnThread);
+    }
+
+    private static boolean shouldUseCache() {
+        return getConfigValue(net.minecraftforge.common.ForgeConfig.COMMON.cachePackAccess);
```
pupnewfster: This doesn't need to use the FQN

src/main/java/net/minecraftforge/resource/PathResourcePack.java
```diff
@@ -36,6 +35,29 @@ public class PathResourcePack extends AbstractPackResources
...
+    }
+
+    private static boolean getConfigValue(java.util.function.Supplier<Boolean> configValue) {
```
pupnewfster: Doesn't need to use the FQN

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -506,4 +506,5 @@ public net.minecraft.world.level.levelgen.feature.treedecorators.TreeDecoratorTy
 protected net.minecraft.world.level.portal.PortalForcer f_77648_ # level
 public net.minecraft.world.level.storage.LevelResource <init>(Ljava/lang/String;)V # constructor
 private-f net.minecraft.world.level.storage.loot.LootPool f_79028_ # rolls
-private-f net.minecraft.world.level.storage.loot.LootPool f_79029_ # bonusRolls
\ No newline at end of file
+private-f net.minecraft.world.level.storage.loot.LootPool f_79029_ # bonusRolls
+public net.minecraft.resources.ResourceLocation m_135841_(Ljava/lang/String;)Z # isValidPath
```
pupnewfster: Run `checkATs` as resources definitely should be somewhere before world.

It may also be worth exposing other helpers in ResourceLocation to devs given this one is being exposed by this change:
- `isValidNamespace`
- `validNamespaceChar`

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,344 @@
+package net.minecraftforge.resource;
```
pupnewfster, line 6: Missing license header

patches/minecraft/net/minecraft/server/packs/VanillaPackResources.java.patch
```diff
@@ -1,5 +1,63 @@
...
++
++   private static boolean getConfigValue(java.util.function.Supplier<Boolean> configValue) {
++      //Yes we catch the early loading error on purpose. This feature needs to be configurable but Mojang loads resources really early.
++      //So when they are loaded early we preserve the default behaviour which is to process it on the main thread.
++      try {
++         return configValue.get();
++      } catch (IllegalStateException e) {
++         return false;
++      }
++   }
```
pupnewfster: Rather than having this great big hunk here and in `PathResourcePack` I feel like it may be nicer to:
a. Add either a helper method or an overload constructor to `ResourceCacheManager` that explicitly takes `BooleanValue`s and then wraps it in the try catch for passing to the current one
b. Expose a helper shouldUseCache method on the manager so that if we do have more pack types at some point then that can just be called rather than having to duplicate more of this hunk again

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,344 @@
...
+            if (!entriesByPathPrefix.containsKey(pathEntry)) {
+                //No cache entry found.
+                return Collections.emptyList();
+            }
+
+            //Since we inject into the cache recursively we can now just grab the map entry that is the prefix and loop over its values.
+            return entriesByPathPrefix.get(pathEntry).stream()
+                    .map(ResourceCacheEntry::resourceLocation)
+                    .filter(filter)
+                    .collect(Collectors.toList());
```
pupnewfster, line 395: If `getResources` doesn't need to be modifiable this potentially should just be `.toList()`

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,344 @@
...
+            //Get the path to the root of the namespace in the current pack.
+            final Path rootPath = pathBuilder.apply(packType, namespace);
+
+            //Build a walkable stream, process it
+            try (final Stream<Path> paths = pathFinder.createWalkingStream(rootPath)) {
+                paths
+                    .parallel() //Run the stream in parallel
+                    .map(rootPath::relativize) //Relative to the given root.
+                    .filter(path -> ResourceLocation.isValidPath(Joiner.on('/').join(path))) //Only process valid paths
+                    .map(path -> new ResourceCacheEntry(packType, namespace, path, new ResourceLocation(namespace, Joiner.on('/').join(path)))) //Create a cache entry.
```
pupnewfster: Rather than having two things that both run joiner on it would probably be more efficient to have the above map also join on `/` in addition to relativizing

pupnewfster (commented):
Seems very nice testing with this and #8822 gives me the following timings for hitting play selected world to getting into world when testing in Mekanism's dev environment:
```
Base: 158 seconds
#8822: 138 seconds
This PR: 28 seconds
This PR combined with #8822: 22 seconds
This PR combined with #8822 (not on my C drive): 16 seconds
```

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
marchermans: The returned collection is actually modified in `DefaultClientPackResources` so it needs to stay like this.

patches/minecraft/net/minecraft/server/packs/VanillaPackResources.java.patch
marchermans: That is really ugly and a lot harder to understand.
I personally think that one extra line shows way better what is happening in this patch file and later in the decompiled source.

pupnewfster: Actually on second thought, opinion on instead of having the patched in if statement also check `path != null` to just put the patched in chunk inside the existing vanilla if statement? And then just have a simple `else` without brackets for whether vanilla's method should run?

src/main/java/net/minecraftforge/resource/PathResourcePack.java
```diff
@@ -50,13 +56,25 @@ public PathResourcePack(String packName, final Path source)
...
+    }
+
     /**
      * Returns the source path containing the resource pack.
      * This is used for error display.
      *
      * @return the root path of the resources.
      */
-    public Path getSource() {
+     public Path getSource() {
```
pupnewfster: Still has a tiny random space being added

src/main/java/net/minecraftforge/resource/PathResourcePack.java
```diff
@@ -36,6 +38,10 @@ public class PathResourcePack extends AbstractPackResources
     private final Path source;
     private final String packName;
 
+    private final ResourceCacheManager cacheManager = new ResourceCacheManager(true, ForgeConfig.COMMON.indexModPackCachesOnThread, (packType, namespace) -> resolve(packType.getDirectory(), namespace).toAbsolutePath());
+
+    private record ResourceCacheEntry(PackType packType, String namespace, Path path, ResourceLocation resourceLocation) {}
```
pupnewfster: This record seems to be unused.

src/main/java/net/minecraftforge/resource/PathResourcePack.java
```diff
@@ -108,6 +126,10 @@ public Collection<ResourceLocation> getResources(PackType type, String resourceN
             Path root = resolve(type.getDirectory(), resourceNamespace).toAbsolutePath();
             Path inputPath = root.getFileSystem().getPath(pathIn);
 
+            if (ResourceCacheManager.shouldUseCache() && this.cacheManager.hasCached(type, resourceNamespace)) {
```
pupnewfster: All call sites of `hasCached` also checks the `shouldUseCache` first, would it be worthwhile to just inline the `shouldUseCache` as the first line in `hasCached`? Or do you think that would cause confusion.

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,375 @@
...
+     * The individual sub managers by pack type and namespace
+     */
+    private final Map<PackTypeAndNamespace, NamespacedResourceCacheManager> managersByNamespace = new ConcurrentHashMap<>();
+
+    /**
+     * Creates a new instance of a resource cache manager.
+     *
+     * @param supportsReloading True to make the namespace specific managers support reloading the cache.
+     * @param indexOffThreadConfig    True to index the file-tree off-thread.
+     * @param pathBuilder       The path builder to use.
```
pupnewfster: Spacing, either should line up or be one space

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,375 @@
...
+import com.google.common.collect.Maps;
+import com.mojang.logging.LogUtils;
+import net.minecraft.Util;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.server.packs.PackType;
+import net.minecraftforge.common.ForgeConfig;
+import net.minecraftforge.common.ForgeConfigSpec;
+import org.slf4j.Logger;
+
+import javax.annotation.Nullable;
```
pupnewfster: Should be using jetbrains nullable annotation

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,375 @@
...
+                //We have null, generally this only happens on the root directory.
+                pathEntry = "";
+            } else {
+                //Use the parent paths, we can not depend on a toString call here, since we need the / as separator.
+                pathEntry = Joiner.on("/").join(parentPath);
+            }
+
+            //Inject into the cache.
+            this.entriesByPathPrefix.putIfAbsent(pathEntry, new CopyOnWriteArrayList<>());
+            this.entriesByPathPrefix.get(pathEntry).add(entry);
```
pupnewfster: Any reason to not just use `this.entriesByPathPrefix.computeIfAbsent(pathEntry, e -> new CopyOnWriteArrayList<>()).add(entry);` This would both avoid initializing the array if there is already a path entry, and avoid an extra lookup call.

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,375 @@
...
+         * @param inputPath The input path prefix to look in.
+         * @param filter The filter which the resources must match.
+         * @return A collection of resource location which match the given path prefix and who match the filter.
+         */
+        public Collection<ResourceLocation> getResources(final Path inputPath, final Predicate<ResourceLocation> filter) {
+            //Use the input path prefix, we can not depend on a toString call here, since we need the / as separator.
+            final String pathEntry = Joiner.on('/').join(inputPath);
+            if (!entriesByPathPrefix.containsKey(pathEntry)) {
+                //No cache entry found.
+                return Collections.emptyList();
```
pupnewfster: Because of the fact that you said the return result of `getResources` is mutable I feel like this shouldn't just be returning an empty list. The easiest solution honestly would be to change the below stream to start as:
```java
return entriesByPathPrefix.getOrDefault(pathEntry, Collections.emptyList()).stream()
```

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,375 @@
...
+         * @return True if the cache has been loaded, false otherwise.
+         */
+        public boolean cacheLoaded() {
+            return cacheLoaded.get();
+        }
+
+        /**
+         * Stream element resource that combines a normalized path and a joined path using the "/" as separator.
+         */
+        private record PathWithLocationPath(Path path, String locationPath) { }
```
pupnewfster: This record could just be a local record rather than a private one as it is only used in a single method as a temporary helper to hold the data while mapping

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,375 @@
...
+         * Can be invoked on- or off-thread.
+         */
+        private void doIndex() {
+            //Get the path to the root of the namespace in the current pack.
+            final Path rootPath = pathBuilder.apply(packType, namespace);
+
+            //Build a walkable stream, process it
+            try (final Stream<Path> paths = pathFinder.createWalkingStream(rootPath)) {
+                paths
+                    .parallel() //Run the stream in parallel
```
pupnewfster: I feel like the initial call which just makes it parallel can be on the same line as `paths`

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,375 @@
...
+    private final ForgeConfigSpec.BooleanValue indexOffThreadConfigOption;
+
+    /**
+     * The path builder (different users have different requirements for how they want to handle this)
+     */
+    private final BiFunction<PackType, String, Path> pathBuilder;
+    /**
+     * The individual sub managers by pack type and namespace
+     */
+    private final Map<PackTypeAndNamespace, NamespacedResourceCacheManager> managersByNamespace = new ConcurrentHashMap<>();
```
sciwhiz12: This could be replaced with a Guava [`Table`](https://github.com/google/guava/wiki/NewCollectionTypesExplained#table) instead, allowing the removal of `PackTypeAndNamespace`. 

However, it would mean that, [`Tables.synchronizedTable`](https://guava.dev/releases/31.1-jre/api/docs/com/google/common/collect/Tables.html#synchronizedTable(com.google.common.collect.Table)) would need to be used to preserve thread-safety, which comes with the caveat that any accesses to its collection views must be manually synchronized on the table instance (as noted in the javadoc of the given method). 

Looking at the current implementation, `#getNamespaces` would be the only part that would need manual synchronization in this case, as it would be implemented as `managersByNamespace.column(packType).keySet()` (plus unmodifiable set or `Set.copyOf`).

Also, do we need the thread-safety and atomicity guarantees, since the map is only mutated in the `index` method and read from everywhere else?

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,375 @@
...
+        private final BiFunction<PackType, String, Path> pathBuilder;
+        /**
+         * The path walker stream factory to use.
+         */
+        private final PathWalkerFactory pathFinder;
+
+        /**
+         * The cache entries for this manager.
+         */
+        private final Map<String, List<ResourceCacheEntry>> entriesByPathPrefix = Maps.newConcurrentMap();
```
sciwhiz12, line 251: Consistency: this uses `Maps.newConcurrectMap` while `new ConcurrentHashMap<>()` was used.

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,375 @@
...
+        return !getConfigValue(this.indexOffThreadConfigOption);
+    }
+
+    public static boolean shouldUseCache() {
+        return getConfigValue(ForgeConfig.COMMON.cachePackAccess);
+    }
+
+    private static boolean getConfigValue(Supplier<Boolean> configValue) {
+        //Yes we catch the early loading error on purpose. This feature needs to be configurable but Mojang loads resources really early.
+        //So when they are loaded early we preserve the default behaviour which is to process it on the main thread.
```
sciwhiz12: Because this method is also used to determine whether to cache pack resource accesses, this comment should be updated to reflect that.

sciwhiz12 (changes requested):
Please revise the placement of braces added in Forge code, from same-line to newline. Also, please revise the formatting of your comments, so there is a space between the comment header and its body (`//comment` to `// comment`).

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
marchermans: No it can't, there is a clear synchronization difference between a synchronized table/map and a concurrent map, and that is related to when it locks access. Synchronized tables (from the documentation) synchronize all methods on the same mutex (as if it had the synchronized access flag) this means that a single thread can read or a single thread can write, which is detrimental to our performance.

As such the concurrent hash map which supports multi-threaded read or single-threaded write is much better for resource access.

marchermans: And yes we need the atomicity, since we support off-thread indexing without holding up the main thread, so it is possible that both index hasCached are invoked simultaniously.

src/main/java/net/minecraftforge/resource/PathResourcePack.java
marchermans: That will cause confusion because the contexts of both methods are different. One is a static call that validates that the caching is even allowed, while the other only indicates if the manager holds on to the cache value currently.

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,406 @@
...
+    public static boolean shouldUseCache()
+    {
+        return getConfigValue(ForgeConfig.COMMON.cachePackAccess);
+    }
+
+    private static boolean getConfigValue(Supplier<Boolean> configValue)
+    {
+        // Yes we catch the early loading error on purpose.
+        // Both the indexing off-thread and if caching is enabled are configurable, so we need to catch the error.
+        // So when they are loaded early we preserve the default behaviour which is to not cache or index on thread.
```
ichttt, line 93: Are these config entries really needed to begin with? IMO they are very technical and if they are not respected under some circumstances they are confusing as well. Maybe just use System.getProperty instead to leave the options there for debugging purposes or maybe move them to the FML config so they load early?

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,406 @@
...
+    {
+        final PackTypeAndNamespace key = new PackTypeAndNamespace(type, resourceNamespace);
+        // Check if we even have a cached manager with the given namespace, else return an empty collection.
+        if (!managersByNamespace.containsKey(key))
+        {
+            return Collections.emptyList();
+        }
+
+        // Request the data from the manager.
+        return managersByNamespace.get(key).getResources(inputPath, filter);
```
ichttt: Instead of calling containsKey followed by a get, you should be able to just call get and null-check the value. That way no double lookup is needed

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,406 @@
...
+     * Indicates if the given namespace is completely cached for the given pack type.
+     *
+     * @param packType  The pack type to check.
+     * @param namespace The namespace to check.
+     * @return True if the namespace is completely cached for the given pack type.
+     */
+    public boolean hasCached(final PackType packType, final String namespace)
+    {
+        final PackTypeAndNamespace key = new PackTypeAndNamespace(packType, namespace);
+        return managersByNamespace.containsKey(key) && managersByNamespace.get(key).cacheLoaded();
```
ichttt: Same here, containsKey followed by get, not really needed to do two lookups

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,406 @@
...
+     * @param path The path to walk.
+     * @return A stream of paths which reside in the given path.
+     * @throws IOException Thrown when the path does not exist or reading of path information failed.
+     */
+    @SuppressWarnings("resource")
+    // We close this automatically later on when the result is actually used. This is just a factory.
+    private Stream<Path> createWalkingStream(final Path path) throws IOException
+    {
+        return Files.walk(path)
+                .filter(resourcePath -> !resourcePath.toString().endsWith(".mcmeta"));
```
ichttt, line 140: Why are mcmeta files excluded? Can you add a comment to this?

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,406 @@
...
+            if (managersByNamespace.containsKey(key) && !supportsReloading)
+            { // If we do not support reloading we just skip this.
+                return;
+            }
+
+            // Create a new manager, overriding the previous one if it exists.
+            this.managersByNamespace.put(key, new NamespacedResourceCacheManager(packType, namespace, this.shouldIndexOffThread(), pathBuilder, this::createWalkingStream));
+
+            // Index the inner manager (which will for sure exist, and will know the pack type and namespace already)
+            this.managersByNamespace.get(key).index();
```
ichttt: Why put and insteantly get when you can just use a local variable for the NamespacedResourceCacheManager and then call index on that?

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,406 @@
...
+
+    /**
+     * Functional callback interface to get a walkable stream of paths.
+     * Supports throwing {@link IOException} if the stream can not be created.
+     */
+    @FunctionalInterface
+    private interface PathWalkerFactory
+    {
+        Stream<Path> createWalkingStream(Path path) throws IOException;
+    }
```
ichttt, line 417: It seems like there is only one implementation for this, and thats createWalkingStream. Is there a reason for this extra layer of abstraction? Seems kinda pointless to me

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,406 @@
...
+
+            // Stream element resource that combines a normalized path and a joined path using the "/" as separator.
+            record PathWithLocationPath(Path path, String locationPath)
+            {
+            }
+
+            // Build a walkable stream, process it
+            try (final Stream<Path> paths = pathFinder.createWalkingStream(rootPath))
+            {
+                paths.parallel() // Run the stream in parallel
```
ichttt, line 311: I'm not really sure if this really needs to be parrallel. The indexing if already on a separate thread pool, which should in theory already exhaust all usable CPU threads, and because of this injectIntoCache has to use a costly CopyOnWriteArrayList instead of a regular ArrayList.
The Comment on CopyOnWriteArrayList also states: `This is ordinarily too costly, but may be more efficient than alternatives when traversal operations vastly outnumber mutations`. And that's not really the case here, as it is not iterated until the cache is build, at which point the list is immutable either way.

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
```diff
@@ -0,0 +1,406 @@
...
+         */
+        public Collection<ResourceLocation> getResources(final Path inputPath, final Predicate<ResourceLocation> filter)
+        {
+            // Use the input path prefix, we can not depend on a toString call here, since we need the / as separator.
+            final String pathEntry = Joiner.on('/').join(inputPath);
+
+            // Since we inject into the cache recursively we can now just grab the map entry that is the prefix and loop over its values.
+            // We use getOrDefault with a stream combination since the returned list needs to be mutable.
+            return entriesByPathPrefix.getOrDefault(pathEntry, Collections.emptyList()).stream()
+                    .map(ResourceCacheEntry::resourceLocation)
```
ichttt, line 393: Is there a reason that this mapping needs to be done? Why not just only store the ResourceLocation in the map? Would save a tiny bit of runtime cost and reduce the memory usage of the cache

ichttt (commented):
I've glimpsed over the ResourceCacheManager and tried to think of it from a performance perspective

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
marchermans: The problem is that pack size plays a role here in the performance gain. There are constructs where the mod asset count vs the vanilla asset count needs to be kept in some form of balance. Yes in most situations it makes sense to for them to be indexed all off-thread. But in a vanilla like scenario where mods are in play which have virtually no assets there are some benefits to be gained from performing mod indexing on thread. The same can be set for vanilla. If you play in a vanilla only environment it might make sense to do the init on thread, wait for it and then completely use in cache only resources.

So I decided to handle it this way. It provides greater flexibility in situations where it makes sense.

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
marchermans: This was default code which was already in place. It has existed since the beginning of the way forge looks up resources without a comment as to why. So no clue, as I have no idea why it is the way it is.

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
marchermans: It might be needed in the future, right now we handle Path and Vanilla only as packs we adapt, but folder and file might also gain benefits from this treatment, although I was not able to completely verify this, which is why this abstraction layer exists for now, and when I or somebody else have more time this system can be extended.

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
marchermans: You are correct that can be an ArrayList if I am not mistaken, at least with the current setup, since the resource access is guarded behind a hasCached check, I will adapt this.

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
marchermans: Note that the filter takes a path, so we need both the path as well as the resourcelocation in the cache
Hence the complete object.

marchermans: No you are right, fixed.

src/main/java/net/minecraftforge/resource/ResourceCacheManager.java
ichttt: Well for that to be safe the stream should *not* be parallel, though (that's why the comment was on that line)

src/main/java/net/minecraftforge/resource/DelegatingResourcePack.java
```diff
@@ -42,6 +41,19 @@ public DelegatingResourcePack(String id, String name, PackMetadataSection packIn
         this.namespacesData = this.buildNamespaceMap(PackType.SERVER_DATA, delegates);
     }
 
+    @Override
+    public void initForNamespace(final String nameSpace) {
```
pupnewfster: Bracket should be on a new line

src/main/java/net/minecraftforge/resource/DelegatingResourcePack.java
```diff
@@ -42,6 +41,19 @@ public DelegatingResourcePack(String id, String name, PackMetadataSection packIn
...
+        this.delegates.forEach(delegate -> delegate.initForNamespace(nameSpace));
+    }
+
+    @Override
+    public void init(final PackType packType) {
```
pupnewfster: Bracket should be on a new line

pupnewfster (approved):

src/main/java/net/minecraftforge/common/extensions/IForgePackResources.java
```diff
@@ -5,10 +5,16 @@
...
 public interface IForgePackResources
 {
     default boolean isHidden()
     {
         return false;
     }
+
+    default void initForNamespace(final String nameSpace) {}
+
+    default void init(final PackType packType) {}
```
cpw, line 19: Could these two default to the behaviour in VanillaPackResources above - to avoid those extra patches there?

cpw (changes requested):

src/main/java/net/minecraftforge/common/extensions/IForgePackResources.java
marchermans: No, since the cache manager is needed to init a given namespace properly, the first one is not changeable. The second one is debatable, be realistically only the vanilla pack has vanilla resources, and as such it makes no sense to have it default to that.

cpw (approved):

embeddedt:
Sorry to hijack a merged PR. Is a backport of this to the 1.18.x branch planned, and if not, would it be accepted if someone were to work on it?

PR #8834, opened 2022-07-08
Pass renderType to IForgeBakedModel.useAmbientOcclusion

bdew:
It would be nice to have RenderType passed to IForgeBakedModel.useAmbientOcclusion so that a model that renders in multiple RenderTypes can have AO in some but not other.

Before the recent rendering rework I had checks for MinecraftForgeClient.getRenderType in useAmbientOcclusion but that's not possible anymore.

### DIFF
--- a/patches/minecraft/net/minecraft/client/resources/model/MultiPartBakedModel.java.patch
+++ b/patches/minecraft/net/minecraft/client/resources/model/MultiPartBakedModel.java.patch
@@ -53,12 +53,16 @@
        }
     }
  
-@@ -77,6 +_,10 @@
+@@ -77,6 +_,14 @@
        return this.f_119453_;
     }
  
 +   public boolean useAmbientOcclusion(BlockState state) {
 +      return this.defaultModel.useAmbientOcclusion(state);
++   }
++
++   public boolean useAmbientOcclusion(BlockState state, net.minecraft.client.renderer.RenderType renderType) {
++      return this.defaultModel.useAmbientOcclusion(state, renderType);
 +   }
 +
     public boolean m_7539_() {
--- a/patches/minecraft/net/minecraft/client/resources/model/WeightedBakedModel.java.patch
+++ b/patches/minecraft/net/minecraft/client/resources/model/WeightedBakedModel.java.patch
@@ -22,13 +22,18 @@
        }).orElse(Collections.emptyList());
     }
  
-@@ -38,6 +_,11 @@
+@@ -38,6 +_,16 @@
        return this.f_119542_.m_7541_();
     }
  
 +   @Override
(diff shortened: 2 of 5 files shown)


marchermans:
Please run the check command, especially checkExcs is needed. Since you addede an override to a vanilla method.

bdew:
I did run checkExcs and it didn't show anything. I don't see a "check" task... anything else i need to check?

marchermans:
Asked internally no new excs is needed it was a misunderstanding.
Assigning.

marchermans (approved):

patches/minecraft/net/minecraft/client/resources/model/MultiPartBakedModel.java.patch
```diff
@@ -57,8 +57,8 @@
        return this.f_119453_;
     }
  
-+   public boolean useAmbientOcclusion(BlockState state) {
-+      return this.defaultModel.useAmbientOcclusion(state);
++   public boolean useAmbientOcclusion(BlockState state, net.minecraft.client.renderer.RenderType renderType) {
++      return this.defaultModel.useAmbientOcclusion(state, renderType);
 +   }
```
marchermans: Leave this method in place. Add the override manually in case this is called already. Preserves compatibility.

patches/minecraft/net/minecraft/client/resources/model/WeightedBakedModel.java.patch
```diff
@@ -27,8 +27,8 @@
     }
  
 +   @Override
-+   public boolean useAmbientOcclusion(BlockState state) {
-+      return this.f_119542_.useAmbientOcclusion(state);
```
marchermans: Leave this method in place. Add the override manually in case this is called already. Preserves compatibility.

src/main/java/net/minecraftforge/client/model/BakedModelWrapper.java
```diff
@@ -52,9 +52,9 @@ public boolean useAmbientOcclusion()
     }
 
     @Override
-    public boolean useAmbientOcclusion(BlockState state)
+    public boolean useAmbientOcclusion(BlockState state, RenderType renderType)
     {
-        return originalModel.useAmbientOcclusion(state);
```
marchermans: Leave this method in place. Add the override manually in case this is called already. Preserves compatibility.

marchermans (changes requested):
Either keep the state only methods, or remove said method and only leave the one with the state and the rendertype.

bdew:
Changed the patch to leave the original methods in place, is that ok now?

marchermans (approved):

PR #8837, opened 2022-07-08
Allow modded blocks overriding canStickTo prevent sticking to vanilla blocks/other modded blocks

### LINKED ISSUE #7122: Ability to hook into a canStickTo in IForgeBlock interface
Please add the ability to hook into this interface method so we can create sticky blocks that don't interact with vanilla sticky blocks (honey and slime)

TelepathicGrunt:
Basically, before, overriding canStickTo in a modded block to try and prevent sticking to vanilla slime blocks would fail when the game checks the stickiness through the vanilla block. This PR makes it so that Forge's blockstate canStickTo checks to make sure both blocks passes their canStickTo checks.

Example modded sticky block that uses canStickTo to prevent sticking to vanilla slime blocks but sticks to everything else
https://user-images.githubusercontent.com/40846040/178077123-354022ad-57f2-4b74-a23a-96cd182cfa43.mp4

This PR was done in a way to try and prevent a breaking change. Though if the other block check is not desirable to be in the blockstate, I can move the check to the patches instead. Also, set non-null on param for blockstate canStickTo because passing in null will crash majority of blocks with the default canStickTo implementation even before this pr.

Fixes #7122.

### DIFF
--- a/patches/minecraft/net/minecraft/world/level/block/piston/PistonStructureResolver.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/block/piston/PistonStructureResolver.java.patch
@@ -46,7 +46,7 @@
 +               oldState = blockstate;
                 blockstate = this.f_60409_.m_8055_(blockpos);
 -               if (blockstate.m_60795_() || !m_155939_(blockstate1, blockstate) || !PistonBaseBlock.m_60204_(blockstate, this.f_60409_, blockpos, this.f_60413_, false, this.f_60413_.m_122424_()) || blockpos.equals(this.f_60410_)) {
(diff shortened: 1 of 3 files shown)


marchermans:
How does this behave with just vanilla blocks?

TelepathicGrunt:
canStickTo is already symmetrical for vanilla blocks because forge has an explicit check for honey touching slime and slime touching honey. So calling canStickTo on the other block changes nothing in regards to vanilla (well, changes nothing in what forge is doing with vanilla blocks). Only new modded blocks can change the behavior.
image

marchermans (approved):

amadornes:
Tagging @KnightMiner on this. If I'm not mistaken, he has some blocks in TiCon that actually take advantage of this only being called on one of the blocks to allow for one-way stickiness.

PR #8838, opened 2022-07-09
Add DamageEntityEvent and onDamageEntity() for post-hurt attack code with access to attack strength.

bconlon1:
This event is similar to `AttackEntityEvent`, however it is called after the enemy has had all hurt effects applied by a player and also passes the player's attack strength into the event as well (which is difficult to access by any other means after an entity has been hurt because of `resetAttackStrengthTicker()` being called prior to hurt-related code in the `attack()` method). This is useful for if an item ability is desired to only apply if an entity has been hurt, but also if the player attacked with full attack strength.

`onDamageEntity()` meanwhile is similar to `onLeftClickEntity()`, however it follows the same use-case as `DamageEntityEvent` as detailed above.


eerussianguy:
Would it be more appropriate to call it PlayerDamageEntityEvent?

bconlon1:
My intention was having naming consistency with AttackEntityEvent due to their similarities, with the main difference being how they're called at different stages of the player attack code.

bconlon1:
Going to try a different implementation to accomplish what this is trying to solve without needing an extra event.

PR #8844, opened 2022-07-09
[1.19.x] Fix MultiPartBakedModel not using model-driven render types

SilverDavidMC:
The title pretty much says it all. `MultiPartBakedModel` doesn't implement `IForgeBakedModel.getRenderTypes`, meaning multipart models can't utilize with model driven render types.

# Implementation
The `getRenderTypes` method for block states will get all render types applicable to the block state passed, based on what models will be rendered. These render types are then cached in `stateRenderTypeCache` for future access.

The `getRenderTypes` method for item stacks will check all possible models for their render types and return all of them. This is not cached since item stacks may use NBT to determine render types.

`getQuads` has an additional check for if the model who's quads we're about to add should be rendered given the render type passed. Without this, all the models will attempt to render under all render types associated with this model, negating the purpose of the above methods.

# Testing

Sample resource pack: [Test Pack.zip](https://github.com/MinecraftForge/MinecraftForge/files/9077852/Test.Pack.zip)
Screenshot of the resource pack in action:

### DIFF
--- a/patches/minecraft/net/minecraft/client/resources/model/MultiPartBakedModel.java.patch
+++ b/patches/minecraft/net/minecraft/client/resources/model/MultiPartBakedModel.java.patch
@@ -32,7 +32,7 @@
        if (p_235050_ == null) {
           return Collections.emptyList();
        } else {
-@@ -60,16 +_,17 @@
+@@ -60,16 +_,18 @@
              this.f_119460_.put(p_235050_, bitset);
           }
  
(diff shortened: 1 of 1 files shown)


marchermans:
Looking at this quickly I think you will need to find a away to clear this cache on resource reload.

SilverDavidMC:
Just had that thought, looking into it now

SilverDavidMC:
It looks to me like the cache should be fine since it's stored in the `MultiPartBakedModel` instance, which gets recreated on resource reload, replacing the old one. I was able to reload assets several times after modifying the resource pack with no issues.

marchermans (approved):

PR #8852, opened 2022-07-10
[1.19.x] Allow block render types to be set in datagen

AnonymousHacker1279:
Since #8786, block render types should be set via model JSON. Most<sup>1</sup> built-in data generators in `BlockStateProvider` currently do not allow for the render type to be set. That is the case for the following block types:
- Axis/log
- Stairs
- Fences
- Fence gates
- Walls
- Panes
- Doors
- Trapdoors

<sup>*1 - Other blocks allow a model file to be specified, meaning the render type can be set there.*</sup>

This PR adds methods which accept a `String` or `ResourceLocation` for the render type. The previous methods which do not accept a render type have not been removed.

The image below shows datagenerated models for the block types mentioned above:
image
All of these models are using the `minecraft:cutout_mipped` render type. The example mod used to generate these can be found [here](https://gist.github.com/AnonymousHacker1279/936fde0626151ee8143c7a7e216981fd).

### DIFF
--- a/src/main/java/net/minecraftforge/client/model/generators/BlockStateProvider.java
+++ b/src/main/java/net/minecraftforge/client/model/generators/BlockStateProvider.java
@@ -204,7 +204,27 @@ public abstract class BlockStateProvider implements DataProvider {
     }
 
     public void axisBlock(RotatedPillarBlock block, ResourceLocation side, ResourceLocation end) {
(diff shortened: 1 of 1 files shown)


marchermans (approved):
Works, I hate that we need that many overloads to make it possible, but oh well.

PR #8854, opened 2022-07-10
Multiple tweaks and fixes to the recent changes in the client refactor PR: Part 2

amadornes:
### Add getter for the component names in an unbaked geometry
The old multipart geometries allowed you to access the children directly, but as it was agreed during the planning of the big refactor that direct access to children before baking should not be part of the API, that entire system was removed. However, while visibility configuration via `IGeometryBakingContext` remains and is the correct way to bake only part of the geometry, there is no longer a way to obtain a list of all the component names to be able to properly control their visibility. This PR re-adds that as per @pupnewfster's request.

### Fix render type hint not being copied in BlockGeometryBakingContext
This fixes an issue that prevents people from using render types with vanilla's generated item models, as they rely on `BlockGeometryBakingContext` being copied, and the render type hint was missed.

### Ensure BlockRenderDispatches's renderSingleBlock uses the correct buffer
The new model-controlled render types mean that this approach is no longer correct. This replaces it with a more appropriate one that visits every render type in the model and properly adapts it to the correct buffer target for dynamic rendering.

### DIFF
--- a/src/main/java/net/minecraftforge/client/model/CompositeModel.java
+++ b/src/main/java/net/minecraftforge/client/model/CompositeModel.java
@@ -125,6 +125,12 @@ public class CompositeModel implements IUnbakedGeometry<CompositeModel>
         return textures;
     }
 
+    @Override
+    public Set<String> getComponentNames(boolean recursive)
+    {
+        return children.keySet();
+    }
+
     public static class Baked implements IDynamicBakedModel
     {
         private final boolean isAmbientOcclusion;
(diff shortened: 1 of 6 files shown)


src/main/java/net/minecraftforge/client/model/CompositeModel.java
```diff
@@ -125,6 +125,12 @@ public Collection<Material> getMaterials(IGeometryBakingContext context, Functio
         return textures;
     }
 
+    @Override
+    public Set<String> getComponentNames(boolean recursive)
+    {
+        return children.keySet();
```
marchermans, line 131: Is it not possible that a composite model is make up out of another composite model?
As far as I can tell that is the case soooo... This should respect the recursive parameter, no?

marchermans (approved):

marchermans (approved):

marchermans (approved):

PR #8855, opened 2022-07-11
[1.18.2] Allow blocks to provide a dynamic MaterialColor for display on maps

XFactHD:
This PR is the LTS backport of #8812

This PR allows blocks to provide the `MaterialColor` for display on maps dependent on dynamic data.

The context needed for this is already present, it's just not forwarded from the `BlockState` to the `BlockState`. This PR adds an extension method forwarding the necessary context to the block.

Example use cases would be:
- Allow mods like [RGB Blocks](https://www.curseforge.com/minecraft/mc-mods/rgb-blocks-forge) to provide the `MaterialColor` that is closest to the actual RGB value of the block
- Allow mods like [FramedBlocks](https://www.curseforge.com/minecraft/mc-mods/framedblocks) that add blocks that mimic other blocks to proxy the `MaterialColor` of the block that is being mimicked

### DIFF
--- a/patches/minecraft/net/minecraft/world/level/block/state/BlockBehaviour.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/block/state/BlockBehaviour.java.patch
@@ -89,7 +89,8 @@
        }
  
        public MaterialColor m_60780_(BlockGetter p_60781_, BlockPos p_60782_) {
-          return this.f_60598_;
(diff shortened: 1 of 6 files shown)


marchermans (approved):

marchermans:
Looks identical to the 1.19 version as far as I can see :lgtm:

PR #8856, opened 2022-07-11
Provide more information in the crashreport on the source of a crash

marchermans:
Output the full path in a crash report so it is easier to find the outer mod when a crash in Jar-In-Jar occurs.

Currently, we only output the file name "foo.jar" which is not very informative when a crash happens inside a jar that is part of another jar.

This PR outputs the full path for the mod file entry in the crash report so it can be easily found.


cpw (commented):
Did you mean to modify the tests in this PR?

PR #8857, opened 2022-07-11
Make a proper initialization state for RemoveBlockGoal

### LINKED ISSUE #8853: ZombieTurtleEggAttackGoal ai broken in 1.19
**Minecraft Version:** 1.19

**Forge Version:** 41.0.85

**Logs:**  N/A

**Steps to Reproduce:**
 1. Go to the nether and teleport away from 0,0 so that 0,0 is not loaded (break your portal so it doesn't force load chunks).
 2. Place down a turtle egg and spawn a zombified piglin
 3. It does NOT attack the turtle egg.
 4. Port back to 0,0 and repeat
 5. It DOES attack the turtle egg.

**Description of issue:**

The issue is with this code used by the ZombieAttackTurtleEggGoal - actually RemoveBlockGoal.canUse()

https://github.com/MinecraftForge/MinecraftForge/blob/05297fd72fa99e9ff020fee3ee365ad51bcaf8f0/src/main/java/net/minecraftforge/common/ForgeHooks.java#L1106

On initial pass, that code gets called before the block to break has been determined. 
This means it will check if 0,0,0 (BlockPos.ZERO in MoveToBlockGoal) is loaded and returns false if not. 
This stops the AI from working away from 0,0

Actually reported by Hsandin7 in the forums:
https://forums.minecraftforge.net/topic/113379-pigman-dont-go-after-turtle-eggs/

marchermans:
This fixes a case where the goal tries to access the 0,0,0 position as that is its initial value.
Since this position is not always loaded it causes an issue where the goal would get stuck.

Closes: #8853

Note: there is an aligned issue where a turtle egg on 0,0,0 would cause zombies to track there regardless of world position. This implicitly fixes this but it is only tangentially related as such no link to a Mojira issue is made.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/ai/goal/RemoveBlockGoal.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/ai/goal/RemoveBlockGoal.java.patch
@@ -1,11 +1,24 @@
 --- a/net/minecraft/world/entity/ai/goal/RemoveBlockGoal.java
 +++ b/net/minecraft/world/entity/ai/goal/RemoveBlockGoal.java
-@@ -34,7 +_,7 @@
+@@ -31,10 +_,11 @@
+       super(p_25841_, p_25842_, 24, p_25843_);
+       this.f_25836_ = p_25840_;
+       this.f_25837_ = p_25841_;
++      this.f_25602_ = null; // Forge: We explicitly set this to null, it gets initialized in the onUse call.
     }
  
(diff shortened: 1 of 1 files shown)


LexManos:
Ya, it's kinda bad that vanilla's null use is not fully implemented, but this change moves things more in line with what we intended.

https://github.com/MinecraftForge/MinecraftForge/commit/5e8c786406d5add1a2982c212f33e9ce849b21a3

PR #8860, opened 2022-07-12
[1.19.x] Block Model Builder Root Transform Support

Rover656:
Allows data generation of the root "transform" node in block model jsons (as present in `ExtendedBlockModelDeserializer`).
Use case discussed in discord: https://discord.com/channels/313125603924639766/983834532904042537/996011350990000129

### DIFF
--- a/src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
+++ b/src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
@@ -25,10 +25,8 @@ import javax.annotation.Nullable;
  * @see ModelProvider
  * @see ModelBuilder
  */
-public class BlockModelBuilder extends ModelBuilder<BlockModelBuilder> {
-
-    private static final Vector3f ONE = new Vector3f(1, 1, 1);
-
+public class BlockModelBuilder extends ModelBuilder<BlockModelBuilder>
+{
     private final RootTransformBuilder rootTransform = new RootTransformBuilder();
 
     public BlockModelBuilder(ResourceLocation outputLocation, ExistingFileHelper existingFileHelper)
@@ -56,43 +54,24 @@ public class BlockModelBuilder extends ModelBuilder<BlockModelBuilder> {
         return json;
     }
 
-    private JsonArray writeVec3(Vector3f vector)
-    {
-        JsonArray array = new JsonArray();
-        array.add(vector.x());
-        array.add(vector.y());
-        array.add(vector.z());
-        return array;
-    }
-
-    private JsonArray writeQuaternion(Quaternion quaternion)
-    {
-        JsonArray array = new JsonArray();
-        array.add(quaternion.i());
-        array.add(quaternion.j());
-        array.add(quaternion.k());
-        array.add(quaternion.r());
-        return array;
-    }
-
     public class RootTransformBuilder
     {
-        public enum Origin implements StringRepresentable
+        public enum TransformOrigin implements StringRepresentable
         {
-            Center(new Vector3f(), "origin"),
-            Corner(new Vector3f(1f, 1f, 1f), "corner"),
-            OpposingCorner(new Vector3f(.5f, .5f, .5f), "opposing_corner");
+            CENTER(new Vector3f(), "origin"),
+            CORNER(new Vector3f(1f, 1f, 1f), "corner"),
+            OPPOSING_CORNER(new Vector3f(.5f, .5f, .5f), "opposing_corner");
 
             private final Vector3f vec;
             private final String name;
 
-            Origin(Vector3f vec, String name)
+            TransformOrigin(Vector3f vec, String name)
             {
                 this.vec = vec;
                 this.name = name;
             }
 
-            public Vector3f vec()
+            public Vector3f getVector()
             {
                 return vec;
             }
@@ -103,35 +82,36 @@ public class BlockModelBuilder extends ModelBuilder<BlockModelBuilder> {
                 return name;
             }
 
-            public static @Nullable Origin fromString(String originName)
+            public static @Nullable TransformOrigin fromString(String originName)
             {
-                if (Center.getSerializedName().equals(originName))
+                if (CENTER.getSerializedName().equals(originName))
                 {
-                    return Center;
+                    return CENTER;
                 }
-                if (Corner.getSerializedName().equals(originName))
+                if (CORNER.getSerializedName().equals(originName))
                 {
-                    return Corner;
+                    return CORNER;
                 }
-                if (OpposingCorner.getSerializedName().equals(originName))
+                if (OPPOSING_CORNER.getSerializedName().equals(originName))
                 {
-                    return OpposingCorner;
+                    return OPPOSING_CORNER;
                 }
                 return null;
             }
         }
+
+        private static final Vector3f ONE = new Vector3f(1, 1, 1);
 
         private Vector3f translation = Vector3f.ZERO;
         private Quaternion leftRotation = Quaternion.ONE.copy();
         private Quaternion rightRotation = Quaternion.ONE.copy();
         private Vector3f scale = ONE;
 
-        private @Nullable Origin origin;
+        private @Nullable TransformOrigin origin;
         private @Nullable Vector3f originVec;
 
         private RootTransformBuilder()
         {
-
         }
 
         /**
@@ -185,7 +165,15 @@ public class BlockModelBuilder extends ModelBuilder<BlockModelBuilder> {
             return this;
         }
 
-        public RootTransformBuilder postRotation(Quaternion postRotation) {
+        /**
+         * Sets the right rotation of the root transform.
+         *
+         * @param postRotation the right rotation
+         * @return this builder
+         * @throws NullPointerException if {@code rightRotation} is {@code null}
+         */
+        public RootTransformBuilder postRotation(Quaternion postRotation)
+        {
             return rightRotation(postRotation);
         }
 
@@ -253,7 +241,7 @@ public class BlockModelBuilder extends ModelBuilder<BlockModelBuilder> {
          * @throws NullPointerException if {@code origin} is {@code null}
          * @throws IllegalArgumentException if {@code origin} is not {@code center}, {@code corner} or {@code opposing-corner}
          */
-        public RootTransformBuilder origin(Origin origin)
+        public RootTransformBuilder origin(TransformOrigin origin)
         {
             this.origin = Preconditions.checkNotNull(origin, "Origin must not be null");;
             this.originVec = null;
@@ -297,12 +285,31 @@ public class BlockModelBuilder extends ModelBuilder<BlockModelBuilder> {
             {
                 transform.addProperty("origin", origin.getSerializedName());
             }
-            else if (originVec != null && originVec != Vector3f.ZERO)
+            else if (originVec != null && !originVec.equals(Vector3f.ZERO))
             {
                 transform.add("origin", writeVec3(originVec));
             }
 
             return transform;
+        }
+
+        private JsonArray writeVec3(Vector3f vector)
+        {
+            JsonArray array = new JsonArray();
+            array.add(vector.x());
+            array.add(vector.y());
+            array.add(vector.z());
+            return array;
+        }
+
+        private JsonArray writeQuaternion(Quaternion quaternion)
+        {
+            JsonArray array = new JsonArray();
+            array.add(quaternion.i());
+            array.add(quaternion.j());
+            array.add(quaternion.k());
+            array.add(quaternion.r());
+            return array;
         }
     }
 }
--- a/src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
+++ b/src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
@@ -90,7 +90,6 @@ public class BlockModelBuilder extends ModelBuilder<BlockModelBuilder>
          * @param y y translation
          * @param z z translation
          * @return this builder
-         * @throws NullPointerException if {@code translation} is {@code null}
          */
         public RootTransformBuilder translation(float x, float y, float z)
         {
--- a/src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
+++ b/src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
@@ -14,6 +14,7 @@ import com.mojang.math.Vector3f;
 import net.minecraft.resources.ResourceLocation;
 import net.minecraft.util.StringRepresentable;
 import net.minecraftforge.common.data.ExistingFileHelper;
+import org.jetbrains.annotations.NotNull;
 
 import javax.annotation.Nullable;
 
--- a/src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
+++ b/src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
@@ -251,9 +251,9 @@ public class BlockModelBuilder extends ModelBuilder<BlockModelBuilder>
 
         public enum TransformOrigin implements StringRepresentable
         {
-            CENTER(new Vector3f(), "origin"),
-            CORNER(new Vector3f(1f, 1f, 1f), "corner"),
-            OPPOSING_CORNER(new Vector3f(.5f, .5f, .5f), "opposing_corner");
+            CENTER(new Vector3f(.5f, .5f, .5f), "origin"),
+            CORNER(Vector3f.ZERO, "corner"),
+            OPPOSING_CORNER(ONE, "opposing_corner");
 
             private final Vector3f vec;
             private final String name;
--- a/src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
+++ b/src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
@@ -5,8 +5,16 @@
 
 package net.minecraftforge.client.model.generators;
 
+import com.google.common.base.Preconditions;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
+import com.mojang.math.Quaternion;
+import com.mojang.math.Transformation;
+import com.mojang.math.Vector3f;
 import net.minecraft.resources.ResourceLocation;
 import net.minecraftforge.common.data.ExistingFileHelper;
+
+import javax.annotation.Nullable;
 
 /**
  * Builder for block models, does not currently provide any additional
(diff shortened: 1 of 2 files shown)


src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -18,7 +26,231 @@
...
+
+        // Write the transform to an object
+        JsonObject transform = new JsonObject();
+
+        if (!rootTransform.translation.equals(Vector3f.ZERO))
+        {
+            transform.add("translation", writeVec3(rootTransform.translation));
+        }
+
+        if (!rootTransform.scale.equals(new Vector3f(1, 1, 1)))
```
sciwhiz12: The `new Vector3f(1, 1, 1)` should be stored in a private constant field.

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -18,7 +26,231 @@
...
+        if (!leftRotation.equals(Quaternion.ONE))
+        {
+            transform.add("left_rotation", writeQuaternion(leftRotation));
+        }
+
+        Quaternion rightRotation = rootTransform.rightRotation;
+        if (!rightRotation.equals(Quaternion.ONE))
+        {
+            transform.add("right_rotation", writeQuaternion(rightRotation));
+        }
```
sciwhiz12: What is more common parlance: `left_rotation` and `right_rotation`, or `rotation` and `post_rotation`? Since the deserializer code in `TransformationHelper` accepts both pairs, we should write out with the name that is more commonly understood/known.

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -18,7 +26,231 @@
...
+        array.add(quaternion.r());
+        return array;
+    }
+
+    public class RootTransformBuilder
+    {
+        private Vector3f translation = Vector3f.ZERO;
+        private Quaternion leftRotation = Quaternion.ONE.copy();
+        private Quaternion rightRotation = Quaternion.ONE.copy();
+        private Vector3f scale = new Vector3f(1, 1, 1);
```
sciwhiz12: Use the previously-suggested constant for `new Vector3f(1, 1, 1)`.

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -18,7 +26,231 @@
...
+        }
+
+        /**
+         * Sets the right rotation of the root transform.
+         *
+         * @param rightRotation the right rotation
+         * @return this builder
+         * @throws NullPointerException if {@code rightRotation} is {@code null}
+         */
+        public RootTransformBuilder rightRotation(Quaternion rightRotation)
```
sciwhiz12, line 159: Should we also provide a `postRotation(Quaternion)` that calls to this method, to mimic the pair of `rotation` and `leftRotation`?

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -18,7 +26,231 @@
...
+         * Sets the translation of the root transform.
+         *
+         * @param translation the translation
+         * @return this builder
+         * @throws NullPointerException if {@code translation} is {@code null}
+         */
+        public RootTransformBuilder translation(Vector3f translation)
+        {
+            Preconditions.checkNotNull(translation, "Translation must not be null");
+            this.translation = translation;
```
sciwhiz12: In general for all these methods, the `checkNotNull` can be inlined to the field assignment as it returns the value passed when not `null`.

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -18,7 +26,231 @@
...
+         *
+         * @param origin the origin name
+         * @return this builder
+         * @throws NullPointerException if {@code origin} is {@code null}
+         * @throws IllegalArgumentException if {@code origin} is not {@code center}, {@code corner} or {@code opposing-corner}
+         */
+        public RootTransformBuilder origin(String origin)
+        {
+            Preconditions.checkNotNull(origin, "Origin must not be null");
+            Preconditions.checkArgument(origin.equals("center") || origin.equals("corner") || origin.equals("opposing-corner"), "Invalid value for origin name.");
```
sciwhiz12: We should be using an enum for this instead of requiring hardcoded strings. The enum should probably be `StringRepresentable`, with the enum values should include the friendly name (these three strings) and the origin vector as seen in the fields of `TransformationHelper.Deserializer`. 

Accordingly, `TransformationHelper.Deserializer#parseOrigin` would be modified to use the names from the enum values rather than hardcoded strings.

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -18,7 +26,231 @@
...
+        JsonArray array = new JsonArray();
+        array.add(quaternion.i());
+        array.add(quaternion.j());
+        array.add(quaternion.k());
+        array.add(quaternion.r());
+        return array;
+    }
+
+    public class RootTransformBuilder
+    {
```
sciwhiz12, line 58: This should have a `private` constructor.

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -18,7 +26,231 @@
...
+        return rootTransform;
+    }
+
+    @Override
+    public JsonObject toJson()
+    {
+        JsonObject json = super.toJson();
+
+        // Write the transform to an object
+        JsonObject transform = new JsonObject();
```
sciwhiz12: I'm thinking that the code that writes the root transform to a JSON object should be a method on `RootTransformBuilder`, which is then called from here.

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
Rover656: Yeah I think rotation and post_rotation are more descriptive.

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
Rover656: An enum does sound like a good idea, is it a good idea to place the enum within TransformationHelper with the rest of the transformation logic, inside that common.util package or in the model package somewhere?

sciwhiz12: Having it either be a nested class of `TransformationHelper` or a nested class of `BlockModelBuilder#RootTransformBuilder` should be fine. The former would be a more logical grouping of transformation-related classes, but the latter would be better in terms of being close to where the enumeration is actually used. I'd probably say the latter, but your choice between the two.

Rover656: In 66ff28c6a3c1cfc610b99932ebe43c937e5d6089 I have added an enum to RootTransformBuilder, let me know if you'd like anything changed with it, I've implemented it as best as I could think :)

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -18,7 +27,282 @@
...
+
+    private JsonArray writeQuaternion(Quaternion quaternion)
+    {
+        JsonArray array = new JsonArray();
+        array.add(quaternion.i());
+        array.add(quaternion.j());
+        array.add(quaternion.k());
+        array.add(quaternion.r());
+        return array;
+    }
```
sciwhiz12: These should be moved to `RootTransformBuilder`, since that is where these methods are used.

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -18,7 +27,282 @@
...
+
+    public class RootTransformBuilder
+    {
+        public enum Origin implements StringRepresentable
+        {
+            Center(new Vector3f(), "origin"),
+            Corner(new Vector3f(1f, 1f, 1f), "corner"),
+            OpposingCorner(new Vector3f(.5f, .5f, .5f), "opposing_corner");
```
sciwhiz12: The convention for naming enum values is that they use `SCREAMING_SNAKE_CASE`, so these should be `CENTER`, `CORNER`, and `OPPOSING_CORNER`.

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -18,7 +27,282 @@
  */
 public class BlockModelBuilder extends ModelBuilder<BlockModelBuilder> {
 
-    public BlockModelBuilder(ResourceLocation outputLocation, ExistingFileHelper existingFileHelper) {
+    private static final Vector3f ONE = new Vector3f(1, 1, 1);
```
sciwhiz12: This should be in `RootTransformBuilder`, where it is used.

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -18,7 +27,282 @@
...
+        private Quaternion leftRotation = Quaternion.ONE.copy();
+        private Quaternion rightRotation = Quaternion.ONE.copy();
+        private Vector3f scale = ONE;
+
+        private @Nullable Origin origin;
+        private @Nullable Vector3f originVec;
+
+        private RootTransformBuilder()
+        {
+
```
sciwhiz12: Unnecessary blank space.

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -18,7 +27,282 @@
...
+         * @return this builder
+         * @throws NullPointerException if {@code rightRotation} is {@code null}
+         */
+        public RootTransformBuilder rightRotation(Quaternion rightRotation)
+        {
+            this.rightRotation = Preconditions.checkNotNull(rightRotation, "Rotation must not be null");
+            return this;
+        }
+
+        public RootTransformBuilder postRotation(Quaternion postRotation) {
```
sciwhiz12: Formatting: braces. Also, this should have javadocs.
```suggestion
        public RootTransformBuilder postRotation(Quaternion postRotation)
        {
```

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -18,7 +27,282 @@
...
+        array.add(quaternion.i());
+        array.add(quaternion.j());
+        array.add(quaternion.k());
+        array.add(quaternion.r());
+        return array;
+    }
+
+    public class RootTransformBuilder
+    {
+        public enum Origin implements StringRepresentable
```
sciwhiz12: Should we name it a bit more particularly, like `TransformOrigin`?

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -18,7 +27,282 @@
...
+            if (!rightRotation.equals(Quaternion.ONE))
+            {
+                transform.add("post_rotation", writeQuaternion(rightRotation));
+            }
+
+            if (origin != null)
+            {
+                transform.addProperty("origin", origin.getSerializedName());
+            }
+            else if (originVec != null && originVec != Vector3f.ZERO)
```
sciwhiz12: Comparing against `Vector3f.ZERO` should be using `.equals(Object)`, not a reference equality check.
```suggestion
            else if (originVec != null && !originVec.equals(Vector3f.ZERO))
```

src/main/java/net/minecraftforge/common/util/TransformationHelper.java
```diff
@@ -108,9 +110,9 @@ public static boolean epsilonEquals(Vector4f v1, Vector4f v2, float epsilon)
 
     public static class Deserializer implements JsonDeserializer<Transformation>
     {
-        private static final Vector3f ORIGIN_CORNER = new Vector3f();
-        private static final Vector3f ORIGIN_OPPOSING_CORNER = new Vector3f(1f, 1f, 1f);
-        private static final Vector3f ORIGIN_CENTER = new Vector3f(.5f, .5f, .5f);
+        private static final Vector3f ORIGIN_CORNER = Origin.Corner.vec();
+        private static final Vector3f ORIGIN_OPPOSING_CORNER = Origin.OpposingCorner.vec();
+        private static final Vector3f ORIGIN_CENTER = Origin.Center.vec();
```
sciwhiz12: It looks like we can remove these constants (which should be of no effect to binary compatibility, since these are `private`) and simply inline the three remaining uses of these constants.

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -18,7 +27,282 @@
...
+            private final Vector3f vec;
+            private final String name;
+
+            Origin(Vector3f vec, String name)
+            {
+                this.vec = vec;
+                this.name = name;
+            }
+
+            public Vector3f vec()
```
sciwhiz12: This should be named in the style of a getter. Probably `getVector()` can suffice, unless there's a better name than just 'vector'.

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
Rover656: I've gone with getVector for now

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
Rover656: I agree, I've gone with that

sciwhiz12: Also, could you please move the enum class nearer to the bottom of the class?

Rover656: can do

Rover656: I've placed it above the private members of RootTransformBuilder to aide anyone reading this as a reference

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -16,9 +25,291 @@
...
+                this.name = name;
+            }
+
+            public Vector3f getVector()
+            {
+                return vec;
+            }
+
+            @Override
+            public String getSerializedName()
```
sciwhiz12: Mark this method as `@NotNull` so the IDEA warning is silenced.
```suggestion
            @Override
            @NotNull
            public String getSerializedName()
```

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -16,9 +25,291 @@
...
+         * Sets the origin of the root transform.
+         *
+         * @param origin the origin name
+         * @return this builder
+         * @throws NullPointerException if {@code origin} is {@code null}
+         * @throws IllegalArgumentException if {@code origin} is not {@code center}, {@code corner} or {@code opposing-corner}
+         */
+        public RootTransformBuilder origin(TransformOrigin origin)
+        {
+            this.origin = Preconditions.checkNotNull(origin, "Origin must not be null");;
```
sciwhiz12: Remove the unnecessary semicolon.
```suggestion
            this.origin = Preconditions.checkNotNull(origin, "Origin must not be null");
```

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -16,9 +25,291 @@
...
+            Preconditions.checkNotNull(transformation, "Transformation must not be null");
+            this.translation = transformation.getTranslation();
+            this.leftRotation = transformation.getLeftRotation();
+            this.rightRotation = transformation.getRightRotation();
+            this.scale = transformation.getScale();
+            return this;
+        }
+
+        /**
+         * Sets the origi of the root transform.
```
sciwhiz12: Typo:
```suggestion
         * Sets the origin of the root transform.
```

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -16,9 +25,291 @@
...
+            }
+
+            return transform;
+        }
+
+        public enum TransformOrigin implements StringRepresentable
+        {
+            CENTER(new Vector3f(), "origin"),
+            CORNER(new Vector3f(1f, 1f, 1f), "corner"),
+            OPPOSING_CORNER(new Vector3f(.5f, .5f, .5f), "opposing_corner");
```
sciwhiz12: It seems like these are different from the now-removed constants in `TransformationHelper`. `CENTER` should be the all-`.5f`, `CORNER` should be the all-zero, and `OPPOSING_CORNER` should be the all-one vector.

Additionally, these should be using the `Vector3f.ZERO` and `RootTransformBuilder.ONE` constants.

sciwhiz12 (changes requested):
I've poked around with this PR by trying to datagenerate the composite model test, and I've hit a pain point while writing up the code: the methods which take in a `Vector3f` should also have overloads which take in `float, float, float`, so the caller does not have to manually construct the `Vector3f` instance themselves.

As a note, I asked the opinion of @TheCurle regarding the storage of quaternion float values in JSON, if it causes a meaningful loss of precision, and they have told me that there is no such loss. 

Though it would've been neat if quaternions are serialized as degrees of rotation on the three axes, I've been advised that (converting between cartesian and polar coordinates) would cause loss of data/precision, so that's ruled out.

Also, perhaps we should provide overloads for the `Quaternion`-accepting methods which accept three values as degrees/radians of rotation for XYZay,, corresponding to `new Quaternion(x, y, z, isDegrees)`.

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
Rover656: Oops, I completely messed the order there, fixed :)

Rover656:
> I've poked around with this PR by trying to datagenerate the composite model test, and I've hit a pain point while writing up the code: the methods which take in a Vector3f should also have overloads which take in float, float, float, so the caller does not have to manually construct the Vector3f instance themselves.

I have added these overloads as requested

> Also, perhaps we should provide overloads for the Quaternion-accepting methods which accept three values as degrees/radians of rotation for XYZay,, corresponding to new Quaternion(x, y, z, isDegrees).

I agree, I've gone ahead and added overloads matching the Quaternion constructor

sciwhiz12 (approved):

src/main/java/net/minecraftforge/client/model/generators/BlockModelBuilder.java
```diff
@@ -16,9 +25,374 @@
...
+            {
+                transform.add("origin", writeVec3(originVec));
+            }
+
+            return transform;
+        }
+
+        public enum TransformOrigin implements StringRepresentable
+        {
+            CENTER(new Vector3f(.5f, .5f, .5f), "origin"),
```
nikita488, line 336: Why the name of origins was changed? Before the CENTER was "center" and OPPOSING_CORNER was "opposing-corner". Now every model that used old names inside .json files is broken in 43.1.9+

Rover656: These are the constants in code, however the third argument on the line highlighted sets the name that is used during serialisation, therefore this will not break compat. However if you are having issues with this PR, opening an issue with a bit more info might be helpful

Rover656: Ooh I have spotted the underscore issue, I'll mention it in forgecord to see if someone's free to issue a fix

nikita488: If anyone had "origin": "center" before it will parse it, but fail because now "center" is called "origin". And yes, same for underscore issue.

Rover656: Yeah just clocked that too, I'm reading on phone right now and I missed it, was going off of memory. I'll try and get this sorted out ASAP. Thanks for reporting this to me, very sorry

nikita488: All good :)

PR #8864, opened 2022-07-13
Multiple tweaks and fixes to the recent changes in the client refactor PR: Part 3

### LINKED ISSUE #8871: [Bug] Fluid Container Model is translucent since 41.0.64
**Minecraft Version:** 1.19

**Forge Version:** 41.0.64 (and later) (tested up to 41.0.98)

**Steps to Reproduce:**
 1. setup an item that uses the `forge:bucket` aka. `forge:fluid_container` model with `"fluid":"minecraft:water"` (or any other fluid that has a translucent texture)
 2. start the game
 3. get the item into your hotbar
 4. select it to hold it in your hand
 5. see through the bucket^^

**Description of issue:**
Since the breaking change of Forge 41.0.64 the `forge:bucket` aka. `forge:fluid_container` model generates translucent item models.

Using this model for an item.
```
{
  "parent": "forge:item/bucket",
  "loader": "forge:bucket",
  "fluid": "minecraft:water"
}
```

Since Forge 41.0.64 the item looks like that:
bucket_64+

In Forge 41.0.63 the item looks like expected:
bucket_63

Here is an example mod to show the issue: https://github.com/cech12/BucketTestMod (two different branches for 41.0.63 and 41.0.64)

It looks like an old issue, that was fixed in 1.16 with this commit: https://github.com/MinecraftForge/MinecraftForge/commit/ce3d8b40cf37924caf1708cdde6842ae6fdcee31

amadornes:
### Fix weighted baked models not respecting children render types
This fixes an oversight leading to weighted blockstate models always falling back to the default render type instead of using the correct one defined by the model they proxy.

### Revise deprecations in ItemBlockRenderTypes and provide replacement helpers
Documentation improvements in `ItemBlockRenderTypes` to clarify deprecations, and a new helper class with methods that are functionally equivalent to those in IBRT, but designed around the new model-defined render types.

### Allow fluid container model to use base texture as particle
This adds yet another fallback case for no particle texture being specified, making it practically impossible to run into the case where no particle is set and causing vanilla to throw a tantrum over it despite the fact that it's an item and as such doesn't need a particle.

### Fix inverted behavior in composite model building
This inverted if statement was causing composite model building to misbehave, resulting in geometry being output out-of-order and using the incorrect `RenderType`. This fixes #8871.

### DIFF
--- a/patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
@@ -89,20 +89,20 @@
 +      }
 +   }
 +
-+   /** @deprecated Set your render type in your model's JSON or override {@link net.minecraft.client.resources.model.BakedModel#getRenderTypes(BlockState, net.minecraft.util.RandomSource, net.minecraftforge.client.model.data.ModelData)} */
++   /** @deprecated Set your render type in your block model's JSON (eg. "render_type": "cutout") or override {@link net.minecraft.client.resources.model.BakedModel#getRenderTypes(BlockState, net.minecraft.util.RandomSource, net.minecraftforge.client.model.data.ModelData)} */
 +   @Deprecated(forRemoval = true, since = "1.19")
 +   public static void setRenderLayer(Block block, RenderType type) {
 +      com.google.common.base.Preconditions.checkArgument(type.getChunkLayerId() >= 0, "The argument must be a valid chunk render type returned by RenderType#chunkBufferLayers().");
 +      setRenderLayer(block, net.minecraftforge.client.ChunkRenderTypeSet.of(type));
 +   }
 +
-+   /** @deprecated Set your render type in your model's JSON or override {@link net.minecraft.client.resources.model.BakedModel#getRenderTypes(BlockState, net.minecraft.util.RandomSource, net.minecraftforge.client.model.data.ModelData)} */
++   /** @deprecated Set your render type in your block model's JSON (eg. "render_type": "cutout") or override {@link net.minecraft.client.resources.model.BakedModel#getRenderTypes(BlockState, net.minecraft.util.RandomSource, net.minecraftforge.client.model.data.ModelData)} */
 +   @Deprecated(forRemoval = true, since = "1.19")
 +   public static synchronized void setRenderLayer(Block block, java.util.function.Predicate<RenderType> predicate) {
 +      setRenderLayer(block, createSetFromPredicate(predicate));
 +   }
 +
-+   /** @deprecated Set your render type in your model's JSON or override {@link net.minecraft.client.resources.model.BakedModel#getRenderTypes(BlockState, net.minecraft.util.RandomSource, net.minecraftforge.client.model.data.ModelData)} */
++   /** @deprecated Set your render type in your block model's JSON (eg. "render_type": "cutout") or override {@link net.minecraft.client.resources.model.BakedModel#getRenderTypes(BlockState, net.minecraft.util.RandomSource, net.minecraftforge.client.model.data.ModelData)} */
 +   @Deprecated(forRemoval = true, since = "1.19")
 +   public static synchronized void setRenderLayer(Block block, net.minecraftforge.client.ChunkRenderTypeSet layers) {
 +      checkClientLoading();
--- a/patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
+++ b/patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
@@ -89,20 +89,20 @@
 +      }
 +   }
 +
-+   /** @deprecated Set your render type in your block model's JSON (eg. "render_type": "cutout") or override {@link net.minecraft.client.resources.model.BakedModel#getRenderTypes(BlockState, net.minecraft.util.RandomSource, net.minecraftforge.client.model.data.ModelData)} */
++   /** @deprecated Set your render type in your block model's JSON (eg. {@code "render_type": "cutout"}) or override {@link net.minecraft.client.resources.model.BakedModel#getRenderTypes(BlockState, net.minecraft.util.RandomSource, net.minecraftforge.client.model.data.ModelData)} */
 +   @Deprecated(forRemoval = true, since = "1.19")
 +   public static void setRenderLayer(Block block, RenderType type) {
 +      com.google.common.base.Preconditions.checkArgument(type.getChunkLayerId() >= 0, "The argument must be a valid chunk render type returned by RenderType#chunkBufferLayers().");
 +      setRenderLayer(block, net.minecraftforge.client.ChunkRenderTypeSet.of(type));
 +   }
 +
-+   /** @deprecated Set your render type in your block model's JSON (eg. "render_type": "cutout") or override {@link net.minecraft.client.resources.model.BakedModel#getRenderTypes(BlockState, net.minecraft.util.RandomSource, net.minecraftforge.client.model.data.ModelData)} */
++   /** @deprecated Set your render type in your block model's JSON (eg. {@code "render_type": "cutout"}) or override {@link net.minecraft.client.resources.model.BakedModel#getRenderTypes(BlockState, net.minecraft.util.RandomSource, net.minecraftforge.client.model.data.ModelData)} */
 +   @Deprecated(forRemoval = true, since = "1.19")
 +   public static synchronized void setRenderLayer(Block block, java.util.function.Predicate<RenderType> predicate) {
 +      setRenderLayer(block, createSetFromPredicate(predicate));
 +   }
 +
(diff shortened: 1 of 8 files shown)


LexManos:
1) Why are you breaking even more API?
2) Why are you ADDING helper functions as API? Things should be at least marked as internal so modders don't call them. These 'helper' functions are for hooks, hence why they are in ForgeHooksClient

XFactHD:
I would argue that at least some of these helpers should indeed be API and not "locked away" as internal.
The `getEntityRenderType()` helper for example is required when you want to properly render a baked model in a BER.

LexManos:
> I would argue that at least some of these helpers should indeed be API and not "locked away" as internal. The `getEntityRenderType()` helper for example is required when you want to properly render a baked model in a BER.

My point is that one is two lines. Which you can easily do yourself.
I would rather not have simple  `helper` API that we have to support in  perpetuity {or cause Major drama like this entire refactor did} for **simple** things modders can do themselves.

amadornes:
I was discussing this with a few people yesterday. There's definitely a need for some of the methods in `ItemBlockRenderTypes`, especially `getRenderType(BlockState, boolean)` like XFact mentioned. Its stack-based counterpart is necessary for fallback behavior, and users may need to manually call if they extend an existing model and can't simply call `super.getRenderTypes(...)` to access the default implementation.

The methods in `ItemBlockRenderTypes` haven't really been useful since Forge added support for multiple render types per block because it's not possible to properly adapt them. At least `getRenderType(...)` has always been "functional enough", though the patched re-implementation results in some edge cases nobody seems to have ever reported, but that I can guarantee are there just by looking at the code.

If you don't think the helpers should be added, I can leave them as internal hooks and add javadoc noting how users should implement a replacement themselves, but since it's a public vanilla API we're obsoleting, I'd say it makes sense for Forge to provide a replacement.

patches/minecraft/net/minecraft/client/resources/model/WeightedBakedModel.java.patch
```diff
@@ -53,6 +53,13 @@
 +
 +   public BakedModel applyTransform(net.minecraft.client.renderer.block.model.ItemTransforms.TransformType transformType, com.mojang.blaze3d.vertex.PoseStack poseStack, boolean applyLeftHandTransform) {
 +      return this.f_119542_.applyTransform(transformType, poseStack, applyLeftHandTransform);
++   }
++
++   @Override // FORGE: Get render types based on the active weighted model
++   public net.minecraftforge.client.ChunkRenderTypeSet getRenderTypes(@org.jetbrains.annotations.NotNull BlockState state, @org.jetbrains.annotations.NotNull RandomSource rand, @org.jetbrains.annotations.NotNull net.minecraftforge.client.model.data.ModelData data) {
++      return WeightedRandom.m_146314_(this.f_119541_, Math.abs((int)rand.m_188505_()) % this.f_119540_)
```
sciwhiz12, line 60: Would there be problems if the baked model used by `getQuads` differs from the baked model used in this method? I'm thinking of the situation where something calls `getQuads` with a random source, and calls `getRenderTypes` with the same random source afterwards, meaning that both methods have a different random number and therefore can select a different baked model from each other.

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
```diff
@@ -16,32 +16,42 @@
...
        RenderType rendertype = RenderType.m_110466_();
        p_109290_.put(Fluids.f_76192_, rendertype);
-@@ -303,6 +_,7 @@
+@@ -303,6 +_,8 @@
     });
     private static boolean f_109277_;
  
-+   @Deprecated // FORGE: Use canRenderInLayer
++   /** @deprecated Forge: Use {@link net.minecraft.client.resources.model.BakedModel#getRenderTypes(BlockState, net.minecraft.util.RandomSource, net.minecraftforge.client.model.data.ModelData)}. */
++   @Deprecated // Note: this method does NOT support model-based render types
```
sciwhiz12, line 24: What are model-based render types?

src/main/java/net/minecraftforge/client/RenderTypeHelper.java
```diff
@@ -0,0 +1,71 @@
+package net.minecraftforge.client;
```
sciwhiz12, line 6: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -953,20 +953,6 @@ public static String onClientSendMessage(String message)
         return MinecraftForge.EVENT_BUS.post(event) ? "" : event.getMessage();
     }
 
-    /**
-     * Mimics the behavior of {@link net.minecraft.client.renderer.ItemBlockRenderTypes#getRenderType(BlockState, boolean)}
-     * for the input {@link RenderType}.
-     */
-    @NotNull
-    public static RenderType getEntityRenderType(RenderType chunkRenderType, boolean fabulous)
```
sciwhiz12, line 961: Should we avoid this breaking change by reinstanting this method, deprecating this method for removal in the next major version/breaking changes window, and redirecting it to `RenderTypeHelper#getEntityRenderType`?

src/main/java/net/minecraftforge/client/RenderTypeHelper.java
```diff
@@ -0,0 +1,71 @@
...
+    public static RenderType getMovingBlockRenderType(RenderType renderType)
+    {
+        if (renderType == RenderType.translucent())
+            return RenderType.translucentMovingBlock();
+        return renderType;
+    }
+
+    /**
+     * Provides a fallback {@link RenderType} for the given {@link ItemStack} in the case that none is explicitly specified.
+     * <p/>
```
sciwhiz12: Javadocs formatting: self-closing paragraph tags are invalid.

sciwhiz12 (changes requested):

patches/minecraft/net/minecraft/client/resources/model/WeightedBakedModel.java.patch
amadornes: While a fair concern, it would be a breach of the standard model contract, where the seed is expected to be reset to a known value before each method call, so it's a non-problem.

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
amadornes: Render types controlled by `BakedModel` instances (the new approach, instead of being managed by `ItemBlockRenderTypes`). The reason it's not supported is because there's no way to faithfully adapt the methods to support multiple render types without partially breaking an implicit contract or causing inconsistent rendering. One of them was adapted in the past, but it had edge cases that would lead to artifacts, as I mentioned in my response to Lex.

src/main/java/net/minecraftforge/client/RenderTypeHelper.java
amadornes: Oh, come on... I did it again >_>

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
amadornes: I wouldn't say it's a breaking change since it's in `ForgeHooksClient` and the whole class is marked as internal, but you can have the final say on it.

marchermans (approved):
@LexManos If this is good to go from your side, I am okey with it. Systemswise this is fine.

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
```diff
@@ -79,20 +89,20 @@
 +      }
 +   }
 +
-+   /** @deprecated Set your render type in your model's JSON or override {@link net.minecraft.client.resources.model.BakedModel#getRenderTypes(BlockState, net.minecraft.util.RandomSource, net.minecraftforge.client.model.data.ModelData)} */
++   /** @deprecated Set your render type in your block model's JSON (eg. "render_type": "cutout") or override {@link net.minecraft.client.resources.model.BakedModel#getRenderTypes(BlockState, net.minecraft.util.RandomSource, net.minecraftforge.client.model.data.ModelData)} */
```
sciwhiz12: As this is a javadoc comment, `{@code}` should enclose the JSON example. (Same for the other two instances.)
```suggestion
+   /** @deprecated Set your render type in your block model's JSON (eg. {@code "render_type": "cutout"}) or override {@link net.minecraft.client.resources.model.BakedModel#getRenderTypes(BlockState, net.minecraft.util.RandomSource, net.minecraftforge.client.model.data.ModelData)} */
```

sciwhiz12 (changes requested):

patches/minecraft/net/minecraft/client/renderer/ItemBlockRenderTypes.java.patch
amadornes: Good point, I didn't consider that.

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
LexManos: Keep and forward.

PR #8126, opened 2021-09-28
Add back the attacking entity to LivingKnockBackEvent 

Sarinsa:
Add back the attacking entity to LivingKnockBackEvent and add nulllable entity parameter for attacking entity to LivingEntity#knockback

Issue #5994, opened 2019-07-31
Add the DamageSource as a member of LivingKnockBackEvent and change its call position


RaycusMX:
Cannot get the immediate damage source from LivingKnockBackEvent currently. Only the true source. So I'd like to have the DamageSource object as a member of LivingKnockBackEvent. In this case the event should be called before the invoking of **knockback** method in **attackEntityFrom** of EntityLivingBase.

This would allow modders to determine whether to knockback or not according to the type of the projectile.

diesieben07:
Like I already told you on discord: `attackEntityFrom` is not the only place `knockBack` is called from. `knockBack` is even public, any random mod can call it at any time.

RaycusMX:
> Like I already told you on discord: `attackEntityFrom` is not the only place `knockBack` is called from. `knockBack` is even public, any random mod can call it at any time.

ok. i didn't understand your meaning. sorry about that.

PR #8865, opened 2022-07-13
[1.19.x] Add more context to LivingKnockBackEvent

Elenterius:
Improves the LivingKnockBackEvent by providing more information to the event in the form of a context object.

### Knockback Context
- contains the **`Cause`** of the knockback 
- may contain the attacker `Entity` that is associated with the knockback (`@Nullable Entity`)
- may contain the `DamageSource` that is associated with the knockback (`@Nullable DamageSource`)
- may contain the attack `damage` that is associated with the knockback (`@Nullable Float`)
- may contain the `ItemStack` that is associated with the knockback (if n/a will return `ItemStack.EMPTY`)

### Knockback  Causes
Knockback  Causes are explicitly expressed as an Enum to allow easier identification what may have caused it. This allows us for example to create a weapon that causes no knockbacks. (setting only the AttackKnockback Attribute to zero would not work due to LivingEntity#hurt(DamageSource, float) causing a knockback)
```java
/**
 * knockback cause is not known or the deprecated ForgeHook was called
 */
UNKNOWN,
/**
 * knockback is caused externally by blocking an attack from an entity
 * @see LivingEntity#blockedByShield(LivingEntity)
 */
SHIELD_BLOCKED_ATTACK_FROM_ENTITY,
/**
 * knockback is caused internally by being damaged
 * @see LivingEntity#hurt(DamageSource, float)
 */
HURT_BY_DAMAGE_SOURCE,
/**
 * knockback is caused externally by an entity attack
 * @see Mob#doHurtTarget(Entity)
 * @see Player#attack(Entity)
 */
ATTACKED_BY_ENTITY,
/**
 * knockback is caused externally by an entity sweep attack
 * @see Player#attack(Entity)
 */
SWEEP_ATTACKED_BY_ENTITY,
/**
 * knockback is caused externally by an entity ramming
 * @see RamTarget#tick(ServerLevel, Goat, long)
 */
RAMMED_BY_ENTITY
```

### Related
- #8126
- #5994


Fuzss:
Could you maybe add an interface to the `KnockbackCause` enum and use it instead of the class? Having this predefined list of 6 causes seems a little limiting to me and giving modders the ability to create their own causes seems like a good idea.

PR #8867, opened 2022-07-14
[1.19.x] fix inability to run 'forge forge test client' and fix inability to load/create worlds

Elenterius:
- removes missing mods from the mod toml file (overlay_layers_test and structure_spawn_list_event_test)
- disables the DataPackRegistriesTest which prevents the creation of worlds
- fixes failing datapack validation (missing JSON keys) caused by ServerWorldCreationTest which prevents the loading of worlds

### DIFF
--- a/src/test/resources/META-INF/mods.toml
+++ b/src/test/resources/META-INF/mods.toml
@@ -117,8 +117,6 @@ modId="living_conversion_event_test"
 [[mods]]
 modId="player_game_mode_event_test"
 [[mods]]
-modId="structure_spawn_list_event_test"
-[[mods]]
 modId="forge_codecs_test"
 [[mods]]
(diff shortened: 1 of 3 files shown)


TheCurle:
the datapack registries test can be fixed by adding the missing fields to the JSONs - disabling it is a poor step

Elenterius:
> the datapack registries test can be fixed by adding the missing fields to the JSONs - disabling it is a poor step

I'm aware that disabling it was a poor step but it was the quickest solution when I was trying to get the world to load.

Anyways i dropped the commit, turns out the DataPackRegistriesTest was not broken after all.
Somehow the error log and behavior of the game made me believe it was one of the issues that prevented world loading 😅

PR #8868, opened 2022-07-14
[1.18.x] Cache resource listing calls in resource packs

embeddedt:
This is a backport of #8829 to the 1.18.x branch. I have tried to keep the patch style as identical as possible.

Note: As 1.18 uses a `String` predicate for filtering, while 1.19 uses a `ResourceLocation`, I had to adjust the filtering logic in the cache manager [here](https://github.com/embeddedt/MinecraftForge/blob/93a0a2b4cc0bcf2a28554f204c4b07afeeecb094/src/main/java/net/minecraftforge/resource/ResourceCacheManager.java#L392-L395). I would appreciate if someone can confirm that the logic here is correct.

### DIFF
--- a/patches/minecraft/net/minecraft/server/packs/resources/MultiPackResourceManager.java.patch
+++ b/patches/minecraft/net/minecraft/server/packs/resources/MultiPackResourceManager.java.patch
@@ -1,10 +1,9 @@
 --- a/net/minecraft/server/packs/resources/MultiPackResourceManager.java
 +++ b/net/minecraft/server/packs/resources/MultiPackResourceManager.java
-@@ -23,7 +_,7 @@
-    public MultiPackResourceManager(PackType p_203797_, List<PackResources> p_203798_) {
+@@ -24,6 +_,7 @@
        this.f_203795_ = List.copyOf(p_203798_);
        Map<String, FallbackResourceManager> map = new HashMap<>();
--
+ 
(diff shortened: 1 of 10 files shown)


patches/minecraft/net/minecraft/server/packs/resources/MultiPackResourceManager.java.patch
```diff
@@ -0,0 +1,11 @@
+--- a/net/minecraft/server/packs/resources/MultiPackResourceManager.java
++++ b/net/minecraft/server/packs/resources/MultiPackResourceManager.java
+@@ -23,7 +_,7 @@
+    public MultiPackResourceManager(PackType p_203797_, List<PackResources> p_203798_) {
+       this.f_203795_ = List.copyOf(p_203798_);
+       Map<String, FallbackResourceManager> map = new HashMap<>();
+-
```
sciwhiz12: This blank line shouldn't be removed.

sciwhiz12 (changes requested):

sciwhiz12 (approved):
Looks fine to me, being very similar to the 1.19 PR barring some changes for the `Predicate<ResourceLocation>` being `Predicate<String>` here. @OrionDevelopment, if you have the time, could you please review this PR to see if nothing else is needed to change to account for the changes between 1.18 and 1.19, thanks.

PR #8872, opened 2022-07-14
[1.19] Add hook for items to remain in the hotbar when picking blocks/entities

### LINKED ISSUE #8869: Issue in Tinkers Construct with Pick Block item swapping
Minecraft Version
1.18.2

Forge Version
40.1.60

Mantle Version
1.9.27

Tinkers' Construct Version
3.5.1.31

Describe your issue
When right clicking on a block Minecraft will swap your currently selected item to that block, if it is in your inventory. If it is not on your hotbar, and your hotbar is full, it will swap that item from your inventory with your current hotbar slot, unless you are currently holding a vanilla enchanted tool. If you have an enchanted tool it will swap with the next non-tool slot on your hotbar.
There is currently no consideration for this if you have a Tinker's tool as your current item. If you have your hotbar full, and have a Tinkers tool in hand, and use Pick Block (middle click by default), the block will replace your Tinkers tool: throwing off what may be your typical tool sorting order.

Steps to Reproduce:

1. Have a Tinkers tool in slot 1
2. Fill your other hotbar slots with other blocks
3. Have your tool in slot 1 currently active
4. Use the Pick Block command (middle-click by default) while looking at a block that is not on your hotbar, but is in your inventory

Expected Result:
The type of item you are looking at when using Pick Block will be swapped from your inventory with your item in the first non-tool slot (slot 2). This is how vanilla works with enchanted tools.
Actual Result:
The type of item you are looking at when using Pick Block swaps from your inventory with your tool, putting it out of your hotbar.

This is a copy of a ticket from the Tinkers Construct GitHub, where the it was determined that it was a Forge issue.
https://github.com/SlimeKnights/TinkersConstruct/issues/4939

diesieben07:
When picking an item using the middle mouse button, the game tries to keep enchanted items in the hotbar, if the hotbar is already full.
This adds a hook for items to customize this behavior and allow themselves to be excluded, even if they are not enchanted.

Closes #8869.
See also: https://github.com/SlimeKnights/TinkersConstruct/issues/4939

### DIFF
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
@@ -830,7 +830,7 @@ public interface IForgeItem
     }
 
     /**
-     * Whether this stack should be excluded (if possible) when selecting the target hotbar slot of a "pick" action.
+     * Whether the given ItemStack should be excluded (if possible) when selecting the target hotbar slot of a "pick" action.
      * By default, this returns true for enchanted stacks.
      *
      * @see Inventory#getSuitableHotbarSlot()
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
@@ -828,4 +828,9 @@ public interface IForgeItem
         return self().getFoodProperties();
     }
 
+    default boolean shouldRemainInHotbarUponPick(ItemStack stack, Player player, int inventorySlot)
+    {
+        return stack.isEnchanted();
+    }
+
 }
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
@@ -838,7 +838,7 @@ public interface IForgeItem
      * @param inventorySlot the inventory slot of the item being up for replacement
      * @return true to leave this stack in the hotbar if possible
      */
-    default boolean isStickyInHotbar(ItemStack stack, Player player, int inventorySlot)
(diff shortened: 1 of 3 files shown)


src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -828,4 +829,18 @@ default FoodProperties getFoodProperties(ItemStack stack, @Nullable LivingEntity
...
+    /**
+     * Whether the given ItemStack should be excluded (if possible) when selecting the target hotbar slot of a "pick" action.
+     * By default, this returns true for enchanted stacks.
+     *
+     * @see Inventory#getSuitableHotbarSlot()
+     * @param player the player performing the picking
+     * @param inventorySlot the inventory slot of the item being up for replacement
+     * @return true to leave this stack in the hotbar if possible
+     */
+    default boolean isStickyInHotbar(ItemStack stack, Player player, int inventorySlot)
```
pupnewfster: This feels like a terrible name, not sure if this is any better but part of me feels like `isLowPriorityHotbarSlot` or something might be at least slightly better? (Still not quite sure if that is a decent name either though)

diesieben07: Yeah, I don't like the name either. Your suggestion sounds better, but I'll wait if someone else has an even better idea.

justliliandev: I suggest isReplaceableByPick or isReplaceableByPickAction (with the return value being inverted) or the negation in its name like isNotReplaceablByPick

diesieben07:
Renamed method to `isNotReplaceableByPickAction`. Still not ideal, but no better names have been suggested so far.

sciwhiz12 (approved):

PR #8875, opened 2022-07-15
Fix packet handling on play messages.

### LINKED ISSUE #8873: [Bug] SpawnEntity packet additional data is not accessible in custom client factory
**Minecraft Version:** 1.19

**Forge Version:** 41.0.98

**Logs:** 
Stack Trace: https://gist.github.com/jglrxavpok/68fa6e4488e9ea7a0fcee842428f717f
latest.log: https://gist.github.com/jglrxavpok/c2f4080d922fd52894804099430d6be3

**Steps to Reproduce:**
1. Setup an Entity implementing IEntityAdditionalSpawnData
2. When spawning the entity, read from the buffer passed as input of `IEntityAdditionalSpawnData#readSpawnData(FriendlyByteBuf)`
3. An exception is thrown because the buffer is no longer accessible.

**Description of issue:**
Additional spawn data cannot be read from the SpawnEntity packet due to the internal buffer not being accessible.
This does not prevent entities to spawn, but no data will be loaded into the entity.

I believe the networking changes of PR #8703 require the buffer to be kept alive until it is used in the enqueueWork callback.

My readSpawnData method for reference:
```kotlin
    override fun readSpawnData(additionalData: FriendlyByteBuf) {
        try {
            val data = additionalData.readNbt()
            readAdditionalSaveData(data!!)
        } catch (t: Throwable) {
            MoarBoats.logger.error("Error in readSpawnData", t)
        }
    }
```

marchermans:
This fixes forges own play messages trying to read data from a FriendlyByteBuf in an enqueueWork section after the code-flow execution of the network thread (which owns and frees said FriendlyByteBuf) has already been completed.

Fixes: #8873

### DIFF
--- a/src/main/java/net/minecraftforge/network/PlayMessages.java
+++ b/src/main/java/net/minecraftforge/network/PlayMessages.java
@@ -5,6 +5,7 @@
 
 package net.minecraftforge.network;
 
+import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.gui.screens.inventory.MenuAccess;
@@ -26,6 +27,7 @@ import net.minecraft.network.chat.Component;
 import net.minecraftforge.entity.IEntityAdditionalSpawnData;
 import net.minecraftforge.network.NetworkEvent;
 
+import java.nio.ByteBuffer;
 import java.util.Locale;
 import java.util.Optional;
 import java.util.UUID;
@@ -109,7 +111,17 @@ public class PlayMessages
             buf.writeShort(msg.velZ);
             if (msg.entity instanceof IEntityAdditionalSpawnData)
             {
-                ((IEntityAdditionalSpawnData) msg.entity).writeSpawnData(buf);
+                final FriendlyByteBuf spawnDataBuffer = new FriendlyByteBuf(Unpooled.buffer());
+
+                ((IEntityAdditionalSpawnData) msg.entity).writeSpawnData(spawnDataBuffer);
+
+                buf.writeVarInt(spawnDataBuffer.readableBytes());
+                buf.writeBytes(spawnDataBuffer);
+
+                spawnDataBuffer.release();
+            }
+            else {
+                buf.writeVarInt(0);
             }
         }
 
@@ -122,8 +134,19 @@ public class PlayMessages
                     buf.readDouble(), buf.readDouble(), buf.readDouble(),
                     buf.readByte(), buf.readByte(), buf.readByte(),
                     buf.readShort(), buf.readShort(), buf.readShort(),
-                    buf
+                    readSpawnDataPacket(buf)
                     );
+        }
+
+        private static FriendlyByteBuf readSpawnDataPacket(FriendlyByteBuf buf) {
+            final int count = buf.readVarInt();
+            if (count > 0) {
+                final FriendlyByteBuf spawnDataBuffer = new FriendlyByteBuf(Unpooled.buffer());
+                spawnDataBuffer.writeBytes(buf, count);
+                return spawnDataBuffer;
+            }
+
+            return new FriendlyByteBuf(Unpooled.buffer());
         }
 
         public static void handle(SpawnEntity msg, Supplier<NetworkEvent.Context> ctx)
@@ -132,6 +155,7 @@ public class PlayMessages
                 EntityType<?> type = Registry.ENTITY_TYPE.byId(msg.typeId);
                 if (type == null)
                 {
+                    msg.buf.release();
                     throw new RuntimeException(String.format(Locale.ENGLISH, "Could not spawn entity (id %d) with unknown type at (%f, %f, %f)", msg.entityId, msg.posX, msg.posY, msg.posZ));
                 }
 
@@ -139,6 +163,7 @@ public class PlayMessages
                 Entity e = world.map(w->type.customClientSpawn(msg, w)).orElse(null);
                 if (e == null)
                 {
+                    msg.buf.release();
                     return;
                 }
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/minecraftforge/network/PlayMessages.java
```diff
@@ -109,7 +111,17 @@ public static void encode(SpawnEntity msg, FriendlyByteBuf buf)
             buf.writeShort(msg.velZ);
             if (msg.entity instanceof IEntityAdditionalSpawnData)
             {
-                ((IEntityAdditionalSpawnData) msg.entity).writeSpawnData(buf);
+                final FriendlyByteBuf spawnDataBuffer = new FriendlyByteBuf(Unpooled.buffer());
+
+                ((IEntityAdditionalSpawnData) msg.entity).writeSpawnData(spawnDataBuffer);
```
sciwhiz12: Since we're already modifying this code, this should be using pattern-matching instanceof.

src/main/java/net/minecraftforge/network/PlayMessages.java
```diff
@@ -109,7 +111,17 @@ public static void encode(SpawnEntity msg, FriendlyByteBuf buf)
...
+
+                buf.writeVarInt(spawnDataBuffer.readableBytes());
+                buf.writeBytes(spawnDataBuffer);
+
+                spawnDataBuffer.release();
+            }
+            else {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
            else
            {
```

src/main/java/net/minecraftforge/network/PlayMessages.java
```diff
@@ -122,23 +134,36 @@ public static SpawnEntity decode(FriendlyByteBuf buf)
...
                     buf.readByte(), buf.readByte(), buf.readByte(),
                     buf.readShort(), buf.readShort(), buf.readShort(),
-                    buf
+                    readSpawnDataPacket(buf)
                     );
         }
 
+        private static FriendlyByteBuf readSpawnDataPacket(FriendlyByteBuf buf) {
+            final int count = buf.readVarInt();
+            if (count > 0) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
        private static FriendlyByteBuf readSpawnDataPacket(FriendlyByteBuf buf)
        {
            final int count = buf.readVarInt();
            if (count > 0)
            {
```

src/main/java/net/minecraftforge/network/PlayMessages.java
```diff
@@ -122,23 +134,36 @@ public static SpawnEntity decode(FriendlyByteBuf buf)
...
+            return new FriendlyByteBuf(Unpooled.buffer());
+        }
+
         public static void handle(SpawnEntity msg, Supplier<NetworkEvent.Context> ctx)
         {
             ctx.get().enqueueWork(() -> {
                 EntityType<?> type = Registry.ENTITY_TYPE.byId(msg.typeId);
                 if (type == null)
                 {
+                    msg.buf.release();
```
sciwhiz12: Perhaps a `try`-`finally` would be better here to ensure the buffer's release, instead of inserting multiple release calls?

src/main/java/net/minecraftforge/network/PlayMessages.java
```diff
@@ -159,6 +184,7 @@ public static void handle(SpawnEntity msg, Supplier<NetworkEvent.Context> ctx)
                 {
                     ((IEntityAdditionalSpawnData) e).readSpawnData(msg.buf);
```
sciwhiz12: Please also convert this into using pattern-matching `instanceof`.

src/main/java/net/minecraftforge/network/PlayMessages.java
```diff
@@ -273,6 +299,7 @@ public static void handle(OpenContainer msg, Supplier<NetworkEvent.Context> ctx)
                 MenuScreens.getScreenFactory(msg.getType(), Minecraft.getInstance(), msg.getWindowId(), msg.getName())
                              .ifPresent(f -> {
                                  AbstractContainerMenu c = msg.getType().create(msg.getWindowId(), Minecraft.getInstance().player.getInventory(), msg.getAdditionalData());
+                                 msg.getAdditionalData().release();
```
sciwhiz12: This is also a candidate for using a `try`-`finally` to ensure the buffer is released.

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/network/PlayMessages.java
```diff
@@ -122,42 +132,67 @@ public static SpawnEntity decode(FriendlyByteBuf buf)
...
-                e.setYBodyRot((msg.headYaw * 360) / 256.0F);
-
-                e.setId(msg.entityId);
-                e.setUUID(msg.uuid);
-                world.filter(ClientLevel.class::isInstance).ifPresent(w->((ClientLevel)w).putNonPlayerEntity(msg.entityId, e));
-                e.lerpMotion(msg.velX / 8000.0, msg.velY / 8000.0, msg.velZ / 8000.0);
-                if (e instanceof IEntityAdditionalSpawnData)
+        public static void handle(SpawnEntity msg, Supplier<NetworkEvent.Context> ctx)
+        {
+            ctx.get().enqueueWork(() ->
```
TheCurle: I beg you to make this function easier to read.

I legitimately cannot parse the control flow or order of parameters being used here.

Please make things one giant line if it benefits, this ridiculous line splitting is horrendous for the rest of us.

src/main/java/net/minecraftforge/network/PlayMessages.java
```diff
@@ -122,42 +132,67 @@ public static SpawnEntity decode(FriendlyByteBuf buf)
...
+            {
+                try {
+                    EntityType<?> type = Registry.ENTITY_TYPE.byId(msg.typeId);
+                    if (type == null)
```
TheCurle: This is a useless check, type cannot be null.

sciwhiz12 (approved):

PR #8880, opened 2022-07-16
Allow registering custom `ColorResolver`s

diesieben07:
`ColorResolver` is used to create biome-based block coloring. The `BlockAndTintGetter`  given to `BlockColor` cannot directly be used to obtain the biome, it must be done through `ColorResolver`, which will allow the game to implement caching and also apply the selected biome blend. Any ColorResolver instance must be registered for this cache though. This PR adds an event to allow mods to register their own.


src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -318,6 +321,23 @@ public static void onItemColorsInit(ItemColors itemColors, BlockColors blockColo
         ModLoader.get().postEvent(new RegisterColorHandlersEvent.Item(itemColors, blockColors));
     }
 
+    private static ImmutableList<ColorResolver> colorResolvers;
```
marchermans: Please move this into its own management class.

marchermans (changes requested):

marchermans (approved):

src/main/java/net/minecraftforge/client/ColorResolverManager.java
```diff
@@ -0,0 +1,51 @@
...
+        colorResolvers = builder.build();
+    }
+
+    /**
+     * Register a {@link BlockTintCache} for every registered {@link ColorResolver} into the given target map.
+     *
+     * @param level the level to use
+     * @param target the map to populate
+     */
+    public static void registerBlockTintCaches(ClientLevel level, Map<ColorResolver, BlockTintCache> target)
```
sciwhiz12, line 40: Should this be marked as `@ApiStatus.Internal`? Is there a reason for a mod to call this method?

src/test/java/net/minecraftforge/debug/client/CustomColorResolverTest.java
```diff
@@ -0,0 +1,60 @@
...
+import net.minecraftforge.eventbus.api.IEventBus;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.RegistryObject;
+
+@Mod(CustomColorResolverTest.MOD_ID)
+public class CustomColorResolverTest
```
sciwhiz12, line 31: Please provide a javadoc for this class which explains what feature this tests, how to test it, and what the expected results are from said test.

This should also be added to the test sourceset's `mods.toml` for now.

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/client/ColorResolverManager.java
diesieben07: My reason for not marking it as internal was for mods that do custom rendering (maybe even something like Optifine) needing to call it.

diesieben07:
Test mod without biome blend

And with biome blend

sciwhiz12 (approved):

diesieben07:
Updated to 1.19.2

sciwhiz12 (approved):

PR #8881, opened 2022-07-17
Add User-Agent header to requests made by the update checker

triphora:
Currently the update checker sends requests with the default Java HTTP client `User-Agent` header - `Java-http-client/<Java version>`. This is rather unhelpful for a couple reasons, principally that it's so generic that it kind of goes against the purpose of a user agent.

This PR changes the user agent header to make it clear that it's a request being made by Forge.

Example: `MinecraftForge/1.19-41.0.99 Java/17.0.1`

### DIFF
--- a/fmlcore/src/main/java/net/minecraftforge/fml/VersionChecker.java
+++ b/fmlcore/src/main/java/net/minecraftforge/fml/VersionChecker.java
@@ -119,7 +119,7 @@ public class VersionChecker
              */
             private String openUrlString(URL url) throws IOException, URISyntaxException, InterruptedException {
                 URL currentUrl = url;
-                final String userAgent = String.format("MinecraftForge/%s Java/%s", FMLLoader.versionInfo().mcAndForgeVersion(), System.getProperty("java.version"));
+                final String userAgent = String.format("Java-http-client/%s MinecraftForge/%s", System.getProperty("java.version"), FMLLoader.versionInfo().mcAndForgeVersion());
                 for (int redirects = 0; redirects < MAX_HTTP_REDIRECTS; redirects++)
                 {
                     var request = HttpRequest.newBuilder()
--- a/fmlcore/src/main/java/net/minecraftforge/fml/VersionChecker.java
+++ b/fmlcore/src/main/java/net/minecraftforge/fml/VersionChecker.java
@@ -119,9 +119,9 @@ public class VersionChecker
              */
             private String openUrlString(URL url) throws IOException, URISyntaxException, InterruptedException {
                 URL currentUrl = url;
+                final String userAgent = String.format("MinecraftForge/%s Java/%s", FMLLoader.versionInfo().mcAndForgeVersion(), System.getProperty("java.version"));
                 for (int redirects = 0; redirects < MAX_HTTP_REDIRECTS; redirects++)
                 {
-                    final String userAgent = String.format("MinecraftForge/%s Java/%s", FMLLoader.versionInfo().mcAndForgeVersion(), System.getProperty("java.version"));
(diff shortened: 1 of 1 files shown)


PaintNinja:
I suggest moving the userAgent variable outside of the for loop - there’s no need to recalculate the string for each Mod.

PaintNinja:
Should the `Java/<java version>` in this PR be replaced with `Java-http-client/<Java version>` (or whatever the *exact* case-sensitive UA was before) to ensure compat with existing web servers that might expect the old UA?

LordRalex:
I would argue changing it anyways is breaking, and that it would need to start with the Java-http-client to avoid breaking compat. Moving it to the middle would mean it's not at the start, and so starts with checks (I'd suspect that's the most common method?) would then not work either.

But I'd also probably argue user agents always change, and anyone who's hardcoded a check probably should not be anyways. But that's my opinion.

PaintNinja:
As far as I'm aware it's usually a contains check rather than a starts with, but either way can be solved by changing the format to `Java-http-client/<Java version> MinecraftForge/<mc and forge version>`, assuming  `Java-http-client/<Java version>` was the exact case-sensitive format used before this PR.

triphora:
> assuming `Java-http-client/<Java version>` was the exact case-sensitive format used before this PR.

Yes, here is the JDK source for that: https://github.com/openjdk/jdk/blob/441c33f0b1e970c82d29a67162e6dceed0fbf44a/src/java.net.http/share/classes/jdk/internal/net/http/HttpRequestImpl.java#L69-L77

Will change it to `Java-http-client/<Java version> MinecraftForge/<mc and forge version>`.

LexManos:
This would actually be kinda useful to do server side statistics gathering.
However, is there any value in also including thee modid/version for the  specific thing being requested, so that mods can also do their own analysis?
So something like: `Java-http-client/<Java version> MinecraftForge/<mc and forge version> <ModId>/<ModVersion>`

fmlcore/src/main/java/net/minecraftforge/fml/VersionChecker.java
```diff
@@ -117,14 +117,22 @@ public void run()
             /**
              * Returns the response body as a String for the given URL while following redirects
              */
-            private String openUrlString(URL url) throws IOException, URISyntaxException, InterruptedException {
+            private String openUrlString(URL url, IModInfo mod) throws IOException, URISyntaxException, InterruptedException {
                 URL currentUrl = url;
+                final String userAgent = String.format(
```
LexManos: StringBuilder not Format

PaintNinja:
Please move the `userAgent.toString()` call outside of the for loop to avoid rebuilding the string on each redirect.

PaintNinja (approved):

sciwhiz12 (approved):

LordRalex:
As a note, Java 8 used `Java/<version>` as it's User-Agent, and changed it to Java-http-client in a later release (Java 10). As such, technically it was already a breaking change to the agent that would have never been noticed. I just now confirmed this by looking at my stats for a service which has the agents, and noticed this. I also confirmed it through more research.

PaintNinja:
We didn't use HTTP Client until MC 1.17.x, which required Java 16 to run (rather than Java 8 in MC 1.16). The switch from UrlHttpConnection to HttpClient was done during the breaking changes window of 1.17.1 (Forge version 37.0.123).

LexManos:
Just o be clear, this is not a exposed/supported API or behavior. This is not part of our 'breaking changes' rules.
This is just a potentially useful feature 3rd parties running web services.
I may/May not change in the future. We make no promises.

PR #8882, opened 2022-07-17
[1.19] Added LivingGravityEvent

MrScautHD:
This PR adding the `LivingGravityEvent` that allow you to change the gravity if the `gravity attribute` not set to `0.08`, (default) that make it easyer to do custom gravity and the player can still change it with the Attribute command.

`TestEvent` is just here that you can check how it works and if this PR get accepted i want to remove it.


TheCurle:
Events are not designed to be fired in anticipation of something; events and handlers are reactionary, and are designed to handle when an event has ALREADY HAPPENED. 

Firing this event, for every entity, on every tick, for no reason, is both poor on performance and poor on API quality.

We'd now have FOUR hooks that do the exact same thing, none of them being "more important" or "easier" than the others, and each with little documentation. 

It'd be confusing for modders, confusing for us, more annoying for us to maintain (since now we have to keep this AND the attribute in line) and worse overall for the quality of Forge's APIs.

Learn how to use attributes instead of doing shit like this.

PR #8884, opened 2022-07-17
Update Gradle to 7.5 (JDK 18 Support)

JT122406:


TheCurle:
We're working on this internally, including reworking ForgeGradle to account for the new features. 

Thanks for the thought, though :P

PR #8890, opened 2022-07-18
Allow faces of an "elements" model to be made emissive

amadornes:
Emissive textures have been a staple of modded Minecraft for some time now, but while we've had the tools to make it happen in code, as well as in OBJ and layered item models, there hasn't been a good way to do it for regular "elements" models.

This PR adds the necessary hooks to `BlockElementFace` and `FaceBakery`, as well as datagen support and a test mod (see image).

Emissivity works natively on both blocks and items that flag their faces as such and does NOT require a special `RenderType`.

image


marchermans (approved):

KnightMiner:
Any reason emissive is a boolean instead of a number 0-15? I have an implementation of this using a custom block model/face bakery which supports a full 0-15 range, allowing you to essentially set the min light level for the block. In other words, if I set to 7, it will show as light level 7 at 0-7, and at the light level from 8-15.

amadornes:
No real reason, no. I may as well do that.

TeamSpen210:
It doesn't look like there's any check on the range of the emissivity value, so users can set it to illegal values, then get an obtuse array-index exception.  Maybe there should be some explicit exceptions raised?

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
```diff
@@ -557,6 +561,15 @@ private static Vector3f getVertexPos(int[] data, int vertex)
         return new Vector3f(x, y, z);
     }
 
+    private static final IQuadTransformer[] EMISSIVE_TRANSFORMERS = Util.make(new IQuadTransformer[16], array -> {
```
marchermans: In the previous commit this was in IQuadTransformer, I would rather see that there again, if at all possible.

Moving it here changes it to an internal API which means that modders need to recreate this trivial logic themselves. It is better to support this natively.

src/main/java/net/minecraftforge/client/model/ItemLayerModel.java
```diff
@@ -95,7 +98,7 @@ public BakedModel bake(IGeometryBakingContext context, ModelBakery bakery, Funct
             TextureAtlasSprite sprite = spriteGetter.apply(textures.get(i));
             var unbaked = UnbakedGeometryHelper.createUnbakedItemElements(i, sprite);
             var quads = UnbakedGeometryHelper.bakeElements(unbaked, $ -> sprite, modelState, modelLocation);
-            if (emissiveLayers.contains(i)) IQuadTransformer.EMISSIVE.processInPlace(quads);
+            if (emissiveLayers.contains(i)) MAX_LIGHTMAP_TRANSFORMER.processInPlace(quads);
```
marchermans: Is this correct?
The emissive layers are emissive if their light level emission strength is not 0.... That should mean we should not just by default inject the max light map, no?

marchermans (changes requested):

src/main/java/net/minecraftforge/client/model/ItemLayerModel.java
amadornes: Yep, it's good. This is for layered items, where it's always been a choice between no emissiveness and max emissiveness (hence why it was originally also a boolean in my PR).

If you check the overall changed files, you'll see this is how it was already implemented. I just reverted it when I took the field out of `IQuadTransformer`. It's been tested and you can see it in action in the mega model test that was part of my original PR.

src/main/java/net/minecraftforge/client/ForgeHooksClient.java
amadornes: As discussed on Discord, I'll put it in a separate helper class to avoid having public fields on the interface.

marchermans (approved):

mcenderdragon:
@amadornes is there a chance we get this also on 1.18 ?

PR #8891, opened 2022-07-18
[1.19.x] Add attack strength field for retaining strength value.

bconlon1:
Different implementation of #8838. Adds a field to LivingEntity that stores the player's attack strength for longer so that it's accessible from more attack-related events instead of getting immediately reset and requiring storing the value from an earlier event to access. This negates the necessity of that workaround and allows the strength to be more accessible from related item methods.


KnightMiner:
The documentation does not explain how this new feature different from the attribute, nor when it is updated so the player can know its up to date. My suspicion is its only valid immediately after `attack` is called and it contains info beyond the attribute, but that may be wrong, and it *should* be in the docs.

bconlon1:
> The documentation does not explain how this new feature different from the attribute, nor when it is updated so the player can know its up to date. My suspicion is its only valid immediately after `attack` is called and it contains info beyond the attribute, but that may be wrong, and it _should_ be in the docs.

Would you be able to explain which attribute you're referring to? Do you mean the `ATTACK_DAMAGE` attribute? Either way though this is separate from that, as this is tied to vanilla's attack strength system (the indicator for whether your weapon swing is on cooldown, which determines whether you do full damage); it is still used similarly to the attribute in that it is used for determining the attack damage, but it's just another factor as can be seen in the `attack` method's code. 

It is updated right after the value for it is set within the `attack` code as seen with the field being set, as well as at the end of the `attack` code so that it is reset properly, this still leaves the value valid during all attack-related events (except for LivingAttackEvent from the Player which occurs before the value is set, but you can still get it in that context from directly calling `getAttackStrengthScale`), which wasn't the case before during the hurt and post-hurt phases of attack-related events because the vanilla code resets the attack strength ticker almost immediately, making `getAttackStrengthScale` unviable.

Shadows-of-Fire:
IMO this is fine but it should probably be called `getLastAttackStrength` to clearly note that it is for the strength of the *previous* attack and not the attack strength at the current time.
You probably also don't need to zero it out, since it should only be being used from an event context shortly after it's set (which can be noted in a javadoc).

bconlon1:
> IMO this is fine but it should probably be called `getLastAttackStrength` to clearly note that it is for the strength of the _previous_ attack and not the attack strength at the current time. You probably also don't need to zero it out, since it should only be being used from an event context shortly after it's set (which can be noted in a javadoc).

I guess it depends how you define previous vs. current attack. Since it is the strength for the current attack phase, but it is also technically the previous attack strength as the player has already attacked and it's also the last attack strength relative to calling `getAttackStrengthScale`. I worry calling it getLastAttackStrength would cause confusion due to the former point, maybe there could be another better name to use for it?

As for the zeroing out I think it's a good precaution to keep in *because* of the narrow context you mention.

Shadows-of-Fire:
The other name I can think of is `getCachedAttackStrength` which is non-temporal (doesnt imply current/prev due to ambiguity) and also doesn't imply it's a "live" (always valid) property like `getAttackStrength` would.

bconlon1:
`getCachedAttackStrength` sounds good so I've switched to use that.

bconlon1:
Fixed patches for 1.19.1.

LexManos:
1) What other events are you talking about that wouldn't be better served by just passing this variable down the stack?
Yes the value gets immediately reset, but why does that matter? What is the use case.
2) Why is a player only field implemented as a Living Entity getter. This is just bad..

bconlon1:
> 1. What other events are you talking about that wouldn't be better served by just passing this variable down the stack?
>    Yes the value gets immediately reset, but why does that matter? What is the use case.
> 2. Why is a player only field implemented as a Living Entity getter. This is just bad..

1. I'm not sure what you mean by the passing the variable down the stack part. But to elaborate on why the value getting immediately reset matters, it means that `getAttackStrengthScale()` is not actually a viable way of directly getting the attack strength during much of the damage event sequence because the `attackStrengthTicker` has been reset to 0. This makes it more difficult to get the attack strength during the sequence and the use-case is to make it more easily accessible than before.
2. The `attackStrengthTicker` field that is used for attack strength is a `LivingEntity` field, it's not player-only. This is why I decided to store the `cachedAttackStrength` field in `LivingEntity` as well to maintain consistency and make it accessible if it is ever needed to be used outside the `Player` in the event of another mod trying to implement attack strength for a `LivingEntity` or something. I can move it to `Player` if wanted but this was the reasoning of putting it in `LivingEntity`.

Shadows-of-Fire:
The event chain of reference here is potentially all of the following
* LivingAttackEvent / AttackEntityEvent
* LivingHurtEvent
* LivingDamageEvent
* LivingDeathEvent
* LivingDropsEvent

Making this field part of each of these events is... difficult, to say the least.  It is best served as a cached field for that reason.

The reasoning for having it be part of LivingEntity is as described above, because Mojang implemented the attack strength on LivingEntity, regardless of where it may currently be used at the moment.

patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
```diff
@@ -649,11 +649,20 @@
...
-@@ -3234,6 +_,64 @@
+@@ -3234,6 +_,73 @@
        this.m_21166_(p_21191_ == InteractionHand.MAIN_HAND ? EquipmentSlot.MAINHAND : EquipmentSlot.OFFHAND);
     }
  
 +   /* ==== FORGE START ==== */
++   /**
++    * Tracks the attack strength of the player as set within {@link net.minecraft.world.entity.player.Player#attack(Entity)} for outside usage
++    */
++   protected float cachedAttackStrength;
```
Shadows-of-Fire: Default this value to `1.0F`

patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
```diff
@@ -295,6 +302,14 @@
                          this.m_21008_(InteractionHand.MAIN_HAND, ItemStack.f_41583_);
                       }
                    }
+@@ -1215,6 +_,7 @@
+ 
+          }
+       }
++      this.cachedAttackStrength = 0.0F; // Forge: Resets the attackStrength field once the attack code is finished
```
Shadows-of-Fire: This reset can be removed, as it is spurious.  Whenever the value is considered "valid", it will have been set, meaning resetting it has no real benefit.

patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
```diff
@@ -649,11 +649,20 @@
...
                 p_21066_.m_7292_(new MobEffectInstance(pair.getFirst()));
              }
-@@ -3234,6 +_,64 @@
+@@ -3234,6 +_,73 @@
        this.m_21166_(p_21191_ == InteractionHand.MAIN_HAND ? EquipmentSlot.MAINHAND : EquipmentSlot.OFFHAND);
     }
  
 +   /* ==== FORGE START ==== */
++   /**
++    * Tracks the attack strength of the player as set within {@link net.minecraft.world.entity.player.Player#attack(Entity)} for outside usage
```
Shadows-of-Fire: This Javadoc needs to mention by name the series of events for which this value is considered valid, which are
* LivingAttackEvent / AttackEntityEvent
* LivingHurtEvent
* LivingDamageEvent
* LivingDeathEvent
* LivingDropsEvent

patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
```diff
@@ -649,11 +649,20 @@
...
++    */
++   protected float cachedAttackStrength;
++
++   public float getCachedAttackStrength() {
```
Shadows-of-Fire, line 672: More Ideally, this getter method also contains the same information as the field's javadocs, or at the very least, `@see LivingEntity#$cachedAttackStrength`

Shadows-of-Fire (changes requested):

patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
bconlon1: I agree with everything here except that I don't think the value would be valid in AttackEntityEvent considering the cached value is set *after* that event is called. However, you can just use `.getAttackStrengthScale()` in AttackEntityEvent because that does have a valid value at the time that event runs.

bconlon1:
Made the requested changes to the PR.

Drullkus:
What if you just moved `this.resetAttackStrengthTicker();` to after the if statement instead?

I checked the code that's currently executed inside `Player.attack()` and there aren't any other calls that'd affect `Player.attackStrengthTicker` besides `getAttackStrengthScale()` & `resetAttackStrengthTicker()`.

bconlon1:
A new PR #9000 has been made with the suggestions by @Drullkus.

PR #8892, opened 2022-07-18
add Biomes.BEACH to Tags

JT122406:


sciwhiz12 (changes requested):
Please run the datagenerator (`:forge:forge_data`) so the tag JSONs are regenerated with the new value.

sciwhiz12 (approved):

PR #8900, opened 2022-07-22
Make mixin work with JarInJar

marchermans:
This fixes a case where URIs which where modified externally did not redirect properly to the correct LayFS Instances.
The fix that is applied here is to force the use of LayFS when a new instance of a mod is created from a subpath.


Zippomoon:
Please backport the fix to 1.18.2.

Peach774:
Also adding that this needs to be backported to 1.18. JarInJar was backported and this fix has not been.

PR #8902, opened 2022-07-24
Fix connecting to certain servers

coehlrich:
Fixes connecting to servers that send characters in channel names in the register or unregister custom packets that aren't valid in `ResourceLocation` anymore.
The issue was caused by #8703 due to the client thread simply ignoring the exception while the netty thread disconnects from the server. This doesn't happen in vanilla for the register and unregister packets since vanilla does nothing with those.

### DIFF
--- a/src/main/java/net/minecraftforge/network/MCRegisterPacketHandler.java
+++ b/src/main/java/net/minecraftforge/network/MCRegisterPacketHandler.java
@@ -5,14 +5,17 @@
 
 package net.minecraftforge.network;
 
+import com.mojang.logging.LogUtils;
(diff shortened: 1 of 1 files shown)


pupnewfster:
Wouldn't it make more sense to use ResourceLocation#tryCreate or tryParse or whatever it called?

coehlrich:
The `tryParse` method returns a null if the `ResourceLocationException` occurs which isn't useful if you want to immediately use it only if it successfully parses (Which is currently the case here since it only needs to be added to a set) or log the exception.

cpw (approved):
This looks good to me.

PR #8903, opened 2022-07-24
[1.19.x] Add ServerOverloadedEvent

PaintNinja:
This PR intends to add a new event that allows mods to detect when the server is overloaded.

This'll allow mods to react to an overloaded server and optionally log data when this occurs, to help diagnose causes of server-side lag. An example usage of this event would be logging the current coordinates and worlds each player is in so that a server owner can investigate possible player-made lag machines.


Sirttas:
I have a bad feeling about it, I easily see a mod using this to "optimize" and just turn off random features vanilla or forge features.

PaintNinja:
> I have a bad feeling about it, I easily see a mod using this to "optimize" and just turn off random features vanilla or forge features.

Do you think discouraging this kind of behaviour and/or giving examples of good usages of this event inside the javadoc would be a good idea?

Sirttas:
> Do you think discouraging this kind of behaviour and/or giving examples of good usages of this event inside the javadoc would be a good idea?

Yes, it cold help. But I don't think its a good idea any mod alter its behavior using this event. your mod should be consistent even under heavy load. I only see this as a way to debug or provide information on server stability.

PaintNinja:
I've updated the javadoc to discourage using this event as a trigger for modifying Vanilla behaviour and provided a couple of intended usage examples. If anyone has any further feedback or would like the wording changed, please let me know.

patches/minecraft/net/minecraft/server/MinecraftServer.java.patch
```diff
@@ -66,6 +66,14 @@
           this.f_129726_ = Util.m_137550_();
           this.f_129757_.m_134908_(Component.m_237113_(this.f_129709_));
           this.f_129757_.m_134912_(new ServerStatus.Version(SharedConstants.m_183709_().getName(), SharedConstants.m_183709_().getProtocolVersion()));
+@@ -633,6 +_,7 @@
+                f_129750_.warn("Can't keep up! Is the server overloaded? Running {}ms or {} ticks behind", i, j);
+                this.f_129726_ += j * 50L;
+                this.f_129718_ = this.f_129726_;
++               net.minecraftforge.server.ServerLifecycleHooks.onServerOverloaded(this, i, j, this.f_129718_);
```
sciwhiz12, line 73: Is it intended that the hook fires after the `lastOverloadWarning` is set? Shouldn't it fire before the hook, so the field contains the previous value (accurately reflecting the time of the last overload warning)?

src/main/java/net/minecraftforge/event/server/ServerOverloadedEvent.java
```diff
@@ -0,0 +1,63 @@
...
+ * <br>
+ * This event is fired via {@link ServerLifecycleHooks#onServerOverloaded(MinecraftServer, long, long, long)},
+ * which is executed from {@link MinecraftServer#runServer()} <br>
+ * <br>
+ * This event is not {@link Cancelable}.<br>
+ * <br>
+ * This event does not have a result. {@link HasResult}<br>
+ * <br>
+ * This event is fired on the {@link MinecraftForge#EVENT_BUS}.
+ */
```
sciwhiz12: I suggest some improvements to this javadoc:

- Using the somewhat standardized format found currently in the client events.
- Using paragraph tags to mark paragraphs (except the first, which is implicitly a paragraph) instead of the line break tag.
- Using an unordered list instead of an ordered list, as the order of the list items doesn't matter.
- Removing the part about where the event is fired from, since that is information not needed by consumers.

```suggestion
/**
 * Fired when the server is running behind ticking schedule and can't keep up. Heavy work should be avoided in listeners
 * for this event, as doing so may cause the {@link net.minecraft.server.dedicated.ServerWatchdog} to timeout and crash
 * the server.
 *
 * <p>Please <strong>do not use this event for disabling vanilla features</strong> - this may cause unexpected behavior
 * for players and hard-to-reproduce compatibility issues with other mods that expect consistent vanilla behavior.</p>
 * 
 * <p>Here are some examples of intended usages of this event:</p>
 * <ul>
 *     <li>to log data that could be useful for diagnosing why the server is overloaded</li>
 *     <li>for sending automated alerts to server staff outside of the game and console</li>
 * </ul>
 *
 * <p>This event is not {@linkplain Cancelable cancellable}, and does not {@linkplain HasResult have a result}.</p>
 *
 * <p>This event is fired on the {@linkplain MinecraftForge#EVENT_BUS main Forge event bus},
 * only on the {@linkplain LogicalSide#SERVER logical server}.</p>
 */
```

src/main/java/net/minecraftforge/event/server/ServerOverloadedEvent.java
```diff
@@ -0,0 +1,63 @@
...
+public class ServerOverloadedEvent extends ServerLifecycleEvent
+{
+    private final long msBehind;
+    private final long ticksBehind;
+    private final long lastOverloadWarningMs;
+
+    public ServerOverloadedEvent(final MinecraftServer server, final long msBehind, final long ticksBehind, final long lastOverloadWarningMs)
```
sciwhiz12, line 38: This event constructor should be marked as `@ApiStatus.Internal`.

src/main/java/net/minecraftforge/event/server/ServerOverloadedEvent.java
```diff
@@ -0,0 +1,63 @@
...
+    {
+        super(server);
+        this.msBehind = msBehind;
+        this.ticksBehind = ticksBehind;
+        this.lastOverloadWarningMs = lastOverloadWarningMs;
+    }
+
+    /**
+     * @return The amount of milliseconds the server is behind schedule
```
sciwhiz12: Use inline `{@return}` tags.
```suggestion
     * {@return the amount of milliseconds the server is behind schedule}
```

src/main/java/net/minecraftforge/event/server/ServerOverloadedEvent.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public long getMsBehind() { return msBehind; }
+
+    /**
+     * @return The number of ticks the server is behind schedule.
```
sciwhiz12: Use inline `{@return}` tags.
```suggestion
     * {@return the number of ticks the server is behind schedule}
```

src/main/java/net/minecraftforge/event/server/ServerOverloadedEvent.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public long getTicksBehind() { return ticksBehind; }
+
+    /**
+     * @return The last time the server warned about being behind schedule.
```
sciwhiz12: Use inline `{@return}` tags.
```suggestion
     * {@return the last time the server warned about being behind schedule}
```

sciwhiz12 (changes requested):

sciwhiz12:
For reference, this PR was rejected by the core team on the basis of there being too many potential ways to abuse this event. Extra logging for debugging or lag diagnosis should preferrably be integrated into Forge, rather than reliant on 3rd-party mods.[^1]

[^1]: [`#github-discussions`, message from `@LexManos`](https://canary.discord.com/channels/313125603924639766/852298000042164244/1006645429762543647)

PR #8905, opened 2022-07-25
AnvilUpdateEvent fix

Luis-St:
This PR fix issue  #7708.

AnvilUpdateEvent is now fired whenever one of the input stacks (left or right) or the name changed


Fuzss:
Would it maybe be possible to use this opportunity to also fix the other issues this event has?

My main issues still is that you have to copy the whole vanilla method for updating the anvil menu output even when doing such simple things as changing the repair item for a tool. Most devs seem to be avoiding the event completely therefore, going straight for a Mixin into the related item’s class. This is definitely something that should be more easily possible via the event. 

I [made a PR that went nowhere](https://github.com/MinecraftForge/MinecraftForge/pull/6586) a while back addressing the issues fixed by this PR, the issue I just mentioned and also other minor details. Would be great if it could be included with this. I could also make a new PR myself if necessary. 

Since this changes the event quite significantly from the behaviour it had for many years now, maybe we could also rename it  to something like `AnvilChangeEvent` to force devs to take a look at the new implementation.

Luis-St:
**Edit**: This are only ideas which could be added in a overhaul

I like the idea of making more modifications to Anvil, but we have currently two AnvilEvents.
I think a global overhaul of both Events is necessary, since both events could have more context/usage.
For example:
1. Both Events could be moved into one class
    - AnvilUpdateEvent -> AnvilEvent.InputChanged
    - AnvilRepairEvent -> AnvilEvent.Repair
2. Add/modify the following sub events to allow more specific modifications
    - AnvilEvent.InputChanged will be fired whenever the the left input ItemStack changed (include name changes),
       Is cancelable to prevent the vanilla logic from being called (current AnvilUpdateEvent)
    - Add Level and BlockPos context to AnvilEvent.Repair
    - Add AnvilEvent.OuputChanged will be fired after the vanilla logic is called, 
      allow modifications on the output item and on the repair cost
    - Add AnvilEvent.NameChanged will be fired whenever the name changes are synced from the client.
       Is cancelable to prevent the item name to change, allow changes of the ItemStack name Component.
       For example this can be used to color the ItemStack name.

Guys what did you think about these Changes?
Any other suggestions which should be added?

Fuzss:
Probably a good solution to have multiple sub-events instead of calling the single one multiple times.

You should really add some context for the player to the event though, to show whether they are in creative mode or not (`this.player.getAbilities().instabuild`). Maybe even the whole player instance, although I don't think that much is necessary.

Luis-St:
Both Events already have a Player instance, did you mean a direct method reference?
Like `isPlayerInstabuild`

Fuzss:
Huh, you're right. Forget what I said, I was looking at some outdated implementation of the event from my own PR.

Luis-St:
~~Let's start with the overhaul~~
~~AnvilEvent.InputChanged~~
~~AnvilEvent.OuputChanged~~
~~AnvilEvent.NameChanged~~
~~AnvilEvent.ItemTake~~

Luis-St:
Since the whole thing probably brings a few major changes with it, this PC will only fix the error described above.
A overhaul of the AnvilEvents should be done in another PR.

Shadows-of-Fire:
So, while I believe this change does make it match the doc as-is, when is it useful to fire the event if only one of the input slots is full?

Luis-St:
To allow modifications of the ItemStack with experience without having a second item which is consumed.
For example: 
I have a Mod which allows the Player to reduce the repair cost of the ItemStack, by placing the item in the first input slot. 
This allows the player reduce the experience cost of further anvil actions. This is currently done with a mixin.

src/test/java/net/minecraftforge/debug/misc/OnAnvilUpdateEventFix.java
```diff
@@ -0,0 +1,29 @@
+package net.minecraftforge.debug.misc;
```
sciwhiz12: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/test/java/net/minecraftforge/debug/misc/OnAnvilUpdateEventFix.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import net.minecraftforge.event.AnvilUpdateEvent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
+
+@Mod(OnAnvilUpdateEventFix.MOD_ID)
+@Mod.EventBusSubscriber(modid = OnAnvilUpdateEventFix.MOD_ID)
+public class OnAnvilUpdateEventFix
```
sciwhiz12: Please provide a javadoc for this class which explains what feature this tests, how to test it, and what the expected results are from said test.

Additionally, this test mod should be expanded to generally test the `AnvilUpdateEvent`' functionality, or removed until someone makes a general functionality test. If expanded, the name should then be `AnvilUpdateEventTest`.

Also, please cleanup the formatting in this class to remove trailing whitespace

src/test/java/net/minecraftforge/debug/misc/OnAnvilUpdateEventFix.java
```diff
@@ -0,0 +1,29 @@
+package net.minecraftforge.debug.misc;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import net.minecraftforge.event.AnvilUpdateEvent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
+
+@Mod(OnAnvilUpdateEventFix.MOD_ID)
+@Mod.EventBusSubscriber(modid = OnAnvilUpdateEventFix.MOD_ID)
```
sciwhiz12: Instead of `EventBusSubscriber`, register the event listener in the mod constructor, blocked behind the `ENABLED` flag.

src/test/java/net/minecraftforge/debug/misc/OnAnvilUpdateEventFix.java
```diff
@@ -0,0 +1,29 @@
...
+public class OnAnvilUpdateEventFix 
+{
+    private static final boolean ENABLED = false;
+    private static final Logger LOGGER = LogManager.getLogger();
```
sciwhiz12: Use SLF4J's `Logger` and Mojang's `LogUtils.getLogger()`.

sciwhiz12 (changes requested):

Luis-St:
I added the wanted changes from sciwhiz12. I hope the description in the java javadoc is enough.
> Also, please cleanup the formatting in this class to remove trailing whitespace

I wasn't exactly sure what you meant by "trailing whitespace", but I think I removed it

src/test/java/net/minecraftforge/debug/misc/AnvilUpdateEventTest.java
```diff
@@ -0,0 +1,41 @@
...
+import net.minecraftforge.event.AnvilUpdateEvent;
+import net.minecraftforge.fml.common.Mod;
+
+/**
+ * This class provides a way to test the AnvilUpdateEvent. <br>
+ * The event is now fired as described in the javadoc (of AnvilUpdateEvent). <br>
+ * "AnvilUpdateEvent is fired whenever the input stacks (left or right) or the name in an anvil changed."
+ */
+@Mod(AnvilUpdateEventTest.MOD_ID)
+@Mod.EventBusSubscriber(modid = AnvilUpdateEventTest.MOD_ID)
```
sciwhiz12: The `@EventBusSubscriber` annotation should be removed.

sciwhiz12 (changes requested):

src/test/java/net/minecraftforge/debug/misc/AnvilUpdateEventTest.java
Luis-St: done

sciwhiz12 (approved):

PR #8914, opened 2022-07-28
[1.19] Add forge tags for tools and armors

Sirttas:
This PR adds forge tags for tools and armors:
- tools: `tools`, `tools/swords`, `tools/axes`, `tools/pickaxes`, `tools/shovels`, `tools/hoes`, `tools/shields`, `tools/bows`, `tools/crossbows`, `tools/fishing_rods`, `tools/tridents`
- armors: `armors`, `armors/helmets`, `armors/chestplates`, `armors/leggings`, `armors/boots`

Please tell me if I have forgotten any item. And should I add modded tool types like paxel/aiot?

### DIFF
--- a/src/generated/resources/data/forge/tags/items/armors.json
+++ b/src/generated/resources/data/forge/tags/items/armors.json
@@ -1,0 +1,8 @@
+{
+  "values": [
+    "#forge:armors/helmets",
+    "#forge:armors/chestplates",
+    "#forge:armors/leggings",
+    "#forge:armors/boots"
+  ]
+}
--- a/src/generated/resources/data/forge/tags/items/armors/boots.json
+++ b/src/generated/resources/data/forge/tags/items/armors/boots.json
@@ -1,0 +1,10 @@
+{
+  "values": [
+    "minecraft:leather_boots",
+    "minecraft:chainmail_boots",
+    "minecraft:iron_boots",
+    "minecraft:golden_boots",
+    "minecraft:diamond_boots",
+    "minecraft:netherite_boots"
+  ]
+}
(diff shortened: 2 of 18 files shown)


pupnewfster:
I wonder if it is worth somehow deprecating the shears tag and suggesting `tools/shears` be used instead? Not sure of the best way to do so though.

marchermans:
Question: why are they not named in singular like most of the others?  Or am I misinformed?

Sirttas:
> I wonder if it is worth somehow deprecating the shears tag and suggesting `tools/shears` be used instead? Not sure of the best way to do so though.

That's a good question, I would agree to deprecate it. TBH shears where out of the scope of the initial feature that motivated me to make this PR.

Sirttas:
> Question: why are they not named in singular like most of the others? Or am I misinformed?

Where do you want to put singular? Tags had always be plural.

KnightMiner:
One concern I have is a lot of these tags overlap with tool actions. As a result, I could see a modder falsely using the tag to determine if a tool is a pickaxe instead of the action

Sirttas:
> One concern I have is a lot of these tags overlap with tool actions. As a result, I could see a modder falsely using the tag to determine if a tool is a pickaxe instead of the action

I agree, and this is not the intended use. Adding comment to redirect people towards tool actions?

Sirttas:
I added javadoc to encourage the use of ToolActions.

KnightMiner:
If tool actions are preferred, what is the usecase of these tags?

Sirttas:
> If tool actions are preferred, what is the usecase of these tags?

Recipes, applying some alterations to tools or armors.

Ex:
- I have in Elemental Craft a feature called jewels where to get the bonus provided by the jewel you need to put it on a peace of gear and you then need to have it equipped. I use a tag to list the tools and armors that can receive a jewel.
- I also have toll infusion, a recipe that add a small bonus (ex: increase an enchantment level by 1), I also use a tag to know which tools must receive which bonus, because the buff is not the same with a sword or a pickaxe).

sciwhiz12 (approved):

PR #8921, opened 2022-07-30
[1.19.1]: Let isRemotePresent check minecraft:register for packet ids from other mod loaders. 

mysticdrew:
Other mod loaders send the registered packet ids on the "minecraft:register" packet. 
This PR lets the SimpleChannel.isRemotePreset check that if it is not in the "fml:conndata" registry. 

I need this because my mod is both optional on the server and the client and is loader agnostic, forge clients can connect to the mod on a fabric or quilt server, fabric clients can connect to the mod if it is on a forger server. 
I do not want to send random packets if the other side does not have the mod.

I will back port this to 1.18.2 when it is ready.

### DIFF
--- a/src/main/java/net/minecraftforge/network/MCRegisterPacketHandler.java
+++ b/src/main/java/net/minecraftforge/network/MCRegisterPacketHandler.java
@@ -37,7 +37,8 @@ public class MCRegisterPacketHandler
             byte[] data = new byte[Math.max(buffer.readableBytes(), 0)];
             buffer.readBytes(data);
             Set<ResourceLocation> oldLocations = this.locations;
-            this.locations = this.remoteLocations = bytesToResLocation(data);
+            this.locations = bytesToResLocation(data);
+            this.remoteLocations = Set.copyOf(this.locations);
             // ensure all locations receive updates, old and new.
             oldLocations.addAll(this.locations);
             oldLocations.stream()
--- a/src/main/java/net/minecraftforge/network/MCRegisterPacketHandler.java
+++ b/src/main/java/net/minecraftforge/network/MCRegisterPacketHandler.java
@@ -32,7 +32,8 @@ public class MCRegisterPacketHandler
 
     public static class ChannelList {
         private Set<ResourceLocation> locations = new HashSet<>();
-        private Set<ResourceLocation> remoteLocations = new HashSet<>();
+        private Set<ResourceLocation> remoteLocations = Set.of();
+
         public void updateFrom(final Supplier<NetworkEvent.Context> source, FriendlyByteBuf buffer, final NetworkEvent.RegistrationChangeType changeType) {
             byte[] data = new byte[Math.max(buffer.readableBytes(), 0)];
             buffer.readBytes(data);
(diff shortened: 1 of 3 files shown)


src/main/java/net/minecraftforge/network/MCRegisterPacketHandler.java
```diff
@@ -76,6 +76,16 @@ private Set<ResourceLocation> bytesToResLocation(byte[] all) {
...
             return rl;
         }
+
+        /**
+         * Provides only the registered locations that the remote side has sent.
+         * This is useful for interacting with other modloaders via the network to inspect registered network ids.
+         *
+         * @return - The remote locations
+         */
+        public Set<ResourceLocation> getRemoteLocations() {
```
pupnewfster, line 86: I am fairly certain this method exposes the ability for modders to mess things up and modify the set of remote locations and also locations (as they are being set to the same instance)

mysticdrew: Good point, I will make a copy of it!

src/main/java/net/minecraftforge/network/MCRegisterPacketHandler.java
```diff
@@ -76,6 +77,16 @@ private Set<ResourceLocation> bytesToResLocation(byte[] all) {
             }
             return rl;
         }
+
+        /**
+         * Provides only the registered locations that the remote side has sent.
+         * This is useful for interacting with other modloaders via the network to inspect registered network ids.
+         *
+         * @return - The remote locations
```
sciwhiz12: Just a bit of cleanup on the javadocs
```suggestion
         * {@return the unmodifiable set of channel locations sent by the remote side}
         * This is useful for interacting with other modloaders via the network to inspect registered network channel IDs.
```

src/main/java/net/minecraftforge/network/MCRegisterPacketHandler.java
```diff
@@ -32,12 +32,13 @@
 
     public static class ChannelList {
         private Set<ResourceLocation> locations = new HashSet<>();
-
+        private Set<ResourceLocation> remoteLocations = new HashSet<>();
```
sciwhiz12: As the value of `remoteLocations` is always assigned to be an unmodifiable set in `updateFrom`, this should also be initialized to an appropriate empty unmodifiable set (`Set.of()`).

Also, the blank space between the method and the fields should be preserved.

```suggestion
        private Set<ResourceLocation> remoteLocations = Set.of();

```

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/network/MCRegisterPacketHandler.java
mysticdrew: Sounds good!

src/main/java/net/minecraftforge/network/MCRegisterPacketHandler.java
mysticdrew: Done!

sciwhiz12 (approved):

PR #8927, opened 2022-07-31
Remove calls to getStepHeight in Player#maybeBackOffFromEdge

### LINKED ISSUE #8922: Separate attribute to control step-down height
Currently step height is controlled as a single attribute, which can affect mods adding step-up enhancements; players will not expect to step down a full block as well, especially while crouching. 

As a reference, this was reported as an issue to Tinker's Construct recently: https://github.com/SlimeKnights/TinkersConstruct/issues/4956  
A video posted there will demonstrate the severity of this issue and how it affects gameplay.

Shadows-of-Fire:
Fixes #8922

Step Height Addition was added to allow for easy modification of how many blocks a player will step-up automatically, but the current implementation has a side effect of making players able to fall off ledges of equal height when sneaking.  This PR reverts the changes to that specific method so sneaking behavior works as expected, instead of allowing you to drop off a cliff if you have increased step height.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
@@ -194,33 +194,14 @@
                       this.m_21008_(p_36159_, ItemStack.f_41583_);
                    }
  
-@@ -1020,7 +_,7 @@
-          double d1 = p_36201_.f_82481_;
-          double d2 = 0.05D;
+@@ -1014,6 +_,7 @@
+       return !this.f_36077_.f_35935_;
+    }
  
--         while(d0 != 0.0D && this.f_19853_.m_45756_(this, this.m_20191_().m_82386_(d0, (double)(-this.f_19793_), 0.0D))) {
-+         while(d0 != 0.0D && this.f_19853_.m_45756_(this, this.m_20191_().m_82386_(d0, (double)(-this.getStepHeight()), 0.0D))) {
-             if (d0 < 0.05D && d0 >= -0.05D) {
-                d0 = 0.0D;
-             } else if (d0 > 0.0D) {
-@@ -1030,7 +_,7 @@
-             }
-          }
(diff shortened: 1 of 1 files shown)


patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
```diff
@@ -194,33 +194,14 @@
...
+@@ -1014,6 +_,7 @@
+       return !this.f_36077_.f_35935_;
+    }
  
--         while(d0 != 0.0D && d1 != 0.0D && this.f_19853_.m_45756_(this, this.m_20191_().m_82386_(d0, (double)(-this.f_19793_), d1))) {
-+         while(d0 != 0.0D && d1 != 0.0D && this.f_19853_.m_45756_(this, this.m_20191_().m_82386_(d0, (double)(-this.getStepHeight()), d1))) {
-             if (d0 < 0.05D && d0 >= -0.05D) {
-                d0 = 0.0D;
-             } else if (d0 > 0.0D) {
++   // Don't update this method to use IForgeEntity#getStepHeight() - https://github.com/MinecraftForge/MinecraftForge/issues/8922
```
sciwhiz12: This should start with `// FORGE: `, so it's easier to find all Forge-added comments by searching for `FORGE:`.

sciwhiz12 (changes requested):
Tested with the `/attribute` command and it works.

/cc @KnightMiner @MarioSMB for the original issue report, @williewillus as the original creator of the step height addition attribute PR

patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
Shadows-of-Fire: fixed

sciwhiz12 (approved):

MarioSMB:
Amazing swift work, thank you!

PR #8928, opened 2022-07-31
[1.19.x] Custom Holderset Types

Commoble:
This PR adds a serializer registry for custom holderset types that can be used in worldgen jsons and implements several builtin holderset types that mods and datapacks can use.

## What's a Holderset?

An interface that represents collection of registry elements and acts as both a set and an indexable list. Holdersets are how mojang implements the "element, list, or tag" format by which elements can be represented in worldgen jsons:

```json5
{
  "a_biome": "desert",
  "some_biomes": ["desert", "swamp", "forest"],
  "biome_tag": "#is_overworld"
}
```

HolderSets use registry-sensitive codecs; when serialized without registry context (RegistryOps), they are flattened to lists and serialized as lists of unregistered values (which aren't particularly useful). Vanilla currently does not use holdersets in any network codec, however if they were, they would likely be serialized without registry context (e.g. synced datapack registries do not currently write/read with registry context when synced).

The tag holdersets are deceptively tricky to work with; they are mutable and have their contents rebound when tags reload (the other two vanilla holderset types are immutable). The mutability of tag holdersets in conjunction with the need to cache an indexable list of holders makes implementing custom composite holdersets slightly convoluted.

This PR expands the existing holderset codecs to allow a fourth format, a type-dispatched holderset object:

```json5
{
   "custom_holderset_type":
  {
    "type": "modid:typename",
    // additional context as needed
  }
}
```

To maintain consistency with the vanilla codec and vanilla netcode, holdersets (including custom types) are always serialized as lists of unregistered values when serialized without registry context (vanilla does not use registry context when syncing datapack registries). It's worth repeating that unregistered registrables are mostly useless and can cause all kinds of problems, and vanilla (perhaps because of this) does not currently use holdersets in any network codec, so it's probably best for mods not to use holdersets in datapack registries' network codecs (or any other netcode).

## APIs
* IForgeHolderSet<T>
  * Extension interface patched onto HolderSet, the sole current purpose of which is to allow tag holdersets to invalidate any composite holdersets that "own" them by invalidating IForgeHolderSet's addInvalidationListener method. Custom holderset types composed of other holdersets should also override this method so that they can be notified of tags rebinding.
* ICustomHolderSet<T> extends HolderSet<T>
  * An interface which all custom holderset types must implement. This provides a HolderSetType getter, which is necessary for implementing the type dispatcher and ensures holdersets are serializable when datagenerating datapack registry elements that have holderset fields (such as biome modifiers).
* HolderSetType
  * A registrable holderset serializer type. Mods that implement custom holderset types should register one of these via DeferredRegister for each ICustomHolderSet-implementing class (We can't simply register a holderset codec because the way that vanilla creates their own holderset codec convolutes this slightly).

## Builtin HolderSet Types
In addition to the three vanilla holderset types (singleton, list, and tag), this PR adds four additional holderset types to forge:

### forge:any
This immutable holderset type represents all holders in a given registry. Its holderset interfaces are implemented by delegating to the registry itself as much as possible and does not create any list/set caches. Json format:
```json5
{
  "type": "forge:any"
}
```

### forge:and
This mutable composite holderset type represents an intersection of other holdersets. It lazily caches a list and set of its holders as needed and invalidates these caches when tags reload. Json format:
```json5
{
  "type": "forge:and",
  "values": [ list of holderset strings, lists, or objects ]
}
```

### forge:or
This mutable composite holderset type represents a union of other holdersets. It lazily caches a list and set of its holders and invalidates these caches when tags reload. Json format:
```json5
{
  "type": "forge:or",
  "values": [ list of holderset strings, lists, or objects ]
}
```

### forge:not
This mutable composite holderset type represents all holders not present in another holderset. It lazily caches a list (but not a set) of its holders and invalidates this list when tags reload. Json format:
```json5
{
  "type": "forge:not",
  "value": holderset string, list, or object
}
```

## Patches
* HolderSetCodec is patched to create a dispatch codec (for the custom holder set types) and an either codec (eithering the dispatch codec with the vanilla tag/list codec) when constructed.
* HolderSetCodec#encode is patched to use the dispatch codec when encoding custom holder sets with registry context. It falls back to the vanilla tag/list codec for vanilla holdersets, and keeps the vanilla behavior of always encoding as a list when encoding without registry context.
* HolderSetCodec#decode is patched to use the either(dispatch, vanilla) codec when decoding holdersets with registry context. It keeps the vanilla behavior of only decoding a list when decoding without registry context.

## Test Datapack / Example Usage
The attached datapack provides a biome modifier that uses all seven vanilla/forge holderset formats to add a placed feature to all non-sandy biomes except swamps, cold mountains, and stony peaks:

[custom_holdersets_test.zip](https://github.com/MinecraftForge/MinecraftForge/files/9274188/custom_holdersets_test.zip)

```json
{
	"type": "forge:add_features",
	"features": "forge:red_wool",
	"step": "underground_ores",
	"biomes":
	{
		"type": "forge:and",
		"values":
		[
			{
				"type": "forge:and",
				"values":
				[
					{
						"type": "forge:any"
					},
					{
						"type": "forge:not",
						"value": "#forge:is_sandy"
					}
				]
			},
			{
				"type": "forge:not",
				"value":
				{
					"type": "forge:or",
					"values":
					[
						["swamp", "mangrove_swamp"],
						"stony_peaks",
						{
							"type": "forge:and",
							"values":
							[
								"#is_mountain",
								"#forge:is_cold"
							]
						}
					]
				}
			}
		]
	}
}
```


src/main/java/net/minecraftforge/common/extensions/IForgeHolderSet.java
```diff
@@ -0,0 +1,32 @@
...
+ * SPDX-License-Identifier: LGPL-2.1-only
+ */
+
+package net.minecraftforge.common.extensions;
+
+import net.minecraft.core.HolderSet;
+
+public interface IForgeHolderSet<T>
+{
+    default public HolderSet<T> self()
```
sciwhiz12: The `self()` method in extension interfaces should be `private`.
```suggestion
    private HolderSet<T> self()
```

src/main/java/net/minecraftforge/common/extensions/IForgeHolderSet.java
```diff
@@ -0,0 +1,32 @@
...
+    {
+        return (HolderSet<T>) this;
+    }
+    
+    /**
+     * <p>Adds a callback to run when this holderset's contents invalidate (i.e. because tags were rebound).
+     * The intended usage and use case is with composite holdersets that need to cache sets/list based on other
+     * holdersets, which may be mutable (because they are tag-based or themselves composite holdersets).
+     * Composite holdersets should use this to add callbacks to each of their component holdersets when constructed.</p>
```
sciwhiz12: If this is all a single paragraph, the paragraph tags can be omitted. 

```suggestion
     * Adds a callback to run when this holderset's contents invalidate (i.e. because tags were rebound).
     * The intended usage and use case is with composite holdersets that need to cache sets/list based on other
     * holdersets, which may be mutable (because they are tag-based or themselves composite holdersets).
     * Composite holdersets should use this to add callbacks to each of their component holdersets when constructed.
```

However, if you decide to split the second and third sentences into their own paragraph (which would probably improve readability a bit), then the opening paragraph tag should be moved to the beginning of the second sentence, and a blank space added to improve source readability.

```suggestion
     * Adds a callback to run when this holderset's contents invalidate (i.e. because tags were rebound).
     *
     * <p>The intended usage and use case is with composite holdersets that need to cache sets/list based on other
     * holdersets, which may be mutable (because they are tag-based or themselves composite holdersets).
     * Composite holdersets should use this to add callbacks to each of their component holdersets when constructed.</p>
```

src/main/java/net/minecraftforge/registries/HolderSetType.java
```diff
@@ -0,0 +1,17 @@
...
+
+package net.minecraftforge.registries;
+
+import com.mojang.serialization.Codec;
+
+import net.minecraft.core.Holder;
+import net.minecraft.core.Registry;
+import net.minecraft.resources.ResourceKey;
+
+public interface HolderSetType
```
sciwhiz12: The builtin Forge implementations of this interface seem to use this as a functional interface. Would there be a case where this interface would not be implemented as such (i.e. as a lambda or method reference)? If not, might be good to mark this as a `@FunctionalInterface`. If yes, then it should be documented that it can be implemented as a method reference.

Also, Forge naming convention calls for all interfaces to be prefixed with `I`.

```suggestion
@FunctionalInterface
public interface IHolderSetType
```

src/main/java/net/minecraftforge/registries/HolderSetType.java
```diff
@@ -0,0 +1,17 @@
...
+{
+    public abstract <T> Codec<? extends ICustomHolderSet<T>> makeCodec(ResourceKey<? extends Registry<T>> registryKey, Codec<Holder<T>> holderCodec, boolean forceList);
```
sciwhiz12: Non-private interface member methods do not need to be marked as `public abstract`.
```suggestion
    <T> Codec<? extends ICustomHolderSet<T>> makeCodec(ResourceKey<? extends Registry<T>> registryKey, Codec<Holder<T>> holderCodec, boolean forceList); 
```

src/main/java/net/minecraftforge/registries/ICustomHolderSet.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Interface for mods' custom holderset types
+ */
+public interface ICustomHolderSet<T> extends HolderSet<T>
+{
+    /**
+     * {@return HolderSetType registered to {@link ForgeRegistries.HOLDER_SET_TYPES}}
+     */
+    public abstract HolderSetType type();
```
sciwhiz12: Non-private interface member methods do not need to be marked as `public abstract`.
```suggestion
    HolderSetType type();
```

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -212,6 +218,31 @@
      */
     public static final RegistryObject<Codec<NoneStructureModifier>> NONE_STRUCTURE_MODIFIER_TYPE = STRUCTURE_MODIFIER_SERIALIZERS.register("none", () -> Codec.unit(NoneStructureModifier.INSTANCE));
 
+    /**
+     * Stock holder set type that represents any/all values in a registry. Can be used in a holderset object with { "type": "forge:any" }
```
sciwhiz12: The JSON code should be wrapped in a `{@code}` tag. (Same comment for each of the below javadocs.)

```suggestion
     * Stock holder set type that represents any/all values in a registry. Can be used in a holderset object with {@code { "type": "forge:any" }}.
```

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -212,6 +218,31 @@
...
+     */
+    public static final RegistryObject<HolderSetType> ANY_HOLDER_SET = HOLDER_SET_TYPES.register("any", () -> AnyHolderSet::codec);
+   
+    /**
+     * Stock holder set type that represents an intersection of other holdersets. Can be used in a holderset object with { "type": "forge:and", "values": [list of holdersets] }
+     */
+    public static final RegistryObject<HolderSetType> AND_HOLDER_SET = HOLDER_SET_TYPES.register("and", () -> AndHolderSet::codec);
+
+    /**
+     * Stock holder set type that represents an intersection of other holdersets. Can be used in a holderset object with { "type": "forge:or", "values": [list of holdersets] }
```
sciwhiz12: This is not an intersection but a union of other holdersets.
```suggestion
     * Stock holder set type that represents a union of other holdersets. Can be used in a holderset object with {@code { "type": "forge:or", "values": [list of holdersets] }}
```

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -212,6 +218,31 @@
...
+     */
+    public static final RegistryObject<HolderSetType> OR_HOLDER_SET = HOLDER_SET_TYPES.register("or", () -> OrHolderSet::codec);
+
+    /**
+     * Stock holder set type that represents values that exist in one holderset but not another. Can be used in a holderset object with { "type": "forge:exclusion", "include": holderset, "exclude": holderset }.
+     */
+    public static final RegistryObject<HolderSetType> EXCLUSION_HOLDER_SET = HOLDER_SET_TYPES.register("exclusion", () -> ExclusionHolderSet::codec);
+
+    /**
+     * Stock holder set type that represents all values in a registry except those in another given set. Can be used in a holderset object with { "type": "forge:not", "value": holderset }
```
sciwhiz12: The javadocs should echo your PR description's recommendation to use the `exclusion` holderset type instead of `and(x, not(y))`.

src/main/java/net/minecraftforge/registries/ForgeHolderSets.java
```diff
@@ -0,0 +1,555 @@
...
+import net.minecraft.resources.HolderSetCodec;
+import net.minecraft.resources.RegistryOps;
+import net.minecraft.resources.ResourceKey;
+import net.minecraft.tags.TagKey;
+import net.minecraft.util.RandomSource;
+import net.minecraftforge.common.ForgeMod;
+
+public final class ForgeHolderSets
+{
+    private ForgeHolderSets() {} // Organizational class
```
sciwhiz12: Should we consider moving these inner classes to a new subpackage as full (outer) classes?

sciwhiz12 (changes requested):
Works fine according to the tests. (Also, may be good to use `json5` as the language for the codeblocks in your PR description so the comments don't appear in alarming red.)

src/main/java/net/minecraftforge/registries/holdersets/AndHolderSet.java
```diff
@@ -0,0 +1,75 @@
+package net.minecraftforge.registries.holdersets;
```
sciwhiz12, line 6: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/registries/holdersets/AnyHolderSet.java
```diff
@@ -0,0 +1,97 @@
+package net.minecraftforge.registries.holdersets;
```
sciwhiz12, line 6: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/registries/holdersets/CompositeHolderSet.java
```diff
@@ -0,0 +1,159 @@
+package net.minecraftforge.registries.holdersets;
```
sciwhiz12, line 6: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/registries/holdersets/ExclusionHolderSet.java
```diff
@@ -0,0 +1,70 @@
+package net.minecraftforge.registries.holdersets;
```
sciwhiz12: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/registries/holdersets/NotHolderSet.java
```diff
@@ -0,0 +1,159 @@
+package net.minecraftforge.registries.holdersets;
```
sciwhiz12, line 6: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/registries/holdersets/OrHolderSet.java
```diff
@@ -0,0 +1,61 @@
+package net.minecraftforge.registries.holdersets;
```
sciwhiz12, line 6: Missing license header, please run the `updateLicenses`/`licenseFormat` task.

sciwhiz12 (changes requested):

sciwhiz12 (approved):

src/main/java/net/minecraftforge/common/extensions/IForgeHolderSet.java
```diff
@@ -0,0 +1,33 @@
...
+ * SPDX-License-Identifier: LGPL-2.1-only
+ */
+
+package net.minecraftforge.common.extensions;
+
+import net.minecraft.core.HolderSet;
+
+public interface IForgeHolderSet<T>
+{
+    private HolderSet<T> self()
```
LexManos: Self is only needed if the extension interface's default implementation itself needs to cast itself.

src/main/java/net/minecraftforge/common/ForgeMod.java
```diff
@@ -212,6 +219,34 @@
...
+
+    /**
+     * Stock holder set type that represents a union of other holdersets. Can be used in a holderset object with {@code { "type": "forge:or", "values": [list of holdersets] }}
+     */
+    public static final RegistryObject<HolderSetType> OR_HOLDER_SET = HOLDER_SET_TYPES.register("or", () -> OrHolderSet::codec);
+
+    /**
+     * Stock holder set type that represents values that exist in one holderset but not another. Can be used in a holderset object with {@code { "type": "forge:exclusion", "include": holderset, "exclude": holderset }}
+     */
+    public static final RegistryObject<HolderSetType> EXCLUSION_HOLDER_SET = HOLDER_SET_TYPES.register("exclusion", () -> ExclusionHolderSet::codec);
```
LexManos: Remove this, as it is just a `and(A, not(B))` would rather not have this as API.

LexManos (changes requested):
Aa couple small notes. And do we have verification that this is never sent over the network/doesn't break vanilla  networking? I dont think HolderSets are sent over the network. The only thing I can think of is recipes, but I think that's resolved into an ItemStack array.

Commoble:
Removed forge:exclusion, datapack in original post has been amended to not have this.

Re; network serialization: Vanilla never syncs holdersets, they're not used in any datapack registry network codecs or any reload listeners. For datapack registries in particular, synced datapack registries are serialized without registry context, which would cause any holderset to be serialized as a list of values and decoded as a list of unregistered values, which would probably cause problems in most contexts, however I have refactored the HolderSetCodec patch so that this behavior is used for custom holderset types in addition to vanilla holdersets, for consistency.

LexManos:
This  all look fine to  me, I couldn't find anywhere in vanilla code where they are serialized over the network so I am pretty sure that this doesn't break anything. Lets give it a try.

PR #8932, opened 2022-08-01
[1.19.x] Custom usage animations for items

tmvkrpxl0:
This is port of #8419 
I have ported changes from the original PR to 1.19.x


src/main/java/net/minecraftforge/client/IForgeArmPose.java
```diff
@@ -0,0 +1,47 @@
...
+import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.client.extensions.common.IClientItemExtensions;
+
+/**
+ * An ArmPose that can be defined by the user.
+ * Register one by creating a custom {@link net.minecraft.client.model.HumanoidModel.ArmPose}
+ * and returning it in {@link IClientItemExtensions#getArmPose(LivingEntity, InteractionHand, ItemStack)}.
+ */
+public interface IForgeArmPose
```
sciwhiz12: The interface nature of this class and its name may suggest that it is an extension interface for `HumanoidModel.ArmPose`. Given that the single method on it applies a transform on the arm, it may be better to name this something like `IArmPoseTransformer`.

Additionally, if there is no reasonable and usual case in which a mod might implement this interface as a class, it may be best to mark this as a `@FunctionalInterface`.

patches/minecraft/net/minecraft/client/model/HumanoidModel.java.patch
```diff
@@ -0,0 +1,52 @@
...
++   public static enum ArmPose implements net.minecraftforge.common.IExtensibleEnum {
+       EMPTY(false),
+       ITEM(false),
+       BLOCK(false),
+@@ -416,5 +_,22 @@
+       public boolean m_102897_() {
+          return this.f_102890_;
+       }
++      // FORGE START
++      private net.minecraftforge.client.IForgeArmPose forgeArmPose;
```
sciwhiz12: This should be `final` and marked `@Nullable`, and the original constructor modified to set this field to `null`.

patches/minecraft/net/minecraft/client/model/HumanoidModel.java.patch
```diff
@@ -0,0 +1,52 @@
...
++
++      private ArmPose(boolean twoHanded, net.minecraftforge.client.IForgeArmPose forgeArmPose) {
++         this(twoHanded);
++         this.forgeArmPose = forgeArmPose;
```
sciwhiz12, line 47: As this constructor is only invoked by mods through the `create` method, this should do a non-null check to ensure that mods cannot pass a `null` here (and subsequently trip on the later condition in `applyTransform`).

src/main/java/net/minecraftforge/client/extensions/common/IClientItemExtensions.java
```diff
@@ -57,6 +59,20 @@ default Font getFont(ItemStack stack, FontContext context)
...
+    /**
+      * This method returns an ArmPose that can be defined using the {@link net.minecraft.client.model.HumanoidModel.ArmPose#create(String, boolean, IForgeArmPose)} method.
+      * This allows for creating custom item use animations.
+      *
+      * @param entityLiving The entity holding the item
+      * @param hand         The hand the ArmPose will be applied to
+      * @param itemStack    The stack being held
+      * @return A custom ArmPose that can be used to define movement of the arm
+      */
+            @Nullable
```
sciwhiz12: Formatting: extra whitespace.
```suggestion
    @Nullable
```

src/test/java/net/minecraftforge/debug/entity/player/ItemUseAnimationTest.java
```diff
@@ -0,0 +1,104 @@
...
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.RegistryObject;
+
+import java.util.function.Consumer;
+
+@Mod(ItemUseAnimationTest.MOD_ID)
+public class ItemUseAnimationTest
```
sciwhiz12, line 35: Please provide a javadoc for this class which explains what feature this tests, how to test it, and what the expected results are from said test.

src/test/resources/META-INF/mods.toml
```diff
@@ -240,7 +240,8 @@ modId="custom_transformtype_test"
 modId="mega_model_test"
 [[mods]]
 modId="emissive_elements_test"
-
```
sciwhiz12, line 249: Please keep the blank line.

src/test/java/net/minecraftforge/debug/entity/player/ItemUseAnimationTest.java
```diff
@@ -0,0 +1,104 @@
...
+{
+
+    public static final String MOD_ID = "item_use_animation_test";
+
+    private static final DeferredRegister<Item> ITEMS = DeferredRegister.create(ForgeRegistries.ITEMS, MOD_ID);
+
+    private static final RegistryObject<Item> THING = ITEMS.register("thing", () -> new ThingItem(new Item.Properties().tab(CreativeModeTab.TAB_TOOLS).food(new FoodProperties.Builder().nutrition(4).build())));
```
sciwhiz12: For convenience of testing, please mark this item's food properties as `alwaysEat()`.

src/test/java/net/minecraftforge/debug/entity/player/ItemUseAnimationTest.java
```diff
@@ -0,0 +1,104 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.debug.entity.player;
```
sciwhiz12: Outdated license header, please run the `updateLicenses`/`licenseFormat` task.

src/main/java/net/minecraftforge/client/IForgeArmPose.java
```diff
@@ -0,0 +1,47 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.client;
```
sciwhiz12: Outdated license header, please run the `updateLicenses`/`licenseFormat` task.

sciwhiz12 (changes requested):
CC @rehwinkel as the author of the original 1.18 PR, for consideration of these review comments to your PR.

src/main/java/net/minecraftforge/client/IForgeArmPose.java
tmvkrpxl0: Done

src/test/java/net/minecraftforge/debug/entity/player/ItemUseAnimationTest.java
tmvkrpxl0: Also done

src/test/java/net/minecraftforge/debug/entity/player/ItemUseAnimationTest.java
tmvkrpxl0: Done

src/test/resources/META-INF/mods.toml
tmvkrpxl0: So that there's 3 blank line between `# ADD ABOVE THIS LINE` and `modId="item_use_animation_test"`?

sciwhiz12: Yep, so there are only additions to this file.

tmvkrpxl0: Done

patches/minecraft/net/minecraft/client/model/HumanoidModel.java.patch
tmvkrpxl0: Should I move `//FORGE START` up? I'm unsure if modified original constructor should be included between `//FORGE START` and `//FORGE END`

src/main/java/net/minecraftforge/client/IForgeArmPose.java
tmvkrpxl0: I'm in for marking it with `FunctionalInterface`, but does it have any downside for mods wanting to implement it as a class?

patches/minecraft/net/minecraft/client/model/HumanoidModel.java.patch
sciwhiz12: The marker should remain where it is, as it denotes the part of the class which is added by Forge. If one wanted to mark the modification to the original constructor, they'd use `// FORGE: `, but it's not needed here since the field it assigns is inside the block.

src/main/java/net/minecraftforge/client/IForgeArmPose.java
sciwhiz12: It shouldn't have any downsides, the annotation is more of a marker to indicate to users that the interface is intended to be a functional interface. (Though, the compiler is mandated by spec to emit errors if an interface annotated with `@FunctionalInterface` is not a functional interface, akin to e.g. `@Override`.)

tmvkrpxl0:
@sciwhiz12 All done!

patches/minecraft/net/minecraft/client/model/HumanoidModel.java.patch
```diff
@@ -0,0 +1,60 @@
...
+          this.f_102890_ = p_102896_;
++         this.forgeArmPose = null;
+       }
+ 
+       public boolean m_102897_() {
+          return this.f_102890_;
+       }
++      // FORGE START
++      @javax.annotation.Nullable
++      private net.minecraftforge.client.IArmPoseTransformer forgeArmPose;
```
sciwhiz12: This should be marked `final`, and correspondingly the patched-in constructor below should be modified to _not_ call the original constructor but set the `twoHandled` field itself.

sciwhiz12 (changes requested):

patches/minecraft/net/minecraft/client/model/HumanoidModel.java.patch
tmvkrpxl0: Ok, changed it now

sciwhiz12 (approved):

tmvkrpxl0:
@amadornes I have applied comments from you

src/test/resources/META-INF/mods.toml
```diff
@@ -243,6 +243,8 @@ modId="mega_model_test"
 [[mods]]
 modId="emissive_elements_test"
 [[mods]]
+modId="anvil_update_event_fix"
+[[mods]]
 modId="item_use_animation_test"
```
tmvkrpxl0, line 248: This is weird, it looks fine on original file https://github.com/tmvkrpxl0/MinecraftForge/blob/itemanim-1.19.x/src/test/resources/META-INF/mods.toml
But it looks weird on commit page

tmvkrpxl0:
@sciwhiz12 I didn't receive any reply from amadornes yet, so If you don't mind, could you take a look at recent changes I made instead?

patches/minecraft/net/minecraft/client/renderer/ItemInHandRenderer.java.patch
```diff
@@ -24,6 +24,14 @@
              boolean flag1 = CrossbowItem.m_40932_(p_109377_);
              boolean flag2 = humanoidarm == HumanoidArm.RIGHT;
              int i = flag2 ? 1 : -1;
+@@ -384,6 +_,7 @@
+             this.m_109322_(p_109372_, p_109377_, flag2 ? ItemTransforms.TransformType.FIRST_PERSON_RIGHT_HAND : ItemTransforms.TransformType.FIRST_PERSON_LEFT_HAND, !flag2, p_109379_, p_109380_, p_109381_);
+          } else {
+             boolean flag3 = humanoidarm == HumanoidArm.RIGHT;
++            if (net.minecraftforge.client.extensions.common.IClientItemExtensions.of(p_109377_).applyForgeHandTransform(p_109379_, f_109299_.f_91074_, humanoidarm, p_109377_, p_109373_, p_109378_, p_109376_)) {} else
```
sciwhiz12: This can be simplified into a `if (!condition)` without the `else`. Additionally, it may be good to put a `// FORGE: ` marker.

patches/minecraft/net/minecraft/world/item/UseAnim.java.patch
```diff
@@ -0,0 +1,10 @@
+--- a/net/minecraft/world/item/UseAnim.java
++++ b/net/minecraft/world/item/UseAnim.java
+@@ -9,5 +_,6 @@
+    SPEAR,
+    CROSSBOW,
+    SPYGLASS,
+-   TOOT_HORN;
++   TOOT_HORN,
++   CUSTOM;
```
sciwhiz12: Could you please clarify for me when this `UseAnim` value should be used by modders?

sciwhiz12 (changes requested):

patches/minecraft/net/minecraft/client/renderer/ItemInHandRenderer.java.patch
tmvkrpxl0: It is to skip any other vanilla transformations and go straight into rendering if it returns true, and that's why there's `else` there
It is meant to be `else if`, combined with if statement right after it.

patches/minecraft/net/minecraft/world/item/UseAnim.java.patch
tmvkrpxl0: It is to avoid vanilla trying to "animate" items that has custom usage animation. Should I add javadoc?

patches/minecraft/net/minecraft/client/renderer/ItemInHandRenderer.java.patch
sciwhiz12: The same can be achieved with `if (!condition)`, because it will take the whole `if-else` chain below as the one statement guarded by the `if (!condition)`.

As in, instead of:
```java
if (condition) {} else
if (first) {
	// ...
} else if (second) {
	// ...
}
```

We can instead do:
```java
if (!condition)
if (first) {
	// ...
} else if (second) {
	// ...
}
```

The above can be reformatted to look like the following (with the braces in comments further illustrating the point:
```java
if (!condition) /* { */
	if (first) {
		// ...
	} else if (second) {
		// ...
	}
/* } */
```

patches/minecraft/net/minecraft/world/item/UseAnim.java.patch
sciwhiz12: Please do so, especially so it is clearly differentiated versus `UseAnim.NONE` and marked as being added by Forge.

patches/minecraft/net/minecraft/client/renderer/ItemInHandRenderer.java.patch
tmvkrpxl0: ah I see what you mean now, thank you

PR #8933, opened 2022-08-01
`LivingEntity#isBlocking` use `ToolActions#SHIELD_BLOCK` instead of `UseAnim#BLOCK`

Sirttas:
This PR alters the call to `LivingEntity#isBlocking` to rely on `ToolAction` insteadof the vanilla `UseAnim`.

Before that you needed to also change your item's `getUseAnimation` to use it as a shield, now only the tool action is required.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
@@ -590,6 +590,15 @@
           if (this.f_20935_.m_41781_()) {
              this.m_21329_();
           }
+@@ -2989,7 +_,7 @@
+    public boolean m_21254_() {
+       if (this.m_6117_() && !this.f_20935_.m_41619_()) {
+          Item item = this.f_20935_.m_41720_();
+-         if (item.m_6164_(this.f_20935_) != UseAnim.BLOCK) {
++         if (this.f_20935_.canPerformAction(net.minecraftforge.common.ToolActions.SHIELD_BLOCK)) {
+             return false;
+          } else {
+             return item.m_8105_(this.f_20935_) - this.f_20936_ >= 5;
 @@ -3115,8 +_,8 @@
        }
  
--- a/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
@@ -595,7 +595,7 @@
        if (this.m_6117_() && !this.f_20935_.m_41619_()) {
           Item item = this.f_20935_.m_41720_();
 -         if (item.m_6164_(this.f_20935_) != UseAnim.BLOCK) {
-+         if (this.f_20935_.canPerformAction(net.minecraftforge.common.ToolActions.SHIELD_BLOCK)) {
++         if (item.m_6164_(this.f_20935_) != UseAnim.BLOCK && !this.f_20935_.canPerformAction(net.minecraftforge.common.ToolActions.SHIELD_BLOCK)) { // TODO 1.20 only use the tool action
              return false;
           } else {
              return item.m_8105_(this.f_20935_) - this.f_20936_ >= 5;


patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
```diff
@@ -590,6 +590,15 @@
           if (this.f_20935_.m_41781_()) {
              this.m_21329_();
           }
+@@ -2989,7 +_,7 @@
+    public boolean m_21254_() {
+       if (this.m_6117_() && !this.f_20935_.m_41619_()) {
+          Item item = this.f_20935_.m_41720_();
+-         if (item.m_6164_(this.f_20935_) != UseAnim.BLOCK) {
++         if (this.f_20935_.canPerformAction(net.minecraftforge.common.ToolActions.SHIELD_BLOCK)) {
```
sciwhiz12: The removal of the original condition would mean items that use `UseAnim.BLOCK` will break. Perhaps it would be better to check this in `Item#getUseAnimation` on the provided item stack and return `UseAnim.BLOCK` if the item supports the `SHIELD_BLOCK` tool item. It would also allow the item to change its use animation if it so chooses.

sciwhiz12 (changes requested):

patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
Sirttas: So you means having the two conditions as a or? 

The two goals of this PR are:
1 having the SHIELD_BLOCK being the sole condition for blocking.
2 decoupling the bock animation from its action allowing people to use custom animation (with for example #8932).

But yes the change as I summited it is a breaking change.

sciwhiz12: Why will the `SHIELD_BLOCK` tool action become the sole determiner of whether the use animation is `BLOCK`? Why do we want to deny the block animation to anything other than shields that may wish to use it? It would unnecessarily break `UseAnim#BLOCK` for those items.

My suggestion is to insert a hook into `Item#getUseAnimation` which checks if the item stack can perform the `SHIELD_BLOCK` action, and return `UseAnim#BLOCK` immediately if so. 

This way, shields implemented through `SHIELD_BLOCK` use the `BLOCK` animation by default, while optionally giving them the ability to override the use animation if they so desire (through a custom `UseAnim` via #8932). Additionally, it would not constitute a breaking change (in that items using `UseAnim#BLOCK` through override still work correctly).

Sirttas: I don't think changing `Item#getUseAnimation` is a good idea because it is also used elsewhere  and won't solve the issue of using a different `UseAnim` than `UseAnim#BLOCK`, I think tool actions are more suited for this role.

sciwhiz12: Oh, wait. 🤦‍♂️ I misread the intention of the PR. I thought this was fixing the use animation to `BLOCK` by default when the item is able to be used as a shield, not that it was fixing `isBlocking` so it doesn't check the use animation to determine whether it is a shield. Sorry for the confusion.

Though, fully replacing the conditional still somewhat constitutes a breaking change, since modded items might have relied on this behavior. It should be at least ORing the current and the new condition, to preserve the current behavior. If we decide that only items that implement the `SHIELD_BLOCK` tool action can block (as determined by `isBlocking`), then it should be marked as a TODO for 1.20.

Sirttas: > Oh, wait. 🤦‍♂️ I misread the intention of the PR.

It may be caused by how bad I am at describing things.

Sirttas:
I added back the use anim, this should be good now.

sciwhiz12 (approved):

PR #8940, opened 2022-08-04
Add support for handling chunk generators in DFU

### LINKED ISSUE #8915: Updating 1.19.0 worlds with modded dimensions to 1.19.1 breaks the world
The twist is that the dimension must be made with a modded chunk generator or biome source to trigger the game's inability to parse the level.dat file. Attached below is the world in 1.19.0 with Bumblezone mod on, the datapack error when trying to access the world in 1.19.1 without Bumblezone mod on, and the missing nether/end when trying to access the world in 1.19.1 with the Bumblezone mod on. Notice how in the logs, it complains about an unsupported key being Bumblezone's chunk generator. I did not change the name of the modded chunk generator at all. My 1.19.1 update is really just a fix for the music disc items as their constructors changed. I did not change anything regarding worldgen.

1.19.0 world entered with Bumblezone on.
World file: [TestBefore.zip](https://github.com/MinecraftForge/MinecraftForge/files/9215519/TestBefore.zip)
image

1.19.0 world entered in 1.19.1 without Bumblezone on.
Log file: https://paste.ee/p/1D8pR - [latest.log](https://github.com/MinecraftForge/MinecraftForge/files/9219138/latest.log)
World file: [TestAfter.zip](https://github.com/MinecraftForge/MinecraftForge/files/9215539/TestAfter.zip)
image

1.19.0 world entered in 1.19.1 with Bumblezone on.
Log file: https://paste.ee/p/Khql0 - [latest.log](https://github.com/MinecraftForge/MinecraftForge/files/9219131/latest.log)
World file: [TestAfterWithMod.zip](https://github.com/MinecraftForge/MinecraftForge/files/9215573/TestAfterWithMod.zip)
My unreleased 1.19.1 version of my mod: [the_bumblezone_forge-6.1.5+1.19.1.zip](https://github.com/MinecraftForge/MinecraftForge/files/9215568/the_bumblezone_forge-6.1.5%2B1.19.1.zip)
image

I cannot release my mod while this issue persist as it would cause too many people to freak out about missing vanilla dimensions or broken worlds. Hence this issue report in hopes that this can be patched by Forge to help both my mod and other modded dimension mods

marchermans:
This adds an event that gets fired when DFU is loaded to register stuff to it.
Right now this only deals with the chunk generators, but could use expansion later.

Delays DFU init to after mod loading.

Closes #8915

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/EntityType.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/EntityType.java.patch
@@ -90,7 +90,7 @@
        private Builder(EntityType.EntityFactory<T> p_20696_, MobCategory p_20697_) {
           this.f_20685_ = p_20696_;
           this.f_20686_ = p_20697_;
-@@ -641,12 +_,36 @@
+@@ -641,12 +_,37 @@
           return this;
        }
  
@@ -119,8 +119,8 @@
 +      }
 +
        public EntityType<T> m_20712_(String p_20713_) {
--         if (this.f_20688_) {
-+         if (this.f_20688_ && false) {
++         if(false) // FORGE: Do not preinitialize DFU via static init of Bootstrap, it is not needed in production.
+          if (this.f_20688_) {
              Util.m_137456_(References.f_16785_, p_20713_);
           }
  
--- a/patches/minecraft/net/minecraft/world/entity/EntityType.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/EntityType.java.patch
@@ -119,7 +119,8 @@
 +      }
 +
        public EntityType<T> m_20712_(String p_20713_) {
-          if (this.f_20688_) {
+-         if (this.f_20688_) {
++         if (this.f_20688_ && false) {
              Util.m_137456_(References.f_16785_, p_20713_);
           }
  
--- a/src/main/java/net/minecraftforge/datafix/ForgeDSL.java
+++ b/src/main/java/net/minecraftforge/datafix/ForgeDSL.java
@@ -1,0 +1,25 @@
+package net.minecraftforge.datafix;
+
+import com.mojang.datafixers.DSL;
+import com.mojang.datafixers.schemas.Schema;
+import com.mojang.datafixers.types.Type;
+import com.mojang.datafixers.types.templates.TaggedChoice;
+import com.mojang.datafixers.types.templates.TypeTemplate;
+import net.minecraftforge.datafix.event.DataFixEvent;
+
+import java.util.Map;
+import java.util.function.Supplier;
+
+public class ForgeDSL
+{
+
+    private ForgeDSL()
+    {
+        throw new IllegalStateException("Can not instantiate an instance of: ForgeDSL. This is a utility class");
+    }
+
+    public static <K> TaggedChoice<K> taggedChoiceLazy(final Class<? extends Schema> schemaClass, final String name, final Type<K> keyType, final Map<K, Supplier<TypeTemplate>> templates) {
+        final Map<K, Supplier<TypeTemplate>> finalValues = DataFixEvent.EventFactory.fireTaggedChoiceFor(schemaClass, name, keyType, templates);
+        return DSL.taggedChoiceLazy(name, keyType, finalValues);
+    }
+}
--- a/src/main/java/net/minecraftforge/datafix/event/DataFixEvent.java
+++ b/src/main/java/net/minecraftforge/datafix/event/DataFixEvent.java
@@ -1,0 +1,94 @@
+package net.minecraftforge.datafix.event;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
+import com.mojang.datafixers.schemas.Schema;
+import com.mojang.datafixers.types.Type;
+import com.mojang.datafixers.types.templates.TypeTemplate;
+import net.minecraftforge.eventbus.api.Event;
+import net.minecraftforge.fml.ModContainer;
+import net.minecraftforge.fml.ModLoader;
+import net.minecraftforge.fml.ModLoadingContext;
+import net.minecraftforge.fml.event.IModBusEvent;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.function.Supplier;
+
+public class DataFixEvent extends Event implements IModBusEvent
+{
+    public static final class EventFactory {
+
+        public static <K> Map<K, Supplier<TypeTemplate>> fireTaggedChoiceFor(final Class<? extends Schema> schemaClass, final String name, final Type<K> keyType, final Map<K, Supplier<TypeTemplate>> templates) {
+            final ConfigureTaggedChoice<K> event = new ConfigureTaggedChoice<>(schemaClass, name, keyType, templates);
+
+            ModLoader.get().postEventWithWrapInModOrder(
+                    event,
+                    (mc, e) -> ModLoadingContext.get().setActiveContainer(mc),
+                    (mc, e) -> ModLoadingContext.get().setActiveContainer(null)
+            );
+
+            return event.build();
+        }
+    }
+
+    public static class SchemaConstruction extends DataFixEvent {
+        private final Class<? extends Schema> schemaClass;
+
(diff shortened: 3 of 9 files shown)


patches/minecraft/net/minecraft/world/entity/EntityType.java.patch
```diff
@@ -119,7 +119,8 @@
 +      }
 +
        public EntityType<T> m_20712_(String p_20713_) {
-          if (this.f_20688_) {
+-         if (this.f_20688_) {
++         if (this.f_20688_ && false) {
```
pupnewfster: Shouldn't the false condition to skip this be either added at the start of this if statement for readability. (potentially adding it as a single line if above to reduce patch size slightly)

src/main/java/net/minecraftforge/datafix/ForgeDSL.java
```diff
@@ -0,0 +1,25 @@
...
+import com.mojang.datafixers.schemas.Schema;
+import com.mojang.datafixers.types.Type;
+import com.mojang.datafixers.types.templates.TaggedChoice;
+import com.mojang.datafixers.types.templates.TypeTemplate;
+import net.minecraftforge.datafix.event.DataFixEvent;
+
+import java.util.Map;
+import java.util.function.Supplier;
+
+public class ForgeDSL
```
alcatrazEscapee, line 13: A new class for a single method which could've just as easily been in `ForgeHooks`, since ya know, it's a hook into patched code from Forge?

src/main/java/net/minecraftforge/datafix/ForgeDSL.java
```diff
@@ -0,0 +1,25 @@
...
+{
+
+    private ForgeDSL()
+    {
+        throw new IllegalStateException("Can not instantiate an instance of: ForgeDSL. This is a utility class");
+    }
+
+    public static <K> TaggedChoice<K> taggedChoiceLazy(final Class<? extends Schema> schemaClass, final String name, final Type<K> keyType, final Map<K, Supplier<TypeTemplate>> templates) {
```
alcatrazEscapee, line 21: Braces on newlines.

src/main/java/net/minecraftforge/datafix/event/DataFixEvent.java
```diff
@@ -0,0 +1,94 @@
...
+import net.minecraftforge.fml.event.IModBusEvent;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.function.Supplier;
+
+public class DataFixEvent extends Event implements IModBusEvent
+{
+    public static final class EventFactory {
```
alcatrazEscapee, line 21: Two different brace styles in three lines.

src/main/java/net/minecraftforge/datafix/event/DataFixEvent.java
```diff
@@ -0,0 +1,94 @@
...
+
+        public static <K> Map<K, Supplier<TypeTemplate>> fireTaggedChoiceFor(final Class<? extends Schema> schemaClass, final String name, final Type<K> keyType, final Map<K, Supplier<TypeTemplate>> templates) {
```
alcatrazEscapee, line 23: More braces, but I'll stop pointing them out.

src/test/java/net/minecraftforge/debug/world/SandboxChunkGeneratorTest.java
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+    }
+
+    public static class TestChunkGenerator extends ChunkGenerator
+    {
+        public static Codec<? extends ChunkGenerator> makeCodec()
+        {
+            Codec<TestChunkGenerator> codec = RecordCodecBuilder.create(builder -> builder.group(
+                                                                                                  RegistryOps.retrieveRegistry(Registry.STRUCTURE_SET_REGISTRY).forGetter(generator -> generator.structureSets),
```
alcatrazEscapee, line 76: Is there more code on this line? Oh, there is, it's just indented far enough it was off my screen.

It's test mod code so it's whatever but why no static single codec instance, like every other chunk generator?

alcatrazEscapee (changes requested):
So if I understand this correctly, this does not fix the bug unless the mod author in question adds the following event handler:

```java
@SubscribeEvent
public final void configueDFUSchema(final DataFixEvent.ConfigureTaggedChoice<String> event) {
    if (event.getSchemaClass() == V2832.class && Objects.equals(event.getName(), "type")) {
        event.register("sandbox:test", DSL::remainder);
    }
}
```

Is there *any other use case* for this event? i.e. could you conceivably use this event to do anything else? Why would you?

You fire the event form `V2832` but then in the event handler check `== V2832.class` - the event is only fired in one location, why would you not make it specific to that? Is there some future DFU use case that making arbitrary tagged choices is a useful bit of infrastructure to put in place?

You add three new events in this to solve one problem, only one of which is ever fired, which it is fired in one exactly location. Is `DataFixEvent` ever going to be fired? Is `SchemaConstructionEvent` ever going to be fired? Are you predicting the introduction of many new events into this hiearchy? Digging into DFU wasn't meant to be a challenge to see if you could overengineer something more than it already did.

There's an `EventFactory` class... for no reason? Because the method that fires an event should be in it's own class, rather than putting it in `ForgeHooks` like pretty much any other event?

---

Here's a random thought: what if, instead of everything here, you ~~patch the immutable map in `V2832` to be a map implementation that returns a default `DSL.remainder()` to any key not in the map?~~  That would be all of one patch, one line of code. Sounds to me like it would work. But what do I know.

Above wouldn't work, would have to be explicit in DFU i.e. `DSL.or(..., DSL::remainder)` or something.

marchermans:
> So if I understand this correctly, this does not fix the bug unless the mod author in question adds the following event handler:
> 
> ```java
> @SubscribeEvent
> public final void configueDFUSchema(final DataFixEvent.ConfigureTaggedChoice<String> event) {
>     if (event.getSchemaClass() == V2832.class && Objects.equals(event.getName(), "type")) {
>         event.register("sandbox:test", DSL::remainder);
>     }
> }
> ```
> 
> Is there _any other use case_ for this event? i.e. could you conceivably use this event to do anything else? Why would you?
> 
> You fire the event form `V2832` but then in the event handler check `== V2832.class` - the event is only fired in one location, why would you not make it specific to that? Is there some future DFU use case that making arbitrary tagged choices is a useful bit of infrastructure to put in place?
> 
> You add three new events in this to solve one problem, only one of which is ever fired, which it is fired in one exactly location. Is `DataFixEvent` ever going to be fired? Is `SchemaConstructionEvent` ever going to be fired? Are you predicting the introduction of many new events into this hiearchy? Digging into DFU wasn't meant to be a challenge to see if you could overengineer something more than it already did.
> 
> There's an `EventFactory` class... for no reason? Because the method that fires an event should be in it's own class, rather than putting it in `ForgeHooks` like pretty much any other event?
> 
> Here's a random thought: what if, instead of everything here, you ~patch the immutable map in `V2832` to be a map implementation that returns a default `DSL.remainder()` to any key not in the map?~ That would be all of one patch, one line of code. Sounds to me like it would work. But what do I know.
> 
> Above wouldn't work, would have to be explicit in DFU i.e. `DSL.or(..., DSL::remainder)` or something.

Your whole idea does not work.
The idea of this PR as it has been discussed internally is to try and figure out a way to make DFU modifyable to modders.
And no simply smacking an `DSL::remainder` call on it solves literally nothing for tagged choices. Since the type is literally looked up to determine what the remaining structure needs to look like (which might then be `DSL::remainder´ or some other Type).

So yes this PR looks over engineered but it gets a way with a tiny patch size, and is designed so we can easily add this to other locations where it might be needed in the future, or even in places where we find it now already.

TLDR: This is not fixable with remainder, nor is it fixable from directly within forge, the keys need to be known, and since the registries are not initialized and we can not delay this to after the registries initialize, we are stuck with this situation. Mojang design this explicitly this way, take it up with them.

src/main/java/net/minecraftforge/datafix/ForgeDSL.java
marchermans: This is ment to expanded in the future and is clearly not a hook but a DFU DSL entry.

TheCurle:
This PR was superseded by #8957.

PR #8941, opened 2022-08-04
Fire `MobEffectEvent.Added` *after* adding effects

mont-40:
Fixes a bug where event firing is false positive when adding a weaker version
Event fires at a time more in line with its class description
Allows mod makers to change `AttributeModifier`s added due to the effect

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
@@ -118,15 +118,21 @@
              iterator.remove();
           }
  
-@@ -904,6 +_,7 @@
-          return false;
-       } else {
-          MobEffectInstance mobeffectinstance = this.f_20945_.get(p_147208_.m_19544_());
-+         net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.entity.living.MobEffectEvent.Added(this, mobeffectinstance, p_147208_, p_147209_));
+@@ -907,17 +_,21 @@
           if (mobeffectinstance == null) {
              this.f_20945_.put(p_147208_.m_19544_(), p_147208_);
              this.m_142540_(p_147208_, p_147209_);
-@@ -918,6 +_,9 @@
+-            return true;
+          } else if (mobeffectinstance.m_19558_(p_147208_)) {
+             this.m_141973_(mobeffectinstance, true, p_147209_);
+-            return true;
+          } else {
+             return false;
+          }
++
++         net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.entity.living.MobEffectEvent.Added(this, mobeffectinstance, p_147208_, p_147209_));
++         return true;
+       }
     }
  
     public boolean m_7301_(MobEffectInstance p_21197_) {


patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
```diff
@@ -118,15 +118,21 @@
...
+-            return true;
+          } else if (mobeffectinstance.m_19558_(p_147208_)) {
+             this.m_141973_(mobeffectinstance, true, p_147209_);
+-            return true;
+          } else {
+             return false;
+          }
++
++         net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.entity.living.MobEffectEvent.Added(this, mobeffectinstance, p_147208_, p_147209_));
++         return true;
```
sciwhiz12, line 134: The consolidtion of the `return true` lines is unnecessary. The hook can simply be inserted before each original `return true`.

However, this change would move the firing of the event to after the callbacks to the entity are fired and, in the case of the `else if` statement, after `MobEffectInstance#update` is called. 

This means that in that situation, the existing mob effect instance passed to the event has been modified (updated based on the new mob effect instance), breaking the contract specified in the javadoc for `MobEffectEvent.Added#getEffectInstance()` and therefore being a breaking change to existing consumers of this event.

Even then, I would think it wouldn't be too useful to consumers to get the state of the original mob effect _after_ it has been updated.

sciwhiz12 (changes requested):
As stated in my review comment, this is currently a breaking change for consumers of the `MobEffectEvent.Added` event. This means we cannot accept this as-is for 1.19.1, as the window for breaking changes has narrowed down to only critical fixes.

If you wish to continue this PR, it will have to wait until the next breaking changes window that allows this kind of behavior change (likely to be 1.20).

patches/minecraft/net/minecraft/world/entity/LivingEntity.java.patch
mont-40: I agree with your former comments. This is a breaking change, it makes sense to continue this PR for the next breaking update. To answer your last comment, I've noted in my commit message that the change "allows mod makers to change `AttributeModifier`s added due to the effect". Currently these modifiers are applied after the event is fired. An example could be weakening or strengthening a modifier based on some additional condition. This is very easy to do if the modifiers are applied before the event is fired, as in this PR.

sciwhiz12: I wonder if there should be a new event fired when an existing effect is updated due to a new effect being applied.

So `MobEffectEvent.Added` would remain to communicate either a new effect being added (where the old mob instance effect is `null`) or there is an existing effect that might be updated (where the old mob instance effect is not `null`) and a new event, let's say `MobEffectEvent.Updated`, communicates that the existing mob instance effect was updated and actually had an effect, as seen in the second `return true`.

This would mean `MobEffectEvent.Added` would remain unmodified and therefore not be a breaking change, while a new event will be added to capture the state of the mob effect after being updated/merged.

mont-40: That suggestion sounds good as well. Alternatively, we could have an `Added` event, as in the current behaviour, and something like an `Applied` event, to notify that the current effect has been applied. Would you like me to create a new PR (or update the current one?)

sciwhiz12: I would prefer this PR be updated, to keep this review thread open for cleaner history purposes.

PR #8948, opened 2022-08-08
Fix Enchantment#doPostHurt and Enchantment#doPostAttack being called twice for players

Shadows-of-Fire:
Title.  Fixes vanilla bug https://bugs.mojang.com/browse/MC-248272

### DIFF
--- a/patches/minecraft/net/minecraft/world/item/enchantment/EnchantmentHelper.java.patch
+++ b/patches/minecraft/net/minecraft/world/item/enchantment/EnchantmentHelper.java.patch
@@ -38,24 +38,22 @@
        }, p_44834_);
        return mutablefloat.floatValue();
     }
-@@ -168,7 +_,7 @@
+@@ -168,6 +_,7 @@
           m_44853_(enchantmenthelper$enchantmentvisitor, p_44824_.m_20158_());
        }
  
--      if (p_44825_ instanceof Player) {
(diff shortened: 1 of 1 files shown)


patches/minecraft/net/minecraft/world/item/enchantment/EnchantmentHelper.java.patch
```diff
@@ -38,6 +38,24 @@
        }, p_44834_);
        return mutablefloat.floatValue();
     }
+@@ -168,7 +_,7 @@
+          m_44853_(enchantmenthelper$enchantmentvisitor, p_44824_.m_20158_());
+       }
+ 
+-      if (p_44825_ instanceof Player) {
++      if (false && p_44825_ instanceof Player) { // Forge: Fix MC-248272
```
sciwhiz12: A simpler patch would be to insert `if (false)` directly above the existing `if` statement. (Same suggestion for the other hunk in the patch.)
```suggestion
+      if (false) // FORGE: Stop calling mainhand item for players twice, fixes MC-248272
       if (p_44825_ instanceof Player) {
```

sciwhiz12 (changes requested):

Shadows-of-Fire:
@sciwhiz12 patches simplified, probably ready for assignment

sciwhiz12 (approved):

PR #7821, opened 2021-06-16
Add (cancelable) ToastAddEvent for use in ToastGui.add

Roundaround:
Gives modders the ability to hook into the ToastGui's toast queue. To prevent the toast from being added, cancel the event. To change the toast, modify it directly or replace it with event.setToast(IToast).

PR #8952, opened 2022-08-10
Add cancellable event firing when Toasts are added (1.19)

eerussianguy:
This is a 1.19 port of #7821

Cancelling this event just causes nothing to happen, it's a safe operation. In relation to the concerns raised by addTimedToast, that provides extra functionality with relation to auto-hiding, it's not a full-fledged second toast handler that we have to handle separately.

### DIFF
--- a/patches/minecraft/net/minecraft/client/gui/components/toasts/ToastComponent.java.patch
+++ b/patches/minecraft/net/minecraft/client/gui/components/toasts/ToastComponent.java.patch
@@ -1,0 +1,13 @@
+--- a/net/minecraft/client/gui/components/toasts/ToastComponent.java
++++ b/net/minecraft/client/gui/components/toasts/ToastComponent.java
+@@ -103,7 +_,9 @@
+    }
+ 
+    public void m_94922_(Toast p_94923_) {
+-      this.f_94916_.add(p_94923_);
++      var event = new net.minecraftforge.client.event.ToastAddEvent(p_94923_);
++      if (net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(event)) return;
++      this.f_94916_.add(event.getToast());
+    }
+ 
+    public Minecraft m_94929_() {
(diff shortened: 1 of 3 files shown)


patches/minecraft/net/minecraft/client/gui/components/toasts/ToastComponent.java.patch
```diff
@@ -0,0 +1,13 @@
+--- a/net/minecraft/client/gui/components/toasts/ToastComponent.java
++++ b/net/minecraft/client/gui/components/toasts/ToastComponent.java
+@@ -103,7 +_,9 @@
+    }
+ 
+    public void m_94922_(Toast p_94923_) {
+-      this.f_94916_.add(p_94923_);
++      var event = new net.minecraftforge.client.event.ToastAddEvent(p_94923_);
++      if (net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(event)) return;
```
marchermans: Since the event does not have a result, make an event factory in the relevant hooks class which like all the others.

marchermans (changes requested):

patches/minecraft/net/minecraft/client/gui/components/toasts/ToastComponent.java.patch
```diff
@@ -0,0 +1,13 @@
+--- a/net/minecraft/client/gui/components/toasts/ToastComponent.java
++++ b/net/minecraft/client/gui/components/toasts/ToastComponent.java
+@@ -103,7 +_,9 @@
+    }
+ 
+    public void m_94922_(Toast p_94923_) {
+-      this.f_94916_.add(p_94923_);
++      var event = net.minecraftforge.client.ForgeHooksClient.onToastAdd(p_94923_);
++      if (event.isCanceled()) return;
++      this.f_94916_.add(event.getToast());
```
marchermans: You are understanding me wrong.
The fired event can not change the toast you passed in.
So it suffices to return a boolean from the hook method, and turn this 4 line patch into a single line patch above the add method.

marchermans (changes requested):

Shadows-of-Fire (approved):

PR #8918, opened 2022-07-29
[1.18.x] Make LivingSetAttackTargetEvent compatible with the Brain/Behavior system.

Budschie:
This PR makes the `LivingSetAttackTargetEvent` compatible with the behavior system. This allows modders to check which entities are being attacked by axolotls, zoglins, hoglins and piglins (which was not possible previously). The example mod included in this PR allows you to be ignored by piglins when holding a stick in your main hand.

The motivation of creating this PR was that I wanted players not to get attacked by piglins whenever they were morphed into one using a morph mod. Achieving this was not possible by just using events, hence I created this PR.

PR #8954, opened 2022-08-10
[1.19.x] Make LivingSetAttackTargetEvent compatible with the Brain/Behavior system. (Port of PR #8918)

Budschie:
This PR makes the `LivingSetAttackTargetEvent` compatible with the behavior system. This allows modders to check which entities are being attacked by axolotls, zoglins, hoglins and piglins (which was not possible previously). The example mod included in this PR allows you to be ignored by piglins when holding a stick in your main hand.

The motivation of creating this PR was that I wanted players not to get attacked by piglins whenever they were morphed into one using a morph mod. Achieving this was not possible by just using events, hence I created this PR.


eerussianguy:
A couple QOL ideas, not blockers:

- Have a helper method in the event that sets the target in the correct memory
- Have a helper method that erases the target
- Have a helper method called like getPlayerTarget() that returns null if the target is not a player

Budschie:
> A couple QOL ideas, not blockers:
> 
> * Have a helper method in the event that sets the target in the correct memory
> 
> * Have a helper method that erases the target
> 
> * Have a helper method called like getPlayerTarget() that returns null if the target is not a player

Good idea! I will probably also replace the LivingSetAttackTargetEvent\#isCausedByBehavior by something more generic like an enum (of course this enum will implement an interface because of extensibility).

Budschie:
> A couple QOL ideas, not blockers:
> 
> * Have a helper method in the event that sets the target in the correct memory
> 
> * Have a helper method that erases the target
> 
> * Have a helper method called like getPlayerTarget() that returns null if the target is not a player

I have now successfully implemented a `LivingSetAttackTargetEvent#setTarget` method which allows you to set a new target or erase the target (by providing `null` as an input to that method).

src/main/java/net/minecraftforge/event/entity/living/LivingSetAttackTargetEvent.java
```diff
@@ -27,16 +35,168 @@
...
+    
+    // Short note on why this is needed: If we were to just let the event "return" data like in the ComputeFovModifierEvent,
+    // this would introduce lots of bugs and breaking changes, so we need to do this mess.
+    // 
+    // Note for any future developer that might stumble across this: Please introduce a breaking change
+    // later down the line (like, when porting forge to 1.20 or 1.21; this is the reaons why all the members of this class are marked for removal) 
+    // and do all of this like it is done in the ComputeFovModifierEvent. By that time, most mod developers will
+    // (hopefully) use LivingSetAttackTargetEvent#setTarget instead of setting the target on their own, which will
+    // minimize the impact of such a breaking change.
+    public static interface ILivingTargetType
```
diesieben07: Why does this need to be public and therefor be subject to breaking changes? I think this interface should just be gone, the enum made private and used directly internally.
To then allow creating the event with the two different types you can introduce static factory methods.

diesieben07 (changes requested):

src/main/java/net/minecraftforge/event/entity/living/LivingSetAttackTargetEvent.java
Budschie: The reason that I used an interface is to make the event extensible for other mods (that maybe add new ways of setting a Target). However, after thinking about it a bit, I must admit that this will probably never happen, so, yeah, I should probably remove the interface.

diesieben07 (changes requested):
Another problem that I have noticed: 
Your call to `setTarget` will cause `LivingSetAttackTarget` to be fired again, offering up a big risk for infinite recursion.
Additionally, due to the way the attack code works, preventing attacks via LivingSetAttackTarget is problematic.
Considering the following scenario:
- A Mob can attack players
- There are 2 players nearby
- Mob AI picks player A to attack
- LivingSetAttackTarget is fired and a mod prevents the attack by using setTarget(null)
- random timeout happens (using `NearestAttackableTargetGoal` as an example)
- Mob AI picks player A to attack again

=> Mob never picks a target to attack, even though it could attack player B

Budschie:
> Another problem that I have noticed: Your call to `setTarget` will cause `LivingSetAttackTarget` to be fired again, offering up a big risk for infinite recursion. Additionally, due to the way the attack code works, preventing attacks via LivingSetAttackTarget is problematic. Considering the following scenario:
> 
> * A Mob can attack players
> 
> * There are 2 players nearby
> 
> * Mob AI picks player A to attack
> 
> * LivingSetAttackTarget is fired and a mod prevents the attack by using setTarget(null)
> 
> * random timeout happens (using `NearestAttackableTargetGoal` as an example)
> 
> * Mob AI picks player A to attack again
> 
> 
> => Mob never picks a target to attack, even though it could attack player B

Yep, that's problematic, and I've come to the conclusion that I should probably remove the stuff about setting targets from this PR and move that into a separate PR (and a separate event).

Budschie:
> => Mob never picks a target to attack, even though it could attack player B

I wonder if this behavior can be avoided. 
It should be pretty doable to integrate something like a `TargetSelectedEvent` into the `NearestAttackableTargetGoal` (although there are a few other target goals in which I would have to hook into, for example the `DefendVillageTargetGoal`), yet I fear that Minecraft's behavior part of the target selection code is designed in a way in which Forge cannot easily hook into.

The `StartAttacking` behavior is simply provided with a Function that can be used to determine the entity which should be attacked; there is no single target selection method to hook into. Instead, I would have to hook into every entities' target selection code separately, which *is* feasible, but it is a high maintenance hassle nonetheless as you would have to check every single MC update for new entities which use the `StartAttacking` behavior. Additionally, it would also put the burden on supporting this event for modded entities on the modders, which is also not great. 

So, all in all, I think that finding an elegant solution to this problem is rather challenging; I would be happy if anyone has any ideas or solutions regarding the difficulties mentioned above.

src/main/java/net/minecraftforge/event/entity/living/LivingChangeTargetEvent.java
```diff
@@ -0,0 +1,74 @@
...
+    public LivingEntity getOriginalTarget()
+    {
+        return originalTarget;
+    }
+    
+    @Override
+    public boolean isCancelable()
+    {
+        return true;
+    }
```
ChampionAsh5357: Use `@Cancelable` annotation, not overriding method on event.

src/main/java/net/minecraftforge/event/entity/living/LivingChangeTargetEvent.java
```diff
@@ -0,0 +1,74 @@
...
+import net.minecraft.world.entity.LivingEntity;
+import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.event.entity.living.LivingSetAttackTargetEvent.ILivingTargetType;
+import net.minecraftforge.eventbus.api.Event.HasResult;
+
+/**
+ * This event allows you to change the target an entity has. <br>
+ * This event is fired before {@link LivingSetAttackTargetEvent}. <br>
+ * <br>
+ * This event is fired via the {@link ForgeHooks#}
```
ChampionAsh5357: Not finished javadoc.

src/main/java/net/minecraftforge/event/entity/living/LivingSetAttackTargetEvent.java
```diff
@@ -27,16 +28,48 @@
...
+    
+    public static interface ILivingTargetType
+    {
+        
+    }
+    
+    public static enum LivingTargetType implements ILivingTargetType
+    {
+        MOB_TARGET,
+        BEHAVIOR_TARGET;
```
ChampionAsh5357: Documentation on the differences between the two targets.

src/main/java/net/minecraftforge/event/entity/living/LivingSetAttackTargetEvent.java
```diff
@@ -27,16 +28,48 @@
  **/
 public class LivingSetAttackTargetEvent extends LivingEvent
 {
-
-    private final LivingEntity target;
+    private final ILivingTargetType targetType;
+    private final LivingEntity originalTarget;
+    
     public LivingSetAttackTargetEvent(LivingEntity entity, LivingEntity target)
```
ChampionAsh5357, line 41: Deprecate, mark for removal.

ChampionAsh5357 (changes requested):
What is the purpose of the `LivingSetAttackTargetEvent` now? It seems like you can just check whether there is a change through `LivingChangeTargetEvent` by scheduling on a lower priority.

Budschie:
> What is the purpose of the `LivingSetAttackTargetEvent` now? It seems like you can just check whether there is a change through `LivingChangeTargetEvent` by scheduling on a lower priority.

The purpose of the `LivingSetAttackTargetEvent` is mostly to provide binary compatibility to previous Forge versions to avoid introducing breaking changes. Thus, I should probably deprecate the event as it can be considered obsolete because of the `LivingChangeTargetEvent`.

src/main/java/net/minecraftforge/event/entity/living/LivingChangeTargetEvent.java
Budschie: Fixed: 214cbd5abc57e331cc7397b1e753bf76a7de625e

src/main/java/net/minecraftforge/event/entity/living/LivingChangeTargetEvent.java
Budschie: Fixed: c8274622fe30303ec1028dfd70d3841605bb9459

src/main/java/net/minecraftforge/event/entity/living/LivingSetAttackTargetEvent.java
Budschie: Fixed: c8274622fe30303ec1028dfd70d3841605bb9459

src/main/java/net/minecraftforge/event/entity/living/LivingSetAttackTargetEvent.java
Budschie: Fixed: d8371baa7a87a75a24e43080473f41f7cb0d75ba

src/main/java/net/minecraftforge/event/entity/living/LivingSetAttackTargetEvent.java
```diff
@@ -7,36 +7,64 @@
...
+ * {@link #getTarget} returns the newly targeted Entity.<br>
+ * {@link #getTargetType()} returns the target type that caused the change of targets.<br>
  * <br>
  * This event is not {@link net.minecraftforge.eventbus.api.Cancelable}.<br>
  * <br>
  * This event does not have a result. {@link HasResult}<br>
  * <br>
  * This event is fired on the {@link MinecraftForge#EVENT_BUS}.
  **/
+@Deprecated
```
ChampionAsh5357: Need to provide version number of deprecation (1.19.2) and set `forRemoval` to true.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -268,6 +270,19 @@ public static void onLivingSetAttackTarget(LivingEntity entity, LivingEntity tar
         MinecraftForge.EVENT_BUS.post(new LivingSetAttackTargetEvent(entity, target));
```
ChampionAsh5357, line 272: Need to deprecate the above method as well since the type is no longer present.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -268,6 +270,19 @@ public static void onLivingSetAttackTarget(LivingEntity entity, LivingEntity tar
...
     }
 
+    public static void onLivingSetAttackTarget(LivingEntity entity, LivingEntity target, ILivingTargetType targetType)
```
ChampionAsh5357, line 277: If event is marked for removal, methods should be as well.

patches/minecraft/net/minecraft/world/entity/Mob.java.patch
```diff
@@ -1,10 +1,15 @@
...
+    }
  
     public void m_6710_(@Nullable LivingEntity p_21544_) {
-       this.f_21362_ = p_21544_;
-+      net.minecraftforge.common.ForgeHooks.onLivingSetAttackTarget(this, p_21544_);
+-      this.f_21362_ = p_21544_;
++      net.minecraftforge.event.entity.living.LivingChangeTargetEvent changeTargetEvent = net.minecraftforge.common.ForgeHooks.onLivingChangeTarget(this, p_21544_, net.minecraftforge.event.entity.living.LivingChangeTargetEvent.LivingTargetType.MOB_TARGET);
++      if(!changeTargetEvent.isCanceled()) {
++          this.f_21362_ = changeTargetEvent.getNewTarget();
++          net.minecraftforge.common.ForgeHooks.onLivingSetAttackTarget(this, this.f_21362_);
```
ChampionAsh5357: Since this is being marked for removal, should add a `TODO: Remove 1.20`

patches/minecraft/net/minecraft/world/entity/ai/behavior/StartAttacking.java.patch
```diff
@@ -0,0 +1,16 @@
...
+    }
+ 
+    public static <E extends Mob> void m_24213_(E p_24214_, LivingEntity p_24215_) {
+-      p_24214_.m_6274_().m_21879_(MemoryModuleType.f_26372_, p_24215_);
+-      p_24214_.m_6274_().m_21936_(MemoryModuleType.f_26326_);
++      net.minecraftforge.event.entity.living.LivingChangeTargetEvent changeTargetEvent = net.minecraftforge.common.ForgeHooks.onLivingChangeTarget(p_24214_, p_24215_, net.minecraftforge.event.entity.living.LivingChangeTargetEvent.LivingTargetType.BEHAVIOR_TARGET);
++      if(!changeTargetEvent.isCanceled()) {
++          p_24214_.m_6274_().m_21879_(MemoryModuleType.f_26372_, changeTargetEvent.getNewTarget());
++          p_24214_.m_6274_().m_21936_(MemoryModuleType.f_26326_);
++          net.minecraftforge.common.ForgeHooks.onLivingSetAttackTarget(p_24214_, changeTargetEvent.getNewTarget(), net.minecraftforge.event.entity.living.LivingChangeTargetEvent.LivingTargetType.BEHAVIOR_TARGET);
```
ChampionAsh5357: Since this is being marked for removal, should add a `TODO: Remove 1.20`

ChampionAsh5357 (changes requested):
I'll get around to testing it later.

src/main/java/net/minecraftforge/event/entity/living/LivingSetAttackTargetEvent.java
Budschie: Out of interest, if I wanted to port this PR back to 1.18.2, should I just use 1.19.2 as a version of deprecation?

patches/minecraft/net/minecraft/world/entity/ai/behavior/StartAttacking.java.patch
Budschie: Just to reassure that I understood you correctly: Should I add that comment to the patch (or really any occurrence of firing the `LivingSetAttackTargetEvent`) *and* the event class or should I add this comment just to the event class?

Budschie: I also have a little code style question. When adding that normal comment to the LivingSetAttackTargetEvent, should I put it above the JavaDoc or beneath it?

src/main/java/net/minecraftforge/event/entity/living/LivingSetAttackTargetEvent.java
ChampionAsh5357: You would change the version of deprecation in the backport to 1.18.2.

patches/minecraft/net/minecraft/world/entity/ai/behavior/StartAttacking.java.patch
ChampionAsh5357: > Just to reassure that I understood you correctly: Should I add that comment to the patch (or really any occurrence of firing the LivingSetAttackTargetEvent) and the event class or should I add this comment just to the event class?

I would add the comment to the event class and calls to the event class since it's easier to double check that everything should be removed.

> I also have a little code style question. When adding that normal comment to the LivingSetAttackTargetEvent, should I put it above the JavaDoc or beneath it?

Same line as the deprecation annotation imo.

Budschie: Ok, I've now added the TODOs and the deprecations with commit 6c510de6a5867b7c612eef839fbd7874dce9fcaf.

patches/minecraft/net/minecraft/world/entity/Mob.java.patch
```diff
@@ -1,10 +1,16 @@
...
  
     public void m_6710_(@Nullable LivingEntity p_21544_) {
-       this.f_21362_ = p_21544_;
-+      net.minecraftforge.common.ForgeHooks.onLivingSetAttackTarget(this, p_21544_);
+-      this.f_21362_ = p_21544_;
++      net.minecraftforge.event.entity.living.LivingChangeTargetEvent changeTargetEvent = net.minecraftforge.common.ForgeHooks.onLivingChangeTarget(this, p_21544_, net.minecraftforge.event.entity.living.LivingChangeTargetEvent.LivingTargetType.MOB_TARGET);
++      if(!changeTargetEvent.isCanceled()) {
++          this.f_21362_ = changeTargetEvent.getNewTarget();
++          // TODO: Remove in 1.20
++          net.minecraftforge.common.ForgeHooks.onLivingSetAttackTarget(this, this.f_21362_);
```
ChampionAsh5357: Inline the message, it doesn't need to be a separate line to patch in:
`<code here> //TODO: ...`

Budschie: I've now inlined the TODO comments in commit a79ed2d588c0d53ef17d43409000a20a4c2c9018.

ChampionAsh5357 (approved):

PR #8955, opened 2022-08-10
[1.19.2] Make tryEmptyContainer respect doDrain 

Edivad99:
This #8318 was not brought into 1.19


sciwhiz12 (approved):

LexManos:
Discussed internally, going to deny. Tempted to revert  8318 because the comment and code explicitly are designed/written to NOT respect simulate. We'll be talking internally about a fluid system change in future versions.

PR #8956, opened 2022-08-10
[1.19.x] add methods with more context to tree growers

Minecraftschurli:
This PR adds a method in `AbstractTreeGrower` and `AbstractMegaTreeGrower` that allows for more control and the usage of RegistryAccess in the grower.
This PR is not intending to fix `MC-251744`

### DIFF
--- a/patches/minecraft/net/minecraft/world/level/block/grower/AbstractMegaTreeGrower.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/block/grower/AbstractMegaTreeGrower.java.patch
@@ -1,6 +1,6 @@
 --- a/net/minecraft/world/level/block/grower/AbstractMegaTreeGrower.java
 +++ b/net/minecraft/world/level/block/grower/AbstractMegaTreeGrower.java
-@@ -26,10 +_,16 @@
+@@ -26,10 +_,17 @@
     }
  
     @Nullable
@@ -9,6 +9,7 @@
 +   }
 +
 +   /** @deprecated Forge: Use {@link #getConfiguredMegaFeature(ServerLevel, ChunkGenerator, BlockPos, BlockState, RandomSource)} instead */
++   @Deprecated
 +   @Nullable
     protected abstract Holder<? extends ConfiguredFeature<?, ?>> m_213566_(RandomSource p_222904_);
  
--- a/patches/minecraft/net/minecraft/world/level/block/grower/AbstractMegaTreeGrower.java.patch
+++ b/patches/minecraft/net/minecraft/world/level/block/grower/AbstractMegaTreeGrower.java.patch
@@ -4,17 +4,17 @@
     }
  
     @Nullable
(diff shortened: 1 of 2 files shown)


eerussianguy:
Typically forge added methods appear below the rest of the class, and have a `/*** FORGE START ***/` label, but in this case it may not be such a big deal.

Minecraftschurli:
@eerussianguy it is intentionally placed there so the patch is only one block and one line and not scattered throughout the file

ChampionAsh5357:
Clarifying question, why would having the granularity control in the grower be better than using the granular control the block already provides when calling the grower?

Minecraftschurli:
@ChampionAsh5357 as stated in the PR description it allows access to the registry access which is necessary if you dont register the feature in the built-in registry but instead datagen it (vanilla uses the hardcoded references to the default values)
the forwarding of the other parameters was suggested by Commoble on [discord](https://discord.com/channels/313125603924639766/983834532904042537/1006972010062172221)

patches/minecraft/net/minecraft/world/level/block/grower/AbstractMegaTreeGrower.java.patch
```diff
@@ -0,0 +1,21 @@
+--- a/net/minecraft/world/level/block/grower/AbstractMegaTreeGrower.java
++++ b/net/minecraft/world/level/block/grower/AbstractMegaTreeGrower.java
+@@ -26,10 +_,17 @@
+    }
+ 
+    @Nullable
++   protected Holder<? extends ConfiguredFeature<?, ?>> getConfiguredMegaFeature(ServerLevel level, ChunkGenerator chunkGenerator, BlockPos pos, BlockState state, RandomSource random) {
++      return m_213566_(random);
```
ChampionAsh5357, line 10: Should have `// Forge: ` markers to indicate it is Forge-added.

patches/minecraft/net/minecraft/world/level/block/grower/AbstractTreeGrower.java.patch
```diff
@@ -0,0 +1,21 @@
+--- a/net/minecraft/world/level/block/grower/AbstractTreeGrower.java
++++ b/net/minecraft/world/level/block/grower/AbstractTreeGrower.java
+@@ -13,10 +_,17 @@
+ 
+ public abstract class AbstractTreeGrower {
+    @Nullable
++   protected Holder<? extends ConfiguredFeature<?, ?>> getConfiguredFeature(ServerLevel level, ChunkGenerator chunkGenerator, BlockPos pos, BlockState state, RandomSource random, boolean hasFlowers) {
++      return m_213888_(random, hasFlowers);
```
ChampionAsh5357, line 10: Should have `// Forge: ` markers to indicate it is Forge-added.

ChampionAsh5357 (changes requested):
After talking it over with other triages, there should be markers on the newly added method to indicate why it was added.
Other than that, the PR looks fine.

ChampionAsh5357 (approved):

PR #8958, opened 2022-08-11
[1.19.2] Add the ability to handle custom stack aware item cooldowns

Sirttas:
This is a reopen of #8013 for 1.19 as I didn't had time to update it then. This PR goal is to offers the ability for mods to add custom cooldown handling on their items. To do so it adds two new methods IForgeItem#isOnCooldown and IForgeItem#getCooldownPercent that expose the vanilla cooldown behavior.

### DIFF
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
@@ -828,4 +828,29 @@ public interface IForgeItem
         return self().getFoodProperties();
     }
 
+    /**
+     * Check if this item is on cooldown for an entity
+     *
+     * @param entity the entity holding the item
+     * @param stack the ItemStack
+     * @return if this item is on cooldow
+     */
+    default boolean isOnCooldown(@NotNull Entity entity, @NotNull ItemStack stack)
+    {
+        return this.getCooldownPercent(entity, stack, 0) > 0;
+    }
+
+    /**
+     * get this item cooldown for an entity
+     *
+     * @param entity the entity holding the item
+     * @param stack the ItemStack
+     * @param partialTick the partial render tick
+     * @return this item cooldow  percent
+     */
+    default float getCooldownPercent(@NotNull Entity entity, @NotNull ItemStack stack, float partialTick)
+    {
+        return entity instanceof Player player ? player.getCooldowns().getCooldownPercent(self(), partialTick) : 0;
+    }
+
 }
--- a/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
@@ -829,28 +829,29 @@ public interface IForgeItem
     }
 
     /**
-     * Check if this item is on cooldown for an entity
+     * Check if this item is on cooldown for a player
      *
-     * @param entity the entity holding the item
+     * @param player the player holding the item
      * @param stack the ItemStack
-     * @return if this item is on cooldow
+     * @return if this item is on cooldown
      */
-    default boolean isOnCooldown(@NotNull Entity entity, @NotNull ItemStack stack)
+    default boolean isOnCooldown(@NotNull Player player, @NotNull ItemStack stack)
     {
-        return this.getCooldownPercent(entity, stack, 0) > 0;
+        return this.getCooldownPercent(player, stack, 0) > 0;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -828,4 +828,29 @@ default FoodProperties getFoodProperties(ItemStack stack, @Nullable LivingEntity
...
+     * get this item cooldown for an entity
+     *
+     * @param entity the entity holding the item
+     * @param stack the ItemStack
+     * @param partialTick the partial render tick
+     * @return this item cooldow  percent
+     */
+    default float getCooldownPercent(@NotNull Entity entity, @NotNull ItemStack stack, float partialTick)
+    {
+        return entity instanceof Player player ? player.getCooldowns().getCooldownPercent(self(), partialTick) : 0;
```
ChampionAsh5357: Cooldowns only exist for the player, so why make it for an entity?

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
```diff
@@ -828,4 +828,29 @@ default FoodProperties getFoodProperties(ItemStack stack, @Nullable LivingEntity
...
+     * @param stack the ItemStack
+     * @return if this item is on cooldow
+     */
+    default boolean isOnCooldown(@NotNull Entity entity, @NotNull ItemStack stack)
+    {
+        return this.getCooldownPercent(entity, stack, 0) > 0;
+    }
+
+    /**
+     * get this item cooldown for an entity
```
ChampionAsh5357: What is the item cooldown? This is not properly documented.

ChampionAsh5357 (changes requested):
So, the main problem of this system has to do with syncing and data storage. Item cooldowns are synced on start and finish which would break and become essentially useless if someone decides to do something custom.

For data storage, this seems to put some of the onus on the item to properly handle and track the cooldown. It is highly likely that this would just cause some modder to implement a timer on their own `Item` instead of making the necessary capability to properly override this system.

Additionally, the `#getCooldownPercent` method is weird in general since it's used in common code for checking if a cooldown exists yet contains a client only parameter.

src/main/java/net/minecraftforge/common/extensions/IForgeItem.java
Sirttas: Because this is the vanilla behavior. It does have a `partialTick`, but it's only used in `ItemRenderer#renderGuiItemDecorations`.

Sirttas: Whoops this answer wasn't meant to be here...

Sirttas:
> So, the main problem of this system has to do with syncing and data storage. Item cooldowns are synced on start and finish which would break and become essentially useless if someone decides to do something custom.

I think it's the responsibility of the mod dev to sync the property they use to handle the cooldown.

> For data storage, this seems to put some of the onus on the item to properly handle and track the cooldown. It is highly likely that this would just cause some modder to implement a timer on their own Item instead of making the necessary capability to properly override this system.

This is something that could happen, but it's not how it should be done. I don't know how I could implements this without completely rewriting the cooldown system. The solution I suggest is to properly document that modders should manage it using a properly synced capability attached to the player. But yeah, I know this could be a deal breaker, and I understand if it would result in the PR getting closed, I'm open to any suggestion on this.

> Additionally, the #getCooldownPercent method is weird in general since it's used in common code for checking if a cooldown exists yet contains a client only parameter.

Because this is the vanilla behavior. It does have a `partialTick`, but it's only used in `ItemRenderer#renderGuiItemDecorations`.

ChampionAsh5357:
Based on some discussion, at a minimum, we need a full writeup on what will need to be done on the modder's part to use the methods effectively. This should also be reflected in the documentation of the methods. There also needs to be a test mod for this feature. Then, there will probably be some more discussion on whether to keep moving towards assigning, take a different path, or simply reject the feature altogether. The concept seems very complex and finicky for modders to use in its current state, otherwise.

Sirttas:
After a bit of thinking I don't think it is the correct implementation for it. I went for this one because it was the simple one to do, and because I already set up everything required on my mod so it felt simple to do. The question is now should I open an issue (and close the PR) to discuses it and the possible implems for it or simply close the PR. The thing is I don't think a lot of people have this need other than me so I don't think it is a good idea to do a lot of changes in forge just for that.

PR #8963, opened 2022-08-13
Allow FakePlayer to report its position

Shadows-of-Fire:
FakePlayers currently cannot report their position through position() and blockPosition(), but can report it through posX()/posY()/posZ().

This means that fakeplayers cannot be used in contexts where the position of the player matters, and the underlying access to the position is through position() or blockPosition().
This limitation is fairly artificial, and I don't think there's a compelling reason for the position to be hidden.

### DIFF
--- a/src/main/java/net/minecraftforge/common/util/FakePlayer.java
+++ b/src/main/java/net/minecraftforge/common/util/FakePlayer.java
@@ -81,8 +81,6 @@ public class FakePlayer extends ServerPlayer
         this.connection = new FakePlayerNetHandler(level.getServer(), this);
     }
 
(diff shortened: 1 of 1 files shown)


pupnewfster (approved):

LexManos:
What cases would position matter?

Shadows-of-Fire:
My specific case is I'm attempting to break a block while holding an item that causes block drops to be teleported to the breaking player.  The break is done with a fakeplayer as to not cause a bunch of re-equip animations / hunger usage / etc to the "real" player, but since the drops attempt to teleport to the player, they teleport to the fake player at `(0, 0, 0)`.

PR #8971, opened 2022-08-15
Give IItemHandler control over multi-item slotless inserts

StevenDoesStuffs:
Currently, `IItemHandler` only has one insertion method, which requires a slot to be specified. This PR adds two additional methods to `IItemHandler` which take a **list** of itemstacks and **doesn't** take a slot. They're almost the same method, except one distributes items by filling the lower slots first (`insertItems`), whereas the other fills existing itemstacks first (`insertItemsStacked`). 

Default implementations are provided so **this is not a breaking change**. 

### Slotless insert

Currently, the way to do slotless insert is to use `ItemHandlerHelper#insertItem`. The way this method works is by looping through all the slots in the inventories and trying to insert the itemstack into each of the slots. Unfortunately, this means that on "non-standard" inventories, simulation isn't guaranteed to be accurate. This is because doing an insertion invalidates all of the previous simulation results. 

Consider the Create basin, which allows any given type of item to occupy only one slot, and each slot can only store 16 items (instead of 64). This means we couldn't have two slots both containing cobblestone for example. Using `ItemHandlerHelper#insertItem` to simulate inserting 32 cobblestone will return that all 32 cobblestone get inserted, which is wrong because only 16 cobblestone can actually be inserted. This is because after inserting the first 16 cobblestone into the first available slot, the basin will no longer take cobblestone in other slots. 

This change would allow non-standard inventories to override the implementation of slotless insert. 

### All-or-nothing insertions

There are some mods which want to do some sort of all-or-nothing insertion, where you have multiple itemstacks, and you only want to actually perform an insert if all of the itemstacks can be fully inserted. A typical use case for this would be autocrafting, where you'd only want to insert the ingredients into a machine if it can accept all of the ingredients. 

Currently, there is no way to do this which is provided by forge, but I assume the standard way to do this would be to loop through the slots, simulating insertion on all of the remaining itemstacks and seeing if there are any left at the end. However, this isn't guaranteed to be accurate, again because doing an insertion invalidates all of the previous simulation results. 

For example, consider an expanding inventory which always reports that it has one empty slot. If I attempt to insert a stack of cobblestone and then a stack of dirt, it will succeed, since the inventory will expand to be able to take the second stack after I insert the first stack. But if I do the simulation as described previously, it'll report that I can only insert one stack of cobblestone, which would cause the all-or-nothing operation to fail. I have a more realistic example if you'd like, but it's a bit more difficult to explain. 

This change would also allow non-standard inventories to override the implementation transaction simulation. 

## Other Changes

I also added a bit to the documentation of `IItemHandler` saying that simulation should be accurate and idempotent. Idempotent means that attempting to insert the remaining items of another insertion right after it happens should return the same remaining items. 

The appropriate `ItemHandlerHelper` methods have been updated to use the new methods in `IItemHandler`.

## Design choices

One thing I'm not really sure about is whether `IItemHandler` should have the `insertItemsStacked` method. It feels like such a small detail and makes the interface more bloated, but allowing inventories to override that behavior is nice, and it doesn't hurt that much to keep it there.

Another thing is taking lists of itemstacks instead of arrays. I chose this because it's fast to convert an array to a list using `Arrays.aslist`, but not the other way around since it almost always requires a copy.

EDIT: change example from two stacks of cobblestone to one stack of cobble + one stack of dirt.


lukebemish:
Your example is doesn't quite make sense to me:
> Using `ItemHandlerHelper#insertItem` to simulate inserting 32 cobblestone will return that all 32 cobblestone get inserted, which is wrong because only 16 cobblestone can actually be inserted. This is because after inserting the first 16 cobblestone into the first available slot, the basin will no longer take cobblestone in other slots.

This seems to mean that Create doesn't implement their IItemHandler correctly. Their IItemHandler should, when simulating, tell the caller that only 16 items will be inserted.

lukebemish:
In the case of an "all-or-nothing" insertion, the IItemHandler should handle all of this on its own. For instance, if the whole stack can't be inserted, it would return that *none* of it can be inserted. (Please correct me if I've misunderstood the issue here)

StevenDoesStuffs:
> Your example is faulty:
> 
> > Using `ItemHandlerHelper#insertItem` to simulate inserting 32 cobblestone will return that all 32 cobblestone get inserted, which is wrong because only 16 cobblestone can actually be inserted. This is because after inserting the first 16 cobblestone into the first available slot, the basin will no longer take cobblestone in other slots.
> 
> This means that Create doesn't implement their IItemHandler correctly. Their IItemHandler should, when simulating, tell the caller that only 16 items will be inserted.

Their IItemHandler is correct, but the problem is that `ItemHandlerHelper` makes some invalid assumptions about IItemHandler.

StevenDoesStuffs:
> In the case of an "all-or-nothing" insertion, the IItemHandler should handle all of this on its own. For instance, if the whole stack can't be inserted, it would return that _none_ of it can be inserted. (Please correct me if I've misunderstood the issue here)

The difficulty is with multi-itemstack all-or-nothing operations, not single stack ones. Also, IItemHandler doesn't handle all-or-nothing insertions. Currently, it only handles best effort single slot insertions, where it attempts to insert as much as possible of an itemstack it's given into a single slot. `IItemHandler#insertItem` then returns the remaining items. 

One way you could do single stack all-or-nothing insertions is to run `IItemHandler#insertItem` in simulation mode, and see if it returns that nothing remains (i.e all of it was inserted). Then if that's the case, do the insertion for real. However, this doesn't work with multiple stacks as of now (which is what the point of the PR is).

lukebemish:
I'm still not convinced that this is necessary, though obviously my opinion means very little. I think I'm seeing what you've done a bit clearer now: my argument is, basically, that this should be implemented by IItemHandlers. And it *can* be implemented there. There's nothing stopping a mod from doing this:
* thing A asks to insert 2 stacks of cobblestone into B at partially empty slot C
* B checks if it will be able to fit all 2 stacks in its inventory, by expanding slots or whatever as necessary
* B returns an appropriate, and accurate simulation
* C chooses to insert or not based on that.

StevenDoesStuffs:
Well... yes IItemHandler should handle those things, but currently it can't. The only insert method IItemHandler has takes only one itemstack and requires a slot to be specified. I should mention that the two stacks of cobblestone is kinda a bad example, what I really mean is stuff like one stack of cobblestone and one stack of dirt for example.

StevenDoesStuffs:
Also, 
> B checks if it will be able to fit all 2 stacks in its inventory, by expanding slots or whatever as necessary

But you specified a slot, which means it will only attempt to insert into that one slot, which can obviously not fit two separate itemstacks.

lukebemish:
Just because you try to insert into one slot, does not mean that the internal implementation of IItemHandler that that mod has will be inserting into only one slot. I do see your point about differing itemstacks; mixed itemstack insertions are something that can't currently be implemented, no. If all the items being inserted are the same type of item, then you could just have one giant stack of stuff that's larger than it's allowed to be.

StevenDoesStuffs:
For those reading this discussion after-the-fact, we've moved to discord ([link](https://canary.discord.com/channels/313125603924639766/852298000042164244/1008852589368070385))

src/main/java/net/minecraftforge/items/IItemHandler.java
```diff
@@ -61,12 +76,134 @@
...
+     *
+     * @param stacks   List of items to insert. Contents must not be null.
+     * @param simulate If true, the insertion is only simulated
+     * @return An array with the same length as the input list representing the remaining itemstacks.
+     *         Each element of the array has the same properties as the return value of
+     *         {@link IItemHandler#insertItem(int, ItemStack, boolean)}
+     *         (non-null, could be the same as its corresponding element in input list, etc.).
+     */
+    @NotNull
+    default ItemStack[] insertItems(@NotNull List<ItemStack> stacks, boolean simulate)
```
Shadows-of-Fire, line 101: Why the mixing of lists and arrays?  This should probably be both `List<ItemStack>` or both `ItemStack[]`.

StevenDoesStuffs: There is a method to turn an array into a list but not the other way around, so taking a list but returning array is the most flexible. Also, returning an array signifies it's "fixed size", since in our case each index in the array corresponds to what remains of the index in the list. 

Anyhow, I can totally change it though if you'd like.

src/main/java/net/minecraftforge/items/IItemHandler.java
```diff
@@ -48,8 +52,19 @@
      * <p>
      * Inserts an ItemStack into the given slot and return the remainder.
      * The ItemStack <em>should not</em> be modified in this function!
+     * Note: This behaviour is subtly different from {@link IFluidHandler#fill(FluidStack, IFluidHandler.FluidAction)}.
+     * </p>
+     * <p>
+     * insertItem must have the four following properties.
+     * First, the simulation mode must not mutate the inventory.
+     * Second, the simulation mode must be accurate, i.e insertItem(slot, stack, true)
```
Technici4n, line 60: This is unfortunately not true, and mods need to deal with that case. Here is an example of an item handler that doesn't follow this guarantee: https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/c0a9dd43df29f6edce051835bb9d662600497d80/src/main/java/appeng/parts/p2p/ItemP2PTunnelPart.java#L69.

(explanation: if an input P2P tunnel is connected to multiple output P2P tunnels on the same inventory, then simulation might allow inserting more than execution).

src/main/java/net/minecraftforge/items/IItemHandler.java
```diff
@@ -48,8 +52,19 @@
...
+     * must be equivalent to insertItem(slot, stack, false).
+     * Third, real insertion must be idempotent, i.e insertItem(slot, insertItem(slot, stack, false), false)
```
Technici4n, line 62: This is not true either, as some item handlers might want to rate-limit the insertion. Here is an example of a **fluid** handler doing it, but I see no reason why item handlers shouldn't be allowed to do the same: https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/c0a9dd43df29f6edce051835bb9d662600497d80/src/main/java/appeng/blockentity/misc/CondenserBlockEntity.java#L268.

src/main/java/net/minecraftforge/items/IItemHandler.java
Technici4n, line 101: This is a weird way of doing it, that doesn't solve the issue of simulations possible allowing more than executions (see my comment above). Correctness in that sense cannot be achieved using the current `IItemHandler` interface, or more generally it cannot be achieved using the `boolean simulate` pattern.

However, this can still be useful from a performance point of view, but in that case the method should take a single `ItemStack` instead of taking an array. See  #8244. This would also solve some correctness issues with for example the infamous storage drawers "fake input slot".

PR #8977, opened 2022-08-17
General Code improvements

JT122406:
Massive Code improvements that should lead to it being faster and using less memory


zml2008:
> being faster and using less memory

these are both measurable metrics - can you supply profiling results to defend this assertion?

ChampionAsh5357:
A lot of the changes here violates Forge's style guide (e.g. new line braces). The changes also seem inconsistent and not standardized by any measure.

Additionally, similar to zml's statement, most of these changes don't seem like they would end up affecting the compiled bytecode and are purely syntax. If you want to make claims about improvements, there needs to be evidence of this.

marchermans:
> Massive Code improvements that should lead to it being faster and using less memory

Please list every change you made with regards to this.
Also please revert all formatting changes which do not reflect the Forge programming scheme.

sekwah41:
I had nothing better to do on the train so I scrolled through every change because I was curious.

There is literally nothing in this pr that I could see making a slight change in performance. It just looks like its been run through an automated script to reformat parts, remove unused imports and replace certain calls with "cleaner" alternatives e.g. a foreach.

Anywhere there would actually be any byte code differences the performance would be exactly the same.

I'm not sure I would list most of these as "Code improvements". One of the sections has a series of - 0 values to make it align better as well as is there for the sake of readability which your pr removes which further points to this being an almost entirely automated pr.

There would be minimal byte code changes though those wouldn't actually cause any performance benefit, and I can almost guarantee there would be little/no memory improvements from what I saw.

sciwhiz12:
I echo the statements made above. To enumerate some my concerns, this PR:
- makes unsubstantiated claims about the effect of its changes on the performance and memory usage;
- reformats parts of many classes without regard for the Forge code style (specifically, braces on newlines);
- removes unused imports _en masse_ likely through automated means;
- removes certain seemingly useless code which are present to improve the code's readability; and
- lumps most of its changes in a _single_ commit, such as but not limited to:
  - reformatting full lambda expressions to single-statement lambda expressions;
  - cleaning up explicit generics to use the diamond operator;
  - rewriting certain conditionals and `if`s to be shorter; and
  - switching out idiomatic `instanceof`-and-cast with pattern-matching `instanceof`.

Quite frankly, this PR seems like the result of running automated scripts, perhaps IDEA inspections, over the codebase and committing that, with minimal human oversight on the changes made therein. (I do believe all of these changes except perhaps the Gradle version upgrade are available as IDEA inspections.)

There is no justification given here for the massive amount of changes, aside from the meager title and the unsubstantiated claim in the PR description about performance improvements. The large amount of changes done by this PR is clumped mostly in a single mega-commit. As the CI reports, this was not even tested to be _compilable_. A reasonable person would expect someone performing large blocks of cleanup to _at the very least_ ensure the code is left in a compilable state before pushing and making a PR.

Given the above reasons, I am closing this PR.

We have plans to draft and published a code style guide to properly standardize the Forge code style, for the benefit of contributors and maintainers alike. Once we have made and published such a guide and other related resources (such as IDE autoformatter files for contributor convenience), we will then embark on cleanup on the whole codebase in a thorough and comprehensive manner.

Until then, it should be noted that we will not accept cleanup that involves either reordering and cleaning up unused imports or reformatting the codebase (especially including those changes that make existing code violate the general code style convention we have).

PR #8978, opened 2022-08-18
IS_DESERT Biome Tag

JT122406:
Is useful for datapacks and mods that add other desert biomes to have a tag so that structures can spawn along with other things like mobs.


eerussianguy:
Why so many white space changes? And why not use the dry overworld tag?

PR #8979, opened 2022-08-18
IS_DESERT Tag

JT122406:
Is useful for datapacks and mods that add other desert biomes to have a tag so that structures can spawn along with other things like mobs.

PR #8980, opened 2022-08-18
Fix Mangrove Swamp tag code

JT122406:
Remove IS_WET AND IS_HOT tags from Mangrove swamp as they are included when using IS_WET_OVERWORLD and IS_HOT_OVERWORLD and don't need to be explicitly written out


Sirttas:
This does not make sense at all, mangroves are wet and hot, plus this is kind of a breaking change, plus you didn't run datagen.

JT122406:
> This does not make sense at all, mangroves are wet and hot, plus this is kind of a breaking change, plus you didn't run datagen.

First of all it is not breaking second of all they are still wet and hot as they have the IS_WET_OVERWORLD and IS_HOT_OVERWORLD tags that automatically adds them to IS_WET and IS_HOT so there is not reason to have them explicitly written into those two since they are automatically added to thoses tags.
And third I ran data but whoever Ran data after making this change must not have because no files were updated after running meaning when someone added these tags they didn't run data so removing them leave the data the same

JT122406:
and if you actually looked at the commit you would see that

TelepathicGrunt:
the is_hot and is_wet tags pulls entries from is_hot/overworld and is_wet/overworld. Just clarifying why they were trying to making it less redundant

TelepathicGrunt (approved):

Sirttas:
Sorry I didn't see it was just a cleanup...

JT122406:
> Sorry I didn't see it was just a cleanup...

All good, sorry for the aggression

sciwhiz12 (approved):
Noting that this merely fixes the code to reflect the actual tags. Indeed, datagenning without this fix causes the mangrove swamp to be added to the `is_wet` and `is_hot` biome tags.

JT122406:
If you guys want I can merge this branch into #8979 So its only 1 PR and one build but up to y'all just lmk here

PR #8988, opened 2022-08-20
Update Gradle to 7.5.1

JT122406:
Fixes some major bugs in 7.5 at the bottom of the release notes it says "We recommend users upgrade to 7.5.1 instead of 7.5."
So we should move to this Version and has no breaking changes
ChangeLog: https://docs.gradle.org/7.5.1/release-notes.html


JT122406:
Not even a reason
Nice

JT122406:
Maybe my logic is just too good for a reason

sciwhiz12:
We do not accept PRs which upgrade the Gradle wrapper version in general. 

We will naturally update the Gradle wrapper during the MC version update and porting process, or when we are affected by a bug which is fixed by a newer version, or when we move to a newer ForgeGradle version which requires a newer Gradle version.

PR #8989, opened 2022-08-21
Add mod mismatch event

robotgryphon:
This event aims to cover a gap where individual World Saved Data files and DataFixer cannot cover - handling data changes across multiple distributed files. It does so by collecting mod version data from FML (the handler that writes warnings to the log about version mismatches) and then firing a collected event about the version differences.

This allows for multiple ways of fixing persisted data without having to introduce complex patterns such as DataFixerUpper or manually run commands. It also, by nature of allowing any mod to hook the event, allows for dedicated "migrator" mods to exist to cover gaps that mod authors themselves cannot or will not cover.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -56,6 +56,7 @@ import net.minecraft.world.item.crafting.RecipeType;
 import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.gameevent.GameEvent;
 import net.minecraft.world.level.material.Fluids;
+import net.minecraft.world.level.storage.LevelStorageSource;
 import net.minecraft.world.level.storage.WorldData;
 import net.minecraft.world.level.storage.loot.LootContext;
 import net.minecraft.world.level.storage.loot.LootTable;
@@ -108,6 +109,7 @@ import net.minecraftforge.event.AnvilUpdateEvent;
 import net.minecraftforge.event.DifficultyChangeEvent;
 import net.minecraftforge.event.ForgeEventFactory;
 import net.minecraftforge.event.ItemAttributeModifierEvent;
+import net.minecraftforge.event.ModVersionsMismatchEvent;
 import net.minecraftforge.event.ServerChatEvent;
 import net.minecraftforge.event.RegisterStructureConversionsEvent;
 import net.minecraftforge.event.VanillaGameEvent;
@@ -168,6 +170,8 @@ import net.minecraft.world.level.LevelAccessor;
 import net.minecraft.world.level.block.state.pattern.BlockInWorld;
 import net.minecraft.world.level.dimension.LevelStem;
 import net.minecraft.world.level.material.Fluid;
+import org.apache.maven.artifact.versioning.ArtifactVersion;
+import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -1319,12 +1323,13 @@ public class ForgeHooks
         levelTag.put("fml", fmlData);
     }
 
-    public static void readAdditionalLevelSaveData(CompoundTag rootTag)
+    public static void readAdditionalLevelSaveData(LevelStorageSource.LevelDirectory levelDirectory, CompoundTag rootTag)
     {
         CompoundTag tag = rootTag.getCompound("fml");
         if (tag.contains("LoadingModList"))
         {
             ListTag modList = tag.getList("LoadingModList", net.minecraft.nbt.Tag.TAG_COMPOUND);
+            Map<String, ArtifactVersion> mismatchedVersions = new HashMap<>(modList.size());
             for (int i = 0; i < modList.size(); i++)
             {
                 CompoundTag mod = modList.getCompound(i);
@@ -1334,17 +1339,27 @@ public class ForgeHooks
                     continue;
                 }
                 String modVersion = mod.getString("ModVersion");
-                Optional<? extends ModContainer> container = ModList.get().getModContainerById(modId);
-                if (container.isEmpty())
-                {
-                    LOGGER.error(WORLDPERSISTENCE,"This world was saved with mod {} which appears to be missing, things may not work well", modId);
-                    continue;
-                }
-                if (!Objects.equals(modVersion, MavenVersionStringHelper.artifactVersionToString(container.get().getModInfo().getVersion())))
-                {
-                    LOGGER.warn(WORLDPERSISTENCE,"This world was saved with mod {} version {} and it is now at version {}, things may not work well", modId, modVersion, MavenVersionStringHelper.artifactVersionToString(container.get().getModInfo().getVersion()));
-                }
+                ModList.get().getModContainerById(modId).ifPresentOrElse(container -> {
+                    final var loadingVersion = container.getModInfo().getVersion();
+                    final var previousVersion = new DefaultArtifactVersion(modVersion);
+                    if (!loadingVersion.equals(previousVersion))
+                    {
+                        // Enqueue mismatched versions for bulk event
+                        mismatchedVersions.put(modId, previousVersion);
+                    }
+                }, () -> LOGGER.error(WORLDPERSISTENCE,"This world was saved with mod {} which appears to be missing, things may not work well", modId));
             }
+
+            final var mismatchEvent = new ModVersionsMismatchEvent(levelDirectory, mismatchedVersions);
+            ModLoader.get().postEvent(mismatchEvent);
+
+            // For mods that did not specify handling, show a warning to users that errors may occur
+            mismatchedVersions.keySet().stream()
+                    .filter(modId -> mismatchEvent.getModState(modId) == ModVersionsMismatchEvent.MismatchHandlingState.UNHANDLED)
+                    .forEach(modId -> {
+                        LOGGER.warn(WORLDPERSISTENCE, "This world was saved with mod {} version {} and it is now at version {}, things may not work well",
+                                modId, mismatchedVersions.get(modId), mismatchEvent.getCurrentVersion(modId));
+                    });
         }
 
         Multimap<ResourceLocation, ResourceLocation> failedElements = null;
--- a/src/main/java/net/minecraftforge/common/ForgeHooks.java
+++ b/src/main/java/net/minecraftforge/common/ForgeHooks.java
@@ -17,6 +17,7 @@ import java.util.regex.Pattern;
 import com.google.common.base.Strings;
 import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Queues;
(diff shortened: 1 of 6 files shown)


src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1334,17 +1339,27 @@ public static void readAdditionalLevelSaveData(CompoundTag rootTag)
...
+                        // Enqueue mismatched versions for bulk event
+                        mismatchedVersions.put(modId, previousVersion);
+                    }
+                }, () -> LOGGER.error(WORLDPERSISTENCE,"This world was saved with mod {} which appears to be missing, things may not work well", modId));
             }
+
+            final var mismatchEvent = new ModVersionsMismatchEvent(levelDirectory, mismatchedVersions);
+            ModLoader.get().postEvent(mismatchEvent);
+
+            // For mods that did not specify handling, show a warning to users that errors may occur
```
sciwhiz12, line 1397: I wonder if we should have a debug logging message for mods which handled the version mismatch, so it is always recorded in the logs.

src/main/java/net/minecraftforge/event/ModVersionsMismatchEvent.java
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright (c) Forge Development LLC and contributors
+ * SPDX-License-Identifier: LGPL-2.1-only
+ */
+
+package net.minecraftforge.event;
+
+
```
sciwhiz12: Extra blank line should be removed.
```suggestion

```

src/main/java/net/minecraftforge/event/ModVersionsMismatchEvent.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * State values for mod version mismatches. Can be queried by any mod listening to the event.
+     */
+    public enum MismatchHandlingState
+    {
+        UNHANDLED,
+        NEEDS_FURTHER_WORK,
+        FULLY_HANDLED
```
sciwhiz12: Could you please explain this enum and the existence of `NEEDS_FURTHER_WORK`, rather than a simple boolean for whether it was handled or not?

src/main/java/net/minecraftforge/event/ModVersionsMismatchEvent.java
```diff
@@ -0,0 +1,121 @@
...
+    {
+        return Optional.ofNullable(this.previousVersions.get(modId));
+    }
+
+    /**
+     * Utility method to fetch current mod version information from the mod list.
+     * @param modId The mod to query.
+     * @return Current mod version information.
+     */
+    public ArtifactVersion getCurrentVersion(String modId)
```
sciwhiz12: I believe we generally avoid utility methods which are easily reimplementable by mods. Even then, a utility method should be `static` since it relies on no state in the event itself, or it should be in some other place like an instance method in `ModList` instead.

src/main/java/net/minecraftforge/event/ModVersionsMismatchEvent.java
```diff
@@ -0,0 +1,121 @@
...
+ * Fires when the mod loader is in the process of loading a world that was last saved
+ * with mod versions that differ from the currently-loaded versions. This can be used to
+ * enqueue work to run at a later point, such as multi-file migration of data.
+ * </p>
+ *
+ * <p>
+ * <b>Note that level and world information has not yet been fully loaded;</b> as such, it is
+ * unsafe to access server or level information during handling of this event.
+ * </p>
+ */
```
sciwhiz12: Suggestion for improvement and standardization of event javadocs:
```suggestion
/**
 * Fires when the mod loader is in the process of loading a world that was last saved
 * with mod versions that differ from the currently-loaded versions. This can be used to
 * enqueue work to run at a later point, such as multi-file migration of data.
 *
 * <p><strong>Note that level and world information has not yet been fully loaded</strong>; as such, it is
 * unsafe to access server or level information during handling of this event.</p>
 *
 * <p>This event is not {@linkplain Cancelable cancellable}, and does not {@linkplain HasResult have a result}.</p>
 *
 * <p>This event is fired on the {@linkplain FMLJavaModLoadingContext#getModEventBus() mod-specific event bus},
 * only on the {@linkplain LogicalSide#SERVER logical server}.</p>
 */
```
Please change the last paragraph if the event is not fired only on the logical server, such as if it is fired on both logical sides (`, on both {@linkplain LogicalSide logical sides}.`).

src/main/java/net/minecraftforge/event/ModVersionsMismatchEvent.java
```diff
@@ -0,0 +1,121 @@
...
+     */
+    private final Map<String, MismatchHandlingState> states;
+
+    /**
+     * Create an event instance to handle mod version mismatches.
+     *
+     * @param levelDirectory The directory of the level being loaded.
+     * @param previousVersions Map of mismatched mod ids and their previously-loaded version.
+     */
+    public ModVersionsMismatchEvent(LevelStorageSource.LevelDirectory levelDirectory, Map<String, ArtifactVersion> previousVersions)
```
sciwhiz12: The javadoc can be omitted, and the event constructor should be annotated with `@ApiStatus.Internal`.

sciwhiz12 (changes requested):

src/main/java/net/minecraftforge/event/ModVersionsMismatchEvent.java
robotgryphon: This is here so that mods can hook the event with a low priority and do additional changes/setup for the mismatches that haven't been fully handled. Since I'd rather not have Forge fire a second event later on (and thus requiring more state to be stored) this was the solution I went for.

src/main/java/net/minecraftforge/event/ModVersionsMismatchEvent.java
robotgryphon: I can look into moving this to ModList; there are a few places I've noticed this kind of code called

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1319,12 +1323,13 @@ public static void writeAdditionalLevelSaveData(WorldData worldData, CompoundTag
         levelTag.put("fml", fmlData);
     }
 
-    public static void readAdditionalLevelSaveData(CompoundTag rootTag)
+    public static void readAdditionalLevelSaveData(LevelStorageSource.LevelDirectory levelDirectory, CompoundTag rootTag)
```
LexManos: This needs bin compatibility as neither this, nor the class is marked as internal.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1334,17 +1339,27 @@ public static void readAdditionalLevelSaveData(CompoundTag rootTag)
...
-                if (container.isEmpty())
-                {
-                    LOGGER.error(WORLDPERSISTENCE,"This world was saved with mod {} which appears to be missing, things may not work well", modId);
-                    continue;
-                }
-                if (!Objects.equals(modVersion, MavenVersionStringHelper.artifactVersionToString(container.get().getModInfo().getVersion())))
-                {
-                    LOGGER.warn(WORLDPERSISTENCE,"This world was saved with mod {} version {} and it is now at version {}, things may not work well", modId, modVersion, MavenVersionStringHelper.artifactVersionToString(container.get().getModInfo().getVersion()));
-                }
+                ModList.get().getModContainerById(modId).ifPresentOrElse(container -> {
```
LexManos: It'd be better to keep the same style, not using ifPresentOrElse, as its far easier to debug, and it'd make your code diff cleaner.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1334,17 +1339,27 @@ public static void readAdditionalLevelSaveData(CompoundTag rootTag)
...
+                    final var loadingVersion = container.getModInfo().getVersion();
+                    final var previousVersion = new DefaultArtifactVersion(modVersion);
+                    if (!loadingVersion.equals(previousVersion))
+                    {
+                        // Enqueue mismatched versions for bulk event
+                        mismatchedVersions.put(modId, previousVersion);
+                    }
+                }, () -> LOGGER.error(WORLDPERSISTENCE,"This world was saved with mod {} which appears to be missing, things may not work well", modId));
```
LexManos: Would it not make sense to store the missing mods as well. Just to give the event full context? "World had ModX at V, but its now missing"

src/main/java/net/minecraftforge/common/ForgeHooks.java
LexManos: Debug information is good yes.

src/main/java/net/minecraftforge/event/ModVersionsMismatchEvent.java
LexManos: Since this is fired on the mod bus, we need to make sure that it's fired in a known state. Is this using the async executor, or the single threaded one? I would assume the single threadeed one as it'd be safer for mods to do work one at a time.

src/main/java/net/minecraftforge/event/ModVersionsMismatchEvent.java
LexManos: Honestly this being here is fine, as it keeps all the logic in one place. It also allows for us to return a known 'NOT HERE' value, instead of throwing, which I would prefer.

src/main/java/net/minecraftforge/event/ModVersionsMismatchEvent.java
LexManos: What are we, as Forge or an external modder supposed to do with this state?

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1319,12 +1326,24 @@ public static void writeAdditionalLevelSaveData(WorldData worldData, CompoundTag
...
+     * @deprecated To be removed in 1.20.
+     * Use {@link #readAdditionalLevelSaveData(LevelStorageSource.LevelDirectory, CompoundTag)} instead.
+     */
+    @Deprecated(forRemoval = true, since = "1.19.2")
+    public static void readAdditionalLevelSaveData(CompoundTag rootTag) {
+        readAdditionalLevelSaveData(null, rootTag);
+    }
+
+    @ApiStatus.Internal
+    public static void readAdditionalLevelSaveData(LevelStorageSource.LevelDirectory levelDirectory, CompoundTag rootTag)
```
SizableShrimp: This should have the level directory marked as Nullable, along with cascading that annotation to the event and any getters. Why exactly are you marking this new method as an internal api?

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1338,28 +1350,72 @@ public static void readAdditionalLevelSaveData(LevelStorageSource.LevelDirectory
...
 
+            StringBuilder resolved = new StringBuilder("The following mods have version differences that were marked resolved:");
+            StringBuilder unresolved = new StringBuilder("The following mods have version differences that were not resolved:");
+
+            // For mods that were marked resolved, log the version resolution and the mod that resolved the mismatch
+            mismatchEvent.getResolved().forEachOrdered((res) ->
+            {
+                final var modid = res.modid();
+                final var diff = res.versionDifference();
+                if(res.wasSelfResolved()) {
```
sciwhiz12: Formatting: space between `if` keyword and parenthesis.
```suggestion
                if (res.wasSelfResolved()) {
```

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1338,28 +1350,72 @@ public static void readAdditionalLevelSaveData(LevelStorageSource.LevelDirectory
...
+                    resolved.append(System.lineSeparator())
+                            .append(diff.isMissing() ?
+                                    "%s (version %s -> MISSING, self-resolved)".formatted(modid, diff.oldVersion()) :
+                                    "%s (version %s -> %s, self-resolved)".formatted(modid, diff.oldVersion(), diff.newVersion())
```
sciwhiz12: I'm not going to request this be changed now, but it might be better readability-wise if the ternary operators are placed at the beginning of the continuation lines, rather than at the end.
```suggestion
                            .append(diff.isMissing()
                                    ? "%s (version %s -> MISSING, self-resolved)".formatted(modid, diff.oldVersion())
                                    : "%s (version %s -> %s, self-resolved)".formatted(modid, diff.oldVersion(), diff.newVersion())
```

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1338,28 +1350,72 @@ public static void readAdditionalLevelSaveData(LevelStorageSource.LevelDirectory
...
+                final var modid = unres.modid();
+                final var diff = unres.versionDifference();
+                unresolved.append(System.lineSeparator())
+                        .append(diff.isMissing() ?
+                                "%s (version %s -> MISSING)".formatted(modid, diff.oldVersion()) :
+                                "%s (version %s -> %s)".formatted(modid, diff.oldVersion(), diff.newVersion())
+                        );
+            });
+
+            if(mismatchEvent.anyResolved())
```
sciwhiz12: Formatting: space between `if` keyword and parenthesis.
```suggestion
            if (mismatchEvent.anyResolved())
```

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1338,28 +1350,72 @@ public static void readAdditionalLevelSaveData(LevelStorageSource.LevelDirectory
...
+            {
+                resolved.append(System.lineSeparator()).append("Things may not work well.");
+                LOGGER.debug(WORLDPERSISTENCE, resolved.toString());
+            }
+
+            if(mismatchEvent.anyUnresolved())
```
sciwhiz12: Formatting: space between `if` keyword and parenthesis.
```suggestion
            if (mismatchEvent.anyUnresolved())
```

src/main/java/net/minecraftforge/event/ModMismatchEvent.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+/**
+ * <p>
+ * Fires when the mod loader is in the process of loading a world that was last saved
+ * with mod versions that differ from the currently-loaded versions. This can be used to
+ * enqueue work to run at a later point, such as multi-file migration of data.
+ * </p>
```
sciwhiz12: Parentheses tags are unneeded for the first paragraph of a javadoc.
```suggestion
 * Fires when the mod loader is in the process of loading a world that was last saved
 * with mod versions that differ from the currently-loaded versions. This can be used to
 * enqueue work to run at a later point, such as multi-file migration of data.
```

src/main/java/net/minecraftforge/event/ModMismatchEvent.java
```diff
@@ -0,0 +1,184 @@
...
+    public record MismatchedVersionInfo(ArtifactVersion oldVersion, @Nullable ArtifactVersion newVersion)
+    {
+        public boolean isMissing()
+        {
+            return newVersion == null;
+        }
+
+        public boolean wasUpgrade()
+        {
+            if(newVersion == null) return false;
```
sciwhiz12: Formatting: space between `if` keyword and parenthesis.
```suggestion
            if (newVersion == null) return false;
```

src/main/java/net/minecraftforge/event/ModMismatchEvent.java
```diff
@@ -0,0 +1,184 @@
...
+
+    /**
+     * Fetch a previous version of a given mod, if it has been mismatched.
+     * @param modId The mod to fetch previous version for.
+     * @return The previously known mod version, or {@link Optional#empty()} if unknown/not found.
+     */
+    @Nullable
+    public ArtifactVersion getPreviousVersion(String modId)
+    {
+        if(this.versionDifferences.containsKey(modId))
```
sciwhiz12: Formatting: space between `if` keyword and parenthesis.
```suggestion
        if (this.versionDifferences.containsKey(modId))
```

src/main/java/net/minecraftforge/event/ModMismatchEvent.java
```diff
@@ -0,0 +1,184 @@
...
+            return this.versionDifferences.get(modId).oldVersion();
+
+        return null;
+    }
+
+    @Nullable
+    public ArtifactVersion getCurrentVersion(String modid)
+    {
+        if(this.versionDifferences.containsKey(modid))
```
sciwhiz12: Formatting: space between `if` keyword and parenthesis.
```suggestion
        if (this.versionDifferences.containsKey(modid))
```

src/main/java/net/minecraftforge/event/ModMismatchEvent.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    public Stream<MismatchResolutionResult> getResolved()
+    {
+        return resolved.keySet().stream()
+                .map(modid -> new MismatchResolutionResult(modid, versionDifferences.get(modid), resolved.get(modid)))
+                .sorted(Comparator.comparing(MismatchResolutionResult::modid));
+    }
+
+    public record MismatchResolutionResult(String modid, MismatchedVersionInfo versionDifference, @Nullable ModContainer resolver) {
```
sciwhiz12: Formatting: braces on newlines.
```suggestion
    public record MismatchResolutionResult(String modid, MismatchedVersionInfo versionDifference, @Nullable ModContainer resolver)
    {
```

src/main/java/net/minecraftforge/event/ModMismatchEvent.java
```diff
@@ -0,0 +1,184 @@
...
+        public boolean wasSelfResolved()
+        {
+            return resolver != null && resolver.getModId().equals(modid);
+        }
+
```
sciwhiz12: Formatting: for consistency, this blank line at the end of the record class should be removed.

sciwhiz12 (changes requested):
Could we have an example of a log snippet with the new messages, to see how they look?

sciwhiz12 (approved):

PR #8990, opened 2022-08-21
Add optional I/O constraints to item handlers

amadornes:
This PR aims to extend the `IItemHandler` specification with opt-in constraints on insertion and extraction operations that provide additional guarantees to callers, greatly expanding their capabilities at no real cost for implementors.

## Motivation

Introducing advanced functionality to item handlers such as transactions has always been a point of contention. While it can provide huge benefits to the API users that want/need that behavior, it also comes with a sizable implementation and runtime cost, and those who don't benefit from it generally have to adapt their code, carrying over some of the complexity of these APIs. Additionally, they may introduce certain potentially unsafe behavior (requiring rollbacks for simulated/cancelled operations, which may be easily missed or forgotten), and require special attention in cases where side effects must be propagated to other systems, or where multiple views of the same underlying storage are exposed and operated on at once to name a few.

## Objective

The goal of this PR is to provide additional contract terms to the insertion and extraction methods in `IItemHandler` on an opt-in basis, affording callers certain guarantees about how these methods will behave and what their side effects will be.

It is, however, **NOT** a goal of this PR to introduce any explicit transactional helpers or behavior, or to support interfacing with multiple item handlers at once.

## I/O Guarantees

There are three proposed levels of I/O guarantees that item handlers may expose:

### None

This reflects the current contract and is returned by default.

Insertion and extraction may affect slots other than the one specified, and the results of these operations may not be reflected in the stack in the queried slot.

*For example, a trash can may immediately void any items put into it.*

### Consistent

The inventory is guaranteed to accurately reflect the outcome of insertion/extraction operations.

Additionally, all slots are guaranteed to be fully independent from each other, meaning changes to one of them must not immediately affect the contents or behavior of any of the others. A delayed side effect (for example, at the end of the tick or next tick) is considered valid behavior.

This means that a third party may safely operate on multiple slots at once guaranteeing the independence of such operations.

*For example, this ensures that, if presented with an empty two-slot inventory, inserting 5 cobblestone into the first slot does not interfere with inserting 8 planks into the second.*

### Strict

This behavior inherits from the previous and is equivalent to that of vanilla's `Container`. Additionally:

Insertion must adhere to the following:
 - If `isItemValid(int, ItemStack)` returns `false` for the provided stack, the stack must be rejected
 - If the slot contains an item and `ItemHandlerHelper#canItemStacksStack(...)` returns `false` when comparing it with the provided stack, the stack must be rejected
 - Otherwise any amount up to `getMaxStackSize(int, ItemStack)` minus the current size of the stack in the slot must be allowed to be inserted, and any additional items must be rejected

Extraction must adhere to the following:
 - If `isExtractionAllowed(int, ItemStack)` returns `false`, no action must be performed and an empty stack must be returned
 - Otherwise any amount up to the current size of the stack in the slot must be allowed to be extracted.

This essentially turns the insertion and extraction methods into a checked version of `set(...)`.

This means that a third party may safely predict the outcome of multiple insertion/extraction operations on the same slot and collapse them down to one or two operations. They may also do this across multiple slots, since they are guaranteed to be independent.

*For example, the slots in a chest are fully independent from each other and follow these insertion and extraction rules.*

## Cost

It's an enum! Implementors just have to pick one of the values for their item handler based on what guarantees they can provide to external callers (defaults to no guarantees).

Code calling into the item handler API doesn't need to be changed, but can be enhanced by querying the I/O guarantees to ensure the safety of certain operations.

## Request for Comments

This is an open call to anyone interested in such a system to leave their feedback and suggestions.

However, there are some key points that should warrant special attention:
 - Should I/O guarantees be per-slot instead of affecting the whole inventory? Note the additional cost for API callers and any inconsistencies that may be introduced with this change
 - Are the 3 provided levels enough to satisfy people's *general* needs? Understand that some behaviors, while nice, might be too niche for their implementation cost

We also have a thread in the github discussions channel of the ForgeCord, which you can find here:
https://discord.com/channels/313125603924639766/1010529560707739689

### DIFF
--- a/src/main/java/net/minecraftforge/items/IItemHandler.java
+++ b/src/main/java/net/minecraftforge/items/IItemHandler.java
@@ -170,7 +170,8 @@ public interface IItemHandler
          */
         NONE(false, false),
         /**
-         * The inventory is guaranteed to accurately reflect the outcome of insertion/extraction operations.
+         * The inventory is guaranteed to accurately reflect the outcome of insertion/extraction operations via
+         * {@link #getStackInSlot(int)}.
          * Additionally, all slots are guaranteed to be fully independent from each other, meaning changes to one of
          * them must not immediately affect the contents or behavior of any of the others. A delayed side effect
          * (for example, at the end of the tick or next tick) is considered valid behavior.
@@ -180,7 +181,8 @@ public interface IItemHandler
          */
         CONSISTENT(true, false),
         /**
-         * A stricter version of {@link #CONSISTENT}, providing additional guarantees on insertion and extraction.
+         * A stricter version of {@link #CONSISTENT} providing additional guarantees on insertion and extraction
+         * equivalent to an implementation of {@link Container}.
          * <p>
          * Insertion must adhere to the following:
          * <ul>
@@ -188,15 +190,15 @@ public interface IItemHandler
          *     be rejected.</li>
          *     <li>If the slot contains an item and {@link ItemHandlerHelper#canItemStacksStack(ItemStack, ItemStack)}
          *     returns false when comparing it with the provided stack, the stack must be rejected.</li>
-         *     <li>Any amount up to {@link #getMaxStackSize(int, ItemStack)} minus the current size of the stack in the
-         *     slot must be allowed to be inserted, and any additional items must be rejected.</li>
+         *     <li>Otherwise any amount up to {@link #getMaxStackSize(int, ItemStack)} minus the current size of the
+         *     stack in the slot must be allowed to be inserted, and any additional items must be rejected.</li>
          * </ul>
          * <p>
          * Extraction must adhere to the following:
          * <ul>
          *     <li>If {@link #isExtractionAllowed(int, ItemStack)} returns {@code false}, no action must be performed
          *     and an empty stack must be returned</li>
-         *     <li>Any amount up to the current size of the stack in the slot must be allowed to be extracted.</li>
+         *     <li>Otherwise any amount up to the current size of the stack in the slot must be allowed to be extracted.</li>
          * </ul>
          * <p>
          * This means that a third party may safely predict the outcome of multiple insertion/extraction operations on
--- a/src/main/java/net/minecraftforge/items/IItemHandler.java
+++ b/src/main/java/net/minecraftforge/items/IItemHandler.java
@@ -166,6 +166,11 @@ public interface IItemHandler
          * Insertion and extraction may affect other slots, and the results of these operations may not be reflected in
          * the stack in the queried slot.
          * <p>
+         * An I/O operation may raise the guarantees to a more restrictive level, but callers do not need to concern
+         * themselves with this, as they can still safely operate assuming they get no guarantees. A case like this may
+         * show up in a block with multiple input slots where setting the first item determines what can go into the
+         * other slots, and behavior becomes consistent/strict as long as that first item remains in the inventory.
+         * <p>
          * This is how item handlers have historically worked.
          */
         NONE(false, false),
--- a/src/main/java/net/minecraftforge/items/IItemHandler.java
+++ b/src/main/java/net/minecraftforge/items/IItemHandler.java
@@ -50,6 +50,8 @@ public interface IItemHandler
      * The ItemStack <em>should not</em> be modified in this function!
      * </p>
      * Note: This behaviour is subtly different from {@link IFluidHandler#fill(FluidStack, IFluidHandler.FluidAction)}
+     * <p>
+     * Note to implementors: this method may have additional constraints based on {@link #getIOGuarantees()}.
      *
      * @param slot     Slot to insert into.
      * @param stack    ItemStack to insert. This must not be modified by the item handler.
@@ -67,6 +69,7 @@ public interface IItemHandler
      * The returned value must be empty if nothing is extracted,
      * otherwise its stack size must be less than or equal to {@code amount} and {@link ItemStack#getMaxStackSize()}.
      * </p>
+     * Note to implementors: this method may have additional constraints based on {@link #getIOGuarantees()}.
      *
      * @param slot     Slot to extract from.
      * @param amount   Amount to extract (may be greater than the current stack's max limit)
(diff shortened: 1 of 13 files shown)


src/main/java/net/minecraftforge/items/IItemHandler.java
```diff
@@ -85,6 +88,19 @@
      */
     int getSlotLimit(int slot);
```
Technici4n, line 89: Should this method be deprecated and eventually phased out? Passing `ItemStack.EMPTY` to `getMaxStackSize` can be defined (and imo should be defined) to return the same as `getSlotLimit`.

src/main/java/net/minecraftforge/items/IItemHandler.java
```diff
@@ -94,15 +110,131 @@
...
+    {
+        /**
+         * Insertion and extraction may affect other slots, and the results of these operations may not be reflected in
+         * the stack in the queried slot.
+         * <p>
+         * This is how item handlers have historically worked.
+         */
+        NONE(false, false),
+        /**
+         * The inventory is guaranteed to accurately reflect the outcome of insertion/extraction operations.
```
Technici4n: Should this be explained a little bit more? (i.e. mentioning getStackInSlot)

src/main/java/net/minecraftforge/items/IItemHandler.java
```diff
@@ -94,15 +110,131 @@
...
+        /**
+         * A stricter version of {@link #CONSISTENT}, providing additional guarantees on insertion and extraction.
+         * <p>
+         * Insertion must adhere to the following:
+         * <ul>
+         *     <li>If {@link #isItemValid(int, ItemStack)} returns {@code false} for the provided stack, the stack must
+         *     be rejected.</li>
+         *     <li>If the slot contains an item and {@link ItemHandlerHelper#canItemStacksStack(ItemStack, ItemStack)}
+         *     returns false when comparing it with the provided stack, the stack must be rejected.</li>
+         *     <li>Any amount up to {@link #getMaxStackSize(int, ItemStack)} minus the current size of the stack in the
```
Technici4n:
```suggestion
         *     <li>Otherwise any amount up to {@link #getMaxStackSize(int, ItemStack)} minus the current size of the stack in the
```
maybe?

src/main/java/net/minecraftforge/items/IItemHandler.java
```diff
@@ -94,15 +110,131 @@
...
+         *     slot must be allowed to be inserted, and any additional items must be rejected.</li>
+         * </ul>
+         * <p>
+         * Extraction must adhere to the following:
+         * <ul>
+         *     <li>If {@link #isExtractionAllowed(int, ItemStack)} returns {@code false}, no action must be performed
+         *     and an empty stack must be returned</li>
+         *     <li>Any amount up to the current size of the stack in the slot must be allowed to be extracted.</li>
```
Technici4n:
```suggestion
         *     <li>Otherwise any amount up to the current size of the stack in the slot must be allowed to be extracted.</li>
```
maybe?

src/main/java/net/minecraftforge/items/IItemHandler.java
```diff
@@ -94,15 +110,131 @@
...
+         * Additionally, all slots are guaranteed to be fully independent from each other, meaning changes to one of
+         * them must not immediately affect the contents or behavior of any of the others. A delayed side effect
+         * (for example, at the end of the tick or next tick) is considered valid behavior.
+         * <p>
+         * This means that a third party may safely operate on multiple slots guaranteeing the independence of such
+         * operations.
+         */
+        CONSISTENT(true, false),
+        /**
+         * A stricter version of {@link #CONSISTENT}, providing additional guarantees on insertion and extraction.
```
Technici4n: Could mention that the inventory behaves like a Container, essentially just a list of slots with insertion and extraction filters?

src/main/java/net/minecraftforge/items/ItemStackHandler.java
```diff
@@ -162,6 +172,12 @@ public boolean isItemValid(int slot, @NotNull ItemStack stack)
         return true;
     }
 
+    @Override
+    public IOGuarantees getIOGuarantees()
```
Technici4n, line 176: I wonder if some implementations might override methods in a way that make the handler not strict anymore. However I don't think we should do anything about it cause it wouldn't be practical.

src/main/java/net/minecraftforge/items/wrapper/CombinedInvWrapper.java
```diff
@@ -126,4 +135,13 @@ public boolean isItemValid(int slot, @NotNull ItemStack stack)
         int localSlot = getSlotFromIndex(slot, index);
         return handler.isItemValid(localSlot, stack);
     }
+
+    @Override
+    public IOGuarantees getIOGuarantees()
```
Technici4n, line 147: This implementation is wrong if some of the inventories are linked, which can't really be tested for.

src/main/java/net/minecraftforge/items/wrapper/SidedInvWrapper.java
```diff
@@ -233,6 +233,19 @@ public int getSlotLimit(int slot)
     public boolean isItemValid(int slot, @NotNull ItemStack stack)
     {
         int slot1 = getSlot(inv, slot, side);
-        return slot1 == -1 ? false : inv.canPlaceItem(slot1, stack);
+        return slot1 == -1 ? false : inv.canPlaceItem(slot1, stack); // TODO: 1.20 - Change this to the face-aware method
```
Technici4n, line 236: Couldn't this be considered a bug in Forge, and as such anyone relying on the old behavior would be doing so wrongly? Hence it wouldn't be a breaking change.

amadornes: It's both a breaking bug and a breaking fix, so I think this one will need attention from triage.

src/main/java/net/minecraftforge/items/ItemStackHandler.java
amadornes: I think being an item stack handler imposes some contracts already on implementors, so I don't think we really need to worry about it.

src/main/java/net/minecraftforge/items/IItemHandler.java
amadornes: I think this one will need some more discussion. I get the reasoning for not having two versions of the method, but I'd like to know if people are generally okay with the semantics of empty stack = generic context.

src/main/java/net/minecraftforge/items/wrapper/CombinedInvWrapper.java
```diff
@@ -30,6 +31,11 @@ public CombinedInvWrapper(IItemHandlerModifiable... itemHandler)
         this.slotCount = index;
     }
 
+    public CombinedInvWrapper(IItemHandlerModifiable... itemHandler)
```
Technici4n, line 35: Might want a `TODO 1.20: remove to encourage broader support of IOGuarantees` ?

amadornes: Good point, I should probably deprecate it for removal

src/main/java/net/minecraftforge/items/IItemHandler.java
thedarkcolour: Should ItemStack.EMPTY be considered a stackable item in that case?

PR #5738, opened 2019-05-13
[1.13.x] Tags for several old oredicts and some extra

Lykrast:
This adds tags equivalent for several old oredicts, and a few others that were used by certain mods or requested [here](https://github.com/MinecraftForge/MinecraftForge/pull/5609#issuecomment-491081669) by @darkhax.

Anything that's unmarked is an old oredict tag.

Item and block:

* `bookshelves` (requested by the Darkhax post)
* `end_stone`
* `gravel`
* `heads` (was "used" (when the item had metadata) in some mods; named heads because more of the variants were named head than skull)
* `netherrack`
* `obsidian`
* `sandstone`

Item:
* `arrows` (requested)
* `bone`
* `crops` (general tag, didn't make a block tag because of some ambiguity on some blocks (like is the crop the pumpkin block or the stem?))
  * `beetroot` (only new crop, didn't add others because of ambiguity with melons, chorus fruits and such)
  * `carrot`
  * `nether_wart`
  * `potato`
  * `wheat`
* `egg` (no turtle egg because it isn't used in the same recipes)
* `ender_pearl`
* `feather`
* `gunpowder`
* `leather`
* `nether_star`
* `seeds` (used by several mods and requested, only kept ones explicitely referred as seeds)
  * `beetroot`
  * `melon`
  * `pumpkin`
  * `wheat`
* `slime_ball`
* `string`

_Footnote for the linked post: beds and rails are already tagged by vanilla according to the wiki._

PR #5609, opened 2019-03-14
[Omnibus] Add common tags

tterrag1098:
# This Is An Omnibus PR

[Open PRs](https://github.com/MinecraftForge/MinecraftForge/pulls?utf8=%E2%9C%93&q=base%3Anoci-omnibus%2Ftags+is%3Aopen) | [Included PRs](https://github.com/MinecraftForge/MinecraftForge/pulls?utf8=%E2%9C%93&q=base%3Anoci-omnibus%2Ftags+is%3Amerged) | [Submit PR](https://github.com/MinecraftForge/MinecraftForge/compare/noci-omnibus/tags...tterrag1098:noci-omnibus/typos)

### What does this Omnibus target?

This PR is targeting the new tag system, making sure that forge has a good base of default tags for modders to work with and standardize on.

### What does this mean?

An omnibus PR is our new way of combining common, small contributions into a single PR so that they can be merged in a single build, while maintaining the simplicity of contributing small fixes piecewise.

### How can I contribute?

To contribute to this omnibus, you must make your PR at the [source branch](https://github.com/MinecraftForge/MinecraftForge/compare/noci-omnibus/tags...tterrag1098:noci-omnibus/typos). You'll need to switch the source repository/branch when using that link. Once your PR is approved it will be merged to the omnibus branch for later merging into the main trunk.

PR #8991, opened 2022-08-22
Added Bookshelves block tag

justliliandev:
This tag was added during 1.13 lifecycle but went missing sometimes later. https://github.com/MinecraftForge/MinecraftForge/pull/5738
As the tag exists for items, I don't see a reason, why there shouldn't be a blocktag

### DIFF
--- a/src/generated/resources/data/forge/tags/blocks/bookshelves.json
+++ b/src/generated/resources/data/forge/tags/blocks/bookshelves.json
@@ -1,0 +1,5 @@
+{
+  "values": [
+    "minecraft:bookshelf"
+  ]
+}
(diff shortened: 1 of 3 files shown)


sciwhiz12:
It looks like the tags omnibus PR #5609, which your linked PR #5738 was targetting, was manually integrated by @LexManos in https://github.com/MinecraftForge/MinecraftForge/commit/2c8c022034ba9967affc7ecf02dc6583f4f3652a, with him omitting some tags which he personally did not agree with without having listed them out. This explains why it looks like it was added but got removed; it never got into the 1.13 branch in the first place.

Could you please give a usecase for the bookshelves block tag, for the record?

justliliandev:
A usecase for this would be a magical ritual that turns bookshelve blocks into dropped enchanted book utems. That could benefit from such a tag, as mutliple mods add bookshelves for their wood variants/vanilla wood types

sciwhiz12 (approved):

LexManos:
Just for note: I typically reject this tag because of its historical usage and justification in using it in game logic which breaks vanilla compatibility. See the discussion in our back channels for more information and my opinions. So it's a no vote from me, and we'll leave it to others to make the final decision.

PR #8711, opened 2022-06-10
[1.19.x] BiomeTags Fixes/Improvements

Flemmli97:
This PR addresses points made by  sciwhiz12 comment [https://github.com/MinecraftForge/MinecraftForge/pull/8564#pullrequestreview-988572064](https://github.com/MinecraftForge/MinecraftForge/pull/8564#pullrequestreview-988572064) that i didnt manage to fix in time before the previous PR was merged.
It removes redundant duplicate entries for dimension specific tags by using a tag reference instead of copy the contents of the sub tags.

It also fixes `minecraft:sparse_jungle` not being in `is_wet/overworld` cause I accidentally used `is_wet` only there.

Also since I'm not too wellversed in biomes idk if `forge:is_coniferous` and `minecraft:is_taiga` should be treated as equal or not and as such left that as it is now: Atm they are not equal.

~~There is also a pending PR #8684 that, depending of whichever gets merged first, then needs to be addressed too with this change~~

PR #8993, opened 2022-08-23
Port IS_BEACH tag to 1.19.x from 1.18.x

JT122406:


Lupicus:
Vanilla already has IS_BEACH.

sciwhiz12:
The Forge-added tag `forge:worldgen/biome/is_beach` was removed by PR #8711 because, as Lupicus notes, vanilla added the `minecraft:worldgen/biome/is_beach` tag. Though their contents differ, we usually side with vanilla's tags. Even then, modders can use their own tag or custom holdersets (#8928) if they wish to also include `minecraft:stony_shore` in some form for their uses.

PR #8999, opened 2022-08-27
[1.19.x] Add IForgeBlock#onTreeGrow to replace IForgeBlock#onPlantGrow from 1.16

thedarkcolour:
This PR adds the IForgeBlock#onTreeGrow method, which allows a block to determine what happens to it when a tree is grown on it, which is useful for modded grass blocks that shouldn't turn into regular dirt when a tree is grown on them.

In vanilla, when a tree is grown on a suitable block (like a grass block), that block is turned into vanilla dirt. In 1.16.5, changing the dirt block to something else could be accomplished by overriding IForgeBlock#onPlantGrow to place the corresponding modded dirt instead. That method was removed in recent versions though, so I hope this new one will serve the same purpose.
Test mod included!


src/test/java/net/minecraftforge/debug/block/OnTreeGrowBlockTest.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+    });
+    public static final RegistryObject<Item> TEST_GRASS_BLOCK_ITEM = ITEMS.register("test_grass_block", () -> new BlockItem(TEST_GRASS_BLOCK.get(), new Item.Properties()));
+    public static final RegistryObject<Item> TEST_DIRT_ITEM = ITEMS.register("test_dirt", () -> new BlockItem(TEST_DIRT.get(), new Item.Properties()));
+
+    public OnTreeGrowBlockTest()
+    {
+        IEventBus modBus = FMLJavaModLoadingContext.get().getModEventBus();
+        BLOCKS.register(modBus);
+        ITEMS.register(modBus);
```
ChampionAsh5357: Should have `ENABLE` flag to enable and disable test mod.

ChampionAsh5357 (changes requested):
What would be the benefit of using this method compared to adding it to the dirt tag and using `Block#updateShape` to update the modded block? Vanilla does define the dirt tag to hold both grass blocks and dirt.

thedarkcolour:
`Block#updateShape` never gets called since the TrunkPlacer sets the dirt before the logs, and there's no way to differentiate between a sapling growing versus a player placing a log on top.

ChampionAsh5357:
I can accept this descision. 

> `Block#updateShape` never gets called since the TrunkPlacer sets the dirt before the logs, and there's no way to differentiate between a sapling growing versus a player placing a log on top.

Fair enough. Only additional comment would be to update the javadoc of the method to give notice that calling the method is not a substitute for adding the blocks to the dirt tag, rather a supplemental for immediate behavior.

ChampionAsh5357 (approved):

PR #9000, opened 2022-08-27
[1.19.x] Moved Player.resetAttackStrengthTicker() to the end of Player.attack()

bconlon1:
This is another take on solving the problem that #8891 is meant to fix. 

The problem is that the player's attack progress scale is reset before they've hurt an entity (`resetAttackStrengthTicker()` resets `attackStrengthTicker`, which is used in calculating the value gotten from `getAttackStrengthScale()`), meaning if a modder wants to know what the player's attack strength scale is *after* they've hurt an entity (through using `getAttackStrengthScale()` as is otherwise possible earlier in the attack sequence), that scale value is always zero. This is fixed by moving the call to reset the value to the end of Player.attack() after the entity hurt code has been called.

One reason a modder may want to know what the attack strength scale is by using `getAttackStrengthScale()` after an entity has been hurt is for checking if the player has performed an attack with full strength (when the attack indicator is full), but also has been verified as having hurt the entity. This could be used for performing an item ability only when a player has injured an entity with a full attack swing.

This change has been tested and does not appear to affect any vanilla behavior; the attack indicator and attack swing recharge still display the same as normal.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
@@ -280,15 +280,14 @@
                          this.m_21008_(InteractionHand.MAIN_HAND, ItemStack.f_41583_);
                       }
                    }
-@@ -1212,7 +_,7 @@
+@@ -1212,6 +_,7 @@
                    }
                 }
              }
--
-+            this.m_36334_();
++            this.m_36334_(); // FORGE: Moved from beginning of attack() so that getAttackStrengthScale() returns an accurate value during all attack events
+ 
           }
        }
-    }
 @@ -1228,7 +_,7 @@
        }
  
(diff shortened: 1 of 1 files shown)


patches/minecraft/net/minecraft/world/entity/player/Player.java.patch
```diff
@@ -276,6 +280,15 @@
                          this.m_21008_(InteractionHand.MAIN_HAND, ItemStack.f_41583_);
                       }
                    }
+@@ -1212,7 +_,7 @@
+                   }
+                }
+             }
+-
++            this.m_36334_();
```
XFactHD: A small comment should be added summarizing what this patch does: `//FORGE: whatever`.
Also, small nitpick: don't remove the empty line, just add the call in a separate line above it.

ChampionAsh5357:
> The problem is that the player's attack strength value is reset before they hurt an entity, meaning if a modder wants to know what the player's attack strength is after they've hurt an entity, the attack strength value is always zero. This is fixed by moving the call to reset the value to the end of Player.attack() after the entity hurt code has been called.

This argument doesn't make much sense. You're saying you want to know the strength of the players attack; however, what you are adjusting is getting the scale of the player's strength, not the strength value itself. Why do you need the strength scalar compared to the computed strength value?

bconlon1:
I have now updated the description to better clarify that I am talking about the player's attack strength *scale*, i.e. the thing represented by the player's attack strength indicator in-game, not actual strength in terms of attack damage if that's what this was being confused with.

ChampionAsh5357 (approved):

PR #9007, opened 2022-09-03
Fix failed entity interactions consuming the click.

### LINKED ISSUE #9006: [1.19.2] Cannot right click while looking at mobs at certain distance
**Minecraft Version:** 1.19.2

**Forge Version:** 43.1.3

**Logs:** https://gist.github.com/BlizzardFenix/6622bedc073c8f532fd981dca42ccc79

**Steps to Reproduce:**
 1. Approach a mob (or other entity)
 2. Hold a shield or any other item with use functionality
 3. Stand around 5 blocks away while looking at the entity
 4. Right click
 5. Nothing will happen

**Description of issue:**
Right clicking while looking at pickable entities from a certain distance (around 5 blocks) does not actually use the held item. This does not happen in vanilla Minecraft.
Also PlayerInteractEvent.RightClickItem does not get called in this case.

Shadows-of-Fire:
Fixes #9006 

In a corner case, an entity may be successfully traced by GameRenderer, but may not actually be within valid reach distance for interaction.  This should result in an empty click occuring, but since there is a `return` statement here, it consumes the click action instead of passing to empty.

### DIFF
--- a/patches/minecraft/net/minecraft/client/Minecraft.java.patch
+++ b/patches/minecraft/net/minecraft/client/Minecraft.java.patch
@@ -302,7 +302,7 @@
                             return;
                          }
  
-+                        if(!this.f_91074_.canInteractWith(entityhitresult.m_82443_(), 0)) return; //Forge: Entity may be traced via attack range, but the player may not have sufficient reach.  No padding in client code.
++                        if(!this.f_91074_.canInteractWith(entityhitresult.m_82443_(), 0)) break; //Forge: Entity may be traced via attack range, but the player may not have sufficient reach.  No padding in client code.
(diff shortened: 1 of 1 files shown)


TheCurle:
Looks like it fixes the issue from my testing.

KostromDan:
This broke Grapple Hook mod. Dev saying it's not his fault and won't fix it.

KostromDan:
https://github.com/yyon/grapplemod/issues/126
https://github.com/yyon/grapplemod/issues/127
https://github.com/yyon/grapplemod/issues/130
https://github.com/yyon/grapplemod/issues/132
https://github.com/yyon/grapplemod/issues/135

PR #9009, opened 2022-09-04
Add value change listening capability to config values

FirEmerald:
allows mod authors to directly listen for when a config value's value changes, mainly useful for when the value is meant to be parsed into a different type, for instance a String to a ResourceLocation, or when some systems need to be updated to account for the changed value.

### DIFF
--- a/src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
+++ b/src/main/java/net/minecraftforge/common/ForgeConfigSpec.java
@@ -21,6 +21,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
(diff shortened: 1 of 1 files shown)


ChampionAsh5357:
What is the benefit of using this over the config events?

Additionally, PRs should be made to latest and then backported to LTS.

TheCurle:
Config Value already supports converting to other types natively, and the Config Events provide a way to be notified of changes.

This PR has no functional value, so if you'd like to try again, please retarget for 1.19 and elaborate on why the current functionality is not sufficient.

FirEmerald:
> Config Value already supports converting to other types natively, and the Config Events provide a way to be notified of changes.
> 
> This PR has no functional value, so if you'd like to try again, please retarget for 1.19 and elaborate on why the current functionality is not sufficient.

Config value actually does not support converting to other types natively. It only supports the types TOML itself does, and attempting to use any other value is a guaranteed ClassCastException. The only way to support other types would be to allow usage of a subtype of ConfigValue, which is currently impossible as the constructor for ConfigValue is default (package only) access. I am going to attempt to write a change on 1.19.x to support this but I may just give up as this has been such a mess anyways, and I have been recommended to just delegate everything to datapacks.

P.S. I am dumb and didn't even notice my listeners wouldn't work anyway as the cached value is only updated if something attempts to grab the value. properly implementing value change listeners would require far more code rewriting than I feel comfortable doing without the possibility of breaking something.

PR #9010, opened 2022-09-04
LTS Backport of ModMistachEvent (#8989)

NertzhulDEV:
Backport of ModMistachEvent to LTS 1.18.x

The reason for the field change from `LevelDirectory` to `Path` is due to the fact that `LevelDirectory` is an abstraction introduced in 1.19 to ease the retrieval of data such as the level data file or the level icon file, so all the operations performed by the LevelDirectory's methods can be manually done by using `Path#resolve()` 

Example:
`levelDirectory.dataFile()` directly translates to `levelPath.resolve("level.dat")`

### DIFF
--- a/src/main/java/net/minecraftforge/event/ModMismatchEvent.java
+++ b/src/main/java/net/minecraftforge/event/ModMismatchEvent.java
@@ -1,0 +1,177 @@
+package net.minecraftforge.event;
+
+import net.minecraftforge.eventbus.api.Cancelable;
+import net.minecraftforge.eventbus.api.Event;
+import net.minecraftforge.fml.LogicalSide;
+import net.minecraftforge.fml.ModContainer;
+import net.minecraftforge.fml.ModList;
+import net.minecraftforge.fml.ModLoadingContext;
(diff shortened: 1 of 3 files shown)


src/main/java/net/minecraftforge/event/ModMismatchEvent.java
```diff
@@ -0,0 +1,177 @@
+package net.minecraftforge.event;
```
ChampionAsh5357, line 6: Missing license header.

ChampionAsh5357 (changes requested):
Your backport should provide information on why the level location is a `Path` compared to 1.19 `LevelDirectory`. (I know it's because the field was changed, but it should be expressed for anyone who reviews this.)

NertzhulDEV:
> Your backport should provide information on why the level location is a `Path` compared to 1.19 `LevelDirectory`. (I know it's because the field was changed, but it should be expressed for anyone who reviews this.)

Added field change explanation to the original post so anyone seeing this can quickly find it

ChampionAsh5357 (approved):

PR #9013, opened 2022-09-04
[1.19] Fix check if entity/block is in range

Meldexun:
#8478 introduced an issue. The check if an entity or block is in interaction/attack range always assumes that the player is looking exactly at the center of the entity or block. But a player can also look at a corner of an entity which might be a lot closer to the eyes of the player.

Here is a picture of the issue:
reachIssue

Another issue is that `Entity#getPickRadius` is ignored in the calculation.

This PR fixes this issue by calculating the distance from the eyes of the player to the nearest point of the entity bounding box inflated by `Entity#getPickRadius`/block position.

### DIFF
--- a/src/main/java/net/minecraftforge/common/extensions/IForgePlayer.java
+++ b/src/main/java/net/minecraftforge/common/extensions/IForgePlayer.java
@@ -5,11 +5,11 @@
 
 package net.minecraftforge.common.extensions;
 
-import java.util.Optional;
-
 import net.minecraft.core.BlockPos;
+import net.minecraft.util.Mth;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.phys.AABB;
 import net.minecraft.world.phys.Vec3;
 import net.minecraftforge.common.ForgeMod;
 
@@ -74,9 +74,9 @@ public interface IForgePlayer
     default boolean isCloseEnough(Entity entity, double dist)
     {
         Vec3 eye = self().getEyePosition();
-        Vec3 targetCenter = entity.getPosition(1.0F).add(0, entity.getBbHeight() / 2, 0);
-        Optional<Vec3> hit = entity.getBoundingBox().clip(eye, targetCenter); //This hit should almost always be present, but we have a fallback just in case.  It likely indicates lag if it is not present.
-        return (hit.isPresent() ? eye.distanceToSqr(hit.get()) : self().distanceToSqr(entity)) < dist * dist;
+        AABB aabb = entity.getBoundingBox().inflate(entity.getPickRadius());
+        Vec3 target = new Vec3(Mth.clamp(eye.x, aabb.minX, aabb.maxX), Mth.clamp(eye.y, aabb.minY, aabb.maxY), Mth.clamp(eye.z, aabb.minZ, aabb.maxZ));
+        return eye.distanceToSqr(target) < dist * dist;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Cixon:
vouch for 1.18

Shadows-of-Fire:
The details of this implementation need to match vanilla, which (from what I remember) does "eye to center" checks on the server for validity. If that isn't the case we can update the implementation to match this.

Meldexun:
Vanilla only checks if the distance between the player and the entity is less than 6 blocks with a bottom-center to bottom-center check. That means (assuming we are playing without any reach increasing mods/datapacks, without lags and without cheats) that basically all attack packets from clients get accepted because clients only work with a reach of 3 blocks and do a raytrace from the eyes to whatever they hit.

The only exception to this is when size of the player or the size of the target entity is a lot bigger than in vanilla.
The image below shows the reach check on the client represented by the green circle and the reach check on the server represented by the red circle. In this example the target entity is 6 blocks high and the client check succeeds indicated by the entity bounding box intersecting with the green circle. But the server check fails indicated by the entity bottom-center position being outside the red circle.
image

So to me this looks like vanilla just assumes that there are no big entities in the game and that the reach of the player also doesn't change and thus the inaccurate bottom-center to bottom-center check on the server is deemed good enough.

But this isn't vanilla and there are mods that add big entities or that change the player's reach. Thus in my opinion it is the best solution for Forge to accept all attacks where the target entity's bounding box intersects with the reach radius of the player (+ some padding to account for lag).

Shadows-of-Fire:
A couple things to address:

1. On the client side, this will effectively increase the attack range (I think), because the code responsible for collecting the `objectMouseOver` uses the max of both reach and range, and then relies on these checks for validity, and this bound is "looser" than the current bound (which is also looser than vanilla).
  a. To fix this, we need to employ a proper `clip()` from the player's eye vector to the target entity on the client side, while performing the "eye to closest corner" check on the server side (which adds greater validation for large entities).
  b. A proper clip may resolve the issue described in #9309 for attack range (the reach distance component is will not fix).
2. If we loosen to "eye to closest corner", do we need to decrease the padding values?  There may be unintended side effects of such an implementation, since the "effective" distance _should_ be shorter to the closest corner, compared to bottom center to center.

marchermans:
We will need a test mod covering the corner cases to consider this issue again.
Could you please provide one?

On top of this, we will need a world that shows the issue (for example, using armor stands and command blocks....)

And a video showing what the exact issue is, how you reproduced it, and how we need to test it would be very useful and speed up this whole process.

Meldexun:
> 1. On the client side, this will effectively increase the attack range (I think), because the code responsible for collecting the `objectMouseOver` uses the max of both reach and range, and then relies on these checks for validity, and this bound is "looser" than the current bound (which is also looser than vanilla).
> a. To fix this, we need to employ a proper `clip()` from the player's eye vector to the target entity on the client side, while performing the "eye to closest corner" check on the server side (which adds greater validation for large entities).
> b. A proper clip may resolve the issue described in [Forge 40.1.22 1.18.2+ has extended player reach #9309](https://github.com/MinecraftForge/MinecraftForge/issues/9309) for attack range (the reach distance component is will not fix).

This would be fixed by #9015.

---

> 2. If we loosen to "eye to closest corner", do we need to decrease the padding values?  There may be unintended side effects of such an implementation, since the "effective" distance _should_ be shorter to the closest corner, compared to bottom center to center.

Too low padding = too many attack requests get denied
Lower padding = more accurate/better against cheaters
Higher padding = more lenience/better when playing with lags
Too high padding = cheating will be easy and more effective

What is the correct padding value now? Well there is no definite answer to that question. I think the current padding values are fine but they could also be changed slightly. Everything between 1 and 4 should be okay.

---

> To consider this issue again, we will need a test mod covering the corner cases.

The issue is reproducible in forge without any mods installed. So I don't really understand how a test mod could look like. (I'm assuming a test mod is just a normal mod that allows you to reproduce the issue easily.)

Shadows-of-Fire:
I think lowering the server padding from 3 to 2 would be reasonable given the introduced leniency.

As for a test mod, generally this would need to be tested by a human due to the nature.  You may be able to make an automated test by doing the following:

1. Create a player and place it in a specified location with a specified facing
2. Place a target at a specified location
3. Query if the player can hit the target
4. Repeat steps 2-3 with different positions (some successful, some unsuccessful), different target types (block/entity), different clicks (left/right), and stat modifiers.

Shadows-of-Fire:
superseded by #9361

PR #9015, opened 2022-09-04
[1.19] Fix client reach checks ignoring hitResult

Meldexun:
#8478 moved the check if an entity is in reach to be able to differentiate between attack range and interaction range. But with this change the location where hit an entity aabb is not used anymore to check if the the attack is valid. Thus the player can attack an entity by targeting parts of the entity aabb which should be out of the players attack range.

In this picture the player can only target the zombie when looking at the top or the bottom of the aabb of the zombie (in this case the top is targeted). The distance between the targeted position of the aabb and the players eyes is ~3.6 blocks. This is greater than the attack range of 3 blocks. But the player is still able to attack the zombie because the location of the hitResult is ignored. Instead it's assumed that the player is targeting the center of the zombies aabb which isn't actually possible because there are blocks in the way.
reachIssue2

This PR fixes this issue by using the location of the hitResult again when calculating if an entity attack/interaction is valid.

### DIFF
--- a/patches/minecraft/net/minecraft/client/Minecraft.java.patch
+++ b/patches/minecraft/net/minecraft/client/Minecraft.java.patch
@@ -293,7 +293,7 @@
 +            var inputEvent = net.minecraftforge.client.ForgeHooksClient.onClickInput(0, this.f_91066_.f_92096_, InteractionHand.MAIN_HAND);
 +            HitResult.Type hitType = this.f_91077_.m_6662_();
 +            if (this.f_91077_ instanceof EntityHitResult entityHit) { // Forge: Perform attack range checks here, instead of in GameRenderer#pick, so extended-reach interactions work.
-+               if (f_91074_.m_146892_().m_82557_(f_91077_.m_82450_()) >= f_91074_.getAttackRange() * f_91074_.getAttackRange()) hitType = HitResult.Type.MISS; // No padding in client code.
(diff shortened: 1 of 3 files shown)


Shadows-of-Fire:
I don't really agree with this solution.  `canHit` should be the source of truth for if a player can attack a target.  If there are issues with either the passed position or lack of information, that data needs to be passed through to it, and the implementation updated.

Meldexun:
I guess I could add a nullable hit location as another argument `canHit(Entity, @Nullable Vec3 hitLocation, double padding)`.

Alternatively I could add a new method which takes a Vec3 instead of an Entity `canHit(Vec3 hitLocation, double padding)` (maybe without the padding because it won't be used by forge anyway).

Shadows-of-Fire:
I think the `canHit(Vec3 hitLocation, double padding) ` overload is the best option.  The Entity variant can delegate to this method using the current logic.

Cixon:
vouch for 1.18

Shadows-of-Fire:
@Meldexun Do you have plans to update this with the desired implementation or is this dead?

Meldexun:
Added the discussed methods now.
Though I haven't changed the implementation of the other methods to make use of the new methods because `isCloseEnough` doesn't always use the eye position of the player. Also my other PR (#9013) is changing the other methods and I don't think I should work on the same code with two different PRs.

Shadows-of-Fire (approved):

Shadows-of-Fire:
superseded by #9361

PR #9016, opened 2022-09-04
Fix reloading event firing during server shutdown and add explicit unloading event instead

cpw:
This fixes the problem that sometimes you might get a Reloading config event because the filewatcher notices the save _after_ the config has been unloaded during server shutdown. This can result in unwelcome "Config is in an invalid state" errors.

Added an explicit Unloading event to compensate in case anyone was somehow relying on this (they shouldn't have been?!)

### DIFF
--- a/fmlcore/src/main/java/net/minecraftforge/fml/config/IConfigEvent.java
+++ b/fmlcore/src/main/java/net/minecraftforge/fml/config/IConfigEvent.java
@@ -11,7 +11,7 @@ import net.minecraftforge.fml.Bindings;
 import java.util.function.Function;
 
 public interface IConfigEvent {
-    record ConfigConfig(Function<ModConfig, IConfigEvent> loading, Function<ModConfig, IConfigEvent> reloading) {}
+    record ConfigConfig(Function<ModConfig, IConfigEvent> loading, Function<ModConfig, IConfigEvent> reloading, Function<ModConfig, IConfigEvent> unloading) {}
     ConfigConfig CONFIGCONFIG = Bindings.getConfigConfiguration().get();
 
     static IConfigEvent reloading(ModConfig modConfig) {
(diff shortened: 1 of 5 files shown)


marchermans (approved):

fmlcore/src/main/java/net/minecraftforge/fml/config/IConfigEvent.java
```diff
@@ -11,7 +11,7 @@
 import java.util.function.Function;
 
 public interface IConfigEvent {
-    record ConfigConfig(Function<ModConfig, IConfigEvent> loading, Function<ModConfig, IConfigEvent> reloading) {}
+    record ConfigConfig(Function<ModConfig, IConfigEvent> loading, Function<ModConfig, IConfigEvent> reloading, Function<ModConfig, IConfigEvent> unloading) {}
```
sciwhiz12: Do we consider this part of the public API? Because if it is, we may need some sort of bouncer for the old two-arg constructor to avoid a breaking change here.

cpw: I debated about this. It's definitely not part of the "standard" public API - it's there to allow FMLonly and Forge to share the code. It's arguable that someone in sponge is using it to supply configs in a completely different way, but I don't have any insight into that. IMO we should not treat it as _public_.

sciwhiz12: How are we to handle changes of a breaking nature in this section of the API? Does this section in particular (or the FML API, however nebulous that may be at the moment, as a whole) share none of the breaking changes guarantees that the main Forge API has?

We could certainly push this now and bank on the (high) chance that nobody else other than Forge uses this. But we should be considering how we will treat breaking changes in this part of the API for the future.

sciwhiz12 (approved):
I'm pushing the decision on whether to treat this as a breaking change to the core team.

fmlcore/src/main/java/net/minecraftforge/fml/config/IConfigEvent.java
LexManos: We already discussed related things before. Now that we are using @ApiStatus.Internal anything without it is covered under the binary breaking rules. So yes this needs a bouncer for the constructor.

PR #8464, opened 2022-02-11
Add bosses tag

MichaelTRoss0:
Add EntityType tag for bosses.

By default, the tag contains Ender Dragon and Wither.

This PR is a prerequisite for https://github.com/MinecraftForge/MinecraftForge/pull/8442

PR #9017, opened 2022-09-04
Create a Forge EntityType Tag for Bosses

### LINKED ISSUE #8400: Add a boss tag
Back in MCP days we mistakenly named `Entity.canChangeDimension` as `Entity.isNonBoss` and most people seriously used that method to determine if something was a boss or not.

Now that we have the actual name, there's no longer a way (that I know of) to determine whether something's a "boss mob" or not.

This ticket proposes an entity tag be added `forge:bosses` that mods can add their bosses to.

GizmoTheMoonPig:
This is a port of the seemingly abandoned PR #8464, and it closes #8400.

A simple PR to create a forge tag for boss mobs. Willie did a good job explaining why this tag should exist in #8400, but I also have a couple specific use cases if you need more examples. 

I recently ported a mod a friend of mine made a long time ago, called Beat Down Stick. The mod adds a stick which does the same amount of damage as an entity's max health, or 10% of their health if they're considered a boss. The stick uses ``` !entity.canChangeDimensions() ``` to check if it was a boss, which, if you know what the method used to be named in MCP, makes complete sense. This leads to a couple different issues. One being that if a mod adds an entity that isn't allowed to go through portals it will be considered a boss and damaging 10% of its health would be strange to end users. The other being that something considered a boss could be allowed to go through portals, and you could easily 1 shot it with a puny stick. A proper check for this would be a godsend. 

Another mod example is a mod another friend of mine made called Boss Rifts. After you kill a boss, it will spawn a rift that you can use to bring you and any entities around you back to your spawn point. Currently the mod has a custom tag for this, so it has to be added to as more bosses are discovered, which is painful. I'm sure a tag like this would be super handy for that. 

Please let me know if there are any mistakes in my PR. I made sure I followed all of the guidelines, but this is my first time making one of these for Forge so I may have missed something. Thank you for hearing me out!

### DIFF
--- a/src/main/java/net/minecraftforge/common/Tags.java
+++ b/src/main/java/net/minecraftforge/common/Tags.java
@@ -12,6 +12,7 @@ import net.minecraft.tags.BlockTags;
 import net.minecraft.tags.FluidTags;
 import net.minecraft.tags.ItemTags;
 import net.minecraft.tags.TagKey;
+import net.minecraft.world.entity.EntityType;
 import net.minecraft.world.item.DyeColor;
 import net.minecraft.world.item.Item;
 import net.minecraft.world.level.biome.Biome;
@@ -24,6 +25,7 @@ public class Tags
     public static void init ()
     {
         Blocks.init();
+        EntityTypes.init();
         Items.init();
         Fluids.init();
         Biomes.init();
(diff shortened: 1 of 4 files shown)


src/main/java/net/minecraftforge/common/Tags.java
```diff
@@ -180,6 +182,19 @@ private static TagKey<Block> tag(String name)
...
+        private static void init() {}
+
+        public static final TagKey<EntityType<?>> BOSSES = tag("bosses");
+
+        private static TagKey<EntityType<?>> tag(String name)
+        {
+            return TagKey.create(Registry.ENTITY_TYPE_REGISTRY, new ResourceLocation("forge", name));
+        }
+    }
+
```
sciwhiz12, line 196: Formatting: there should only be one blank line, not two.

sciwhiz12 (changes requested):
Looks fine to me, just needs a tiny formatting change. :shipit:

src/main/java/net/minecraftforge/common/Tags.java
GizmoTheMoonPig: fixed!

sciwhiz12 (approved):

PR #9018, opened 2022-09-05
[1.19] Fix EntityEvent.Size not being fired

Meldexun:
Continuation of #8667

This is not a direct port of the old PR and instead as described previously I split the `EntityEvent.Size` into `EntityEvent.Size` and `EntityEvent.EyeHeight`.

### DIFF
--- a/patches/minecraft/net/minecraft/world/entity/Entity.java.patch
+++ b/patches/minecraft/net/minecraft/world/entity/Entity.java.patch
@@ -43,14 +43,13 @@
        this.f_19847_ = p_19870_;
        this.f_19853_ = p_19871_;
        this.f_19815_ = p_19870_.m_20680_();
-@@ -250,7 +_,11 @@
+@@ -250,7 +_,10 @@
        this.f_19804_.m_135372_(f_146800_, 0);
        this.m_8097_();
        this.m_6034_(0.0D, 0.0D, 0.0D);
 -      this.f_19816_ = this.m_6380_(Pose.STANDING, this.f_19815_);
-+      net.minecraftforge.event.entity.EntityEvent.Size sizeEvent = net.minecraftforge.event.ForgeEventFactory.getEntitySizeForge(this, Pose.STANDING, this.f_19815_, this.m_6380_(Pose.STANDING, this.f_19815_));
-+      this.f_19815_ = sizeEvent.getNewSize();
-+      this.f_19816_ = sizeEvent.getNewEyeHeight();
++      this.f_19815_ = this.getDimensionsForge(Pose.STANDING);
++      this.f_19816_ = this.getEyeHeightForge(Pose.STANDING, this.f_19815_);
 +      net.minecraftforge.common.MinecraftForge.EVENT_BUS.post(new net.minecraftforge.event.entity.EntityEvent.EntityConstructing(this));
 +      this.gatherCapabilities();
     }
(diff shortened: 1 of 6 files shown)


ChampionAsh5357:
Currently, the entire PR would be a breaking change as the contract of the event would change and the methods would be removed. You can restore the previous behavior and deprecate everything that needs to be removed and leave Forge markers for what needs to be changed. If you believe that this isn't worthwhile, I would recommend waiting until the next version release to submit this change.

Shadows-of-Fire:
This can be merged in the breaking changes window of 1.19.3 if it is rebased in time

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -595,7 +651,7 @@
 +   /**
 +    * Accessor method for {@link #getEyeHeight(Pose, EntityDimensions)}
 +    */
-+   public float getEyeHeightAccess(Pose pose, EntityDimensions size) {
++   public final float getEyeHeightAccess(Pose pose, EntityDimensions size) {
 +      return this.m_6380_(pose, size);
 +   }
 +
```
Shadows-of-Fire, line 657: Given that `getEyeHeight(Pose, Dims)` is deprecated now, this method can probably be removed, since it should not be being used externally in favor of the event-computed value (which is accessible via `getEyeHeight()`)

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -414,6 +432,33 @@
...
+       Vec3 vec3 = new Vec3(this.m_20185_() - (double)f, this.m_20186_(), this.m_20189_() - (double)f);
+       Vec3 vec31 = new Vec3(this.m_20185_() + (double)f, this.m_20186_() + (double)entitydimensions.f_20378_, this.m_20189_() + (double)f);
+@@ -2570,12 +_,16 @@
+       this.f_19828_ = p_20012_;
+    }
+ 
++   /**
++    * @deprecated Can be overridden. Use {@link #getEyeHeightForge(Pose, EntitySize)} instead.
++    */
++   @Deprecated
```
Shadows-of-Fire, line 451: May be better to read
"Should be overriden, but callers should invoke `#getEyeHeightForge(Pose, Dims)` to compute the true eye height.  If you do not want to compute a new value, use `#getEyeHeight()`"

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -519,6 +564,17 @@
...
     }
  
+@@ -2903,6 +_,10 @@
+       return new ClientboundAddEntityPacket(this);
+    }
+ 
++   /**
++    * @deprecated Can be overridden. Use {@link #getDimensionsForge(Pose)} instead.
++    */
++   @Deprecated
```
Shadows-of-Fire, line 574: Same comments as on getEyeHeight(Pose, Dims)

Shadows-of-Fire: Actually, we don't appear to have an accessor for this, so we may want to add one.  Otherwise the event-computed dimensions aren't really exposed.

Shadows-of-Fire (changes requested):

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
Meldexun: This accessor is needed because `getEyeHeight(Pose, EntityDimensions)` is protected and can't be called in `IForgeEntity#getEyeHeightForge(Pose, EntityDimensions)`.

patches/minecraft/net/minecraft/world/entity/Entity.java.patch
Meldexun: This method is public so I don't know what we could need an accessor for?

PR #9019, opened 2022-09-05
Allow faces of an "elements" model to have disabled ambient occlusion

nikita488:
Minecraft allows to disable ambient occlusion on entire model (but in a horrible way, because it will only work if it's disabled in the top-level model). Sometimes you need to disable ao for specific faces, especially when you want to make one layer fully emissive (which will look too dark and not emissive with ao enabled, especially in corners) and overlay layer that will look too bright if you disable ao for the entire model.
This PR allows to specify ao flags separately for each element face via "ambientocclusion": true|false entry. This is done by patching ModelBlockRenderer, BakedQuad, BlockElementFace and FaceBakery classes.
Also it comes with support for data gen and test mod is also provided.
Tested with default Minecraft pipeline and Forge's experimental pipeline.
image


src/test/resources/assets/ambient_occlusion_elements_test/models/block/ambient_occlusion.json
```diff
@@ -0,0 +1,35 @@
...
+  "render_type": "cutout",
+  "textures": {
+    "all": "minecraft:block/white_concrete",
+    "overlay": "minecraft:block/glass"
+  },
+  "elements": [
+    {
+      "from": [ 0, 0, 0 ],
+      "to": [ 16, 16, 16 ],
+      "shade": false,
```
marchermans: This does not work. Check your own deserialized again.

marchermans (changes requested):

src/test/resources/assets/ambient_occlusion_elements_test/models/block/ambient_occlusion.json
nikita488: I just tested it and it worked for me. But i can see that it's hard to see the difference, so i'll expand test mod to have a 4 blocks with different shade and ao settings and pure white texture so the result will be more clear to see and compare

patches/minecraft/net/minecraft/client/renderer/block/model/BlockElementFace.java.patch
```diff
@@ -1,33 +1,40 @@
...
+@@ -37,7 +_,11 @@
           int i = this.m_111368_(jsonobject);
           String s = this.m_111370_(jsonobject);
           BlockFaceUV blockfaceuv = p_111367_.deserialize(jsonobject, BlockFaceUV.class);
 -         return new BlockElementFace(direction, i, s, blockfaceuv);
 +         int emissivity = GsonHelper.m_13824_(jsonobject, "emissivity", 0);
 +         if (emissivity != net.minecraft.util.Mth.m_14045_(emissivity, 0, 15))
 +            throw new JsonParseException("The emissivity value must be between 0 and 15. Found: " + emissivity);
-+         return new BlockElementFace(direction, i, s, blockfaceuv, emissivity);
++         boolean hasAmbientOcclusion = GsonHelper.m_13855_(jsonobject, "ambientocclusion", true);
```
OrionDevelopment, line 36: Based on this line the serializer, the json entries should be "ambientocclusion" not "shade" as you use later on.

Please elaborate on how this works in your current testing environment?

marchermans (approved):

PR #9021, opened 2022-09-06
Allow mods to specify shader import namespace

FiniteReality:
Previously any included shaders would be hardcoded to search using the `minecraft` namespace, limiting options for organisation options for modders. This change allows modders to specify which namespace that their shader imports should be loaded from, by prefixing their import with the desired mod namespace.

To be completely honest, I am not a huge fan of this approach, but it appears to be the least invasive, and also avoids a double-lookup where we attempt to search the mod's namespace before falling back to the `minecraft` namespace.


patches/minecraft/net/minecraft/client/renderer/ShaderInstance.java.patch
```diff
@@ -30,3 +30,31 @@
...
+@@ -232,12 +_,13 @@
+                private final Set<String> f_173369_ = Sets.newHashSet();
+ 
+                public String m_142138_(boolean p_173374_, String p_173375_) {
+-                  p_173375_ = FileUtil.m_179924_((p_173374_ ? s1 : "shaders/include/") + p_173375_);
+-                  if (!this.f_173369_.add(p_173375_)) {
++                  // FORGE: use the mod's namespace to look up resources if specified
++                  ResourceLocation resourcelocation = new ResourceLocation(p_173375_);
++                  resourcelocation = new ResourceLocation(resourcelocation.m_135827_(),
++                     FileUtil.m_179924_((p_173374_ ? s1 : "shaders/include/") + resourcelocation.m_135815_()));
```
ChampionAsh5357: This could be made into a Forge client hook to reduce the patch size.

src/test/java/net/minecraftforge/debug/client/rendering/ShaderResourcesTest.java
```diff
@@ -0,0 +1,66 @@
...
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+
+@Mod(ShaderResourcesTest.MODID)
+public class ShaderResourcesTest
+{
+    private static Logger LOGGER;
+
+    public static final String MODID = "shader_resources_test";
+    private static final boolean ENABLED = false;
```
ChampionAsh5357: I believe the convention Forge uses is `ENABLE`.

src/test/java/net/minecraftforge/debug/client/rendering/ShaderResourcesTest.java
```diff
@@ -0,0 +1,66 @@
...
+            LOGGER = LogUtils.getLogger();
+
+            final var modEventBus = FMLJavaModLoadingContext.get().getModEventBus();
+            modEventBus.addListener(ShaderResourcesTest::registerShaders);
+        }
+    }
+
+    public static void registerShaders(final RegisterShadersEvent event)
+    {
+        if (!ENABLED)
```
ChampionAsh5357: Test mod will fail on the server. Please fix sideness.

ChampionAsh5357 (changes requested):

src/test/java/net/minecraftforge/debug/client/rendering/ShaderResourcesTest.java
FiniteReality: Oh right, I didn't check this on the server side -  originally I was using the subscribe annotations but those weren't working so I changed it to manually adding the listeners, and forgot to add the client-side check. My bad!

marchermans (approved):

marchermans:
@ChampionAsh5357 This is good to go form my end. Anything to add?

patches/minecraft/net/minecraft/client/renderer/ShaderInstance.java.patch
```diff
@@ -30,3 +30,29 @@
...
+                public String m_142138_(boolean p_173374_, String p_173375_) {
+-                  p_173375_ = FileUtil.m_179924_((p_173374_ ? s1 : "shaders/include/") + p_173375_);
+-                  if (!this.f_173369_.add(p_173375_)) {
++                  // FORGE: use the mod's namespace to look up resources if specified
++                  ResourceLocation resourcelocation = net.minecraftforge.client.ForgeHooksClient.getShaderImportLocation(s1, p_173374_, p_173375_);
++                  if (!this.f_173369_.add(resourcelocation.toString())) {
+                      return null;
+                   } else {
+-                     ResourceLocation resourcelocation = new ResourceLocation(p_173375_);
+-
```
ChampionAsh5357: Unnecessary removal

src/test/resources/assets/shader_resources_test/shaders/include/cubemap_includes.glsl
```diff
@@ -0,0 +1,5 @@
+#version 150
+
+in vec3 vertexPos;
+
+out vec4 fragColor;
```
ChampionAsh5357: New line at end of file.

ChampionAsh5357 (changes requested):
Two quick fixes on formatting, but otherwise looks and works fine.

patches/minecraft/net/minecraft/client/renderer/ShaderInstance.java.patch
FiniteReality: I've added back the empty line - more Eclipse default workspace setting woes that I missed. The removal of the ResourceLocation is still necessary as it's computed earlier for duplicate import checks.

ChampionAsh5357 (approved):

PR #9024, opened 2022-09-08
Spawn placement register event

eerussianguy:
This event allows replacing or modifying the spawn predicate of entities at the type level. Modders are expected to register their own for their own mobs, but there's currently no system for modifying vanilla's predicates, or that of other mods in say, a compatibility addon.

My use case: I allow users to define some spawn parameters through json reloadable data. Currently I am only able to do this for my own mobs, as that's what SpawnPlacements are registered for. I cannot restrict just based on biomes, as I have extra chunk data parameters (climate) that determine spawning.

Q: what's the change in CommandSourceStack?
A: according to discord, this is due to an AT by sciwhiz that was not accounted for properly

Q: why not use the check spawn event?
A: Looking at NaturalSpawner, this provides a lot more organized control over broadly what should spawn and how, especially with regards to things like heightmaps, because it's checked earlier and uses the heightmaps and spawn placement types. The current event is also unable to prevent the vanilla spawn placement checks from firing as those are checked before the check spawn event is fired.


patches/minecraft/net/minecraft/commands/CommandSourceStack.java.patch
```diff
@@ -8,4 +8,4 @@
 +public class CommandSourceStack implements SharedSuggestionProvider, net.minecraftforge.common.extensions.IForgeCommandSourceStack {
     public static final SimpleCommandExceptionType f_81286_ = new SimpleCommandExceptionType(Component.m_237115_("permissions.requires.player"));
     public static final SimpleCommandExceptionType f_81287_ = new SimpleCommandExceptionType(Component.m_237115_("permissions.requires.entity"));
-    private final CommandSource f_81288_;
+    public final CommandSource f_81288_;
```
ChampionAsh5357, line 11: You got sci's funky change in here.

ChampionAsh5357 (changes requested):
So, imo, this is not how I would do the event. I would rather have the event fire on the mod event bus probably around the same time as the entity attribute events and have fire once, and then for each entity you can register a predicate in a method or add a predicate in a separate method that is either ANDed, ORed, or REPLACEs what's currently there. If you want to replace, you would do it during the lowest priority, all others can go on either priorities.

For how ANDs and ORs would work, ORs would be based upon the initial predicate while ANDs are attached around (e.g. with condition A initial, B to AND, C to OR, and D to OR, (A || C || D) & B).

I will restate this is my opinion on how the event should be drawn up since I don't think this event makes much sense as it is now.

patches/minecraft/net/minecraft/commands/CommandSourceStack.java.patch
sciwhiz12: Keep this fix in, please, have mercy on my soul.

eerussianguy: I am a generous contributor.

eerussianguy:
Right now I think it’s best to think of this as basically an inject with some helper methods. It just intercepts each spawn placement and allows custom behavior. In terms of that not making sense I will need to hear more elaboration. I think this is a pretty clean way to do it, for almost every use case. Collecting a list of predicates beforehand is just more to maintain and more code because you then have to reinterpret everything just to recreate this behavior that we have now.

Further, spawn placements are registered during world load, meaning collecting this stuff too far beforehand is not acceptable as it means it’s less exposed to datapacks.

ChampionAsh5357:
> In terms of that not making sense I will need to hear more elaboration.

Injection makes sense on a simple level if all the predicates could do would have the same binary operation. However, the issue becomes how do you resolve someone who wishes to OR their predicate compared to someone who wishes to AND compared to someone who wants to REPLACE? There is no consistent method.

The second issue comes from when the event is fired. First, the spawn placements are registered during common setup, so the event should be on the mod bus. However, this is the case where an event is called within an event, specifically in the common setup event, which if a modder does wrong means that the event will be executed in parallel with itself at times causing weird collisions.

This is why I suggest making it like the entity attribute events and force all registration through that singular event since it not only consolidates bookkeeping but also fixes any dumb issues of modders doing who knows what where.

eerussianguy:
I was being dumb about reg time. That said I mean I can do it. It’s a pain in my ass but I could. I still think OR and AND and REPLACE are very possible, you have access to the predicate as modified (I could add access to the original predicate too) and since it’s a predicate you can just && or || or replace wholesale

ChampionAsh5357:
> you have access to the predicate as modified (I could add access to the original predicate too) and since it’s a predicate you can just && or || or replace wholesale

Yes but let's say I have condition A, B, and C from three mods where B is ORed and C is ANDed. There are six possible combinations:
```
A | B & C = C & B | A
A & C | B = B | A & C
B | C & A = C & A | B
```

So, basically, there are three different behaviors that could occur depending on which mod is loaded first. This is not consistent API.

eerussianguy:
Is the position that mods are not responsible for their load order against each other anymore? If I’m adding on to X mod I have to guarantee that I am after then or else I expect this to not work

ChampionAsh5357:
You're forgetting that anyone can add onto vanilla mobs using this event.

patches/minecraft/net/minecraft/world/entity/SpawnPlacements.java.patch
```diff
@@ -26,5 +34,20 @@
...
++       Map<EntityType<?>, net.minecraftforge.event.entity.SpawnPlacementRegisterEvent.MergedSpawnPredicate> map = Maps.newHashMap();
++       f_21750_.forEach((type, data) -> {
++           map.put(type, new net.minecraftforge.event.entity.SpawnPlacementRegisterEvent.MergedSpawnPredicate(data.f_21769_, data.f_21768_, data.f_21767_));
++       });
++       net.minecraftforge.common.ForgeHooks.onSpawnPlacementRegister(map);
++   }
++
++   public static void registerAll(Map<EntityType<?>, net.minecraftforge.event.entity.SpawnPlacementRegisterEvent.MergedSpawnPredicate> map)
++   {
++       map.forEach(((entityType, merged) -> f_21750_.put(entityType, new SpawnPlacements.Data(merged.getHeightmapType(), merged.getSpawnType(), merged.build()))));
```
ChampionAsh5357: Does not need to be such a large patch. Can be simplified with a call to the hooks class:

```java
public static void modifyPlacements() {
    Map<EntityType, MergedSpawnPred> map = ForgeHooks.buildPlacements(DATA_BY_TYPE);
    map.forEach(DATA_BY_TYPE.put());
}
```

src/main/java/net/minecraftforge/event/entity/SpawnPlacementRegisterEvent.java
```diff
@@ -0,0 +1,204 @@
...
+                replacementPredicate = predicate;
+            }
+            if (spawnType != null)
+            {
+                this.spawnType = spawnType;
+            }
+            if (heightmapType != null)
+            {
+                this.heightmapType = heightmapType;
+            }
```
ChampionAsh5357: Should these be modifiable in only replace? I don't know when else you would change this.

src/main/java/net/minecraftforge/event/entity/SpawnPlacementRegisterEvent.java
```diff
@@ -0,0 +1,204 @@
...
+        {
+            map.get(entityType).merge(operation, predicate, placementType, heightmap);
+        }
+    }
+
+    @ApiStatus.Internal
+    public void registerAll()
+    {
+        SpawnPlacements.registerAll(map);
+    }
```
ChampionAsh5357: You don't need this method since the map itself is mutable, so you already have an instance.

src/test/java/net/minecraftforge/debug/entity/SpawnPlacementTest.java
```diff
@@ -0,0 +1,44 @@
+package net.minecraftforge.debug.entity;
```
ChampionAsh5357, line 6: Missing license header.

src/main/java/net/minecraftforge/event/entity/SpawnPlacementRegisterEvent.java
```diff
@@ -0,0 +1,204 @@
...
+
+    public enum Operation
+    {
+        AND,
+        OR,
+        REPLACE
+    }
```
ChampionAsh5357: Probably should document operations here as well.

ChampionAsh5357 (changes requested):

src/main/java/net/minecraftforge/event/entity/SpawnPlacementRegisterEvent.java
eerussianguy: i was going back and forth on this, and I think I agree with your reading of it. I'll change it back

patches/minecraft/net/minecraft/world/entity/SpawnPlacements.java.patch
eerussianguy: yes and no, MergedSP and SP.Data are not convertible between each other just by reference, and SP.Data is private. So I cannot read anything about the SP.Data in ForgeHooks.

Definitely can be made smaller though

patches/minecraft/net/minecraft/world/entity/SpawnPlacements.java.patch
```diff
@@ -26,5 +34,18 @@
...
++   public static void fireSpawnPlacementEvent()
++   {
++       Map<EntityType<?>, net.minecraftforge.event.entity.SpawnPlacementRegisterEvent.MergedSpawnPredicate> map = Maps.newHashMap();
++       f_21750_.forEach((type, data) -> map.put(type, new net.minecraftforge.event.entity.SpawnPlacementRegisterEvent.MergedSpawnPredicate(data.f_21769_, data.f_21768_, data.f_21767_)));
++       net.minecraftforge.common.ForgeHooks.onSpawnPlacementRegister(map);
++   }
++
++   public static void registerAll(Map<EntityType<?>, net.minecraftforge.event.entity.SpawnPlacementRegisterEvent.MergedSpawnPredicate> map)
++   {
++       map.forEach(((entityType, merged) -> f_21750_.put(entityType, new SpawnPlacements.Data(merged.getHeightmapType(), merged.getSpawnType(), merged.build()))));
```
ChampionAsh5357: I still don't think you need this second method. You could just add the map loop as the last statement in the `fireSpawnPlacementEvent` since the map instance is the same throughout the calls.

src/main/java/net/minecraftforge/common/ForgeHooks.java
```diff
@@ -1166,6 +1168,13 @@ public static int canEntitySpawn(Mob entity, LevelAccessor world, double x, doub
         return res == Result.DEFAULT ? 0 : res == Result.DENY ? -1 : 1;
     }
 
+    public static void onSpawnPlacementRegister(Map<EntityType<?>, SpawnPlacementRegisterEvent.MergedSpawnPredicate> map)
+    {
+        var event = new SpawnPlacementRegisterEvent(map);
+        ModLoader.get().postEvent(event);
+        SpawnPlacements.registerAll(map);
```
ChampionAsh5357: This can be removed with my above comment.

src/main/java/net/minecraftforge/event/entity/SpawnPlacementRegisterEvent.java
```diff
@@ -0,0 +1,207 @@
...
+        {
+            return spawnType;
+        }
+
+        public Heightmap.Types getHeightmapType()
+        {
+            return heightmapType;
+        }
+
+        private void merge(Operation operation, SpawnPlacements.SpawnPredicate<?> predicate, @Nullable SpawnPlacements.Type spawnType, @Nullable Heightmap.Types heightmapType)
```
ChampionAsh5357: Should this method throw an error if the spawn and heightmap are nonnull when performing an AND or OR operation?

src/main/java/net/minecraftforge/event/entity/SpawnPlacementRegisterEvent.java
```diff
@@ -0,0 +1,207 @@
...
+    {
+        register(entityType, null, null, predicate, operation);
+    }
+
+    /**
+     * Register a {@code predicate} for a given {@code entityType} and {@code operation}
+     * With the option of changing the {@code placementType} and {@code heightmap}. These are only applied if {@link Operation#REPLACE} is used.
+     * Use {@code null} for the placement or heightmap to leave them as is (which should be done in almost every case)
+     */
+    public void register(EntityType<?> entityType, @Nullable SpawnPlacements.Type placementType, @Nullable Heightmap.Types heightmap, SpawnPlacements.SpawnPredicate<Entity> predicate, Operation operation)
```
ChampionAsh5357: You can genericize these methods so that the spawn predicate can use the correct instance of the entity without casting. Additionally, should this be only limited to mobs, or all entities (only used for mobs currently).

ChampionAsh5357 (changes requested):
Few more things I noticed when looking over it again.

src/main/java/net/minecraftforge/event/entity/SpawnPlacementRegisterEvent.java
eerussianguy: gonna leave the type bound as entity as that's the type bound for SpawnPlacements

ChampionAsh5357 (approved):
Looks fine, keeping in the licenses and change sci requested.

PR #9026, opened 2022-09-09
[1.18] Revert resource caching 

### LINKED ISSUE #8962: [1.18.2] - `cachePackAccess = true` increase significant (5x) loading time of mod packs.
**Minecraft Version:** 1.18.2

**Forge Version:** 1.18.2 - 40.1.73

**Steps to Reproduce:**
 1. Load a huge mod pack with `cachePackAccess = true` and stop the loading time.
 2. Change `cachePackAccess = false` and stop the loading time 
 3. Compare the two loading times.

**Description of issue:**
 `cachePackAccess = true` ~ 10:58.755 min
 `cachePackAccess = false` ~ 02:21.096 min

I would expect that the new option (which is enabled by default) is not causing such a increase (5x) in the loading times.
I assume that is maybe caused by long timeouts during scanning, loading or parsing the relevant resources.

The issue is not happening with former forge versions, so it's seems to be related to the following change:
https://github.com/MinecraftForge/MinecraftForge/commit/4df024ef43524d8ec48928c70f561903f7a8e184

I'm using the Curse Forge Launcher for loading the mod pack like:
https://www.curseforge.com/minecraft/modpacks/bos-adventure-world/files/3932653

Shadows-of-Fire:
This reverts commit 4df024ef43524d8ec48928c70f561903f7a8e184
Fixes #8962


eerussianguy:
This is a breaking change for mods that make their own PathResourcePack. A situation I could see is subclassing and having to instantiate a new resource cache manager since the original one is private. Not sure if we are considering that public API but it’s worth noting.

Shadows-of-Fire:
If this reversion is breaking, then wouldn't the introduction of it have been breaking in the first place?

In general, private fields are not public api.

eerussianguy:
Well no, because if you add something and then people use it and then you remove it then you broke something when you did not before

marchermans:
A revert is not the way to go.
Please just set the default option for the config to disabled.

Shadows-of-Fire:
Changing the config option does not address the logging issue, which occurs regardless.  Unless it is fixed, this must be reverted, and I would rather it be reverted *until* it is fixed.

marchermans:
You can easily prevent the logging if you added a fix like this:
```java
if (!Files.exists(path)) {
   return Stream.of();
}
```
here: https://github.com/MinecraftForge/MinecraftForge/pull/9026/files#diff-89a22bac53549dc6bdb669a650a792b5802b001d2d7df9a7f9a4b6013cdd1808L139

With the config option disabled.

tmvkrpxl0:
and I feel it's bit mean to just revert it...

TheCurle:
> You can easily prevent the logging if you added a fix like this:
> 
> if (!Files.exists(path)) {
   return Stream.of();
}

Please don't do this, that's a major performance issue for most of us.

marchermans:
> > You can easily prevent the logging if you added a fix like this:
> > if (!Files.exists(path)) {
> > return Stream.of();
> > }
> 
> Please don't do this, that's a major performance issue for most of us.

That does not matter, the exception that causes the performance hit is the one that is being caught and causing the log line to appear.

We could discuss just Nooping the log line and as such the performance hit, or making it a config option.
Regardless simply reverting is not an option, I have asked for repro cases now for several days already, I have the repro case for 1.18.2 and but not for 1.19. If anybody could provide me that I can easily provide a fix.

TheCurle:
Open any modpack.

That's the reproduction case.

Thousands of reports, I've reproduced it about three dozen times picking random modpacks from CurseForge and looking at the logs.

It's really not a special case, this is happening to everyone.

marchermans:
> Thousands of reports, I've reproduced it about three dozen times picking random modpacks from CurseForge and looking at the logs.

Okey, but funnily enough it does not happen in dev environment.....
I will check later today and push a fix.

NertzhulDEV:
> > Thousands of reports, I've reproduced it about three dozen times picking random modpacks from CurseForge and looking at the logs.
> 
> Okey, but funnily enough it does not happen in dev environment..... I will check later today and push a fix.

I can confirm that this does in fact happen in dev enviroment.
Currently using forge 40.1.73

https://gist.github.com/NertzhulDEV/ed76d3ab64167243d4edfb9293d9c91d
image

marchermans:
> > > Thousands of reports, I've reproduced it about three dozen times picking random modpacks from CurseForge and looking at the logs.
> > 
> > Okey, but funnily enough it does not happen in dev environment..... I will check later today and push a fix.
> 
> I can confirm that this does in fact happen in dev enviroment.
> Currently using forge 40.1.73
> 
> https://gist.github.com/NertzhulDEV/ed76d3ab64167243d4edfb9293d9c91d
> image

Yeah on 1.18.2 it is known. But not on 1.19 where the logs still pile up

Xetaxheb:
Not exactly endorsing a total revert/removal if there is an interim kludge that can be applied to disable the pressing issue, like this https://github.com/MinecraftForge/MinecraftForge/pull/9026#issuecomment-1241557505 (and assuming it also fixes any performance issues when the feature is disabled)

**_However,_** this feature should not have been put in, in the first place, without proper testing. Testing in a dev environment is not testing. A dev environment is to work out the bugs and it should have been used in various test cases outside of the dev environment, because this should have been caught instantly. If you can't throw a change at something like whatever relevant flavor of all-the-mods pack and not cause issues (say, launch a version without your changes and launch a version with your changes and diff the logs) then it's not ready for commit.

Arguing that something should stay in essentially from the viewpoint of "because it's my child" is not a good argument. There needs to be reasons something should stay and really the potential for world loading is good enough... however the increase in game loading _**is not, in my opinion, good enough to warrant it on by default, yet.**_

_Either way the performance impact from merely existing as an option, despite being disabled, needs to be resolved asap._ If you want it not reverted then come up with a solution since it's your guys (orion and russian) implementation that's the problem?
Are there even any examples of it being used by third-parties yet on 1.18.2, since it's only been out a month? It shouldn't be in 1.18.2 if it's having issues in 1.18.2 until they're resolved. Maybe disable the offending code and throw an exception when it's attempted to be used, letting users and devs know it's been removed in the interim.

LexManos:
Executive decision, pulling this because I'm annoyed that Orion is letting this stagnate. It should be a simple fix to just disable it and fix whatever logging issue but it seems to not be the case.

As for any mods that will break, that is a shame. But the major outweighs the few.

TheCurle:
As a corollary, this decision was made by the both of us.

This issue is causing large performance deficits just by existing, by virtue of the nature of exceptions as control flow, combined with intense logging, combined with disk bandwidth.

Every modpack running 1.18.2 with these versions of Forge was encountering log spam, varying from a few lines serving as an annoyance, to more than 55 gigabytes worth of the same message being printed over and over.

Simply writing all of that to disk was causing substantial load time performance issues, combined with the nature of the caching slowing loading even further, and we have a catastrophe.

For reference, that happened in my own userdev environment. Loading the game took approximately 20 minutes that time around. On a version before this caching was added, it took approximately 40 seconds. 

Let that sink in. 

I've asked Orion to make a hotfix multiple times, the response being "I have a fix in my workspace but I cant test it", which is honestly bogus. I haven't found an environment that hasn't had this problem to some extent.

So, we came to the conclusion that a full revert is the only action left.

We'll work on helping those mods that used the new features, and we apologise to the developers.

However, we fully believe the benefits outweigh the negatives.

PR #9027, opened 2022-09-09
[1.19.2] Copy bookshelves block tag into item tag

Jonathing:
Addendum to #8991.

This is a bit of a small PR and is more-so for consistency than anything else, but I thought it would be worth noting regardless. Basically, in `ForgeItemTagsProvider`, block tags are usually copied into item tags, but that was not the case with the bookshelves item tag. This means that additional blocks added to the bookshelves tag would need to have their block items manually added as well.


ChampionAsh5357 (approved):

LexManos:
This is a non-functional difference... But whatever i'll combine it with all the trivial other tags.

kaiden-lcfc:
Im new to this kinda stuff so can you please help me by telling me how to get to the file that you have changed?

ChampionAsh5357:
> Im new to this kinda stuff so can you please help me by telling me how to get to the file that you have changed?

This question should be asked on the [Forge discord](https://discord.gg/forge) or forums as it does not have anything to do with the PR or implementation of the feature.

PR #9028, opened 2022-09-10
[1.19.x] (RecipeType, Item)->RecipeBookCategories table to manually sort recipes into categories

bconlon1:
In vanilla, `ClientRecipeBook` determines what recipe book category (the tabs on the side of the recipe book) that a recipe should be sorted into through fairly hardcoded means, especially the crafting table. Recipes are sorted into categories in the crafting table's recipe book based on the result item's creative tab, e.g. if a block is in the vanilla redstone creative tab, it gets sorted into the redstone category in the recipe book. This becomes an issue for mods that use their own creative tabs to store their items, as since they don't belong to a vanilla creative tab, their recipes automatically get sorted into the miscellaneous category.

This PR solves this by adding in a table that correlates RecipeBookCategories to modded items and recipe types, so when `ClientRecipeBook` looks for where to sort a recipe based on its result item and recipe type, it'll check if those first exist in the map and then return the category set by the modder. This is mainly meant to fix the issue in the case of the crafting table menu, but it is capable of working for other recipe categories and types if a modder desires.

The implementation of this is just handled by calling `RecipeBookManager.addItemToCategory(RecipeType<?>, RecipeBookCategories, Item...)` any stage after registry (`FMLCommonSetup` works just fine).

### DIFF
--- a/patches/minecraft/net/minecraft/client/ClientRecipeBook.java.patch
+++ b/patches/minecraft/net/minecraft/client/ClientRecipeBook.java.patch
@@ -9,6 +9,17 @@
              if (s.isEmpty()) {
                 map.computeIfAbsent(recipebookcategories, (p_90645_) -> {
                    return Lists.newArrayList();
+@@ -76,6 +_,10 @@
+ 
+    private static RecipeBookCategories m_90646_(Recipe<?> p_90647_) {
+       RecipeType<?> recipetype = p_90647_.m_6671_();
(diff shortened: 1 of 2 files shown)


siasur:
This would be so awesome! Just a few days ago I had asked on the Discord about this, and noticed the relation to the creative tab. I worked around this with setting the tab to the tools tab, and overwriting `Item#allowedIn(CreativeModeTab creativeTab)` so It would still show up in the creative Menu Tab for my mod. Felt kinda janky.

bconlon1:
Closing as #9054 is the new iteration of this.

